Issue #1200, opened 2019-01-25
Using go1.11 breaks memcache DNS lookups


gouthamve:
The net library in `go1.11+` cannot read SRV responses from kube-dns `v1.14.10` or below. This is because kube-dns prior to `1.14.11` used to send compressed SRV responses which is against the spec. See: https://github.com/golang/go/issues/27546 for more details.

We use it here: https://github.com/cortexproject/cortex/blob/master/pkg/chunk/cache/memcached_client.go#L103

This means that we are right now stuck to go1.10. Upgrading cortex to `go1.11+` would break memcache lookups among others. The fix for this would be to use https://github.com/miekg/dns like Prometheus does.

/cc @codesome

codesome:
Found another place where it is used https://github.com/cortexproject/cortex/blob/a888416ea1cc069e09c1e690f378769cd7dcc0f9/pkg/alertmanager/peers.go#L65

codesome:
Looks like I need to update in 1 more place. Thanks @bboreham!

codesome:
I don't know about this part of Golang, but would this make this file to be compiled only with go1.8?
https://github.com/cortexproject/cortex/blob/a888416ea1cc069e09c1e690f378769cd7dcc0f9/vendor/google.golang.org/grpc/naming/go18.go#L1

bboreham:
It means 1.8 and above. See https://golang.org/pkg/go/build/

codesome:
Yes you are right. I should have searched before asking :)

gouthamve:
This is only a problem when running kubernetes `v1.11` or below and compiling Cortex with go v1.11 and above.

gouthamve:
Having said that bumping up `kube-dns` is straight forward and will fix this. We're on k8s 1.11 but with kube-dns at `1.14.13`

bboreham:
So there is a more specific problem description involving kube-dns ?

bwplotka:
That's funny, we found similar issue on Thanos as well: https://github.com/improbable-eng/thanos/issues/1015 obviously same thing with go 1.12

We consider moving to https://github.com/miekg/dns as it allows much much more flexibility. Have you considered this?

bboreham:
@bwplotka see #1206 - in the end we decided that upgrading kube-dns would be a better route than changing the software.

bwplotka:
Awesome, thanks for this link, super useful.

bwplotka:
[...]

PR #1206, opened 2019-01-26
Use miekg/dns for SRV lookup

codesome:
Fixes https://github.com/cortexproject/cortex/issues/1200

EDIT (by @gouthamve): This uses miekg/dns library to do the SRV lookups. We are using this in the grpc codebase as there is a A record fallback when SRV lookups fail.

### DIFF
--- a/pkg/util/dns.go
+++ b/pkg/util/dns.go
@@ -3,6 +3,7 @@ package util
 import (
 	"fmt"
 
+	"github.com/go-kit/kit/log/level"
 	"github.com/miekg/dns"
 )
 
@@ -31,15 +32,19 @@ func LookupSRV(service, proto, name string) ([]*dns.SRV, error) {
 	name = "_" + service + "._" + proto + "." + name
 	msg.SetQuestion(dns.Fqdn(name), dns.TypeSRV)
 
+	client := dns.Client{
+		Net: "tcp",
+	}
 	var result []*dns.SRV
 	dnsResolved := false
 	for _, serverAddr := range dnsServers {
-		resMsg, err := dns.Exchange(msg, serverAddr)
+		resMsg, _, err := client.Exchange(msg, serverAddr)
 		if err != nil {
+			level.Warn(Logger).Log("msg", "DNS exchange failed", "err", err)
 			continue
 		}
 		dnsResolved = true
-		for _, ans := range resMsg.Ns {
+		for _, ans := range resMsg.Answer {
 			if srvRecord, ok := ans.(*dns.SRV); ok {
 				result = append(result, srvRecord)
 			}
--- a/pkg/util/dns.go
+++ b/pkg/util/dns.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"fmt"
+	"net"
 
 	"github.com/go-kit/kit/log/level"
 	"github.com/miekg/dns"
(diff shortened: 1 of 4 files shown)


tomwilkie:
Looks good so far; lint needs fixing, and there was some mention of gRPC also using LookupSRV for querier -> frontend comms.

codesome:
Yes, it was [this](https://github.com/cortexproject/cortex/issues/1200#issuecomment-457941931) comment by @bboreham. Need to fix that too. Will do when I get free from academic stuff.

tomwilkie:
I opened https://github.com/grpc/grpc-go/pull/2634 to help with the gRPC side.

codesome:
Thanks @tomwilkie!

pkg/util/dns.go
```diff
@@ -0,0 +1,57 @@
...
+	msg := &dns.Msg{}
+	name = "_" + service + "._" + proto + "." + name
+	msg.SetQuestion(dns.Fqdn(name), dns.TypeSRV)
+
+	var result []*dns.SRV
+	dnsResolved := false
+	for _, serverAddr := range dnsServers {
+		resMsg, err := dns.Exchange(msg, serverAddr)
+		if err != nil {
+			continue
```
gouthamve: Please log the error here. Deploying this, gives this error: `dial udp: address 10.47.240.10: missing port in address`

bboreham:
Note the gRPC PR was rejected

gouthamve:
Hmm, as far as I see, the grpc resolver uses `grpclb` port name to resolve the services, https://github.com/grpc/grpc-go/blob/master/naming/dns_resolver.go#L216-L220

If that fails, it falls back to simple A record lookup: https://github.com/grpc/grpc-go/blob/master/naming/dns_resolver.go#L261-L267

I found that in the clusters we run, we don't expose a port-named `grpclb` and the resolver falls back to A records. Given that it works nevertheless, I wonder if it's worth the pain to copy the `dns_resolver.go` file from `grpc-go`. Note that this doesn't exist in the latest kubernetes versions and only is a problem when running cortex with `go 1.11` with an old kubernetes version. Even then nothing will break.

bboreham:
I'm happy with A record lookup.
Can someone update #1200 with exactly what works and doesn't work?
(e.g. what's "an old kubernetes version")

gouthamve:
A: We don't need to do it if we're okay asking people to upgrade kube-dns
B: The TCP-UDP change has been reverted in the last commit by me. The current state of the PR could be described as:

Use miekg/dns instead of native DNS library to be compatible with kube-dns `<1.14.11` and golang `>=1.11`. This is what Prometheus does for DNS-SD

C: The problem is that `go1.11`'s net library cannot read SRV records sent by kube-dns `<1.14.11`

bboreham:
Still feels like a bad idea to me. 
Note I have two questions outstanding on the go modules PR #1232 for 25 days.

tomwilkie:
> Still feels like a bad idea to me.

What feels bad about it?  Can I infer from your comments you'd prefer to upgrade go without this, effectively making Cortex not work on kube-dns <1.14.11?

> Note I have two questions outstanding on the go modules PR #1232 for 25 days.

Thats a different PR; lets address that there, it shouldn't block this.

bboreham:
> What feels bad about it?

It's adding 70 lines of code in a client to work around an issue in a server which was fixed months ago.

Are there reports from people who use or would like to use Cortex and cannot upgrade kube-dns?

Also I think there should be comments in the code
 - why we are doing this
 - why do it in some places and not others
 - why you shouldn't call this code very often
 - why it's ok to have a relatively inefficient implementation

gouthamve:
Hmm, I'm going to hold off on this PR until someone actually raises an issue and proceed with Go 1.12 upgrade. We might just be optimising for something that has been fixed already as you point out.

tomwilkie:
@khaines on the call yesterday we decided we probably didn't need this, but we just want to check you've got an update kubedns?

tomwilkie:
No word from @khaines. I'll close, we can reopen if needs be.

Issue #1151, opened 2019-01-02
Implement Querier.LabelNames


tomwilkie:
See #1139

codesome:
I would like to work on it. Started reading the codebase for that already.

PR #1207, opened 2019-01-27
Implement Querier.LabelNames

codesome:
Closes https://github.com/cortexproject/cortex/issues/1151


pkg/querier/querier.go
```diff
@@ -158,7 +158,20 @@ func (q querier) LabelValues(name string) ([]string, error) {
 }
 
 func (q querier) LabelNames() ([]string, error) {
-	return nil, nil
+	return q.distributor.LabelNames(q.ctx)
+
+	// TODO(codesome): Would the above make sense or the below?
```
codesome: I have 2 solutions for this function. Would `q.distributor.LabelNames(q.ctx)` be fine or the commented solution is the way to do?

tomwilkie: For now there is no way to implement it for the chunkstore, so what you have above should be fine.

pkg/util/merger.go
```diff
@@ -43,3 +43,40 @@ func MergeNSampleSets(sampleSets ...[]model.SamplePair) []model.SamplePair {
 	right := MergeNSampleSets(sampleSets[n:]...)
 	return MergeSampleSets(left, right)
 }
+
+func MergeSortedStringLists(lnss [][]string) []string {
```
tomwilkie: I was hoping we already had an implementation of this, but I can't find it.

We might want to look at `LabelValuesForLabelName` and use the same method there?

tomwilkie:
Looks good @codesome!  Need to fix the lint.

pkg/util/merger.go
codesome: We have something similar in here, but for `model.LabelNames`. Direct typecast was not possible, so decided to have another one for `[]string`.
https://github.com/cortexproject/cortex/blob/f6627227636f447b29cd46e0ec3f432ed0a1a0d3/pkg/ingester/index/index.go#L299

`LabelValuesForLabelName` uses map to filter out, and should be also apt for label names as they are less in number. I think I will remove this function and use maps everywhere instead.

codesome:
@tomwilkie done!

codesome:
@tomwilkie ping

tomwilkie:
LGTM from me.

PR #1210, opened 2019-01-29
remove the redundant error check

yeya24:
remove the redundant error check


bboreham:
Thanks for the PR.

Is there some background to this change?  This is a bit of a pattern in Go:

```
...
if err != nil {
   return err
}
...
if err != nil {
   return err
}
return nil
```

yeya24:
> Thanks for the PR.
> 
> Is there some background to this change? This is a bit of a pattern in Go:
> 
> ```
> ...
> if err != nil {
>    return err
> }
> ...
> if err != nil {
>    return err
> }
> return nil
> ```

Thanks! I just thought it is kind of redundant here. I will close this PR

PR #1214, opened 2019-01-31
Make enforcement of the presence of metic name on pushes optional.

tomwilkie:

### DIFF
--- a/pkg/util/validation/validate.go
+++ b/pkg/util/validation/validate.go
@@ -78,7 +78,7 @@ func (cfg *Overrides) ValidateLabels(userID string, ls []client.LabelPair) error
 	numLabelNames := len(ls)
 	if numLabelNames > cfg.MaxLabelNamesPerSeries(userID) {
 		DiscardedSamples.WithLabelValues(maxLabelNamesPerSeries, userID).Inc()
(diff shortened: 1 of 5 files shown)


bboreham:
Would be nice to have the background to this change.

pkg/util/validation/validate.go
```diff
@@ -65,12 +65,14 @@ func (cfg *Overrides) ValidateSample(userID string, metricName []byte, s client.
 // ValidateLabels returns an err if the labels are invalid.
 func (cfg *Overrides) ValidateLabels(userID string, ls []client.LabelPair) error {
 	metricName, err := extract.MetricNameFromLabelPairs(ls)
-	if err != nil {
-		return httpgrpc.Errorf(http.StatusBadRequest, errMissingMetricName)
-	}
+	if cfg.EnforceMetricName(userID) {
```
bboreham: In the case where we don't enforce, and it is blank, the error returns below (lines 81 and 105) become less usable.

tomwilkie:
> Would be nice to have the background to this change.

We're using the same label validation code for Loki; but logs don't have metric names.

pkg/util/validation/validate.go
tomwilkie: Yeah, bit of a shame.  Could return the whole label set there - WDYT?

Issue #322, opened 2017-03-09
Document configs service


jml:
It's used for storing & retrieving alertmanager configs and rules. We should document it & its API (in case folk want to use a different implementation).

gouthamve:
Closed in #1215

Issue #619, opened 2017-12-06
Split alertmanager & ruler configs


jml:
Currently there's one table, `configs`, keyed by (`id`, `orgID`), where `id` is a version number. The value is a JSON blob which has both the alertmanager config and the rules. Further, although there are separate endpoints, the implementations behind these endpoints are identical.

Instead, the config systems for alertmanager and ruler should be fully disjoint.

This means we should have:
- one table with rules
- one table with alertmanager configs
- endpoints for alertmanager (that have nothing to do with ruler) for
  - getting a config
  - setting a config
  - getting all configs
  - getting all configs since a version
- likewise endpoints for ruler configs
 
We should also separate the client libraries, embedding them in their respective call-sites (move ruler config client to ruler, alertmanager config client to alertmanager). 

This will need to be multiple PRs in order to handle backwards compatibility between UI <-> configs <-> DB.

At a guess:

1. New endpoints
    - create new endpoints with desired behaviour
    - update clients in cortex to use & rely on new behaviour
    - still write to existing conflated DB structure
    - old endpoints still work as intended
2. Update Weave Cloud UI to use new endpoints
3. Remove old endpoints
4. New tables
    - create new tables
    - write to new tables
    - read from new tables and fall back to old
5. Migrate old data
    - wait until data only being written to new tables
    - migrate old data to new tables
6. Remove old tables
    - remove old tables
    - remove fallback code

LMK if that needs refining (can we do it in fewer steps? would even these steps have backwards compat problems?).

I'd kind of like to kill off the configs service completely by moving the endpoints to alertmanager and ruler respectively, and having both of those services connect directly to the database. The most convenient place to make that decision is in PR 1 above, since we'll have to update UI endpoints. It will also require more PRs for juggling flags to our deployed versions.

Other stuff that we might want to consider:
- use gRPC rather than REST
  - not a massive advantage, as we would still have to re-parse configs at client side, because the configs will still be strings
  - since we'll want new endpoints, might as well have new endpoints with better protocol
  - probably don't want to use gRPC if we're moving config endpoints to ruler & alertmanager, because we'll need to talk to them from JS anyway
- could theoretically split DBs, or even move to different DB backend (e.g. DynamoDB)
  - not under this GH issue

bboreham:
What is the expected gain from this change?

jml:
Not a great deal (which is why we've not done it yet).

Gains are:
 1. conceptual simplicity
[...]

PR #1215, opened 2019-01-31
add configs api docs

agrahamlincoln:
Signed-off-by: Graham Rounds <graham@platform9.com>

---

Wrote this when exploring ruler/configs functionality and figured it would be valuable the official repo. Please correct anything incorrect and/or misleading I may have overlooked.

### DIFF
--- a/docs/configs-api.md
+++ b/docs/configs-api.md
@@ -98,9 +98,3 @@ will use the config with the highest number.
 These API endpoints will disable/enable the current Rule and Alertmanager configuration for a tenant.
 
 Note that setting a new config will effectively "re-enable" the Rules and Alertmanager configuration for a tenant.
-
-### Internal APIs
-
-`GET /private/api/prom/configs/rules` - Get current rule file
-
-`GET /private/api/prom/configs/alertmanager` - Get current Alertmanager config
--- a/docs/configs-api.md
+++ b/docs/configs-api.md
@@ -22,7 +22,7 @@ Example Payload:
     "id": 99,
     "rule_format_version": "2",
     "config": {
-        "rules_files": { <see below> }
+        "rules_files": { <see below> },
         "alertmanager_config":"<standard alertmanager.yaml config>"
     }
 }
@@ -45,12 +45,12 @@ Example:
 {
     "id": 99,
     "config": {
-        "alertmanager_config":"<see above>"
+        "alertmanager_config":"<see above>",
         "rules_files": {
             "rules.yaml": "<standard rules.yaml config>",
             "rules2.yaml": "<standard rules.yaml config>",
         },
-    }
+    },
     "rule_format_version": "2",
 }
 ```
@@ -74,8 +74,8 @@ Prometheus version you want to match.
 {
     "id": 99,
     "config": {
-        "alertmanager_config":"<see above>"
-        "rules_files": { <see above> }
+        "alertmanager_config":"<see above>",
+        "rules_files": { <see above> },
         "template_files": {
             "templates.tmpl": "<standard template file>",
             "templates2.tmpl": "<standard template file>"
--- a/docs/configs-api.md
+++ b/docs/configs-api.md
@@ -4,6 +4,9 @@ The configs service provides an API-driven multi-tenant approach to handling var
 
 Each tenant will have it's own set of rule files, Alertmanager config, and templates. A POST operation will effectively replace the existing copy with the configs provided in the request body.
 
+At the current time of writing, the API is part-way through a migration from a single Configs service that handled all three sets of data to a split API. [Tracking issue](https://github.com/cortexproject/cortex/issues/619). So, unfortunately, all APIs take and return all sets of data.
+
+
 ### Manage Alertmanager
 
 `GET /api/prom/configs/alertmanager` - Get current Alertmanager config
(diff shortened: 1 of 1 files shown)


tomwilkie:
Thanks @agrahamlincoln, this LGTM!  We need a second +1 to merge though.  @gouthamve @bboreham @csmarchbanks ?

docs/configs-api.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Manage Rules
+
+`GET /api/prom/configs/rules` - Get current rule file
+
+`POST /api/prom/configs/rules` - Replace current rule file
+
+Returns and Accepts a JSON object representing a [Prometheus rule file](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules). 
+
+Example:
```
csmarchbanks: It's been quite awhile since I used this api, but I though these were wrapped with some extra data (rule file version, then a map of all the rule files). I think the relevant struct is [`configs.RulesConfig`](https://github.com/cortexproject/cortex/blob/master/pkg/configs/configs.go#L139).

docs/configs-api.md
```diff
@@ -0,0 +1,95 @@
+# Configs API
+
+The configs api provides API-driven multi-tenant approach to prometheus' Rule files and Alertmanager configs.
+
+Each tenant has a single Rule file and Alertmanager Configuration. A POST operation will effectively replace the existing copy with the configs provided in the request body.
+
+Although Prometheus works with YAML files by default, these HTTP API's only accept JSON. You may convert your YAML to JSON using a tool such as [yaml2json](https://github.com/bronze1man/yaml2json).
```
csmarchbanks: I remember succeeding in posting a raw YAML file at one point... that said I think using JSON makes the requests much simpler to figure out.

jtlisi:
@csmarchbanks It does take a different input. Here is an example:
```json
{
  "new_config": {
    "format_version": "2",
    "files": {
      "example_rules.yaml": "groups:\n  - name: test\n    rules:\n    - record: test_ruler_sum_up\n      expr: sum(up)"
    }
  },
  "old_config": {
    "format_version": "2",
    "files": {
      "example_rules.yaml": ""
    }
  }
}```

File: docs/configs-api.md (hunk 2 above)
Comment (khaines) [reply docs/configs-api.md]: I'm certain we post yaml files in our configuration system to this api. @rsteneteg , can you provide some comments here?

Comment (rsteneteg) [reply docs/configs-api.md]: The part about only accepting JSON is correct, however just to be clear it's only the request to config service that needs to be in JSON format, the contents of your rule files should still be YAML if using the prometheus v2 rules format

File: docs/configs-api.md [hunk 3]
```diff
@@ -0,0 +1,95 @@
+# Configs API
+
+The configs api provides API-driven multi-tenant approach to prometheus' Rule files and Alertmanager configs.
+
+Each tenant has a single Rule file and Alertmanager Configuration. A POST operation will effectively replace the existing copy with the configs provided in the request body.
```
rsteneteg: You can have multiple rule files

docs/configs-api.md
```diff
@@ -0,0 +1,95 @@
+# Configs API
+
+The configs api provides API-driven multi-tenant approach to prometheus' Rule files and Alertmanager configs.
```
bboreham: minor point but I would add something like "backed by a database" - this doc mostly talks about files but the whole reason this code exists is because we wanted to use a database and the comparable code in Prometheus is hard-coded to only read files from a disk path.

docs/configs-api.md
```diff
@@ -0,0 +1,95 @@
...
+                "match": {
+                    "team": "frontend"
+                }
+            }
+        ]
+    }
+}
+```
+
+**Note:** Template files and Email Notifications are not supported in Cortex yet.
```
rsteneteg: this PR adds support for template files https://github.com/cortexproject/cortex/pull/1237

agrahamlincoln:
Thanks for all the feedback! I've added the templates API endpoints, and updated the examples to be accurate. Let me know if there is anything else incorrect or that could use more elaboration.

rsteneteg:
@agrahamlincoln, I read the updated document, and it sounds like you can send a partial update with only templates or only rules etc.. but that is not correct, each post to any of the endpoints will replace the config with whatever was in the request. So you will need to post everything AM config, rules, templates in each request for all endpoints.

rsteneteg:
It would however be nice if you could post a partial update with only templates or rules etc, we should create a task to fix that

bboreham:
I think partial updates are listed as part of #619

bboreham:
I took the liberty of pushing to your branch some updates which I think bring it more into line with what the code implements.

Also I don't think we should include internal APIs in the same doc.

bboreham:
This can probably close #322

agrahamlincoln:
> I took the liberty of pushing to your branch some updates which I think bring it more into line with what the code implements.

Thanks! I appreciate you bringing together links to paint some color around why the behavior is what it is today.

agrahamlincoln:
Given that POST requests use the same payload, I consolidated them all into a section and used the extra real-estate to document expected return codes.

I'm pretty comfortable with this doc at this point - but definitely open to any more feedback

bboreham:
Let's merge this now, then any further updates can be a new PR.

docs/configs-api.md
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+}
+```
+
+### Formatting
+
+`id` - should be incremented every time data is updated; Cortex will use the config with the highest number.
+
+`rule_format_version` - allows compatability for tenants with config in Prometheus V1 format. Pass "1" or "2" according to which Prometheus version you want to match.
```
bboreham: "compatability" is triggering a CI failure - better fix that or CI will be broken on master

PR #1216, opened 2019-02-01
More efficient memory layout of bigchunk chunks

bboreham:
Instead of three parallel slices, use a slice of structs.
Saves 2x24-byte slice headers per `bigchunk`, and some GC overhead.

also `UnmarshalFromBuf()` becomes more efficient because it was only pre-sizing one of the three slices.


tomwilkie:
LGTM!

PR #1218, opened 2019-02-01
Strip Chunk interface indirection on bigchunk sub-chunks

bboreham:
The `Chunk` indirection doubles the pointer size for every sub-chunk, which we don't need because there is only one kind of Chunk.

Note this will clash with #1216 but I wanted to separate the ideas.  Can rebase or merge the PRs if appropriate.


tomwilkie:
LGTM after rebase.

PR #1219, opened 2019-02-02
Add options for client-side rate limiting and backoff on all gRPC clients.

tomwilkie:
The Bigtable table client does O(n) `TableInfo` calls for every `ListTables`, to check the column family exists.  Run it for long enough and Google starts rate limiting you when you have 100s of tables.

### DIFF
--- a/pkg/chunk/gcp/gcs_object_client.go
+++ b/pkg/chunk/gcp/gcs_object_client.go
@@ -64,6 +64,11 @@ func (s *gcsObjectClient) PutChunks(ctx context.Context, chunks []chunk.Chunk) e
 			return err
 		}
 		writer := s.bucket.Object(chunk.ExternalKey()).NewWriter(ctx)
+		// Default GCSChunkSize is 8M and for each call, 8M is allocated xD
+		// By setting it to 0, we just upload the object in a single a request
+		// which should work for our chunk sizes.
+		writer.ChunkSize = 0
+
 		if _, err := writer.Write(buf); err != nil {
 			return err
 		}
--- a/pkg/chunk/gcp/gcs_object_client.go
+++ b/pkg/chunk/gcp/gcs_object_client.go
@@ -31,7 +31,7 @@ func (cfg *GCSConfig) RegisterFlags(f *flag.FlagSet) {
 
 // NewGCSObjectClient makes a new chunk.ObjectClient that writes chunks to GCS.
 func NewGCSObjectClient(ctx context.Context, cfg GCSConfig, schemaCfg chunk.SchemaConfig) (chunk.ObjectClient, error) {
-	client, err := storage.NewClient(ctx, instrumentation()...)
+	client, err := storage.NewClient(ctx)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/chunk/gcp/gcs_object_client.go
+++ b/pkg/chunk/gcp/gcs_object_client.go
@@ -21,12 +21,14 @@ type gcsObjectClient struct {
 
 // GCSConfig is config for the GCS Chunk Client.
 type GCSConfig struct {
-	BucketName string `yaml:"bucket_name"`
+	BucketName      string `yaml:"bucket_name"`
+	ChunkBufferSize int    `yaml:"chunk_buffer_size"`
 }
 
 // RegisterFlags registers flags.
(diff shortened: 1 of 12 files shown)


tomwilkie:
Added backoff & retry too, which I should have done to begin with.  Would like to leave the rate limit code in there, as it could be useful too, but open to taking it out.

pkg/util/grpcclient/grpcclient.go
```diff
@@ -32,6 +44,18 @@ func (cfg *Config) CallOptions() []grpc.CallOption {
...
 // DialOption returns the config as a grpc.DialOptions.
-func (cfg *Config) DialOption() grpc.DialOption {
-	return grpc.WithDefaultCallOptions(cfg.CallOptions()...)
+func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) []grpc.DialOption {
+	if cfg.BackoffOnRatelimits {
+		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{NewBackoffRetry(cfg.BackoffConfig)}, unaryClientInterceptors...)
+	}
+
+	if cfg.RateLimit > 0 {
+		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{NewRateLimiter(cfg.RateLimit, cfg.RateLimitBurst)}, unaryClientInterceptors...)
```
bboreham: I might set `RateLimitBurst` here to `RateLimit` if it is still zero, otherwise you get a really weird failure mode.

pkg/chunk/gcp/instrumentation.go
```diff
@@ -1,40 +1,87 @@
...
+type instrumentedTransport struct {
+	observer prometheus.ObserverVec
+	next     http.RoundTripper
+}
+
+func (i instrumentedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	start := time.Now()
+	resp, err := i.next.RoundTrip(req)
+	if err == nil {
+		i.observer.WithLabelValues(req.URL.Path, strconv.Itoa(resp.StatusCode)).Observe(time.Since(start).Seconds())
```
gouthamve: This will blow up with GCS client. The path is `/<bucket>/<object-path>`

jtlisi: Maybe it would be better to instrument the object and bigtable client separately.

pkg/chunk/gcp/instrumentation.go
```diff
@@ -1,40 +1,87 @@
...
+type instrumentedTransport struct {
+	observer prometheus.ObserverVec
+	next     http.RoundTripper
+}
+
+func (i instrumentedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	start := time.Now()
+	resp, err := i.next.RoundTrip(req)
+	if err == nil {
+		i.observer.WithLabelValues(req.Method, strconv.Itoa(resp.StatusCode)).Observe(time.Since(start).Seconds())
```
gouthamve: The label `operation` corresponds to `GET/POST` now. I wonder if we should change the label name to `method`?

tomwilkie: Nah, I think this is fine.

tomwilkie:
Thanks for feedback @bboreham, and thanks for fixing the method @gouthamve.  Any chance we can get this merged?

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -64,6 +64,11 @@ func (s *gcsObjectClient) PutChunks(ctx context.Context, chunks []chunk.Chunk) e
 			return err
 		}
 		writer := s.bucket.Object(chunk.ExternalKey()).NewWriter(ctx)
+		// Default GCSChunkSize is 8M and for each call, 8M is allocated xD
+		// By setting it to 0, we just upload the object in a single a request
+		// which should work for our chunk sizes.
+		writer.ChunkSize = 0
```
tomwilkie: Can we make this a flag?  We use for Loki and some buffering might be good?

gouthamve: Done

tomwilkie:
LGTM from me too ;-)

PR #1221, opened 2019-02-03
WIP: Compute remainingSamples when unmarshalling bigchunks

### LINKED ISSUE #1220: Ingester creates new bigchunk subchunk immediately after handover
Bigchunks are held as a slice of sub-chunks, and a new sub-chunk is added every N samples by reference to `remainingSamples` which counts down from the maximum (currently 120).

`remainingSamples` is not initialized in `Unmarshal()`, so after an ingester hand-over a new sub-chunk is created immediately.

Leaving the smaller sub-chunk reduces overall compression ratios and slows everything down a bit.

~While it would be expensive to count `remainingSamples` accurately, it seems to me we could estimate it based on the byte size of the last sub-chunk.~ We can count samples at the same time as we determine first and last sample for each sub-chunk.

bboreham:
Fixes #1220 

On handover this allows ingesters to continue adding samples to the last subchunk, rather than starting a new one.

### DIFF
--- a/pkg/chunk/encoding/bigchunk.go
+++ b/pkg/chunk/encoding/bigchunk.go
@@ -128,7 +128,7 @@ func (b *bigchunk) UnmarshalFromBuf(buf []byte) error {
 		b.chunks = append(b.chunks, chunk)
 		b.starts = append(b.starts, start)
 		b.ends = append(b.ends, end)
-		if i == numChunks-1 {
+		if i == numChunks-1 && count < samplesPerChunk {
 			b.remainingSamples = samplesPerChunk - count
 		}
 	}
--- a/pkg/chunk/encoding/bigchunk.go
+++ b/pkg/chunk/encoding/bigchunk.go
@@ -118,7 +118,7 @@ func (b *bigchunk) UnmarshalFromBuf(buf []byte) error {
 			return err
 		}
 
-		start, end, err := firstAndLastTimes(chunk)
+		start, end, count, err := firstAndLastTimesAndCount(chunk)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


bboreham:
This is causing all queries to fail when I try it.  Can't see why.  All I get in logs are:

```
level=warn ts=2019-02-04T09:21:37.638691773Z caller=gokit.go:46 method=/cortex.Ingester/Query duration=282.947µs err=EOF msg="gRPC\n"
level=warn ts=2019-02-04T09:21:37.646512093Z caller=gokit.go:46 method=/cortex.Ingester/Query duration=216.571µs err=EOF msg="gRPC\n"
level=warn ts=2019-02-04T09:21:37.652662593Z caller=gokit.go:46 method=/cortex.Ingester/Query duration=161.303µs err=EOF msg="gRPC\n"
```

tomwilkie:
Weird its causing queries to fail; nothing obviously looks wrong.  Let me know if you want me to dig in.

bboreham:
Yes, if you can figure it out that would be good. To expand on the “queries fail” I’m pretty sure it created corrupt chunks in the DB because queries over that time period continued to fail after rollback. 

I guess we need a unit test for this case but I’m unlikely to get to it myself - too much going on.

bboreham:
I traced the immediate cause of the failure - [`chunkenc.FromData()`](https://github.com/prometheus/tsdb/blob/0b72f9af4c7a7953f89c993e69bbf81b12af49c3/chunkenc/chunk.go#L55) sets `count` to `0` so the next sample is not written at the correct point.

I couldn't immediately see how to reconstruct the missing data.  It ought to be possible, given that we can read all the samples.

PR #1223, opened 2019-02-05
add migration schemas to ruler docker image

jtlisi:
The ruler can currently interact with a postgres db directly without a configdb. However, it does not have the migration files. This PR adds the migration files to the ruler image.

If anyone has an idea of a place to put these files so they can live in the docker build context of the configdb, ruler, and lite images let me know. I wasn't sure the best way to do that without having the make process copy them there.


tomwilkie:
> I wasn't sure the best way to do that without having the make process copy them there.

This I think.  Can you add some makefile magic to copy them in please?

jtlisi:
@tomwilkie 
This should be good for a second look. I squashed it into a single commit.

tomwilkie:
LGTM; might want to tweak it so you don't have to mount the migrations into the root of the build container, but not too fussed.

@bboreham @gouthamve need a second pair of eyes.

Makefile
```diff
@@ -98,7 +102,7 @@ configs-integration-test: build-image/$(UPTODATE)
 		-v $(shell pwd)/.cache:/go/cache \
 		-v $(shell pwd)/.pkg:/go/pkg \
 		-v $(shell pwd):/go/src/github.com/cortexproject/cortex \
-		-v $(shell pwd)/cmd/configs/migrations:/migrations \
+		-v $(shell pwd)/migrations:/migrations \
```
jtlisi: @tomwilkie this change is actually only in the in the integration test container. The build container does not mount the migrations

tomwilkie: +1

Makefile
```diff
@@ -34,13 +34,17 @@ images:
 PROTO_DEFS := $(shell find . $(DONT_FIND) -type f -name '*.proto' -print)
 PROTO_GOS := $(patsubst %.proto,%.pb.go,$(PROTO_DEFS))
 
+# Ensure a each image that requires a migration directory mounted has the updated migrations directory in their
```
bboreham: It took me a while to work out what was going on.  Maybe point to the `cp -r` line lower down which does the actual work?

jtlisi:
@bboreham Changed as requested, let me know if there is anything else

Issue #990, opened 2018-09-06
How should we handle traffic from HA pairs of Prometheus?


tomwilkie:
We should be advising people to run their Prometheus instance in HA, so they don't loose data if the machine Prometheus is running on fails.  However, in a nieve setup, this will lead to a high percentage of out-of-order samples.  The questions is: what to do about this?

Simpilest thing would be to ask for external labels (`replica: 0`, `replica: 1` etc) but this doesn't help at query time.  Can we do something better?  Also, can we come up with a way to solve this on ingestion so we don't have to store everything twice?

We should look at the heuristics thanos use.

tomwilkie:
Enumerating the solutions I can think of, in order of the retrival/write/read pipeline:

1. We could stick to our guns around single source of truth, and provide a better failover solution at the client side.

   This would look a bit like Vulcan's sharded scraping, and would move Cortex away from using Prometheus as the client's agent.  We could have a gossip'd DHT cluster of agents, each scraping 1/Nth of the targets.  When one scraper fails, we could detect the fairly quickly via gossip & something like a phi-accrual failure detector (ala Cassandra) and fail the scraping over to the other scrapers.  Partitions would results in two sets of samples being sent and deduped on the Cortex side (see #3).

   Pros:
   - only one set of sample sent under normal conditions
   - quick failover
   - single source of truth won't make query performance worse

   Cons:
   - complexity
   - moves away from Prometheus as the client, no more client side rule or alert evalutation
   - probably need a solution to #3 consistency issues under partitions

2. As #1, but get rid of any gossip/DHT and recommend a single agent per host.

   Pros:
   - Simpler 
   - Failure modes easy to understand: if a host fails, no more metrics from the jobs on that host.

   Cons:
   - Pain to deploy?

3. Send both HA pairs, with the same external labels, and rely on ordering guarantees to remove duplicates.

   This would work reasonably well; if one of the pair fails the other would instantly "take over" and start having samples accepted.  The downside is twice as much traffic coming through and the errors Prometheus would see on writes.   We might be able to detect and suppress these if we nominate a specific label to identify each replica, don't include said label in the series labels, but track the most recently written label with some timeout - ie, we accept writes from replica "A", ignoring all others, until we haven't seen a write for ~1m, then we accept what ever shard gets to us first, replica "B" etc.  Then When we reject writes from the other shard then it won't be a error.

[...]

PR #1226, opened 2019-02-08
Support receiving samples from HA Prometheus pairs.

cstyan:
WIP, just been running tests with the mock or with sample programs using the HA tracker that just talked to consul on my laptop.

I've copied the the kvstore interface from `pkg/ring` to it's own package. That can go away and I can make the necessary changes to the code in the ring package, or I can continue and make the changes in the ring package to have it use the new kvstore package.

cc @tomwilkie @gouthamve @jtlisi 


tomwilkie:
Design doc: https://docs.google.com/document/d/1UunkvlE5W3p_Zz7g6Q0Uov_maSvLejvrGXe324B3Pgk/edit

Issue: #990

pkg/distributor/distributor.go
```diff
@@ -241,18 +253,41 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		return nil, err
 	}
 
+	d.haFlag = true
```
tomwilkie: Can you add this to the `Config` struct and register it as a flag appropriately.

pkg/distributor/distributor.go
```diff
@@ -241,18 +253,41 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
+		if err != nil && (cluster != "" && instance != "") {
+			return nil, err
+		}
+		req.Timeseries[0].Labels = ts.Labels
+
+		// Lookup the cluster/instance here to see if we want to accept this sample
+		if inst := d.instances.LookupInstance(cluster, instance); inst.instance != instance && (cluster != "" && instance != "") { // Don't accept the sample.
+			// If we assume all samples will have the same cluster/instance labels we can return an error here.
+			// Should we just silently drop here?
+			return nil, fmt.Errorf("dropping sample")
```
tomwilkie: I think we do want to silently drop, but we probably want to check acceptance (or not) after the validation loop below.  And the you can check all the timeseries have the same cluster/instance label in that loop too, for extra points.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,243 @@
...
+
+	"github.com/weaveworks/common/mtime"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/kvstore"
+)
+
+// Treat these as consts.
+var haInstanceBytes = []byte("prom_ha_instance")
+var haClusterBytes = []byte("prom_ha_cluster")
```
tomwilkie: These probably need to be configurable (in the config struct), and although I don't want to encourage bikeshedding, I'd go with `cluster` and `replica` as default.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,243 @@
...
+		client = kvstore.NewInMemoryKVClient(codec)
+	} else {
+		client, err = kvstore.NewConsulClient(kvstore.ConsulConfig{
+			Host: "localhost:8500",
+		}, codec)
+
+		if err != nil {
+			os.Exit(1)
+		}
+	}
```
tomwilkie: Can you use `ring.newKVStore` and pass the codec to that too?

pkg/distributor/distributor.go
```diff
@@ -244,6 +264,36 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
+		// If the sample doesn't have either HA label, accept it.
+		if cluster == "" && replica == "" {
+			goto acceptSample
+		}
+		// Right now removeHALabels will only return an error if it finds one but not both of the labels.
+		if err != nil {
+			if d.cfg.HARejectOne {
+				return nil, err
+			}
+			goto acceptSample
```
tomwilkie: Don't use a goto!  Put this all in a function.

pkg/distributor/ha_tracker.go
tomwilkie: Also, split out the `ring.Config` into a `KVConfig` that embeds the `Store string` and the `ConsulConfig`, and then put this in your config.  Then have `ring.newKVStore()` take this config.

cmd/distributor/main.go
```diff
@@ -78,6 +79,9 @@ func main() {
 		os.Exit(1)
 	}
 
+	// consulConfig := kvstore.ConsulConfig{}
+	// consulConfig = ringConfig.Consul.(kvstore.ConsulConfig)
+	// distributorConfig.ConsulConfig = ringConfig.Consul
```
tomwilkie: Lets remove this code now the config is plumbed in correctly.

pkg/distributor/distributor.go
```diff
@@ -79,6 +79,12 @@ var (
 	})
 )
 
+type clusterTracker interface {
+	LookupInstance(ctx context.Context, cluster string, instance string) *instanceTracker
+	setTimeout(overwrite, write int64)
```
tomwilkie: This seems to never be called?

pkg/distributor/distributor.go
```diff
@@ -98,8 +107,13 @@ type Distributor struct {
 // create a Distributor
 type Config struct {
 	EnableBilling bool
+	HAReplicas    bool
+	HARejectOne   bool
+	KVStore       string
 	BillingConfig billing.Config
 	PoolConfig    ingester_client.PoolConfig
+	TrackerConfig HATrackerConfig
+	ConsulConfig  ring.ConsulConfig
```
tomwilkie: `ConsulConfig` (and eventually `KVConfig`, see below) should be part of `TrackerConfig` IMHO.

pkg/distributor/distributor.go
```diff
@@ -98,8 +107,13 @@ type Distributor struct {
 // create a Distributor
 type Config struct {
 	EnableBilling bool
+	HAReplicas    bool
+	HARejectOne   bool
+	KVStore       string
```
tomwilkie: These should be part of `TrackerConfig`, no?

pkg/distributor/distributor.go
```diff
@@ -244,6 +264,36 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
+	if d.cfg.HAReplicas && len(req.Timeseries) > 0 {
+		ts := req.Timeseries[0]
+		cluster, replica, err := findHALabels([]byte(d.cfg.TrackerConfig.ReplicaLabel), []byte(d.cfg.TrackerConfig.ClusterLabel), ts.Labels)
+		// If the sample doesn't have either HA label, accept it.
+		if cluster == "" && replica == "" {
+			goto acceptSample
+		}
+		// Right now removeHALabels will only return an error if it finds one but not both of the labels.
+		if err != nil {
+			if d.cfg.HARejectOne {
```
tomwilkie: When would we ever turn this on?  Perhaps if this was per-use, but right now I'd say just remove it.

pkg/distributor/distributor.go
```diff
@@ -244,6 +264,36 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
+				return nil, err
+			}
+			goto acceptSample
+		}
+
+		// If there's no error we found both HA labels and should lookup whether we
+		// want to accept samples from that replica for the given cluster.
+		req.Timeseries[0].Labels = ts.Labels
```
tomwilkie: Is this a noop?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,291 @@
...
+func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.ReplicaLabel, "ha-tracker.replica", "replica", "Prometheus label to look for in samples to identify a Proemtheus HA replica.")
+	f.StringVar(&cfg.ClusterLabel, "ha-tracker.cluster", "cluster", "Prometheus label to look for in samples to identify a Poemtheus HA cluster.")
+	f.StringVar(&cfg.Prefix, "ha-tracker.prefix", "prom_ha", "Prefix to store HA cluster accept instance information under within KV store.")
+	// Need to copy these flags since we don't want to break compatability of the existing 'consul.<var>' flags
+	// but we want the ability to use different Consul instances for the ring and for HA cluster tracking.
+	f.StringVar(&cfg.consul.Host, "ha-tracker.consul-hostname", "localhost:8500", "Hostname and port of Consul.")
+	f.StringVar(&cfg.consul.ACLToken, "ha-tracker.consul-acltoken", "", "ACL Token used to interact with Consul.")
+	f.DurationVar(&cfg.consul.HTTPClientTimeout, "ha-tracker.consul-client-timeout", 2*longPollDuration, "HTTP timeout when talking to consul")
+	f.BoolVar(&cfg.consul.ConsistentReads, "ha-tracker.consul-consistent-reads", true, "Enable consistent reads to consul.")
```
tomwilkie: You should use `cfg.consul.RegisterFlags()` and update it to take an (optional) prefix.

pkg/distributor/distributor.go
tomwilkie: Looks like there is already a ConsulConfig in `TrackerConfig`.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,291 @@
...
+
+		if err != nil {
+			os.Exit(1)
+		}
+	}
+	var ctx context.Context
+	ctx, quit := context.WithCancel(context.Background())
+	t := &tracker{
+		logger:           util.Logger,
+		elected:          make(map[string]instanceTracker),
```
tomwilkie: We prefer `map[string]instanceTracker{}` when you don't know the size.

pkg/distributor/distributor.go
cstyan: I'm using it in `ha_tracker_test.go`. I can remove it and just modify the values on the struct in line, or remove the function from the interface and just have it local within the test file.

pkg/distributor/distributor.go
cstyan: I think it makes sense to keep `HAReplicas` as part of distributor config. The rest can move.

pkg/distributor/distributor.go
cstyan: I don't know why this was added

pkg/distributor/distributor.go
tomwilkie: > or remove the function from the interface and just have it local within the test file.

Yeah, this sounds better.

cmd/distributor/main.go
```diff
@@ -48,6 +48,7 @@ func main() {
 			},
 			ExcludeRequestInLog: true,
 		}
+		// consulConfig
```
tomwilkie: I think this can be removed.

pkg/distributor/distributor.go
```diff
@@ -78,6 +78,12 @@ var (
 	})
 )
 
+type clusterTracker interface {
+	LookupReplica(ctx context.Context, cluster string, replica string) *replicaTracker
+	setTimeout(overwrite, write int64)
+	loop(ctx context.Context)
```
tomwilkie: Lets have the constructor do the `go loop()` so it doesn't have to be in the interface.  You'll want to add a `Stop` though.

pkg/distributor/distributor.go
```diff
@@ -218,6 +238,44 @@ func shardByAllLabels(userID string, labels []client.LabelPair) (uint32, error)
...
+		if pair.Name.Equal(replica) {
+			*labels = append((*labels)[:i], (*labels)[i+1:]...)
+			return
+		}
+	}
+}
+
+// Returns an error if we should not accept samples based on the HA labels found in ts. Boolean value
+// is whether we found both HA labels. We use the boolean value to know if we want to remove the
+// replica label.
```
tomwilkie: I don't remember is we discussed this: why are we removing the HA labels?  Should we leave them in?  It will make failover between replicas visible to the users (and cause huge amount of series churn).  OTOH it might be useful.  WDYT?

pkg/distributor/distributor.go
```diff
@@ -218,6 +238,44 @@ func shardByAllLabels(userID string, labels []client.LabelPair) (uint32, error)
...
+func (d *Distributor) checkSample(ctx context.Context, ts client.PreallocTimeseries) (bool, error) {
```
tomwilkie: This error returned is never used, and is only one of two values - nil or `fmt.Errorf...`.  Probably want to use a bool.

pkg/distributor/distributor.go
```diff
@@ -218,6 +238,44 @@ func shardByAllLabels(userID string, labels []client.LabelPair) (uint32, error)
...
+	cluster, replica, err := findHALabels([]byte(d.cfg.TrackerConfig.ReplicaLabel), []byte(d.cfg.TrackerConfig.ClusterLabel), ts.Labels)
```
tomwilkie: This `err` is also never used - might want to consider just not returning it, and using empty strings when you don't find the labels.

pkg/distributor/distributor.go
```diff
@@ -218,6 +238,44 @@ func shardByAllLabels(userID string, labels []client.LabelPair) (uint32, error)
...
+
+	// If the sample doesn't have either HA label, accept it.
+	if cluster == "" && replica == "" {
```
tomwilkie: `||` not `&&`?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,316 @@
...
+		quit:             quit,
+	}
+	go t.loop(ctx)
+	return t, nil
+}
+
+// Follows pattern used by ring for WatchKey.
+func (c *tracker) loop(ctx context.Context) {
+	defer close(c.done)
+	c.client.WatchPrefix(context.Background(), fmt.Sprintf("%s", c.prefix), func(value interface{}) bool {
```
tomwilkie: Why is `loop` in the interface?

And if you've going to take a ctx, you'll want to use it...

Also the sprintf is a noop.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,316 @@
...
+			return nil, err
+		}
+	case "inmemory":
+		// TODO: do we want to use the same init do for sync.Once as we do in the ring?
+		client = ring.NewInMemoryKVClient(codec)
+	default:
+		return nil, fmt.Errorf("invalid KV store type: %s", cfg.KVStore.Store)
+	}
+
+	ctx, quit := context.WithCancel(context.Background())
```
tomwilkie: Can you call `quit` `cancel` pls?  And in the struct.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,316 @@
...
+	elected     map[string]replicaTracker
+	done        chan struct{}
+	quit        context.CancelFunc
+	// The labels to look for as bytes.
+	clusterBytes []byte
+	replicaBytes []byte
+	prefix       string
+
+	// We should only update the timestamp this often.
+	writeTimeout int64
```
tomwilkie: Why not `time.Duration`?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,316 @@
...
+
+	consulLock       sync.RWMutex
+	overwriteTimeout int64 // Timeout that has to occur before we overwrite to a different replica.
```
tomwilkie: Why not `time.Duration`?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// Replicas we are accepting samples from.
+	electedLock sync.RWMutex
+	elected     map[string]replicaTracker
+	done        chan struct{}
+	quit        context.CancelFunc
+	// The labels to look for as bytes.
+	clusterBytes []byte
+	replicaBytes []byte
+	prefix       string
```
tomwilkie: Can you embed the cfg in this struct?  And then not duplicate these values.

pkg/ring/kvstore.go
```diff
@@ -18,35 +19,55 @@ var inmemoryStore KVClient
...
+	f.StringVar(&cfg.Store, prefix+"store", "consul", "Backend storage to use for the ring (consul, inmemory).")
+}
+
 // CASCallback is the type of the callback to CAS.  If err is nil, out must be non-nil.
 type CASCallback func(in interface{}) (out interface{}, retry bool, err error)
 
-func newKVStore(cfg Config) (KVClient, error) {
+// NewKVStore creates a new KVstore client (inmemory or consul) based on the config,
+// encodes and decodes data for storage using the codec.
+func NewKVStore(cfg Config, codec Codec) (KVClient, error) {
```
tomwilkie: This should take a `KVConfig`, no?

pkg/distributor/distributor.go
cstyan: If we find both and want to accept the current sample, we want to send the sample on to the rest of Cortex for storage, but if we kept both we would end up with a series per replica after we do any failovers between replicas. 

If we find both labels but only send the sample on for storage with the `cluster` label we would only ever have one series per cluster, regardless of the # of replicas.

pkg/distributor/ha_tracker.go
cstyan: Timestamps within samples are int64. We could import the Prometheus timestamp package and convert back to a time type?

pkg/ring/kvstore.go
cstyan: I think so, but the ring config is what contains the `Mock` value. Where should we move that?

tomwilkie: I'd move the mock in to KVConfig too.

pkg/distributor/ha_tracker.go
tomwilkie: Oh I see.  Lets keep these a time.Durations, and do the conversion on the Prometheus timestamps.

Note Prometheus timestamps are milliseconds, not nano.

pkg/distributor/distributor.go
tomwilkie: Sure, but only one series would have a value at any one time, so in aggregations people wouldn't notice the difference, right?

cstyan: Right, but that would potentially require changes to any queries they're currently using against cortex wouldn't it?

pkg/distributor/distributor.go
cstyan: Do we want to look for a way to have `Push` return a 202 when we don't want to accept the sample? I've just had it return `return &client.WriteResponse{}, nil` like we're accepting the sample which AFAICT results in a 200.

I didn't see a way to do this with the http grpc stuff you mentioned.

pkg/distributor/distributor.go
cstyan: Yep. Leftover from previous code that wasn't accepting samples that only had one HA label.

pkg/distributor/distributor.go
cstyan: :+1:

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,312 @@
...
+			return desc, false, nil
+		}
+
+		// TODO: resolve this below
+		// I think we want to get rid of this block. The CAS should only write to Consul if necessary, and then return
+		// the replica desc for the replica we want to accept samples from, or nil.
+		// Or should we return a new error type.
+		if desc.Replica != replica {
+			level.Info(util.Logger).Log("msg", "replicas did not match in CAS call")
+			return nil, false, fmt.Errorf("replicas did not match, rejecting sample: %s != %s", replica, desc.Replica)
```
cstyan: I need some time to think about what we should be doing here. I want to make sure we return a value that indicates that we shouldn't actually CAS (returning desc would do that), but we also want to get the value of desc out of this and store it locally. 

Right now I don't know how if we're accepting samples from replica2, but the current distributor doesn't know that yet, would make the earlier call to consul to ask it what replica we should accept from and save that information locally but still reject the sample from replica1. Other than getting `WatchPrefix` actually working.

pkg/distributor/distributor.go
cstyan: Let me know what you want to do here.

cstyan:
@tomwilkie PTAL

I think I've addressed all of the previous review, the outstanding items I wanted to fix, and the refactor of `LookupReplica`/checking the cache. 

The cache check now only determines if the incoming sample should be accepted, and if the value stored in the cache is still valid. `LookupReplica` is responsible for handling those decisions. I think `setReplica` and `casWrapper` are still called the same amount of time but in places that make more sense.
Anywhere I've left a TODO comment is generally something I'm thinking can be changed, so if you have thoughts let me know.

cc @gouthamve

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+type HATrackerConfig struct {
+	ReplicaLabel string
+	ClusterLabel string
+
+	KVStore ring.KVConfig
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.ReplicaLabel, "ha-tracker.replica", "replica", "Prometheus label to look for in samples to identify a Proemtheus HA replica.")
```
jtlisi: Since this label is for internal use by cortex can we default to something like `__replica`? That way it doesn't interfere with any series that already use this label.

csmarchbanks: I could be not interpreting this correctly, but this is the external label Prometheus is writing correct? Similar to `--query.replica-label` in Thanos? If so, has there been thought about making this configurable per customer for people with different configurations?

As far as defaults go, I vote `prometheus_replica` since that is what the Prometheus operator uses by default.

cstyan: Correct, this is the external label you would add to your Prometheus config. You're right it probably should be configurable per customer. At the same time, I'm thinking we can actually just drop both the cluster and replica label before storing the data.

pkg/ring/consul_client.go
```diff
@@ -203,6 +207,55 @@ func (c *consulClient) WatchKey(ctx context.Context, key string, f func(interfac
...
+		for _, kvp := range kvps {
+			out, err := c.codec.Decode(kvp.Value)
+			if err != nil {
+				level.Error(util.Logger).Log("msg", "error decoding list of values for prefix:key", "prefix", prefix, "key", kvp.Key, "err", err)
+				continue
+			}
+			if !f(KVEntry{
+				Cluster:      kvp.Key,
+				ReplicaEntry: out,
+			}) {
```
tomwilkie: Can we pass `f` a key and value instead of reconstructing this?  Signature probably wants to be `func(string, interface{})`.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+
+// Follows pattern used by ring for WatchKey.
+func (c *tracker) loop(ctx context.Context) {
+	defer close(c.done)
+	// The KVStore config we gave when creating c should have contained a prefix,
+	// which would have given us a prefixed KVStore client. So, we can pass empty string here.
+	c.client.WatchPrefix(ctx, "", func(value interface{}) bool {
+		if value == nil {
+			return true
+		}
```
tomwilkie: Under what circumstances is `value` ever `nil`?

pkg/ring/consul_client.go
```diff
@@ -245,6 +298,11 @@ func (c *prefixedConsulClient) WatchKey(ctx context.Context, key string, f func(
 	c.consul.WatchKey(ctx, c.prefix+key, f)
 }
 
+// WatchPrefix watches a prefix. For a prefix client it appends the prefix argument to the clients prefix.
+func (c *prefixedConsulClient) WatchPrefix(ctx context.Context, prefix string, f func(interface{}) bool) {
+	c.consul.WatchPrefix(ctx, fmt.Sprintf("%s/%s", c.prefix, prefix), f)
```
tomwilkie: Once `f` receives a path, it would make sense for the prefix to be removed here.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (e *ErrDontOverwrite) Error() string {
+	return fmt.Sprintf("don't overwrite")
+}
+
+// Track the most recent timestamp for the replica we're
+// accepting samples from in a Prometheus HA setup.
+type replicaTracker struct {
+	replica   string
+	timestamp time.Time
```
tomwilkie: What is `timestamp`?  It this the last updated time?  Or is the timestamp from the sample?  Is this the "freshness horizon" or "timeout" for this value?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+		// update what's in the KVStore so we return nil for the interface.
+		return nil, false, &ErrDontOverwrite{}
+	})
+	return err
+}
+
+// Do we want to actually do this all the time, or only do it when the write timeout is reached?
+// Wraps updating the timestamp for an replica in the local cache so that we can grab the
+// lock and then check if we should actually update it. The timestamp should only ever go up.
+func (c *tracker) updateTimestamp(cluster, replica string, ts time.Time) {
```
tomwilkie: Unused, pls remove.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+		if _, ok := err.(*ErrDontOverwrite); !ok {
+			level.Error(c.logger).Log("msg", "non ErrDontOverwrite error", "err", err)
+			return nil
+		}
+	}
+	// There was no error or error was ErrDontOvewrite, so we still want to accept the sample.
+	c.setReplica(cluster, replica, now)
+	c.electedLock.RLock()
+	ret = c.elected[cluster]
+	c.electedLock.RUnlock()
```
tomwilkie: As discussed: lets not update the local cache in this case - WatchPrefix might be behind for instance, and will clobber this.  Just rely on watch prefix.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+type ErrDontOverwrite struct {
+}
+
+func (e *ErrDontOverwrite) Error() string {
+	return fmt.Sprintf("don't overwrite")
+}
+
+// Track the most recent timestamp for the replica we're
+// accepting samples from in a Prometheus HA setup.
+type replicaTracker struct {
```
tomwilkie: Can we get rid of this and just use `ReplicaDescription` from the generated protos?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+	err := c.client.CAS(ctx, fmt.Sprintf("/%s", cluster), func(in interface{}) (out interface{}, retry bool, err error) {
+		desc, ok := in.(*ReplicaDesc)
+		// There was either invalid or no data for the key, so we now accept samples
+		// from this replica.
+		if !ok || desc == nil {
+			desc = &ReplicaDesc{
+				Replica:   replica,
+				Timestamp: timestamp.FromTime(now),
+			}
+			return desc, false, nil
```
tomwilkie: Why do you not want to retry here?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+
+		// As far as the timestamp in the KVStore is concerned none of the Distributors have
+		// received a sample from the stored replica recently, so we should overwrite the
+		// replica in the KVStore and accept this sample.
+		if now.Sub(timestamp.Time(desc.Timestamp)) > c.overwriteTimeout {
+			desc = &ReplicaDesc{
+				Replica:   replica,
+				Timestamp: timestamp.FromTime(now),
+			}
+			return desc, false, nil
```
tomwilkie: Why not retry?

pkg/distributor/ha_tracker.go
csmarchbanks: I think replica label we should definitely drop, otherwise you would have some weird disjointed series whenever you switch replicas.

I don't think cluster should be dropped since it is common for multiple clusters to send to a cortex and would need to be distinguished. Speaking of cluster, if someone only has one cluster but wants to use deduping they would have to specify the cluster label correct?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+			level.Info(util.Logger).Log("msg", "replicas did not match in CAS call")
+			return nil, false, fmt.Errorf("replicas did not match, rejecting sample: %s != %s", replica, desc.Replica)
+		}
+
+		// As far as the timestamp in the KVStore is concerned none of the Distributors have
+		// updated the timestamp for this cluster/replica in more than the write timeout.
+		// We should update the timestamp so that we don't potentially failover to accepting
+		// samples from another replica when we don't need to.
+		// This may only catch local CAS' updating the KVStore, not other Distributors.
+		if now.Sub(timestamp.Time(desc.Timestamp)) > c.writeTimeout {
```
tomwilkie: Could be pick better names than writetimeout and overwritetimeout?  Maybe update and failover?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,296 @@
...
+			desc = &ReplicaDesc{
+				Replica:   replica,
+				Timestamp: timestamp.FromTime(now),
+			}
+			return desc, false, nil
+		}
+		// TODO: should this just return nil for an error as well? It results in an error log when
+		// we return ErrDontOverwrite. Here the cluster/replica matched, but we just don't want to
+		// update what's in the KVStore so we return nil for the interface.
+		return nil, false, &ErrDontOverwrite{}
```
tomwilkie: Still don't like `ErrDontOverwrite`, I think updating CAS so a nil value means dont' write is more correct.  Errors are for errors, not success.

pkg/distributor/ha_tracker.go
tomwilkie: I think returning an error here is fine, but it should be `httpgrp.Errorf(http.StatusOK, message...)`.

pkg/distributor/distributor.go
```diff
@@ -78,6 +78,11 @@ var (
 	})
 )
 
+type clusterTracker interface {
+	LookupReplica(ctx context.Context, cluster string, replica string) *replicaTracker
```
tomwilkie: This need to return an error.  You never swallow errors.

tomwilkie: Lookup implies this is side effect free.  Probably want to go with `CheckReplica` or something?

pkg/distributor/ha_tracker.proto
```diff
@@ -0,0 +1,17 @@
...
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+
+message ClustersDesc {
+    map<string, int64> clusters = 1 [(gogoproto.nullable) = false];
+}
+
+message ReplicaDesc {
+    string replica = 1;
+    int64 timestamp = 2;
```
tomwilkie: Use gogoproto time here.  I'll update this for you.

pkg/distributor/distributor.go
tomwilkie: And probably not return the replica, as its checking it validity.  Just return an error?

pkg/distributor/model.go
```diff
@@ -0,0 +1,18 @@
+package distributor
+
+import (
+	// "sort"
+	// "time"
```
tomwilkie: :-(

pkg/distributor/ha_tracker_test.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	for _, c := range cases {
+		cluster, replica := findHALabels([]byte(cfg.ReplicaLabel), []byte(cfg.ClusterLabel), c.labelsIn)
+		assert.Equal(t, c.expected.cluster, cluster)
+		assert.Equal(t, c.expected.replica, replica)
+	}
+}
+
+func Test_checkLocal(t *testing.T) {
```
tomwilkie: No `_` in function names pls.

pkg/distributor/ha_tracker.go
cstyan: The only reason I can think _to retry_ is when there's an error talking to Consul. I supposed this could be one of those cases?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+	client ring.KVClient
+	// We should only update the timestamp if the difference
+	// between the stored timestamp and the time we received a sample at
+	// is more than this duration.
+	updateTimeout time.Duration
+	// We should only failover to accepting samples from a replica
+	// other than the replica written in the KVStore if the difference
+	// between the stored timestamp and the time we received a sample is
+	// more than this duration
+	failoverTimeout time.Duration
```
tomwilkie: Move these to config?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+// Wraps calls to client CAS with locking and error checking
+func (c *tracker) casWrapper(ctx context.Context, cluster, replica string, now time.Time) error {
+	// We have should a prefixed KVStore client, so we want to pass /cluster instead of just cluster.
+	// Prefixed client does prefix+key without separating the values.
+	err := c.client.CAS(ctx, fmt.Sprintf("/%s", cluster), func(in interface{}) (out interface{}, retry bool, err error) {
+		desc, ok := in.(*ReplicaDesc)
+		// There was either invalid or no data for the key, so we now accept samples
+		// from this replica.
+		if !ok || desc == nil {
+			return &ReplicaDesc{Replica: replica, ReceivedAt: timestamp.FromTime(now)}, false, nil
```
tomwilkie: Why is this a pointer-to-a ReplicaDesc?

tomwilkie: Also, can we have newlines in the struct instantiation pls.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+		// As far as the timestamp in the KVStore is concerned none of the Distributors have
+		// updated the timestamp for this cluster/replica in more than the write timeout.
+		// We should update the timestamp so that we don't potentially failover to accepting
+		// samples from another replica when we don't need to.
+		// This may only catch local CAS' updating the KVStore, not other Distributors.
+		if now.Sub(timestamp.Time(desc.ReceivedAt)) > c.updateTimeout {
+			desc.ReceivedAt = timestamp.FromTime(now)
+			return desc, false, nil
+		}
+		return nil, false, nil
```
tomwilkie: Stick a comment here saying `nil, false, nil` indicated success and will stop CAS overwriting the value.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+
+	if cas {
+		return c.casWrapper(ctx, cluster, replica, now)
+	}
+	return nil
+}
+
+// Returns an error if we don't want to accept the sample.
+func (c *tracker) checkKVStore(ctx context.Context, cluster, replica string, now time.Time) error {
+	return c.casWrapper(ctx, cluster, replica, now)
```
tomwilkie: Its a code smell to have a function which just call another one and does nothing else.  Can we remove this?  `s/casWrapper/checkKVStore/`

cstyan:
@tomwilkie @gouthamve  @jtlisi PTAL

pkg/distributor/distributor.go
```diff
@@ -77,6 +78,11 @@ var (
 	})
 )
 
+type clusterTracker interface {
+	CheckReplica(ctx context.Context, cluster string, replica string) error
+	Stop()
+}
```
tomwilkie: As the implementation is internal to this package, and we didn't go with the layered approach, probably don't need an interface for this.

pkg/distributor/distributor.go
```diff
@@ -141,6 +154,11 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 	replicationFactor.Set(float64(ring.ReplicationFactor()))
 	cfg.PoolConfig.RemoteTimeout = cfg.RemoteTimeout
 
+	replicas, err := NewClusterTracker(cfg.TrackerConfig)
```
tomwilkie: Probably only want to instantiate this is cfg.HAReplicas is true?  Won't it fail if consul flags aren't specified?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+		// from this replica. Invalid could mean that the timestamp in the KV store was
+		// out of date based on the update and failover timeouts when compared to now.
+		return &ReplicaDesc{Replica: replica, ReceivedAt: timestamp.FromTime(now)}, true, nil
+	})
+}
+
+// Modifies the labels parameter in place, removing labels that match
+// the replica or cluster label and returning their values. Returns an error
+// if we find one but not both of the labels.
+func findHALabels(replicaLabel, clusterLabel []byte, labels []client.LabelPair) (string, string) {
```
tomwilkie: labels are now strings in master.

pkg/distributor/distributor.go
```diff
@@ -217,6 +237,42 @@ func shardByAllLabels(userID string, labels []client.LabelAdapter) (uint32, erro
...
+	if cluster == "" || replica == "" {
+		return false, nil
+	}
+
+	// At this point we know we have both HA labels, we should lookup
+	// the cluster/instance here to see if we want to accept this sample.
+	err := d.replicas.CheckReplica(ctx, cluster, replica)
+	// CheckReplica should only have returned an error if there was a real error talking to Consul, or if the replica labels don't match.
+	if err != nil { // Don't accept the sample.
+		level.Error(util.Logger).Log("error", err)
```
tomwilkie: Can you add `"msg", "rejecting push due to replication"`?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	logger log.Logger
+	cfg    HATrackerConfig
+
+	// Replicas we are accepting samples from.
+	electedLock sync.RWMutex
+	elected     map[string]ReplicaDesc
+	done        chan struct{}
+	cancel      context.CancelFunc
+
+	client ring.KVClient
```
tomwilkie: super nit pick: move `client` up after `cfg` pls.z

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	// more than this duration
+	FailoverTimeout time.Duration
+
+	KVStore ring.KVConfig
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
+	f.DurationVar(&cfg.UpdateTimeout, "ha-tracker.update-timeout", time.Second, "Update the timestamp in the KV store for a given cluster/replica only after this amount of time has passed since the current stored timestamp.")
+	f.DurationVar(&cfg.FailoverTimeout, "ha-tracker.failover-timeout", 10*time.Second, "If we don't receive any samples from the accepted replica for a cluster in this amount of time we will failover to the next replica we recieve a sample from.")
```
tomwilkie: Might want to specify this should be g.t. than `update-timeout`.

Also, is 10s a good timeout?  How does this relate to scrape interval?

pkg/distributor/ha_tracker.go
tomwilkie: :+1:, looks like Thanos defaults to `replica` for this label.  I do think we should use `__replica__` or `__prometheus_replica__`, as the change a random timeseries has a `replica` label is quite high.

pkg/ring/kvstore.go
```diff
@@ -18,35 +19,56 @@ var inmemoryStore KVClient
...
+	Mock KVClient
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *KVConfig) RegisterFlags(f *flag.FlagSet, prefix string) {
+	cfg.Consul.RegisterFlags(f, prefix)
+	if prefix == "" {
+		prefix = "ring"
+	}
+	prefix = prefix + "."
```
tomwilkie: This seems a it redundant - we should just expect any prefix to end with a `.`.  And can we not default to `ring`, but instead add that in the appropriate place?

pkg/distributor/ha_tracker.go
tomwilkie: Ping?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	if ok && entry.Replica == replica && now.Sub(timestamp.Time(entry.ReceivedAt)) < c.cfg.UpdateTimeout {
+		return nil
+	}
+	return c.checkKVStore(ctx, cluster, replica, now)
+}
+
+func (c *tracker) checkKVStore(ctx context.Context, cluster, replica string, now time.Time) error {
+	// We have should a prefixed KVStore client, so we want to pass /cluster instead of just cluster.
+	// Prefixed client does prefix+key without separating the values.
+	return c.client.CAS(ctx, fmt.Sprintf("/%s", cluster), func(in interface{}) (out interface{}, retry bool, err error) {
```
tomwilkie: Don't we need to include the userID in this?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// CheckReplica checks the cluster and replica against the backing KVStore and local cache in the
+// tracker c to see if we should accept the incomming sample. It will return an error if the sample
+// should not be accepted. Note that internally this function does checks against the stored values
+// and may modify the stored data, for example to failover between replicas after a certain period of time.
+func (c *tracker) CheckReplica(ctx context.Context, cluster, replica string) error {
+	now := mtime.Now()
+	c.electedLock.RLock()
+	entry, ok := c.elected[cluster]
```
tomwilkie: Don't we need to include the user ID in here?

pkg/distributor/distributor.go
tomwilkie: This is also going to spam the hell out of the logs.  Can you include a check for the status code, and not log 202s?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	// between the stored timestamp and the time we received a sample is
+	// more than this duration
+	FailoverTimeout time.Duration
+
+	KVStore ring.KVConfig
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
+	f.DurationVar(&cfg.UpdateTimeout, "ha-tracker.update-timeout", time.Second, "Update the timestamp in the KV store for a given cluster/replica only after this amount of time has passed since the current stored timestamp.")
```
tomwilkie: This means we're doing 1 QPS per user per distributor.  I'd say we're limited to 100QPS to consul.  Our biggest cluster has ~10 and ~60 active users.  So this is 600QPS.  Might need to tweak this down...

pkg/distributor/model.go
```diff
@@ -0,0 +1,15 @@
+package distributor
+
+import (
+	"github.com/golang/protobuf/proto"
+)
+
+// ProtoReplicaDescFactory makes new InstanceDescs
+func ProtoReplicaDescFactory() proto.Message {
```
tomwilkie: Nit: just stuff this in ha_tracker.go.  No need for another file.

pkg/ring/consul_client.go
```diff
@@ -29,12 +30,15 @@ type ConsulConfig struct {
...
-func (cfg *ConsulConfig) RegisterFlags(f *flag.FlagSet) {
-	f.StringVar(&cfg.Host, "consul.hostname", "localhost:8500", "Hostname and port of Consul.")
-	f.StringVar(&cfg.Prefix, "consul.prefix", "collectors/", "Prefix for keys in Consul.")
-	f.StringVar(&cfg.ACLToken, "consul.acltoken", "", "ACL Token used to interact with Consul.")
-	f.DurationVar(&cfg.HTTPClientTimeout, "consul.client-timeout", 2*longPollDuration, "HTTP timeout when talking to consul")
-	f.BoolVar(&cfg.ConsistentReads, "consul.consistent-reads", true, "Enable consistent reads to consul.")
+func (cfg *ConsulConfig) RegisterFlags(f *flag.FlagSet, prefix string) {
+	if prefix != "" {
+		prefix = prefix + "."
+	}
```
tomwilkie: Same as on `KVClientConfig`, lets not do this and instead ask callers to append a dot.

pkg/ring/consul_client.go
```diff
@@ -203,6 +209,57 @@ func (c *consulClient) WatchKey(ctx context.Context, key string, f func(interfac
...
+
+		index = meta.LastIndex
+		for _, kvp := range kvps {
+			out, err := c.codec.Decode(kvp.Value)
+			if err != nil {
+				level.Error(util.Logger).Log("msg", "error decoding list of values for prefix:key", "prefix", prefix, "key", kvp.Key, "err", err)
+				continue
+			}
+			// We should strip the prefix from the front of the key.
+			chunks := strings.Split(kvp.Key, prefix)
```
tomwilkie: Probably want to us https://golang.org/pkg/strings/#TrimPrefix here.

pkg/ring/consul_client.go
tomwilkie: Ping?

pkg/ring/ring.go
```diff
@@ -56,19 +56,17 @@ var ErrEmptyRing = errors.New("empty ring")
...
 	ReplicationFactor int           `yaml:"replication_factor,omitempty"`
 
 	Mock KVClient
 }
 
-// RegisterFlags adds the flags required to config this to the given FlagSet
+// RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
-	cfg.Consul.RegisterFlags(f)
+	cfg.KVStore.RegisterFlags(f, "")
```
tomwilkie: Should this be "ring"?

tomwilkie: Hmmm actually no it shouldn't.  The flags were `consul...` before, and after this change shouldn't be `ring.consul...` or you'll break all the previous deployments.

pkg/ring/kvstore.go
tomwilkie: It looks like the `ring` prefix wasn't there before, and we don't want to rename flags (it'll break upgrades).

tomwilkie:
Need to fix that build and address these comments, but I promise this is the last - its almost there!

pkg/distributor/distributor.go
cstyan: Is there any way to check the status code `202` other than `httpgrpc.HTTPResponseFromError`? We probably want to check this in `consul_client.go`, we log the error in `cas` if retry is false.

pkg/ring/kvstore.go
cstyan: Right, that's why I'm not defaulting prefix from `""` to `"ring"` until after we register the Consul flags.

```cfg.Consul.RegisterFlags(f, prefix)
	if prefix == "" {
		prefix = "ring"
	}```

File: pkg/distributor/ha_tracker.go (hunk 52 above)
Comment (cstyan) [reply pkg/distributor/ha_tracker.go]: Right. `UpdateTimeout` should probably be at least greater than 2x your scrape interval. `FailoverTimeout` again greater than update. Should we assume everyone has scrape intervals around 30s for the defaults?

File: pkg/distributor/distributor.go (hunk 16 above)
Comment (tomwilkie) [reply pkg/distributor/distributor.go]: Yeah lets remove, we can't deal with the series chunk.

File: pkg/distributor/distributor.go (hunk 50 above)
Comment (tomwilkie) [reply pkg/distributor/distributor.go]: No, no other way.  Thats fine too.

File: pkg/distributor/distributor.go [hunk 61]
```diff
@@ -96,8 +105,10 @@ type Distributor struct {
 // create a Distributor
 type Config struct {
 	EnableBilling bool
+	HAReplicas    bool
```
tomwilkie: Can you make this per-customer?  Put it in validation.Limits.

pkg/distributor/ha_tracker.go
tomwilkie: Lets leave as `replica`, and make this configurable per user.

pkg/ring/consul_client.go
tomwilkie:
```go
c.consul.WatchPrefix(ctx, fmt.Sprintf("%s/%s", c.prefix, prefix), func(k string, i interface{}) bool {
  return f(strings.TrimPrefix(key, c,prefix), i)
})
```

pkg/ring/kvstore.go
tomwilkie: Right I get it now thanks.

pkg/distributor/distributor.go
cstyan: Did you want to keep this config flag in addition to the per user "accept ha samples" flag

pkg/distributor/distributor.go
```diff
@@ -141,6 +147,11 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 	replicationFactor.Set(float64(ring.ReplicationFactor()))
 	cfg.PoolConfig.RemoteTimeout = cfg.RemoteTimeout
 
+	replicas, err := newClusterTracker(cfg.TrackerConfig)
+	if err != nil {
+		return nil, err
+	}
```
tomwilkie: Have we lost the ability for this to be disabled by default?

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,191 @@
...
+
+		if desc, ok := in.(*ReplicaDesc); ok {
+
+			// We don't need to CAS and update the timestamp in the KV store if the timestamp we've received
+			// this sample at is less than updateTimeout amount of time since the timestamp in the KV store.
+			if desc.Replica == replica && now.Sub(timestamp.Time(desc.ReceivedAt)) < c.cfg.UpdateTimeout {
+				return nil, false, nil
+			}
+
+			// We shouldn't failover to accepting to a new replica if the timestamp we've received this sample at
```
tomwilkie:
```suggestion
			// We shouldn't failover to accepting a new replica if the timestamp we've received this sample at
```

pkg/distributor/http_server.go
```diff
@@ -36,7 +36,9 @@ func (d *Distributor) PushHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if _, err := d.Push(r.Context(), &req.WriteRequest); err != nil {
-		level.Error(logger).Log("msg", "push error", "err", err)
+		if resp, ok := httpgrpc.HTTPResponseFromError(err); ok && resp.GetCode() != 202 {
+			level.Error(logger).Log("msg", "push error", "err", err)
+		}
```
tomwilkie: Can we combine these two `httpgrpc.HTTPResponseFromError(err)`s?  Eg

```golang
esp, ok := httpgrpc.HTTPResponseFromError(err)
if !ok {
  return
}

// 202 happens for rejected samples from HA pairs - its expected, so we don't want to spam the logs.
if resp.GetCode() != 202 {
  level.Error(logger).Log("msg", "push error", "err", err)
}

http.Error(w, string(httpResp.Body), int(httpResp.Code))
```

```

File: pkg/ring/consul_client.go [hunk 65]
```diff
@@ -110,7 +113,9 @@ func (c *consulClient) cas(ctx context.Context, key string, f CASCallback) error
 		if kvp != nil {
 			out, err := c.codec.Decode(kvp.Value)
 			if err != nil {
-				level.Error(util.Logger).Log("msg", "error decoding key", "key", key, "err", err)
+				if resp, ok := httpgrpc.HTTPResponseFromError(err); ok && resp.GetCode() != 202 {
+					level.Error(util.Logger).Log("msg", "error decoding key", "key", key, "err", err)
+				}
```
tomwilkie: I don't see how this if will ever not be true - none of the codecs return a httpgrpc error.

pkg/util/validation/limits.go
```diff
@@ -40,6 +43,9 @@ type Limits struct {
 func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&l.IngestionRate, "distributor.ingestion-rate-limit", 25000, "Per-user ingestion rate limit in samples per second.")
 	f.IntVar(&l.IngestionBurstSize, "distributor.ingestion-burst-size", 50000, "Per-user allowed ingestion burst size (in number of samples). Warning, very high limits will be reset every -distributor.limiter-reload-period.")
+	f.BoolVar(&l.AcceptHASamples, "distributor.accept-ha-samples", false, "Per-user flag to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup.")
+	f.StringVar(&l.HAReplicaLabel, "ha-tracker.replica", "__replica", "Prometheus label to look for in samples to identify a Proemtheus HA replica.")
+	f.StringVar(&l.HAClusterLabel, "ha-tracker.cluster", "__cluster", "Prometheus label to look for in samples to identify a Poemtheus HA cluster.")
```
tomwilkie: Sorry to rake this up again, but can you set this to `cluster` by default?

pkg/util/validation/limits.go
```diff
@@ -40,6 +43,9 @@ type Limits struct {
 func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&l.IngestionRate, "distributor.ingestion-rate-limit", 25000, "Per-user ingestion rate limit in samples per second.")
 	f.IntVar(&l.IngestionBurstSize, "distributor.ingestion-burst-size", 50000, "Per-user allowed ingestion burst size (in number of samples). Warning, very high limits will be reset every -distributor.limiter-reload-period.")
+	f.BoolVar(&l.AcceptHASamples, "distributor.accept-ha-samples", false, "Per-user flag to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup.")
+	f.StringVar(&l.HAReplicaLabel, "ha-tracker.replica", "__replica", "Prometheus label to look for in samples to identify a Proemtheus HA replica.")
```
tomwilkie: Please make this either `replica` or `__replica__`.

tomwilkie:
Thats the final round of feedback!  LGTM once this is addressed.

pkg/ring/consul_client.go
cstyan: Yeah I think I accidentally added this here but meant to add the check in the http code, which I did 5 minutes later. I'll remove this and deploy all the changes from the review. If there isn't logging here for 202's then I'd mistakenly added this check :)

pkg/distributor/distributor.go
```diff
@@ -96,8 +100,10 @@ type Distributor struct {
 // create a Distributor
 type Config struct {
 	EnableBilling bool
+	HAReplicas    bool
```
gouthamve: Can you make it `EnableHAReplicas` inline with the other config options?

pkg/distributor/distributor.go
```diff
@@ -96,8 +100,10 @@ type Distributor struct {
...
 	BillingConfig billing.Config
 	PoolConfig    ingester_client.PoolConfig
+	TrackerConfig HATrackerConfig
```
gouthamve: `TrackerConfig` --> `HATrackerConfig`

pkg/ring/consul_client.go
```diff
@@ -203,6 +210,54 @@ func (c *consulClient) WatchKey(ctx context.Context, key string, f func(interfac
...
 }
 
+// WatchPrefix will watch a given prefix in Consul for new keys and changes to existing keys under that prefix.
+// When the value under said key changes, the f callback is called with the deserialised value.
+// To construct the deserialised value, a factory function should be supplied which generates an empty
+// struct for WatchKey to deserialise into. Note that for WatchPrefix it's a struct containing the Consul key and
+// value KVPair that is passed to the callback f. See the KVEntry struct in kvstore.go. Values in Consul are assumed
+// to be JSON. This function blocks until the context is cancelled.
+func (c *consulClient) WatchPrefix(ctx context.Context, prefix string, f func(string, interface{}) bool) {
+
```
gouthamve: Extra whitespace.

pkg/ring/consul_client.go
```diff
@@ -203,6 +210,54 @@ func (c *consulClient) WatchKey(ctx context.Context, key string, f func(interfac
 	}
 }
 
+// WatchPrefix will watch a given prefix in Consul for new keys and changes to existing keys under that prefix.
+// When the value under said key changes, the f callback is called with the deserialised value.
+// To construct the deserialised value, a factory function should be supplied which generates an empty
```
gouthamve: > To construct the deserialised value, a factory function should be supplied which generates an empty struct for WatchKey to deserialise into.

This statement can be omitted. It's not passed in this function and only stands to confuse the user.

pkg/distributor/ha_tracker.go
```diff
@@ -0,0 +1,191 @@
...
+
+	if ok && entry.Replica == replica && now.Sub(timestamp.Time(entry.ReceivedAt)) < c.cfg.UpdateTimeout {
+		return nil
+	}
+	return c.checkKVStore(ctx, key, replica, now)
+}
+
+func (c *haTracker) checkKVStore(ctx context.Context, key, replica string, now time.Time) error {
+	return c.client.CAS(ctx, key, func(in interface{}) (out interface{}, retry bool, err error) {
+
```
gouthamve: Extra whitespace.

tomwilkie:
I've rebased & squashed, and consolidated the commit message into something more long-form.

PR #1228, opened 2019-02-08
Factor out the batch write logic from the distributor to the ring package.

tomwilkie:
Mainly as I want to reuse it in Loki.  But I also like the refactoring and separation of concerns, makes it easier to see whats going on IMO.


tomwilkie:
Thanks folks!  

> LGTM - I would have thought the background context stuff should also move.

Thats a cortex opinion - I'm not even convinced its the right thing.  But thats for another day...

PR #1231, opened 2019-02-12
Store series label-value sets as strings rather than slices

bboreham:
This should reduce the memory size of ingesters a bit: slices are 24 bytes plus contents while strings are 16 bytes.

It might not seem like much but there are typically 40 of them per series, so 300MB saved per million series.

This change does put more work onto the query side, where we will allocate byte slices in the process of serialising the answers.  Also when transferring chunks from a leaving to a joining ingester.

### DIFF
--- a/pkg/ingester/client/compat.go
+++ b/pkg/ingester/client/compat.go
@@ -248,6 +248,18 @@ func FromLabelPairsToLabels(labelPairs []LabelPair) labels.Labels {
 	return ls
 }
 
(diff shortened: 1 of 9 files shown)


tomwilkie:
Sorry for the slow review on this @bboreham!  Will give it a thorough review tomorrow.

tomwilkie:
Sorry this took so long @bboreham. I was a little surprised to see this could be done without any changes to the write path! LGTM

PR #1233, opened 2019-02-20
Don't block lock in forSeriesMatching callback

codesome:
`stream.Send` here https://github.com/cortexproject/cortex/blob/05b35bbf0d9e545c58de0acce47b94983d567fc1/pkg/ingester/ingester.go#L411-L413 is time consuming, in-turn blocking the lock in `forSeriesMatching`. 

This PR breaks the callbacks into 2 callbacks, where second callback is called in intervals outside the critical sections of the lock. 

- [ ] Unit tests

/cc @tomwilkie

### DIFF
--- a/pkg/ingester/user_state.go
+++ b/pkg/ingester/user_state.go
@@ -257,13 +257,18 @@ func (u *userState) removeSeries(fp model.Fingerprint, metric labelPairs) {
 	memSeries.Dec()
 }
 
-// forSeriesMatching passes all series matching the given matchers to the provided callback.
-// Deals with locking and the quirks of zero-length matcher values.
-// There are 2 callbacks. callback1 is called for each series, where the lock is held.
-// callback2 is called at certain intervals specified by callback2Interval,
-// i.e. callback2() after callback2Interval calls of callback1().
-func (u *userState) forSeriesMatching(ctx context.Context, allMatchers []*labels.Matcher, append func(context.Context, model.Fingerprint, *memorySeries) error,
-	send func(context.Context) error, batchSize int) error {
+// forSeriesMatching passes all series matching the given matchers to the
+// provided callback. Deals with locking and the quirks of zero-length matcher
+// values. There are 2 callbacks:
+// - The `add` callback is called for each series while the lock is held, and
+//   is intend to be used by the caller to build a batch.
+// - The `send` callback is called at certain intervals specified by batchSize
+//   with no locks held, and is intended to be used by the caller to send the
+//   built batches.
+func (u *userState) forSeriesMatching(ctx context.Context, allMatchers []*labels.Matcher,
+	add func(context.Context, model.Fingerprint, *memorySeries) error,
+	send func(context.Context) error, batchSize int,
+) error {
 	log, ctx := spanlogger.New(ctx, "forSeriesMatching")
 	defer log.Finish()
 
@@ -296,7 +301,7 @@ outer:
 			}
 		}
 
-		err := append(ctx, fp, series)
+		err := add(ctx, fp, series)
 		u.fpLocker.Unlock(fp)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


pkg/ingester/user_state.go
```diff
@@ -259,7 +259,11 @@ func (u *userState) removeSeries(fp model.Fingerprint, metric labelPairs) {
 
 // forSeriesMatching passes all series matching the given matchers to the provided callback.
 // Deals with locking and the quirks of zero-length matcher values.
-func (u *userState) forSeriesMatching(ctx context.Context, allMatchers []*labels.Matcher, callback func(context.Context, model.Fingerprint, *memorySeries) error) error {
+// There are 2 callbacks. callback1 is called for each series, where the lock is held.
+// callback2 is called at certain intervals specified by callback2Interval,
+// i.e. callback2() after callback2Interval calls of callback1().
+func (u *userState) forSeriesMatching(ctx context.Context, allMatchers []*labels.Matcher, callback1 func(context.Context, model.Fingerprint, *memorySeries) error,
+	callback2 func(context.Context) error, callback2Interval int) error {
```
tomwilkie: Lets have better names for the callbacks - `append` and `send` would work for me.

pkg/ingester/user_state.go
```diff
@@ -271,6 +275,10 @@ func (u *userState) forSeriesMatching(ctx context.Context, allMatchers []*labels
 
 	level.Debug(log).Log("series", len(fps))
 
+	iter := 0
+	if callback2Interval <= 0 {
+		callback2Interval = 1
+	}
```
tomwilkie: This should be called `batchSize` IMO.

pkg/ingester/user_state.go
```diff
@@ -271,6 +275,10 @@ func (u *userState) forSeriesMatching(ctx context.Context, allMatchers []*labels
...
 	// fps is sorted, lock them in order to prevent deadlocks
 outer:
 	for _, fp := range fps {
```
tomwilkie: Instead of `iter`, you could use `i` from this loop and do `i % batchSize == 0` for the test.

pkg/ingester/user_state.go
```diff
@@ -292,11 +300,19 @@ outer:
...
 			return err
 		}
+
+		if iter >= callback2Interval && callback2 != nil {
+			if err = callback2(ctx); err != nil {
+				return nil
+			}
+			iter = 0
+		}
 	}
```
tomwilkie: Need to call `callback2` again to catch the last batch.

tomwilkie:
Looks good @codesome! Left some feedback.  I'm actually going to deploy and test this, so might do the feedback myself if thats okay.

codesome:
@tomwilkie I'll be unavailable for ~35h from now (got a long flight ahead), so please go ahead!

codesome:
PS: As this function was used elsewhere too where there was no notion of batch, send etc, I tried to keep the names more general.

pkg/ingester/user_state.go
```diff
@@ -292,12 +296,21 @@ outer:
 			}
 		}
 
-		err := callback(ctx, fp, series)
+		err := append(ctx, fp, series)
```
bboreham: `append` is a well-known function in Go; I think it's too confusing to re-use the name here

pkg/ingester/user_state.go
```diff
@@ -259,7 +259,11 @@ func (u *userState) removeSeries(fp model.Fingerprint, metric labelPairs) {
 
 // forSeriesMatching passes all series matching the given matchers to the provided callback.
 // Deals with locking and the quirks of zero-length matcher values.
-func (u *userState) forSeriesMatching(ctx context.Context, allMatchers []*labels.Matcher, callback func(context.Context, model.Fingerprint, *memorySeries) error) error {
+// There are 2 callbacks. callback1 is called for each series, where the lock is held.
+// callback2 is called at certain intervals specified by callback2Interval,
+// i.e. callback2() after callback2Interval calls of callback1().
```
bboreham: names of callbacks need to match the code; also I would be explicit that the lock is not held while the second callback is called

PR #1237, opened 2019-02-22
adding alert templates support for AlertManager

rsteneteg:
Signed-off-by: Roger Steneteg <rsteneteg@ea.com>

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -149,18 +149,20 @@ func New(cfg *Config) (*Alertmanager, error) {
 }
 
 // ApplyConfig applies a new configuration to an Alertmanager.
-func (am *Alertmanager) ApplyConfig(conf *config.Config) error {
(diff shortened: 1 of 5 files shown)


rsteneteg:
This PR adds the ability to add reusable templates in the Alert Managers notification output to the receivers.  As described here https://prometheus.io/docs/alerting/notification_examples/#defining-reusable-templates

This can be used to avoid repeating alert templates in the configuration in the alertmanager config, or moving long multiline alert templates into separate files to make the alert manager configuration file leaner

bboreham:
Just looked this over again - it actually writes out disk files for the templates?
That's a bit different from what we do for other ruler stuff, but I guess it's workable because the contents of the files are rewritten any time the config is reloaded.

rsteneteg:
@bboreham yes, the template files are written to disk, just as silences are. The alertmanager uses this Prometheus package to load templates from files on disk "github.com/prometheus/alertmanager/template".

And as you said, they are rewritten and reloaded each time the config changes.

PR #1239, opened 2019-02-22
Optionally prepend a hash of row key to improve distribution.

tomwilkie:

### DIFF
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -259,50 +259,6 @@ func (s *storageClientColumnKey) QueryPages(ctx context.Context, queries []chunk
 	return lastErr
 }
 
-func (s *storageClientColumnKey) query(ctx context.Context, query chunk.IndexQuery, callback func(result chunk.ReadBatch) (shouldContinue bool)) error {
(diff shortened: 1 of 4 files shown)


pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -247,50 +259,6 @@ func (s *storageClientColumnKey) QueryPages(ctx context.Context, queries []chunk
 	return lastErr
 }
 
-func (s *storageClientColumnKey) query(ctx context.Context, query chunk.IndexQuery, callback func(result chunk.ReadBatch) (shouldContinue bool)) error {
```
bboreham: How does removal of this `query()` function relate to the rest of the commit?

gouthamve: Doesn't. But that codepath is not being called. Probably should have been another PR.

gouthamve:
LGTM to me!

tomwilkie:
Got this deployed out to our big test cluster.  Before the change:

screenshot 2019-02-26 at 07 34 29

After the change:

screenshot 2019-02-26 at 07 34 23


And 90th percentile latency of bigtable reads on this cluster has gone from 129ms to 60ms.  No change in 99th, but I suspect thats limited by big rows.

pkg/chunk/gcp/bigtable_index_client.go
bboreham: As a separate commit it might be acceptable, but surely having all changes in one commit consistent and described by the commit message is a very basic standard to set?

PR #1240, opened 2019-02-23
Add some more logging to the index cache, 

tomwilkie:
And make the metric count the number of gets correctly so you can calculate hit rate.

### DIFF
--- a/pkg/chunk/storage/caching_index_client.go
+++ b/pkg/chunk/storage/caching_index_client.go
@@ -9,6 +9,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/chunk/cache"
 	chunk_util "github.com/cortexproject/cortex/pkg/chunk/util"
 	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 	"github.com/go-kit/kit/log/level"
 	proto "github.com/golang/protobuf/proto"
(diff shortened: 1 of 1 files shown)


pkg/chunk/storage/caching_index_client.go
```diff
@@ -233,14 +237,15 @@ func (s *cachingIndexClient) cacheFetch(ctx context.Context, keys []string) (bat
...
-			level.Warn(util.Logger).Log("msg", "error unmarshalling index entry from cache", "err", err)
+			level.Warn(log).Log("msg", "error unmarshalling index entry from cache", "err", err)
 			cacheCorruptErrs.Inc()
 			continue
 		}
 
 		// Make sure the hash(key) is not a collision in the cache by looking at the
 		// key in the value.
 		if key != readBatch.Key || (readBatch.Expiry != 0 && time.Now().After(time.Unix(0, readBatch.Expiry))) {
+			level.Warn(log).Log("msg", "collision/expiration in entry cache", "key", key, "readBatch.Key", readBatch.Key, "expiry", time.Unix(0, readBatch.Expiry))
```
bboreham: Why is this a warning?
Can we have the message be more actionable for the operator who reads it?

tomwilkie: I've split these two reasons into separate ifs & logs, and made it debug level.  WDYT?

PR #1242, opened 2019-02-24
v10 schema builds on v9 but shards index rows to break up very big rows.

tomwilkie:
We have some very-high-cardinality series; 300k+ series per metric.  This results in bigtable rows that are over 100MB in size.  This change splits all rows into 16 rows, distributing entries based on the series ID.  Each read becomes 16 reads, done in parallel.


tomwilkie:
We deployed this and saw ~50% decrease in 99%-ile on bigtable queries.  So FTW!

PR #1247, opened 2019-02-26
add jsonnet templates for deployment

agrahamlincoln:
This change adds jsonnet templates for configuring cortex. These
templates are intended to be extensible and easily configurable for
vendoring into a repo and layering any tweaks on top.

There are two entrypoints to the templates
- cortex.jsonnet - provides all cortex components
- demo.jsonnet - provides all cortex deployments and the basic
configuration of prometheus & nginx for testing and demo purposes

The demo includes a prometheus which is configured to scrape all cortex
components and leverage remote storage into the cortex deployment.

Signed-off-by: Graham Rounds <graham@platform9.com>

---

I was unsure of what we want to do with `k8s/` once this is merged. the way I see it, we can generate static manifests from these templates and drop them in there, or we can remove the directory altogether. Thoughts?


bboreham:
Speaking personally, jsonnet is incomprehensible to me so if you remove the YAMLs I will not be able to comment on the contents.

khaines:
I'm in the same boat as @bboreham. Jsonnet just isn't something I'm familiar with, but I know @tomwilkie is a big fan :). Hopefully, he can have a look at this.

I don't see anything wrong here, but I can't confirm if it, in fact, is correct either.

deploy/images.libsonnet
```diff
@@ -0,0 +1,15 @@
+// last updated: 1/28/2019
+{
+    _images+:: {
+        alertmanager: "quay.io/cortexproject/alertmanager:master-0ef1ce95",
```
csmarchbanks: Would it be possible to pull the image hashes from `_config`?

agrahamlincoln:
> Would it be possible to add memcached as well?

Yeah, memcached is absolutely necessary. We layer this on top now, but I think it makes sense to include with the base templates as well.

deploy/images.libsonnet
agrahamlincoln: sure. though there is something nice about keeping images all in one area, so you can glance at all the various versions running for other services such as memcached, postgres, consul, etc.

I guess we could do `$._config.images.alertmanager` instead of putting the image field within `$._config.alertmanager.image` for example. What do you think?

nicklan: You should make the master branch tag a local var so it doesn't have be changed in so many places to update everything at once

deploy/lib/ingester.libsonnet
```diff
@@ -0,0 +1,99 @@
...
+{
+    ingester_deployment:
+        local name = $._config.ingester.name;
+
+        # Arguments
+        local extraArgs = $._config.ingester.extraArgs;
+        local consul_uri = $._config.consul.name + '.' + $._config.namespace + '.svc.cluster.local:8500';
+        local args = [
+            '-ingester.join-after=30s',
+            '-ingester.claim-on-rollout=false',
```
nicklan: this should be true

deploy/lib/ingester.libsonnet
```diff
@@ -0,0 +1,99 @@
...
+        # Pod
+        local ingesterPod = kube.PodSpec + {
+            containers_: {
+                ingester: ingesterContainer,
+            },
+            # Give ingesters 40 minutes to flush chunks and exit cleanly
+            # They will exit early when completely flushed
+            terminationGracePeriodSeconds: 2400,
+            volumes_: {
+                config_volume: schemaConfigVolume
```
nicklan: nit: include trailing comma

deploy/lib/querier.libsonnet
```diff
@@ -0,0 +1,89 @@
...
+            '-querier.batch-iterators=true',
+            '-querier.ingester-streaming=true',
+            '-config-yaml=/etc/cortex/schemaConfig.yaml',
+            '-consul.hostname=' + consul_uri,
+        ];
+        
+        # Environment Variables
+        local env = $._config.ingester.env;
+        local envKVMixin = $._config.querier.envKVMixin;
+        local extraEnv = [{name: key, value: extraEnv[key]} for key in std.objectFields(envKVMixin)];
```
nicklan: should be: `local extraEnv = [{name: key, value: envKVMixin[key]} for key in std.objectFields(envKVMixin)];`

deploy/lib/querier.libsonnet
```diff
@@ -0,0 +1,89 @@
...
+            resources+: $._config.querier.resources,
+        };
+
+        # Pod
+        local querierPod = kube.PodSpec + {
+            containers_: {
+                querier: querierContainer,
+            },
+            volumes_: {
+                config_volume: schemaConfigVolume
```
nicklan: nit: include trailing comma

deploy/lib/ruler.libsonnet
```diff
@@ -0,0 +1,86 @@
...
+            '-config-yaml=/etc/cortex/schemaConfig.yaml',
+            '-consul.hostname=' + consul_uri,
+            '-ruler.alertmanager-url=' + alertmanager_uri,
+            '-ruler.configs.url=' + configs_uri,
+        ];
+
+        # Environment Variables
+        local env = $._config.ruler.env;
+        local envKVMixin = $._config.ruler.envKVMixin;
+        local extraEnv = [{name: key, value: extraEnv[key]} for key in std.objectFields(envKVMixin)];
```
nicklan: should be: `local extraEnv = [{name: key, value: envKVMixin[key]} for key in std.objectFields(envKVMixin)];`

deploy/lib/table-manager.libsonnet
```diff
@@ -0,0 +1,73 @@
...
+        # Arguments
+        local extraArgs = $._config.tableManager.extraArgs;
+        local args = [
+            '-config-yaml=/etc/cortex/schemaConfig.yaml',
+        ];
+
+        # Environment Variables
+        local env = $._config.tableManager.env;
+        local envKVMixin = $._config.tableManager.envKVMixin;
+        local extraEnv = [{name: key, value: extraEnv[key]} for key in std.objectFields(envKVMixin)];
```
nicklan: should be: `local extraEnv = [{name: key, value: envKVMixin[key]} for key in std.objectFields(envKVMixin)];`

deploy/demo.jsonnet
```diff
@@ -0,0 +1,53 @@
+(import 'cortex.jsonnet') +
+(import 'lib/dynamodb.libsonnet') +
+(import 'lib/prometheus.libsonnet') +
+(import 'lib/nginx.libsonnet') + 
+{
+    _config+:: {
+        local dynamodb_uri = 'dynamodb://user:pass@' + $._config.dynamodb.name + '.' + $._config.namespace + '.svc.cluster.local:8000',
```
nicklan: this way of passing credentials fails for me.  I have to set them via `envKVMixin`

deploy/images.libsonnet
agrahamlincoln: These templates need to be updated to use the single binary. When this is done, I think we can easily roll this up into `_config` since there will only be one image

deploy/lib/ingester.libsonnet
agrahamlincoln: thanks for this! We've been running our cortex this way and it's been OK, but I agree that setting this to `true` would be better for production use.

deploy/demo.jsonnet
agrahamlincoln: I'm interested to hear why this didn't work. I'm able to build the demo without any problems with it setup this way. What are you seeing when you try to build this?

agrahamlincoln:
Apologies for all of the force pushing, needed to shake out the signed commit messages, squash a large number of commits, and update the repo to the latest `master` branch.

I believe I've addressed all reviewer concerns, and at the same time updated the templates to match the current cortex images.

khaines:
Is the intention to merge these templates in this repo are should the path be to do something similar to https://github.com/cortexproject/cortex-helm-chart?

agrahamlincoln:
I'm closing this out in favor of https://github.com/grafana/cortex-jsonnet/

PR #1250, opened 2019-02-28
Limit parallelism in a single query to 14 sub queries.

### LINKED ISSUE #1148: Long query handling
The query-frontend slices big queries up into days, and then runs them in parallel.  It also limits the number of queries a user can have queued to 100.  

The problem is we therefore can't run queries longer than 100days right now.  And very long queries convoy shorter ones.

Ideas:
- make it only run a specific number (14?) in parallel.  This will make queries longer than 14 days slower, but at least they will return.  
- increase the queue length.  Probably always want some fixed upper limit.  Doesn't help with convoying.
- make the queue a priority queue, where the priority is inverse of the number of shards a query is split into - long queries will never block short ones.

Probably don't want to have backoff/retry putting stuff on the queue - thats what the queue is for, and it will make it hard to understand the behaviour.  Client (grafana) will retry.

WDYT?

tomwilkie:
There is a limit of 100 outstanding queries at once per customer.  Due to day splitting, this makes queries longer than 100 days fail.
This PR limits the number of 'sub queries' per query that can be executed in parallel to 14.

Fixes #1148

### DIFF
--- a/pkg/querier/frontend/results_cache.go
+++ b/pkg/querier/frontend/results_cache.go
@@ -110,7 +110,7 @@ func (s resultsCache) handleHit(ctx context.Context, r *QueryRangeRequest, exten
 		return response, nil, err
 	}
 
-	reqResps, err = doRequests(ctx, s.next, requests)
+	reqResps, err = doRequests(ctx, s.next, requests, len(requests))
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/querier/frontend/results_cache.go
+++ b/pkg/querier/frontend/results_cache.go
@@ -110,7 +110,7 @@ func (s resultsCache) handleHit(ctx context.Context, r *QueryRangeRequest, exten
 		return response, nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/querier/frontend/results_cache.go
```diff
@@ -110,7 +110,7 @@ func (s resultsCache) handleHit(ctx context.Context, r *QueryRangeRequest, exten
 		return response, nil, err
 	}
 
-	reqResps, err = doRequests(ctx, s.next, requests)
+	reqResps, err = doRequests(ctx, s.next, requests, len(requests))
```
csmarchbanks: Why is this `len(requests)` instead of `maxParallelism`? If it can be `maxParallelism` then I think the function signature will not even need to change.

tomwilkie: In this case I thought it would be easier to ensure the requests would not have to wait, but I guess we can simplify.

csmarchbanks: :+1:

Issue #1249, opened 2019-02-28
Index Tables constantly being created if GCS storage is used.


jtlisi:
Table manager keeps trying to recreate tables that already exist. Can be an issue with certain schemas that have the same table.

```
...
level=info ts=2019-03-01T18:48:12.674935659Z caller=table_manager.go:354 msg="creating table" table=cortex-ops_index_2564
level=info ts=2019-03-01T18:48:13.102494276Z caller=table_manager.go:354 msg="creating table" table=cortex-ops_index_2565
level=info ts=2019-03-01T18:53:12.677863813Z caller=table_manager.go:354 msg="creating table" table=cortex-ops_index_2564
level=info ts=2019-03-01T18:53:13.104331145Z caller=table_manager.go:354 msg="creating table" table=cortex-ops_index_2565
level=info ts=2019-03-01T18:58:12.676143101Z caller=table_manager.go:354 msg="creating table" table=cortex-ops_index_2564
level=info ts=2019-03-01T18:58:13.103416555Z caller=table_manager.go:354 msg="creating table" table=cortex-ops_index_2565
...
```

Causes an error in bigtable when the family already exists.

PR #1253, opened 2019-03-01
Handle schema transition which don't align to periodic table transistions.

tomwilkie:
Schema transitions can happen on any day; table transitions happen weekly.  When they don't align, the table manager will try and create table twice.

This PR fixes the logic to handle these duplicates.


tomwilkie:
Part of #1249

Issue #416, opened 2017-05-10
Make queries involving no metric name work


tomwilkie:
Currently queries involving no metric name (such as `count({__name__=~".+"}) by (__name__)`) almost always fail with some think like:

```
Error: RequestError: send request failed caused by: Get https://weaveworks-dev-chunks.s3.amazonaws.com/2/1a9d5d42bc35a818%3A15bf13d7a4b%3A15bf16ca09b%3Afa6f6c42: dial tcp 52.216.0.88:443: connect: cannot assign requested address
```

I suspect this is because we're trying to fetch too many chunks simultaneously, and we're running out of client ports.

Also, the v7 schema isn't rolled out to prod yet.

And they are hella slow - but lets get them working first!

aaron7:
https://docs.google.com/a/weave.works/document/d/1dIpy2XAB9seML5JjCGyauBPBy_86dig9Vef1k8rpBOM/edit?usp=sharing

tomwilkie:
Design doc: https://docs.google.com/document/d/1dIpy2XAB9seML5JjCGyauBPBy_86dig9Vef1k8rpBOM/edit#

bboreham:
Current status: it's rolled out but unusably slow

tomwilkie:
There was a ‘fix’ in Prometheus, but it turns out it was incorrect. Whole
thing needs a rethink.

bboreham:
Rolled back from v8 to v6 today; write throughput seems much better.
cc @khaines

bboreham:
Tried the query `count({__name__=~".+"}) by (__name__)` on Weave Cloud today; got:

```
Error: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (215173530 vs. 67108864)
```

On a very small instance I do get results.

tomwilkie:
#958 deletes the schemas that support this.

With the schema store, there are 12x less index entries.  We might be able to consider removing metric name from the hash key, and having a proper `__name__` index row, allowing us to do queries withing metric names.   This would get hit on every write, so we'd need #607 to not hot spot, and it would be included in every read, so we'd need #947 to not hotspot there.  

Even then, with millions of series, this row might be too big...

gouthamve:
I have a design doc here to make this happen: https://docs.google.com/document/d/19gnA8vHKSnqhnly6hA2W8Hwhiib15toFGmD4nM582-s/edit#

pracucci:
[...]

PR #1255, opened 2019-03-02
Don't block queries with no metric name if no-op

bboreham:
Queries within the ingester retention time are satisfied without going to the chunk store, so we shouldn't reject ones that the chunk store can't implement.

In practical terms, move the "query must contain metric name" check after the time checks, so that we may return with `shortcut=true` even on those queries.  The ingester query code is ok with queries that don't have a metric name.

Partial fix for #416


tomwilkie:
LGTM!

Issue #1213, opened 2019-01-31
Make ruler scalable


bboreham:
Ruler is coded as a singleton; at some point you run out of power on one machine to run all the rules.

Instead, run them in more than one process.
Related: #310

Issue #310, opened 2017-03-01
Make ruler HA


jml:
Currently the ruler is not HA. There's only one instance that runs all the rules. If it goes down, rules are not being processed.

jml:
Current plan:

- store state about how long condition has been firing for somewhere else (probably dynamodb on weaveworks deployment)
- have rule evaluator that gets sent a rule (group?) and then evaluates it, storing state about firing conditions in thing above, and 
- have single scheduler that sends requests for rule evaluation to evaluators, relying on k8s to do load balancing for us

We'll sort out the config mess later.

Not personally going to do any of this any time soon, although I would personally like to do it. I also think this should wait until #555 is done.

jml:
https://github.com/prometheus/prometheus/issues/422 is relevant here. AFAICT, much of the discussion on the ticket assumes that Prometheus's metric store is the only acceptable place to store 'for' state.

Cortex doesn't have that constraint, but ideally the work would be done such that the storage is behind a golang interface, such that the current in-memory approach and a hypothetical TSDB-backed approach would both work, in addition to whatever dynamodb / bigtable thing we do.

jml:
Lots of this came up from a conversation I had w/ @juliusv a while back. The good ideas are probably his, the bad ones mine.

juliusv:
I don't think this is a concern in terms of ruler statefulness. We get all the samples of a rule evaluation at the same time anyway, so we may as well send them as one request instead of N. Nothing we need to persist over a long time. I'm already doing that for my Prometheus 2.0 port, where `Add()` just adds samples to the appender internally and `Commit()` sends them out.

juliusv:
Ah damn, commented on the wrong issue. Will delete here and repost on the right one.

tomwilkie:
Fixed in #1258. Ruler is now horizontally scalable, and rules are sharded using the ring.  I believe this should be enough.

PR #1258, opened 2019-03-05
Horizontally Scalable Ruler

jtlisi:
This PR provides a flag that allows the ruler to shard ownership of which rule groups to evaluate by re-using the ring code. When enabled each ruler will act as a Lifecycler and interact with consul to create a ring.

## TO DO
- [x] Add tests for the ruler ring
- [x] Add better instrumentation for monitoring the distributed ruler

### DIFF
--- a/pkg/ring/lifecycler.go
+++ b/pkg/ring/lifecycler.go
@@ -18,23 +18,23 @@ import (
 )
 
 var (
-	consulHeartbeats = promauto.NewCounter(prometheus.CounterOpts{
-		Name: "cortex_ingester_consul_heartbeats_total",
+	consulHeartbeats = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "cortex_member_consul_heartbeats_total",
 		Help: "The total number of heartbeats sent to consul.",
-	})
-	tokensOwned = promauto.NewGauge(prometheus.GaugeOpts{
-		Name: "cortex_ingester_ring_tokens_owned",
+	}, []string{"name"})
+	tokensOwned = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "cortex_member_ring_tokens_owned",
 		Help: "The number of tokens owned in the ring.",
-	})
-	tokensToOwn = promauto.NewGauge(prometheus.GaugeOpts{
-		Name: "cortex_ingester_ring_tokens_to_own",
+	}, []string{"name"})
+	tokensToOwn = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "cortex_member_ring_tokens_to_own",
 		Help: "The number of tokens to own in the ring.",
-	})
+	}, []string{"name"})
 	shutdownDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Name:    "cortex_shutdown_duration_seconds",
 		Help:    "Duration (in seconds) of cortex shutdown procedure (ie transfer or flush).",
 		Buckets: prometheus.ExponentialBuckets(10, 2, 8), // Biggest bucket is 10*2^(9-1) = 2560, or 42 mins.
-	}, []string{"op", "status"})
+	}, []string{"op", "status", "name"})
 )
 
 // LifecyclerConfig is the config to build a Lifecycler.
@@ -61,15 +61,20 @@ type LifecyclerConfig struct {
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *LifecyclerConfig) RegisterFlags(f *flag.FlagSet) {
-	cfg.RingConfig.RegisterFlags(f)
+	cfg.RegisterFlagsWithPrefix("", f)
+}
 
-	f.IntVar(&cfg.NumTokens, "ingester.num-tokens", 128, "Number of tokens for each ingester.")
-	f.DurationVar(&cfg.HeartbeatPeriod, "ingester.heartbeat-period", 5*time.Second, "Period at which to heartbeat to consul.")
-	f.DurationVar(&cfg.JoinAfter, "ingester.join-after", 0*time.Second, "Period to wait for a claim from another ingester; will join automatically after this.")
-	f.DurationVar(&cfg.MinReadyDuration, "ingester.min-ready-duration", 1*time.Minute, "Minimum duration to wait before becoming ready. This is to work around race conditions with ingesters exiting and updating the ring.")
-	f.BoolVar(&cfg.ClaimOnRollout, "ingester.claim-on-rollout", false, "Send chunks to PENDING ingesters on exit.")
-	f.BoolVar(&cfg.NormaliseTokens, "ingester.normalise-tokens", false, "Store tokens in a normalised fashion to reduce allocations.")
-	f.DurationVar(&cfg.FinalSleep, "ingester.final-sleep", 30*time.Second, "Duration to sleep for before exiting, to ensure metrics are scraped.")
+// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
+func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	cfg.RingConfig.RegisterFlagsWithPrefix(prefix, f)
+
+	f.IntVar(&cfg.NumTokens, prefix+"num-tokens", 128, "Number of tokens for each ingester.")
+	f.DurationVar(&cfg.HeartbeatPeriod, prefix+"heartbeat-period", 5*time.Second, "Period at which to heartbeat to consul.")
+	f.DurationVar(&cfg.JoinAfter, prefix+"join-after", 0*time.Second, "Period to wait for a claim from another member; will join automatically after this.")
+	f.DurationVar(&cfg.MinReadyDuration, prefix+"min-ready-duration", 1*time.Minute, "Minimum duration to wait before becoming ready. This is to work around race conditions with ingesters exiting and updating the ring.")
+	f.BoolVar(&cfg.ClaimOnRollout, prefix+"claim-on-rollout", false, "Send chunks to PENDING ingesters on exit.")
+	f.BoolVar(&cfg.NormaliseTokens, prefix+"normalise-tokens", false, "Store tokens in a normalised fashion to reduce allocations.")
+	f.DurationVar(&cfg.FinalSleep, prefix+"final-sleep", 30*time.Second, "Duration to sleep for before exiting, to ensure metrics are scraped.")
 
 	hostname, err := os.Hostname()
 	if err != nil {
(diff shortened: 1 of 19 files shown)


khaines:
Thank you for working on this change! I'll be very happy to test this when it is ready :)

@rsteneteg FYI, since we've discussed this as a scaling contention point in the past

tomwilkie:
Does this also fix #310?

bboreham:
Fixes #1213.

cmd/ruler/main.go
```diff
@@ -59,28 +58,29 @@ func main() {
...
-	util.CheckFatal("initializing ruler", err)
-	defer rlr.Stop()
 
-	rulesAPI, err := ruler.NewRulesAPI(configStoreConfig)
-	util.CheckFatal("initializing rules API", err)
-
-	rulerServer, err := ruler.NewServer(rulerConfig, rlr, rulesAPI)
-	util.CheckFatal("initializing ruler server", err)
-	defer rulerServer.Stop()
+	rulerConfig.LifecyclerConfig = lifecyclerConfig
```
tomwilkie: Can we have the rulerConfig recursively register the lifecycler, so we don't have to do this?  This has been error prone in the past.

cmd/ruler/main.go
```diff
@@ -59,28 +58,29 @@ func main() {
 	util.CheckFatal("", err)
 	defer chunkStore.Stop()
 
-	r, err := ring.New(ringConfig)
+	r, err := ring.New(lifecyclerConfig.RingConfig)
```
tomwilkie: I'd like us to have two ringConfigs here - one for the ingesters and one for the rulers.  This will allow us to store them in different consuls in the future.  They should register their flags with different prefixes.

pkg/ring/lifecycler.go
```diff
@@ -19,15 +19,15 @@ import (
 
 var (
 	consulHeartbeats = promauto.NewCounter(prometheus.CounterOpts{
-		Name: "cortex_ingester_consul_heartbeats_total",
+		Name: "cortex_member_consul_heartbeats_total",
```
tomwilkie: We should add a label to these metrics so they can differentiate between different rings.

pkg/ring/lifecycler.go
```diff
@@ -97,7 +97,7 @@ type Lifecycler struct {
 
 	// These values are initialised at startup, and never change
 	ID   string
-	addr string
+	Addr string
```
tomwilkie: Why was this necessary?

pkg/ring/ring.go
```diff
@@ -100,30 +101,34 @@ func New(cfg Config) (*Ring, error) {
 		return nil, err
 	}
 
+	if cfg.Name == "" {
+		cfg.Name = "ingester"
+	}
```
tomwilkie: Lets not do this.  Also, I don't think the name is part of the config - I think they're defined by the code.  Name should be a string passed to the constructor, and should not be the empty string.

pkg/ring/ring.go
```diff
@@ -100,30 +101,34 @@ func New(cfg Config) (*Ring, error) {
...
 		done:     make(chan struct{}),
 		ringDesc: &Desc{},
-		ingesterOwnershipDesc: prometheus.NewDesc(
-			"cortex_ring_ingester_ownership_percent",
-			"The percent ownership of the ring by ingester",
-			[]string{"ingester"}, nil,
+		memberOwnershipDesc: prometheus.NewDesc(
+			"cortex_ring_member_ownership_percent",
+			"The percent ownership of the ring by member",
+			[]string{"member", "type"}, nil,
```
tomwilkie: s/type/name/?

pkg/ruler/lifecycle.go
```diff
@@ -0,0 +1,90 @@
...
+				time.Sleep(r.SearchPendingFor / pendingSearchIterations)
+				continue
+			} else {
+				level.Warn(util.Logger).Log("msg", "Could not find pending ruler before deadline", "err", err)
+				return nil, err
+			}
+		}
+		return ingester, nil
+	}
+}
```
tomwilkie: Is this actually used?  Does it need to exist?  I think transfer in the rulers case is a noop.

pkg/ruler/ring.go
```diff
@@ -0,0 +1,22 @@
...
+	consulCfg.Prefix = "rulers/"
+
+	return ring.Config{
+		Name:              "ruler", // Ensures separate ring metrics for ruler ring
+		Consul:            consulCfg,
+		Store:             cfg.Store,
+		HeartbeatTimeout:  cfg.HeartbeatTimeout,
+		ReplicationFactor: 1, // Ruler will never use a RF > 1
+		Mock:              cfg.Mock,
+	}
```
tomwilkie: Lets have the ring's registerflags func take a prefix so we can have multiple of them, and then lets use an independant one for the ruler.  What we've got here won't give us the operation flexibility.

pkg/ruler/ruler.go
```diff
@@ -116,92 +120,43 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
-func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d *distributor.Distributor) (*Ruler, error) {
+func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d *distributor.Distributor, rulesAPI RuleStore) (*Ruler, error) {
 	ncfg, err := buildNotifierConfig(&cfg)
 	if err != nil {
 		return nil, err
 	}
-	return &Ruler{
+
+	ruler := &Ruler{
+		cfg:           cfg,
```
tomwilkie: Now you're embedding the config (good), can you remove the fields that are just copied out of config?

pkg/ruler/ruler.go
```diff
@@ -210,71 +165,63 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
...
+			return nil, err
 		}
 	}
-	amConfig := &config.AlertmanagerConfig{
-		Scheme:                 u.Scheme,
-		PathPrefix:             u.Path,
-		Timeout:                model.Duration(rulerConfig.NotificationTimeout),
-		ServiceDiscoveryConfig: sdConfig,
+
+	if cfg.NumWorkers <= 0 {
```
tomwilkie: Can you put this kind of validation first pls?

pkg/ruler/ruler.go
```diff
@@ -210,71 +165,63 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
...
+	r.notifiersMtx.Unlock()
 
-	return promConfig, nil
+	r.scheduler.Stop()
+	for _, w := range r.workers {
+		w.Stop()
+	}
+
+	r.lifecycler.Shutdown()
+	r.ring.Stop()
```
tomwilkie: These will be nil if sharding isn't enabled, right?

pkg/ruler/ruler.go
```diff
@@ -365,6 +312,13 @@ func (r *Ruler) getOrCreateNotifier(userID string) (*notifier.Manager, error) {
 func (r *Ruler) Evaluate(userID string, item *workItem) {
 	ctx := user.InjectOrgID(context.Background(), userID)
 	logger := util.WithContext(ctx, util.Logger)
+	if r.cfg.EnableSharding {
+		owned := r.checkRule(item.hash)
+		if !owned {
```
tomwilkie: You can compress this into one if.  I don't like nested ifs:

```go
if r.cfg.EnableSharding && !r.checkRule(item.hash) {
...
```

pkg/ruler/ruler.go
```diff
@@ -384,109 +338,38 @@ func (r *Ruler) Evaluate(userID string, item *workItem) {
...
-		go w.Run()
-	}
-	level.Info(util.Logger).Log("msg", "ruler up and running")
-}
-
-// Stop the server.
-func (s *Server) Stop() {
-	for _, w := range s.workers {
-		w.Stop()
+func (r *Ruler) checkRule(hash uint32) bool {
```
tomwilkie: Might want to call this "ownsRule"?

pkg/ruler/ruler.go
```diff
@@ -384,109 +338,38 @@ func (r *Ruler) Evaluate(userID string, item *workItem) {
...
+	rlrs, err := r.ring.Get(hash, ring.Read)
+	// If an error occurs keep evaluating a rule as if it is owned
+	// better to have extra datapoints for a rule than none at all
+	if err != nil {
+		ringCheckErrors.Inc()
```
tomwilkie: Can you log here too pls?

pkg/ruler/scheduler.go
```diff
@@ -90,7 +96,7 @@ type userConfig struct {
 type groupFactory func(userID string, groupName string, rls []rules.Rule) (*group, error)
 
 type scheduler struct {
-	rulesAPI           RulesAPI
+	ruleStore          RuleStore
```
tomwilkie: For simplicity, might be best not to rename this type?  Feel free to do it in another PR.

pkg/ruler/scheduler.go
```diff
@@ -134,17 +139,17 @@ func (s *scheduler) Run() {
...
 			return
 		}
 	}
 }
 
 func (s *scheduler) Stop() {
-	close(s.stop)
+	close(s.done)
 	s.q.Close()
-	<-s.done
```
tomwilkie: This double check is to give the scheduler time to shutdown.  Why remove it?

pkg/ruler/scheduler.go
```diff
@@ -262,19 +268,29 @@ func (s *scheduler) addUserConfig(now time.Time, hasher hash.Hash64, generation
 				level.Warn(util.Logger).Log("msg", "scheduler: failed to create group for user", "user_id", userID, "group", group, "err", err)
 				return
 			}
-			workItems = append(workItems, workItem{userID, group, g, evalTime, generation})
+			ringHasher.Reset()
+			ringHasher.Write([]byte(userID + ":" + group))
```
tomwilkie: I'd just do:
```
ringHasher.Write([]byte(userID))
ringHasher.Write([]byte(group))
``

The separate doesn't add anything here.

File: pkg/ruler/worker.go [hunk 18]
```diff
@@ -0,0 +1,63 @@
...
+	Run()
+	Stop()
+}
+
+type worker struct {
+	scheduler *scheduler
+	ruler     *Ruler
+
+	done       chan struct{}
+	terminated chan struct{}
```
tomwilkie: `quit` & `done` are the names we tend to use.  lets be consistent.

tomwilkie:
This looks really good!  Lots of small nits and we need to chat about how we handle config on Cortex.  When you get this green & all the points address, give me a ping and we can do a final pass.

pkg/ruler/lifecycle.go
jtlisi: This basically just hangs the ruler til a new one joins the ring. A noop would probably work just as well.

pkg/ring/lifecycler.go
jtlisi: Addr is used by the ruler to verify ownership, I would rather use ID but that was not in the IngesterDesc.

pkg/ruler/scheduler.go
jtlisi: It's not needed since the scheduler is gracefully shut down later by the ruler during the flush stage of the lifecycle. It empties the scheduling queue as a noop currently and once the scheduling_queue closes it's next channel or a set timeout is hit, it will complete the flush and finish shutting down.

jtlisi:
@tomwilkie I think this is ready for another look

pkg/ruler/worker.go
```diff
@@ -0,0 +1,63 @@
...
+func (w *worker) Run() {
+	defer close(w.done)
+	for {
+		select {
+		case <-w.quit:
+			return
+		default:
+		}
+		waitStart := time.Now()
+		blockedWorkers.Inc()
```
gouthamve: I see these metrics (`blockedWorkers`, `evalLatency`, `workerIdleTime`) are defined in `ruler.go`, can you move them here?

jtlisi: 👍

jtlisi:
I rebased this on top of master to work with a single binary. This should be ready for another look @tomwilkie

tomwilkie:
LGTM! Thanks for you work on this Jacob.

bboreham:
I have edited the description to mention some breaking changes to metrics.

PR #1260, opened 2019-03-06
Check ingester has tokens before reporting ready

### LINKED ISSUE #1112: Add a check that we have tokens to the ingester ready handler
This would be done in lifecycler.go probably.

bboreham:
Fixes #1112 

Also state the reason why an ingester is not reporting ready - the reason will be visible if you curl the `/ready` endpoint. This is to remove some head-scratching when debugging.

### DIFF
--- a/pkg/ring/lifecycler.go
+++ b/pkg/ring/lifecycler.go
@@ -153,30 +153,35 @@ func NewLifecycler(cfg LifecyclerConfig, flushTransferer FlushTransferer) (*Life
 	return l, nil
 }
 
-// IsReady is used to rate limit the number of ingesters that can be coming or
+// CheckReady is used to rate limit the number of ingesters that can be coming or
 // going at any one time, by only returning true if all ingesters are active.
-func (i *Lifecycler) IsReady(ctx context.Context) bool {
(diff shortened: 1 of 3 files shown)


pkg/ring/lifecycler.go
```diff
@@ -153,30 +153,38 @@ func NewLifecycler(cfg LifecyclerConfig, flushTransferer FlushTransferer) (*Life
...
 	if i.ready {
-		return true
+		return nil
 	}
 
 	// Ingester always take at least minReadyDuration to become ready to work
 	// around race conditions with ingesters exiting and updating the ring
 	if time.Now().Sub(i.startTime) < i.cfg.MinReadyDuration {
-		return false
+		return fmt.Errorf("Not ready: waiting for %v after startup", i.cfg.MinReadyDuration)
```
tomwilkie: Generally considered that errors should be lowercase (https://github.com/golang/go/wiki/Errors).  The argument is that they'll be prefixed by the caller, so you might want to put the `"Not ready: %v"` bit in the HTTP handler?

bboreham: done

PR #1261, opened 2019-03-06
Report the stack-trace from errors at top level

bboreham:
Errors created with `github.com/pkg/errors` may have extra info such as a stack trace, but we need to ask for it specially.

Localise this code in one place in the `utils` package.

Sweet sweet -145 lines of code.


bboreham:
To be extra-clear, the substantive change is to add `Sprintf("%+v", err)`.

PR #1262, opened 2019-03-06
Easy to use Cortex: Single binary, single process

### LINKED ISSUE #1136: RFC: single binary cortex / make cortex easier to get started with
- I'd like to put Cortex back to a single binary; see #151 for the split into multiple binaries.
- You'd still be able to run as a ingester, distributor etc; that would be a flag.
- I want to introduce reading flags from yaml (see #1127); this would allow us to leave the flag defaults as is for backwards compat, whilst shipping a couple of yamls in the repo for getting started.
- I want to unify the dependancies of the various "components" into a single place, and programatically initialise them in the right order.  This should reduce the duplication in the various mains.
- it will ensure the lite version never breaks due to flag (although we have a test now, so maybe not an issue)

I've got a WIP prototype of this based on the code in #1127, which I'm trying to publish in time for kubecon.  WDYT?

tomwilkie:
Fixes #1136

Build a single binary Cortex which can be told to start as an ingester, distributor, querier etc.  Declare the various dependancies between modules and use code to start and stop them in the right order.  Support the loading of config from a YAML file, and ship a config that allows you to start as a single process.

This replaces all the previous binaries and `lite`, and avoids the problem of having two (or more) copies of the init code that can get out of sync.

The follow flags have been removed:
- `-ingester.max-concurrent-streams`: use `-server.grpc-max-concurrent-streams`
- `-query-frontend.max-recv-message-size-bytes`: use `-server.grpc-max-recv-msg-size-bytes`
- `-memcached.*` on the the frontend: use `-frontend.memcached.*`

And you need to add the following flag when running as a set of microservices:
- `-target={distributor,ingester,querier,query-frontend,table-manager}`

TODO:
- [x] Single binary
- [x] Figure out what to do about ruler and alertmanager.
- [x] Test.
- [x] Config for single process mode.
- [x] Add getting started guide.
- [x] Test it all still works as a bunch of microservices.


docs/getting_started.md
```diff
@@ -0,0 +1,80 @@
...
+$ docker run -p 3000:3000 grafana/grafana
+```
+
+In [the Grafana UI](http://localhost:3000) (username/password admin/admin), add a Prometheus datasource for Cortex (`http://host.docker.internal:9009/api/prom`).
+
+## Scale out
+
+Next we're going to show how you can run a scale out Cortex cluster using Docker.
+We'll first make a Docker network to put these containers on so they can resolve each other by name,
+run a single nodeSonsul to coordinate the cluster and build the Cortex image:
```
weeco: "nodeSonsul" typo

docs/getting_started.md
```diff
@@ -0,0 +1,80 @@
...
+Your Prometheus instance will now start pushing data to Cortex.  To query that data, start a Grafana instance:
+
+```sh
+$ docker run -p 3000:3000 grafana/grafana
+```
+
+In [the Grafana UI](http://localhost:3000) (username/password admin/admin), add a Prometheus datasource for Cortex (`http://host.docker.internal:9009/api/prom`).
+
+## Scale out
+
```
weeco: Maybe a little section about prerequisites to scale out cortex? It seems Cortex needs Consul for the cluster coordination. Does it need something else?

pkg/configs/client/client.go
```diff
@@ -0,0 +1,151 @@
...
+	"time"
+
+	"github.com/cortexproject/cortex/pkg/configs"
+	"github.com/cortexproject/cortex/pkg/configs/db"
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/go-kit/kit/log/level"
+)
+
+// Client is what the ruler and altermanger needs from a config store to process rules.
+type Client interface {
```
khaines: This interface looks to be missing a `GetTemplates` method for alert template files, which were added in PR #1237.

docs/getting_started.md
```diff
@@ -0,0 +1,80 @@
+# Getting Started
+
+Cortex can be runs as a single binary or as multiple independent microservices.
```
weeco: Can you be more precie about this? You could clarify the advantages/disadvantages of each approach. Right now I could only guess, that multiple independent microservices offer the advantage to be independently scalable.

Also what 3rd party dependencies do each of these approaches have? Does one always need memcache, consul & a timeseries DB such as Bigtable?

pkg/cortex/cortex.go
```diff
@@ -0,0 +1,239 @@
...
+	ruler        *ruler.Ruler
+	rulerServer  *ruler.Server
+	configAPI    *api.API
+	configDB     db.DB
+	alertmanager *alertmanager.MultitenantAlertmanager
+
+	httpAuthMiddleware middleware.Interface
+}
+
+// New makes a new Loki.
```
khaines: copy/paste typo.

docs/getting_started.md
weeco: Also what do you think about adding a link to https://github.com/cortexproject/cortex/blob/master/docs/architecture.md ? This gave me a good high level overview about how cortex works and thus helped me understanding it.

pkg/cortex/cortex.go
```diff
@@ -0,0 +1,239 @@
...
+	level.Info(util.Logger).Log("msg", "initialising", "module", m)
+	if modules[m].init != nil {
+		if err := modules[m].init(t); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("error initialising module: %s", m))
+		}
+	}
+	return nil
+}
+
+// Run starts Loki running, and blocks until a signal is received.
```
khaines: copy/paste typo

pkg/cortex/cortex.go
```diff
@@ -0,0 +1,239 @@
...
+func (t *Cortex) Run() error {
+	return t.server.Run()
+}
+
+// Stop gracefully stops a Loki.
```
khaines: copy/paste typo

docs/getting_started.md
tomwilkie: Done!  Thanks for the feedback.

docs/getting_started.md
tomwilkie: I've made the following sentence a list of requirements.

pkg/configs/client/client.go
tomwilkie: Hi @khaines - I don't understand - I don't see a `GetTemplates` method in that PR (or anywhere in the codebase).  What am I missing?

tomwilkie: Yeah, I think we're good - GetAlerts returns a ConfigsResponse which contains the templates, like before.

cmd/cortex/main.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/weaveworks/common/tracing"
+	"gopkg.in/yaml.v2"
+
+	"github.com/cortexproject/cortex/pkg/cortex"
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
+)
+
+func init() {
+	prometheus.MustRegister(version.NewCollector("loki"))
```
gouthamve: `cortex`

cmd/cortex/main.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	// Parse a second time, as command line flags should take precedent over the config file.
+	flag.Parse()
+
+	util.InitLogger(&cfg.Server)
+	util.InitEvents(eventSampleRate)
+
+	// Setting the environment variable JAEGER_AGENT_HOST enables tracing
+	trace := tracing.NewFromEnv(cfg.Target.String())
```
gouthamve: Can we prefix it `cortex-`? It's just going to be a mess when there are other services called `ingester, distributor, etc.`

cmd/cortex/main.go
tomwilkie: :-)

cmd/cortex/main.go
tomwilkie: We can!  It'll be a change, but don't think it should break too much.

tomwilkie:
Questions form the slack channel:

> Is the yaml config optional ?

Yes, command line flags will continue to work and take precedence.

> `-target` doesn't seem a good way to express "take on this function"

Target was chosen for its similarity to make - is run this module and all its dependancies.  I'm not particularly wedded to it though

> can it have more than one target ?

As the code stands "no", although there is nothing to stop us doing that in the future.

> should I be able to switch over service by service from my current running system ?  I.e. are all inter-process APIs the same?

Yes & yes

pkg/cortex/modules.go
```diff
@@ -0,0 +1,462 @@
...
+	if err != nil {
+		return
+	}
+
+	rulesAPI, err := config_client.New(cfg.ConfigStore)
+	if err != nil {
+		return err
+	}
+
+	t.rulerServer, err = ruler.NewServer(cfg.Ruler, t.ruler, rulesAPI)
```
bboreham: check error?

pkg/cortex/modules.go
```diff
@@ -0,0 +1,462 @@
...
+	},
+
+	TableManager: {
+		deps: []moduleName{Server},
+		init: (*Cortex).initTableManager,
+		stop: (*Cortex).stopTableManager,
+	},
+
+	Ruler: {
+		deps: []moduleName{Distributor},
```
bboreham: `Ruler` should depend on `Store`

bboreham:
This flag has also disappeared: `-alertmanager.configs.url`

tomwilkie:
`grafana/cortex:single-binary-4ac3bf5bb` is upload for testing, should fix issues with the alertmanager flags.

cmd/querier/main.go
```diff
@@ -1,115 +0,0 @@
...
-		regexp.MustCompile(".*"),
-	)
-	promRouter := route.New().WithPrefix("/api/prom/api/v1")
-	api.Register(promRouter)
-
-	subrouter := server.HTTP.PathPrefix("/api/prom").Subrouter()
-	subrouter.PathPrefix("/api/v1").Handler(middleware.AuthenticateUser.Wrap(promRouter))
-	subrouter.Path("/read").Handler(middleware.AuthenticateUser.Wrap(querier.RemoteReadHandler(queryable)))
-	subrouter.Path("/validate_expr").Handler(middleware.AuthenticateUser.Wrap(http.HandlerFunc(dist.ValidateExprHandler)))
-	subrouter.Path("/user_stats").Handler(middleware.AuthenticateUser.Wrap(http.HandlerFunc(dist.UserStatsHandler)))
```
bboreham: This path seems to have been dropped

tomwilkie: Good catch!

bboreham: Unfortunately you added it back on distributor not querier.

PR #1267, opened 2019-03-07
Fix fmt.Errorf() error message

JoeWrightss:
"unmarshaling" to "unmarshalling".
"TestUnmarshalingCorruptedDeltaReturnsAnError" to "TestUnmarshallingCorruptedDeltaReturnsAnError".


bboreham:
BTW I would expect these to be caught by `misspell` run from the `tools/lint` script - would be good to look into why that didn't happen, if you feel like it.

PR #1271, opened 2019-03-09
Integration test running ingesters

bboreham:
A first cut.  Attempting to keep things lightweight, e.g. no Kubernetes.

The test client is at https://github.com/bboreham/avalanche/tree/remote-write

You may say "that's an ugly pile of shell script, Bryan", and I might well agree with you.
It lets me iterate quickly.

### DIFF
--- a/integration-tests/test-flush.sh
+++ b/integration-tests/test-flush.sh
@@ -1,0 +1,30 @@
+#! /bin/sh
+
+set -e
+
+. "$(dirname "$0")/common.sh"
+
+"$(dirname "$0")/start-components.sh"
+
+# TODO: wait for DynamoDB ready
+sleep 5
+echo Start table-manager
+docker run $RUN_ARGS -d --name=tm --hostname=tm --restart=always quay.io/cortexproject/cortex:$IMAGE_TAG -target=table-manager $STORAGE_ARGS -table-manager.retention-period=168h
+
+echo Start ingester
+docker run $RUN_ARGS -d --name=i1 --hostname=i1 quay.io/cortexproject/cortex:$IMAGE_TAG -target=ingester $INGESTER_ARGS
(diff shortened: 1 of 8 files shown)


csmarchbanks:
I am excited to see how this turns out. Also, if you want to make a PR for the avalanche work ping me  in it and I will take a look. Since there seems to be some interest in Avalanche we will do some updates to it soon.

integration-tests/test-flush.sh
```diff
@@ -0,0 +1,30 @@
...
+I1_ADDR=$(container_ip i1)
+wait_for "curl -s -f -m 3 $I1_ADDR/ready" "ingester ready"
+
+has_tokens_owned() {
+    curl -s -f -m 3 $1/metrics | grep 'cortex_ring_tokens_owned' >/dev/null
+}
+DIST_ADDR=$(container_ip distributor)
+wait_for "has_tokens_owned $DIST_ADDR" "distributor to see ingester in ring"
+
+docker run $RUN_ARGS --rm weaveworks/avalanche:remote-write --metric-count=2 --label-count=2 --series-count=2 --remote-url=http://distributor/api/prom/push
```
csmarchbanks: Perhaps use an upstream tag from: https://quay.io/repository/freshtracks.io/avalanche?tab=tags

bboreham:
> why `sh` rather than using go?

I'm very used to working this way, from experience with Weave Net; it lets me iterate quickly.  Part of it is I don't have a good framework in Go for starting processes, waiting for status, etc., and part I don't have a good toolset for interrupting Go programs, inspecting, status, re-running parts, etc.

Happy to take suggestions for tooling.  It's only 100 lines of shell script so should be possible to rewrite in Go in less than 1,000 😜

PR #1274, opened 2019-03-12
vendor: Update prometheus to pull in promql improvements

gouthamve:
Pulls in https://github.com/prometheus/prometheus/pull/5189 and https://github.com/prometheus/prometheus/pull/5290

Updates to AzureSDK and k8s libs seem to be most of the additions.


khaines:
pulling in  
https://github.com/prometheus/prometheus/pull/5189, will certainly help some confusion we get from users.

PR #1275, opened 2019-03-12
Inject 256 splits into every bigtable at creation time.

tomwilkie:
After #1239 and #1242, we have nice small rows which are very evenly balanced in BigTable.  But we're still seeing high latencies on some requests.  The suggestion is that our weekly tables are too small - index tables in prod are only ~150GB on average - and that it too small for Bigtable to split them into enough tablets to load balance across our servers.

This change injects 256 splits into each table at creation time.  This should result in better distribution of tablets amongst servers and lower latency.

### DIFF
--- a/pkg/chunk/gcp/table_client.go
+++ b/pkg/chunk/gcp/table_client.go
@@ -2,6 +2,7 @@ package gcp
 
 import (
 	"context"
+	"fmt"
 
 	"google.golang.org/grpc/codes"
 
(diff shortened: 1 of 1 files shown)


gouthamve:
I'd rather not do this until we know this is the cause for our slowness.

tomwilkie:
How will we know without trying it? I was going to experiment in ops-tools and see if this helps; I see it as relatively risk free.

tomwilkie:
We ran this change for a week and didn't see any change in bigtable latency.  The bigtable cluster we ran on only had ~4 nodes, and this probably only effects bigger clusters. A bit of me thinks it will still help, but I can justify it.  So closing.

PR #1278, opened 2019-03-13
Intern label strings

### LINKED ISSUE #1257: Only hold one copy of each label name and value in ingesters
[Just jotting down an idea which came to me; haven't thought it through very much]

Prometheus metrics will generally come in "families" that share a lot of the same labels, with one or two differing between timeseries.

Currently the ingester will have a copy of each label name and value for each series, which adds up to a lot.

It occurs to me that the inverted index should provide a place where we know we have one copy of each string.

bboreham:
Fixes #1257 (mostly)

Currently, we have a separate copy of each label names and value string (e.g. `instance`, `kubernetes_namespace`, `get`) for each timeseries.  This consumes about 25% of all ingester memory in profiles I've looked at.

The inverted index is keyed by those strings, so we have a ready-made place to reduce down to one copy (per index shard per user). 

Unfortunately Go doesn't let us access the key of a map on lookup, so we have to add another reference to the string in the map value.  This makes data access a bit more ugly, and also costs another 16 bytes per index entry, but unless you have a shedload of unique short label values I think we'll come out ahead.

Benchmarks (Go 1.12; best of 5 runs):
Before:
```
BenchmarkIngesterPush-2   	      20	  87524048 ns/op	 3086538 B/op	   13751 allocs/op
```
After:
```
BenchmarkIngesterPush-2   	      20	  86766945 ns/op	 2843284 B/op	   12347 allocs/op
```

### DIFF
--- a/pkg/ingester/index/index.go
+++ b/pkg/ingester/index/index.go
@@ -24,7 +24,7 @@ type InvertedIndex struct {
 func New() *InvertedIndex {
 	shards := make([]indexShard, indexShards)
 	for i := 0; i < indexShards; i++ {
-		shards[i].idx = map[model.LabelName]map[model.LabelValue][]model.Fingerprint{}
+		shards[i].idx = map[model.LabelName]indexEntry{}
 	}
 	return &InvertedIndex{
 		shards: shards,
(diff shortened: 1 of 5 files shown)


bboreham:
I tried this out in our staging environment and Go heap size went down around 10%.
The total memory allocated in `copyValuesAndSort` was 18%, so a bit less than I was remembering above.

tomwilkie:
LGTM!  Only (optional) nit is that it might be time to remove the `model.LabelName/Value` casts, and make these strings.

bboreham:
Created #1299 to remove casts

PR #1282, opened 2019-03-14
Parallelise calls from the querier -> ingester in an attempt to reduce latency.

tomwilkie:
Even after #1233, we're still seeing long tail latencies on QueryStream calls to the ingester.  This PR splits calls from the queries to the the ingesters up into a configurable number of parallel calls.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -104,7 +104,8 @@ type Config struct {
 	ExtraQueryDelay     time.Duration
 	LimiterReloadPeriod time.Duration
 
-	ShardByAllLabels bool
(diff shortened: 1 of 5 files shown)


bboreham:
When I look in Jaeger it seems those calls are already parallelised. This change is to make them less parallel?

tomwilkie:
> When I look in Jaeger it seems those calls are already parallelised. This change is to make them less parallel?

They're parallelised to multiple ingesters, but each ingester only has a single stream.  This change gives each ingester multiple streams.

Not clear its worth it after #1306 though.  And if we did, I'd want to integrate this with the inverted index shards.

PR #1291, opened 2019-03-19
Allow extra space to add new bigchunk data after Unmarshal

bboreham:
As noted at #1220, after a transfer of chunks from one ingester to another, a new `smallChunk` is created immediately for any new samples that arrive. Allowing room for this avoids a copy and doubling in size of the slice of `smallChunk`s.

We go through the same code path on queries, so they will get slightly larger in memory - 24 bytes per chunk.

This change wouldn't be so useful if #1221 could be made to work.


tomwilkie:
LGTM x 2

PR #1293, opened 2019-03-20
Fix table lookup when chunk time aligns with table

gouthamve:
When `Chunk.From` and `Config.From` are same, then the current code fails with:
```
msg="failed to flush user" err="no chunk table found for time 1551830400"
```

### DIFF
--- a/pkg/chunk/schema_config_test.go
+++ b/pkg/chunk/schema_config_test.go
@@ -3,8 +3,10 @@ package chunk
 import (
 	"reflect"
 	"testing"
+	"time"
 
 	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/require"
 )
 
 func TestHourlyBuckets(t *testing.T) {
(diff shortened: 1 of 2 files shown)


gouthamve:
Fixed to Unix epoch

pkg/chunk/schema_config_test.go
```diff
@@ -178,3 +180,101 @@ func TestDailyBuckets(t *testing.T) {
...
+		ts, err := time.Parse(time.RFC3339, tc.timeStr)
+		require.NoError(t, err)
+
+		table, err := schemaCfg.ChunkTableFor(model.TimeFromUnix(ts.Unix()))
+		require.NoError(t, err)
+
+		require.Equal(t, tc.chunkTable, table)
+	}
+
+	return
```
tomwilkie: This `return` is redundant, right?

tomwilkie:
Can we have an issue for this too?  I'l like to do things like:
- Keep a per-chunk counter of number of flush attempts, and alert when it gets too high, to catch this case of chunks failing to flush forever.
- Have a hook to slurp all the chunks out of an ingester so we have a way of recovering this.

PR #1295, opened 2019-03-21
Fix modulo arithmetic to prevent ingesters OOMing.

tomwilkie:
Previously would only call send at the end of the function, causing the batches to be massive.

### DIFF
--- a/pkg/ingester/user_state_test.go
+++ b/pkg/ingester/user_state_test.go
@@ -1,6 +1,7 @@
 package ingester
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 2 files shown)


pkg/ingester/user_state_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+
+	"github.com/stretchr/testify/require"
+	"github.com/weaveworks/common/user"
+	"golang.org/x/net/context"
+)
+
+func TestForSeriesMatching(t *testing.T) {
```
gouthamve: Can you name it `TestForSeriesMatchingBatching` as batching only is being tested?

tomwilkie: Sure!

PR #1297, opened 2019-03-22
fix bug of caller location in log messages

### LINKED ISSUE #1230: Log messages do not give useful location
Example:
```
level=warn ts=2019-02-04T09:21:37.646512093Z caller=gokit.go:46 method=/cortex.Ingester/Query duration=216.571µs err=EOF msg="gRPC\n"
```

`gokit.46` is utility code; we want to know where it was called from.

I believe the code does some counting of stack frames to arrive at the `caller`, so we need to set things up in a different order or adjust the count. 

Note some other messages do work as expected:
```
level=info ts=2019-02-04T09:21:35.849706809Z caller=lifecycler.go:441 msg="changing ingester state from" old_state=JOINING new_state=ACTIVE
```

Very similar to #655

yiyuanliu:
This PR fix #1230 

DefaultCaller assumes the depth of callstack is 3. Gokit wraps go-kit log function increases callstack depth to 4. Because gokit's methods are declared with value receiver, when used as a instance of logging.Interface, there's an autogenerated function in callstack. So, the depth of callstack when log with gokit should be 5.

Signed-off-by: Yiyuan Liu <yiyuanliu1997@gmail.com>


pkg/util/log.go
```diff
@@ -44,8 +44,8 @@ func InitLogger(cfg *server.Config) {
 		panic(err)
 	}
 
-	Logger = l
-	cfg.Log = logging.GoKit(l)
+	Logger = log.With(l, "caller", log.DefaultCaller)
+	cfg.Log = logging.GoKit(log.With(l, "caller", log.Caller(5)))
```
bboreham: Do we really want both `DefaultCaller` and `log.Caller` ?

yiyuanliu: They have different stack depth. `DefaultCaller` is equivalent to `log.Caller(3)`.

bboreham: OK, but do we want two of them on every log line?

bboreham: oh, these are going to different places?

yiyuanliu: Yes. There won't be two caller information in a single log message.

pkg/util/log.go
```diff
@@ -69,7 +69,7 @@ func NewPrometheusLogger(l logging.Level) (log.Logger, error) {
 	}
 
 	// DefaultCaller must be the last wrapper
-	logger = log.With(logger, "ts", log.DefaultTimestampUTC, "caller", log.DefaultCaller)
+	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
```
bboreham: comment on line above doesn't make sense here any more.

yiyuanliu: Comment has updated, tell me if there are something unclear.

bboreham:
Now I'm getting:

```
level=warn ts=2019-03-28T11:02:28.071677812Z caller=chain.go:39 method=/cortex.Ingester/Query duration=127.475902ms err="context canceled" msg="gRPC\n"
```

which doesn't seem right either.  That file is `vendor/github.com/grpc-ecosystem/go-grpc-middleware/chain.go`; the real caller is `vendor/github.com/weaveworks/common/middleware/grpc_logging.go`

PR #1305, opened 2019-03-27
log with system time zone

### LINKED ISSUE #1304: log with system time zone
For [code](https://github.com/cortexproject/cortex/blob/master/pkg/util/log.go#L72), it always use UTC time zone, is it possible to use system time zone, it would be better to read for users from any time zone.

daixiang0:
Fix #1304 

Signed-off-by: Xiang Dai <764524258@qq.com>

### DIFF
--- a/pkg/util/log.go
+++ b/pkg/util/log.go
@@ -69,7 +69,7 @@ func NewPrometheusLogger(l logging.Level) (log.Logger, error) {
 	}
 
 	// DefaultCaller must be the last wrapper
-	logger = log.With(logger, "ts", log.DefaultTimestampUTC, "caller", log.DefaultCaller)
+	logger = log.With(logger, "ts", log.DefaultTimestamp, "caller", log.DefaultCaller)
 	return logger, nil
 }
 


bboreham:
Thanks for the PR.

When I have to support servers in multiple timezones, I prefer UTC.
I guess it could be an option.

daixiang0:
Do you mean add a func with this option?

bboreham:
I meant a command-line option. 

If you look around, how do you see this done in other server software?

daixiang0:
They do not use UTC but system time zone, for softwares which call `pkg/util/log`, do not get how to switch to system time zone.

daixiang0:
@bboreham any input? Do you have any idea about switch UTC to system time zone when call `pkg/util/log`.

bboreham:
I am starting to wonder if this makes no practical difference to me, since we run Cortex inside containers and containers default to UTC. 

For yourself, how do you propose to run Cortex such that timezone information is available?  Outside of containers?

What do other maintainers think?  @tomwilkie @khaines @gouthamve

daixiang0:
In my case, since my time zone is not UTC, the log time is useless for me, so i want to mount /etc/localtime then i can get time in my time zone. But Cortex uses UTC as default and have no option to do it now.

khaines:
My local timezone is not UTC, but all of our machines & containers deployed around the world have UTC set as their TZ. 

I guess I'm feeling that this sort of TZ conversion should be done outside of cortex. The log processing tools we use allow for viewing logs in local time and it handles that conversion at display time vs recording time.

daixiang0:
@khaines could you give me an example, i do not know how to use local time when call `corex/pkg/util/log`.

khaines:
@daixiang0 What I was saying is that I log everything in UTC, though I do not live in the UTC zone; so there is no code example for cortex that I can give you.

I use external tools such as fluentd/logstash/Loki/Elasticsearch to process log files. These tools do the TZ conversion for me, so when I'm reading logs the timestamps appear in my local TZ.

PR #1306, opened 2019-03-28
Optimise Ingester.QueryStream

tomwilkie:
- Build a simple Benchmark that uses QueryStream to fetch 1e6 series.
- Remove copying of labels by directly marshalling `labels.Lables` (warning: uses `unsafe`)
- Remove the FP sort in index.Lookup, AFAICT its not needed for locking.
- Tweak some more allocations and increase batch size.

For me, this reduces the QueryStream run time from ~1.7s to ~1s.  ~Still got work to do in other packages, and not happy with so many different representations of labels - need to tidy this up.~


tomwilkie:
Before:

Screenshot 2019-03-28 at 10 38 52

After:

Screenshot 2019-03-28 at 11 22 35

pkg/ingester/client/timeseries.go
```diff
@@ -41,3 +47,63 @@ func (p *PreallocTimeseries) Unmarshal(dAtA []byte) error {
...
+
+type LabelAdapter labels.Label
+
+func (bs *LabelAdapter) Marshal() ([]byte, error) {
+	buf := make([]byte, bs.Size())
+	_, err := bs.MarshalTo(buf)
+	return buf, err
+}
+
+func (bs *LabelAdapter) MarshalTo(buf []byte) (n int, err error) {
```
gouthamve: Where is this function being used? Is this used by grpc/proto or somewhere else? Also needs comments.

pkg/ingester/client/compat.go
```diff
@@ -236,6 +236,31 @@ func FromLabelPairs(labelPairs []LabelPair) model.Metric {
...
+
+func FromLabelAdaptersToMetric(ls []LabelAdapter) model.Metric {
+	metric := make(model.Metric, len(ls))
+	for _, l := range ls {
+		metric[model.LabelName(l.Name)] = model.LabelValue(l.Value)
+	}
+	return metric
+}
+
+// FromLabelPairs unpack a []LabelPair to a model.Metric
```
gouthamve: `FromLabelPairs` --> `FromLabelPairAdapaters`

pkg/ingester/client/timeseries.go
tomwilkie: Will comment, for sure.  This is the implementation of the Protobuf encoding.

tomwilkie:
See the notes on `Big change: Remove the label copying in QueryStream.` for more on this.

I've run out other benchmarks and don't see any regressions.

tomwilkie:
Deployed this out to one of our (under provisioned) test envs, saw P99 QueryStream drop by ~50%.  Also say a reduction in push latency, but the new image includes a bunch of optimisations so I wouldn't read too much into that.

Screenshot 2019-04-01 at 11 40 18

pkg/ingester/locker.go
```diff
@@ -30,7 +30,7 @@ type paddedMutex struct {
 // the same mutex twice).
 type fingerprintLocker struct {
 	fpMtxs    []paddedMutex
-	numFpMtxs uint
+	numFpMtxs uint32
```
gouthamve: Does this really make enough difference? Just curious if it does.

pkg/ingester/index/index.go
```diff
@@ -49,7 +49,6 @@ func (ii *InvertedIndex) Lookup(matchers []*labels.Matcher) []model.Fingerprint
 		result = append(result, fps...)
 	}
 
-	sort.Sort(fingerprints(result))
```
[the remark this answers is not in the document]
tomwilkie: "Remove the sort from the inverted index Lookup, its not actually needed."

I also add a comment in the place we call this discussion the locking.  Previously we used to ensure we locked the series in order for a query, to prevent deadlocks.  Now we only ever hold one series lock at once, so deadlocks can't happen.  Hence we don't need the sort.

pkg/ingester/locker.go
tomwilkie: This makes no difference.  its there as I pulled in the fnv32 implementation to avoid copying a string to a []byte, and reused it here.

tomwilkie:
> If I didn't miss anything, the main change here is to move away from proto generated LabelPair to custom type LabelAdapter labels.Label and using unsafe to do away with any copying? 

Thats right; the commit comment on the last change goes into more detail.  Removing the copying is pretty much the only change here, but it was quite involved.

> Also reducing allocs by using yoloString in the unmarshalling?

We previously used LabelPair, which avoided the copy by storing the labels as []byte.  As I've moved everything to labels.Labels (via LabelAdapater), we need a different technique to avoid the copy - `yoloString`.   If I've got this right, it should be safe as the only reference to the labels is held in the index - and this is a copy.

tomwilkie:
I've got this deployed to our dev/test envs.  Will leave it there for a day or so before I merge this.

tomwilkie:
Thanks Ken! That comments was out of date - I've plumbed this change through everywhere and had it in test for the last few days.

PR #1309, opened 2019-03-29
Document the logic behind cardinality limits.

gouthamve:
~We're not loading the chunks in the initial queries, instead just
figuring out which series have to be loaded. I don't think it makes
sense to enforce cardinality limits on queries that return seriesIDs
which will be intersected away.~

~Here, I made the changes to make sure only the final intersected result
is used for the cardinality limits. As only these chunks are loaded, I
think we should be good just checking the cardinality limits here.~

EDIT: Documented the logic behind cardinality limits.

### DIFF
--- a/pkg/chunk/series_store.go
+++ b/pkg/chunk/series_store.go
@@ -210,6 +210,10 @@ func (c *seriesStore) lookupSeriesByMetricNameMatchers(ctx context.Context, from
 				ids = intersectStrings(ids, incoming)
 			}
 		case err := <-incomingErrors:
+			// The idea is that if we have 2 matchers, and if one returns a lot of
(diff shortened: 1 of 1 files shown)


bboreham:
I think you've missed the intent.

Suppose we have three matchers A, B and C, that gave 1,000,000, 10 and 2 matches.
We want to find A in the cache, skip doing A at all, then intersect B and C.

gouthamve:
So yes, I did miss the intent :) I'll add comments instead.

bboreham:
Seems fair enough, lgtm.  Probably better to get approval from @tomwilkie who wrote the code.

tomwilkie:
:+1:

Issue #1276, opened 2019-03-13
Buffered datapoints returned after last sample has been pushed


rsteneteg:
When datapoints for a time series stops being pushed, Cortex will continue to return the last datapoint for additional time period (the full lookback delta) even if no sample was pushed to Cortex during that window.

<img width="447" alt="image" src="https://user-images.githubusercontent.com/1894640/54256151-087e9580-4529-11e9-8d38-f75fd1f5bc4e.png">

In the above picture I pushed 1 sample to Cortex at 13:43:15, yet it returned 8 datapoints over a 2m period (I currently have the lookback delta set to 2m, if I use the default lookback Cortex returns datapoints for 5m instead).

Since I only pushed in 1 sample I would have thought I would only see 1 datapoint.

By looking at the code it is the BufferedSeriesIterator in the Prometheus code that returns these extra datapoints, since it uses the lookback delta here https://github.com/cortexproject/cortex/blob/master/vendor/github.com/prometheus/prometheus/storage/buffer.go#L79

I did a quick look at the code in Prometheus master as well and they look similar, yet if I make Prometheus scrape 1 sample, Prometheus will only return datapoints for the duration of the Prometheus scrape interval (not the full lookback delta).

rsteneteg:
FYI @khaines

rsteneteg:
It seems like Prometheus advances the Buffered iterator here https://github.com/prometheus/prometheus/blob/master/storage/buffer.go#L97, while Cortex don't

tomwilkie:
Thanks the the report @rsteneteg! I'm wondering if this has to do with staleness handling.  When you say "if I make Prometheus scrape 1 sample, Prometheus will only return datapoints for the duration of the Prometheus scrape interval" how did you achieve this?  Because if you removed the target or series, I believe Prometheus inserts a special "end of timeseries" marker.

Cortex should honor them, if they get propagated.  But if they don't, if will fall back to the legacy behaviour and return the old value for the staleness period.

rsteneteg:
As metrics flows through my metrics pipeline they pass through a custom prometheus exporter that I built that will remove all metrics after it has been scraped, similar to how the blackbox exporter works by rebuilding the prometheus registry for each scrape.

I found last night that Cortex uses the `concreteSeriesIterator`, while prometheus uses a different iterator called `chunkSeriesIterator` which seems to behave differently

rogst:
[...]

PR #1310, opened 2019-03-30
modified concreteChunkIterator to behave like Prometheus with adding …

rogst:
…an end marker to indicate stale metrics to avoid repeating buffered datapoints after the last sample

Signed-off-by: Roger Steneteg <roger@steneteg.org>

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -17,6 +17,12 @@ import (
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
+var (
+	// EndMarkerInjectDelta determines how long after the last sample for a time serie within the
+	// lookback delta to inject a NaN marker to stop repeating buffered datapoints
+	EndMarkerDelta = 1 * time.Minute
+)
+
 // Config contains the configuration require to create a querier
 type Config struct {
 	MaxConcurrent            int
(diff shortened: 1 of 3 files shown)


rogst:
Fixes #1276

rogst:
Added a configurable duration for how long after the last sample within the lookback period to add the NaN end marker, similar to what Prometheus does for series that is removed from a scrape target. Since Cortex does not scrape we need to configure how soon after the NaN value should be.

The EndMarkerDelta should be set to the lowest query step size you want to use.

The behavior is that the BufferedSeriesIterator will fill in missing samples using the previous sample up until the last real sample within the lookback delta.

tomwilkie:
Had a chat with Roger on slack.  I'd like to understand if/why the NaN is not being propagated from the scraping Prometheus, or not being store correctly by us, first.  We shouldn't be inserting NaN like this if we can get away with it.

rogst:
Thanks Tom, for your insight on this. I have verified it now and cortex does honor the propagated NaNs received from Prometheus with the remote write.
The reason I got the trailing datapoints was that I tried to send in sparse metrics using a custom component with Prometheus remote writes which did not send in NaNs.
I'm closing this PR

PR #1312, opened 2019-04-03
In query-frontend, log details of each request to OpenTracing

bboreham:
This is to help with debugging: the top-level request that comes in and what is sent to queriers are logged to the current span, so you can see the effects of splitting, caching, etc.

### DIFF
--- a/pkg/querier/frontend/query_range.go
+++ b/pkg/querier/frontend/query_range.go
@@ -94,10 +94,10 @@ func (q QueryRangeRequest) toHTTPRequest(ctx context.Context) (*http.Request, er
 
 func (q QueryRangeRequest) logToSpan(ctx context.Context) {
 	if span := opentracing.SpanFromContext(ctx); span != nil {
-		span.LogFields(otlog.String("query", q.Query))
-		span.LogFields(otlog.String("start", timestamp.Time(q.Start).String()))
-		span.LogFields(otlog.String("end", timestamp.Time(q.End).String()))
-		span.LogFields(otlog.Int64("step (ms)", q.Step))
+		span.LogFields(otlog.String("query", q.Query),
+			otlog.String("start", timestamp.Time(q.Start).String()),
+			otlog.String("end", timestamp.Time(q.End).String()),
+			otlog.Int64("step (ms)", q.Step))
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/querier/frontend/frontend.go
```diff
@@ -316,6 +316,9 @@ func (f *Frontend) Process(server Frontend_ProcessServer) error {
 		}
 
 		originalCtx := request.originalCtx
+		if request.request != nil && request.request.QueryRangeRequest != nil {
```
tomwilkie: Under what circumstances will request.request be nil?  I can't think of any.

bboreham: Idk it crashed in the unit tests and I couldn’t be bothered figuring out which one was nil.

tomwilkie: I removed `request.request != nil &&` and the tests passed...

Either way this is in the wrong place; if you look at L172, `QueryRangeRequest` is never initialised in `ProcessRequest`.  I don't even recall why its there.  We should probably remove it, and put this call into `queryRangeTerminator`.

bboreham: Thanks for the tip!

bboreham:
Looks like this:
image

pkg/querier/frontend/query_range.go
```diff
@@ -91,6 +92,15 @@ func (q QueryRangeRequest) toHTTPRequest(ctx context.Context) (*http.Request, er
 	return req.WithContext(ctx), nil
 }
 
+func (q QueryRangeRequest) logToSpan(ctx context.Context) {
+	if span := opentracing.SpanFromContext(ctx); span != nil {
+		span.LogFields(otlog.String("query", q.Query))
+		span.LogFields(otlog.String("start", timestamp.Time(q.Start).String()))
+		span.LogFields(otlog.String("end", timestamp.Time(q.End).String()))
+		span.LogFields(otlog.Int64("step (ms)", q.Step))
```
tomwilkie: Nit: should this be one call to LogFields (its variadic)?  Not sure what the best practices here are.

tomwilkie:
LGTM with one nit.

Issue #349, opened 2017-03-16
QoD caused by start=<huge-negative-value>&end=<positive-value> (huge start-end range)


bwplotka:
Follow up of: https://github.com/weaveworks/cortex/issues/327

My QoD is still valid. I dived into the issue and it seems that the confusion here is because Prometheus wrongly parses my mistaken `start=1489025420000` timestamp. From query perspective it is not really start > time problem, but actually underlying int64 overflow I fixed here:  https://github.com/prometheus/prometheus/pull/2501 caused start to be huge negative value.

This means that QoD is still possible and is caused not by `start > end`, but actually huge range (e.g `start=MinInt64&end=MaxInt64`)

@tomwilkie do you have any idea how to handle this properly? IMO Some limiter or further validation is needed.

tomwilkie:
Yeah, we'll just have to limit the size of the biggest query to something like 13months.

Bplotka:
Or configured by flag

juliusv:
Is it really a query of death though? It should at least just time out after 2 minutes (default `promql.Engine` timeout), right?

tomwilkie:
I wonder if we honour the context cancelled stuff in the chunk store...

juliusv:
As far as I can see, we don't care about the context done-ness anywhere in the chunk store :-/

I filed another issue for that because that's a bigger undertaking to make all of the chunk store context cancellation aware: https://github.com/weaveworks/cortex/issues/381

bboreham:
I think a limit on length of date range should be done in the Prometheus code that first unpacks the query request. Cortex doesn't see the dates until the query comes down into the chunk store.

It would be appropriate to have a limit in the chunk store on things that are uniquely chunk store concerns, e.g. how many buckets the query touches.

PR #1059, opened 2018-10-04
Optionally limit size of queries to the chunk store.

tomwilkie:
Fixes #1057 

When combined with day-by-day query parallelism in the query frontend, this limit in the chunk store will limit vector windows (ie `rate(foo[20d])`) whilst not limit query lengths (ie still being able to do 1yr range queries).

Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>

PR #1314, opened 2019-04-03
Limit number of days a query can be split into.

tomwilkie:

### DIFF
--- a/pkg/querier/frontend/query_range.go
+++ b/pkg/querier/frontend/query_range.go
@@ -127,6 +127,11 @@ func encodeDurationMs(d int64) string {
 	return strconv.FormatFloat(float64(d)/float64(time.Second/time.Millisecond), 'f', -1, 64)
 }
 
+func msToTime(t int64) time.Time {
+	const msPerS = int64(time.Second / time.Millisecond)
+	return time.Unix(t/msPerS, t%msPerS)
+}
+
 const statusSuccess = "success"
 
 func parseQueryRangeResponse(ctx context.Context, r *http.Response) (*APIResponse, error) {
--- a/pkg/querier/frontend/query_range.go
+++ b/pkg/querier/frontend/query_range.go
@@ -127,11 +127,6 @@ func encodeDurationMs(d int64) string {
 	return strconv.FormatFloat(float64(d)/float64(time.Second/time.Millisecond), 'f', -1, 64)
 }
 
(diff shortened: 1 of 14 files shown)


bboreham:
Fixes #349 ?

Why limit this to when split-by-days is on - wouldn't it be more understandable to have a simple length limit?

tomwilkie:
> Why limit this to when split-by-days is on - wouldn't it be more understandable to have a simple length limit?

That already covered by the limits in the chunks store (see #1059) - when you don't split there is very little chance of abuse in the frontend.  This limit is here to protect the memcaches from abuse, for instance.

I do want to me this configurable on a per-user basis though.

tomwilkie:
@bboreham I've updated this to use the same flag we use in the querier, and as a result made it configurable per user.

pkg/querier/frontend/query_range.go
```diff
@@ -127,6 +127,11 @@ func encodeDurationMs(d int64) string {
 	return strconv.FormatFloat(float64(d)/float64(time.Second/time.Millisecond), 'f', -1, 64)
 }
 
+func msToTime(t int64) time.Time {
```
bboreham: I used Prometheus `timestamp.Time()` for this in #1312 - we should be consistent.

pkg/util/validation/limits.go
```diff
@@ -55,6 +56,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 
 	f.IntVar(&l.MaxChunksPerQuery, "store.query-chunk-limit", 2e6, "Maximum number of chunks that can be fetched in a single query.")
 	f.DurationVar(&l.MaxQueryLength, "store.max-query-length", 0, "Limit to length of chunk store queries, 0 to disable.")
+	f.IntVar(&l.MaxQueryParallelism, "querier.max-query-parallelism", 14, "Maximum number of queries will be scheduled in parallel by the frontend.")
```
bboreham: Please mention this change in the commit description.  Ideally a separate commit.

bboreham:
Seems OK apart from nits.

pkg/util/validation/limits.go
tomwilkie: :+1:

pkg/querier/frontend/query_range.go
tomwilkie: :+1:

PR #1315, opened 2019-04-04
Use the correct ip address for minikube

heliodor:
Explains that Minikube might not be using 192.168.99.100 and how to find the correct ip address.


bboreham:
Note that Cortex, like all CNCF-hosted projects, requires that you “sign off” your commits. See https://github.com/cortexproject/cortex/runs/92593952

PR #1319, opened 2019-04-05
Default per-user limits to those specified on the command line.

tomwilkie:
Not to command line defaults.  Eg if I specify `-validation.reject-old-samples=true` on the CLI, I want this to apply to all users unless explicitly disabled.   


tomwilkie:
(NB this is a bit of hack, I can't think of a nicer way though)

PR #1320, opened 2019-04-05
Improve index cardinality limiting.

tomwilkie:
We currently have a way of limiting load on the backends by caching the size of rows and preventing re-reading of very large rows. This is called the cardinality cache, and is currently in-process. As you add more queriers, it becomes less useful.

This PR adds support for "external" cardinality caching using the index read cache.  We also make query cardinality limit configurable per user. 

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -60,7 +60,6 @@ type StoreConfig struct {
 	MinChunkAge              time.Duration
 	CardinalityCacheSize     int
 	CardinalityCacheValidity time.Duration
-	CardinalityLimit         int
 
 	CacheLookupsOlderThan time.Duration
 }
@@ -74,7 +73,6 @@ func (cfg *StoreConfig) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.MinChunkAge, "store.min-chunk-age", 0, "Minimum time between chunk update and being saved to the store.")
 	f.IntVar(&cfg.CardinalityCacheSize, "store.cardinality-cache-size", 0, "Size of in-memory cardinality cache, 0 to disable.")
 	f.DurationVar(&cfg.CardinalityCacheValidity, "store.cardinality-cache-validity", 1*time.Hour, "Period for which entries in the cardinality cache are valid.")
-	f.IntVar(&cfg.CardinalityLimit, "store.cardinality-limit", 1e5, "Cardinality limit for index queries.")
 	f.DurationVar(&cfg.CacheLookupsOlderThan, "store.cache-lookups-older-than", 0, "Cache index entries older than this period. 0 to disable.")
 }
 
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/chunk/cache"
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/extract"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 	"github.com/cortexproject/cortex/pkg/util/validation"
 	"github.com/weaveworks/common/httpgrpc"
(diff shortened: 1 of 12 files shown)


pkg/chunk/chunk_store.go
```diff
@@ -57,25 +57,21 @@ type StoreConfig struct {
...
 	cfg.WriteDedupeCacheConfig.RegisterFlagsWithPrefix("store.index-cache-write.", "Cache config for index entry writing. ", f)
 
 	f.DurationVar(&cfg.MinChunkAge, "store.min-chunk-age", 0, "Minimum time between chunk update and being saved to the store.")
-	f.IntVar(&cfg.CardinalityCacheSize, "store.cardinality-cache-size", 0, "Size of in-memory cardinality cache, 0 to disable.")
-	f.DurationVar(&cfg.CardinalityCacheValidity, "store.cardinality-cache-validity", 1*time.Hour, "Period for which entries in the cardinality cache are valid.")
-	f.IntVar(&cfg.CardinalityLimit, "store.cardinality-limit", 1e5, "Cardinality limit for index queries.")
 	f.DurationVar(&cfg.CacheLookupsOlderThan, "store.cache-lookups-older-than", 0, "Cache index entries older than this period. 0 to disable.")
+
+	// Deprecated.
+	f.Int("store.cardinality-cache-size", 0, "DEPRECATED. store.cardinality-cache.enable-fifocache and store.cardinality-cache.fifocache.size.")
```
khaines: Possible Nit: The flag text could be more descriptive for these two deprecated flags

bboreham: This isn’t “deprecated” it’s “disabled”. 
It should print a warning a least if you use it.

pkg/chunk/storage/caching_index_client.go
```diff
@@ -135,18 +150,29 @@ func (s *cachingIndexClient) QueryPages(ctx context.Context, queries []chunk.Ind
 		defer resultsMtx.Unlock()
 		keys := make([]string, 0, len(results))
 		batches := make([]ReadBatch, 0, len(results))
+		var cadinalityErr error
```
bboreham: typo `cadinalityErr` -> `cardinalityErr`

pkg/chunk/chunk_store.go
```diff
@@ -57,23 +57,21 @@ type StoreConfig struct {
...
-
 	cfg.WriteDedupeCacheConfig.RegisterFlagsWithPrefix("store.index-cache-write.", "Cache config for index entry writing. ", f)
 
 	f.DurationVar(&cfg.MinChunkAge, "store.min-chunk-age", 0, "Minimum time between chunk update and being saved to the store.")
-	f.IntVar(&cfg.CardinalityCacheSize, "store.cardinality-cache-size", 0, "Size of in-memory cardinality cache, 0 to disable.")
-	f.DurationVar(&cfg.CardinalityCacheValidity, "store.cardinality-cache-validity", 1*time.Hour, "Period for which entries in the cardinality cache are valid.")
 	f.DurationVar(&cfg.CacheLookupsOlderThan, "store.cache-lookups-older-than", 0, "Cache index entries older than this period. 0 to disable.")
+
+	// Deprecated.
+	f.Int("store.cardinality-cache-size", 0, "DEPRECATED. store.cardinality-cache.enable-fifocache and store.cardinality-cache.fifocache.size.")
```
bboreham: Should this change be part of the previous commit?

tomwilkie: No I don't think so - the previous one was about the cardinality limit; this is about the cache we used to store the cardinalities.

pkg/chunk/storage/caching_index_client.go
```diff
@@ -135,18 +150,29 @@ func (s *cachingIndexClient) QueryPages(ctx context.Context, queries []chunk.Ind
 		defer resultsMtx.Unlock()
 		keys := make([]string, 0, len(results))
 		batches := make([]ReadBatch, 0, len(results))
+		var cardinalityErr error
 		for key, batch := range results {
+			cardinality := int32(len(batch.Entries))
+			if cardinalityLimit > 0 && cardinality > cardinalityLimit {
+				batch.Cardinality = cardinality
+				batch.Entries = nil
```
gouthamve: This is dangerous, this assumes that the cardinality limit is static, but it can be dynamically changed with a file. And given we cache the entries forever, this won't go away until we potentially rotate the caches.

tomwilkie: Fair point!  I don't want to cache very large entries (it will slow down the fetch, just to discard them).  Perhaps we should throw away empty cache entries?  Please file an issue.

PR #1321, opened 2019-04-05
Support external cardinality caching.

tomwilkie:
We currently have a way of limiting load on the backends by caching the size of rows and preventing re-reading of very large rows.  This is called the cardinality cache, and is currently in-process.  As you add more queriers, it becomes less useful.

This PR adds support for "external" (ie, memcached) cardinality caching.


tomwilkie:
Thinking about this, we already have a memcached keyed by the index rows hash value, so we could actually use that?  One less thing to manage etc.

tomwilkie:
Yeah that look a bit nicer.  Going to merge this with #1320 as they're so related.

PR #1327, opened 2019-04-10
Set the default interval for promql to eval subqueries.

### LINKED ISSUE #1265: "unexpected error: runtime error: integer divide by zero" doing subquery
```
quantile_over_time(0.99, sum(rate(cortex_gw_request_duration_seconds_count{job=~"(.+)/cortex-gw", route="cortex-read"}[5m]))[24h:])
```

NB this doesn't happen with prometheus 2.8rc0.  Either we need to update vendor or we're missing some config.

gouthamve:
When you have a subquery of the form `[1h:]`, the second parameter is substituted to be the default evaluation interval by Prometheus. We don't set that in Cortex.

This is a little contentious, because if users change the default in their prometheus, their graphs in local vs cloud would be different.

Fixes #1265

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -27,6 +27,11 @@ type Config struct {
 	MaxSamples               int
 	IngesterMaxQueryLookback time.Duration
 
+	// The default evaluation interval for the promql engine.
+	// Needs to be configured for subqueries to work as it is the default
+	// step if not specified.
+	DefaultEvaluationInterval time.Duration
+
 	// For testing, to prevent re-registration of metrics in the promql engine.
 	metricsRegisterer prometheus.Registerer
 }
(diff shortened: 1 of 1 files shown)


pkg/querier/querier.go
```diff
@@ -43,6 +48,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.IngesterStreaming, "querier.ingester-streaming", false, "Use streaming RPCs to query ingester.")
 	f.IntVar(&cfg.MaxSamples, "querier.max-samples", 50e6, "Maximum number of samples a single query can load into memory.")
 	f.DurationVar(&cfg.IngesterMaxQueryLookback, "querier.query-ingesters-within", 0, "Maximum lookback beyond which queries are not sent to ingester. 0 means all queries are sent to ingester.")
+	f.DurationVar(&cfg.DefaultEvaluationInterval, "querier.default-evaluation-interval", time.Minute, "The default evaluation interval or step size for subqueries.")
```
bboreham: I would slightly prefer this defaults to 15 seconds, but not going to make a big thing of it

gouthamve: The default for it in prometheus is `1m` hence I made it `1m` here too.

tomwilkie:
QQ: should this be overridable per user?  Or should we cross that bridge when we get there?

pkg/querier/querier.go
tomwilkie: Sticking with Prometheus defaults seems the right thing to do.

gouthamve:
Well making it per user would make things much more complicated as that would mean changes upstream. Currently the eval interval is a global variable.

gouthamve:
Merging this as-is as I think we should cross the bridge when we get there :)

PR #1330, opened 2019-04-15
Nuke some old deprecated flags

gouthamve:
We have some deprecated flags in our code base for long and I think it's time we cleaned up some of them :)

Deleted the useless flags (they do nothing even if set):
```
bigtable.max-recv-msg-size
ingester.client.max-recv-message-size
ingester.client.compress-to-ingester
distributor.compress-to-ingester
ingester.search-pending-for
ingester.reject-old-samples
ingester.reject-old-samples.max-age
ingester.validation.max-length-label-name
ingester.validation.max-length-label-value
ingester.max-label-names-per-series
```

Other flags (use `-store.index-cache-read.*`):
```
store.index-cache-size
index.memcached.*
```

```

### DIFF
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -46,9 +46,6 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Instance, "bigtable.instance", "", "Bigtable instance ID.")
 
 	cfg.GRPCClientConfig.RegisterFlags("bigtable", f)
-
-	// Deprecated.
-	f.Int("bigtable.max-recv-msg-size", 100<<20, "DEPRECATED. Bigtable grpc max receive message size.")
 }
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by bboreham** [comment]
`ingester.search-pending-for` was only removed a month ago.

Upon checking, I see we have two more of these in our config; silently removing options is a really bad idea and is not what "deprecated" means.

**Comment by gouthamve** [comment]
I'd like to disagree here. Having the flags set causes the operator to think that they still work when they don't. I prefer breaking things in this case so we can fix them. Case in point which drove me to do this (in our ksonnet):

```
-      'ingester.search-pending-for': '5m',
+      'ingester.max-transfer-retries': 60,  // Each retry is backed off by 5s, so 5mins for new ingester to come up.
```

Before we used to wait for `5m` and now we wait less than `50s` (before I made the changes above, a month after I should have).

Changing the behavior like above is a breaking change and making them explicit is much better than implicit.

bboreham:
A warning message is better than silent.
The radical change in wait time is a bug.

PR #1337, opened 2019-04-18
Adds the ability the fetch label values from the store

Kuqd:

This is required to advance https://github.com/grafana/loki/issues/453 in Loki.

I also need to add `LabelNames(metricName, start, end)` but since this is my first PR I wanted to have a quick feedback on this first.

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -188,22 +188,22 @@ func (c *store) Get(ctx context.Context, from, through model.Time, allMatchers .
 	return c.getMetricNameChunks(ctx, from, through, matchers, metricName)
 }
 
-func (c *store) validateQuery(ctx context.Context, from model.Time, through *model.Time, matchers []*labels.Matcher) (string, []*labels.Matcher, bool, error) {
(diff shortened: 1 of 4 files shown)


bboreham:
Thanks for the PR; it looks promising.  This isn't the same as the Prometheus API `LabelValues`, right?  Because that one doesn't have a metric name.  I think you should mention that in a comment, and possibly use a different name.

I would probably also split the change into two commits containing the refactor of `validateQueryTimeRange()` and then the new code.

Kuqd:
Not exactly like https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values as it requires the metric name like you said.

I'll see how I can split the commit. I actually intentionally squash rebased it to a single commit, since I was used to it for other project.

Kuqd:
Could not find a better name but I did add a documentation.

PTAL

bboreham:
Given that we have `LabelValuesForLabelName()` elsewhere, how about `LabelValuesForMetricName()` here?

bboreham:
Besides the name, LGTM

tomwilkie:
Nice work Cyril!  LGTM too.

PR #1338, opened 2019-04-19
Run all proto generation in a single container

bboreham:
Add a new phony target `generated` so we can save time by not starting and stopping a bunch of containers to run make N times.

Do the same for compilation of `exes` under the default `all` target.


tomwilkie:
LGTM!

tomwilkie:
Looks like this saves about 5s on CI, but loads more on my Mac.  Pretty happy!

PR #1341, opened 2019-04-22
Updated table deletes to ignore empty prefixes

### LINKED ISSUE #1340: Table manager ignores prefixes for retention deletes
If retention is enabled on table-manager `-table-manager.retention-deletes-enabled` & `-table-manager.retention-period=<>`, table manager tries to delete all tables in DynamoDB if the retention criterion is met irrespective of the prefixes (set by `-dynamodb.chunk-table.prefix=cortex_chunks_` & `-dynamodb.periodic-table.prefix=cortex_weekly_`).

Issue is the logic [here](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/schema_config.go#L112) will always create a `v1` schema config (if you're using legacy schema based on CLI args) whose ChunkTables `PeriodConfig` prefix is never set.

Then the [table manager](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/table_manager.go#L335) code tries to prefix match for an empty string which always matches and all DynamoDB tables get added to the `toDelete` list.

nrjpoddar:
Fixes #1340

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -333,8 +333,12 @@ func (m *TableManager) partitionTables(ctx context.Context, descriptions []Table
 		// Ensure we only delete tables which have a prefix managed by Cortex.
 		tablePrefixes := map[string]struct{}{}
 		for _, cfg := range m.schemaCfg.Configs {
-			tablePrefixes[cfg.IndexTables.Prefix] = struct{}{}
-			tablePrefixes[cfg.ChunkTables.Prefix] = struct{}{}
+			if cfg.IndexTables.Prefix != "" {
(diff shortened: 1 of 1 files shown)


nrjpoddar:
@bboreham PTAL

nrjpoddar:
@bboreham can I get your feedback on this? We have to currently run a fork of upstream cortex as without this fix the table manager tries to delete all DynamoDB tables.

nrjpoddar:
@bboreham can you merge this? I will add a second PR with unit tests as you requested but we kind of need this image asap. Thanks!

nrjpoddar:
Thanks @gouthamve

PR #1343, opened 2019-04-23
Put Distributor hash algorithm back the way it was before 922a3b989a

bboreham:
#1306 moved the hash implementation, but at the same time flipped it from FNV-1 to FNV-1a.
Put it back the way it was before, and add some comments noting the difference.

Note: without this, the effect is to re-shard all series to a different ingester, causing a spike in active series and (in shard-by-metric-name mode) breaking queries.


gouthamve:
LGTM :+1:

PR #1346, opened 2019-04-23
Add LabelNamesForMetricName for the series store

Kuqd:
This adds LabelNamesForMetricName for the series store only, I wanted to first get your feedback if this seems correct.

- I'm getting all series ids for a given metric name
- lookupSingleChunkBySeries which takes the first index query and lookup index entries then take the first index entry and parse the chunk key.
- fetch a single chunk for each series
- merge all label names possible except metric name.

I would like a special attention to that lookupSingleChunkBySeries and its usage, not sure if this is the right way to approach the problem. (May be I should loop though queries and entries until I get a result ?)

I still need to implement this for the chunk store but I'm not sure what would be the most efficient way of doing this.

The PR is rebased on my other work to include the split of the time range validation. So don't mind the two other commit they won't be part of this PR.

Thank you !

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -250,12 +250,12 @@ func (c *store) LabelNamesForMetricName(ctx context.Context, from, through model
 	if err != nil {
 		return nil, err
 	}
-	level.Debug(log).Log("Chunks in index", len(chunks))
+	level.Debug(log).Log("msg", "Chunks in index", "chunks", len(chunks))
 
 	// Filter out chunks that are not in the selected time range and keep a single chunk per fingerprint
 	filtered := filterChunksByTime(from, through, chunks)
-	filtered, keys := filterChunksByUniqueFingerPrint(filtered)
-	level.Debug(log).Log("Chunks post filtering", len(chunks))
+	filtered, keys := filterChunksByUniqueFingerprint(filtered)
+	level.Debug(log).Log("msg", "Chunks post filtering", "chunks", len(chunks))
 
 	// Now fetch the actual chunk data from Memcache / S3
 	allChunks, err := c.FetchChunks(ctx, filtered, keys)
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -263,17 +263,7 @@ func (c *store) LabelNamesForMetricName(ctx context.Context, from, through model
 		level.Error(log).Log("msg", "FetchChunks", "err", err)
 		return nil, err
 	}
-	var result []string
-	for _, c := range allChunks {
-		for _, l := range c.Metric {
-			if l.Name != model.MetricNameLabel {
-				result = append(result, string(l.Name))
-			}
-		}
-	}
-	sort.Strings(result)
-	result = uniqueStrings(result)
-	return result, nil
+	return labelNamesFromChunks(allChunks), nil
 }
 
 func (c *store) validateQueryTimeRange(ctx context.Context, from *model.Time, through *model.Time) (bool, error) {
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -228,7 +228,49 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, from, through mode
 		}
 		result = append(result, string(labelValue))
 	}
+	sort.Strings(result)
+	result = uniqueStrings(result)
+	return result, nil
+}
 
+// LabelNamesForMetricName retrieves all label names for a metric name.
+func (c *store) LabelNamesForMetricName(ctx context.Context, from, through model.Time, metricName string) ([]string, error) {
+	log, ctx := spanlogger.New(ctx, "ChunkStore.LabelNamesForMetricName")
+	defer log.Span.Finish()
+	level.Debug(log).Log("from", from, "through", through, "metricName", metricName)
+
+	shortcut, err := c.validateQueryTimeRange(ctx, &from, &through)
+	if err != nil {
+		return nil, err
+	} else if shortcut {
+		return nil, nil
+	}
+
+	chunks, err := c.lookupChunksByMetricName(ctx, from, through, nil, metricName)
+	if err != nil {
+		return nil, err
+	}
+	level.Debug(log).Log("Chunks in index", len(chunks))
+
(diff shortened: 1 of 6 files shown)


pkg/chunk/series_store.go
```diff
@@ -298,6 +351,38 @@ func (c *seriesStore) lookupChunksBySeries(ctx context.Context, from, through mo
 	return result, err
 }
 
+func (c *seriesStore) lookupSingleChunkBySeries(ctx context.Context, from, through model.Time, seriesIDs []string) ([]string, error) {
```
Kuqd: does this seems ok to do ?

bboreham:
Seems kinda gross :grimacing: 
I can't think of a better way to do it in the current schema; maybe we should think again about writing this data to the index (e.g. a metricname->labelnames row). 

Back with the code as you have it, it may be worthwhile to pass a hint down to the DB interface that you only want one chunk, and/or avoid reading the same series across multiple day buckets.

pkg/chunk/series_store.go
```diff
@@ -298,6 +351,38 @@ func (c *seriesStore) lookupChunksBySeries(ctx context.Context, from, through mo
...
+	}
+	level.Debug(log).Log("seriesIDs", len(seriesIDs))
+	var result []string
+	for _, seriesID := range seriesIDs {
+		qs, err := c.schema.GetChunksForSeries(from, through, userID, []byte(seriesID))
+		if err != nil {
+			return nil, err
+		}
+		if len(qs) > 0 {
+			entries, err := c.lookupEntriesByQueries(ctx, []IndexQuery{qs[0]})
```
gouthamve: This is not valid, you should get all the entries (chunks) using all the queries and then use the first chunk. This is because if the store can return no chunks on the first query but chunks on the second one.

Further, I think the selection of which chunk to load should happen here: https://github.com/cortexproject/cortex/pull/1346/files#diff-1d35ed616880c655543a534fed562675R198 after `filterChunksByTime`.

Kuqd: Yeah I was wondering about chunk missing too, ok let's try another approach thanks.

Kuqd:
So I tried another implementation, I'm fetching all index but only one chunk per metric fingerprint.

I had to create two implementation to make it works (series+chunk store)

Kuqd:
I've rebased my branch, I think this is good to go

pkg/chunk/chunk_store.go
```diff
@@ -228,7 +228,49 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, from, through mode
...
+		return nil, err
+	} else if shortcut {
+		return nil, nil
+	}
+
+	chunks, err := c.lookupChunksByMetricName(ctx, from, through, nil, metricName)
+	if err != nil {
+		return nil, err
+	}
+	level.Debug(log).Log("Chunks in index", len(chunks))
```
cstyan: I don't think this will ever print the `len(chunks)`, you probably need something like `Log("msg", "Chunks found", "num", len(chunks)). The go-kit logger is fun like that.

pkg/chunk/chunk_store.go
```diff
@@ -228,7 +228,49 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, from, through mode
 		}
 		result = append(result, string(labelValue))
 	}
+	sort.Strings(result)
```
cstyan: why did this need to change?

pkg/chunk/chunk_store.go
```diff
@@ -228,7 +228,49 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, from, through mode
...
+	chunks, err := c.lookupChunksByMetricName(ctx, from, through, nil, metricName)
+	if err != nil {
+		return nil, err
+	}
+	level.Debug(log).Log("Chunks in index", len(chunks))
+
+	// Filter out chunks that are not in the selected time range and keep a single chunk per fingerprint
+	filtered := filterChunksByTime(from, through, chunks)
+	filtered, keys := filterChunksByUniqueFingerPrint(filtered)
+	level.Debug(log).Log("Chunks post filtering", len(chunks))
```
cstyan: same here

pkg/chunk/chunk_store.go
Kuqd: the diff makes you think I did but in fact it's just the same 3 last line for both method, so nothing changed here.

Kuqd:
please @tomwilkie @bboreham would you be able to take a look ? From my point of view this is good to go.

pkg/chunk/chunk_store.go
```diff
@@ -228,7 +228,49 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, from, through mode
...
+	level.Debug(log).Log("msg", "Chunks post filtering", "chunks", len(chunks))
+
+	// Now fetch the actual chunk data from Memcache / S3
+	allChunks, err := c.FetchChunks(ctx, filtered, keys)
+	if err != nil {
+		level.Error(log).Log("msg", "FetchChunks", "err", err)
+		return nil, err
+	}
+	var result []string
+	for _, c := range allChunks {
```
bboreham: This bit seems likely to be inefficient - most chunks will have the same label names, but potentially thousands of unique fingerprints, so you will add thousands of copies of the same string then sort them then dedupe them.
Suggest using a `map` instead.

pkg/chunk/series_store.go
```diff
@@ -191,6 +192,72 @@ func (c *seriesStore) GetChunkRefs(ctx context.Context, from, through model.Time
...
+
+	// Now fetch the actual chunk data from Memcache / S3
+	allChunks, err := c.FetchChunks(ctx, filtered, keys)
+	if err != nil {
+		level.Error(log).Log("msg", "FetchChunks", "err", err)
+		return nil, err
+	}
+
+	var result []string
+	for _, c := range allChunks {
```
bboreham: same inefficiency as in chunk_store.

pkg/chunk/chunk_store.go
Kuqd: actually I'm doing this just above. I'm only fetching a single chunks per unique fingerprints 257:

```
	filtered, keys := filterChunksByUniqueFingerprint(filtered)
```

Kuqd: Voila !

tomwilkie:
LGTM from me!

PR #1353, opened 2019-04-28
deps: Updated client_golang and pinned to stable release.

bwplotka:
Prerequisite to https://github.com/cortexproject/cortex/pull/1332/files

Signed-off-by: Bartek Plotka <bwplotka@gmail.com>

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -6,16 +6,10 @@
   name = "github.com/aws/aws-sdk-go"
   version = "v1.15.90"
 
-# Need an override on jaeger-lib as it has a ^0.8.0 contraint on client_golang.
-[[override]]
-  name = "github.com/uber/jaeger-lib"
-  version = ">=v1.5.0"
-
-# Pin to master branch until there is a more recent stable release:
-#   https://github.com/prometheus/client_golang/issues/375
+# Need an override on client_golang as it has a ^0.8.0 contraint on uber/jaeger-lib.
 [[override]]
   name = "github.com/prometheus/client_golang"
-  branch = "master"
+  version = "v0.9.2"
 
 [[constraint]]
   name = "github.com/prometheus/alertmanager"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -772,18 +772,19 @@
   revision = "fb713f6d8239b57c646cae30f78e8b4b8861a1aa"
 
 [[projects]]
-  branch = "master"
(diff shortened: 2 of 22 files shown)


bwplotka:
I pinned jaeger-lib to 1.5.0 to not change anything else than prometheus/golang_client in this PR. The comment above jaeger-lib overrdie was lying so removed it.

bwplotka:
Any ideas what is with Cortex linter? 

I literally run this: `docker run -it -v `pwd`:/go/src/github.com/cortexproject/cortex --workdir=/go/src/github.com/cortexproject/cortex --entrypoint='dep' cortexproject/build-image:master-b645241c check` and all is fine. 

I installed exact 0.5.0 version of dep and all is the same as well?

bboreham:
So why do we need an override?

Gopkg.toml
```diff
@@ -6,16 +6,14 @@
...
 
-# Need an override on jaeger-lib as it has a ^0.8.0 contraint on client_golang.
 [[override]]
   name = "github.com/uber/jaeger-lib"
-  version = ">=v1.5.0"
+  version = "v1.5.0"
 
-# Pin to master branch until there is a more recent stable release:
-#   https://github.com/prometheus/client_golang/issues/375
+# Need an override on client_golang as it has a ^0.8.0 contraint on uber/jaeger-lib.
```
bwplotka: @bboreham see comment here to answer your question. Have you seen this? (:

bboreham:
The context of my remark is you removing a comment on the override on `jaeger-lib`.
Why do we need an override on `jaeger-lib` ?

bwplotka:
@bboreham sorry, misundersdoot you.

So client_golang needs override due to jaegor-lib's constrain on older client_golang.

But I totally don't get this: `Need an override on jaeger-lib as it has a ^0.8.0 contraint on client_golang.` ..  it does not make sense to mey, that's why removing it, let me know if I am wrong.

I initially left override though to not revert to old version by accident (as without this override any dep ensure with fresh .lock will just revert version). I am not sure how you manage deps, so maybe this is fine to remove override at all. It's not needed to client_golang that's for sure. We locked jaeger-lib on 1.5.0 for now.

Sorry for delay! Hopefully this makes sense to you. I added `make BUILD_IN_CONTAINER=false dep` makefile command as well.

tomwilkie:
@bwplotka we just moved to go modules :-)

Can you update this PR for those?

bwplotka:
Nice, will do once back on caching work, closing for now.

Issue #1354, opened 2019-04-29
Panic in querier


gouthamve:
Panic in querier:

```
[...]

PR #1355, opened 2019-04-29
Return an empty iterator when no chunks are present

gouthamve:
Fixes: https://github.com/cortexproject/cortex/issues/1354

### DIFF
--- a/pkg/chunk/encoding/bigchunk.go
+++ b/pkg/chunk/encoding/bigchunk.go
@@ -170,6 +170,10 @@ func (b *bigchunk) Size() int {
 }
 
 func (b *bigchunk) NewIterator() Iterator {
+	if len(b.chunks) == 0 {
+		return emptyIterator{}
+	}
+
 	return &bigchunkIterator{
 		bigchunk: b,
 		curr:     b.chunks[0].Iterator(),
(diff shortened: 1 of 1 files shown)


pkg/chunk/encoding/bigchunk.go
```diff
@@ -332,3 +336,12 @@ func firstAndLastTimes(c chunkenc.Chunk) (int64, int64, error) {
 	}
 	return first, last, iter.Err()
 }
+
+// emptyIterator has no samples.
+type emptyIterator struct{}
```
bboreham: How about using `NewNopIterator()` from TSDB instead of this?
AFAICS only `Batch()` would behave differently, and that might indicate a bug.

gouthamve: `NewNopIterator()` returns a `tsdb.Iterator` as seen here: https://github.com/prometheus/tsdb/blob/719b57db44f5e5dbad7d813a5645c222c3ba7647/chunkenc/chunk.go#L55-L65

Not sure how it'll be simpler to use that than just add a few lines here.

bboreham: because then you won't add any lines here?

I'm pretty sure it returns a `chunkenc.Iterator`, i.e. it fits in `curr`.

bboreham: I created #1461 to make clear what I meant.

bboreham:
Closed in favour of #1461

PR #1356, opened 2019-04-30
Expose the cache config as YAML

gouthamve:
Allow setting the fields of the cache config via YAML.


gouthamve:
Removed the WIP tag, this works.

tomwilkie:
LGTM!

PR #1361, opened 2019-05-01
Use a rate-limiter to back off on DynamoDB throttling

bboreham:
By using a rate-limiter we can slow down in proportion to the amount we are being throttled - when approaching its capacity limit DynamoDB may only reject one or two items out of a batch but previously we were not slowing down until the entire batch was rejected.

DynamoDB internal metrics refer to "Throttled events" being part of a batch and "Throttled requests" meaning the entire batch was rejected. `ThrottleLimit` in the code effectively becomes the rate at which "Throttled events" will be seen when under load, per process.

This change makes the ingester behave much more nicely when throttled, to the extent that we see hardly any `ProvisionedThroughputExceeded` errors, which means that the auto-scaling driven by those errors no longer works.  Add a metric for throttling and use that instead.

I did wonder whether to remove the `backoff` object, but it's still handling max retries and cancellation, so left it in place.

### DIFF
--- a/pkg/chunk/aws/metrics_autoscaling.go
+++ b/pkg/chunk/aws/metrics_autoscaling.go
@@ -49,6 +49,8 @@ type MetricsAutoScalingConfig struct {
 	UsageQuery       string  // Promql query to fetch write capacity usage per table
 	ReadUsageQuery   string  // Promql query to fetch read usage per table
 	ReadErrorQuery   string  // Promql query to fetch read errors per table
+
+	deprecatedErrorRateQuery string
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
@@ -61,6 +63,8 @@ func (cfg *MetricsAutoScalingConfig) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.UsageQuery, "metrics.usage-query", defaultUsageQuery, "query to fetch write capacity usage per table")
 	f.StringVar(&cfg.ReadUsageQuery, "metrics.read-usage-query", defaultReadUsageQuery, "query to fetch read capacity usage per table")
 	f.StringVar(&cfg.ReadErrorQuery, "metrics.read-error-query", defaultReadErrorQuery, "query to fetch read errors per table")
+
+	f.StringVar(&cfg.deprecatedErrorRateQuery, "metrics.error-rate-query", "", "DEPRECATED: use -metrics.write-throttle-query instead")
 }
 
 type metricsData struct {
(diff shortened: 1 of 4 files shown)


pkg/chunk/aws/metrics_autoscaling.go
```diff
@@ -57,7 +57,7 @@ func (cfg *MetricsAutoScalingConfig) RegisterFlags(f *flag.FlagSet) {
 	f.Int64Var(&cfg.TargetQueueLen, "metrics.target-queue-length", 100000, "Queue length above which we will scale up capacity")
 	f.Float64Var(&cfg.ScaleUpFactor, "metrics.scale-up-factor", 1.3, "Scale up capacity by this multiple")
 	f.StringVar(&cfg.QueueLengthQuery, "metrics.queue-length-query", defaultQueueLenQuery, "query to fetch ingester queue length")
-	f.StringVar(&cfg.ErrorRateQuery, "metrics.error-rate-query", defaultErrorRateQuery, "query to fetch error rates per table")
+	f.StringVar(&cfg.ThrottleQuery, "metrics.write-throttle-query", defaultThrottleRateQuery, "query to fetch throttle rates per table")
```
khaines: It's a good & necessary change, but it will also break any CI/CD systems. @rsteneteg FYI when we're wondering what changed after this is merged.

rsteneteg: What about keeping the old `metrics.error-rate-query` parameter but add `DEPRECATED` in the help string, and have both set `cfg.ThrottleQuery`, and then add a check during after parsing the flags, to look if any deprecated param has been set, and warn about it?

rsteneteg: if the code that looks for usage of deprecated flags also had a metric for how many deprecated flags are used, we could have an alert for it that lets us know that we need to look at the logs and see which param has been deprecated so we get time to update our CI/CD pipeline before the change actually break stuff

bboreham: Reporting via a metric seems a fine change to `pkg/util/flagext/deprecated.go` but beyond the scope of this PR.

bboreham:
Rebased and added back the deprecated flag

khaines:
@bboreham if you are happy with this, I think it's ready to merge.

bboreham:
Going over this again, I think it will break auto-scaling on reads.
My commit messages seem to say I meant to change writes but not reads.  I'll make some changes to get there.

PR #1362, opened 2019-05-01
Remove cortex_dynamo_query_retry_count histogram

bboreham:
Because the backoff is reset on any successful query, almost every data point is 0 and the histogram is not useful.


gouthamve:
LGTM!

gouthamve:
LGTM!

PR #1366, opened 2019-05-01
Support setting table tags in yaml

bboreham:
Includes drive-by refactoring since it seemed odd to have schema_config.go depend on table_manager.go, and `tags.go` works ok as a self-contained file.


gouthamve:
LGTM!

PR #1367, opened 2019-05-01
Ensure Limits Overrides metric is registered explicitly once

jtlisi:
This metric should not be registered by default. It is an important metric that signals an error and it is in a widely imported package. This code ensure this metric is not registered automatically. It allows for mutliple calls to NewOverrides instead of causing a panic to allow for tests to function correctly. However, it signals a warning because multiple override structs should not be instantiated since they act as a singleton.

### DIFF
--- a/pkg/util/validation/override.go
+++ b/pkg/util/validation/override.go
@@ -18,6 +18,10 @@ var overridesReloadSuccess = promauto.NewGauge(prometheus.GaugeOpts{
 	Help: "Whether the last overrides reload attempt was successful.",
 })
 
+func init() {
+	overridesReloadSuccess.Set(1) // Default to 1
+}
+
 // When we load YAML from disk, we want the various per-customer limits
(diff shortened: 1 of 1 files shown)


bboreham:
Am I understanding the problem correctly: the metric is designed to report `1` on success and `0` on failure, but gets initialized to `0` on first creation.

So even with your change there is a window between creation and first tick where it will signal an error.
You could `Set(1)` on creation, or redesign the metric entirely.

gouthamve:
Hrm, I agree, setting it to `1` in `init()` makes more sense tbh.

tomwilkie:
The downside to setting it to `1` is that we're indicating success when nothing has happened.  I would have probably preferred a `limits_enables` metric that we could use in alerts to ignore this metric when they're not enabled.

That being said, thats probably too much effort.  So LGTM.

PR #1372, opened 2019-05-03
Add arguments to docs

weeco:
Still work in progress, added documentation of existing ingester flags. I intend to add all existent flags of all cortex microservices.

Due to the amount of flags I switched it to a table instead of just bullets. Are you fine with this?


weeco:
I'll close this PR as it seems there are lots of flag names and prefixing which must be fixed. Before that has happened, it is in my oppinion barely possible to create a proper and easy to understand documentation for all existent flags.

PR #1007, opened 2018-09-14
Ingester: defer converting labelPairs to Metric until flush

bboreham:
Another go at #997 , in which the code to delete blank labels was wrong.
I have added a unit test with ~a blank label~ _two_ blank labels, since I got it wrong for that case the second and third attempt.
Also added a `String()` function so `labelPairs` print out like Metrics.

PR #1377, opened 2019-05-08
 Change Chunk.Metric from a map to a slice, for performance

bboreham:
This continues what was started in #1007, so now we never create a `map` for labels in the ingester.

I started on this when I noticed that index caching was taking a lot of time in `Metric.String()`.  This PR reduces that a bit, but there are further improvements that can be made to `Labels.String()`.

While fixing up the tests I noticed we go through all kinds of indirections, from chunks to matrix to series set and back again, and I started to remove some of those.  That's a tough job though, so I stopped after a few.

Benchmarks (best of 3):
Before:
```
BenchmarkEncode      	  300000	      4137 ns/op	    2978 B/op	      10 allocs/op
BenchmarkDecode1     	   50000	     29984 ns/op	  152483 B/op	      57 allocs/op
BenchmarkDecode100   	    2000	    692953 ns/op	  619859 B/op	    5234 allocs/op
BenchmarkDecode10000 	      20	 102851651 ns/op	49305985 B/op	  543190 allocs/op
BenchmarkIndexCaching 	   30000	     60016 ns/op	   12816 B/op	     188 allocs/op
```

After:
```
BenchmarkEncode      	  500000	      3699 ns/op	    2850 B/op	       8 allocs/op
BenchmarkDecode1     	   50000	     24661 ns/op	  151576 B/op	      56 allocs/op
BenchmarkDecode100   	    2000	    543184 ns/op	  548370 B/op	    5303 allocs/op
BenchmarkDecode10000 	      20	  73501908 ns/op	40227785 B/op	  530003 allocs/op
BenchmarkIndexCaching 	   30000	     43642 ns/op	   14209 B/op	     128 allocs/op
```

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -141,9 +141,9 @@ func (c *store) PutOne(ctx context.Context, from, through model.Time, chunk Chun
 func (c *store) calculateIndexEntries(userID string, from, through model.Time, chunk Chunk) (WriteBatch, error) {
 	seenIndexEntries := map[string]struct{}{}
 
-	metricName, err := extract.MetricNameFromMetric(chunk.Metric)
-	if err != nil {
-		return nil, err
+	metricName := chunk.Metric.Get(labels.MetricName)
+	if metricName == "" {
+		return nil, fmt.Errorf("no MetricNameLabel for chunk")
 	}
 
 	entries, err := c.schema.GetWriteEntries(from, through, userID, metricName, chunk.Metric, chunk.ExternalKey())
@@ -201,7 +201,7 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, from, through mode
 		return nil, nil
 	}
 
-	queries, err := c.schema.GetReadQueriesForMetricLabel(from, through, userID, model.LabelValue(metricName), model.LabelName(labelName))
+	queries, err := c.schema.GetReadQueriesForMetricLabel(from, through, userID, metricName, labelName)
 	if err != nil {
 		return nil, err
 	}
@@ -336,7 +336,7 @@ func (c *store) lookupChunksByMetricName(ctx context.Context, from, through mode
 
 	// Just get chunks for metric if there are no matchers
 	if len(matchers) == 0 {
-		queries, err := c.schema.GetReadQueriesForMetric(from, through, userID, model.LabelValue(metricName))
+		queries, err := c.schema.GetReadQueriesForMetric(from, through, userID, metricName)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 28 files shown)


pkg/util/labels.go
```diff
@@ -0,0 +1,15 @@
+package util
+
+import (
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+// LabelsToMetric converts a Labels to Metric
+func LabelsToMetric(ls labels.Labels) model.Metric {
```
tomwilkie: We have an almost exact copy of this function here: https://github.com/cortexproject/cortex/blob/master/pkg/ingester/client/compat.go#L211, can we dedupe?

pkg/chunk/chunk_store.go
```diff
@@ -141,11 +141,7 @@ func (c *store) PutOne(ctx context.Context, from, through model.Time, chunk Chun
 func (c *store) calculateIndexEntries(userID string, from, through model.Time, chunk Chunk) (WriteBatch, error) {
 	seenIndexEntries := map[string]struct{}{}
 
-	metricName, err := extract.MetricNameFromMetric(chunk.Metric)
-	if err != nil {
-		return nil, err
-	}
-
+	metricName := chunk.Metric.Get(labels.MetricName)
```
tomwilkie: This would fail if there was no metric name - should we preserve this?  I'm not sure TBH - the empty string could be considered a valid metric name...

pkg/chunk/chunk_store_test.go
```diff
@@ -65,35 +65,40 @@ var stores = []struct {
...
-func newTestChunkStore(t *testing.T, schemaName string) Store {
+func newTestChunkStore(schemaName string) (Store, error) {
 	var storeCfg StoreConfig
 	flagext.DefaultValues(&storeCfg)
-	return newTestChunkStoreConfig(t, schemaName, storeCfg)
+	return newTestChunkStoreConfig(schemaName, storeCfg)
 }
 
-func newTestChunkStoreConfig(t *testing.T, schemaName string, storeCfg StoreConfig) Store {
+func newTestChunkStoreConfig(schemaName string, storeCfg StoreConfig) (Store, error) {
```
tomwilkie: Instead of littering this function with `if err != nil ...`, could you change `t` to a `require.TestingT`? https://godoc.org/github.com/stretchr/testify/require#TestingT

tomwilkie: (That would allow this function to be used in tests and benchmarks)

tomwilkie:
Could we have a test to confirm that labels.Labels can be unmarsalled from a JSON map?  Something in chunk_test.go that unmarshals a string into a chunk, to show we haven't broken compatibility?

bboreham:
> unmarshals a string into a chunk

a hard-coded string generated from the previous code?

pkg/chunk/chunk_store.go
bboreham: How does this relate to #1214, which suggests (a) we already enforce that everything has a metric name in Cortex and (b) in other situations you don't have a metric name.

pkg/util/labels.go
bboreham: It feels bad adding a dependency on `pkg/ingester/client` in places that aren't ingester clients.
(although I did that in a few places for `client.Fingerprint()` to avoid moving the FNV code)

However I can easily have `client.FromLabelAdaptersToMetric()` call this one.

tomwilkie:
> a hard-coded string generated from the previous code?

Yeah, I can't think of a better way!  Or download a chunk and stick it in the repo as a file.

pkg/chunk/chunk_store.go
bboreham: I added back the check.

bboreham:
I have addressed review comments, added a backwards-compatibility test and fixed the bug that it revealed.

tomwilkie:
LGTM, sorry for the delay!

PR #1378, opened 2019-05-09
Update go version to 1.12.5

tomwilkie:


gouthamve:
Can we do it two PRs one for bumping and one for updating the image? The issue is that the `update-go` branch will be deleted and we won't have a way to see what was in the image?

tomwilkie:
Sure, can do.

Issue #1381, opened 2019-05-09
TTL for object stores(S3, GCS)


sandlis:
As of now, Table Manager does the retention, which is the only component that knows about retention policy and it only handles the retention of index tables.

Retention in object stores like S3 and GCS can be done by setting a TTL.
Should this be done manually by someone before configuring object store into Cortex or should Table Manager get object store clients to set the TTL?

gouthamve:
My opinion is that we let the operator handle it and not table-manager. This is because we have a single bucket and even that bucket is created by the operator and not via table-manager.

gouthamve:
Closing this as I see this as the purview of the operator and not the table-manager.

PR #1383, opened 2019-05-10
Retention changes for boltdb and filesystem

sandlis:
Currently, retention does not delete expired data from boltdb and filesystem.
Added retention support for boltdb by implementing methods in table client.
Table manager enforces retention periodically in a goroutine by checking mtime of all the files in the configured directory.
For object stores(GCS and S3), there is another issue open [here](https://github.com/cortexproject/cortex/issues/1381) which is under discussion.

### DIFF
--- a/pkg/chunk/local/boltdb_index_client.go
+++ b/pkg/chunk/local/boltdb_index_client.go
@@ -8,6 +8,10 @@ import (
 	"os"
 	"path"
 	"sync"
+	"time"
+
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/go-kit/kit/log/level"
 
 	"github.com/etcd-io/bbolt"
 
@@ -18,8 +22,9 @@ import (
 var bucketName = []byte("index")
 
 const (
-	separator = "\000"
-	null      = string('\xff')
+	separator      = "\000"
+	null           = string('\xff')
+	dbReloadPeriod = 1 * time.Minute
 )
 
 // BoltDBConfig for a BoltDB index client.
@@ -37,6 +42,8 @@ type boltIndexClient struct {
 
 	dbsMtx sync.RWMutex
 	dbs    map[string]*bbolt.DB
+	done   chan struct{}
+	wait   sync.WaitGroup
 }
 
 // NewBoltDBIndexClient creates a new IndexClient that used BoltDB.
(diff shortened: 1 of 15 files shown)


pkg/chunk/local/boltdb_index_client.go
```diff
@@ -45,18 +52,71 @@ func NewBoltDBIndexClient(cfg BoltDBConfig) (chunk.IndexClient, error) {
...
+		select {
+		case <-ticker.C:
+			b.reload()
+		case <-b.done:
+			return
+		}
+	}
+}
+
+func (b *boltIndexClient) reload() {
```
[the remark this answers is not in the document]
sandlis: > We open any unopened dbs and add them to current set.
Do we want to do this? Not all the dbs needs to be opened I feel since queriers would be limited to only a few dbs.

We still would have to lock `dbsMtx`, for finding deleted dbs we can have read lock, and if there are any deleted dbs then we can have exclusive lock. What do you think?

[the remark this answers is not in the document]
gouthamve: Yes, just that we don't block queries unless we need to.

pkg/chunk/local/boltdb_index_client.go
```diff
@@ -8,6 +8,10 @@ import (
 	"os"
 	"path"
 	"sync"
+	"time"
+
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/go-kit/kit/log/level"
```
tomwilkie: We tend to split imports into 3 sections: stdlib, non-cortex, and cortex.  Could you put these two imports in the right place please?

pkg/chunk/local/boltdb_table_client.go
```diff
@@ -2,27 +2,45 @@ package local
...
+	"os"
+	"path/filepath"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
 )
 
-type tableClient struct{}
+type tableClient struct {
+	directory string
+}
```
tomwilkie: Question: would it make sense to have a single implementation (ie singleton) that implements both TableClient, and StorageClient?  That way DeleteTable can close the active table instance properly.

pkg/chunk/local/fs_object_client.go
```diff
@@ -70,3 +76,25 @@ func (f *fsObjectClient) getChunk(_ context.Context, decodeContext *chunk.Decode
 
 	return c, nil
 }
+
+// NewBucketClient makes a chunk.BucketClient which stores chunks as files in the local filesystem.
+func NewBucketClient(cfg FSConfig) (chunk.BucketClient, error) {
+	bucketClient, err := NewFSObjectClient(cfg)
```
tomwilkie: Can we expose the `fsObjectClient` type, and then return that?  That way we don't need the cast (or the extra constructor).

pkg/cortex/modules.go
```diff
@@ -298,7 +298,14 @@ func (t *Cortex) initTableManager(cfg *Config) error {
 		return err
 	}
 
-	t.tableManager, err = chunk.NewTableManager(cfg.TableManager, cfg.Schema, cfg.Ingester.MaxChunkAge, tableClient)
+	var bucketClient chunk.BucketClient
+	if cfg.Storage.FSConfig.Directory != "" {
```
tomwilkie: I feel like this logic is misplaced a will become brittle.  Could we always call NewBucketClient here, and have the big switch in that function no when to return nil and when to return an error?

pkg/chunk/local/boltdb_table_client.go
gouthamve: Hrm, while that would work if the same client is being used for queries and table-manager, but I don't think that is the case. I think two different clients are instantiated for store and table-manager.

tomwilkie:
LGTM, thanks Sandeep!

PR #1385, opened 2019-05-13
Make sure only series with chunks are sent

gouthamve:
Could fix https://github.com/cortexproject/cortex/issues/1354

### DIFF
--- a/pkg/ingester/ingester_test.go
+++ b/pkg/ingester/ingester_test.go
@@ -123,11 +123,15 @@ func matrixToSamples(m model.Matrix) []model.Sample {
 }
 
 func runTestQuery(ctx context.Context, t *testing.T, ing *Ingester, ty labels.MatchType, n, v string) (model.Matrix, *client.QueryRequest, error) {
+	return runTestQueryTimes(ctx, t, ing, ty, n, v, model.Earliest, model.Latest)
+}
+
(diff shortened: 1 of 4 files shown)


tomwilkie:
Panic in #1354 seems to be on the chunk store path, not the ingester path.

Also, might be easier to put this check in the querier side of this interface, so you can test without rolling new ingesters?

gouthamve:
> seems to be on the chunk store path, not the ingester path.

Not really, https://github.com/cortexproject/cortex/blob/master/pkg/querier/ingester_streaming_queryable.go#L99-L103, the ingester path uses code from that file, but its still the ingester streaming path.

> Also, might be easier to put this check in the querier side of this interface, so you can test without rolling new ingesters?

Done.

pkg/ingester/ingester_test.go
```diff
@@ -187,6 +191,32 @@ func TestIngesterAppend(t *testing.T) {
...
+		err = ing.QueryStream(req, &s)
+		require.NoError(t, err)
+
+		// Nothing should be selected.
+		require.Equal(t, 0, len(s.responses))
+	}
+
+	// Read samples back via chunk store.
+	ing.Shutdown()
+	//store.checkData(t, userIDs, testData)
```
tomwilkie: Lets not commit commented-out code.

tomwilkie:
LGTM!

PR #1390, opened 2019-05-15
Make sure ObjectClient also support bigtable-hashed

gouthamve:
This means we don't crash when users specify `store: bigtable-hashed`.


tomwilkie:
It would be good to turn this into a "proper" enum with correct marshalling, and generate these lists.  As thats more than a 1 line change, you can either do it or file an issue for it.  Up to you. 

LGTM

PR #1394, opened 2019-05-16
Add context parameter to DB interface functions

jtlisi:
This does not change the functionality of the DB in any way. It only passes the context as a parameter. I am working on an object store backend for rules files in GCS and want to pass a context variable. I also think it is useful to have the context in postgres if we want to add lower level instrumentation.

### DIFF
--- a/pkg/configs/client/client.go
+++ b/pkg/configs/client/client.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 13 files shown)


pkg/configs/client/client.go
```diff
@@ -119,20 +120,20 @@ type dbStore struct {
 // GetRules implements ConfigClient.
 func (d dbStore) GetRules(since configs.ID) (map[string]configs.VersionedRulesConfig, error) {
```
tomwilkie: Can you add a context here too please?

jtlisi: Yes, I wanted to wait til #1258 was merged so I could avoid more merge conflicts, but I'll add it here.

tomwilkie:
LGTM!

PR #1397, opened 2019-05-17
Add helm chart

cyriltovena:
This adds a helm chart using the single process binary and a stateful set.

Todo:
- [ ] Documentation
- [ ] Push to a repository
- [ ] Correct the image tag, currently we don't have a master/release tag so I used a specific tag.

You can test it with:

```bash
helm template ./k8s/helm/cortex
```

or install it with
```bash
helm install ./k8s/helm/cortex
```


Kuqd:
Will work on documentation while we figure how and where we want to package this chart. I'm also waiting for feedback on the image tag to use.

Thank you !

khaines:
I've been working on a distributed process helm chart, but I'd gladly convert my work to be an extension of this. Thank you for the submission.

k8s/helm/cortex/values.yaml
```diff
@@ -0,0 +1,165 @@
...
+    port: http-metrics
+  initialDelaySeconds: 45
+
+
+## ref: https://kubernetes.io/docs/user-guide/node-selection/
+nodeSelector: {}
+
+## ref: http://kubernetes.io/docs/user-guide/persistent-volumes/
+## If you set enabled as "True", you need :
+## - create a pv which above 10Gi and has same namespace with cortex
```
khaines: Since the statefulset is using a volume claim template when there is no existing claim specified, I think line 90 is not accurate. The only requirement I understand in this case is to have a storage class defined, as the pv & claim will be generated automatically from the template.

Kuqd: you're right the comment is outdated. I'll remove it. For the test I'm not sure but I think this was a flaky test.

khaines: Yes, you are right about the test. It was corrected in a different commit

k8s/helm/cortex/templates/pvc.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) }}
+apiVersion: v1
+kind: PersistentVolumeClaim
```
steven-sheehy: This can be removed right since the StatefulSet has the PVC template.

Kuqd:
@khaines I've updated the PR, you can either merge this one directly or close this PR with yours.

khaines:
@Kuqd thanks Kuqd... PRs need two maintainers to sign off before they can be merged, so it is eligible after another has a look.

khaines:
Going to close this older PR as helm chart development for cortex has been moved to https://github.com/cortexproject/cortex-helm-chart. There were some features in this PR that would be a great addition in the new repo and would encourage migrating them.

cyriltovena:
The current cortex chart deploy the micro-service architecture, I think for some users it would make sense to have only the single binary, however I'm not sure where we should have that.

khaines:
Yes, I agree with the need for a single binary option; however, I feel that can be done in the helm chart repo alongside the microservice topology, using a template option to switch between them as needed.

PR #1398, opened 2019-05-18
Update k8s manifests with target flags

weeco:
After https://github.com/cortexproject/cortex/pull/1262 has been merged, one has to set target flags for each microservice. I updated the k8s manifests to do so.

While doing so I noticed an inconsistency of the target flags. `alertmanager` had been used while beforehand `table-manager` and `query-frontend` were used. Therefore I also updated the alertmanager module name to achieve consistent target flags - before it's too late.


pkg/cortex/modules.go
```diff
@@ -78,7 +78,7 @@ func (m moduleName) String() string {
 	case Configs:
 		return "configs"
 	case AlertManager:
-		return "alertmanager"
+		return "alert-manager"
```
jtlisi: Why are you hyphenating here?

weeco: Please see my comment in the PR submit

weeco: Actually I believe you are right, Alertmanager is just one word. I was confused due to the variable name with an uppercased "Manager".

weeco:
Done

Issue #1403, opened 2019-05-19
Default single-process-config.yaml file is invalid


khaines:
Following the Getting Started documentation, there are these 3 steps:

+ make protos # Build all the protobuffer definitions.
+ go build ./cmd/cortex
+ ./cortex -config.file=./cmd/cortex/single-process-config.yaml

The problem is that the ./cmd/cortex/single-process-config.yaml file is invalid for the current master branch and results in this error:

+ ./cortex -config.file=./cmd/cortex/single-process-config.yaml
error loading config from ./cmd/cortex/single-process-config.yaml: yaml: unmarshal errors:
  line 44: field store not found in type ring.Config

After tracing the config structs, it appears there is field missing in the file. A pull request will be made to fix this shortly.

PR #1404, opened 2019-05-19
Updating default/example yaml file with expected ring config structure

khaines:
Small change to fix a fatal error in the default/demo yaml config for cortex. 

Addresses #1403


tomwilkie:
Nice one, thanks Ken.

PR #1409, opened 2019-05-21
Exporting metrics for alerting when there is some problem with table manager creating table or enforcing retention

sandeepsukhani:
Exporting table numbers, retention period and periodic table config for alerting.
This would be helpful to setup alerts for missing required tables or existence of expired tables not deleted by table manager.

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"math/rand"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
@@ -201,7 +200,7 @@ func (m *TableManager) loop(ctx context.Context) error {
 	ticker := time.NewTicker(m.cfg.DynamoDBPollInterval)
 	defer ticker.Stop()
 
-	if err := instrument.CollectedRequest(context.Background(), "TableManager.SyncTables", syncTableDuration, instrument.ErrorCode, func(ctx context.Context) error {
+	if err := instrument.CollectedRequest(context.Background(), "TableManager.SyncTables", instrument.NewHistogramCollector(m.metrics.syncTableDuration), instrument.ErrorCode, func(ctx context.Context) error {
 		return m.SyncTables(ctx)
 	}); err != nil {
 		level.Error(util.Logger).Log("msg", "error syncing tables", "err", err)
@@ -217,7 +216,7 @@ func (m *TableManager) loop(ctx context.Context) error {
 	for {
 		select {
 		case <-ticker.C:
-			if err := instrument.CollectedRequest(context.Background(), "TableManager.SyncTables", syncTableDuration, instrument.ErrorCode, func(ctx context.Context) error {
+			if err := instrument.CollectedRequest(context.Background(), "TableManager.SyncTables", instrument.NewHistogramCollector(m.metrics.syncTableDuration), instrument.ErrorCode, func(ctx context.Context) error {
 				return m.SyncTables(ctx)
 			}); err != nil {
 				level.Error(util.Logger).Log("msg", "error syncing tables", "err", err)
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -28,48 +28,49 @@ const (
 	bucketRetentionEnforcementInterval = 12 * time.Hour
 )
 
-var (
-	syncTableDuration = instrument.NewHistogramCollector(prometheus.NewHistogramVec(prometheus.HistogramOpts{
+type tableManagerMetrics struct {
+	syncTableDuration *prometheus.HistogramVec
+	tableCapacity     *prometheus.GaugeVec
+	createFailures    prometheus.Gauge
+	deleteFailures    prometheus.Gauge
+}
+
+func newTableManagerMetrics(r prometheus.Registerer) *tableManagerMetrics {
+	m := tableManagerMetrics{}
+	m.syncTableDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: "cortex",
 		Name:      "dynamo_sync_tables_seconds",
 		Help:      "Time spent doing SyncTables.",
 		Buckets:   prometheus.DefBuckets,
-	}, []string{"operation", "status_code"}))
-	tableCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	}, []string{"operation", "status_code"})
+
+	m.tableCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
 		Name:      "dynamo_table_capacity_units",
 		Help:      "Per-table DynamoDB capacity, measured in DynamoDB capacity units.",
 	}, []string{"op", "table"})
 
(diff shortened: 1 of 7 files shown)


pkg/chunk/table_manager.go
```diff
@@ -37,11 +43,29 @@ var (
 		Name:      "dynamo_table_capacity_units",
 		Help:      "Per-table DynamoDB capacity, measured in DynamoDB capacity units.",
 	}, []string{"op", "table"})
+	retentionPeriodConfig = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "retention_period",
```
csmarchbanks: Would you add units to this metric please?

pkg/chunk/table_manager.go
```diff
@@ -37,11 +43,29 @@ var (
...
+		Help:      "Configured retention period",
+	})
+	tableNumber = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
csmarchbanks: What do you buy from having the `extreme` and `kind` labels? The Prometheus rule of thumb is that it should be meaningful to aggregate across all labels, and I am not convinced of that right now.

pkg/chunk/table_manager.go
```diff
@@ -37,11 +43,29 @@ var (
...
+		Namespace: "cortex",
+		Name:      "table_number",
+		Help:      "Min or Max table number for chunks or index kind",
+	}, []string{"extreme", "kind"})
+	tablePeriodConfig = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
csmarchbanks: How does this metric help with alerting? Would a more useful metric be something for "oldest possible data"?

sandlis:
> Left a few comments, but I am not sure of all your use cases right now. What specific conditions would you like to alert upon by using these metrics?

The idea here is to check whether right tables exists and expired tables are deleted.
Using current time, retention period and periodic table size we can calculate which should be the current lowest table number and compare it with actual minimum table number. If difference is more than we can tolerate then we can send an alert.
I think I should remove the highest table number since it won't be adding much value.

pkg/chunk/table_manager.go
sandlis: I am going to remove extreme label. Do you think there should be different guages for chunk and index table instead of having a single guage with `kind` label?

csmarchbanks:
Would it be simpler to just have metrics for the oldest possible data, and retention. Then you could alert when the difference between those numbers is more than your tolerance. It wouldn't include the periodic table part, so your tolerance would have to include that.

Just an idea, I am curious what you think!

sandlis:
Do you mean instead of calculating expected table minimum in alert evaluation, add it to a metric and avoid pushing configuration? Sounds good to me.

csmarchbanks:
Yep, I would love to see if it simplifies this at all. Thanks!

khaines:
@sandlis is this PR still something you consider in progress and have the intention to merge? I'm trying to give some of the older PRs a scan through for anything state or no longer desired.

sandlis:
@khaines sorry I didn't get to finish this, will try to finish this soon.
Thanks!

pracucci:
@sandeepsukhani Is there still interest into this PR? Otherwise we may close it.

pkg/chunk/table_manager.go
```diff
@@ -39,11 +40,36 @@ var (
 		Name:      "dynamo_table_capacity_units",
 		Help:      "Per-table DynamoDB capacity, measured in DynamoDB capacity units.",
 	}, []string{"op", "table"})
+
+	expectedMinIndexTableNumber = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "expected_min_index_table_number",
```
pracucci: We're currently prefixing new metrics with the service name. I would suggest to add `table_manager_` prefix to all these metrics.

pkg/chunk/table_manager.go
```diff
@@ -39,11 +40,36 @@ var (
...
 )
 
 func init() {
 	prometheus.MustRegister(tableCapacity)
 	syncTableDuration.Register()
+	prometheus.MustRegister(expectedMinIndexTableNumber)
+	prometheus.MustRegister(actualMinIndexTableNumber)
+	prometheus.MustRegister(expectedMinChunkTableNumber)
+	prometheus.MustRegister(actualMinChunkTableNumber)
 }
```
pracucci: Could you do an extra effort to remove this `init()`  function at all and switch to the new pattern, where we pick `registerer Prometheus.Registerer` in the `NewTableManager()` and we register metrics on that? This way we wouldn't have table manager metrics exposed by Cortex replicas not running the table manager.

pkg/chunk/table_manager.go
pracucci: The `_number` suffix is something I've never seen for gauges. What's that?

pkg/chunk/table_manager.go
```diff
@@ -197,6 +223,90 @@ func (m *TableManager) loop() {
 	}
 }
 
+func (m *TableManager) setRetentionMetrics() error {
```
pracucci: I think this is over complicated. I believe it goes a bit against what metrics monitoring should be: expose a metric gatherest in the most easieast and safest way. Here we do replicate some of the table manager to compute values (ie. `expectedMinIndexTableNumber`) and I'm not sure it's the right approach.

_I will reach you offline to discuss on what we want to really achieve with this PR, to offer some alternative way to do it._

pkg/chunk/table_manager.go
```diff
@@ -368,9 +396,12 @@ func (m *TableManager) createTables(ctx context.Context, descriptions []TableDes
 		level.Info(util.Logger).Log("msg", "creating table", "table", desc.Name)
 		err := m.client.CreateTable(ctx, desc)
 		if err != nil {
+			m.metrics.createFailures.Add(1)
```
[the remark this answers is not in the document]
sandeepsukhani: It does sound better. Will change it. Thanks!

pkg/chunk/table_manager.go
```diff
@@ -358,17 +386,29 @@ func (m *TableManager) partitionTables(ctx context.Context, descriptions []Table
...
+	numFailures := 0
+	var firstErr error
+
 	for _, desc := range descriptions {
 		level.Info(util.Logger).Log("msg", "creating table", "table", desc.Name)
 		err := m.client.CreateTable(ctx, desc)
 		if err != nil {
-			return err
+			numFailures++
+			level.Error(util.Logger).Log("msg", "error creating table", "table", desc.Name, "err", err)
```
pracucci: This will lead to double logging (because the returned error is also logged in `loop()`). I understand you wanna to it because we will just return the first error, but I think a cleaner way to handle is:
1. Remove the log here
2. Import `tsdberrors "github.com/prometheus/prometheus/tsdb/errors"`  and use it:
  - `merr := tsdberrors.MultiError{}`
  - Then for each error, `merr.Add(err)`
  - At the end you can `return merr.Err()` (will return `nil` if no error has been registered)

What's your take?

_Same comment applies to `deleteTables()`_

sandeepsukhani: Ohh I didn't know we had something like that. Will fix it. Thanks!

PR #1413, opened 2019-05-23
Add the `ingester.` prefix back to some flags

csmarchbanks:
https://github.com/cortexproject/cortex/pull/1258 got rid of the hardcoded `ingester.` which broke an ingester upgrade for me today. Note that new flags like `ruler.consul.hostname` will not work after this.

There is only one binary now, so use that in all the k8s/ manifests components.

Changes:
* Lifecycle flags for the ingester will have the `ingester.` prefix again
* `ruler.consul.*` flags will become `consul.*`
* `ruler.ring.heartbeat-timeout` and `ruler.distributor.replication-factor` lose the ruler prefix (according to Jacob these are not needed for the ruler anyway)


pkg/ring/ring.go
```diff
@@ -61,17 +62,12 @@ type Config struct {
...
 }
 
-// RegisterFlags adds the flags required to config this to the given FlagSet with a specified prefix
+// RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
-	cfg.RegisterFlagsWithPrefix("", f)
-}
-
-// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet with a specified prefix
-func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
```
jtlisi: Removing this will break the HA ruler because it requires the replication factor to be set to 1. `ruler.ring.heartbeat-timeout` is not necessarily needed but changing the replication factor is important so two rulers don't evaluate the same rule. Also the ruler ring and ingester ring will no longer be able to use separate KV stores if need be.

For all intensive purposes the ruler ring and ingester ring are essentially unrelated. That's why totally separate sets of flags are used.

csmarchbanks: Isn't that only the case if you run an ingester and ruler in the same process? You can have separate configs for the ruler if it is running as its own microservice.

jtlisi: Yea that would be the only case. Honestly the easiest solution would be to have the ruler hard code it's own lifecycler config and the ruler will only register flags for the ring config embedded in the lifecycler

csmarchbanks: I think you will actually be running both rings in the ruler since you need to query from and write to the ingester. Third try - just defaulting the lifecycler configs to have the `ingester.` prefix if none is specified. What do you think of that?

jtlisi: That sounds good to me

tomwilkie:
Sorry about that!

Issue #1375, opened 2019-05-06
Update/enhance architecture documentation


weeco:
The architecture documentation helped me understanding how Cortex works. However while working on the deployment of Cortex I noticed there may be some parts which have not been documented yet or left out (for a reason?).

**1. Configs Deployment**

I am not sure what this deployment is supposed to do or how it interacts with the other microservices. Apparently it has another dependency on a Database (Postgres only?). 

**2. Table Manager**

Same here, what is this microservice responsible for? Do I need it when I want to use Cassandra/BigTable instead of DynamoDB?

**3. Ruler**

Is the Ruler dependency required to run Cortex? In case I do not want to use any of the alert manager features.

pracucci:
Closed by #1421 and #1945.

PR #1421, opened 2019-05-27
New architecture drawing with alertmanager and ruler

weeco:
I added a new architecture drawing, which is supposed to bring the following improvements

- Added further components such as Grafana, ruler/alertmanager and Configs API
- Is a little more granular with write and query paths between the components
- Has been created with free software (draw.io) and therefore can be edited by anyone


bzub:
Seems related to https://github.com/cortexproject/cortex/issues/1375 :)

PR #1423, opened 2019-05-28
Fix Rulers Lifecycle

jtlisi:
This fixes a bug in the ruler which caused it to panic when shutting down and not exit the ring. When I rebased the HA ruler PR onto the single binary PR I added back a channel that wasn't meant to be added.

This PR refactors the ruler shutdown to use a WaitGroup to ensure all of the workers can complete before returning. Instead of having a shutdown process for each worker, the workers will be signaled to shutdown by the scheduler.

### DIFF
--- a/pkg/ruler/ruler.go
+++ b/pkg/ruler/ruler.go
@@ -119,7 +119,6 @@ type Ruler struct {
 	notifierCfg *config.Config
 
 	scheduler *scheduler
-	workers   []worker
 	workerWG  *sync.WaitGroup
 
 	lifecycler *ring.Lifecycler
@@ -168,24 +167,22 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d 
 		}
 	}
 
-	workers := make([]worker, cfg.NumWorkers)
 	for i := 0; i < cfg.NumWorkers; i++ {
-		workers[i] = newWorker(ruler)
+		// initialize each worker in a function that signals when
+		// the worker has completed
+		ruler.workerWG.Add(1)
+		go func() {
+			w := newWorker(ruler)
+			w.Run()
+			ruler.workerWG.Done()
+		}()
 	}
 
-	ruler.workers = workers
-	go ruler.run()
+	go ruler.scheduler.Run()
+
+	level.Info(util.Logger).Log("msg", "ruler up and running")
 
 	return ruler, nil
-}
-
-// Run the ruler.
-func (r *Ruler) run() {
-	go r.scheduler.Run()
-	for _, w := range r.workers {
-		go w.Run()
-	}
-	level.Info(util.Logger).Log("msg", "ruler up and running")
 }
 
 // Stop stops the Ruler.
@@ -197,7 +194,7 @@ func (r *Ruler) Stop() {
 	}
 	r.notifiersMtx.Unlock()
 
-	level.Info(util.Logger).Log("msg", "ruler shutting down scheduler")
+	level.Info(util.Logger).Log("msg", "shutting down rules scheduler")
 	r.scheduler.Stop()
 
 	level.Info(util.Logger).Log("msg", "waiting for workers to finish")
--- a/pkg/ruler/ruler.go
+++ b/pkg/ruler/ruler.go
@@ -120,6 +120,7 @@ type Ruler struct {
 
 	scheduler *scheduler
 	workers   []worker
+	workerWG  *sync.WaitGroup
 
 	lifecycler *ring.Lifecycler
 	ring       *ring.Ring
@@ -148,6 +149,7 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d 
 		alertURL:    cfg.ExternalURL.URL,
 		notifierCfg: ncfg,
 		notifiers:   map[string]*rulerNotifier{},
+		workerWG:    &sync.WaitGroup{},
 	}
 
 	ruler.scheduler = newScheduler(rulesAPI, cfg.EvaluationInterval, cfg.EvaluationInterval, ruler.newGroup)
@@ -187,6 +189,7 @@ func (r *Ruler) run() {
 }
 
 // Stop stops the Ruler.
+// Each function of the ruler is terminated before leaving the ring
 func (r *Ruler) Stop() {
 	r.notifiersMtx.Lock()
 	for _, n := range r.notifiers {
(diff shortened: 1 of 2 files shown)


csmarchbanks:
I don't really like each worker reaching back into the `ruler` to change the workgroup. What about adding to/subtracting from the workgroup in `ruler.run()`?

jtlisi:
> I don't really like each worker reaching back into the `ruler` to change the waitgroup. What about adding to/subtracting from the workgroup in `ruler.run()`?

You can add to the waitgroup there, but, you will still need to subtract from the waitgroup when the worker exits.
```
func (w *worker) Run() {
	defer w.ruler.workerWG.Done()
```

We can pass the waitgroup to worker.Run(). However, the worker already carries a reference to the ruler as is, so it doesn't seem necessary to pass the the reference to the waitgroup unless we plan on removing the ruler from the worker struct later.

csmarchbanks:
I was thinking something like:
```
workerWG.Add(len(r.workers))
for _, w := range r.workers {
    go func() {
        w.Run()
        workerWG.Done()
    }()
}
```
The extra anonymous function is annoying though. I think I am ok approving this as is, but curious what you think.

Also, I would love to remove the ruler from the worker struct at some point. It doesn't seem like a worker should have to know about the entire ruler.

jtlisi:
I think the anonymous function would work great. I also think removing the ruler from the worker makes sense. Instead the worker can evaluate the rule itself. The only change required would be initializing the worker with a reference to the ring instead of the ruler.

pkg/ruler/ruler.go
```diff
@@ -171,36 +172,40 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
...
-func (r *Ruler) run() {
-	go r.scheduler.Run()
-	for _, w := range r.workers {
-		go w.Run()
+	go ruler.scheduler.Run()
+	for i := range workers {
+		// initialize each worker in a function that signals when
+		// the worker has completed
+		go func(w worker) {
+			ruler.workers.Add(1)
```
csmarchbanks: `Add` should always be called outside of the goroutine being created. Otherwise, you can end up with some weird race conditions where workers could be created after `Wait()` returns.

csmarchbanks: The race condition I am describing is probably also why the tests are failing.

jtlisi: Good catch, it should be good now

pkg/ruler/ruler.go
```diff
@@ -171,36 +172,40 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
...
-	return ruler, nil
-}
-
-// Run the ruler.
-func (r *Ruler) run() {
-	go r.scheduler.Run()
-	for _, w := range r.workers {
-		go w.Run()
+	go ruler.scheduler.Run()
+	for i := range workers {
```
csmarchbanks: Could this and the loop above be simplified to something like:
```
ruler.workers.Add(cfg.NumWorkers)
for i := 0; i < cfg.NumWorkers; i++ {
    go func() {
        newWorker(ruler).Run()
        ruler.workers.Done()
    }()
}
```
It doesn't look like we actually need the slice of workers anymore.

jtlisi: True, I updated and removed the slice entirely

csmarchbanks: :+1:

csmarchbanks:
This just needs another Maintainer approval. Perhaps @tomwilkie or @gouthamve could take a look?

pkg/ruler/ruler.go
```diff
@@ -166,41 +167,43 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
...
 // Stop stops the Ruler.
+// Each function of the ruler is terminated before leaving the ring
 func (r *Ruler) Stop() {
 	r.notifiersMtx.Lock()
 	for _, n := range r.notifiers {
 		n.stop()
 	}
 	r.notifiersMtx.Unlock()
 
+	level.Info(util.Logger).Log("msg", "ruler shutting down scheduler")
```
gouthamve: Do we need the `ruler` here? The logs are coming from a ruler right?

pkg/ruler/ruler.go
```diff
@@ -119,7 +119,7 @@ type Ruler struct {
 	notifierCfg *config.Config
 
 	scheduler *scheduler
-	workers   []worker
+	workers   *sync.WaitGroup
```
gouthamve: Hrm, the naming is a little off here. I would expect this would have the workers. `workersWG` is better?

pkg/ruler/ruler.go
csmarchbanks: It might be nice for it to say `ruler` for single binary mode. The log line should say `ruler.go`, but that has been broken in the past.

pkg/ruler/ruler.go
```diff
@@ -119,7 +119,7 @@ type Ruler struct {
 	notifierCfg *config.Config
 
 	scheduler *scheduler
-	workers   []worker
+	workerWG  *sync.WaitGroup
```
tomwilkie: Nit: you don't need the pointer.

pkg/ruler/ruler.go
```diff
@@ -148,6 +148,7 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
 		alertURL:    cfg.ExternalURL.URL,
 		notifierCfg: ncfg,
 		notifiers:   map[string]*rulerNotifier{},
+		workerWG:    &sync.WaitGroup{},
```
tomwilkie: Nit: zero value is fine.

pkg/ruler/ruler.go
```diff
@@ -166,41 +167,43 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
 		}
 	}
 
-	workers := make([]worker, cfg.NumWorkers)
 	for i := 0; i < cfg.NumWorkers; i++ {
-		workers[i] = newWorker(ruler)
+		// initialize each worker in a function that signals when
+		// the worker has completed
+		ruler.workerWG.Add(1)
```
tomwilkie: Nit: I'd do `ruler.workerWG.Add(cfg.NumWorkers)` before the loop.

tomwilkie:
Got to this too late & had a few nits, sorry Jacob.

pkg/ruler/ruler.go
tomwilkie: Nit: don't need the pointer

pkg/ruler/ruler.go
tomwilkie: Nit: zero value is enough

pkg/ruler/ruler.go
tomwilkie: Nit: I'd do `ruler.workerWG.Add(cfg.NumWorkers)` before the loop.

PR #1424, opened 2019-05-28
Diskcache Windows Support

### LINKED ISSUE #1336: Cortex does not compile for Windows
refer [issue](https://github.com/grafana/loki/issues/193)

sh0rez:
This PR adds Windows `mmap()` support, make [diskcache.go](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/cache/diskcache.go) (and thus Cortex) to compile on Windows. Closes #1336 

Previously, the file had a dependency on `unix.Mmap()`. This has been refactored into a separate package `util/mmap` and using built constrains either the original `golang.org/x/sys/unix` implementation or the windows specific `golang.org/x/sys/windows` one is used. 

The Windows implementation is based on https://github.com/edsrzf/mmap-go/blob/master/mmap_windows.go and has been slimmed down and refactored to keep API compatibility with the original `unix.Mmap` implementation.

The tests succeed when being run natively on Windows, so it looks like as if cortex is now Windows compatible .. in case somebody ever needs this.

The CircleCI Pipeline has been extended by a new target `build-windows`, which builds only the windows binary, just to see whether it works. No container, just the binary for now.

While it might not add a lot of value to cortex, it allows dependent projects to compile on Windows, for example Loki's `promtail` grafana/loki#567

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,6 +100,12 @@ jobs:
         paths:
         - images/
 
+    - run:
+        name: Build on Windows
+        command: |
+          export GOOS=windows
+          make BUILD_IN_CONTAINER=false exes
+
   deploy:
     <<: *defaults
     steps:
--- a/pkg/chunk/cache/diskcache.go
+++ b/pkg/chunk/cache/diskcache.go
@@ -16,9 +16,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/tsdb/fileutil"
-	"golang.org/x/sys/unix"
 
 	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/mmap"
 )
 
 var (
@@ -102,7 +102,7 @@ func newDiskcache(cfg DiskcacheConfig) (*Diskcache, error) {
 		return nil, errors.Wrap(err, "stat")
 	}
 
-	buf, err := unix.Mmap(int(f.Fd()), 0, int(info.Size()), unix.PROT_READ|unix.PROT_WRITE, unix.MAP_SHARED)
+	buf, err := mmap.Mmap(f.Fd(), 0, int(info.Size()), mmap.READ|mmap.WRITE, mmap.SHARED)
 	if err != nil {
 		f.Close()
 		return nil, err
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -100,6 +100,18 @@ jobs:
         paths:
         - images/
 
+
+  build-windows:
```
bboreham: This would need to be included in `workflows` above, if you want it to run each time.
It's probably faster just to add the `GOOS=windows` step to `build` after `save_cache`.

Makefile
```diff
@@ -118,7 +119,7 @@ else
 exes: $(EXES)
 
 $(EXES):
-	CGO_ENABLED=0 go build $(GO_FLAGS) -o $@ ./$(@D)
+	CGO_ENABLED=0 GOOS=$(GOOS) go build $(GO_FLAGS) -o $@ ./$(@D)
```
bboreham: why not just `export GOOS=windows` before running `make` then you don't need these two lines.

sh0rez:
> @bboreham 
> Since it seems to be adapted from a cross-platform library, can we just use that library from Cortex?

This is correct, all of the low level code has been taken from [mmap-go](/edsrzf/mmap-go). However, the library includes a lot of (unnecessary?) boilerplate, uses a different API and seems to be unmaintained (PR's unreviewed for multiple months, etc). I thought depending on this one would not be beneficial to cortex, so it seemed reasonable to just take the required windows low level code and build a minimal wrapper around it, to keep `unix.Mmap()` API compatibility.

I am no Windows developer either, so I cannot really tell whether it is a good implementation .. but it passed cortex' tests on windows and seems to work.

Using the library nevertheless is okay for me as well, if you prefer this.

jtlisi:
I'm in the same boat in that I don't know much about the underlying windows implementation. I think this looks fine to me at least for the unix implementation.

Another option would be to keep the unix disk cache as is and just create a skeleton for a windows disk_cache, with build tags, that returns an `not supported` error if enabled. I think that would be the easiest approach since the disk cache is not a commonly used feature to begin with. If down the line someone deploying cortex on windows wants to use the diskcache they can provide their own implementation later.

tomwilkie:
I'd even support the removal of the disk cache - we don't use it any more.  Lets mark it as a failed experiment.

sh0rez:
> @tomwilkie 
> I'd even support the removal of the disk cache - we don't use it any more. Lets mark it as a failed experiment.

This would arguably be the simplest approach .. if diskcache is reduntant I will file another PR to remove it

PR #1426, opened 2019-05-29
Add GCS object client specific timeout

gouthamve:
We're seeing issues where GCS is rate-limiting our requests. Now the SDK keeps retrying the requests until the context is cancelled. And in this case the context is cancelled after 2mins, which means the query is stuck for 2mins and this causes a huge query queue pileup.

By setting a smaller timeout, we will fail the queries early atleast.

### DIFF
--- a/pkg/chunk/gcp/gcs_object_client.go
+++ b/pkg/chunk/gcp/gcs_object_client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"io/ioutil"
+	"time"
 
 	"cloud.google.com/go/storage"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -21,14 +22,16 @@ type gcsObjectClient struct {
...
+	BucketName      string        `yaml:"bucket_name"`
+	ChunkBufferSize int           `yaml:"chunk_buffer_size"`
+	RequestTimeout  time.Duration `yaml:"request_timeout"`
 }
 
 // RegisterFlags registers flags.
 func (cfg *GCSConfig) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.BucketName, "gcs.bucketname", "", "Name of GCS bucket to put chunks in.")
 	f.IntVar(&cfg.ChunkBufferSize, "gcs.chunk-buffer-size", 0, "The size of the buffer that GCS client for each PUT request. 0 to disable buffering.")
+	f.DurationVar(&cfg.RequestTimeout, "gcs.request_timeout", 0, "The duration after which the requests to GCS should be timed out.")
```
tomwilkie: This should probably be a hyphen for consistency.

tomwilkie:
LGTM with 1 minor nit.

PR #1427, opened 2019-05-29
Use prometheus/tsdb in ingester.

tomwilkie:
This PR is the first attempt at using prometheus/tsdb in the ingester.

There is lots missing/TODO:
- [ ] Add series limits to TSDB to match existing Cortex limits.
- [ ] Upstream TSDB changes for exposing chunks on series interface.
- [ ] Figure out how to do per user metrics for TSDB (number of series, ingestion rate etc).
- [ ] Ensure TSDB errors for things like out-of-order etc are being handled correctly.
- [ ] Figure out how to handle transfers for TSDB.
- [ ] Figure out how to do upgrades.
- [ ] Implement flushing TSDB to object store
- [ ] Allow index entries to point to sub-objects
- [ ] Benchmark Bigtable vs in-object-store index.

### DIFF
--- a/vendor/github.com/prometheus/tsdb/querier.go
+++ b/vendor/github.com/prometheus/tsdb/querier.go
@@ -53,6 +53,9 @@ type Series interface {
 
 	// Iterator returns a new iterator of the data of the series.
 	Iterator() SeriesIterator
+
+	// Chunks returns a copy of the compressed chunks that make up this series.
+	Chunks() []chunks.Meta
 }
 
 // querier aggregates querying results from time blocks within
@@ -332,15 +335,6 @@ func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings, 
 	it := index.Intersect(its...)
 
 	for _, n := range notIts {
-		if _, ok := n.(*index.ListPostings); !ok {
-			// Best to pre-calculate the merged lists via next rather than have a ton
-			// of seeks in Without.
-			pl, err := index.ExpandPostings(n)
-			if err != nil {
-				return nil, err
-			}
-			n = index.NewListPostings(pl)
-		}
 		it = index.Without(it, n)
 	}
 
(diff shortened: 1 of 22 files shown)


codesome:
I can pick TSDB changes to be made 1 by 1, starting with series limit.

codesome:
Limit series in TSDB: https://github.com/prometheus/tsdb/pull/617

cstyan:
`Figure out how to do per user metrics for TSDB (number of series, ingestion rate etc).` Could we achieve this by just having a separate TSDB data dir per user ID?

codesome:
@cstyan Yes, after I opened the PR in TSDB to have limits on number of series, it looks unnecessary to have it in TSDB. We should be able to handle that in cortex (with or without TSDB per user ID). 
Also, wouldn't it be too many database instances to handle if we have TSDB per user ID?

vendor/github.com/prometheus/tsdb/querier.go
```diff
@@ -53,6 +53,9 @@ type Series interface {
 
 	// Iterator returns a new iterator of the data of the series.
 	Iterator() SeriesIterator
+
+	// Chunks returns a copy of the compressed chunks that make up this series.
+	Chunks() []chunks.Meta
```
bwplotka: Where is this coming from? I cannot see this on tip of TSDB :thinking: nor 0.8.0 which `go mod` vendors here. 

Is this manually modifed TSDB code in vendor? Cannot see any `replace` directive on `go.mod` as well so you don't use fork as well...

tomwilkie: I just added it here; opening a TSDB PR now.

tomwilkie: https://github.com/prometheus/tsdb/pull/659

tomwilkie:
Closing in favour of #1695.

PR #1430, opened 2019-05-29
Encode write-dedupe keys as they are created

bboreham:
This simplifies the code and avoids some allocations.

I suspect it can be simplified further because it's just the SHA that needs encoding, but if we change the encoding that would trigger extra work on rollout as the existing cache wouldn't match.


tomwilkie:
LGTM, nice one.

PR #1432, opened 2019-05-30
Make series IDs backwards-compatible again

bboreham:
#1377 caused an unforseen change in the formatting of label strings, which in turn means that the SHA is different so the series ID changes.  It's more efficient to continue with the same IDs across an
upgrade.

I noticed there was an unused function `metricSeriesID()` and a test for backwards-compatibility which would have caught this mistake, so have pressed them back into service.

### DIFF
--- a/pkg/chunk/schema.go
+++ b/pkg/chunk/schema.go
@@ -131,7 +131,7 @@ func (s schema) GetLabelEntryCacheKeys(from, through model.Time, userID string, 
 		key := strings.Join([]string{
 			bucket.tableName,
 			bucket.hashKey,
-			string(sha256bytes(labels.String())),
+			string(labelsSeriesID(labels)),
 		},
 			"-",
 		)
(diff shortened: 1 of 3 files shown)


gouthamve:
I am not sure if we want to maintain this given the impact will be diminished overtime. Having said that if we're using the series-store, the impact on the index will be 2x on the database for queries whose time-range overlaps that of the deploy here.

I have a slight preference for not doing this and just going with the new series-ids, but don't have a strong preference. This PR looks good otherwise!

bboreham:
Just to be clear, it's the function `labelsString()` you don't want to maintain?
It's broadly the same as `labels.Labels.String()` from Prometheus, apart from checking `MetricName` and some tricks to reduce memory allocations.

tomwilkie:
I'm happy to go with this change; 2x more entries on a migration day will hurt anyone on DynamoDB.  

I minor nit: I feel like the test is too brittle though - can we test the schema directly, and not a function the schema uses which might accidentally get refactored away?

LGTM so no one is blocked on me.

khaines:
we haven't rolled #1377 into our prod environments yet, so (selfishly) we are not affected by the dynamo impact in this.

LGTM also

bboreham:
I agree about having a better test, but I'm not very clear how to go about that.
I'd like to merge the code change and leave improving the test to later.

tomwilkie:
Go for it @bboreham - please file an issue to improve the test so we don't forget.

PR #1434, opened 2019-05-31
frontend: Make sure no samples are repeating

gouthamve:

### DIFF
--- a/pkg/querier/frontend/query_range.go
+++ b/pkg/querier/frontend/query_range.go
@@ -295,6 +295,13 @@ func matrixMerge(resps []*APIResponse) []SampleStream {
 					Labels: stream.Labels,
 				}
 			}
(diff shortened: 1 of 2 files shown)


bboreham:
Can you clarify what causes the issue?
If it comes from sharding queries into days, maybe subtract one milisecond from the end-time of each shard except the last one so the time slices don't overlap?

gouthamve:
It comes from the splitting of queries and how extents work. See: https://github.com/cortexproject/cortex/blob/master/pkg/querier/frontend/results_cache.go#L170-L175 and https://github.com/cortexproject/cortex/blob/master/pkg/querier/frontend/results_cache.go#L182-L187

I think it comes from here too, and maybe even from the splitting by day. I think the fix here is cleaner than manipulating timestamps to be off by a ms.

tomwilkie:
LTGM!

PR #1435, opened 2019-05-31
 Avoid duplicate label index writes by combining LabelEntryCacheKeys and LabelWriteEntries

### LINKED ISSUE #1428: Avoid writing label index entries twice each day
The [index write dedupe code](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/series_store.go#L399-L408) checks if _any_ buckets are missing, and if so writes _all_ buckets.

This means that, when a series continues from one day to the next, when writing the chunk that spans the day boundary all the previous day's entries are written again.

It may be this way because the interfaces don't talk about buckets, but this seems like a significant win for those of us who are iops limited.

bboreham:
Fixes #1428 

Move the write dedupe cache lookup down to `schema` so the labels for each bucket can be omitted if they've already been done.  This avoids re-writing yesterday's labels when we write a chunk that spans the day boundary.

It's now calling the cache separately for each bucket, but I felt the loss of clarity wasn't worth optimising this.

### DIFF
--- a/pkg/chunk/schema.go
+++ b/pkg/chunk/schema.go
@@ -34,9 +34,9 @@ type Schema interface {
 	GetWriteEntries(from, through model.Time, userID string, metricName string, labels labels.Labels, chunkID string) ([]IndexEntry, error)
 
 	// Should only be used with the seriesStore. TODO: Make seriesStore implement a different interface altogether.
-	GetLabelWriteEntries(from, through model.Time, userID string, metricName string, labels labels.Labels, chunkID string) ([]IndexEntry, error)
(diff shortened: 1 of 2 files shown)


tomwilkie:
Not a big fan of schema having side effects; I've always considered the schemas to be deterministic.

Perhaps having `GetLabelWriteEntries` return the cache keys (as this PR does) and having the filtering/checking done in the caller would be cleaner?  We would need to comment on the interface that the write entries and cache entries need to be guaranteed to be the same length.

bboreham:
Changed as you suggested

bboreham:
I ran this up in our staging environment, and the rate of index writes after midnight halved.

PR #1439, opened 2019-06-02
Add kustomization.yaml

bzub:
This will allow users to deploy cortex demo manifests with modifications layered on top.

```
kustomize build github.com/cortexproject/cortex | kubectl apply -f -
```

Example of this elsewhere: https://github.com/coreos/kube-prometheus/blob/master/kustomization.yaml


tomwilkie:
Thanks @bzub!

Would it be possible to put this in a subdirectory?  Does that work for kustomise?  I'd like to put all this kinda stuff in a `production` folder.

Also, could you add something to the README or perhaps a new doc with the command from the PR?

bzub:
@tomwilkie Yah it can go anywhere in the repo, just have to change the relative path to the manifests. Docs would be good, I'll add those.

bzub:
Hey @tomwilkie. I was mistaken about where the `kustomization.yaml` can live. It should live in a parent directory of the manifests.

I've updated this PR with `kustomization.yaml` and a README in the `k8s` directory. I've set things up so that when a user runs the commands in the README a `cortex` namespace is created and populated with the resources in the manifests.

Finally, I've updated the `kustomization.yaml` so it uses a certain image tag for all the `quay.io/cortexproject/cortex` containers. In the future this image tag could be updated with each release to the release version of cortex. It also serves as a good example of how users can create their own `kustomization.yaml` to dictate the version of cortex that's deployed.

Let me know what you think.

khaines:
@tomwilkie tagging you, since the latest updates here are in response to some of your comments. Is this PR good for a review? I don't use kustomize, so I can't give it a valuable review.

pracucci:
@bzub I would like to revive this PR if you still have interest into it. If so, the next steps could be:

1. Rebase `master`
2. Change the image tag from `master` to the latest stable release `v0.6.0`
3. Add a note to `RELEASE.md` (in the "Prepare your release" section) to mention to update the version to the kustomize file too

pracucci:
_I'm going to close this PR due to inactivity. Feel free to reopen it in case you're still interested into working on this._

PR #1440, opened 2019-06-03
Fix bucket bounds on query metrics

gouthamve:
Previously we aren't having as many buckets as we wanted to have in the comments.


bboreham:
Very similar code here: https://github.com/cortexproject/cortex/blob/master/pkg/ingester/ingester.go#L62-L75

Maybe this style of comment is clearer?  https://github.com/cortexproject/cortex/blob/master/pkg/ingester/flush.go#L37

PR #1441, opened 2019-06-03
Allow the query-frontend code to proxy queries to 'vanilla' Prometheus.

tomwilkie:
This allows us to get all the benefits of the query-frontend (caching, splitting, aligning, retires etc) with OSS Prometheus.

eg
```
$ ./cortex --config.file=./docs/prometheus-frontend.yml  -frontend.downstream-url=http://localhost:9090
```


pkg/querier/frontend/frontend.go
```diff
@@ -99,6 +103,7 @@ func New(cfg Config, log log.Logger, limits *validation.Overrides) (*Frontend, e
 		log:    log,
 		queues: map[string]chan *request{},
 	}
+	f.cond = sync.NewCond(&f.mtx)
```
tomwilkie: This is just a move, used to be at bottom of the function.  Prefer it here, but can do in different PR if necessary.

PR #1442, opened 2019-06-03
Fix the version of protobuf and pin tools in the build image.

### LINKED ISSUE #1370: We don't know which gogoproto we are using
We don't pin it in the build container, so as far as I can tell we are using whichever one was in the Docker build cache for whoever built the build container.

This is a bit scary for debugging / reproducability.

tomwilkie:
Fixes #1370

### DIFF
--- a/build-image/Dockerfile
+++ b/build-image/Dockerfile
@@ -8,16 +8,14 @@ RUN curl -fsSLo shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt
 	echo "b1925c2c405458811f0c227266402cf1868b4de529f114722c2e3a5af4ac7bb2  shfmt" | sha256sum -c && \
 	chmod +x shfmt && \
 	mv shfmt /usr/bin
-RUN go get -tags netgo \
+RUN GO111MODULE=on go get -tags netgo \
 		github.com/fzipp/gocyclo \
 		golang.org/x/lint/golint \
-		github.com/kisielk/errcheck \
-		github.com/mjibson/esc \
-		github.com/client9/misspell/cmd/misspell \
-		github.com/jteeuwen/go-bindata/go-bindata \
-		github.com/golang/protobuf/protoc-gen-go \
-		github.com/gogo/protobuf/protoc-gen-gogoslick \
-		github.com/gogo/protobuf/gogoproto && \
+		github.com/kisielk/errcheck@v1.2.0 \
+		github.com/client9/misspell/cmd/misspell@v0.3.4 \
+		github.com/golang/protobuf/protoc-gen-go@v1.3.1 \
+		github.com/gogo/protobuf/protoc-gen-gogoslick@v1.2.1 \
+		github.com/gogo/protobuf/gogoproto@v1.2.1 && \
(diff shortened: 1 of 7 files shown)


build-image/Dockerfile
```diff
@@ -8,16 +8,14 @@ RUN curl -fsSLo shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt
...
 		golang.org/x/lint/golint \
-		github.com/kisielk/errcheck \
-		github.com/mjibson/esc \
-		github.com/client9/misspell/cmd/misspell \
-		github.com/jteeuwen/go-bindata/go-bindata \
-		github.com/golang/protobuf/protoc-gen-go \
-		github.com/gogo/protobuf/protoc-gen-gogoslick \
-		github.com/gogo/protobuf/gogoproto && \
+		github.com/kisielk/errcheck@v1.2.0 \
+		github.com/client9/misspell/cmd/misspell@v0.3.4 \
```
bboreham: why did we drop `esc`?

build-image/Dockerfile
```diff
@@ -8,16 +8,14 @@ RUN curl -fsSLo shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt
...
+		github.com/golang/protobuf/protoc-gen-go@v1.3.1 \
+		github.com/gogo/protobuf/protoc-gen-gogoslick@v1.2.1 \
```
bboreham: why did we drop `go-bindata` ?

pkg/chunk/storage/caching_index_client.pb.go
```diff
@@ -420,13 +420,8 @@ func (this *ReadBatch) String() string {
...
 		return "nil"
 	}
-	repeatedStringForEntries := "[]Entry{"
-	for _, f := range this.Entries {
-		repeatedStringForEntries += strings.Replace(strings.Replace(f.String(), "Entry", "Entry", 1), `&`, ``, 1) + ","
-	}
-	repeatedStringForEntries += "}"
 	s := strings.Join([]string{`&ReadBatch{`,
-		`Entries:` + repeatedStringForEntries + `,`,
+		`Entries:` + strings.Replace(strings.Replace(fmt.Sprintf("%v", this.Entries), "Entry", "Entry", 1), `&`, ``, 1) + `,`,
```
bboreham: This change is removing https://github.com/gogo/protobuf/pull/557 - do we want to go backwards?

build-image/Dockerfile
tomwilkie: We don't use it AFAICT.

build-image/Dockerfile
tomwilkie: We don't use it AFAICT.

pkg/chunk/storage/caching_index_client.pb.go
tomwilkie: There is no gogo release with that fix in it. 

We're not using the String function for any production use (I don't think displaying the proto on /ring is affected adversely.)

tomwilkie:
@bboreham do those responses assuage your concerns?

PR #1443, opened 2019-06-04
Fix wait time on DynamoDB throttling

bboreham:
Slight bug in #1361: the `UnprocessedItems` field that comes back is a map from table to items, so we need to count all the items and wait that long, rather than just waiting one unit per table.

### DIFF
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -271,7 +271,7 @@ func (a dynamoDBStorageClient) BatchWrite(ctx context.Context, input chunk.Write
 
 		// If there are unprocessed items, retry those items.
 		unprocessedItems := dynamoDBWriteBatch(resp.UnprocessedItems)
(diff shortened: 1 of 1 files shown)


pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -270,9 +270,10 @@ func (a dynamoDBStorageClient) BatchWrite(ctx context.Context, input chunk.Write
 		}
 
 		// If there are unprocessed items, retry those items.
-		if unprocessedItems := resp.UnprocessedItems; unprocessedItems != nil && dynamoDBWriteBatch(unprocessedItems).Len() > 0 {
-			logWriteRetry(ctx, dynamoDBWriteBatch(unprocessedItems))
-			a.writeThrottle.WaitN(ctx, len(unprocessedItems))
+		unprocessedItems := dynamoDBWriteBatch(resp.UnprocessedItems)
+		if unprocessedItems != nil && len(unprocessedItems) > 0 {
```
csmarchbanks: `len(unprocessedItems) > 0` covers the `nil` case as well

Issue #1382, opened 2019-05-09
Retention with multiple store configs


sandeepsukhani:
As of now, Table Manager creates Table Client only for newest config as seen here:
https://github.com/cortexproject/cortex/blob/master/cmd/table-manager/main.go#L47-L62
Retention code calls deleteTable only on this table client for newest config.
If someone would use different config for different time span then Table Manager would not delete expired index tables from previous configs.
Should this be fixed or its never going to be a case where people would have multiple configs?

bboreham:
It should be fixed.
It should even support cases like "started on DynamoDB, moved to Bigtable".
Not just retention, but all the provisioning stuff too.

friedrichg:
Not valid anymore, Chunks was deprecated and removed

PR #1446, opened 2019-06-05
TableClient for each PeriodConfig in Schema

codesome:
Attempt to fix https://github.com/cortexproject/cortex/issues/1382

This PR creates a TableClient for all the past configs and deletes tables from all the clients in `SyncTables`.

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -2,6 +2,7 @@ package chunk
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"sort"
@@ -123,9 +124,14 @@ type TableManager struct {
 	bucketClient BucketClient
 }
 
-// NewTableManager makes a new TableManager
+// NewTableManager makes a new TableManager.
+// The i'th TableClient in 'tableClients' should correspond to the i'th PeriodConfig in schemaCfg.Config.
+// The TableClient for the last client should not be nil.
 func NewTableManager(cfg TableManagerConfig, schemaCfg SchemaConfig, maxChunkAge time.Duration, tableClients []TableClient,
 	objectClient BucketClient) (*TableManager, error) {
+	if len(schemaCfg.Configs) != len(tableClients) {
+		return nil, errors.New("Mismatch in number of PeriodConfig and TableClient")
+	}
 	return &TableManager{
 		cfg:          cfg,
 		schemaCfg:    schemaCfg,
@@ -202,6 +208,9 @@ func (m *TableManager) bucketRetentionLoop() {
 // not and update those that need it.  It is exposed for testing.
 func (m *TableManager) SyncTables(ctx context.Context) error {
 	for i, client := range m.clients {
+		if client == nil {
+			continue
+		}
 		lastClient := i == len(m.clients)-1
 		expected := m.calculateExpectedTables(i)
 		if lastClient {
@@ -245,9 +254,15 @@ func (m *TableManager) calculateExpectedTables(idx int) []TableDesc {
 		return result
 	}
 
-	// Collect tables from the first config till idx'th config.
-	for i, config := range m.schemaCfg.Configs[:idx+1] {
+	// Use only the idx'th config for idx'th client.
+	startIdx := idx
+	if idx == len(m.schemaCfg.Configs)-1 {
+		// For the last client, collect tables from all the configs.
+		startIdx = 0
+	}
+	for i, config := range m.schemaCfg.Configs[startIdx : idx+1] {
 		if config.From.Time.Time().After(mtime.Now()) {
+			sort.Sort(byName(result))
 			return result
 		}
 		if config.IndexTables.Period == 0 { // non-periodic table
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -114,7 +114,7 @@ func (cfg *ProvisionConfig) RegisterFlags(argPrefix string, f *flag.FlagSet) {
 
 // TableManager creates and manages the provisioned throughput on DynamoDB tables
 type TableManager struct {
-	client       TableClient
+	clients      []TableClient // TableClient for each config in SchemaConfig.Configs, in same order.
 	cfg          TableManagerConfig
 	schemaCfg    SchemaConfig
 	maxChunkAge  time.Duration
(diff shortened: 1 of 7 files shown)


pkg/chunk/table_manager.go
```diff
@@ -205,20 +204,31 @@ func (m *TableManager) SyncTables(ctx context.Context) error {
 	expected := m.calculateExpectedTables()
 	level.Info(util.Logger).Log("msg", "synching tables", "num_expected_tables", len(expected), "expected_tables", len(expected))
 
-	toCreate, toCheckThroughput, toDelete, err := m.partitionTables(ctx, expected)
-	if err != nil {
-		return err
-	}
+	for i, client := range m.clients {
+		toCreate, toCheckThroughput, toDelete, err := m.partitionTables(ctx, client, expected)
```
gouthamve: This doesn't seem right. This would mean that every single client would create a bunch of random tables (ones that belong to other clients).

Can we pass `config` to `calculateExpectedTables()` and then use that for each of the clients?

codesome: I was thinking the same. But it is the same behavior before apparently. `calculateExpectedTables` parses all the configs, and creates/updates/deletes from the last client. Here I was only deleting from old client, but left the behavior of the last client unchanged.

Do you want me to change it to per client for `calculateExpectedTables`?

gouthamve: But the delete calls will fail, as they are trying to delete tables that don't exist. Yeah, changing `calculateExpectedTables` makes sense.

codesome: Tests failed for the change described. While I am still digging into the test failure, from past implementation my understanding is that `i th` client uses all config from the starting till the `i th` config to create tables.

codesome: I changed the implementation to use `0,1,...,i'th` config for the `i'th` client. Tests pass for that. It that the expected behaviour then?

bboreham: Doesn't sound right.  Why wouldn't the i'th client just use the i'th config?

codesome: What you said sounds right. Apparently the old code used all the configs for the client. I will update the tests then.

codesome: Now I see why all the configs were parsed in the old code
https://github.com/cortexproject/cortex/blob/53438dfee1beec34876b14fc977f345158c351dc/pkg/chunk/table_manager.go#L244-L255

The old configs are parsed and checked for grace period w.r.t new configs. Also provisions read/write accordingly if the config is active/inactive. I guess we would need parsing the old configs if we want to delete the tables from the client if tables were created because of the grace period. So `0,1,...,i'th` configs are only parsed inside `calculateExpectedTables` for the `i'th` client, and not necessarily used.

bboreham: Could we make a rule that grace period only applies to the latest config?

codesome: Done.

khaines:
@gouthamve @bboreham Since you two had looked at this PR previously, can you please look at it again? I'm taking a scan through old PRs to clear out older items laying about.

pkg/chunk/table_manager.go
```diff
@@ -202,36 +207,65 @@ func (m *TableManager) bucketRetentionLoop() {
...
+		}
 
-	if err := m.deleteTables(ctx, toDelete); err != nil {
-		return err
-	}
+		if err := m.deleteTables(ctx, client, toDelete); err != nil {
+			if lastClient {
+				return err
+			}
+			level.Error(util.Logger).Log("msg", "error in deleting tables from TableClient")
```
khaines: This log output states there was an error, but offers no context or the error itself. Can you add more details to this line in order to make it useful for troubleshooting?

pkg/cortex/modules.go
```diff
@@ -293,15 +290,27 @@ func (t *Cortex) initTableManager(cfg *Config) error {
 		os.Exit(1)
 	}
 
-	tableClient, err := storage.NewTableClient(lastConfig.IndexType, cfg.Storage)
-	if err != nil {
-		return err
+	var tableClients []chunk.TableClient
```
jtlisi: Is there any reason we can't generate the table and bucket clients inside the NewTableManager function? Instead of instantiating the clients and passing them as a parameter would could just pass the storage config as a parameter. I think the way things are now is probably fine. However, the columnar relationship between the schema cfg and clients makes me think it might be better to encapsulate this logic inside the function. The comment I'm referring to:

```
// The i'th TableClient in 'tableClients' should correspond to the i'th PeriodConfig in schemaCfg.Config.
// The TableClient for the last client should not be nil.
func NewTableManager(cfg TableManagerConfig, schemaCfg SchemaConfig, maxChunkAge time.Duration, tableClients []TableClient, objectClient BucketClient) (*TableManager, error) {objectClient BucketClient) (*TableManager, error) {
```

pkg/chunk/table_manager.go
jtlisi: 👍 It would be good to print the error in this log statement

pkg/chunk/table_manager.go
```diff
@@ -202,36 +207,65 @@ func (m *TableManager) bucketRetentionLoop() {
...
-	expected := m.calculateExpectedTables()
-	level.Info(util.Logger).Log("msg", "synching tables", "num_expected_tables", len(expected), "expected_tables", len(expected))
+	for i, client := range m.clients {
+		if client == nil {
+			continue
+		}
+		lastClient := i == len(m.clients)-1
+		expected := m.calculateExpectedTables(i)
+		if lastClient {
+			level.Info(util.Logger).Log("msg", "synching tables", "num_expected_tables", len(expected), "expected_tables", len(expected))
```
jtlisi: What is the difference between `num_expected_tables` and `expected_tables` in these two log statements?

pkg/chunk/table_manager.go
```diff
@@ -202,36 +207,65 @@ func (m *TableManager) bucketRetentionLoop() {
...
+	if idx < 0 || idx >= len(m.schemaCfg.Configs) {
+		return result
+	}
 
-	for i, config := range m.schemaCfg.Configs {
+	// Use only the idx'th config for idx'th client.
+	startIdx := idx
+	if idx == len(m.schemaCfg.Configs)-1 {
+		// For the last client, collect tables from all the configs.
+		startIdx = 0
```
jtlisi: Why are we collecting all of the tables for the final config?

pkg/chunk/table_manager.go
codesome: Ah I already opened a PR to fix it upstream before looking at this comment https://github.com/cortexproject/cortex/pull/1754, they are the same. Noticed this log when running Loki.

Will fix it here too.

pracucci:
@gouthamve @jtlisi @bboreham What's the state of this PR? Is it something we wanna put some effort to reach a mergeable state, or should we close it?

codesome:
Closing this to let someone else take it up. I don't plan to work on this further at the moment.

Issue #1454, opened 2019-06-09
Allow custom authenticator for cassandra clusters


CH-JosephBironas:
We're doing audit logging using the ecAudit cassandra module and getting this error from table-manager:
level=error ts=2019-06-09T16:34:17.500073012Z caller=log.go:141 msg="error initializing cortex" err="gocql: unable to create session: unable to discover protocol version: unexpected authenticator \"com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator\"\n

Looking at this gocql issue for reference, it looks like it's possible to create a new authProvider object that allows override of gocql's default list.
https://github.com/gocql/gocql/issues/1109#issuecomment-439044906

gouthamve:
Closing as it's fixed by https://github.com/gocql/gocql/pull/1321

PR #1456, opened 2019-06-10
Cassandra Auth Plugin fix

CH-JosephBironas:
gocql maintains a whitelist of accepted authentication plugins, and ecAudit is not on the list. They were kind enough to offer an interface so that you can write your own however. In this case, though, it's just a reimplementation of the basic username/password without the whitelist as the ecAudit plugin is a dropin replacement for cassandra's native username/password mechanism.

This is a fix for issue #1454.

### DIFF
--- a/pkg/chunk/cassandra/storage_client.go
+++ b/pkg/chunk/cassandra/storage_client.go
@@ -36,27 +36,6 @@ type Config struct {
 	ConnectTimeout           time.Duration `yaml:"connect_timeout,omitempty"`
 }
 
-// PasswordAuthenticator creates the gocql Authenticator object
-type passwordAuthenticator struct {
-	Username string
-	Password string
-}
-
-// Challenge implements the Authenticator interface for gocql
-func (p passwordAuthenticator) Challenge(req []byte) ([]byte, gocql.Authenticator, error) {
-	resp := make([]byte, 2+len(p.Username)+len(p.Password))
-	resp[0] = 0
-	copy(resp[1:], p.Username)
-	resp[len(p.Username)+1] = 0
-	copy(resp[2+len(p.Username):], p.Password)
-	return resp, nil, nil
-}
-
-// Success implementes the Authenticator interface for gqcql
-func (p PasswordAuthenticator) Success(data []byte) error {
-	return nil
-}
-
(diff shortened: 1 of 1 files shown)


CH-JosephBironas:
Arg. git-- I have no idea how the additional files got pulled into this PR.

bboreham:
It looks like your branch starts from one of Tom's PRs?

You may be able to `git rebase master`, or throw away your current state with `git reset --hard master` and `git cherry-pick` each commit you do want.

CH-JosephBironas:
`git rebase master` did the trick. Thanks!

pkg/chunk/cassandra/storage_client.go
```diff
@@ -121,6 +121,27 @@ func (cfg *Config) createKeyspace() error {
...
+// PasswordAuthenticator creates the gocql Authenticator object
+type PasswordAuthenticator struct {
+	Username string
+	Password string
+}
+
+// Challenge implements the Authenticator interface for gocql
+func (p PasswordAuthenticator) Challenge(req []byte) ([]byte, gocql.Authenticator, error) {
+	resp := make([]byte, 2+len(p.Username)+len(p.Password))
+	resp[0] = 0
```
bboreham: Is there some reference for the particular format of this byte slice? Might be good to put that in a comment.

pkg/chunk/cassandra/storage_client.go
```diff
@@ -121,6 +121,27 @@ func (cfg *Config) createKeyspace() error {
...
+	copy(resp[1:], p.Username)
+	resp[len(p.Username)+1] = 0
+	copy(resp[2+len(p.Username):], p.Password)
+	return resp, nil, nil
+}
+
+// Success implements the Authenticator interface for gqcql
+func (p PasswordAuthenticator) Success(data []byte) error {
+	return nil
```
bboreham: I know nothing about Cassandra, so maybe this is obvious, but I found this line surprising.
We always return no error without checking anything?

CH-JosephBironas:
Holding off on this for the time being to see of the plugin can be added to the upstream library whitelist since this impacts any project using gocql.

CH-JosephBironas:
Witdrawing now that https://github.com/gocql/gocql/pull/1321 is merged.

Issue #1450, opened 2019-06-06
Queries that hit size limits are retried


gouthamve:
We are seeing a lot of the following errors in our queriers:
```
level=error ts=2019-06-06T17:42:31.656508688Z caller=worker.go:176 msg="error processing requests" err="rpc error: code = ResourceExhausted desc = trying to send message larger than max (398218677 vs. 104857600)"
```

Now this same query is being retried 5 times, as the front-end sees a 500 error. We should have a limit on the number of samples being returned and then return a 4xx for it.

bboreham:
Related: #1178

Issue #1178, opened 2019-01-08
Poor Error Response for Max Query Length (Instant Queries)


jtlisi:
When running an instant query with a length greater than the `MaxQueryLength` set in cortex the returned error is not helpful.

Error: `Query failed after 5 retries.`

When run on a range query an explicit message is returned showing the reason the query failed: 
```json
{
  "status": "error",
  "errorType": "execution",
  "error": "rpc error: code = Code(400) desc = invalid query, length > limit (789h0m0s > 720h0m0s)",
  "message": "rpc error: code = Code(400) desc = invalid query, length > limit (789h0m0s > 720h0m0s)"
}
```

Ideally the error should be the same for both range and instant queries.

bboreham:
I hit the same problem after #1255 merged - underlying error was `"rpc error: code = Code(413) desc = exceeded maximum number of series in a query"`

It seems to me that an error in that category should not be retried, since there is no reason to suppose it will work better the next time.

gouthamve:
Right, got to the bottom of this. https://github.com/cortexproject/cortex/blob/master/pkg/querier/ingester_streaming_queryable.go#L51-L53 this makes all the errors thrown by the ingesters, 500 errors cuz of this: https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L399-L400 which are then retried.

rfratto:
@jtlisi Looks like this was fixed at some point. I ran `rate(consul_http_request[40d])` against Cortex and both `query` and `query_range` returned the expected error:

```
{"status":"error","errorType":"execution","error":"rpc error: code = Code(400) desc = invalid query, length \u003e limit (961h0m0s \u003e 744h0m0s)"}
```

jtlisi:
^

PR #1459, opened 2019-06-10
Make sure we send a 400 for large responses.

gouthamve:
While this is not super ideal, we will still catch 99% of the cases
through this.

Fixes: https://github.com/cortexproject/cortex/issues/1450


bboreham:
Why is "send a 400" useful in this case but not in #1178 ?

gouthamve:
See: https://github.com/cortexproject/cortex/issues/1178#issuecomment-500811028

PR #1460, opened 2019-06-11
Remove redundant op name middleware.

gouthamve:
The weaveworks/common already has the middleware on by default.
See: https://github.com/weaveworks/common/blob/master/server/server.go#L162-L164
and https://github.com/weaveworks/common/pull/126

Else we are ending up URIs in the op names.


gouthamve:
I think it's legacy that is used only for only one path (`/api/prom/push`) in the old code. I think we're going to get the same result even in the distributors with this.

PR #1464, opened 2019-06-12
Fix ingester startup with a normalised ring

### LINKED ISSUE #1311: Ingester restart with -ingester.normalise-tokens on results in not-ready state
We had an ingester running on a node which spontaneously restarted.
Kubernetes started the ingester process, but it refused to go Ready.
We have excellent debugging for this condition now:

```
$ curl http://10.244.202.176/ready
Not ready: this ingester owns no tokens
```

At the top of the ingester logs on restart was:

```
level=info ts=2019-04-01T10:26:31.676674615Z caller=lifecycler.go:383 msg="existing entry found in ring" state=ACTIVE tokens=0
```

I believe this is because that code doesn't handle the normalise-tokens case.

bboreham:
Fixes #1311

### DIFF
--- a/pkg/ring/lifecycler_test.go
+++ b/pkg/ring/lifecycler_test.go
@@ -101,3 +101,51 @@ func TestRingNormaliseMigration(t *testing.T) {
 			d.(*Desc).Ingesters["ing2"].Tokens[0] == token
 	})
 }
+
+type nopFlushTransferer struct{}
+
+func (f *nopFlushTransferer) StopIncomingRequests() {}
+func (f *nopFlushTransferer) Flush()                {}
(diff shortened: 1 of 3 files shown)


csmarchbanks:
Code looks good, but would it be possible to add a test case in/similar to `TestRingNormaliseMigration`?

bboreham:
I added a test which led me to realise my original approach was flawed.
New approach is to do the denormalise every time we need `TokensFor()` an ingester, which is in `initRing()` and `autoJoin()`.
(and hence I realise `autoJoin()` didn't check for clashing tokens in normalised mode)

`migrateRing()` is now rather poorly named, but I felt we could move that later.

PR #1465, opened 2019-06-13
Add documentation for HA tracker flags.

cstyan:
Do we want to include the kvstore flags that are nested under `ha-tracker`?

@gouthamve @tomwilkie


weeco:
Can you add a couple words about what is necessary beside the available flags?

E. g. setting the external labels in prometheus. What happens if you enable the HA config in Cortex but your tenant/user does not set the replica flag or sets it incorrectly (wrong property name)?

gouthamve:
One thing that struck me. Why is it not nested as `distributor.ha-tracker`?

docs/arguments.md
```diff
@@ -92,6 +92,38 @@ The ingester query API was improved over time, but defaults to the old behaviour
 - `-distributor.extra-query-delay`
    This is used by a component with an embedded distributor (Querier and Ruler) to control how long to wait until sending more than the minimum amount of queries needed for a successful response.
 
+- `distributor.accept-ha-samples`
+   Per-user flag to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup. This defaults to false, and is technically defined in the Distributor limits.
+
+- `distributor.accept-ha-labels`
```
gouthamve: Can we make it `enable-ha-tracker`? Or

docs/arguments.md
```diff
@@ -92,6 +92,38 @@ The ingester query API was improved over time, but defaults to the old behaviour
 - `-distributor.extra-query-delay`
    This is used by a component with an embedded distributor (Querier and Ruler) to control how long to wait until sending more than the minimum amount of queries needed for a successful response.
 
+- `distributor.accept-ha-samples`
```
gouthamve: How is this a per-user flag, it's global for all user no? Can we rename it to be something more descriptive?

docs/arguments.md
cstyan: I think that makes sense.

docs/arguments.md
cstyan: It's defined in limits, so you can set an override per user. Is the description not accurate? 

We can rename to anything, any suggestions as to what would be more descriptive?

cstyan:
> Why is it not nested as distributor.ha-tracker?

I don't remember any reason for it not being nested from the review, it might have just slipped through then.

weeco:
@cstyan Looks good for me now, if you want a really good documentation you might want to explain at a high level how the HA feature works (maybe in the architecture.md). Does it dedupe the replicas in the distributor/ingester? Should we expect an increased RAM usage in the distributor or ingester (e. g. at the factor of n where n is the number of replicas)?

docs/arguments.md
```diff
@@ -92,6 +92,43 @@ The ingester query API was improved over time, but defaults to the old behaviour
...
+- `ha-tracker.consul.acltoken`
+   ACL Token used to interact with Consul.
+- `ha-tracker.consul.client-timeout`
+   HTTP timeout when talking to consul (default 20s)
+- `ha-tracker.consul.consistent-reads`
+   Enable consistent reads to consul. (default true)
+- `ha-tracker.consul.hostname`
+   Hostname and port of Consul. (default "localhost:8500")
+- `ha-tracker.consul.prefix`
+   Prefix for keys in Consul. Should end with a /. (default "collectors/")
```
csmarchbanks: Perhaps add a not here to set this value to something other than the default if you are using the same consul as for the ring. Or change the default to something like `collectors/HA/`?

cstyan: Hmm. I don't think _not_ changing the value to something other than the prefix used for the ring would cause any issues. The keys for HA tracking are `user ID/cluster ID`.

Unfortunately without some more refactoring we wouldn't be able to change the default. We just call the `RegisterFlags` function for the other config structs, passing an optional prefix.

csmarchbanks: I don't think it will cause issues unless you happen to have a tenant named `ring`. Mostly it was nice to debug that everything was looking correct for them to be in their own folder.

cstyan: Agreed. Best practice imo would be to use different prefixes.

docs/arguments.md
```diff
@@ -92,6 +92,43 @@ The ingester query API was improved over time, but defaults to the old behaviour
...
+
+### HA Tracker
+
+HA tracking has two of it's own flags:
+- `ha-tracker.cluster`
+   Prometheus label to look for in samples to identify a Prometheus HA cluster. (default "cluster")
+- `ha-tracker.replica`
+   Prometheus label to look for in samples to identify a Prometheus HA replica. (default "__replica__")
+
+It's reasonable to assume people probably already have a `cluster` label, or something similar. If not, they should add one along with `__replica__`
```
weeco: We don't use external labels in our prometheus config, as each prometheus gets a different UserID / TenantID. Does Cortex still require the "cluster" external label then if I only have one prometheus cluster per tenant or does it use the default "cluster" if no label is specified?

cstyan: Both labels have to be present for Cortex to dedup samples from HA prometheus instances. Whether the labels are already in your series or added via external labels doesn't matter. Cortex just needs a way of identifying each HA cluster, and each replica within that cluster.

weeco: Assuming I am just about to enable the HA tracker feature and I have a single prometheus instance for each tenant/user ID which does not set either of these two labels. Does this cause any problems, will it reject samples without these labels?

docs/arguments.md
```diff
@@ -92,6 +92,43 @@ The ingester query API was improved over time, but defaults to the old behaviour
 - `-distributor.extra-query-delay`
    This is used by a component with an embedded distributor (Querier and Ruler) to control how long to wait until sending more than the minimum amount of queries needed for a successful response.
 
+- `distributor.accept-ha-samples`
+   Per-user flag to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup. This defaults to false, and is technically defined in the Distributor limits.
+
+- `distributor.enable-ha-tracker` 
+   Enable the distributors HA tracker so that it can accept samples from Prometheus HA replicas gracefully (requires labels). Global (for distributors), this ensures that the necessary internal data structures for the HA handling are created. The option `accept-ha-samples` is still needed to enable ingestion of HA samples per user ID until that flag is defaulted to true.
```
csmarchbanks: I don't think you need "until that flag is defaulted to true"

docs/arguments.md
```diff
@@ -92,6 +92,43 @@ The ingester query API was improved over time, but defaults to the old behaviour
...
+- `ha-tracker.consul.consistent-reads`
+   Enable consistent reads to consul. (default true)
+- `ha-tracker.consul.hostname`
+   Hostname and port of Consul. (default "localhost:8500")
+- `ha-tracker.consul.prefix`
+   Prefix for keys in Consul. Should end with a /. (default "collectors/")
+- `ha-tracker.failover-timeout`
+   If we don't receive any samples from the accepted replica for a cluster in this amount of time we will failover to the next replica we receive a sample from. This value must be greater than the update timeout (default 30s)
+- `ha-tracker.store`
+   Backend storage to use for the ring (consul, inmemory). (default "consul")
```
csmarchbanks: etcd is also an option here now.

cstyan:
Could use another review @csmarchbanks  
cc @tomwilkie

Added a note about HA tracking in the architecture doc, and I also added info about the `store` flags in arguments.md since there weren't any in there.

docs/arguments.md
```diff
@@ -92,6 +92,61 @@ The ingester query API was improved over time, but defaults to the old behaviour
...
+- `distributor.accept-ha-samples`
+   Per-user flag to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup. This defaults to false, and is technically defined in the Distributor limits.
+
+- `distributor.enable-ha-tracker` 
+   Enable the distributors HA tracker so that it can accept samples from Prometheus HA replicas gracefully (requires labels). Global (for distributors), this ensures that the necessary internal data structures for the HA handling are created. The option `accept-ha-samples` is still needed to enable ingestion of HA samples per user ID.
+
+### Ring/HA Tracker Store
+
+The KVStore client is used by both the Ring and HA Tracker.
+- `prefix`
```
csmarchbanks: Some sort of syntax like `{ring,ha-tracker}.prefix` would help, from these docs I would not know to add the prefix.

docs/arguments.md
```diff
@@ -92,6 +92,61 @@ The ingester query API was improved over time, but defaults to the old behaviour
...
+   The prefix for the keys in the store. Should end with a /. For example with a preifx of foo, the key bar would be stored under foo/bar.
+- `store`
+   Backend storage to use for the ring (consul, etcd, inmemory).
+
+#### Consul
+
+- `consul.hostname`
```
csmarchbanks: Maybe a note to add `ha-tracker.` before these to configure the ha-tracker?

docs/architecture.md
```diff
@@ -24,6 +24,8 @@ The **distributor** service is responsible for handling samples written by Prome
 
 Distributors communicate with ingesters via [gRPC](https://grpc.io). They are stateless and can be scaled up and down as needed.
 
+If the HA Tracker is enabled, the Distributor will deduplicate incomming samples that contain both a cluster and replica label. It talks to a KVStore to store state about which replica per cluster it's accepting samples from for a given user ID. Samples with one or neither of these labels will be accepted by default.
```
csmarchbanks: s/incomming/incoming

docs/arguments.md
```diff
@@ -92,6 +92,67 @@ The ingester query API was improved over time, but defaults to the old behaviour
...
+   Per-user flag to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup. This defaults to false, and is technically defined in the Distributor limits.
+
+- `distributor.enable-ha-tracker` 
+   Enable the distributors HA tracker so that it can accept samples from Prometheus HA replicas gracefully (requires labels). Global (for distributors), this ensures that the necessary internal data structures for the HA handling are created. The option `accept-ha-samples` is still needed to enable ingestion of HA samples per user ID.
+
+### Ring/HA Tracker Store
+
+The KVStore client is used by both the Ring and HA Tracker.
+- `{ring,ha-tracker}.prefix`
+   The prefix for the keys in the store. Should end with a /. For example with a preifx of foo, the key bar would be stored under foo/bar.
```
csmarchbanks: s/preifx/prefix, also, should probably say "prefix of foo/" since the prefix should end with a /.

cstyan:
bump @csmarchbanks @gouthamve

docs/arguments.md
gouthamve: I think calling it "per user" is very misleading. We can call it "The default value to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup. This defaults to false, and is technically defined in the Distributor limits"

Also, list it in the limits section like `distributor.ingestion-rate-limit`

pkg/distributor/distributor.go
```diff
@@ -123,7 +123,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.HATrackerConfig.RegisterFlags(f)
 
 	f.BoolVar(&cfg.EnableBilling, "distributor.enable-billing", false, "Report number of ingested samples to billing system.")
-	f.BoolVar(&cfg.EnableHAReplicas, "distributor.accept-ha-labels", false, "Accept samples from Prometheus HA replicas gracefully (requires labels).")
+	f.BoolVar(&cfg.EnableHATracker, "distributor.enable-ha-tracker", false, "Enable the distributors HA tracker so that it can accept samples from Prometheus HA replicas gracefully (requires labels).")
```
csmarchbanks: Should this be `distributor.ha-tracker.enable` to keep everything under `distributor.ha-tracker.*`?

pkg/util/validation/limits.go
```diff
@@ -44,8 +44,8 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&l.IngestionRate, "distributor.ingestion-rate-limit", 25000, "Per-user ingestion rate limit in samples per second.")
 	f.IntVar(&l.IngestionBurstSize, "distributor.ingestion-burst-size", 50000, "Per-user allowed ingestion burst size (in number of samples). Warning, very high limits will be reset every -distributor.limiter-reload-period.")
 	f.BoolVar(&l.AcceptHASamples, "distributor.accept-ha-samples", false, "Per-user flag to enable handling of samples with external labels for identifying replicas in an HA Prometheus setup.")
```
csmarchbanks: Perhaps this should be `distributor.ha-tracker.accept-users` or something like that? Since turning it on here will turn it on for all users.

tomwilkie:
This LGTM too.  I see enough LGTMs on this to merge now.

PR #1469, opened 2019-06-19
Drop chunks with less than min samples

csmarchbanks:
Allow a user to configure the minimum number of samples needed for a
chunk to be flushed to the store. This is to avoid your store being
overloaded by lots of tiny chunks that are unlikely to have useful
information.

Signed-off-by: Chris Marchbanks <csmarchbanks@gmail.com>

### DIFF
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -64,6 +64,10 @@ var (
 		Name: "cortex_ingester_flush_reasons",
 		Help: "Total number of series scheduled for flushing, with reasons.",
 	}, []string{"reason"})
+	droppedChunks = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "cortex_ingester_dropped_chunks_total",
+		Help: "Total number of chunks dropped from flushing because they have too few samples.",
+	})
 )
 
 // Flush triggers a flush of all the chunks and closes the flush queues.
@@ -272,6 +276,29 @@ func (i *Ingester) flushUserSeries(flushQueueIndex int, userID string, fp model.
 	} else {
 		chunks = chunks[:len(chunks)-1]
 	}
+
+	if reason == reasonIdle && series.headChunkClosed && i.cfg.MinChunkLength > 0 {
+		chunkLength := 0
+		for _, c := range chunks {
+			chunkLength += c.C.Len()
+		}
+		if chunkLength < i.cfg.MinChunkLength {
+			userState.removeSeries(fp, series.metric)
+			memoryChunks.Sub(float64(len(chunks)))
+			droppedChunks.Add(float64(len(chunks)))
+			util.Event().Log(
+				"msg", "dropped chunks",
+				"userID", userID,
+				"numChunks", len(chunks),
+				"chunkLength", chunkLength,
+				"fp", fp,
+				"series", series.metric,
+				"queue", flushQueueIndex,
+			)
+			chunks = nil
+		}
+	}
+
 	userState.fpLocker.Unlock(fp)
 
 	if len(chunks) == 0 {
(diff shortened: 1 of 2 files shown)


cboggs:
On the topic of extra safety in the form of, "let's be sure this is really just a lonely series with too little data"... I wonder if we could make use of the index cache to look up a series' fingerprint before deciding to drop chunks, so that we can see if any chunks for that series have been recently(-ish) flushed.

I _think_ that if we find a chunk with < N samples, then look up the fingerprint for that chunk's series and get a cache miss, we can safely assume "this chunk represent a teensy island in the sea, drop it".

Thoughts?

pkg/ingester/flush.go
```diff
@@ -333,6 +337,10 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, metric
 
 	wireChunks := make([]chunk.Chunk, 0, len(chunkDescs))
 	for _, chunkDesc := range chunkDescs {
+		// Do not write chunks below the minimum chunk duration.
+		if chunkDesc.C.Len() < i.cfg.MinChunkLength {
```
bboreham: nit: `Len()` is expensive on most chunk types (not bigchunk), so you could re-use the value 15 lines later

pkg/ingester/flush.go
```diff
@@ -349,6 +357,10 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, metric
 	// Record statistsics only when actual put request did not return error.
 	for _, chunkDesc := range chunkDescs {
 		utilization, length, size := chunkDesc.C.Utilization(), chunkDesc.C.Len(), chunkDesc.C.Size()
+		if length < i.cfg.MinChunkLength {
+			droppedChunks.Inc()
+			continue
```
bboreham: I'd want these passed to `util.Event()` for analysis.

csmarchbanks: :+1:

pkg/ingester/flush.go
csmarchbanks: I ended up moving the code around in order to only drop idle chunks that have been closed. I don't think it is as easy to reuse the value anymore.

csmarchbanks:
This drops 625 chunks per second in our cluster (for chunks with 6 or fewer samples, which is 3 minutes for us). That is a fair bit of AWS savings if you don't care about those short lived series.

pkg/ingester/ingester.go
```diff
@@ -111,6 +112,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.MaxChunkAge, "ingester.max-chunk-age", 12*time.Hour, "Maximum chunk age before flushing.")
 	f.DurationVar(&cfg.ChunkAgeJitter, "ingester.chunk-age-jitter", 20*time.Minute, "Range of time to subtract from MaxChunkAge to spread out flushes")
 	f.BoolVar(&cfg.SpreadFlushes, "ingester.spread-flushes", false, "If true, spread series flushes across the whole period of MaxChunkAge")
+	f.IntVar(&cfg.MinChunkLength, "ingester.min-chunk-length", 0, "Minimum number of samples in a chunk to flush it to the store. Use with care, if chunks are less than this size they will be discarded.")
```
bboreham: This is now slightly inaccurate - a chunk can be less than this size and not discarded if it adjoins another chunk in the same series. Also if the ingester goes into 'flush immediately' mode it doesn't apply.

Though I don't suggest you fit all this into the CLI help.

csmarchbanks: I added the word "idle" to the description to make it a bit more accurate.

Issue #1402, opened 2019-05-18
RFC: Release cadence, process and ideas


gouthamve:
We've decided to do a cortex release soon, but after the first release, I think we should have some process and cadence for future releases. I made a Google docs with initial suggestions here: https://docs.google.com/document/d/1Ma5upv_lPG0KaJySqm8AY20ZjicNrB2qZS5WIlDqbNw/edit#

This is largely inspired by the Prometheus release cadence and process.

avicshw:
@gouthamve Is there any progress?

gouthamve:
Closing as this is done in #1470

PR #1470, opened 2019-06-20
Provide instructions and ci job for releases

csmarchbanks:
Lets just do it!

These instructions are based on the Prometheus instructions with a few small modifications according to https://github.com/cortexproject/cortex/issues/1402. I set the first release date to be on Tuesday in order for to collect some feedback.

Updated the circle ci deploy job to use the `CIRCLE_TAG` env var if it is present rather than generating the image name from the commit sha.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,11 +9,24 @@ defaults: &defaults
 workflows:
   version: 2
   test-build-deploy:
+    # You must specify a tag filter for each job that deploy depends on.
     jobs:
-    - lint
-    - test
-    - integration
-    - build
+    - lint:
+        filters:
+          tags:
+            only: /^v[0-9]+(\.[0-9]+){2}(-.+|[^-.]*)$/
+    - test:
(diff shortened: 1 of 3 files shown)


cboggs:
This seems like a good place to start. Accommodates shepherding from outside Grafana/Weaveworks, which makes sense overall.

LGTM, we'll iterate as we go.

weeco:
Maybe we should also wait for https://github.com/cortexproject/cortex/pull/1465 to be merged before the release. It renames a flag.

csmarchbanks:
> Maybe we should also wait for #1465 to be merged before the release. It renames a flag.
I don't think that should be a blocker, 0.1.x -> 0.2.x is allowed to have breaking changes in it. We will just make sure to document them.

Would love another maintainer or two to take a look at this before I update the dates and do the first release! (@bboreham, @tomwilkie, @gouthamve, @khaines)

cboggs:
LGTM :+1:

PR #1471, opened 2019-06-21
add options for boltdb_index_client to avoid dead loop

mizeng:
As part of solution for https://github.com/grafana/loki/issues/685, the purpose is to avoid opening boltdb fall into dead loop.

### DIFF
--- a/pkg/chunk/local/boltdb_index_client.go
+++ b/pkg/chunk/local/boltdb_index_client.go
@@ -140,7 +140,8 @@ func (b *boltIndexClient) getDB(name string) (*bbolt.DB, error) {
 	}
 
 	// Open the database.
-	db, err := bbolt.Open(path.Join(b.cfg.Directory, name), 0666, nil)
+	// Set Timeout to avoid obtaining file lock wait indefinitely.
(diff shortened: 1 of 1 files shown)


bboreham:
It seems that your issue is about accessing Boltdb from more than one process.
The Boltdb README says: "multiple processes cannot open the same database at the same time".

So I am not clear why a timeout will help.

mizeng:
@bboreham If the one request in the process hang on BoltDB for any reason, it should not block any incoming requests rather than keep hanging. I think that's why this boltdb client has timeout parameter.
Using default parameter anyway looks like improper for fast failed.

Quote from the boltdb description in code for this parameter:
> // Timeout is the amount of time to wait to obtain a file lock.
	// When set to zero it will wait indefinitely. This option is only
	// available on Darwin and Linux.

bboreham:
So your aim is to fail every request from then on?

mizeng:
yes. Fast fail looks better then indefinite wait.

mizeng:
fair enough.

csmarchbanks:
Thanks!

PR #1473, opened 2019-06-25
Add BenchmarkChunkQueryableFromTar

codesome:
`BenchmarkChunkQueryableFromTar` is same as `TestChunkTar` in `pkg/querier/chunk_tar_test.go`. This benchmark would come handy in investigating a query offline.

### DIFF
--- a/pkg/querier/chunk_tar_test.go
+++ b/pkg/querier/chunk_tar_test.go
@@ -20,41 +20,43 @@ import (
 	"github.com/weaveworks/common/user"
 )
 
-func TestChunkTar(t *testing.T) {
-	chunksFilename := os.Getenv("CHUNKS")
-	if len(chunksFilename) == 0 {
-		return
+func getTarDataFromEnv(t testing.TB) (query string, from, through time.Time, step time.Duration, store ChunkStore) {
+	var (
+		err            error
+		chunksFilename = os.Getenv("CHUNKS")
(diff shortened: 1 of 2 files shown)


bboreham:
Can you factor out the common code?

codesome:
Done

bboreham:
What is the reason the remaining code is different between the benchmark and the test?
Looks like a limit and a timeout.

codesome:
Yes, I was not sure if I should alter those limits (samples and timeout) for `TestChunkTar`. And I was hitting both the limits when I was doing some benchmarks locally, so increased the limits for the benchmark.

Also, I would need calling `b.ResetTimer()` and `b.ReportAllocs()` for the benchmark, so the common code would be the one before that.

bboreham:
If you can explain why the limits need to be different, please put that in a comment.
If not, remove the difference.

I don't usually put `ReportAllocs()` in the code because it is available as a command-line option `-benchmem`

codesome:
Made the limits same, and removed `b.ReportAllocs()`.

bboreham:
so more of the code can be shared now?

codesome:
`b.ResetTimer()` was in between. I will try sharing more code.

codesome:
And done!

tomwilkie:
:+1: LGTM

PR #1480, opened 2019-06-27
 Check that the last sample value is valid before complaining about it

### LINKED ISSUE #1438: Rolling update on ingester causes duplicate value for timestamps
**The problem**

Every time I perform a rolling update on ingesters I can see lots of discarded samples (~1% of all ingested samples) due to "new-value-for-timestamp". With each updated ingester the number of discarded samples seems to increase. After all ingesters had been updated it takes something like 30-45minutes until the discarded samples slowly decreases towards 0 again:

Cortex Dashboard

This issue is 100% reproducible across several Cortex versions. I checked the logs and they all have in common that they say, the previous value for that timestamp seems to be 0 (which I suspect to be wrong).

If I can provide any further information to figure out the problem, please let me know.

bboreham:
The `lastSampleValue` member may not be initialized, e.g. after a transfer from one ingester to another.  We already have a flag which is set when the value is valid, so use it.

Fixes #1438

### DIFF
--- a/pkg/ingester/series.go
+++ b/pkg/ingester/series.go
@@ -63,17 +63,17 @@ func newMemorySeries(m labels.Labels) *memorySeries {
 //
 // The caller must have locked the fingerprint of the series.
 func (s *memorySeries) add(v model.SamplePair) error {
+	// If sender has repeated the same timestamp, check more closely and perhaps return error.
 	if v.Timestamp == s.lastTime {
-		// Don't report "no-op appends", i.e. where timestamp and sample
-		// value are the same as for the last append, as they are a
-		// common occurrence when using client-side timestamps
-		// (e.g. Pushgateway or federation).
-		if s.lastSampleValueSet && v.Value.Equal(s.lastSampleValue) {
-			return nil
-		}
-		// If we don't know what the last sample value is, silently discard as for no-op.
+		// If we don't know what the last sample value is, silently discard.
 		// This will mask some errors but better than complaining when we don't really know.
 		if !s.lastSampleValueSet {
+			return nil
+		}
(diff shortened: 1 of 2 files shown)


csmarchbanks:
I guess I have a small concern that it will now be possible to push out of order timestamps if it happens to be the first value after being transferred to a new ingester?

bboreham:
Yeah, after thinking about it I think we should silently drop all repeated timestamps.
If the value is different the culprit will presumably do it again some other time and get caught.

bboreham:
I re-did it to discard all samples with duplicate timestamp, when we don't know the last value.
Also extended the test to fail on the original version and on my previous effort.

pkg/ingester/series.go
```diff
@@ -63,16 +63,19 @@ func newMemorySeries(m labels.Labels) *memorySeries {
...
-		v.Timestamp == s.lastTime &&
-		v.Value.Equal(s.lastSampleValue) {
-		return nil
-	}
 	if v.Timestamp == s.lastTime {
+		// Don't report "no-op appends", i.e. where timestamp and sample
+		// value are the same as for the last append, as they are a
+		// common occurrence when using client-side timestamps
+		// (e.g. Pushgateway or federation).
+		if s.lastSampleValueSet && v.Value.Equal(s.lastSampleValue) {
```
csmarchbanks: Swapping the order of these two ifs would allow you to get rid of the `s.lastSampleValueSet` clause.

I guess it could even be collapsed to `if !s.lastSampleValueSet || v.Value.Equal(s.lastSampleValue)` but I think the comments explaining each case are nice.

bboreham: I've done the reordering.

PR #1482, opened 2019-06-30
Optionally write stub entries to the chunk cache from ingesters

bboreham:
This an optional optimisation for #1475 - if we have a dedicated cache just for de-duplicating writes then we don't need to store the chunk data in it, just a placeholder for the key.

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -64,11 +64,15 @@ type StoreConfig struct {
 
 	// Limits query start time to be greater than now() - MaxLookBackPeriod, if set.
 	MaxLookBackPeriod time.Duration `yaml:"max_look_back_period"`
+
+	// Not visible in yaml because the setting shouldn't be common between ingesters and queriers
+	chunkCacheStubs bool // don't write the full chunk to cache, just a stub entry
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *StoreConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.ChunkCacheConfig.RegisterFlagsWithPrefix("", "Cache config for chunks. ", f)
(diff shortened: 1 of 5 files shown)


pkg/chunk/chunk_store.go
```diff
@@ -69,6 +70,7 @@ type StoreConfig struct {
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *StoreConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.ChunkCacheConfig.RegisterFlagsWithPrefix("", "Cache config for chunks. ", f)
+	f.BoolVar(&cfg.ChunkCacheStubs, "store.chunk-cache-stubs", false, "If true, don't write the full chunk to cache, just a stub entry.")
```
csmarchbanks: This should only be set on the ingester correct? Would prefixing it with `ingester` make sense even though it is in the store config?

pkg/chunk/chunk_store.go
```diff
@@ -57,6 +57,7 @@ func init() {
 // StoreConfig specifies config for a ChunkStore
 type StoreConfig struct {
 	ChunkCacheConfig       cache.Config `yaml:"chunk_cache_config,omitempty"`
+	ChunkCacheStubs        bool         // don't write the full chunk to cache, just a stub entry
```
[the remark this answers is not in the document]
bboreham: Yes, I was thinking it would only be set via command-line as you wouldn't want this in a config you shared with queriers.

pkg/chunk/chunk_store.go
bboreham: I see what you're getting at, but not sure about the suggestion.
Yes, it's only supposed to be set when storing chunks, which we only do in the ingester today.

bboreham:
I renamed the field so it won't be visible in yaml

pkg/chunk/chunk_store.go
csmarchbanks: Yeah, I wasn't sure of my suggestion either. I am happy with it as is.

PR #1483, opened 2019-06-30
Replace outdated comment in Distributor.Push()

bboreham:
`sampleTracker` has moved inside `ring.DoBatch()`, so the comment doesn't make sense any more.
To replace it, I tried to summarise what goes on in the loop.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -291,7 +291,8 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/distributor/distributor.go
```diff
@@ -291,7 +291,8 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		}
 	}
 
-	// Build slice of sampleTrackers, one per timeseries.
+	// For each timeseries, compute a hash to distribute across ingesters;
+	// check each sample and discard if outside limits
```
csmarchbanks: Full stop at end of sentence please.

PR #1485, opened 2019-07-01
Optimize mergeStream

codesome:
In my quest of optimizing query performance, I found that `result.append` was taking a significant amount of time in `mergeStream` (2s+ of around 4s of `mergeStream`).

The query and data in action is described here https://github.com/prometheus/prometheus/pull/5707#issuecomment-505440230

In this PR I have inlined the appending logic with the merging logic. This saves anywhere from `1s-2.5s` for the above data.

### DIFF
--- a/pkg/querier/batch/stream.go
+++ b/pkg/querier/batch/stream.go
@@ -19,6 +19,8 @@ func (bs batchStream) print() {
 	fmt.Println("]")
 }
 
+// reset, hasNext, next, atTime etc are all inlined in go1.11.
+
 func (bs *batchStream) reset() {
 	for i := range *bs {
 		(*bs)[i].Index = 0
@@ -73,6 +75,7 @@ func mergeBatches(batches batchStream, result batchStream, size int) batchStream
 }
 
 func mergeStreams(left, right batchStream, result batchStream, size int) batchStream {
+	// Ensure that 'result' has enough capacity of left and right added together.
 	if cap(result) >= len(left)+len(right) {
 		for i := range result {
 			result[i].Index = 0
@@ -84,17 +87,17 @@ func mergeStreams(left, right batchStream, result batchStream, size int) batchSt
 	} else {
 		result = make([]promchunk.Batch, len(left)+len(right))
 	}
-	l := 1
+	resultLen := 1
 	b := &result[0]
 
 	for left.hasNext() && right.hasNext() {
 		if b.Index == size {
 			b.Length = b.Index
-			l++
-			if l > len(result) {
+			resultLen++
+			if resultLen > len(result) {
 				result = append(result, promchunk.Batch{})
 			}
-			b = &result[l-1]
+			b = &result[resultLen-1]
 		}
 		t1, t2 := left.atTime(), right.atTime()
 		if t1 < t2 {
@@ -114,11 +117,11 @@ func mergeStreams(left, right batchStream, result batchStream, size int) batchSt
 	for ; left.hasNext(); left.next() {
 		if b.Index == size {
 			b.Length = b.Index
-			l++
-			if l > len(result) {
+			resultLen++
+			if resultLen > len(result) {
 				result = append(result, promchunk.Batch{})
 			}
-			b = &result[l-1]
+			b = &result[resultLen-1]
 		}
 		b.Timestamps[b.Index], b.Values[b.Index] = left.at()
 		b.Index++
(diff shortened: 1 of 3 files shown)


bboreham:
suggests this comment wasn't true?
```
// append, reset, hasNext, next, atTime etc are all inlined in go1.11.	
```

pkg/querier/batch/stream.go
```diff
@@ -89,33 +73,74 @@ func mergeBatches(batches batchStream, result batchStream, size int) batchStream
 }
 
 func mergeStreams(left, right batchStream, result batchStream, size int) batchStream {
-	result.reset()
-	result = result[:0]
+	if cap(result) >= len(left)+len(right) {
```
bboreham: I would appreciate a comment here like "ensure that `result` has enough space to hold left and right added together".

pkg/querier/batch/stream.go
```diff
@@ -89,33 +73,74 @@ func mergeBatches(batches batchStream, result batchStream, size int) batchStream
...
+			result[i].Index = 0
+			result[i].Length = 0
+		}
+		for len(result) < len(left)+len(right) {
+			result = append(result, promchunk.Batch{})
+		}
+	} else {
+		result = make([]promchunk.Batch, len(left)+len(right))
+	}
+	l := 1
```
bboreham: lowecase ell is just too similar to digit one to use as a variable

codesome:
Ah yup, then I guess the improvement was by efficient use of `result` slice I believe. There are fewer accesses to the `result` slice (only when going to next batch as opposed to accessing it for every sample), and direct assignment of values to the batch.

Also avoiding `result = append(result, promchunk.Batch{})` for the first few `Batch` (till capacity) and just setting `Index` and `Length` to 0 also seemed to help.

codesome:
> I'd like some more information on why the repeated code cannot be a function, before approving.

I will club repeated code into functions and see if there is any difference in perf. If Go is inlining them, I believe there shouldn't be much difference.

pkg/querier/batch/stream.go
```diff
@@ -61,7 +47,7 @@ func (bs *batchStream) at() (int64, float64) {
 	return b.Timestamps[b.Index], b.Values[b.Index]
 }
 
-// mergeBatches assumes the contents of batches are overlapping and unstorted.
+// mergeBatches assumes the contents of batches are overlapping and unsorted.
 // Merge them together into a sorted, non-overlapping stream in result.
 // Caller must guarantee result is big enough.  Return value will always be a
```
bboreham: Can this comment be relied on?  Seems the code would be a lot simpler if it were.

bboreham:
This comment that was there before:

    // Caller must guarantee result is big enough.  Return value will always be a	

Can this comment be relied on? Seems the code would be a lot simpler if it were.

codesome:
I considered that, but here are the reasons why I infer `result` can fall short.

1. `mergeStreams` is also called here https://github.com/cortexproject/cortex/blob/795dd596d4d877d40cb9f3533e6d89af106cc7f8/pkg/querier/batch/merge.go#L117 so the guarantee asked by `mergeBatches` need not hold true here.

2. Tests failed if I assume that, because it was thought result length would be taken care of?

bboreham:
At least, the comment should be changed.
Right now I feel the code is too complicated for me to approve.  Possibly if I spent a few hours poking at it I would understand what is going on.

codesome:
I will try to simplify and add more comments to make it more readable.

> At least, the comment should be changed.

I was thinking we can remove `mergeBatch` totally as we are not using it anywhere.
```
➜  cortex git:(merge-stream) ✗ grep -rni mergeBatches
Binary file .pkg/linux_amd64/github.com/cortexproject/cortex/pkg/querier/batch.a matches
pkg/querier/batch/stream.go:50:// mergeBatches assumes the contents of batches are overlapping and unsorted.
pkg/querier/batch/stream.go:53:// slice pointing to the same underly array as result, allowing mergeBatches
pkg/querier/batch/stream.go:55:func mergeBatches(batches batchStream, result batchStream, size int) batchStream {
pkg/querier/batch/stream.go:66:		left := mergeBatches(batches[n:], result[n:], size)
pkg/querier/batch/stream.go:67:		right := mergeBatches(batches[:n], result[:n], size)
pkg/querier/batch/stream_test.go:48:			result = mergeBatches(tc.input, result, promchunk.BatchSize)
```

codesome:
I have simplified some part of `mergeStreams` (the initilization at the beginning) and also added some comments. Additionally, removed `mergeBatchs` as it was not used anywhere.

PR #1489, opened 2019-07-02
Add optional "Heap ballast" to alter GC behaviour.

bboreham:
This is a terrible, terrible thing.  I apologise.

It allows the heap size to be made bigger, to make Go's GC run less often and use less CPU.  See https://github.com/golang/go/issues/23044
Particularly aimed at distributors.

### DIFF
--- a/cmd/cortex/main.go
+++ b/cmd/cortex/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"runtime"
 
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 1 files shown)


cmd/cortex/main.go
```diff
@@ -60,6 +66,7 @@ func main() {
 		level.Error(util.Logger).Log("msg", "error running Cortex", "err", err)
 	}
 
+	runtime.KeepAlive(ballast)
```
csmarchbanks: Aha! I tried doing this in a different project awhile ago and did not know about KeepAlive.

PR #1491, opened 2019-07-06
Allow HTTP pushes directly to ingesters.

tomwilkie:
This is mostly for benchmark comparisons.


bboreham:
We should drop that "billing" stuff.  Unfortunately I probably need to put something else in its place or I'll spook the natives.

tomwilkie:
Agreed re: billing.  We use the per-user metrics for billing now.

tomwilkie:
@bboreham I've remove the billing code, WDYT?

bboreham:
I’m still at “ need to put something else in its place”. 
Also you should mention this change in the title and in changelog.

tomwilkie:
> I’m still at “ need to put something else in its place”.

Do you want us to wait?  When do you think you'll have a chance to do this?

pkg/cortex/modules.go
```diff
@@ -341,6 +342,7 @@ func (t *Cortex) initIngester(cfg *Config) (err error) {
 	t.server.HTTP.Path("/ready").Handler(http.HandlerFunc(t.ingester.ReadinessHandler))
 	t.server.HTTP.Path("/flush").Handler(http.HandlerFunc(t.ingester.FlushHandler))
 	t.server.HTTP.Path("/shutdown").Handler(http.HandlerFunc(t.ingester.ShutdownHandler))
+	t.server.HTTP.Handle("/push", t.httpAuthMiddleware.Wrap(push.Handler(cfg.Distributor, t.ingester.Push)))
```
pracucci: Two things:
1. I'm a bit concerned about the inconsistency in the path between distributors (`/api/prom/push`) and ingesters (`/push`). Have you thought about it?
2. Could you document it to `docs/apis.md`, please?

bboreham:
> > I’m still at “ need to put something else in its place”.
> Do you want us to wait? When do you think you'll have a chance to do this?

I have written the program.  I don't think it's essential that you wait.

pkg/cortex/modules.go
tomwilkie: > I'm a bit concerned about the inconsistency in the path between distributors (/api/prom/push) and ingesters (/push). Have you thought about it?

Yeah, we can have them at the same path as the paths will collide in the single binary case.

I put it at `/push` to indicate its a private API; the /api/prom is for public endpoints.

PR #1492, opened 2019-07-07
Ensure -dynamodb.url is set when calling NewTableClient() in aws mode

pracucci:
Loki depends on Cortex. When loki is started without `-dynamodb.url` (or a properly configured config file) it panics in `NewTableClient()` (`aws` is the default mode). This PR fixes it.

Fixes https://github.com/grafana/loki/issues/714.


marco-m:
@pracucci I am impressed by how thoroughly you root-cause fixed grafana/loki#714. Kudos.

csmarchbanks:
Thanks!

Issue #1301, opened 2019-03-25
Ingestion latency spikes when discarding metrics due to limits


gouthamve:
![image](https://user-images.githubusercontent.com/7354143/54903852-dc88ea00-4f02-11e9-9e77-619fc91d0676.png)

We're seeing consistent spikes in ingester `/cortex.Ingester/Push` latency when there is an uptick in the samples dropped due to limits.

Needs to be investigated further.

bboreham:
@gouthamve do you see any improvement after #1497 ?

bboreham:
I have a theory what causes this.

For each rejection, `userState.getSeries()` calls `httpgrpc.ErrorFromHTTPResponse()`, which does a lot of memory-allocating, then returns it to `Ingester.Push()` which calls `httpgrpc.HTTPResponseFromError()` which does even more memory-allocating.  Then `Push()` throws it all away except for the last error.

I recommend we use an error object that just holds a reference to the labels until called upon to format itself.

PR #1497, opened 2019-07-10
Reduce contention on discarded samples counters

bboreham:
Pre-create a vector of counters per user, so we don't lock across all users when the counts are increased.

This should help #1301


tomwilkie:
LGTM

PR #1498, opened 2019-07-10
Count samples silently discarded for duplicate timestamp

bboreham:
Make sure all cases where ingesters discard data are covered in counters.

I have a case where distributors are seeing data but it's not showing up in queries, so trying to cover all possibilities I can think of why it could be discarded.


tomwilkie:
LGTM

Issue #157, opened 2016-11-24
Come up with a better scheme for dealing with consul outages


tomwilkie:
We should be able to operate if consul goes away.  This probably means heartbeating the ingesters ourselves, or something like that.

bboreham:
I was thinking about this - in a Kubernetes install we could just store the ring as an object in the api-server, and forget Consul.

bboreham:
I imagine the benefit of Consul (or equivalent) is that all callers have a consistent view of the world.  If callers have an inconsistent view then queries can go to ingesters that don't have all the samples.

So what does "heartbeating the ingesters ourselves" mean? "Use a last-known-consistent copy of the ring, and check heartbeats from ingesters are still consistent with that" ?

bboreham:
OK, so "Use a last-known-consistent copy of the ring, make heartbeat calls directly to ingesters in that ring"?

rverma-nikiai:
+1 for kubernetes crd

gouthamve:
One solution: We have healthcheck / heartbeats in the distributors to the ingesters. if consul is down, use the healthcheck info to coast until it is back up.

friedrichg:
With memberlist and dynamodb we can close this as done. No longer needed to run single consul.

PR #1499, opened 2019-07-10
Honor -consul.consistent-reads on watches too.

tomwilkie:
We're only run Consul now - don't need the HA as its ephemeral state, and more outages came from consul clusters outages than anything else.

-consul.consistent-reads=false massively reduces the load on the Consul server, but most of GET load is from watches.  So we should honor this flag there.


bboreham:
Related: #157

PR #1510, opened 2019-07-12
Vendor update

codesome:
It includes [this](https://github.com/prometheus/prometheus/pull/5707) query optimization.


codesome:
I have opened https://github.com/weaveworks/common/pull/161 for some conflicts.

codesome:
The `TestChunkStoreRandom` which failed in the test seems to be flaky. I ran it on my machine and it failed for the first time and then it did not fail later.

gouthamve:
Do you think the flakyness is caused by the vendor update?

codesome:
I plan to investigate that sometime soon. If it was flaky before we would have seen it in the tests, so it might be the vendor update.

go.mod
```diff
@@ -27,39 +27,34 @@ require (
...
 	github.com/golang/snappy v0.0.1
-	github.com/google/gofuzz v0.0.0-20161122191042-44d81051d367 // indirect
-	github.com/googleapis/gnostic v0.2.0 // indirect
 	github.com/gophercloud/gophercloud v0.0.0-20190307220656-fe1ba5ce12dd // indirect
 	github.com/gorilla/mux v1.6.2
 	github.com/gorilla/websocket v1.4.0 // indirect
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
 	github.com/hailocab/go-hostpool v0.0.0-20160125115350-e80d13ce29ed // indirect
-	github.com/hashicorp/consul v1.4.4
+	github.com/hashicorp/consul/api v1.1.0
```
csmarchbanks: Why did this get downgraded?

go.mod
```diff
@@ -27,39 +27,34 @@ require (
...
 	github.com/opentracing/opentracing-go v1.0.2
-	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/philhofer/fwd v0.0.0-20160129035939-98c11a7a6ec8 // indirect
 	github.com/pkg/errors v0.8.1
-	github.com/prometheus/alertmanager v0.13.0
-	github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829
-	github.com/prometheus/common v0.3.0
-	github.com/prometheus/prometheus v0.0.0-20190417125241-3cc5f9d88062
-	github.com/prometheus/tsdb v0.7.2-0.20190506134726-2ae028114c89
+	github.com/prometheus/alertmanager v0.17.0
```
csmarchbanks: This is a breaking change. All of the gossip/mesh protocols have changed in this upgrade. At a minimum I would want to see what is necessary to upgrade the alertmanager after this.

codesome: Prometheus uses v0.17.0 of alertmanager, hence this gets updated automatically if we want to update prometheus. Also the changes that I did to `pkg/alertmanager` was not final (was randomly making changes to fit the new API). It still needs testing and verifying.

go.mod
codesome: I did not see that! Let me check (also the package changed here).

codesome: So `v1.4.4` is of the consul package. But the latest release of `consul/api` is `v1.1.0` https://github.com/hashicorp/consul/releases/tag/api%2Fv1.1.0

codesome:
[This PR is still in a Draft stage as I have to verify the changes to alertmanager package and see that it works fine]

codesome:
Plugging this here https://github.com/prometheus/prometheus/issues/5813.

I am going to decouple alertmanager from prometheus and vendor it here instead of bumping the alertmanager version.

codesome:
Now I have vendored against [`remove-alertmanager`](https://github.com/prometheus/prometheus/tree/remove-alertmanager) branch of Prometheus. It only contains [this commit](https://github.com/prometheus/prometheus/commit/63ed2e28f1acfb8d880e5e95fbce4ab79a9379f7)  on top of Prometheus v2.11.1 which removes the hard dependency on alertmanager by copying only the required models.

This keeps the alertmanager version of cortex unchanged.

codesome:
In the same way, alertmanager 0.13.0 was not totally compatible with client_golang 1.0.0. And the latest Prometheus would pull it in.

I created another branch of alertmanager called [`ganesh-0.13`](https://github.com/prometheus/alertmanager/tree/ganesh-0.13) based on alertmanager 0.13.0, and added [this commit](https://github.com/prometheus/alertmanager/commit/6e135a0112f5c9ea6ebeea0d67db34df03da86ee) on top of it for missing parts (very hacky).

pkg/ruler/notifier.go
```diff
@@ -117,6 +117,7 @@ func buildNotifierConfig(rulerConfig *Config) (*config.Config, error) {
 		}
 	}
 	amConfig := &config.AlertmanagerConfig{
+		APIVersion:             config.AlertmanagerAPIVersionV2,
```
bboreham: this appears to break alerts completely, since the v2 endpoint is not wired up

PR #1511, opened 2019-07-13
feat(gcs): Add Service Account Support

grahamar:
To use GCP service accounts, the scope needs to be provided in the transport options.

Currently when trying to use this GCS client via Loki, by mounting the service account `credentials.json` file from a secret and using the `GOOGLE_APPLICATION_CREDENTIALS` environment variable, results in the following error response:

```
err="Post https://www.googleapis.com/upload/storage/v1/b/my-loki-bucket/o?alt=json&prettyPrint=false&projection=full&uploadType=multipart: oauth2: cannot fetch token: 400 Bad Request\nResponse: {\n  \"error\": \"invalid_scope\",\n  \"error_description\": \"Empty or missing scope not allowed.\"\n}"
```

Signed-off-by: Graham Rhodes <1964861+grahamar@users.noreply.github.com>


grahamar:
Great thanks @bboreham does it need more approvals or can it be merged by someone with write access? 😃

bboreham:
Two maintainers is standard rule.

csmarchbanks:
Thanks!

PR #1514, opened 2019-07-17
Add an ADOPTERS.md

gouthamve:
Some inspiration from: https://github.com/rook/rook/blob/master/ADOPTERS.md and https://github.com/jaegertracing/jaeger/blob/master/ADOPTERS.md

We can also split it into service providers and end users, but I prefer it this way tbh. I've added Aspen Mesh (blog post), EA and MayaData (CNCF presentation) as they are already public but, pinging @nrjpoddar @khaines @kmova for permission.

I'll ping others I know are using it for permission too.

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -1,0 +1,9 @@
+# Cortex Adopters
+
+This is the list of organisations that are using Cortex in **production environments** to power their metrics and monitoring systems. Please send PRs to add or remove organisations.
+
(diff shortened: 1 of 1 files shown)


kmova:
Thanks for already adding MayaData @gouthamve. The volume analytics provided on MayaOnline 
 (or as some of us would like to call it - OpenEBS Director) are powered by Cortex.

nrjpoddar:
Thanks @gouthamve for including Aspen Mesh. LGTM.

PR #1516, opened 2019-07-18
Reduce extra calls to Consul in HA mode

csmarchbanks:
Only check the value in Consul when the cached elected entry is not found or is older than the UpdateTimeout.

cc @cstyan @tomwilkie


cstyan:
This seems reasonable to me.

PR #1519, opened 2019-07-18
Add GoJek to the list of users

gouthamve:
Added with permission from @achilles42 who runs the cortex cluster there. They use it to collect their telegraf and prometheus metrics.


achilles42:
Thanks @gouthamve 👍

weeco:
You can also add REWE Digital GmbH - rewe-digital.com to the list.

gouthamve:
I've added it :)

weeco:
@gouthamve Thank you, the 'd' in "digital" should be small though :P

PR #1520, opened 2019-07-18
Add more tracing detail when parsing incoming requests

bboreham:
Change the sub-span into an event to reduce clutter.

The log lines show up as small black vertical lines in Jaeger - example of how it looks:
image


bboreham:
Yeah I started to file an issue about the flaky test but then got distracted.

PR #1526, opened 2019-07-23
Use pools to reduce allocs in distributor write path

gouthamve:
Motivated by:
```
(pprof) top
Showing nodes accounting for 12563.18GB, 75.10% of 16728.81GB total
Dropped 686 nodes (cum <= 83.64GB)
Showing top 10 nodes out of 107
      flat  flat%   sum%        cum   cum%
 3033.42GB 18.13% 18.13%  3033.42GB 18.13%  github.com/cortexproject/cortex/pkg/ingester/client.(*PreallocTimeseries).Unmarshal
 2412.37GB 14.42% 32.55%  2412.37GB 14.42%  github.com/cortexproject/cortex/pkg/ingester/client.(*WriteRequest).Marshal
 2264.33GB 13.54% 46.09%  2270.93GB 13.57%  github.com/cortexproject/cortex/vendor/google.golang.org/grpc/internal/transport.(*http2Client).Write
 1579.87GB  9.44% 55.53%  1579.87GB  9.44%  github.com/cortexproject/cortex/vendor/github.com/golang/snappy.Decode
  874.42GB  5.23% 60.76%   874.42GB  5.23%  bytes.makeSlice
  814.28GB  4.87% 65.63%   814.28GB  4.87%  regexp.(*bitState).reset
  465.64GB  2.78% 68.41%  3499.06GB 20.92%  github.com/cortexproject/cortex/pkg/ingester/client.(*WriteRequest).Unmarshal
  394.94GB  2.36% 70.77%  1184.50GB  7.08%  github.com/cortexproject/cortex/pkg/ring.DoBatch
  362.11GB  2.16% 72.94%   723.91GB  4.33%  github.com/cortexproject/cortex/pkg/ring.(*Ring).getInternal
  361.80GB  2.16% 75.10%   361.80GB  2.16%  github.com/cortexproject/cortex/pkg/ring.(*Ring).replicationStrategy
(pprof) list PreallocTimeseries
Total: 16.34TB
ROUTINE ======================== github.com/cortexproject/cortex/pkg/ingester/client.(*PreallocTimeseries).Unmarshal in github.com/cortexproject/cortex/pkg/ingester/client/timeseries.go
    2.96TB     2.96TB (flat, cum) 18.13% of Total
         .          .     43:	TimeSeries
         .          .     44:}
         .          .     45:
         .          .     46:// Unmarshal implements proto.Message.
         .          .     47:func (p *PreallocTimeseries) Unmarshal(dAtA []byte) error {
    2.37TB     2.37TB     48:	p.Labels = make([]LabelAdapter, 0, expectedLabels)
  606.91GB   606.91GB     49:	p.Samples = make([]Sample, 0, expectedSamplesPerSeries)
         .   512.62kB     50:	return p.TimeSeries.Unmarshal(dAtA)
         .          .     51:}
         .          .     52:
         .          .     53:// LabelAdapter is a labels.Label that can be marshalled to/from protos.
         .          .     54:type LabelAdapter labels.Label
```

This is from one of our distributor and it only gets worse as the traffic increases. I don't know if this is a sane approach though.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -391,7 +391,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 			localCtx = opentracing.ContextWithSpan(localCtx, sp)
 		}
 		return d.sendSamples(localCtx, ingester, timeseries)
-	})
+	}, func() { client.ReuseTS(req.Timeseries) })
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -391,7 +391,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 			localCtx = opentracing.ContextWithSpan(localCtx, sp)
 		}
 		return d.sendSamples(localCtx, ingester, timeseries)
-	}, func() { client.ReuseTS(req.Timeseries) })
+	}, func() { client.ReuseSlice(req.Timeseries) })
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/ingester/client/timeseries.go
+++ b/pkg/ingester/client/timeseries.go
@@ -16,24 +16,18 @@ var (
 	expectedLabels           = 20
 	expectedSamplesPerSeries = 10
 
-	// TSPool is the pool to return the timeseries to.
-	TSPool = sync.Pool{
+	slicePool = sync.Pool{
 		New: func() interface{} {
 			return make([]PreallocTimeseries, 0, expectedTimeseries)
 		},
 	}
 
-	// LabelAdapterPool is the pool to return the labels to.
-	LabelAdapterPool = sync.Pool{
+	timeSeriesPool = sync.Pool{
 		New: func() interface{} {
-			return make([]LabelAdapter, 0, expectedLabels)
-		},
-	}
-
-	// SamplePool is the pool to return the samples to.
-	SamplePool = sync.Pool{
-		New: func() interface{} {
-			return make([]Sample, 0, expectedSamplesPerSeries)
+			return TimeSeries{
+				Labels:  make([]LabelAdapter, 0, expectedLabels),
+				Samples: make([]Sample, 0, expectedSamplesPerSeries),
+			}
 		},
 	}
 )
@@ -56,7 +50,7 @@ type PreallocWriteRequest struct {
 
 // Unmarshal implements proto.Message.
 func (p *PreallocWriteRequest) Unmarshal(dAtA []byte) error {
-	p.Timeseries = TSPool.Get().([]PreallocTimeseries)
+	p.Timeseries = slicePool.Get().([]PreallocTimeseries)
 	return p.WriteRequest.Unmarshal(dAtA)
 }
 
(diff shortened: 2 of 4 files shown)


bboreham:
I think this could be OK.  You know the objects aren't used outside of `distributor.Push()`.

However:
 * `ingester.Push()` also unmarshals these objects; I think you can pool there too but it may need closer examination, particularly `Labels`
 * I don't like the asymmetry - can you put `ReuseTS()` as a method on `PreallocWriteRequest` ?

pkg/distributor/distributor.go
```diff
@@ -276,6 +276,8 @@ func (d *Distributor) checkSample(ctx context.Context, userID string, ts client.
 
 // Push implements client.IngesterServer
 func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*client.WriteResponse, error) {
+	defer client.ReuseTS(req.Timeseries)
```
csmarchbanks: All the pushes are very unlikely to all be done when this function finishes. This is due to us returning once a quorum of ingesters have been successfully written to, but we do not cancel the remaining pushes so that our replication factor is still accurate.

bboreham: Good point.  It is highly likely that all the data has been marshalled by this time, but not guaranteed. 
I'd think we do know enough in the tracker to fire the `ReuseTS` when all calls are done, though perhaps an added complication.

bboreham:
@gouthamve can you file an issue?  I have a fix in mind for the `ReadAll()` line too.

pkg/distributor/distributor.go
tomwilkie: I suspect this is actually relatively safe - the buf is used to serialise the outgoing request to the ingesters, which is most likely all done by here, even if the request is outstanding.

That being said, its not definitely safe.  You could attach the reuse to the `batchTracker` in `ring`, and have that call a callback when all the RPCs are guaranteed to have finished.

gouthamve:
image


Before and after this PR ^, I see a consistent 10% drop in the allocations from the improvements above. The lint is failing on `XXX_Unmarshal`, don't have a good idea on how to fix that.

tomwilkie:
@gouthamve what can we do about the lint?  Why do we need to use `XXX_Unmarshal`?

pkg/ingester/client/timeseries.go
```diff
@@ -14,6 +15,27 @@ var (
...
+			return make([]LabelAdapter, 0, expectedLabels)
+		},
+	}
+
+	// SamplePool is the pool to return the samples to.
+	SamplePool = sync.Pool{
+		New: func() interface{} {
+			return make([]Sample, 0, expectedSamplesPerSeries)
+		},
+	}
```
tomwilkie: Do we need 3 pools here?  Or can we get away with a single pool of TimeSeries, as they hold references to the Labels and Samples?

gouthamve: So, I think we'll still need 3 pools because, in the unmarshalling code we have:

https://github.com/cortexproject/cortex/blob/master/pkg/ingester/client/cortex.pb.go#L4515-L4518

This'll create a new `Timeseries` and append it. We need the preallocs happening on Unmarshal. Not sure how I can do that with a single pool.

tomwilkie: This is just a thought (you're probably right): in Timeseries.Unmarshall, if the slices are nill, we allocate new non-zero length ones.

These newly allocated timeseries will be added to slices that eventually end up in the pool (via the ReuseTS function).

I might test this out to see if it has the same effect.   Given remote write sends a fixes number of series, and we prealloc that number, I suspect the line you reference is never called.

gouthamve:
We need `XXX_Unmarshal` for this: https://github.com/gogo/protobuf/blob/4651555b1566cab497c790283c1cabd0b8ff0efc/proto/decode.go#L336-L338

Because `PreallocTimeseries` embeds TimeSeries, which implements `XXX_Unmarshal`, we'll never call the `Unmarshal`. Hence we'll never prealloc the timeseries. And that saves us 3% or more iirc.

Issue #1512, opened 2019-07-14
Embed metrics into structs and optional initialization


jtlisi:
The single binary version of cortex ends up reporting metrics for each component in cortex regardless of whether the module is enabled or not. At this point it would make sense to start embedding metrics into structs so that the initialization of those metrics is actually dependent on whether that module is deployed or not.

This can be done piecemeal one module or package at a time.

bboreham:
I support this idea; came here to add that we have some other metrics that come in via `promauto` from vendor'd code that never gets called.  Example `prometheus_remote_storage_highest_timestamp_in_seconds`

jtlisi:
Relevant Design Document: https://docs.google.com/document/d/100HSvgvaswVR96Bmz2BWIL3d9ds6wqNTVI7_vESgOtE/edit?usp=sharing

pracucci:
I think the solution we've already started adopting is to pass `prometheus.Registerer` to services and register metrics on that, avoiding `promauto`.

bboreham:
This issue says not to use `promauto`, while #2204 says to use `promauto`.

This one got closed as stale; does that mean `promauto` won?

pracucci:
> This one got closed as stale; does that mean `promauto` won?

I think we can close this issue. The pattern we're currently using (and looks working fine with downstream projects) is to take the `prometheus.Registerer` in input (eg. `NewSomething(reg prometheus.Registerer)` and then registering metrics via `promauto.With(reg)`.

bboreham:
Agreed: if we create the metrics in `NewSomething()` that achieves the aim of this issue.

Nitpick: the key is we are _creating_ metrics via `promauto.With()`, hence not needing to think about whether they need registering.

PR #1527, opened 2019-07-23
Embed Ingester metrics into struct

jtlisi:
Relates To: #1512

This PR embeds the metrics from the ingester into the ingester struct. This means they will only be initialized if an ingester struct is initialized.

I avoided embedding a prometheus Registerer into the config and passing it because we do not currently use non-default registerers anywhere in cortex. Down the line we may want to consider this approach.

### DIFF
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -219,7 +219,7 @@ func (t *Cortex) stopQuerier() error {
 
 func (t *Cortex) initIngester(cfg *Config) (err error) {
 	cfg.Ingester.LifecyclerConfig.ListenPort = &cfg.Server.GRPCListenPort
-	t.ingester, err = ingester.New(cfg.Ingester, cfg.IngesterClient, t.overrides, t.store)
+	t.ingester, err = ingester.New(cfg.Ingester, cfg.IngesterClient, t.overrides, t.store, prometheus.DefaultRegisterer)
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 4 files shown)


beorn7:
I think this change only addresses a small part of the problems. It's essentially only moving the registration into the constructor, but still registers everything with the global registry. That's certainly an improvement, but you could have it much simpler (by not using `promauto` metrics but the normal metrics from the `prometheus` package, and then explicitly register them with `prometheus.Register` in the constructor). The problem now is that you will get a panic if you call `ingester.New` twice (e.g. in a test), with no way of intercepting it.

There are many ways to do it, but I see a certain convergence onto something like in prometheus/tsdb. Check out https://github.com/prometheus/tsdb/blob/master/db.go#L426 and take it from there. They are also using a metrics `struct` there. Note how you can pass in `nil` as the `prometheus.Registerer`, which will just not register anything.

jtlisi:
@beorn7 Thanks for the feedback, I updated this PR to reflect your suggestions.

beorn7:
I'll have a detailed look ASAP (but perhaps not tonight… :sleeping: ).

csmarchbanks:
I like the pattern, but what is the scope you want to remove metrics for? It looks like you only removed the metrics from `ingester.go`, but there are lots of other metrics being registered by promauto or init in the ingester package.

beorn7:
I think this is just meant as an incremental step to try things out, with the goal of converting all metrics in the package to the same pattern eventually.

Issue #1529, opened 2019-07-24
Metrics for HA deduping


gouthamve:
As we start rolling out HA deduping, it's super hard to see if things are actually working or not. There is no visibility into it and we should have metrics as to how many samples were deduped, how many times the replica leader changed etc.

gouthamve:
Fixed in #1530

PR #1530, opened 2019-07-25
Add HA tracker metrics

cstyan:
cc @gouthamve

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -321,7 +321,9 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 			return nil, err
 		}
 		// If there wasn't an error but removeReplica is false that means we didn't find both HA labels
-		nonHASamples.WithLabelValues(userID).Add(float64(numSamples))
+		if !removeReplica {
+			nonHASamples.WithLabelValues(userID).Add(float64(numSamples))
+		}
 	}
 
 	// For each timeseries, compute a hash to distribute across ingesters;
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -270,6 +270,7 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica 
 	// If the sample doesn't have either HA label, accept it.
 	// At the moment we want to accept these samples by default.
 	if cluster == "" || replica == "" {
+		nonHASamples.WithLabelValues(userID).Inc()
 		return false, nil
 	}
 
@@ -298,22 +299,26 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	var lastPartialErr error
 	removeReplica := false
 
+	numSamples := 0
+	for _, ts := range req.Timeseries {
+		numSamples += len(ts.Samples)
+	}
+	receivedSamples.WithLabelValues(userID).Add(float64(numSamples))
+
 	if d.cfg.EnableHAReplicas && d.limits.AcceptHASamples(userID) && len(req.Timeseries) > 0 {
 		cluster, replica := findHALabels(d.limits.HAReplicaLabel(userID), d.limits.HAClusterLabel(userID), req.Timeseries[0].Labels)
 		removeReplica, err = d.checkSample(ctx, userID, cluster, replica)
 		if err != nil {
-			if _, ok := httpgrpc.HTTPResponseFromError(err); ok {
+			if resp, ok := httpgrpc.HTTPResponseFromError(err); ok && resp.GetCode() == 202 {
 				// These samples have been deduped.
+				numSamples := 0
 				for _, ts := range req.Timeseries {
-					dedupedSamples.WithLabelValues(userID, cluster).Add(float64(len(ts.Samples)))
+					numSamples += len(ts.Samples)
 				}
+				dedupedSamples.WithLabelValues(userID, cluster).Add(float64(numSamples))
 			}
+
 			return nil, err
-		}
-		// These samples were sent for a user ID that we've configured to accept HA samples from,
-		// but the samples didn't contain one or both of the required HA labels.
-		for _, ts := range req.Timeseries {
-			nonHASamples.WithLabelValues(userID).Add(float64(len(ts.Samples)))
 		}
 	}
 
@@ -321,7 +326,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	// check each sample and discard if outside limits.
 	validatedTimeseries := make([]client.PreallocTimeseries, 0, len(req.Timeseries))
 	keys := make([]uint32, 0, len(req.Timeseries))
-	numSamples := 0
+	numSamples = 0
 	for _, ts := range req.Timeseries {
 		// If we found both the cluster and replica labels, we only want to include the cluster label when
 		// storing series in Cortex. If we kept the replica label we would end up with another series for the same
@@ -360,7 +365,6 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 
 		numSamples += len(ts.Samples)
 	}
-	receivedSamples.WithLabelValues(userID).Add(float64(numSamples))
 
 	if len(keys) == 0 {
 		return &client.WriteResponse{}, lastPartialErr
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -13,7 +13,6 @@ import (
 	"golang.org/x/time/rate"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
-	"github.com/go-kit/kit/log/level"
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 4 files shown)


weeco:
https://github.com/cortexproject/cortex/issues/1529 should probably be referenced here

pkg/distributor/distributor.go
```diff
@@ -284,18 +292,34 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	var lastPartialErr error
 	removeReplica := false
 
+	numSamples := 0
+	for _, ts := range req.Timeseries {
+		numSamples += len(ts.Samples)
+	}
+	receivedSamples.WithLabelValues(userID).Add(float64(numSamples))
```
tomwilkie: We use this metric for billing.  By moving it here, customers will get charged for samples which are rejected.

tomwilkie: Given that, it looks like the whole ha tracking block might need moving below the validation block.

pkg/distributor/distributor.go
```diff
@@ -284,18 +292,34 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
+			if resp, ok := httpgrpc.HTTPResponseFromError(err); ok && resp.GetCode() == 202 {
+				// These samples have been deduped.
+				dedupedSamples.WithLabelValues(userID, cluster).Add(float64(numSamples))
+			}
+
 			return nil, err
 		}
+
+		if !removeReplica {
+			nonHASamples.WithLabelValues(userID).Add(float64(numSamples))
```
tomwilkie: I wonder wether it would be easier to maintain these metrics in the `checkSample` method?

pkg/ring/kv/etcd/etcd.go
```diff
@@ -82,6 +83,9 @@ func (c *Client) CAS(ctx context.Context, key string, f func(in interface{}) (ou
 		if err != nil {
 			level.Error(util.Logger).Log("msg", "error CASing", "key", key, "err", err)
 			if !retry {
+				if resp, ok := httpgrpc.HTTPResponseFromError(err); ok && resp.GetCode() != 202 {
+					level.Error(util.Logger).Log("msg", "error CASing", "key", key, "err", err)
+				}
```
tomwilkie: This looks like a layering violation; why should our etcd client special case 202s?  This should go in the ha tracker - that way it would behave the same with consul too.

cstyan: How could we move this into the HA tracker? With the current structure the returning of the 202 from `f(...)`has to be an error, right?

pkg/distributor/distributor.go
cstyan: The idea was to have a metric to compare total samples in vs # of samples deduped. I can add a new metric instead of using `receivedSamples`. 

Why would we need to move the HA tracking block below validation? For the deduped samples counting, so we only count samples if they user is valid/not above their limit? We'd still want the current block to know if we want to remove the replica label prior to iterating through each series in `req.Timeseries`.

pkg/distributor/ha_tracker.go
```diff
@@ -114,7 +141,13 @@ func (c *haTracker) loop(ctx context.Context) {
 		replica := value.(*ReplicaDesc)
 		c.electedLock.Lock()
 		defer c.electedLock.Unlock()
+		chunks := strings.SplitN(key, "/", 2)
+		if replica.Replica != c.elected[key].Replica {
+			electedReplicaChanges.WithLabelValues(chunks[0], chunks[1])
```
gouthamve: Needs an `Inc()`

pkg/distributor/distributor.go
```diff
@@ -285,10 +293,22 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
+				// These samples have been deduped.
+				for _, ts := range req.Timeseries {
+					dedupedSamples.WithLabelValues(userID, cluster).Add(float64(len(ts.Samples)))
+				}
+			}
 			return nil, err
 		}
+		// These samples were sent for a user ID that we've configured to accept HA samples from,
+		// but the samples didn't contain one or both of the required HA labels.
+		for _, ts := range req.Timeseries {
```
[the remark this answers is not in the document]
cstyan: I've added a new metric to count total samples in, before deduping or validation

pkg/distributor/distributor.go
```diff
@@ -45,6 +45,21 @@ var (
 		Name:      "distributor_received_samples_total",
 		Help:      "The total number of received samples.",
 	}, []string{"user"})
+	incommingSamples = promauto.NewCounterVec(prometheus.CounterOpts{
```
csmarchbanks: s/incomming/incoming

csmarchbanks: What do you think of updating the help text in `receivedSamples` to say that it only includes samples that have not been deduped?

pkg/distributor/distributor.go
```diff
@@ -301,7 +327,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	// check each sample and discard if outside limits.
 	validatedTimeseries := make([]client.PreallocTimeseries, 0, len(req.Timeseries))
 	keys := make([]uint32, 0, len(req.Timeseries))
-	numSamples := 0
+	numSamples = 0
```
csmarchbanks: Does `numSamples` need to be recounted? Is there any way that it would differ from the above calculation?

pkg/distributor/distributor.go
```diff
@@ -45,6 +45,21 @@ var (
 		Name:      "distributor_received_samples_total",
 		Help:      "The total number of received samples.",
 	}, []string{"user"})
+	incommingSamples = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "distributor_samples_in_total",
+		Help:      "The total number of samples that have come in to the distributor, including rejected or deduped samples.",
+	}, []string{"user"})
+	nonHASamples = promauto.NewCounterVec(prometheus.CounterOpts{
```
csmarchbanks: Unless I am misreading the behavior, this will only increment once per payload, so is really the total number of received requests that didn't contain both HA labels.

pkg/distributor/distributor.go
cstyan: Honestly I think this metric should be `receivedSamples` and the current `receivedSamples` should be `acceptedSamples` or `storedSamples`.

csmarchbanks: :+1: Good idea, that naming pattern is much more clear. I like `acceptedSamples` if the metric is increased before the push, and `storedSamples` if it is only increased when there are no errors writing to the ingesters. Which of those two options do you think would be more valuable?

pkg/distributor/distributor.go
```diff
@@ -45,6 +45,21 @@ var (
 		Name:      "distributor_received_samples_total",
 		Help:      "The total number of received samples.",
```
tomwilkie: Update this description to mention doesn't include rejected samples or deduped samples.

pkg/distributor/distributor.go
```diff
@@ -337,8 +363,6 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 				Samples: samples,
 			},
 		})
-
-		numSamples += len(ts.Samples)
```
tomwilkie: Not sure its safe to remove this; we previously didn't count invalid samples for instance.

pkg/distributor/distributor.go
```diff
@@ -290,18 +303,31 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
+	incomingSamples.WithLabelValues(userID).Add(float64(numSamples))
+
 	if d.cfg.EnableHATracker && d.limits.AcceptHASamples(userID) && len(req.Timeseries) > 0 {
-		removeReplica, err = d.checkSample(ctx, userID, req.Timeseries[0])
+		cluster, replica := findHALabels(d.limits.HAReplicaLabel(userID), d.limits.HAClusterLabel(userID), req.Timeseries[0].Labels)
+		removeReplica, err = d.checkSample(ctx, userID, cluster, replica)
 		if err != nil {
+			if resp, ok := httpgrpc.HTTPResponseFromError(err); ok && resp.GetCode() == 202 {
+				// These samples have been deduped.
+				dedupedSamples.WithLabelValues(userID, cluster).Add(float64(numSamples))
```
tomwilkie: Would this be clearer and less code to move it into `checkSample`?

pkg/ring/kv/etcd/etcd.go
tomwilkie: As discussed, lets decouple and move this to HA tracker, and not log here.

pkg/distributor/distributor.go
cstyan: fixed, we count total samples and valid samples now

pkg/distributor/distributor.go
cstyan: We'd need to pass `numSamples` into checkSample, and we'll always have to check for a response code of 202. If you feel strongly about moving this I'll change it.

tomwilkie: Oh yes I see.  Now this is only checked in one place, this is fine.

PR #1531, opened 2019-07-25
Update LICENSE

dankohn:
Please update the License text so that GitHub (which uses https://github.com/licensee/licensee)
can properly detect the license.


bboreham:
@dankohn thanks for the change.  Would you care to apply a signed-off line to your commit?

dankohn:
Fixed, thanks.

PR #1533, opened 2019-07-26
flush boltdb files to object stores

sandeepsukhani:
This can be useful for running cortex using just boltdb and gcs/s3.
Few details about implementation:
* Files are stored in a folder per periodic table and are named after ingester and startup timestamp so that integers can keep pushing latest changes by overriding same files
* Flushes files every 15 mins and before shutting down to make index available to other services
* New stores can be implemented easily by implementing methods required by store interface
* Ingester to also query store when using boltdb. Limits query to only last 30 mins when flushing boltdb to store is enabled.
* Queriers would keep syncing new and updated index files in a local folder which would be cleaned up by default every 24 hours.


pracucci:
@sandeepsukhani Could you rebase `master`, fix the linter and tests before I start the review, please?

pkg/chunk/local/archiver.go
```diff
@@ -0,0 +1,435 @@
...
+		err := a.syncTable(ctx, tableToSync, true)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (a *Archiver) syncTable(ctx context.Context, tableName string, sendUpdates bool) error {
```
pracucci: Do we really need this `sendUpdates` instead of always doing it?

pkg/ingester/ingester.go
```diff
@@ -454,18 +460,55 @@ func (i *Ingester) Query(ctx old_ctx.Context, req *client.QueryRequest) (*client
...
-			})
+			Samples: modelSamplePairsToClientSamples(values),
 		}
+
 		result.Timeseries = append(result.Timeseries, ts)
 		return nil
 	}, nil, 0)
-	i.metrics.queriedSeries.Observe(float64(numSeries))
+
+	if i.cfg.QueryStore {
```
pracucci: Could you explain me why we do want all this logic, please?

pkg/chunk/local/archiver.go
```diff
@@ -0,0 +1,435 @@
...
+		if isOK && !uploadedFileMtime.Before(file.mtime) {
+			continue
+		}
+
+		f, err := os.Open(file.path)
+		if err != nil {
+			return err
+		}
+
+		buf, err := ioutil.ReadAll(f)
```
pracucci: We're reading a file while BoltDB may simultaneously write on it. How can we guarantee the uploaded file is not corrupted?

pkg/chunk/local/archiver.go
```diff
@@ -0,0 +1,435 @@
...
+	downloadedTablesMtx sync.RWMutex
+
+	updatesChan           chan Update
+	uploadedFilesMtime    map[string]time.Time
+	uploadedFilesMtimeMtx sync.RWMutex
+	done                  chan struct{}
+
+	// We would use ingester name and startup timestamp for naming files while uploading so that
+	// ingester does not override old files when using same id
+	ingesterNameAndStartUpTs string
```
pracucci: We may rename it to something like `ingesterUniqueID` (the fact it's built upon ingester name + timestamp is an implementation detail, but what we need to store in the variable is a unique ID for the ingester).

sandeepsukhani:
Since Cortex now has TSDB support, this is not needed.
Support for custom index clients is added in #2049 for this to be implemented in Loki instead.

PR #1538, opened 2019-07-29
Adds v11 schema

cyriltovena:

We ran Loki for a while using the new `LabelNamesForMetricName` to retrieve all possibles labels. Turned out that it does pull too many chunks for a single query spanning over 6h (12k~)

This introduces a new schema where we store the metric (labels set) within the `IndexEntry.Value`. Other schema keep the same implementation.

Design doc & reason for this change: https://docs.google.com/document/d/1sVZHtQACLQZfiKnnFhXSiqyRx6YSX945s9jPjc2js6o/edit?usp=sharing

/cc @gouthamve

### DIFF
--- a/pkg/chunk/chunk_store_utils.go
+++ b/pkg/chunk/chunk_store_utils.go
@@ -42,11 +42,9 @@ func labelNamesFromChunks(chunks []Chunk) []string {
 	var result []string
 	for _, c := range chunks {
 		for _, l := range c.Metric {
-			if l.Name != model.MetricNameLabel {
-				if _, ok := keys[string(l.Name)]; !ok {
-					keys[string(l.Name)] = struct{}{}
-					result = append(result, string(l.Name))
-				}
+			if _, ok := keys[string(l.Name)]; !ok {
+				keys[string(l.Name)] = struct{}{}
+				result = append(result, string(l.Name))
 			}
 		}
 	}
--- a/pkg/chunk/schema.go
+++ b/pkg/chunk/schema.go
@@ -24,7 +24,7 @@ var (
 	seriesRangeKeyV1      = []byte{'7'}
 	labelSeriesRangeKeyV1 = []byte{'8'}
 	// For v11 schema
-	metricConstRangeKeyV1 = []byte{'9'}
+	labelNamesRangeKeyV1 = []byte{'9'}
 
 	// ErrNotSupported when a schema doesn't support that particular lookup.
 	ErrNotSupported = errors.New("not supported")
@@ -48,7 +48,7 @@ type Schema interface {
 
 	// If the query resulted in series IDs, use this method to find chunks.
 	GetChunksForSeries(from, through model.Time, userID string, seriesID []byte) ([]IndexQuery, error)
-	// Returns queries to retrieve all labels of multiple series by id.
+	// Returns queries to retrieve all label names of multiple series by id.
 	GetLabelNamesForSeries(from, through model.Time, userID string, seriesID []byte) ([]IndexQuery, error)
 }
 
@@ -687,30 +687,12 @@ func (s v10Entries) GetLabelWriteEntries(bucket Bucket, metricName string, label
 	// read first 32 bits of the hash and use this to calculate the shard
 	shard := binary.BigEndian.Uint32(seriesID) % s.rowShards
 
-	labelNames := make([]string, 0, len(labels))
-	for _, l := range labels {
-		if l.Name == model.MetricNameLabel {
-			continue
-		}
-		labelNames = append(labelNames, l.Name)
-	}
-	data, err := jsoniter.ConfigFastest.Marshal(labelNames)
-	if err != nil {
-		return nil, err
-	}
 	entries := []IndexEntry{
 		// Entry for metricName -> seriesID
 		{
 			TableName:  bucket.tableName,
 			HashValue:  fmt.Sprintf("%02d:%s:%s", shard, bucket.hashKey, metricName),
 			RangeValue: encodeRangeKey(seriesID, nil, nil, seriesRangeKeyV1),
-		},
-		// Entry for seriesID -> labels
-		{
-			TableName:  bucket.tableName,
-			HashValue:  string(seriesID),
-			RangeValue: encodeRangeKey(nil, nil, nil, metricConstRangeKeyV1),
-			Value:      data,
 		},
 	}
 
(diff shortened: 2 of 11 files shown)


weeco:
@cyriltovena Can you make the design doc accessible for the public?

cyriltovena:
Oups I though it was already sorry !

pkg/chunk/schema_test.go
```diff
@@ -333,3 +335,55 @@ func TestSchemaRangeKey(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkEncodeLabelsJson(b *testing.B) {
```
cyriltovena: This was a benchmark I used to decide if I should use json or string encoding for the labels set.

`.String()`/`promql.ParseMetric` is faster than `json.Marshal`/`json.Unmarshal`.

bboreham: In case you didn't notice, `pkg/chunk` has special code to speed up the latter via `github.com/json-iterator/go`

cyriltovena: I'll give it a look, thanks @bboreham .

cyriltovena:
This would also help advance #416 !

pkg/chunk/schema_test.go
cyriltovena: Wow 
```
BenchmarkEncodeLabelsString-16    	    1000	     13343 ns/op	    2029 B/op	      35 allocs/op
BenchmarkEncodeLabelsJson-16    	    1000	      1074 ns/op	     586 B/op	      18 allocs/op
```

@gouthamve I see you already approved I'm going to change to use json-iterator as suggested by @bboreham

cyriltovena: done !

pkg/chunk/schema.go
```diff
@@ -734,3 +774,47 @@ func (v10Entries) GetChunksForSeries(bucket Bucket, seriesID []byte) ([]IndexQue
...
+	}
+
+	return entries, nil
+}
+
+func (v11Entries) GetMetricsForSeries(bucket Bucket, seriesID []byte) ([]IndexQuery, error) {
+	return []IndexQuery{
+		{
+			TableName: bucket.tableName,
+			HashValue: bucket.hashKey + ":" + string(seriesID) + ":metric",
```
tomwilkie: Do you need to specify a range value here?

pkg/distributor/ha_tracker.go
```diff
@@ -172,6 +174,10 @@ func (c *haTracker) checkKVStore(ctx context.Context, key, replica string, now t
 	})
 }
 
+func replicasNotMatchError(replica, elected string) error {
+	return httpgrpc.Errorf(http.StatusAccepted, "replicas did not mach, rejecting sample: replica=%s, elected=%s", replica, elected)
+}
```
tomwilkie: Spurious change?

tomwilkie:
Is this covered by any end-to-end test?

cyriltovena:
It is covered by unit tests, could you point me at the end-to-end test ?

pkg/chunk/schema.go
cyriltovena: I added a `+ ":metric"` at the end of the HasValue, so I don't think I need the RangeValue to query. (From the doc above it means I'm getting all of them)

pkg/distributor/ha_tracker.go
cyriltovena: This appeared after I rebased amend a commit, I wanted to trick the CI to trigger a build as it was failing on a flaky test on the ruler.

So this is from master

pkg/chunk/schema.go
cyriltovena: However the `IndexEntry` does have a RangeValue as it seems required for validation. But it is just a constant.

pkg/distributor/ha_tracker.go
cyriltovena: I rebased it again and its gone !

pkg/chunk/schema.go
```diff
@@ -734,3 +774,47 @@ func (v10Entries) GetChunksForSeries(bucket Bucket, seriesID []byte) ([]IndexQue
...
+	}
+	entries := []IndexEntry{
+		// Entry for seriesID -> chunkID
+		{
+			TableName:  bucket.tableName,
+			HashValue:  bucket.hashKey + ":" + string(seriesID),
+			RangeValue: encodeRangeKey(encodedThroughBytes, nil, []byte(chunkID), chunkTimeRangeKeyV3),
+		},
+		// Entry for seriesID -> labels
+		{
```
gouthamve: We should move this into `GetLabelWriteEntries` as those will be cached and deduped. I don't see a reason for this to be in `GetChunkWriteEntries`.

cyriltovena:
@gouthamve I've added some improvements:

- The index entry is now only returned via GetLabelWriteEntries which means the entry is created only for a unique label set.
- I'm storing now only label names (not values) and skipping __name__ as it will always be there.
- deduping using a map.

pkg/chunk/schema.go
cyriltovena: done

gouthamve:
Rebased and calculated the overhead. This adds 7% to the index size: https://docs.google.com/spreadsheets/d/1admjIY_SyBN8q7FJdhumNtsM85yOWxTgfCUinRtYLwY/edit

Calculated using v10 as base: https://github.com/cortexproject/cortex/blob/98db0a4e4f007c2dbce35efae23bb37b5ee650db/pkg/chunk/schema.go#L642-L673

bboreham:
TIL that Google Sheets doesn't colour in the cells in a formula when you don't have edit permission.
So it's a bit hard to follow.

Came here to say that DynamoDB charges 100 bytes overhead per record, so the storage cost in both cases is quite a bit higher than the raw byte count, and the relative cost of 130 extra bytes in one record is much less.

gouthamve:
I've made the sheet editable, but given that it's even better for Dynamo, can this get an LGTM? I wanna first roll this out and then experiment with a v12 that doesn't require a `__name__`.

bboreham:
Should we have something in the docs saying which schemas are solid and which are experimental / not recommended ?

pkg/chunk/chunk_store_utils.go
```diff
@@ -42,11 +42,9 @@ func labelNamesFromChunks(chunks []Chunk) []string {
 	var result []string
 	for _, c := range chunks {
 		for _, l := range c.Metric {
-			if l.Name != model.MetricNameLabel {
-				if _, ok := keys[string(l.Name)]; !ok {
-					keys[string(l.Name)] = struct{}{}
-					result = append(result, string(l.Name))
-				}
+			if _, ok := keys[string(l.Name)]; !ok {
```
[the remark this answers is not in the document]
gouthamve: This is a separate change yes, this is to maintain some compatibility with the Prometheus API which returns `__name__`

pkg/chunk/schema_config.go
```diff
@@ -219,8 +219,12 @@ func (cfg *SchemaConfig) ForEachAfter(t model.Time, f func(config *PeriodConfig)
 
 // CreateSchema returns the schema defined by the PeriodConfig
 func (cfg PeriodConfig) CreateSchema() Schema {
-	var e entries
+	rowShards := uint32(16)
```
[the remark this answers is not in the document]
gouthamve: Hrm, so we load the schema from YAML. Now, we are defaulting it to `16`, but only here. On the other hand, we could loop through all the configs we loaded and set it to `16` after Unmarshaling, but I am not sure that is more cleaner than this.

gouthamve:
> Also, am I right in thinking the labels data is never cached?

No, it should be cached as far as I can see. Atleast the index lookups will be cached and that should be enough, no?

cyriltovena:
> > Also, am I right in thinking the labels data is never cached?
> 
> No, it should be cached as far as I can see. Atleast the index lookups will be cached and that should be enough, no?

I think index cache is enough, since we store the data in the index only.

pkg/chunk/chunk_store_utils.go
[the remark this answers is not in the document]
bboreham: Needs to be in the description and the changelog, no?

bboreham:
OK, I think I understand how the caching works now.
I would still like to see the blind alleys removed from the commits.

pkg/chunk/chunk_store_utils.go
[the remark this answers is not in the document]
gouthamve: There's no user facing changes here. The API is not being used by Cortex yet and I don't see the reason to update the changelog.

PR #1541, opened 2019-07-30
Add trace IDs to cached extents to help track down query gaps.

tomwilkie:
This is to help track down gaps in the queries.


pkg/querier/frontend/results_cache.go
```diff
@@ -107,6 +112,8 @@ func (s resultsCache) handleHit(ctx context.Context, r *QueryRangeRequest, exten
 		reqResps []requestResponse
 		err      error
 	)
+	log, ctx := spanlogger.New(ctx, "handleHit")
```
gouthamve: Do we want to add the same line to `handleMiss`?

pkg/querier/frontend/frontend.proto
```diff
@@ -62,5 +62,6 @@ message CachedResponse  {
 message Extent  {
 	int64 start = 1 [(gogoproto.jsontag) = "start"];
 	int64 end = 2 [(gogoproto.jsontag) = "end"];
-	APIResponse response = 3 [(gogoproto.jsontag) = "response"];
+  APIResponse response = 3 [(gogoproto.jsontag) = "response"];
```
bboreham: spurious change?

vendor/modules.txt
```diff
@@ -386,12 +386,8 @@ github.com/tinylib/msgp/msgp
 # github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5
 github.com/tmc/grpc-websocket-proxy/wsproxy
 # github.com/uber/jaeger-client-go v2.16.0+incompatible
-github.com/uber/jaeger-client-go/config
 github.com/uber/jaeger-client-go
-github.com/uber/jaeger-client-go/internal/baggage/remote
-github.com/uber/jaeger-client-go/internal/throttler/remote
-github.com/uber/jaeger-client-go/rpcmetrics
-github.com/uber/jaeger-client-go/transport
+github.com/uber/jaeger-client-go/config
```
bboreham: what happened here? looks like the file was reordered?

pkg/querier/frontend/results_cache.go
tomwilkie: I don't think so - the miss never modifies an existing extent.  In a hit we might, so we log the old trace ids.

vendor/modules.txt
tomwilkie: I needed this to pass lint.  No idea why.

PR #1543, opened 2019-07-31
Cut release v0.1.0-rc.0

csmarchbanks:
I added some of the notable changes that have happend over the last 4 weeks. Let me know if there is anything else you would like added to the first changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+## master / unreleased
+
+## 0.1.0-rc.0 / 2019-07-31
+
+* [CHANGE] HA Tracker flags were renamed to provide more clarity #1465
+  * `distributor.accept-ha-labels` is now `distributor.ha-tracker.enable`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+## master / unreleased
+
+## 0.1.0-rc.0 / 2019-07-31
+
+* [CHANGE] HA Tracker flags were renamed to provide more clarity #1465
```
weeco: Maybe explicitly mark this as Breaking Change?

csmarchbanks: The goal is for [CHANGE] to indicate that something is breaking. I followed how Prometheus does annotations, for example config being removed or metrics being renamed recently: https://github.com/prometheus/prometheus/blob/master/CHANGELOG.md

I guess there are different levels of breaking (metrics changing, vs something potentially not starting), but I think it could be hard to differentiate levels of changes.

PR #1545, opened 2019-07-31
roll back embedded metrics change

jtlisi:
This rolls back #1527, since this will lead to 3 separate approaches for instrumentation in cortex and this should be something discussed and agreed upon before moving forward with it.


csmarchbanks:
I added this as an agenda item for the community call tomorrow.

Issue #1522, opened 2019-07-19
Test flake in TestChunkStoreRandom


bboreham:
Failed at:
https://circleci.com/gh/cortexproject/cortex/7713
https://circleci.com/gh/cortexproject/cortex/7830
https://circleci.com/gh/cortexproject/cortex/8092

PR #1546, opened 2019-07-31
Add admin page for HA tracker information.

cstyan:
It wasn't clear to me what the right way to serve the page was, and whether for other pages (ex: the ring pages) if we serve the page from every distributor/ingester or not.

### DIFF
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -170,6 +170,7 @@ func (t *Cortex) initDistributor(cfg *Config) (err error) {
 
 	t.server.HTTP.HandleFunc("/all_user_stats", t.distributor.AllUserStatsHandler)
 	t.server.HTTP.Handle("/api/prom/push", t.httpAuthMiddleware.Wrap(http.HandlerFunc(t.distributor.PushHandler)))
+	t.server.HTTP.Handle("/ha-tracker", t.distributor.Replicas)
 	return
 }
 
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -139,7 +139,7 @@ type Config struct {
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.BillingConfig.RegisterFlags(f)
 	cfg.PoolConfig.RegisterFlags(f)
-	cfg.HATrackerConfig.RegisterFlags(f)
+	cfg.HATrackerConfig.RegisterFlags("distributor", f)
 
 	f.BoolVar(&cfg.EnableBilling, "distributor.enable-billing", false, "Report number of ingested samples to billing system.")
 	f.DurationVar(&cfg.RemoteTimeout, "distributor.remote-timeout", 2*time.Second, "Timeout for downstream ingesters.")
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -123,7 +123,6 @@ type Config struct {
 	BillingConfig billing.Config             `yaml:"billing,omitempty"`
 	PoolConfig    ingester_client.PoolConfig `yaml:"pool,omitempty"`
 
-	EnableHATracker bool            `yaml:"enable_ha_tracker,omitempty"`
 	HATrackerConfig HATrackerConfig `yaml:"ha_tracker,omitempty"`
 
 	RemoteTimeout       time.Duration `yaml:"remote_timeout,omitempty"`
@@ -143,7 +142,6 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.HATrackerConfig.RegisterFlags(f)
 
 	f.BoolVar(&cfg.EnableBilling, "distributor.enable-billing", false, "Report number of ingested samples to billing system.")
-	f.BoolVar(&cfg.EnableHATracker, "distributor.ha-tracker.enable", false, "Enable the distributors HA tracker so that it can accept samples from Prometheus HA replicas gracefully (requires labels).")
 	f.DurationVar(&cfg.RemoteTimeout, "distributor.remote-timeout", 2*time.Second, "Timeout for downstream ingesters.")
 	f.DurationVar(&cfg.ExtraQueryDelay, "distributor.extra-query-delay", 0, "Time to wait before sending more than the minimum successful query requests.")
 	f.DurationVar(&cfg.LimiterReloadPeriod, "distributor.limiter-reload-period", 5*time.Minute, "Period at which to reload user ingestion limits.")
@@ -180,13 +178,11 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 		quit:           make(chan struct{}),
 	}
 
-	if cfg.EnableHATracker {
-		replicas, err := newClusterTracker(cfg.HATrackerConfig)
-		if err != nil {
-			return nil, err
-		}
-		d.Replicas = replicas
+	replicas, err := newClusterTracker(cfg.HATrackerConfig)
+	if err != nil {
+		return nil, err
 	}
+	d.Replicas = replicas
 
 	go d.loop()
 
@@ -218,9 +214,7 @@ func (d *Distributor) loop() {
 func (d *Distributor) Stop() {
 	close(d.quit)
 	d.ingesterPool.Stop()
-	if d.cfg.EnableHATracker {
-		d.replicas.stop()
-	}
+	d.Replicas.stop()
 }
 
 func (d *Distributor) tokenForLabels(userID string, labels []client.LabelAdapter) (uint32, error) {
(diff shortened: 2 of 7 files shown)


cstyan:
cc @tomwilkie @gouthamve @jtlisi

pkg/cortex/modules.go
```diff
@@ -170,6 +170,7 @@ func (t *Cortex) initDistributor(cfg *Config) (err error) {
 
 	t.server.HTTP.HandleFunc("/all_user_stats", t.distributor.AllUserStatsHandler)
 	t.server.HTTP.Handle("/api/prom/push", t.httpAuthMiddleware.Wrap(http.HandlerFunc(t.distributor.PushHandler)))
+	t.server.HTTP.Handle("/ha-tracker", t.distributor.Replicas)
```
tomwilkie: `Replicas` is nil is HA tracking is disabled, so this will panic.

tomwilkie: In general, having this be nil when HA tracking is disabled is a pain.  I would always create the struct and internalise the enabled state within that struct.

pkg/distributor/ha_tracker_http.go
```diff
@@ -0,0 +1,81 @@
...
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<title>Cortex HA Tracker Status</title>
+	</head>
+	<body>
+		<h1>Cortex HA Tracker Status</h1>
+		<p>Current time: {{ .Now }}</p>
+		<form action="" method="POST">
+			<input type="hidden" name="csrf_token" value="$__CSRF_TOKEN_PLACEHOLDER__">
```
tomwilkie: This is only needed if the page has an "action" - ie a button.  Same with the form.

pkg/distributor/ha_tracker_http.go
```diff
@@ -0,0 +1,81 @@
...
+			<table width="100%" border="1">
+				<thead>
+					<tr>
+						<th>User ID</th>
+						<th>Cluster</th>
+						<th>Replica</th>
+						<th>Timestamp</th>
```
tomwilkie: When the lease runs out or when we last took the lease?  Both would be good.

pkg/distributor/ha_tracker_http.go
```diff
@@ -0,0 +1,81 @@
...
+					</tr>
+				</thead>
+				<tbody>
+					{{ range .Elected }}
+					<tr>
+						<td>{{ .UserID }}</td>
+						<td>{{ .Cluster }}</td>
+						<td>{{ .Replica }}</td>
+						<td>{{ .Timestamp }}</td>
```
tomwilkie: Does this render in a useable form?

tomwilkie:
Looks really good - couple minor nits, but hopefully merge tomorrow.

pkg/distributor/distributor.go
```diff
@@ -166,13 +164,11 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
...
-		replicas, err := newClusterTracker(cfg.HATrackerConfig)
-		if err != nil {
-			return nil, err
-		}
-		d.replicas = replicas
+	replicas, err := newClusterTracker(cfg.HATrackerConfig)
+	if err != nil {
+		return nil, err
 	}
+	d.Replicas = replicas
```
tomwilkie: Nit: I prefer if this goes above the building of the Distributor now.

pkg/distributor/distributor.go
```diff
@@ -109,7 +109,6 @@ type Config struct {
 	BillingConfig billing.Config             `yaml:"billing,omitempty"`
 	PoolConfig    ingester_client.PoolConfig `yaml:"pool,omitempty"`
 
-	EnableHATracker bool            `yaml:"enable_ha_tracker,omitempty"`
```
tomwilkie: Not sure you can remove this completely - the distributor won't start if you don't provide consul/etc address for the ha tracker.

The idea was to move this flag into the HA tracker config instead, and have that struct 'know' if it were enabled or not - and not start the consul client, loop etc if it is.

pkg/distributor/ha_tracker_http.go
```diff
@@ -0,0 +1,86 @@
...
+		electedReplicas = append(electedReplicas, struct {
+			UserID, Cluster, Replica            string
+			ElectedAt, UpdateTime, FailoverTime time.Time
+		}{
+			UserID:       chunks[0],
+			Cluster:      chunks[1],
+			Replica:      desc.Replica,
+			ElectedAt:    timestamp.Time(desc.ReceivedAt),
+			UpdateTime:   timestamp.Time(desc.ReceivedAt).Add(h.cfg.UpdateTimeout),
+			FailoverTime: timestamp.Time(desc.ReceivedAt).Add(h.cfg.FailoverTimeout),
```
tomwilkie: Might be worth having these as durations, so they print out a bit more concisely?

Are UpdateTimeout and FailoverTimeout configurable per user?

tomwilkie:
Couple more nits.

pkg/distributor/distributor.go
cstyan: :man_facepalming:

pkg/distributor/ha_tracker_http.go
cstyan: Update/FailoverTimeout aren't configurable per user. Whether we accept HA samples from a user and what cluster/replica flags to look for are.

cstyan: Durations via what call? `time.Until()`?

cstyan:
@tomwilkie PTAL. In the most recent commit I've reverted the removal of/moved the `EnableHATracker` cfg flag to the HA tracker cfg from the Distributor cfg, added a prefix flag to the HA tracker flag registration (so we don't have to manually prefix each of it's flags with `distributor.`), and moved any checks for `if EnableHATracker` from distributor.go to ha_tracker.go. 

This means the HA Tracker struct is always created, but it only creates the client to it's KV store if `EnableHATracker == true`. Also if there's a misconfiguration of `EnableHATracker == false` but `acceptHASamples == true` for a user ID, we'll still accept samples that contain both HA tracking labels (cluster/replica), the HA tracker `checkReplica` will tell the distributor to accept any sample if `EnableHATracker == false`.

pkg/distributor/ha_tracker_http.go
tomwilkie: Yeah, `Until`.

pkg/distributor/ha_tracker.go
```diff
@@ -80,27 +80,32 @@ type haTracker struct {
...
-	FailoverTimeout time.Duration
+	FailoverTimeout time.Duration `yaml:"ha_tracker_failover_timeout"`
 
 	KVStore kv.Config
 }
 
-// RegisterFlags adds the flags required to config this to the given FlagSet
-func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
+// RegisterFlags adds the flags required to config this to the given FlagSet. If prefix is not an empty string it should end with a period.
+func (cfg *HATrackerConfig) RegisterFlags(prefix string, f *flag.FlagSet) {
```
tomwilkie: TBH not sure the prefix is the right way to go here - its not like we're planning on using the HA tracker anywhere else, are we?

cstyan: Not that I can think of, but it makes the flag registration a bit cleaner and less chance of new flags missing the `distributor.` prefix imo. I can revert this if you disagree.

pkg/distributor/distributor.go
cstyan: I'm assuming you mean move the creation of the HA tracker to above the creation of the distributor, but leave this assignment. I've done that in the latest commit.

cstyan:
test failures look like a flake?

bboreham:
`TestChunkStoreRandom` failures noted at #1522

It _could_ be a real problem, but not one introduced by this PR.

pkg/distributor/ha_tracker.go
tomwilkie: We have to be really careful changing flag names.  For instance you dropped the dot between distributor and ha-tracker.  This has the potential to break rollouts and piss off users.

pkg/distributor/distributor.go
```diff
@@ -139,7 +139,7 @@ type Config struct {
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.BillingConfig.RegisterFlags(f)
 	cfg.PoolConfig.RegisterFlags(f)
-	cfg.HATrackerConfig.RegisterFlags(f)
+	cfg.HATrackerConfig.RegisterFlags("distributor.", f)
```
cstyan: @tomwilkie I'm confused, I had the prefix as `distributor.` but your commit removing the prefix shows it was just `distributor`

pkg/distributor/ha_tracker.go
cstyan: https://github.com/cortexproject/cortex/pull/1546/commits/8ffe4b9064485e5cc293892209bdb391c8257e9a#r312024987

pkg/distributor/distributor.go
tomwilkie: `¯\_(ツ)_/¯`

pkg/distributor/ha_tracker_http.go
```diff
@@ -0,0 +1,100 @@
...
+
+var trackerTmpl *template.Template
+
+func init() {
+	trackerTmpl = template.Must(template.New("ha-tracker").Parse(trackerTpl))
+}
+
+func (h *haTracker) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	h.electedLock.RLock()
+	defer h.electedLock.RUnlock()
```
gouthamve: Can we unlock right after we use it? I see we hold this lock until the network transfer is over and I see it as a potential DOS target, with someone deliberately holding the connection, as holding the lock will mean requests fail.

PR #1549, opened 2019-08-01
Refactored limits to be able to define them dynamically when Cortex is vendored in another project

sandlis:
This would be useful to define limits specific to Loki while still being able to use the same code for managing limits and per tenant overwrites.

Changes include:
1. Limits are now stored as key-value pairs, where the key is a string and values are interfaces for being able to support dynamic limits
2. Limits need to be registered before being able to use them
3. Sample and Labels validation methods now accept any type which implements required methods to get required limits for validation
4. Added Setter methods to change default limit values at runtime for testing purpose

### DIFF
--- a/pkg/util/validation/limits.go
+++ b/pkg/util/validation/limits.go
@@ -232,6 +232,8 @@ func (o *Overrides) CardinalityLimit(userID string) int {
 // Loads overrides and returns the limits as an interface to store them in OverridesManager.
 // We need to implement it here since OverridesManager must store type Limits in an interface but
 // it doesn't know its definition to initialize it.
+// We could have used yamlv3.Node for this but there is no way to enforce strict decoding due to a bug in it
+// TODO: Use yamlv3.Node to move this to OverridesManager after https://github.com/go-yaml/yaml/issues/460 is fixed
 func loadOverrides(filename string) (map[string]interface{}, error) {
 	f, err := os.Open(filename)
 	if err != nil {
--- a/pkg/util/validation/limits.go
+++ b/pkg/util/validation/limits.go
@@ -76,6 +76,8 @@ func (l *Limits) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// We want to set c to the defaults and then overwrite it with the input.
 	// To make unmarshal fill the plain data struct rather than calling UnmarshalYAML
 	// again, we have to hide it using a type indirection.  See prometheus/config.
+
+	// During startup we wont have a default value so we don't want to overwrite them
 	if defaultLimits != nil {
 		*l = *defaultLimits
 	}
@@ -222,6 +224,9 @@ func (o *Overrides) CardinalityLimit(userID string) int {
 	return o.overridesManager.GetLimits(userID).(*Limits).CardinalityLimit
 }
 
+// Loads overrides and returns the limits as an interface to store them in OverridesManager.
+// We need to implement it here since OverridesManager must store type Limits in an interface but
+// it doesn't know its definition to initialize it.
 func loadOverrides(filename string) (map[string]interface{}, error) {
 	f, err := os.Open(filename)
 	if err != nil {
@@ -239,8 +244,8 @@ func loadOverrides(filename string) (map[string]interface{}, error) {
 	}
 
 	overridesAsInterface := map[string]interface{}{}
-	for k := range overrides.Overrides {
-		overridesAsInterface[k] = overrides.Overrides[k]
+	for userID := range overrides.Overrides {
+		overridesAsInterface[userID] = overrides.Overrides[userID]
 	}
 
 	return overridesAsInterface, nil
--- a/pkg/util/validation/limits.go
+++ b/pkg/util/validation/limits.go
@@ -103,9 +103,14 @@ type Overrides struct {
 // become the new global defaults.
 func NewOverrides(defaults Limits) (*Overrides, error) {
 	defaultLimits = &defaults
+	overridesManagerConfig := OverridesManagerConfig{
+		OverridesReloadPeriod: defaults.PerTenantOverridePeriod,
+		OverridesLoadPath:     defaults.PerTenantOverrideConfig,
+		OverridesLoader:       loadOverrides,
+		Defaults:              &defaults,
+	}
 
-	overridesManager, err := NewOverridesManager(defaults.PerTenantOverridePeriod, defaults.PerTenantOverrideConfig,
-		loadOverrides, &defaults)
+	overridesManager, err := NewOverridesManager(overridesManagerConfig)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/util/validation/limits.go
+++ b/pkg/util/validation/limits.go
@@ -2,7 +2,10 @@ package validation
 
 import (
 	"flag"
+	"os"
 	"time"
+
+	"gopkg.in/yaml.v2"
 )
 
 // Limits describe all the limits for users; can be used to describe global default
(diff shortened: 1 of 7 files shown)


bboreham:
Can you say a bit more about what has been done in the 800 lines of changes? 
It looks like some words were renamed - does this relate to making things dynamic?

sandlis:
Exact usecase is being able to manage limits in Loki using same code as cortex. Loki uses some of the limits from Cortex and few additional ones specific to Loki.
`pkg/util/limits/limits.go` is to manage limit types and per tenant overwrites. 
`pkg/util/limits/validation.go` provides methods to access specific limits.
With this change Loki would use code for managing limits from limits.go and provide its own methods for specific limits.

bboreham:
For instance, it looks like you changed the package name "validation" to "limits".
Is that true?  If so, what was the reason?
Please ensure that your commit message describes all the changes made.

sandlis:
> what was the reason?

Actually, I wrote a new package with the name "limits" because most of its use case was managing and accessing limits. I had copied parts of code from the previous package so it looks like just a rename.
If you feel that it made more sense with the previous name, then I can rename it to validation.

> Please ensure that your commit message describes all the changes made.

Sure will do and will make sure I take care of them going forward. Sorry about that.

bboreham:
This isn't about the name, it's about you explaining what you did so it can be reviewed.
As it stands it's an 800+ line change.  Maybe if you didn't rename so much it would be a 100-line change and much easier to review.

sandlis:
@bboreham renamed and refactored the code to make the changes minimal. Please let me know if it looks good or whether I can do any more improvements to it.

pkg/util/validation/validate.go
```diff
@@ -65,8 +73,16 @@ func (cfg *Overrides) ValidateSample(userID string, metricName string, s client.
 	return nil
 }
 
+// LabelValidationConfig helps with getting required config to validate labels
+type LabelValidationConfig interface {
```
gouthamve: `*Config` suggests a struct and not an interface. Can we rename it to `LabelValidator`. Similarly for `SampleValidator`

pkg/util/validation/limits.go
```diff
@@ -2,78 +2,247 @@ package validation
...
+
+	return nil
+}
+
+func (l *Limits) getValue(limitName string, userID string) interface{} {
+	l.perTenantOverridesMtx.RLock()
+	defer l.perTenantOverridesMtx.RUnlock()
+
+	if l.perTenantOverrides != nil {
+		override, ok := l.perTenantOverrides[userID]
```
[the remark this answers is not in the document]
sandlis: There is a check for that in the previous line, is that what you meant?

sandlis:
@bboreham @tomwilkie Can you please have a look at this and suggest any changes if required to get another approval?

tomwilkie:
As discussed offline, not super happy with the amount of strings here.  Quick sketch of an alternative:

```go
package foo

type Limits interface {
    RegisterFlags()
    UnmarhsalYAML()
}

type Overrides struct {
    factory    func() Limits
    defaults   Limits
    overrides map[string]Limits
}

func (o Overrides) register() {]}
    o.defaults = o.factory()
    o.defaults.RegisterFlags()
}


func (o *Overrides) getFloat64(userID string, f func(*Limits) float64) float64 {
	o.overridesMtx.RLock()
	defer o.overridesMtx.RUnlock()
	override, ok := o.overrides[userID]
	if !ok {
		return f(&o.Defaults)
	}
	return f(override)
}

func loop() {
    // load the overrides file in the map above.
}

// Cortex 

type CortexLimits struct {
    IngestionRate          float64       `yaml:"ingestion_rate"`
    ...
}

type CortexOverrider struct {
    Overrides
}

func (o *CortexOverrider) IngestionRate(userID string) float64 {
	return o.getFloat64(userID, func(l *Limits) float64 {
        cl := l.(CortexLimits)
		return l.IngestionRate
	})
}

// Loki

type LokiLimits struct {
    ....
}

type LokiOverrider astruct {
    Overrides
}
```

pkg/util/validation/limits.go
```diff
@@ -73,7 +76,172 @@ func (l *Limits) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// We want to set c to the defaults and then overwrite it with the input.
 	// To make unmarshal fill the plain data struct rather than calling UnmarshalYAML
 	// again, we have to hide it using a type indirection.  See prometheus/config.
-	*l = defaultLimits
+	if defaultLimits != nil {
```
sandlis: Without this values were getting reset to 0 while loading cortex config which had just per tenant overrides config set.

pkg/util/validation/validate.go
```diff
@@ -50,8 +51,15 @@ func init() {
 	prometheus.MustRegister(DiscardedSamples)
 }
 
+// SampleValidationConfig helps with getting required config to validate sample.
+type SampleValidationConfig interface {
```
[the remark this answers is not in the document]
sandlis: We won't be able to use it in Loki then I guess.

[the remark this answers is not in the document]
tomwilkie: I quite like these interfaces actually!

pkg/util/validation/override.go
```diff
@@ -22,290 +19,86 @@ func init() {
...
-			overrides: map[string]*Limits{},
-			quit:      make(chan struct{}),
-		}, nil
-	}
-
-	overrides, err := loadOverrides(defaults.PerTenantOverrideConfig)
-	if err != nil {
-		return nil, err
+// NewOverridesManager creates an instance of OverridesManager and starts reload overrides loop based on config
+func NewOverridesManager(perTenantOverridePeriod time.Duration, overridesReloadPath string, overridesLoader OverridesLoader,
```
tomwilkie: I think this would be nicer as a config struct, not individual args.

sandlis:
@tomwilkie Thanks for the review! Done the suggested changes.
Waiting for a fix for broken modules. Will rebase this when it is merged in master to run the checks.

pkg/util/validation/limits.go
```diff
@@ -73,7 +76,177 @@ func (l *Limits) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		return nil, err
+	}
+
+	overridesAsInterface := map[string]interface{}{}
+	for userID := range overrides.Overrides {
+		overridesAsInterface[userID] = overrides.Overrides[userID]
+	}
+
+	return overridesAsInterface, nil
+}
```
tomwilkie: Can we make this take a "factory" for the `Limits` type?  Then have the code to make a map vs the defaults would be in override.go.

I think this is possible with `yaml.v3` and `Node`: https://godoc.org/gopkg.in/yaml.v3

tomwilkie: As discussed offline, `Node.Decode` doesn't honour `KnownFields`: https://github.com/go-yaml/yaml/issues/460

pkg/util/validation/override.go
```diff
@@ -22,290 +19,87 @@ func init() {
...
-	o.overridesMtx.RLock()
-	defer o.overridesMtx.RUnlock()
-	override, ok := o.overrides[userID]
-	if !ok {
-		return f(&o.Defaults)
-	}
-	return f(override)
-}
+// GetLimits returns Limits for a specific userID if its set otherwise the default Limits
+func (om *OverridesManager) GetLimits(userID string) interface{} {
```
tomwilkie: What happends when the overrides are only partially specified?

ie if defaults are:
```yaml
a: 1
b: 2
```

and the overrides are:
```yaml
foo:
  b: 3
```

`GetLimits("foo")` will return:
```yaml
a: 0
b: 3
```
not
```yaml
a: 1
b: 3
```

tomwilkie: Not it won't ignore me.

pkg/util/validation/limits.go
tomwilkie: Why did this not happen before?

tomwilkie: @sandlis explained offline:
1. we init the flags, which sets up the default in a Limits struct in `cortex` pkg.
2. we then load YAML, which will overwrite them with 0s, as `defaultLimits` has not been inited.
3. we Parse the flags, which set only ones we've specifices
4. we then set this `defaultLimits` to the half-zero'd struct.

Issue #731, opened 2018-03-02
After N-1 ingester crashes, query results are unstable


bboreham:
We replicate to three ingesters but after losing two of them in an Unfortunate Incident, queries would sometimes contain results from just the two restarted instances which didn't have recent history.

It rights itself once those chunks age out of the remaining ingester into the store, but not great in the meantime.

csmarchbanks:
This should be fixed by https://github.com/weaveworks/cortex/pull/732 correct?

bboreham:
Different problem - #732 is about dropping samples on writes, while this is about reads from ingesters that have crashed and restarted.

bboreham:
To improve the user experience, we could get the remaining ingesters to flush those chunks for which they hold the only remaining copy.

So, if I'm ingester X, and I know ingesters A and B have restarted, I iterate over all series in memory, run the distributor hash, and every series that maps to (A, B, X) I should flush.

How do I get to know A and B have restarted?  I guess that could be written as metadata into the ring, or we could just have a human type it in via an admin entrypoint.

Slight wrinkle: we will get the wrong hash for series with a blank label value, since we discard that information on entry to the ingester.

bboreham:
How about:

We have ingesters remember the time their data starts, and hand that time over on a transfer.
If an ingester gets a query covering further back than that time, they return http status 206 - partial content.
Change the querier to not cancel remaining requests on 206.

bboreham:
I think we can get into the same bad situation if you add a new ingester and then one old ingester restarts - there are now two places to get an incomplete answer.  And my suggestion at https://github.com/cortexproject/cortex/issues/731#issuecomment-435362741 helps there too.

gouthamve:
Fixed by https://github.com/cortexproject/cortex/pull/1103

gouthamve:
One possible fix is to use `shard-by-all-labels`, but even with that, if you're having 3 ingesters, and 2 ingesters go down and come back, we can ignore a (we only wait for 2 replys) different ingester in different queries, hence will have unstable queries.

harry671003:
> Fixed by #1103

@gouthamve Was this issues fixed only for blocks store?

bboreham:
The WAL can be used for chunks, but I think Goutham’s subsequent comment is saying it isn’t fixed.

khaines:
[...]

Issue #612, opened 2017-11-23
Make distributors/ingesters aware of availability zones


bboreham:
Suppose your replication-factor is about the same as the number of AZs.  You'd prefer that replicas for any particular write were spread across all the AZs.

I have no idea how hard this is.

tomwilkie:
Cassandra does this with topology-aware snitches[1] and replication strategies[2]

[1] https://docs.datastax.com/en/cassandra/2.1/cassandra/architecture/architectureSnitchesAbout_c.html
[2] https://docs.datastax.com/en/cassandra/2.1/cassandra/architecture/architectureDataDistributeReplication_c.html

weeco:
I believe Kafka's approach is similarly to Cassandra's. They attach a Rack ID on each broker pod (which would be the identifier for a AV Zone): https://issues.apache.org/jira/browse/KAFKA-1215

khaines:
going to revive this issue and do some investigating

PR #1553, opened 2019-08-02
A start at a guide to running Cortex in production

bboreham:
This is something a lot of people ask about.

Probably it will take months to get into decent shape but the sooner we get started the better, so I jotted down some points.

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -11,5 +11,5 @@ interface and humans sending queries from GUIs, supply credentials
 which identify them and confirm they are authorised.
 
 When configuring the remote_write API in Prometheus there is no way to
-add extra headers. The http user and password fields can be user to
-convey tenant ID and/or authentication credentials.
+add extra headers. The user and password fields of http Basic auth, or
+Bearer token, can be used to convey tenant ID and/or credentials.
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -1,0 +1,15 @@
+# Authentication and Authorisation
+
+All Cortex components take the tenant ID from a header `X-Scope-OrgID`
+on each request. They trust this value completely: if you need to
+protect your Cortex installation from accidental or malicious calls
+then you must add an additional layer of protection.
+
+Typically this means you run Cortex behind a reverse proxy, and ensure
+that all callers, both machines sending data over the remote_write
+interface and humans sending queries from GUIs, supply credentials
+which identify them and confirm they are authorised.
+
+When configuring the remote_write API in Prometheus there is no way to
+add extra headers. The http user and password fields can be user to
+convey tenant ID and/or authentication credentials.
--- a/docs/running.md
+++ b/docs/running.md
@@ -103,7 +103,8 @@ Now, some rules of thumb:
 
 If you turn on compression between distributors and ingesters (for
 example to save on inter-zone bandwidth charges at AWS) they will use
-significantly more CPU.
+significantly more CPU (approx 100% more for distributor and 50% more
+for ingester).
 
 ### Caching
 
--- a/docs/running.md
+++ b/docs/running.md
@@ -39,7 +39,7 @@ Memcached is not essential but highly recommended.
 ### Ingester replication factor
 
 The standard replication factor is three, so that we can drop one
-sample and be unconcerned, as we still have two copies of the data
+replica and be unconcerned, as we still have two copies of the data
 left for redundancy. This is configurable: you can run with more
 redundancy or less, depending on your risk appetite.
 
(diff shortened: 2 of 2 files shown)


docs/running.md
```diff
@@ -0,0 +1,143 @@
...
+ of series in ingesters is number of active series times the
+ replication factor. This is with the default of 12-hour chunks - RAM
+ required will reduce if you set `-ingester.max-chunk-age` lower
+ (trading off more back-end database IO)
+ 2. Each 100,000 samples/sec arriving takes 1 CPU in distributors.
+ Distributors don't need much RAM.
+
+If you turn on compression between distributors and ingesters (for
+example to save on inter-zone bandwidth charges at AWS) they will use
+significantly more CPU.
```
weeco: Maybe add a ballpark for the additional CPU usage? For me it roughly doubled the CPU usage.

docs/running.md
```diff
@@ -0,0 +1,143 @@
...
+            memory: 10Gi
+```
+
+The specific values here should be adjusted based on your own
+experiences running Cortex - they are very dependent on rate of data
+arriving and other factors such as series churn.
+
+### Spread out ingesters
+
+Don't run multiple ingesters on the same node, as that raises the risk
```
weeco: What do you think about adding a quick note how ingesters transfer data (one pod will be spawned additionally, while the other is in the terminating state and transfer the data to the new pod)? Additionally I'd mention here that an operator always wants to avoid losing or moving ingesters as they are semi stateful. This impacts how you want to distribute the ingesters across zones, nodes and what machines you want to use (e. g. no preemptible/spot instances etc).

bboreham: I think hand-over would go along with basic documentation on ingesters, which we also need; it's not particularly a production thing.
Agreed on the rest.

bboreham: #1560 for hand-over doc

docs/running.md
```diff
@@ -0,0 +1,173 @@
...
+[architecture](architecture.md) document.
+
+## Planning
+
+### Tenants
+
+If you will run Cortex as a multi-tenant system, you need to give each
+tenant a unique ID - this can be any string. Managing tenants and
+allocating IDs must be done outside of Cortex. You must also configure
+[Authentication and Authorisation](auth.md).
```
csmarchbanks: I don't see `auth.md` in master or in this PR?

docs/running.md
```diff
@@ -0,0 +1,173 @@
...
+Cortex requires a scalable storage back-end.  Commercial cloud options
+are DynamoDB and Bigtable: the advantage is you don't have to know how
+to manage them, but the downside is they have specific costs.
+Alternatively you can choose Cassandra, which you will have to install
+and manage.
+
+### Components
+
+Every Cortex installation will need Distributor, Ingester and Querier.
+Alert-manager, Ruler and Query-frontend are optional.
```
csmarchbanks: Nit: Alertmanager is usually one word, happens in the next section as well.

docs/running.md
```diff
@@ -0,0 +1,173 @@
...
+If you want to configure recording and alerting rules (i.e. if you
+will run the Ruler and Alert-manager components) then a Postgres
+database is required to store configs.
+
+Memcached is not essential but highly recommended.
+
+### Ingester replication factor
+
+The standard replication factor is three, so that we can drop one
+sample and be unconcerned, as we still have two copies of the data
```
rade: Did you mean 'replica' here instead of 'sample'?

docs/running.md
bboreham: bah, forgot to `git add`

bboreham:
I think I addressed all review comments.

docs/running.md
```diff
@@ -0,0 +1,209 @@
...
+
+```
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 80
+```
+
+We do not recommend configuring a liveness probe on ingesters -
+killing them is a last resort and should not be left to a machine.
```
weeco: Since you are talking about how you want to avoid losing Ingesters, what do you think about adding a note about the impacts if you still lose more than `sampleReplicas` ingesters. Either temporarily (e. g. ingester pods have network issues (zone failure) and therefore can temporarily not be queried and ingested to) or permanently (e. g. one zone has a power outage).

Speaking of zones you might want to add a short note about availability across zones and probably link that issue too: https://github.com/cortexproject/cortex/issues/612 .

docs/auth.md
```diff
@@ -0,0 +1,15 @@
...
+on each request. They trust this value completely: if you need to
+protect your Cortex installation from accidental or malicious calls
+then you must add an additional layer of protection.
+
+Typically this means you run Cortex behind a reverse proxy, and ensure
+that all callers, both machines sending data over the remote_write
+interface and humans sending queries from GUIs, supply credentials
+which identify them and confirm they are authorised.
+
+When configuring the remote_write API in Prometheus there is no way to
```
weeco: Implicitly you can define headers in the remote write API in prometheus by setting a bearer token (which then will be used in the `Authorization` header). This is what we use and we also embed the tenant id inside of it and sign the token on our cortex gateway.

bboreham: Is that different to "http user and password" ?

docs/running.md
bboreham: #731 makes this embarrassing to write about.  Typically the pattern is I fix an issue rather than document it.

docs/auth.md
```diff
@@ -0,0 +1,15 @@
...
+add extra headers. The http user and password fields can be user to
```
csmarchbanks: s/user/used

docs/auth.md
csmarchbanks: username/password and bearer tokens are just different forms of data in the Authorization header. I guess to be complete you could say "The bearer_token or username and password fields can be set to convey..."

bboreham:
From #1560 we should also say what the "unhealthy" state of ingesters means and what to do about it.
Also mention that auto-scaling ingesters is, er, brave.

bboreham:
I'm going to merge this even though it's very incomplete, so it gives some value.
I invite everyone to add small amounts to improve it.

PR #1559, opened 2019-08-06
Lock around alertmanagers map everywhere

bboreham:
Seems odd to have a mutex and to use it in some places but not others.


tomwilkie:
LGTM

Issue #1307, opened 2019-03-28
Ingesters do not wait long enough for transfer


bboreham:
We used to tell ours to wait for two minutes, but that was torn up by #1185 

Today I see this:
```
level=info ts=2019-03-28T16:41:45.673567131Z caller=lifecycler.go:456 msg="changing ingester state from" old_state=ACTIVE new_state=LEAVING
level=error ts=2019-03-28T16:41:45.688616369Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:45.695897001Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:45.746735562Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:45.807703134Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:46.334636223Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:46.46819466Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:47.212946695Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:47.370324186Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:48.672456187Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:51.35543268Z caller=transfer.go:199 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"
level=error ts=2019-03-28T16:41:51.355469649Z caller=lifecycler.go:466 msg="Failed to transfer chunks to another ingester" err="terminated after 10 retries"
```

so it isn't even waiting the 100ms minimum I expect from cursory examination of the code.

Possibly related: #792

[...]

PR #1560, opened 2019-08-06
Document the ingester hand-over process

bboreham:
Suggested at https://github.com/cortexproject/cortex/pull/1553#discussion_r310160832

### DIFF
--- a/docs/ingester-handover.md
+++ b/docs/ingester-handover.md
@@ -1,0 +1,44 @@
+# Ingester Hand-over
+
+The [ingester](architecture.md#ingester) holds several hours of sample
+data in memory. When we want to shut down an ingester, either for
+software version update or to drain a node for maintenance, this data
+must not be discarded.
+
+Each ingester goes through different states in its lifecycle. When
+working normally, the state is `ACTIVE`.
+
+On start-up, an ingester first goes into state `PENDING`. After a
+short time, if nothing happens, it adds itself to the ring and goes
+into state ACTIVE.
+
+A running ingester is notified to shut down by Unix signal
+`SIGINT`. On receipt of this signal it goes into state `LEAVING` and
+looks for an ingester in state `PENDING`. If it finds one, that
+ingester goes into state `JOINING` and the leaver transfers all its
+in-memory data over to the joiner. On successful transfer the leaver
+removes itself from the ring and exits and the joiner changes to
+`ACTIVE`, taking over ownership of the leaver's
+[ring tokens](architecture.md#hashing).
+
+If a leaving ingester does not find a pending ingester, it will flush
+all of its chunks to the backing database, then remove itself from the
+ring and exit. This may take tens of minutes to complete.
+
+During hand-over, neither the leaving nor joining ingesters will
+accept new samples. Distributors are aware of this, and "spill" the
+samples to the next ingester in the ring. This creates a set of extra
+"spilled" chunks which will idle out and flush after hand-over is
+complete. The sudden increase in flush queue can be alarming!
+
+The following metrics can be used to observe this process:
+
+ - `cortex_member_ring_tokens_owned` - how many tokens each ingester thinks it owns
+ - `cortex_ring_tokens_owned` - how many tokens each ingester is seen to own by other components
+ - `cortex_ring_member_ownership_percent` same as `cortex_ring_tokens_owned` but expressed as a percentage
+ - `cortex_ring_members` - how many ingesters can be seen in each state, by other components
+ - `cortex_ingester_sent_chunks` - number of chunks sent by leaving ingester
+ - `cortex_ingester_received_chunks` - number of chunks received by joining ingester
+ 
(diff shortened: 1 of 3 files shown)


docs/ingester-handover.md
```diff
@@ -0,0 +1,44 @@
...
+must not be discarded.
+
+Each ingester goes through different states in its lifecycle. When
+working normally, the state is `ACTIVE`.
+
+On start-up, an ingester first goes into state `PENDING`. After a
+short time, if nothing happens, it adds itself to the ring and goes
+into state ACTIVE.
+
+A running ingester is notified to shut down by Unix signal
```
csmarchbanks: This paragraph should mention the `-ingester.claim-on-rollout` flag.

docs/ingester-handover.md
```diff
@@ -0,0 +1,44 @@
+# Ingester Hand-over
+
+The [ingester](architecture.md#ingester) holds several hours of sample
+data in memory. When we want to shut down an ingester, either for
+software version update or to drain a node for maintenance, this data
+must not be discarded.
+
+Each ingester goes through different states in its lifecycle. When
+working normally, the state is `ACTIVE`.
```
tomwilkie: May want to mention these states can be seen on the /ring page on the querier and distributor.

docs/ingester-handover.md
```diff
@@ -0,0 +1,44 @@
...
+A running ingester is notified to shut down by Unix signal
+`SIGINT`. On receipt of this signal it goes into state `LEAVING` and
+looks for an ingester in state `PENDING`. If it finds one, that
+ingester goes into state `JOINING` and the leaver transfers all its
+in-memory data over to the joiner. On successful transfer the leaver
+removes itself from the ring and exits and the joiner changes to
+`ACTIVE`, taking over ownership of the leaver's
+[ring tokens](architecture.md#hashing).
+
+If a leaving ingester does not find a pending ingester, it will flush
```
tomwilkie: May want to mention this timeout is tuneable with a flag (`-ingester.join-after`

docs/ingester-handover.md
```diff
@@ -0,0 +1,44 @@
...
+data in memory. When we want to shut down an ingester, either for
+software version update or to drain a node for maintenance, this data
+must not be discarded.
+
+Each ingester goes through different states in its lifecycle. When
+working normally, the state is `ACTIVE`.
+
+On start-up, an ingester first goes into state `PENDING`. After a
+short time, if nothing happens, it adds itself to the ring and goes
+into state ACTIVE.
```
tomwilkie: Tuneable with `-ingester.join-after`

docs/ingester-handover.md
tomwilkie: Ah no, its `ingester.max-transfer-retries` - it will try, backoff for 5 secs (max).

tomwilkie:
LGTM with a few nits.

docs/ingester-handover.md
weeco: If you add that property can you tell under which circumstances you may want to tune it?
Why wouldn't I like to have a new ingester to immediately join the ring (i. e. setting `-ingester.join-after` to 0). Would that introduce the risk the ingester is considered as additional ingester (rather than the new ingester which is supposed to receive the data from the terminating ingester?) or why does that property even exist?

I am missing some more context to understand what is happening there.

docs/ingester-handover.md
bboreham: Yeah if you win the lottery.  See #1307

docs/ingester-handover.md
bboreham: The leaving ingester probes periodically for a joining one - need to allow time for this to happen.  I guess it's tuneable because we don't trust our initial guesses.

Personally I don't want to litter the main narrative with all the tuning options - maybe add a pointer to their description in `arguments.md` ?

csmarchbanks: I think linking to `arguments.md` makes a lot of sense for the detailed parameters.

docs/ingester-handover.md
bboreham: #1566 to remove it

bboreham:
I added the two args.

docs/ingester-handover.md
```diff
@@ -0,0 +1,45 @@
...
+[short time](arguments.md#ingester), if nothing happens, it adds
+itself to the ring and goes into state ACTIVE.
+
+A running ingester is notified to shut down by Unix signal
+`SIGINT`. On receipt of this signal it goes into state `LEAVING` and
+looks for an ingester in state `PENDING`. If it finds one, that
+ingester goes into state `JOINING` and the leaver transfers all its
+in-memory data over to the joiner. On successful transfer the leaver
+removes itself from the ring and exits and the joiner changes to
+`ACTIVE`, taking over ownership of the leaver's
```
achilles42: Should we also mention about `unhealthy` ingester state as well?

PR #1561, opened 2019-08-07
Add VERSION file

csmarchbanks:
This will mean you must add a commit when releasing different versions which will cause a different cache in the build to be used.


gouthamve:
:+1:

bboreham:
How exactly does this cause a different cache to be used?

PR #955, opened 2018-08-24
Don't succeed a transfer if no series are received.

tomwilkie:
Fixes #954

We don't know how we got in this state initially, but I suspect it was caused by an ingester being killed quickly - it started a transfer, but didn't actually get a chance to send any series.  This triggered the EOF break, and the new ingester joined the ring with 0 tokens.  

These checks should cause the new ingester to reject the transfer request if it doesn't receive a ingester ID or any series, leaving the ingester to join the ring itself.

Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>

PR #1563, opened 2019-08-08
Recieving no series is a valid case

gouthamve:
If we're doing an upgrade in a completely new cluster, the ingesters are
not going to send anything and this'll just slow the rollout process and
cause un-necessary alerts.


tomwilkie:
Hmmm, we added that because of a cascading failure: #955

bboreham:
There's a test to stop a completely blank ingester sending anything: https://github.com/cortexproject/cortex/blob/master/pkg/ingester/transfer.go#L210

Can you explain how you get into this state?  Maybe create a test?

PR #1564, opened 2019-08-08
check for retention period to be a multiple of periodic table duration and relevant test

sandlis:
When the retention period is not a multiple of periodic table duration, we need to error out since Cortex doesn't yet support partial deletion of tables.
Added a check to validate retention against the latest config(assuming the last one is the latest) and relevant test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [FEATURE] Add option to use jump hashing to load balance requests to memcached #1554
 * [FEATURE] Add status page for HA tracker to distributors #1546
+* [CHANGE] Added check for retention period to be a multiple of periodic table duration #1564
 
 ## 0.1.0 / 2019-08-07
 
@@ -13,3 +14,4 @@
 * [FEATURE] You can specify "heap ballast" to reduce Go GC Churn #1489
 * [BUGFIX] HA Tracker no longer always makes a request to Consul/Etcd when a request is not from the active replica #1516
 * [BUGFIX] Queries are now correctly cancelled by the query-frontend #1508
+* [CHANGE] Added check for retention period to be a multiple of periodic table duration
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## master / unreleased
 
+* [CHANGE] Retention period should now be a multiple of periodic table duration #1564
 * [FEATURE] Add option to use jump hashing to load balance requests to memcached #1554
 * [FEATURE] Add status page for HA tracker to distributors #1546
-* [CHANGE] Retention period should now be a multiple of periodic table duration #1564
 
 ## 0.1.0 / 2019-08-07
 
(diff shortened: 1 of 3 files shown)


bboreham:
Rather than silently changing what the user asked for, we could just make this an error.

CHANGELOG.md
```diff
@@ -10,3 +10,4 @@
 * [FEATURE] You can specify "heap ballast" to reduce Go GC Churn #1489
 * [BUGFIX] HA Tracker no longer always makes a request to Consul/Etcd when a request is not from the active replica #1516
 * [BUGFIX] Queries are now correctly cancelled by the query-frontend #1508
+* [CHANGE] Added check for retention period to be a multiple of periodic table duration
```
gouthamve: This should be under `master / unreleased` and not under `0.1.0`.

sandlis: Sorry, my bad. Fixed it.

pkg/chunk/table_manager.go
```diff
@@ -126,6 +127,11 @@ type TableManager struct {
 // NewTableManager makes a new TableManager
 func NewTableManager(cfg TableManagerConfig, schemaCfg SchemaConfig, maxChunkAge time.Duration, tableClient TableClient,
 	objectClient BucketClient) (*TableManager, error) {
+	// Assume the newest config is the one to use for validation of retention
+	if cfg.RetentionPeriod != 0 && cfg.RetentionPeriod%schemaCfg.Configs[len(schemaCfg.Configs)-1].IndexTables.Period != 0 {
```
tomwilkie: This checks they are not zero.  Should it test to ensure they are multiples, like the error says?

sandlis: Sorry, I think you missed seeing that `%` operator between all the text or maybe I am not able to see the problem that you are seeing?

tomwilkie: Yeah you are right.  Sorry.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * [FEATURE] Add option to use jump hashing to load balance requests to memcached #1554
 * [FEATURE] Add status page for HA tracker to distributors #1546
+* [CHANGE] Added check for retention period to be a multiple of periodic table duration #1564
```
gouthamve: This is user-facing so something like `Retention period should now be a multiple of periodic table duration` should better.

CHANGELOG.md
```diff
@@ -13,3 +14,4 @@
 * [FEATURE] You can specify "heap ballast" to reduce Go GC Churn #1489
 * [BUGFIX] HA Tracker no longer always makes a request to Consul/Etcd when a request is not from the active replica #1516
 * [BUGFIX] Queries are now correctly cancelled by the query-frontend #1508
+* [CHANGE] Added check for retention period to be a multiple of periodic table duration
```
gouthamve: It's back

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * [FEATURE] Add option to use jump hashing to load balance requests to memcached #1554
 * [FEATURE] Add status page for HA tracker to distributors #1546
+* [CHANGE] Retention period should now be a multiple of periodic table duration #1564
```
gouthamve: Changes should be placed before `[FEATURE]`

PR #325, opened 2017-03-09
Faster rolling upgrades.

tomwilkie:
Fixes #321 

It works!  And its quick...

TODO
- [x] move ingester lifecycle code into ingester, work out the coupling a bit better
- [x] move ingester shutdown code into ingester (from main.go)
- [x] fallback to flush it transfer fails.
- [x] naming
- [x] check ring logic for new ingester states
- ~ensure ingester flushing stops before transferring chunks~ - shouldn't do this, as we don't want chunks that won't flush to block the transfer.
- ~add an ever-extending timeout on the receiving ingester, when we do timeout auto-join~ - going to leave this for now, can always kill ingesters if they get stuck, will fall back to flushing.

PR #1566, opened 2019-08-08
Remove claim-on-rollout option

bboreham:
As discussed in [Slack](https://cloud-native.slack.com/archives/CCYDASBLP/p1565106143201700)

> **Chris Marchbanks**   Related to #1560, What do people think of defaulting the flag `-ingester.claim-on-rollout` to true, maybe even deprecating the flag. Does anyone run with that flag off?
> **Bryan Boreham**  I've just looked back to see that the flag came in with #325, where the whole rolling updates thing was added.
> So it was reasonable at the time to have a way to turn it off, but I think it's past its usefulness.

Leave the flag and yaml names for now to avoid crashing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.1.0 / 2019-08-07
 
+* [CHANGE] --claim-on-rollout flag deprecated; feature is now always on #1566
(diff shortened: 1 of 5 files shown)


tomwilkie:
Test failure looks legit?

tomwilkie:
Changelog entry?  Otherwise merge away.

PR #1569, opened 2019-08-09
Add a histogram for cached value sizes.

tomwilkie:


pkg/chunk/chunk_store_utils.go
```diff
@@ -108,6 +108,7 @@ type decodeResponse struct {
 
 // NewChunkFetcher makes a new ChunkFetcher.
 func NewChunkFetcher(cfg cache.Config, cacheStubs bool, storage ObjectClient) (*Fetcher, error) {
+	cfg.Prefix = "chunks"
```
bboreham: This has the effect that the `name` label in `cortex_cache_hits` changes from "memcache" to "chunksmemcache".
Which is a (minor) breaking change that should go in the changelog.

Issue #1396, opened 2019-05-17
Ruler should be protected against high-cardinality output


bboreham:
Suppose someone creates a rule, either recording rule or alert, that generates 100,000 output series.

Currently, all series will be sent in one request, which will hit the distributor rate-limit (defaults to 50,000 burst size) and be dropped.

This creates problems:
 - ruler consumes a lot of CPU and memory
 - it's for the operator hard to figure out what happened.  The log line (`caller=manager.go:539 msg="rule sample appending failed" err="rpc error: code = Code(429) desc = ingestion rate limit (50000) exceeded while adding 100000 samples"`) doesn't include the tenant ID
 - it's completely invisible to the user of the tenant

I'm thinking ruler should cap the size of its output, and generate some signal (a synthetic series, perhaps?) that can be used to know when the cap was hit.

If we _want_ to handle outputs from rules in the hundreds of thousands, we should batch them up so they don't choke the distributor.

The channel to alertmanager is also limited: `caller=notifier.go:371 msg="Alert batch larger than queue capacity, dropping alerts" num_dropped=30973`

jtlisi:
I agree! The implementation is going to be a bit tricky I believe. I think we are going to have to write our own rule manager instead of using the prom upstream.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

bboreham:
Perhaps this could be implemented via [engineQueryFunc](https://github.com/cortexproject/cortex/blob/106aa42093a6286852fcb5cddffb6e8b978c9d1e/pkg/ruler/compat.go#L71).
Limits could be integrated with the existing limit/overrides system.

bboreham:
Possibly addressed by prometheus/prometheus#9260

jeromeinsf:
@krishnateja325 something you are looking at?

krishnateja325:
yes, pulled-in https://github.com/prometheus/prometheus/pull/9260 and added support for limit field in this PR: https://github.com/cortexproject/cortex/pull/5528

PR #1571, opened 2019-08-09
Update Ruler to use upstream Prom Rule Manager

### LINKED ISSUE #477: Ruler doesn't report success/failure of rule evaluation
This is partly due to the underlying Prometheus library API. We should at least suggest some upstream changes. 

Having this information would make debugging issues in a running cortex deployment much easier.

### LINKED ISSUE #493: Config updates cause alerting rules to forget firing state
If a config set is updated, all the rules it declares will forget their firing state, even if the rules are unchanged.

jtlisi:
This PR is a refactor of #1532 to utilize the [Prometheus Rule Manager](https://github.com/prometheus/prometheus/blob/master/rules/manager.go) to schedule and evaluate rule groups.

Overview
---------
* Utilize the upstream prometheus manager to scheduler and evaluate rules
* Sync and map rules from an external service (configdb) to a temporary file which is synced with the users prometheus rules manager
* Utilize [afero](https://github.com/spf13/afero) to map rule files to allow for mocked tests using in memory filesystems and potentially using an in-memory file system in production if upstream changes can be made
* Utilize a protobuf interchange format for rules to allow for the effectively have a versionable format for storing and communicating rules

Fixes #477
Fixes #493


pkg/configs/client/client.go
```diff
@@ -163,3 +119,52 @@ func (c ConfigsResponse) GetLatestConfigID() configs.ID {
 	}
 	return latest
 }
+
+// ListAllRuleGroups polls the configdb server and returns the updated rule groups
+func (c *ConfigDBClient) ListAllRuleGroups(ctx context.Context) (map[string]rules.RuleGroupList, error) {
```
joe-elliott: Did we drop the since functionality for a reason?  Any concern for extremely large rule sets that this will return very large amounts of data.  Otherwise I like removing it for simplicity.

pkg/configs/client/client.go
```diff
@@ -163,3 +119,52 @@ func (c ConfigsResponse) GetLatestConfigID() configs.ID {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return newRules, nil
+}
+
+// decomposeGroupSlug breaks the group slug from Parse
+// into it's group name and file name
+func decomposeGroupSlug(slug string) (string, string) {
```
joe-elliott: unreferenced?

pkg/configs/client/client.go
joe-elliott: Should this method continue returning `map[string]configs.VersionedRulesConfig` and the translation to the new type occur in `pkg/ruler/rules`?

pkg/configs/client/client.go
```diff
@@ -163,3 +119,52 @@ func (c ConfigsResponse) GetLatestConfigID() configs.ID {
...
+		if cfg != nil {
+			configs[id] = *cfg
+		}
+	}
+
+	newRules := map[string]rules.RuleGroupList{}
+
+	for user, cfg := range configs {
+		userRules := rules.RuleGroupList{}
+		if cfg.IsDeleted() {
```
joe-elliott: Is this like a diff of rules you're parsing and rebuilding the final state?

pkg/ruler/ruler.go
```diff
@@ -188,66 +162,50 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
...
-
-	if n, ok := r.notifiers[userID]; ok {
-		n.stop()
+	level.Info(util.Logger).Log("msg", "stopping user managers")
+	wg := sync.WaitGroup{}
+	r.userManagerMtx.Lock()
+	for user, manager := range r.userManagers {
+		level.Debug(util.Logger).Log("msg", "shutting down user  manager", "user", user)
+		go func(manager *promRules.Manager, user string) {
+			wg.Add(1)
```
joe-elliott: is this a race condition?  should `wg.Add(1)` be outside the go routine so we are guaranteed it runs before `wg.Wait()`?

pkg/ruler/ruler.go
```diff
@@ -188,66 +162,50 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable storage.Queryable, d
 // Stop stops the Ruler.
 // Each function of the ruler is terminated before leaving the ring
 func (r *Ruler) Stop() {
+	close(r.done)
+	<-r.terminated
```
joe-elliott: this line waits on the terminated channel to stop before proceeding.  this channel is closed by the `run()` method which calls `loadRules()` in a loop.  `loadRules()` and `Stop()` both use the `userManagerMtx` to lock each other out of the `userManagers` map.  Is this necessary?  If `Stop()` is guaranteed to run independent of `loadRules()` I think dropping the mutex for clarity would be good.

pkg/configs/client/client.go
bboreham: I would say `since` was for large numbers of tenants, regardless of size of rule set.  And I am a bit nervous about dropping it.

pkg/ruler/ruler.go
```diff
@@ -368,6 +297,143 @@ func (r *Ruler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		case <-tick.C:
+			r.loadRules()
+		}
+	}
+}
+
+func (r *Ruler) loadRules() {
+	ringHasher := fnv.New32a()
+
+	configs, err := r.store.ListAllRuleGroups(context.Background())
```
joe-elliott: `loadRules()` could create a context at the top and pass it into `newManager()` so that the same context is used in both places since this functionality is connected.  unsure if that would connect traces or do anything valuable.

pkg/ruler/compat.go
```diff
@@ -22,17 +24,21 @@ type Pusher interface {
...
 	samples []client.Sample
+	userID  string
 }
 
 func (a *appendableAppender) Appender() (storage.Appender, error) {
 	return a, nil
 }
 
 func (a *appendableAppender) Add(l labels.Labels, t int64, v float64) (uint64, error) {
+	a.Lock()
```
joe-elliott: Why are we locking here? None of the Appender implementations in prometheus lock suggesting these functions are not reentrant.

pkg/configs/client/client.go
jtlisi: I do understand the concern. The main reason since was dropped is because the entire ruleset must be hashed when generating mapped files. Otherwise changes to the ring will not be reflected in the scheduled evaluations. However, polling such a large payload is not ideal. One option would be for the config client to cache the previous response and have an internal usage of the since variable. That way it can keep an up-to-date set of the active rules configs and only poll for changes.

pkg/configs/client/client.go
jtlisi: Good catch, I'll remove this.

pkg/configs/client/client.go
jtlisi: Yes, that is essentially what is going on.

pkg/ruler/ruler.go

pkg/ruler/ruler.go
jtlisi: I think in this case it would be best to maintain the mutex. Especially because it will likely be needed to implement the `/api/v1/rules` prometheus endpoint which returns the following json:

```json
{
    "status": "success",
    "data": {
        "groups": [
            {
                "name": "alertmanager.rules",
                "file": "/etc/prometheus/alerts.rules",
                "rules": [
                    {
                        "name": "AlertmanagerFailedReload",
                        "query": "alertmanager_config_last_reload_successful{job=\"default/alertmanager\"} == 0",
                        "duration": 600,
                        "labels": {
                            "severity": "warning"
                        },
                        "annotations": {
                            "message": "Reloading Alertmanager's configuration has failed for {{ $labels.namespace }}/{{ $labels.pod}}."
                        },
                        "alerts": [],
                        "health": "ok",
                        "type": "alerting"
                    },
...
```

In order to get the health of indvidual rules we are going to have to allow for concurrent access to the userManager map.

pkg/ruler/compat.go
jtlisi: A lock is required here because we pool the samples for a user to the same appendable.

pkg/ruler/ruler.go
jtlisi: 👍 I'm not sure there is any benefits currently but down the line we could consider using a context for top level cancellation.

pkg/configs/client/client.go
joe-elliott: Per discussions I'm re-adding `since` support now.

joe-elliott:
@jtlisi I restored the configs Client functionality to what it was before and made a `ConfigRuleStore` in `rules/store.go`.  This implements the `RuleStore` interface and bridges the gap to the config Client by maintaining state.

This will be easy to extend in the future by adding other implementations such as `GCSRuleStore` or whatever.

pkg/ruler/rules/store.go
```diff
@@ -0,0 +1,106 @@
...
+			continue
+		}
+		ruleMap[g.Namespace] = append(ruleMap[g.Namespace], FromProto(g))
+
+	}
+	return ruleMap
+}
+
+// ConfigRuleStore is a concrete implementation of RuleStore that sources rules from the config service
+type ConfigRuleStore struct {
```
jtlisi: This implementation looks good to me. @bboreham are you ok with abstracting `since` into a concrete type that lives behind the interface? That way a full set of rules can be returned on each poll. However, only new rules will be polled from the config service? I don't think we can work around polling the entire ruleset currently since the horizontally sharded ruler will need to hash each rule group to ensure it is evaluating the appropriate set of rules.

bboreham: The thing that I worry about is like this: say we have 10,000 rule groups across all tenants, and one tenant changes one of them, does the program do 10,000 things or 1 thing?

joe-elliott: Feel free to correct any errors @jtlisi.

As it is currently designed, once a polling cycle each ruler will calculate a hash for every rule group to determine which groups it should process locally.  This will happen 10,000 times per ruler in your scenario.

Next it will this subset of the rule groups and compare them to locally stored files on disk only updating those files on disk that have changed.  This will happen 10,000 / n times per ruler.  Where n is the number of rulers.

Then, if any files have changed or been added for a given user, it will clear the old prometheus rules manager for that user and build a new one pointed at the new set of files.

Even if we do not perform this process once a polling cycle we would need to at least do this process when certain events happen (such as a ruler joining or leaving the ring).  I believe @jtlisi was preferring the straightforward nature of this approach.

joe-elliott:
@jtlisi I rewrote the `MapRules` method as discussed.  Please review when you get a chance.

jtlisi:
> @jtlisi I rewrote the `MapRules` method as discussed. Please review when you get a chance.

LGTM

pkg/configs/client/client.go
```diff
@@ -76,24 +81,29 @@ func (c configsClient) GetRules(ctx context.Context, since configs.ID) (map[stri
...
 	req, err := http.NewRequest("GET", endpoint, nil)
 	if err != nil {
 		return nil, err
 	}
 
 	client := &http.Client{Timeout: timeout}
+
+	start := time.Now()
 	resp, err := client.Do(req)
+	if resp != nil {
```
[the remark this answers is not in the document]
joe-elliott: I assume it was written this way to include failed requests in the `configs_request_duration_seconds` metric.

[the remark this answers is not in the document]
gouthamve: Yes, but `resp` is nil when `err != nil` :) Atleast in the cases where we are using it :)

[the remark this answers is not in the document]
joe-elliott: I see.  You're correct.  

Pushing a change that moves the metric after the err check and removes the `resp != nil` check

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
 * [ENHANCEMENT] Allocation improvements in adding samples to Chunk. #1706
 * [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
 * [ENHANCEMENT] Consul client can optionally rate-limit itself during Watch (used e.g. by ring watchers) and WatchPrefix (used by HA feature) operations. Rate limiting is disabled by default. New flags added: `--consul.watch-rate-limit`, and `--consul.watch-burst-size`. #1708
+* [CHANGE] Flags changed with transition to upstream Prometheus rules manager:
+  * `ruler.client-timeout` is now `ruler.configs.client-timeout` in order to match `ruler.configs.url`
+  * `ruler.group-timeout`has been removed
+  * `ruler.num-workers` has been removed
+  * `ruler.rule-path` has been added to specify where the prometheus rule manager will sync rule files
+  * `ruler.storage.type` has beem added to specify the rule store backend type, currently only the configdb.
```
pracucci: `beem` > ` been`

pkg/configs/configs.go
```diff
@@ -182,6 +182,23 @@ func (c RulesConfig) Parse() (map[string][]rules.Rule, error) {
 	}
 }
 
+// ParseFormatted returns the rulefmt map of a users rules configs. It allows
+// for rules to be mapped to disk and read by the prometheus rules manager.
+func (c RulesConfig) ParseFormatted() (map[string]rulefmt.RuleGroups, error) {
```
bboreham: Does this deal with rules format "v1" (as the old `Parse()` function did) ?

pkg/ruler/compat.go
bboreham: I think previously in Cortex we created a separate `appendableAppender` for each group, which was run on a single goroutine.  So there must also be some change that means we have more goroutines talking to the same `appendableAppender` now?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
 * [ENHANCEMENT] Allocation improvements in adding samples to Chunk. #1706
 * [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
 * [ENHANCEMENT] Consul client can optionally rate-limit itself during Watch (used e.g. by ring watchers) and WatchPrefix (used by HA feature) operations. Rate limiting is disabled by default. New flags added: `--consul.watch-rate-limit`, and `--consul.watch-burst-size`. #1708
+* [CHANGE] Flags changed with transition to upstream Prometheus rules manager:
+  * `ruler.client-timeout` is now `ruler.configs.client-timeout` in order to match `ruler.configs.url`
+  * `ruler.group-timeout`has been removed
+  * `ruler.num-workers` has been removed
```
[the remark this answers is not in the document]
jtlisi: I updated it so the flags aren't removed and are instead deprecated with a message.

pkg/configs/configs.go
jtlisi: I added V1 support with tests cases.

pkg/configs/client/client.go
```diff
@@ -64,27 +81,30 @@ func (c configsClient) GetRules(ctx context.Context, since configs.ID) (map[stri
...
 	}
 
 	client := &http.Client{Timeout: timeout}
+
+	start := time.Now()
 	resp, err := client.Do(req)
 	if err != nil {
 		return nil, err
 	}
+	configsRequestDuration.WithLabelValues(operation, resp.Status).Observe(time.Since(start).Seconds())
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/rules/rules.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
jtlisi: The use of protos made more sense before I split this out from a larger PR a few months back. The proto format is used to store rule groups in a denormalized way in an object store backend. It also is used to communicate between rulers to fulfill the `/api/v1/rules` endpoint that reports the status of rules with their rule health. Since each ruler only knows the state of rules it is currently responsible for it needs to communicate with each ruler in the ring to get a complete view of rule health. To implement this feature a GRPC service is implemented by each ruler.

pkg/ruler/compat.go
jtlisi: Yea the rule groups for each user will all share an appendableAppender. There are some advantages  to this approach. Primarily, it will make  #1396 easier to solve since output limits for a user can be configured in the same place.

CHANGELOG.md
```diff
@@ -12,6 +12,13 @@
 * [ENHANCEMENT] Allocation improvements in adding samples to Chunk. #1706
 * [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
 * [ENHANCEMENT] Consul client can optionally rate-limit itself during Watch (used e.g. by ring watchers) and WatchPrefix (used by HA feature) operations. Rate limiting is disabled by default. New flags added: `--consul.watch-rate-limit`, and `--consul.watch-burst-size`. #1708
+* [CHANGE] Flags changed with transition to upstream Prometheus rules manager:
```
gouthamve: Changes should be on top.

jtlisi: All set

jtlisi:
@bboreham I fixed V1 rule loading and refactored based on your comments. This should be good for a second look.

pkg/cortex/cortex.go
```diff
@@ -109,7 +109,7 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 
 	c.Ruler.RegisterFlags(f)
 	c.ConfigDB.RegisterFlags(f)
-	c.ConfigStore.RegisterFlags(f)
+	c.ConfigStore.RegisterFlagsWithPrefix("alertmanager.", f)
```
khaines: changing flag names is a breaking change for existing users. Is this really necessary to change? If it is, can we please call it out in the changelog so the release owner can highlight this fact in the appropriate release notes?

jtlisi: The actual flags don't change in this case: https://github.com/cortexproject/cortex/pull/1571/files#diff-e8adc6214ca82bcbffe97a18e4753bc3L29

Only the functions in which they are registered, so this should be a noop for users.

khaines: found the deprecated flags later in the review... resolving

pkg/ruler/ruler.go
```diff
@@ -92,70 +68,84 @@ type Config struct {
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.LifecyclerConfig.RegisterFlagsWithPrefix("ruler.", f)
+	cfg.StoreConfig.RegisterFlags(f)
+
+	// Deprecated Flags that will be maintained to avoid user disruption
+	flagext.DeprecatedFlag(f, "ruler.client-timeout", "This flag has been renamed to ruler.configs.client-timeout")
```
khaines: `ruler.configs.url` appears to be missing from this list of deprecated flags. It was part of the deleted `pkg/configs/client/config.go` file.

jtlisi: This flag also still exists. It got move a bit and is registered in `pkg/ruler/storage.go` https://github.com/cortexproject/cortex/pull/1571/files#diff-16c509ab46b783eb193e10999f09ed31R21

PR #1103, opened 2018-10-31
Use TSDB's WAL for writes.

tomwilkie:
As per https://docs.google.com/document/d/1n1HcdgmsqaVwqxKg_POZqIREOSqZIAlrxuyx4YYV_pA/edit

Fixes #12

PR #1575, opened 2019-08-12
Remove pkg/ingester/wal.pb.go

codesome:
This got added in https://github.com/cortexproject/cortex/pull/1510 by mistake (File originates from https://github.com/cortexproject/cortex/pull/1103)


codesome:
The test fail seems to be flaky. Have seen the same test fail in other PRs.

PR #1585, opened 2019-08-16
s3: force tls

thorfour:
DigitalOcean spaces while S3 compatible doesn't play nice with using http as the scheme with which it is accessed.

This changes adds a `s3.force-tls` flag to change the s3config endpoint from `http` to `https`

Flag turned off by default for backwards compatibility.

### DIFF
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -128,6 +128,7 @@ type StorageConfig struct {
 	DynamoDBConfig
 	S3               flagext.URLValue
 	S3ForcePathStyle bool
+	S3ForceTLS       bool
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
@@ -137,6 +138,7 @@ func (cfg *StorageConfig) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&cfg.S3, "s3.url", "S3 endpoint URL with escaped Key and Secret encoded. "+
 		"If only region is specified as a host, proper endpoint will be deduced. Use inmemory:///<bucket-name> to use a mock in-memory implementation.")
 	f.BoolVar(&cfg.S3ForcePathStyle, "s3.force-path-style", false, "Set this to `true` to force the request to use path-style addressing.")
+	f.BoolVar(&cfg.S3ForceTLS, "s3.force-tls", false, "Set this to `true` to force s3 requests to be made over https")
 }
 
 type dynamoDBStorageClient struct {
--- a/pkg/chunk/aws/s3_storage_client.go
+++ b/pkg/chunk/aws/s3_storage_client.go
@@ -47,6 +47,12 @@ func NewS3ObjectClient(cfg StorageConfig, schemaCfg chunk.SchemaConfig) (chunk.O
 		return nil, err
 	}
 
+	if cfg.S3ForceTLS {
+		if strings.Contains(cfg.S3.URL.Host, ".") {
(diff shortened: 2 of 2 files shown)


pkg/chunk/aws/s3_storage_client.go
```diff
@@ -47,6 +47,12 @@ func NewS3ObjectClient(cfg StorageConfig, schemaCfg chunk.SchemaConfig) (chunk.O
 		return nil, err
 	}
 
+	if cfg.S3ForceTLS {
+		if strings.Contains(cfg.S3.URL.Host, ".") {
```
csmarchbanks: Should this check for presence of http/https in the url already?

blockloop:
> Does specifying `-s3.url=https://<whatever>` from the command line work?

it doesn't, because [common/aws](https://github.com/weaveworks/common/blob/ddeaa31513fde49351b563d2f37343906eae06d3/aws/config.go#L48) ignores the scheme and always attaches `http`

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -137,6 +138,7 @@ func (cfg *StorageConfig) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&cfg.S3, "s3.url", "S3 endpoint URL with escaped Key and Secret encoded. "+
 		"If only region is specified as a host, proper endpoint will be deduced. Use inmemory:///<bucket-name> to use a mock in-memory implementation.")
 	f.BoolVar(&cfg.S3ForcePathStyle, "s3.force-path-style", false, "Set this to `true` to force the request to use path-style addressing.")
+	f.BoolVar(&cfg.S3ForceTLS, "s3.force-tls", false, "Set this to `true` to force s3 requests to be made over https")
```
tomwilkie: Nit: `Set this to true to.." is a bit redundant, could just say "Force S3e requests to be made over HTTPS." 

If you change it, please change the one above to.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -47,6 +47,12 @@ func NewS3ObjectClient(cfg StorageConfig, schemaCfg chunk.SchemaConfig) (chunk.O
...
+			s3Config = s3Config.WithEndpoint(fmt.Sprintf("https://%s", cfg.S3.URL.Host))
+		}
+	}
```
tomwilkie: I'd say this should be part of `awscommon.ConfigFromURL` - WDYT?

tomwilkie: ie I'm suggesting you should check the URL provided and preserve the scheme.

pkg/chunk/aws/s3_storage_client.go
thorfour: I tried to keep it consistent with how `ConfigFromURL` sets the endpoint which is almost identical to that if statement, but with https instead

pkg/chunk/aws/dynamodb_storage_client.go
thorfour: Fair, was trying to be consisted with what was above :)

pkg/chunk/aws/s3_storage_client.go
thorfour: Hm well given the example url `s3://abc:123@s3.default.svc.cluster.local:4569` would it just default to http then? 

So if it's https it's respected but all other schemes are http? This feels a little more obvious to me, but I don't disagree that it probably should be apart of `ConfigFromURL`.

thorfour:
Closing this to make the change in the common package instead. Will open a new PR with a vendor change once that's merged.

PR #1587, opened 2019-08-19
Spread the sync load over time

gouthamve:
If we have several table-managers that start at the same time, they will all start syncing at once causing limits to be hit. This adds some randomness so that the load is spread evenly.

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"math/rand"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


pkg/chunk/table_manager.go
```diff
@@ -166,6 +167,10 @@ func (m *TableManager) Stop() {
 func (m *TableManager) loop() {
 	defer m.wait.Done()
 
+	// Sleep for a bit to spread the sync load across different times if the tablemanagers are all started at once.
+	rand.Seed(time.Now().UnixNano())
```
bboreham: I think this is unwise anywhere outside of a main program.  Sometimes you want repeatable random numbers, e.g. to reproduce a bug.

Also if `loop()` were to be called twice in the same program at the same time (within clock resolution) you'd get the opposite of what you intended.

gouthamve: I've moved this one level up. Let me know if that makes more sense?

bboreham: You're still calling `rand.Seed()` in a place that is not the `main()`, and I still think that is unwise.

PR #1593, opened 2019-08-20
Use remove-alertmanager-2.11 branch of Prometheus

codesome:
I updated `remove-alertmanager` branch of Prometheus for some other PR which cortex is using currently. This broke the vendoring.

I have created another branch `remove-alertmanager-2.11` with base version of Prometheus suffixed so that the same mistake is not repeated.

(Refer https://github.com/cortexproject/cortex/pull/1510#issuecomment-516895345 for the previous Prometheus vendoring)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,7 @@ require (
 	github.com/prometheus/alertmanager v0.15.1
 	github.com/prometheus/client_golang v1.0.0
 	github.com/prometheus/common v0.4.1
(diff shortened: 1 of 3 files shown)


gouthamve:
Why do we need 2 branches?

codesome:
Forgot to mention: now cortex would only use `remove-alertmanager-2.11`, and won't use `remove-alertmanager`. I will delete the old branches from Prometheus accordingly.

codesome:
Restored `remove-alertmanager` to the old commit instead. This PR is no longer needed.

PR #1595, opened 2019-08-21
fix crashing OverridesManager and updated the tests

sandlis:
OverridesManager was crashing while stopping because its quit channel was not initialized. Fixed it and updated some tests including calling Stop method of OverridesManager

### DIFF
--- a/pkg/util/validation/override_test.go
+++ b/pkg/util/validation/override_test.go
@@ -18,7 +18,7 @@ var defaultTestLimits *TestLimits
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
 func (l *TestLimits) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	if defaultLimits != nil {
+	if defaultTestLimits != nil {
 		*l = *defaultTestLimits
 	}
 	type plain TestLimits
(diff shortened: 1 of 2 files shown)


bboreham:
Can we have a test for `NewOverridesManager()` ?

sandlis:
Tests actually are using OverridesManager. You want me to do any changes to it o? https://github.com/cortexproject/cortex/blob/870fe73107bd2b7fe35976e3c310ae1a21ad76be/pkg/util/validation/override_test.go#L61

gouthamve:
The test now fails without the changes due to the new `overridesManager.Stop()` at the end @bboreham.

bboreham:
I mean: let's have a test that would have failed on the previous code.

sandlis:
I have added tests for both the scenarios that are fixed in this PR.
[This](https://github.com/cortexproject/cortex/pull/1595/files#diff-b8d9f32b50189a47769b187f2e71b0c8R72) validates the loop.
[This](https://github.com/cortexproject/cortex/pull/1595/files#diff-b8d9f32b50189a47769b187f2e71b0c8R78) validates the stopping of OverridesManager.

The loop test would wait indefinitely if it would have been there for previous code. I will do some changes to fail early instead of waiting until timeout or NewOverridesManager() to return.

pkg/util/validation/override_test.go
```diff
@@ -49,6 +50,54 @@ func testLoadOverrides(filename string) (map[string]interface{}, error) {
...
+		require.NoError(t, tempFile.Close())
+		require.NoError(t, os.Remove(tempFile.Name()))
+	}()
+
+	_, err = tempFile.WriteString(`overrides:
+  user1:
+    limit2: 150`)
+	require.NoError(t, err)
+
+	defaultTestLimits = &TestLimits{Limit1: 100}
```
bboreham: We should really have a loud warning on this global variable that you can't expect to run multiple different tests at the same time.  Or fix the design.

pkg/util/validation/override_test.go
```diff
@@ -49,6 +50,54 @@ func testLoadOverrides(filename string) (map[string]interface{}, error) {
...
+		OverridesLoadPath:     tempFile.Name(),
+		OverridesLoader:       testLoadOverrides,
+		Defaults:              defaultTestLimits,
+	}
+
+	var overridesManager *OverridesManager
+	done := make(chan struct{})
+
+	// timeout for getting a response from NewOverridesManager()
+	timeout := time.NewTicker(1 * time.Second)
```
bboreham: `time.After()` is simpler.

pkg/util/validation/override_test.go
sandlis: I do not see any way to get rid of Global variable from tests so added a Warning.

PR #1574, opened 2019-08-12
Vendor tsdb and reuse iterators

codesome:
This includes https://github.com/prometheus/tsdb/pull/642. I have added similar re-use of iterators here. There is a possibility of allocs improvements in ingester push and I am trying to realize that.

I will post the benchmark results soon.

PR #1597, opened 2019-08-21
Vendor prometheus 2.12.0 and alertmanager 0.19.0

codesome:
With this, we no longer vendor prometheus/tsdb repo as it was merged with prometheus.

prometheus/alertmanager is also upgraded to 0.18.0 as a part of this. I have managed to fix all the API changes except 1, still trying to figure out what has to be done for that.


codesome:
With the last commit I believe I fixed all alertmanager API changes. But now alertmanager 0.18.0 uses an older version of prometheus/common (0.4.1), and is not compatible with the latest (0.6.0) which the prometheus pulls. It is fixed in alertmanager master, but vendoring that is downgrading prometheus unusually.

codesome:
Opened this PR https://github.com/prometheus/alertmanager/pull/2008 to fix the prometheus downgrade.

pkg/alertmanager/alertmanager.go
```diff
@@ -110,28 +120,31 @@ func New(cfg *Config) (*Alertmanager, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create api: %v", err)
+	}
 
 	am.router = route.New()
 
 	webReload := make(chan chan error)
 	ui.Register(am.router.WithPrefix(am.cfg.ExternalURL.Path), webReload, log.With(am.logger, "component", "ui"))
-	am.api.Register(am.router.WithPrefix(path.Join(am.cfg.ExternalURL.Path, "/api/v1")))
+	am.api.Register(am.router, path.Join(am.cfg.ExternalURL.Path, "/api/v1"))
```
jtlisi: This `Register` will handle registering the correct api prefix, no need to pass this.

```suggestion
	am.api.Register(am.router, am.cfg.ExternalURL.Path)
```

pkg/alertmanager/alertmanager.go
```diff
@@ -233,3 +244,67 @@ func (am *Alertmanager) Stop() {
...
+			return nil, err
+		}
+		integrationsMap[rcv.Name] = integrations
+	}
+	return integrationsMap, err
+}
+
+// buildReceiverIntegrations builds a list of integration notifiers off of a
+// receiver config.
+// Taken from https://github.com/prometheus/alertmanager/blob/94d875f1227b29abece661db1a68c001122d1da5/cmd/alertmanager/main.go#L112-L159.
```
jtlisi: We should look into exporting this upstream function if we can.

codesome: Looking at old alertmanager releases, it was actually exported in some different package. It was un-exported and moved there in recent releases. I will attempt exporting from alertmanager.

codesome: Tried to export it in upstream by putting it in the right place (`alertmanager/notify`), but running into a lot of cyclic import conflicts (tried to remove most, but hit a dead-end to fix it without duplicating some code). If its fine we could leave it as it is here.

codesome:
This PR is ready for another review now. I have vendored prometheus 2.12.0 and alertmanager 0.19.0-rc.0 (will do 0.19.0 in this PR when it's out but I don't expect it to be different), and finally no conflicts with prometheus 2.12.0!

pkg/alertmanager/multitenant.go
```diff
@@ -327,17 +328,25 @@ func (am *MultitenantAlertmanager) transformConfig(userID string, amConfig *amco
 	if am.cfg.AutoSlackRoot != "" {
 		for _, r := range amConfig.Receivers {
 			for _, s := range r.SlackConfigs {
-				if s.APIURL == autoSlackURL {
-					s.APIURL = amconfig.Secret(am.cfg.AutoSlackRoot + "/" + userID + "/monitor")
+				if s.APIURL.String() == autoSlackURL {
```
bboreham: I suspect Weaveworks was the only user of this code and we've moved to just webhook, so we could remove it.

pkg/chunk/encoding/bigchunk.go
```diff
@@ -281,7 +281,7 @@ func (it *bigchunkIterator) Scan() bool {
 
 	for it.i < len(it.chunks)-1 {
 		it.i++
-		it.curr = it.chunks[it.i].Iterator()
+		it.curr = it.chunks[it.i].Iterator(nil)
```
bboreham:
```suggestion
		it.curr = it.chunks[it.i].Iterator(&it.curr)
```

pkg/chunk/encoding/bigchunk.go
```diff
@@ -257,9 +257,9 @@ func (it *bigchunkIterator) FindAtOrAfter(target model.Time) bool {
 	}
 
 	if it.curr == nil {
-		it.curr = it.chunks[it.i].Iterator()
+		it.curr = it.chunks[it.i].Iterator(nil)
 	} else if t, _ := it.curr.At(); int64(target) <= t {
-		it.curr = it.chunks[it.i].Iterator()
+		it.curr = it.chunks[it.i].Iterator(nil)
```
bboreham:
```suggestion
		it.curr = it.chunks[it.i].Iterator(&it.curr)
```

pkg/chunk/encoding/bigchunk.go
codesome: I had another PR open to take care of many iterator re-use https://github.com/cortexproject/cortex/pull/1574, hence I did not include any in this PR. Would you like to add it in this PR itself?

codesome: Will add this part in this PR itself

codesome:
Updated alertmanager to 0.19.0 (no changes from rc0) and addressed all reviews. Waiting to cut 0.2 I believe.

codesome:
@bboreham Done

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## master / unreleased
 
+* [ENHANCEMENT] Upgraded Prometheus to 2.12.0 and Alertmanager to 0.19.0 #1597
+  * [CHANGE] No longer supports auto Slack root, hence `--alertmanager.configs.auto-slack-root` flag was also dropped.
```
gouthamve: Change comes before Enhancement.

codesome: I had actually indented it as it was part of the same PR. I will make those as 2 entries instead.

codesome: Modified

bboreham:
Was it essential here to update jsoniter ?

codesome:
Did not really see into that and did not expect the bugs that it's causing. Sorry about that! I see @gouthamve already has some advancements on this https://github.com/prometheus/prometheus/pull/6030

PR #1602, opened 2019-08-22
added interfaces for store specific limits

sandlis:
`NewStore` method in `pkg/storage` uses Overrides type defined in Cortex itself.
Added an interface to let consumers of that method pass any type which implements required methods returning Store specific limits.
This is done as part of way to let projects using Cortex, define their own limits

### DIFF
--- a/pkg/chunk/composite_store.go
+++ b/pkg/chunk/composite_store.go
@@ -3,12 +3,17 @@ package chunk
 import (
 	"context"
 	"sort"
+	"time"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
-
-	"github.com/cortexproject/cortex/pkg/util/validation"
 )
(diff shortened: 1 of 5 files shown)


pkg/chunk/composite_store.go
```diff
@@ -3,13 +3,18 @@ package chunk
...
+	"time"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
-
-	"github.com/cortexproject/cortex/pkg/util/validation"
 )
 
+// CompositeStoreLimits helps get Limits specific to Queries for Stores in CompositeStore
+type CompositeStoreLimits interface {
```
bboreham: This seems misnamed, since it's used in `store`, `CompositeStore` and `seriesStore`.  Suggest renaming to `StoreLimits`.

PR #1610, opened 2019-08-23
bugfix: memcachedClient.updateLoop() does not exit infinite loop

dmitsh:
bugfix: memcachedClient.updateLoop() does not exit infinite loop
when the client is stopped.

### DIFF
--- a/pkg/chunk/cache/memcached_test.go
+++ b/pkg/chunk/cache/memcached_test.go
@@ -3,10 +3,8 @@ package cache_test
 import (
 	"context"
 	"errors"
-	"sync"
 	"sync/atomic"
 	"testing"
-	"time"
 
 	"github.com/bradfitz/gomemcache/memcache"
 	"github.com/cortexproject/cortex/pkg/chunk/cache"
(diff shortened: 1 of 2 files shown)


gouthamve:
Can we have a quick test for this if it's not too cumbersome?

dmitsh:
> Can we have a quick test for this if it's not too cumbersome?

I've added a unit test (cc @gouthamve)

pkg/chunk/cache/memcached_test.go
```diff
@@ -151,3 +153,45 @@ func testMemcacheFailing(t *testing.T, memcache *cache.Memcached) {
 		}
 	}
 }
+
+// TestUpdateLoop simulates memcachedClient.updateLoop and verifies its completeness
```
gouthamve: Hrm, this test simulates it but is not a good unit test. For example, if in the future, if we change the logic of the code in `updateLoop`, we will never know. I don't think having this test adds any value, and if testing the `updateLoop` itself is too much, lets just chuck the test.

Issue #1333, opened 2019-04-16
Support Redis beside Memcached


weeco:
As devops who has:

1. Much more experience in operating Redis instead of Memcached
2. Can use managed redis services like Google's memory store

I wish cortex would support Redis beside/instead of memcached too, so that I don't need to gain experience in operating Memcached clusters and/or can even use existing redis clusters for Cortex.

krystianity:
+1

tomwilkie:
I'd be happy to merge support for this; you just need to implement the Cache interface here: https://github.com/cortexproject/cortex/blob/master/pkg/chunk/cache/cache.go#L10

dmitsh:
@weeco @tomwilkie please take a look at #1612 and share your feedback. Thanks!

bboreham:
Closed by #1612

PR #1612, opened 2019-08-24
Add Redis support

dmitsh:
Adding support for Redis.
Reference: issue #1333


dmitsh:
This PR is not complete yet. I would need to add unit tests (at the very least).
I'm publishing it now as is mostly to get the feedback.

tomwilkie:
Hi @dmitsh!  This PR looks pretty good to me so far.  I wonder why you've choosen to separate out the `redis` type from the `redisClient` type?  I think they could be all one type.

dmitsh:
@tomwilkie The main rationale was to follow `memcached` model. But I can see your point. Let me address that shortly.

dmitsh:
@tomwilkie please take a look. I also added a unit test.

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,244 @@
...
+		Buckets: prometheus.ExponentialBuckets(0.000016, 4, 8),
+	}, []string{"method", "status_code", "name"})
+
+	errRedisQueryTimeout = errors.New("redis query timeout")
+)
+
+// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
+func (cfg *RedisConfig) RegisterFlagsWithPrefix(prefix, description string, f *flag.FlagSet) {
+	f.StringVar(&cfg.Endpoint, prefix+"redis.endpoint", "", description+"Redis service endpoint to use when caching chunks. If empty, no redis will be used.")
+	f.DurationVar(&cfg.Timeout, prefix+"redis.timeout", 100*time.Millisecond, description+"Maximum time to wait before giving up on redis requests.")
```
bboreham: Is there some way we could share these flags with memcached?
E.g. have a memcached name and a neutral name both pointing to the same variable so we can deprecate and eventually remove the cache-specific names?

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,244 @@
...
+	pool *redis.Pool
+}
+
+var (
+	redisRequestDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "cortex",
+		Name:      "redis_request_duration_seconds",
+		Help:      "Total time spent in seconds doing redis requests.",
+		// Redis requests are very quick: smallest bucket is 16us, biggest is 1s
+		Buckets: prometheus.ExponentialBuckets(0.000016, 4, 8),
```
bboreham: microseconds?

bboreham:
Oh, and you should add a line to CHANGELOG for the release.

pkg/chunk/cache/redis_cache.go
dmitsh: I agree that it doesn't make sense to have both `redis` and `memcached`. I should fix that.
For example, we can have `--cache.store.type=(memcached|redis)` parameter to select the cache type.
But could you elaborate about cache-agnostic parameters? Some parameters are cache-specific, for example `batch_size` in `memcached`. How do you envision the command line would look like?
Do you mean something like that:
```
--cache.store.type=memcached --cache.store.batch_size=0 --cache.store.parallelism=10

--cache.store.type=redis --cache.store.host=redis.default:6379
```

pkg/chunk/cache/redis_cache.go
dmitsh: Yes, same as for `memcached`, I'm using a histogram with 8 buckets: 16us, 64us, 256us, 1ms, 4ms, 16ms, 64ms, 256ms.
Indeed, 16us and 64us might be too short, but 256us looks reasonable to me.
I suggest I change it to `prometheus.ExponentialBuckets(0.00025, 4, 8)`. Would that make sense?

dmitsh:
@bboreham I've implemented the changes we discussed.

pkg/chunk/cache/cache.go
```diff
@@ -37,19 +47,52 @@ type Config struct {
...
+	f.IntVar(&cfg.Parallelism, prefix+"cache.store.parallelism", 100, description+"Maximum active requests to cache.")
+	f.StringVar(&cfg.Host, prefix+"cache.store.hostname", "", description+"Hostname for cache service to use when caching chunks.")
+	f.StringVar(&cfg.Service, prefix+"cached.store.service", "", description+"SRV service used to discover cache servers.")
+	f.IntVar(&cfg.MaxIdleConns, prefix+"cache.store.max-idle-conns", 16, description+"Maximum number of idle connections in pool.")
+	f.IntVar(&cfg.MaxActiveConns, prefix+"cache.store.max-active-conns", 0, description+"Maximum number of active connections in pool.")
+	f.DurationVar(&cfg.Timeout, prefix+"cache.store.timeout", 100*time.Millisecond, description+"Maximum time to wait before giving up on memcached requests.")
+	f.DurationVar(&cfg.UpdateInterval, prefix+"cache.store.update-interval", 1*time.Minute, description+"Period with which to poll DNS for cache servers (if applicable).")
+	f.BoolVar(&cfg.ConsistentHash, prefix+"cached.store.consistent-hash", false, description+"Use consistent hashing to distribute to cache servers (if applicable).")
```
gouthamve: Why are some `cached.` and others `cache.`?

pkg/chunk/cache/cache.go
```diff
@@ -37,19 +47,52 @@ type Config struct {
...
 	f.BoolVar(&cfg.EnableFifoCache, prefix+"cache.enable-fifocache", false, description+"Enable in-memory cache.")
 	f.DurationVar(&cfg.DefaultValidity, prefix+"default-validity", 0, description+"The default validity of entries for caches unless overridden.")
 
 	cfg.Prefix = prefix
 }
 
+// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
+func (cfg *StoreConfig) RegisterFlagsWithPrefix(prefix, description string, f *flag.FlagSet) {
+	f.StringVar(&cfg.Type, prefix+"cache.store.type", "", description+"Type of storage cache.")
+	f.DurationVar(&cfg.Expiration, prefix+"cache.store.expiration", 0, description+"How long keys stay in the cache.")
```
gouthamve: I'm not sure this is the best way to do things tbh. Some of the options here are for one store and then others are for others stores.

I'd rather have:
```
StoreConfig {
   Type string
   Redis RedisConfig
   Memcached MemcachedConfig
   NewCacheType  NewCacheTypeConfig 
}
```

That'll make things easier and we'll clearly know what to set for what.

Further the flags would be:
`cache.type`, `cache.memcache.*`, `cache.redis.*`.

gouthamve: Also makes things like `NewMemcachedClient()` take only `Memcached` parameters.

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+		}
+		_, err = conn.Do("EXEC")
+	}()
+
+	select {
+	case err := <-res:
+		return err
+	case <-ctx.Done():
+		return errRedisQueryTimeout
```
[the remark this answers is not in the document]
dmitsh: I'm calling `defer conn.Close()` right after getting the connection.

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,232 @@
...
+	go func() {
+		bufs, err := redis.ByteSlices(conn.Do("MGET", intf...))
+		res <- &mgetResult{bufs: bufs, err: err}
+	}()
+
+	select {
+	case dat := <-res:
+		return dat.bufs, dat.err
+	case <-ctx.Done():
+		return nil, errRedisQueryTimeout
```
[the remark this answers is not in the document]
dmitsh: I'm calling `defer conn.Close()` right after getting the connection.

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,232 @@
...
+			_, err = redis.String(pong, err)
+		}
+		res <- err
+	}()
+
+	select {
+	case err := <-res:
+		return err
+	case <-ctx.Done():
+		return errRedisQueryTimeout
```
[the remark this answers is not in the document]
dmitsh: I'm calling `defer conn.Close()` right after getting the connection.

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,232 @@
...
+		name:       name,
+		client:     client,
+		requestDuration: observableVecCollector{
+			v: redisRequestDuration.MustCurryWith(prometheus.Labels{
+				"name": name,
+			}),
+		},
+	}
+
+	if err := cache.ping(context.Background()); err != nil {
```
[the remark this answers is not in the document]
dmitsh: the timeout is applied inside ping()

pkg/chunk/cache/redis_cache.go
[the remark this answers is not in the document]
gouthamve: Oops, nvm :)

dmitsh:
@gouthamve per our discussion I have made requested changed.
I've also rebased the branch and updated dependencies.

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,251 @@
...
+func (c *RedisCache) mset(ctx context.Context, keys []string, bufs [][]byte, ttl int) error {
+	res := make(chan error, 1)
+
+	conn := c.client.Connection()
+	defer conn.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, c.timeout)
+	defer cancel()
+
+	go func() {
```
bboreham: I think this deserves a comment.
(guessing it might be because Redis library doesn't respect context cancellation?)

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// mset adds key-value pairs to the cache.
+func (c *RedisCache) mset(ctx context.Context, keys []string, bufs [][]byte, ttl int) error {
+	res := make(chan error, 1)
+
+	conn := c.client.Connection()
+	defer conn.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, c.timeout)
```
bboreham: Why isn't this using the Redis library timeout facility? 
https://github.com/gomodule/redigo/blob/bb5990dc771ab24e169aaea90ee3b9ad50cffce9/redis/redis.go#L82

pkg/chunk/cache/redis_cache.go
```diff
@@ -0,0 +1,251 @@
...
+		} else {
+			missed = append(missed, key)
+		}
+	}
+	return
+}
+
+// Store stores the key in the cache.
+func (c *RedisCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
+	err := instr.CollectedRequest(ctx, "store", c.requestDuration, redisStatusCode, func(ctx context.Context) error {
```
bboreham: This seems to add nothing over the `instrumentedCache` wrapper.

Issue #759, opened 2018-03-19
Tidy up 'ring' admin page


bboreham:
Distributors answer an http call to `/ring` with an html page showing current status of all ingesters and all tokens.  I find I quite often want the status and have never needed the tokens, which form 99.5% of the bulk of the page.

exerpt:
```
tokens: <
  token: 4287485801
  ingester: "ingester-5cd48d4477-p22fk"
>
tokens: <
  token: 4288738658
  ingester: "ingester-5cd48d4477-j7bhg"
>
```

Each token could go on one line; we could have a button on the page to request the full detail.

ankit1ank:
What details would this new button show? Would all the tokens still be printed on the same page as they are now?

bboreham:
The page would show just the ingesters table by default, then if you press the button you get to see the tokens like on the current page.

PR #1621, opened 2019-08-25
Use different colors for table rows on ring status page

ankit1ank:
Ring Status Page Table With Colors

This PR colors alternate table rows of the ingesters table on the ring status page.

Signed-off-by: Ankit Goel <ankit.goel@go-jek.com>


bboreham:
Thanks for the PR!  If you're in the mood could you consider #759 ?

PR #1624, opened 2019-08-28
Optimise ring.DoBatch()

bboreham:
`ring.DoBatch` contributes 20% of the memory allocation in my production distributors.

I started looking at this and realised the existing benchmark was not exercising the functionality - the `cas()` didn't update the ring so it was actually empty, and some test fixture code was also quite far up the profile.

Having fixed all that and made the benchmark call the higher-level function `DoBatch()`, the numbers before are:

```
BenchmarkBatch10x100-2     	   10000	    175156 ns/op	   56648 B/op	     334 allocs/op
BenchmarkBatch100x100-2    	    5000	    256146 ns/op	   87675 B/op	     720 allocs/op
BenchmarkBatch100x1000-2   	    1000	   1838020 ns/op	  578726 B/op	    3278 allocs/op
```

after:

```
BenchmarkBatch10x100-2     	   10000	    149385 ns/op	   12061 B/op	      29 allocs/op
BenchmarkBatch100x100-2    	   10000	    208285 ns/op	   30909 B/op	     238 allocs/op
BenchmarkBatch100x1000-2   	    1000	   1558331 ns/op	  112601 B/op	     217 allocs/op
```

The `ReadRing` function signatures have changed a bit, which may impact downstream projects using this code.

### DIFF
--- a/pkg/ring/replication_strategy.go
+++ b/pkg/ring/replication_strategy.go
@@ -10,6 +10,7 @@ import (
 // tolerate.
 // - Filters out dead ingesters so the one doesn't even try to write to them.
 // - Checks there is enough ingesters for an operation to succeed.
+// The ingesters argument may be overwritten.
 func (r *Ring) replicationStrategy(ingesters []IngesterDesc, op Operation) (
 	liveIngesters []IngesterDesc, maxFailure int, err error,
 ) {
@@ -26,14 +27,15 @@ func (r *Ring) replicationStrategy(ingesters []IngesterDesc, op Operation) (
 	// Skip those that have not heartbeated in a while. NB these are still
 	// included in the calculation of minSuccess, so if too many failed ingesters
 	// will cause the whole write to fail.
-	liveIngesters = make([]IngesterDesc, 0, len(ingesters))
-	for _, ingester := range ingesters {
-		if r.IsHealthy(&ingester, op) {
-			liveIngesters = append(liveIngesters, ingester)
+	for i := 0; i < len(ingesters); {
+		if r.IsHealthy(&ingesters[i], op) {
+			i++
 		} else {
+			ingesters = append(ingesters[:i], ingesters[i+1:]...)
 			maxFailure--
 		}
 	}
+	liveIngesters = ingesters
 
 	// This is just a shortcut - if there are not minSuccess available ingesters,
 	// after filtering out dead ones, don't even bother trying.
--- a/pkg/ring/replication_strategy.go
+++ b/pkg/ring/replication_strategy.go
@@ -12,8 +12,7 @@ import (
 // - Checks there is enough ingesters for an operation to succeed.
 // The ingesters argument may be overwritten.
 func (r *Ring) replicationStrategy(ingesters []IngesterDesc, op Operation) (
-	liveIngesters []IngesterDesc, maxFailure int, err error,
-) {
+	[]IngesterDesc, int, error) {
 	// We need a response from a quorum of ingesters, which is n/2 + 1.  In the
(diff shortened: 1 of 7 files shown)


pkg/ring/ring.go
```diff
@@ -25,16 +25,18 @@ const (
...
+	maxExpectedReplicationSet = 5 // used to pre-allocate some arrays
 )
 
 // ReadRing represents the read inferface to the ring.
 type ReadRing interface {
 	prometheus.Collector
 
-	Get(key uint32, op Operation) (ReplicationSet, error)
-	BatchGet(keys []uint32, op Operation) ([]ReplicationSet, error)
+	Get(key uint32, op Operation, descs []IngesterDesc) (ReplicationSet, error)
```
tomwilkie: Can you document the new param is there to preserve allocations, and that its contents are not important - if thats true of course...

pkg/ring/ring.go
```diff
@@ -181,37 +183,16 @@ func migrateRing(desc *Desc) []TokenDesc {
...
-	r.mtx.RLock()
-	defer r.mtx.RUnlock()
-	return r.getInternal(key, op)
-}
-
-// BatchGet returns ReplicationFactor (or more) ingesters which form the replicas
-// for the given keys. The order of the result matches the order of the input.
-func (r *Ring) BatchGet(keys []uint32, op Operation) ([]ReplicationSet, error) {
+// ingesters should be a 0-length slice which can be overwritten (or nil)
+func (r *Ring) Get(key uint32, op Operation, ingesters []IngesterDesc) (ReplicationSet, error) {
```
tomwilkie: nit: `ingesters` here, `descs` in interface.

pkg/ring/replication_strategy.go
```diff
@@ -26,14 +27,15 @@ func (r *Ring) replicationStrategy(ingesters []IngesterDesc, op Operation) (
...
-			liveIngesters = append(liveIngesters, ingester)
+	for i := 0; i < len(ingesters); {
+		if r.IsHealthy(&ingesters[i], op) {
+			i++
 		} else {
+			ingesters = append(ingesters[:i], ingesters[i+1:]...)
 			maxFailure--
 		}
 	}
+	liveIngesters = ingesters
```
tomwilkie: I would fully support the removal of the use of named return parameters; I find them hard to follow in the review...

pkg/ring/ring.go
```diff
@@ -25,16 +25,18 @@ const (
 
 	// ConsulKey is the key under which we store the ring in consul.
 	ConsulKey = "ring"
+
+	maxExpectedReplicationSet = 5 // used to pre-allocate some arrays
```
tomwilkie: Can we put this in the file where it is used?

Also, IIUC, this could be called "max replication factor plus one" or something similar?

Finally, sentences for comments pls.

tomwilkie:
Nice set of optimisations!  Sorry I've been so nit picky.

Once small concern: we're now taking a (read) lock in a loop.  Not super concerned about consistency (the ring might update mid processing a batch) but do you think this will impact CPU usage?

pkg/ring/ring.go
bboreham: The code returns more ingesters than the replication factor under circumstances I don't fully understand. I'm not at all certain that "plus one" is correct otherwise I would have made the constant 4.

bboreham:
> do you think this will impact CPU usage?

The benchmarks show that CPU overall comes down.  Profiling suggests that `sort.Search()` looking for each token is far more CPU-intensive than the lock.

I have addressed review comments; PTAL.

Issue #1629, opened 2019-08-29
query frontend doesn't stop cache on exit


dmitsh:
When the query frontend exits, it does not invoke `Stop` method of `Cache` interface.

bboreham:
What are the bad consequences of this?

dmitsh:
> What are the bad consequences of this?
In case of memcached, we will have `updateLoop` go-routine running indefinitely.
In case of Redis (#1612) we won't properly clean up and close connection pool.

bboreham:
You’re saying the updateLoop continues running past the process’ decision to exit?

dmitsh:
> You’re saying the updateLoop continues running past the process’ decision to exit?

No, when a process exits, obviously all running go-routines belonging to that process are terminated.
What I'm saying is that `Stop` method `Cache` interface is never called, despite being declared.
We might meed to gracefully stop the cache before exiting.

bboreham:
Ok, I like to be precise. 
So the issue is that there are no bad consequences but there might be, under some different circumstances in future?

dmitsh:
You are correct. There are no immediate consequences. There is a bug in `memcached_client.go`, which I submitted PR #1610 for, but since `Stop` is never called, this bug hasn't been exposed.

I've been working on adding Redis cache support (PR #1612), and for that case calling `Stop` would be desirable, as it would properly dispose connection pool.

bboreham:
>  it would properly dispose connection pool

What is the value of this when the program is exiting?

jtlisi:
If the `Stop` function in the cache interface is never called and there are currently no consequences, that sounds s like a good reason to remove the `Stop` function from the interface.

PR #1630, opened 2019-08-29
frontend to clean up cache on exit

dmitsh:
Fixes issue #1629

### DIFF
--- a/pkg/querier/queryrange/results_cache.go
+++ b/pkg/querier/queryrange/results_cache.go
@@ -41,10 +41,10 @@ type resultsCache struct {
 }
 
 // NewResultsCacheMiddleware creates results cache middleware from config.
-func NewResultsCacheMiddleware(logger log.Logger, cfg ResultsCacheConfig, limits Limits) (Middleware, error) {
+func NewResultsCacheMiddleware(logger log.Logger, cfg ResultsCacheConfig, limits Limits) (Middleware, cache.Cache, error) {
 	c, err := cache.New(cfg.CacheConfig)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	return MiddlewareFunc(func(next Handler) Handler {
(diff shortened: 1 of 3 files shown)


pkg/querier/queryrange/results_cache.go
```diff
@@ -41,10 +41,10 @@ type resultsCache struct {
 }
 
 // NewResultsCacheMiddleware creates results cache middleware from config.
-func NewResultsCacheMiddleware(logger log.Logger, cfg ResultsCacheConfig, limits Limits) (Middleware, error) {
+func NewResultsCacheMiddleware(logger log.Logger, cfg ResultsCacheConfig, limits Limits) (Middleware, cache.Cache, error) {
```
bboreham: I think I would add a `Stop()` method to the `Middleware` return rather than exposing the implementation.

dmitsh: > I think I would add a `Stop()` method to the `Middleware` return rather than exposing the implementation.

I agree, let me take another swing at it.

dmitsh: @bboreham, I looked at this once again. `Frontend` does not have direct access to `queryrange.Middleware` member. Instead, it has `roundTripper` interface member of type `http.RoundTripper`. `queryrange.NewRoundTripper` returns `queryrange.roundTripper` object that implements `http.RoundTripper` but hides `Middleware`s internally. So, adding a new method to `queryrange.Handler` would break parity between `http.RoundTripper` and `queryrange.roundTripper`, hence the design.
I would argue that Cache is a special component that requires separate treatment, different from `RoundTripper`.
I'm open to an idea of adding `Close' method to `queryrange.Middleware` but IMO this PR in its current form offers sufficient and isolated changes.

khaines: I had a read through this and agree with @dmitsh 's findings. It might be a good point to revaluate some of the logic  in order to refactor away the implementation, but that seems like a larger effort than just addressing the issue at hand.

dmitsh:
Closing in favor of #1879 (rebased)

PR #1634, opened 2019-09-01
feature: redis storage client

blockloop:
Redis storage client is similar to the S3 storage client. It stores using the `ExternalKey()` as the key and sets a TTL based on the flag `-redis.chunk-retention`. The TTL is set in Redis so that the key expires when the through date of the chunk is reaches the retention time.


blockloop:
Closing in favor of #1612

PR #1637, opened 2019-09-01
Make Ruler Push() call get timeseries from the pool

bboreham:
Because `Distributor.Push()` returns them to the pool (after #1526).

In a typical deployment this doesn't actually break, since nothing in `Ruler` gets things out of the pool and the pool is emptied on GC, but it's clearly wrong and also inefficient.

The main change in this PR is to make `ToWriteRequest()` get from the pool - besides the one call in Ruler this is only used in tests, where it doesn't matter that the memory is never returned to the pool.

Along the way I re-coded `ToWriteRequest()` to not go via `Metric` - we were creating a map on every ruler sample just to read the data back out again.

Oh, and a drive-by spelling correction.

### DIFF
--- a/pkg/ruler/compat.go
+++ b/pkg/ruler/compat.go
@@ -3,7 +3,6 @@ package ruler
 import (
 	"context"
 
-	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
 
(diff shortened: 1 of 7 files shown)


pkg/ruler/compat.go
```diff
@@ -51,12 +47,14 @@ func (a *appendableAppender) AddFast(l labels.Labels, ref uint64, t int64, v flo
 }
 
 func (a *appendableAppender) Commit() error {
-	_, err := a.pusher.Push(a.ctx, client.ToWriteRequest(a.samples, client.RULE))
+	_, err := a.pusher.Push(a.ctx, client.ToWriteRequest(a.labels, a.samples, client.RULE))
+	a.labels = nil
```
gouthamve: Curious why we're not doing `a.labels = a.labels[:0]`?

bboreham: This new line simply parallels the old `a.samples = nil` below.
I believe making them both do `[:0]`, as you suggest, is valid, and likely to be an improvement.

PR #1638, opened 2019-09-01
 Return request data to the pool in Ingester.Push()

bboreham:
After #1526 the unmarshall code is trying to get memory out of the pool, but in ingesters it is never returned so the `pool.Get()` call just slows things down.

After making this change it still wasn't that great, because the pool was boxing `TimeSeries` objects.  So change those to `*TimeSeries`.

before:
```
BenchmarkIngesterPush-2   	     100	  23525939 ns/op	 3881766 B/op	   42317 allocs/op
```

after:
```
BenchmarkIngesterPush-2   	     100	  22937648 ns/op	 3273382 B/op	   32482 allocs/op
```

(note out of the 3-4MB per iteration, 1MB is overhead from the `FingerprintLocker`)

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -348,7 +348,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 
 		keys = append(keys, key)
 		validatedTimeseries = append(validatedTimeseries, client.PreallocTimeseries{
-			TimeSeries: client.TimeSeries{
+			TimeSeries: &client.TimeSeries{
 				Labels:  ts.Labels,
 				Samples: samples,
 			},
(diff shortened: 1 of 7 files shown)


bboreham:
I'm not so sure now what those benchmarks were run against.  Here's a new set with more permutations, on a different machine (go version go1.12 linux/amd64):

```
Current master 7447f75e9a97072e0b7b07d51174e55974d1baa4:
BenchmarkIngesterPush 	     100	  19752857 ns/op	 3280511 B/op	   31703 allocs/op
Before this PR db09d07e546e9c558154cfcf7d6766e06b0466cf: 
BenchmarkIngesterPush 	     100	  21412348 ns/op	 3291262 B/op	   31719 allocs/op
After changing benchmark f35f374c84991f7501da80ff550209f8f5118a9a:
BenchmarkIngesterPush 	      50	  33123151 ns/op	12155937 B/op	   62149 allocs/op
This PR fab5ced5e3c071a930966781b6eb786d8d34cc83:
BenchmarkIngesterPush 	     100	  26417682 ns/op	 3178133 B/op	   32073 allocs/op
Rebased against master d886f6e85a0356e4abbfd76cb685fddb69a8b626:
BenchmarkIngesterPush 	     100	  24700828 ns/op	 3174894 B/op	   32078 allocs/op
```

I'll merge since it's approved and check we get the expected drop in allocations in my staging environment.

PR #1648, opened 2019-09-03
cache table info for bigtable tables in tablemanager

jtlisi:

### DIFF
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"strings"
+	"time"
 
 	"cloud.google.com/go/bigtable"
 	ot "github.com/opentracing/opentracing-go"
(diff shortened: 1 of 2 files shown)


bboreham:
I don't quite get the point - you can already control how fast table-manager refreshes, so why would a cache with expiry be better?

bboreham:
I note the DynamoDB code has a rate-limiter - do you think that would be any help on GCP?

jtlisi:
> We should not cache them forever but rather have an expiry there.

@gouthamve I planned on adding cache expiration, just haven't gotten to it yet.

@bboreham Rate limiting would also solve this problem, but I think caching may be a better solution for bigtable. Bigtable has a GCP project-wide rate limit for table operations, so the problem we are facing is there are too many requests to bigtable from all of our clusters and we end up saturating the rate limit. However, since bigtable tables don't need any maintenance once created, it's simple enough to statically cache the table info.

bboreham:
Maybe set a time before which you don't fetch the detail on tables?
`--table-manager.retention-period` nearly achieves this.
I regularly move the `from` date forward on my table-manager to cut the work it is doing.

jtlisi:
@bboreham I thought a bit more about this and took a look at what you do in Dynamo. I think rate limiting is the solution here. It will lead to a similar approach so it will be easier to abstract this logic up to the table manager code one day. Also I personally think caching leads to potentially strange errors/edge cases.

PR #1649, opened 2019-09-04
Shrink bigchunk data structure

bboreham:
Remove pointer indirection on the underlying `XORChunk`, which will save memory and a few CPU cycles.
We have to be careful to create the Appender after the chunk is appended to our slice, since it has a pointer to the chunk memory.

And stop storing the end time of every small chunk - we can look at the start time of the next one since they don't overlap. This makes the data structure slightly smaller, and speeds up unmarshalling since we don't need to seek through every value.

We don't _need_ to store the start time either, since it can be fetched from the underlying chunk, but that takes longer.

Estimate of the impact:  Suppose chunks are avg 6 hours long then each one has 12 smallchunks and we save 16 bytes per = 192MB per million series, plus Go heap overheads.

### DIFF
--- a/pkg/chunk/encoding/bigchunk.go
+++ b/pkg/chunk/encoding/bigchunk.go
@@ -17,13 +17,13 @@ var errOutOfBounds = errors.New("out of bounds")
 type smallChunk struct {
 	chunkenc.XORChunk
 	start int64
-	end   int64
 }
 
 // bigchunk is a set of prometheus/tsdb chunks.  It grows over time and has no
 // upperbound on number of samples it can contain.
 type bigchunk struct {
 	chunks []smallChunk
+	end    int64
 
 	appender         chunkenc.Appender
 	remainingSamples int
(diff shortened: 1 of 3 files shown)


pkg/chunk/encoding/bigchunk.go
```diff
@@ -131,16 +128,15 @@ func (b *bigchunk) UnmarshalFromBuf(buf []byte) error {
 			return err
 		}
 
-		var start, end int64
-		start, end, reuseIter, err = firstAndLastTimes(chunk, reuseIter)
+		var start int64
+		start, reuseIter, err = firstTime(chunk, reuseIter)
```
csmarchbanks: Will this function now need to populate `b.end` somewhere?

bboreham: That seems a good shout. 
And a missing test.

bboreham: It turns out `end` is completely unnecessary - it is only used to shortcut the case where `FindAtOrAfter()` is called with a target past the end of the chunk, and this never happens because we have from/through times on each chunk at a higher level.

I extended the unit tests to fail on the previous code.

csmarchbanks: :+1: Thanks for looking into that!

PR #1651, opened 2019-09-04
Cache Bigtable Table Info

jtlisi:
Continuation of #1648 which I closed because I didn't fully understand the issue I was trying to solve.

### Overview
* Bigtable has a global rate limit for admin operations across a project.
  * Each operation counts towards this limit which can very quickly choke the endpoint if multiple . table managers request info within a few minutes.
* To ensure a schema is properly constructed in bigtable, a call to bigtable is required per table.
* We utilize a static consistent schema config across all of our services and creating a one off schema for the table manager is not desirable
* Once created a cortex table in bigtable does not need to be updated unless it is changed externally

Based on these parameters, caching is the best solution because it dramatically reduces the number of admin operations to bigtable. Whereas, rate limiting will only spread them out. Currently there is rate-limiting/backoff in the bigtable grpc client library, which has not been able to alleviate the problem.

### DIFF
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -38,12 +38,15 @@ type Config struct {
 
 	ColumnKey      bool
 	DistributeKeys bool
+
+	CacheTableInfo bool
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Project, "bigtable.project", "", "Bigtable project ID.")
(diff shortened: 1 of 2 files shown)


slim-bean:
I would agree with @jtlisi in regards to caching, rate limiting helps but it will ultimately reach a wall if you try to run many Cortex/Loki instances within one Google project.  The current code for enumerating tables will list ALL the tables even if they aren't for the current Cortex/Loki instance.  With 5 Cortex/Loki instances running in the same project each is calling GetTable on 300+ tables and quickly causes us to bump up against the 1000 requests every 100s rate limit.

Because each schema iteration can have a different prefix for both the index table and chunk table, trying to make the code calling GetTable prefix aware would quickly become very complicated, caching seems like a simpler solution.

jtlisi:
@bboreham Thanks for the suggestions, it should be good for another look

PR #1653, opened 2019-09-05
Make sure generated tokens are unique

pstibrany:
GenerateTokens could generate duplicate tokens previously, which kind of defeats the purpose I guess. This patch fixes it.

### DIFF
--- a/pkg/ring/util.go
+++ b/pkg/ring/util.go
@@ -10,18 +10,18 @@ import (
 func GenerateTokens(numTokens int, takenTokens []uint32) []uint32 {
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 
-	added := make(map[uint32]bool)
+	used := make(map[uint32]bool)
 	for _, v := range takenTokens {
-		added[v] = true
+		used[v] = true
 	}
 
 	tokens := []uint32{}
 	for i := 0; i < numTokens; {
 		candidate := r.Uint32()
-		if added[candidate] {
+		if used[candidate] {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


bboreham:
Thanks for the PR.  Good spot!
Did you consider adding all of `takenTokens` into the map first, so you don't have two kinds of check?
I think that would make the code simpler.  
Not a big deal, just something I thought of.

pstibrany:
> Thanks for the PR. Good spot!
> Did you consider adding all of `takenTokens` into the map first, so you don't have two kinds of check?
> I think that would make the code simpler.
> Not a big deal, just something I thought of.

You're absolutely correct :) I'll change it.

pkg/ring/util.go
```diff
@@ -2,23 +2,26 @@ package ring
...
 )
 
-// GenerateTokens make numTokens random tokens, none of which clash
-// with takenTokens.  Assumes takenTokens is sorted.
+// GenerateTokens make numTokens unique random tokens, none of which clash
+// with takenTokens.
 func GenerateTokens(numTokens int, takenTokens []uint32) []uint32 {
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	added := make(map[uint32]bool)
```
csmarchbanks: Nit: I would a prefer a name like `taken` or `existing`; to me `added` means that we have added them in this round.

pstibrany: What about 'used' then? It's shortest I can come up with.

bboreham: `used` sgtm

PR #1657, opened 2019-09-09
Turn off golang proxy

codesome:
Proxy is making the fetching of modules fail https://circleci.com/gh/cortexproject/cortex/9872, also confirmed in local machine


bboreham:
#1660 should be a better fix.

codesome:
Thanks

PR #1666, opened 2019-09-12
Drain memcached connections

tomwilkie:
If a connection times out, instead of closing it run a goroutine to drain anything thats left on the connection and then close it.

We occasionally see lots (10s of thousands) of sockets on the memcached server in CLOSE_WAIT, all with data in their send buffers. This looks like a bug in memcache, and this is a work around on the client side.

See https://github.com/bradfitz/gomemcache/pull/108 for upstream changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -84,3 +84,5 @@ require (
 	gopkg.in/fsnotify/fsnotify.v1 v1.4.7 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 )
+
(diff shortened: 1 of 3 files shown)


bboreham:
Could you file an issue in case this PR gets abandoned?

gouthamve:
I think there are now issues and this changes things in vendor itself, and iirc, this didn't help too much. Closing

PR #1674, opened 2019-09-17
add tags in manifests

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

The manifests in /k8s don't have tag in the cortex image, and in quay.io/ortexproject/project there is no tag `latest`. Here add the tag v0.2.0 so users can deploy directly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,21 +51,6 @@ To run the test suite:
 make test
 ```
 
-## Playing in `minikube`
-
-First, start `minikube`.
-
-Then run Cortex in minikube:
-```
-kubectl apply -f ./k8s
-```
-
-Cortex will sit behind an nginx instance exposed on port 30080.  A job is deployed to scrape itself.  Try it:
-
-http://192.168.99.100:30080/api/prom/api/v1/query?query=up
-
-If that doesn't work, your Minikube might be using a different ip address. Check with `minikube status`.
-
 ### Dependency management
 
 We uses [Go modules](https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more) to manage dependencies on external packages.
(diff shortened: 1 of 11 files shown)


bboreham:
Thanks for the PR!

For background, the existing example manifests were written to aid someone building from source, who would have `latest` tags, and there were instructions in the README to use them that way.  Right now I see no instructions mentioning the manifests, so maybe we should have some?

Also there is a `prime-minikube` target in the Makefile that is not consistent with changing them to a specific version tag.

Finally the release instructions should include updating the version.

yeya24:
Oh, sorry. I missed this Makefile. But I think uses will always try to look at the yamls first instead of Makefile

bboreham:
I described some inconsistencies between this PR and other files in the repo.
If you want to fix up those inconsistencies we can proceed.

anoopl:
I think the version should be v0.3.0

bboreham:
The version should be the latest released version, which is why I suggested to add “update the version in the example yamls” to release instructions.

yeya24:
@bboreham Updated.  PTAL

bboreham:
You don’t seem to have addressed any of my comments. 

Maybe I missed an explanation?

pracucci:
@yeya24 Are you still interested in working on this PR? If yes, may you rebase `master`, reiterate on Bryan feedback (see the discussion) and also update the version to `v0.6.0` (released today), please?

yeya24:
Sure, I will rebase it soon. Could you please give me some hints about how to modify the Makefile?

pracucci:
> Could you please give me some hints about how to modify the Makefile?

What would you like to do specifically? Address [Bryan's feedback](https://github.com/cortexproject/cortex/pull/1674#issuecomment-532584306) and add a target to update the version in all K8S spec files?

yeya24:
Sorry for this delay. If using specific image tags in manifests, do we still need the targets `save-images`, `load-images`, `prime-minikube` in Makefile? Can I delete them all?

Issue #1628, opened 2019-08-29
distributor dies when enabling ha-tracker


friedrich-at-adobe:
**Steps**
1. Start consul as daemon
2. Start one ingester and kill it
3. Start distributor with ha-tracker enabled

**Results**
- Distributors crashes

**Expected**
- No crashes from distributor

```
$ docker run -d -p 8500:8500 consul:1.6.0 agent  -ui -server -client=0.0.0.0 -bootstrap
$ go run cmd/cortex/main.go -target=ingester -dynamodb.url=inmemory:///cortex -chunk.storage-client=inmemory
level=info ts=2019-08-29T10:04:42.186978Z caller=cortex.go:198 msg=initialising module=overrides
level=info ts=2019-08-29T10:04:42.187492Z caller=override.go:53 msg="per-tenant overrides disabled"
level=info ts=2019-08-29T10:04:42.187525Z caller=cortex.go:198 msg=initialising module=store
level=info ts=2019-08-29T10:04:42.187601Z caller=cortex.go:198 msg=initialising module=server
level=info ts=2019-08-29T10:04:42.18787Z caller=server.go:117 http=[::]:80 grpc=[::]:9095 msg="server listening on addresses"
level=info ts=2019-08-29T10:04:42.188101Z caller=cortex.go:198 msg=initialising module=ingester
level=warn ts=2019-08-29T10:04:42.188262Z caller=net.go:15 msg="error getting interface" inf=eth0 err="route ip+net: no such network interface"
level=info ts=2019-08-29T10:04:42.188601Z caller=main.go:68 msg="Starting Cortex" version="(version=, branch=, revision=)"
level=info ts=2019-08-29T10:04:42.194753Z caller=lifecycler.go:387 msg="entry not found in ring, adding with no tokens"
level=info ts=2019-08-29T10:04:42.202317Z caller=lifecycler.go:317 msg="auto-joining cluster after timeout"
^Clevel=info ts=2019-08-29T10:05:25.553445Z caller=signals.go:54 msg="=== received SIGINT/SIGTERM ===\n*** exiting"
level=info ts=2019-08-29T10:05:25.553543Z caller=cortex.go:224 msg=stopping module=ingester
level=info ts=2019-08-29T10:05:25.553572Z caller=lifecycler.go:470 msg="changing ingester state from" old_state=ACTIVE new_state=LEAVING
level=info ts=2019-08-29T10:05:25.568159Z caller=transfer.go:212 msg="nothing to transfer"
[...]

PR #1677, opened 2019-09-18
Make sure the HA tracker only handles valid keys

### LINKED ISSUE #1652: enable ha-tracker,ERR
my configure
```
        - -consul.hostname=consul:8500
        - -distributor.replication-factor=3
        - -distributor.ha-tracker.enable=true
        - -distributor.ha-tracker.enable-for-all-users=true
        - -distributor.ha-tracker.consul.hostname=consul:8500
        - -distributor.ha-tracker.prefix=ha/
        - -distributor.ha-tracker.store=consul
        - -distributor.ha-tracker.cluster="prometheus"
        - -distributor.ha-tracker.replica="prometheus_replica"
```
distributor log err:
caller=client.go:226 msg="error getting path" prefix=ha/ err=null

use cortex:v0.2.0-rc.0

gouthamve:
If the ring and the HA tracker share the same prefix then we will end up
handling invalid keys.

Fixes #1652 and #1628

### DIFF
--- a/pkg/distributor/ha_tracker.go
+++ b/pkg/distributor/ha_tracker.go
@@ -150,6 +150,13 @@ func (c *haTracker) loop(ctx context.Context) {
 		c.electedLock.Lock()
 		defer c.electedLock.Unlock()
 		chunks := strings.SplitN(key, "/", 2)
+
+		// The prefix has already been stripped, so a valid key would look like cluster/replica,
+		// and a key without a / such as `ring` would be invalid.
+		if len(chunks) != 2 {
+			return true
+		}
+
 		if replica.Replica != c.elected[key].Replica {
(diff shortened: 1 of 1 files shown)


pkg/distributor/ha_tracker.go
```diff
@@ -150,6 +150,12 @@ func (c *haTracker) loop(ctx context.Context) {
 		c.electedLock.Lock()
 		defer c.electedLock.Unlock()
 		chunks := strings.SplitN(key, "/", 2)
+
+		// Only handle valid keys.
```
cstyan: Can we leave a comment here or somewhere nearby like `The prefix has already been stripped, so a valid key would look like cluster/replica, and a key without a / such as ring would be invalid.`

PR #1678, opened 2019-09-18
Always push build-image

gouthamve:
Also, remove dep as we don't need it. I've verified that cortex still
builds after removing dep.


gouthamve:
This won't work as we're inside the build-image and won't see the image itself.

PR #1680, opened 2019-09-18
consul: Print warning not error when path is empty

### LINKED ISSUE #1652: enable ha-tracker,ERR

gouthamve:

Fixes #1652

### DIFF
--- a/pkg/ring/kv/consul/client.go
+++ b/pkg/ring/kv/consul/client.go
@@ -223,7 +223,12 @@ func (c *Client) WatchPrefix(ctx context.Context, prefix string, f func(string, 
 
 		kvps, meta, err := c.kv.List(prefix, queryOptions.WithContext(ctx))
 		if err != nil || kvps == nil {
-			level.Error(util.Logger).Log("msg", "error getting path", "prefix", prefix, "err", err)
+			if kvps == nil {
+				level.Warn(util.Logger).Log("msg", "no items in path", "prefix", prefix)
+			} else {
+				level.Error(util.Logger).Log("msg", "error getting path", "prefix", prefix, "err", err)
+			}
+
 			backoff.Wait()
 			continue
 		}


pkg/ring/kv/consul/client.go
```diff
@@ -223,7 +223,12 @@ func (c *Client) WatchPrefix(ctx context.Context, prefix string, f func(string,
 
 		kvps, meta, err := c.kv.List(prefix, queryOptions.WithContext(ctx))
 		if err != nil || kvps == nil {
-			level.Error(util.Logger).Log("msg", "error getting path", "prefix", prefix, "err", err)
+			if kvps == nil {
+				level.Warn(util.Logger).Log("msg", "no items in path", "prefix", prefix)
```
bboreham: Can we make this warning actionable?  What would an administrator do, upon seeing this?

jtlisi: I don't think this needs to be actionable since an empty ring does not necessarily mean anything is wrong. Maybe this should be info level instead of a warning.

khaines: If there's no action to be taken, this very much sounds like an info message to me

gouthamve:
Closing this in favour of https://github.com/cortexproject/cortex/pull/1756

Issue #1647, opened 2019-09-03
Static File Backend for Alertmanager and Ruler


jtlisi:
Ideally both the alertmanager& ruler could utilize static files for user configurations. This would allow new cortex users to start using the project without coupling postgres as a dependency. All that would be required is a filesystem where configs can be stored. All that would need to be provided would be a path to a directory where user configs would exist

Tenants can be separated with a unique naming scheme. Either each tenants configs exists in a directory named after their user id or a prefix naming scheme could be devised.

csmarchbanks:
To be similar with the other storage options for Cortex, would it make more sense to be able to store user configurations in s3/GCS? Perhaps that would be in addition to static files?

jtlisi:
Yes in addition. We have a branch currently running using GCS.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

bboreham:
To clarify, you would have no API in front of this, just leave it up to the admin to modify the files?

jtlisi:
Yes, read only

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

pracucci:
@jtlisi Has #2125 fixed this, or this issue should be kept open?

jtlisi:
#2125 fixed this

Issue #1244, opened 2019-02-24
Future of the Config Service


jtlisi:
The **Config Service** is a central component to the Alertmanager and Ruler. It receives JSON data with an embedded yaml blobs that represent prometheus config files. Preliminary work to decouple the Ruler from the config service has already started. Instead of polling the config service the ruler will host it's own api and interact with the Postgres database directly. If the alertmanager moves in this direction that leaves little reason to keep the config service. Below I added some goals to keep in mind when considering the Config service. I would be interested to know what people think and what the general sentiment is around deprecating the config db in favor of embedding the libs within the alertmanager and ruler.

**Goals for Configuration**

1. Maintain the ability to upload prometheus config files

In my opinion the ability to upload and export prometheus config files is essential to the cortex user experience. Users should have a portable configuration that they can move from prometheus to cortex and vice versa. However, this does not mean we need to store each user information in a json blob. Structuring user configs into a more refined format could be necessary for user friendly CRUD features related to alerts/rules in the future.

2. Validate User Configs in API

Validation features that ensure users have properly formatted configuration files would increase the user experience with the ruler. It would be beneficial for users to know they have an inoperable group of rules as a response to an api call rather than an internal error when that rule group is polled by the ruler.

csmarchbanks:
This seems very related to: https://github.com/cortexproject/cortex/issues/619

Personally, I would be in favor of not requiring the config service since I don't believe it provides much (any?) benefit over having the relevant api calls directly on the ruler/alertmanager and allow those components to be fully in charge of their data.

bboreham:
Re point 2, that is supposed to happen [here](https://github.com/cortexproject/cortex/blob/6fc5b0ae2421f133f3cc5a9db9de65c23532a58d/pkg/configs/api/api.go#L117-L126).  Interested to know what error you are referring to.

jtlisi:
@bboreham The way the ruler is currently set up it does not have the same validation when it uses it's own API rather than the config service:
https://github.com/cortexproject/cortex/blob/005412977f765d25610e1ec04e9726072a2a4989/pkg/ruler/api.go#L87

This leads to the requirement of ensuring a parseable rule set later: 

https://github.com/cortexproject/cortex/blob/005412977f765d25610e1ec04e9726072a2a4989/pkg/ruler/scheduler.go#L233

[...]

PR #1686, opened 2019-09-20
Add Alertmanager Integration Tests and Static File Backend

jtlisi:
## Overview
* Abstract the ConfigDB client behind the alert store interface
* Add Integration tests for the Alertmanager that ensures the proper functionality of the v1 api
* Add a static file config backend for the alertmanager

## Related
#1647
#1244 
#1513 
#619


bboreham:
Where does the new protobuf come in?

jtlisi:
I chose to use a protobuf for the polling interface boundary to provide a easy to use storage format that wasn't yaml based and easy to version. This would make any additional backend based on cloud provider object stores much easier to add, while not reducing any of the current functionality.

pkg/alertmanager/alerts/store.go
```diff
@@ -0,0 +1,66 @@
...
+
+		var templates []*TemplateDesc
+		for fn, template := range cfg.Config.TemplateFiles {
+			templates = append(templates, &TemplateDesc{
+				Filename: fn,
+				Body:     template,
+			})
+		}
+
+		userRules := AlertConfigDesc{
```
pracucci: Can be simplified as `c.alertConfigs[user] = AlertConfigDesc{`

pkg/alertmanager/multitenant.go
```diff
@@ -198,13 +193,18 @@ func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, cfgCfg confi
 		go peer.Settle(context.Background(), cluster.DefaultGossipInterval)
```
pracucci: I can't comment on non changed lines, but I've noticed above we do `os.Exit(1)`. I don't think it's correct. We should return the the error.

pkg/alertmanager/multitenant.go
```diff
@@ -270,33 +270,24 @@ func (am *MultitenantAlertmanager) updateConfigs(now time.Time) error {
 }
 
 // poll the configuration server. Not re-entrant.
-func (am *MultitenantAlertmanager) poll() (map[string]configs.View, error) {
-	configID := am.latestConfig
-	cfgs, err := am.configsAPI.GetAlerts(context.Background(), configID)
+func (am *MultitenantAlertmanager) poll() (map[string]alerts.AlertConfigDesc, error) {
+	cfgs, err := am.store.ListAlertConfigs(context.Background())
 	if err != nil {
 		level.Warn(util.Logger).Log("msg", "MultitenantAlertmanager: configs server poll failed", "err", err)
```
pracucci: We should rename `server` with `store` because it's not guaranteed to be a DB anymore. The word `server` is also mentioned in a couple of comments in this file.

pkg/alertmanager/multitenant.go
```diff
@@ -270,33 +270,24 @@ func (am *MultitenantAlertmanager) updateConfigs(now time.Time) error {
...
 	}
-	am.latestMutex.Lock()
-	am.latestConfig = cfgs.GetLatestConfigID()
-	am.latestMutex.Unlock()
-	return cfgs.Configs, nil
+	return cfgs, nil
 }
 
-func (am *MultitenantAlertmanager) addNewConfigs(cfgs map[string]configs.View) {
+func (am *MultitenantAlertmanager) addNewConfigs(cfgs map[string]alerts.AlertConfigDesc) {
```
pracucci: The function name doesn't look correct to me. It doesn't just "add", but also "update" existing configs. Maybe `syncConfigs()` may be more appropriate.

pkg/alertmanager/multitenant.go
```diff
@@ -270,33 +270,24 @@ func (am *MultitenantAlertmanager) updateConfigs(now time.Time) error {
...
 	// TODO: instrument how many configs we have, both valid & invalid.
 	level.Debug(util.Logger).Log("msg", "adding configurations", "num_configs", len(cfgs))
-	for userID, config := range cfgs {
-		if config.IsDeleted() {
```
pracucci: Before this PR, we would have removed a config no more existing. In this PR, we filter out removed users in the store, which means the `cfgs` input parameter here does not contain removed users, but then we never reconcile `am.alertmanagers` and `am.cfgs` to remove users no more in the `cfgs`.

Moreover the `deleteUser()` is not called by anyone now.

pkg/alertmanager/multitenant.go
```diff
@@ -270,33 +270,24 @@ func (am *MultitenantAlertmanager) updateConfigs(now time.Time) error {
 }
 
 // poll the configuration server. Not re-entrant.
-func (am *MultitenantAlertmanager) poll() (map[string]configs.View, error) {
-	configID := am.latestConfig
-	cfgs, err := am.configsAPI.GetAlerts(context.Background(), configID)
+func (am *MultitenantAlertmanager) poll() (map[string]alerts.AlertConfigDesc, error) {
```
pracucci: Given `alerts.AlertStore` is an interface easily mockable, would make much sense add unit tests to cover at least a basic scenario:

- `loadAllConfigs()` with 2 users (A and B)
- Assert the state is as expected
- Add user C and remove user B
- `poll()`
- Assert the state is as expected (C added, B removed)

pkg/alertmanager/alerts/compat.go
```diff
@@ -0,0 +1,26 @@
+package alerts
+
+// ToProto con
+func ToProto(cfg string, templates map[string]string, user string) (*AlertConfigDesc, error) {
```
pracucci: Who uses it?

pkg/alertmanager/alerts/compat.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+	return &AlertConfigDesc{
+		User:      user,
+		RawConfig: cfg,
+		Templates: tmpls,
+	}, nil
+}
+
+// ParseTemplates returns a alertmanager config object
+func ParseTemplates(cfg *AlertConfigDesc) map[string]string {
```
pracucci: Who uses it?

pkg/alertmanager/alerts/alerts.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
```
pracucci: >> Where does the new protobuf come in?
> I chose to use a protobuf for the polling interface boundary to provide a easy to use storage format that wasn't yaml based and easy to version. This would make any additional backend based on cloud provider object stores much easier to add, while not reducing any of the current functionality.

Could you re-explain to me (which I'm slow to understand :)) why we do need protobuf given the current use case? Let's say, without considering what the future could hold for us...

pkg/alertmanager/multitenant.go
jtlisi: I agree, not something I noticed in the past but I can take care of this in a separate PR

pkg/alertmanager/storage.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/cortexproject/cortex/pkg/alertmanager/alerts"
+	"github.com/cortexproject/cortex/pkg/alertmanager/local"
+	"github.com/cortexproject/cortex/pkg/configs/client"
+)
+
+// AlertStoreConfig configures the alertmanager backend
+type AlertStoreConfig struct {
+	Type     string `yaml:"type"`
+	ConfigDB client.Config
```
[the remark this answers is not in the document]
jtlisi: I think local is a better fit. It's what we use to designate the boltdb backend for chunks

pkg/alertmanager/storage.go
```diff
@@ -0,0 +1,40 @@
...
+	File     local.FileAlertStoreConfig `yaml:"file_config"`
+}
+
+// RegisterFlags registers flags.
+func (cfg *AlertStoreConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.File.RegisterFlags(f)
+	cfg.ConfigDB.RegisterFlagsWithPrefix("alertmanager.", f)
+	f.StringVar(&cfg.Type, "alertmanager.storage.type", "configdb", "Method to use for backend rule storage (configdb, file)")
```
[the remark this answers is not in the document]
jtlisi: I think everything related to backend config storage should have a `alertmanager.storage.`. However, I wanted to avoid making a breaking change in this PR and updating the prefix for `alertmanager.configs` to `alertmanager.storage.configs` would force users to change their flags. I can add a TODO to make that change in the future and go throough a change/deprecation process in a future PR?

pkg/alertmanager/storage.go
```diff
@@ -0,0 +1,40 @@
...
+	case "configdb":
+		c, err := client.New(cfg.ConfigDB)
+		if err != nil {
+			return nil, err
+		}
+		return alerts.NewConfigAlertStore(c), nil
+	case "file":
+		return local.NewFileAlertStore(cfg.File)
+	default:
+		return nil, fmt.Errorf("Unrecognized rule storage mode %v, choose one of: configdb, file", cfg.Type)
```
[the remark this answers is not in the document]
jtlisi: Good catch, a lot of this code is based on the Ruler.

pkg/alertmanager/multitenant.go
jtlisi: I opened #2047 to fix this.

pkg/alertmanager/multitenant.go
jtlisi: I updated this function to include logic to do this. I also have removed deleteUser in order to avoid obfuscating when the mutex is being used.

pkg/alertmanager/multitenant.go
jtlisi: I agree

pkg/alertmanager/alerts/compat.go
jtlisi: Not needed in this PR, I'll remove this file

pkg/alertmanager/alerts/compat.go
jtlisi: Not needed in this PR, I'll remove this file

pkg/alertmanager/alerts/store.go
jtlisi: Will do

pkg/alertmanager/alerts/alerts.proto
jtlisi: It will allow for more robust serialization and versioning. This will ensure a stable format for communicating alertmanager configs.

pkg/alertmanager/multitenant.go
jtlisi: +1

pkg/alertmanager/multitenant.go
jtlisi: +1

jtlisi:
@pracucci this should be good for a second look

pkg/alertmanager/multitenant.go
```diff
@@ -263,45 +263,47 @@ func (am *MultitenantAlertmanager) loadAllConfigs() map[string]configs.View {
...
+	}
 
+	am.alertmanagersMtx.Lock()
+	defer am.alertmanagersMtx.Unlock()
+	for user, userAM := range am.alertmanagers {
+		if _, exists := cfgs[user]; !exists {
+			userAM.Stop()
+			delete(am.alertmanagers, user)
+			delete(am.cfgs, user)
+			level.Info(util.Logger).Log("msg", "deleting alertmanager", "user", user)
```
pracucci: I know I'm annoying. Technically this should go before `userAM.Stop()`. Why? Because if any critical issue occurs (ie. a panic) we see what it was doing, otherwise it's not guaranteed the execution reaches this point.

Here may eventually have a `deleted alertmanager` log (given deletions should not be that frequest, looks good knowing when it started and completed).

pkg/alertmanager/multitenant.go
```diff
@@ -366,70 +368,55 @@ func (am *MultitenantAlertmanager) setConfig(userID string, config configs.Confi
...
 
-	if amConfig, err = am.transformConfig(userID, amConfig); err != nil {
+	if userAmConfig, err = am.transformConfig(cfg.User, userAmConfig); err != nil {
 		return err
 	}
 
 	// If no Alertmanager instance exists for this user yet, start one.
 	if !hasExisting {
-		newAM, err := am.newAlertmanager(userID, amConfig)
+		level.Debug(util.Logger).Log("msg", "MultitenantAlertmanager: initializing new alertmanager tenant", "user", cfg.User)
```
pracucci: I'm seeing several logs calling it "alertmanager tenant". I personally find it a bit unclear. Maybe checking all the logs in this file and referring to it as the "per-tenant alertmanager" may be more clear? _I'm not a native english, so feel free to ignore this comment if you think the current one is good._

pkg/alertmanager/multitenant_test.go
```diff
@@ -0,0 +1,107 @@
...
+		stop:             make(chan struct{}),
+		done:             make(chan struct{}),
+	}
+
+	// Ensure the configs are synced correctly
+	err = am.updateConfigs()
+	require.NoError(t, err)
+	require.Len(t, am.alertmanagers, 2)
+
+	currentConfig, exists := am.cfgs["user1"]
```
pracucci: May we assert on `user2` too?

pkg/alertmanager/multitenant_test.go
```diff
@@ -0,0 +1,107 @@
...
+		store:            mockStore,
+		cfgs:             map[string]alerts.AlertConfigDesc{},
+		alertmanagersMtx: sync.Mutex{},
+		alertmanagers:    map[string]*Alertmanager{},
+		stop:             make(chan struct{}),
+		done:             make(chan struct{}),
+	}
+
+	// Ensure the configs are synced correctly
+	err = am.updateConfigs()
```
pracucci: Can be compacted into `require.NoError(t, am.updateConfigs())`.

_Same comment applies below._

pkg/alertmanager/multitenant_test.go
```diff
@@ -0,0 +1,107 @@
...
+		RawConfig: simpleConfigTwo,
+		Templates: []*alerts.TemplateDesc{},
+	}
+	err = am.updateConfigs()
+	require.NoError(t, err)
+
+	currentConfig, exists = am.cfgs["user1"]
+	require.True(t, exists)
+	require.Equal(t, simpleConfigTwo, currentConfig.RawConfig)
+}
```
pracucci: May we also remove one of the 3 users and assert on it?

pkg/alertmanager/storage.go
```diff
@@ -0,0 +1,40 @@
...
+	Type     string            `yaml:"type"`
+	ConfigDB client.Config     `yaml:"configdb"`
+	Local    local.StoreConfig `yaml:"local"`
+}
+
+// RegisterFlags registers flags.
+func (cfg *AlertStoreConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.Local.RegisterFlags(f)
+	cfg.ConfigDB.RegisterFlagsWithPrefix("alertmanager.", f)
+	f.StringVar(&cfg.Type, "alertmanager.storage.type", "configdb", "Method to use for backend alertmanager configs storage (configdb, file)")
```
pracucci: > Method to use for backend alertmanager configs storage

Maybe it's more clear if we describe it like `Type of backend to use to store alertmanager configs. Supported values are: "configdb", "local".` (pay attention "local", not "file")

pkg/alertmanager/storage.go
```diff
@@ -0,0 +1,40 @@
...
+	case "configdb":
+		c, err := client.New(cfg.ConfigDB)
+		if err != nil {
+			return nil, err
+		}
+		return alerts.NewConfigAlertStore(c), nil
+	case "local":
+		return local.NewStore(cfg.Local)
+	default:
+		return nil, fmt.Errorf("Unrecognized alertmanager storage mode %v, choose one of: configdb, local", cfg.Type)
```
pracucci: `mode` > ` file`

pkg/alertmanager/alerts/store.go
```diff
@@ -0,0 +1,64 @@
+package alerts
```
pracucci: Being the configdb store, may make sense to move it to `pkg/alertmanager/alerts/configdb/store.go` and just move the `AlertStore` interface back into `pkg/alertmanager/alerts`?

pkg/alertmanager/alerts/store.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/cortexproject/cortex/pkg/configs/client"
+)
+
+// AlertStore stores and configures users rule configs
+type AlertStore interface {
+	ListAlertConfigs(ctx context.Context) (map[string]AlertConfigDesc, error)
+}
+
+// ConfigAlertStore is a concrete implementation of RuleStore that sources rules from the config service
+type ConfigAlertStore struct {
```
pracucci: `ConfigDBAlertStore` would be more appropriate? _Just asking, up to you._

pkg/alertmanager/alerts/store.go
```diff
@@ -0,0 +1,64 @@
...
+	configClient client.Client
+	since        configs.ID
+	alertConfigs map[string]AlertConfigDesc
```
pracucci: Why do we need to retain it? Looking at `ListAlertConfigs()` implementation looks not useful to me.

pkg/alertmanager/local/store.go
```diff
@@ -0,0 +1,71 @@
...
+			if err != nil {
+				return err
+			}
+
+			content, err := ioutil.ReadFile(f.cfg.Path + info.Name())
+			if err != nil {
+				return err
+			}
+
+			user := strings.TrimSuffix(info.Name(), ext)
```
pracucci: As previously mentioned, I would add a comment to explain that the filename **must** match the user ID.

pkg/alertmanager/local/store.go
```diff
@@ -0,0 +1,71 @@
...
+func (f *Store) ListAlertConfigs(ctx context.Context) (map[string]alerts.AlertConfigDesc, error) {
+	configs := map[string]alerts.AlertConfigDesc{}
+	err := filepath.Walk(f.cfg.Path, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		ext := filepath.Ext(info.Name())
+
+		if !info.IsDir() && (ext == ".yml" || ext == ".yaml") {
```
pracucci: As previously mentioned, I would reverse this into a guard. Looks easier to read the code, if we `return nil` in case it's not a file or not a YAML extension.

I would also suggest to log with a `Warn` any skipped entry, because we wouldn't expect any spurious file/directory within `f.cfg.Path`, right?

pkg/cortex/modules.go
```diff
@@ -446,7 +446,7 @@ func (t *Cortex) stopConfigs() error {
 }
 
 func (t *Cortex) initAlertmanager(cfg *Config) (err error) {
-	t.alertmanager, err = alertmanager.NewMultitenantAlertmanager(&cfg.Alertmanager, cfg.ConfigStore)
+	t.alertmanager, err = alertmanager.NewMultitenantAlertmanager(&cfg.Alertmanager)
 	if err != nil {
 		return
```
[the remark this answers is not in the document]
jtlisi: We specify the `err` variable in the function signature and utilize a naked return pattern throughout the modules.go file. I personally like to explicitly return the error so I'll update it.

pkg/alertmanager/multitenant_test.go
jtlisi: 👍

pkg/alertmanager/multitenant_test.go
jtlisi: Will do

pkg/alertmanager/alerts/store.go
jtlisi: Makes sense, I will also move the `pkg/alertmananger/local` package to be `pkg/alertmanager/alerts/local`

pkg/alertmanager/alerts/store.go
jtlisi: `configs, err := c.configClient.GetAlerts(ctx, c.since)`

Will only return configs that have `IDs` greater than `c.since`. This avoids the configdb having to do a full pull of it's table.

pkg/alertmanager/alerts/store.go
jtlisi: SGTM

pkg/alertmanager/local/store.go
jtlisi: Sorry I didn't rectify this previously

pkg/alertmanager/local/store.go
jtlisi: Sorry I didn't rectify this previously

pkg/alertmanager/multitenant.go
```diff
@@ -263,45 +263,47 @@ func (am *MultitenantAlertmanager) loadAllConfigs() map[string]configs.View {
...
-// poll the configuration server. Not re-entrant.
-func (am *MultitenantAlertmanager) poll() (map[string]configs.View, error) {
-	configID := am.latestConfig
-	cfgs, err := am.configsAPI.GetAlerts(context.Background(), configID)
+// poll the alert store. Not re-entrant.
+func (am *MultitenantAlertmanager) poll() (map[string]alerts.AlertConfigDesc, error) {
+	cfgs, err := am.store.ListAlertConfigs(context.Background())
 	if err != nil {
-		level.Warn(util.Logger).Log("msg", "MultitenantAlertmanager: configs server poll failed", "err", err)
+		level.Warn(util.Logger).Log("msg", "MultitenantAlertmanager: configs alert store poll failed", "err", err)
```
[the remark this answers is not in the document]
jtlisi: I'll use an embedded logger with a component field.

pkg/alertmanager/multitenant.go
```diff
@@ -263,45 +263,47 @@ func (am *MultitenantAlertmanager) loadAllConfigs() map[string]configs.View {
...
-	am.latestMutex.Lock()
-	am.latestConfig = cfgs.GetLatestConfigID()
-	am.latestMutex.Unlock()
-	return cfgs.Configs, nil
+	return cfgs, nil
 }
 
-func (am *MultitenantAlertmanager) addNewConfigs(cfgs map[string]configs.View) {
+func (am *MultitenantAlertmanager) syncConfigs(cfgs map[string]alerts.AlertConfigDesc) {
 	// TODO: instrument how many configs we have, both valid & invalid.
```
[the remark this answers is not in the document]
jtlisi: I broke the `totalConfigs` metric up to have a `status` label

pkg/alertmanager/multitenant.go
jtlisi: I moved this log line and added a `deleted` log line to clarify

pkg/alertmanager/multitenant.go
jtlisi: I like `per-tenant alertmanager`. I'll update it.

pkg/alertmanager/multitenant.go
```diff
@@ -366,70 +368,55 @@ func (am *MultitenantAlertmanager) setConfig(userID string, config configs.Confi
...
 		if err != nil {
 			return err
 		}
 		am.alertmanagersMtx.Lock()
-		am.alertmanagers[userID] = newAM
+		am.alertmanagers[cfg.User] = newAM
 		am.alertmanagersMtx.Unlock()
-	} else if am.cfgs[userID].AlertmanagerConfig != config.AlertmanagerConfig || hasTemplateChanges {
+	} else if am.cfgs[cfg.User].RawConfig != cfg.RawConfig || hasTemplateChanges {
+		level.Debug(util.Logger).Log("msg", "MultitenantAlertmanager: updating new alertmanager tenant", "user", cfg.User)
```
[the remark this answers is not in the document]
jtlisi: Agreed.

jtlisi:
Closing in favor of #2125

jtlisi:
Reopening since #2128 has been merged.

Issue #1477, opened 2019-06-27
Elastic as a storage backend


roopakparikh:
Elastic is probably the most popular opensource document storage out there.  I would like to add Elastic as backend storage for cortex. I would also like to hear any opinions out there on that subject.

csmarchbanks:
I am fine with someone adding elastic as a backend (and offering to support it). Would it make sense to add a pluggable storage instead though? Similar to the approach Jaeger is taking.

friedrichg:
Elastic made sense when cortex needed a KV DB for the indexes (Cassandra or Dynamodb ) in chunks.
With chunks deprecated and removed in v1.14.0. This is no longer possible/needed.

PR #1694, opened 2019-09-24
Add ElasticSearch as a new Index Client

mizeng:
Currently Cortex only support GCP, AWS Dynamo, Cassandra for Chunk Index. Here add a new choice, which is ElasticSearch since it's a very popular NoSQL storage.

Implement Hint:
Only write some new go code in **pkg/chunk/elastic**, other file changes are introduced by `go mod tidy` and `go mod vendor`.

Sample Config to Use ElasticSearch in https addr(will skip tls verify by default, user can change it with passing some config) and http addr:
```
schema_config:
        configs:
        - from: 2018-04-15
          store: elastic
          object_store: s3
          schema: v9
          index:
            prefix: index_
            period: 168h

      storage_config:
        elastic:
          address: https://es_addr:443
          user: user
          password: password
```

or 

```
schema_config:
        configs:
        - from: 2018-04-15
          store: elastic
          object_store: s3
          schema: v9
          index:
            prefix: index_
            period: 168h

      storage_config:
        elastic:
          address: https://es_addr:443
          user: user
          password: password
          tls_skip_verify: false
          cert_file: cert_file_addr
          key_file: key_file_addr
          ca_file: ca_file
```

and 

```
schema_config:
        configs:
        - from: 2018-04-15
          store: elastic
          object_store: s3
          schema: v9
          index:
            prefix: index_
            period: 168h

      storage_config:
        elastic:
          address: http://es_addr
```


mizeng:
why the ci lint failed by unchanged code? can it pass for current master code?

bboreham:
Yes, CI passes for current master.  You can view test runs at https://circleci.com/gh/cortexproject/cortex/tree/master
The `lint` script is giving you these hints:
```
./pkg/chunk/elastic/es_index_client_test.go: run gofmt -s -w ./pkg/chunk/elastic/es_index_client_test.go
./pkg/chunk/elastic/es_table_client.go: run gofmt -s -w ./pkg/chunk/elastic/es_table_client.go
./pkg/chunk/elastic/es_index_client.go: run gofmt -s -w ./pkg/chunk/elastic/es_index_client.go
./pkg/chunk/storage/factory.go: run gofmt -s -w ./pkg/chunk/storage/factory.go
```

The complexity warning is not considered fatal by the script.

mizeng:
@bboreham  Thanks for your reply. But I randomly check one https://circleci.com/gh/cortexproject/cortex/10263, seems it has no lint step compared to mine: https://circleci.com/gh/cortexproject/cortex/10277.

mizeng:
However I finally understood below message, so it tells me to format golang code..
```
./pkg/chunk/elastic/es_index_client_test.go: run gofmt -s -w ./pkg/chunk/elastic/es_index_client_test.go
./pkg/chunk/elastic/es_table_client.go: run gofmt -s -w ./pkg/chunk/elastic/es_table_client.go
./pkg/chunk/elastic/es_index_client.go: run gofmt -s -w ./pkg/chunk/elastic/es_index_client.go
./pkg/chunk/storage/factory.go: run gofmt -s -w ./pkg/chunk/storage/factory.go
```

I will fix it now.

mizeng:
now failed on "Check vendor directory is consistent." I'm a little confused. Could someone help to guide me how to fix this?

gouthamve:
You now need to run `go mod tidy; go mod vendor`

mizeng:
tried `go mod tidy; go mod vendor`, but still failed

codesome:
@mizeng try rebasing with master and run those command again.

gouthamve:
Also make sure you `go1.13`

mizeng:
already did both, still fail.

codesome:
Tried myself now. Rebasing the PR with master and running tidy and vendor does have some difference.

mizeng:
Thanks a lot. I may need run `go mod tidy` and `go mod vendor` again and see if it resolves the failure.

mizeng:
Finally passed CI.. Summary as below:

1. Cortex uses go **1.13** now, need to upgrade local go version too.
2. If adding or modifying any dependency, need to run `go mod tidy` and `go mod vendor` to sync it in vendor.

Thanks for all the help from @codesome  @bboreham @gouthamve.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -276,14 +279,25 @@ func NewESIndexClient(cfg Config) (chunk.IndexClient, error) {
 }
 
 func newES(cfg Config) (*elastic.Client, error) {
+	//fix x509: certificate signed by unknown authority
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
zetaab: in my opinion this is not correct way to handle "certificates". If you use insecure always it does not identify the elasticsearch server at all. If people want use insecure it should be startup flag or configuration option somewhere, but default value insecure is wrong

mizeng: make sense. I will fix it.

mizeng: done, plz review again.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+	CaFile        string `yaml:"ca_file"`
+}
+
+// RegisterFlags registers flags.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.Address, "elastic.address", "http://127.0.0.1:9200", "Address of ElasticSearch.")
+	f.StringVar(&cfg.IndexType, "elastic.index_type", "lokiindex", "Index Type used in ElasticSearch.")
+	f.StringVar(&cfg.User, "elastic.user", "user", "User used in ElasticSearch Basic Auth.")
+	f.StringVar(&cfg.Password, "elastic.password", "password", "Password used in ElasticSearch Basic Auth.")
+	f.BoolVar(&cfg.TLSSkipVerify, "elastic.tls_skip_verify", true, "Skip tls verify or not. Default is skip.")
```
zetaab: in my opinion the default value for tls skip verify should be false. False is the securest option and if someone really wants to use something else, he knows what he does.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+
+func newES(cfg Config) (*elastic.Client, error) {
+	//fix x509: certificate signed by unknown authority
+	var tr *http.Transport
+	if cfg.TLSSkipVerify { // if skip TLS Verify
+		tr = &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+	} else { // not skip TLS Verify
+		cert, err := tls.LoadX509KeyPair(cfg.CertFile, cfg.KeyFile)
```
zetaab: no option to use with valid certificates that are already in rootcabundle? Now this assumes 1) insecure 2) certs provided in start flags.

mizeng: I haven't used the way you mentioned. Could you share some details?

pkg/chunk/elastic/es_index_client.go
mizeng: In my opinion, most of the customers are not care about tls verify and what they want is to use https the same as http, which means they need not do anything.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+		RangeValue: rangeValue,
+		Value:      value,
+	})
+}
+
+func (e *esClient) BatchWrite(ctx context.Context, batch chunk.WriteBatch) error {
+	b := batch.(*writeBatch)
+
+	indexName := b.entries[0].TableName
+	exists, err := e.client.IndexExists(indexName).Do(ctx)
```
khaines: A check for an index on every write seems inefficient in the pattern that cortex uses. I'm only starting to work through the review on this, however I know that Elasticsearch has an auto-create index feature and templates. This can be used to create the index if it doesn't already exist on the first bulk index request, plus having the mapping template applied automatically. 

The management of the template(s) would be something I would expect the tablemanager to do.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+type esClient struct {
+	cfg    Config
+	client *elastic.Client
+}
+
+func (e *esClient) Stop() {
+	e.client.Stop()
+}
+
+// ES batching isn't really useful in this case, its more to do multiple
```
khaines: Nit: This comment could be refactored to note that the code is performing a bulk index request, vs parallel writes.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+		valueTermQuery = elastic.NewTermQuery("value", query.ValueEqual)
+
+	case query.ValueEqual != nil:
+		valueTermQuery = elastic.NewTermQuery("value", query.ValueEqual)
+
+	case query.ValueEqual == nil:
+		break
+	}
+
+	exists, err := e.client.IndexExists(query.TableName).Do(ctx)
```
khaines: There's no functional benefit to checking if the index exists or not. In a positive case, you are performing 2 HTTP calls per query or 1 on failure.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+
+func (e *esClient) BatchWrite(ctx context.Context, batch chunk.WriteBatch) error {
+	b := batch.(*writeBatch)
+
+	indexName := b.entries[0].TableName
+	exists, err := e.client.IndexExists(indexName).Do(ctx)
+	if err != nil {
+		// Handle error
+		level.Error(util.Logger).Log("msg", fmt.Sprintf("IndexName %s exists check has error!", indexName))
+		panic(err)
```
khaines: calling panic from inside the ingester process will cause data loss.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+		valueTermQuery = elastic.NewTermQuery("value", query.ValueEqual)
+
+	case query.ValueEqual == nil:
+		break
+	}
+
+	exists, err := e.client.IndexExists(query.TableName).Do(ctx)
+	if err != nil {
+		// Handle error
+		fmt.Println(exists)
```
khaines: This error handling doesn't log data through the go-kit library, but it also does not return the error upstream. this will lead to a panic later in the function

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+		baseQuery = baseQuery.Query(valueTermQuery)
+	}
+	if rangeQuery != nil {
+		baseQuery = baseQuery.Query(rangeQuery)
+	}
+
+	// Search with a term query
+	searchResult, err := baseQuery.
+		Sort("range", true). // sort by "range" field, ascending
+		From(0).Size(maxFetchDocs).
```
khaines: There seems to be a fixed limit here of 1000 records.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+		Pretty(true). // pretty print request and response JSON
```
khaines: I'm unsure about the benefit this option provides, however it does cause more processing on the server side.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,344 @@
...
+		Do(ctx)       // execute
+
+	if searchResult == nil || searchResult.Hits == nil {
+		return nil
+	}
+
+	if err != nil {
+		// Handle error
+		level.Error(util.Logger).Log("msg", fmt.Sprintf("Query in index %s met error!", query.TableName))
+		panic(err)
```
khaines: Panicing seems to be a bad behaviour during a query.

pkg/chunk/elastic/es_table_client.go
```diff
@@ -0,0 +1,77 @@
...
+	response, err := c.client.IndexNames()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	return response, nil
+}
+
+func (c *tableClient) CreateTable(ctx context.Context, desc chunk.TableDesc) error {
+	//// Use the IndexExists service to check if a specified index exists.
```
khaines: If you intend this to be a no-op call and use Elastics auto index management with templates, please clear out this commented code and replace with a comment note on how tables are created.

mizeng:
@khaines Thanks for your comments, I will look through them and reply.

pkg/chunk/elastic/es_index_client.go
mizeng: you're right. I will change it.

pkg/chunk/elastic/es_index_client.go
mizeng: good catch!

pkg/chunk/elastic/es_index_client.go
mizeng: Let me try without `Pretty(true)`.

pkg/chunk/elastic/es_index_client.go
mizeng: Will change it to configurable.

pkg/chunk/elastic/es_index_client.go
mizeng: Indeed. I forgot this.

mizeng:
@khaines  Sorry for late reply, just get some time to improve my PR based on your comments. Could you help to review again?

pkg/chunk/elastic/es_index_client.go
khaines: It is a better security practice to have this as false by default.

pkg/chunk/elastic/es_index_client.go
khaines: While this is now configurable, your design is limited to return only that number of results total. Also, it cannot be larger than `index.max_result_window` which by default is set to 10K in ES. 

In order to be an effective at scale index store, the storage layer needs to be able to return many tens of thousands of results. You probably need to look at this search portion of code and add pagination for larger result sets, otherwise, there is a decent risk of results being truncated from anything but small result sets.

mizeng:
@khaines modified based on your comments. could you please review again?

zetaab:
@mizeng could you rebase against master

@khaines could you please re-review this PR. At least we are waiting for this

mizeng:
close by mistake, reopen it.
Add prometheus metrics for ES Index Client.
Use ES Scroll in Query to improve performance.

pkg/chunk/elastic/es_index_client_test.go
```diff
@@ -0,0 +1,56 @@
+package elastic
+
+// There's no good embedded ElasticSearch, so we use a real ElasticSearch instance.
```
khaines: There's no embedded ES and this header is a great summary of how to get a local ES in docker. However, these tests can be uncommented but protected from running normally using a build flag. There are other integration tests that work against sql using `// +build integration`, so that does offer an example of how to do it for the ES tests. I would chose a different flag however, since Elasticsearch isn't (for at least now) likely to be part of the core integration tests.

pkg/chunk/elastic/es_table_client.go
```diff
@@ -0,0 +1,99 @@
...
+	// Instead, we create Template.
+	templateName := "loki-index"
+	// check if template exists, and create one if not exist
+	exists, err := c.client.IndexTemplateExists(templateName).Do(ctx)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	if !exists {
+		tmpl := `{
+		"index_patterns":["lokiindex_*"],
```
khaines: this template pattern should not be hard coded. Also this seems to be leaking a loki detail into cortex.

pkg/chunk/elastic/es_table_client.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil, errors.WithStack(err)
+	}
+
+	return response, nil
+}
+
+func (c *tableClient) CreateTable(ctx context.Context, desc chunk.TableDesc) error {
+	// Here we use ElasticSearch auto index management with templates, no need to create table manually.
+	// Instead, we create Template.
+	templateName := "loki-index"
```
khaines: template name should be configurable, and not contain loki in the default.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,316 @@
...
+
+const (
+	null = string('\xff')
+)
+
+var esRequestDuration = instrument.NewHistogramCollector(prometheus.NewHistogramVec(prometheus.HistogramOpts{
+	Name: "es_request_duration_seconds",
+	Help: "Time spent doing ElasticSearch requests.",
+
+	// from 0us to 10s. TODO: Confirm that this is the case for ApplicationAutoScaling.
```
khaines: Please revist this comment. It appears to be a copy & paste where some details don't apply to ES

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,316 @@
...
+	TLSSkipVerify bool   `yaml:"tls_skip_verify"`
+	CertFile      string `yaml:"cert_file"`
+	KeyFile       string `yaml:"key_file"`
+	CaFile        string `yaml:"ca_file"`
+}
+
+// RegisterFlags registers flags.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.Address, "elastic.address", "http://127.0.0.1:9200", "Address of ElasticSearch.")
+	f.StringVar(&cfg.IndexType, "elastic.index_type", "lokiindex", "Index Type used in ElasticSearch.")
```
khaines: Please choose a different default. This is cortex :)

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,316 @@
...
+
+func (e *esClient) query(ctx context.Context, query chunk.IndexQuery, callback func(chunk.ReadBatch) (shouldContinue bool)) error {
+	var rangeQuery *elastic.RangeQuery
+	var valueTermQuery *elastic.TermQuery
+
+	level.Debug(util.Logger).Log("msg", fmt.Sprintf(
+		"hash [%s], rangeValuePrefix [%s], rangeValueStart [%s]", query.HashValue, query.RangeValuePrefix, query.RangeValueStart))
+	hashTermQuery := elastic.NewTermQuery("hash", query.HashValue)
+	switch {
+	case len(query.RangeValuePrefix) > 0 && query.ValueEqual == nil:
```
khaines: Comments for each case would assist others reading this to understand the intent of the code.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,316 @@
...
+		valueTermQuery = elastic.NewTermQuery("value", query.ValueEqual)
+
+	case len(query.RangeValueStart) > 0 && query.ValueEqual == nil:
+		rangeQuery = elastic.NewRangeQuery("range").Gte(string(query.RangeValueStart))
+
+	case len(query.RangeValueStart) > 0 && query.ValueEqual != nil:
+		rangeQuery = elastic.NewRangeQuery("range").Gte(string(query.RangeValueStart))
+		valueTermQuery = elastic.NewTermQuery("value", query.ValueEqual)
+
+	case query.ValueEqual != nil:
```
khaines: these value filters should be in its own block. It will reduce the duplication in cases above and separate this back into a set of range cases and value cases.

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,316 @@
...
+		valueTermQuery = elastic.NewTermQuery("value", query.ValueEqual)
+
+	case query.ValueEqual == nil:
+		break
+	}
+
+	// Build the query
+	baseQuery := e.client.Scroll(query.TableName).Size(e.cfg.MaxFetchDocs).
```
khaines: Scroll is not the correct API to use for these queries. It certainly can return a large number of results, but it isn't meant for real-time user queries. 

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-scroll

See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-from-size instead for how to do this with the `search` endpoint. Note there is also limitations here, based on the configuration of ES itself.

khaines (changes requested):
Sorry for the delay in looking these over again. I've left some more feedback regarding the changes that I feel needs addressing.

Feel free to ping me on slack if you have any questions

mizeng:
Thanks for the comments. I will take a look when I get some time after this week since I'm busy in a whole week training.

pkg/chunk/elastic/es_index_client.go
mizeng: Considering big amount of data in my case of testing, I prefer to use scroll rather than search-from-size.

pkg/chunk/elastic/es_index_client.go
mizeng: I'm sorry for I don't get your point. Could you explain more? Hence similar cases is used in cassandra client.

pkg/chunk/elastic/es_index_client.go
mizeng: let me try search_after. https://www.elastic.co/guide/en/elasticsearch/reference/master/search-request-body.html#request-body-search-search-after

khaines: That sounds like a better approach than Scroll as it saves the stateful nightmare that scroll puts on a cluster under load. However since it is stateless, depending on your caching configuration the query will be executed multiple times. Ordering can change also as new records are added, but I would expect the sort to be on time and the ingesters are writing data chronologically so this shouldn't be as much of an issue.

mizeng: Yeah, considering current approaches: search_from, scroll, search_after, seems `search_after` is the most suitable one for realtime, large data set.

pracucci:
Fixes #1477.

pkg/chunk/elastic/es_index_client.go
khaines: Disregard. I don't recall why I put this here.

khaines:
@mizeng are you still intending to complete this PR? The last commit was in the beginning of December.

mizeng:
Sorry for I was quite busy for last 3 months due to my baby born. I will try to continue this PR this week.

Wing924:
BTW, Elasticsearch itself have very fast reverse index function.
Why not let Elasticsearch store chunks and let it mange index?

mizeng:
@khaines  I've modified based on your comments. could you help to review again?

khaines (changes requested):
Thank you @mizeng for getting on the changes. This is looking really good, but there are just a few important but minor changes to do and then I think this is ready to be an experimental feature.

docs/configuration/config-file-reference.md
```diff
@@ -1499,6 +1499,55 @@ filesystem:
   # CLI flag: -local.chunk-directory
   [directory: <string> | default = ""]
 
+elastic:
+  # Address of ElasticSearch.
+  # CLI flag: -elastic.address
+  [address: <string> | default = "http://127.0.0.1:9200"]
+
+  # Template name in ElasticSearch.
+  # CLI flag: -elastic.template_name
```
pracucci: Please make sure CLI flags are `-` separated, while config file are `_` separated. In this case, the config file option name `template_name` is fine, while the CLI flag should be `-elastic.template-name`.

docs/configuration/config-file-reference.md
```diff
@@ -1499,6 +1499,55 @@ filesystem:
...
+  # CLI flag: -elastic.number_of_shards
+  [number_of_shards: <int> | default = 3]
+
+  # Number of replicas in creating template.
+  # CLI flag: -elastic.number_of_replicas
+  [number_of_replicas: <int> | default = 1]
+
+  # User used in ElasticSearch Basic Auth.
+  # CLI flag: -elastic.user
+  [user: <string> | default = ""]
```
pracucci: What do you think about `username`?

pkg/chunk/elastic/es_index_client.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+	return indexClient, nil
+}
+
+func newES(cfg Config) (*elastic.Client, error) {
+	//fix x509: certificate signed by unknown authority
+	var tr *http.Transport
+	if !strings.Contains(cfg.Address, "https") || cfg.TLSSkipVerify {
+		tr = &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
mizeng: so you mean even for http, we'll have option to not skip tls verify?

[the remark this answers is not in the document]
mizeng: I think I've already have it configurable. If user use an ES addr without "https" or set "tls_skip_verify" to true, it will use the skip verify condition.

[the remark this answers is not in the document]
khaines: Yes you did. I must have missed the if statement above

khaines (approved):
Thanks @mizeng! I think this is ready to be an experimental feature, so I've approved. Before it can be merged, it requires a 2nd approval by a maintainer.

pracucci (commented):
Thanks @mizeng for your work! I have no personal experience with ElasticSearch, so I feel I'm not the best person to do the 2nd review for this PR.

Generally speaking, I would be very glad to see an integration test for the ElasticSearch support. An option could be adding ElasticSearch case to the test `integration/chunks_storage_backends_test.go`. May you take a look at how the integration test works and see if it's feasible adding ElasticSearch case too, please?

mizeng:
> Thanks @mizeng for your work! I have no personal experience with ElasticSearch, so I feel I'm not the best person to do the 2nd review for this PR.
> 
> Generally speaking, I would be very glad to see an integration test for the ElasticSearch support. An option could be adding ElasticSearch case to the test `integration/chunks_storage_backends_test.go`. May you take a look at how the integration test works and see if it's feasible adding ElasticSearch case too, please?

Seems only DynamoDB and BigTab are included in `integration/chunks_storage_backends_test.go` ? Only adding ElasticSearch case seems weird. I prefer a seperate PR to include all existing chunk store/index store in the e2e testing together.

pracucci:
> Seems only DynamoDB and BigTab are included in integration/chunks_storage_backends_test.go ? Only adding ElasticSearch case seems weird. I prefer a seperate PR to include all existing chunk store/index store in the e2e testing together.

If you rebase `master` you will also get Cassandra, so basically all 3 supported index stores are now in that integration test. I believe we should cover ElasticSearch there as well within this PR.

mizeng:
> > Seems only DynamoDB and BigTab are included in integration/chunks_storage_backends_test.go ? Only adding ElasticSearch case seems weird. I prefer a seperate PR to include all existing chunk store/index store in the e2e testing together.
> 
> If you rebase `master` you will also get Cassandra, so basically all 3 supported index stores are now in that integration test. I believe we should cover ElasticSearch there as well within this PR.

Fair enough. I'll see what I can do.

mizeng:
@pracucci I can not run the test successfully. Is it working? I can not even find the image "quay.io/cortexproject/cortex:latest".

mizeng:
Also current testing will pull a cortex image, while my code is not in that cortex image. Dead lock.
So I suggest merge this pr first, build a cortex docker image in the repo "quay.io/cortexproject/cortex", then continue to write e2e testing.

pracucci:
> Also current testing will pull a cortex image, while my code is not in that cortex image. Dead lock. So I suggest merge this pr first, build a cortex docker image in the repo "quay.io/cortexproject/cortex", then continue to write e2e testing.

@mizeng There's no dead lock. Let me explain how integration tests work first, so you can work on it.

When integration tests run in CI, we build the Cortex docker image based on the PR code and then run the integration tests against it.

When running tests locally (or while developing integration tests) you should first build the Docker image running:

```
make ./cmd/cortex/.uptodate
```

This will locally build the `quay.io/cortexproject/cortex:latest` image used by integration tests. Whenever the Cortex code changes (`cmd/`, `pkg/` or vendors) you should re-run `make ./cmd/cortex/.uptodate` to rebuild the image, while it's **not** necessary to rebuild the image while developing integration tests.

Once the Docker image is built, you can run integration tests with the following command:

```
go test -v -tags=requires_docker ./integration/...
```

If you want to run a single test you can use a filter. For example, to only run `TestChunksStorageAllIndexBackends` you can use:

```
go test -tags=requires_docker -v ./integration -run ^(TestChunksStorageAllIndexBackends)$
```

Talking specifically about this PR, a good starting point may be looking at how `TestChunksStorageAllIndexBackends` works and integrating ElasticSearch there (as yet another entry in the schema config).

mizeng:
> > Also current testing will pull a cortex image, while my code is not in that cortex image. Dead lock. So I suggest merge this pr first, build a cortex docker image in the repo "quay.io/cortexproject/cortex", then continue to write e2e testing.
> 
> @mizeng There's no dead lock. Let me explain how integration tests work first, so you can work on it.
> 
> When integration tests run in CI, we build the Cortex docker image based on the PR code and then run the integration tests against it.
> 
> When running tests locally (or while developing integration tests) you should first build the Docker image running:
> 
> ```
> make ./cmd/cortex/.uptodate
> ```
> 
> This will locally build the `quay.io/cortexproject/cortex:latest` image used by integration tests. Whenever the Cortex code changes (`cmd/`, `pkg/` or vendors) you should re-run `make ./cmd/cortex/.uptodate` to rebuild the image, while it's **not** necessary to rebuild the image while developing integration tests.
> 
> Once the Docker image is built, you can run integration tests with the following command:
> 
> ```
> go test -v -tags=requires_docker ./integration/...
> ```
> 
> If you want to run a single test you can use a filter. For example, to only run `TestChunksStorageAllIndexBackends` you can use:
> 
> ```
> go test -tags=requires_docker -v ./integration -run ^(TestChunksStorageAllIndexBackends)$
> ```
> 
> Talking specifically about this PR, a good starting point may be looking at how `TestChunksStorageAllIndexBackends` works and integrating ElasticSearch there (as yet another entry in the schema config).

Thanks, it's very helpful.

mizeng:
BTW, my dokcer start up need to pass `-e "discovery.type=single-node"`.
How to do it with e2e.newHttpService? I've tried below, but not work.

```
return e2e.NewHTTPService(
		"elasticsearch",
		// If you change the image tag, remember to update it in the preloading done
		// by CircleCI too (see .circleci/config.yml).
		"docker.elastic.co/elasticsearch/elasticsearch:6.4.3",
		e2e.NewCommand("-e", "discovery.type=single-node"),
		// readiness probe inspired from https://github.com/kubernetes/examples/blob/b86c9d50be45eaf5ce74dee7159ce38b0e149d38/cassandra/image/files/ready-probe.sh
		e2e.NewHTTPReadinessProbe(9200, "/", 200),
		9200,
	)
```

pracucci:
> Still met error like below
> 
> ```
> ingester: level=error ts=2020-03-24T09:49:04.2524663Z caller=log.go:141 msg="error initializing cortex" err="Unrecognized storage client elastic, choose one of: aws, azure, cassandra, inmemory, gcp, bigtable, bigtable-hashed\n
> ```
> 
> this is not related to .uptodate, but to the same storage config that index and chunk used. I may need to separate them.

The entry you will add to the schema config will need to use elastic for the index and another storage (ie. S3) for the chunks.

mizeng:
while trying use cassandra for the trunks, I'm alaways facing this problem:
```
Ports for container: e2e-cortex-test-table-manager Mapping: map[80:33091]
table-manager: 2020-03-25 07:30:48.104000 I | gocql: dns error: lookup : no such host
table-manager: level=error ts=2020-03-25T07:30:48.1082501Z caller=log.go:141 msg="error initializing cortex" err="gocql: unable to create session: failed to resolve any of the provided hostnames\ngithub.com/cortexproject/cortex/pkg/chunk/cassandra.(*Config).session\n\t/go/src/github.com/cortexproject/cortex/pkg/chunk/cassandra/storage_client.go:105\ngithub.com/cortexproject/cortex/pkg/chunk/cassandra.NewTableClient\n\t/go/src/github.com/cortexproject/cortex/pkg/chunk/cassandra/table_client.go:20\ngithub.com/cortexproject/cortex/pkg/chunk/storage.NewTableClient\n\t/go/src/github.com/cortexproject/cortex/pkg/chunk/storage/factory.go:239\ngithub.com/cortexproject/cortex/pkg/cortex.(*Cortex).initTableManager\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/modules.go:409\ngithub.com/cortexproject/cortex/pkg/cortex.(*Cortex).initModuleServices\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:283\ngithub.com/cortexproject/cortex/pkg/cortex.New\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:222\nmain.main\n\t/go/src/github.com/cortexproject/cortex/cmd/cortex/main.go:106\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1357\ngithub.com/cortexproject/cortex/pkg/chunk/cassandra.NewTableClient\n\t/go/src/github.com/cortexproject/cortex/pkg/chunk/cassandra/table_client.go:22\ngithub.com/cortexproject/cortex/pkg/chunk/storage.NewTableClient\n\t/go/src/github.com/cortexproject/cortex/pkg/chunk/storage/factory.go:239\ngithub.com/cortexproject/cortex/pkg/cortex.(*Cortex).initTableManager\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/modules.go:409\ngithub.com/cortexproject/cortex/pkg/cortex.(*Cortex).initModuleServices\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:283\ngithub.com/cortexproject/cortex/pkg/cortex.New\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:222\nmain.main\n\t/go/src/github.com/cortexproject/cortex/cmd/cortex/main.go:106\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1357\nerror initialising module: table-manager\ngithub.com/cortexproject/cortex/pkg/cortex.(*Cortex).initModuleServices\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:285\ngithub.com/cortexproject/cortex/pkg/cortex.New\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:222\nmain.main\n\t/go/src/github.com/cortexproject/cortex/cmd/cortex/main.go:106\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1357"
```

pracucci:
> while trying use cassandra for the trunks, I'm always facing this problem:

Looks the hostname configured for cassandra is not correct. Feel free to reach me out in the CNCF Slack workspace (I'm "Marco Pracucci") and I can provide a quicker feedback, or please share more details here about how the schema is configured.

integration/chunks_storage_backends_test.go
```diff
@@ -24,27 +24,44 @@ func TestChunksStorageAllIndexBackends(t *testing.T) {
...
+			IndexStore: indexStore, ObjectStore: objStore}
+		i++
 	}
 
 	storageFlags := mergeFlags(ChunksStorageFlags, map[string]string{
 		"-cassandra.addresses": cassandra.NetworkHTTPEndpoint(),
 		"-cassandra.keyspace":  "tests", // keyspace gets created on startup if it does not exist
 	})
 
+	storageFlags = mergeFlags(ChunksStorageFlags, map[string]string{
```
pracucci: You can merge these flags along with the cassandra flags above.

mizeng:
> Any specific reason why we're not using https://hub.docker.com/_/elasticsearch ?

No specific reason. I just pick one image of 6.X from an official site which can work in my dev machine last year.

integration/chunks_storage_backends_test.go
```diff
@@ -24,27 +24,44 @@ func TestChunksStorageAllIndexBackends(t *testing.T) {
 	dynamo := e2edb.NewDynamoDB()
 	bigtable := e2edb.NewBigtable()
 	cassandra := e2edb.NewCassandra()
-
-	stores := []string{"aws-dynamo", "bigtable", "cassandra"}
+	elasticsearch := e2edb.NewElasticSearch()
+
+	// index store : chunk store
+	stores := map[string]string{
```
pracucci: Maps do not guarantee ordering. Please switch back to a slice using a custom struct, ie.

```
stores := []struct{
   IndexStore string
   ChunkStore string
}{
  // ...
}
```

mizeng: though I don't see any ordering problem here, but fair enough, I will change to struct array.

mizeng:
suspect the port map missing part.
Since I can see error like 
`table-manager: level=error ts=2020-03-25T09:06:12.1298848Z caller=table_manager.go:252 msg="error syncing tables" err="Get http://e2e-cortex-test-elasticsearch:9200/_all/_settings: dial tcp 192.168.160.3:9200: connect: connection refused"`
When I add a debug point before table-manager, I can not visit "http://e2e-cortex-test-elasticsearch:9200". 

So do we need to add port map for e2e testing?

mizeng:
previous assumption is wrong. I see the publish port. Maybe it’s related to ElasticSearch running in docker network.

mizeng:
got stucked. I added env "network.host=0.0.0.0" as suggested in StackOverflow, and can access "e2e-cortex-test-elasticsearch:9200" in container "e2e-cortex-test-cassandra" while debugging. But still met error "Get http://e2e-cortex-test-elasticsearch:9200/_all/_settings: dial tcp 192.168.160.3:9200: connect: connection refused" in testing.

mizeng:
Another problem is that when Index Store (ElastciSearch) and Chunk Store (e.g.Bigtable) is different, table manager can not create table for chunk store, thus caused future testing error like "table not exist".

I really need some expert's help on this e2e things.

mizeng:
After one day's trial, finally I worked out with only using `filesystem` as chunk storage to avoid any of the dependency issue like cassandra/bigtable/etc.  

@pracucci could you plz help review again?

mizeng:
though I don't know why integration test fails again. I can pass it in my local.

mizeng:
finally!!! found the testing failed due to **timeout**. The stack trace is really not obvious.
Fixed it.

mizeng:
2 weeks passed. May I get any progress on the merge?

bboreham:
Hi, sorry about the delay in getting back to this.
Since we added storage plugins via gRPC in #2220, it would be better to re-do this as a plugin.

We appreciate you put a lot of work into this PR, but it's unlikely it will get merged.

Issue #1698, opened 2019-09-25
Long-term index caching misses chunks on lookups


gouthamve:
In [`GetChunksForSeries`](https://github.com/cortexproject/cortex/blob/78e0607b405eba31c151d34d333de201f76a9c33/pkg/chunk/schema.go#L728-L733), we do a `RangeValueStart` on `bucket.from`. Now I thought if the `bucket.hashKey` is same, the `bucket.from` is also same, turns out that is soo not true.

[`bucket.from`](https://github.com/cortexproject/cortex/blob/78e0607b405eba31c151d34d333de201f76a9c33/pkg/chunk/schema_config.go#L293-L306) is the relative milliseconds from the actual bucket start time.

Now I did not realise this while building the long-term caching approach, I thought if the hashKey and RangeKey match, everything else matches: https://github.com/cortexproject/cortex/blob/78e0607b405eba31c151d34d333de201f76a9c33/pkg/chunk/schema_caching.go#L99-L100

But when we split the ranges, we split the actual from and to, to "cacheable" and "active" ranges here: https://github.com/cortexproject/cortex/blob/78e0607b405eba31c151d34d333de201f76a9c33/pkg/chunk/schema_caching.go#L66

and end up picking only the "active" range query on merge. This means the `bucket.from` is higher than it should be and we end up filtering chunks out. This causes gaps in the queries and we're dropping entire chunks to the floor.

PR #1699, opened 2019-09-25
Simplify long-term caching

gouthamve:
An attempt to fix https://github.com/cortexproject/cortex/issues/1698

We don't mix things when the time-range for the query overlaps the
"active" time-range. We consider all index entries as active. This is
because the `IndexQuery` fields depend on the `from` value and changing
it might mess things up.

This is kinda only effective when paired with query-frontend as most
queries issued fall in the active-range, but the query-frontend with
it's splitting would make sure the queriers actually only see some
queries that are totally in the non-active range.


pkg/chunk/schema_caching.go
```diff
@@ -90,24 +50,25 @@ func splitTimesByCacheability(from, through model.Time, cacheBefore model.Time)
 	return from, cacheBefore, cacheBefore, through
 }
 
-func mergeCacheableAndActiveQueries(cacheableQueries []IndexQuery, activeQueries []IndexQuery) []IndexQuery {
-	finalQueries := make([]IndexQuery, 0, len(cacheableQueries)+len(activeQueries))
+func (s *schemaCaching) setImmutability(from, through model.Time, f func(from, through model.Time) ([]IndexQuery, error)) ([]IndexQuery, error) {
+	queries, err := f(from, through)
```
bboreham: I can't see why `f()` is passed in as a closure.  Why not pass `queries` in?

pkg/chunk/schema_caching.go
```diff
@@ -90,24 +50,25 @@ func splitTimesByCacheability(from, through model.Time, cacheBefore model.Time)
...
-	for _, cq := range cacheableQueries {
-		for _, aq := range activeQueries {
-			// When deduping, the bucket values only influence TableName and HashValue
-			// and just checking those is enough.
-			if cq.TableName == aq.TableName && cq.HashValue == aq.HashValue {
-				continue Outer
-			}
+	// If the entire query is cacheable then cache it. We can't do partial query
+	// caching as we'd need bucket level info as to which bucket is immutable.
+	// And a previous attempt at it was buggy: https://github.com/cortexproject/cortex/issues/1698
```
bboreham: I don't think this amount of narrative is necessary.

pkg/chunk/schema_caching.go
```diff
@@ -90,24 +50,25 @@ func splitTimesByCacheability(from, through model.Time, cacheBefore model.Time)
...
-Outer:
-	for _, cq := range cacheableQueries {
-		for _, aq := range activeQueries {
-			// When deduping, the bucket values only influence TableName and HashValue
-			// and just checking those is enough.
-			if cq.TableName == aq.TableName && cq.HashValue == aq.HashValue {
-				continue Outer
-			}
+	// If the entire query is cacheable then cache it. We can't do partial query
+	// caching as we'd need bucket level info as to which bucket is immutable.
```
bboreham: This reasoning doesn't seem like a slam-dunk.  Why couldn't we require that all buckets before a certain time are immutable?

gouthamve: Hrm, Its hard from a separation of concerns perspective. Right now, we are using a `cachingSchema` which wraps a normal `Schema`, buckets are schema internal, and looking at things at a bucket level would mean we move the caching logic into normal Schema itself.

bboreham: Maybe the comment could say something like "we don't know the start/end time of a bucket" ?
(If I understood your point)

gouthamve: Done!

PR #1706, opened 2019-10-01
Refactor enocding.Chunk.Add

codesome:
Previously `encoding.Chunk.Add` used to return a slice of chunks including the chunk which called the `Add` method. This added a lot of unnecessary allocations to create new slices if no new chunk is created. This PR changes it to only return the new chunk if it is created, `nil` otherwise. There is no slice returned because there wont be more than 1 new chunk created for any single `Add` call.

Benchmark results for Ingester Push:
(This is with 10000 series)
```
benchmark                                        old ns/op      new ns/op      delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     2241738857     2112765505     -5.75%
BenchmarkIngesterPush/encoding=Varbit-8          2232805325     2102434453     -5.84%
BenchmarkIngesterPush/encoding=Bigchunk-8        2236451103     2084783678     -6.78%

benchmark                                        old allocs     new allocs     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     2783307        763322         -72.57%
BenchmarkIngesterPush/encoding=Varbit-8          2782899        762856         -72.59%
BenchmarkIngesterPush/encoding=Bigchunk-8        2782862        762844         -72.59%

benchmark                                        old bytes     new bytes     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     170320192     121377536     -28.74%
BenchmarkIngesterPush/encoding=Varbit-8          170238200     120911304     -28.98%
BenchmarkIngesterPush/encoding=Bigchunk-8        170210496     121198592     -28.79%
```

**Edit**
Benchmark results after reverting back to 100 series
```
benchmark                                        old ns/op     new ns/op     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     16431496      15754699      -4.12%
BenchmarkIngesterPush/encoding=Varbit-8          16342108      15671923      -4.10%
BenchmarkIngesterPush/encoding=Bigchunk-8        16409123      15624513      -4.78%

benchmark                                        old allocs     new allocs     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     32348          12139          -62.47%
BenchmarkIngesterPush/encoding=Varbit-8          32341          12138          -62.47%
BenchmarkIngesterPush/encoding=Bigchunk-8        32344          12137          -62.48%

benchmark                                        old bytes     new bytes     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     3324933       2768749       -16.73%
BenchmarkIngesterPush/encoding=Varbit-8          3341311       2764124       -17.27%
BenchmarkIngesterPush/encoding=Bigchunk-8        3338947       2769880       -17.04%
```


pkg/chunk/encoding/doubledelta.go
```diff
@@ -133,26 +139,28 @@ func (c doubleDeltaEncodedChunk) Add(s model.SamplePair) ([]Chunk, error) {
 		}
 	}
 	if tb != ntb || vb != nvb || c.isInt() != nInt {
-		if len(c)*2 < cap(c) {
-			return transcodeAndAdd(newDoubleDeltaEncodedChunk(ntb, nvb, nInt, cap(c)), &c, s)
-		}
+		// TODO(codesome): How to take care of this case?
+		// if len(c.b)*2 < cap(c.b) {
+		// 	return transcodeAndAdd(newDoubleDeltaEncodedChunk(ntb, nvb, nInt, cap(c.b)), &c, s)
```
codesome: I have commented out all the `transcodeAndAdd` for now as they return a slice of chunks. Given they are called within the same chunk type, can it be assumed that the single destination chunk can hold the existing chunk data and the sample (in short a slice of only 1 chunk, which looks like the case wherever it is called)?

pkg/chunk/encoding/chunk_test.go
```diff
@@ -28,26 +28,33 @@ import (
...
+			nc, err := c.Add(model.SamplePair{
 				Timestamp: model.Time(i),
 				Value:     model.SampleValue(i),
 			})
-			c = cs[0]
+			require.NoError(t, err)
+			if nc != nil {
+				chunks = append(chunks, nc)
+				c = nc
+			}
```
bboreham: This test seems to have changed into a completely different test, and I don't understand why.  Not that it was a great test before, but it seems really twisted now.
If you still want to change it, can you write a sentence or two describing what it is doing?

pkg/chunk/encoding/delta.go
```diff
@@ -76,21 +76,21 @@ func newDeltaEncodedChunk(tb, vb deltaBytes, isInt bool, length int) *deltaEncod
 }
 
 // Add implements chunk.
-func (c deltaEncodedChunk) Add(s model.SamplePair) ([]Chunk, error) {
+func (c *deltaEncodedChunk) Add(s model.SamplePair) (Chunk, error) {
```
bboreham: Alternative suggestion: delete this whole file.
It's a really old encoding from Prometheus, which nobody has a reason to use.

pkg/chunk/encoding/doubledelta.go
```diff
@@ -133,26 +135,43 @@ func (c doubleDeltaEncodedChunk) Add(s model.SamplePair) ([]Chunk, error) {
...
+				*c = *c0
+				return nil, nil
+			} else if len(result) == 2 {
+				// Replace the current chunk with the new bigger chunk
+				// and return the additional chunk.
+				c0 := result[0].(*doubleDeltaEncodedChunk)
+				c1 := result[1].(*doubleDeltaEncodedChunk)
+				*c = *c0
+				return c1, nil
+			}
```
bboreham: If this is an intentional fallthrough then change the comment to reflect.

pkg/distributor/distributor_test.go
```diff
@@ -576,14 +576,14 @@ func (i *mockIngester) QueryStream(ctx context.Context, req *client.QueryRequest
 
 		c := encoding.New()
 		for _, sample := range ts.Samples {
-			cs, err := c.Add(model.SamplePair{
+			// TODO(codesome): consider additional chunks.
```
bboreham: This comment doesn't mean much to me.

pkg/ingester/ingester_test.go
```diff
@@ -524,7 +525,7 @@ func BenchmarkIngesterPush(b *testing.B) {
 	limits := defaultLimitsTestConfig()
 
 	const (
-		series  = 100
+		series  = 10000
```
bboreham: This is testing a radically different mix of operations, and in particular very different from the default Prometheus config. I don't think it's justified.

codesome: I don't think I understand what you mean here

pkg/chunk/encoding/chunk.go
```diff
@@ -38,12 +38,10 @@ var (
 // goroutine-safe.
 type Chunk interface {
 	// Add adds a SamplePair to the chunks, performs any necessary
-	// re-encoding, and adds any necessary overflow chunks. It returns the
-	// new version of the original chunk, followed by overflow chunks, if
-	// any. The first chunk returned might be the same as the original one
-	// or a newly allocated version. In any case, take the returned chunk as
-	// the relevant one and discard the original chunk.
-	Add(sample model.SamplePair) ([]Chunk, error)
+	// re-encoding, and adds any necessary overflow chunks.
```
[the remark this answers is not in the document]
codesome: Also did s/adds/creates/

pkg/chunk/encoding/chunk_test.go
codesome: That was before I fixed my TODO for `transcodeAndAdd`. I brought back the test to the original state with some modifications for the new API.

pkg/chunk/encoding/delta.go
codesome: Done :)

pkg/chunk/encoding/doubledelta.go
codesome: Added the comment, both here and in `varbit.go`

pkg/distributor/distributor_test.go
codesome: That was a TODO for me which I forgot to fix. I have updated it now to handle any additional chunks.

pkg/ingester/ingester_test.go
bboreham: The setup/teardown costs are divided over 100,000 series instead of 100.

Put it another way: why did you make this change?

pkg/chunk/encoding/bigchunk_test.go
```diff
@@ -12,12 +12,11 @@ import (
 func TestSliceBiggerChunk(t *testing.T) {
 	var c Chunk = newBigchunk()
 	for i := 0; i < 12*3600/15; i++ {
-		cs, err := c.Add(model.SamplePair{
+		_, err := c.Add(model.SamplePair{
```
tomwilkie: Lets check this field is nil (ie add a `require.Nil(t, nc)`)

tomwilkie: Ditto in all the other places you've used `_`.

tomwilkie:
Can we have a changelog entry making it clear we're removing the DELTA encoding please?

pkg/ingester/ingester_test.go
codesome: I didn't think much before making the change to 10,000 series, I thought it would make the test more intensive. I have reverted back to 100 series anyway and have updated the benchmark results in the PR description

Here is the updated benchmark result:
```
benchmark                                        old ns/op     new ns/op     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     16431496      15754699      -4.12%
BenchmarkIngesterPush/encoding=Varbit-8          16342108      15671923      -4.10%
BenchmarkIngesterPush/encoding=Bigchunk-8        16409123      15624513      -4.78%

benchmark                                        old allocs     new allocs     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     32348          12139          -62.47%
BenchmarkIngesterPush/encoding=Varbit-8          32341          12138          -62.47%
BenchmarkIngesterPush/encoding=Bigchunk-8        32344          12137          -62.48%

benchmark                                        old bytes     new bytes     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     3324933       2768749       -16.73%
BenchmarkIngesterPush/encoding=Varbit-8          3341311       2764124       -17.27%
BenchmarkIngesterPush/encoding=Bigchunk-8        3338947       2769880       -17.04%
```

pkg/chunk/encoding/bigchunk_test.go
codesome: Done

pkg/chunk/encoding/factory.go
```diff
@@ -35,7 +35,8 @@ func (e Encoding) String() string {
 }
 
 const (
-	// Delta encoding
+	// Delta encoding is no longer supported and will be automatically changed to DoubleDelta.
```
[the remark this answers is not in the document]
codesome: I mistook this code to take care of it https://github.com/cortexproject/cortex/blob/857bb8476e596b28c784b9c01e4ced48c4f5f75a/pkg/chunk/chunk.go#L306-L310

I have now updated the flag initialization to take care of changing Delta to DoubleDelta. Thanks!

pkg/ingester/ingester_test.go
bboreham: Thanks! 

Go benchmarks are repeated enough times to make it run for more than 1 second, so you didn't make it more intensive overall.

bboreham:
Could you fix the changelog conflict please?

codesome:
Done. Also merged all commits into 1.

pkg/chunk/encoding/doubledelta.go
```diff
@@ -133,26 +135,47 @@ func (c doubleDeltaEncodedChunk) Add(s model.SamplePair) ([]Chunk, error) {
...
 		}
+
 		// Chunk is already half full. Better create a new one and save the transcoding efforts.
-		return addToOverflowChunk(&c, s)
+		// We also perform this if `transcodeAndAdd` resulted in >2 chunks.
+		return addToOverflowChunk(s)
```
bboreham: This bit seems to be repeated four times - can we move it inside `transcodeAndAdd()` ?

codesome: Given that there is typecast to specific types, some parts would again get repeated inside `transcodeAndAdd` for each type. Does that sound fine?

bboreham: OK leave it as it is.

bboreham:
Personally I don't believe there are any users on Delta encoding.  The default in Cortex was DoubleDelta. I think Delta is just a left-over from ancient Prometheus history - we copied all those files when Prometheus v2 moved to different encoding - and Cortex never used it.

codesome:
I had forgotten to do the cleanup after the latest changes. Did it in the last commit (after the approvals)

Issue #1672, opened 2019-09-17
proposal: Moving Caching part of query-frontend to separate project.


bwplotka:
Hi :wave: 

A month ago @tomwilkie [merged a PR](https://github.com/cortexproject/cortex/pull/1441) that makes `query-frontend` capable to cache responses for the queries against any Prometheus API. Details were presented at Prometheus London Meetup:
* Slides: https://speakerdeck.com/grafana/blazin-fast-promql 
* Watch the talk here: https://youtu.be/eyBbImSDOrI

Now, this is amazing piece of work as it allows simple and clear Cortex response caching (with days splitting!) to be used against any Prometheus-based backend. Requests against metric backends are often expensive, have small result output, are simultaneous and repetitive, so it makes sense to treat such caching component as must-have - even for vanilla Prometheus. As the [Thanos](http://thanos.io) maintainers we were looking exactly for something like [this](https://github.com/thanos-io/thanos/issues/1006) for some time. Overall it definitely looks like both Cortex and Thanos are looking to solve a very similar goal. 

**From Thanos side we want to make it a default caching solution that we want to recommend, document and maintain.**

However, still, such caching is heavily bound to Cortex. It has quite a complex Queuing engine that already [was proposed to be extracted from caching](https://github.com/cortexproject/cortex/issues/1150). I believe that splitting caching into a separate project (`promcache`?), in some common org like https://github.com/prometheus-community can have many advantages around contributing, clarity and adoption. I enumerated some benefits further down.

## Proposal

1. Move `query-frontend` caching logic to separate Go module (plus cmd to run it) e.g  https://github.com/prometheus-community/promcache
   * Name of the project is to be defined ( :
2. Add maintainers who want to help from both Cortex and Thanos as the project owners. 
3. Make it clear that this is a caching project for Prometheus API, Cortex, and Thanos backends.
    * Open questions: 
         * What if other backends want something extra? VM, M3DB?
         * Should we embed retries and limits as well? (IMO yes)
4. Allow Cortex to use it either as a library in `query-frontend` or just point to `query-frontend` (without caching)
5. Allow Thanos to use it as a library in Querier (potentially) or spin up on top of Querier (must-have)

If we agree on this, we (Thanos team) are happy to spin this project up, prepare repo, go module, initial docs and extract caching logic from query-frontend. Then we can focus on embedded caching in existing components like Querier or Query-frontend and use `promcache` as a library if needed.

[...]

PR #1711, opened 2019-10-02
Refactor: Separate Frontend and Queryrange middleware

cyriltovena:
This separates any `queryrange` package logic from the `frontend` package. Interaction between the two packages is made via a tripperware chain. This makes it easier to implement a totally different tripperware while still being able to use the GRPC/HTTP frontend.

> Note for reviewers the configuration has changed !

I wanted to make the SplitByDay works with `Duration` but it seems that the caching middleware is hardcoded to also cache by day.

This should help #1672 as you can now use queryrange middlewares with your own roundtripper if you don't want to use the frontend or you can use your own middleware if you want to use the frontend.

I've also cleaned up the protobuf, I believe some properties were not used.

/cc @tomwilkie @bwplotka


cyriltovena:
Not sure what's going on with the lint:

```
#!/bin/bash -eo pipefail
make BUILD_IN_CONTAINER=false lint
./tools/lint -notestpackage -novet -ignorespelling queriers -ignorespelling Queriers .
./pkg/ingester/client/timeseries.go: higher than 25 cyclomatic complexity - 32 client (*LabelAdapter).Unmarshal ./pkg/ingester/client/timeseries.go:99:1
./pkg/cortex/modules.go: run gofmt -s -w ./pkg/cortex/modules.go
./pkg/chunk/encoding/delta.go: higher than 25 cyclomatic complexity - 30 encoding (deltaEncodedChunk).Add ./pkg/chunk/encoding/delta.go:79:1
./pkg/chunk/encoding/doubledelta.go: higher than 25 cyclomatic complexity - 30 encoding (doubleDeltaEncodedChunk).Add ./pkg/chunk/encoding/doubledelta.go:87:1
Makefile:132: recipe for target 'lint' failed
```

Locally it tries to lint all vendor files.

Issue #1708, opened 2019-10-02
Upskill the consul watch code for best practices


tomwilkie:
Hashicorp have some nice advice in the docs: https://www.consul.io/api/features/blocking.html#implementation-details

- Check for the index going backward
- Check the index >0
- Ratelimit watches.

gouthamve:
Closed in #1713

PR #1713, opened 2019-10-03
Implement Hashicorps's recommendations when watching consul key

pstibrany:
This PR implements Hashicorp's recommendations for using blocking queries at https://www.consul.io/api/features/blocking.html:

- rate limiting when watching for value changes. Current limit is 1 read/second, with bursts up to 5/second.
- handle LastIndex value going down or being zero by restarting the blocking loop from 0 or 1 respectively

Fixes issue #1708.


pstibrany:
Thanks for your review.

I couldn't find any other fitting rate limiter in dependencies that Cortex already has, and I didn't want to introduce new dependency. I'm open to switching to a different one if you have a good suggestion, although I quite like how simple this one is.

tomwilkie:
@pstibrany have you seen https://godoc.org/golang.org/x/time/rate?

pstibrany:
Replaced with rate limiter from golang.org/x/time/rate. Not as nice, but it works and we avoid copying.

pkg/ring/kv/consul/client.go
```diff
@@ -31,6 +32,9 @@ var (
 		MinBackoff: 1 * time.Second,
 		MaxBackoff: 1 * time.Minute,
 	}
+
+	watchKeyRate  = rate.Limit(1)
+	watchKeyBurst = 3
```
tomwilkie: Can we have these as part of the Config struct below?  And have a way for the default values to be a no-op?  I want this to be opt-in.

pstibrany:
Oh, I guess rate-limit should apply for WatchPrefix as well, not just WatchKey. I missed that.

pkg/ring/kv/consul/client.go
pstibrany: Moved WatchKeyRate and WatchKeyBurst to Config. Default value to WatchKeyRate (0) disables rate-limiting. Burst is set to 1 if rate-limiting is enabled and Burst is <= 1.

pkg/ring/kv/consul/client.go
```diff
@@ -264,3 +289,32 @@ func (c *Client) Get(ctx context.Context, key string) (interface{}, error) {
...
+		// guaranteed to be the same as last time
+		return metaLastIndex, true
+	} else {
+		return metaLastIndex, false
+	}
+}
+
+func (c *Client) createRateLimiter() *rate.Limiter {
+	if c.cfg.WatchKeyRate <= 0 {
+		return nil
```
gouthamve: Just return `rate.Inf`, this way we won't have the `nil` and code will be cleaner?

pkg/ring/kv/consul/client_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return []byte(s), nil
+}
+
+func (c stringCodec) Decode(d []byte) (interface{}, error) {
+	return string(d), nil
+}
+
+var _ codec.Codec = &stringCodec{}
```
gouthamve: What is this doing?

pkg/ring/kv/consul/client.go
```diff
@@ -62,6 +65,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.StringVar(&cfg.ACLToken, prefix+"consul.acltoken", "", "ACL Token used to interact with Consul.")
 	f.DurationVar(&cfg.HTTPClientTimeout, prefix+"consul.client-timeout", 2*longPollDuration, "HTTP timeout when talking to Consul")
 	f.BoolVar(&cfg.ConsistentReads, prefix+"consul.consistent-reads", true, "Enable consistent reads to Consul.")
+	f.Float64Var(&cfg.WatchKeyRate, prefix+"consul.watch-rate", 0, "Rate limit when watching key or prefix in Consul, in requests per second. Zero disables the rate limit.")
```
pracucci: Have you considered naming it `consul.watch-rate-limit`?

pkg/ring/kv/consul/client.go
pstibrany: Done

pkg/ring/kv/consul/client_test.go
pstibrany: By adding that line, IntelliJ then offers me to add missing methods from the interface. I forgot to remove it afterwards.

pkg/ring/kv/consul/client.go
```diff
@@ -264,3 +289,32 @@ func (c *Client) Get(ctx context.Context, key string) (interface{}, error) {
...
+		return metaLastIndex, false
+	}
+}
+
+func (c *Client) createRateLimiter() *rate.Limiter {
+	if c.cfg.WatchKeyRate <= 0 {
+		return nil
+	}
+	burst := c.cfg.WatchKeyBurst
+	if burst < 1 {
```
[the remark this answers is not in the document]
pstibrany: Oh, I was about to add this check. I think it makes sense, but you seem to have changed your mind?

pkg/ring/kv/consul/client.go
pstibrany: Sounds good. Thanks.

pkg/ring/kv/consul/client.go
```diff
@@ -62,6 +65,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.StringVar(&cfg.ACLToken, prefix+"consul.acltoken", "", "ACL Token used to interact with Consul.")
 	f.DurationVar(&cfg.HTTPClientTimeout, prefix+"consul.client-timeout", 2*longPollDuration, "HTTP timeout when talking to Consul")
 	f.BoolVar(&cfg.ConsistentReads, prefix+"consul.consistent-reads", true, "Enable consistent reads to Consul.")
+	f.Float64Var(&cfg.WatchKeyRate, prefix+"consul.watch-rate", 0, "Rate limit when watching key or prefix in Consul, in requests per second. Zero disables the rate limit.")
+	f.IntVar(&cfg.WatchKeyBurst, prefix+"consul.watch-burst", 1, "Burst size used in rate limit. Values less than 1 are treated as 1.")
```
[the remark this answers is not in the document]
pstibrany: I'd use `watch-burst-size` maybe?

pkg/ring/kv/consul/client.go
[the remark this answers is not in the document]
pracucci: I've rechecked the rate limiter implementation and it's not an issue as far as the burst is at least 1, because here you always consume 1 token each time.

pkg/ring/kv/consul/client.go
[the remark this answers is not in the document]
pracucci: Technically it's the max number of tokens you can consume in a single request. Given here you always consume 1, I'm wondering if it it's really useful to allow people to configure it, instead of simply hardcoding it to `1`.

pkg/ring/kv/consul/client.go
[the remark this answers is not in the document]
pracucci: What I mean is that we always use `limiter.Wait()` here, which internally calls `WaitN()` with `n = 1` (`n` is the number of tokens to reserve), while `burst` is the max number of tokens you can consume in a single operation, which in our case is always `1`, so an hard-coded `1` value should be fine (if I'm not missing anything).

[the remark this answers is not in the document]
pstibrany: Burst being the max number of tokens you can consume is not the only reason to use burst. Another is to temporarily raise the limit, if limiter wasn't used recently (or at the beginning of the operation). See here: https://play.golang.org/p/zBtoOUf5fSl ... notice how first 5 operations are allowed immediately, and only subsequent operations are limited to 1 per second.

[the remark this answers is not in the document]
pracucci: You're right.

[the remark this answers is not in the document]
pstibrany: On the other hand, you're right in that checking that burst is higher than limit doesn't help much – if you use limit 5, and burst 1, client using Wait() is still allowed 5 operations / second. So all is fine here. I will resolve it now.

pkg/ring/kv/consul/client.go
```diff
@@ -62,6 +65,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.StringVar(&cfg.ACLToken, prefix+"consul.acltoken", "", "ACL Token used to interact with Consul.")
 	f.DurationVar(&cfg.HTTPClientTimeout, prefix+"consul.client-timeout", 2*longPollDuration, "HTTP timeout when talking to Consul")
 	f.BoolVar(&cfg.ConsistentReads, prefix+"consul.consistent-reads", true, "Enable consistent reads to Consul.")
+	f.Float64Var(&cfg.WatchKeyRateLimit, prefix+"consul.watch-rate-limit", 0, "Rate limit when watching key or prefix in Consul, in requests per second. Zero disables the rate limit.")
```
jtlisi:
```suggestion
	f.Float64Var(&cfg.WatchKeyRateLimit, prefix+"consul.watch-rate-limit", 0, "Rate limit when watching key or prefix in Consul, in requests per second. 0 disables the rate limit.")
```

Just a slight nit here

gouthamve:
@pstibrany Could you add a changelog entry and I'll merge this after that and Jacob's nit.

pstibrany:
Added changelog, fixed "Zero" -> "0", and rebased on top of master.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ This release adds support for Redis as an alternative to Memcached, and also inc
 * [ENHANCEMENT] Upgraded Prometheus to 2.12.0 and Alertmanager to 0.19.0. #1597
 * [ENHANCEMENT] Cortex is now built with Go 1.13 #1675, #1676, #1679
 * [ENHANCEMENT] Many optimisations, mostly impacting ingester and querier: #1574, #1624, #1638, #1644, #1649, #1654, #1702
+* [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
```
gouthamve: Th updates should go into unreleased and not 0.3.0 :)

PR #1714, opened 2019-10-04
Validate schema config to ensure table period is a multiple of buckets period

pracucci:
The `schema_config` uses `dailyBuckets()` (starting from `v2`) to generate the buckets where indexes/chunks should be stored. The bucket also contains `tableName`, so it effectively drives to which table data is read//written.

Due to how `dailyBuckets()` is implemented, it constructs the `tableName` as function of the day bucket (with time `00:00:00`). Because of this, reads/writes occurs on index/chunk tables whose name is constructed with a daily-approximated numbering.

If the table period is not a multiple of `24h`, the table manager will create/update/delete tables with a naming not matching the real one (generated by `dailyBuckets()`).

After a [brief discussion with Brian on Slack](https://cloud-native.slack.com/archives/CCYDASBLP/p1570008064266600), in this PR I'm proposing to introduce a validation which prevents cortex to start if the table period is not a multiple of the buckets period.

### DIFF
--- a/pkg/chunk/schema_config.go
+++ b/pkg/chunk/schema_config.go
@@ -1,6 +1,7 @@
 package chunk
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -21,6 +22,11 @@ const (
 	secondsInDay       = int64(24 * time.Hour / time.Second)
 	millisecondsInHour = int64(time.Hour / time.Millisecond)
 	millisecondsInDay  = int64(24 * time.Hour / time.Millisecond)
+)
+
+var (
+	errInvalidSchemaVersion = errors.New("invalid schema version")
+	errInvalidTablePeriod   = errors.New("the table period must be a multiple of 24h (1h for schema v1)")
 )
 
 // PeriodConfig defines the schema and tables to use for a period of time
@@ -116,8 +122,7 @@ func (cfg *LegacySchemaConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.ChunkTables.RegisterFlags("dynamodb.chunk-table", "cortex_chunks_", f)
 }
 
-// translate from command-line parameters into new config data structure
-func (cfg *SchemaConfig) translate() error {
+func (cfg *SchemaConfig) loadFromFlags() error {
 	cfg.Configs = []PeriodConfig{}
 
 	add := func(t string, f model.Time) {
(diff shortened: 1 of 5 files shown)


khaines:
Thanks, @pracucci for the work on this! This is very needed as more people use cortex and are unaware of the dependency dailyBuckets() has on the period tables

pkg/cortex/cortex.go
```diff
@@ -108,6 +108,17 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	flag.IntVar(&chunk_util.QueryParallelism, "querier.query-parallelism", 100, "Max subqueries run in parallel per higher-level query.")
 }
 
+// Validate the cortex config and returns an error if the validation
+// doesn't pass
+func (c *Config) Validate() error {
+	err := c.Schema.Validate()
```
bboreham: I can't see how `Load()` is called before we get here - won't it just be blank?

pracucci: To my understanding, there are three ways to configure the schema config in Cortex:
1. Everything passed as CLI flag
2. Everything passed as CLI flag, but schema config as file via `-config-yaml`
3. The whole config is a yaml via `-config.file`

In the case 3, the schema config is already loaded at this stage and we should validate it here, because the `SchemaConfig.Load()` (called later) will be a no-op (it skips if `len(cfg.Configs) > 0`).

Am I missing anything?

pkg/chunk/schema_config.go
```diff
@@ -116,8 +122,8 @@ func (cfg *LegacySchemaConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.ChunkTables.RegisterFlags("dynamodb.chunk-table", "cortex_chunks_", f)
 }
 
-// translate from command-line parameters into new config data structure
-func (cfg *SchemaConfig) translate() error {
+// loadFromFlags loads the schema config from command-line flags
```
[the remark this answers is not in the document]
pracucci: You're right. Removed it.

pkg/cortex/cortex.go
bboreham: I didn’t even know it was possible to put the schema in the same file, ie case 3. 
What happens in case 1? Does that deserve a comment?

pracucci: > What happens in case 1?

In case 1, all flags are parsed by `flag.Parse()` (`main.go`) and the schema config validation is triggered by `SchemaConfig.Load()`.

However, you comment made me realize that my change set doesn't trigger the overall config validation in case of config passed as CLI flags in all cases, so I've pushed a change to fix it (_shouldn't be an issue with the current change set, cause we validate only the schema config, but this PR should be considered the scaffolding for further validations, it's better to have it right first_).

bboreham: Ah, the other thing I was confused about is that it isn't possible to set the period from the CLI, so case 1 doesn't need checking.

PR #1719, opened 2019-10-07
In Distribtor, pre-allocate buffer for reading protobufs

bboreham:
ReadAll() will start at 512 bytes then successively double the buffer, copying the contents each time. If the caller supplies a size we can be much more efficient. It's not a massive win, but every little helps.

(However, doing this creates a new DoS vector - supplying an fake size that is way too big. I'm thinking we should have a max input size, similar to gRPC)

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -125,6 +125,7 @@ type Config struct {
 
 	HATrackerConfig HATrackerConfig `yaml:"ha_tracker,omitempty"`
 
+	MaxRecvMsgSize      int           `yaml:"max_send_msg_size"`
 	RemoteTimeout       time.Duration `yaml:"remote_timeout,omitempty"`
 	ExtraQueryDelay     time.Duration `yaml:"extra_queue_delay,omitempty"`
 	LimiterReloadPeriod time.Duration `yaml:"limiter_reload_period,omitempty"`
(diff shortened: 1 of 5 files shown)


tomwilkie:
>  I'm thinking we should have a max input size, similar to gRPC

+1

Do we have a benchmark that shows an improvement?

bboreham:
Nope, no benchmark.  In my prod distributors this `ReadAll()` is 5.5% of all garbage, and this change brings that down over half, so expect a 3% improvement overall?

bboreham:
This is the compressed data - if we could use a pool of buffers for the uncompressed bytes, that would be much bigger impact, but I'm not sure whether `Unmarshal()` retains any of the memory.
(and it gives the buffer to the "billing" code but I want to remove that anyway)

khaines:
>  I'm thinking we should have a max input size, similar to gRPC

also +1 on this.

bboreham:
I added a size limit.

pkg/distributor/distributor.go
```diff
@@ -125,6 +125,7 @@ type Config struct {
 
 	HATrackerConfig HATrackerConfig `yaml:"ha_tracker,omitempty"`
 
+	MaxRecvMsgSize      int           `yaml:"max_send_msg_size"`
```
pracucci: The yaml here is `max_send_msg_size`, while the CLI flag is `max-recv-msg-size`. Isn't contradictory?

gouthamve: Good catch, could you open a PR/issue so that we don't lose track of this?

bboreham: oops, sorry!

Issue #771, opened 2018-03-26
RFC: Tools to block and/or drop data which is bad


bboreham:
Suppose a user suddenly starts sending thousands of samples where the labels change each time: the ingester will create new series for each of them and quickly grow in memory. (cf #665)

Many installations of Cortex will have an upstream component to validate users, so we perhaps don't need to block inside Cortex, but we may want to have a button to erase all series from the flush queue for a particular user.

bboreham:
Another variant: drop all series for a particular metric as they arrive, because it has been observed to be very high-cardinality.

gouthamve:
I think this is fixed with the `-ingester.min-chunk-length`

bboreham:
Some aspects are improved, but this feature is to block before the data gets into ingesters.

gouthamve:
Closing this in favour of #1507

PR #1726, opened 2019-10-09
Add a limits config option to allow for dropping of the cluster label

### LINKED ISSUE #1724: Add ability to drop cluster label after dedupe
A couple of users have asked for this; as we need a cluster label for deduping in the distributors, some people don't have this and adding it will break their graphs.

By "break their graphs", they will cause things like discontinuations in things like recording rules, which for LTS systems is undesirable.

@cstyan is this something we can make an option?

cstyan:
Fixes #1724 

Adds an optional flag per user to drop the cluster label, in addition to dropping the replica label, when ingesting HA samples. Should only be used if that user has their own set of more than two labels that they use to uniquely identify Prometheus replicas.

There's a bit of a refactor here as well:
- I noticed that in the config the ordering of the values was cluster label then replica label, but in RegisterFlags the order is replica then cluster. I moved them around in RegisterFlags to match the order in the struct.
- `distributor.Push` was getting pretty long, I moved the internals of the validation loop into a function `validateSeries`. This also had the side effect of making that portion of the code testable. I've only added tests for the label removal bits. I assume the actual validation functions `ValidateLabels` and `ValidateSamples` are tested already.
- renamed removeReplicaLabel -> removeLabel, it wasn't doing anything replica label specific anyways

cc @tomwilkie @gouthamve

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -144,11 +144,13 @@ prefix these flags with `distributor.ha-tracker.`
 
 ### HA Tracker
 
-HA tracking has two of it's own flags:
+HA tracking has three of its own flags:
 - `distributor.ha-tracker.cluster`
    Prometheus label to look for in samples to identify a Prometheus HA cluster. (default "cluster")
 - `distributor.ha-tracker.replica`
    Prometheus label to look for in samples to identify a Prometheus HA replica. (default "`__replica__`")
+- `distributor.ha-tracker.drop-cluster-label`
+   Drops the cluster label before ingesting samples. The replica label is always dropped. This flag should only be used for users who already have some combination of more than two labels to uniquely identify a Prometheus replica, but for whom it is not feasible to switch over to a new set of just two labels.
 
 It's reasonable to assume people probably already have a `cluster` label, or something similar. If not, they should add one along with `__replica__` via external labels in their Prometheus config. If you stick to these default values your Prometheus config could look like this (`POD_NAME` is an environment variable which must be set by you):
 
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -108,6 +108,9 @@ The ingester query API was improved over time, but defaults to the old behaviour
 - `distributor.ha-tracker.enable` 
    Enable the distributors HA tracker so that it can accept samples from Prometheus HA replicas gracefully (requires labels). Global (for distributors), this ensures that the necessary internal data structures for the HA handling are created. The option `enable-for-all-users` is still needed to enable ingestion of HA samples for all users.
 
+- `distributor.drop-label`
+   This flag can be used to specify label names that to drop during sample ingestion within the distributor and can be repeated in order to drop multiple labels.
+
 ### Ring/HA Tracker Store
 
 The KVStore client is used by both the Ring and HA Tracker.
@@ -144,13 +147,11 @@ prefix these flags with `distributor.ha-tracker.`
 
 ### HA Tracker
 
-HA tracking has three of its own flags:
+HA tracking has two of its own flags:
 - `distributor.ha-tracker.cluster`
    Prometheus label to look for in samples to identify a Prometheus HA cluster. (default "cluster")
 - `distributor.ha-tracker.replica`
    Prometheus label to look for in samples to identify a Prometheus HA replica. (default "`__replica__`")
-- `distributor.ha-tracker.drop-cluster-label`
-   Drops the cluster label before ingesting samples. The replica label is always dropped. This flag should only be used for users who already have some combination of more than two labels to uniquely identify a Prometheus replica, but for whom it is not feasible to switch over to a new set of just two labels.
 
 It's reasonable to assume people probably already have a `cluster` label, or something similar. If not, they should add one along with `__replica__` via external labels in their Prometheus config. If you stick to these default values your Prometheus config could look like this (`POD_NAME` is an environment variable which must be set by you):
 
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -96,6 +96,7 @@ var (
 		Name:      "distributor_replication_factor",
 		Help:      "The configured replication factor.",
 	})
+	emptyPreallocSeries = ingester_client.PreallocTimeseries{}
 )
 
 // Distributor is a storage.SampleAppender and a client.Querier which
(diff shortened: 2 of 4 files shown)


docs/arguments.md
```diff
@@ -144,11 +144,13 @@ prefix these flags with `distributor.ha-tracker.`
 
 ### HA Tracker
 
-HA tracking has two of it's own flags:
+HA tracking has three of it's own flags:
```
bboreham:
```suggestion
HA tracking has three of its own flags:
```

bboreham:
I find it easier to review PRs like this when you split the pure-refactoring part from the behaviour-changing part, into separate commits. Otherwise I have to figure out your intention on every line and that takes longer.

pkg/distributor/distributor.go
```diff
@@ -280,14 +281,55 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
 	return true, nil
 }
 
+// Validates a single series from a write request. Will remove HA labels if necessary.
```
bboreham: State what the return values are.
Include why we have return values both via pointer and by value.

pkg/util/validation/limits.go
```diff
@@ -48,8 +49,9 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
...
 	f.BoolVar(&l.AcceptHASamples, "distributor.ha-tracker.enable-for-all-users", false, "Flag to enable, for all users, handling of samples with external labels identifying replicas in an HA Prometheus setup.")
-	f.StringVar(&l.HAReplicaLabel, "distributor.ha-tracker.replica", "__replica__", "Prometheus label to look for in samples to identify a Prometheus HA replica.")
 	f.StringVar(&l.HAClusterLabel, "distributor.ha-tracker.cluster", "cluster", "Prometheus label to look for in samples to identify a Prometheus HA cluster.")
+	f.StringVar(&l.HAReplicaLabel, "distributor.ha-tracker.replica", "__replica__", "Prometheus label to look for in samples to identify a Prometheus HA replica.")
+	f.BoolVar(&l.HADropClusterLabel, "distributor.ha-tracker.drop-cluster-label", false, "Enable this flag to drop the cluster label when ingesting HA samples for a user, in addition to dropping the replica label. Note that this flag should only be used when users already have their own combination of more than two labels to uniquely identify Prometheus replicas.")
```
bboreham: The 'when' part doesn't appear to be true - it is done unconditionally.

cstyan: fixed

pkg/distributor/distributor.go
cstyan: Done. I've simplified the function as well so the pointer isn't needed anymore.

pkg/distributor/distributor.go
```diff
@@ -249,11 +250,11 @@ func shardByAllLabels(userID string, labels []client.LabelAdapter) (uint32, erro
 	return h, nil
 }
 
-// Remove the replica label from a slice of LabelPairs if it exists.
-func removeReplicaLabel(replica string, labels *[]client.LabelAdapter) {
+// Remove the label labelname from a slice of LabelPairs if it exists.
+func removeLabel(labelName string, labels *[]client.LabelAdapter) {
```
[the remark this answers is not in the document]
cstyan: I'll have a look at the labelsPairs function in the morning.

cstyan:
@bboreham thanks for the review! I've split out the refactoring into another commit. I also changed the `validateSeries` function to no longer require the error pointer argument in addition to returning an error.

bboreham:
Given that it isn't tied to HA, could the feature be done as a general label-removal mechanism? This would give us a portion of #771.

cstyan:
I moved the cluster label dropping inside the check for `removeReplica`, which is tied to HA. I can move that back out and turn the label removal into a more general non-HA dependent thing, maybe in another PR? if you'd like.

bboreham:
Given that it doesn't cost me anything, yes I would prefer a generic "for tenant X, everywhere you see label Y, drop it".

nbellowe:
@cstyan Any updates with this PR? Its much appreciated, we're waiting on this to enable HA writing to cortex.

docs/arguments.md
```diff
@@ -144,11 +144,13 @@ prefix these flags with `distributor.ha-tracker.`
 
 ### HA Tracker
 
-HA tracking has two of it's own flags:
+HA tracking has three of its own flags:
 - `distributor.ha-tracker.cluster`
    Prometheus label to look for in samples to identify a Prometheus HA cluster. (default "cluster")
 - `distributor.ha-tracker.replica`
    Prometheus label to look for in samples to identify a Prometheus HA replica. (default "`__replica__`")
+- `distributor.ha-tracker.drop-cluster-label`
```
bboreham: does this name need to be updated?

pkg/util/validation/limits.go
```diff
@@ -48,8 +51,9 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&l.IngestionRate, "distributor.ingestion-rate-limit", 25000, "Per-user ingestion rate limit in samples per second.")
 	f.IntVar(&l.IngestionBurstSize, "distributor.ingestion-burst-size", 50000, "Per-user allowed ingestion burst size (in number of samples). Warning, very high limits will be reset every -distributor.limiter-reload-period.")
 	f.BoolVar(&l.AcceptHASamples, "distributor.ha-tracker.enable-for-all-users", false, "Flag to enable, for all users, handling of samples with external labels identifying replicas in an HA Prometheus setup.")
-	f.StringVar(&l.HAReplicaLabel, "distributor.ha-tracker.replica", "__replica__", "Prometheus label to look for in samples to identify a Prometheus HA replica.")
 	f.StringVar(&l.HAClusterLabel, "distributor.ha-tracker.cluster", "cluster", "Prometheus label to look for in samples to identify a Prometheus HA cluster.")
+	f.StringVar(&l.HAReplicaLabel, "distributor.ha-tracker.replica", "__replica__", "Prometheus label to look for in samples to identify a Prometheus HA replica.")
```
bboreham: It's not obvious to me why this moved.  No big deal.

docs/arguments.md
cstyan: Yes, thanks for catching that.

pkg/util/validation/limits.go
cstyan: Sorry for the noise. I just rearranged the calls to `flags.VarType` here for the cluster and replica labels to match the order those values are specified in the Limits struct.

pkg/distributor/distributor.go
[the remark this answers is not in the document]
cstyan: `labelPairs` is nested within the `ingester` package. If we're going to merge the two functions should that exist in a new package under `util`?

And the reason for preserving order is only that I've generally seen that be the case within Prometheus itself. For example within remote write, when processing a sample through external labelse and relabel rules, we preserve the sorted order of the labels.

pkg/distributor/distributor.go
```diff
@@ -280,6 +281,48 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
 	return true, nil
 }
 
+// Validates a single series from a write request. Will remove HA labels if necessary.
+// Takes a pointer for a partial error so that we can get partial errors, errors during validation
```
gouthamve: Can you update the comment to remove the reference to the pointer?

pkg/util/validation/limits.go
```diff
@@ -142,14 +146,19 @@ func (o *Overrides) AcceptHASamples(userID string) bool {
...
+
 // HAReplicaLabel returns the replica label to look for when deciding whether to accept a sample from a Prometheus HA replica.
 func (o *Overrides) HAReplicaLabel(userID string) string {
 	return o.overridesManager.GetLimits(userID).(*Limits).HAReplicaLabel
 }
 
-// HAClusterLabel returns the cluster label to look for when deciding whether to accept a sample from a Prometheus HA replica.
-func (o *Overrides) HAClusterLabel(userID string) string {
-	return o.overridesManager.GetLimits(userID).(*Limits).HAClusterLabel
+// DropLabels returns whether the cluster label should be dropped when ingesting HA samples for the user.
```
gouthamve: Returns the list of labels to be dropped?

pkg/distributor/distributor.go
```diff
@@ -280,6 +281,48 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
+// of a single sample, back from the function without adding an additional return param.
+// Returns a token for the series if it is valid, the validated series with it's labels/samples, and any fatal error.
+func (d *Distributor) validateSeries(key uint32, ts ingester_client.PreallocTimeseries, userID string, removeReplica bool) (client.PreallocTimeseries, error) {
+	// If we found both the cluster and replica labels, we only want to include the cluster label when
+	// storing series in Cortex. If we kept the replica label we would end up with another series for the same
+	// series we're trying to dedupe when HA tracking moves over to a different replica.
+	if removeReplica {
+		removeLabel(d.limits.HAReplicaLabel(userID), &ts.Labels)
+	}
+	for _, s := range d.limits.DropLabels(userID) {
```
gouthamve: `s` is a weird variable name in this instance, maybe call it `label`/`labelName`?

pkg/distributor/distributor.go
```diff
@@ -280,6 +281,48 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
+		removeLabel(s, &ts.Labels)
+	}
+	key, err := d.tokenForLabels(userID, ts.Labels)
```
gouthamve: Also check and handle if the series has no labels anymore after the removing of labels.

nbellowe:
Thank you @cstyan!

PR #1728, opened 2019-10-10
Remove `dynamo` as a valid storage option

rsteneteg:
Signed-off-by: Roger Steneteg <rsteneteg@ea.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [CHANGE] Remove direct DB/API access from the ruler
 * [CHANGE] Removed `Delta` encoding. Any old chunks with `Delta` encoding cannot be read anymore. If `ingester.chunk-encoding` is set to `Delta` the ingester will fail to start. #1706
 * [CHANGE] Setting `-ingester.max-transfer-retries` to 0 now disables hand-over when ingester is shutting down. Previously, zero meant infinite number of attempts. #1771
+* [CHANGE] `dynamo` has been removed as a valid storage name to make it consistent for all components. `aws` and `aws-dynamo` remain as valid storage names. 
 * [FEATURE] Global limit on the max series per user and metric #1760
   * `-ingester.max-global-series-per-user`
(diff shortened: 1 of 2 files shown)


rsteneteg:
This allows the usage of `store: dynamo` which otherwise would work for the NewIndexClient, but would fail for table-manager when calling the NewTableClient.

bboreham:
Another way to be consistent would be to remove "dynamo", if it never actually worked.
"aws-dynamo" is identical, as far as I can see.
@tomwilkie do you recall why "dynamo" was half-added in #1120 ?

gouthamve:
:+1: to removing `dynamo` if `aws-dynamo` is actually the same thing. @rsteneteg Would you be able to do the change?

rsteneteg:
@gouthamve @bboreham My only concern with removing `dynamo` is that it may break users current configuration.

gouthamve:
We could make a clear note in the changelog? This is a relatively minor
change (easy to fix) and something that is super noticable (rollouts will
fail).

rsteneteg:
For chunks the ObjectClient accepts `aws` and `s3` for S3 storage, since there is no `aws-s3` and for bigtable the option is called just `bigtable` not `gcp-bigtable`, so I think it makes more sense to keep the `dynamo` option over removing it.

bboreham:
The case you are trying to fix is where someone has configured Querier or Ingester with `dynamo` and Tablemanager with `aws-dynamo`, and you want to let them say `dynamo` everywhere ?

I just don't like adding options which aren't necessary.

rsteneteg:
ok I have have updated the PR to remove the `dynamo` option.

gouthamve:
@rsteneteg Could you update the changelog with this change? It might break some users and is best called out.

rsteneteg:
Ok, have updated the change log as well now, the reason I did not add it before is that the breaking change where `dynamo` was no longer supported by table manager had already been done.

PR #1730, opened 2019-10-11
Executing child queries on queriers

owen-d:
## Disclaimer
This is not finished, but I'd like feedback on the design.

## What
These changes aim to introduce a path towards further distributing queries. Currently Cortex dispatches queries to backend queriers, but as the throughput of metrics increase, running entire queries on a single querier can become a bottleneck.

#### Sharding
The v10 schema introduced a shard factor for data which spreads series across n shards. This is a prerequisite for allowing us to query data from these shards in parallel.

### Problem
Although the v10 schema change introduces a shard factor, all shards must still be processed on a single querier. This is compounded by the fact that query ASTs are not broken down before execution. Therefore, while sharding lets us split up data at it's source location, we still re-aggregate it and process it all in one place. The two goals of this PR are to fix these, by

1) splitting up child ASTs for individual execution
2) Splitting up selectors into n separate selectors where n = shard factor

### Mapping ASTs
Firstly, we introduce a package called `astmapper` with the interface:

```
// ASTMapper is the exported interface for mapping between multiple AST representations
type ASTMapper interface {
	Map(node promql.Node) (promql.Node, error)
}
```

This interface is used to map ASTs into different ASTs. We use this to turn i.e.

`sum by (foo) (rate(bar{baz=”blip”}[1m]))` ->

```
sum by (foo) (
  sum by (foo) (rate(bar{baz=”blip”,__cortex_shard__=”0of16”}[1m])) or
  sum by (foo) (rate(bar{baz=”blip”,__cortex_shard__=”1of16”}[1m])) or
  ...
  sum by (foo) (rate(bar{baz=”blip”,__cortex_shard__=”15of16”}[1m]))
)
```

This works largely because summing the sub-sums is equivalent. The principle should hold true for any merge operation that is associative, but sums are the initial focus.

### Hijacking Queryable + embedding queries
Queries are executed in unison by a `promql.Engine` and a `storage.Queryable`. Since the Engine is a concrete type, it's implementation is locked. This means that the remaining option is to hijack the queryable to dispatch child queries. However, the queryable is only called to retrieve vector and matrix selector nodes -- all aggregations, functions, etc are handled by the Engine itself. Therefore, in order to regain control of these entire subtrees, we must encode them in vector/matrix selectors. This is done by stringifying an entire subtree and replacing the node with a vector or matrix selector. Currently queries are hex-encoded, but this could be something more human-friendly.

Using our previous example with a shard-factor of 3, `sum by(foo) (rate(bar1{baz="blip"}[1m]))` is turned into 
```
sum by(foo) (
  __embedded_query__{__cortex_query__="73756d20627928666f6f2920287261746528626172317b5f5f636f727465785f73686172645f5f3d22305f6f665f33222c62617a3d22626c6970227d5b316d5d2929"} or
  __embedded_query__{__cortex_query__="73756d20627928666f6f2920287261746528626172317b5f5f636f727465785f73686172645f5f3d22315f6f665f33222c62617a3d22626c6970227d5b316d5d2929"} or
  __embedded_query__{__cortex_query__="73756d20627928666f6f2920287261746528626172317b5f5f636f727465785f73686172645f5f3d22325f6f665f33222c62617a3d22626c6970227d5b316d5d2929"}
)
```

The queryable implementation will look for these `__embedded_query__` and `__cortex_query__` labels and upon finding them, shell out to a downstream querier with the encoded query. The Engine will then reassemble the resulting SeriesSet, applying parent operations and merging multiple child queries as necessary.

### label grouping should include shard labels
- [x] implemented

`sum by (foo)` when sharded will return a vector with labels that only include `foo=<value>`. Due to the merge behavior of the union operator, ([_or_](https://prometheus.io/docs/prometheus/latest/querying/operators/)), this would result in discarding data in later vectors which have the same label value for _foo_. Therefore, we need to turn these into:
```
sum by (foo) (
  sum by (foo, __cortex_shard__) (rate(bar{baz="blip", __cortex_shard__="0_of_3"})) or
  ...
)
```

#### Splitting non-sum queries
- [x] implemented

We need a way to handle shard splitting for non-parallelized queries. In the sum example, we introduce `__cortex_shard__` labels in the AST and parallelize them. We need to ensure we're querying the right shards for non-sum queries as well. This may be handled either in the AST (ideal as it isolates logic) or in the backend (i.e. a backend could detect which queries do not have `__cortex_shard__` labels and fan-out/collect them at that level)

#### Better logic for determining which subtrees to execute
- [ ] ~~implemented~~ this optimization will be left for a later pr

Currently, parallelizable sums will be executed on a downstream querier, but non-sum subtrees will not. In the example `rate(<selector>[5m])`, the selector matrix would be dispatched to a querier, but the rate would be computed over the entire series by the frontend. This is certainly suboptimal and I'll be adding more logic to correct this.

#### Custom impls for specific functions
- [ ] ~~implemented~~ this optimization will be left for a later pr

As an example, the `average` function is not associative and is thus difficult to combine across shards without knowing the number of data points the average was calculated over. However, average may be remapped to be a `sum(count)`, which is associative and parallelizes nicely.

#### Injecting shard configurations into the frontend
- [x] implemented

This could be done similarly to how `PeriodConfigs` are used to create composite stores (i.e. as a configuration in a yaml file) or some other method.

## Nice to haves/Ideal Solution/etc
- It would be nice to have an interface for `promql.Engine` so that evaluation wouldn't need to be handled in the `storage.Queryable` interface. It would remove the need to splice in selectors with encoded queries and allow for a cleaner, less obfuscated implementation. This may be a significant lift, though. As an example, the interface may look something like
```
type Engine Interface {
  // signatures for every AST node type
  EvalBinOp(left promql.Expr, right promql.Expr)  (promql.Expr, error)
  // others
}
```


pkg/querier/astmapper/shard_summer_test.go
```diff
@@ -0,0 +1,79 @@
...
+			  sum by(foo) (
+				sum by(foo) (rate(foo{__cortex_shard__="0_of_2",baz="blip"}[1m])) or
+				sum by(foo) (rate(foo{__cortex_shard__="1_of_2",baz="blip"}[1m]))
+			  )
+			)`,
+		},
+		// This is currently redundant: sums split into sharded versions, including summed sums.
+		{
+			2,
+			`sum(sum by(foo) (rate(bar1{baz="blip"}[1m])))`,
```
cyriltovena: can you add example with `sum without()` and `sum(rate(bar1{baz="blip"}[1m]))` ?

pkg/querier/querysharding/queryable.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Select returns a set of series that matches the given label matchers.
+func (q *DownstreamQuerier) Select(
+	_ *storage.SelectParams,
+	matchers ...*labels.Matcher,
+) (storage.SeriesSet, storage.Warnings, error) {
+	var embeddedQuery string
+	var isEmbedded bool
+	for _, matcher := range matchers {
+		if matcher.Name == "__name__" && matcher.Value == astmapper.EMBEDDED_QUERY_FLAG {
```
cyriltovena:
```suggestion
		if matcher.Name == labels.MetricName && matcher.Value == astmapper.EMBEDDED_QUERY_FLAG {
```

pkg/querier/querysharding/queryable.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// handleEmbeddedQuery defers execution of an encoded query to a downstream Handler
+func (q *DownstreamQuerier) handleEmbeddedQuery(encoded string) (storage.SeriesSet, storage.Warnings, error) {
+	decoded, err := hex.DecodeString(encoded)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	resp, err := q.Handler.Do(q.Ctx, ReplaceQuery(*q.Req, string(decoded)))
```
cyriltovena: This is going to be called sequentially by the engine. You need to implement a lazy `storage.SeriesSet` that doesn't block the engine waiting for the result.

pkg/querier/astmapper/parallel.go
```diff
@@ -0,0 +1,81 @@
...
+		return true
+
+	case *promql.EvalStmt:
+		return CanParallel(n.Expr)
+
+	case *promql.MatrixSelector, *promql.NumberLiteral, *promql.StringLiteral, *promql.VectorSelector:
+		return true
+
+	default:
+		panic(errors.Errorf("CloneNode: unhandled node type %T", node))
```
cyriltovena:
```suggestion
		panic(errors.Errorf("CanParallel: unhandled node type %T", node))
```

pkg/querier/querysharding/middleware.go
```diff
@@ -0,0 +1,132 @@
...
+	)
+
+	if err != nil {
+		return nil, err
+	}
+
+	qry, err := qs.engine.NewRangeQuery(
+		queryable,
+		mappedQuery.String(),
+		time.Unix(r.Start, 0), time.Unix(r.End, 0),
```
cyriltovena: Just so you don't forget about this: `r.Start`, `r.End` and `r.Step` are in milliseconds.

pkg/querier/astmapper/embedded.go
```diff
@@ -0,0 +1,84 @@
...
+	embedded_query, err := labels.NewMatcher(labels.MatchEqual, QUERY_LABEL, encoded)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if isMatrix {
+		return &promql.MatrixSelector{
+			Name:          EMBEDDED_QUERY_FLAG,
+			Range:         time.Minute,
```
cyriltovena: Range is actually required for performing operation like rate see https://github.com/prometheus/prometheus/blob/master/promql/functions.go#L132.

I think you need to convey that information if the current node is a Matrix.

pkg/querier/astmapper/parallel.go
owen-d: My obvious copy/past jobs are embarassing >.<

pkg/querier/querysharding/queryable.go
owen-d: Thanks for the heads up. This should be resolved by 46f7dfe and 9c15c0f!

pkg/querier/querysharding/middleware.go
owen-d: This should be resolved by 2f00289.

pkg/querier/astmapper/shard_summer_test.go
owen-d: added in b892aca

pkg/chunk/series_store.go
```diff
@@ -281,10 +282,25 @@ func (c *seriesStore) lookupSeriesByMetricNameMatchers(ctx context.Context, from
 		return nil, err
 	}
 
+	level.Debug(log).Log("msg", "lookupSeriesByMetricNameMatchers:shard", "shard", shard, "label_index", shardLabelIndex, "matchers", len(matchers))
+
 	if shard != nil {
 		matchers = append(matchers[:shardLabelIndex], matchers[shardLabelIndex+1:]...)
+		// Just get series for metric if there are no matchers
```
owen-d: we'd still need to filter the resulting seriesIDs here. I think it can more be more succinctly expressed by calculating the shard/splicing shard labels then running the `len(matchers) == 0` logic once.

owen-d:
deprecated in favor of https://github.com/cortexproject/cortex/pull/1878

PR #1734, opened 2019-10-16
Frontend middlewares refactoring

cyriltovena:
This makes the `frontend` and `queryrange` package easier to use by external projects. The behaviour of both packages hasn't changed.

Middlewares now uses `Request` and `Response` interfaces . This allows to create a roundtripper that will encode/decode HTTP requests/responses (using a `Codec`) to wrap the middleware chain.

To support this new Response interface the cache protobuf message has been updated with an [`Any`](https://developers.google.com/protocol-buffers/docs/proto3#any) field. I have also updated the code so that if the field doesn't exist, it will be trigger a cache miss.

I've added some tests for the cache migration but also tested this in our dev environment.

The queryrange configuration has moved but flag definition remains the same, this allows seamless update if you use flag configuration. However configuration by file must be updated to new field `query_range` at root.

/cc @owen-d @gouthamve can you please review this.

Next steps:
- Change the cache to support time range different than one day.
- Allows to split queries by a configured duration for more parallelism.


pkg/querier/queryrange/results_cache.go
```diff
@@ -235,6 +307,13 @@ func (s resultsCache) get(ctx context.Context, key string) ([]Extent, bool) {
 		return nil, false
 	}
 
+	// Refreshes the cache if it contains an old proto schema.
+	for _, e := range resp.Extents {
```
owen-d: This is not explicitly related to this pr, but I'm curious why a single cache hit maps to multiple entries. Since we already split by day, why isn't there a 1:1 mapping between cache keys and Extents?

pkg/querier/queryrange/roundtrip.go
```diff
@@ -72,59 +81,73 @@ func MergeMiddlewares(middleware ...Middleware) Middleware {
...
+func NewTripperware(cfg Config, log log.Logger, limits Limits, codec Codec, cacheExtractor Extractor) (frontend.Tripperware, error) {
+	queryRangeMiddleware := []Middleware{LimitsMiddleware(limits)}
+	if cfg.AlignQueriesWithStep {
+		queryRangeMiddleware = append(queryRangeMiddleware, InstrumentMiddleware("step_align"), StepAlignMiddleware)
+	}
+	if cfg.SplitQueriesByDay {
+		queryRangeMiddleware = append(queryRangeMiddleware, InstrumentMiddleware("split_by_day"), SplitByDayMiddleware(limits, codec))
+	}
+	if cfg.CacheResults {
+		queryCacheMiddleware, err := NewResultsCacheMiddleware(log, cfg.ResultsCacheConfig, limits, codec, cacheExtractor)
```
owen-d: Just to confirm, the utilization of `limits` in multiple middlewares will implement that parallelism at _each_ middleware function (n^m total max parallelism where `n = parallelism factor` and `m = number of middleware utilizing doRequests`). This is only until we have something more sophisticated to control centralized limits, yes?

Note: Currently cortex uses `doRequests` in [multiple places](https://github.com/cortexproject/cortex/search?q=doRequests&unscoped_q=doRequests) so it's not like you're introducing this behavior, but it's something I wanted to confirm.

cyriltovena: Correct !

pkg/querier/queryrange/results_cache.go
cyriltovena: Its because the cache key is per day for a query. However for a given query you might not ask a full day, e.g 2 days and 12 hours.

So we still store partial result as extent (12hours), if another query for the same day arrives again but still not a full day just 5h hour more, then we keep adding to the extent list. (12h+5h extent).

So basically the cache is building up by day as more request comes in, and we store result for a given day inside the extent list. Of course you could have a single extent with one full day of data when you query the first time for 30days, that's guaranteed.

This is not from me, and personally I found this very smart.

owen-d: Clever, thanks for the explanation.

pkg/querier/queryrange/queryrange.proto
```diff
@@ -46,7 +47,8 @@ message CachedResponse  {
 
 message Extent  {
 	int64 start = 1 [(gogoproto.jsontag) = "start"];
-	int64 end = 2 [(gogoproto.jsontag) = "end"];
-  APIResponse response = 3 [(gogoproto.jsontag) = "response"];
+  int64 end = 2 [(gogoproto.jsontag) = "end"];
+  reserved 3;
```
gouthamve: Why is this reserved?

pkg/querier/queryrange/results_cache.go
```diff
@@ -154,63 +192,97 @@ func (s resultsCache) handleHit(ctx context.Context, r *Request, extents []Exten
 
 		accumulator.TraceId = jaegerTraceID(ctx)
 		accumulator.End = extents[i].End
-		accumulator.Response, err = mergeAPIResponses([]*APIResponse{accumulator.Response, extents[i].Response})
+		accRes, err := accumulator.toResponse()
```
gouthamve: Hrm, we are serializing _and_ serializing for every iteration here. Seems wasteful, can we optimize?

pkg/querier/queryrange/query_range.go
```diff
@@ -4,18 +4,18 @@ import (
...
-	"math"
+	math "math"
 	"net/http"
 	"net/url"
 	"sort"
 	"strconv"
-	"time"
+	time "time"
 
+	proto "github.com/gogo/protobuf/proto"
```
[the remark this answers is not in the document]
cyriltovena: yeah thanks for catching this, not sure what is going I will make sure formatting is fixed.

pkg/querier/queryrange/queryrange.proto
cyriltovena: This is to ease the cache transition. We ignore the old response.

pkg/querier/frontend/frontend_test.go
```diff
@@ -150,7 +149,6 @@ func testFrontend(t *testing.T, handler http.Handler, test func(addr string)) {
 		workerConfig WorkerConfig
 	)
 	flagext.DefaultValues(&config, &workerConfig)
-	config.SplitQueriesByDay = true
```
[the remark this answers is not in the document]
cyriltovena: The frontend doesn't have the config anymore this is handle in the queryrange package now.

The idea was to separate concern.

pkg/querier/queryrange/queryrange.proto
gouthamve: Hrm, curious how it's done. Could you also add a comment to make it clear to future readers?

pkg/querier/frontend/frontend_test.go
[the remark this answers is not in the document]
gouthamve: Hrm, fair enough. But it's always good to have a test that tests end-to-end, and the wiring up.

pkg/querier/queryrange/queryrange.proto
cyriltovena: well it simple, since it's reserved and the new property uses another key index, the new code won't catch old entry values and so update the cache accordingly.

I will add a comment.

pkg/querier/queryrange/query_range.go
```diff
@@ -4,18 +4,18 @@ import (
 	"bytes"
 	"context"
 	"io/ioutil"
-	"math"
+	math "math"
```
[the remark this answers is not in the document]
cyriltovena: Turned out this is the new language server doing this.

pkg/querier/queryrange/results_cache.go
cyriltovena: I'll check if I can use different type for accumulating, so I don't loose the type.

pkg/querier/frontend/frontend_test.go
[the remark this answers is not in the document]
cyriltovena: I added back the test but in the queryrange package this time.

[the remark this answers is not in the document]
cyriltovena: see https://github.com/cortexproject/cortex/pull/1734/commits/6d171addd0c4c37407dbc051239c4075b5cb63ec

pkg/querier/queryrange/results_cache.go
cyriltovena: I've updated the code to avoid the accumulator serializing and deserializing. Let me know if that works for you.

cyriltovena:
Can we retrigger the build, it had a glitch. circleci ....

Makefile
```diff
@@ -126,7 +126,7 @@ $(EXES):
 protos: $(PROTO_GOS)
 
 %.pb.go:
-	protoc -I $(GOPATH)/src:./vendor:./$(@D) --gogoslick_out=plugins=grpc:./$(@D) ./$(patsubst %.pb.go,%.proto,$@)
+		protoc -I $(GOPATH)/src:./vendor:./$(@D) --gogoslick_out=plugins=grpc,Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,:./$(@D) ./$(patsubst %.pb.go,%.proto,$@)
```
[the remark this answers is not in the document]
cyriltovena: Yep sorry this was a typo.

pkg/querier/queryrange/roundtrip.go
```diff
@@ -72,60 +82,75 @@ func MergeMiddlewares(middleware ...Middleware) Middleware {
...
-	if err != nil {
-		return nil, err
-	}
-
-	maxQueryLen := q.limits.MaxQueryLength(userid)
-	queryLen := timestamp.Time(request.End).Sub(timestamp.Time(request.Start))
-	if maxQueryLen != 0 && queryLen > maxQueryLen {
-		return nil, httpgrpc.Errorf(http.StatusBadRequest, validation.ErrQueryTooLong, queryLen, maxQueryLen)
-	}
+	LogToSpan(r.Context(), request)
```
cyriltovena: FYI @jtlisi @gouthamve  I had to rebased to master and re-introduce this.

see https://github.com/cortexproject/cortex/pull/1734/files#diff-19934b07a15ed5212e1637c543973bd1R79

PR #1735, opened 2019-10-17
Guard agasint divide by 0 error when the rings ingester count is 0

thorfour:


thorfour:
> A simple test would be good for this though.

Done

PR #1740, opened 2019-10-18
Removed unused code

pracucci:
While reviewing the distributor implementation, I've noticed the `sendDuration` metric was defined but unused. I've took the chance to quickly run a linter to remove unused variables and code.

_Don't know how this kind of PRs are seen, but in case there's any interest, here it is._

### DIFF
--- a/pkg/chunk/cache/redis_cache.go
+++ b/pkg/chunk/cache/redis_cache.go
@@ -141,14 +141,3 @@ func (c *RedisCache) ping(ctx context.Context) error {
 	}
 	return err
 }
-
-func redisStatusCode(err error) string {
-	switch err {
-	case nil:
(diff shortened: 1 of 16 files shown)


pkg/chunk/cache/redis_cache.go
```diff
@@ -141,14 +141,3 @@ func (c *RedisCache) ping(ctx context.Context) error {
 	}
 	return err
 }
-
-func redisStatusCode(err error) string {
```
bboreham: @dmitsh do we want to use this?
There could be a field on `instrumentedCache` to call either this function or `instr.ErrorCode()` for Redis or Memcached.
But I'm not 100% sure we need to decode `ErrNil` now.

Issue #1567, opened 2019-08-08
Add user(tenant) info in the metrics exposed by query frontend


achilles42:
Currently, for the metrics which are exposed by query frontend does not have user(tenant) info in the labels, it gets really difficult to figure out which tenant's queries are getting stuck.

I would suggest, we should add user info on the following metrics-
1. query_frontend_queue_duration_seconds - Time spend by requests queued.
2. query_frontend_queue_length - number of queries in the queue.
3. frontend_query_range_duration_seconds - Total time spent in seconds doing query range requests.

Please let me know your thoughts on this. I would love to raise PR.

csmarchbanks:
I think it definitely makes sense for queue_length to be per tenant so it is easy to see which tenant is backing up.

I am less convinced by the other metrics, but not ruling them out. An alternative would be slow query logging so it is also easier to see which queries are slow as well.

achilles42:
@csmarchbanks  Sure.

For the `query_frontend_queue_duration_seconds ` and `frontend_query_range_duration_seconds `, I agree that the slow query logs can give us exact info about which query is not performing as expected or taking more time.
But I am thinking from a consistency point of view where one metric exposes user_id and others don't. IMO, if we add the user_id info on other metrics, then the users can have cumulative details(by doing sum(query_frontend_queue_duration_seconds{}))as well as per tenant. Please let me know WDYT.

For enabling a slow query, I am thinking of having a flag called `log-queries-after`(s) and print the queries in the logs which are taking more time than defined on `log-queries-after`.

bboreham:
FWIW we use Jaeger tracing to see slow queries.  We don't get that many, compared to everything else, so can sample a high percentage.

gouthamve:
So, one of the reasons we are not keen on adding user-id to the other metrics is that they are histograms and things quickly add up with user-ids.

The `log-queries-longer-than` flag makes sense tbh. I see a lot of use-cases for it internally.

gouthamve:
Closing this as we now have slow query logigng.

PR #1744, opened 2019-10-19
Add support to log slow queries in Frontend

achilles42:
This PR has a feature to log slow queries in Frontend based on `querier.log-queries-longer-than` parameter. As per this discussion on issue [1567](https://github.com/cortexproject/cortex/issues/1567)
cc @gouthamve

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,10 @@
   * `-ingester.max-global-series-per-metric`
 * [FEATURE] Flush chunks with stale markers early with `ingester.max-stale-chunk-idle`. #1759
 * [FEATURE] EXPERIMENTAL: Added new KV Store backend based on memberlist library. Components can gossip about tokens and ingester states, instead of using Consul or Etcd. #1721
+* [FEATURE] Allow Query Frontend to log slow queries #1744
 * [ENHANCEMENT] Allocation improvements in adding samples to Chunk. #1706
 * [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
 * [ENHANCEMENT] Consul client can optionally rate-limit itself during Watch (used e.g. by ring watchers) and WatchPrefix (used by HA feature) operations. Rate limiting is disabled by default. New flags added: `--consul.watch-rate-limit`, and `--consul.watch-burst-size`. #1708
-* [FEATURE] Allow Query Frontend to log slow queries #1744
 
 ## 0.3.0 / 2019-10-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [ENHANCEMENT] Allocation improvements in adding samples to Chunk. #1706
 * [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
 * [ENHANCEMENT] Consul client can optionally rate-limit itself during Watch (used e.g. by ring watchers) and WatchPrefix (used by HA feature) operations. Rate limiting is disabled by default. New flags added: `--consul.watch-rate-limit`, and `--consul.watch-burst-size`. #1708
+* [FEATURE] Allow Query Frontend to log slow queries #1744
 
 ## 0.3.0 / 2019-10-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
   * `-ingester.max-global-series-per-metric`
 * [FEATURE] Flush chunks with stale markers early with `ingester.max-stale-chunk-idle`. #1759
 * [FEATURE] EXPERIMENTAL: Added new KV Store backend based on memberlist library. Components can gossip about tokens and ingester states, instead of using Consul or Etcd. #1721
-* [FEATURE] Allow Query Frontend to log slow queries #1744
+* [FEATURE] Allow Query Frontend to log slow queries with `frontend.log-queries-longer-than`. #1744
 * [ENHANCEMENT] Allocation improvements in adding samples to Chunk. #1706
 * [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
 * [ENHANCEMENT] Consul client can optionally rate-limit itself during Watch (used e.g. by ring watchers) and WatchPrefix (used by HA feature) operations. Rate limiting is disabled by default. New flags added: `--consul.watch-rate-limit`, and `--consul.watch-burst-size`. #1708
--- a/docs/prometheus-frontend.yml
+++ b/docs/prometheus-frontend.yml
@@ -16,7 +16,7 @@ server:
   http_listen_port: 9091
 
 frontend:
-  log_queries_longer_than: 1 # in seconds
+  log_queries_longer_than: 1s
   split_queries_by_day: true
   align_queries_with_step: true
   cache_results: true
--- a/docs/prometheus-frontend.yml
+++ b/docs/prometheus-frontend.yml
@@ -16,6 +16,7 @@ server:
   http_listen_port: 9091
 
 frontend:
+  log_queries_longer_than: 1 # in seconds
   split_queries_by_day: true
   align_queries_with_step: true
   cache_results: true
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -55,7 +55,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxOutstandingPerTenant, "querier.max-outstanding-requests-per-tenant", 100, "Maximum number of outstanding requests per tenant per frontend; requests beyond this error with HTTP 429.")
 	f.BoolVar(&cfg.CompressResponses, "querier.compress-http-responses", false, "Compress HTTP responses.")
 	f.StringVar(&cfg.DownstreamURL, "frontend.downstream-url", "", "URL of downstream Prometheus.")
-	f.DurationVar(&cfg.LogQueriesLongerThan, "querier.log-queries-longer-than", 1*time.Second, "Log slow queries")
+	f.DurationVar(&cfg.LogQueriesLongerThan, "querier.log-queries-longer-than", 0, "Log slow queries")
 }
 
 // Frontend queues HTTP requests, dispatches them to backends, and handles retries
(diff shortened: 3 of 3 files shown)


pkg/querier/frontend/frontend.go
```diff
@@ -180,7 +184,14 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
+	startTime := time.Now()
 	resp, err := f.roundTripper.RoundTrip(r)
+	queryResponseTime := time.Now().Sub(startTime)
+
+	if queryResponseTime > f.cfg.LogQueriesLongerThan {
+		level.Info(f.log).Log("msg", "slow query", "url", fmt.Sprintf("http://%s/", r.URL.Host+r.URL.Path+r.URL.RawQuery), "time", queryResponseTime.String())
```
khaines: A few comments on this line:
- Is INFO the correct level for the output? Considering this only outputs for queries that exceed a time threshold set by the operator to be notified about, perhaps this should be WARN.
- Please revisit the output for the `url` field. The string format template and the values concatenated do not appear to rebuild the original url as it was sent i, if that is the intention.
- `time` may not a be a good field name for a duration filed. Please consider `time-taken`
- The `orgID` the query was executed against may be a valuable piece of information to make this data actionable by anyone attempting to reproduce the query load on the backend store.

docs/prometheus-frontend.yml
```diff
@@ -16,6 +16,7 @@ server:
   http_listen_port: 9091
 
 frontend:
+  log_queries_longer_than: 1 # in seconds
```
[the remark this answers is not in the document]
achilles42: Ah sorry, should be 1s.

pkg/querier/frontend/frontend.go
achilles42: Thanks @khaines For feedback. 

> Is INFO the correct level for the output? Considering this only outputs for queries that exceed a time threshold set by the operator to be notified about, perhaps this should be WARN.

Done.

> Please revisit the output for the url field. The string format template and the values concatenated do not appear to rebuild the original url as it was sent i, if that is the intention.

Updated, now using `rHost+r.RequestURI` to get the original request URL.

> time may not a be a good field name for a duration filed. Please consider time-taken

Done.

> The orgID the query was executed against may be a valuable piece of information to make this data actionable by anyone attempting to reproduce the query load on the backend store.

Done.

gouthamve:
Could you also add a changelog entry?

pkg/querier/frontend/frontend.go
```diff
@@ -180,7 +184,20 @@ func (f *Frontend) Handler() http.Handler {
...
 	resp, err := f.roundTripper.RoundTrip(r)
+	queryResponseTime := time.Now().Sub(startTime)
+
+	userID, err := user.ExtractOrgID(r.Context())
+	if err != nil {
+		server.WriteError(w, err)
+		return
+	}
+
+	if queryResponseTime > f.cfg.LogQueriesLongerThan {
```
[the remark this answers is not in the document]
achilles42: done

achilles42:
> Could you also add a changelog entry?
Updated the Changelog.

pkg/querier/frontend/frontend.go
```diff
@@ -42,16 +44,18 @@ var (
...
+	DownstreamURL           string        `yaml:"downstream"`
+	LogQueriesLongerThan    time.Duration `yaml:"log_queries_longer_than"`
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxOutstandingPerTenant, "querier.max-outstanding-requests-per-tenant", 100, "Maximum number of outstanding requests per tenant per frontend; requests beyond this error with HTTP 429.")
 	f.BoolVar(&cfg.CompressResponses, "querier.compress-http-responses", false, "Compress HTTP responses.")
 	f.StringVar(&cfg.DownstreamURL, "frontend.downstream-url", "", "URL of downstream Prometheus.")
+	f.DurationVar(&cfg.LogQueriesLongerThan, "querier.log-queries-longer-than", 0, "Log slow queries")
```
pracucci: Shouldn't this be `frontend.log-queries-longer-than`? Otherwise we have a mismatch between the yaml config and the CLi flags.

pkg/querier/frontend/frontend.go
```diff
@@ -139,7 +143,20 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
+	startTime := time.Now()
 	resp, err := f.roundTripper.RoundTrip(r)
+	queryResponseTime := time.Now().Sub(startTime)
+
+	userID, err := user.ExtractOrgID(r.Context())
```
pracucci: Would be better to move it at the beginning of the function, to avoid running the query and get the results if then we trash them because of this error

pkg/querier/frontend/frontend.go
```diff
@@ -139,7 +143,20 @@ func (f *Frontend) Handler() http.Handler {
...
+	if err != nil {
+		server.WriteError(w, err)
+		return
+	}
+
+	if f.cfg.LogQueriesLongerThan > 0 && queryResponseTime > f.cfg.LogQueriesLongerThan {
+		level.Warn(f.log).Log("msg", "slow query", "orgID", userID, "url", fmt.Sprintf("http://%s", r.Host+r.RequestURI), "time-taken", queryResponseTime.String())
```
pracucci: I'm a bit dubious about the `Warn` level. A warning or error would mean that it's actionable. This log is not necessarily actionable (think about a multi-tenant system and the query is implicitly slow because complex and/or spanning over a large time range). I would see `Info` more appropriate here.

pracucci: Please also replace `"orgID"` with `"user"`.

gouthamve: :+1: to both. The info level logging is pretty much non-existent for the frontend and this won't get lost afaics.

gouthamve: Actually, I prefer `orgID/tenant` over user. And `orgID` is the majority term in use right now. Maybe just variable from `userID` to `orgID`?

pracucci: `tenant` looks the right naming to me, but I've mentioned `user` to keep consistency with the other logs / metrics (it's usually called "user"). I let you @gouthamve say the last word on this.

pkg/querier/frontend/frontend.go
```diff
@@ -72,6 +75,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.CompressResponses, "querier.compress-http-responses", false, "Compress HTTP responses.")
 	cfg.ResultsCacheConfig.RegisterFlags(f)
 	f.StringVar(&cfg.DownstreamURL, "frontend.downstream-url", "", "URL of downstream Prometheus.")
+	f.DurationVar(&cfg.LogQueriesLongerThan, "querier.log-queries-longer-than", 1*time.Second, "Log slow queries")
```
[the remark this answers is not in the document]
achilles42: done

pkg/querier/frontend/frontend.go
achilles42: Does that also mean `max-outstanding-requests-per-tenant` and `compress-http-responses` should be on frontend instead of querier?

pracucci: Maybe should, but - as a general rule - please do not change anything outside this PR's scope.

achilles42: sure. won't be changing anything outside this PR scope. I just wanted to confirm. 

Made the changes. Thanks.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [ENHANCEMENT] Allocation improvements in adding samples to Chunk. #1706
 * [ENHANCEMENT] Consul client now follows recommended practices for blocking queries wrt returned Index value. #1708
 * [ENHANCEMENT] Consul client can optionally rate-limit itself during Watch (used e.g. by ring watchers) and WatchPrefix (used by HA feature) operations. Rate limiting is disabled by default. New flags added: `--consul.watch-rate-limit`, and `--consul.watch-burst-size`. #1708
+* [FEATURE] Allow Query Frontend to log slow queries #1744
```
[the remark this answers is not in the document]
achilles42: Ah sorry, Updated now.

pkg/querier/frontend/frontend.go
achilles42: done.

pkg/querier/frontend/frontend.go
gouthamve: Argh, I just did a quick grep and we use `org_id` and `user` interchangeably everywhere. I think it's best to standardise towards `org_id` rather than user, see https://github.com/cortexproject/cortex/blob/a5ecafc0782b4c480c3902f228e3409a647d82cb/pkg/util/log.go#L119-L124

Also, sadly tenant is not used anywhere.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
   * `-ingester.max-global-series-per-metric`
 * [FEATURE] Flush chunks with stale markers early with `ingester.max-stale-chunk-idle`. #1759
 * [FEATURE] EXPERIMENTAL: Added new KV Store backend based on memberlist library. Components can gossip about tokens and ingester states, instead of using Consul or Etcd. #1721
+* [FEATURE] Allow Query Frontend to log slow queries #1744
```
gouthamve: Could you mention the flag here, see the flush chunks with stale markers above.

pkg/querier/frontend/frontend.go
```diff
@@ -42,16 +44,18 @@ var (
...
+	DownstreamURL           string        `yaml:"downstream"`
+	LogQueriesLongerThan    time.Duration `yaml:"log_queries_longer_than"`
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxOutstandingPerTenant, "querier.max-outstanding-requests-per-tenant", 100, "Maximum number of outstanding requests per tenant per frontend; requests beyond this error with HTTP 429.")
 	f.BoolVar(&cfg.CompressResponses, "querier.compress-http-responses", false, "Compress HTTP responses.")
 	f.StringVar(&cfg.DownstreamURL, "frontend.downstream-url", "", "URL of downstream Prometheus.")
+	f.DurationVar(&cfg.LogQueriesLongerThan, "frontend.log-queries-longer-than", 0, "Log slow queries")
```
gouthamve: Maybe something like: `Log queries that are slower than the specified duration. 0 to disable.`.

pkg/querier/frontend/frontend.go
achilles42: yeah, that makes sense to standardize towards `org_id` everywhere.

Done, updated

CHANGELOG.md
achilles42: added flag.

pkg/querier/frontend/frontend.go
achilles42: updated.

pkg/querier/frontend/frontend.go
achilles42: Thanks @gouthamve

PR #1746, opened 2019-10-21
HTTP handler for ingester shutdown

codesome:
This is based on [this design](https://docs.google.com/document/d/1n1HcdgmsqaVwqxKg_POZqIREOSqZIAlrxuyx4YYV_pA/edit#heading=h.ut9lekjstiqc) for scaling down using the WAL. This would not block #1103 but this is required for scale downs.

I have yet to verify how it will interact with Kubernetes health checks and pod restarts if the endpoint is hit.


pkg/ingester/ingester.go
```diff
@@ -244,6 +244,15 @@ func (i *Ingester) Shutdown() {
 	i.lifecycler.Shutdown()
 }
 
+// ShutdownHandler triggers the following set of operations in order:
+//     * Change the state of ring to stop accepting writes.
+//     * Flush all the chunks.
+func (i *Ingester) ShutdownHandler(w http.ResponseWriter, r *http.Request) {
+	i.Shutdown()
```
gouthamve: So when we call this, we will first look for ingesters to transfer to and then flush the chunks anyways. I don't think we need a `i.Flush()` after. Also please test this in dev once.

codesome: Ah yes. The flushing was somewhere deep inside the code, found it. Will change and also test it in dev.

codesome:
Tested it and it works as intended. Ingester goes to LEAVING states, flushes after waiting for other ingesters, and then leaves the ring.

I also added another commit to avoid panic on multiple shutdown calls (during graceful shutdown after having called `/shutdown` already)

bboreham:
Could you add a description of this API to `docs/apis.md` ?

Also it seems worth mentioning in CHANGELOG.md.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## master / unreleased
 
 * [CHANGE] Removed `Delta` encoding. Any old chunks with `Delta` encoding cannot be read anymore. If `ingester.chunk-encoding` is set to `Delta` the ingester will fail to start. #1706
+* [FEATURE] Added `/shutdown` endpoint for ingester to shutdown all operations of the ingester. #1746
```
jtlisi: Put this at the top of the changleog

bboreham:
Needs rebase for changelog conflict now.

docs/apis.md
```diff
@@ -127,3 +127,10 @@ Error Response: BadRequest(400)
 These API endpoints will disable/enable the current Rule and Alertmanager configuration for a tenant.
 
 Note that setting a new config will effectively "re-enable" the Rules and Alertmanager configuration for a tenant.
+
+#### Ingester Shutdown
+
+`POST /shutdown` - Shutdown all operations of an ingester. Shutdown operations performed are similar to when an ingester is gracefully shutting down, including flushing of chunks if no other ingester is in `PENDING` state. Ingester does not terminate after calling this endpoint.
+
+- Normal Response Codes: NoContent(204)
+- Error Response Codes: Unauthorized(401), NotFound(404)
```
gouthamve: Why would it ever return a 404?

codesome: It would not

PR #1747, opened 2019-10-21
Flush mode for ingester

codesome:
This is as per [this design](https://docs.google.com/document/d/1n1HcdgmsqaVwqxKg_POZqIREOSqZIAlrxuyx4YYV_pA/edit#heading=h.ut9lekjstiqc). Ingester would be run in flush mode in case someone forgets to hit the shutdown endpoint as described in https://github.com/cortexproject/cortex/pull/1746 before scaling down the ingesters with WAL.

Still in draft mode, as I have to check about the ingester quitting gracefully and not just stopping the operations.


pkg/ingester/ingester.go
```diff
@@ -610,6 +625,9 @@ func (i *Ingester) Watch(in *grpc_health_v1.HealthCheckRequest, stream grpc_heal
 // the addition removal of another ingester. Returns 204 when the ingester is
 // ready, 500 otherwise.
 func (i *Ingester) ReadinessHandler(w http.ResponseWriter, r *http.Request) {
+	if i.cfg.FlushMode {
+		w.WriteHeader(http.StatusNoContent)
```
bboreham: did you mean to return after this?

pkg/cortex/cortex.go
```diff
@@ -218,6 +218,10 @@ func (t *Cortex) initModule(cfg *Config, m moduleName) error {
 
 // Run starts Cortex running, and blocks until a signal is received.
 func (t *Cortex) Run() error {
+	if t.ingester != nil && t.ingester.Config().FlushMode {
+		// If it is a flush mode ingester, skip running the server.
```
bboreham: also skips serving metrics?
This bit doesn't feel right to me.

codesome: The flush mode would run as a short-lived job. Do we need metrics to keep track of the flushing?

pkg/ingester/ingester.go
codesome: Yes. Thanks.

pkg/cortex/cortex.go
codesome: Also, it was easier this way to run it as a job and not try and give a signal. The way it would go it the WAL is read during `cortex.New(...)`. `Run()` returns immediately. Flushing happens during `cortex.Stop()`.

gouthamve: We need the metrics please. Metrics are super important to debug when things are not working, and I would prefer having metrics here.

pkg/ingester/ingester.go
```diff
@@ -204,6 +207,18 @@ func New(cfg Config, clientConfig client.Config, limits *validation.Overrides, c
...
+	i.flushQueuesDone.Add(cfg.ConcurrentFlushes)
+	for j := 0; j < cfg.ConcurrentFlushes; j++ {
+		i.flushQueues[j] = util.NewPriorityQueue(i.metrics.flushQueueLength)
+		go i.flushLoop(j)
+	}
+
+	if cfg.FlushMode {
+		// Flush and don't update the ring.
+		i.Flush()
+		return i, nil
```
jtlisi: Could we initiate the shutdown here?

cmd/cortex/main.go
```diff
@@ -75,12 +75,30 @@ func main() {
...
+		if err := t.Run(); err != nil {
+			level.Error(util.Logger).Log("msg", "error running Cortex", "err", err)
+		}
+		close(done)
+	}()
 
 	runtime.KeepAlive(ballast)
-	err = t.Stop()
+
+	if t.IsIngesterJob() {
```
jtlisi: Is it possible to refactor this logic into the ingester module?

codesome:
> Would it be possible to refactor this code so the shutdown is initiated by the ingester module?

One blocker that I faced was that the server was started by the cortex package, and the `Run()` function is blocked on that. I could not find a way to stop the server from the ingester, would it be sane to pass the server object into the ingester which can stop it there?

codesome:
I have a clarification to make here: The flushing happens even before the server starts. But would the metrics be served via the default registry even before the server starts? If not, then the code needs to be refactored so that the server starts before flushing.

jtlisi:
> I have a clarification to make here: The flushing happens even before the server starts. But would the metrics be served via the default registry even before the server starts? If not, then the code needs to be refactored so that the server starts before flushing.

Yea we are going to need the server to be running if we want metrics from this job.

pkg/ingester/ingester.go
codesome: I have been staring at this for a while and I might need some pointers on how cortex can be shut down within ingester. I can only see signals as an option.

pkg/cortex/cortex.go
```diff
@@ -249,6 +254,7 @@ func (t *Cortex) Stop() error {
 	for i := len(deps) - 1; i >= 0; i-- {
 		t.stopModule(deps[i])
 	}
+	t.server.Stop()
```
gouthamve: Shouldn't it be called in `stopModule` above?

pkg/ingester/ingester.go
gouthamve: Hrm, let me see. Can we pass `t.Stop` to the ingester when initiating and then just call that?

pkg/ingester/ingester.go
```diff
@@ -263,8 +272,11 @@ func (i *Ingester) Shutdown() {
 		close(i.quit)
 		i.done.Wait()
 
-		// Next initiate our graceful exit from the ring.
-		i.lifecycler.Shutdown()
+		if !i.cfg.FlushMode {
+			// Next initiate our graceful exit from the ring.
+			// If ingester was in flush mode, then it would not be added in the ring.
+			i.lifecycler.Shutdown()
```
gouthamve: I wonder what happens if we call `i.lifecycler.Shutdown()` anyways?

pkg/cortex/cortex.go
codesome: The server module only creates server object and initiates the shutdown. The server is being started and stopped manually in the cortex module. Maybe it should be moved before we call any stopModule.

pkg/ingester/ingester.go
codesome: Will try that! Should work

pkg/ingester/ingester.go
codesome: nil pointer exception. `i.lifecycler` is not set if it's flush mode, because it need not.

pkg/cortex/cortex.go
codesome: Looks like it should. Will remove it.

pkg/ingester/ingester.go
codesome: Causes an initialization loop :/

codesome: Also, I think it's not safe to stop all modules while we are in the process of initing all the modules (additionally, the server is started after all modules are inited).

codesome:
I am curious how can we scrape the metrics here even if we have the server running. Reason: The flushing happens in `cortex.New` when the ingester is being created. The server is stopped right after it was started, leaving little chance for a scrape to gather any metrics.

pkg/ingester/ingester.go
codesome: I tried another modification of my earlier attempt - the modules propagate the info of shutting down via a bool instead of cortex explicitly knowing about the ingester job.

codesome:
I tested this with #1103 and seems to be working fine.

jtlisi:
@codesome Ideally we use something like a push gateway for something like this. However, that may take a significant bit of refactoring. Otherwise we could add a wait period before shutting down.

codesome:
I have now rebased this PR on top of https://github.com/cortexproject/cortex/pull/1885 with a small fix

codesome:
Tested it. Worked fine.

bboreham:
Sorry to come up with another suggestion after so many other discussions, but it seemed cleaner to me: how about doing it from a different `-target` ?  That way `Cortex.Run()` doesn't need to change.

codesome:
Superseded by https://github.com/cortexproject/cortex/pull/2075

Issue #1534, opened 2019-07-29
Add jitter to HA deduping heartbeats


gouthamve:
Right now there is a thundering herd every `15s` (updateTimeout) where all distributors will be CASing the ring. With 25+ distributors, I think this puts undue load on etcd. I think we can do better by adding jitter to the heartbeats.

adnxn:
@gouthamve will take a look at this one

gouthamve:
Closed this in #1748

PR #1748, opened 2019-10-22
Add jitter to HA deduping heartbeats (#1543)

adnxn:
addresses #1534, adds a new `distributor.ha-tracker.update-timeout-jitter` config to add a +/- jitter to the UpdateTimeout for the HA dedupe heartbeat.

### DIFF
--- a/pkg/distributor/ha_tracker.go
+++ b/pkg/distributor/ha_tracker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"strings"
 	"sync"
@@ -62,9 +63,10 @@ func NewReplicaDesc() *ReplicaDesc {
 // Track the replica we're accepting samples from
 // for each HA cluster we know about.
 type haTracker struct {
-	logger log.Logger
-	cfg    HATrackerConfig
-	client kv.Client
+	logger              log.Logger
+	cfg                 HATrackerConfig
+	client              kv.Client
+	updateTimeoutJitter time.Duration
 
 	// Replicas we are accepting samples from.
 	electedLock sync.RWMutex
@@ -80,7 +82,8 @@ type HATrackerConfig struct {
 	// We should only update the timestamp if the difference
 	// between the stored timestamp and the time we received a sample at
 	// is more than this duration.
-	UpdateTimeout time.Duration `yaml:"ha_tracker_update_timeout"`
+	UpdateTimeout          time.Duration `yaml:"ha_tracker_update_timeout"`
+	UpdateTimeoutJitterMax time.Duration `yaml:"ha_tracker_update_timeout_jitter_max"`
 	// We should only failover to accepting samples from a replica
 	// other than the replica written in the KVStore if the difference
 	// between the stored timestamp and the time we received a sample is
@@ -100,6 +103,10 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
 		"distributor.ha-tracker.update-timeout",
 		15*time.Second,
 		"Update the timestamp in the KV store for a given cluster/replica only after this amount of time has passed since the current stored timestamp.")
+	f.DurationVar(&cfg.UpdateTimeoutJitterMax,
+		"distributor.ha-tracker.update-timeout-jitter-max",
+		5*time.Second,
+		"To spread the HA deduping heartbeats out over time.")
 	f.DurationVar(&cfg.FailoverTimeout,
 		"distributor.ha-tracker.failover-timeout",
 		30*time.Second,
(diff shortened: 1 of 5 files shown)


pracucci:
Can you elaborate which specific problem you're addressing in this PR? I've checked out the linked #1543 but I'm not sure I'm following it.

Thanks!

adnxn:
@pracucci: oops, typo. meant #1534.

pkg/distributor/ha_tracker.go
```diff
@@ -211,9 +217,15 @@ func (c *haTracker) checkKVStore(ctx context.Context, key, replica string, now t
 	return c.client.CAS(ctx, key, func(in interface{}) (out interface{}, retry bool, err error) {
 		if desc, ok := in.(*ReplicaDesc); ok {
 
+			// Adjust update timeout slightly to spread heartbeats out over time with +/- UpdateTimeoutJitter
+			var jitter time.Duration
+			if c.cfg.UpdateTimeoutJitter != 0 {
+				jitter = time.Duration(rand.Int63n(int64(2*c.cfg.UpdateTimeoutJitter))) - c.cfg.UpdateTimeoutJitter
```
cstyan: I feel like we might be able to set the jitter for the distributor instead of per request. At least, I'd like to see what the write load reduction is doing it that way. @gouthamve thoughts?

gouthamve:
Hey @adnxn, so this is pretty good, and almost done! 

But as @cstyan mentioned, best to instantiate the amount of jitter when we create the distributor itself and not call `rand()` per request.
With the above change and validating config (as suggested by Callum), I think this is good to go!

pkg/distributor/ha_tracker.go
```diff
@@ -113,9 +120,14 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
 func newClusterTracker(cfg HATrackerConfig) (*haTracker, error) {
 	codec := codec.Proto{Factory: ProtoReplicaDescFactory}
 
-	if cfg.FailoverTimeout <= cfg.UpdateTimeout {
-		return nil, fmt.Errorf("HA Tracker failover timeout must be greater than update timeout, %d is <= %d", cfg.FailoverTimeout, cfg.UpdateTimeout)
+	if cfg.UpdateTimeoutJitterMax != 0 {
+		cfg.UpdateTimeoutJitter = time.Duration(rand.Int63n(int64(2*cfg.UpdateTimeoutJitterMax))) - cfg.UpdateTimeoutJitterMax
```
gouthamve: Right, this field shouldn't be part of `cfg`, but rather `haTracker`. Maybe something like `updateTimeoutJitter := ..` and then use the variable in L136 as part of `t := haTrcker{}``

pkg/distributor/ha_tracker.go
```diff
@@ -113,9 +120,14 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
...
 	}
+
+	if cfg.FailoverTimeout <= cfg.UpdateTimeout+cfg.UpdateTimeoutJitterMax {
```
gouthamve: Instead of just making sure of `<=`, lets make it atleast one second bigger? `cfg.FailoverTimeout + time.Second <= `?

pkg/distributor/ha_tracker.go
```diff
@@ -113,16 +120,24 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
...
 	codec := codec.Proto{Factory: ProtoReplicaDescFactory}
 
-	if cfg.FailoverTimeout <= cfg.UpdateTimeout {
-		return nil, fmt.Errorf("HA Tracker failover timeout must be greater than update timeout, %d is <= %d", cfg.FailoverTimeout, cfg.UpdateTimeout)
+	var jitter time.Duration
+	if cfg.UpdateTimeoutJitterMax != 0 {
+		jitter = time.Duration(rand.Int63n(int64(2*cfg.UpdateTimeoutJitterMax))) - cfg.UpdateTimeoutJitterMax
 	}
+
+	if cfg.FailoverTimeout <= cfg.UpdateTimeout+cfg.UpdateTimeoutJitterMax+time.Second {
```
pracucci: This is not the right place to validate the config. Please add `func (cfg *HATrackerConfig) Validate() error` and move the validation there (should return an error on validation failure, `nil` otherwise), add the `Validate()` function to `distributor.Config` (this should call `HATrackerConfig.Validate()`) and the call the `distributor.Config.Validate()` method from the `Validate()` function in `cortex.go` (looks complex, but it's a very easy chain of `Validate()` calls).

pkg/distributor/ha_tracker.go
```diff
@@ -113,16 +120,24 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
...
+		return nil, fmt.Errorf("HA Tracker failover timeout must be 1s greater than update timeout, %d is <= %d",
```
pracucci: You can make the message more clear this way:

```
fmt.Errorf("HA Tracker failover timeout (%v) must be at least 1s greater than update timeout (%v)", cfg.FailoverTimeout, cfg.UpdateTimeout+cfg.UpdateTimeoutJitterMax)
```

Moreover, given the error message, the check should be `<` instead of `<=`, so:

```
if cfg.FailoverTimeout < cfg.UpdateTimeout+cfg.UpdateTimeoutJitterMax+time.Second {
```

pkg/distributor/ha_tracker.go
```diff
@@ -113,16 +120,24 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
 func newClusterTracker(cfg HATrackerConfig) (*haTracker, error) {
 	codec := codec.Proto{Factory: ProtoReplicaDescFactory}
 
-	if cfg.FailoverTimeout <= cfg.UpdateTimeout {
-		return nil, fmt.Errorf("HA Tracker failover timeout must be greater than update timeout, %d is <= %d", cfg.FailoverTimeout, cfg.UpdateTimeout)
+	var jitter time.Duration
+	if cfg.UpdateTimeoutJitterMax != 0 {
```
pracucci: Better `> 0` for safety.

pkg/distributor/ha_tracker.go
adnxn: im okay with making this change - however wondering if it would make more sense to abstract this out as we add more validations. right now it is only validating a single field. but yea, lmk.

pracucci: I would get stick to the PR's scope. Config validation support has been introduced very recently and will be adopted more over the time. My suggestion is to validate only the new config option introduced with this PR, while we'll improve the rest of config over the time.

Issue #1525, opened 2019-07-22
Come up with a way of migrating from consul to etcd.


tomwilkie:
We probably want to be able to mirror the ring into etcd, and then have a way of switching over to it.

Will be tricky to get right, so some though required.

tomwilkie:
I've asked @pstibrany to look into this.

pstibrany:
Sent PR #1749 to support such migration. Details in PR.

pstibrany:
#1749 is now in Cortex 0.6.0, so this issue can be closed.

PR #1749, opened 2019-10-23
Added KV Store client that mirrors one backend to another

pstibrany:
Since we support multiple KV Store backends now (Consul, Etcd, memberlist-gossiping soon), we want to be able to migrate from one backend to a different one gracefully, without shutting down all components at once. (Issue #1525) This PR implements one way of doing that.

It adds yet another KV Store backend, called multiClient. This client is configured with several backends, one of them is designated as primary. Primary one is used for all client operations like Get, CAS, WatchKey, and WatchPrefix. ~At the same time, multiClient reads all changes done in primary backend and copies them to secondary backend. (That is done using WatchPrefix, with "" prefix [update: now configurable])~ Updates via CAS are also written to secondary store, if mirroring is enabled.

It is possible to switch the primary backend in the runtime by using existing overrides mechanism. This mechanism has been extended and refactored a bit. [update: it's now also possible to enable/disable mirroring in runtime via overrides]

Cortex has new "runtimeConfigValues" struct that is reloaded using OverridesManager (now `runtime_config.Manager`). runtimeConfigValues currently contains limit overrides and multi-client primary store. `runtime_config.Manager` now lives in its own `runtime_config` package, and its job is to reload runtime YAML file (`overrides.yaml`), and provide new value to clients and listeners.

Components interested in using runtimeConfigValues setup themselves in main Cortex setup code (cortex.go, modules.go). At the moment, interested components are Overrides (using limits) and Ring (when using multi-client, for listening on primary store changes).

After all this introduction, how is this supposed to work? In scenario, where one wants to migrate from Consul to Etcd, here are the steps for migration:

1) setup all components to use multi-client, with consul as primary and etcd as secondary store, and restart them. Components will keep using consul, but will also mirror the ring to etcd.
2) switch primary store to etcd in runtime (eg. by pushing new config map)
3) once done, all components are using etcd (and mirroring to consul)
4) remove multi-client configuration, and let components use etcd only

All this can be done without any downtime. ~Cost of this is higher load on consul/etcd when components do mirroring, but that can be allieviated by using rate limits on consul client. [update: mirroring itself now has rate limit]~ [Watch+mirroring functionality has been replaced with simple update of values in secondary stores when performing CAS operation]


pstibrany:
TODO: think about how to do this without killing consul/etcd with too many requests. If every component (distributor/ingester/querier) does the mirroring all the time, that's probably little too much.

pstibrany:
Cleaned up the code a little bit, and added: 
- rate limit for watching for changes in the primary store
- config option for watch prefix
- flag for enabling/disabling mirroring in runtime (via overrides)

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+		} else if stored {
+			level.Info(util.Logger).Log("msg", "stored updated value to secondary store", "key", key, "primary", primary.name, "secondary", kvc.name)
+		}
+	}
+}
+
+func (m *multiClient) Stop() {
+	m.cancel()
+
+	// TODO: call Stop method of all clients from gossiping ring branch.
```
jtlisi: Are you planning on fixing this in this PR? Since both of these clients are only called within the MultClient it shouldn't be a major issue but it is likely good practice.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+				level.Debug(util.Logger).Log("msg", "no change", "key", key)
+				return nil, false, nil
+			}
+
+			// try once
+			return in, false, nil
+		})
+
+		if err != nil {
+			level.Error(util.Logger).Log("msg", "failed to update value in secondary store", "key", key, "err", err, "primary", primary.name, "secondary", kvc.name)
```
jtlisi: If we fail to update here does that mean going forward it will break the migration to the new ring? Is their any redundancies or does this mean we need to start fresh and roll out a new set of ingesters? That should be fine since this is a rare procedure just good to know.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+func (m *multiClient) WatchPrefix(ctx context.Context, prefix string, f func(string, interface{}) bool) {
+	_ = m.runWithPrimaryClient(ctx, func(newCtx context.Context, primary kvclient) error {
+		primary.client.WatchPrefix(newCtx, prefix, f)
+		return newCtx.Err()
+	})
+}
+
+func (m *multiClient) watchChanges(prefix string) {
```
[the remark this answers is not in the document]
pstibrany: Done.

pkg/ring/kv/multi.go
pstibrany: That TODO refers to new Stop method introduced by PR #1727. It's not part of Client interface on this branch yet.

Perhaps I should just remove Stop method here, and introduce it again when either merging this, or merging PR #1727. It kind of depends on which PR is merged first :-(

pkg/ring/kv/multi.go
pstibrany: If this fails, something bad happened, e.g. Consul/Etcd is down, or CAS operation failed too many times. However, watching still continues, because WatchPrefix only ends when either context is done, or 'f' returns false (which we don't do in watchChanges).

As for redundancies – idea is that ALL clients accessing the ring will be configured to use multiClient and mirroring at the same time, so if at least one is still alive, mirroring is done. This is also reason why we only try to update the value once.

pkg/cortex/modules.go
```diff
@@ -151,6 +155,10 @@ func (t *Cortex) stopServer() (err error) {
 }
 
 func (t *Cortex) initRing(cfg *Config) (err error) {
+	if cfg.Ingester.LifecyclerConfig.RingConfig.KVStore.Store == kv.MultiClientName {
+		cfg.Ingester.LifecyclerConfig.RingConfig.KVStore.Multi.RuntimeConfigChannel = multiClientRuntimeConfigChannel(t.runtimeConfig)
+	}
+
```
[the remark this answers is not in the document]
pstibrany: I agree that this is ugly. Passing `t.runtimeConfig` to `ring.New` would however introduce cyclic dependency, unless hidden behind some interface. (Also, this runtimeConfig is specific to Cortex, and Loki will need its own). I'll try to implement that.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+	"sync"
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/uber-go/atomic"
+	"golang.org/x/time/rate"
+
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+type multiConfig struct {
```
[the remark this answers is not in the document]
pstibrany: I can do that. I'm a firm believer in minimizing your API surface, but if people are crazy enough to depend on API exported by some random project on GitHub (ehm, Loki, ehm), that's their problem :)

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+	Primary   string
+	Secondary string
+
+	MirrorRateLimit float64
+	MirrorRateBurst int
+	MirrorEnabled   bool
+	MirrorPrefix    string
+
```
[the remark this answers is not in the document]
pstibrany: Yeah, my bad. I got confused by the fact that `consul.Config` doesn't have yaml tags, but then I discovered that we configure those values by command line flags. Even without yaml tags, they can be used by yaml, with lowercased field names (which is not as nice). I'll add tags here.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+			cancelFn()
+		}
+		if cancelFnID > 0 {
+			m.unregisterCancelFn(cancelFnID)
+		}
+
+		pid, kv := m.getPrimaryClient()
+
+		var cancelCtx context.Context
+		// This only loops if switchover to a new primary backend happens, which is very rare.
```
pracucci: Is this the right place for this comment? I would have expected to see it right before the `for {`.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+
+	delete(m.inProgress, id)
+}
+
+// Runs supplied fn with current primary client. If primary client changes, fn is restarted.
+// When fn finishes (with or without error), this method returns given error value.
+func (m *multiClient) runWithPrimaryClient(origCtx context.Context, fn func(newCtx context.Context, primary kvclient) error) error {
+	cancelFn := context.CancelFunc(nil)
+	cancelFnID := 0
+	defer func() {
```
pracucci: Since this logic is also within the loop, you may define a cleanup function and call it from both places, something like:

```
// Call it to ensure context resources are correctly released
func cleanup() {
		if cancelFn != nil {
			cancelFn()
		}
		if cancelFnID > 0 {
			m.unregisterCancelFn(cancelFnID)
		}
	}

	defer cleanup()

	for {
		cleanup()
```

pkg/cortex/modules.go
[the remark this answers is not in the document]
pstibrany: So I've tried to use interface, but that interface would have single function -- to get channel. So now I'm passing that function instead. It's still in the struct though, because MultiClient config should not leak to ring.New API. Take a look and tell me what you think.

pkg/util/override.go
```diff
@@ -82,26 +100,35 @@ func (om *OverridesManager) loadOverrides() error {
 	}
 	overridesReloadSuccess.Set(1)
 
+	om.setOverrides(overrides)
+	om.callListeners(overrides)
```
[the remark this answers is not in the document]
pstibrany: Yes, there is some low-hanging fruit here. I didn't want to change OverridesManager too much from it's existing version, as I wanted to get some feedback on the whole 'runtime configuration' idea.

However, even if we add detection of file change, we may call listeners that are only interested in part of the configuration that hasn't changed. So listeners must be prepared for that situation.

pkg/ring/kv/client.go
```diff
@@ -11,19 +11,29 @@ import (
...
+}
+
 // Config is config for a KVStore currently used by ring and HA tracker,
 // where store can be consul or inmemory.
 type Config struct {
-	Store  string        `yaml:"store,omitempty"`
-	Consul consul.Config `yaml:"consul,omitempty"`
-	Etcd   etcd.Config   `yaml:"etcd,omitempty"`
-	Prefix string        `yaml:"prefix,omitempty"`
+	StoreConfig
```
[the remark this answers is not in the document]
pstibrany: I certainly hope not! But I better verify that. Thanks!

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// MultiRuntimeConfig has values that can change in runtime (via overrides)
+type MultiRuntimeConfig struct {
+	// Primary store used by multiClient. Can be updated in runtime to switch to a different store (eg. consul -> etcd,
+	// or to gossip). Doing this allows nice migration between stores. Empty values are ignored.
+	PrimaryStore string `yaml:"primary"`
+
+	// Put "enabled", "enable" or "true" here to enable. Any other value = disabled. Empty = ignore.
+	Mirroring string `yaml:"mirroring"`
```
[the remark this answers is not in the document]
pstibrany: I've added comment here now. Reason is that if that boolean was false, I wouldn't be able to distinguish between value missing in yaml file (in which case I don't want to change mirroring status), or value set to false (disabled).

pkg/ring/kv/client.go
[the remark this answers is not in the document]
pracucci: I thought you had to use the following, but I may be wrong:

```
StoreConfig `yaml:",inline"`
```

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+
+	MirrorRateLimit float64
+	MirrorRateBurst int
+	MirrorEnabled   bool
+	MirrorPrefix    string
+
+	// Channel used by multiClient to get runtime config updates
+	RuntimeConfigChannel <-chan MultiRuntimeConfig
+
+	clients []kvclient
```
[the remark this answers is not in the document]
pstibrany: You're right, that will be nicer.

pkg/ring/kv/multi.go
pstibrany: Fair enough :)

pkg/ring/kv/multi.go
pstibrany: You're 100% right. I was just looking at it and thinking the same.

pkg/ring/kv/client.go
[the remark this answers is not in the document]
pstibrany: You're correct! It broke the Yaml marshalling :-( Thanks' for catching this... I better add a test for this!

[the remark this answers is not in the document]
pstibrany: Fixed, with a test.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,335 @@
...
+	f.StringVar(&cfg.Primary, prefix+"multi.primary", "", "Primary backend storage used by multi-client.")
+	f.StringVar(&cfg.Secondary, prefix+"multi.secondary", "", "Secondary backend storage used by multi-client.")
+	f.StringVar(&cfg.MirrorPrefix, prefix+"multi.mirror-prefix", "", "Prefix for keys that should be mirrored")
+	f.BoolVar(&cfg.MirrorEnabled, prefix+"multi.mirror-enabled", false, "Mirror values to secondary store")
+	f.Float64Var(&cfg.MirrorRateLimit, prefix+"multi.mirror-rate-limit", 1, "Rate limit used for mirroring. 0 disables rate limit.")
+	f.IntVar(&cfg.MirrorRateBurst, prefix+"multi.mirror-burst-size", 1, "Burst size used by rate limiter. Values less than 1 are treated as 1.")
+}
+
+// MultiRuntimeConfig has values that can change in runtime (via overrides)
+type MultiRuntimeConfig struct {
```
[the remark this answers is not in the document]
pstibrany: I haven't noticed Validate before, thanks. I don't see much point in moving it there, as it would be the same code, just at a different place. If you insist, I don't mind, but I fail to see the value of it.

pkg/ring/kv/multi.go
pstibrany: I've renamed method to 'stop' and removed TODO. We be handled later, when merging with PR 1727. Compilation error (eg. missing Stop method when merging 1727) will make sure it's not forgotten.

pkg/ring/kv/multi.go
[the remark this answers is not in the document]
pracucci: Oh, this is not part of the standard config (it's in overrides). Please ignore my comment, and sorry for the noise.

pkg/ring/kv/multi.go
jtlisi: 👍

pstibrany:
Rebased on top of master, and squashed to single commit, to ease possible future rebases.

pkg/util/validation/limits.go
```diff
@@ -114,184 +111,153 @@ func (l *Limits) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	defaultLimits *Limits
+	tenantLimits  TenantLimits
 }
 
 // NewOverrides makes a new Overrides.
 // We store the supplied limits in a global variable to ensure per-tenant limits
 // are defaulted to those values.  As such, the last call to NewOverrides will
 // become the new global defaults.
-func NewOverrides(defaults Limits) (*Overrides, error) {
-	defaultLimits = &defaults
```
sandlis: This needs to be retained since it sets the default values to the global variable `defaultLimits` without which the limits which are not set in overrides would be set to zero values. Try removing line https://github.com/cortexproject/cortex/blob/d6edf4b2cacf8a19e026b8d7e86144491143334a/pkg/util/override_test.go#L101 if you want to see the impact.

Maybe to avoid this from happening again, do you want to add an end to end test somewhere which uses the same global variable from this package?

pstibrany: Oh, you're correct! I broke it :-( Fixing now. I'll add a test for this, to make sure it doesn't get broken again.

pstibrany: Thanks for catching this!

pstibrany: Done.

pstibrany: Thinking about it... perhaps we should just have SetDefaultLimits method in that package instead?

pstibrany: I've moved it to separate function now, called `SetDefaultLimitsForYAMLUnmarshalling`.

sandlis: I think it would be better calling `SetDefaultLimitsForYAMLUnmarshalling` from `limits.NewOverrides` itself. It would be easy to miss calling it.

pstibrany: I prefer code to be explicit, rather than relying on hidden side-effects like that. It is only used in single place in Cortex (not counting tests, which don't rely on it) and I've added call to that place. Of course, we need to add that to Loki, when it is updated to use this code, but there are more changes related to this PR that will be required in Loki.

pkg/util/override.go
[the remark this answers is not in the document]
pstibrany: I'll resolve this. I agree with you, but would prefer to do more changes to OverridesManager in a different PR, as this PR is about something else.

pkg/ring/kv/multi.go
pstibrany: With gossiping code in master now, I've rebased and added this code.

pkg/util/validation/limits.go
pracucci: My perspective:
- Looking at the func name `NewOverrides` I would expect it just create new overrides, without touching the global variable storing default limits. For this reason, I would suggest to be explicit and call the func to set default limits outside `NewOverrides`
- The name `SetDefaultLimitsForYAMLUnmarshalling()` looks a bit weird to me, because it doesn't have logic which is specific of the fact that the default limits come from YAML unmarshalling. I would simply call it `SetDefaultLimits()` and then explain - with a code comment in `modules.go` - that we're calling it to set default limits from value read from YAML

pkg/ring/kv/multi.go
```diff
@@ -316,9 +316,13 @@ func (m *MultiClient) keyUpdated(primary kvclient, key string, newValue interfac
...
 }
 
-// Stop the multiClient (mirroring part).
-func (m *MultiClient) stop() {
+// Stop the multiClient (mirroring part), and all configured clients.
+func (m *MultiClient) Stop() {
 	m.cancel()
+
+	for _, kv := range m.clients {
+		kv.client.Stop()
```
pracucci: Consul and etcd `Stop()` are no-op, so who is using it right now? Has this been introduced because of the Gossip client? If so, do you need to wait before the `cancel()` takes effect (the on-going operations have been cancelled) or not necessary?

pkg/ring/kv/client.go
```diff
@@ -20,14 +20,22 @@ import (
...
+}
+
+// Config is config for a KVStore currently used by ring and HA tracker,
+// where store can be consul or inmemory.
+type Config struct {
+	Store       string `yaml:"store,omitempty"`
+	Prefix      string `yaml:"prefix,omitempty"`
+	StoreConfig `yaml:",inline"`
 
 	Mock Client
```
pracucci: I known you didn't add it, but this should have `yaml:"-"`, to exclude it from any marshalling/unmarshalling.

pkg/ring/kv/multi.go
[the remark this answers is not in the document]
pracucci: Have you investigated if you could do the same with a `*bool` data type? Should be `nil` if not explicitly set. I'm not a big fan of pointers in the config, but I've the feeling that enabling it via a string is a sort of new pattern to Cortex.

pkg/util/validation/limits.go
pstibrany: > * The name `SetDefaultLimitsForYAMLUnmarshalling()` looks a bit weird to me

That was my goal... to make it look weird to make sure people take a deeper look before using/changing it. It's not setting default limits (which `SetDefaultLimits` would suggest) -- default limits are passed to the `NewOverrides` as a first argument. This function is setting default limits that are ONLY USED during YAML unmarshal, nowhere else. I think calling it just `SetDefaultLimits` would be misleading.

pkg/ring/kv/multi.go
pstibrany: This is a new method introduced by Gossiping work. It's no-op in Consul and Etcd (although they could do some cleanup, if necessary), while memberlist client uses it to leave the cluster gracefully... it is especially useful when ingester or distributor wants to remove itself from the ring, before leaving -- this gives gossiping client a chance to send it out.

Cancel only stops the mirroring part [update: and now also watching of the config channel]. It doesn't cancel ongoing operations.

pkg/ring/kv/client.go
pstibrany: Done.

pkg/cortex/runtime_config.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/cortexproject/cortex/pkg/util/validation"
+)
+
+// runtimeConfigValues are values that can be reloaded from configuration file while Cortex is running.
+// Reloading is done by OverridesManager, which also keeps currently loaded config.
+// These values are then pushed to the components that are interested in them.
+type runtimeConfigValues struct {
+	TenantLimits map[string]*validation.Limits `yaml:"overrides"`
+
+	Multi kv.MultiRuntimeConfig `yaml:"multi"`
```
gouthamve: `yaml:"multi_kv_config"`?

pkg/ring/kv/client_test.go
```diff
@@ -0,0 +1,31 @@
...
+	cfg := Config{}
+
+	err := yaml.Unmarshal([]byte(conf), &cfg)
+	require.NoError(t, err)
+	require.Equal(t, "consul", cfg.Store)
+	require.Equal(t, "test/", cfg.Prefix)
+	require.Equal(t, "consul:8500", cfg.Consul.Host)
+	require.Equal(t, "consul", cfg.Multi.Primary)
+	require.Equal(t, "etcd", cfg.Multi.Secondary)
+}
```
gouthamve: Could we have a couple of tests for bad config too? Like not setting secondary, setting both to the same thing etc.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,349 @@
...
+
+func (m *MultiClient) watchConfigChannel(ctx context.Context, configChannel <-chan MultiRuntimeConfig) {
+	for {
+		select {
+		case cfg, ok := <-configChannel:
+			if !ok {
+				return
+			}
+
+			if cfg.Mirroring != nil {
```
gouthamve: I would make it a simple bool, not `*bool`. The Swap would be a no-op then.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,349 @@
...
+				enable := *cfg.Mirroring
+				old := m.mirroringEnabled.Swap(enable)
+				if old != enable {
+					level.Info(util.Logger).Log("msg", "toggled mirroring", "enabled", enable)
+				}
+			}
+
+			if cfg.PrimaryStore != "" {
```
gouthamve: This should never happen. Maybe log an error?

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,341 @@
...
+func (m *MultiClient) unregisterCancelFn(id int) {
+	m.inProgressMu.Lock()
+	defer m.inProgressMu.Unlock()
+
+	delete(m.inProgress, id)
+}
+
+// Runs supplied fn with current primary client. If primary client changes, fn is restarted.
+// When fn finishes (with or without error), this method returns given error value.
+func (m *MultiClient) runWithPrimaryClient(origCtx context.Context, fn func(newCtx context.Context, primary kvclient) error) error {
```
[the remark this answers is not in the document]
pstibrany: `WatchKey` and `WatchPrefix` operations are long-running, basically designed to never return (eg. see usage in pkg/ring/ring.go:155 or pkg/distributor/ha_tracker.go:144). So we need to either preserve that behaviour (what this PR is trying to do), or modify clients to restart if it returns early.

`Get` and `CAS` are short operations, and switching midway isn't necessary there.

pkg/util/validation/limits.go
pracucci: Ok, thanks for the explanation. I'm good with the current naming and design then.

pkg/ring/kv/multi.go
```diff
@@ -43,10 +43,8 @@ type MultiRuntimeConfig struct {
 	// or to gossip). Doing this allows nice migration between stores. Empty values are ignored.
 	PrimaryStore string `yaml:"primary"`
 
-	// Put "enabled", "enable" or "true" here to enable. Any other value = disabled. Empty = ignore.
-	// We don't use bool here, because it would not be possible to distinguish between false = missing value,
-	// or false = disabled.
-	Mirroring string `yaml:"mirroring"`
+	// Mirroring enabled or not. Nil = no change.
+	Mirroring *bool `yaml:"mirroring"`
```
pracucci: Naming wise, shouldn't this reflect `MultiConfig`, so naming it `MirrorEnabled` / ``yaml:"mirror-enabled"`?

pkg/ring/kv/multi.go
pracucci: If it's a `bool` there's no way to distinguish it between the default value (`false`) and whether has not been set. It's correct it's a `*bool` from my perspective.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,341 @@
...
+// WatchPrefix is a part of kv.Client interface
+func (m *MultiClient) WatchPrefix(ctx context.Context, prefix string, f func(string, interface{}) bool) {
+	_ = m.runWithPrimaryClient(ctx, func(newCtx context.Context, primary kvclient) error {
+		primary.client.WatchPrefix(newCtx, prefix, f)
+		return newCtx.Err()
+	})
+}
+
+// watchChanges performs mirroring -- it watches changes in the primary client, and puts them into secondary.
+func (m *MultiClient) watchChanges(prefix string) {
```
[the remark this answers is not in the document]
pstibrany: I think it would achieve the same, at the cost of making CAS slower for clients. But that is probably just fine. (There is question of what to do when CAS to primary succeeds, but CAS to secondary fails, but the only reasonable thing we can do is to ignore it... we should not call `f` again)

pkg/ring/kv/client_test.go
pstibrany: This test doesn't actually run any new code, it is just doing YAML parsing. Reason is that by extracting `StoreConfig` from `Config` struct, I broke YAML parsing of existing configurations, and wanted to make sure it's ok now. I can add tests for config verification as well, but it seems like overkill.

pkg/ring/kv/multi.go
pstibrany: If it is just `bool`, then you cannot distinguish between value missing from overrides.yaml (in which case, Mirroring should not change its 'enabled' status) or value being `false` (which should disable mirroring). Hence the `*bool` and unit test verifying that `*bool` is parsed properly in each of these three states (missing, true, false).

pkg/ring/kv/multi.go
pstibrany: It will happen if primary store is not defined in overrides.yaml. This entire method is dealing with values loaded from overrides.

pstibrany:
> So my big comment would be around simplifying the entire thing to be: Write to both, read from primary. Also, forget about inflight ops, just do it on the primary client when we initiated the call.

I've changed functionality to follow this logic: writes via CAS function are now mirrored (if enabled) to secondary store, Get always goes to primary. WatchKey and WatchPrefix are left as they were [https://github.com/cortexproject/cortex/pull/1749#discussion_r342954874 for explanation].

I've removed mirroring goroutine and rate limiting, and added metrics. I've also introduced new config value, MirrorTimeout, used when forwarding write to secondary store. (I've observed very long delays when doing Consul -> Etcd mirroring and Etcd was down -- over minute, and ingester got to Unhealthy state.)

pkg/util/override.go
```diff
@@ -1,10 +1,9 @@
-package validation
+package util
```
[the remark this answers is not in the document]
pstibrany: Two reasons:
1) it has nothing to do with validation, and is more generic type. Specifically, I wanted to avoid any references to Limit from OverridesManager.
2) Since this changes functionality of this type, I wanted to break existing client code, to make sure it is updated.

pstibrany:
> LGTM
> 
> The KV code was much clearer than the last time I reviewed this PR. The logic around the KV store looks good and has good associated metrics/logs.
> 
> I want to note I would prefer if the OverridesManager and associated code lived in it's own package. We don't have a strong style guide around these types of things as a project. However, since a significant part of Cortex is going to be consumed by downstream projects I think it is important to maintain well named, clearly defined packages throughout the code base.

Thanks @jtlisi for your review.

"much cleaner" -- I think I mostly removed code from KV code since your last review. It's now not doing the "mirroring" as a background task, but updates values as part of CAS.

re OverridesManager: I've now moved it and renamed to runtime_config.Manager. Originally I didn't want to do big changes to it, as this PR is primarily about something else, but I also think this is a good change.

pkg/util/override.go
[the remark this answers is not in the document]
pstibrany: Now it lives in pkg/util/runtime_config/manager.go (as runtime_config.Manager type).

pkg/cortex/cortex.go
```diff
@@ -78,6 +79,9 @@ type Config struct {
 	ConfigDB     db.Config                                  `yaml:"configdb,omitempty"`
 	ConfigStore  config_client.Config                       `yaml:"config_store,omitempty"`
 	Alertmanager alertmanager.MultitenantAlertmanagerConfig `yaml:"alertmanager,omitempty"`
+
+	RuntimeConfigFile       string        `yaml:"runtime_config_file"`
```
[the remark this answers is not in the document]
pstibrany: Please take a look and latest commit.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,357 @@
...
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/uber-go/atomic"
+
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+var (
+	primaryStoreGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "consul_multikv_primary_store",
```
pracucci: It sounds a bit weird prefixing these metrics with `consul_`. What's the rationale behind it?

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,357 @@
...
+func init() {
+	prometheus.MustRegister(primaryStoreGauge, mirrorEnabled, mirrorWrites, mirrorFailures)
+}
+
+// MultiConfig is a configuration for MultiClient.
+type MultiConfig struct {
+	Primary   string `yaml:"primary"`
+	Secondary string `yaml:"secondary"`
+
+	MirrorEnabled bool          `yaml:"mirror-enabled"`
```
pracucci: To keep consistency, YAML config name should have `_` instead of `-` (same comment applies to `MirrorTimeout`)

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// MultiRuntimeConfig has values that can change in runtime (via overrides)
+type MultiRuntimeConfig struct {
+	// Primary store used by MultiClient. Can be updated in runtime to switch to a different store (eg. consul -> etcd,
+	// or to gossip). Doing this allows nice migration between stores. Empty values are ignored.
+	PrimaryStore string `yaml:"primary"`
+
+	// Mirroring enabled or not. Nil = no change.
+	Mirroring *bool `yaml:"mirror-enabled"`
```
pracucci: `mirror-enabled` > `mirror_enabled`

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,357 @@
...
+func (m *MultiClient) watchConfigChannel(ctx context.Context, configChannel <-chan MultiRuntimeConfig) {
+	for {
+		select {
+		case cfg, ok := <-configChannel:
+			if !ok {
+				return
+			}
+
+			if cfg.Mirroring != nil {
+				enable := *cfg.Mirroring
```
pracucci: Maybe `enable` > `enabled` is more accurate

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,357 @@
...
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+var (
+	primaryStoreGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "consul_multikv_primary_store",
+		Help: "Selected primary KV store",
+	}, []string{"store"})
+
+	mirrorEnabled = prometheus.NewGauge(prometheus.GaugeOpts{
```
pracucci: `primaryStoreGauge` ends with `Gauge`. I personally find it more clear when reading the code if you add `Gauge` and `Counter` suffix to the other metrics too.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,357 @@
...
+
+		return err
+	}
+}
+
+// Get is a part of kv.Client interface.
+func (m *MultiClient) Get(ctx context.Context, key string) (interface{}, error) {
+	_, kv := m.getPrimaryClient()
+	val, err := kv.client.Get(ctx, key)
+	level.Info(util.Logger).Log("key", key, "value", val, "store", kv.name, "err", err)
```
pracucci: Do we wanna keep this log?

pkg/ring/kv/multi.go
pstibrany: Ehm, that was supposed to be `cortex` :)

pstibrany:
I've added two small changes:
1) when server starts, we reload overrides after setting YAML defaults for config. Otherwise they would only be applied after next reload. This is to fix bug observed in production, when in first 10 seconds of ingester running, incorrect limits were applied.

2) Communicate configuration updates via channels: runtimeconfig.Manager no longer spawns new goroutine to tell Listener about new configuration, but sends value to channel. If there is no receiver, or channel buffer is full, update is discarded.

pstibrany:
TODO:
- changelog
- document change in naming of arguments [old arguments still supported]

pstibrany:
Updated CHANGELOG.md and documented runtime config section and updated format of overrides/runtime config  yaml file.

pkg/ring/kv/multi.go
```diff
@@ -0,0 +1,356 @@
...
+		case cfg, ok := <-configChannel:
+			if !ok {
+				return
+			}
+
+			if cfg.Mirroring != nil {
+				enabled := *cfg.Mirroring
+				old := m.mirroringEnabled.Swap(enabled)
+				if old != enabled {
+					level.Info(util.Logger).Log("msg", "toggled mirroring", "enabled", enabled)
```
gouthamve: Can we be more descriptive in the log here. Maybe add the component (multi-kv)?

pkg/util/runtimeconfig/manager.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	if cfg.LoadPath != "" {
+		if err := mgr.loadConfig(); err != nil {
+			// Log but don't stop on error - we don't want to halt all ingesters because of a typo
+			level.Error(util.Logger).Log("msg", "failed to load config", "err", err)
+		}
+		go mgr.loop()
+	} else {
+		level.Info(util.Logger).Log("msg", "config disabled")
```
gouthamve: rutime config disabled.

docs/arguments.md
```diff
@@ -177,6 +177,32 @@ Flags for configuring KV store based on memberlist library. This feature is expe
...
    Log debug transport messages. Note: global log.level must be at debug level as well.
+   
+#### Multi KV
+
+This is a special key-value implementation that uses two other KV stores (eg. consul, etcd or memberlist). One of them is always marked as primary, and all reads and writes go to primary store. Other one, secondary, is only used for writes. The idea is that operator can use multi KV store to migrate from primary to secondary store in runtime. For example, migration from Consul to Memberlist would look like this:
+
+- Set `ring.store` to use `multi` store. Set `-multi.primary=consul` and `-multi.secondary=memberlist`. All consul and memberlist settings must still be specified.
+- Start all Cortex microservices. They will still use Consul as primary KV, but they will also write share ring via memberlist.
+- Operator can now use "runtime config" mechanism to switch primary store to memberlist.
+- After all Cortex microservices have picked up new primary store, and everything looks correct, operator can now shut down Consul, and modify Cortex configuration to use `-ring.store=memberlist` only.
```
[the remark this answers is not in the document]
pracucci: Why don't we generally refer to the key-store instead of naming consul/etcd? It may look misleading that it only apply to migrating from consul to memberlist/etcd, while you can migrate from any store to any other store, right?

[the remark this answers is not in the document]
pstibrany: I like to be specific in the example, and it also starts with "For example, migration from Consul to memberlist would look like this:". I'll modify it to use Consul -> Etcd as an example.

pkg/util/runtimeconfig/manager.go
pstibrany: what about `"runtime config file not specified, reload disabled"` to explain "why" as well.

pstibrany: or rather `runtime config disabled: file not specified`

pkg/ring/kv/multi.go
pstibrany: I've added logger with component=multikv that is used for all logging from this file.

PR #1750, opened 2019-10-23
Flush tokens to disk

codesome:
This PR implements flushing of tokens to the disk and reading them back again when a pod is restarted/upgraded. This goes with [this design](https://docs.google.com/document/d/1n1HcdgmsqaVwqxKg_POZqIREOSqZIAlrxuyx4YYV_pA/edit#heading=h.58ivdnk96o85) of retaining tokens required for WAL in ingesters.


codesome:
And it is ready for review

pkg/ingester/tokens
```diff
@@ -0,0 +1 @@
+�\�
```
pracucci: Is this file expected to be committed?

pkg/ring/lifecycler.go
```diff
@@ -52,6 +60,7 @@ type LifecyclerConfig struct {
 	NormaliseTokens  bool          `yaml:"normalise_tokens,omitempty"`
 	InfNames         []string      `yaml:"interface_names"`
 	FinalSleep       time.Duration `yaml:"final_sleep"`
+	TokenFileDir     string        `yaml:"token_file_dir,omitempty"`
```
pracucci: Why not `TokensFilePath`? We always create only 1 file (called `tokens`) inside the dir, which looks more natural to allow to configure the full filepath.

BTW, I would call it `TokensFile...` instead of `TokenFile...` since it contains all tokens.

pkg/ring/lifecycler.go
```diff
@@ -242,12 +252,59 @@ func (i *Lifecycler) getTokens() []uint32 {
...
+	return tokens, nil
+}
+
 func (i *Lifecycler) setTokens(tokens []uint32) {
 	tokensOwned.WithLabelValues(i.RingName).Set(float64(len(tokens)))
 
 	i.stateMtx.Lock()
 	defer i.stateMtx.Unlock()
 	i.tokens = tokens
+	i.flushTokensToFile()
```
pracucci: Instead of loking while flushing tokens to disk, you may pass `tokens` to `flushTokensToFile()`, like this:

```
i.stateMtx.Lock()
i.tokens = tokens
i.stateMtx.Unlock()

i.flushTokensToFile(tokens)
```

pkg/ruler/tokens
```diff
@@ -0,0 +1 @@
+��z�
```
pracucci: Is this file expected to be committed? (same for `pkg/ring/tokens`)

pkg/ring/lifecycler.go
```diff
@@ -242,12 +252,59 @@ func (i *Lifecycler) getTokens() []uint32 {
...
+		return
+	}
+	tokenFilePath := path.Join(i.cfg.TokenFileDir, tokensFileName)
+	f, err := os.OpenFile(tokenFilePath, os.O_WRONLY|os.O_CREATE, 0666)
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "error in creating token file", "err", err)
+		return
+	}
+
+	b := make([]byte, 4*len(i.tokens))
```
pracucci: I would decouple the tokens storage format from the ring lifecycler logic. I would suggest to create a dedicated `type` for the tokens with serialization and deserialization functions. I would also suggest to add a version number to the format, to keep the door open to future changes.

pkg/ingester/tokens
codesome: Definitely not. Forgot to push a follow-up cleanup PR.

pkg/ring/lifecycler.go
codesome: Yup makes sense. I had thought with the lock we can assure that tokens were into the disk. But I guess we don't need that guarantee.

pkg/ring/lifecycler.go
codesome: Do you mean the entire path including the file name?

pracucci: Yes. If it's going to be just 1 file, configuring the whole path looks just more clear to me.

codesome: I feel leaving the file name to the user would lead to problems or confusion (if names are changed). By only taking the directory we can be consistent with the file name.

pracucci: Ok. What's about calling it `TokensFileDir`?

pkg/ring/lifecycler.go
codesome: If we were to have versions for tokens here, would we also need to store the version in the ring too?

codesome: But I don't see how we will have different versions of tokens in the future.

pkg/ring/lifecycler.go
codesome: Sure. Will change.

pkg/ring/lifecycler.go
pracucci: The comment was about two distinct things:
1. I would suggest to create a dedicated type for the tokens with serialization and deserialization functions
2. I would also suggest to add a version number to the format, to keep the door open to future changes.

I'm still keen on the comment #1. It helps to keep a separation of concerns and the code cleaner and easier to unit test.

About #2 I see your point. My point is that it's hard to foresee the future, and adding a byte with a version number at the beginning of the file looks a minimal effort compared to the case in the future we'll have to change the format (ie. add other data) and versioning will be handy.

codesome: Will add the point 1. About point 2, I will add the type part in this PR to store in the disk but we might need to change it in the ring too which can be a follow-up PR with some migration path.

pracucci: About point 2, ask other people feedback, please. Don't want to be pushy on it. Realistically, there may not be that need and may be not worth.

pstibrany: > But I don't see how we will have different versions of tokens in the future.

Not sure if that would need to be stored, but #1764 is introducing State to the tokens.

If you need extensible serialization format, perhaps just use protobuf.

codesome: Added an interface to handle tokens https://github.com/cortexproject/cortex/pull/1750/commits/7ab02ffe583ea85795353e8db0909d68b9794378. Might be an initial iteration towards the final version, is it what is intended? I tried to have `type Token uint32` but that would have led to a lot of changes and I would like to keep that for a separate PR in itself.

pkg/ring/lifecycler.go
```diff
@@ -687,3 +752,38 @@ func (i *Lifecycler) unregister(ctx context.Context) error {
...
 }
+
+// TokenVersion1 is the version is a simple list of tokens.
+const TokenVersion1 = 1
+
+// Tokens is a simple list of tokens.
+type Tokens []uint32
+
+// MarshalTokens encodes given tokens into JSON.
+func MarshalTokens(tokens Tokens) ([]byte, error) {
```
pracucci: Any reason why `MarshalTokens` and `UnmarshalTokens` are not functions of the `Tokens` type (in that case should be just named `Marshal` and `Unmarshal`? Looking at the code base looks a common pattern.

pkg/ring/lifecycler.go
pracucci: I left my feedback here:
https://github.com/cortexproject/cortex/pull/1750#pullrequestreview-311208119

codesome:
> Can you rebase with `master`? I think there have been several changes related to the ring in the meanwhile.

I had already rebased with `master` to include the ring changes. Rebased again in case something was left.

> Please add unit tests for your logic (serialization / deserialization / changes to lifecycler)

I had added `TestTokensOnDisk`. Any case that is missing in that?

> Actually it's what I see you did.

Yeah. I made some changes following @gouthamve's suggestion after I put that comment :)  

Regarding persisting the lifecycler state, that would include more than just the tokens. I would prefer restoring other states (like the `IngesterState`) by looking at the ring and not the file and only persist the tokens. The idea was to only keep track of the tokens and the rest to be taken care of as usual. WDYT?

pkg/ring/lifecycler.go
codesome: Looks like it can be a part of `Tokens`, will update.

pkg/ingester/lifecycle_test.go
```diff
@@ -94,11 +95,17 @@ func TestIngesterTransfer(t *testing.T) {
 	limits, err := validation.NewOverrides(defaultLimitsTestConfig())
 	require.NoError(t, err)
 
+	tokenDir1, err := ioutil.TempDir(os.TempDir(), "ingester_transfer")
```
gouthamve: Please clean these up in a defer please.

pkg/ruler/tokens
gouthamve: I don't this is needed anymore?

pkg/ring/tokens.go
```diff
@@ -0,0 +1,83 @@
...
+	"os"
+	"path"
+)
+
+const (
+	// TokenVersion1 is the version is a simple list of tokens.
+	TokenVersion1 = 1
+
+	// Name of the file in which tokens is stored.
+	tokensFileName = "tokens"
```
gouthamve: Can we make it take a full path rather than have a directory and then hardcode the file name?

pkg/ring/lifecycler.go
```diff
@@ -84,6 +85,7 @@ func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.Flag
 	flagext.DeprecatedFlag(f, prefix+"claim-on-rollout", "DEPRECATED. This feature is no longer optional.")
 	f.BoolVar(&cfg.NormaliseTokens, prefix+"normalise-tokens", false, "Store tokens in a normalised fashion to reduce allocations.")
 	f.DurationVar(&cfg.FinalSleep, prefix+"final-sleep", 30*time.Second, "Duration to sleep for before exiting, to ensure metrics are scraped.")
+	f.StringVar(&cfg.TokensFileDir, prefix+"tokens-file-dir", "", "Directory in which the tokens file is stored. If empty, tokens are not stored at shutdown and restored at startup.")
```
gouthamve: Take a filepath instead of a directory.

pkg/ring/model.go
```diff
@@ -79,14 +79,14 @@ func (d *Desc) RemoveIngester(id string) {
...
-	var result []uint32
+func (d *Desc) ClaimTokens(from, to string, normaliseTokens bool) Tokens {
+	result := Tokens{}
 
 	if normaliseTokens {
 
 		// If the ingester we are claiming from is normalising, get its tokens then erase them from the ring.
 		if fromDesc, found := d.Ingesters[from]; found {
-			result = fromDesc.Tokens
+			result = append(result, fromDesc.Tokens...)
```
gouthamve: Why are we not doing `result = Tokens(fromDesc.Tokens)`? That doesn't allocate, but I think this does.

pkg/ring/model.go
```diff
@@ -111,7 +111,7 @@ func (d *Desc) ClaimTokens(from, to string, normaliseTokens bool) []uint32 {
 	} else {
 		// If source ingester is normalising, copy its tokens to d.Tokens, and set new owner
 		if fromDesc, found := d.Ingesters[from]; found {
-			result = fromDesc.Tokens
+			result = append(result, fromDesc.Tokens...)
```
gouthamve: Same here.

pkg/ring/tokens.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// Tokens is a simple list of tokens.
+type Tokens []uint32
+
+// StoreToFile stores the tokens in the given directory.
+func (t Tokens) StoreToFile(dir string) error {
```
gouthamve: Always write to a temp file and then do a `mv`.

pkg/ring/tokens.go
codesome: @pracucci had the same suggestion. [this](https://github.com/cortexproject/cortex/pull/1750#discussion_r341284749) was my comment for it. Does it make sense?

pkg/ring/lifecycler.go
codesome: https://github.com/cortexproject/cortex/pull/1750#discussion_r344639731

pkg/ring/model.go
codesome: Ah, this is something left behind with so many design changes in between. Will clean this up.

pkg/ring/lifecycler.go
codesome: With some discussions offline with @gouthamve, filepath wins

pkg/ruler/tokens
codesome: Looks like this is being created in some test. Will find it out and use a temporary dir instead.

pkg/ring/lifecycler.go
```diff
@@ -246,18 +248,25 @@ func (i *Lifecycler) ChangeState(ctx context.Context, state IngesterState) error
 	return <-err
 }
 
-func (i *Lifecycler) getTokens() []uint32 {
+func (i *Lifecycler) getTokens() Tokens {
 	i.stateMtx.Lock()
 	defer i.stateMtx.Unlock()
+	if i.tokens == nil {
```
pracucci: I'm missing the need of this `if`. Isn't `return i.tokens` enough?

pkg/ring/lifecycler.go
```diff
@@ -246,18 +248,25 @@ func (i *Lifecycler) ChangeState(ctx context.Context, state IngesterState) error
...
-func (i *Lifecycler) setTokens(tokens []uint32) {
+func (i *Lifecycler) setTokens(tokens Tokens) {
 	tokensOwned.WithLabelValues(i.RingName).Set(float64(len(tokens)))
 
 	i.stateMtx.Lock()
-	defer i.stateMtx.Unlock()
 	i.tokens = tokens
+	i.stateMtx.Unlock()
+
+	if err := i.tokens.StoreToFile(i.cfg.TokensFilePath); err != nil {
```
pracucci: I'm not sure it's a good idea accessing `i.tokens` outside the lock. Is it safe if we call `tokens.StoreToFile()` instead of `i.tokens.StoreToFile()`?

pkg/ring/lifecycler.go
```diff
@@ -246,18 +248,25 @@ func (i *Lifecycler) ChangeState(ctx context.Context, state IngesterState) error
...
+		level.Error(util.Logger).Log("msg", "error storing tokens to disk", "err", err)
```
pracucci: We should add the filepath here to help people debug the issue.

pkg/ring/model_test.go
```diff
@@ -104,23 +104,23 @@ func TestClaimTokensFromNormalizedToNormalized(t *testing.T) {
 	r := normalizedSource()
 	result := r.ClaimTokens("first", "second", true)
 
-	assert.Equal(t, []uint32{100, 200, 300}, result)
+	assert.Equal(t, []uint32{100, 200, 300}, []uint32(result))
```
pracucci: Why don't we use `Tokens` data type, like:

```
assert.Equal(t, Tokens{100, 200, 300}, result)
```

_Same comment applies to the rest of the file._

pkg/ring/tokens.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Tokens is a simple list of tokens.
+type Tokens []uint32
+
+// StoreToFile stores the tokens in the given directory.
+func (t Tokens) StoreToFile(tokenFilePath string) error {
+	if tokenFilePath == "" {
+		return errors.New("path is empty")
+	}
+	tmp := tokenFilePath + ".tmp"
```
pracucci: This is not safe if `StoreToFile()` is called concurrently (we will have two go routines writing to the same `.tmp` file). You should use https://golang.org/pkg/io/ioutil/#TempFile

pkg/ring/tokens.go
```diff
@@ -0,0 +1,95 @@
...
+
+	b, err := t.Marshal()
+	if err != nil {
+		return err
+	}
+	if _, err = f.Write(b); err != nil {
+		return err
+	}
+
+	if err := f.Close(); err != nil {
```
pracucci: Shouldn't we `f.Close()` even before any `return err` above?

pkg/ring/tokens.go
```diff
@@ -0,0 +1,95 @@
...
+	"os"
+
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/prometheus/tsdb/fileutil"
+)
+
+const (
+	// TokenVersion1 is the version is a simple list of tokens.
+	TokenVersion1 = 1
```
pracucci: `TokenVersion1` > `TokensVersion1`

pkg/ring/tokens.go
```diff
@@ -0,0 +1,95 @@
...
+	tj := tokensJSON{}
+	if err := json.Unmarshal(b, &tj); err != nil {
+		return err
+	}
+	switch tj.Version {
+	case TokenVersion1:
+		*t = Tokens(tj.Tokens)
+		return nil
+	default:
+		return errors.New("invalid token type")
```
pracucci: `invalid token type` > `invalid token version`. Moreover you should also add the version load from file for reference in the error.

pkg/ring/tokens.go
```diff
@@ -0,0 +1,95 @@
...
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+	return t.Unmarshal(b)
+}
+
+// Marshal encodes the tokens into JSON.
+func (t Tokens) Marshal() ([]byte, error) {
```
pracucci: I would suggest to add a serialization test. You can use `output = Unmarshal(Marshal(input))` (pseudocode) and make sure that `output` is equal `input`. It's generally a simple yet useful unit test to add whenever you work with serialization.

pkg/ring/lifecycler_test.go
```diff
@@ -246,17 +256,90 @@ func TestCheckReady(t *testing.T) {
...
+	r, err := New(ringConfig, "ingester")
+	require.NoError(t, err)
+	defer r.Stop()
+
+	lifecyclerConfig := testLifecyclerConfig(t, ringConfig, "ing1")
+	lifecyclerConfig.NumTokens = 512
+
+	// Start first ingester.
+	ft := &noopFlushTransferer{}
+	l1, err := NewLifecycler(lifecyclerConfig, ft, "ingester")
```
pracucci: You can directly pass `&noopFlushTransferer{}` instead of `ft`, like do for the ingester 2.

pkg/ring/lifecycler_test.go
```diff
@@ -246,17 +256,90 @@ func TestCheckReady(t *testing.T) {
...
+}
+
+func (f *noopFlushTransferer) StopIncomingRequests()                 {}
+func (f *noopFlushTransferer) Flush()                                {}
+func (f *noopFlushTransferer) TransferOut(ctx context.Context) error { return nil }
+
+func TestTokensOnDisk(t *testing.T) {
+	var ringConfig Config
+	flagext.DefaultValues(&ringConfig)
+	ringConfig.KVStore.Mock = consul.NewInMemoryClient(GetCodec())
```
pracucci: You should set `ringConfig.NormaliseTokens = true`, because we're going to get rid of this setting at all (see https://github.com/cortexproject/cortex/pull/1809) and setting it by default.

pkg/ring/lifecycler_test.go
```diff
@@ -246,17 +256,90 @@ func TestCheckReady(t *testing.T) {
...
+		}
+		return ok &&
+			len(desc.Ingesters) == 1 &&
+			desc.Ingesters["ing2"].State == ACTIVE &&
+			len(desc.Ingesters["ing2"].Tokens) == 0 &&
+			len(desc.Tokens) == 512
+	})
+
+	// Check for same tokens.
+	for i := 0; i < 512; i++ {
```
pracucci: Why don't you replace it with `require.Equal(t, expTokens, actTokens)`?

pkg/ring/lifecycler_test.go
```diff
@@ -35,6 +43,8 @@ func testLifecyclerConfig(ringConfig Config, id string) LifecyclerConfig {
 	lifecyclerConfig.NumTokens = 1
 	lifecyclerConfig.ID = id
 	lifecyclerConfig.FinalSleep = 0
+	lifecyclerConfig.TokensFilePath = tokenDir
```
pracucci: Do we need it in the default config used for tests? I would say we want it into in the specific test we have introduced for the tokens persistency.

pkg/ingester/lifecycle_test.go
```diff
@@ -94,11 +95,23 @@ func TestIngesterTransfer(t *testing.T) {
...
```
pracucci: Could you explain why we need to add `TokensFilePath` config option to these tests? Looks it make tests harder to read without adding any value to the tests itself.

pkg/ring/lifecycler.go
codesome: Lots of missed cleanups with changes in designs :)

pkg/ring/lifecycler.go
codesome: There might be race in storing tokens in file in both cases. Best case to avoid this race of writing to files would be having it under the lock. Given that this happens during start-up, writing a small file inside the lock should be fine. WDYT?

pkg/ring/tokens.go
codesome: If we are doing https://github.com/cortexproject/cortex/pull/1750#discussion_r346005496, it fixes this. Else will use the random temp file.

pkg/ingester/lifecycle_test.go
codesome: 1. To test that everything works with tokens on disk (I think we need tests for both on and off disk).
2. So that we can clean it up later. Else it leaves behind the token files in the project (like the ones I accidentally pushed)

pkg/ring/lifecycler_test.go
codesome: Because it is `ing1` in `expTokens[i].Ingester` and `ing2` in `actTokens[i].Ingester`. So that is never going to be true. Hence checking only the tokens.

pkg/ring/lifecycler_test.go
codesome: I just realized that this doesn't make sense at all. I am deleting the directory in the `defer` :man_facepalming:. I will modify to only include it in the new test.

pkg/ingester/lifecycle_test.go
codesome: Yeah, I realized now that we don't need it. Earlier I thought that all the functionalities need to be tested with tokens on disk to make sure that it works fine. But figured that testing only the tokens on disk would suffice.

pkg/ring/lifecycler_test.go
codesome: It goes in LifecyclerConfig, so I will add that in `testLifecyclerConfig` instead

codesome: Apparently `false` is required for `TestRingNormaliseMigration`. Will keep it only for `TestTokensOnDisk`. And good call! `TestTokensOnDisk` fails for `true`.

pkg/ring/lifecycler_test.go
codesome: It would be possible now with normalization turned on.

pkg/ring/lifecycler.go
pracucci: I see two possible race conditions:
1. You're accessing `i.tokens` outside the lock and another concurrent `setTokens()` may conflict
2. Two concurrent `setTokens()` will both try to store the file to disk and 1 will randomly win

I think the #2 is not a problem we should care because it's really remote and you use the move file operation in `StoreToFile`, so one will randomly win but the file integrity is guaranteed.

I believe #1 should be fixed. Given `i.tokens` is never manipulated (correct me if I'm wrong), only overwritten, it should be enough changing `i.tokens.StoreToFile()` to `tokens.StoreToFile()`.

pkg/ring/lifecycler_test.go
```diff
@@ -26,7 +29,7 @@ func (f *flushTransferer) TransferOut(ctx context.Context) error {
 	return f.lifecycler.ChangeState(ctx, ACTIVE)
 }
 
-func testLifecyclerConfig(ringConfig Config, id string) LifecyclerConfig {
+func testLifecyclerConfig(t *testing.T, ringConfig Config, id string) LifecyclerConfig {
```
pracucci: You don't need to pass `t` here. Can you rollback it, please?

pkg/ring/lifecycler_test.go
```diff
@@ -246,17 +249,99 @@ func TestCheckReady(t *testing.T) {
 	flagext.DefaultValues(&ringConfig)
 	ringConfig.KVStore.Mock = &MockClient{}
 
+	tokens := Tokens([]uint32{1})
```
pracucci: Why do you need to define this variable, instead of simply doing the following below?

```
l1.setTokens(Tokens([]uint32{1}))
```

pkg/ring/lifecycler_test.go
pracucci: You've marked the comment as resolved, but doesn't look resolved to me.

> Apparently false is required for TestRingNormaliseMigration.

Yes. I think @pstibrany has a PR for it (you should just enable the normalization in your test).

pkg/ring/tokens.go
pracucci: I wouldn't advocate to wrap the `StoreToFile()` within the lock. I would personally suggest to use `ioutil.TempFile()` here.

pkg/ring/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	tmp := tokenFilePath + ".tmp"
+	f, err := os.OpenFile(tmp, os.O_WRONLY|os.O_CREATE, 0666)
+	if err != nil {
+		return err
+	}
+
+	fileClosed := false
+	defer func() {
+		if !fileClosed {
```
pracucci: This looks overcomplicated to me. The `f.Close()` executed in the defer will just cover the case an error occurred before the function will call `f.Close()` at line `55`. In any case, we don't want to return the `f.Close()` but always the original error (you have a check but looks superfluos to me, cause there's no cause we want to return the error from `f.Close()`).

From my perspective, this logic could be simplified doing:
- Remove `fileClosed`
- Always call `f.Close()` in the defer and ignore its error

pkg/ring/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+			if err := f.Close(); e == nil && err != nil {
+				e = err
+			}
+		}
+		// RemoveAll returns no error when tmp doesn't exist so it is safe to always run it.
+		if err := os.RemoveAll(tmp); err != nil {
+			level.Error(util.Logger).Log("msg", "error deleting temporary file", "err", err)
```
pracucci: This looks more a `level.Warn` to me, because the even if this error occurs the application is not negatively affected (we just miss the opportunity to cleanup a little file from disk).

pkg/ring/tokens.go
pstibrany: This is actually correct logic for closing files. Ignoring `f.Close()` errors is wrong. However, if fileClosed is false, and we're closing in defer, then there must be already other error.

pkg/ring/lifecycler_test.go
codesome: Another cleanup left from the previous :/

pkg/ring/lifecycler.go
codesome: > 1 will randomly win

If the tokens written to the disk is not the winner, then is that not a problem? Because if we are reading the tokens after a restart of ingester, then the tokens that it read won't be all the tokens the is used.

Regarding point 1, yes, that should be fixed, but I think the order of file writes matter, or more like the tokens that are set at the end is should be the one reflected in the file. I will take another look at the functions calls to see if that is a problem.

(PS: using hash with a number will end up linking that PR/issue with this PR)

pkg/ring/tokens.go
codesome: Agree with both the comments. There should already be an error if the file is not closed yet.

pkg/ring/tokens.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	defer func() {
+		// If the file was not closed, then there must already be an error, hence ignore
+		// the error (if any) from f.Close(). If the file was already closed, then
+		// we would ignore the error in that case too.
+		f.Close()
+		// RemoveAll returns no error when tmp doesn't exist so it is safe to always run it.
+		if err := os.RemoveAll(f.Name()); err != nil {
```
pracucci: What if an error occurred during the `ioutil.TempFile()`? `f` is `nil`, so here you should `RemoveAll()` only if `f != nil`.

pkg/ring/lifecycler.go
pracucci: Ok. Then let's be on the safest side (agree) and wrap the `StoreToFile` within the lock (you should just to rollback to `defer i.stateMtx.Unlock()`).

pkg/ring/tokens.go
codesome: We never reach this defer statement if `ioutil.TempFile` errored out :)

pstibrany:
Just out of curiosity... cannot ingester simply reuse its own tokens from the ring? That would assume that it keeps its identity, but otherwise it should already work.

pkg/ring/lifecycler.go
```diff
@@ -456,6 +462,21 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 
 		ingesterDesc, ok := ringDesc.Ingesters[i.ID]
 		if !ok {
+			var tokens Tokens
+			// We load the tokens from the file only if it does not exist in the ring yet.
+			err := tokens.LoadFromFile(i.cfg.TokensFilePath)
```
pstibrany: Longer the CAS operation takes, the higher chance of it failing is, and then entire load needs to be repeated. Perhaps this decision can be moved outside of CAS, so that Load is only performed once?

pkg/ring/lifecycler.go
```diff
@@ -564,11 +585,12 @@ func (i *Lifecycler) autoJoin(ctx context.Context, targetState IngesterState) er
 
 		newTokens := GenerateTokens(i.cfg.NumTokens-len(myTokens), takenTokens)
 		i.setState(targetState)
+
 		ringDesc.AddIngester(i.ID, i.Addr, newTokens, i.GetState(), i.cfg.NormaliseTokens)
 
-		tokens := append(myTokens, newTokens...)
-		sort.Sort(sortableUint32(tokens))
-		i.setTokens(tokens)
+		myTokens = append(myTokens, newTokens...)
```
pstibrany: This code deals with similar situation than saving/loading tokens -- existing tokens in the ring, that belong to the current ingester. Perhaps this can be reused instead of using files?

(There is of course a bug in that AddIngester should not be called with `newTokens` only, but with `myTokens`)

codesome:
> cannot ingester simply reuse its own tokens from the ring

There was some discussion about this. It would work, but, it would be dependant on how reliable is the ring, as it's possible that the ring can go away for a while or it might be restarted sometime. Having it in the file beside the WAL would also make it easy to relate the tokens with the WAL beside it, and makes those tokens on disk as reliable as WAL being present on the disk. Though the ring is reliable enough, I would personally push forward having the tokens on disk. (If ring, we need to have some state other than the current ones while upgrading the ingester else it is likely to mess up with other operations. Playing with local file feels less complex than playing with the ring)

pkg/ring/lifecycler.go
codesome: Right. Didn't notice that part. Thanks!

pkg/ring/lifecycler.go
codesome: I don't think I understand. The current flow is to read the file only if we don't exist in the ring yet. So if do exist in the ring, we are re-using those tokens and we never read from the file.

pstibrany: This was just another comment about the same suggestion of reusing ring (instead of writing to file), to which you have already replied. [If you want to fix call to `AddIngester` with wrong tokens, feel free, but if not, it doesn't like anyone is complaining about is]

pstibrany:
> There was some discussion about this. It would work, but, it would be dependant on how reliable is the ring, as it's possible that the ring can go away for a while or it might be restarted sometime. Having it in the file beside the WAL would also make it easy to relate the tokens with the WAL beside it, and makes those tokens on disk as reliable as WAL being present on the disk.

Fair enough. I just wanted to make sure this option was considered. Thanks.

pkg/ring/lifecycler.go
```diff
@@ -246,18 +248,22 @@ func (i *Lifecycler) ChangeState(ctx context.Context, state IngesterState) error
...
 
-func (i *Lifecycler) setTokens(tokens []uint32) {
+func (i *Lifecycler) setTokens(tokens Tokens) {
 	tokensOwned.WithLabelValues(i.RingName).Set(float64(len(tokens)))
 
 	i.stateMtx.Lock()
 	defer i.stateMtx.Unlock()
+
 	i.tokens = tokens
+	if err := i.tokens.StoreToFile(i.cfg.TokensFilePath); err != nil {
```
pstibrany: I think there should be a check for `i.cfg.TokensFilePath` being non-empty here. Otherwise `StoreToFile` returns error (correct, since it cannot store to empty filename), but on this level, it's not really an error but disabled option.

pkg/ring/tokens.go
```diff
@@ -0,0 +1,99 @@
...
+	case TokensVersion1:
+		*t = Tokens(tj.Tokens)
+		return nil
+	default:
+		return errors.New("invalid token version")
+	}
+}
+
+type tokensJSON struct {
+	Version int      `json:"version"`
```
pstibrany: Nitpick: Json is already easily extensible format, I see little need for versioning it. If you use totally incompatible tokens in the future, just write them into 'tokens2' field.

pracucci: The versioning originated from a comment I made when the previous format was binary. I do agree that json is pretty extensible (compared to a custom binary format) and thus the versioning can be now removed.

pkg/ring/lifecycler.go
pracucci: I missed it while doing the review, but Peter is right. This should be fixed.

pkg/ring/tokens.go
codesome: The idea of having version is to make sense of what the uint32 that we store in the JSON mean. For example, version 1 is storing the actual tokens. If needed for any reason in the future, version 2 would store the deltas to reduce the size. Looks like we can remove the version as the file is not bound to be very large, but would it be good to keep the version anyway if there is a possibility of changing how/what we store?

pkg/ring/tokens.go
```diff
@@ -0,0 +1,99 @@
...
+	if _, err = f.Write(b); err != nil {
+		return err
+	}
+
+	if err := f.Close(); err != nil {
+		return err
+	}
+
+	// Block successfully written, make visible and remove old ones.
+	return fileutil.Replace(f.Name(), tokenFilePath)
```
pstibrany: I think we should just use `os.Rename` here. `fileutil.Replace` has additional step: if target is a directory, it **removes** the directory, and then calls `os.Rename`, but I would argue that if target is indeed a directory (?), it's better to let writing of the token fail, keep the directory and report the error. Operator should handle this situation, cortex shouldn't just delete directories it doesn't like.

pkg/ring/tokens.go
pstibrany: Do we need to delete the temp file? If writing of the token fails, I think it would be better to keep the temp file around, so that it can be inspected by operator and moved to target 'tokens' file. So basically, either temp file is renamed to target location, or we keep it. WDYT?

pkg/ring/tokens.go
pstibrany: If we store deltas, it would use different field name, eg. "deltas". By using fields, new code can actually write both fields: `tokens` for old reades, `deltas` for new readers, thus having both forward and backward compatibility. By using version number, old readers cannot support new versions, so you can only have backward compatibility (new readers reading old version).

It's no big deal, but with JSON, versioning is redundant and just adds extra unnecessary code.

pkg/ring/tokens.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// LoadFromFile loads tokens from given directory.
+func (t *Tokens) LoadFromFile(tokenFilePath string) error {
```
pstibrany: This should be a standalone function, returning loaded tokens or error.

pkg/ring/lifecycler.go
```diff
@@ -447,6 +455,11 @@ heartbeatLoop:
 func (i *Lifecycler) initRing(ctx context.Context) error {
 	var ringDesc *Desc
 
+	var tokensFromFile Tokens
+	if err := tokensFromFile.LoadFromFile(i.cfg.TokensFilePath); err != nil {
```
pstibrany: Currently this code handles `TokensFilePath` being empty by trying to open the file with empty filename, and (hopefully) discovering that it doesn't exist. It is a strange pattern. It would be clearer to just check `TokensFilePath` for emptiness and skip loading of tokens in such case. If tokens file doesn't exist, it would be nice to log that information as a warning, but continue with no tokens.

pkg/ring/tokens.go
codesome: Will remove the versioning

pkg/ring/tokens.go
codesome: It is a file and not a directory. And I see that `fileutil.Replace` takes care of syncing the changes, so I would prefer keeping it as it is. (And I realised that I copied the comment from Prometheus while I was hunting for this function without modifying it :) )

pkg/ring/tokens.go
codesome: Makes sense, we can keep the temporary file hanging around.

pkg/ring/tokens.go
pstibrany: My point was that `os.Rename` and `fileutil.Replace` differ in functionality only if target is a directory (which it isn't here), but you also point out sync on parent dir. I'm not convinced it's enough to justify dependency on fileutil package here.

codesome: I don't have any strong opinion here, will change it to `os.Rename`.

codesome:
@pstibrany I have addressed all your reviews

pkg/ring/lifecycler.go
```diff
@@ -445,9 +453,22 @@ heartbeatLoop:
...
+		err            error
+	)
+
+	if i.cfg.TokensFilePath != "" {
+		tokensFromFile, err = LoadTokensFromFile(i.cfg.TokensFilePath)
+		if err != nil {
+			level.Error(util.Logger).Log("msg", "error in getting tokens from file", "err", err)
+		}
+	} else {
+		level.Warn(util.Logger).Log("msg", "not loading tokens from file, tokens file path is empty")
```
pstibrany: This is expected behavior, level.Info would be fine.

pkg/ring/lifecycler.go
```diff
@@ -456,6 +477,17 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 
 		ingesterDesc, ok := ringDesc.Ingesters[i.ID]
 		if !ok {
+			// We use the tokens from the file only if it does not exist in the ring yet.
+			if len(tokensFromFile) > 0 {
+				level.Info(util.Logger).Log("msg", "adding tokens from file", "num_tokens", len(tokensFromFile))
+				if len(tokensFromFile) == i.cfg.NumTokens {
+					i.setState(ACTIVE)
```
pstibrany: what exactly happens if there isn't enough of them? Ingester stays in PENDING state, and then what?

pkg/ring/lifecycler.go
```diff
@@ -564,11 +596,12 @@ func (i *Lifecycler) autoJoin(ctx context.Context, targetState IngesterState) er
 
 		newTokens := GenerateTokens(i.cfg.NumTokens-len(myTokens), takenTokens)
 		i.setState(targetState)
+
 		ringDesc.AddIngester(i.ID, i.Addr, newTokens, i.GetState(), i.cfg.NormaliseTokens)
```
pstibrany: We only write newly generated `newTokens` to the ring, but if `myTokens` wasn't empty, there is a mismatch between what's in the ring, and what we keep in memory (`myTokens`).

(This is old bug, but since you're changing this code, perhaps we should fix it)

pkg/ring/tokens.go
```diff
@@ -0,0 +1,80 @@
...
+	f, err := ioutil.TempFile(os.TempDir(), "tokens")
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		// If the file was not closed, then there must already be an error, hence ignore
+		// the error (if any) from f.Close(). If the file was already closed, then
+		// we would ignore the error in that case too.
+		f.Close()
```
pstibrany: Please use `_ = f.Close()` to silence checkers that verify unused returned error.

pkg/ring/lifecycler.go
gouthamve: Do we need a warning for this? Maybe debug?

pkg/ring/tokens.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Tokens successfully written, replace the temporary file with the actual file path.
+	return os.Rename(f.Name(), tokenFilePath)
+}
+
+// LoadTokensFromFile loads tokens from given file path.
+func LoadTokensFromFile(tokenFilePath string) (Tokens, error) {
+	b, err := ioutil.ReadFile(tokenFilePath)
+	if err != nil {
+		if os.IsNotExist(err) {
```
pstibrany: I assume that this was ignored previously to handle "" path. Now it is a valid error, that should be returned, and then logged by the caller.

pkg/ring/lifecycler.go
```diff
@@ -564,11 +596,12 @@ func (i *Lifecycler) autoJoin(ctx context.Context, targetState IngesterState) er
...
 
-		tokens := append(myTokens, newTokens...)
-		sort.Sort(sortableUint32(tokens))
-		i.setTokens(tokens)
+		myTokens = append(myTokens, newTokens...)
+		sort.Sort(sortableUint32(myTokens))
```
gouthamve: Now that we have a dedicated type, we should be able to get rid of `sortableUint32` :)

pkg/ring/lifecycler.go
codesome: The `autoJoin` takes care of adding more tokenshttps://github.com/cortexproject/cortex/blob/41d37d8f58c8ca5ba0ea44666fcac8e2a90e5290/pkg/ring/lifecycler.go#L580-L615

Now this makes me wonder what will happen if we have more tokens in the file than `i.cfg.NumTokens` - that is not handled yet. Should we ignore the file completely in that case or take a subset of it?

pkg/ring/lifecycler.go
codesome: Last time I checked, `AddIngester` appended the passed tokens to the existing tokens and didn't overwrite the old ones. But looking at the code I see it is appended in one case and replace in another https://github.com/cortexproject/cortex/blob/41d37d8f58c8ca5ba0ea44666fcac8e2a90e5290/pkg/ring/model.go#L38-L63

pkg/ring/lifecycler.go
pstibrany: > The `autoJoin` takes care of adding more tokenshttps://github.com/cortexproject/cortex/blob/41d37d8f58c8ca5ba0ea44666fcac8e2a90e5290/pkg/ring/lifecycler.go#L580-L615

Ah, so fixing the "old bug" gets more important :)

> Now this makes me wonder what will happen if we have more tokens in the file than `i.cfg.NumTokens` - that is not handled yet. Should we ignore the file completely in that case or take a subset of it?

This is part of WAL work... would it be safe to take subset of tokens only in that context?

codesome: Yup, should be fine, will update accordingly

pkg/ring/tokens.go
codesome: Right

pkg/ring/lifecycler.go
pstibrany: You're right. This is a bug only when using normalization. I'll fix it in #1787 then.

pkg/ring/lifecycler.go
codesome: Info should be fine as @pstibrany suggested. It would be good to have that info without enabling debug level.

pkg/ring/lifecycler.go
```diff
@@ -456,6 +477,18 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 
 		ingesterDesc, ok := ringDesc.Ingesters[i.ID]
 		if !ok {
+			// We use the tokens from the file only if it does not exist in the ring yet.
+			if len(tokensFromFile) > 0 {
+				level.Info(util.Logger).Log("msg", "adding tokens from file", "num_tokens", len(tokensFromFile))
+				if len(tokensFromFile) >= i.cfg.NumTokens {
+					tokensFromFile = tokensFromFile[:i.cfg.NumTokens]
```
gouthamve: Interesting, we should do this if we find tokens in the ring too no? Why are we only doing this for `tokensFromFile`? Also, while decreasing the tokens operation is super rare, if we do this, as the tokens are sorted, I can imagine how we will be ignoring everything at the end and cause huge imbalances if I am not wrong.

gouthamve: Can you not handle this here and open an issue for it rather, as not handling is the what we do currently if I am not wrong?

codesome: Sounds fine, will do that

Issue #1709, opened 2019-10-02
Startup message can have blank version


bboreham:
I see this at the top of my logs:

`caller=main.go:68 msg="Starting Cortex" version="(version=, branch=, revision=)"`

I guess something is supposed to put a value into those strings?
Maybe we would consider this approach, new with Go Modules: https://github.com/dlespiau/go-main-module-version

This is running image `cortexproject/cortex:master-de231804` from Dockerhub.

ariefrahmansyah:
Hi, I would like to pick this up. I have two solutions in mind:

### Proposed solution 1: We still use the current approach and fix missing value.

From my understanding, the current code is using Prometheus' common [version](https://github.com/prometheus/common/blob/master/version/info.go) package and we can use `-ldflags` to set these build information value. We can put it on `Makefile` and it will be something like this:

```
VERSION=$(shell cat VERSION)
GIT_REVISION=$(shell git rev-parse HEAD)
GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)

GO_BUILD_FLAGS := -X github.com/cortexproject/cortex/vendor/github.com/prometheus/common/version.Version=${VERSION} \
        -X github.com/cortexproject/cortex/vendor/github.com/prometheus/common/version.Revision=${GIT_REVISION} \
        -X github.com/cortexproject/cortex/vendor/github.com/prometheus/common/version.Branch=${GIT_BRANCH}

GO_FLAGS := -ldflags "${GO_BUILD_FLAGS} -extldflags \"-static\" -s -w" -tags netgo

$(EXES):
	CGO_ENABLED=0 go build $(GO_FLAGS) -o $@ ./$(@D)
	$(NETGO_CHECK)
```

### Proposed solution 2: Read Build Information from inside the code.

We can use your proposed approach here by using:
1. [`ReadBuildInfo`](https://tip.golang.org/pkg/runtime/debug/#ReadBuildInfo) as we can see on https://github.com/dlespiau/go-main-module-version
2. Execute `git` command using `exec.Command()` as we can see on https://github.com/prometheus/promu/blob/master/pkg/repository/info.go

---

Personally, I prefer to use **solution 1**. Even though Prometheus' common package is made to be shared across Prometheus components and libraries, the `version` package itself is generic enough to be used by other projects. Another benefit is we don't need to modify any Cortex Golang code.

bboreham:
Please go ahead, @ariefrahmansyah 

[...]

PR #1751, opened 2019-10-23
Add GO_LDFLAGS for Golang binary build information

ariefrahmansyah:
Signed-off-by: Arief Rahmansyah <ariefrahmansyah@hotmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ UPTODATE := .uptodate
 
 # Boiler plate for building Cortex's Golang binary.
 VERSION := $(shell cat VERSION)
-GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
+GIT_BRANCH := $(shell ./tools/image-tag)
 GIT_REVISION := $(shell git rev-parse HEAD)
 GO_LDFLAGS := -X github.com/prometheus/common/version.Version=$(VERSION) \
 	-X github.com/prometheus/common/version.Branch=${GIT_BRANCH} \
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,24 @@
 IMAGE_PREFIX ?= quay.io/cortexproject/
 # Use CIRCLE_TAG if present for releases.
 IMAGE_TAG ?= $(if $(CIRCLE_TAG),$(CIRCLE_TAG),$(shell ./tools/image-tag))
-GIT_REVISION := $(shell git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


ariefrahmansyah:
#1709

ariefrahmansyah:
Notes:

1. `GIT_BRANCH` instead of [`/tools/image-tag@master`](https://github.com/cortexproject/cortex/blob/master/tools/image-tag), because `image-tag` produces `branch_name-git_revision[-WIP]` while we only need branch name.

2. I keep `VERSION` file. I agree that the `VERSION` file value is often not in sync with the commit. But since it's still only bumped for every release

bboreham:
> GIT_BRANCH instead of /tools/image-tag@master, because image-tag produces branch_name-git_revision[-WIP] while we only need branch name.

To me, the `WIP` part can be very interesting. 
40 chars of revision is not; 12 or 16 is plenty.

bboreham:
Am I reading this right: you now have short commit sha in the 'branch' field, and long sha in 'revision' ?

Is this what anyone wants?

Makefile
```diff
@@ -6,9 +6,25 @@
...
 # Use CIRCLE_TAG if present for releases.
 IMAGE_TAG ?= $(if $(CIRCLE_TAG),$(CIRCLE_TAG),$(shell ./tools/image-tag))
-GIT_REVISION := $(shell git rev-parse HEAD)
 UPTODATE := .uptodate
 
+# Boiler plate for building Cortex's Golang binary.
+VERSION := $(shell cat VERSION)
+GIT_BRANCH := $(shell ./tools/image-tag)
+GIT_REVISION := $(shell git rev-parse HEAD)
+GO_LDFLAGS := -X github.com/prometheus/common/version.Version=$(VERSION) \
```
pstibrany: Note that this will break in Go 1.14, which will use vendor directory by default, if present and go.mod specifies go 1.14. Vendored dependencies need to be specified as `github.com/cortexproject/cortex/vendor/github.com/prometheus/common/version` :-(

I think easiest solution to handle both scenarios is to have simple build.go in the Cortex project, that will then set Prometheus values in `init` method.

pracucci:
@ariefrahmansyah Are you still interested into this PR? If so, I would suggest you to rebase `master` and reiterate based on the received feedback.

pracucci:
_I'm going to close this PR due to inactivity. Feel free to reopen it in case you're still interested into working on this._

PR #1756, opened 2019-10-24
Various fixes in Consul client and mockKV ("inmemory" KV store)

pstibrany:
- Added tests for WatchKey and WatchKeyPrefix kv.Client methods.
- Removed backoff from Consul client when Consul reports no value for given key or prefix. Since Consul also reports index, next query blocks as expected, and no backoff is necessary. (Introduced by #560, see also next item)
- Fixed Get call in mockKV to block for non-existant key if WaitIndex > 0. This is how Consul works, and this inconsistency was primary reason for PR #560, which introduced backoff in such case instead (which then delays everything, especially tests). Added test for this.
- Changed WatchKeyPrefix in Consul client to not report values that were not modified. Consul returns them even if they haven't changed.
- Fixed mockKV to always start with Index 1. This in line with how Consul works and what our client expects. It treats returned index 0 as invalid value (per Hashicorps recommendations).
- Respect context deadline in mockKV.List function
- mockKV.List only returns values with supplied prefix now

### DIFF
--- a/pkg/ring/kv/consul/client.go
+++ b/pkg/ring/kv/consul/client.go
@@ -168,8 +168,7 @@ func (c *Client) cas(ctx context.Context, key string, f func(in interface{}) (ou
 // under said key changes, the f callback is called with the deserialised
 // value. To construct the deserialised value, a factory function should be
 // supplied which generates an empty struct for WatchKey to deserialise
-// into. Values in Consul are assumed to be JSON. This function blocks until
-// the context is cancelled.
+// into. This function blocks until the context is cancelled or f returns false.
 func (c *Client) WatchKey(ctx context.Context, key string, f func(interface{}) bool) {
 	var (
 		backoff = util.NewBackoff(ctx, backoffConfig)
--- a/pkg/ring/kv/consul/client.go
+++ b/pkg/ring/kv/consul/client.go
@@ -5,7 +5,6 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/go-kit/kit/log/level"
@@ -194,7 +193,10 @@ func (c *Client) WatchKey(ctx context.Context, key string, f func(interface{}) b
 		}
 
 		kvp, meta, err := c.kv.Get(key, queryOptions.WithContext(ctx))
-		if err != nil || kvp == nil {
+
+		// Don't backoff if value is not found (kvp == nil). In that case, Consul still returns index value,
+		// and next call to Get will block as expected. We handle missing value below.
+		if err != nil {
 			level.Error(util.Logger).Log("msg", "error getting path", "key", key, "err", err)
 			backoff.Wait()
 			continue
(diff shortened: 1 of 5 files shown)


pkg/ring/kv/consul/client.go
```diff
@@ -247,7 +247,9 @@ func (c *Client) WatchPrefix(ctx context.Context, prefix string, f func(string,
 		}
 
 		kvps, meta, err := c.kv.List(prefix, queryOptions.WithContext(ctx))
-		if err != nil || kvps == nil {
+		// kvps being nil here is not an error -- quite the opposite. Consul returns index,
+		// which makes next query blocking, so there is no need to detect this and act on it.
+		if err != nil {
```
pstibrany: TODO: add test

pstibrany: Modified TestWatchPrefix test to start watching with empty KV store.

pkg/ring/kv/consul/client.go
```diff
@@ -207,13 +208,15 @@ func (c *Client) WatchKey(ctx context.Context, key string, f func(interface{}) b
...
 		}
 
-		out, err := c.codec.Decode(kvp.Value)
-		if err != nil {
-			level.Error(util.Logger).Log("msg", "error decoding key", "key", key, "err", err)
-			continue
-		}
-		if !f(out) {
-			return
+		if kvp != nil {
```
gouthamve: Could be outdented.
```
if kp == nil {
  log.Info("value is nil")
  continue
}

//logic.
```

Notice that I also added a info log line.

pstibrany: Done

pkg/ring/kv/consul/client.go
```diff
@@ -243,31 +249,38 @@ func (c *Client) WatchPrefix(ctx context.Context, prefix string, f func(string,
...
+				continue
+			}
+
 			out, err := c.codec.Decode(kvp.Value)
 			if err != nil {
 				level.Error(util.Logger).Log("msg", "error decoding list of values for prefix:key", "prefix", prefix, "key", kvp.Key, "err", err)
 				continue
 			}
-			// We should strip the prefix from the front of the key.
-			key := strings.TrimPrefix(kvp.Key, prefix)
```
pracucci: Removing this, isn't a breaking change?

pstibrany: Yes, but it also brings consistency between Consul and Etcd clients. We (I) need to find out if it was relied on. According to @cstyan: "I think that was for ha tracking for something". Fortunately, this method is only used in HA tracking code, so hopefully it shouldn't be that hard.

pracucci: Right. It's used only by the HA tracker, and it watches the prefix `""`, so the change should be harmless 👍

pstibrany: You're right. Thanks for checking!

bboreham:
Do you have an opinion whether we need both a rate-limiter and a backoff in the watch routine?

pstibrany:
They are used differently. Backoff when things go bad (eg. Consul being down), rate limit when things go too good (many ingesters updating ring too often). At the moment, backoff is hardoded cannot be disabled or changed, rate limit is disabled by default and can be used on larger installations when needed.

We could replace backoff with rate limit, but then we would need to enable it for everyone and find a default value that works for both purposes.

PR #1759, opened 2019-10-25
Ability to flush chunks with stale markers early.

gouthamve:

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -175,6 +175,14 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
 
   The maximum duration of a timeseries chunk in memory. If a timeseries runs for longer than this the current chunk will be flushed to the store and a new chunk created. (default 12h)
 
+- `-ingester.max-chunk-idle`
+  
+  If a series doesn't receive a sample for this duration, it is flushed and removed from memory.
+
+- `-ingester.max-stale-chunk-idle`
+  
+  If a series receives a [staleness marker](https://www.robustperception.io/staleness-and-promql), then we wait for this duration to get another sample before we close and flush this series, removing it from memory. You want it to be atleast 2x the scrape interval as you don't want a single failed scrape to cause a chunk flush. 
+
 - `-ingester.chunk-age-jitter`
 
   To reduce load on the database exactly 12 hours after starting, the age limit is reduced by a varying amount up to this. (default 20m)
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -181,7 +181,7 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
 
 - `-ingester.max-stale-chunk-idle`
   
-  If a series receives a [staleness marker](https://www.robustperception.io/staleness-and-promql), then we wait for this duration to get another sample before we close and flush this series, removing it from memory. You want it to be atleast 2x the scrape interval as you don't want a single failed scrape to cause a chunk flush. 
+  If a series receives a [staleness marker](https://www.robustperception.io/staleness-and-promql), then we wait for this duration to get another sample before we close and flush this series, removing it from memory. You want it to be at least 2x the scrape interval as you don't want a single failed scrape to cause a chunk flush. 
 
 - `-ingester.chunk-age-jitter`
 
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -13,6 +13,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/value"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
 	"github.com/cortexproject/cortex/pkg/util"
@@ -128,6 +129,7 @@ const (
 	reasonMultipleChunksInSeries
 	reasonAged
 	reasonIdle
+	reasonStaleIdle
 )
 
 func (f flushReason) String() string {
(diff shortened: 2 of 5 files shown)


pkg/ingester/flush.go
```diff
@@ -200,11 +204,18 @@ func (i *Ingester) shouldFlushChunk(c *desc, fp model.Fingerprint) flushReason {
...
-	// Chunk should be flushed if their last update is older then MaxChunkIdle
+	// Chunk should be flushed if their last update is older then MaxChunkIdle.
 	if model.Now().Sub(c.LastUpdate) > i.cfg.MaxChunkIdle {
 		return reasonIdle
 	}
 
+	// A chunk that has a stale can be flushed if possible.
+	if i.cfg.MaxStaleChunkIdle > 0 &&
+		value.IsStaleNaN(float64(c.LastValue)) &&
+		model.Now().Sub(c.LastUpdate) > i.cfg.MaxStaleChunkIdle {
```
codesome: Looking at this, we would reach here only if `MaxChunkIdle > MaxStaleChunkIdle`. So in that case if staleness is observed before `MaxChunkIdle` is reached, why not just check for staleness instead of additional `MaxStaleChunkIdle`?

gouthamve: We don't want to flush immediately for just one scrape failure. And it doesn't make sense to set `MaxChunkIdle < MaxStaleChunkIdle`

codesome: Isn't a series marked stale after like 5 minutes of no samples and not after a single scrape failure?

bboreham: Brian has you covered https://www.robustperception.io/staleness-and-promql

codesome: Oh, my bad, so the 5m comes from something like `MaxStaleChunkIdle` inside Prometheus. I thought the stale sample was added after 5m.

pkg/ingester/series.go
```diff
@@ -211,19 +211,21 @@ func (s *memorySeries) setChunks(descs []*desc) error {
...
 }
 
-func newDesc(c encoding.Chunk, firstTime model.Time, lastTime model.Time) *desc {
+func newDesc(c encoding.Chunk, firstTime model.Time, lastTime model.Time, lastValue model.SampleValue) *desc {
 	return &desc{
 		C:          c,
 		FirstTime:  firstTime,
 		LastTime:   lastTime,
 		LastUpdate: model.Now(),
+		LastValue:  lastValue,
```
bboreham: can you try to use `memorySeries.lastSampleValue` instead of adding a similar member here?

gouthamve:
@bboreham Using the lastSampleValue from memorySeries now, nice catch! Also added this to the docs.

pkg/ingester/flush.go
```diff
@@ -142,6 +144,8 @@ func (f flushReason) String() string {
 		return "Aged"
 	case reasonIdle:
 		return "Idle"
+	case reasonStaleIdle:
+		return "StaleIdle"
```
bboreham: Why not just "Stale" ?

docs/arguments.md
```diff
@@ -175,6 +175,14 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
 
   The maximum duration of a timeseries chunk in memory. If a timeseries runs for longer than this the current chunk will be flushed to the store and a new chunk created. (default 12h)
 
+- `-ingester.max-chunk-idle`
+  
+  If a series doesn't receive a sample for this duration, it is flushed and removed from memory.
+
+- `-ingester.max-stale-chunk-idle`
+  
+  If a series receives a [staleness marker](https://www.robustperception.io/staleness-and-promql), then we wait for this duration to get another sample before we close and flush this series, removing it from memory. You want it to be atleast 2x the scrape interval as you don't want a single failed scrape to cause a chunk flush.
```
pracucci: `atleast` > `at least`

pkg/ingester/flush.go
```diff
@@ -179,13 +183,13 @@ func (i *Ingester) shouldFlushSeries(series *memorySeries, fp model.Fingerprint,
 		return reasonMultipleChunksInSeries
 	} else if len(series.chunkDescs) > 0 {
 		// Otherwise look in more detail at the first chunk
-		return i.shouldFlushChunk(series.chunkDescs[0], fp)
+		return i.shouldFlushChunk(series.chunkDescs[0], fp, value.IsStaleNaN(float64(series.lastSampleValue)))
```
pracucci: Should you also check `series.lastSampleValueSet`?

_Same comment applies below in `flushUserSeries()`_

pkg/ingester/ingester.go
```diff
@@ -129,6 +130,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.RetainPeriod, "ingester.retain-period", 5*time.Minute, "Period chunks will remain in memory after flushing.")
 	f.DurationVar(&cfg.FlushOpTimeout, "ingester.flush-op-timeout", 1*time.Minute, "Timeout for individual flush operations.")
 	f.DurationVar(&cfg.MaxChunkIdle, "ingester.max-chunk-idle", 5*time.Minute, "Maximum chunk idle time before flushing.")
+	f.DurationVar(&cfg.MaxChunkIdle, "ingester.max-stale-chunk-idle", 0, "Maximum chunk idle time for chunks terminating in stale markers before flushing.")
```
pracucci: I would also say that `0 disables it and a stale series is not flushed until the max-chunk-idle timeout is reached`.

pkg/ingester/flush.go
pracucci: Also, may be more readable if you add a simple function `series.stale()` which returns a `bool`.

PR #1760, opened 2019-10-25
Add global limit to the max series per user and metric

pracucci:
Following up to [this design doc](https://docs.google.com/document/d/1QjPWSobORQ_14_Geaxve1ee5rlfb_nk2GQYl6OZCglc/edit), in this PR I propose to introduce a global limit on the maximum number of series per user and metric.

The global limit has been implemented as a local (per-ingester) limit configured based on the number of ingesters in the cluster + replication factor.

Notes:
- The ingester now requires `-distributor.shard-by-all-labels` config option as well, otherwise it doesn't know it. Not super nice, but I haven't find a better way to do it (better ideas?)
- ~~The ingester now depends on the read ring, to count the number of ingesters across the cluster~~
- The local limit has been kept for backward compatibility, but can be disabled with a value of `0`
- The global limit is disabled by default


pkg/cortex/modules.go
```diff
@@ -419,7 +421,7 @@ var modules = map[moduleName]module{
 	},
 
 	Ingester: {
-		deps: []moduleName{Overrides, Store, Server},
+		deps: []moduleName{Overrides, Store, Ring, Server},
```
pstibrany: Ingester already has access to the ring, doesn't need to initialize Ring component. (in Lifecycler.KVStore)

pracucci: In the ingester I need `ring.RingRead`, while the `Lifecycler` has a different interface (because they both cover different use cases on the same ring). This is the reason why I've initialized the ring here.

Do you mean refactoring `ring.New()` to accept `kv.Client` in input, so that we can reuse the `Lifecycler.KVStore`?

pstibrany: I'm just pointing out that ingester is already reading and updating the ring via its lifecycler, and  it may be possible to use that (esp. if it is only to discover number of other ingesters), instead of introducing dependency on Ring component. Ring component watches the KV for updates constantly, which puts quite a load on the KV store. If all ingesters are going to do that as well, the load will undoubtedly increase.

Initializing two KV clients in memory will also cause issues when using memberlist-based KV store, and that will need to be handled somehow. (Eg. by using our suggestion -- both ring.New and Lifecycler using the same KVClient instance)

I'm not sure what the correct answer is here. I just wanted to bring it up as a potential issue, but if you have already considered alternatives and selected this, perhaps it is a correct answer.

pracucci: Thanks @pstibrany for your feedback. I've [commited a change](https://github.com/cortexproject/cortex/pull/1760/commits/24260ba9b20a0531068ce040c08a1b2a9c0be66f) to rollback the introduction of the `ReadRing` for the ingesters and count the number of healthy ingesters from the `Lifecycler`. May you re-review it, please?

pkg/ingester/ingester.go
```diff
@@ -195,6 +196,13 @@ func New(cfg Config, clientConfig client.Config, limits *validation.Overrides, c
 		return nil, err
 	}
 
+	// Init the limter and instantiate the user states which depend on it
+	i.limiter = NewSeriesLimiter(limits, i.lifecycler, cfg.LifecyclerConfig.RingConfig.ReplicationFactor, cfg.ShardByAllLabels)
+	i.userStates = newUserStates(i.limiter, cfg)
+
+	// Now that user states have been created, we can start the lifecycler
+	i.lifecycler.Start()
```
gouthamve: Curious why this needs to be done after the user states have been created. Why can't we just start when we create the lifecycler as before?

pkg/util/validation/limits.go
```diff
@@ -73,6 +82,18 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&l.PerTenantOverridePeriod, "limits.per-user-override-period", 10*time.Second, "Period with which to reload the overrides.")
 }
 
+// Validate the limits config and returns an error if the validation
+// doesn't pass
+func (l *Limits) Validate(shardByAllLabels bool) error {
+	// The ingester.max-global-series-per-user metric is not support
```
gouthamve: `/s/support/supported`

pkg/util/validation/limits.go
```diff
@@ -1,13 +1,18 @@
...
 import (
+	"errors"
 	"flag"
 	"os"
 	"time"
 
 	"gopkg.in/yaml.v2"
 )
 
+const (
```
gouthamve: A common pattern is to just declare the error here:
`var errMaxGlobalSeriesPerUserValidation = errors.New(test)`

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 * [CHANGE] Remove direct DB/API access from the ruler
 * [CHANGE] Removed `Delta` encoding. Any old chunks with `Delta` encoding cannot be read anymore. If `ingester.chunk-encoding` is set to `Delta` the ingester will fail to start. #1706
+* [FEATURE] Global limit on the max series per user and metric:
```
gouthamve: Need the PR number too :P

pkg/ring/lifecycler.go
```diff
@@ -429,6 +451,8 @@ func (i *Lifecycler) autoJoin(ctx context.Context) error {
 		sort.Sort(sortableUint32(tokens))
 		i.setTokens(tokens)
 
+		i.updateCounters(ringDesc)
```
gouthamve: Right I don't think we should update the counter here, the ring updation might or might not have succeeded.

gouthamve: It should be done in `loop()`.

pkg/ingester/ingester.go
pracucci: The ingester depends on the lifecycler, but the `Lifecycler` depends on the ingester instance too for the flush + transfer. Due to this cyclic dependency, we currently create the ingester instance first and pass it to the lifecycler.

Once the lifecycler is created, it also start the `loop()`, which may potentially lead to some ingester's code requiring `i.userStates` which is not guaranteed to be instantiated yet (after changes in this PR).

**Right now, this should never happen**, but in the future we may do code changes leading to it, which may be difficult to spot in code reviews.

The change proposed is to be more on the safe side, but I'm OK rolling it back if you don't see the value.

pkg/ring/lifecycler.go
pracucci: I've followed a different path instead of doing it in the `loop()`, to keep the loop clear more clean. May you check it out, please?

pkg/ring/lifecycler.go
```diff
@@ -478,6 +519,22 @@ func (i *Lifecycler) changeState(ctx context.Context, state IngesterState) error
 	return i.updateConsul(ctx)
 }
 
+func (i *Lifecycler) updateCounters(ringDesc *Desc) {
+	i.countersLock.Lock()
```
pstibrany: Technically speaking, there is no need to keep the lock while counting healthy ingesters.

pkg/ingester/user_state.go
```diff
@@ -43,13 +43,13 @@ var (
 )
 
 type userStates struct {
```
pstibrany: It would be nice to add a comment to this struct, to explain what is it used for. (I know that you haven't introduced it, so feel free to ignore)

pkg/ingester/limiter.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/cortexproject/cortex/pkg/util/validation"
+)
+
+const (
+	errMaxSeriesPerMetricLimitExceeded = "per-metric series limit (local: %d global: %d actual local: %d) exceeded"
```
pstibrany: I like that you provide details in error messages!

PR #1762, opened 2019-10-25
Add some words about what makes a good PR

bboreham:
I realised I was complaining about people doing it wrong and we haven't set down what 'good' looks like.

Probably there are some good versions of this in other projects, but I couldn't find one I liked on a quick Google.


pracucci:
Well done @bboreham. Have you also considered setting up a simple PR template with a simple checklist like:

- [ ] Tests updated
- [ ] CHANGELOG entry added

_Just as a "reminder" of what should be done._

bboreham:
I hadn't thought of that, but it could be useful.
I've never looked at how templates work on GitHub.

pracucci:
@bboreham Should be enough creating a markdown file called `pull_request_template.md` in the repository root. Inside you just write the content of the template you want to display on PRs.

Issue #1090, opened 2018-10-23
Investigate centralised rate-limiting system


gouthamve:
I am now tweaking the limits in our system, and it's confusing. The limits right now are per distributor/ingester. Which means the limits for each user keeps changing as we scale distributors and ingesters. Further if we want to know how much to set it, it should be `ActualLimit / # distributors` for the distributor limits and for ingester limits it is `ActualLimit * 3 / # ingesters` where is 3 is replication.

This is super confusing and having global limits makes configuration simpler. Looking at tools available to do that, they rely on consistent stores like etcd (https://github.com/youtube/doorman). Not sure if we want that complexity yet.

Having said that, this tool sounds interesting: https://github.com/tsenart/patrol

bboreham:
This was discussed at https://github.com/cortexproject/cortex/pull/844#issuecomment-397826678

gouthamve:
Yes, that is per-ingester user limits, which ideally should scale with the number of ingesters. This is put in to make sure one user doesn't overload an ingester.

But given we hash on ALL the metric labels (including name) now, I think it is fair to assume that the distribution will be even. What we need is `global` series limits for each user and an ingester limit for the number of series for all users. i.e, max number of series per ingester.

tomwilkie:
Also see #255

pracucci:
I've drafted a design document with a proposal to introduce a global rate limiter:
https://docs.google.com/document/d/1QjPWSobORQ_14_Geaxve1ee5rlfb_nk2GQYl6OZCglc/edit

PR #1766, opened 2019-10-29
Add global ingestion rate limiter to distributors

pracucci:
Following up [this design doc](https://docs.google.com/document/d/1QjPWSobORQ_14_Geaxve1ee5rlfb_nk2GQYl6OZCglc/edit#heading=h.bpywb1qpvw13), in this PR I'm proposing to introduce a global ingestion rate limiter implemented as a local rate limiter configured with `limit / N`, where `N` is the actual number of distributors.

Fixes https://github.com/cortexproject/cortex/issues/1090

### Notes about the distributors ring

We currently don't have a ring of distributors, so I had to introduce it.

~~For this purpose, in this PR I'm proposing to introduce a generic simple service discovery based on the KVStore, instead of the ring itself. It's an opinionated change. If there's no consensus, I can rollback to reuse the ring implementation for distributors too.~~

~~From my perspective, a generic lightweight internal service discovery/registry would be just easier to plug compared to the current ring, because of:~~
- ~~No need to have a real ring (ie. no need for tokens)~~
- ~~No need to deal with `IngesterState`~~
- ~~No need to have `FlushTransferer`~~

### Notes about alternatives

- I've excluded to use Kubernetes API to figure out the current number of distributors replica, to not introduce Kubernetes as an hard dependency.


gouthamve:
> For this purpose, in this PR I'm proposing to introduce a generic simple service discovery based on the KVStore, instead of the ring itself. It's an opinionated change. If there's no consensus, I can rollback to reuse the ring implementation for distributors too.

I don't want to maintain another way of doing service discovery tbh. I'd would just use the current one again. There is no proof that the current method is limiting us in a anyway and my guess would be that people would be inclined to use the gossip backend for rate-limiting SD which should be even lesser IO.

pkg/util/servicediscovery/registry.proto
```diff
@@ -0,0 +1,17 @@
...
+package servicediscovery;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+
+message ServiceInstanceDesc {
+	string id = 1;
+	int64 timestamp = 2;
```
pstibrany: What exactly is the timestamp? In the ring it's UnixSeconds, here it's UnixNanos. Should be documented, to avoid confusion.

pstibrany:
> I don't want to maintain another way of doing service discovery tbh. I'd would just use the current one again. There is no proof that the current method is limiting us in a anyway and my guess would be that people would be inclined to use the gossip backend for rate-limiting SD which should be even lesser IO.

This code is reusing existing KVStore, but is storing a different structure into it, under a different key. I think that makes sense, instead of making ring structure bigger. It's written in a generic way (I don't see a need for that, for now, but is harmless otherwise) and called "service discovery", but idea is the same as in the ring (modulo tokens).

pracucci:
@pstibrany and @gouthamve thanks for your comments. I've addressed the feedback:
1. Removed the service discovery in favour of the ring
2. The `FlushTransferer` in the ring is no more required (ie. distributors don't need it)
3. I've generalized the `ring.Lifecycler` logs, removing any reference to "ingester" or "consul" (from the logs) because misleading (not necessarily a ring of ingesters, or backed by consul)
4. I've masked the `LifecyclerConfig` behind a custom ring config for the distributors, so that we expose only the config options / CLI flags which make sense for the distributor (see `distributor_ring.go`)

May you do another round of review, please?

pkg/ring/lifecycler.go
```diff
@@ -38,6 +38,11 @@ var (
 	}, []string{"op", "status", "name"})
 )
 
+// ReadLifecycler represents the read interface to the lifecycler.
+type ReadLifecycler interface {
```
pstibrany: This interface should be defined at the place of consumption (ingestion_rate_strategy.go)

tomwilkie:
Ping @gouthamve?

pkg/distributor/distributor_ring.go
```diff
@@ -0,0 +1,87 @@
...
+	InstanceInterfaceNames []string `yaml:"instance_interface_names"`
+	InstancePort           int      `yaml:"instance_port"`
+	InstanceAddr           string   `yaml:"instance_addr"`
+
+	// Injected internally
+	ListenPort int `yaml:"-"`
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *RingConfig) RegisterFlags(f *flag.FlagSet) {
```
jtlisi: Something similar for this would be nice for the ruler. The main difference being the number of tokens.

pkg/ring/flush.go
```diff
@@ -0,0 +1,30 @@
...
+	StopIncomingRequests()
+	Flush()
+	TransferOut(ctx context.Context) error
+}
+
+// NoopFlushTransferer is a FlushTransferer which does nothing and can
+// be used in cases we don't need one
+type NoopFlushTransferer struct{}
+
+// NewNoopFlushTransferer makes a new NoopFlushTransferer
```
jtlisi: The ruler has identical functionality in `pkg/ruler/lifecycle.go` it would be cool to delete that and reuse this instead.

pkg/distributor/distributor.go
```diff
@@ -119,28 +120,30 @@ type Config struct {
...
 	cfg.BillingConfig.RegisterFlags(f)
 	cfg.PoolConfig.RegisterFlags(f)
 	cfg.HATrackerConfig.RegisterFlags(f)
+	cfg.DistributorRing.RegisterFlags(f)
 
 	f.BoolVar(&cfg.EnableBilling, "distributor.enable-billing", false, "Report number of ingested samples to billing system.")
 	f.IntVar(&cfg.MaxRecvMsgSize, "distributor.max-recv-msg-size", 100<<20, "remote_write API max receive message size (bytes).")
 	f.DurationVar(&cfg.RemoteTimeout, "distributor.remote-timeout", 2*time.Second, "Timeout for downstream ingesters.")
 	f.DurationVar(&cfg.ExtraQueryDelay, "distributor.extra-query-delay", 0, "Time to wait before sending more than the minimum successful query requests.")
-	f.DurationVar(&cfg.LimiterReloadPeriod, "distributor.limiter-reload-period", 5*time.Minute, "Period at which to reload user ingestion limits.")
```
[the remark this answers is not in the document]
pracucci: Sure, done!

pkg/distributor/distributor_ring.go
pracucci: Right. I would suggest to keep ruler's refactoring as a separate PR, given it's not related to this work. _I will add it to my backlog._

pkg/ring/flush.go
pracucci: The ruler `Flush()` is not a noop. Currently the `NoopFlushTransferer` is a adopt all or nothing: how would you see reusing it in the ruler?

jtlisi: True, I forgot the change to make the ruler flush has not been merged yet:
https://github.com/cortexproject/cortex/pull/1571/files#diff-95b5a43683ab83429e93fef5c5daf87fL26

pracucci: No problem. We can use the `NoopFlushTransferer` in the ruler as soon as the PR 1571 will be merged.

owen-d:
Nice work -- I like the ring re-use.

pkg/distributor/distributor.go
```diff
@@ -100,52 +101,52 @@ type Distributor struct {
...
 	f.BoolVar(&cfg.ShardByAllLabels, "distributor.shard-by-all-labels", false, "Distribute samples based on all labels, as opposed to solely by user and metric name.")
 }
 
 // New constructs a new Distributor
-func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Overrides, ring ring.ReadRing) (*Distributor, error) {
+func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Overrides, ring ring.ReadRing, internalDependency bool) (*Distributor, error) {
 	if cfg.ingesterClientFactory == nil {
```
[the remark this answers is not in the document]
gouthamve: Can we rename it to `canJoinDistributorRing`, because if it follow the `ingestersRing` argument, then it seems reasonable that it joins the ingester ring and not a ring of its own.

pkg/distributor/distributor.go
```diff
@@ -166,55 +171,57 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
...
+	// it's an internal dependency and can't join the distributors ring, we skip rate
+	// limiting.
+	var ingestionRateStrategy limiter.RateLimiterStrategy
+	var distributorsRing *ring.Lifecycler
 
-func (d *Distributor) loop() {
-	if d.cfg.LimiterReloadPeriod == 0 {
-		return
-	}
+	if !canJoinRing {
```
gouthamve: Can be simplified to:

```
	if !canJoinRing {
		ingestionRateStrategy = newLocalIngestionRateStrategy(limits)
	} else if limits.IngestionRateStrategy() == validation.GlobalIngestionRateStrategy {
		distributorsRing, err = ring.NewLifecycler(cfg.DistributorRing.ToLifecyclerConfig(), nil, "distributor", ring.DistributorRingKey)
		if err != nil {
			return nil, err
		}

		distributorsRing.Start()

		ingestionRateStrategy = newGlobalIngestionRateStrategy(limits, distributorsRing)
	}
```

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [FEATURE] The frontend split and cache intervals can now be configured using the respective flag `--querier.split-queries-by-interval` and `--frontend.cache-split-interval`.
   * If `--querier.split-queries-by-interval` is not provided request splitting is disabled by default.
   * __`--querier.split-queries-by-day` is still accepted for backward compatibility but has been deprecated. You should now use `--querier.split-queries-by-interval`. We recommend a to use a multiple of 24 hours.__
+* [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
```
gouthamve: The deprecation should also be noted in a `CHANGE`

pkg/distributor/distributor.go
gouthamve: I think this means that it's never going to be infinite, maybe double the logic once?

pracucci: Not sure. I think the current logic is safer and the intention is more clear (when it's an internal dependency, do not configure any rate limit at all).

pkg/distributor/distributor_test.go
```diff
@@ -655,16 +748,18 @@ func prepare(t *testing.T, numIngesters, happyIngesters int, queryDelay time.Dur
...
 	}
-	limits.IngestionRate = 20
-	limits.IngestionBurstSize = 20
 	cfg.ingesterClientFactory = factory
 	cfg.ShardByAllLabels = shardByAllLabels
 	cfg.ExtraQueryDelay = 50 * time.Millisecond
+	cfg.DistributorRing.HeartbeatPeriod = 100 * time.Millisecond
+	cfg.DistributorRing.InstanceID = strconv.Itoa(rand.Int())
+	cfg.DistributorRing.KVStore.Mock = kvStore
+	cfg.DistributorRing.InstanceInterfaceNames = []string{"eth0", "en0", "lo0"}
```
codesome: `TestDistributor_PushIngestionRateLimiter` fails in my local machine because none of these interfaces match :/ I will open a PR to get it from the system so that tests pass on all machines.

codesome: https://github.com/cortexproject/cortex/pull/1915

PR #1770, opened 2019-10-31
Fix for flaky test.

pstibrany:
Test failure reported by @gouthamve, after recent merge of PR #1756:

```
--- FAIL: TestWatchPrefix (3.86s)
    --- FAIL: TestWatchPrefix/etcd (1.86s)
        kv_test.go:191: key test/97 has incorrect value 0
        kv_test.go:191: key test/98 has incorrect value 0
        kv_test.go:191: key test/99 has incorrect value 0
```

~Test was too strict. Now it checks that we receive most notifications, but not necessarily all.~

Test watches for changes under prefix "test/" and expect single change in every key between 0-100. 
It failed because it didn't wait long enough. Now it waits for longer time, but stops as soon it receives all notifications. Also watching for changes starts before generator starts, so that first changes can't possibly be missed.

### DIFF
--- a/pkg/ring/kv/kv_test.go
+++ b/pkg/ring/kv/kv_test.go
@@ -183,18 +183,26 @@ func TestWatchPrefix(t *testing.T) {
 		<-ch1
 		<-ch2
 
-		// verify that each key was reported once, and keys outside prefix were not reported
+		// verify that each key was reported at most once, and keys outside prefix were not reported
+		// also make sure that we get most notifications, if not all.
+		count := 0
 		for i := 0; i < max; i++ {
 			key := fmt.Sprintf("%s%d", prefix, i)
 
-			if observedKeys[key] != 1 {
+			if observedKeys[key] > 1 {
 				t.Errorf("key %s has incorrect value %d", key, observedKeys[key])
 			}
+			count += observedKeys[key]
 			delete(observedKeys, key)
 		}
 
 		if len(observedKeys) > 0 {
 			t.Errorf("unexpected keys reported: %v", observedKeys)
+		}
+
+		const expectedFactor = 0.75
+		if count < expectedFactor*max {
+			t.Errorf("expected at least %.0f%% observed values, got %.0f%% (observed count: %d)", 100*expectedFactor, 100*float64(count)/max, count)
 		}
 	})
 }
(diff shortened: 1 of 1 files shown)


pkg/ring/kv/kv_test.go
```diff
@@ -183,18 +183,26 @@ func TestWatchPrefix(t *testing.T) {
...
 			}
+			count += observedKeys[key]
 			delete(observedKeys, key)
 		}
 
 		if len(observedKeys) > 0 {
 			t.Errorf("unexpected keys reported: %v", observedKeys)
 		}
+
+		const expectedFactor = 0.75
```
pracucci: I'm not a big fan of this, cause it doesn't guarantee it works as expected (ie. no key has been missed). A trick may be waiting - right before the `for` until the number of observed keys doesn't increase for some time (ie. 500ms). You may do something like this:

```
		// wait until the WatchPrefix() stop gets called, which means all the
		// keys have been observed
		lastObservedKeysCount := len(observedKeys)
		lastObservedKeysChangeAt := time.Now()

		test.Poll(t, 2*time.Second, true, func() interface{} {
			count := len(observedKeys)
			elapsed := time.Now().Sub(lastObservedKeysChangeAt)

			if count == lastObservedKeysCount {
				return elapsed >= 250*time.Millisecond
			}

			lastObservedKeysCount = count
			lastObservedKeysChangeAt = time.Now()
			return false
		})
```

Does it make sense?

pstibrany: I've tried to implement your suggestion. Test is strict again (each key once), but more forgiving in terms of timeouts. But if nothing is received in 500ms, we stop.

pkg/ring/kv/kv_test.go
```diff
@@ -165,23 +178,35 @@ func TestWatchPrefix(t *testing.T) {
...
-		client.WatchPrefix(ctx, prefix, func(key string, val interface{}) bool {
-			observedKeys[key] = observedKeys[key] + 1
-			return true
-		})
 
-		// wait until updaters finish (should be done by now)
-		<-ch1
-		<-ch2
+		totalDeadline := time.After(totalTestTimeout)
+		var nextTimer = time.NewTimer(nextKeyDeadline)
```
pracucci: To simplify it, what if we remove the `nextTimer` at all, and we simply set `watching = false` when `len(observedKeys)` reach `max`?

pstibrany: Sounds good

pstibrany:
> Thanks @pstibrany to patiently address my feedback. LGTM!

Thanks for your review and feedback!

PR #1773, opened 2019-11-01
tsdb/blocks: shard buckets by time

thorfour:

This adds horizontal scaling to S3 buckets for the block storage path.
It does this by accepting multiple S3 endpoints with a date attached after a `|` 

`ex: mybucket.eu-west1.amazonaws.com|2019-01-01`

This will starting sharding blocks based on ULID to that bucket on Jan 1st 2019.


gouthamve:
Could you explain which path is bottlenecking here? Are we hitting limits on the read path or the write path?

At what QPS are we seeing rate-limits?

thorfour:
@gouthamve It's entirely dependent on workload and S3 implementation. 

Ex: AWS S3 us 3500 Writes/s and 55 reads/s, DigitalOcean Spaces is 740 rps. These are per bucket.

So this is a horizontally scaling release valve should limits be hit. It'll reduce the overall rps sent to each bucket.

thorfour:
@pracucci would love your review to see if this is compatible with your GCS changes.

pkg/ingester/bucket.go
```diff
@@ -2,67 +2,238 @@ package ingester
...
+	"github.com/pkg/errors"
+	"github.com/thanos-io/thanos/pkg/block"
 	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/objstore/s3"
 	"golang.org/x/net/context"
 )
 
-// Bucket is a wrapper around a objstore.Bucket that prepends writes with a userID
+// Bucket is a wrapper around a slice of objstore.Bucket that prepends writes with a userID and shards them based on root name
 type Bucket struct {
```
pracucci: From a design perspective, this looks weird to me. A `Bucket` struct contains multiple `PeriodBucket`. Naming wise, to me a `Bucket` is a single bucket and then you have a storage abstraction which - under the hood is composed by multiple buckets.

I would suggest to rework a bit on this. An option may be keeping the  `Bucket` as is in master (all in all, a bucket is a bucket) and then add an abstraction over it, which incapsulate a "multiple buckets storage".

Following up, this way we may also keep the current implementation as is and opt-in into the multi-buckets only if multiple endpoints are configured. The idea is to keep two backend storages implementation: one single bucket and one multi-bucket. The multi-bucket is based on the single-bucket, for the single bucket implementation.

pkg/storage/tsdb/config.go
```diff
@@ -15,12 +15,13 @@ const (
...
-	Retention    time.Duration `yaml:"retention_period"`
-	ShipInterval time.Duration `yaml:"ship_interval"`
-	Backend      string        `yaml:"backend"`
+	Dir            string        `yaml:"dir"`
+	SyncDir        string        `yaml:"sync_dir"`
+	BlockRanges    time.Duration `yaml:"block_ranges_period"`
+	Retention      time.Duration `yaml:"retention_period"`
+	ShipInterval   time.Duration `yaml:"ship_interval"`
+	Backend        string        `yaml:"backend"`
+	ConcurrentSync int           `yaml:"concurrent_sync"`
```
pracucci: I would rename this to `ConcurrentSyncLimit` / `concurrent_sync_limit` / `experimental.tsdb.concurrent-sync-limit`.

pkg/storage/tsdb/backend/s3/config.go
```diff
@@ -1,21 +1,58 @@
...
+	return nil
 }
 
 // RegisterFlags registers the flags for TSDB s3 storage
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.AccessKeyID, "experimental.tsdb.s3.access-key-id", "", "S3 access key ID")
 	f.StringVar(&cfg.SecretAccessKey, "experimental.tsdb.s3.secret-access-key", "", "S3 secret access key")
-	f.StringVar(&cfg.BucketName, "experimental.tsdb.s3.bucket-name", "", "S3 bucket name")
-	f.StringVar(&cfg.Endpoint, "experimental.tsdb.s3.endpoint", "", "S3 endpoint without schema")
+	f.Var(&cfg.Endpoints, "experimental.tsdb.s3.endpoint-from", "endpoint names with a bucket prefix and a date to start using the endpoint from (ex: bucket.endpoint|2019-01-01)")
```
pracucci: Few things:
1. The yaml config option is called `endpoints`, while here it's called `endpoint-from`. We should keep consistency
2. It's S3 specific, while we should move this endpoint config to the generic TSDB config and then use it both in S3 and GCS config
3. Have you considered renaming `endpoints` into `buckets`? From a user perspective, may look easier to understand
4. The overall config should work in single bucket mode (no date). I would also suggest to add unit tests to cover both cases
5. I've mixed feelings the way the bucket name and endpoint have been merged. For example, if the bucket name contains dots (ie. it was used in my previous company) it will break. A safer way to handle it would be through a dedicated config block (ie. `BucketConfig`) which contains `bucket_name`, `endpoint` and `starting_date`, and the `buckets` is a list of `BucketConfig`. CLI flags may support only the case of the single bucket, and enforce the use of the YAML config for multi buckets setup (similar to Cortex schema config, which CLI flags support only some legacy use cases).

thorfour:
Since this is going to require a large refactor with @pracucci latest changes, I'm going to close this PR and open a new one once that's complete.

PR #1776, opened 2019-11-04
Export Unix timestamp of oldest unflushed chunk in the memory.

### LINKED ISSUE #1386: Alerting on number of chunks per series is noisy
We alert when there are more than 1.3 chunks per series per ingester.  This works most of the time, but when a new ingester is added (without transfers), all the series in that ingester are created, and flushed, at the same time.  This means for a period of at least 15 mins, there is almost 2 chunks per series for this ingester.

I propose instead of alert on this, we export the age of the oldest chunk, and alert on that.  This will also catch cases where chunks persistently fail to flush.  The question is, how do we implement this metrics without scanning all chunks every scrape interval?

pstibrany:
This PR implements a simple addition: export of timestamp of oldest unflushed chunk in memory. This can be used for alerting, if timestamp is older than expected.

Fixes #1386

### DIFF
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -67,6 +67,10 @@ var (
 		Name: "cortex_ingester_dropped_chunks_total",
 		Help: "Total number of chunks dropped from flushing because they have too few samples.",
 	})
+	oldestUnflushedChunkTimestamp = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "cortex_oldest_unflushed_chunk_timestamp",
+		Help: "Unix timestamp of the oldest unflushed chunk in the memory",
+	})
 )
 
 // Flush triggers a flush of all the chunks and closes the flush queues.
@@ -110,14 +114,23 @@ func (i *Ingester) sweepUsers(immediate bool) {
 		return
 	}
 
+	oldest := model.Time(0)
+
 	for id, state := range i.userStates.cp() {
 		for pair := range state.fpToSeries.iter() {
 			state.fpLocker.Lock(pair.fp)
 			i.sweepSeries(id, pair.fp, pair.series, immediate)
 			i.removeFlushedChunks(state, pair.fp, pair.series)
+			first := pair.series.firstUnflushedChunkTime()
 			state.fpLocker.Unlock(pair.fp)
+
+			if first > 0 && (oldest == 0 || first < oldest) {
+				oldest = first
+			}
 		}
 	}
+
+	oldestUnflushedChunkTimestamp.Set(float64(oldest.Unix()))
 }
 
 type flushReason int
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -68,7 +68,7 @@ var (
 		Help: "Total number of chunks dropped from flushing because they have too few samples.",
 	})
 	oldestUnflushedChunkTimestamp = promauto.NewGauge(prometheus.GaugeOpts{
-		Name: "cortex_oldest_unflushed_chunk_timestamp",
+		Name: "cortex_oldest_unflushed_chunk_timestamp_seconds",
 		Help: "Unix timestamp of the oldest unflushed chunk in the memory",
 	})
 )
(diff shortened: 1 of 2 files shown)


pkg/ingester/flush.go
```diff
@@ -67,6 +67,10 @@ var (
 		Name: "cortex_ingester_dropped_chunks_total",
 		Help: "Total number of chunks dropped from flushing because they have too few samples.",
 	})
+	oldestUnflushedChunkTimestamp = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "cortex_oldest_unflushed_chunk_timestamp",
```
gouthamve: Per the naming best practices, it should be `cortex_oldest_unflushed_chunk_time_seconds`: https://prometheus.io/docs/practices/naming/ and a similar metric is `process_start_time_seconds`

pstibrany: Prometheus doesn't seem to be consistent either, with several 'timestamp_in_seconds' metrics:

http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=%7B__name__%3D~%22.%2Btime.*%22%2C%20job%3D%22prometheus%22%7D&g0.tab=1

pstibrany: I'm tempted to rename it to `cortex_oldest_unflushed_chunk_timestamp_in_seconds`, would that be fine? Or do you strongly prefer `cortex_oldest_unflushed_chunk_time_seconds`?

gouthamve: `cortex_oldest_unflushed_chunk_timestamp_seconds` seems the best. `_in_` is redundant as that is implicit when ending the metric with units in Prometheus.

pkg/ingester/flush.go
```diff
@@ -110,14 +114,23 @@ func (i *Ingester) sweepUsers(immediate bool) {
...
+
 	for id, state := range i.userStates.cp() {
 		for pair := range state.fpToSeries.iter() {
 			state.fpLocker.Lock(pair.fp)
 			i.sweepSeries(id, pair.fp, pair.series, immediate)
 			i.removeFlushedChunks(state, pair.fp, pair.series)
+			first := pair.series.firstUnflushedChunkTime()
 			state.fpLocker.Unlock(pair.fp)
+
+			if first > 0 && (oldest == 0 || first < oldest) {
```
bboreham: This would be simplified by initializing `oldest` to maxInt, perhaps?

How does your alerting rule look when the metric is sometimes zero?

pkg/ingester/series.go
```diff
@@ -151,6 +151,23 @@ func (s *memorySeries) firstTime() model.Time {
 	return s.chunkDescs[0].FirstTime
 }
 
+// Returns time of oldest chunk in the series, that isn't flushed. If there are
+// no chunks, or all chunks are flushed, returns 0.
+// The caller must have locked the fingerprint of the memorySeries.
+func (s *memorySeries) firstUnflushedChunkTime() model.Time {
+	if len(s.chunkDescs) == 0 {
```
bboreham: seems unnecessary

pkg/ingester/flush.go
pstibrany: I think something like this can work (12 hours check):

`(cortex_oldest_unflushed_chunk_timestamp_seconds >bool 0) * (time() - cortex_oldest_unflushed_chunk_timestamp_seconds) > 12*60*60`

pstibrany: You're right that it can be simplified by using MaxInt64, but I want to return 0 in case of no chunks. I think that's better than 9.223372036854776e+18, which is math.MaxInt64 converted to float64.

pkg/ingester/series.go
pstibrany: Removed.

bboreham:
I wonder if this information can be used to fix a thorny issue with ingester resilience, as described at https://github.com/cortexproject/cortex/issues/731#issuecomment-435362741

(not 100% sure that is a valid fix, but it could be worth exploring)

PR #1778, opened 2019-11-04
Fix duplicated series when querying both ingesters and TSDB storage

pracucci:
**What this PR does**:
The current (experimental) TSDB integration has two issues related to the external label added by the shipper:

1. The label name is `"user"`, which can overlap with a metric label
2. The external label is not removed when querying the TSDB storage, thus leading to duplicated series: the series read from the storage have the `"user"` label, while the series read by ingesters don't

In this PR, I'm proposing to:
1. Rename `"user"` external label to `__org_id__` to avoid collisions
2. Remove the external label when querying the storage
3. Abort a TSDB query in case of partial response (we want full consistency)

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -81,17 +81,14 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 	// Walk the samples, appending them to the users database
 	app := db.Appender()
 	for _, ts := range req.Timeseries {
+		// Convert labels to the type expected by TSDB
+		lset := cortex_tsdb.FromLabelAdaptersToLabels(ts.Labels)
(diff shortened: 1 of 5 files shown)


pracucci:
@thorfour What's your take?

pracucci:
Open question: should we validate if the pushed series contain the `__org_id__` label in the distributor, when TSDB storage is enabled, or it's an edge case doesn't really make sense to check?

@gouthamve @thorfour thoughts?

thorfour:
We should probably either check it in distributor, or at the very least document somewhere that `__org_id__` isn't a valid label to send Cortex with tsdb enabled, just so there isn't an undocumented surprise.

PR #1786, opened 2019-11-05
Refactor: change index key markers from slice to single byte

bboreham:
I was concerned that `parseChunkTimeRangeValue()` is getting slower and slower as it linear-searches the choices with the newest ones at the end.

This is more idiomatic Go and more efficient because it doesn't have to loop over the slice.

### DIFF
--- a/pkg/chunk/schema.go
+++ b/pkg/chunk/schema.go
@@ -12,21 +12,23 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 )
 
-var (
-	chunkTimeRangeKeyV1a = []byte{1}
-	chunkTimeRangeKeyV1  = []byte{'1'}
-	chunkTimeRangeKeyV2  = []byte{'2'}
-	chunkTimeRangeKeyV3  = []byte{'3'}
-	chunkTimeRangeKeyV4  = []byte{'4'}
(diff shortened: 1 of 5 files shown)


pkg/chunk/schema.go
```diff
@@ -12,21 +12,23 @@ import (
...
 
-var (
-	chunkTimeRangeKeyV1a = []byte{1}
-	chunkTimeRangeKeyV1  = []byte{'1'}
-	chunkTimeRangeKeyV2  = []byte{'2'}
-	chunkTimeRangeKeyV3  = []byte{'3'}
-	chunkTimeRangeKeyV4  = []byte{'4'}
-	chunkTimeRangeKeyV5  = []byte{'5'}
-	metricNameRangeKeyV1 = []byte{'6'}
+const (
```
pracucci: After this refactoring, the version is required to be single character. We've already hit version `9`, so should we rollback the change on next version or switch to letters? _Just asking to better understand the impact of this change to the future maintenance._

bboreham: It’s a good question, but there are (255-11) byte values available so I think we should be ok. 
I’d guess the next choice would be 'A'.

pracucci:
@bboreham Do we wanna merge it?

PR #1790, opened 2019-11-05
Fixes store duplicate label names and values.

cyriltovena:
When we query label names or values across multiple stores from different schema period we can lose the unicity or ordering.

This changes sort and filter label in the composite store. I've also added some test to verify the output.

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -204,17 +204,15 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, userID string, fro
 		return nil, err
 	}
 
-	var result []string
-	values := map[string]struct{}{}
+	var result UniqueStrings
 	for _, entry := range entries {
 		_, labelValue, _, _, err := parseChunkTimeRangeValue(entry.RangeValue, entry.Value)
 		if err != nil {
 			return nil, err
 		}
-		result = appendUniqueStrings(result, values, string(labelValue))
+		result.Add(string(labelValue))
 	}
-	sort.Strings(result)
-	return result, nil
+	return result.Strings(), nil
 }
 
 // LabelNamesForMetricName retrieves all label names for a metric name.
@@ -462,7 +460,6 @@ func (c *store) lookupEntriesByQueries(ctx context.Context, queries []IndexQuery
 
 func (c *store) parseIndexEntries(ctx context.Context, entries []IndexEntry, matcher *labels.Matcher) ([]string, error) {
 	result := make([]string, 0, len(entries))
-	values := map[string]struct{}{}
 	for _, entry := range entries {
 		chunkKey, labelValue, _, _, err := parseChunkTimeRangeValue(entry.RangeValue, entry.Value)
 		if err != nil {
@@ -472,10 +469,11 @@ func (c *store) parseIndexEntries(ctx context.Context, entries []IndexEntry, mat
 		if matcher != nil && !matcher.Matches(string(labelValue)) {
 			continue
 		}
-		result = appendUniqueStrings(result, values, chunkKey)
+		result = append(result, chunkKey)
 	}
 	// Return ids sorted and deduped because they will be merged with other sets.
 	sort.Strings(result)
+	result = uniqueStrings(result)
 	return result, nil
 }
 
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -205,15 +205,15 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, userID string, fro
 	}
 
 	var result []string
+	values := map[string]struct{}{}
 	for _, entry := range entries {
 		_, labelValue, _, _, err := parseChunkTimeRangeValue(entry.RangeValue, entry.Value)
 		if err != nil {
 			return nil, err
 		}
-		result = append(result, string(labelValue))
+		result = appendUniqueStrings(result, values, string(labelValue))
 	}
 	sort.Strings(result)
-	result = uniqueStrings(result)
 	return result, nil
 }
 
@@ -462,7 +462,7 @@ func (c *store) lookupEntriesByQueries(ctx context.Context, queries []IndexQuery
 
 func (c *store) parseIndexEntries(ctx context.Context, entries []IndexEntry, matcher *labels.Matcher) ([]string, error) {
 	result := make([]string, 0, len(entries))
-
+	values := map[string]struct{}{}
 	for _, entry := range entries {
 		chunkKey, labelValue, _, _, err := parseChunkTimeRangeValue(entry.RangeValue, entry.Value)
 		if err != nil {
@@ -472,12 +472,10 @@ func (c *store) parseIndexEntries(ctx context.Context, entries []IndexEntry, mat
 		if matcher != nil && !matcher.Matches(string(labelValue)) {
 			continue
 		}
-		result = append(result, chunkKey)
+		result = appendUniqueStrings(result, values, chunkKey)
 	}
-
 	// Return ids sorted and deduped because they will be merged with other sets.
 	sort.Strings(result)
-	result = uniqueStrings(result)
 	return result, nil
 }
 
--- a/pkg/chunk/composite_store.go
+++ b/pkg/chunk/composite_store.go
@@ -101,28 +101,32 @@ func (c compositeStore) Get(ctx context.Context, userID string, from, through mo
 // LabelValuesForMetricName retrieves all label values for a single label name and metric name.
 func (c compositeStore) LabelValuesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string, labelName string) ([]string, error) {
 	var result []string
+	uniqueValues := map[string]struct{}{}
 	err := c.forStores(from, through, func(from, through model.Time, store Store) error {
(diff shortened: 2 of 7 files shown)


cyriltovena:
see https://github.com/grafana/loki/issues/1220.

pkg/chunk/chunk_store.go
```diff
@@ -205,15 +205,19 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, userID string, fro
...
 		}
-		result = append(result, string(labelValue))
+		v := string(labelValue)
+		if _, ok := values[v]; !ok {
+			values[v] = struct{}{}
+			result = append(result, v)
+		}
+
 	}
-	sort.Strings(result)
```
pracucci: To my understanding, the changes in this function are to remove the sorting, because - according to the rest of the changes - values will be sorted later on in the composite store. Am I correct?

If correct, it looks weird to me that `LabelValuesForMetricName()` in the chunk store and composite store have a different behaviour. The first one returns non-sorted values, while the second one does. However both implements the same `Store` interface and - by contract - should behave the same. Thoughts?

pkg/chunk/composite_store.go
```diff
@@ -101,28 +101,42 @@ func (c compositeStore) Get(ctx context.Context, userID string, from, through mo
...
 func (c compositeStore) LabelValuesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string, labelName string) ([]string, error) {
 	var result []string
+	uniqueValues := map[string]struct{}{}
 	err := c.forStores(from, through, func(from, through model.Time, store Store) error {
 		labelValues, err := store.LabelValuesForMetricName(ctx, userID, from, through, metricName, labelName)
 		if err != nil {
 			return err
 		}
-		result = append(result, labelValues...)
+		for _, l := range labelValues {
```
pracucci: Since this piece of code is getting duplicated in few places, have you considered creating an utility function `appendUniqueStrings(result []string, uniqueValues map[string)struct{}, values []string) []string`?

I would be easily called like this:
```
result = appendUniqueStrings(result, uniqueVaLues, labelValues)
```

pkg/chunk/composite_store_test.go
```diff
@@ -180,3 +180,64 @@ func TestCompositeStore(t *testing.T) {
...
+}
+
+func TestCompositeStoreLabels(t *testing.T) {
+	t.Parallel()
+
+	cs := compositeStore{
+		stores: []compositeStoreEntry{
+			{model.TimeFromUnix(0), mockStore(1)},
+			{model.TimeFromUnix(20), mockStoreLabel{mockStore(1), []string{"b", "e", "c"}}},
+			{model.TimeFromUnix(40), mockStoreLabel{mockStore(1), []string{"f", "a", "b", "c"}}},
```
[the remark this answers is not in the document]
cyriltovena: Well my idea was that I still need to do ordering in the composite store so to save some time, I removed it from the store implementation.

I can keep ordering and uniqueness in each store if that's better ?

pkg/chunk/chunk_store.go
cyriltovena: yes you're correct that was my idea. If we're not worried about performance, I'll add back sorting and uniquess on each storage.

cyriltovena:
> I'm not very sure what's going on here.

Have you read the PR description ?

pkg/chunk/chunk_store.go
```diff
@@ -205,15 +205,20 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, userID string, fro
...
 	var result []string
+	values := map[string]struct{}{}
 	for _, entry := range entries {
 		_, labelValue, _, _, err := parseChunkTimeRangeValue(entry.RangeValue, entry.Value)
 		if err != nil {
 			return nil, err
 		}
-		result = append(result, string(labelValue))
+		v := string(labelValue)
+		if _, ok := values[v]; !ok {
```
pracucci: I guess you can use `result = appendUniqueStrings(result, values, []string{v})` or you see it inefficient?

cyriltovena: I was actually trying to find a way to use it here, there's another place also and no I think it is a good idea.

pkg/ingester/flush.go
```diff
@@ -67,6 +67,10 @@ var (
 		Name: "cortex_ingester_dropped_chunks_total",
 		Help: "Total number of chunks dropped from flushing because they have too few samples.",
 	})
+	oldestUnflushedChunkTimestamp = promauto.NewGauge(prometheus.GaugeOpts{
```
pracucci: I don't know what happened, but looks something went wrong with the last merge/rebase you did.

pkg/chunk/chunk_store.go
bboreham: We are worried about performance, but worried more about correctness.
It is necessary to highlight inconsistencies and clearly explain in comments why it's ok to break a contract or an invariant.

pkg/chunk/chunk_store.go
```diff
@@ -205,15 +205,16 @@ func (c *store) LabelValuesForMetricName(ctx context.Context, userID string, fro
...
 
 	var result []string
+	values := map[string]struct{}{}
 	for _, entry := range entries {
 		_, labelValue, _, _, err := parseChunkTimeRangeValue(entry.RangeValue, entry.Value)
 		if err != nil {
 			return nil, err
 		}
-		result = append(result, string(labelValue))
+		result = appendUniqueStrings(result, values, string(labelValue))
```
bboreham: Since this change is (now) unrelated to the bugfix, I would prefer it goes in a different commit with its own explanation.

You may want to make the same change in `parseIndexEntries()`, which would let you remove `uniqueStrings()`.

It's worth checking the benchmark numbers and/or making sure the benchmarks are using a realistic data set.

cyriltovena: good catch ! I'll remove it.

cyriltovena: voila !

cyriltovena: before :

```
pkg: github.com/cortexproject/cortex/pkg/chunk
BenchmarkParseIndexEntries500-16      	    4562	    257222 ns/op	   96427 B/op	    1503 allocs/op
BenchmarkParseIndexEntries2500-16     	     847	   1366432 ns/op	  482388 B/op	    7503 allocs/op
BenchmarkParseIndexEntries10000-16    	     198	   5846147 ns/op	 1928891 B/op	   30005 allocs/op
BenchmarkParseIndexEntries50000-16    	      38	  30553792 ns/op	 9607698 B/op	  150006 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/chunk	7.953s
```

after:

```
pkg: github.com/cortexproject/cortex/pkg/chunk
BenchmarkParseIndexEntries500-16      	    3678	    318019 ns/op	  130810 B/op	    1522 allocs/op
BenchmarkParseIndexEntries2500-16     	     747	   1632040 ns/op	  612466 B/op	    7563 allocs/op
BenchmarkParseIndexEntries10000-16    	     172	   6779048 ns/op	 2441477 B/op	   30219 allocs/op
BenchmarkParseIndexEntries50000-16    	      31	  36258715 ns/op	11602340 B/op	  151867 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/chunk	8.108s
```

cyriltovena: Adding half duplicate in the mix:

before:

```
pkg: github.com/cortexproject/cortex/pkg/chunk
BenchmarkParseIndexEntries500-16      	    4485	    262828 ns/op	   96346 B/op	    1503 allocs/op
BenchmarkParseIndexEntries2500-16     	     838	   1379892 ns/op	  482316 B/op	    7503 allocs/op
BenchmarkParseIndexEntries10000-16    	     201	   5907672 ns/op	 1928821 B/op	   30005 allocs/op
BenchmarkParseIndexEntries50000-16    	      39	  30373427 ns/op	 9607658 B/op	  150007 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/chunk	8.056s
```

after:
```
pkg: github.com/cortexproject/cortex/pkg/chunk
BenchmarkParseIndexEntries500-16      	    4454	    265512 ns/op	  109322 B/op	    1514 allocs/op
BenchmarkParseIndexEntries2500-16     	     878	   1356005 ns/op	  526739 B/op	    7536 allocs/op
BenchmarkParseIndexEntries10000-16    	     211	   5586487 ns/op	 2107155 B/op	   30116 allocs/op
BenchmarkParseIndexEntries50000-16    	      37	  29155700 ns/op	10205072 B/op	  150868 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/chunk	7.917s
```

cyriltovena: So it looks like if there is duplicates the new append with the map unique values  is faster, no surprise here. However if there is no duplicates the old version is faster.

Memory wise I think the increase is negligible. (25B per op).

I will submit the benchmark now.

cyriltovena: After some discussions with Goutham, let's not take any risk and leave the `parseIndexEntries` as is. Other functions (Fetching labels) are not used by Cortex and should be fine to change.

cyriltovena: done ! I think we should be good to go here.

PR #1794, opened 2019-11-07
indexcache with memcached

thorfour:
This adds the option to use the memcached config for the index cache in the user bucket stores.

### DIFF
--- a/pkg/querier/block_store.go
+++ b/pkg/querier/block_store.go
@@ -16,7 +16,6 @@ import (
 	"github.com/thanos-io/thanos/pkg/model"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/store"
-	storecache "github.com/thanos-io/thanos/pkg/store/cache"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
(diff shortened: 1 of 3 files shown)


thorfour:
This wasn't intended to be a performance enhancement for query speeds but a way to reduce memory footprint of queriers by offloading their index cache into memcached. I'll run some benchmarks to see what the reduction might be.

pracucci:
> This wasn't intended to be a performance enhancement for query speeds but a way to reduce memory footprint of queriers by offloading their index cache into memcached. I'll run some benchmarks to see what the reduction might be.

I don't want to make it complicated, but I would like to better understand the use case.

Right now, we do setup a per-tenant in-memory index cache limited to 250MB. If you have 1K tenants, it's 244GB which I understand it's problematic. Introducing memcached you have no per-ingester index cache, while using a centralized one. However, if we realize accessing the centralized cache is slower than having no cache at all, would be better setting the per-ingester index cache to 0B?

thorfour:
> . However, if we realize accessing the centralized cache is slower than having no cache at all, would be better setting the per-ingester index cache to 0B

If that ends up being the case then definitely, and I'll gladly abandon this PR. I just need to benchmark those times.

PR #1798, opened 2019-11-08
Don't return 500's for KV CAS calls if the CAS returned an HTTP status code.

cstyan:
CAS calls to the KV store can return HTTP status codes since the addition of the HA tracking code. 

If the return error of the CAS has an HTTP status code, we should return that. This will resolve the issue of the KV metrics showing 500's for all writes from the non-elected Prometheus replicas.

@tomwilkie

### DIFF
--- a/pkg/ring/kv/metrics.go
+++ b/pkg/ring/kv/metrics.go
@@ -2,8 +2,10 @@ package kv
 
 import (
 	"context"
+	"strconv"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/weaveworks/common/httpgrpc"
 	"github.com/weaveworks/common/instrument"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/ring/kv/metrics.go
```diff
@@ -18,6 +20,17 @@ func init() {
...
 
+// errorCode converts an error into an HTTP status code, modified from weaveworks/common/instrument
+func errorCode(err error) string {
+	if err == nil {
+		return "200"
+	}
+	if resp, ok := httpgrpc.HTTPResponseFromError(err); ok {
+		return strconv.Itoa(int(resp.GetCode()))
+	}
+	return "500"
```
pstibrany: This is a very strong assumption, that is not true at all when using memberlist-based client. (There are no HTTP calls at all)

pkg/ring/kv/metrics.go
```diff
@@ -18,6 +20,17 @@ func init() {
 	requestDuration.Register()
 }
 
+// errorCode converts an error into an HTTP status code, modified from weaveworks/common/instrument
+func errorCode(err error) string {
+	if err == nil {
+		return "200"
+	}
+	if resp, ok := httpgrpc.HTTPResponseFromError(err); ok {
```
pstibrany: This only works for errors from GRPC layer. I don't think etcd or consul do.

cstyan:
@pstibrany This change is to allow us to differentiate between actual errors and short circuit exits from CAS calls within the HA tracking code: https://github.com/cortexproject/cortex/blob/master/pkg/distributor/ha_tracker.go#L237-L239
https://github.com/cortexproject/cortex/blob/master/pkg/distributor/ha_tracker.go#L220-L225

Issue #1800, opened 2019-11-08
TestWatchKey Flake


bboreham:
Observed at https://circleci.com/gh/cortexproject/cortex/12434
Test passed on subsequent build.

pstibrany:
It's my new test, I'll make it bit less strict. Thanks.

pracucci:
The PR has been merged, so I'm closing this issue.

PR #1808, opened 2019-11-11
Modified TestWatchKey test to make it less strict.

pstibrany:
Test now uses longer waits between time updates, has longer total time (3s) and expects to see fewer values (75%).

At the same time, test uses same structure as TestWatchPrefix:
- start watcher early, before generator even starts
- report observed values through buffered channel
- end test early if we have seen enough values

Fixes issue #1800.

### DIFF
--- a/pkg/ring/kv/kv_test.go
+++ b/pkg/ring/kv/kv_test.go
@@ -92,12 +92,26 @@ func TestNilCAS(t *testing.T) {
 func TestWatchKey(t *testing.T) {
 	const key = "test"
 	const max = 100
-	const sleep = 10 * time.Millisecond
+	const sleep = 15 * time.Millisecond
+	const totalTestTimeout = 3 * max * sleep
+	const expectedFactor = 0.75 // we may not see every single value
 
 	withFixtures(t, func(t *testing.T, client Client) {
-		ch := make(chan struct{})
+		observedValuesCh := make(chan string, max)
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
 		go func() {
-			defer close(ch)
+			// Start watching before we even start generating values.
+			// Values will be buffered in the channel.
+			client.WatchKey(ctx, key, func(value interface{}) bool {
+				observedValuesCh <- value.(string)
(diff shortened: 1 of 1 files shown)


pstibrany:
> I'm wondering if we could rewrite the test using a `sync.Cond` used by `client.CAS()` goroutine and `client.WatchKey()` gorouting. We CAS once then `cond.Wait()`, then the `client.WatchKey()` will file and we assert there and call `cond.Signal()` so that another CAS will occur, and so on. The expected behaviour is that any `CAS()` will have triggered a `WatchKey()`. Makes sense?

I think we can achieve the same behaviour by using unbuffered channel.

pracucci:
> I think we can achieve the same behaviour by using unbuffered channel.

Right, should be the same, agree.

pstibrany:
> Right, should be the same, agree.

Actually, it's not... your suggestion is about syncing CAS and WatchKey, while buffered channel would sync watcher and reporting goroutine.

Anyway, question is what do we even test here. When doing watching test, I think we need to make sure that 1) we are receiving *some* watch notifications, 2) we have seen the latest one [we don't check for this].

You are right that it may still flake under bad conditions. But I think adding sync between sender and notifications defeats the purpose of the test.

pracucci:
> But I think adding sync between sender and notifications defeats the purpose of the test.

I see your point. I don't have a better suggestion, and the proposed change improve the situation, so consider it still approved by me 😉

pracucci:
@pstibrany Could you rebase just to make sure tests still pass? I remember I did extensively review this (non trivial) code change and LGTM.

pstibrany:
Rebased. 🤞

pracucci:
Today we've got [another CI run failing because of this](https://circleci.com/gh/cortexproject/cortex/16291):
```
--- FAIL: TestWatchKey (4.18s)
    --- FAIL: TestWatchKey/etcd (2.18s)
        kv_test.go:141: expected at least 90% observed values, got 89% (observed count: 89)
```

I think this PR would solve (or at least reduce the change of) this flaky test.

Issue #1787, opened 2019-11-05
Remove support for storing ring with denormalised tokens


pstibrany:
New code should only store normalised tokens to the ring. I'd like to remove support for `-ingester.normalise-tokens` flag by silently ignoring it and treating it as true. Reading from ring would still understand denormalised tokens as upgrade path from older version, but all other code handling denormalised tokens would be removed.

pstibrany:
I totally forgot there is an issue for this. Closed by #1809 and #2034.

PR #1809, opened 2019-11-11
Removed support for writing denormalised tokens in the ring.

### LINKED ISSUE #1877: Inconsistent behaviour of `Desc.AddIngester`
Before the introduction of normalized tokens, `Desc.AddIngester` used to append tokens to the existing ones. But currently, it works as `append` when it's not normalized tokens and `replace` when it's normalized tokens.

https://github.com/cortexproject/cortex/blob/08ddf88e082352a89a6daa08fad6c7064d7a731a/pkg/ring/model.go#L38-L63

Looking at places where `Desc.AddIngester` is used, `append` seems to be the desired behaviour. Discovered in #1750 and @pstibrany plans to fix it in #1809

pstibrany:
Removed support for writing denormalised tokens in the ring.

Still supported:
- claiming tokens from another ingester, that is using denormalised tokens
- finding denormalised tokens in the ring, when joining

Other ingesters can still use denormalised tokens, migrateRing takes care of that.

Targeted for Cortex 0.5.0. ~No changelog written yet.~

Issue #1787
Fixes #1877

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
+* [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored.
 
 ## 0.4.0 / 2019-12-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
-* [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored. #1809
+* [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored. If you want to switch back to using denormalised tokens, you need to downgrade to Cortex 0.4.0. Previous versions don't handle claiming tokens from normalised ingesters correctly. #1809
 
 ## 0.4.0 / 2019-12-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
-* [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored.
+* [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored. #1809
 
 ## 0.4.0 / 2019-12-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@
   * `ruler.poll-interval` has been added to specify the interval in which to poll new rule groups.
 * [CHANGE] Use relative links from /ring page to make it work when used behind reverse proxy. #1896
 * [CHANGE] Deprecated `-distributor.limiter-reload-period` flag. #1766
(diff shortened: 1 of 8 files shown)


docs/arguments.md
```diff
@@ -234,9 +234,9 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
 
 - `-ingester.normalise-tokens`
 
-   Write out "normalised" tokens to the ring.  Normalised tokens consume less memory to encode and decode; as the ring is unmarshalled regularly, this significantly reduces memory usage of anything that watches the ring.
+   Deprecated. New ingesters always write "normalised" tokens to the ring. Normalised tokens consume less memory to encode and decode; as the ring is unmarshalled regularly, this significantly reduces memory usage of anything that watches the ring.
```
pracucci: This change should be also clearly stated out in the changelog.

pkg/ring/lifecycler.go
```diff
@@ -82,7 +82,7 @@ func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.Flag
 	f.DurationVar(&cfg.ObservePeriod, prefix+"observe-period", 0*time.Second, "Observe tokens after generating to resolve collisions. Useful when using gossiping ring.")
 	f.DurationVar(&cfg.MinReadyDuration, prefix+"min-ready-duration", 1*time.Minute, "Minimum duration to wait before becoming ready. This is to work around race conditions with ingesters exiting and updating the ring.")
 	flagext.DeprecatedFlag(f, prefix+"claim-on-rollout", "DEPRECATED. This feature is no longer optional.")
-	f.BoolVar(&cfg.NormaliseTokens, prefix+"normalise-tokens", false, "Store tokens in a normalised fashion to reduce allocations.")
+	flagext.DeprecatedFlag(f, prefix+"normalise-tokens", "DEPRECATED. This feature is no longer optional")
```
pracucci: Add a `.` at the end of the description sentence, please.

pkg/ring/model.go
```diff
@@ -45,18 +45,17 @@ func (d *Desc) AddIngester(id, addr string, tokens []uint32, state IngesterState
...
 
-	if normaliseTokens {
-		ingester.Tokens = tokens
-	} else {
-		for _, token := range tokens {
-			d.Tokens = append(d.Tokens, TokenDesc{
-				Token:    token,
-				Ingester: id,
-			})
+	// delete any reference from d.Tokens (can happen if denormalized tokens for this ingester
```
pracucci: I'm missing some context. Can you help me understand this change, so that I can make a more conscious review, please?

docs/arguments.md
pstibrany: I don't want to update changelog yet, as 0.4.0 is going to be released soon (13th nov), and then I'll need to change it anyway. But you're right, this needs to go to changelog.

pkg/ring/model.go
pstibrany: See pkg/ring/lifecycler.go:549, autoJoin method.

When ingester finds that it already owns some tokens during join, it will use those tokens, instead of generating new ones. Those tokens could have been normalized or not. If they were not normalized (i.e. stored in d.Tokens), we need to remove them, otherwise they would be stored at two places (denormalized in `d.Tokens`, but also normalized tokens in `Ingesters[id].Tokens`).

pstibrany: Also, this behavior is used by updated test.

pracucci: I would rephrase the comment for clarity, explaining we do `delete any denormalized token for this ingester, given all ingester tokens are going to be stored normalized. It may happen to find denormalized tokens when joining the ring (ie. the previous instance of the same ingester was running with denormalized tokens).`

pracucci: Right, got it, thanks!

pstibrany: I've updated comment, and also added test for this. It turns out that my previous test was checking different scenario (takeover of denormalised tokens by different ingester), and commenting out this code didn't make it fail.

docs/arguments.md
pstibrany: Added changelog entry now.

pkg/ring/model.go
```diff
@@ -45,18 +45,18 @@ func (d *Desc) AddIngester(id, addr string, tokens []uint32, state IngesterState
 		Addr:      addr,
 		Timestamp: time.Now().Unix(),
 		State:     state,
+		Tokens:    tokens,
```
codesome: Created an issue for this just to not miss out: https://github.com/cortexproject/cortex/issues/1877

pstibrany: I've added comment to AddIngester about tokens being replaced and test to make sure it will not break.

pstibrany:
@gouthamve why 'type/breaking' label? We know that upgrade to normalised tokens works fine, and so making them mandatory won't break anything. (Perhaps if someone is upgrading from VERY old Cortex versions that didn't support normalisation at all?)

pkg/ring/lifecycler.go
```diff
@@ -50,7 +50,7 @@ type LifecyclerConfig struct {
 	JoinAfter        time.Duration `yaml:"join_after,omitempty"`
 	MinReadyDuration time.Duration `yaml:"min_ready_duration,omitempty"`
 	UnusedFlag       bool          `yaml:"claim_on_rollout,omitempty"` // DEPRECATED - left for backwards-compatibility
```
gouthamve: Can we move the unused flags to bottom of the struct definition?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
   * `ruler.poll-interval` has been added to specify the interval in which to poll new rule groups.
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
+* [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored. If you want to switch back to using denormalised tokens, you need to downgrade to Cortex 0.4.0. Previous versions don't handle claiming tokens from normalised ingesters correctly. #1809
```
gouthamve: Change should be on top.

pkg/ring/lifecycler.go
pstibrany: Done.

CHANGELOG.md
pstibrany: Moved.

PR #1811, opened 2019-11-12
Added user label to cortex_ingester_memory_series metric

pracucci:
**What this PR does**:
In the PR #1760 I've introduced a global limit on the max series per user and metric, but unfortunately we don't have a per-tenant metric to give us an overview on the current per-tenant series utilization, in order to fine tune the global limit.

For this reason, in this PR I'm suggesting to add the `user` label to the `cortex_ingester_memory_series` metric, like we already have for `cortex_ingester_memory_series_created_total` and `cortex_ingester_memory_series_removed_total`.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
Approving, but we really need to find a different way of having access to this information. Adding user label to more metrics is a cause to metrics explosion :-(

pracucci:
Closing cause - with a lower resolution - should be able to approximate via `cortex_ingester_memory_series_created_total - cortex_ingester_memory_series_removed_total`

PR #1814, opened 2019-11-12
use tempdir when xfering tsdb

thorfour:

Uses a temp directory when writing a transfer, to avoid corruptions errors. Will use a different temp dir during each xfer retry.

### DIFF
--- a/pkg/ingester/ingester_test.go
+++ b/pkg/ingester/ingester_test.go
@@ -3,8 +3,11 @@ package ingester
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"math"
+	"math/rand"
 	"net/http"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"sync"
@@ -571,4 +574,22 @@ func BenchmarkIngesterPush(b *testing.B) {
 		})
 	}
 
+}
+
+func TestRemoveEmptyDir(t *testing.T) {
+
+	// remove dir that dne
+	require.NoError(t, removeEmptyDir(fmt.Sprintf("%v", rand.Int63())))
+
+	// remove empty dir
+	dir, err := ioutil.TempDir("", "TestRemoveEmptyDir")
+	require.NoError(t, err)
+	require.NoError(t, removeEmptyDir(dir))
+
+	// remove non-empty dir
+	dir, err = ioutil.TempDir("", "TestRemoveEmptyDir")
+	require.NoError(t, err)
+
+	ioutil.WriteFile(filepath.Join(dir, "tempfile"), []byte("hello world"), 0777)
+	require.NotNil(t, removeEmptyDir(dir))
 }
--- a/pkg/ingester/lifecycle_test.go
+++ b/pkg/ingester/lifecycle_test.go
@@ -1,9 +1,12 @@
 package ingester
 
 import (
+	"fmt"
 	"io"
-	"io/ioutil"
 	"math"
+	"math/rand"
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
@@ -419,10 +422,8 @@ func TestV2IngesterTransfer(t *testing.T) {
 	limits, err := validation.NewOverrides(defaultLimitsTestConfig())
 	require.NoError(t, err)
 
-	dir1, err := ioutil.TempDir("", "tsdb")
-	require.NoError(t, err)
-	dir2, err := ioutil.TempDir("", "tsdb")
-	require.NoError(t, err)
+	dir1 := filepath.Join(os.TempDir(), fmt.Sprintf("TestV2IngesterTransfer-%v", rand.Int63()))
+	dir2 := filepath.Join(os.TempDir(), fmt.Sprintf("TestV2IngesterTransfer-%v", rand.Int63()))
 
 	// Start the first ingester, and get it into ACTIVE state.
 	cfg1 := defaultIngesterTestConfig()
--- a/pkg/ingester/lifecycle_test.go
+++ b/pkg/ingester/lifecycle_test.go
@@ -1,12 +1,10 @@
 package ingester
 
 import (
-	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
-	"math/rand"
 	"os"
-	"path/filepath"
 	"testing"
 	"time"
 
@@ -422,8 +420,11 @@ func TestV2IngesterTransfer(t *testing.T) {
 	limits, err := validation.NewOverrides(defaultLimitsTestConfig())
 	require.NoError(t, err)
 
-	dir1 := filepath.Join(os.TempDir(), fmt.Sprintf("TestV2IngesterTransfer-%v", rand.Int63()))
-	dir2 := filepath.Join(os.TempDir(), fmt.Sprintf("TestV2IngesterTransfer-%v", rand.Int63()))
+	dir1, err := ioutil.TempDir("", "tsdb")
+	require.NoError(t, err)
+	dir2, err := ioutil.TempDir("", "tsdb")
+	require.NoError(t, err)
+	require.NoError(t, os.Remove(dir2)) // remove the destination dir so there isn't a move conflict
 
 	// Start the first ingester, and get it into ACTIVE state.
 	cfg1 := defaultIngesterTestConfig()
@@ -522,4 +523,10 @@ func TestV2IngesterTransfer(t *testing.T) {
 	response, err = ing2.Query(ctx, request)
 	require.NoError(t, err)
 	assert.Equal(t, expectedResponse, response)
+
+	// Assert the data is in the expected location of dir2
+	files, err := ioutil.ReadDir(dir2)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(files))
+	require.Equal(t, "1", files[0].Name())
 }
--- a/pkg/ingester/transfer.go
+++ b/pkg/ingester/transfer.go
@@ -38,6 +38,10 @@ var (
 		Name: "cortex_ingester_sent_files",
 		Help: "The total number of files sent by this ingester whilst leaving.",
 	})
+	receivedFiles = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "cortex_ingester_received_files",
+		Help: "The total number of files received by this ingester whilst joining",
+	})
 	receivedBytes = prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "cortex_ingester_received_bytes",
 		Help: "The total number of bytes received by this ingester whilst joining",
@@ -52,6 +56,7 @@ func init() {
 	prometheus.MustRegister(receivedChunks)
 	prometheus.MustRegister(sentFiles)
 	prometheus.MustRegister(receivedBytes)
+	prometheus.MustRegister(receivedFiles)
 }
 
 // TransferChunks receives all the chunks from another ingester.
@@ -221,7 +226,8 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 		}
 		defer os.RemoveAll(tmpDir)
 
-		totalXfer := 0
+		bytesXfer := 0
+		filesXfer := 0
 
 		files := make(map[string]*os.File)
 		defer func() {
@@ -249,7 +255,7 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 					return err
 				}
 			}
-			totalXfer += len(f.Data)
+			bytesXfer += len(f.Data)
 
 			createfile := func(f *client.TimeSeriesFile) (*os.File, error) {
 				dir := filepath.Join(tmpDir, filepath.Dir(f.Filename))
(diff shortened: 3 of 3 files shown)


pkg/ingester/transfer.go
```diff
@@ -214,6 +214,13 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 	fromIngesterID := ""
 
 	xfer := func() error {
+
+		dir, err := ioutil.TempDir("", "tsdb_xfer")
```
pracucci: I would suggest to let configure this as well (ie. `-experimental.tsdb.transfer-temp-dir` / `transfer_temp_dir`). We should also `Validate()` that `TransferTempDir != Dir`.

pkg/ingester/transfer.go
```diff
@@ -284,7 +290,8 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 		receivedFiles.Add(float64(filesXfer))
 		level.Error(util.Logger).Log("msg", "Total files xfer", "from_ingester", fromIngesterID, "num", filesXfer)
```
pracucci: Unrelated from this changes, but I believe this should be an `Info`

pkg/ingester/transfer.go
```diff
@@ -244,9 +251,8 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 			}
 			filesXfer++
```
pracucci: Unrelated from these changes, but shouldn't this counter incredemented only when a new file is created instead of each time a data chunk is transferred?

pkg/ingester/transfer.go
```diff
@@ -284,7 +290,8 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 		receivedFiles.Add(float64(filesXfer))
 		level.Error(util.Logger).Log("msg", "Total files xfer", "from_ingester", fromIngesterID, "num", filesXfer)
 
-		return nil
+		// Move the tmpdir to the final location
+		return os.Rename(dir, i.cfg.TSDBConfig.Dir)
```
pracucci: From the golang doc:

> Rename renames (moves) oldpath to newpath. If newpath already exists and is not a directory, Rename replaces it.

What if the newpath (`i.cfg.TSDBConfig.Dir`) already exists and is a directory?

pkg/ingester/transfer.go
pracucci: I would suggest to rename `dir` to `tempDir` for clarity.

pkg/ingester/transfer.go
```diff
@@ -214,6 +214,13 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 	fromIngesterID := ""
 
 	xfer := func() error {
+
+		dir, err := ioutil.TempDir("", "tsdb_xfer")
+		if err != nil {
+			return err
+		}
+		defer os.Remove(dir)
```
pracucci: Two things:
1. This expects that `dir` is empty. I think you should use `RemoveAll()` in this case.
2. Do we really want to remove it? On success shouldn't exist anymore because it has been moved, while on failure may be handy to keep them for debugging? _It's an open question, not sure_

pkg/ingester/transfer.go
thorfour: yup fixed.

pkg/ingester/transfer.go
thorfour: gonna change this to be a bytes xfer counter instead.

pkg/ingester/transfer.go
thorfour: Yea, removall would be correct. I'm not sure what would be gained debug wise from keeping it though.

pkg/ingester/transfer.go
thorfour: done

thorfour: seems like an unnecessary configuration option, `TempDir` ensures that there isn't a naming conflict so it wont have that collision.

pkg/ingester/transfer.go
pracucci: Do you mind adding the count of transferred files, please? I believe both are useful for debugging (I'm currently debugging a transfer which lead to WAL loss).

pkg/ingester/transfer.go
pracucci: Ok, `RemoveAll()` would be then 😉

pkg/ingester/transfer.go
thorfour: I don't know what the correct behaviour should be here. If I had a leftover ingester that rejoined I would expect it to overwrite what it had with it's new data

pkg/ingester/transfer.go
thorfour: Changed it to removeall

pkg/ingester/transfer.go
pracucci: Ok

pkg/ingester/transfer.go
pracucci: Does `Rename()` replace the destination if it's a non empty directory? I just would like to double check it does what we expect it to do.

pkg/ingester/transfer.go
```diff
@@ -38,9 +38,9 @@ var (
 		Name: "cortex_ingester_sent_files",
 		Help: "The total number of files sent by this ingester whilst leaving.",
 	})
-	receivedFiles = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "cortex_ingester_received_files",
-		Help: "The total number of files received by this ingester whilst joining",
+	receivedBytes = prometheus.NewCounter(prometheus.CounterOpts{
```
pracucci: I'm not much convinced about this change, because of two reasons:
1. Drift between `cortex_ingester_sent_files` and `cortex_ingester_received_bytes `
2. Drift between chunks storage (`cortex_ingester_sent_chunks` and `cortex_ingester_received_chunks`) which count chunks and TSDB which count bytes

I would suggest to rollback to count files (and fix the count).

thorfour: File count is kinda weird though since they're variable sized right? Chunks are a static size so one can infer how much data was actually sent. I guess we could count number of batches sent? I guess my question is what type of info are we trying to glean here? I would imagine it would be much more useful to see how much data is being sent on xfer, vs number of arbitrarily sized files.

pracucci: Given we're in doubt and given this metric cardinality is low, why don't we keep the current one and we add bytes as well?

thorfour: Should the former one be batches transferred or files. If the latter we have to do some tracking of filenames to determine the correct file count.

pracucci: I would suggest to count files and bytes. The files count should be incremented only when we create a new file in `TransferTSDB()` (should just be a matter of placing the counter increment in the right place, no extra logic should be required).

pkg/ingester/transfer.go
thorfour: Confirmed it errors out on a directory (empty or not)

pkg/ingester/transfer.go
thorfour: Good call. Done

pkg/ingester/transfer.go
thorfour: So should it attempt to remove any preexisting directory?

pracucci: I'm dubious cause it may lead to data loss (ie. think about restarting a crashed ingester). From my perspective the ideal logic should be: OK if the destination directory not exist or empty, fail if not empty.

pkg/ingester/lifecycle_test.go
```diff
@@ -419,10 +422,8 @@ func TestV2IngesterTransfer(t *testing.T) {
 	limits, err := validation.NewOverrides(defaultLimitsTestConfig())
 	require.NoError(t, err)
 
-	dir1, err := ioutil.TempDir("", "tsdb")
-	require.NoError(t, err)
-	dir2, err := ioutil.TempDir("", "tsdb")
-	require.NoError(t, err)
+	dir1 := filepath.Join(os.TempDir(), fmt.Sprintf("TestV2IngesterTransfer-%v", rand.Int63()))
```
pracucci: Out of curiosity: why do you need this change? Did you get failing or flaky tests with the previous code?

pkg/ingester/transfer.go
```diff
@@ -281,10 +293,12 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
...
 		}
 
+		receivedBytes.Add(float64(bytesXfer))
 		receivedFiles.Add(float64(filesXfer))
-		level.Error(util.Logger).Log("msg", "Total files xfer", "from_ingester", fromIngesterID, "num", filesXfer)
+		level.Info(util.Logger).Log("msg", "Total xfer", "from_ingester", fromIngesterID, "files", filesXfer, "bytes", bytesXfer)
 
-		return nil
+		// Move the tmpdir to the final location
+		return os.Rename(tmpDir, i.cfg.TSDBConfig.Dir)
```
pracucci: For the sake of correctness, I would suggest you to add an assertion to `TestV2IngesterTransfer`, checking:
1. `i.cfg.TSDBConfig.Dir` is empty (or not existent) before transfer
2. `i.cfg.TSDBConfig.Dir` is not empty and contains the expected userIDs after transfer

If I'm not missing anything, right now that test doesn't guarantee that the transfer occurred to the right location.

pkg/ingester/lifecycle_test.go
thorfour: ugh that was when I was debugging and I forgot to revert that. Will revert

pkg/ingester/transfer.go
thorfour: Added assertion the data is where expected.

pkg/ingester/transfer.go
thorfour: Added behaviour to delete an empty dir, but throw an error on a non-empty dir.

pkg/ingester/ingester_test.go
```diff
@@ -572,3 +575,21 @@ func BenchmarkIngesterPush(b *testing.B) {
...
+	dir, err := ioutil.TempDir("", "TestRemoveEmptyDir")
+	require.NoError(t, err)
+	require.NoError(t, removeEmptyDir(dir))
+
+	// remove non-empty dir
+	dir, err = ioutil.TempDir("", "TestRemoveEmptyDir")
+	require.NoError(t, err)
+
+	ioutil.WriteFile(filepath.Join(dir, "tempfile"), []byte("hello world"), 0777)
+	require.NotNil(t, removeEmptyDir(dir))
```
pracucci: You can use `require.Error()` here, to signal more clearly the intention.

thorfour: Yea but that looks for a specific error, and I got lazy

pracucci: Oh you're right! Then your is better :)

thorfour: :)

pkg/ingester/transfer.go
```diff
@@ -42,6 +42,10 @@ var (
 		Name: "cortex_ingester_received_files",
 		Help: "The total number of files received by this ingester whilst joining",
 	})
+	receivedBytes = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "cortex_ingester_received_bytes",
```
gouthamve: `_bytes_total`

pkg/ingester/transfer.go
```diff
@@ -51,6 +55,7 @@ func init() {
 	prometheus.MustRegister(sentChunks)
 	prometheus.MustRegister(receivedChunks)
 	prometheus.MustRegister(sentFiles)
+	prometheus.MustRegister(receivedBytes)
```
gouthamve: Curious why there is no `sentBytes`

pkg/ingester/transfer.go
```diff
@@ -214,6 +219,20 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 	fromIngesterID := ""
 
 	xfer := func() error {
+
+		// Validate the final directory is empty, if it exists and is empty delete it so a move can succeed
+		err := removeEmptyDir(i.cfg.TSDBConfig.Dir)
+		if err != nil {
+			return errors.Wrap(err, "Non-empty tsdb directory")
```
gouthamve: I think, "remove existing TSDB directory" might be appropriate, as there could be several reasons for the error, not just non empty.

pkg/ingester/transfer.go
thorfour: done.

pkg/ingester/transfer.go
thorfour: Simply overlooked. Added.

pkg/ingester/transfer.go
thorfour: Good call; changed.

PR #1816, opened 2019-11-12
blocks/tsdb: support block range lists

thorfour:
Adds support for user defined tsdb block ranges.

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -263,16 +263,11 @@ func (i *Ingester) getOrCreateTSDB(userID string) (*tsdb.DB, error) {
 
 			udir := i.userDir(userID)
 
-			var ranges []int64
-			for _, t := range i.cfg.TSDBConfig.BlockRanges {
-				ranges = append(ranges, int64(t/time.Millisecond))
-			}
-
 			// Create a new user database
 			var err error
 			db, err = tsdb.Open(udir, util.Logger, nil, &tsdb.Options{
 				RetentionDuration: uint64(i.cfg.TSDBConfig.Retention / time.Millisecond),
-				BlockRanges:       ranges,
+				BlockRanges:       i.cfg.TSDBConfig.BlockRanges.ToMillisecondRanges(),
 				NoLockfile:        true,
 			})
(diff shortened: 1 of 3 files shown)


pkg/storage/tsdb/config.go
```diff
@@ -34,14 +35,34 @@ type Config struct {
...
 }
 
+// RangeList is the block ranges for a tsdb
+type RangeList []time.Duration
+
+// String implements the flag.Var interface
+func (b RangeList) String() string { return "RangeList is the block ranges for a tsdb" }
+
+// Set implements the flag.Var interface
+func (b RangeList) Set(s string) error {
```
pracucci: Would you mind adding unit tests on this config, please?

pkg/storage/tsdb/config.go
```diff
@@ -34,14 +35,34 @@ type Config struct {
 	GCS gcs.Config `yaml:"gcs"`
 }
 
+// RangeList is the block ranges for a tsdb
+type RangeList []time.Duration
```
pracucci: What do you think about naming the table `DurationList` since it's just a list of `time.Duration`?

pkg/ingester/ingester_v2.go
```diff
@@ -263,11 +263,16 @@ func (i *Ingester) getOrCreateTSDB(userID string) (*tsdb.DB, error) {
 
 			udir := i.userDir(userID)
 
+			var ranges []int64
+			for _, t := range i.cfg.TSDBConfig.BlockRanges {
```
pracucci: What do you think about adding a `ToMillisecondRanges() []int64` function to the data type used in the config, so that here we just do `BlockRanges: i.cfg.TSDBConfig.BlockRanges.ToMillisecondRanges()`? Makes sense?

thorfour: Makes sense

pkg/storage/tsdb/config.go
thorfour: Fair enough

pkg/storage/tsdb/config.go
thorfour: Will do

PR #1818, opened 2019-11-13
Fix TSDB creation conflicting with transfer

pracucci:
**What this PR does**:
In this PR:
- Do not allow to query PENDING ingesters (consider them unhealthy)
- Do not create TSDB if not exist when querying an ingester
- Do not allow to create TSDB if the ingester is not in ACTIVE state

_Suggest to enable "Hide whitespace changes" when reviewing for an easier diff._

**Why?**
I found an issue leading to data loss and/or TSDB wal corruption, which should be fixed by this PR. What happens:

- Right now, a TSDB is created also when querying (not need, fixed)
- The querier consider healthy an ingester in the `PENDING` state, so a query request can hit an ingester even if the `ACTIVE` state has not been reached yet (not desired, fixed)

If a query hit a new ingester in the `PENDING` state before the transfer starts (it switches to `JOINING` then), the query creates a TSDB opening the wal segment `00000000`. The subsequent transfer will overwrite the wal segments, while the `00000000` file reference is still kept by the opened TSDB.

Moreover, if it get transferred a wal where a checkpoint already occurred, the `00000000` segment is not part of the transfer but the segment 0 is created by the opened TSDB which leads to an inconsistent segments numbering on disk (ie. 0, checkpoint.6, 7, 8, 9) which cause TSDB to fail a subsequent re-opening.

_No reported issue._

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -322,13 +322,17 @@ func (i *Ingester) getOrCreateTSDB(userID string) (*tsdb.DB, error) {
 		return nil, err
 	}
 
-	// Create a new shipper for this database
+	// Thanos shipper requires at least 1 external label to be set. For this reason,
+	// we set the tenant ID as external label and we'll filter it out when reading
+	// the series from the storage.
 	l := lbls.Labels{
 		{
 			Name:  cortex_tsdb.TenantIDExternalLabel,
 			Value: userID,
 		},
 	}
+
+	// Create a new shipper for this database
 	s := shipper.New(util.Logger, nil, udir, &Bucket{userID, i.TSDBState.bucket}, func() lbls.Labels { return l }, metadata.ReceiveSource)
 	i.done.Add(1)
 	go func() {
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -325,7 +325,7 @@ func (i *Ingester) getOrCreateTSDB(userID string) (*tsdb.DB, error) {
 	// Create a new shipper for this database
 	l := lbls.Labels{
 		{
-			Name:  "user",
+			Name:  cortex_tsdb.TenantIDExternalLabel,
 			Value: userID,
 		},
 	}
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -3,7 +3,6 @@ package ingester
 import (
 	"fmt"
 	"net/http"
-	"path/filepath"
 	"time"
 
 	"github.com/cortexproject/cortex/pkg/ingester/client"
@@ -24,6 +23,10 @@ import (
 	"github.com/weaveworks/common/user"
 	"golang.org/x/net/context"
 	old_ctx "golang.org/x/net/context"
+)
+
+const (
+	errTSDBCreateIncompatibleState = "cannot create a new TSDB while the ingester is not in active state (current state: %s)"
 )
 
 // TSDBState holds data structures used by the TSDB storage engine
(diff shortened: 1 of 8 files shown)


pkg/ingester/transfer.go
```diff
@@ -284,6 +284,30 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
...
+		userIDs, err := ioutil.ReadDir(i.cfg.TSDBConfig.Dir)
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("unable to list TSDB users in %s", i.cfg.TSDBConfig.Dir))
+		}
+
+		for _, user := range userIDs {
+			userID := user.Name()
+
+			level.Info(util.Logger).Log("msg", fmt.Sprintf("Loading TSDB for user %s", userID))
+			_, err = i.getOrCreateTSDB(userID, true)
```
thorfour: Do we maybe want to do this concurrently?

pracucci: I agree, but I would suggest to keep it for a dedicated PR (it's in my backlog) in order to keep this PR a bit easier (running it in parallel means we need a queue to limit concurrent jobs, ideally let configure the queue size, etc... so some further code will be required).

thorfour: Sounds good

pkg/ingester/transfer.go
```diff
@@ -284,6 +284,30 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 		receivedFiles.Add(float64(filesXfer))
 		level.Error(util.Logger).Log("msg", "Total files xfer", "from_ingester", fromIngesterID, "num", filesXfer)
 
+		// At this point all TSDBs have been received, so we can proceed loading TSDBs in memory.
+		// This is required because of two reasons:
+		// 1. No WAL replay performance penalty once the ingester switches to ACTIVE state
+		// 2. If a query is received on user X, for which the TSDB has been transferred, before
+		//    the first series is ingested, if we don't open the TSDB the query will return an
+		//    empty result (because the TSDB is opened only on first push or transfer)
+		userIDs, err := ioutil.ReadDir(i.cfg.TSDBConfig.Dir)
```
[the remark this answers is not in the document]
pracucci: I tried to reproduce the case with a test and I've realized the transfer doesn't start if there are no TSDBs, due to this:

```
func (i *Ingester) v2TransferOut(ctx context.Context) error {
	if len(i.TSDBState.dbs) == 0 {
		level.Info(util.Logger).Log("msg", "nothing to transfer")
		return nil
	}
```

Do you think is it still a case to cover/fix? It's trivial adding an "if directory exist", but if not required I would keep the correct code (which fails) as a signal of "something went unexpected".

Thoughts?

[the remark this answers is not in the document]
thorfour: Ah forgot that was in there. Okay, I think that's okay to leave as is then.

pkg/ingester/ingester_v2.go
```diff
@@ -247,58 +250,72 @@ func (i *Ingester) getTSDB(userID string) *tsdb.DB {
...
-			i.TSDBState.dbs[userID] = db
+	// We're ready to create the TSDB, however we must be sure that the ingester
+	// is in the ACTIVE state, otherwise it may conflict with the transfer in/out.
+	// The TSDB is created when the first series is pushed and this shouldn't happen
+	// to a non-ACTIVE ingester, however we want to protect from any bug, cause we
+	// may have data loss or TSDB WAL corruption if the TSDB is created before/during
+	// a transfer in occurs.
+	if !force {
+		ingesterState := i.lifecycler.GetState()
+		if ingesterState != ring.ACTIVE {
```
codesome: Nit: you can combine these 2, also combine it with the above `if !force` check to be single `if` statement.

pracucci: Sure, done.

pkg/ring/model.go
```diff
@@ -184,7 +184,7 @@ func (d *Desc) TokensFor(id string) (tokens, other []uint32) {
 func (i *IngesterDesc) IsHealthy(op Operation, heartbeatTimeout time.Duration) bool {
 	if op == Write && i.State != ACTIVE {
 		return false
-	} else if op == Read && i.State == JOINING {
+	} else if op == Read && i.State != ACTIVE && i.State != LEAVING {
```
tomwilkie: IIRC pending ingesters shouldn't have any tokens; how did you trigger this?

pracucci: It should occur in the paths where we use `forAllIngesters()`. I think one path is:
1. Querier receive a `LabelNames` request
2. Querier forwards the request to the distributor
3. The distributor runs `client.LabelNames()` in `distributor.forAllIngesters()`
4. `distributor.forAllIngesters()` calls `ring.GetAll()`
5. `ring.GetAll()` returns all ingesters which satisfy the `IsHealthy()`

pkg/ingester/ingester_v2.go
```diff
@@ -42,7 +45,7 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 		clientConfig: clientConfig,
 		metrics:      newIngesterMetrics(registerer),
 		limits:       limits,
-		chunkStore:   chunkStore,
+		chunkStore:   nil,
```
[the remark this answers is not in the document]
pracucci: I introduced intentionally to signal the future reader (and myself) that we didn't forget to init the chunkStore, but we intentionally wanna it to be nil.

pkg/ingester/ingester_v2_test.go
```diff
@@ -0,0 +1,180 @@
...
+		},
+		"allow to create TSDB if in ACTIVE state": {
+			state:       ring.ACTIVE,
+			expectedErr: nil,
+		},
+	}
+
+	for testName, testData := range tests {
+		testData := testData
+
```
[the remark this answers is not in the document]
pracucci: Correct, removed.

pkg/ring/model.go
pracucci: @bboreham What's your take on this? _If this change is a problem, I could rollback it in order to move one with this PR._

owen-d: Thanks for enumerating these.

owen-d:
Nice work! I appreciate the refactoring on the read path (`getTSDB` instead of `getOrCreateTSDB`).

codesome:
You got some merge conflicts now

pracucci:
@gouthamve I've fixed the conflicts after rebasing. May you help me moving forward with this PR, please?

codesome:
Active development going on here :) you got more conflicts

PR #1819, opened 2019-11-13
Ingester Push improvements

codesome:
Found some (small) possible optimization while debugging the WAL code. Reuse of some objects created in `ingester.append` by creating them in `ingester.Push` and passing them on has some improvements in push time (no changes in allocs or B/op). Along with that, updating metrics only once for `Push` call instead of in each `append` call.
```
benchmark                                        old ns/op     new ns/op     delta
BenchmarkIngesterPush/encoding=DoubleDelta-8     15628904      14997580      -4.04%
BenchmarkIngesterPush/encoding=Varbit-8          15608147      14949577      -4.22%
BenchmarkIngesterPush/encoding=Bigchunk-8        15625913      15132356      -3.16%
```
If the readability if getting bad with this change, we can only reuse the user state for now.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -287,11 +287,22 @@ func (i *Ingester) Push(ctx old_ctx.Context, req *client.WriteRequest) (*client.
 		i.userStatesMtx.RUnlock()
 		return &client.WriteResponse{}, nil
 	}
+	var discarded, totalAppends int
 	for _, ts := range req.Timeseries {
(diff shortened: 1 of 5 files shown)


pracucci:
I took a brief look at the change set and I'm wondering if the complexity in the changes is enough to justify a small performance improvement.

codesome:
I guess not. But I could change the code to only re-use the `state`, which has the maximum contribution to the reduced latency.

PR #1822, opened 2019-11-14
Fixes ARM build with const overflow.

cyriltovena:
Since #1727  ARM build fails with an error:

```
# github.com/cortexproject/cortex/pkg/ring/kv/memberlist
--
130 | vendor/github.com/cortexproject/cortex/pkg/ring/kv/memberlist/memberlist_client.go:619:21: constant 4294967295 overflows int
```

see https://github.com/grpc/grpc-go/pull/1471, it seems that this is cause by missing int conversion with MaxUint32.

I also run all tests `GOARCH=386 make test` all good now.

### DIFF
--- a/pkg/ring/kv/memberlist/memberlist_client.go
+++ b/pkg/ring/kv/memberlist/memberlist_client.go
@@ -616,7 +616,7 @@ func (m *Client) LocalState(join bool) []byte {
 			level.Error(util.Logger).Log("msg", "key too long", "key", key)
 			continue
 		}
-		if len(val.value) > math.MaxUint32 {
+		if uint(len(val.value)) > math.MaxUint32 {
 			level.Error(util.Logger).Log("msg", "value too long", "key", key, "value_length", len(val.value))
 			continue
 		}


bboreham:
This test is now a no-op on a 32-bit platform: it can never be true. 

I’m not immediately convinced that this is the right way to go.

pstibrany:
> This test is now a no-op on a 32-bit platform: it can never be true.
> 
> I’m not immediately convinced that this is the right way to go.

This is fine, because you're unlikely to have more than 4gb array in memory on 32-bit platform, so this check will never fail. This check is there only to make sure that we don't overflow 32-bit serialized value few lines below.

gouthamve:
I'm merging this as it's a Golang quirk.

pstibrany:
> bh, I am not super sure why the _compilation_ fails and what the compiler is doing.

[`len` function](https://golang.org/pkg/builtin/#len) returns `int`, which has different size on 32-bit and 64-bit systems. We compare output of `len` against `math.MaxUint32`, which doesn't fit into 32-bit int (it would fit into `uint32` though). That's why compiler complains when compiling for 32-bit systems.

Fix changes left side of comparison to `uint`, which makes compiler to convert `math.MaxUint32` constant to `uint` as well, which now fits.

PR #1830, opened 2019-11-15
Add /series support to TSDB storage

pracucci:
**What this PR does**:
Currently the `/series` API endpoint doesn't work with the TSDB storage, due to missing implementation of `MetricsForLabelMatchers()`. In this PR I'm addressing this issue.

_No issue created._

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -13,7 +13,6 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/validation"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/tsdb"
 	lbls "github.com/prometheus/prometheus/tsdb/labels"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
@@ -275,7 +274,7 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
 	defer q.Close()
 
 	// Run a query for each matchers set and collect all the results
-	added := model.FingerprintSet{}
+	added := map[string]struct{}{}
 	result := &client.MetricsForLabelMatchersResponse{
 		Metric: make([]*client.Metric, 0),
 	}
@@ -300,8 +299,8 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
 			// return the unique set of matching series, we do check if the series has
 			// already been added to the result
 			ls := seriesSet.At().Labels()
-			fp := client.Fingerprint(cortex_tsdb.FromLabelsToLegacyLabels(ls))
-			if _, ok := added[fp]; ok {
+			key := ls.String()
+			if _, ok := added[key]; ok {
 				continue
 			}
 
@@ -309,7 +308,7 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
 				Labels: cortex_tsdb.FromLabelsToLabelAdapters(ls),
 			})
 
-			added[fp] = struct{}{}
+			added[key] = struct{}{}
 		}
 
 		// In case of any error while iterating the series, we break
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -13,7 +13,6 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/validation"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
 	lbls "github.com/prometheus/prometheus/tsdb/labels"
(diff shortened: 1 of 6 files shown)


pracucci:
@thorfour and @codesome may you add to review this PR to your backlog, please?

pkg/ingester/ingester_v2.go
```diff
@@ -240,6 +218,78 @@ func (i *Ingester) v2LabelNames(ctx old_ctx.Context, req *client.LabelNamesReque
...
+	}
+
+	// Parse the request
+	from, to, matchersSet, err := client.FromMetricsForLabelMatchersRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a new instance of the TSDB querier
+	q, err := db.Querier(from.Unix()*1000, to.Unix()*1000)
```
thorfour: `nit` prefer `time.Microsecond` here instead

pracucci: It's actually even easier: `from`  and `to` are `model.Time` (not the Go time pkg), and are already expressed in milliseconds, so we've just to cast it to `int64`.

What's your take?

thorfour: that sounds good

pkg/storage/tsdb/labels.go
```diff
@@ -23,3 +26,48 @@ func FromLabelAdaptersToLabels(input []client.LabelAdapter) labels.Labels {
 
 	return result
 }
+
+// FromLabelsToLabelAdapters converts TSDB labels.labels to []LabelAdapter.
```
pstibrany: Is all this duplication of labels and conversions necessary? :-(

pkg/ingester/ingester_v2.go
```diff
@@ -240,6 +218,78 @@ func (i *Ingester) v2LabelNames(ctx old_ctx.Context, req *client.LabelNamesReque
...
+
+	// Create a new instance of the TSDB querier
+	q, err := db.Querier(int64(from), int64(to))
+	if err != nil {
+		return nil, err
+	}
+	defer q.Close()
+
+	// Run a query for each matchers set and collect all the results
+	added := model.FingerprintSet{}
```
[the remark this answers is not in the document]
pracucci: I could fix it here, but wouldn't be enough, because the result of this method is then merged by the [distributor](https://github.com/cortexproject/cortex/blob/master/pkg/distributor/distributor.go#L524) which uses a fingerprint-based map to build the result. I believe this fingerprint hash collision issue should be discussed and fixed in a separate PR, cause it's quite wide spread in Cortex.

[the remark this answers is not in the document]
pstibrany: > I believe this fingerprint hash collision issue should be discussed and fixed in a separate PR, cause it's quite wide spread in Cortex.

It's actually pretty well handled elsewhere in the Cortex, except one small bug when fetching labels. I'll take a look at what you're pointing to to understand your comment, but on the first sight, this code introduces new bug.

pkg/storage/tsdb/labels.go
pracucci: Thanks for questioning it. I re-checked it, and it should be safe to do a data type conversion only here.

What's about now?

pstibrany: I was questioning the need to have labels.Labels vs client.LabelAdapter vs legacy_labels.Labels... they all seem pretty much the same to me. I don't like it, but I assume there is a good reason :-(

pracucci: As of today, we need it in order to reuse the large part of Cortex which works with `client.LabelAdapter`. At least now we just do a data type conversion.

pkg/ingester/ingester_v2.go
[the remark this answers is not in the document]
pracucci: I've switched this specific part to a map of strings, avoiding to use any hashing at all. What's your take?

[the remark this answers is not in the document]
pstibrany: 👍  Thanks!

pkg/ingester/ingester_v2.go
```diff
@@ -272,6 +249,78 @@ func (i *Ingester) v2LabelNames(ctx old_ctx.Context, req *client.LabelNamesReque
...
+			if seriesSet.Err() != nil {
+				break
+			}
+
+			// Given the same series can be matched by multiple matchers and we want to
+			// return the unique set of matching series, we do check if the series has
+			// already been added to the result
+			ls := seriesSet.At().Labels()
+			key := ls.String()
+			if _, ok := added[key]; ok {
```
gouthamve: Just to be consistent with upstream, can we use this: https://github.com/prometheus/prometheus/blob/0ae4899c47fbcdab1b0bebedf213b4424f50f760/web/api/v1/api.go#L546-L550

It's an exposed function and I see it fitting perfectly here.

pracucci: I checked it. The `storage.NewMergeSeriesSet()` you pointed out works with the `storage` pkg structs, while here we work with `tsdb` structs (incompatible types). I checked the merge functions in `tsdb` pkg and looks a bit overkilled for what we need here: they merge series samples, while here we just need a unique set of labels.

_I would suggest to keep this code as is._

PR #1837, opened 2019-11-18
Fail query if an error occurs while converting TSDB series to chunks

pracucci:
**What this PR does**:
While doing tests on Cortex TSDB, I've realized a query doesn't fail if some series fail to be converted to chunks. There are a couple of warning logs, but from the user perspective we end up with a partial query result, which we would like to avoid in favour of correctness.

In this PR:
- Fail a query if an error occurs while converting TSDB series to chunks
- Handle `chunkend.Iterator.Err()` (was unchecked)

_No issue was created_

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/block.go
+++ b/pkg/querier/block.go
@@ -2,6 +2,7 @@ package querier
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"time"
 
@@ -142,7 +143,7 @@ func seriesToChunks(userID string, series *storepb.Series) ([]chunk.Chunk, error
 
 		enc, err := chunkenc.FromData(chunkenc.EncXOR, c.Raw.Data)
 		if err != nil {
-			return nil, errors.Wrap(err, "failed to initialize chunk from XOR encoded raw data")
+			return nil, errors.Wrap(err, fmt.Sprintf("failed to initialize chunk from XOR encoded raw data (series: %v min time: %d max time: %d)", lbls, c.MinTime, c.MaxTime))
 		}
 
 		it := enc.Iterator(nil)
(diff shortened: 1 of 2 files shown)


pracucci:
- @thorfour and @codesome may you review it, please?
- @tomwilkie may you add this PR to https://github.com/cortexproject/cortex/projects/4 , please?

bboreham:
I think the code was written that way because bugs were expected in encoding, and we didn't want a single error in the stored data to "poison" a whole set of results.

After this change, supposing such a bug happens again, what is the process to recover?
Perhaps the chunk reference should be logged so the operator can delete it manually?

pracucci:
> I think the code was written that way because bugs were expected in encoding, and we didn't want a single error in the stored data to "poison" a whole set of results.

@bboreham I agree enriching the error message and ensuring it gets logged.

To be on same page, please note this is a new function [recently introduced](https://github.com/cortexproject/cortex/pull/1695/files#diff-4ae88dc2abc8554f83d0aa2804c9d95dR117) with the experimental TSDB support, it's not something which was used before that.

pracucci:
> Perhaps the chunk reference should be logged so the operator can delete it manually?

@bboreham I've enhanced the error messages and logged them. A log example is:
```
level=error ts=2019-11-21T09:26:09.0919771Z caller=block.go:116 msg="failed converting TSDB series to Cortex chunks" err="failed to initialize chunk from XOR encoded raw data (series: {__name__=\"avalanche_metric_mmmmm_0_1\", cycle_id=\"0\", label_key_kk_0=\"label_val_vv_0\", label_key_kk_1=\"label_val_vv_1\", label_key_kk_2=\"label_val_vv_2\", label_key_kk_3=\"label_val_vv_3\", label_key_kk_4=\"label_val_vv_4\", series_id=\"0\"} min time: 1573829811392 max time: 1573829819389)"
```

What's your take?

PR #1838, opened 2019-11-18
Added tests covering TSDB read path in the ingester

pracucci:
**What this PR does**:
While working on PR #1830 I've realized we don't have any test on the TSDB read path in the ingester. In this PR I'm adding them.

_No issue for this_

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2_test.go
+++ b/pkg/ingester/ingester_v2_test.go
@@ -7,11 +7,15 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/ring"
+	"github.com/cortexproject/cortex/pkg/util/test"
 	"github.com/cortexproject/cortex/pkg/util/validation"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


pracucci:
- @thorfour and @codesome may you take a look, please?
- @tomwilkie may you add this PR to https://github.com/cortexproject/cortex/projects/4 , please?

pkg/ingester/ingester_v2_test.go
```diff
@@ -0,0 +1,288 @@
...
+		lbls      labels.Labels
+		value     float64
+		timestamp int64
+	}{
+		{labels.Labels{{Name: labels.MetricName, Value: "test_1"}, {Name: "status", Value: "200"}, {Name: "route", Value: "get_user"}}, 1, 100000},
+		{labels.Labels{{Name: labels.MetricName, Value: "test_1"}, {Name: "status", Value: "500"}, {Name: "route", Value: "get_user"}}, 1, 110000},
+		{labels.Labels{{Name: labels.MetricName, Value: "test_2"}}, 2, 200000},
+	}
+
+	expected := []string{"__name__", "status", "route"}
```
codesome: The result of `LabelNames()` is always sorted in Prometheus. If we were to follow that in Cortex too, then this should be fixed.

pkg/ingester/ingester_v2_test.go
```diff
@@ -0,0 +1,288 @@
...
+	// Push series
+	ctx := user.InjectOrgID(context.Background(), "test")
+
+	for _, series := range series {
+		req, _ := mockWriteRequest(series.lbls, series.value, series.timestamp)
+		_, err := i.v2Push(ctx, req)
+		require.NoError(t, err)
+	}
+
+	// Get label values
```
codesome: Wrong comment, I guess that was a copy paste from above :)

pkg/ingester/ingester_v2_test.go
```diff
@@ -0,0 +1,288 @@
...
+	expectedResponse := &client.QueryResponse{
+		Timeseries: []client.TimeSeries{
+			{
+				Labels:  client.FromLabelsToLabelAdapters(lbls),
+				Samples: samples,
+			},
+		},
+	}
+
+	return req, expectedResponse
```
codesome: I don't see `expectedResponse` being used anywhere

pkg/ingester/ingester_v2_test.go
pracucci: Ops! Yes, I've copied previous test as template. Fixed.

pkg/ingester/ingester_v2_test.go
pracucci: This function is a copy and paste from https://github.com/cortexproject/cortex/pull/1830 (still under review). Keeping the copy and paste will help me rebasing once one of the two PRs will get merged.

pkg/ingester/ingester_v2_test.go
pracucci: Yes, but this is the ingester, not the querier, so ordering is not required at this stage. The querier fetch label names from all ingesters, and then sort it. Technically it's implemented in the distributor (which runs as internal dependency of the querier) and the [specific code is here](https://github.com/cortexproject/cortex/blob/master/pkg/distributor/distributor.go#L499).

_I'm going to resolve this conversation, but feel free to re-open it if something is missing._

codesome: Ah got it!

pracucci:
@gouthamve I've fixed the conflicts after rebasing. May you help me moving forward with this PR, please?

PR #1839, opened 2019-11-19
Migrate integration tests to Go testing framework

pracucci:
**What this PR does**:
This PR introduces an easy-to-use framework to write Cortex integration tests in Go. The current integration tests written in Bash have been ported (and enriched) into Go, and a couple of new tests have been added:
1. Backward compatibility test
2. Ingester transfer with blocks storage

Worth to note:
1. In CircleCI, the configs DB integration tests and the new tests have been splitted into 2 pipelines, to run them concurrently
2. I've tried to fine-tune the Minio and Consul config to reduce the startup time (I haven't found any possibile optimization on DynamoDB local)
3. `make test` doesn't run tests in the `integration/` folder
4. Being based on the Go testing framework, I can run integration tests directly in the IDE (VSCode) as far as the `cortex:latest` Docker image is updated. On my local machine a single integration test takes about 10s and re-iterating on it (while developing it) looks pretty handy

_No issue_

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


bboreham:
I support the idea of moving to Go, but Compose seems to be adding a lot of weight - is it justified?
It looks like we will need another 50-line Compose file for each test?

pracucci:
> Compose seems to be adding a lot of weight - is it justified?
> It looks like we will need another 50-line Compose file for each test?

@bboreham It's a good question. My idea (which is not visible in this PoC) is that you need a compose file for each **scenario**, not test suite. A scenario is a cluster setup (ie. storage based on Dynamo+S3 or another scenario with gossip ring instead of consul, etc), upon which you can run multiple tests. Each test suite runs against a scenario and within a single suite you can have multiple semi-isolated tests running each test with a different tenant ID (so the "framework" may give each test a unique tenant ID to use).

integration/framework/scenario.go
```diff
@@ -0,0 +1,292 @@
...
+	return s.StartService(NewService(
+		"minio",
+		// If you change the image tag, remember to update it in the preloading done
+		// by CircleCI too (see .circleci/config.yml).
+		"minio/minio:RELEASE.2019-12-30T05-45-39Z",
+		s.networkName,
+		[]int{9000},
+		map[string]string{
+			"MINIO_ACCESS_KEY": "cortex",
+			"MINIO_SECRET_KEY": "supersecret",
```
pstibrany: It would be nice to extract these hardcoded credentials and other shared values into constants, and use them from integration/configs.go and other places.

integration/framework/service.go
```diff
@@ -0,0 +1,330 @@
...
+		readiness:                    readiness,
+		retryBackoff: util.NewBackoff(context.Background(), util.BackoffConfig{
+			MinBackoff: 100 * time.Millisecond,
+			MaxBackoff: 200 * time.Millisecond,
+			MaxRetries: 100, // Sometimes the CI is slow
+		}),
+	}
+}
+
+func (s *Service) Start() error {
```
pstibrany: Another issue I've run into is that container already existed. Would it be possible to delete them beforehand? Or even better, use dynamic container names, but static hostnames on the network? (That's like V2 request)

integration/framework/service.go
```diff
@@ -0,0 +1,330 @@
...
+
+	if out, err := RunCommandAndGetOutput("docker", "stop", "--time=0", s.name); err != nil {
+		fmt.Println(string(out))
+		return err
+	}
+
+	return nil
+}
+
+func (s *Service) Endpoint(port int) string {
```
pstibrany: This `port` parameter is also unused.

integration/framework/scenario.go
```diff
@@ -0,0 +1,292 @@
...
+
+func NewScenario() (*Scenario, error) {
+	s := &Scenario{
+		id:          nextScenarioID,
+		networkName: "cortex-" + strconv.Itoa(nextScenarioID),
+		services:    []*Service{},
+	}
+
+	// Setup the docker network
+	if out, err := RunCommandAndGetOutput("docker", "network", "create", s.networkName); err != nil {
```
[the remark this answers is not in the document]
pracucci: The problem is not removing or reusing the network if already exists, but after that error you will get an error for already existing containers. You get this error if the previous test run panicked (ie. test timeout) and automatically cleaning up containers in a safe way is non trivial.

Please let me know if you ended up into this situation without a panic.

integration/framework/service.go
```diff
@@ -0,0 +1,330 @@
...
+
+		s.retryBackoff.Wait()
+	}
+
+	return fmt.Errorf("the service %s is not ready", s.name)
+}
+
+func (s *Service) WaitMetric(port int, metric string, value float64) error {
+	for s.retryBackoff.Reset(); s.retryBackoff.Ongoing(); {
+		metrics, err := s.Metrics(80)
```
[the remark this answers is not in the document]
pracucci: Well spotted. Yes, my mistake. It worked because so far it was always `80`.

integration/framework/service.go
```diff
@@ -0,0 +1,330 @@
...
+			return err
+		}
+
+		s.retryBackoff.Wait()
+	}
+
+	return fmt.Errorf("the service %s is not ready", s.name)
+}
+
+func (s *Service) WaitMetric(port int, metric string, value float64) error {
```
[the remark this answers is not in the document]
pracucci: We should actually load metrics from `port` and not the hardcoded `80`. Fixed.

integration/framework/scenario.go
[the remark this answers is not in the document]
pstibrany: It was pretty common to run into problem with container left behind, when I didn't have CORTEX_IMAGE and  CORTEX_INTEGRATION_DIR set correctly.

With unset CORTEX_INTEGRATION_DIR, docker starts minio container, but minio then fails. However, since this is reported as error in the `StartService` code, service is not appended to list of started services, and is then not cleaned up.

integration/framework/service.go
pracucci: Damn. Thanks! You spotted another issue. It should have been `s.networkPortsContainerToLocal[port]` (this is the result of a bunch of refactories I did while building it).

integration/framework/scenario.go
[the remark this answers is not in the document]
pracucci: After talking to Peter, we actually found an issue. If `service.Start()` did return `error` (for any reason) the container may have been created but it was not cleaned up. I've pushed a fix for this.

integration/framework/service.go
pracucci: I thought about it. I would be more glad to fix the issues leading to the "container already existed". Since we're using a shared Docker environment, I would prefer to not remove containers just because of a matching name (ie. it's not guaranteed that the "dynamodb" container is actually the integration test container if we find it already running on the system).

> Or even better, use dynamic container names, but static hostnames on the network?

I thought about it, but makes things even more complicated. I'm wondering if we could keep it for the future if we see this being an issue after the fix done [in this commit](https://github.com/cortexproject/cortex/pull/1839/commits/7d7be2d5f911802cc2f06697b74abbf132245d03).

What's your take?

pstibrany: I'm fine pushing dynamic container names for (possibly much) later. With latest fix, I cannot reproduce problem with unremoved containers.

integration/framework/scenario.go
```diff
@@ -0,0 +1,297 @@
...
+
+// getDefaultCortexImage returns the Docker image to use to run Cortex.
+func getDefaultCortexImage() string {
+	// Get the cortex image from the CORTEX_IMAGE env variable,
+	// falling back to "cortex:latest"
+	if os.Getenv("CORTEX_IMAGE") != "" {
+		return os.Getenv("CORTEX_IMAGE")
+	}
+
+	return "cortex:latest"
```
bboreham: `cortex:latest` doesn't exist on my machine; 
```suggestion
	return "quay.io/cortexproject/cortex:latest"
```

integration/framework/service.go
```diff
@@ -0,0 +1,345 @@
...
+	defer func() {
+		if err != nil {
+			_, _ = RunCommandAndGetOutput("docker", "rm", "--force", s.name)
+		}
+	}()
+
+	cmd := exec.Command("docker", s.buildDockerRunArgs()...)
+	cmd.Stdout = &LinePrefixWriter{prefix: s.name + ": ", wrapped: os.Stdout}
+	cmd.Stderr = &LinePrefixWriter{prefix: s.name + ": ", wrapped: os.Stderr}
+	if err = cmd.Start(); err != nil {
```
bboreham: This does not seem to detect some failures:
```
Starting table-manager
table-manager: Unable to find image 'cortex:latest' locally
table-manager: docker: Error response from daemon: pull access denied for cortex, repository does not exist or may require 'docker login'.
table-manager: See 'docker run --help'.
panic: test timed out after 5m0s
```

integration/framework/scenario.go
```diff
@@ -0,0 +1,297 @@
...
+		s.networkName,
+		[]int{},
+		nil,
+		// Run consul in "dev" mode so that the initial leader election is immediate
+		NewCommand("agent", "-server", "-client=0.0.0.0", "-dev", "-log-level=err"),
+		nil,
+	))
+}
+
+func (s *Scenario) StartMinio() error {
```
bboreham: Could comment this like "Used as a local replacement for S3", since I hadn't heard of Minio.

integration/framework/scenario.go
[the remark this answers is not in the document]
bboreham: In the previous version I force-removed all containers and the network before running each scenario.
(That is what `stop-components.sh` did)

integration/framework/scenario.go
pstibrany: I use it for local development. Works fine here. But I learned about it from Marco some time ago, so ...

integration/framework/service.go
pracucci: Thanks for spotting it! There was a bug in `WaitStarted()`. I was calling `backoff.Reset()` instead of `backoff.Wait()` 🤦‍♂

pracucci:
> more importantly I haven't had a clean run yet

@bboreham According to the output, I can see:

```
        	Error:      	Received unexpected error:
        	            	the service minio is not ready
        	Test:       	TestIngesterHandOverWithBlocksStorage
```

And looking at minio logs (starting with `minio:`) I can see:

```
minio: Error: disk path full
```

Not sure about the root cause tho.

integration/framework/scenario.go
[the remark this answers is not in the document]
pracucci: Ok. I've added the logic to cleanup the Docker containers + network before each test, so that we can now automatically recover from a spurious situation left from the previous run. The change is in [this commit](https://github.com/cortexproject/cortex/pull/1839/commits/f81d2825a8c723902c1bdadef639cfa03afd2b54).

PR #1845, opened 2019-11-20
Feat/blocks grpc

### LINKED ISSUE #1824: Ingester query streaming needs implementing for TSDB

thorfour:
**What this PR does**: Implements gRPC streaming from TSDB based ingesters. 

Fixes #1824 

**Checklist**
- [X ] Tests updated
- [ ] Documentation added


pkg/querier/querier.go
```diff
@@ -46,6 +47,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.Iterators, "querier.iterators", false, "Use iterators to execute query, as opposed to fully materialising the series in memory.")
 	f.BoolVar(&cfg.BatchIterators, "querier.batch-iterators", false, "Use batch iterators to execute query, as opposed to fully materialising the series in memory.  Takes precedent over the -querier.iterators flag.")
 	f.BoolVar(&cfg.IngesterStreaming, "querier.ingester-streaming", false, "Use streaming RPCs to query ingester.")
+	f.BoolVar(&cfg.TSDBIngesterStreaming, "querier.tsdb-ingester-streaming", false, "Use TSDB streaming RPCs to query ingester.")
```
pracucci: I'm missing the rationale on adding a new flag instead of reusing the `cfg.IngesterStreaming`. Is there any specific reason?

pkg/ingester/client/cortex.proto
```diff
@@ -13,6 +13,7 @@ service Ingester {
   rpc Push(WriteRequest) returns (WriteResponse) {};
   rpc Query(QueryRequest) returns (QueryResponse) {};
   rpc QueryStream(QueryRequest) returns (stream QueryStreamResponse) {};
+  rpc QueryTSDBStream(QueryRequest) returns (stream QueryTSDBStreamResponse) {};
```
pracucci: I'm a bit dubious about this duplication, instead of using `QueryStream`. Is it hard to fit the TSDB response into `QueryStreamResponse`? If I'm not missing anything, it would allow use to reuse the code on the distributor side, instead of the custom `ingesterTSDBQueryable`.

thorfour: So my reasoning was it didn't make sense to re-purpose it, even though that would have been ideal. Since a `QueryStreamResponse` returns a slice of `TimeSeriesChunk` and the TSDB select returns a `SeriesSet` it didn't make sense to me to then re-encode that data into chunks just to reuse a function.

pkg/querier/querier.go
thorfour: Since there's a separate path for TSDB streaming I needed a way to  select that path, and we didn't pass in the storage config to the querier. But to reuse it, that's what we'd have to do. I don't have a strong preference.

pkg/ingester/client/cortex.proto
pracucci: I see your point. However, I think it's not just a function reusing, but the large amount of the new code introduced in this PR. Am I wrong?

thorfour: I don't think fear of new code is a good reason to make a bad design decision. If it's "better" to re-encode into chunks before streaming then it would be worth it

pracucci: I don't fear new code, but I believe when we - as software engineers - evaluate options, we should balance different things, including whether it's worth or not to fit in the current design.

Given we're in the discussion phase (when there's no design doc, the code review is also the place where tradeoffs in design are evaluated), I would be glad if we could openly evaluate pros and cons.

That being said, I do understand your perspective, but I've the feeling that re-using Cortex chunks is - from a design perspective - what we already did in `BlockQuerier.Get()` when we call the `seriesToChunks()`?

thorfour: I'm not arguing we shouldn't have this conversation here :) we definitely should. 

I definitely wanted to re-use as much code as possible, but since nothing else in the TSDB path operates in chunks it felt a diversion from that, as well as a waste of CPU cycles to encode it back to chunks.

pracucci: I'm not against your perspective. Let me just understand, before I do another round of review. Would be different than what we already did in BlockQuerier.Get() when we call the seriesToChunks() or would be the same? _Just to understand, maybe I'm missing something._

thorfour: It would be the same in concept, but different in implementation. `seriesToChunks` is converting one set of chunks to a different type of chunks. We'd have to write a timeseriesToChunks function to convert the timeseries slice. 

The block querier already deals with chunks, so it felt like less of a stretch from a high-level perspective.

pkg/querier/querier.go
```diff
@@ -24,6 +24,7 @@ type Config struct {
 	Iterators                bool
 	BatchIterators           bool
 	IngesterStreaming        bool
+	TSDBIngesterStreaming    bool // TODO THOR
```
pstibrany: What is this TODO about?

pkg/querier/querier.go
pstibrany: I would have preferred if streaming API call handled both cases, perhaps extending `QueryStreamResponse` with `TimeSeries` field, and then selecting appropriate path (chunks-based, or timeseries-based) in the querier. To me TSDB or Blocks setting should only affect ingesters. (See also discussion at proto file)

pkg/distributor/query.go
```diff
@@ -173,3 +191,56 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	hashToSeries := map[model.Fingerprint]ingester_client.TimeSeries{}
+	for _, result := range results {
+		for _, response := range result.([]*ingester_client.QueryTSDBStreamResponse) {
+			for _, series := range response.Timeseries {
+				hash := client.FastFingerprint(series.Labels)
```
pstibrany: This usage of fast fingerprint introduces a bug (it's also present in `queryIngesterStream`), where different timeseries with colliding fast fingerprint are merged together. We should fix both places.

pkg/ingester/client/cortex.proto
pstibrany: Ingester with chunks simply sends entire encoded chunks to the querier, which is interesting design choice, most likely done to avoid doing extra work inside ingester.

We obviously cannot do the same with TSDB, since its API is different. We need to extract data from TSDB, but encoding to chunks is extra work – and we try to put extra work in queriers instead. I think a good compromise would be to unify the API and return both results in `QueryStreamResponse` (in different fields), and let querier handle those differences.

This way, querier doesn't need to know whether ingester uses chunks of TSDB beforehand (to choose API call), but will simply work with whatever data ingester returns (chunks or timeseries).

pkg/querier/querier.go
thorfour: removed.

pkg/querier/querier.go
thorfour: This config option is removed.

pkg/ingester/client/cortex.proto
thorfour: @pstibrany that's a good idea. I've implemented a first pass of that idea. API now returns a `QueryStreamResponse` which contains fields for a slice of Timeseries or Chunkseries. 

I considered leveraging grpc using the oneof field
```
message QueryStreamResponse {
  repeated QueryStreamData series = 1;
}

message QueryStreamData {
  oneof type {
    TimeSeriesChunk chunkseries = 1 [(gogoproto.nullable) = false];
    TimeSeries timeseries = 2 [(gogoproto.nullable) = false];
  }
}
```
which has the benefit it guarantees only one of those structs can exist in the response, but I felt like it made the code more obtuse. Thoughts?

pkg/distributor/query.go
thorfour: What's the correct function to user here?

pstibrany: Easiest solution is to skip any fingerprinting, and just convert labels to a string, and use that string as a key to hash map.

pkg/ingester/client/cortex.proto
pstibrany: I cannot see generated code when using `oneOf`, so cannot comment on complexity, but I wouldn't worry about `oneOf` too much. Putting a comment into proto file explaining the contract (only one of the fields is set) would be enough imho, as this is internal API.

pstibrany: I've found this: https://developers.google.com/protocol-buffers/docs/reference/go-generated#oneof, it doesn't look *that* terrible :)

pkg/distributor/query.go
pstibrany: Note: if you are going to make this change, it should be done in `queryIngesterStream` as well. Perhaps a separate PR is a better idea.

pkg/ingester/client/cortex.proto
thorfour: alright, talked me into it :)

thorfour: Nevermind, it quickly became a mess getting it to work with our existing code. Leaving as is for now.

pkg/distributor/query.go
thorfour: I think this should be brought up in a separate PR. Going to defer for now.

pkg/ingester/client/cortex.proto
```diff
@@ -166,3 +167,4 @@ message TimeSeriesFile {
 }
 
 message TransferTSDBResponse {}
+
```
pstibrany: nit: This extra newline can go

thorfour: Fixed.

pkg/ingester/client/cortex.proto
```diff
@@ -56,9 +56,10 @@ message QueryResponse {
   repeated TimeSeries timeseries = 1 [(gogoproto.nullable) = false];
 }
 
-// QueryStreamResponse contains a batch of timeseries chunks.
+// QueryStreamResponse contains a batch of timeseries chunks or timeseries. Only one of these series will be populated.
 message QueryStreamResponse {
-  repeated TimeSeriesChunk timeseries = 1 [(gogoproto.nullable) = false];
+  repeated TimeSeriesChunk chunkseries = 1 [(gogoproto.nullable) = false];
```
pracucci: I'm personally not a big fan of this name. Have you considered `timeseries_chunk` which then gets generated into `TimeseriesChunk`?

pkg/ingester/ingester_v2.go
```diff
@@ -390,6 +390,85 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
 	return result, nil
 }
 
+// QueryTSDBStream streams metrics from a TSDB. This implements the client.IngesterServer interface
+func (i *Ingester) QueryTSDBStream(req *client.QueryRequest, stream client.Ingester_QueryStreamServer) error {
```
pracucci: So far we've prefixed the TSDB functions as `v2<name>`, so I would get stick to it for consistency (`v2QueryStream()`). I'm not a big fan of `v2` at all, but we can do a mass renaming in a dedicated PR.

pkg/ingester/client/cortex.proto
thorfour: Hadn't, but I'm happy to change it to that if it's preferred.

pkg/ingester/ingester_v2.go
thorfour: Yea that sounds like good consistency

pkg/ingester/client/cortex.proto
pracucci: Don't have strong feelings, but to me it looks a bit more clear (being equal to the data type).

pkg/querier/ingester_streaming_queryable.go
```diff
@@ -47,8 +47,8 @@ func (i ingesterQueryable) Get(ctx context.Context, userID string, from, through
 		return nil, promql.ErrStorage{Err: err}
 	}
 
-	chunks := make([]chunk.Chunk, 0, len(results))
-	for _, result := range results {
+	chunks := make([]chunk.Chunk, 0, len(results.Chunkseries))
```
pracucci: I manually tested this PR and querying the ingesters does not work. I think it's because `results.Timeseries` is not handled here in the `ingesterQueryable.Get()`.

_We also need better unit tests coverage here, to ensure this issue is covered too._

pkg/querier/tsdb_ingester_streaming_queryable.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"unsafe"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// TimeSeriesSeriesSet is a wrapper around a client.TimeSeries slice to implement to SeriesSet interface
+type TimeSeriesSeriesSet struct {
```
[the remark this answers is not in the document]
thorfour: Because it is a wrapper around a `TimeSeries` to implement the `SeriesSet` interface. So I thought that name although a little ugly better indicated what it was.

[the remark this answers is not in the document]
pracucci: Ok. Thanks for the explaining the rationale.

pkg/ingester/ingester_v2.go
```diff
@@ -390,6 +390,85 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
 	return result, nil
 }
 
+// v2QueryStream streams metrics from a TSDB. This implements the client.IngesterServer interface
+func (i *Ingester) v2QueryStream(req *client.QueryRequest, stream client.Ingester_QueryStreamServer) error {
+
```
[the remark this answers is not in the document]
thorfour: Done.

pkg/ingester/ingester_v2.go
```diff
@@ -390,6 +390,85 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
...
+		return err
+	}
+
+	result := &client.QueryResponse{}
+	count := 0
+	for ss.Next() {
+		series := ss.At()
+
+		// convert labels to LabelAdapter
+		var adapters []client.LabelAdapter
```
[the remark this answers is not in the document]
thorfour: Done. (these changes predate that which is why it was missed)

pkg/ingester/ingester_v2.go
```diff
@@ -390,6 +390,85 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
...
+	}
+	defer q.Close()
+
+	ss, err := q.Select(matchers...)
+	if err != nil {
+		return err
+	}
+
+	result := &client.QueryResponse{}
+	count := 0
```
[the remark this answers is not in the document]
thorfour: Done.

pkg/ingester/ingester_v2.go
```diff
@@ -390,6 +390,85 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
...
+		result.Timeseries = append(result.Timeseries, ts)
+		count++
+		if count >= queryStreamBatchSize {
+			err = stream.Send(&client.QueryStreamResponse{
+				Timeseries: result.Timeseries,
+			})
+			if err != nil {
+				return err
+			}
+			count = 0
```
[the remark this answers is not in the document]
thorfour: Fixed.

pkg/ingester/ingester_v2.go
```diff
@@ -390,6 +390,85 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
...
+		return err
+	}
+	defer q.Close()
+
+	ss, err := q.Select(matchers...)
+	if err != nil {
+		return err
+	}
+
+	result := &client.QueryResponse{}
```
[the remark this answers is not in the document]
thorfour: Fixed.

pkg/ingester/ingester_v2.go
```diff
@@ -390,6 +390,85 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
...
+	// Final flush any existing metrics
+	if count != 0 {
+		err = stream.Send(&client.QueryStreamResponse{
+			Timeseries: result.Timeseries,
+		})
+		if err != nil {
+			return err
+		}
+	}
+
```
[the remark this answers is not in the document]
thorfour: Added.

pkg/distributor/query.go
```diff
@@ -138,38 +138,56 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
...
+		// Parse any time series
+		for _, series := range response.Timeseries {
+			hash := client.FastFingerprint(series.Labels)
+			existing := hashToTimeSeries[hash]
+			existing.Labels = series.Labels
+			existing.Samples = append(existing.Samples, series.Samples...)
+			hashToTimeSeries[hash] = existing
 		}
 	}
-	result := make([]client.TimeSeriesChunk, 0, len(hashToSeries))
```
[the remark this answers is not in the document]
thorfour: Added.

pkg/querier/tsdb_ingester_streaming_queryable.go
```diff
@@ -0,0 +1,66 @@
+package querier
```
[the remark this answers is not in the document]
thorfour: Done.

pkg/querier/ingester_streaming_queryable.go
thorfour: Yea you're right. An unfortunate result from merged the results and trying to reuse the `ingesterStreamingQueryable` as common code. It actually looks like the `Select` function is never utilized since `newUnifiedChunkQueryable.Select` calls `Get` instead of `Select`. 

So to resolve that I'll have to convert the `client.TimeSeries` into chunks, which is what I was trying to avoid.

pracucci: @pstibrany Do you have thoughts about this? Looks unfortunate if we have to convert timeseries into chunks. I haven't deeply looked into alternatives, so I'm asking if you have any idea.

pkg/querier/tsdb_ingester_streaming_queryable.go
```diff
@@ -0,0 +1,66 @@
...
+type TimeSeriesSeriesIterator struct {
+	ts *client.TimeSeries
+	i  int
+}
+
+// Next implements SeriesSet interface
+func (t *TimeSeriesSeriesSet) Next() bool { t.i++; return t.i <= len(t.ts) }
+
+// At implements SeriesSet interface
+func (t *TimeSeriesSeriesSet) At() storage.Series { return &TimeSeriesSeriesSet{ts: t.ts, i: t.i} }
```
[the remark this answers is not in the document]
thorfour: fixed.

pkg/querier/tsdb_ingester_streaming_queryable.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// TimeSeriesSeriesSet is a wrapper around a client.TimeSeries slice to implement to SeriesSet interface
+type TimeSeriesSeriesSet struct {
+	ts []client.TimeSeries
+	i  int
```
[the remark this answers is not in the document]
thorfour: Changed.

pkg/querier/tsdb_ingester_streaming_queryable.go
```diff
@@ -0,0 +1,66 @@
...
+// Iterator implements the Series interface
+func (t *TimeSeriesSeriesSet) Iterator() storage.SeriesIterator {
+	return &TimeSeriesSeriesIterator{
+		ts: &t.ts[t.i-1],
+	}
+}
+
+// Seek implements SeriesIterator interface
+func (t *TimeSeriesSeriesIterator) Seek(s int64) bool {
+	for i, sample := range t.ts.Samples {
```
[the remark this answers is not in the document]
thorfour: Done.

pkg/querier/tsdb_ingester_streaming_queryable.go
```diff
@@ -0,0 +1,66 @@
...
+
+// At implements SeriesSet interface
+func (t *TimeSeriesSeriesSet) At() storage.Series { return &TimeSeriesSeriesSet{ts: t.ts, i: t.i} }
+
+// Err implements SeriesSet interface
+func (t *TimeSeriesSeriesSet) Err() error { return nil }
+
+// Labels implements the Series interface
+func (t *TimeSeriesSeriesSet) Labels() labels.Labels {
+	lbls := client.FromLabelAdaptersToLabels(t.ts[t.i-1].Labels)
```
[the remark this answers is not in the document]
thorfour: Old code. Used to be two different packages. Fixed.

pkg/querier/ingester_streaming_queryable.go
```diff
@@ -87,8 +87,14 @@ func (q *ingesterStreamingQuerier) Select(sp *storage.SelectParams, matchers ...
 		return nil, nil, promql.ErrStorage{Err: err}
 	}
 
-	serieses := make([]storage.Series, 0, len(results))
-	for _, result := range results {
+	// results returned in time series format
+	if len(results.Timeseries) != 0 {
+		return &TimeSeriesSeriesSet{ts: results.Timeseries}, nil, nil
```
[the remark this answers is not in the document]
thorfour: added sorted TimeSeries by labels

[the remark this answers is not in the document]
pracucci: Could you add a comment to explain why is it required? I'm not 100% sure it's required, so would be great signaling it to future maintainers / contributors.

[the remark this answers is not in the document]
thorfour: I honestly have no idea. But for know I figured I'd keep it 1:1 with the old methodology.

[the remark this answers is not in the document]
pracucci: @bboreham Do you remember why we do sort series by labels in `newConcreteSeriesSet()`?

pkg/querier/tsdb_ingester_streaming_queryable.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Iterator implements the Series interface
+func (t *TimeSeriesSeriesSet) Iterator() storage.SeriesIterator {
+	return &TimeSeriesSeriesIterator{
+		ts: &t.ts[t.i-1],
+	}
+}
+
+// Seek implements SeriesIterator interface
+func (t *TimeSeriesSeriesIterator) Seek(s int64) bool {
```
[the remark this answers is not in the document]
thorfour: Maybe I'm missing something but that appears to be the current behaviour of `queryIngesterStream` in master, and I don't see it sorting the samples anywhere either?

pkg/querier/ingester_streaming_queryable.go
pstibrany: This is only used by chunks-based querier to unify access to different chunk sources. TSDB based querier doesn't use this. We only convert chunk-based series into `[]chunk.Chunk` here.

pstibrany: Ignore my comment pls, it's clearly wrong.

pstibrany: I think with current design of Querier, it's difficult to avoid copying and adapting to expected return value (We already do the same conversion in BlockQuerier). Long-term, we should refactor querier to be based on Queryables more than chunk stores.

pkg/querier/ingester_streaming_queryable.go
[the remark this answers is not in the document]
pstibrany: While sorting was added to `newTimeSeriesSeriesSet`, it's only invoked from tests.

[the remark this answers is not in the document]
pstibrany: Re sorting requirement: this is required by `storage.NewMergeSeriesSet` call (eg. at pkg/querier/querier.go:167), although it's not documented properly (https://github.com/prometheus/prometheus/issues/6717).

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"sort"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// TimeSeriesSeriesSet is a wrapper around a client.TimeSeries slice to implement to SeriesSet interface
+type TimeSeriesSeriesSet struct {
```
pstibrany: No need to export this.

pkg/querier/ingester_streaming_queryable.go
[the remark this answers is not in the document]
pracucci: I would suggest to add a comment to explain why we need sorting. It may be something like this (feel free to improve it):

```
We need to sort series because this SeriesSet may be used as input to merge it with other SeriesSet (ie. using storage.NewMergeSeriesSet) and such merging requires each SeriesSet to be already sorted.
```

pkg/querier/tsdb_ingester_streaming_queryable.go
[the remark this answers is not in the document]
pracucci: Currently `newConcreteSeriesSet()` is called passing in input a slice of `chunkSeries`. We need to check if a `chunkSeries.Iterator()` returns an interator which guarantee the `Seek()` works fine:

```
func (s *chunkSeries) Iterator() storage.SeriesIterator {
	return s.chunkIteratorFunc(s.chunks, model.Time(s.mint), model.Time(s.maxt))
}
```

There are two `chunkIteratorFunc`:
- `batch.NewChunkMergeIterator`
- `iterators.NewChunkMergeIterator`

Looking at both implementations, they look good to me. Am I missing anything?

pkg/querier/timeseries_series_set.go
thorfour: unexported.

thorfour:
Thanks to @pstibrany's work, there's no longer mangling to return streamed series from either implementation. The code is now just a simple switch statement.
```
    switch resp := results.Series.(type) {
    case *client.QueryStreamResponse_TimeSeries:
        if resp.TimeSeries == nil {
            return nil, nil, nil
        }

        return newTimeSeriesSeriesSet(resp.TimeSeries.Timeseries), nil, nil

    case *client.QueryStreamResponse_ChunkSeries:
        if resp.ChunkSeries == nil {
            return nil, nil, nil
        }

        serieses := make([]storage.Series, 0, len(resp.ChunkSeries.Chunkseries))
        for _, result := range resp.ChunkSeries.Chunkseries {
            // Sometimes the ingester can send series that have no data.
            if len(result.Chunks) == 0 {
                continue
            }

            ls := client.FromLabelAdaptersToLabels(result.Labels)
            sort.Sort(ls)

            chunks, err := chunkcompat.FromChunks(userID, ls, result.Chunks)
            if err != nil {
            }

            series := &chunkSeries{
                labels:            ls,
                chunks:            chunks,
                chunkIteratorFunc: q.chunkIterFn,
            }
            serieses = append(serieses, series)
        }

        return newConcreteSeriesSet(serieses), nil, nil
```

pkg/querier/distributor_queryable.go
```diff
@@ -84,8 +84,12 @@ func (q *distributorQuerier) streamingSelect(sp storage.SelectParams, matchers [
 		return nil, nil, promql.ErrStorage{Err: err}
 	}
 
-	serieses := make([]storage.Series, 0, len(results))
-	for _, result := range results {
+	if len(results.Timeseries) != 0 {
```
thorfour: I had recently change the QueryStreamResponse to use a `oneof` label so that a response could only return TimeSeries or ChunkSeries. However that came back to the distributor, which combines different return types, which comes back to this, where even though we may have two return types, we only return one type. 

Allowing the distributor to return an conjoined set of ChunkSeries and TimeSeries is more complicated, but potentially allows a easier migration path from old chunk format to block format. Since a querier could be deployed that supports both ingesters. But it will require more complexity here since I believe we'll need to do some merging between the two sets. 

Is that potential migration path the one we might expect to go? Or am I over thinking this, and we'd require a hard cutover of ingesters with streaming, in which case I can reimplement the oneof behavior and returning whichever payload here would be acceptable. @pstibrany @pracucci

pracucci: > Is that potential migration path the one we might expect to go?

No. The support for `Timeseries` is temporary. We plan to drop it at some point, as soon as TSDB will allow to directly query chunks (there's an on-going work from Bartek). The reason is that we don't want to pay the CPU cost to decode chunks in the ingesters but only in the queriers, so we want to ship TSDB chunks from the ingesters to the querier like we do for the chunks storage and then decode the chunks in the querier. It's not possible to do yet, but that's the final goal.

thorfour: Sounds good. I'll reimplement with oneof then, and have it conditionally only return a single type

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/prometheus/prometheus/storage"
+)
+
+// timeSeriesSeriesSet is a wrapper around a client.TimeSeries slice to implement to SeriesSet interface
+type timeSeriesSeriesSet struct {
+	ts []client.TimeSeries
+	i  int
+}
+
+// Timeseries is a type wrapper that implements the storage.Series interface
```
pstibrany: timeSeriesSeriesSet type method should be here, not later in the file.

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,89 @@
...
+		ts: series,
+		i:  -1,
+	}
+}
+
+// Next implements SeriesSet interface
+func (t *timeSeriesSeriesSet) Next() bool { t.i++; return t.i < len(t.ts) }
+
+// At implements SeriesSet interface
+func (t *timeSeriesSeriesSet) At() storage.Series { return &Timeseries{series: t.ts[t.i]} }
```
pstibrany: Can we return nil, if `t.i < 0`? Instead of panicking. Just in case.

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,89 @@
...
+		t.i = idx
+		return true
+	}
+
+	return false
+}
+
+// At implements the SeriesIterator interface
+func (t *TimeSeriesSeriesIterator) At() (int64, float64) {
+	return t.ts.series.Samples[t.i].TimestampMs, t.ts.series.Samples[t.i].Value
```
pstibrany: Can we return zeroes, if `t.i < 0`? Instead of panicking. Just in case.

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,89 @@
...
+
+// At implements SeriesSet interface
+func (t *timeSeriesSeriesSet) At() storage.Series { return &Timeseries{series: t.ts[t.i]} }
+
+// Err implements SeriesSet interface
+func (t *timeSeriesSeriesSet) Err() error { return nil }
+
+// Labels implements the storage.Series interface
+func (t *Timeseries) Labels() labels.Labels {
+	return client.FromLabelAdaptersToLabels(t.series.Labels)
```
pstibrany: I would add a comment saying that this conversion is safe, because t.series.Labels is set in ingester by calling `client.FromLabelsToLabelAdapters(series.Labels())`, so labels are guaranteed to be sorted.

pkg/querier/timeseries_series_set.go
thorfour: moved.

pkg/querier/timeseries_series_set.go
thorfour: done.

pkg/querier/timeseries_series_set.go
thorfour: done.

pkg/querier/timeseries_series_set.go
thorfour: done.

thorfour:
@pstibrany I've addressed your comments. If this is something that is desired soon, feel free to merge it. I think using oneof would be cleaner, but I'm currently taking paternity leave so I may not get around to that for a bit.

pkg/distributor/query.go
```diff
@@ -138,38 +138,59 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
...
+			existing.Chunks = append(existing.Chunks, series.Chunks...)
+			hashToChunkseries[hash] = existing
+		}
+
+		// Parse any time series
+		for _, series := range response.Timeseries {
+			hash := client.FastFingerprint(series.Labels)
+			existing := hashToTimeSeries[hash]
+			existing.Labels = series.Labels
+			existing.Samples = append(existing.Samples, series.Samples...)
```
pracucci: Isn't samples ordering required here? Because then series these goes into `newTimeSeriesSeriesSet()`, whose iterator `Seek()` requires samples sorted, but I can't find where such sorting occurs.

/cc @pstibrany

pkg/ingester/ingester_v2.go
```diff
@@ -536,6 +537,88 @@ func createUserStats(db *userTSDB) *client.UserStatsResponse {
...
+			})
+			if err != nil {
+				return err
+			}
+
+			batchSize = 0
+			timeseries = timeseries[:0]
+		}
+	}
+
```
pracucci: At this point, we should check `ss.Err()` and return an error if `!= nil`, because the `for ss.Next()` could also stop iterating if an error is encountered.

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,99 @@
...
+	idx := sort.Search(len(t.ts.series.Samples), func(i int) bool {
+		return t.ts.series.Samples[i].TimestampMs >= s
+	})
+
+	if idx < len(t.ts.series.Samples) {
+		t.i = idx
+		return true
+	}
+
+	return false
```
pracucci: To keep consistency with other iterator implementations, if a `Seek()` doesn't find any sample, `t.i` should be updated anyway.

Which means the function could be implemented as:
```
	t.i = sort.Search(len(t.ts.series.Samples), func(i int) bool {
		return t.ts.series.Samples[i].TimestampMs >= s
	})

        return t.i < len(t.ts.series.Samples)
```

_Please also add a test to cover this case._

pkg/distributor/query.go
pstibrany: Good catch Marco!

My assumption was always that entire `SeriesIterator` (and now `chunkenc.Iterator`) expects samples to be sorted, but interestingly, it doesn't say so in the documentation. Hmm.

pstibrany: Created https://github.com/prometheus/prometheus/issues/6876 to get clarification.

pkg/querier/timeseries_series_set.go
pstibrany: Actually, this implementation of `Seek` can go backwards, which is weird. `Seek` is only supposed to go forward. (On Prometheus master, this is now explicit: `Seek advances the iterator forward to the first sample with the timestamp equal or greater than t.`)

pracucci:
@thorfour I've also pushed an integration test: it passes on this PR, but fails on `master` (as expected). Please take a look.

pkg/distributor/query.go
brian-brazil: Yes, it must be sorted.

pstibrany: Thanks Brian!

thorfour:
@pracucci nice those integration tests look awesome!

pkg/ingester/ingester_v2.go
thorfour: Added.

pkg/querier/timeseries_series_set.go
thorfour: I've updated seek to only advance, and added unit tests around Seek.

pkg/distributor/query.go
thorfour: So we need to sort these before returning? Wouldn't the chunks need to have been sorted before my changes as well?

thorfour: I've added a sort before returning the timeseries

pstibrany: > So we need to sort these before returning? Wouldn't the chunks need to have been sorted before my changes as well?

This is a great question! In case of chunks, we have three different "chunk merging" functions (see `getChunksIteratorFunction` function in querier.go). Two of them (`batch.NewChunkMergeIterator` and `iterators.NewChunkMergeIterator`) sort chunks by From time before using them, while third one (`mergeChunks`) simply gets all the samples from chunks, and then sorts samples).

So extra sorting is not needed for chunks here, and is done later when merging chunks together.

pracucci:
_I've rebased `master`_

thorfour:
Sounds good to me

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil
+	}
+	return &Timeseries{series: t.ts[t.i]}
+}
+
+// Err implements SeriesSet interface
+func (t *timeSeriesSeriesSet) Err() error { return nil }
+
+// Timeseries is a type wrapper that implements the storage.Series interface
+type Timeseries struct {
```
pstibrany: This doesn't need to be exported.

pkg/querier/timeseries_series_set.go
```diff
@@ -0,0 +1,99 @@
...
+	series client.TimeSeries
+}
+
+// TimeSeriesSeriesIterator is a wrapper around a client.TimeSeries to implement the SeriesIterator interface
+type TimeSeriesSeriesIterator struct {
```
pstibrany: This doesn't need to be exported.

pkg/querier/timeseries_series_set.go
thorfour: Fixed.

pkg/querier/timeseries_series_set.go
thorfour: Fixed.

PR #1847, opened 2019-11-21
Memberlist: Include reason for CAS failure in the error message.

pstibrany:
Enhance error message in `memberlist.Client.CAS` method to include why CAS failed.

While doing that, this change also makes logic around retries more consistent (eg. if `f` or merge returned error, there was no retry, even if `retry` was true). Furthermore fixes bug when context was done -- previously, iteration continued because `break` only broke select statement, but not `for`.

Added tests.

[This code only affects memberlist-based KV client, which is still marked as experimental]

### DIFF
--- a/pkg/ring/kv/memberlist/memberlist_client.go
+++ b/pkg/ring/kv/memberlist/memberlist_client.go
@@ -455,7 +455,7 @@ outer:
 			}
 		}
 
-		change, newver, err, retry := m.trySingleCas(key, f)
+		change, newver, retry, err := m.trySingleCas(key, f)
 		if err != nil {
 			level.Debug(util.Logger).Log("msg", "CAS attempt failed", "err", err, "retry", retry)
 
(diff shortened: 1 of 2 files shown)


pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -445,6 +446,7 @@ func (m *Client) CAS(ctx context.Context, key string, f func(in interface{}) (ou
 			case <-time.After(1 * time.Second):
 				// ok
 			case <-ctx.Done():
+				failReason = "context done"
```
pracucci: Maybe `context timed out or cancelled` may be more clear to the final user?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -433,6 +433,7 @@ func (m *Client) CAS(ctx context.Context, key string, f func(in interface{}) (ou
 	}
 
 	sleep := false
+	failReason := "too many retries"
```
pracucci: May you enrich it saying how many retries.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -480,6 +482,7 @@ func (m *Client) CAS(ctx context.Context, key string, f func(in interface{}) (ou
 		if newver == 0 {
 			// 'f' didn't do any update that we can merge. Let's give it a new chance?
 			if !retry {
+				failReason = "no retry"
```
pracucci: I'm just wondering. Is this case a real error? To my understanding this occurs when nothing has changed, and I'm not sure it's fair to consider it a failure.

pstibrany: In this case, it is a failure actually. If nothing has changed, `f` should return nil. If it has returned non-nil value, then there was a change, but merge mechanism in memberlist client failed to detect the change. I can improve the error message in this case.

pkg/ring/kv/memberlist/memberlist_client.go
pstibrany: there is hardcoded 10 retries in the code :) let's not point that fact out too much.

pkg/ring/kv/memberlist/memberlist_client.go
pstibrany: addressed in updated code.

pkg/ring/kv/memberlist/memberlist_client.go
pstibrany: your comment lead me to realization that retry logic here is not consistent with other clients, and other cases. eg. if 'f' returns error, we can still retry. new update addresses this.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -122,7 +123,9 @@ type valueDesc struct {
 
 var (
 	// if merge fails because of CAS version mismatch, this error is returned. CAS operation reacts on it
```
jtlisi: You can probably trim this commend, these error messages are self documenting for all intents and purposes

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -432,70 +435,96 @@ func (m *Client) CAS(ctx context.Context, key string, f func(in interface{}) (ou
...
-		if newver == 0 {
-			// 'f' didn't do any update that we can merge. Let's give it a new chance?
+			lastError = err
 			if !retry {
 				break
 			}
-
-			// If merge failed because of version mismatch, don't sleep.
-			sleep = (err != errVersionMismatch)
 			continue
```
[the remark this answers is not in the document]
pstibrany: Collision happens when update to value is received via gossip during CAS call. Those gossips can be received at any time and are not strictly related to timing of changes, so sleeping will not help in that situation. Fortunately, since CAS operates only locally and is generally fast, collisions are rare.

The sleep is there for a different reason -- to make CAS be able to detect the change. This is done via Merge function of the ring, which uses Unix timestamps with one second resolution to detect changes. That's why we wait for one second, if no change is detected.

PR #1848, opened 2019-11-22
Removed extra check from checkFromIngesterIsInLeavingState in transfer.go

pstibrany:
It checked for the same condition as IF above it. Also removed "TransferChunks:" prefix from error messages, and added correct prefixes at calling sites.

### DIFF
--- a/pkg/ingester/transfer.go
+++ b/pkg/ingester/transfer.go
@@ -80,7 +80,7 @@ func (i *Ingester) TransferChunks(stream client.Ingester_TransferChunksServer) e
 				// Before transfer, make sure 'from' ingester is in correct state to call ClaimTokensFor later
 				err := i.checkFromIngesterIsInLeavingState(stream.Context(), fromIngesterID)
 				if err != nil {
-					return err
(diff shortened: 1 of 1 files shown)


pkg/ingester/transfer.go
```diff
@@ -149,24 +149,18 @@ func (i *Ingester) TransferChunks(stream client.Ingester_TransferChunksServer) e
...
-	}
-
-	if r.Ingesters == nil || r.Ingesters[fromIngesterID].State != ring.LEAVING {
-		return fmt.Errorf("TransferChunks: source ingester is not in a LEAVING state, found state=%v", r.Ingesters[fromIngesterID].State)
+		return fmt.Errorf("ring not found, got %T", v)
 	}
 
 	if r.Ingesters == nil || r.Ingesters[fromIngesterID].State != ring.LEAVING {
-		err = fmt.Errorf("source ingester is not in a LEAVING state, found state=%v", r.Ingesters[fromIngesterID].State)
-		util.Logger.Log("msg", "TransferChunks error", "err", err)
```
pracucci: Note to myself: removing this log is OK, cause the error will be logged by `Ingester.TransferOut()`.

PR #1853, opened 2019-11-25
Move to golangci-lint 

bboreham:
This is faster and does more checks.

* Remove home-grown `lint` script.
* Install the `golangci-lint` binary in the build container.
* Remove individual tools `gocyclo`, `golint`, `errcheck` which are included in `golangci-lint`.

Run in `--new-from-rev` mode so it doesn't fail the build on a bunch of warnings that were pre-existing.

Still running `go vet` separately because I couldn't see that the `vet` tool inside `golangci-lint` is the same thing.

Also run `misspell` specifically on the docs because `golangci-lint` only does `*.go`.

Also remove a couple of unnecessary steps from the build image.


Makefile
```diff
@@ -120,11 +120,11 @@ protos: $(PROTO_GOS)
...
 
 lint:
-	./tools/lint -notestpackage -novet -ignorespelling queriers -ignorespelling Queriers .
-
 	# -stdmethods=false disables checks for non-standard signatures for methods with familiar names.
 	# This is needed because the Prometheus storage interface requires a non-standard Seek() method.
-	go vet -stdmethods=false ./pkg/...
+	go vet -tags netgo -stdmethods=false ./pkg/...
+	misspell -error docs
+	golangci-lint run --new-from-rev ed7c302fd968 --build-tags netgo --timeout=5m --enable golint --enable misspell --enable gofmt
```
pracucci: This is particularly slow when running on OSX (with Docker for Mac) due to the shared volumes performances. Do you see any problem adding `:delegated` to the 3 related volumes, switching

```
	@$(SUDO) time docker run $(RM) $(TTY) -i \
		-v $(shell pwd)/.cache:/go/cache \
		-v $(shell pwd)/.pkg:/go/pkg \
		-v $(shell pwd):/go/src/github.com/cortexproject/cortex \
		$(BUILD_IMAGE) $@;
```

to

```
	@$(SUDO) time docker run $(RM) $(TTY) -i \
		-v $(shell pwd)/.cache:/go/cache:delegated \
		-v $(shell pwd)/.pkg:/go/pkg:delegated \
		-v $(shell pwd):/go/src/github.com/cortexproject/cortex:delegated \
		$(BUILD_IMAGE) $@;
```

In my case, the execution time difference is:

- Current: `8m8.369s`
- With delegated: `1m38.258s`

Makefile
```diff
@@ -120,11 +120,11 @@ protos: $(PROTO_GOS)
 	protoc -I $(GOPATH)/src:./vendor:./$(@D) --gogoslick_out=plugins=grpc,Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,:./$(@D) ./$(patsubst %.pb.go,%.proto,$@)
 
 lint:
-	./tools/lint -notestpackage -novet -ignorespelling queriers -ignorespelling Queriers .
-
 	# -stdmethods=false disables checks for non-standard signatures for methods with familiar names.
 	# This is needed because the Prometheus storage interface requires a non-standard Seek() method.
-	go vet -stdmethods=false ./pkg/...
+	go vet -tags netgo -stdmethods=false ./pkg/...
```
cyriltovena: I think go vet is on by default. We use it on Loki.

>govet (vet, vetshadow): Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string [fast: true, auto-fix: false]

see https://github.com/golangci/golangci-lint#quick-start

bboreham: I did this test: change a printf format string to invalid, e.g. `%i`.
I get a warning from `go vet` but not from `golangci-lint`.
So I am un-convinced.

Makefile
bboreham: What does `:delegated` do?
Would we want it on all similar lines?

pracucci: > What does :delegated do?

The `:delegated` tells Docker for Mac that the container’s view is authoritative (permit delays before updates on the container appear in the host). Details here: https://docs.docker.com/docker-for-mac/osxfs-caching/

I don't have a Linux box right now to play with, but it should just be ignored on Linux.

> Would we want it on all similar lines?

I would say yes.

Makefile
cyriltovena: Interesting I have the same issue in loki but if I ran it for the specific package it works. Tried to clean my cache also using `golangci-lint cache clean`.

cyriltovena: This is failing `golangci-lint run -v ./pkg/querier/queryrange`

but not 

`golangci-lint run -v ./...`

cyriltovena: Found the issue on Loki side, one file was producing warning.

```
WARN [runner] Can't run linter goanalysis_metalinter: misspell: can't get file /Users/ctovena/go/src/github.com/grafana/loki/pkg/logql/pkg/logql/expr.y contents: can't read file /Users/ctovena/go/src/github.com/grafana/loki/pkg/logql/pkg/logql/expr.y: open /Users/ctovena/go/src/github.com/grafana/loki/pkg/logql/pkg/logql/expr.y: no such file or directory
```

This was caused by a comment `//line expr.y`, I cleanup the file and it now catches %i in fmt.

cyriltovena: FYI it was also causing some linter to be skipped.

cyriltovena: You have the same kind of warning.

```
WARN [runner] Can't run linter goanalysis_metalinter: fact_purity: failed prerequisites: buildssa@github.com/cortexproject/cortex/pkg/chunk [github.com/cortexproject/cortex/pkg/chunk/cache.test] 
```

Not sure why the exit code is not 1.

bboreham: I found various issues relating to this - https://github.com/golangci/golangci-lint/issues/866, https://github.com/golangci/golangci-lint/issues/827

bboreham:
I'm now running `make lint` twice in an attempt to work around https://github.com/golangci/golangci-lint/issues/866, https://github.com/golangci/golangci-lint/issues/827

Once I do that I am reassured that it does indeed run `go vet` so have taken that out of our commands.

bboreham:
Rebased and we have a bunch of lint failures from code added since I started this PR!

I wonder if we should disable the return value check on log calls (or everywhere)?

Makefile
pracucci: Have you considered adding `:delegated` to volume mounts?

pracucci:
> I wonder if we should disable the return value check on log calls (or everywhere)?

My personal take:
- Definitely disable it for log calls
- I still see a value for other use cases (errors should be checked/handled)

bboreham:
I figured out how to disable errcheck on individual methods.

Makefile
bboreham: Have done this now.

pracucci:
@bboreham What's the state of this PR? Is there anything missing?

bboreham:
Ideally it would get an approval after the latest set of changes.

PR #1854, opened 2019-11-25
Do not close and re-open TSDBs during ingester transfer

### LINKED ISSUE #1829: Do not close and re-open TSDBs during ingester transfer
When the ingester is leaving the ring and begins transferring TSDB blocks and WAL to a joining ingester, currently we do close the TSDB to flush it to disk before transferring. However, if a query hit the ingester after the TSDB has been closed and before the transfer is completed, the leaving ingester will re-open the TSDB (including WAL replay) which lead to an high resources consumption and query latency.

pracucci:
**What this PR does**:
The Cortex TSDB experimental integration currently close all TSDBs during the ingester shutdown, before blocks and WALs are transferred to a joining ingester. This has two main issues:
1. Query requests received while closing TSDBs fail with error `open index reader: block is closing`
2. Query requests received after the TSDBs have been closed issue a re-opening of TSDBs, which means re-playing the WAL (expensive operation)

Me and @codesome checked out the TSDB implementation and came to the conclusion which is safe to transfer blocks as WAL without closing the TSDB as far as:
1. No more writes occur
2. No compaction is on-going or will occur

In this PR I've introduced some changes to avoid closing and re-opening TSDBs on queries received during ingester blocks/WAL transfer, as well as some logic to guarantee no writes occur to the TSDB once the transfer starts.

I've tested the changes in our TSDB test cluster and looks working as expected (at least, experimentally).

**Suggestion to the reviewer**: enable "Hide whitespace changes" because there are a couple of large code blocks for which it has changed only the indentation.

Fixes #1829

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -94,16 +94,16 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 
 	// Ensure the ingester shutdown procedure hasn't started
 	i.userStatesMtx.RLock()
-	stopped := i.stopped
-	i.userStatesMtx.RUnlock()
 
-	if stopped {
+	if i.stopped {
+		i.userStatesMtx.RUnlock()
 		return nil, fmt.Errorf("ingester stopping")
 	}
 
 	// Keep track of in-flight requests, in order to safely start blocks transfer
 	// (at shutdown) only once all in-flight write requests have completed
 	i.TSDBState.inflightWriteReqs.Add(1)
+	i.userStatesMtx.RUnlock()
 	defer i.TSDBState.inflightWriteReqs.Done()
 
 	// Keep track of some stats which are tracked only if the samples will be
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -3,8 +3,8 @@ package ingester
 import (
 	"fmt"
 	"net/http"
-	"time"
 	"sync"
+	"time"
 
 	"github.com/cortexproject/cortex/pkg/ingester/client"
 	"github.com/cortexproject/cortex/pkg/ring"
@@ -414,4 +414,35 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error) 
 	i.TSDBState.dbs[userID] = db
 
 	return db, nil
+}
+
+func (i *Ingester) closeAllTSDB() {
+	i.userStatesMtx.Lock()
+
+	wg := &sync.WaitGroup{}
+	wg.Add(len(i.TSDBState.dbs))
+
+	// Concurrently close all users TSDB
+	for userID, db := range i.TSDBState.dbs {
+		userID := userID
+
+		go func(db *tsdb.DB) {
+			defer wg.Done()
+
+			if err := db.Close(); err != nil {
+				level.Warn(util.Logger).Log("msg", "unable to close TSDB", "err", err, "user", userID)
+				return
+			}
+
+			// Now that the TSDB has been closed, we should remove it from the
+			// set of open ones.
+			i.userStatesMtx.Lock()
+			delete(i.TSDBState.dbs, userID)
+			i.userStatesMtx.Unlock()
+		}(db)
+	}
+
+	// Wait until all Close() completed
+	i.userStatesMtx.Unlock()
+	wg.Wait()
 }
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -435,7 +435,9 @@ func (i *Ingester) closeAllTSDB() {
 			}
 
 			// Now that the TSDB has been closed, we should remove it from the
-			// set of open ones.
+			// set of open ones. This lock acquisition doesn't deadlock with the
+			// outer one, because the outer one is released as soon as all go
+			// routines are started.
 			i.userStatesMtx.Lock()
 			delete(i.TSDBState.dbs, userID)
 			i.userStatesMtx.Unlock()
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"time"
+	"sync"
 
 	"github.com/cortexproject/cortex/pkg/ingester/client"
 	"github.com/cortexproject/cortex/pkg/ring"
@@ -32,6 +33,13 @@ const (
 type TSDBState struct {
 	dbs    map[string]*tsdb.DB // tsdb sharded by userID
 	bucket objstore.Bucket
+
+	// Keeps count of in-flight requests
+	inflightWriteReqs sync.WaitGroup
+
+	// Used to run only once operations at shutdown, during the blocks/wal
+	// transferring to a joining ingester
+	transferOnce sync.Once
 }
 
 // NewV2 returns a new Ingester that uses prometheus block storage instead of chunk storage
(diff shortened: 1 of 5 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -78,20 +86,30 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
...
+	// Ensure the ingester shutdown procedure hasn't started
+	i.userStatesMtx.RLock()
+	stopped := i.stopped
+	i.userStatesMtx.RUnlock()
+
+	if stopped {
+		return nil, fmt.Errorf("ingester stopping")
+	}
+
+	// Keep track of in-flight requests, in order to safely start blocks transfer
```
thorfour: I believe there's a race condition here. If the ingester isn't stopped, gets passed the check, and gets stopped before incrementing the `inflightWriteReqs` we can start a transfer and a write can that continue on.

pracucci: Right! I think the safest way would be incrementing the `sync.WaitGroup` within the lock. What's about now?

pkg/ingester/transfer.go
```diff
@@ -446,32 +446,58 @@ func (i *Ingester) transferOut(ctx context.Context) error {
...
+	// some initialization only once.
+	i.TSDBState.transferOnce.Do(func() {
+		// In order to transfer TSDB WAL without closing the TSDB itself - which is a
+		// pre-requisite to continue serving read requests while transferring - we need
+		// to make sure no more series will be written to the TSDB. For this reason, we
+		// wait until all in-flight write requests have been completed. No new write
+		// requests will be accepted because the "stopped" flag has already been set.
+		level.Info(util.Logger).Log("msg", "waiting for in-flight write requests to complete")
+
+		if !util.WaitGroupWithTimeout(&i.TSDBState.inflightWriteReqs, 10*time.Second) {
```
pstibrany: Instead of using 10 seconds, we should also respect deadline from context. Perhaps it would be best to use `context.WithTimeout(ctx, 10*time.Second)` here and pass new context to `WaitGroupWithTimeout`, and then react on ctx.Done channel inside that function?

pkg/ingester/transfer.go
```diff
@@ -446,32 +446,58 @@ func (i *Ingester) transferOut(ctx context.Context) error {
...
+			level.Warn(util.Logger).Log("msg", "timeout expired while waiting in-flight write requests to complete, transfer will continue anyway")
```
pstibrany: I don't understand this decision. On one hand, you have found that transferring TSDB is only safe when there are no ongoing writes, on the other hand, this code continues even if there are still writes in progress. Wouldn't it be safer to just error out in such case?

(When using context, it may also be done/cancelled by now)

pracucci: > Wouldn't it be safer to just error out in such case?

From my perspective, in the unlikely event that the in-flight requests will take long time (or will never go down to zero) it more important to transfer all the data (and loose data points pushed by in-flight requests) than no transfer anything at all. The very last thing we want is having an ingester leaving without having transferred its state.

pkg/ingester/transfer.go
pracucci: I think it's a good idea. What's about now?

pkg/ingester/transfer.go
```diff
@@ -446,32 +446,61 @@ func (i *Ingester) transferOut(ctx context.Context) error {
...
+		// pre-requisite to continue serving read requests while transferring - we need
+		// to make sure no more series will be written to the TSDB. For this reason, we
+		// wait until all in-flight write requests have been completed. No new write
+		// requests will be accepted because the "stopped" flag has already been set.
+		level.Info(util.Logger).Log("msg", "waiting for in-flight write requests to complete")
+
+		waitCtx, waitCancel := context.WithTimeout(ctx, 10*time.Second)
+		defer waitCancel()
+
+		if !util.WaitGroup(waitCtx, &i.TSDBState.inflightWriteReqs) {
```
pstibrany: Now if `WaitGroup` returns false, it may indicate various things. If `ctx.Err()` (original context) returns non-nil value here, transfer should not continue.

pracucci: For the records, we discuss it though Slack and we came to the conclusion that we shouldn't use the parent context cause we don't want to interrupt while waiting for in-flight requests to complete if the parent context is cancelled, given this logic run only once.

@pstibrany What's about now?

pstibrany: LGTM.

pkg/ingester/ingester_v2.go
thorfour: hmm I think I was wrong and that doesn't fix it. Since inflightwritereqs is being waited on outside the lock I believe the race condition still exists.

pracucci: Here we check if the flag `stopped` is set. If not, we do proceed increasing the `inflightWriteReqs` within the lock wrapping the `stopped`.

At ingester shutdown, when the `v2TransferOut()` will be called, the `stopped` flag is already set to `true` (it uses a write lock), which means that when the `stopped` flag is set to `true` there's the guarantee no other routine will be within the read lock used in `v2Push()`. Then, outside the lock, down in the execution path, we do wait on `inflightWriteReqs` but there's already the guarantee it won't be incremented because the `stopped` flag is already set to `true`.

Am I missing anything?

thorfour: okay you're right, I think I just got lost in my head. Thanks for fixing it!

pracucci:
@thorfour May you check it out again and - if you don't have any further comment and you're good with it - approve it, please? Otherwise feel free to take more time to think about it or comment.

thorfour:
@pracucci will take a look first thing Monday

pracucci:
@gouthamve and @thorfour I've fixed the conflicts after rebasing. May you take a look and help this PR moving forward, please?

pkg/ingester/ingester_v2.go
```diff
@@ -82,6 +90,20 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
...
+
+	if i.stopped {
+		i.userStatesMtx.RUnlock()
+		return nil, fmt.Errorf("ingester stopping")
+	}
+
+	// Keep track of in-flight requests, in order to safely start blocks transfer
+	// (at shutdown) only once all in-flight write requests have completed
+	i.TSDBState.inflightWriteReqs.Add(1)
+	i.userStatesMtx.RUnlock()
```
gouthamve: Do we need to do this with a lock? Can we not add it outside the lock?

Can this be simplified to:
```
i.userStatesMtx.RLock()
stopped := i.stopped
i.userStatesMtx.RUnLock()

if stopped {
}
```

pkg/ingester/ingester_v2.go
```diff
@@ -334,3 +352,34 @@ func (i *Ingester) getOrCreateTSDB(userID string) (*tsdb.DB, error) {
...
+			defer wg.Done()
+
+			if err := db.Close(); err != nil {
+				level.Warn(util.Logger).Log("msg", "unable to close TSDB", "err", err, "user", userID)
+				return
+			}
+
+			// Now that the TSDB has been closed, we should remove it from the
+			// set of open ones.
+			i.userStatesMtx.Lock()
```
gouthamve: We already acquire the lock at the beginning of the parent function this is going to deadlock no?

EDIT: No, they'll just block until the loop exits, but maybe add a comment?

pkg/ingester/ingester_v2.go
pracucci: Unfortunately not because - in order to avoid any race condition - we also need to `i.TSDBState.inflightWriteReqs.Add(1)` within the lock.

pkg/ingester/ingester_v2.go
pracucci: As you stated, there's no deadlock. They block, but realistically the `for` loop is fast and the lock is acquired after `db.Close()` which may be a slower operation.

pracucci: > but maybe add a comment?

Done

PR #1856, opened 2019-11-27
Add unit tests for distributor's MetricsForLabelMatchers()

pracucci:
**What this PR does**:
Today I was investigating if the distributor's `MetricsForLabelMatchers()` looks susceptible of FNV-1a hash collisions and I believe it is. In this PR I'm not fixing it, but I'm adding some unit tests on it (I will try to fix it in a separate PR, but I would like to find a FNV-1a hash collision first).

_No issue_

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor_test.go
+++ b/pkg/distributor/distributor_test.go
@@ -24,10 +24,12 @@ import (
 
 	"github.com/cortexproject/cortex/pkg/chunk/encoding"
 	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/prom1/storage/metric"
 	"github.com/cortexproject/cortex/pkg/ring"
 	"github.com/cortexproject/cortex/pkg/ring/kv"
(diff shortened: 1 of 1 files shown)


pracucci:
Now that we know a FNV-1a hash collision (thanks to @pstibrany) I've added - in my local branch - a test case for that and I confirm the `MetricsForLabelMatchers` is **not** hash collision safe. I will fix it in a separate PR.

```
// The two following series have the same FNV-1a=4efc6755d61ec064
{labels.Labels{{Name: labels.MetricName, Value: "up"}, {Name: "instance", Value: "hello"}, {Name: "_", Value: "ypfajYg2lsv"}}, 1, 300000},
{labels.Labels{{Name: labels.MetricName, Value: "up"}, {Name: "instance", Value: "hello"}, {Name: "_", Value: "KiqbryhzUpn"}}, 1, 300000},
```

/cc @gouthamve

PR #1857, opened 2019-11-27
memberlist: Use advertised address when sending packets

pstibrany:
Change TCP transport to use advertised address when sending "packets" (term used by memberlist lib). This was needed to stop memberlist complaining about not being able to ping other members of the cluster when running on internal Grafana k8s cluster.

(Memberlist is still experimental)

### DIFF
--- a/pkg/ring/kv/memberlist/tcp_transport.go
+++ b/pkg/ring/kv/memberlist/tcp_transport.go
@@ -351,6 +351,7 @@ func (t *TCPTransport) FinalAdvertiseAddr(ip string, port int) (net.IP, int, err
 		advertisePort = t.GetAutoBindPort()
 	}
 
+	level.Debug(util.Logger).Log("msg", "FinalAdvertiseAddr", "advertiseAddr", advertiseAddr.String(), "advertisePort", advertisePort)
 	return advertiseAddr, advertisePort, nil
 }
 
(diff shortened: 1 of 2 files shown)


bboreham:
Seems kinda nuts that we need a 600-line file just to add a _transport_ to use a library.

I read your comments about `mesh` in the design doc; those 'todo's were written 2-3 years ago and it's extremely unlikely it will change except to fix bugs.

pkg/ring/kv/memberlist/tcp_transport.go
```diff
@@ -404,7 +423,11 @@ func (t *TCPTransport) writeTo(b []byte, addr string) error {
 	// We need to send our address to the other side, otherwise other side can only see IP and port from TCP header.
 	// But that doesn't match our node address (source port is assigned automatically), which confuses memberlist.
 	// We will announce first listener's address as our address. This is what memberlist's net_transport.go does as well.
-	ourAddr := t.tcpListeners[0].Addr().String()
+	// ourAddr := t.tcpListeners[0].Addr().String()
```
pracucci: Remove me

pstibrany: Yeah, the comment is now wrong too.

pstibrany:
> Seems kinda nuts that we need a 600-line file just to add a _transport_ to use a library.

It would be nice if TCP-based transport was part of memberlist. There are some comments here and there indicating that people are using it with TCP transport, but nothing is included in the lib itself.

> I read your comments about `mesh` in the design doc; those 'todo's were written 2-3 years ago and it's extremely unlikely it will change except to fix bugs.

If you want to encourage more people to use it (and maybe you don't, which is fine), it may be worthwhile to remove TODO's from the code, clean up the API little bit and add some documentation to it.

PR #1859, opened 2019-11-27
tsdb/block_store: expose block store config settings

thorfour:

**What this PR does**: Exposes the bucket store configuration that was previously hardcoded to configurable values.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -99,18 +99,16 @@ type BucketStoreConfig struct {
 	MaxChunkPoolBytes    uint64 `yaml:"max_chunk_pool_bytes"`
 	MaxSampleCount       uint64 `yaml:"max_sample_count"`
 	MaxConcurrent        int    `yaml:"max_concurrent"`
-	DebugLogging         bool   `yaml:"debug_logging"`
 	BlockSyncConcurrency int    `yaml:"block_sync_concurrency"`
 }
 
 // RegisterFlags registers the BucketStore flags
 func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 
-	f.StringVar(&cfg.Dir, "experimental.tsdb.bucket-store.sync-dir", "tsdb-sync", "Directory to place synced tsdb indicies.")
+	f.StringVar(&cfg.SyncDir, "experimental.tsdb.bucket-store.sync-dir", "tsdb-sync", "Directory to place synced tsdb indicies.")
 	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(250*units.Mebibyte), "Size of index cache in bytes per tenant.")
 	f.Uint64Var(&cfg.MaxChunkPoolBytes, "experimental.tsdb.bucket-store.max-chunk-pool-bytes", uint64(2*units.Gibibyte), "Max size of chunk pool in bytes per tenant.")
 	f.Uint64Var(&cfg.MaxSampleCount, "experimental.tsdb.bucket-store.max-sample-count", 0, "Max number of samples (0 is no limit) per query when loading series from storage.")
 	f.IntVar(&cfg.MaxConcurrent, "experimental.tsdb.bucket-store.max-concurrent", 20, "Max number of concurrent queries to the storage per tenant.")
-	f.BoolVar(&cfg.DebugLogging, "experimental.tsdb.bucket-store.debug-logging", false, "Turn on debug logging.")
 	f.IntVar(&cfg.BlockSyncConcurrency, "experimental.tsdb.bucket-store.block-sync-concurrency", 20, "Number of Go routines to use when syncing blocks from object storage per tenant.")
 }
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/alecthomas/units"
 	"github.com/cortexproject/cortex/pkg/storage/tsdb/backend/gcs"
 	"github.com/cortexproject/cortex/pkg/storage/tsdb/backend/s3"
 )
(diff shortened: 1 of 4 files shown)


pkg/storage/tsdb/config.go
```diff
@@ -23,12 +24,12 @@ var (
...
-	BlockRanges  DurationList  `yaml:"block_ranges_period"`
-	Retention    time.Duration `yaml:"retention_period"`
-	ShipInterval time.Duration `yaml:"ship_interval"`
-	Backend      string        `yaml:"backend"`
+	Dir            string            `yaml:"dir"`
+	BlockRanges    DurationList      `yaml:"block_ranges_period"`
+	Retention      time.Duration     `yaml:"retention_period"`
+	ShipInterval   time.Duration     `yaml:"ship_interval"`
+	Backend        string            `yaml:"backend"`
+	BucketStoreCfg BucketStoreConfig `yaml:"bucket_store_config"`
```
pracucci: I'm not sure the `Cfg` / `_config` suffix brings a real value here. I would simplify it to:

```
BucketStore BucketStoreConfig `yaml:"bucket_store"`
```

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
 
 	return nil
 }
+
+// BucketStoreConfig holds the config information for Bucket Stores used by the querier
+type BucketStoreConfig struct {
+	Dir                  string `yaml:"dir"`
```
pracucci: I would still call this `SyncDir` / `sync_dir` for clarity.

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
...
+	IndexCacheSizeBytes  uint64 `yaml:"index_cache_size_bytes"`
+	MaxChunkPoolBytes    uint64 `yaml:"max_chunk_pool_bytes"`
+	MaxSampleCount       uint64 `yaml:"max_sample_count"`
+	MaxConcurrent        int    `yaml:"max_concurrent"`
+	DebugLogging         bool   `yaml:"debug_logging"`
+	BlockSyncConcurrency int    `yaml:"block_sync_concurrency"`
+}
+
+// RegisterFlags registers the BucketStore flags
+func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
```
pracucci: I would suggest to change `experimental.tsdb.bucket-store-config.` prefix to `experimental.tsdb.bucket-store.` given the `-config` doesn't bring a real value (all in all, any CLI flag is "config").

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
...
+
+	f.StringVar(&cfg.Dir, "experimental.tsdb.bucket-store-config.dir", "tsdb-sync", "directory to place synced tsdb indicies")
+	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store-config.index-cache-size", uint64(250*units.Mebibyte), "size of index cache in bytes")
```
pracucci: Add `-bytes` suffix to the CLI name, like you did for `max-chunk-pool-bytes`.

pkg/storage/tsdb/config.go
pracucci: By convention, all CLI flags description should start with an upper case and end with `.`

pkg/storage/tsdb/config.go
pracucci: I would add to the description that this cache is per-tenant.

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
...
+	f.Uint64Var(&cfg.MaxChunkPoolBytes, "experimental.tsdb.bucket-store-config.max-chunk-pool-bytes", uint64(2*units.Gibibyte), "max size of chunk pool in bytes")
```
pracucci: I would add to the description that this pool is per-tenant.

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
...
+	f.Uint64Var(&cfg.MaxSampleCount, "experimental.tsdb.bucket-store-config.max-sample-count", 0, "max number of samples (0 is no limit)")
```
pracucci: I would explain this limit is applied to queries, when loading series from storage. To my understanding, the limit is per-query.

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
...
+	f.IntVar(&cfg.MaxConcurrent, "experimental.tsdb.bucket-store-config.max-concurrent", 20, "max number of concurrent go routines")
```
pracucci: I would improve the description. Isn't this the max number of concurrent queries executed against the storage? This is also per-tenant.

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
...
+	f.BoolVar(&cfg.DebugLogging, "experimental.tsdb.bucket-store-config.debug-logging", false, "turn on debug logging")
```
pracucci: What do you think if we enable it only when the Cortex log level is debug, instead of exposing a config option? Would it be too limiting?

pkg/storage/tsdb/config.go
```diff
@@ -90,3 +91,26 @@ func (cfg *Config) Validate() error {
...
+	f.IntVar(&cfg.BlockSyncConcurrency, "experimental.tsdb.bucket-store-config.block-sync-concurrency", 20, "max number of go routines to perform block sync functions")
```
pracucci: This is also per-tenant.

pkg/storage/tsdb/config.go
thorfour: No, that's probably a better pattern.

thorfour:
@pracucci Thanks for the review, I think I've addressed all your comments.

PR #1861, opened 2019-11-28
Fixed HA Tracker jitter causing unnecessary CAS operations

pracucci:
**What this PR does**:
Today me and @pstibrany spent few hours debugging an expected pattern on the number of CAS operations done by the distributors when the HA tracker is enabled.

We found out that the PR #1748 - which introduced the update timeout jitter - also introduced a time window (long as much as the jitter) during which every request does a CAS but the CAS operation itself doesn't update the replica updated timestamp, because the CAS is triggered if `now - receivedAt >= updateTimeout` but then the CAS function is a noop if `now - receivedAt < updateTimeout + jitter`.

While adding tests, I've also realized that the `checkReplicaTimestamp()` was broken (tests were actually failing) so I've fixed it.

_No issue_

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/ha_tracker_test.go
+++ b/pkg/distributor/ha_tracker_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/ring/kv"
 	"github.com/cortexproject/cortex/pkg/ring/kv/codec"
 	"github.com/cortexproject/cortex/pkg/ring/kv/consul"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/distributor/ha_tracker_test.go
```diff
@@ -35,19 +40,22 @@ outer:
...
 			c.electedLock.RUnlock()
 
+			// If the replica or the timestamp don't match, we save the error
+			// to return if the expected match is not found within the timeout period
 			if r.GetReplica() != replica {
 				err = fmt.Errorf("replicas did not match: %s != %s", r.GetReplica(), replica)
 				continue outer
-			}
-			if timestamp.Time(r.GetReceivedAt()).Equal(expected) {
+			} else if !timestamp.Time(r.GetReceivedAt()).Equal(expected) {
```
pstibrany: Tiny detail, but using 'else if' when previous 'if' body does continue, break or return is not very nice. Previous version was clearer.

pracucci: Right! Fixed

pkg/distributor/ha_tracker_test.go
```diff
@@ -35,79 +41,82 @@ outer:
...
 			c.electedLock.RUnlock()
 
+			// If the replica or the timestamp don't match, we save the error
+			// to return if the expected match is not found within the timeout period
 			if r.GetReplica() != replica {
 				err = fmt.Errorf("replicas did not match: %s != %s", r.GetReplica(), replica)
 				continue outer
 			}
-			if timestamp.Time(r.GetReceivedAt()).Equal(expected) {
+			if !timestamp.Time(r.GetReceivedAt()).Equal(expected) {
```
pstibrany: Nice catch! :)

pkg/distributor/ha_tracker_test.go
```diff
@@ -120,19 +129,22 @@ func TestWatchPrefixAssignment(t *testing.T) {
...
+		EnableHATracker:        true,
+		KVStore:                kv.Config{Mock: mock},
+		UpdateTimeout:          time.Millisecond,
+		UpdateTimeoutJitterMax: 0,
+		FailoverTimeout:        time.Millisecond * 2,
 	})
 	assert.NoError(t, err)
 
 	// Write the first time.
+	mtime.NowForce(start)
```
cstyan: was there something that necessitated this change?

pkg/distributor/ha_tracker_test.go
```diff
@@ -246,54 +261,61 @@ func TestCheckReplicaMultiClusterTimeout(t *testing.T) {
...
+		EnableHATracker:        true,
+		KVStore:                kv.Config{Mock: mock},
+		UpdateTimeout:          time.Second,
+		UpdateTimeoutJitterMax: 0,
+		FailoverTimeout:        time.Second,
 	})
 	assert.NoError(t, err)
 
 	// Write the first time.
+	mtime.NowForce(startTime)
```
cstyan: same here

pkg/distributor/ha_tracker_test.go
pracucci: It avoids flaky tests.

pkg/distributor/ha_tracker_test.go
pracucci: It avoids flaky tests.

pracucci:
@gouthamve Do we want this in the `0.4.0`?

PR #1862, opened 2019-11-29
memberlist: fix metric names and expose internal memberlist metrics

pstibrany:
This PR fixes some metrics names used by `memberlist.Client` (eg. `_size` -> `_bytes`), and configures "github.com/armon/go-metrics" library used by memberlist to write its metrics to Prometheus.

### DIFF
--- a/pkg/ring/kv/memberlist/metrics.go
+++ b/pkg/ring/kv/memberlist/metrics.go
@@ -17,7 +17,7 @@ func (m *Client) createAndRegisterMetrics() {
 	m.totalSizeOfReceivedMessages = prometheus.NewCounter(prometheus.CounterOpts{
 		Namespace: m.cfg.MetricsNamespace,
 		Subsystem: subsystem,
-		Name:      "received_broadcasts_size",
+		Name:      "received_broadcasts_bytes",
 		Help:      "Total size of received broadcast user messages",
 	})
 
@@ -68,7 +68,7 @@ func (m *Client) createAndRegisterMetrics() {
 	m.totalSizeOfBroadcastMessagesInQueue = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: m.cfg.MetricsNamespace,
 		Subsystem: subsystem,
-		Name:      "messages_in_broadcast_size_total",
+		Name:      "messages_in_broadcast_queue_bytes",
 		Help:      "Total size of messages waiting in the broadcast queue",
 	})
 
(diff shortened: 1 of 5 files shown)


pkg/ring/kv/memberlist/metrics.go
```diff
@@ -17,7 +22,7 @@ func (m *Client) createAndRegisterMetrics() {
 	m.totalSizeOfReceivedMessages = prometheus.NewCounter(prometheus.CounterOpts{
 		Namespace: m.cfg.MetricsNamespace,
 		Subsystem: subsystem,
-		Name:      "received_broadcasts_size",
+		Name:      "received_broadcasts_bytes",
```
pracucci: A counter should have the `_total` suffix according to [Prometheus conventions](https://prometheus.io/docs/practices/naming/). Same comment applies to other counters in this file.

_I'm mentioning it because memberlist is still experimental, so introducing breaking changes in the metrics name shouldn't be an issue._

pkg/ring/kv/memberlist/metrics.go
```diff
@@ -161,13 +160,28 @@ func (m *Client) createAndRegisterMetrics() {
...
+		cfg := armonmetrics.DefaultConfig("")
+		cfg.EnableHostname = false         // no need to put hostname into metric
+		cfg.EnableHostnameLabel = false    // no need to put hostname into labels
+		cfg.EnableRuntimeMetrics = false   // metrics about Go runtime already provided by prometheus
+		cfg.EnableTypePrefix = true        // to make better sense of internal memberlist metrics
+		cfg.TimerGranularity = time.Second // timers are in seconds in prometheus world
+
+		armonmetrics.NewGlobal(cfg, sink)
+	} else {
+		fmt.Println("Error while registering ")
```
pracucci: I guess we want a log here, not `fmt.Println()`

pkg/ring/kv/memberlist/metrics.go
pstibrany: > A counter should have the _total suffix according to Prometheus conventions. Same comment applies to other counters in this file.

Fixed (I don't like this specific convention at all).

> I'm mentioning it because memberlist is still experimental, so introducing breaking changes in the metrics name shouldn't be an issue.

That's why this PR is fixing it now.

pracucci: Should we change metrics in `tcp_transport.go` as well, to keep consistency?

pstibrany: > Should we change metrics in `tcp_transport.go` as well, to keep consistency?

You're adding more salt to the wound now :) OK, ok, ...

pracucci: Unintentionally. I'm OK if you prefer to remove the `_total` suffix, but if we add it, we should do it in all memberlist metrics.

PR #1863, opened 2019-11-29
Reuse request timeseries slice on error in the ingester

pracucci:
**What this PR does**:
If I'm not missing anything, this should be a safe change to do, to make sure a request timeseries slice gets reused (in the ingester) even in case of error.

_No issue_

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


bboreham:
Which kinds of error did you have in mind?
The benefit seems marginal, given `defer` has a cost on the happy path.

pracucci:
You're right. Re-checking the code, soft errors (like out of order timestamp) did reuse the slice, so slices were not reused only in case of hard errors, which are hard.

_Closing it, thanks for pointing out._

PR #1865, opened 2019-12-02
Added fingerprint collisions to the test.

pstibrany:
This PR adds one test, which shows collisions in Fingerprint function. We knew that Fingerprint function has collisions, but now we have specific examples, and simple way to generate more.

### DIFF
--- a/pkg/ingester/client/compat_test.go
+++ b/pkg/ingester/client/compat_test.go
@@ -130,6 +130,12 @@ func TestFingerprintCollisions(t *testing.T) {
 	c1 = metric.Set("A", Alabel1).Labels()
 	c2 = metric.Set("A", Alabel2).Labels()
 	verifyCollision(t, true, c1, c2)
+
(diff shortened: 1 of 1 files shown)


gouthamve:
Umm, what is this test doing and why should it be included in the code-base? Just curious as I don't have a lot of context on the whole collision discussions.

pstibrany:
> Umm, what is this test doing and why should it be included in the code-base? Just curious as I don't have a lot of context on the whole collision discussions.

We have couple of places in the code, where we don't handle collisions properly. We have discussed whether it's worth fixing them or how often they show up. I was trying to show that they are in fact real, and this test shows some exact colliding label sets, and even gives you hint how to create any number of them. (by using those colliding "_" and "A" labels from the test).

gouthamve:
Ah makes sense.

> It's mostly for documentation purposes.

Could you put the info in a comment above the test so that in the future we know why it is there?

PR #1866, opened 2019-12-02
Do not consider PENDING state as healthy

pracucci:
**What this PR does**:
While working on the PR #1818 I've realized that an ingester is considered healthy for read operations while `PENDING` but not while `JOINING`. From my perspective, an ingester shouldn't be considered healthy while `PENDING`, so I'm suggesting to:

1. Do not consider `PENDING` state as healthy
2. Explicitly enumerate healthy ingester states for the read path

In the read path, there are some API endpoints like `LabelNames` and `LabelValuesForLabelName` for which we query all healthy ingesters (see `distributor.forAllIngesters()`). A `PENDING` ingester shouldn't hold any data yet, so shouldn't be required to be hit.

A comment I've received in a previous discussion on this topic is that this change may introduce quorum issues during ingesters rollout. However, I can't see a real difference compared to when the ingester switches from `PENDING` to `JOINING`, considered that we already consider the `JOINING` state as unhealthy.

_No issue_

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/model.go
+++ b/pkg/ring/model.go
@@ -184,7 +184,7 @@ func (d *Desc) TokensFor(id string) (tokens, other []uint32) {
 func (i *IngesterDesc) IsHealthy(op Operation, heartbeatTimeout time.Duration) bool {
 	if op == Write && i.State != ACTIVE {
 		return false
-	} else if op == Read && i.State == JOINING {
+	} else if op == Read && i.State != ACTIVE && i.State != LEAVING {
 		return false
 	}
 	return time.Now().Sub(time.Unix(i.Timestamp, 0)) <= heartbeatTimeout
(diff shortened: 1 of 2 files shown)


pstibrany:
Personally, I would find it easier to read `IsHealthy` method, if "true" cases were more explicit. As it is now, each time I read it, I find it confusing. :-(

csmarchbanks:
I am concerned that this could break queries after a single ingester failure. If an ingester pod vanishes (node issue or something), it will still be unhealthy in the ring, and now the new pending pod will also be considered unhealthy causing all reads to fail. Would you be able to test that?

bboreham:
I think Chris is correct, and I think the solution is not to count pending ingesters as an error.
That implies a change to `Ring.GetAll()`.
Maybe we can lose `reallyAll` at the same time.

pracucci:
_Closing for now_

PR #1867, opened 2019-12-02
Declass ingester fingerprint hash collision log to debug level

pracucci:
**What this PR does**:
We're aware that there are fingerprint hash collisions and we handle them through the mapper. Is it really worth log them as Info level or can we declass it to Debug?

_A train of these logs is quite common during ingesters rollout, logged by the joining ingester while receiving all series from the leaving one._

_No issue_

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
Btw, this mapper is used with 'fast fingerprints', totally different function ;-) (much more collision-prone).

PR #1878, opened 2019-12-04
Distributing sum queries

owen-d:
## What
This PR (squashed from https://github.com/cortexproject/cortex/pull/1730 for simplicity) introduces a new query-frontend flag, `querier.sum-shards`, which allows cortex to fan out and process queries in parallel where appropriate. In order for a query to be parallelizeable, the query itself must be parallelizeable and it must use a sharded schema (v10+).

#### Sharding
The v10 schema introduced a shard factor which spreads series across n shards. This is a prerequisite for allowing us to query data from these shards in parallel.

### Problem
Although the v10 schema change introduces a shard factor, all shards must still be processed on a single querier. This is compounded by the fact that query ASTs are not broken down before execution. Therefore, while sharding lets us split up and process data at it's source location, we still re-aggregate it and process it all in one place. The two goals of this PR are to fix these by splitting up child ASTs into `n` merge-able queries for individual execution, where `n = shard factor`.

### Mapping ASTs
Firstly, we introduce a package called `astmapper` with the interface:

```
// ASTMapper is the exported interface for mapping between multiple AST representations
type ASTMapper interface {
	Map(node promql.Node) (promql.Node, error)
}
```

This interface is used to map ASTs into different ASTs. We use this to turn i.e.

`sum by (foo) (rate(bar{baz=”blip”}[1m]))` into approximately

```
sum by (foo) (
  sum by (foo) (rate(bar{baz=”blip”,__cortex_shard__=”0of16”}[1m])) or
  sum by (foo) (rate(bar{baz=”blip”,__cortex_shard__=”1of16”}[1m])) or
  ...
  sum by (foo) (rate(bar{baz=”blip”,__cortex_shard__=”15of16”}[1m]))
)
```

There are more complex optimizations to our sum mapping, but this is a simplified example for the PR description.

This works largely because summing the sub-sums is equivalent. The principle should hold true for any merge operation that is associative, but sums are the initial focus.

### `querier.sum-shard` (bool)
This flag enables shard summing behavior

### Schema config on frontend
The query-frontend now requires a schema config to determine how/when to shard queries, either from a file or from flags (i.e. by the `config-yaml` CLI flag). This is the same schema config the queriers consume. Queries are currently shardable when they only span a single shard-enabled schema definition (v10+) although this could be improved in later work.

### Concurrency controls
It's also advised to increase downstream concurrency controls as well to account for more queries of smaller sizes:
  * `querier.max-outstanding-requests-per-tenant`
  * `querier.max-query-parallelism`
  * `querier.max-concurrent`
  * `server.grpc-max-concurrent-streams`

It is also worth noting that instrumentation (traces) also scale with the number of sharded queries. We've had success scaling up our `JAEGER_REPORTER_MAX_QUEUE_SIZE` in order to handle the new volume.

### Work distribution
With query sharding enabled, the query-frontends do more work than previously. This runs afoul of the fact that they aren't horizontally scalable yet (per tenant queues are per process). Therefore, it's highly suggested to vertically scale them compared to before. In the future, this can be mitigated by the following:
- making the query-frontends horizontally scalable (co-opt @pracucci's centralized limits or similar for the frontends)
- Support recursive parallelization for queriers (detailed later).

### parallelization vs ast delegation
In addition to parallelizing sum operations, this PR still distributes non-parallelized operations to downstream queriers. In the following example, we parallelize the `sum(rate...))` and distribute the largest possible child ast (the `sum(histogram_quantile...))` in this case) to a querier.

```
sum(histogram_quantile(0.5, rate(selector[1m]))) +
sum(rate(selector[1m]))
```
-> 
shard `sum(rate(selector[1m]))` and delegate `sum(histogram_quantile(0.5, rate(selector[1m])))` to a querier.

#### Parallelize merge operations
This PR only parallelizes sums, but it should be possible for any [_associative_](https://en.wikipedia.org/wiki/Associative_property) merge operation. Some examples could be:

- `avg(selector)` -> `sum(selector)/count(selector)`, which can be parallelized in turn by their shard factors
- `rate(selector[1m])` -> `rate(selector{shard=0}[1m]) + rate(selector{shard=1}[1m])`

#### pipe dream: recursive parallelization
Theoretically, queriers could perform a single aggregation and parallelize/recursively delegate to other queriers. It would likely require a planner/scheduler and could be extended to support intermediate caching. However, it is beyond the scope of this PR.

## Benchmarks
Here are some benchmark results for different query types, shard factors, and artificial delay introduced (to simulate downstream executions):

```
go test -test.bench BenchmarkQuerySharding
goos: darwin
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/querier/queryrange
BenchmarkQuerySharding/desc:[sum_nogroup]---shards:[1]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                        2         668514448 ns/op
BenchmarkQuerySharding/desc:[sum_nogroup]---shards:[2]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                        3         345328972 ns/op
BenchmarkQuerySharding/desc:[sum_nogroup]---shards:[4]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                        6         173694813 ns/op

BenchmarkQuerySharding/desc:[sum_nogroup]---shards:[1]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                     1        1120235887 ns/op
BenchmarkQuerySharding/desc:[sum_nogroup]---shards:[2]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                     2         576303220 ns/op
BenchmarkQuerySharding/desc:[sum_nogroup]---shards:[4]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                     4         289560953 ns/op

--------------------------------

BenchmarkQuerySharding/desc:[sum_by]---shards:[1]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                             2         807234470 ns/op
BenchmarkQuerySharding/desc:[sum_by]---shards:[2]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                             3         398730618 ns/op
BenchmarkQuerySharding/desc:[sum_by]---shards:[4]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                             5         202423816 ns/op

BenchmarkQuerySharding/desc:[sum_by]---shards:[1]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                          1        1250864781 ns/op
BenchmarkQuerySharding/desc:[sum_by]---shards:[2]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                          2         624703704 ns/op
BenchmarkQuerySharding/desc:[sum_by]---shards:[4]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                          4         324303066 ns/op

--------------------------------

BenchmarkQuerySharding/desc:[sum_without]---shards:[1]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                        2         893648764 ns/op
BenchmarkQuerySharding/desc:[sum_without]---shards:[2]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                        2         777097324 ns/op
BenchmarkQuerySharding/desc:[sum_without]---shards:[4]---series:[4096]---delayPerSeries:[0s]---samplesPerSeries:[100]-8                        2         627139368 ns/op

BenchmarkQuerySharding/desc:[sum_without]---shards:[1]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                     1        1367469450 ns/op
BenchmarkQuerySharding/desc:[sum_without]---shards:[2]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                     1        1012079071 ns/op
BenchmarkQuerySharding/desc:[sum_without]---shards:[4]---series:[4096]---delayPerSeries:[100µs]---samplesPerSeries:[100]-8                     2         791027808 ns/op

--------------------------------

PASS
```

## Tooling:
I've also added a little tool, `query-audit`, which can run a series of queries against multiple prometheus backends and compare the results (I can add make targets if requested). I've used this to empirically test differences in query results across sharded and non-shard enabled backends. Output is below.
```
$ go install ./tools/query-audit/ && query-audit -f ~/grafana/tmp/equivalence-config.yaml

0.000000% avg diff for:
        query: sum(rate(container_cpu_usage_seconds_total[5m]))
        series: 1
        samples: 289
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-28 00:00:00 +0000 UTC
        step: 15m0s

0.000000% avg diff for:
        query: sum(rate(container_cpu_usage_seconds_total[5m])) by (container_name)
        series: 95
        samples: 25877
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-28 00:00:00 +0000 UTC
        step: 15m0s

0.000000% avg diff for:
        query: sum(rate(container_cpu_usage_seconds_total[5m])) without (container_name)
        series: 4308
        samples: 374989
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-26 00:00:00 +0000 UTC
        step: 15m0s

0.000000% avg diff for:
        query: histogram_quantile(0.9, sum(rate(cortex_cache_value_size_bytes_bucket[5m])) by (le, job))
        series: 13
        samples: 325
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-25 06:00:00 +0000 UTC
        step: 15m0s

0.000000% avg diff for:
        query: sum without (instance, job) (rate(cortex_query_frontend_queue_length[5m])) or sum by (job) (rate(cortex_query_frontend_queue_length[5m]))
        series: 21
        samples: 525
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-25 06:00:00 +0000 UTC
        step: 15m0s

0.000000% avg diff for:
        query: sum without (instance, job) (rate(cortex_cache_request_duration_seconds_count[5m])) or rate(cortex_cache_request_duration_seconds_count[5m])
        series: 942
        samples: 23550
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-25 06:00:00 +0000 UTC
        step: 15m0s

0.000000% avg diff for:
        query: sum by (namespace) (predict_linear(container_cpu_usage_seconds_total[5m], 10))
        series: 16
        samples: 400
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-25 06:00:00 +0000 UTC
        step: 15m0s

0.000000% avg diff for:
        query: sum by (namespace) (avg_over_time((rate(container_cpu_usage_seconds_total[5m]))[10m:]) > 1)
        series: 4
        samples: 52
        start: 2019-11-25 00:00:00 +0000 UTC
        end: 2019-11-25 01:00:00 +0000 UTC
        step: 5m0s
```


pkg/cortex/modules.go
```diff
@@ -215,7 +218,12 @@ func (t *Cortex) stopDistributor() (err error) {
 }
 
 func (t *Cortex) initQuerier(cfg *Config) (err error) {
-	var store querier.ChunkStore
+	t.worker, err = frontend.NewWorker(cfg.Worker, httpgrpc_server.NewServer(t.server.HTTPServer.Handler), util.Logger)
+	if err != nil {
+		return
+	}
+
+	var store chunkstore.ChunkStore
```
rfratto: I'm not sure I understand why `t.worker` is initialized twice in this function.

pkg/chunk/chunk_store.go
```diff
@@ -459,6 +462,9 @@ func (c *store) lookupEntriesByQueries(ctx context.Context, queries []IndexQuery
 }
 
 func (c *store) parseIndexEntries(ctx context.Context, entries []IndexEntry, matcher *labels.Matcher) ([]string, error) {
+	log, ctx := spanlogger.New(ctx, "store.parseIndexEntries")
```
cyriltovena: Do we really need this span ? This should be pretty fast since it is cpu bound.

pkg/chunk/schema_config.go
```diff
@@ -192,15 +192,29 @@ func (cfg *SchemaConfig) loadFromFile() error {
...
 		if err := periodCfg.validate(); err != nil {
 			return err
 		}
-	}
 
+		// apply default row shards
+		if periodCfg.RowShards == 0 {
+			periodCfg.RowShards = defaultRowShards(periodCfg.Schema)
+			cfg.Configs[i] = periodCfg
+		}
```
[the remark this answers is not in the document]
cyriltovena: I think it's a good idea.

pkg/chunk/series_store.go
```diff
@@ -131,6 +132,15 @@ func (c *seriesStore) Get(ctx context.Context, userID string, from, through mode
 		return nil, err
 	}
 
+	// inject artificial __cortex_shard__ labels if present in the query. GetChunkRefs guarantees any chunk refs match the shard.
+	shard, _, err := astmapper.ShardFromMatchers(allMatchers)
```
cyriltovena: You might want to do that in `GetChunkRefs` in case someone (Loki for instance does that) uses the refs first then fetches chunks after.

cyriltovena: Well might not work because `fetcher.FetchChunks` might validate the metric. To be verified.

pkg/chunk/series_store.go
```diff
@@ -252,10 +262,21 @@ func (c *seriesStore) lookupSeriesByMetricNameMatchers(ctx context.Context, from
 	log, ctx := spanlogger.New(ctx, "SeriesStore.lookupSeriesByMetricNameMatchers", "metricName", metricName, "matchers", len(matchers))
 	defer log.Span.Finish()
 
+	// Check if one of the labels is a shard annotation, pass that information to lookupSeriesByMetricNameMatcher,
+	// and remove the label.
```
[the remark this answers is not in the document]
cyriltovena: We need to remove the matcher because it doesn't exist in the store. if we leave it `lookupSeriesByMetricNameMatcher` would returns no result. Basically the matchers is used to convey the information until here.

pkg/chunk/storage/caching_index_client.go
```diff
@@ -66,6 +66,9 @@ func (s *cachingIndexClient) Stop() {
 }
 
 func (s *cachingIndexClient) QueryPages(ctx context.Context, queries []chunk.IndexQuery, callback func(chunk.IndexQuery, chunk.ReadBatch) (shouldContinue bool)) error {
+	log, ctx := spanlogger.New(ctx, "cachingIndexClient.QueryPages")
```
cyriltovena: I think can remove that trace too.

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## master / unreleased
 
+* [FEATURE] Fan out parallelizable queries to backend queriers concurrently.
+  * `-querier.sum-shards` (bool)
+  * Requires a shard-compatible schema (v10+)
+  * This causes the number of traces to increase accordingly.
+  * The query-frontend now requires a schema config to determine how/when to shard queries, either from a file or from flags (i.e. by the `config-yaml` CLI flag). This is the same schema config the queriers consume.
```
cyriltovena: IMHO this specific line should be a `[CHANGE]`, because the frontend won't work anymore without it.

owen-d: This should still be fine. The schema load path will automatically add a `v1` schema if overrides (config file, etc) are not specified.

pkg/querier/astmapper/embedded.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Squash reduces an AST into a single vector or matrix query which can be hijacked by a Queryable impl.
+func Squash(codec Codec, isMatrix bool, nodes ...promql.Node) (promql.Expr, error) {
+	if codec == nil {
+		return nil, errors.Errorf("nil Codec")
+	}
+
+	// concat OR legs
+	var strs []string
```
cyriltovena:
```suggestion
	strs := make([]string, 0, len(nodes))
```

pkg/querier/astmapper/parallel.go
```diff
@@ -0,0 +1,91 @@
...
+var summableAggregates = map[promql.ItemType]struct{}{
+	promql.ItemSum:     {},
+	promql.ItemMin:     {},
+	promql.ItemMax:     {},
+	promql.ItemTopK:    {},
+	promql.ItemBottomK: {},
+	promql.ItemCount:   {},
+}
+
+// CanParallel tests is a subtree is parallelizable..
```
cyriltovena:
```suggestion
// CanParallel tests if a subtree is parallelizable..
```

pkg/querier/astmapper/parallel.go
```diff
@@ -0,0 +1,91 @@
...
+
+	default:
+		panic(errors.Errorf("CanParallel: unhandled node type %T", node))
+	}
+
+}
+
+// ParallelFunc ensures that a promql function can be part of a parallel query.
+func ParallelFunc(f promql.Function) bool {
+	unallowed := []string{
```
cyriltovena: you might be better with a map[string]struct and a global one like `summableAggregates`.

pkg/querier/chunkstore/chunkstore.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+// ChunkStore is the read-interface to the Chunk Store.  Made an interface here
+// to reduce package coupling.
+type ChunkStore interface {
```
cyriltovena: Since go interfaces are implemented implicitly you usually want to declare them where you need them. 

I might be wrong but you could copy the interface where it was required or coupled. I think this would be better.

pkg/chunk/series_store.go
owen-d: This is injecting a `cortex_shard` label into each of the resulting series because mapped queries expect its existence.

If we query `metric{__cortex_shard__="1_of_2"}`, we expect the responses from the querier to include this `__cortex_shard__="1_of_2"` label.

`GetChunkRefs` does only return the shard-specific series if a shard label is present in the query. This logic is handled below in `lookupSeriesByMetricNameMatcher{s,}`

pkg/querier/chunkstore/chunkstore.go
owen-d: I moved this code to prevent a circular pkg dependency after refactoring the `LazyQueryable` into its own package (in order for consumption by multiple other packages) . It currently exists here:
https://github.com/cortexproject/cortex/blob/f3d3ddba04d7ec30af68f3a11e7f7b25976a7c2f/pkg/querier/querier.go#L55-L59

I don't mind redefining it in multiple places if that's preferred -- let me know.

cyriltovena: No big opinion here, just giving you an alternative.

pkg/querier/queryrange/querysharding.go
```diff
@@ -0,0 +1,311 @@
...
+
+// NewQueryShardMiddleware creates a middleware which downstreams queries after AST mapping and query encoding.
+func NewQueryShardMiddleware(
+	logger log.Logger,
+	engine *promql.Engine,
+	confs ShardingConfigs,
+	codec Codec,
+	minShardingLookback time.Duration,
+) Middleware {
+	passthrough := MiddlewareFunc(func(next Handler) Handler {
```
cyriltovena: This could be a global variable.

cyriltovena: I would even make it public for other packages extending the middleware chain.

pkg/querier/queryrange/querysharding.go
```diff
@@ -0,0 +1,311 @@
...
+	if noshards {
+		level.Warn(logger).Log(
+			"middleware", "QueryShard",
+			"msg", "no configuration with shard found",
+			"confs", fmt.Sprintf("%+v", confs),
+		)
+		return passthrough
+	}
+
+	getConf := func(r Request) (chunk.PeriodConfig, error) {
```
cyriltovena: I think this could be a function of the alias `ShardingConfigs` created line 24. This should simplify the code by using the alias as dependency instead of a unnamed function that you had to defined 3 time.

pkg/querier/queryrange/querysharding.go
```diff
@@ -0,0 +1,311 @@
...
+	shardCh := make(chan Response, 1)
+	errCh := make(chan error, 2)
+
+	go func() {
+		res, err := splitter.next.Do(ctx, nonsharded)
+		if err != nil {
+			errCh <- err
+			return
+		}
+		nonshardCh <- res
```
cyriltovena: Should you select on sending the result or context canceled ? I guess if we get cancelled this shouldn't stay blocked on sending.

same goes for line 266.

pkg/querier/queryrange/queryable.go
```diff
@@ -0,0 +1,120 @@
...
+			if err != nil {
+				errCh <- err
+				return
+			}
+			streams, err := ResponseToSamples(resp)
+			if err != nil {
+				errCh <- err
+				return
+			}
+			samplesCh <- streams
```
cyriltovena: You might also need to select on the `context.Done()`.

pkg/querier/queryrange/value.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/promql"
+)
+
+// FromResult transforms a promql query result into a samplestream
+func FromResult(res *promql.Result) ([]SampleStream, error) {
+	if res.Err != nil {
```
cyriltovena: should you add a nil check ? not sure about the context of this when called though.

tools/query-audit/auditor.go
```diff
@@ -0,0 +1,92 @@
+package main
+
+import (
```
cyriltovena: Not sure why this tools/query-audit was added. May be a rebase ?

cyriltovena:
We might also want to add some metrics instrumentations to know how many requests has been // by shard sums, or even tag an existing metrics to compare shard vs non shard in the past, up to you.

pkg/querier/queryrange/querysharding.go
owen-d: great idea!

pkg/querier/queryrange/querysharding.go
owen-d: the channels are buffered here so it should be fine

pkg/querier/queryrange/queryable.go
owen-d: These channels are also buffered here

tools/query-audit/auditor.go
owen-d: This was added to get the diffs between results when querying sharded and non sharded versions of the read path. The output is included in the _Tooling_ section of the PR description.

pkg/querier/queryrange/value.go
owen-d: This comes from the prometheus promql engine and should be fine:
```
type Query interface {
	// Exec processes the query. Can only be called once.
	Exec(ctx context.Context) *Result
	// Close recovers memory used by the query result.
	Close()
	// Statement returns the parsed statement of the query.
	Statement() Statement
	// Stats returns statistics about the lifetime of the query.
	Stats() *stats.QueryTimers
	// Cancel signals that a running query execution should be aborted.
	Cancel()
}
```

tools/query-audit/auditor.go
cyriltovena: Sorry I missed this. will take a look

pkg/chunk/schema_util.go
```diff
@@ -48,7 +48,8 @@ func labelsString(ls labels.Labels) string {
 	return b.String()
 }
 
-func labelsSeriesID(ls labels.Labels) []byte {
+// LabelsSeriesID hashes a label set
+func LabelsSeriesID(ls labels.Labels) []byte {
```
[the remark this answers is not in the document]
owen-d: You're correct. It was made public in an earlier iteration when ingesters had shard-aware reads. I'll make it private again.

pkg/chunk/chunk_store_utils.go
```diff
@@ -252,3 +256,10 @@ func (c *Fetcher) processCacheResponse(chunks []Chunk, keys []string, bufs [][]b
 	}
 	return found, missing, err
 }
+
+func injectShardLabels(chunks []Chunk, shard astmapper.ShardAnnotation) {
```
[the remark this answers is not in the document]
owen-d: I agree, but put it here in order to avoid a circular dependency between `astmapper` and `chunk` pkgs. I can investigate a way to do this via injection, a sidecar package, etc, but I felt this would be cumbersome and another layer of confusion in an already large PR. Let me know what you'd prefer.

pkg/querier/astmapper/embedded.go
```diff
@@ -0,0 +1,120 @@
...
+	EmbeddedQueryFlag = "__embedded_queries__"
+)
+
+// EmbeddedQueries is a wrapper type for encoding queries
+type EmbeddedQueries struct {
+	Concat []string `json:"Concat"`
+}
+
+// JSONCodec is a Codec impl that uses JSON representations of EmbeddedQueries structs
+var JSONCodec Codec = jsonCodec{}
```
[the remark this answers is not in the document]
owen-d: Good catch. This was from an earlier iteration where I tried a few different codec approaches. I'll remove this for simplicity.

pkg/querier/astmapper/embedded.go
```diff
@@ -0,0 +1,120 @@
...
+
+	encoded := codec.Encode(strs)
+
+	embeddedQuery, err := labels.NewMatcher(labels.MatchEqual, QueryLabel, encoded)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if isMatrix {
```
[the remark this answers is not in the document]
owen-d: No, it should not. I'll replace it with an error case.

pkg/chunk/schema.go
```diff
@@ -226,13 +235,23 @@ type entries interface {
 	GetReadMetricLabelValueQueries(bucket Bucket, metricName string, labelName string, labelValue string) ([]IndexQuery, error)
 	GetChunksForSeries(bucket Bucket, seriesID []byte) ([]IndexQuery, error)
 	GetLabelNamesForSeries(bucket Bucket, seriesID []byte) ([]IndexQuery, error)
+	FilterReadQueries(queries []IndexQuery, shard *astmapper.ShardAnnotation) []IndexQuery
+}
+
+// noops is a placeholder which can be embedded to provide default implementations
```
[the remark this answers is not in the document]
owen-d: Sure.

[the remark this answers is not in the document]
owen-d: I personally feel the struct embedding method is a cleaner way to do this, but I'll provide methods on each schema instead.

pkg/chunk/chunk_store_utils.go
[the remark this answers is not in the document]
jtlisi: It's fine to keep it here for now.

jtlisi:
@owen-d can you rebase this on to master? The vendored version of prometheus recently changed and I think some variable names you use in astmapper need to be updated.

pkg/querier/astmapper/embedded.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return &promql.VectorSelector{
+		Name:          EmbeddedQueryFlag,
+		LabelMatchers: []*labels.Matcher{embeddedQuery},
+	}, nil
+
+}
+
+// OrSquasher is a custom squasher which mimics the intuitive but less efficient OR'ing of sharded vectors.
+func OrSquasher(nodes ...promql.Node) (promql.Expr, error) {
```
jtlisi: Is this only used in testing? Do we need to keep it?

pkg/querier/queryrange/querysharding.go
```diff
@@ -0,0 +1,307 @@
...
+func (splitter *shardSplitter) parallel(ctx context.Context, sharded, nonsharded Request) (Response, error) {
+	if sharded == nil {
+		return splitter.next.Do(ctx, nonsharded)
+	}
+
+	if nonsharded == nil {
+		return splitter.shardingware.Do(ctx, sharded)
+	}
+
+	nonshardCh := make(chan Response, 1)
```
jtlisi: None of these channels are ever closed. In the event of a timeout I think they will remain open forever. This will likely cause a memory leak over time.

jtlisi: I think it should be fine, the GC should take care of it actually

pkg/cortex/modules.go
```diff
@@ -332,11 +336,32 @@ func (t *Cortex) stopStore() error {
 }
 
 func (t *Cortex) initQueryFrontend(cfg *Config) (err error) {
+	err = cfg.Schema.Load()
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
owen-d: This follows the [naked return](https://tour.golang.org/basics/7) pattern. This pattern is also used in `initStore`.

pkg/ingester/ingester.go
```diff
@@ -519,6 +519,7 @@ func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_
 		}
 
 		numChunks += len(wireChunks)
+
```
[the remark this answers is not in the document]
owen-d: This is undoubtedly a relic from when this PR included ingester-aware query sharding. I assume the lines with code were removed during the refactor to simply not shard queries during the ingester lookback duration.

pkg/chunk/schema.go
```diff
@@ -847,4 +892,5 @@ func (v11Entries) GetLabelNamesForSeries(bucket Bucket, seriesID []byte) ([]Inde
 			HashValue: string(seriesID),
 		},
 	}, nil
+
```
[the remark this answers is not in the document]
owen-d: This is probably due to the refactoring you requested where I removed the `noop` embedded type which contained default implementations for the schema entries interface in favor of deliberately implementing that function on each schema. This errant line must have made it through during that change.

pkg/chunk/schema.go
```diff
@@ -784,6 +813,22 @@ func (v10Entries) GetLabelNamesForSeries(_ Bucket, _ []byte) ([]IndexQuery, erro
...
 
+// FilterReadQueries will return only queries that match a certain shard
+func (v10Entries) FilterReadQueries(queries []IndexQuery, shard *astmapper.ShardAnnotation) (matches []IndexQuery) {
+	if shard == nil {
+		return queries
+	}
+
+	for _, query := range queries {
+		s := strings.Split(query.HashValue, ":")[0]
+		n, err := strconv.Atoi(s)
```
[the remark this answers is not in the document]
owen-d: Sure. It shouldn't be possible for this to incorrectly parse without mangling our Indices (this is particular to the `v10Entries`).

pkg/querier/astmapper/embedded.go
owen-d: Good catch - it should definitely not be outside of the test package. It's a relic of an earlier implementation, but is handy for tests because of its intuitive & human readable output:
https://github.com/cortexproject/cortex/pull/1878/files/fb6f97e796ebbae2d89f207aa08a0b61d269a595#diff-b84e4282a45b40a259c28a087b872fdbR83-R91

pkg/cortex/modules.go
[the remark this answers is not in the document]
jtlisi: Gotcha, I was thrown off by the explicit return in the line below

pkg/chunk/chunk_store_utils.go
```diff
@@ -252,3 +256,10 @@ func (c *Fetcher) processCacheResponse(chunks []Chunk, keys []string, bufs [][]b
...
+	for i, chunk := range chunks {
+		chunk.Metric = append(chunk.Metric, shard.Label())
```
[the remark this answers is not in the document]
pstibrany: Agree with Marco – better be safe and adhere to contract then break it. Suggested code looks good, although given that labels are supposedly sorted already, simple iteration to find the insertion point would do the trick as well. `__cortex_shard__` will most likely go first anyway (but may not).

docs/configuration/arguments.md
```diff
@@ -68,6 +68,28 @@ The ingester query API was improved over time, but defaults to the old behaviour
...
+   ```
+   When enabled, the query-frontend requires a schema config to determine how/when to shard queries, either from a file or from flags (i.e. by the `config-yaml` CLI flag). This is the same schema config the queriers consume.
+   It's also advised to increase downstream concurrency controls as well to account for more queries of smaller sizes:
+
+   - `querier.max-outstanding-requests-per-tenant`
+   - `querier.max-query-parallelism`
+   - `querier.max-concurrent`
+   Furthermore, both querier and query-frontend components require the `querier.query-ingesters-within` parameter to know when to start sharding requests (ingester queries are not sharded). It's recommended to align this with `ingester.max-chunk-age`.
+
+   Instrumentation (traces) also scale with the number of sharded queries and it's suggested to account for increased throughput there as well.
```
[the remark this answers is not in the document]
owen-d: I opted to leave this out as there are multiple ways to configure this, but will add this as an example.

pkg/chunk/schema_config.go
```diff
@@ -309,6 +317,14 @@ func (cfg *SchemaConfig) Load() error {
 		return err
 	}
 
+	for i, periodCfg := range cfg.Configs {
+		// apply default row shards
+		if periodCfg.RowShards == 0 {
+			periodCfg.RowShards = defaultRowShards(periodCfg.Schema)
+			cfg.Configs[i] = periodCfg
+		}
+	}
```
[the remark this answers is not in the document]
owen-d: good catch - it'll be much cleaner with that logic deduped.

pkg/querier/astmapper/embedded.go
```diff
@@ -0,0 +1,85 @@
...
+type jsonCodec struct{}
+
+func (c jsonCodec) Encode(queries []string) string {
+	embedded := EmbeddedQueries{
+		Concat: queries,
+	}
+	b, err := json.Marshal(embedded)
+
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
owen-d: good call!

pkg/querier/astmapper/parallel.go
```diff
@@ -0,0 +1,93 @@
...
+		return true
+
+	case *promql.EvalStmt:
+		return CanParallel(n.Expr)
+
+	case *promql.MatrixSelector, *promql.NumberLiteral, *promql.StringLiteral, *promql.VectorSelector:
+		return true
+
+	default:
+		panic(errors.Errorf("CanParallel: unhandled node type %T", node))
```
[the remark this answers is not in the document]
owen-d: I can understand that. I opted to panic here because a panic forces us to account for new node types in the future whereas errors are likely to get logged & ignored. Perhaps it's a bit heavy handed though - I really wish I knew a way to get compile time guarantees for this in our type system.

At the end of the day I think it's a choice between being forced to account for promql extensions or not. I'll change it to log & return false.

pkg/querier/astmapper/shard_summer.go
```diff
@@ -0,0 +1,295 @@
...
+			selector.LabelMatchers...,
+		),
+	}, nil
+}
+
+// ParseShard will extract the shard information encoded in ShardLabelFmt
+func ParseShard(input string) (parsed ShardAnnotation, err error) {
+	if !ShardLabelRE.MatchString(input) {
+		return parsed, errors.Errorf("Invalid ShardLabel value: [%s]", input)
+	}
```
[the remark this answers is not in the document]
owen-d: I personally think the readability outweighs the costs here (it's nice to be able to read snake case in distributed tracing spans/logs). If you still think it's worth the change, we can make it.

pkg/chunk/chunk_store_utils.go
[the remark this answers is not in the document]
owen-d: Thanks for the example code, I'll definitely incorporate it. I believe a reason why this has not come up is because these shard labels are later filtered out (we don't return `__cortex_shard__` labels in responses).

owen-d:
Hey @pracucci, thanks for the kind words and the review. I've addressed all but one of your comments which I've pushed back on. 


Regarding
> The query-frontend now requires a schema config to determine how/when to shard queries, either from a file or from flags (i.e. by the config-yaml CLI flag).
This is only required if one wishes to use the new optimizations. If no config file is specified, the configuration is loaded via flags, which by default [adds a v1 schema](https://github.com/owen-d/cortex/blob/feature/query-sharding-squashed/pkg/chunk/schema_config.go#L140).

However, if a config file is specified, but it doesn't have any period configs, loading the frontend will now error. Should this case be handled or is this the right logic?

pkg/querier/astmapper/shard_summer.go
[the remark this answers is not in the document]
pracucci: Sounds good. I'm wondering if my comments still applies and we can get rid of `ShardLabelRE` anyway just doing `matches := strings.Split(input, "_of_")`

pracucci:
> However, if a config file is specified, but it doesn't have any period configs, loading the frontend will now error. Should this case be handled differently or is this the right logic?

This would be a problem both because of 3rd parties using the `query-frontend` in front of non-Cortex instances (ie. Thanos) and because of the new blocks storage (you can run the `query-frontend` in front of Cortex with the blocks storage even if the sum shard won't be supported).

However, I can't find a way to reproduce it and even looking at the code it's not clear to me which specific change has introduced this "breaking change". Could you share more details, please?

pracucci:
> Regarding the configs breaking change, I too couldn't find it. cfg.Schema.Load() will return nil if there are no configs :)

I've enhanced the `query-frontend` tests in #2167 (pending review) and I've also locally merged the PR #2167 into this PR and re-run integration tests: all green 💚.

pkg/querier/queryrange/roundtrip.go
```diff
@@ -50,6 +58,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.SplitQueriesByInterval, "querier.split-queries-by-interval", 0, "Split queries by an interval and execute in parallel, 0 disables it. You should use an a multiple of 24 hours (same as the storage bucketing scheme), to avoid queriers downloading and processing the same chunks. This also determines how cache keys are chosen when result caching is enabled")
 	f.BoolVar(&cfg.AlignQueriesWithStep, "querier.align-querier-with-step", false, "Mutate incoming queries to align their start and end with their step.")
 	f.BoolVar(&cfg.CacheResults, "querier.cache-results", false, "Cache query results.")
+	f.BoolVar(&cfg.SumShards, "querier.sum-shards", false, "Parse the ast and parallelize sums by shard.")
```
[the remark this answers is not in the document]
owen-d: Good call. Yes, there definitely are (I'm working on a cleaner impl for Loki because we control the `logql` pkg which should cover significantly more optimizations).

tools/query-audit/auditor.go
```diff
@@ -0,0 +1,92 @@
+package main
```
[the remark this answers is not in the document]
owen-d: This is some new tooling used to diff responses from different read endpoints and is used for correctness testing. It's included in the PR description, but I'll add some docs.

pkg/querier/queryrange/roundtrip.go
```diff
@@ -63,7 +63,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.SplitQueriesByInterval, "querier.split-queries-by-interval", 0, "Split queries by an interval and execute in parallel, 0 disables it. You should use an a multiple of 24 hours (same as the storage bucketing scheme), to avoid queriers downloading and processing the same chunks. This also determines how cache keys are chosen when result caching is enabled")
 	f.BoolVar(&cfg.AlignQueriesWithStep, "querier.align-querier-with-step", false, "Mutate incoming queries to align their start and end with their step.")
 	f.BoolVar(&cfg.CacheResults, "querier.cache-results", false, "Cache query results.")
-	f.BoolVar(&cfg.SumShards, "querier.sum-shards", false, "Parse the ast and parallelize sums by shard.")
+	f.BoolVar(&cfg.ShardedQueries, "querier.parallelise-shardable-queries", false, "Perform query parallelisations based on storage sharding configuration and query ASTs.")
```
pracucci: > Perform query parallelisations based on storage sharding configuration and query ASTs.

Please also something like `This feature is supported only by the chunks storage engine.`

docs/operations/query-auditor.md
```diff
@@ -0,0 +1,139 @@
+---
+title: "Query Auditor (tool)"
+linkTitle: "query auditor (tool)"
+weight: 1
```
pracucci: Upper case title and weight 2, please.

```
linkTitle: "Query Auditor (tool)"
weight: 2
```

docs/operations/query-auditor.md
```diff
@@ -0,0 +1,139 @@
...
+slug: query-auditor
+---
+
+The query auditor is a tool bundled in the cortex repo, but **not** included in docker images -- this must be built from source. It's primarily useful for those _developing_ cortex, but can be helpful to operators as well during certain scenarios (backend migrations come to mind).
```
pracucci: `repo` > `repository`

pracucci: `docker` > `Docker`

docs/operations/query-auditor.md
```diff
@@ -0,0 +1,139 @@
...
+
+## How it works
+
+The `query-audit` tool performs a set of queries against two backends that expose the prometheus read API. This is generally the `query-frontend` component of two cortex deployments. It will then compare the differences in the responses to determine the average difference for each query. It does this by:
+ - Ensuring the resulting label sets match
```
pracucci: Add a newline before this otherwise it's not formatted as list.

docs/operations/query-auditor.md
```diff
@@ -0,0 +1,139 @@
...
+linkTitle: "query auditor (tool)"
+weight: 1
+slug: query-auditor
+---
+
+The query auditor is a tool bundled in the cortex repo, but **not** included in docker images -- this must be built from source. It's primarily useful for those _developing_ cortex, but can be helpful to operators as well during certain scenarios (backend migrations come to mind).
+
+## How it works
+
+The `query-audit` tool performs a set of queries against two backends that expose the prometheus read API. This is generally the `query-frontend` component of two cortex deployments. It will then compare the differences in the responses to determine the average difference for each query. It does this by:
```
pracucci: `prometheus` > `Prometheus`

docs/operations/query-auditor.md
pracucci: `cortex` > `Cortex` (everywhere)

docs/operations/query-auditor.md
```diff
@@ -0,0 +1,139 @@
...
+ - Ensuring the resulting label sets match
+ - For each label set, ensuring they contain the same number of samples as their pair from the other backend
+ - For each sample, calculates their difference against it's pair from the other backend/label set.
+ - Calculates the average diff per query from the above diffs.
```
pracucci: No ending `.` here or you add an ending `.` to the other list items above

docs/operations/query-auditor.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Limitations
+
+It currently only supports queries with `Matrix` response types, but should be simple to extend to `Vector`s as well, should the need arise.
```
pracucci: I would just say:

> It currently only supports queries with `Matrix` response types

_The rest is a technical detail not useful to the final user._

docs/operations/query-auditor.md
```diff
@@ -0,0 +1,139 @@
...
+    end: 2019-11-25T06:00:00Z
+    step_size: 15m
+```
+
+### Example Output
+
+Under ideal circumstances, you'll see output like the following:
+
+```
+$ go install ./tools/query-audit/ && query-audit -f ~/grafana/tmp/equivalence-config.yaml
```
pracucci: Have you considered simplifying it into:
`go run ./tools/query-audit/ -f config.yaml`

PR #1885, opened 2019-12-05
Add flush job mode to ingester.

gouthamve:
This is my attempt to simplify things compared to #1747

/cc @jtlisi @codesome

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -143,6 +143,8 @@ func (c *Config) Validate() error {
 
 // Cortex is the root datastructure for Cortex.
 type Cortex struct {
+	cfg Config
+
 	target             moduleName
 	httpAuthMiddleware middleware.Interface
 
@@ -173,6 +175,7 @@ func New(cfg Config) (*Cortex, error) {
 	}
 
 	cortex := &Cortex{
+		cfg:    cfg,
 		target: cfg.Target,
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/cortex/cortex.go
```diff
@@ -238,7 +241,20 @@ func (t *Cortex) initModule(cfg *Config, m moduleName) error {
 
 // Run starts Cortex running, and blocks until a signal is received.
 func (t *Cortex) Run() error {
-	return t.server.Run()
+	var err error
+	done := make(chan struct{})
```
codesome: We are not waiting for this anywhere (I will create a follow-up PR to your branch, just a note to myself)

gouthamve:
> This might create it's own problem

Can you think of cases? I see this flush-job as something you run after the cluster has stabilised and I don't see any issues with it. Even if the cluster is in flux, I don't see it causing an major issues.

> There is a long wait before we start flushing even when we know it's a flush job

How / Where?

codesome:
> you run after the cluster has stabilised

If we assume this then should be fine. But it is still an assumption. I am afraid it might interfere with transfer otherwise? I prefer running this job without worrying about the state of the cluster, as that just adds another dependency for the job.

> How / Where?

Apparently it does not after the `TransferOut` is disabled :)

pkg/cortex/cortex.go
```diff
@@ -238,7 +241,22 @@ func (t *Cortex) initModule(cfg *Config, m moduleName) error {
...
+	go func() {
+		err = t.server.Run()
+
+		close(done)
+	}()
+
+	// Initiate shutdown if its a job to flush data.
+	if t.cfg.Ingester.IsFlushJob {
+		_ = t.Stop()
```
codesome: We need to make sure that calling `t.Stop()` twice is not a problem. Currently, `main.go` would also call this and try stopping the modules again.

codesome:
I have rebased https://github.com/cortexproject/cortex/pull/1747 on top of this with a small fix. I think we can close this now.

gouthamve:
Closing this as most of the changes have been incorporated in #1747

PR #1888, opened 2019-12-06
Fixed per-series sharding token generation in the distributor

pracucci:
**What this PR does**:
The PR #1726 introduced some changes in the distributor to allow to drop labels on a per-tenant basis. These changes also unintentionally changed the actual per-series sharding token: we noticed it while rolling out in one of our clusters, where an almost full resharding occurred (switching the cluster from 5M active series to 9M active series until the old sharing chunks have been flushed).

Before the PR #1726, the token was generated based on the labels **without** the HA replica label. After the PR #1726, the token is generated based on the labels **with** the HA replica label.

In this PR I'm fixing it, rolling back to the old behaviour, while honoring the intention of the PR #1726: the token should be generated based on the labels **without** the HA replica label **and without** the dropped ones.

_I haven't updated the changelog because the PR #1726 has been merged after the `0.4.0` release, so it's still in the unreleased limbo._

_No issue_

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -2,8 +2,8 @@ package distributor
 
 import (
 	"context"
-	"errors"
 	"flag"
+	"fmt"
 	"net/http"
 	"sort"
 	"strings"
@@ -90,8 +90,7 @@ var (
 		Name:      "distributor_replication_factor",
 		Help:      "The configured replication factor.",
 	})
-	emptyPreallocSeries     = ingester_client.PreallocTimeseries{}
-	errShardLabelsNotSorted = errors.New("Labels not sorted")
+	emptyPreallocSeries = ingester_client.PreallocTimeseries{}
 )
 
 // Distributor is a storage.SampleAppender and a client.Querier which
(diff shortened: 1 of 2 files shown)


pracucci:
> As PR #1726 hasn't been merged yet

@pstibrany It is merged.

pstibrany:
> @pstibrany It is merged.

You're right. So no problem there. This is a tricky bug, good that you've found it.

pkg/distributor/distributor.go
```diff
@@ -367,12 +371,14 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	keys := make([]uint32, 0, len(req.Timeseries))
 	validatedSamples := 0
 	for _, ts := range req.Timeseries {
-		key, err := d.tokenForLabels(userID, ts.Labels)
-		if err != nil {
+		key, validatedSeries, err := d.validateSeries(ts, userID, removeReplica)
+
+		// In case of a fatal error (ie. logic bug) we should fail the entire
+		// request and not consider it a partial error.
+		if err == errShardLabelsNotSorted {
```
tomwilkie: This seem like it might be a little brittle to me - previously all errors from `validateSeries` were "soft", now some are special.

pracucci: Right. From a design perspective, I think the main issue is that both labels removal and sharding token generation is done inside the `validateSeries()` which sounds "weird" (it does more than "validation").

I've pushed a commit where both labels removal and sharding token generation has been moved outside.

What's your take now?

bboreham:
Since we've managed to re-shard the world at least twice in the past year, would it be worth having a unit test that fails if it happens again?

pracucci:
> Since we've managed to re-shard the world at least twice in the past year, would it be worth having a unit test that fails if it happens again?

Agree @bboreham. I've added a test hardcoding some expected tokens, given we don't want them to change over the time. What's about now?

PR #1889, opened 2019-12-06
Lower log level on ingester transfer failed because no pending ingester is found

pracucci:
**What this PR does**:
While rolling out ingesters it's usual to get a train of errors `no pending ingesters` while the leaving ingester has started the shutdown procedure but there's still no joining ingester. This is not really an issue, as far as the joining ingester is found before the end of all retries.

In this PR I'm suggesting to lower the log level to Warn while retrying, while logging it as error if persists once all retries have been done.

_No issue_

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/transfer.go
+++ b/pkg/ingester/transfer.go
@@ -3,7 +3,6 @@ package ingester
 import (
 	"bytes"
 	"context"
-	go_errors "errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -12,7 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/ingester/transfer.go
```diff
@@ -377,16 +381,30 @@ func (i *Ingester) TransferOut(ctx context.Context) error {
...
+		err = i.transferOut(ctx)
 		if err == nil {
 			return nil
 		}
 
-		level.Error(util.Logger).Log("msg", "transfer failed", "err", err)
+		// Log just with an info level errors which are expected to happen
+		var lvl log.Logger
+		if go_errors.Is(err, errTransferNoPendingIngesters) {
+			lvl = level.Info(util.Logger)
```
pstibrany: Let's just log them all as warnings, and last one as error? This code seems to be too smart.

pracucci: Probably. What's about now?

pstibrany: I think that's fine. Thanks.

PR #1890, opened 2019-12-06
Explicit healty states

pstibrany:
**What this PR does**:
Inspired by PR #1866, this PR changes `IsHealthy` function to be explicit about which states are healthy. There is no change in functionality.

There is no bug.

**Checklist**
- [x] Tests updated
- [ ] No documentation changes
- [ ] `CHANGELOG.md` updated - no functionality change

### DIFF
--- a/pkg/ring/model.go
+++ b/pkg/ring/model.go
@@ -182,12 +182,17 @@ func (d *Desc) TokensFor(id string) (tokens, other []uint32) {
 
 // IsHealthy checks whether the ingester appears to be alive and heartbeating
 func (i *IngesterDesc) IsHealthy(op Operation, heartbeatTimeout time.Duration) bool {
-	if op == Write && i.State != ACTIVE {
-		return false
-	} else if op == Read && i.State == JOINING {
-		return false
+	healthy := false
+
+	if op == Write && i.State == ACTIVE {
+		healthy = true
 	}
-	return time.Now().Sub(time.Unix(i.Timestamp, 0)) <= heartbeatTimeout
+
+	if op == Read && (i.State == ACTIVE || i.State == LEAVING || i.State == PENDING) {
+		healthy = true
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/ring/model.go
```diff
@@ -182,12 +182,20 @@ func (d *Desc) TokensFor(id string) (tokens, other []uint32) {
...
+
+	switch op {
+	case Write:
+		healthy = i.State == ACTIVE
+
+	case Read:
+		healthy = i.State == ACTIVE || i.State == LEAVING || i.State == PENDING
+
+	case Reporting:
+		healthy = true
```
codesome: If we want 1:1 mapping of the functionality, then this should be `default` case.

pstibrany: Then it would never be set to false.

pstibrany: Ah, sorry, forget my previous comment. It would of course. But there is no other `op`, so having Reporting case here covers all cases.

codesome: How? It would fall back to `default` only if the `op` was not `Write` or `Read`.

codesome: Yeah, it looks fine. There is no hard requirement from Go that it should be only those 3 values, but for our use case it should be fine I guess.

pkg/ring/model.go
```diff
@@ -182,12 +182,20 @@ func (d *Desc) TokensFor(id string) (tokens, other []uint32) {
...
 func (i *IngesterDesc) IsHealthy(op Operation, heartbeatTimeout time.Duration) bool {
-	if op == Write && i.State != ACTIVE {
-		return false
-	} else if op == Read && i.State == JOINING {
-		return false
+	healthy := false
+
+	switch op {
+	case Write:
+		healthy = i.State == ACTIVE
```
gouthamve: Can we have brackets here `(i.State == ACTIVE)` and more important for the next condition.

We had a big outage sometime ago with a `op` b `op` c and since then I want to use brackets if we're doing things like that :)

PR #1892, opened 2019-12-09
Pass Request to DecodeResponse in case it is required.

cyriltovena:
Loki has the concept of log Direction which is essential to know when
merging responses. This change allows Loki to inject the direction in the
response which can be then used when merging multiple responses.

Cortex doesn't need to use this.

### DIFF
--- a/pkg/querier/queryrange/query_range.go
+++ b/pkg/querier/queryrange/query_range.go
@@ -40,7 +40,9 @@ type Codec interface {
 	Merger
 	// DecodeRequest decodes a Request from an http request.
 	DecodeRequest(context.Context, *http.Request) (Request, error)
-	// DecodeResponse decodes a Response from an http response..
+	// DecodeResponse decodes a Response from an http response.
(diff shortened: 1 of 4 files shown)


pkg/querier/queryrange/query_range.go
```diff
@@ -198,7 +198,7 @@ func (prometheusCodec) EncodeRequest(ctx context.Context, r Request) (*http.Requ
 	return req.WithContext(ctx), nil
 }
 
-func (prometheusCodec) DecodeResponse(ctx context.Context, r *http.Response) (Response, error) {
+func (prometheusCodec) DecodeResponse(ctx context.Context, r *http.Response, _ Request) (Response, error) {
```
pracucci: I would suggest to add a comment to this function, explaining why we have `_ Request`, otherwise there's the risk we'll clean it up in the future because considered dead code.

cyriltovena: I added a note, but on the interface, I don't here it matters.

PR #1893, opened 2019-12-09
querier: option to skip sending queries to long term storage

thorfour:
**What this PR does**: Moves the MinChunkAge option from chunk store into the querier to allow that behaviour regardless of store backend.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## master / unreleased
 
 * [CHANGE] Removed remaining support for using denormalised tokens in the ring. If you're still running ingesters with denormalised tokens (Cortex 0.4 or earlier, with `-ingester.normalise-tokens=false`), such ingesters will now be completely invisible to distributors and need to be either switched to Cortex 0.6.0 or later, or be configured to use normalised tokens. #2034
+* [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893 
+  * `--store.min-chunk-age` has been removed
+  * `--querier.query-store-after` has been added in it's place.
 * [ENHANCEMENT] Experimental TSDB: Export TSDB Syncer metrics from Compactor component, they are prefixed with `cortex_compactor_`. #2023
 * [ENHANCEMENT] Experimental TSDB: Added dedicated flag `-experimental.tsdb.bucket-store.tenant-sync-concurrency` to configure the maximum number of concurrent tenants for which blocks are synched. #2026
 * [ENHANCEMENT] Experimental TSDB: Expose metrics for objstore operations (prefixed with `cortex_<component>_thanos_objstore_`, component being one of `ingester`, `querier` and `compactor`). #2027
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -2,6 +2,7 @@ package querier
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"time"
 
@@ -19,13 +20,16 @@ import (
 
 // Config contains the configuration require to create a querier
 type Config struct {
-	MaxConcurrent            int
-	Timeout                  time.Duration
-	Iterators                bool
-	BatchIterators           bool
-	IngesterStreaming        bool
-	MaxSamples               int
-	IngesterMaxQueryLookback time.Duration
+	MaxConcurrent        int
+	Timeout              time.Duration
+	Iterators            bool
+	BatchIterators       bool
+	IngesterStreaming    bool
+	MaxSamples           int
+	QueryIngestersWithin time.Duration `yaml:"query_ingesters_within"`
+
+	// QueryStoreAfter the time after which queries should also be sent to the store and not just ingesters.
+	QueryStoreAfter time.Duration `yaml:"query_store_after"`
 
 	// The default evaluation interval for the promql engine.
 	// Needs to be configured for subqueries to work as it is the default
@@ -35,6 +39,10 @@ type Config struct {
 	// For testing, to prevent re-registration of metrics in the promql engine.
 	metricsRegisterer prometheus.Registerer `yaml:"-"`
 }
+
+var (
+	errBadLookbackConfigs = errors.New("bad settings, query_store_after >= query_ingesters_within which can result in queries not being sent")
+)
 
 // RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
@@ -47,9 +55,23 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.BatchIterators, "querier.batch-iterators", false, "Use batch iterators to execute query, as opposed to fully materialising the series in memory.  Takes precedent over the -querier.iterators flag.")
 	f.BoolVar(&cfg.IngesterStreaming, "querier.ingester-streaming", false, "Use streaming RPCs to query ingester.")
 	f.IntVar(&cfg.MaxSamples, "querier.max-samples", 50e6, "Maximum number of samples a single query can load into memory.")
-	f.DurationVar(&cfg.IngesterMaxQueryLookback, "querier.query-ingesters-within", 0, "Maximum lookback beyond which queries are not sent to ingester. 0 means all queries are sent to ingester.")
+	f.DurationVar(&cfg.QueryIngestersWithin, "querier.query-ingesters-within", 0, "Maximum lookback beyond which queries are not sent to ingester. 0 means all queries are sent to ingester.")
 	f.DurationVar(&cfg.DefaultEvaluationInterval, "querier.default-evaluation-interval", time.Minute, "The default evaluation interval or step size for subqueries.")
+	f.DurationVar(&cfg.QueryStoreAfter, "querier.query-store-after", 0, "The time after which a metric should only be queried from storage and not just ingesters. 0 means all queries are sent to store.")
 	cfg.metricsRegisterer = prometheus.DefaultRegisterer
+}
+
+// Validate the config
+func (cfg *Config) Validate() error {
+
+	// Ensure the config wont create a situation where no queriers are returned.
+	if cfg.QueryIngestersWithin != 0 && cfg.QueryStoreAfter != 0 {
+		if cfg.QueryStoreAfter >= cfg.QueryIngestersWithin {
+			return errBadLookbackConfigs
+		}
+	}
+
+	return nil
 }
 
 // ChunkStore is the read-interface to the Chunk Store.  Made an interface here
@@ -70,11 +92,11 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
 	var queryable storage.Queryable
 	if cfg.IngesterStreaming {
 		dq := newIngesterStreamingQueryable(distributor, iteratorFunc)
-		queryable = newUnifiedChunkQueryable(dq, chunkStore, distributor, iteratorFunc, cfg.IngesterMaxQueryLookback)
+		queryable = newUnifiedChunkQueryable(dq, chunkStore, distributor, iteratorFunc, cfg)
 	} else {
 		cq := newChunkStoreQueryable(chunkStore, iteratorFunc)
 		dq := newDistributorQueryable(distributor)
-		queryable = NewQueryable(dq, cq, distributor, cfg.IngesterMaxQueryLookback)
+		queryable = NewQueryable(dq, cq, distributor, cfg)
 	}
 
 	lazyQueryable := storage.QueryableFunc(func(ctx context.Context, mint int64, maxt int64) (storage.Querier, error) {
@@ -97,28 +119,33 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
 }
 
 // NewQueryable creates a new Queryable for cortex.
-func NewQueryable(dq, cq storage.Queryable, distributor Distributor, ingesterMaxQueryLookback time.Duration) storage.Queryable {
+func NewQueryable(dq, cq storage.Queryable, distributor Distributor, cfg Config) storage.Queryable {
 	return storage.QueryableFunc(func(ctx context.Context, mint, maxt int64) (storage.Querier, error) {
-		cqr, err := cq.Querier(ctx, mint, maxt)
-		if err != nil {
-			return nil, err
-		}
-
 		q := querier{
-			queriers:    []storage.Querier{cqr},
 			distributor: distributor,
 			ctx:         ctx,
 			mint:        mint,
 			maxt:        maxt,
 		}
 
-		// Include ingester only if maxt is within ingesterMaxQueryLookback w.r.t. current time.
-		if ingesterMaxQueryLookback == 0 || maxt >= time.Now().Add(-ingesterMaxQueryLookback).UnixNano()/1e6 {
+		// Include ingester only if maxt is within queryIngestersWithin w.r.t. current time.
+		now := model.Now()
+		if cfg.QueryIngestersWithin == 0 || maxt >= int64(now.Add(-cfg.QueryIngestersWithin)) {
 			dqr, err := dq.Querier(ctx, mint, maxt)
 			if err != nil {
 				return nil, err
 			}
 			q.queriers = append(q.queriers, dqr)
+		}
+
+		// Include store only if mint is within QueryStoreAfter w.r.t current time.
+		if cfg.QueryStoreAfter == 0 || mint <= int64(now.Add(-cfg.QueryStoreAfter)) {
+			cqr, err := cq.Querier(ctx, mint, maxt)
+			if err != nil {
+				return nil, err
+			}
+
+			q.queriers = append(q.queriers, cqr)
 		}
 
 		return q, nil
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -508,7 +508,12 @@ The `querier_config` configures the Cortex querier.
 # Maximum lookback beyond which queries are not sent to ingester. 0 means all
 # queries are sent to ingester.
 # CLI flag: -querier.query-ingesters-within
-[ingestermaxquerylookback: <duration> | default = 0s]
+[query_ingesters_within: <duration> | default = 0s]
+
+# The time after which a metric should only be queried from storage and not just
+# ingesters. 0 means all queries are sent to store.
+# CLI flag: -querier.query-store-after
+[query_store_after: <duration> | default = 0s]
 
 # The default evaluation interval or step size for subqueries.
 # CLI flag: -querier.default-evaluation-interval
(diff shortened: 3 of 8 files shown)


pkg/querier/querier.go
```diff
@@ -27,6 +27,9 @@ type Config struct {
 	MaxSamples               int
 	IngesterMaxQueryLookback time.Duration
 
+	// IngesterOnlyQueriesAllowed sends queries that fall within IngesterMaxQueryLookback to only ingesters and skips LTS
+	IngesterOnlyQueriesAllowed bool
```
jtlisi: Where is this configured? Are you going to add flag?

thorfour: It would just be apart of the yaml config. I guess I should add the yaml annotation for clarity. I normally would have added a flag but I think the current POR is to deprecate the usage of flags for configs.

bboreham:
The chunk store already has this feature here: 
https://github.com/cortexproject/cortex/blob/fcbc186d495aea15650d353775677b787d53863c/pkg/chunk/chunk_store.go#L272-L275

so you shouldn't need to add it at the querier level.

thorfour:
> The chunk store already has this feature here

That's true, but I would then need to add the feature to the block querier, it made sense to have it as a top level feature instead of backend specific.

Is it preferred to be implemented in each backend or on the top level @bboreham?

bboreham:
I guess you can have a go at _moving_ the feature, but don't add something that duplicates it.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
   * `ruler.storage.type` has beem added to specify the rule store backend type, currently only the configdb.
   * `ruler.poll-interval` has been added to specify the interval in which to poll new rule groups.
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
+* [FEATURE] The Querier can now send queries to Ingesters only if the query fits within MaxLookback
```
pracucci: A couple of things:
1. Please add the PR number
2. `MaxLookback` looks unclear to me. We should mention the specific config option
3. I would also mention the new flag/config option introduced to enable it

pkg/querier/querier.go
pracucci: Given @bboreham mentioned we already have such logic in the `chunk_store.go` and he agrees on moving the check in the querier, do we still need this option or we can just keep it enabled?

If we keep the option, I would be glad if you could add the CLI flag too. At Grafana we're not ready to switch to config files, while I would like to use this flag too.

pkg/querier/querier.go
```diff
@@ -97,30 +100,36 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
 		// Include ingester only if maxt is within ingesterMaxQueryLookback w.r.t. current time.
-		if ingesterMaxQueryLookback == 0 || maxt >= time.Now().Add(-ingesterMaxQueryLookback).UnixNano()/1e6 {
+		if cfg.IngesterMaxQueryLookback == 0 || maxt >= time.Now().Add(-cfg.IngesterMaxQueryLookback).UnixNano()/1e6 {
 			dqr, err := dq.Querier(ctx, mint, maxt)
 			if err != nil {
 				return nil, err
 			}
 			q.queriers = append(q.queriers, dqr)
+
+			// query fits within IngesterMaxQueryLookback we can skip sending queries to LTS
```
pracucci: I guess I'm missing something. How can you assert the query fits within `IngesterMaxQueryLookback`? We have just checked `maxt` but not `mint`.

_Same comment applies to `newUnifiedChunkQueryable()`_

thorfour: Fixed. I've moved the MinChunkAge logic into the querier as IngesterMetricMinAge

CHANGELOG.md
thorfour: Fixed

pkg/querier/querier.go
thorfour: I've moved the previous config, and added a flag option.

pkg/querier/querier.go
```diff
@@ -27,6 +27,9 @@ type Config struct {
 	MaxSamples               int
 	IngesterMaxQueryLookback time.Duration
 
+	// IngesterMetricMinAge minimum time between a metric being ingested and saved to the store.
+	IngesterMetricMinAge time.Duration `yaml:"ingester_metric_min_age"`
```
pracucci: I find this config option name difficult to understand (ie. it's called "min age" but looks more a "max age for queries to be sent to ingesters only". Do you have any better option for the name?

pracucci: How does this option relates to ingester's `MaxChunkIdle` and `MaxStaleChunkIdle`? The problem I see is that if this option is > than `min(MaxChunkIdle, MaxStaleChunkIdle)` then you may end up in weird bugs hiding series which have been flushed to the storage.

If I'm not missing anything, we shouldn't mine Cortex correctness in any way and we should protect from it.

thorfour: Yea I tried to keep it similar to the option I moved, `MinChunkAge` so it refers to the minimum age of a metric before sending to the store. But I don't really like the name either and would love a better idea.

thorfour: I don't think it changes that relation at all right? If a user had set `MinChunkAge` to be `1h` but set `MaxChunkIdle` to `1s` it could leave a gap during query time. I'm not sure it's a bug but a known configuration pattern.

pracucci: It looks a bug to me (given it's a quite tricky thing even to correctly configure), but I would be happy to be contradicted. @gouthamve @bboreham any thought?

gouthamve: I think it's a combination of `min(MaxChunkIdle, MaxStaleChunkIdle)` + `-ingester.retain-period`. As long as we have the retain period, we don't need to hit the store. Though I agree its hard to configure and get right.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
 * [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored. If you want to switch back to using denormalised tokens, you need to downgrade to Cortex 0.4.0. Previous versions don't handle claiming tokens from normalised ingesters correctly. #1809
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
+* [FEATURE] Moves MinChunkAge to the Querier config as IngesterMetricMinAge, allowing the store to be skipped during query time if the metrics wouldn't be found.#1893
```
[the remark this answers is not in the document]
thorfour: Moved it to CHANGE.

thorfour:
@gouthamve Do you have a better name for the flag in mind? `min-ingester-lookback-time`? 
`skip-store-before-time` ?

pkg/querier/querier.go
```diff
@@ -97,30 +101,37 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
+		now := model.Now()
+		if cfg.IngesterMaxQueryLookback == 0 || maxt >= int64(now.Add(-cfg.IngesterMaxQueryLookback)) {
 			dqr, err := dq.Querier(ctx, mint, maxt)
 			if err != nil {
 				return nil, err
 			}
 			q.queriers = append(q.queriers, dqr)
+
+			// metrics wont have been sent to store yet, skip store query
+			if model.Time(mint).After(now.Add(-cfg.IngesterMetricMinAge)) {
```
pracucci: I think having this `if` here makes the code a bit difficult to read, because the parent `if` block is used to determine if a query should hit the ingesters, while this `if` block is used to determine if a query should hit the store.

Instead of having a guard here, I would suggest to wrap the store querier within a `if`, to keep the intention more explicit.

_Same comment applies to `pkg/querier/unified_querier.go`_

pkg/querier/querier.go
pracucci: To me, this new option is similar to `IngesterMaxQueryLookback`, in the sense it does the same thing but applied to the store. A better naming may be `store_min_query_lookback` (`-querier.store-min-query-lookback`), in the sense it's the minimum query lookback time required to hit the storage.

thorfour: I like that name much better. Changed.

pkg/querier/querier.go
thorfour: Good call, I like that better too.

thorfour:
Setting has been renamed to `StoreMinQueryLookback`

pkg/querier/querier.go
```diff
@@ -97,30 +101,35 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
+		if cfg.IngesterMaxQueryLookback == 0 || maxt >= int64(now.Add(-cfg.IngesterMaxQueryLookback)) {
 			dqr, err := dq.Querier(ctx, mint, maxt)
 			if err != nil {
 				return nil, err
 			}
 			q.queriers = append(q.queriers, dqr)
 		}
 
+		// Include store only if mint is within StoreMinQueryLookback w.r.t current time.
+		if cfg.StoreMinQueryLookback == 0 || model.Time(mint).Before(now.Add(-cfg.StoreMinQueryLookback)) {
```
pracucci: `Before()` is `<`, while I think `<=` would be safer. Maybe we can just use `mint <= int64(now.Add(-cfg.StoreMinQueryLookback))` like you did for the `IngesterMaxQueryLookback `?

_If agree, same comment applies to `pkg/querier/unified_querier.go` too_

pkg/querier/querier.go
```diff
@@ -97,30 +101,35 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
+			cqr, err := cq.Querier(ctx, mint, maxt)
+			if err != nil {
+				return nil, err
+			}
+
+			q.queriers = append(q.queriers, cqr)
+		}
+
```
pracucci: Ideally we should make sure it's not possible to reach this point with `len(q.queriers) == 0`. We could error out if there are no queriers, but it would be probably safer (and easier for the user) to add a `Config.Validate()` func to make sure `StoreMinQueryLookback` is never greater than `IngesterMaxQueryLookback`, when both are enabled (non zero).

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
 * [CHANGE] Use relative links from /ring page to make it work when used behind reverse proxy. #1896
 * [CHANGE] Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored. If you want to switch back to using denormalised tokens, you need to downgrade to Cortex 0.4.0. Previous versions don't handle claiming tokens from normalised ingesters correctly. #1809
+* [CHANGE] Moves MinChunkAge to the Querier config as StoreMinQueryLookback, allowing the store to be skipped during query time if the metrics wouldn't be found.#1893
```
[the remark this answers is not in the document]
thorfour: Fixed.

pkg/querier/querier.go
thorfour: I don't have a strong preference, so I changed it to as you've suggested.

pkg/querier/querier.go
thorfour: Added a validate function and have it called it in `cortex.go`

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
 * [CHANGE] Use relative links from /ring page to make it work when used behind reverse proxy. #1896
 * [CHANGE] Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [CHANGE] Ingesters now write only normalised tokens to the ring, although they can still read denormalised tokens used by other ingesters. `-ingester.normalise-tokens` is now deprecated, and ignored. If you want to switch back to using denormalised tokens, you need to downgrade to Cortex 0.4.0. Previous versions don't handle claiming tokens from normalised ingesters correctly. #1809
+* [CHANGE] Moved MinChunkAge to the Querier config as StoreMinQueryLookback, allowing the store to be skipped during query time if the metrics wouldn't be found.#1893 
+  * `store.min-chunk-age` has been removed
+  * `querier.store-min-query-lookback` has been added in it's place.
```
gouthamve: This is again opaque tbh. `IngesterMaxQueryLookback` has the flag `querier.query-ingesters-within`, maybe renaming it to `querier.query-store-after` is better?

thorfour: @pracucci thoughts?

pracucci: > IngesterMaxQueryLookback has the flag querier.query-ingesters-within

I didn't notice the CLI flag for `IngesterMaxQueryLookback` was different. YAML config file name and CLI flags should always match, otherwise we introduce yet another inconsistency. If we wanna keep the `-querier.query-ingesters-within` naming, than I would:

1. Rename `IngesterMaxQueryLookback` to `QueryIngestersWithin` and set its YAML config name to `query_ingesters_within`
2. Then I agree with Goutham about naming it `-querier.query-store-after` and - as a consequence - name the YAML `query_store_after`

thorfour: Done!

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
...
 * [CHANGE] Overrides mechanism has been renamed to "runtime config", and is now separate from limits. Runtime config is simply a file that is reloaded by Cortex every couple of seconds. Limits and now also multi KV use this mechanism.<br />New arguments were introduced: `-runtime-config.file` (defaults to empty) and `-runtime-config.reload-period` (defaults to 10 seconds), which replace previously used `-limits.per-user-override-config` and `-limits.per-user-override-period` options. Old options are still used if `-runtime-config.file` is not specified. This change is also reflected in YAML configuration, where old `limits.per_tenant_override_config` and `limits.per_tenant_override_period` fields are replaced with `runtime_config.file` and `runtime_config.period` respectively. #1749
+* [CHANGE] Moved MinChunkAge to the Querier config as QueryStoreAfter, allowing the store to be skipped during query time if the metrics wouldn't be found.#1893
```
pracucci: `MinChunkAge` and `QueryStoreAfter` are internal names and not the names of the config options. I would replace them with the CLI flags, to make them easier to understand for the final user.

Please also mention that the `ingestermaxquerylookback` YAML option has been renamed to `query_ingesters_within` (in case anyone is configuring it through YAML).

pkg/querier/querier.go
```diff
@@ -19,13 +20,16 @@ import (
...
+	MaxConcurrent        int
+	Timeout              time.Duration
+	Iterators            bool
+	BatchIterators       bool
+	IngesterStreaming    bool
+	MaxSamples           int
+	QueryIngestersWithin time.Duration `yaml:"query_ingesters_within"`
+
+	// QueryStoreAfter the time after which queries should also be sent to the store and not just ingesters.
+	QueryStoreAfter time.Duration `yaml:"store_min_query_lookback"`
```
pracucci: `store_min_query_lookback` > `query_store_after`

pkg/querier/querier.go
```diff
@@ -36,6 +40,10 @@ type Config struct {
 	metricsRegisterer prometheus.Registerer
 }
 
+var (
+	errBadLookbackConfigs = errors.New("bad settings, QueryStoreAfter >= QueryIngestersWithin which can result in queries not being sent")
```
pracucci: We should also replace `QueryStoreAfter` and `QueryIngestersWithin` with the user-configurable options here as well.

CHANGELOG.md
thorfour: done.

pkg/querier/querier.go
thorfour: done.

pkg/querier/querier.go
thorfour: done.

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ instructions below to upgrade your Postgres.
...
 * [CHANGE] Cortex now rejects data with duplicate labels. Previously, such data was accepted, with duplicate labels removed with only one value left. #1964
 * [CHANGE] Changed the default value for `-distributor.ha-tracker.prefix` from `collectors/` to `ha-tracker/` in order to not clash with other keys (ie. ring) stored in the same key-value store. #1940
+* [CHANGE] Moved `store-min-chunk-age` to the Querier config as `querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. IngesterMaxQueryLookback as been renamed to QueryIngestersWithin to make it match its flag and yaml settings.#1893
```
pracucci: > IngesterMaxQueryLookback as been renamed to QueryIngestersWithin to make it match its flag and yaml settings

This is an important information for the user because it changes the YAML config option, so I would say something like:

```
The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag.
```

pracucci: `store-min-chunk-age` > `store.min-chunk-age` and please prefix flags with `-` to make it more clear they're CLI flags.

pkg/querier/querier.go
```diff
@@ -19,13 +20,16 @@ import (
...
-	IngesterStreaming        bool
-	MaxSamples               int
-	IngesterMaxQueryLookback time.Duration
+	MaxConcurrent        int
+	Timeout              time.Duration
+	Iterators            bool
+	BatchIterators       bool
+	IngesterStreaming    bool
+	MaxSamples           int
+	QueryIngestersWithin time.Duration `yaml:"query_ingesters_within"`
```
pracucci: Maybe `QueryIngestersBefore` is more clear? /cc @gouthamve

CHANGELOG.md
thorfour: Fixed.

thorfour: Changed.

pkg/querier/querier.go
jtlisi: `QueryIngestersWithin` seems reasonable to me

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ instructions below to upgrade your Postgres.
...
 * [CHANGE] Cortex now rejects data with duplicate labels. Previously, such data was accepted, with duplicate labels removed with only one value left. #1964
 * [CHANGE] Changed the default value for `-distributor.ha-tracker.prefix` from `collectors/` to `ha-tracker/` in order to not clash with other keys (ie. ring) stored in the same key-value store. #1940
+* [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingester_max_query_lookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893
```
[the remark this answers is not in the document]
thorfour: Changed.

pracucci:
@thorfour Sorry to bother you. We've merge the release `0.6.0` and you would need to move your `CHANGELOG.md` entry to the top of the file (remember to keep entries grouped by type).

thorfour:
Yea forgot we had cut 0.6.0 and didn't really look at the merge conflicts when I rebased it.

Issue #778, opened 2018-03-27
Support reverse proxy in admin http pages


bboreham:
The `/ring` page (maybe others too) does a redirect which will break if you put it behind a reverse proxy that changes the path.

Prometheus has a `--web.external-url` option to address this.

pstibrany:
What places need to be fixed, apart from /ring? Perhaps we can do it without --web.external-url ? (By using relative paths everywhere)

bboreham:
I don't know, nor did I ever understand what the "PRG pattern" is that needed the redirect.

pstibrany:
> I don't know, nor did I ever understand what the "PRG pattern" is that needed the redirect.

PRG is used to make REFRESH in the browser work correctly after POST request. Without this, refreshing the page would do another POST and browser would ask the user if he really wants to do the refresh, which is annoying. With redirect to GET, refreshing the page simply repeats the GET.

I don't quite see how it is relevant to cross-site request forgery (CSRF) attack though.

gouthamve:
We looked at this in our bug scrub and decided that it was closed by #1896

pracucci:
Closed by #1896

PR #1896, opened 2019-12-10
Fix redirects on /ring page by using relative targets.

pstibrany:
**What this PR does**: This PR changes redirects from /ring page to use relative links, instead of path-absolute. This is useful when running distributor HTTP server behind reverse proxy, which maps it to a different path than root.

**Checklist**
- [ ] Tests updated - nope, tested manually
- [ ] Documentation added - nope


bboreham:
Fixes #778, at least in part.

Issue #1852, opened 2019-11-23
Memory leak in ingester after it has failed to write indixes on memcache


weeco:
**The problem:**

Even though I run cortex v0.3.0 for a couple weeks already I face OOM killed ingester replicas more frequently in the recent time (roughly 1-2 times every day, which is quite a lot given that I only run 5 big ingesters).

Pod RAM request/limit in k8s: 40GB
Average RAM usage: 14-18GB
First OOM kill at: 11:29:45am
Second OOM kill at: 11:32:15am

**Incident details (RAM usage, log messages etc.)**

1. Go Heap allocated bytes before it gets OOM killed (in this screenshot you see two gaps because it got OOM killed twice):

![ram_cortex](https://user-images.githubusercontent.com/23424570/69482892-3b4f8c00-0e21-11ea-80fa-2e7dc5bd7695.png)

2. Number of Go routines:

![go_routines](https://user-images.githubusercontent.com/23424570/69482925-a4370400-0e21-11ea-9fec-1bc7f820fa4d.png)

3. The number of log messages (seem to correlate with the restarts)

![log_messages](https://user-images.githubusercontent.com/23424570/69482936-ca5ca400-0e21-11ea-87ec-5d765bda5ec9.png)

**Log messages:**

Since there is an unusual amount of log messages, I was interested what kind of log messages these were. I had to filter the noisy messages anyways to get to the actually interesting log messages:

4k of the 22k log messages were "context cancalled" (QueryStream error)
3k messsages because of "fingerprint collision detected, mapping to new fingerprint"
15k messages because of "sample timestamp out of order" (which was likely caused by this faulty ingester)

After filtering out these messages I had 31 messages left:

![kibana-logs](https://user-images.githubusercontent.com/23424570/69483155-30e2c180-0e24-11ea-96fd-2efe4cf62c7b.png)

**In short:**

Every time I see these ingester OOM kills it seems to start with memcached timeouts. In this case it couldn't put it's indexes on the "index-cache-write" memcache because it timedout. Afterwards the memory seems to leak very quickly, which makes it hard to provide a heapdump. If the information provided in this issue is not sufficient to track the leak, I'd be happy to provide whatever is needed to figure it out - please let me know what else I can do.

![kibana-failed-memcached](https://user-images.githubusercontent.com/23424570/69483194-83bc7900-0e24-11ea-9ee4-312085179441.png)

bboreham:
I have seen similar symptoms, and I don't know the cause.

[...]

PR #1897, opened 2019-12-10
Ensure metric name is not pointing into gRPC buffer

bboreham:
This relates to #1852 - not sure if it's the only problem but I think this will help.

The chain of ownership:
Strings are pulled directly from the gRPC buffer without copying: https://github.com/cortexproject/cortex/blob/fcbc186d495aea15650d353775677b787d53863c/pkg/ingester/client/timeseries.go#L188

This value is pulled out by `MetricNameFromLabelAdapters()`:
https://github.com/cortexproject/cortex/blob/fcbc186d495aea15650d353775677b787d53863c/pkg/util/extract/extract.go#L15

`metricName` is passed down: https://github.com/cortexproject/cortex/blob/fcbc186d495aea15650d353775677b787d53863c/pkg/ingester/user_state.go#L209

and used as a key into a map:
https://github.com/cortexproject/cortex/blob/fcbc186d495aea15650d353775677b787d53863c/pkg/ingester/user_state.go#L236

### DIFF
--- a/pkg/util/extract/extract.go
+++ b/pkg/util/extract/extract.go
@@ -12,7 +12,9 @@ import (
 func MetricNameFromLabelAdapters(labels []client.LabelAdapter) (string, error) {
 	for _, label := range labels {
 		if label.Name == model.MetricNameLabel {
-			return label.Value, nil
+			// Force a string copy since LabelAdapter is often a pointer into
+			// a large gRPC buffer which we don't want to keep alive on the heap.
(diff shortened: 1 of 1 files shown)


bboreham:
Having inspected all the other places where `MetricNameFromLabelAdapters()` is called, the copy is unnecessary.

However, given the damage from the one place it was necessary, and the time it took to find, I'm ok with a bit more waste until we can see that it causes a problem.  I'm open to argument on this point though.

Issue #1895, opened 2019-12-10
Distributors OOM on a single slow ingester in the cluster


pracucci:
Yesterday we've got all distributors continuously `OOMKilled` in one of our Cortex clusters. The root cause analysis outlined this issue has been caused by a single ingester which was running on a failing Kubernetes node which was running but very slow.

This issue is due to how the quorum works. When the distributors receive a `Push()` request, the time series are sharded and then sent to 3 ingesters (we have a replication factor of `3`). The distributor's `Push()` request completes as soon as all series are pushed to at least 2 ingesters.

In the case of a very slow ingester, the distributor piles up the number of in-flight requests towards the slow ingester, while the inbound `Push()` request is completed as soon as the other ingesters successfully complete the ingestion.

This causes the memory used by the distributors to increase due to the in-flight requests towards the slow ingester.

In a high traffic Cortex cluster, distributors can hit the memory limit before the timeout of the in-flight requests towards the slow ingester is expired, causing all distributors to be `OOMKilled` (and subsequent distributors restarts will OOM again until the very slow ingester is not removed from the ring).

tomwilkie:
I think this might be caused by https://github.com/cortexproject/cortex/pull/736 - we used to cancel the outstanding request, now we pile them up as you said.

bboreham:
#858 talks about a similar situation - we need to limit the number of "backgrounded" requests.

It sounds like a smaller timeout would help.

#736 was done for good reasons, and is essential to the efficiency gain from #1578.

pstibrany:
> #736 was done for good reasons, and is essential to the efficiency gain from #1578.

Can you please elaborate on how #1578 is related to #736? Is it to make sure that each ingesters gets exactly the same data, and not only part of it due to parent context being cancelled/timed out?

bboreham:
Yes, if you cancel the 3rd push every time then each ingester will have a random sprinkling of holes in the data, so the checksums won't match.

bboreham:
BTW I just added a link to a blog post in #1578 that describes the efficiency gains.

pstibrany:
> BTW I just added a link to a blog post in #1578 that describes the efficiency gains.

Thanks. I just wanted to make sure I understand it correctly, as I was just adding similar thing to Loki earlier today and hope to see similar benefits. (Loki already uses #736 change, so all is good there).

weeco:
Today we ran into the same issue which caused an outage of the write path in our prod environment. 

[...]

PR #1898, opened 2019-12-10
Fix reuse slice in distributor.Push()

pracucci:
**What this PR does**:
While investigating the issue #1895, I've realized the request slice is not reused in case the request gets skipped because of HA dedup or rate limiting. In case all incoming requests have the HA enabled, it's 50% of requests. This PR fixes it.

_No issue_

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci:
> I am wondering if we should do it in a defer statement if there was an error?

@codesome In case of error it's OK doing it before `return`, so I don't think there's any real value paying the cost of a `defer`. If you mean adding a `defer` at the beginning of the function, to cover all cases at once, unfortunately we can't do it because of the async nature of `ring.DoBatch()`: when the execution reaches that point, the `req` needs to be alive even after the `distributor.Push()` completes.

PR #1802, opened 2019-11-08
Upgrade golang-migrate/migrate to v4 (#1384)

adnxn:
**What this PR does**:
changes `mattes/migrate` path to `golang-migrate/migrate`, upgrades to [v4.7.0](https://github.com/golang-migrate/migrate/releases/tag/v4.7.0), and corresponding changes to migrate api use in `pkg/configs/db/postgres/postgres.go`

ran `go mod vendor` and `go mod tidy`

**Which issue(s) this PR fixes**:
Fixes #1384

PR #1899, opened 2019-12-10
prefix migrations dir to URI

jtlisi:
**What this PR does**:

Ensures the proper file URI scheme for the provided migrations directory.

Fixes https://github.com/cortexproject/cortex/pull/1802#issuecomment-562556715

### DIFF
--- a/pkg/configs/db/postgres/postgres.go
+++ b/pkg/configs/db/postgres/postgres.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Masterminds/squirrel"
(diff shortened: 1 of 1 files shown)


pkg/configs/db/postgres/postgres.go
```diff
@@ -75,6 +76,13 @@ func New(uri, migrationsDir string) (DB, error) {
 	}
 
 	if migrationsDir != "" {
+		i := strings.Index(migrationsDir, ":")
+
+		// Add file scheme if no scheme is present
+		if i < 1 {
```
bboreham: `!strings.Contains()` is shorter

jtlisi: I ended up using `strings.HasPrefix`

PR #1900, opened 2019-12-10
Doing an initial synch and then sleep to space out subsequent cycles.…

### LINKED ISSUE #1618: Tablemanager sleep at startup could confuse new adopters
The sleep at startup added in #1587 will mean that people who download Cortex, fire it up on default settings and do a query within the first two minutes could get an error message that the table doesn't exist.

I guess this is a bit of a rare scenario, but it might be better to do an initial sync then sleep to space out subsequent cycles.

tcl00taco:
**What this PR does**:
Adding sleep after initial SynchTable operation to spread out subsequent synch.
Fixes #1618 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -167,9 +167,6 @@ func (m *TableManager) Stop() {
 func (m *TableManager) loop() {
 	defer m.wait.Done()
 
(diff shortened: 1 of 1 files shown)


gouthamve:
Fixes #1618

PR #1906, opened 2019-12-12
Delete series 

sandeepsukhani:
**What this PR does**:
Adds delete series support. Also supports cancelling the delete requests which are not picked up for processing yet. It does query time filtering of data to let users make an informed decision about continuing with delete request or cancelling it.
Since we support caching of results, index and chunks, this PR also takes care of cache invalidation using generation number. All the cache keys will be suffixed by a generation number, which would be changed over the lifetime of delete requests.

*Implementation details*
* New entities 
  * `DeleteStore` - Handles storing and fetching delete requests, delete plans, cache gen numbers
  * `TombstonesLoader` - Helps loading delete requests and cache gen. Also keeps checking for updates.
  * `DataPurger` - Builds delete plans for requests older than 24h and executes them using worker pool
  * `CacheGenMiddleware` - Adds gen number as a prefix to cache keys before doing Store/Fetch operations.

* Delete Request Lifecycle
Delete request could have one following states at a time:
1. `DeleteRequestStatusReceived` - No actions are done on request yet, just doing query time filtering
2. `DeleteRequestStatusBuildingPlan` - Request picked up for processing and building plans for it, still doing query time filtering
3. `DeleteRequestStatusDeleting` - Plans built already, running delete operations and still doing query time filtering
4. `DeleteRequestStatusProcessed` - All requested data deleted, not considering this for query time filtering

Delete requests keeps moving forward from one state to another if user does not request cancellation of it while it is in `DeleteRequestStatusReceived` state. They are moved from `DeleteRequestStatusReceived` to `DeleteRequestStatusBuildingPlan` state only after they are 24+ hours old, after which it keeps moving forward as it gets processed.

* Delete Plans
Process performing delete operation could die in the middle which can cause issues in queries if chunks are deleted first or end up in stale chunks if an index is deleted first. To avoid such issues a delete plan is built which include labels and chunks which are supposed to be deleted. To perform deletion in parallel, delete plan is sharded by day.

*New APIs*
* `/delete_series` - Accepts Prometheus style delete request for deleting series
* `/cancel_delete_request` - Cancel pending delete request
* `/get_all_delete_requests` - Get all delete requests

*Tradeoffs*
* Parallel execution of delete requests from same users is not supported - Parallel execution can invalidate another delete request's delete plan if they are touching same chunks.

I have added comments wherever required to explain the logic. I would add more implementation details to design doc [here](https://docs.google.com/document/d/1PeKwP3aGo3xVrR-2qJdoFdzTJxT8FcAbLm2ew_6UQyQ/edit). It would be great if someone can have a look an do a first pass to validate the approach and point out any major design issues. I have also added some ToDo's to be addressed next by me and do some refactoring wherever needed but would avoid doing so until someone can have a look over it and share any thoughts.
I would start adding tests, metrics and traces when my approach looks good.

Fixes #<issue number>

**Checklist**
- [ ] Traces Added
- [ ] Metrics Added
- [ ] Tests updated
- [ ] Documentation added


pkg/chunk/storage_client.go
```diff
@@ -30,6 +31,9 @@ type ObjectAndIndexClient interface {
 // WriteBatch represents a batch of writes.
 type WriteBatch interface {
 	Add(tableName, hashValue string, rangeValue []byte, value []byte)
+	Increment(tableName, hashValue string, rangeValue []byte, value int64)
```
jtlisi: I think you should consider creating a separate client for deletions instead of using the index client.

First off it will avoid having a bunch of clients with deletes unimplemented. Also using a dedicated client will allow for a cleaner interface to be designed. Also it will make it easier to design something that can be used more easily with a the internal KVStore which could be a good fit for this use case.

pkg/util/time.go
```diff
@@ -0,0 +1,23 @@
...
+	"math"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/weaveworks/common/httpgrpc"
+)
+
+// ParseTime parses the string into an int64, milliseconds since epoch.
+func ParseTime(s string) (int64, error) {
```
jtlisi: This code seems be copied from `pkg/querier/queryrange/query_range.go`. Deduplicating would be good.

pkg/chunk/chunk_store.go
```diff
@@ -343,6 +353,42 @@ func (c *store) getMetricNameChunks(ctx context.Context, userID string, from, th
 
 	// Filter out chunks based on the empty matchers in the query.
 	filteredChunks := filterChunksByMatchers(allChunks, filters)
+	return c.applyTombstonesToChunks(userID, filteredChunks, from, through)
+}
+
+// filters out completely deleted chunks and sets deletedIntervals for partially deleted chunks
+func (c *store) applyTombstonesToChunks(userID string, chunks []Chunk, queryRangeFrom, queryRangeThrough model.Time) ([]Chunk, error) {
+	tombstonesSet, err := c.tombstonesLoader.GetPendingTombstones(userID)
```
jtlisi: If we are unable to load Tombstones queries will be broken.

pkg/chunk/chunk_store.go
```diff
@@ -343,6 +353,42 @@ func (c *store) getMetricNameChunks(ctx context.Context, userID string, from, th
 
 	// Filter out chunks based on the empty matchers in the query.
 	filteredChunks := filterChunksByMatchers(allChunks, filters)
+	return c.applyTombstonesToChunks(userID, filteredChunks, from, through)
+}
+
+// filters out completely deleted chunks and sets deletedIntervals for partially deleted chunks
+func (c *store) applyTombstonesToChunks(userID string, chunks []Chunk, queryRangeFrom, queryRangeThrough model.Time) ([]Chunk, error) {
```
jtlisi: What impact will this have on query performance? If there is a large number of deletion jobs will that have a significant impact? If so we may want to consider isolating tombstones are per user basis.

sandeepsukhani:
Thanks for the initial review, I really appreciate it!

> It took me a while to get a handle on all of the new components being added in this PR. I think some components of this PR are further along than others.
> 
> One thing that really stood out to me is potential for the `DeleteStore` to break writes and reads. I personally do not think the delete store should ever cause reads/writes to fail. Prometheus allows for Warnings in the returned response. IMO failures along the delete store path should be propagated there. Also considering the generation of a cache is returned on a polled basis anyway I don't think we care that much about strong consistency.

Yeah, you are right, we don't have to break read/writes since we aim to be eventually consistent. I will fix it.
> As far as features go, Tombstoning/Cache-Invalidation are not essential to the implementation of this feature. However, they add the bulk of the complexity to this PR. The also are the source for a number of potential bugs and failure modes. I recommend removing them from this PR and adding them later in a future PR. The other option would be putting them behind feature flags. As things stand they are too embedded in critical paths throughout Cortex and have the potential to cause a number of issues for operators/users.

As far as I remember what was discussed in cortex community calls, users would need to make informed decisions about whether they want to continue with a delete request or cancel it which can only be done by query time filtering. If we do not have query time filtering I doubt many users would use it. We sure can put whole delete feature, including purger and query filtering behind a feature flag.
> 
> I also think there is a significant amount of complexity added by trying to utilize the index client for all delete operations. I think a seperate client would help make things a bit cleaner and avoid all the unimplemented panics.

I had thought about it but went with reusing same code because it saves the efforts of the user/operator in managing additional dependency. Also, I will be implementing all the unimplemented methods which panic currently, I had left it out to intentionally to get an initial review first on my approach. Do you have any other thoughts on it or do we want to continue with it?

pkg/chunk/delete_store.go
```diff
@@ -0,0 +1,366 @@
...
+	StartTime model.Time          `json:"start_time"`
+	EndTime   model.Time          `json:"end_time"`
+	Selectors []string            `json:"selectors"`
+	Status    string              `json:"status"`
+	Matchers  [][]*labels.Matcher `json:"-"`
+	CreatedAt model.Time          `json:"created_at"`
+}
+
+// DeleteStore provides all the methods required to manage lifecycle of delete request and things related to it
+type DeleteStore interface {
```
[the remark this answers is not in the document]
sandeepsukhani: Yes, I had added it for some other use-case but ended up refactoring the code. Will remove it for now and add it back if I need it for tests.

pkg/chunk/purger/purger.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	if pb == nil {
+		level.Debug(util.Logger).Log("msg", "plan not found, must be executed already",
+			"request-id", requestID, "user-id", userID, "plan-no", planNo)
+		// this means plan was already executed and got removed. Do nothing.
+		return nil
+	}
+
+	level.Debug(util.Logger).Log("msg", "executing plan",
```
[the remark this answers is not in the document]
sandeepsukhani: I didn't want too much noise in logs so I used info level only for important logs since a delete request could touch thousands of series and chunks. Will go through them again and switch some of them to info level until it gets stable.

pkg/chunk/purger/purger.go
```diff
@@ -0,0 +1,494 @@
...
+	// and break the purge plan for other requests
+	inProcessRequestIDs    map[string]string
+	inProcessRequestIDsMtx sync.RWMutex
+
+	quit chan struct{}
+	wg   sync.WaitGroup
+}
+
+// NewDataPurger creates a new DataPurger
+func NewDataPurger(cfg DataPurgerConfig, deleteStore chunk.DeleteStore, chunkStore chunk.Store) (*DataPurger, error) {
```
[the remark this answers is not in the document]
sandeepsukhani: yes, it should be run as a single instance only like table manager. You can run multiple worker goroutines with it(defaults to 2 now) to execute delete plan concurrently.

pkg/chunk/purger/purger.go
```diff
@@ -0,0 +1,494 @@
...
+
+	dp.wg.Add(1)
+
+	go func() {
+
+		for {
+			select {
+			case jobExecutionStatus := <-dp.workerJobExecutionStatusChan:
+				if jobExecutionStatus.err != nil {
+					level.Error(util.Logger).Log("msg", "error executing delete plan", "request-id", jobExecutionStatus.deleteRequestID, "plan-no", jobExecutionStatus.planNo, "err", jobExecutionStatus.err)
```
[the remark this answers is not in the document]
sandeepsukhani: yes it will and none of the other delete operations would start for that user until this one finishes.

pkg/chunk/purger/purger.go
```diff
@@ -0,0 +1,494 @@
...
+	deleteRequestID string
+}
+
+type workerJobExecutionStatus struct {
+	workerJob
+	err error
+}
+
+// DataPurger does the purging of data which is requested to be deleted
+type DataPurger struct {
```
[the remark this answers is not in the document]
sandeepsukhani: By instrumentation do you mean exposing metrics and traces? If yes then I haven't added them anywhere yet because I wanted to get the approach validated first.

pkg/chunk/azure/blob_storage_client.go
```diff
@@ -141,3 +141,7 @@ func (b *BlobStorage) getBlobURL(blobID string) (azblob.BlockBlobURL, error) {
 		},
 	})), nil
 }
+
+func (b *BlobStorage) DeleteChunk(ctx context.Context, chunkID string) error {
+	panic("implement me")
```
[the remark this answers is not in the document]
sandeepsukhani: I am going to implement all new methods for the stores once we make sure this is what we are going to stick to.

pkg/chunk/storage_client.go
sandeepsukhani: KVStore was an option but I didn't want to add an additional dependency so I went with reusing existing code. Please let me know what do you think about it.

pkg/chunk/chunk_store.go
sandeepsukhani: Yes, there would be impact due to this on the query side. After discussing this in community call we are okay with it given it is only going to impact users who have pending delete requests.

sandeepsukhani:
One more problem I can think of when we would still continue to return results if loading tombstones/cache gen numbers fail is wrong results might get cached with the latest gen number. I feel being eventually consistent is different than never being consistent since loading tombstones could fail forever until someone fixes what the issue is and we will also have to nuke whole cache to remove incorrectly cached results. I think not being able to load any of them would be an issue which would need someone's attention. What do you think?

pkg/chunk/purger/purger.go
```diff
@@ -0,0 +1,494 @@
...
+	requestsWithBuildingPlanStatus, err := dp.deleteStore.GetDeleteRequestsByStatus(context.Background(), chunk.DeleteRequestStatusBuildingPlan)
+	if err != nil {
+		return err
+	}
+
+	for _, deleteRequest := range requestsWithBuildingPlanStatus {
+		level.Debug(util.Logger).Log("msg", "loaded in process delete requests with status building plan", "request-id", deleteRequest.RequestID, "user-id", deleteRequest.UserID)
+
+		dp.inProcessRequestIDs[deleteRequest.UserID] = dp.inProcessRequestIDs[deleteRequest.RequestID]
+		dp.buildPlanChan <- deleteRequest
```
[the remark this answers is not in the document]
sandeepsukhani: yes, will add a check for it, thanks!

pkg/chunk/purger/purger.go
```diff
@@ -0,0 +1,494 @@
...
+		deleteStore:                  deleteStore,
+		chunkStore:                   chunkStore,
+		inProcessRequestIDs:          map[string]string{},
+		buildPlanChan:                make(chan chunk.DeleteRequest, 50),
+		executePlansChan:             make(chan chunk.DeleteRequest, 50),
+		workerJobChan:                make(chan workerJob, 50),
+		workerJobExecutionStatusChan: make(chan workerJobExecutionStatus, 50),
+	}
+
+	err := dataPurger.loadInprocessDeleteRequests()
```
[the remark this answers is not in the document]
sandeepsukhani: will fix it, thanks!

pkg/chunk/delete_store.go
```diff
@@ -0,0 +1,366 @@
...
+	"fmt"
+	"hash/fnv"
+	"strings"
+	"time"
+
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+const (
```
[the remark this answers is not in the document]
sandeepsukhani: I am not sure what advantage that would have except have smaller names for each status type.
I prefer this because it the names have enough context to be used outside of the package

pkg/chunk/storage_client.go
jtlisi: What do you mean by an additional dependency? Couldn't this be implemented on top of a ETCD/Consul/Gossip kvstore, which already exist?

jtlisi:
I think we can agree that the tombstone loader should not in any way affect writes. However, the answer may be less clear on reads. IMO access to the tombstone loader should not cause an error on reads. I think we should treat it similar to how we treat caching, i.e. we don't let cache errors propagate through the system. I understand this could mean cortex will serve data that was meant to be deleted. However, I would rather data that was meant to be deleted be served temporarily instead of breaking reads entirely.

pkg/chunk/storage_client.go
sandeepsukhani: Maybe gen numbers can be stored but we also have to store delete plans and delete requests which would be too much for gossip at least. Also I wouldn't consider them as reliable as index stores for storing data which won't get lost.

pkg/chunk/purger/purger.go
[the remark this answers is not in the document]
jtlisi: I think any time a plan is being exectuted/completed we should log.

pkg/chunk/purger/purger.go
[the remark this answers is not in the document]
jtlisi: That's fine if you want to wait to implement them.

pkg/chunk/delete_store.go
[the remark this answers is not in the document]
jtlisi: Generally named types make it easier to understand a function signature. I know you are using these as part of the index client. However, I don't think you should be doing that. I think the delete client should have a separate interface with a function that does something along the lines of `SetStatus(ctx context.Context, planName string, status PlanStatus) error`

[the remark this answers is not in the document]
jtlisi: Enums in general make understanding the code much easier. You can immediately understand what options you have when using a function by finding the named type. When you use a `string` it makes it much harder to understand what options are provided by the API. It also makes it much easier to misuse.

pkg/chunk/storage_client.go
jtlisi: Cache Gen numbers should be stored in a KVStore. We can take advantage of the `CAS` interface. We can also do it in a way where the read/write path should know nothing about the cache gen number outside of the cache package. All that would be needed is another layer of cache with a ETCD/Consul KVStore embedded in it. This same KVStore would also be used in the Purger.

pkg/chunk/delete_store.go
```diff
@@ -0,0 +1,366 @@
...
+	EndTime   model.Time          `json:"end_time"`
+	Selectors []string            `json:"selectors"`
+	Status    string              `json:"status"`
+	Matchers  [][]*labels.Matcher `json:"-"`
+	CreatedAt model.Time          `json:"created_at"`
+}
+
+// DeleteStore provides all the methods required to manage lifecycle of delete request and things related to it
+type DeleteStore interface {
+	Add(ctx context.Context, userID string, startTime, endTime model.Time, selectors []string) error
```
jtlisi: This function should be called `AddDeletePlan` or `CreateDeletePlan`

pkg/chunk/delete_store.go
```diff
@@ -0,0 +1,366 @@
...
+	GetDeleteRequestsByStatus(ctx context.Context, status string) ([]DeleteRequest, error)
+	GetDeleteRequestsForUserByStatus(ctx context.Context, userID string, status string) ([]DeleteRequest, error)
+	GetPendingDeleteRequestsForUser(ctx context.Context, userID string) ([]DeleteRequest, error)
+	UpdateStatus(ctx context.Context, userID, requestID string, newStatus string) error
+	RemoveDeleteRequest(ctx context.Context, userID, requestID string, createdAt, startTime, endTime model.Time) error
```
jtlisi: Is it difficult to combine `RemoveDeleteRequest` and `RemoveDeletePlan` into a single function? Kind of a code smell there.

jtlisi: It looks like `RemoveDeletePlan` is unused?

pkg/cortex/cortex.go
```diff
@@ -59,22 +60,24 @@ type Config struct {
...
+	Schema           chunk.SchemaConfig       `yaml:"schema,omitempty" doc:"hidden"` // Doc generation tool doesn't support it because part of the SchemaConfig doesn't support CLI flags (needs manual documentation)
+	LimitsConfig     validation.Limits        `yaml:"limits,omitempty"`
+	Prealloc         client.PreallocConfig    `yaml:"prealloc,omitempty" doc:"hidden"`
+	Worker           frontend.WorkerConfig    `yaml:"frontend_worker,omitempty"`
+	Frontend         frontend.Config          `yaml:"frontend,omitempty"`
+	QueryRange       queryrange.Config        `yaml:"query_range,omitempty"`
+	TableManager     chunk.TableManagerConfig `yaml:"table_manager,omitempty"`
+	Encoding         encoding.Config          `yaml:"-"` // No yaml for this, it only works with flags.
+	TSDB             tsdb.Config              `yaml:"tsdb" doc:"hidden"`
+	DeletesConfig    chunk.DeleteStoreConfig  `yaml:"deletes,omitempty"`
```
pracucci: Can we find a better name for the YAML config? `deletes` looks a bit unclear to me, if I look at it with the eyes of the first time user.

Also, is it necessary to have this config at root level? Is there a better place inside any other root config block we already have? _Thinking loudly_

pkg/chunk/delete_store.go
sandeepsukhani: Actually whole delete request has a status not delete plan. If a delete plan is there in db then it needs to be executed. When a delete request has no delete plan its marked as complete. We might not want to add an overhead of managing statuses for a delete plan.

pkg/chunk/delete_store.go
sandeepsukhani: `Add` function adds a new delete request so I guess you mean `AddDeleteRequest`

pkg/chunk/storage_client.go
jtlisi: Starting a separate comment thread to talk about the usage of the IndexClient separately from CacheGen.  IMO this is a misuse of the IndexClient. It adds a number of functions to the WriteBatch interface that are entirely unrelated to the Index. It reuses parameters with the names `hashValue` and `rangeValue` which don't any real relation to what we are trying to do when storing Deletion Plans.

A separate client for deletes is a much cleaner way to do this. We don't need a `WriteBatch` for storing deletion plans or updating their status. From my perspective, the only thing we would need is a few functions, specifically:`CreatePlan`, `ListPlans` and `UpdatePlan`. This would make it obvious what the client is doing.

It also means the `ObjectClient` and `IndexClient` only have to implement one function respectively to support deletes. If the `DeleteClient`/`DeletePlanClient` isn't supported in a particular backend, that's fine because they could always use another backend they do support concurrently to serve as their delete client. I am also suggesting this because I think the range of clients that could support deletes is much broader that ones that support the index. We could be implementing this on top of S3/GCS, or a KVStore for example.

pracucci:
@sandeepsukhani Does this PR fixes 1449?

sandeepsukhani:
Closing this since we are doing separate smaller PRs.

Issue #1234, opened 2019-02-21
Add support for Azure Blob Storage


huydinhle:
Add support for Azure blob storage along side with GCS and S3

tamalsaha:
I wonder if it makes sense to use something like https://github.com/google/go-cloud/tree/master/blob to reimplement the [ObjectClient](https://github.com/cortexproject/cortex/blob/0ef1ce95aed22bd2bfbb1b53bacba20766e35c79/pkg/chunk/storage_client.go#L18) ?

tomwilkie:
@huydinhle let me know if you need any pointers.  The [ObjectClient](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/storage_client.go#L18) interface is the one to implement (as @tamalsaha mentioned), and you'll need to plumb it into the factory in [NewObjectClient](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/storage/factory.go#L153).

@tamalsaha could do!  Its super important we have batching in the interface, which they don't seem to have.  And we've need to tweak buffer sizes on the GCS client - not sure how to do that with go-cloud.  I think if that library had existed when we started Cortex, we would have used it - but we've got almost all the implementations we'd need, so unless its super simple, I'm not sure we'd be that keen.

bboreham:
Is anyone seriously looking at this, or it's just an idea?
What are the advantages, compared to other options?
(I now see this request came from Loki; there is a bit more background there but not much)
Should there be a related request to support indexing on Azure?

henningzahn:
Is anyone working on this topic? Or are there shortcomings on Azure side that cannot be overcome? We would like to adopt Loki but cannot start because of the missing Azure support.

santhoshparepu:
Is this going to be supported soon? We are Azure shop and without having the support for Azure blob there are challenges in adopting this.

gouthamve:
You can use this for now: https://az.minio.io/ to provide an S3 interface for Azure Blob storage. Meanwhile you can also watch this issue for a more direct implementation: https://github.com/cortexproject/cortex/issues/1681

khaines:
https://docs.min.io/docs/deploy-minio-on-kubernetes.html  may also be useful if one is running in AKS , as the Azure marketplace setup uses at minimum 2 isolated VMs.

[...]

PR #1913, opened 2019-12-13
Adding Microsoft Azure Blob storage support for chunks

khaines:
**What this PR does**: Adds support for Microsoft Azure blob storage to be used for storing data chunks.

Fixes https://github.com/cortexproject/cortex/issues/1234

**Checklist**


pkg/chunk/azure/blob_storage_client.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/chunk/util"
+)
+
+const blobURLFmt = "https://%s.blob.core.windows.net/%s/%s"
+
+// BlobStorageConfig defines the configurable flags that can be defined when using azure blob storage.
+type BlobStorageConfig struct {
```
pracucci: By convention we use lowercase dash-separated yaml config options (ie. `containerName` to `container_name`). May you fix it, please?

khaines: done. Thanks for calling that out.

khaines: Fixed. Thank you for bringing that up.

PR #1914, opened 2019-12-13
Added rate limiter for BigTable writes. #1739

### LINKED ISSUE #1739: Rate limit Chunk & Index writes to Bigtable
As far as I know there's currently no way to rate limit the write requests from Cortex -> BigTable.

We currently use a single dev node which can run handle up to 25k writes / sec. After a rolling update the ingesters try to flush lots of serieses to BigTable. In my case this takes approx 1 hour. The ingesters do that without any rate limits, which quickly results in a very busy Cortex. During that time my read requests mostly fail because BigTable is so saturated. Even with bigger clusters I can see this happening for some time.

Therefore I propose an option to rate limit the writes, to ensure BigTable is not overstrained handling the read requests at the same time.

_For reference, here's Bryan's comment from Slack:_

```
Bryan Boreham 8:26 PM
I don't think that's in the code at the moment.  The DynamoDB client has a rate-limiter, based on throttling responses that come back from DynamoDB.
I think it would be quite simple to put a golang.org/x/time/rate.Limiter on the write calls.
```

tcl00taco:
Signed-off-by: Samiur Arif <samiur.arif@gmail.com>

**What this PR does**:
1. Added rate limiters for BigTable writes.
2. Please note default values are configurable and are waiting for suggestion for tweaks. 

Fixes #1739

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


bboreham:
Have you done any research into the combination of rate-limiter and backoff?

In the AWS code I found it quite nice just to use a rate-limiter, though DynamoDB gives us a guide by saying explictly how much the request has been throttled.

tcl00taco:
@bboreham I will have to dig more, my assumptions were based on single cluster one Bigtable setup that I have created locally and tried 100,000 (max limit) for 20 minutes to simulate the problem @weeco  mentioned which does create lack of query response or slowness. I have noticed with the combination it seems to improve the situation at least I can query but then again it is not a real world load. I will dig more and understand what is going on, meantime any pointers will be appreciated.

weeco:
@tcl00taco At the time I faced the issue I was not using a memcache for BigTable index writes. Also I was just using a single node cluster (dev). I believe due to the missing cache I was exceeding these limits which were above the average write loads.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -13,34 +13,36 @@ import (
...
-	columnPrefix = columnFamily + ":"
-	column       = "c"
-	separator    = "\000"
-	maxRowReads  = 100
+	columnFamily              = "f"
+	columnPrefix              = columnFamily + ":"
+	column                    = "c"
+	separator                 = "\000"
+	maxRowReads               = 100
+	bigtableMaxWriteBatchSize = 10000
```
pracucci: Why is hit hardcoded? How the value has been picked?

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -179,7 +188,10 @@ func (s *storageClientColumnKey) BatchWrite(ctx context.Context, batch chunk.Wri
 			rowKeys = append(rowKeys, rowKey)
 			muts = append(muts, mut)
 		}
-
+		if s.writeLimiter.Allow() == false {
+			s.writeLimiter.WaitN(ctx, len(muts))
+			backoff.Wait()
```
pracucci: Why do we need a backoff and `writeLimiter.WaitN()` is not enough?

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -49,16 +51,19 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Instance, "bigtable.instance", "", "Bigtable instance ID.")
 	f.BoolVar(&cfg.TableCacheEnabled, "bigtable.table-cache.enabled", true, "If enabled, once a tables info is fetched, it is cached.")
 	f.DurationVar(&cfg.TableCacheExpiration, "bigtable.table-cache.expiration", 30*time.Minute, "Duration to cache tables before checking again.")
-
+	f.Float64Var(&cfg.WriteLimit, "bigtable.throttle-limit", 10.0, "BigTable rate cap to back off when throttled.")
```
pracucci: I would suggest to:
1. Call it `bigtable.write-rate-limit`
2. Defaults it to `0` (disabled)
3. Improve the flag description explaining it's the max number of keys written per second to bigtable, once this rate is exceeded writes are throttled and that 0 disables it

pkg/chunk/gcp/bigtable_index_client.go
jtlisi: Agreed, @tcl00taco are these values based on some constants related to bigtable. From what I understand the max batch size would be `100000` not `10000`

tcl00taco: @pracucci and @jtlisi Will remove hard coded value and make it configurable. The values were based on single cluster BigTable local test where I have not seen any query time out / delay. Thanks for the feedback and will make the changes.

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: @pracucci  We do not need back off, will remove it.

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: @pracucci this is exactly what I was looking for, some pointers, will make changes and test it again.

tcl00taco:
To all sorry for the delay, I have just realized today that I did not push the appropriate commit before leaving for Christmas break. Thanks for your patience and understanding.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -179,7 +197,10 @@ func (s *storageClientColumnKey) BatchWrite(ctx context.Context, batch chunk.Wri
 			rowKeys = append(rowKeys, rowKey)
 			muts = append(muts, mut)
 		}
-
+		if s.writeLimiter.Allow() == false {
```
pracucci: `Allow()` is a shorthand for `AllowN(time.Now(), 1)`, which increments by 1 the number of "consumed" tokens in the rate limiter. Using it here is not correct, because we don't want to increment by 1 the number of tokens, and then increment again in the `WaitN()`.

I would suggest to remove the `Allow()` at all, and just call the `WaitN()`. You will not be able to track your `rateLimitedWriteRequest` metric, but I don't see any non-hacky wait to do it.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -34,31 +37,44 @@ type Config struct {
...
 	TableCacheEnabled    bool
 	TableCacheExpiration time.Duration
+	WriteLimit           float64
+	MaxWriteBurstSize    int
 }
 
+var (
+	rateLimitedWriteRequest = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "rate_limited_write_requests_total",
```
pracucci: I think it would be better name the metric with `bigtable` prefix, similarly to what we did for DynamoDB. For example, `bigtable_rate_limited_requests_total`.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -34,31 +37,44 @@ type Config struct {
...
+		Help:      "Total count of rate limited write requests.",
+	}, []string{"name"})
```
pracucci: I would call it `operation` (like we did for DynamoDB) and have values like `write-index` and `write-chunk`.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -179,7 +197,10 @@ func (s *storageClientColumnKey) BatchWrite(ctx context.Context, batch chunk.Wri
 			rowKeys = append(rowKeys, rowKey)
 			muts = append(muts, mut)
 		}
-
+		if s.writeLimiter.Allow() == false {
+			s.writeLimiter.WaitN(ctx, len(muts))
```
pracucci: What do we want to rate limit on? Mutations? Keys?

pkg/chunk/gcp/bigtable_object_client.go
```diff
@@ -65,6 +68,10 @@ func (s *bigtableObjectClient) PutChunks(ctx context.Context, chunks []chunk.Chu
 
 	for tableName := range keys {
 		table := s.client.Open(tableName)
+		if s.writeLimiter.Allow() == false {
+			s.writeLimiter.WaitN(ctx, len(muts))
```
pracucci: `len(muts)` doesn't look correct to me. Shouldn't be `len(muts[tableName])` to rate limit on the mutations? (assuming we want to limit on the mutations).

docs/guides/running.md
```diff
@@ -232,6 +232,35 @@ We do not recommend configuring a liveness probe on ingesters -
 killing them is a last resort and should not be left to a machine.
 
 
+### Remote writing Prometheus
```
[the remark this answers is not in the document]
tcl00taco: Will Do!..

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -34,31 +37,44 @@ type Config struct {
 	Project  string `yaml:"project"`
 	Instance string `yaml:"instance"`
 
-	GRPCClientConfig grpcclient.Config `yaml:"grpc_client_config"`
-
-	ColumnKey      bool
-	DistributeKeys bool
-
+	GRPCClientConfig     grpcclient.Config `yaml:"grpc_client_config"`
+	BackoffConfig        util.BackoffConfig
```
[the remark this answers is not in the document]
tcl00taco: @pracucci done!

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -34,31 +37,44 @@ type Config struct {
...
+)
+
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Project, "bigtable.project", "", "Bigtable project ID.")
 	f.StringVar(&cfg.Instance, "bigtable.instance", "", "Bigtable instance ID.")
 	f.BoolVar(&cfg.TableCacheEnabled, "bigtable.table-cache.enabled", true, "If enabled, once a tables info is fetched, it is cached.")
 	f.DurationVar(&cfg.TableCacheExpiration, "bigtable.table-cache.expiration", 30*time.Minute, "Duration to cache tables before checking again.")
-
+	f.Float64Var(&cfg.WriteLimit, "bigtable.write-rate-limit", 0.0, "BigTable rate limiter, default it is disabled.")
```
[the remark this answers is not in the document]
tcl00taco: agree...done!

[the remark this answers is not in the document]
tcl00taco: okay.

pkg/chunk/gcp/bigtable_object_client.go
tcl00taco: Good catch, it should be len(muts[tableName]) to rate limit on the mutations.

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: okay!

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: agree, changed.

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: I got it, would have been great if we could track metric with waitN.

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: Mutations!.

pracucci:
@tcl00taco I've seen you replied to my comments, but I can't see any code change. Maybe you forgot to `git push` or it failed?

tcl00taco:
@pracucci Apologies again, they should be here now.

docs/guides/running.md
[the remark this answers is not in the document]
pracucci: FYI, it's still here. Please double check the changed file after you `git push` to ensure the rebase worked as expected (same issue for `pkg/configs/legacy_promql/testdata/functions.test`).

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -180,6 +198,9 @@ func (s *storageClientColumnKey) BatchWrite(ctx context.Context, batch chunk.Wri
 			muts = append(muts, mut)
 		}
 
+		s.writeLimiter.WaitN(ctx, len(muts))
+		//rateLimitedWriteRequest.WithLabelValues("write-index").Inc()
```
pracucci: Please remove any commented code

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -41,24 +44,37 @@ type Config struct {
...
+
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Project, "bigtable.project", "", "Bigtable project ID.")
 	f.StringVar(&cfg.Instance, "bigtable.instance", "", "Bigtable instance ID.")
 	f.BoolVar(&cfg.TableCacheEnabled, "bigtable.table-cache.enabled", true, "If enabled, once a tables info is fetched, it is cached.")
 	f.DurationVar(&cfg.TableCacheExpiration, "bigtable.table-cache.expiration", 30*time.Minute, "Duration to cache tables before checking again.")
-
+	f.IntVar(&cfg.WriteLimit, "bigtable.write-rate-limit", 0, "BigTable rate limiter for write operations, by default it is disabled.")
+	f.IntVar(&cfg.MaxWriteBurstSize, "bigtable.write-limit-burst", 0, "BigTable maximum burst size for write operations, by default it is disabled.")
```
pracucci: A couple of tiny things to keep more consistency with the rest of the config:
1. `bigtable.write-rate-limit-burst-size`
2. `BigTable maximum burst size for write operations when the rate limit is enabled`
3. Replace `by default it is disabled`, with `0 is disabled` (both CLI flags)

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -180,6 +198,9 @@ func (s *storageClientColumnKey) BatchWrite(ctx context.Context, batch chunk.Wri
 			muts = append(muts, mut)
 		}
 
+		s.writeLimiter.WaitN(ctx, len(muts))
```
pracucci: Given we're using the `WaitN()`, does the `ctx` have a deadline set? If not, it may be risky cause it could potentially hung for a long time, while a fast failing may be more desiderable.

pkg/chunk/gcp/bigtable_object_client.go
```diff
@@ -65,6 +68,10 @@ func (s *bigtableObjectClient) PutChunks(ctx context.Context, chunks []chunk.Chu
 
 	for tableName := range keys {
 		table := s.client.Open(tableName)
+		if s.writeLimiter.Allow() == false {
```
pracucci: Using `Allow()` suffers the issue I've already commented about, reason why you've changed the `pkg/chunk/gcp/bigtable_index_client.go` implementation.

tcl00taco: okay! Made similar changes as bigtable_index_client go. Since, we no longer have the ability to write metrics for these rate limiters, should I remove the metrics altogether?

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: I have a default deadline is not set.

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: done.

pkg/chunk/gcp/bigtable_index_client.go
tcl00taco: done.

tcl00taco:
@pracucci Any idea how I can fix mixing signoff for a particular commit? I have tried the following but did not work.
1. git commit --amend -s -c e770ad1b8   # where e770d1b8 is the commit hash which was not signed off properly.

2. git push -f origin

tcl00taco:
> @pracucci Any idea how I can fix mixing signoff for a particular commit? I have tried the following but did not work.
> 
>     1. git commit --amend -s -c [e770ad1](https://github.com/cortexproject/cortex/commit/e770ad1b8fe025599866b4afd199bc58f3049a13)   # where e770d1b8 is the commit hash which was not signed off properly.
> 
>     2. git push -f origin

This is fixed.


@pracucci  Any more comments? Also, I am not sure why generated doc checks are failing... It seems to be adding these extra lines to configure.md. 

 # BigTable rate limiter for write operations, with 0 is disable.
  # CLI flag: -bigtable.write-rate-limit
  [writelimit: <int> | default = 0]

  # BigTable maximum burst size for write operations when the rate limit is
  # enabled, with 0 is disabled.
  # CLI flag: -bigtable.write-limit-burst-size
  [maxwriteburstsize: <int> | default = 0]

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -50,16 +53,18 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Instance, "bigtable.instance", "", "Bigtable instance ID.")
 	f.BoolVar(&cfg.TableCacheEnabled, "bigtable.table-cache.enabled", true, "If enabled, once a tables info is fetched, it is cached.")
 	f.DurationVar(&cfg.TableCacheExpiration, "bigtable.table-cache.expiration", 30*time.Minute, "Duration to cache tables before checking again.")
-
+	f.IntVar(&cfg.WriteLimit, "bigtable.write-rate-limit", 0, "BigTable rate limiter for write operations, with 0 is disable.")
```
csmarchbanks: `0 is disabled` rather than `with 0 is disabled`? On this line and the line after?

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -184,6 +191,22 @@ func (s *storageClientColumnKey) BatchWrite(ctx context.Context, batch chunk.Wri
 			muts = append(muts, mut)
 		}
 
+		if s.cfg.WriteLimit > 0 {
+
```
csmarchbanks: Nit: Extra whitespace.

pkg/chunk/gcp/bigtable_object_client.go
```diff
@@ -65,6 +69,23 @@ func (s *bigtableObjectClient) PutChunks(ctx context.Context, chunks []chunk.Chu
 
 	for tableName := range keys {
 		table := s.client.Open(tableName)
+
+		if s.cfg.WriteLimit > 0 {
+
```
csmarchbanks: Nit: Another extra line.

PR #1915, opened 2019-12-16
Replace interface names with local ip in distributor tests

codesome:
Hardcoding interface names cause the tests to fail on systems that have interface names different than the ones hardcoded.


pkg/distributor/distributor_test.go
```diff
@@ -754,7 +755,14 @@ func prepare(t *testing.T, numIngesters, happyIngesters int, queryDelay time.Dur
 	cfg.DistributorRing.HeartbeatPeriod = 100 * time.Millisecond
 	cfg.DistributorRing.InstanceID = strconv.Itoa(rand.Int())
 	cfg.DistributorRing.KVStore.Mock = kvStore
-	cfg.DistributorRing.InstanceInterfaceNames = []string{"eth0", "en0", "lo0"}
```
pracucci: I think you can make it even easier. The Interfaces are used only when `InstanceAddr` is not configured. You could just configure `InstanceAddr` to `127.0.0.1`. May you give it a try, please?

codesome: Yes it works. I have updated the PR.

PR #1917, opened 2019-12-16
Limit concurrent replays of TSDB on startup

thorfour:
**What this PR does**: Experimental TSDB change only. On startup walk the TSDB dir, and open a TSDB under each user dir. Limiting the number of concurrently opening TSDB. 

Large memory spikes were observed when Ingesters would startup that had a large number of users with WALs. 
image

This was causing oom kills on ingesters that would otherwise be able to run just fine. This change limits the number of TSDBs that can be opening at once, by opening all pre-existing user TSDBs at startup.

It does not open a TSDB for an empty user dir, which means it has no interaction with a transfer, since transfers delete empty dirs, and error out on non-empty user dirs.

Edit: While this change is still useful it doesn't address the memory usage spikes seen above, and because of a missed lock when looking over the code WAL's weren't being replayed in parallel. WAL replay still eats a large amount of memory regardless of how many are happening in parallel.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
 * [FEATURE] EXPERIMENTAL: Added `/series` API endpoint support with TSDB blocks storage. #1830
 * [FEATURE] Added "multi" KV store that can interact with two other KV stores, primary one for all reads and writes, and secondary one, which only receives writes. Primary/secondary store can be modified in runtime via runtime-config mechanism (previously "overrides"). #1749
 * [ENHANCEMENT] Added `password` and `enable_tls` options to redis cache configuration. Enables usage of Microsoft Azure Cache for Redis service.
+* [ENHANCEMENT] Experimental TSDB: Open existing TSDB on startup to prevent ingester from becoming ready before it can accept writes. #1917
+  * `--experimental.tsdb.max-tsdb-opening-concurrency-on-startup`
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
 * [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
 * [BUGFIX] TSDB: Fixed handling of out of order/bound samples in ingesters with the experimental TSDB blocks storage. #1864
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -2,7 +2,10 @@ package ingester
 
 import (
 	"fmt"
+	"io"
 	"net/http"
+	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -14,6 +17,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/pkg/gate"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
@@ -72,6 +76,11 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 	// Init the limter and instantiate the user states which depend on it
 	i.limiter = NewSeriesLimiter(limits, i.lifecycler, cfg.LifecyclerConfig.RingConfig.ReplicationFactor, cfg.ShardByAllLabels)
 	i.userStates = newUserStates(i.limiter, cfg)
+
+	// Scan and open TSDB's that already exist on disk
+	if err := i.openExistingTSDB(context.Background()); err != nil {
+		return nil, err
+	}
 
 	// Now that user states have been created, we can start the lifecycler
 	i.lifecycler.Start()
@@ -382,11 +391,23 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error) 
 		return nil, fmt.Errorf(errTSDBCreateIncompatibleState, ingesterState)
 	}
 
+	// Create the database and a shipper for a user
+	db, err := i.createTSDB(userID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add the db to list of user databases
+	i.TSDBState.dbs[userID] = db
+	return db, nil
+}
+
+// createTSDB creates a TSDB for a given userID, and returns the created db.
+func (i *Ingester) createTSDB(userID string) (*tsdb.DB, error) {
 	udir := i.cfg.TSDBConfig.BlocksDir(userID)
 
 	// Create a new user database
-	var err error
-	db, err = tsdb.Open(udir, util.Logger, nil, &tsdb.Options{
+	db, err := tsdb.Open(udir, util.Logger, nil, &tsdb.Options{
 		RetentionDuration: uint64(i.cfg.TSDBConfig.Retention / time.Millisecond),
 		BlockRanges:       i.cfg.TSDBConfig.BlockRanges.ToMillisecondRanges(),
 		NoLockfile:        true,
@@ -422,8 +443,6 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error) 
 		}()
 	}
 
-	i.TSDBState.dbs[userID] = db
-
 	return db, nil
 }
 
@@ -458,4 +477,74 @@ func (i *Ingester) closeAllTSDB() {
 	// Wait until all Close() completed
 	i.userStatesMtx.Unlock()
 	wg.Wait()
+}
+
+// openExistingTSDB walks the user tsdb dir, and opens a tsdb for each user. This may start a WAL replay, so we limit the number of
+// concurrently opening TSDB.
+func (i *Ingester) openExistingTSDB(ctx context.Context) error {
+	level.Info(util.Logger).Log("msg", "opening existing TSDBs")
+	wg := &sync.WaitGroup{}
+	openGate := gate.New(i.cfg.TSDBConfig.MaxTSDBOpeningConcurrencyOnStartup)
+
+	err := filepath.Walk(i.cfg.TSDBConfig.Dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return filepath.SkipDir
+		}
+
+		// Skip root dir and all other files
+		if path == i.cfg.TSDBConfig.Dir || !info.IsDir() {
+			return nil
+		}
+
+		// Top level directories are assumed to be user TSDBs
+		userID := info.Name()
+		f, err := os.Open(path)
+		if err != nil {
+			level.Error(util.Logger).Log("msg", "unable to open user TSDB dir", "err", err, "user", userID, "path", path)
+			return filepath.SkipDir
+		}
+		defer f.Close()
+
+		// If the dir is empty skip it
+		if _, err := f.Readdirnames(1); err != nil {
+			if err != io.EOF {
+				level.Error(util.Logger).Log("msg", "unable to read TSDB dir", "err", err, "user", userID, "path", path)
+			}
+
+			return filepath.SkipDir
+		}
+
+		// Limit the number of TSDB's opening concurrently. Start blocks until there's a free spot available or the context is cancelled.
+		if err := openGate.Start(ctx); err != nil {
+			return err
+		}
+
+		wg.Add(1)
+		go func(userID string) {
+			defer wg.Done()
+			defer openGate.Done()
+			db, err := i.createTSDB(userID)
+			if err != nil {
+				level.Error(util.Logger).Log("msg", "unable to open user TSDB", "err", err, "user", userID)
+				return
+			}
+
+			// Add the database to the map of user databases
+			i.userStatesMtx.Lock()
+			i.TSDBState.dbs[userID] = db
+			i.userStatesMtx.Unlock()
+
+		}(userID)
+
+		return filepath.SkipDir // Don't descend into directories
+	})
+
+	// Wait for all opening routines to finish
+	wg.Wait()
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "error while opening existing TSDBs")
+	} else {
+		level.Info(util.Logger).Log("msg", "successfully opened existing TSDBs")
+	}
+	return err
 }
--- a/pkg/ingester/ingester_v2_test.go
+++ b/pkg/ingester/ingester_v2_test.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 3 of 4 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+
+	err := filepath.Walk(i.cfg.TSDBConfig.Dir, func(path string, info os.FileInfo, err error) error {
+		if path == i.cfg.TSDBConfig.Dir { // Nothing to do for root
+			return nil
+		}
+
+		// Top level directories are assumed to be user TSDBs
+		if info.IsDir() {
+
+			userID := info.Name()
```
codesome: Do we want to verify if the directory name actually follows the pattern of a user ID instead of assuming? (i.e. just an integer)

pracucci: The user ID could any string (not necessarily a number). For example, the default user is called `fake`. I would advise on no validation here, cause we can't do any assumption on it.

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+	wg := &sync.WaitGroup{}
+	openGate := gate.New(i.cfg.TSDBConfig.MaxOpeningTSDBOnStartup)
+
+	err := filepath.Walk(i.cfg.TSDBConfig.Dir, func(path string, info os.FileInfo, err error) error {
+		if path == i.cfg.TSDBConfig.Dir { // Nothing to do for root
+			return nil
+		}
+
+		// Top level directories are assumed to be user TSDBs
+		if info.IsDir() {
```
pracucci: I would simplify this with a guard:

```
if !info.isDir() {
  return nil
}
```

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
 	i.userStatesMtx.Unlock()
 	wg.Wait()
 }
+
+// openExistingTSDB walks the user tsdb dir, and opens a tsdb for each user. This may start a WAL replay, so we limit the number of
+// concurrently opening TSDB.
+func (i *Ingester) openExistingTSDB(ctx context.Context) error {
```
pracucci: I would add a Info log at the beginning and one once completed.

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+			return nil
+		}
+
+		// Top level directories are assumed to be user TSDBs
+		if info.IsDir() {
+
+			userID := info.Name()
+			f, err := os.Open(path)
+			if err != nil {
+				level.Error(util.Logger).Log("msg", "unable to open user TSDB dir", "err", err, "user", userID)
```
pracucci: What do you think if we also log the `path`, for a quicker understanding?

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+				return filepath.SkipDir
+			}
+			defer f.Close()
+
+			// If the dir is empty skip it
+			if _, err := f.Readdirnames(1); err != nil {
+				if err != io.EOF {
+					level.Error(util.Logger).Log("msg", "unable to read TSDB dir", "err", err, "user", userID)
```
pracucci: What do you think if we also log the `path`, for a quicker understanding?

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+					return filepath.SkipDir
+				}
+
+				// Empty dir
+				return filepath.SkipDir
+			}
+
+			// Limit the number of TSDB's opening concurrently
+			if err := openGate.Start(ctx); err != nil {
+				return err
```
pracucci: Shouldn't we log here as well?

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+				if err != io.EOF {
+					level.Error(util.Logger).Log("msg", "unable to read TSDB dir", "err", err, "user", userID)
+					return filepath.SkipDir
+				}
+
+				// Empty dir
+				return filepath.SkipDir
+			}
+
+			// Limit the number of TSDB's opening concurrently
```
pracucci: I would better explain that `Start()` will block until there's a free spot or the context is done (just to help the future reader).

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+			if err := openGate.Start(ctx); err != nil {
+				return err
+			}
+
+			wg.Add(1)
```
pracucci: I believe we should add before the call to `Start()`, otherwise you may potentially end up with a race condition where the routines have `Done()` and the gate hasn't started the new one yet (looks an edge case but can be potentially easy to reproduce with a concurrency of `1`).

pkg/ingester/ingester_v2.go
codesome: Okay

pkg/ingester/ingester_v2.go
codesome: The `filepath.Walk` call won't return till we have crossed this line, i.e. all TSDBs have initiated to open. So, that race condition would not happen I believe.

pracucci: You're right, but actually this comment made me realize that maybe we should move the `openGate.Start()` inside the go routine, so that we can continue iterating the filesystem and scheduling goroutines without being blocked by the gate?

codesome: I don't see any problem in being blocked by the gate as walking through the files would be fairly cheap compared to replaying the WAL. Additionally, moving it inside the goroutine will make it hard to handle any error caused by `openGate.Start()`. But I am on the fence here.

pracucci: Right. I was primarely thinking loudly.

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +461,63 @@ func (i *Ingester) closeAllTSDB() {
...
+			go func(userID string) {
+				defer wg.Done()
+				defer openGate.Done()
+				_, err := i.getOrCreateTSDB(userID, true) // force create the TSDB due to the lifecycler not having started yet
```
pracucci: I'm a bit confused. Given `getOrCreateTSDB()` wraps `tsdb.Open()` within the `userStatesMtx.Lock()`, and the WAL reply is done by `tsdb.Open()` before returning, how does the parallelization works and how did you hit the issue described in the PR anyway?

thorfour: Huh I totally missed that. The behavior I observed pointed to WAL replay eating memory, but you're right we can't be opening more than 1 at a time already. Guess I nees to try and recreate and get a memory profile.

pracucci: > get a memory profile

That would be very helpful. You can download an heap profile from the ingester's HTTP endpoint `/debug/pprof/`.

> I totally missed that

Please don't trash this PR. In any case I believe that pre-loading all TSDBs before an ingester is active is the desired behaviour, but right now it's not really parallelized.

pkg/storage/tsdb/config.go
```diff
@@ -32,6 +32,9 @@ type Config struct {
 	Backend      string            `yaml:"backend"`
 	BucketStore  BucketStoreConfig `yaml:"bucket_store"`
 
+	// MaxOpeningTSDBOnStartup limits the number of concurrently opening TSDB's during startup
+	MaxOpeningTSDBOnStartup int `yaml:"max_opening_tsdb_on_startup"`
```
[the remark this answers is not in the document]
thorfour: Fixed.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [FEATURE] Added support for Microsoft Azure blob storage to be used for storing chunk data. #1913
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
+* [ENHANCEMENT] Experimental TSDB: Open TSDB on startup to avoid large memory spikes with WAL replay immediately after joining. #1917
```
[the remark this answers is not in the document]
thorfour: Fixed.

pkg/ingester/ingester_v2.go
thorfour: Fixed.

pkg/ingester/ingester_v2.go
thorfour: Fixed

pkg/ingester/ingester_v2.go
thorfour: Added.

pkg/ingester/ingester_v2.go
thorfour: Added.

pkg/ingester/ingester_v2.go
thorfour: Don't think there's anything useful information to log here. It's a context error, and the caller likely should log the returned error.

pkg/ingester/ingester_v2.go
thorfour: Good call

pkg/ingester/ingester_v2.go
pracucci: You're right. All good then.

pkg/ingester/ingester_v2.go
pracucci: > In any case I believe that pre-loading all TSDBs before an ingester is active is the desired behaviour, but right now it's not really parallelized.

If end up to be complicated, we can also come up removing the concurrency option and pre-loading all TSDBs sequentially (which to me is better than now) and then add concurrency in a subsequent PR.

pkg/ingester/ingester_v2.go
thorfour: I think I plan to leave as is, because there doesn't seem to be much gain in changing it.

pkg/ingester/ingester_v2.go
thorfour: I refactored `getOrCreateTSDB` so that we can call `createTSDB` in the go routines and then only lock to add it to the db list once open.

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +471,68 @@ func (i *Ingester) closeAllTSDB() {
...
+			i.TSDBState.dbs[userID] = db
+
+		}(userID)
+
+		return filepath.SkipDir // Don't descend into directories
+	})
+
+	// Wait for all opening routines to finish
+	wg.Wait()
+	level.Info(util.Logger).Log("msg", "completed opening existing TSDBs")
```
pracucci: This log message will look weird in case `err != nil`. I would suggest to check `err` and log differently based on the case an error occurred or not.

pkg/ingester/ingester_v2.go
```diff
@@ -452,3 +471,68 @@ func (i *Ingester) closeAllTSDB() {
...
+			defer openGate.Done()
+			db, err := i.createTSDB(userID)
+			if err != nil {
+				level.Error(util.Logger).Log("msg", "unable to open user TSDB", "err", err, "user", userID)
+				return
+			}
+
+			// Add the database to the map of user databases
+			i.userStatesMtx.Lock()
+			defer i.userStatesMtx.Unlock()
```
pracucci: `defer` doesn't look to add much value here. We could just move the `Unlock()` after the next line.

pkg/ingester/ingester_v2_test.go
```diff
@@ -743,3 +744,85 @@ func newIngesterMockWithTSDBStorage(ingesterCfg Config, registerer prometheus.Re
 
 	return ingester, cleanup, nil
 }
+
+func TestIngester_V2Startup(t *testing.T) {
```
pracucci: I would name this test to better explain we're testing the loading of TSDB on startup, so something like `TestIngester_v2LoadTSDBOnStartup`

pkg/ingester/ingester_v2_test.go
```diff
@@ -743,3 +744,85 @@ func newIngesterMockWithTSDBStorage(ingesterCfg Config, registerer prometheus.Re
...
+	t.Parallel()
+
+	tests := []struct {
+		name  string
+		seed  func(string)
```
pracucci: Maybe calling this `setup` could be more clear.

pkg/ingester/ingester_v2_test.go
```diff
@@ -743,3 +744,85 @@ func newIngesterMockWithTSDBStorage(ingesterCfg Config, registerer prometheus.Re
 
 	return ingester, cleanup, nil
 }
+
+func TestIngester_V2Startup(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name  string
```
pracucci: You could remove the `name` just using a `tests := map[string]struct{}`. Resulting test cases may be more clear to read.

pkg/ingester/ingester_v2_test.go
```diff
@@ -743,3 +744,85 @@ func newIngesterMockWithTSDBStorage(ingesterCfg Config, registerer prometheus.Re
...
+			ingesterCfg.TSDBConfig.S3.Endpoint = "localhost"
+
+			// Seed the tsdbs dir
+			testData.seed(tempDir)
+
+			ingester, err := NewV2(ingesterCfg, clientCfg, overrides, nil)
+			require.NoError(t, err)
+
+			defer ingester.Shutdown()
+			defer os.RemoveAll(tempDir)
```
pracucci: This should be moved right after:

```
tempDir, err := ioutil.TempDir("", "tsdb")
require.NoError(t, err)
```

pkg/ingester/ingester_v2_test.go
```diff
@@ -743,3 +744,85 @@ func newIngesterMockWithTSDBStorage(ingesterCfg Config, registerer prometheus.Re
...
+
+			testData.check(ingester)
```
pracucci: I think you should pass `t` both to `seed()` and `check()` because the instance of `t` here is different than the outer one (look at `t.Run()` implementation for further details).

pkg/ingester/ingester_v2.go
thorfour: I personally always defer, since in the future someone might add something in between those locks that can panic (defensive programming). But I understand that right now it has no benefit. Changed.

pkg/ingester/ingester_v2.go
thorfour: Fixed.

pkg/ingester/ingester_v2_test.go
thorfour: Fixed

pkg/ingester/ingester_v2_test.go
thorfour: Fixed

pkg/ingester/ingester_v2_test.go
thorfour: Fixed

pkg/ingester/ingester_v2_test.go
thorfour: Fixed

pkg/ingester/ingester_v2_test.go
thorfour: Fixed

pkg/ingester/ingester_v2.go
```diff
@@ -445,3 +464,72 @@ func (i *Ingester) closeAllTSDB() {
...
+				return filepath.SkipDir
+			}
+
+			// Empty dir
+			return filepath.SkipDir
+		}
+
+		// Limit the number of TSDB's opening concurrently. Start blocks until there's a free spot available or the context is cancelled.
+		if err := openGate.Start(ctx); err != nil {
+			return err
```
codesome: Should this be `return filepath.SkipDir` too?

pkg/ingester/ingester_v2.go
```diff
@@ -445,3 +464,72 @@ func (i *Ingester) closeAllTSDB() {
...
+			level.Error(util.Logger).Log("msg", "unable to open user TSDB dir", "err", err, "user", userID, "path", path)
+			return filepath.SkipDir
+		}
+		defer f.Close()
+
+		// If the dir is empty skip it
+		if _, err := f.Readdirnames(1); err != nil {
+			if err != io.EOF {
+				level.Error(util.Logger).Log("msg", "unable to read TSDB dir", "err", err, "user", userID, "path", path)
+				return filepath.SkipDir
```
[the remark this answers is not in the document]
thorfour: Removed.

CHANGELOG.md
```diff
@@ -24,7 +24,9 @@
 * [BUGFIX] TSDB: Fixed querying ingesters in `LEAVING` state with the experimental TSDB blocks storage. #1854
 * [BUGFIX] TSDB: Fixed error handling in the series to chunks conversion with the experimental TSDB blocks storage. #1837
 * [BUGFIX] TSDB: Fixed TSDB creation conflict with blocks transfer in a `JOINING` ingester with the experimental TSDB blocks storage. #1818
-
+* [ENHANCEMENT] Experimental TSDB: Open existing TSDB on startup to prevent ingester from becoming ready before it can accept writes. #1917
+  * --experimental.tsdb.max-tsdb-opening-concurrency-on-startup
```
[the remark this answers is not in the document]
thorfour: Fixed.

pkg/ingester/ingester_v2.go
thorfour: No, because if the gate returns an error it means the context was cancelled and we should respect that and exit.

pracucci:
@thorfour May you fix the failing test, please?

thorfour:
@pracucci should be fixed.

pracucci:
@thorfour Other tests are failing now. May you take a look, please?

CHANGELOG.md
```diff
@@ -20,14 +20,16 @@
...
+  * `--experimental.tsdb.max-tsdb-opening-concurrency-on-startup`
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
 * [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
 * [BUGFIX] TSDB: Fixed handling of out of order/bound samples in ingesters with the experimental TSDB blocks storage. #1864
 * [BUGFIX] TSDB: Fixed querying ingesters in `LEAVING` state with the experimental TSDB blocks storage. #1854
 * [BUGFIX] TSDB: Fixed error handling in the series to chunks conversion with the experimental TSDB blocks storage. #1837
 * [BUGFIX] TSDB: Fixed TSDB creation conflict with blocks transfer in a `JOINING` ingester with the experimental TSDB blocks storage. #1818
 * [BUGFIX] TSDB: `experimental.tsdb.ship-interval` of <=0 treated as disabled instead of allowing panic. #1975
-
+* [BUGFIX] TSDB: ShipInterval of <=0 treated as disabled instead of allowing panic. #1975
```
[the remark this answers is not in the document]
thorfour: Done.

PR #1919, opened 2019-12-17
pkg/ring/kv/etcd: fix etcd watch api usage.

johncming:
1. There is no need to use backoff in etcd watch api. and missing `Wait` in current implements.
2. check watch response error .

### DIFF
--- a/pkg/ring/kv/etcd/etcd.go
+++ b/pkg/ring/kv/etcd/etcd.go
@@ -125,21 +125,35 @@ func (c *Client) CAS(ctx context.Context, key string, f func(in interface{}) (ou
 
 // WatchKey implements kv.Client.
 func (c *Client) WatchKey(ctx context.Context, key string, f func(interface{}) bool) {
-	watchChan := c.cli.Watch(ctx, key)
-	for resp := range watchChan {
-		if err := resp.Err(); err != nil {
-			level.Error(util.Logger).Log("msg", "watch chan error", "key", key, "err", err)
-		}
+	backoff := util.NewBackoff(ctx, util.BackoffConfig{
+		MinBackoff: 1 * time.Second,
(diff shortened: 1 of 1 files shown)


johncming:
> > There is no need to use backoff in etcd watch api. and missing Wait in current implements.
> 
> @johncming Could you explain why the backoff is not needed? There are some failure conditions in the etcd client `Watch()` for which a closed channel is returned. In such cases, we don't want to frantically loop over and - I believe - a backoff is desired.

Backoff is used for collision detection when get tokens, the first kv implementment consul has no watch api. Etcd watch api worked as subscription way.

tomwilkie:
The `Watch` API in KV store only expected to return when the context is cancelled; there are plenty of situations where the channel return by the etcd client's `Watch` function is closed (see https://github.com/etcd-io/etcd/blob/d6a3c995cf86b479cb5a44b48d000feb33e3d8f8/clientv3/watch.go#L332).  In this situation we want a retry loop, and it is right to use a backoff here in my opinion.

Does this behaviour cause problems for you @johncming?

pkg/ring/kv/etcd/etcd.go
```diff
@@ -129,13 +129,19 @@ func (c *Client) WatchKey(ctx context.Context, key string, f func(interface{}) b
 		MinBackoff: 1 * time.Second,
 		MaxBackoff: 1 * time.Minute,
 	})
+
+	// Ensure the context used by the Watch is always cancelled.
+	watchCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
pracucci: Why do we need this?

tomwilkie: The callback can return false causing us to exit the loop - we need to ensure the context used by the Watch is cancelled in this case.

PR #1921, opened 2019-12-17
Fix/lifecycler/stuck in leaving

thorfour:
**What this PR does**: I've observed ingesters that were evicted and then rescheduled. They were unable to gracefully shutdown and left their state in the ring as LEAVING. On reschedule the ingester found itself in the LEAVING state. Set itself to the LEAVING state and was then unable to join the cluster. 

This allows an ingester that finds itself in the LEAVING state with tokens to mark itself as ACTIVE.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/lifecycler.go
+++ b/pkg/ring/lifecycler.go
@@ -497,6 +497,12 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 			return ringDesc, true, nil
 		}
 
+		// If the ingester failed to clean it's ring entry up in can leave it's state in LEAVING.
+		// Move it into ACTIVE to ensure the ingester joins the ring.
+		if ingesterDesc.State == LEAVING && len(ingesterDesc.Tokens) == i.cfg.NumTokens {
+			ingesterDesc.State = ACTIVE
+		}
+
 		// We exist in the ring, so assume the ring is right and copy out tokens & state out of there.
 		i.setState(ingesterDesc.State)
 		tokens, _ := ringDesc.TokensFor(i.ID)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [FEATURE] Added support for Microsoft Azure blob storage to be used for storing chunk data. #1913
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
+* [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
   
 ## 0.4.0 / 2019-12-02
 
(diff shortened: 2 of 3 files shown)


pkg/ring/lifecycler.go
```diff
@@ -354,6 +354,7 @@ func (i *Lifecycler) loop() {
 
 loop:
 	for {
+
```
codesome: Unnecessary

pkg/ring/lifecycler.go
```diff
@@ -497,6 +498,12 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 			return ringDesc, true, nil
 		}
 
+		// If the ingester failed to clean it's ring entry up in can leave it's state in LEAVING.
+		// Move it into ACTIVE to ensure the ingester joins the ring.
+		if ingesterDesc.State == LEAVING && len(ingesterDesc.Tokens) != 0 {
+			ingesterDesc.State = ACTIVE
```
codesome: We should not set it to `ACTIVE` till we have confirmed that we have enough tokens as set in the config.

codesome: (I think so looking at the existing code, not sure if it is the requirement)

pstibrany: I am not sure if PENDING would be correct, as pending ingesters can be contacted to take over tokens from another ingester. We could use JOINING state, but that requires extra code to call autoJoin from this state (or using observe period flag).

I think ACTIVE is fine, if we don't care too much about number of tokens (I don't think it's a big deal) or it hasn't changed between restarts.

pkg/ring/lifecycler.go
thorfour: Typo. Fixed.

pkg/ring/lifecycler.go
thorfour: I've updated it to check for NumTokens instead of just non-zero amount of tokens. Though I'm not sure it would make much of a difference.

codesome: Given that we don't change number of tokens often, this should not make a difference. If you increase the tokens between rollout, this might need a different handling (but not sure if we need to handle that in this PR).

khaines:
This will certainly help the statefulset ingester setup when recovering from an unclean shutdown. Thank you!

canghai118:
If ingester in deploy use k8s deployment,  every time ingester recreate by k8s,  it will register itself with different hostname and address.   the zombie ingester will stay in ring for ever,  and finally the clients try to read/write will report `too many failed ingesters`.  

If the ingester is kill by oom-killer, it won't start the quit process, and will leave state in `ACTIVE`, I just encounter this problems on ruler #1956 

By the way ,  after the success transfer or flush, ingester will [sleep for `-ingester.final-sleep`](https://github.com/cortexproject/cortex/blob/v0.4.0/pkg/ring/lifecycler.go#L673),  after that it unregister it self from ring. Default value for `-ingester.final-sleep` is 30seconds, the default  grace termination period for k8s pod is 30s, it will probably cause trouble if both parameter use default value.
 
the same problem happens on ruler two.

May be we need a away to cleanup dead ingester from ring.

@thorfour @khaines @pstibrany

thorfour:
@canghai118 I'm not sure there's a safe way to determine if an ingester is a "zombie" or if it's just crash looping and will come back eventually. I'd worry that trying to perform cleanups on what we believe to be a zombie ingester might cause more problems than it solves.

PR #1929, opened 2019-12-19
Ignore the queries in the future in the ingesters

gouthamve:
We ignore queries in the future now.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/ingester/ingester.go
```diff
@@ -459,6 +459,20 @@ func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_
...
+	if from.After(now) {
+		// time-span start is in future ... regard as legal
+		level.Info(log).Log("msg", "whole timerange in future, yield empty resultset", "through", through, "from", from, "now", now)
+		return nil
+	}
+
+	maxFutureThrough := now.Add(5 * time.Minute)
+	if through.After(maxFutureThrough) {
+		// time-span end is in future ... regard as legal
+		level.Info(log).Log("msg", "adjusting end timerange from future to now", "old_through", through, "new_through", maxFutureThrough)
```
cstyan: in the message should we say `from future to now+5 minutes` just to be explicitly clear?

pkg/ingester/ingester.go
```diff
@@ -459,6 +459,20 @@ func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_
 		return err
 	}
 
+	now := model.Now()
+	if from.After(now) {
+		// time-span start is in future ... regard as legal
```
codesome: What is `regard as legal`?

pkg/ingester/ingester.go
```diff
@@ -459,6 +459,20 @@ func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_
...
+		level.Info(log).Log("msg", "whole timerange in future, yield empty resultset", "through", through, "from", from, "now", now)
+		return nil
+	}
+
+	maxFutureThrough := now.Add(5 * time.Minute)
```
pracucci: I think would be safe if replace `5 * time.Minute` with a shared constant between ingester and chunk store, to keep consistency.

pracucci:
> Why don't we enforce it in the querier, instead of duplicating the logic between the chunk store and ingester?

I'm talking specifically about `querier.NewQueryable()` and `querier.newUnifiedChunkQueryable()`. To me looks more natural to validate the query time range at the beginning of the read path.

bboreham:
My sense is that the chunk store does it to avoid generating error messages about "missing tables" when the problem is in the query parameters.
I would agree that the querier is a better place to shortcut, but not sure if I would remove the protection entirely from the chunk store for the above reason.

pkg/chunk/chunk_store.go
```diff
@@ -281,10 +281,11 @@ func (c *store) validateQueryTimeRange(ctx context.Context, userID string, from
...
 	}
 
-	if through.After(now.Add(5 * time.Minute)) {
+	maxFutureThrough := now.Add(5 * time.Minute)
+	if through.After(maxFutureThrough) {
 		// time-span end is in future ... regard as legal
-		level.Info(log).Log("msg", "adjusting end timerange from future to now", "old_through", through, "new_through", now)
-		*through = now // Avoid processing future part - otherwise some schemas could fail with eg non-existent table gripes
+		level.Info(log).Log("msg", "adjusting end timerange from future to now", "old_through", through, "new_through", maxFutureThrough)
+		*through = maxFutureThrough // Avoid processing future part - otherwise some schemas could fail with eg non-existent table gripes
```
owen-d: There's no guarantee that after this operation `from` < `through`. If we specify `from: now().Add(10min)` and `through: now().Add(15min)`, it'd would violate the guarantee we're erroring on earlier (that `from` < `through`). 

It probably needs another check/error condition here like `errors.New("Adjusted end time would be before start time")`

owen-d:
What about wrapping the distributor queryable in order to short circuit these before they call the ingesters? Maybe here:
https://github.com/cortexproject/cortex/blob/master/pkg/querier/querier.go#L117

pkg/querier/querier.go
```diff
@@ -97,8 +99,21 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
+			maxQueryTime := util.TimeMilliseconds(now.Add(maxQueryIntoFuture))
+
+			if mint > maxQueryTime {
+				return storage.NoopQuerier(), nil
+			}
+			if maxt > maxQueryTime {
+				maxt = maxQueryTime
+			}
+		}
+
```
cstyan: is there somewhere in the querier package for this to live so it doesn't have to be duplicated in `newUnifiedChunkQueryable`?

pkg/querier/unified_querier.go
```diff
@@ -10,12 +10,26 @@ import (
 	"github.com/weaveworks/common/user"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/util"
 )
 
-func newUnifiedChunkQueryable(ds, cs ChunkStore, distributor Distributor, chunkIteratorFunc chunkIteratorFunc, ingesterMaxQueryLookback time.Duration) storage.Queryable {
+func newUnifiedChunkQueryable(ingesterStore, chunkStore ChunkStore, distributor Distributor, chunkIteratorFunc chunkIteratorFunc, ingesterMaxQueryLookback, maxQueryIntoFuture time.Duration) storage.Queryable {
```
cstyan: :+1:  for more descriptive names on the `ChunkStore` parameters

pkg/util/time.go
```diff
@@ -0,0 +1,7 @@
+package util
+
+import "time"
+
+func TimeMilliseconds(t time.Time) int64 {
+	return t.UnixNano() / int64(time.Millisecond/time.Nanosecond)
```
pracucci: `time.Nanosecond` is `1` so it's enough to do `/ int64(time.Millisecond)`.

pkg/querier/querier_test.go
```diff
@@ -228,6 +228,93 @@ func TestNoHistoricalQueryToIngester(t *testing.T) {
...
+			maxt:               time.Now().Add(-59 * time.Minute),
+			hitStores:          true,
+			maxQueryIntoFuture: 10 * time.Minute,
+		},
+		{ // Skipping stores is disabled.
+			name:               "hit-test2",
+			mint:               time.Now().Add(-5 * time.Hour),
+			maxt:               time.Now().Add(-50 * time.Minute),
+			hitStores:          true,
+			maxQueryIntoFuture: 0,
```
pracucci: In this test you would hit the store even if `maxQueryIntoFuture` is enabled, because the `mint` and `maxt` are in the past. You should set them in the future to assert that when `maxQueryIntoFuture` is disabled the stores will be hit anyway.

pkg/querier/querier_test.go
```diff
@@ -228,6 +228,93 @@ func TestNoHistoricalQueryToIngester(t *testing.T) {
...
+		},
+	}
+
+	engine := promql.NewEngine(promql.EngineOpts{
+		Logger:        util.Logger,
+		MaxConcurrent: 10,
+		MaxSamples:    1e6,
+		Timeout:       1 * time.Minute,
+	})
+	cfg := Config{}
```
pracucci: It's always a good practice to initialize it with defaults. You can just do `flagext.DefaultValues(&cfg)`.

pkg/util/time.go
gouthamve: It's not something we want to rely on. It's an implementation detail and this is already going to be optimised by the compiler anyways.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
     * `server.grpc-max-concurrent-streams` (for both query-frontends and queriers)
 * [CHANGE] The frontend http server will now send 502 in case of deadline exceeded and 499 if the user requested cancellation. #2156
 * [CHANGE] Config file changed to remove top level `config_store` field in favor of a nested `configdb` field. #2125
+* [CHANGE] We now ignore queries that are more than 10mins into the future. #1929
+  * Can be configured using `querier.max-query-into-future`.
```
pracucci: I'm wondering if it would be more clear:

```suggestion
* [CHANGE] We now enforce queries to be up to `-querier.max-query-into-future` into the future (defaults to 10m). #1929
```

Because we don't "ignore queries" but we actually clamp the max timestamp.

pkg/querier/querier_test.go
```diff
@@ -237,6 +237,96 @@ func TestNoHistoricalQueryToIngester(t *testing.T) {
...
+		},
+		{
+			name:               "hit-test2",
+			mint:               time.Now().Add(-5 * time.Hour),
+			maxt:               time.Now().Add(-59 * time.Minute),
+			hitStores:          true,
+			maxQueryIntoFuture: 10 * time.Minute,
+		},
+		{ // Skipping stores is disabled.
+			name:               "hit-test2",
```
pracucci: Clashing name with the previous one. The name may be something like "max-query-into-future-disabled".

PR #1934, opened 2019-12-23
Added /ready endpoint to queriers

pracucci:
**What this PR does**:
The querier - when running with the TSDB experimental storage - does the initial sync of all TSDB blocks' index header, which may take some time (order of minutes). Until successfully completed, it will not be ready to serve queriers.

In this PR:
- Add `/ready` endpoint to the querier, which always return `204`, once the initialization has been done
- Start the query frontend worker only after the store and querier have been successfully initialized

_N/A_

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [FEATURE] Added support for Microsoft Azure blob storage to be used for storing chunk data. #1913
+* [FEATURE] Added readiness probe endpoint`/ready` to queriers. #1934
 * [ENHANCEMENT] Added `password` and `enable_tls` options to redis cache configuration. Enables usage of Microsoft Azure Cache for Redis service.
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
 * [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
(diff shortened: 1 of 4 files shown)


pracucci:
The failing `lint` looks unrelated.

pkg/cortex/modules.go
```diff
@@ -236,6 +231,20 @@ func (t *Cortex) initQuerier(cfg *Config) (err error) {
...
+	// Start the query frontend worker once the query engine and the store
+	// have been successfully initialized.
+	t.worker, err = frontend.NewWorker(cfg.Worker, httpgrpc_server.NewServer(t.server.HTTPServer.Handler), util.Logger)
+	if err != nil {
+		return
+	}
+
+	// Once the execution reaches this point, all synchronous initialization has been
+	// done and the querier is ready to serve queries, so we're just returning a 202.
+	t.server.HTTP.Path("/ready").Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
thorfour: `nit` I think I'd rather see it return something like a 503 before its ready, instead of timing out. And then switch to the 204 once it gets here. But I don't think it's a big deal

pracucci: Yeah. The reason why I didn't do it, is because the HTTP server is not opened until the `cortex.Run()` is executed, which occurs after all the modules have been initialized.

The cleaner solution would be not having any synchronous initialization in the modules initialization (so that modules are fast to initialize), while making sure we do start the querier worker only after the asynchronous initial sync is completed, but this would require some significative refactoring which I'm not sure it's worth at this stage.

Thoughts?

thorfour: Definitely not worth it right now. This change is good

pracucci:
The failed lint looks unrelated:

```
go: finding sigs.k8s.io/structured-merge-diff v0.0.0-20190525122527-15d366b2352e
github.com/apache/thrift@v0.12.0: stream error: stream ID 609; INTERNAL_ERROR
Makefile:140: recipe for target 'mod-check' failed
make: *** [mod-check] Error 1

Exited with code exit status 2
```

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [FEATURE] Added support for Microsoft Azure blob storage to be used for storing chunk data. #1913
+* [FEATURE] Added readiness probe endpoint`/ready` to queriers. #1934
+* [ENHANCEMENT] Start the query frontend worker - in the querier target - after the querier has been successfully initialized. #1934
```
gouthamve: This is the internals and has no operator impact. Shoudn't be in the changelog.

PR #1940, opened 2019-12-27
Changed the default value for -distributor.ha-tracker.prefix in order to not clash with other keys

### LINKED ISSUE #1936: error decoding list of values for prefix:key
#docker-compose.yml 

> services:
  cortex:
    image: cortexproject/cortex:v0.4.0
    restart: always
    hostname: prometheus-db-1
    container_name: cortex
    ports:
      - "9009:9009"
      - "9095:9095"
    volumes:
      - "/opt/data/cortex/data:/data"
      - "./config/single-process-config.yaml:/etc/single-process-config.yaml"
    command:
      - "-config.file=/etc/single-process-config.yaml"
      - "-ring.store=consul"
      - "-consul.hostname=consul-cluster:8500"
      - "-distributor.replication-factor=3"
      - "-distributor.ha-tracker.enable"
      - "-distributor.ha-tracker.enable-for-all-users"
      - "-distributor.ha-tracker.cluster=cluster"
      - "-distributor.ha-tracker.replica=replica"
      - "-distributor.ha-tracker.consul.hostname=consul-cluster:8500"

#./config/single-process-config.yaml as like [https://github.com/kanuahs/cortex-demo/blob/master/docker-demo/cortex-configs/single-process-config.yaml](url)

**why does it reports the following error?**

`level=error ts=2019-12-26T10:03:54.746577132Z caller=client.go:275 msg="error decoding list of values for prefix:key" prefix=collectors/ key=collectors/ring err="proto: wrong wireType = 2 for field ReceivedAt"`

pracucci:
**What this PR does**:

Using the default flags/config, the HA Tracker clashes with the ring if the both share the same KVStore backend (ie. same consul cluster). This is due to the fact that the HA Tracker watches all keys on prefix `""` and decode each of them, but can't decode the ring one.

In this PR I'm proposing to change the default value of the HA Tracker KVStore prefix, in order to not have them clashing just using the default settings out of the box.

Fixes #1936

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci:
@gouthamve I've rebased it to fix conflicts.

PR #1942, opened 2019-12-31
Introduce TSDB blocks compactor

pracucci:
**What this PR does**:
Following this [design doc](https://docs.google.com/document/d/1U7SWi44pLsZrI2awWbz7gurn8kBHL7bmww3943mFC_w/edit), this PR introduces a new component - called `compactor` - which uses the Thanos bucket compactor to compact TSDB blocks stored in the bucket.

I've got the compactor running in a dev cluster and - as expected - it significantly reduce the memory used by the querier, due to better optimized Thanos index headers. For example, if the following chart (showing queriers memory usage), the compactor dropped from about 1.8K to 50 (1 per day, 50 days retention so far):

Screen Shot 2019-12-31 at 12 24 27

There are few things left out from this PR, that will be addressed in subsequent PRs:
- Horizontal scalability (run 2+ compactors at the same time)
- Export Thanos compactor metrics (see `TODO` in `pkg/compactor/compactor.go`)

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/compactor/compactor.go
```diff
@@ -0,0 +1,284 @@
...
+	c.compactUsersWithRetries(c.ctx)
+
+	ticker := time.NewTicker(c.compactorCfg.CompactionInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			c.compactUsersWithRetries(c.ctx)
+		case <-c.quit:
```
thorfour: could we get rid of the quit channel and simply use `<- ctx.Done()` ?

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,284 @@
...
+	users, err := c.discoverUsers(ctx)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to discover users from bucket", "err", err)
+		return err
+	}
+	level.Info(c.logger).Log("msg", "discovered users from bucket", "users", len(users))
+
+	for _, userID := range users {
+		// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
+		if ctx.Err() != nil {
```
thorfour: this statement looks funny to my eyes. Can this be a switch with `<- ctx.Done()` ?

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,284 @@
...
+
+	return nil
+}
+
+func (c *Compactor) compactUser(ctx context.Context, userID string) error {
+	bucket := cortex_tsdb.NewUserBucketClient(userID, c.bucketClient)
+
+	syncer, err := compact.NewSyncer(
+		c.logger,
+		nil, // TODO(pracucci) we should pass the prometheus registerer, but we would need to inject the user label to each metric, otherwise we have clashing metrics
```
thorfour: I've found this to be troublesome since we likely don't want per-user metrics anyways but want a rollup of all the metrics across users.

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,284 @@
...
+		[]*relabel.Config{})
+	if err != nil {
+		return errors.Wrap(err, "failed to create syncer")
+	}
+
+	compactor, err := compact.NewBucketCompactor(
+		c.logger,
+		syncer,
+		c.tsdbCompactor,
+		path.Join(c.compactorCfg.DataDir, "compact"),
```
thorfour: does this end up writing blocks to `data/compact` ?, and should it have the userID prefixed in there?

pkg/compactor/compactor.go
pracucci: Definitely, done.

pkg/compactor/compactor.go
pracucci: You mean changing it to something like the following? If yes, what's the real benefit? To me it looks we're writing the same thing using more lines of code.

```
select {
case <-c.ctx.Done():
  return c.ctx.Err()
default:
}
```

pkg/compactor/compactor.go
pracucci: I see your point. Don't have an answer yet. As stated in the PR description, I would prefer to defer it to a subsequent PR, to not block the compactor because of this.

What's your take? Do you have any idea to solve this?

pkg/compactor/compactor.go
pracucci: > does this end up writing blocks to data/compact ?

Yes

> and should it have the userID prefixed in there?

It's not necessary, cause the bucket compactor clean up the dir at the beginning (and end) of each `Compact()`.

pkg/compactor/compactor.go
thorfour: Yea it just felt more idiomatic to me to write it that way. But I have no qualms leaving it the way it is.

pkg/compactor/compactor.go
thorfour: Oh definitely don't block this PR for that. I too punted on a solution for that for the other user wrapped thanos components. Just wanted to add my thoughts to the comment. 

I'm not sure there can be a good solution for this without changing the Thanos code to only call the register function once.

pkg/compactor/compactor.go
thorfour: Okay great!

pkg/compactor/compactor.go
pracucci: An option - in Thanos - would be exposing `compact.syncerMetrics` and picking its instance in input in `compact.NewSyncer()` so what we could create it once in Cortex and pass the same `syncerMetrics` instance to multiple `Syncer`.

@bwplotka would you see feasible having such refactoring in Thanos, just to help Cortex? Would you see a better way to do it?

bwplotka: We want to treat any Thanos package as a library, so if the use case of a struct in our package makes sense, then we are ok with it. (: 

We can definitely allow passing metrics for syncer. Also, note that we changed syncer a bit and introduced `block.MetadataFetcher`. I assume you use different syncers because of many buckets?

pkg/compactor/compactor.go
pstibrany: Personally I would only use `select` version if there were more things to select on. `ctx.Err()` returns non-nil value only when context is finished, so I think this is fine.

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,281 @@
...
+	)
+	if err != nil {
+		return errors.Wrap(err, "failed to create bucket compactor")
+	}
+
+	return compactor.Compact(ctx)
+}
+
+func (c *Compactor) discoverUsers(ctx context.Context) ([]string, error) {
+	users := make([]string, 0)
```
codesome: Nit. Just `var users []string` would suffice. I am not sure how Go works in this case, but I vaguely remember that there are no allocations if just declared and not appended to it yet (also, it looks cleaner).

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,281 @@
...
+	retries := util.NewBackoff(ctx, util.BackoffConfig{
+		MinBackoff: c.compactorCfg.retryMinBackoff,
+		MaxBackoff: c.compactorCfg.retryMaxBackoff,
+		MaxRetries: c.compactorCfg.CompactionRetries,
+	})
+
+	c.compactionRunsStarted.Inc()
+
+	for retries.Ongoing() {
+		if err := c.compactUsers(ctx); err == nil {
```
codesome: The error should not be ignored here I think though we are logging it inside the function (context error is not logged). To foolproof it for the future, I would prefer wrapping the error inside `compactUsers` and logging error here.

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,281 @@
...
+	compactor, err := compact.NewBucketCompactor(
+		c.logger,
+		syncer,
+		c.tsdbCompactor,
+		path.Join(c.compactorCfg.DataDir, "compact"),
+		bucket,
+		// No compaction concurrency. Due to how Cortex works we don't
+		// expect to have multiple block groups per tenant, so setting
+		// a value higher than 1 would be useless.
+		1,
```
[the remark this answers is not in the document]
pracucci: The block groups here are the Thanos one.

Blocks are grouped based on the external labels (a Thanos thing). For each group, all their block meta files are written in a temporary directory and then the TSDB compactor `Plan()` will run and finally the compaction based on the result of the plan.

For each Thanos blocks group (based on external labels) there are no multiple compactions running at the same time. Given we have only 1 Thanos blocks group per tenant, having a value higher than 1 wouldn't improve performances.

pkg/compactor/compactor.go
pracucci: Few notes:
1. It's logged in `compactUsers()`
2. The reason why I haven't wrapped it and logged here it because the `compactUsers()` already log several `Info()` and - to me - looks weird reading that function code and seeing that info are logged and error not

> context error is not logged

Good point. Added a log.

pkg/compactor/compactor.go
pracucci: Changed even if such kind of changes for any non-hot path are not really a valuable thing.

pkg/compactor/compactor.go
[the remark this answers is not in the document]
codesome: Oh, okay. I was thinking of a case where there would be multiple plans and they could be run in parallel. But given how `Plan()` works, I guess we cannot get remaining plans before the current plan is compacted. Sounds fine then. (I am missing the "Mark Resolved" button on GitHub)

pracucci:
@thorfour @pstibrany @codesome May you take another look and eventually approve if you don't have further comments and you believe the current PR is good to be merged, please? It would help as a signal for maintainers.

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,282 @@
...
+	tsdbCompactor, err := tsdb.NewLeveledCompactor(ctx, registerer, logger, compactorCfg.BlockRanges.ToMillisecond(), downsample.NewPool())
+	if err != nil {
+		cancelCtx()
+		return nil, errors.Wrap(err, "failed to create TSDB compactor")
+	}
+
+	return newCompactor(ctx, cancelCtx, compactorCfg, storageCfg, bucketClient, tsdbCompactor, logger, registerer)
+}
+
+func newCompactor(ctx context.Context, cancelCtx context.CancelFunc, compactorCfg Config, storageCfg cortex_tsdb.Config, bucketClient objstore.Bucket, tsdbCompactor tsdb.Compactor, logger log.Logger, registerer prometheus.Registerer) (*Compactor, error) {
```
codesome: Would it be more readable on the editor if all arguments are in separate lines?

pracucci: Definitely. Done.

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,291 @@
...
+	retryMaxBackoff time.Duration `yaml:"-"`
+}
+
+// RegisterFlags registers the Compactor flags.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.BlockRanges = cortex_tsdb.DurationList{2 * time.Hour, 12 * time.Hour, 24 * time.Hour}
+	cfg.retryMinBackoff = 10 * time.Second
+	cfg.retryMaxBackoff = time.Minute
+
+	f.Var(
```
pstibrany: We don't use this style of flags definitions anywhere else.

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,291 @@
...
+		&cfg.BlockRanges,
+		"compactor.block-ranges",
+		"Comma separated list of compaction ranges expressed in the time duration format")
+
+	f.DurationVar(
+		&cfg.ConsistencyDelay,
+		"compactor.consistency-delay",
+		30*time.Minute,
```
pstibrany: Is 30 minutes related to `compact.MinimumAgeForRemoval`? Should we use that constant here?

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,291 @@
...
+
+	c.compactionRunsFailed.Inc()
+}
+
+func (c *Compactor) compactUsers(ctx context.Context) error {
+	level.Info(c.logger).Log("msg", "discovering users from bucket")
+	users, err := c.discoverUsers(ctx)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to discover users from bucket", "err", err)
+		return err
```
pstibrany: Since error is not used by the caller, can we return bool instead? It's just weird to see error being logged, but also returned.

pracucci:
Integration tests are failing because of this unrelated issue (looks a temporary networking issue):
```
Unable to find image 'quay.io/freshtracks.io/avalanche:master-2019-07-28-47b00e3' locally
docker: Error response from daemon: Get https://quay.io/v2/freshtracks.io/avalanche/manifests/master-2019-07-28-47b00e3: net/http: TLS handshake timeout.
```

pkg/storage/tsdb/config.go
```diff
@@ -48,31 +48,38 @@ type Config struct {
...
-// ToMillisecondRanges returns the duration list in milliseconds
-func (d DurationList) ToMillisecondRanges() []int64 {
-	ranges := make([]int64, 0, len(d))
-	for _, t := range d {
-		ranges = append(ranges, int64(t/time.Millisecond))
+// ToMillisecond returns the duration list in milliseconds
+func (d *DurationList) ToMillisecond() []int64 {
+	values := make([]int64, 0, len(*d))
+	for _, t := range *d {
+		values = append(values, int64(t/time.Millisecond))
```
pstibrany:
```suggestion
		values = append(values, t.Milliseconds())
```

pkg/storage/tsdb/config.go
```diff
@@ -48,31 +48,38 @@ type Config struct {
...
+		values = append(values, v.String())
+	}
+
+	return strings.Join(values, ",")
+}
+
+// Set implements the flag.Value interface
+func (d *DurationList) Set(s string) error {
+	values := strings.Split(s, ",")
+	*d = make([]time.Duration, 0, len(values)) // flag.Parse may be called twice, so overwrite instead of append
```
pstibrany: This is tricky. flag package supports calling same parameter multiple times, in which case it will also call `Set` method multiple times. More flag-compatible way of using it would therefore be: `-compactor.block-ranges=2h -compactor.block-ranges=48h`. It's ok if we want to support list flag, but we should probably still append ranges from multiple `Set` calls.

In either case, should these ranges be validated for being in correct order, before they are used?

pkg/storage/tsdb/config.go
```diff
@@ -48,31 +48,38 @@ type Config struct {
...
 	}
 	return nil
 }
 
-// ToMillisecondRanges returns the duration list in milliseconds
-func (d DurationList) ToMillisecondRanges() []int64 {
-	ranges := make([]int64, 0, len(d))
-	for _, t := range d {
-		ranges = append(ranges, int64(t/time.Millisecond))
+// ToMillisecond returns the duration list in milliseconds
```
pstibrany: nit: Should be plural? It's not just one millisecond, it's slice of milliseconds.

pkg/compactor/compactor.go
pracucci: @codesome Commented on this too, so I guess I can't ignore it 😉

pkg/compactor/compactor.go
pracucci: They're two different things.

The consistency delay is the minimum time of a block before it's considered for compaction and garbage collection. If this delays is less than 30m, then in the Thanos garbage collector the min delay is enforced to `30m` (`MinimumAgeForRemoval`).

pkg/storage/tsdb/config.go
pracucci: This way done by @thorfour. I've just fixed it because didn't work the CLI flags parsing (was passed by copy and not reference).

@thorfour What's the use case because you did this? I'm missing some context to safely change it.

pracucci:
Thanks @pstibrany for your review. I've addressed all the comments, except the one about `DurationList.Set()` because I'm missing a bit of context on the reason of the current logic.

pkg/storage/tsdb/config.go
thorfour: I guess I was hedging a little too much against obviousness of behaviour. I thought if you had multiple flags with duration, it wasn't obvious in which order they are applied. But looking back on it now, it's pretty obvious they'll be sorted by time. I think it's probably a good change to get rid of that, and just use multiple flags if desired.

thorfour: The clearing of the duration list was because the flag parse function is called twice from the main package, which was resulting in a double list being created, despite the user only submitting a single list.

pstibrany: I like single comma-separated parameter option as well, but think that multiple parameters should be supported, and "append" is a correct option to use here, instead of overwrite.

About sorting -- where does that happen? I haven't found validation nor sorting (but haven't looked that hard)

pstibrany: > The clearing of the duration list was because the flag parse function is called twice from the main package, which was resulting in a double list being created, despite the user only submitting a single list.

Oh really? That would actually explain one weird thing in memberlist implementation I was seeing and never connected to flags being parsed twice. We should fix that, if possible.

thorfour: Sorry, was kind of talking out loud. If we moved away from a comma separated parameter list and just did a single time duration flag that could be passed multiple times, we'd want to sort all of those values so we didn't end up with block ranges that made no sense. 

I think you're right in that "append" is a correct behaviour if given two flags, it just gets weird when `flag.Parse` is called twice from main and there's only one flag.

pstibrany: I will send PR that fixes flag.Parse() being called twice on same config soon.

pracucci: This discussion is very interesting. I've seen the CLI flags are parsed twice in the `main.go` but I'm not sure about the reason (I agree should be fixed, not sure about the implications). In any case looks out of the scope of this PR, so I would keep it for a separate PR.

pstibrany: The only reason for double-parsing is that we need -config.file option to parse config file, and then reparse command line flags so that they take precendence. Easy to fix. I'll send separate PR. It came up here, because this `Set` method is using incorrect semantics (replacing instead of appending flags).

pstibrany: I've sent #1997 now.

pkg/compactor/compactor.go
```diff
@@ -0,0 +1,263 @@
...
+	retryMaxBackoff time.Duration `yaml:"-"`
+}
+
+// RegisterFlags registers the Compactor flags.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.BlockRanges = cortex_tsdb.DurationList{2 * time.Hour, 12 * time.Hour, 24 * time.Hour}
+	cfg.retryMinBackoff = 10 * time.Second
+	cfg.retryMaxBackoff = time.Minute
+
+	f.Var(&cfg.BlockRanges, "compactor.block-ranges", "Comma separated list of compaction ranges expressed in the time duration format")
```
jtlisi: Could you add the default values to the help text?

pracucci: There actually no need. When the `--help` is generated, defaults are displayed reading them from the values set:

```
 -compactor.block-ranges value
        Comma separated list of compaction ranges expressed in the time duration format (default 2h0m0s,12h0m0s,24h0m0s)
```

jtlisi: Awesome

pkg/storage/tsdb/config.go
pstibrany: Now that #1997 is merged (with subsequent fix as well), it would make sense to use `*d = append(*d, values...)` here, to make it work like standard flag. What do you think?

pstibrany: Ah, I guess I'm too late. I can send separate PR later. :) Great to see this merged!

pracucci: I didn't wait to merge this PR before addressing this because it's not a regression introduced by this PR. I would be glad if you could take care of that 🙏

PR #1945, opened 2020-01-02
Updated Cortex architecture doc and documented the experimental blocks storage

pracucci:
**What this PR does**:
In this PR:
- Added missing TSDB-related changes (already merged) to changelog
- Updated Cortex architecture doc 
- Documented the experimental blocks storage

**Notes to the reviewer:**

- I suggest to read the updated architecture doc entirely and not from the diff, to have a better overview of how has been covered
- Mentioned the WAL, since I do expect to be merged soon.
- Mentioned the blocks compactor, which I hope to see merged soon (PR https://github.com/cortexproject/cortex/pull/1942).
- I've specified the blocks storage config giving preference to YAML and mentioning each related CLI flag. I've used the Prometheus config documentation syntax (also used by Loki). If there are no objection, I would like to submit a subsequent PR to document the entire Cortex config the same way.

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -28,15 +28,15 @@ The two engines mostly share the same Cortex architecture with few differences o
 
 ### Chunks storage (default)
 
-The chunks storage stores each single time series into a separate object called _chunk_. Each chunk contains the timestamp-value pairs for a given period (defaults to 12 hours) of the series. chunks are then indexed by time range and labels, in order to provide a fast lookup across many (over millions) chunks.
+The chunks storage stores each single time series into a separate object called _Chunk_. Each Chunk contains the timestamp-value pairs for a given period (defaults to 12 hours) of the series. Chunks are then indexed by time range and labels, in order to provide a fast lookup across many (over millions) Chunks.
 
 For this reason, the chunks storage consists of:
 
-* An index for the chunks. This index can be backed by:
+* An index for the Chunks. This index can be backed by:
     * [Amazon DynamoDB](https://aws.amazon.com/dynamodb)
     * [Google Bigtable](https://cloud.google.com/bigtable)
     * [Apache Cassandra](https://cassandra.apache.org)
-* An object store for the chunk data itself, which can be:
+* An object store for the Chunk data itself, which can be:
     * [Amazon DynamoDB](https://aws.amazon.com/dynamodb)
     * [Google Bigtable](https://cloud.google.com/bigtable)
     * [Apache Cassandra](https://cassandra.apache.org)
@@ -51,11 +51,11 @@ The current schema recommendation is the **v10 schema** (v11 is still experiment
 
 ### Blocks storage (experimental)
 
-The blocks storage is based on [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/storage/): it stores each tenant's time series into their own TSDB which write out their series to a on-disk block (defaults to 2h block range periods). Each block is composed by few files storing the chunks and the block index.
+The blocks storage is based on [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/storage/): it stores each tenant's time series into their own TSDB which write out their series to a on-disk Block (defaults to 2h block range periods). Each Block is composed by few files storing the chunks and the block index.
 
-The TSDB chunk files contain the timestamp-value pairs for multiple series. The series inside the chunks are then indexed by a per-block index, which indexes metric names and labels to time series in the chunk files.
+The TSDB chunk files contain the timestamp-value pairs for multiple series. The series inside the Chunks are then indexed by a per-block index, which indexes metric names and labels to time series in the chunk files.
 
-The blocks storage doesn't require a dedicated storage backend for the index. The only requirement is an object store for the block files, which can be:
+The blocks storage doesn't require a dedicated storage backend for the index. The only requirement is an object store for the Block files, which can be:
 
 * [Amazon S3](https://aws.amazon.com/s3)
 * [Google Cloud Storage](https://cloud.google.com/storage/)
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -28,7 +28,7 @@ The two engines mostly share the same Cortex architecture with few differences o
 
 ### Chunks storage (default)
 
-The chunks storage stores each single time series into a separate object called chunk. Each chunk contains the timestamp-value pairs for a given period (defaults to 12 hours) of the series. Series chunks are then indexed by time range period and metric labels, in order to provide a fast lookup across millions - or even billions - of chunks.
+The chunks storage stores each single time series into a separate object called _chunk_. Each chunk contains the timestamp-value pairs for a given period (defaults to 12 hours) of the series. chunks are then indexed by time range and labels, in order to provide a fast lookup across many (over millions) chunks.
 
 For this reason, the chunks storage consists of:
 
@@ -45,26 +45,26 @@ For this reason, the chunks storage consists of:
 
 Internally, the access to the chunks storage relies on a unified interface called "chunks store". Unlike other Cortex components, the chunk store is not a separate service, but rather a library embedded in the services that need to access the long-term storage: [ingester](#ingester), [querier](#querier) and [ruler](#ruler).
 
-The chunk and index data format is versioned, in order to allow Cortex operators to upgrade the cluster to the latest stable version and take advantage of new features and improvements - which may introduce changes in the storage format - without requiring any downtime or complex procedures to rewrite the stored data.
+The chunk and index format are versioned, this allows Cortex operators to upgrade the cluster to take advantage of new features and improvements. This strategy enables changes in the storage format without requiring any downtime or complex procedures to rewrite the stored data. A set of schemas are used to map the version while reading and writing time series belonging to a specific period of time.
 
-A set of storage schemas are used to map the version to use to read and write time series belonging to a specific period of time. The current schema recommendation is the **v10 schema** (v11 is still experimental). For more information about the schema, please check out the "[Schema](guides/running.md#schema)" documentation.
+The current schema recommendation is the **v10 schema** (v11 is still experimental). For more information about the schema, please check out the [Schema](guides/running.md#schema) documentation.
 
 ### Blocks storage (experimental)
 
-The blocks storage is based on [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/storage/): it stores all the time series for a given tenant and period (defaults to 2 hours) into a single TSDB block. Each block is composed by few files storing the chunks and the block index.
+The blocks storage is based on [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/storage/): it stores each tenant's time series into their own TSDB which write out their series to a on-disk block (defaults to 2h block range periods). Each block is composed by few files storing the chunks and the block index.
 
-The chunk files contain the timestamp-value pairs for multiple series. The series inside the chunks are then indexed by a per-block index, which indexes metric names and labels to time series in the chunk files.
+The TSDB chunk files contain the timestamp-value pairs for multiple series. The series inside the chunks are then indexed by a per-block index, which indexes metric names and labels to time series in the chunk files.
 
-The blocks storage doesn't require a dedicated backend for the index, so it only requires an object store for the block files, which can be:
+The blocks storage doesn't require a dedicated storage backend for the index. The only requirement is an object store for the block files, which can be:
 
 * [Amazon S3](https://aws.amazon.com/s3)
 * [Google Cloud Storage](https://cloud.google.com/storage/)
 
-For more information, please check out the "[Blocks storage](operations/blocks-storage.md)" documentation.
+For more information, please check out the [Blocks storage](operations/blocks-storage.md) documentation.
 
 ## Services
 
-Cortex has a service-based architecture, in which the overall system is split up into a variety of components that perform specific tasks and run separately and in parallel.
+Cortex has a service-based architecture, in which the overall system is split up into a variety of components that perform a specific task, run separately, and in parallel.
 
 Cortex is, for the most part, a shared-nothing system. Each layer of the system can run multiple instances of each component and they don't coordinate or communicate with each other within that layer.
 
@@ -79,28 +79,35 @@ The Cortex services are:
 
 ### Distributor
 
-The **distributor** service is responsible for handling incoming series by Prometheus. It's the first stop in the write path for series samples. Once the distributor receives samples from Prometheus, each series is validated for correctness and to ensure that it is within the configured tenant (or global) limits. Valid samples and then split into batches and sent to multiple [ingesters](#ingester) in parallel.
+The **distributor** service is responsible for handling incoming series from Prometheus. It's the first stop in the write path for series samples. Once the distributor receives samples from Prometheus, each series is validated for correctness and to ensure that it is within the configured tenant limits, falling back to default ones in case limits have not been overridden for the specific tenant. Valid samples and then split into batches and sent to multiple [ingesters](#ingester) in parallel.
+
+The validation done by the distributor includes:
+
+- The metric name is formally correct
+- The configured max number of labels per metric is respected
+- The configured max length of a label name and value is respected
+- The timestamp is not older/newer than the configured min/max time range
 
 Distributors are **stateless** and can be scaled up and down as needed.
 
-#### HA Tracker
+#### High Availability Tracker
 
-The distributor features an **HA Tracker**. When enabled, the distributor deduplicates incoming samples from redundant Prometheus servers. This allows you to have multiple HA replicas of the same Prometheus servers, writing the same series to Cortex and then deduplicate these series in the Cortex distributor.
+The distributor features a **High Availability (HA) Tracker**. When enabled, the distributor deduplicates incoming samples from redundant Prometheus servers. This allows you to have multiple HA replicas of the same Prometheus servers, writing the same series to Cortex and then deduplicate these series in the Cortex distributor.
 
-The HA Tracker deduplicates incoming samples based on a cluster and replica label. The cluster label uniquely identifies the cluster of redundant Prometheus servers for a given tenant, while the replica label uniquely identify the replica within the Prometheus servers cluster. Incoming samples are considered duplicated - and thus skipped - if received by any replica which is not the current primary within a cluster.
+The HA Tracker deduplicates incoming samples based on a cluster and replica label. The cluster label uniquely identifies the cluster of redundant Prometheus servers for a given tenant, while the replica label uniquely identifies the replica within the Prometheus cluster. Incoming samples are considered duplicated - and thus dropped - if received by any replica which is not the current primary within a cluster.
 
-The HA Tracker requires a key-value (KV) store to coordinate which is the currently elected primary replica from which the distributor accepts the samples. Samples with one or neither of the replica and cluster labels are accepted by default and never deduplicated.
+The HA Tracker requires a key-value (KV) store to coordinate which replica is currently elected. The distributor will only accept samples from the current leader. Samples with one or no labels (of the replica and cluster) are accepted by default and never deduplicated.
 
 The supported KV stores for the HA tracker are:
 
 * [Consul](https://www.consul.io)
 * [Etcd](https://etcd.io)
 
-For more information, please check out the "[Config for sending HA pairs data to Cortex](guides/ha-pair-handling.md)" documentation.
+For more information, please refer to [config for sending HA pairs data to Cortex](guides/ha-pair-handling.md) in the documentation.
 
 #### Hashing
 
-Distributors use consistent hashing, in conjunction with a configurable replication factor, to determine which instances of the ingester service should receive a given series.
+Distributors use consistent hashing, in conjunction with a configurable replication factor, to determine which ingester instances should receive a given series.
 
 Cortex supports two hashing strategies:
 
@@ -108,6 +115,10 @@ Cortex supports two hashing strategies:
 2. Hash the metric name, labels and tenant ID (recommended, enabled with `-distributor.shard-by-all-labels=true`)
 
 The trade-off associated with the latter is that writes are more balanced across ingesters but each query needs to talk to any ingester since a metric could be spread across any ingester given different label sets. The first hashing strategy (default) was originally chosen to reduce the number of required ingesters on the query path, but running Cortex in production showed that an evenly distributed write load on the ingesters is preferable at the cost of always querying all ingesters.
+
+The trade-off associated with the latter is that writes are more balanced across ingesters, but requires each query to look across all ingesters, since a metric could be spread across any ingester given different label sets.
+
+The first hashing strategy (default) was originally chosen to reduce the number of required ingesters on the query path, but running Cortex in production showed that an evenly distributed write load on the ingesters is preferable at the cost of always querying most of the ingesters.
 
 #### The hash ring
 
@@ -151,7 +162,7 @@ Ingesters contain a **lifecycler** which manages the lifecycle of an ingester an
 
 5. `UNHEALTHY` is an ingester's state when it has failed to heartbeat to the ring's KV Store. While in this state, distributors skip the ingester while building the replication set for incoming series and the ingester does not receive write or read requests.
 
-For more information about the hand-over process, please check out the "[Ingester hand-over](guides/ingester-handover.md)" documentation.
+For more information about the hand-over process, please check out the [Ingester hand-over](guides/ingester-handover.md) documentation.
 
 Ingesters are **semi-stateful**.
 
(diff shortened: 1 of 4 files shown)


pracucci:
Specific requests:
- @codesome  may you double check the WAL section, please?
- @owen-d may you double check the query frontend section, please?
- @jtlisi may you help me enriching the ruler and alert manager doc, please?
- @thorfour may you review the TSDB related stuff?

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+Cortex currently supports two storage engines to store and query the time series:
+
+- Chunks (default, stable)
+- Blocks (experimental)
+
+The two engines mostly share the same Cortex architecture with few differences outlined in the rest of the document.
+
+### Chunks storage (default)
+
+The chunks storage stores each single time series into a separate object called chunk. Each chunk contains the timestamp-value pairs for a given period (defaults to 12 hours) of the series. Series chunks are then indexed by time range period and metric labels, in order to provide a fast lookup across millions - or even billions - of chunks.
```
gotjosh:
```suggestion
The Chunks storage stores each single time series into a separate object called _Chunk_. Each Chunk contains the timestamp-value pairs for a given period (defaults to 12 hours) of the series. Chunks are then indexed by time range and labels, in order to provide a fast lookup across many (over millions) chunks.
```

gotjosh: Sorry if this is a bit invasive, in this paragraph you're describing a Chunk and its purpose. Seems like in this case a Chunk becomes the subject of the paragraph.

[Prometheus describes](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/chunks.md#chunk) a Chunk as a proper noun. The only case where it doesn't is as it refers to chunk files. I know these are not TSDB Chunks but I'm using that documentation as an example.

owen-d:
Nice work with the docs. Frontend section is looking good.

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
 
-The query frontend job accepts gRPC streaming requests from the queriers, which then "pull" requests from the frontend. For high availability it's recommended that you run multiple frontends; the queriers will connect to—and pull requests from—all of them. To reap the benefit of fair scheduling, it is recommended that you run fewer frontends than queriers. Two should suffice in most cases.
+The query frontend internally performs some query mangling and push execution jobs for the queriers to an internal queue. In this setup, queriers act as workers picking jobs from the query frontend queue, executing the jobs and pushing back the result which is then aggregated by the query frontend before returning it to the client who performed the query request. Queriers need to be configured with the query frontend address - via the `-querier.frontend-address` CLI flag - in order to allow them to connect to the query frontends and pick up jobs to execute.
```
owen-d:
```suggestion
The query frontend internally performs some query adjustments and holds queries in an internal queue. In this setup, queriers act as workers which pull jobs from the queue, execute them, and return them to the query-frontend for aggregation. Queriers need to be configured with the query frontend address - via the `-querier.frontend-address` CLI flag in order to allow them to connect to the query frontends.
```

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+* An object store for the chunk data itself, which can be:
+    * [Amazon DynamoDB](https://aws.amazon.com/dynamodb)
+    * [Google Bigtable](https://cloud.google.com/bigtable)
+    * [Apache Cassandra](https://cassandra.apache.org)
+    * [Amazon S3](https://aws.amazon.com/s3)
+    * [Google Cloud Storage](https://cloud.google.com/storage/)
+
+Internally, the access to the chunks storage relies on a unified interface called "chunks store". Unlike other Cortex components, the chunk store is not a separate service, but rather a library embedded in the services that need to access the long-term storage: [ingester](#ingester), [querier](#querier) and [ruler](#ruler).
+
+The chunk and index data format is versioned, in order to allow Cortex operators to upgrade the cluster to the latest stable version and take advantage of new features and improvements - which may introduce changes in the storage format - without requiring any downtime or complex procedures to rewrite the stored data.
```
gotjosh:
```suggestion
The chunk and index format are versioned, this allows Cortex operators to upgrade the cluster to take advantage of new features and improvements. This strategy enables changes in the storage format without requiring any downtime or complex procedures to rewrite the stored data.
```

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
 
-> Unlike the other core components of Cortex, the chunk store is not a separate service, job, or process, but rather a library embedded in the three services that need to access Cortex data: the [ingester](#ingester), [querier](#querier), and [ruler](#ruler).
+#### Caching
 
-The chunk store relies on a unified interface to the "[NoSQL](https://en.wikipedia.org/wiki/NoSQL)" stores—DynamoDB, Bigtable, and Cassandra—that can be used to back the chunk store index. This interface assumes that the index is a collection of entries keyed by:
+The query frontend supports Memcached as backend to cache query results and reuses them on subsequent queries. If the cached results are incomplete, the query frontend calculates the required subqueries and executes them in parallel on downstream queriers. The query frontend can optionally align queries with their step parameter to improve the cacheability of the query results.
```
owen-d: I believe it's compatible with any of our caching backends as it uses `cache.Config` [internally](https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/results_cache.go#L27). The list would be `memcached`, `redis`, `memory`.

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+    * [Apache Cassandra](https://cassandra.apache.org)
+    * [Amazon S3](https://aws.amazon.com/s3)
+    * [Google Cloud Storage](https://cloud.google.com/storage/)
+
+Internally, the access to the chunks storage relies on a unified interface called "chunks store". Unlike other Cortex components, the chunk store is not a separate service, but rather a library embedded in the services that need to access the long-term storage: [ingester](#ingester), [querier](#querier) and [ruler](#ruler).
+
+The chunk and index data format is versioned, in order to allow Cortex operators to upgrade the cluster to the latest stable version and take advantage of new features and improvements - which may introduce changes in the storage format - without requiring any downtime or complex procedures to rewrite the stored data.
+
+A set of storage schemas are used to map the version to use to read and write time series belonging to a specific period of time. The current schema recommendation is the **v10 schema** (v11 is still experimental). For more information about the schema, please check out the "[Schema](guides/running.md#schema)" documentation.
```
gotjosh:
```suggestion
A set of schemas are used to map the version while reading and writing time series belonging to a specific period of time. 

The current stable schema is the **v10 schema** (v11 is still experimental). For more information about the schema, please check out the "[Schema](guides/running.md#schema)" documentation.
```

docs/architecture.md
owen-d:
```suggestion
The query frontend supports caching query results and reuses them on subsequent queries. If the cached results are incomplete, the query frontend calculates the required subqueries and executes them in parallel on downstream queriers. The query frontend can optionally align queries with their step parameter to improve the cacheability of the query results. The result cache is compatible with any cortex caching backend (currently memcached, redis, and an in-memory cache).
```

I believe it's compatible with any of our caching backends as it uses `cache.Config` [internally](https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/results_cache.go#L27). The list would be `memcached`, `redis`, `memory`.

docs/architecture.md
gotjosh: How do you feel about this split? Furthermore, I think the first paragraph here needs to be incorporated into the previous one. In the previous paragraph, we talk about the benefits of the schema but then on this paragraph we define how the schema is used if my understanding is correct which feels counterintuitive.

I think we could do something like:

> The chunk and index format are versioned, this allows Cortex operators to upgrade the cluster to take advantage of new features and improvements. This strategy enables changes in the storage format without requiring any downtime or complex procedures to rewrite the stored data. The set of schemas are used to map the version while reading and writing time series belonging to a specific period of time.

> The current stable schema is the **v10 schema** (v11 is still experimental). For more information about the schema, please refer to the [Schema documentation](guides/running.md#schema).

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+
+### Blocks storage (experimental)
+
+The blocks storage is based on [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/storage/): it stores all the time series for a given tenant and period (defaults to 2 hours) into a single TSDB block. Each block is composed by few files storing the chunks and the block index.
```
gotjosh:
```suggestion
The blocks storage is based on [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/storage/): It stores all the time series for a given tenant and period (defaults to 2 hours) into a single TSDB block. Each Block is composed of files storing the Chunks and the Block Index. You can refer to the [Prometheus documentation for a reference of the files used](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/README.md). 
```

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+
+The chunk files contain the timestamp-value pairs for multiple series. The series inside the chunks are then indexed by a per-block index, which indexes metric names and labels to time series in the chunk files.
```
gotjosh: When we talk about chunks, are we referring to the same chunks previously defined or TSDB chunks? Would be good to clear up that distinction and link to https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/chunks.md if necessary.

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+
+The blocks storage doesn't require a dedicated backend for the index, so it only requires an object store for the block files, which can be:
+
+* [Amazon S3](https://aws.amazon.com/s3)
+* [Google Cloud Storage](https://cloud.google.com/storage/)
+
+For more information, please check out the "[Blocks storage](operations/blocks-storage.md)" documentation.
```
gotjosh:
```suggestion
For more information, please check out the [Blocks storage](operations/blocks-storage.md) documentation.
```

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+
+A set of storage schemas are used to map the version to use to read and write time series belonging to a specific period of time. The current schema recommendation is the **v10 schema** (v11 is still experimental). For more information about the schema, please check out the "[Schema](guides/running.md#schema)" documentation.
+
+### Blocks storage (experimental)
+
+The blocks storage is based on [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/storage/): it stores all the time series for a given tenant and period (defaults to 2 hours) into a single TSDB block. Each block is composed by few files storing the chunks and the block index.
+
+The chunk files contain the timestamp-value pairs for multiple series. The series inside the chunks are then indexed by a per-block index, which indexes metric names and labels to time series in the chunk files.
+
+The blocks storage doesn't require a dedicated backend for the index, so it only requires an object store for the block files, which can be:
```
gotjosh:
```suggestion
The Blocks storage doesn't require a dedicated storage backend for the index. The only requirement is an object store for the block files, which can be:
```

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+
+* [Amazon S3](https://aws.amazon.com/s3)
+* [Google Cloud Storage](https://cloud.google.com/storage/)
+
+For more information, please check out the "[Blocks storage](operations/blocks-storage.md)" documentation.
 
 ## Services
 
-Cortex has a service-based architecture, in which the overall system is split up into a variety of components that perform specific tasks and run separately (and potentially in parallel).
+Cortex has a service-based architecture, in which the overall system is split up into a variety of components that perform specific tasks and run separately and in parallel.
```
gotjosh:
```suggestion
Cortex has a service-based architecture, in which the overall system is split up into a variety of components that perform a specific task, run separately, and in parallel.
```

docs/architecture.md
thorfour: > time series for a given tenant and period (defaults to 2 hours) into a single TSDB block

The wording on this reads weird to me, makes it sound like all users are in a single TSDB. Maybe change to something like "it stores each tenant's time series into their own TSDB which write out their metrics to a on-disk block"

pracucci: Resolving the comment, cause Thor suggested to completely rephrase it (see new version). Please re-open the comment if you have further comments on it.

docs/architecture.md
pracucci: Good point, thanks!

docs/architecture.md
pracucci: The grouping you propose makes sense to me. I picked your updated sentence and re-grouped it in the doc.

docs/architecture.md
pracucci: The chunks in the blocks storage are TSDB chunks. Conceptually they're not much different from the "chunks storage" given they both contain timestamp-value pairs, but in TSDB a single chunks file contain multiple series (many).

I've changed it to "TSDB chunk files" without repeating "TSDB chunk" in every single place, cause looks a bit overkilled to me.

pracucci:
Thanks @owen-d and @gotjosh. I've addressed your comments. May you re-check it, please?

> Is this is the kind of feedback you expected? Let me know and I'll keep going.

@gotjosh Yes, definitely. I'm just a bit dubious about the "Blocks" and "Chunks" uppercase letter. If we decide to go into this direction, it's better if we decide the rationale and I take care of all the changes cause there are many (more than then the places you edited). So far, I've kept them lowercase.

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+2. Hash the metric name, labels and tenant ID (recommended, enabled with `-distributor.shard-by-all-labels=true`)
+
+The trade-off associated with the latter is that writes are more balanced across ingesters but each query needs to talk to any ingester since a metric could be spread across any ingester given different label sets. The first hashing strategy (default) was originally chosen to reduce the number of required ingesters on the query path, but running Cortex in production showed that an evenly distributed write load on the ingesters is preferable at the cost of always querying all ingesters.
 
 #### The hash ring
 
-A consistent hash ring is stored in [Consul](https://www.consul.io/) as a single key-value pair, with the ring data structure also encoded as a [Protobuf](https://developers.google.com/protocol-buffers/) message. The consistent hash ring consists of a list of tokens and ingesters. Hashed values are looked up in the ring; the replication set is built for the closest unique ingesters by token. One of the benefits of this system is that adding and remove ingesters results in only 1/_N_ of the series being moved (where _N_ is the number of ingesters).
```
gotjosh: In my opinion, this section is one slightly more complex pieces of Cortex. You've done a fantastic job trying to expand on it, but I'm not sure about the way it reads.

If you feel strongly about it we can leave it as is, if you have doubts - I'm happy to take a stab and try to reword it and perhaps add a small diagram to it. What do you think?

gotjosh:
> @gotjosh Yes, definitely. I'm just a bit dubious about the "Blocks" and "Chunks" uppercase letter. If we decide to go into this direction, it's better if we decide the rationale and I take care of all the changes cause there are many (more than then the places you edited). So far, I've kept them lowercase.

@pracucci I'll leave that in your hands then. My opinion is that if these components have their own semantics they should be consider proper nouns e.g. A Block within Cortex does not fall under the regular definition of the word.

As I liked in the Prometheus doc, not capitalizing makes sense if we refer to them as adjectives e.g. block files or index files - in every other case, they should.

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+- [Ingester](#ingester)
+- [Querier](#querier)
+- [Query frontend](#query-frontend) (optional)
+- [Ruler](#ruler) (optional)
+- [Alertmanager](#alertmanager) (optional)
+
 ### Distributor
 
-The **distributor** service is responsible for handling samples written by Prometheus. It's essentially the "first stop" in the write path for Prometheus samples. Once the distributor receives samples from Prometheus, it splits them into batches and then sends them to multiple [ingesters](#ingester) in parallel.
+The **distributor** service is responsible for handling incoming series by Prometheus. It's the first stop in the write path for series samples. Once the distributor receives samples from Prometheus, each series is validated for correctness and to ensure that it is within the configured tenant (or global) limits. Valid samples and then split into batches and sent to multiple [ingesters](#ingester) in parallel.
```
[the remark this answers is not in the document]
pracucci: I've tried to address both feedback. What's about now?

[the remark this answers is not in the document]
gotjosh: Looks wonderful 👌

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+Distributors are **stateless** and can be scaled up and down as needed.
 
-If the HA Tracker is enabled, the Distributor will deduplicate incoming samples that contain both a cluster and replica label. It talks to a KVStore to store state about which replica per cluster it's accepting samples from for a given user ID. Samples with one or neither of these labels will be accepted by default.
+#### HA Tracker
 
-#### Hashing
+The distributor features an **HA Tracker**. When enabled, the distributor deduplicates incoming samples from redundant Prometheus servers. This allows you to have multiple HA replicas of the same Prometheus servers, writing the same series to Cortex and then deduplicate these series in the Cortex distributor.
+
+The HA Tracker deduplicates incoming samples based on a cluster and replica label. The cluster label uniquely identifies the cluster of redundant Prometheus servers for a given tenant, while the replica label uniquely identify the replica within the Prometheus servers cluster. Incoming samples are considered duplicated - and thus skipped - if received by any replica which is not the current primary within a cluster.
```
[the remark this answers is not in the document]
pracucci: I think dropped is more correct. I change it, thanks.

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
-2. All the series labels and tenant ID
+* [Consul](https://www.consul.io)
+* [Etcd](https://etcd.io)
+
+For more information, please check out the [Config for sending HA pairs data to Cortex](guides/ha-pair-handling.md) documentation.
+
+#### Hashing
 
-The trade-off associated with the latter is that writes are more balanced but they must involve every ingester in each query.
+Distributors use consistent hashing, in conjunction with a configurable replication factor, to determine which instances of the ingester service should receive a given series.
```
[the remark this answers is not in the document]
pracucci: Due to the replication factor they're multiple instances, so I think `instances` is more appropriate. I change it, then if you have further feedback please comment.

docs/architecture.md
```diff
@@ -15,139 +15,207 @@ Prometheus instances scrape samples from various targets and then push them to C
...
 
-> This hashing scheme was originally chosen to reduce the number of required ingesters on the query path. The trade-off, however, is that the write load on the ingesters is less even.
+Cortex supports two hashing strategies:
+
+1. Hash the metric name and tenant ID (default)
+2. Hash the metric name, labels and tenant ID (recommended, enabled with `-distributor.shard-by-all-labels=true`)
+
+The trade-off associated with the latter is that writes are more balanced across ingesters but each query needs to talk to any ingester since a metric could be spread across any ingester given different label sets. The first hashing strategy (default) was originally chosen to reduce the number of required ingesters on the query path, but running Cortex in production showed that an evenly distributed write load on the ingesters is preferable at the cost of always querying all ingesters.
```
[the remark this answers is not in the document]
pracucci: I agree on being clear about the reason.

> increases the likelihood of a query needing

It's not that it increases the likelihood, but we literally always query all ingesters. I've cherry picked your change and changed that part.

Feel free to further comment if you have any other feedback.

docs/architecture.md
pracucci: I never feel very strong about something, and I'm always open to discussion 😉 That being said, I would appreciate if could express the same concept in a easier way.

About the diagram, sure. I also thought about it, but I didn't (yet) because I'm not sure how to draw a diagram in a way which will be easily editable by other contributors in the future (ie. which tool to use? where should we share the sources?).

pracucci:
> As I liked in the Prometheus doc, not capitalizing makes sense if we refer to them as adjectives e.g. block files or index files - in every other case, they should.

Ok @gotjosh. May you check it again, please? I've capitalized them - when they're nouns (at least I believe they are) - in the commit https://github.com/cortexproject/cortex/pull/1945/commits/0e481f36bb84acab1d259ff99c5a534d1f3bb9a1

docs/architecture.md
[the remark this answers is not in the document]
gotjosh: I would use instance(s) then, as there is a chance (when is 1) that they only end up in 1.

docs/architecture.md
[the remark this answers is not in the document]
gotjosh: Thank you for clearing that up! Looks good 👌 

The diff still shows the previous paragraph, not sure if this is indented or not but was worth mentioning.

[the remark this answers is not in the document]
gotjosh: one tiny nit: Is that we use the word "spread" followed by "any" which suggests a division by 1. My suggestion is:

> since a metric could be spread across multiple ingesters

If my understanding of how this works is correct.

docs/architecture.md
gotjosh: I tend to use [mermaid](https://mermaidjs.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbkFbQ2hyaXN0bWFzXSAtLT58R2V0IG1vbmV5fCBCKEdvIHNob3BwaW5nKVxuQiAtLT4gQ3tMZXQgbWUgdGhpbmt9XG5DIC0tPnxPbmV8IERbTGFwdG9wXVxuQyAtLT58VHdvfCBFW2lQaG9uZV1cbkMgLS0-fFRocmVlfCBGW2ZhOmZhLWNhciBDYXJdXG4iLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ) for that exact same reasons. The source code is extremely simple which makes it easy to check it into version control and they have a live editor which means anyone can copy/paste then start editing right away.

Let's merge this as is and I'll create a follow-up issue assigned to me to come up with something.

gotjosh: I have created #1949 to record this conversation. Let me know what you think.

docs/architecture.md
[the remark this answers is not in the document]
pracucci: Done

docs/architecture.md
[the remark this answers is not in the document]
pracucci: > The diff still shows the previous paragraph, not sure if this is indented or not but was worth mentioning.

Ops, fixed. Thanks!

> since a metric could be spread across multiple ingesters

Way better.

docs/architecture.md
```diff
@@ -15,139 +15,218 @@ Prometheus instances scrape samples from various targets and then push them to C
...
+
+* [Amazon S3](https://aws.amazon.com/s3)
+* [Google Cloud Storage](https://cloud.google.com/storage/)
+
+For more information, please check out the [Blocks storage](operations/blocks-storage.md) documentation.
 
 ## Services
 
-Cortex has a service-based architecture, in which the overall system is split up into a variety of components that perform specific tasks and run separately (and potentially in parallel).
+Cortex has a service-based architecture, in which the overall system is split up into a variety of components that perform a specific task, run separately, and in parallel.
```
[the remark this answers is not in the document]
pracucci: Right. What's about now?

docs/architecture.md
```diff
@@ -15,139 +15,218 @@ Prometheus instances scrape samples from various targets and then push them to C
...
 
-The trade-off associated with the latter is that writes are more balanced but they must involve every ingester in each query.
+The trade-off associated with the latter is that writes are more balanced across ingesters, but requires each query to look across all ingesters, since a metric could be spread across any ingester given different label sets.
 
-> This hashing scheme was originally chosen to reduce the number of required ingesters on the query path. The trade-off, however, is that the write load on the ingesters is less even.
+The first hashing strategy (default) was originally chosen to reduce the number of required ingesters on the query path, but running Cortex in production showed that an evenly distributed write load on the ingesters is preferable at the cost of always querying most of the ingesters.
```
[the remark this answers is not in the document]
pracucci: I simplified it, removing "opinions" and just leaving the trade-off mention.

docs/architecture.md
```diff
@@ -15,139 +15,218 @@ Prometheus instances scrape samples from various targets and then push them to C
...
 
-[Upstream Docs](https://prometheus.io/docs/alerting/alertmanager/).
+The **write-ahead log** (WAL) is used to write to a persistent local disk all incoming series samples until they're flushed to the long-term storage. In the event of an ingester failure, a subsequent process restart will replay the WAL and recover the in-memory series samples.
 
-### Query frontend
+Contrary to the sole replication and given the persistent local disk data is not lost, in the event of multiple ingesters failure each ingester will recover the in-memory series samples from WAL upon subsequent restart. The replication is still recommended in order to ensure no temporary failures on the read path in the event of a single ingester failure.
```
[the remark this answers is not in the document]
pracucci: It means, "differently then when you setup a cluster only with the replication and without the WAL"

pracucci:
@gouthamve I've addressed your comments and replied to a couple. May you take another look, please?

docs/architecture.md
pracucci: I think it's good not blocking this PR with the improved ring description. I will comment the issue.

PR #1947, opened 2020-01-02
limits: distributor user subrings

thorfour:

**Experimental TSDB Change Only**

**What this PR does**:  Opening a TSDB per user on each Ingester presents a problem with memory consumption. The series hashtable that TSDB pre-allocates per user is roughly 1.5MB.
image
 Which means that at a large user count it eats up a large amount of memory per Ingester. More importantly when the number of ingesters/nodes are scaled out the memory consumption stays the same per node since the same number of TSDB's are allocated on each node. 

Just as a demonstration I reduced the stripe size in the TSDB code. 
```
+++ b/vendor/github.com/prometheus/prometheus/tsdb/head.go
@@ -1465,7 +1465,7 @@ type stripeSeries struct {
 }
 const (
-       stripeSize = 1 << 14
+       stripeSize = 1 << 8
        stripeMask = stripeSize - 1
 )
```
Which reduced overall memory consumption from ~94GiB to ~23Gib
image

Reducing the stripe size has performance implications, and doesn't resolve the root problem of metric footprint not scaling with node count. 

This change adds an optional step when distributing metrics, that first creates a subring of ingesters based on the user ID, and then uses that subring to hash as normal. This means that for any given user, that user only has an open TSDB on N given Ingesters reducing the overall memory footprint and horizontally scaling with nodes. The setting is apart of the util.Limits code to provide a default and importantly per-user limits to prevent hot-spots being created by extra heavy users. 

You can see in the image below the memory usage before (red arrow) with the reduced stripe size, and the memory usage after (green arrow) with a user sub ring of 4 and replication 3. So a user is allocated to 4 of the 16 nodes in the cluster. 

image

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893 
   * `--store.min-chunk-age` has been removed
   * `--querier.query-store-after` has been added in it's place.
+* [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
+  * `--experimental.distributor.user-subring-size`
 * [ENHANCEMENT] Experimental TSDB: Export TSDB Syncer metrics from Compactor component, they are prefixed with `cortex_compactor_`. #2023
 * [ENHANCEMENT] Experimental TSDB: Added dedicated flag `-experimental.tsdb.bucket-store.tenant-sync-concurrency` to configure the maximum number of concurrent tenants for which blocks are synched. #2026
 * [ENHANCEMENT] Experimental TSDB: Expose metrics for objstore operations (prefixed with `cortex_<component>_thanos_objstore_`, component being one of `ingester`, `querier` and `compactor`). #2027
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -1812,6 +1812,10 @@ The `limits_config` configures default and per-tenant limits imposed by Cortex s
 # CLI flag: -validation.enforce-metric-name
 [enforce_metric_name: <boolean> | default = true]
 
+# Per-user subring to shard metrics to ingesters. 0 is disabled.
+# CLI flag: -experimental.distributor.user-subring-size
+[user_subring_size: <int> | default = 0]
+
 # The maximum number of series that a query can return.
 # CLI flag: -ingester.max-series-per-query
 [max_series_per_query: <int> | default = 100000]
--- a/pkg/ring/ring.go
+++ b/pkg/ring/ring.go
@@ -388,16 +388,18 @@ func (r *Ring) Subring(key uint32, n int) (ReadRing, error) {
 		return nil, ErrEmptyRing
 	}
 
-	if n > len(r.ringDesc.Ingesters) {
-		return nil, fmt.Errorf("subring too large")
-	}
-
 	var (
 		ingesters     = make(map[string]IngesterDesc, n)
 		distinctHosts = map[string]struct{}{}
 		start         = r.search(key)
 		iterations    = 0
 	)
+
+	// Subring exceeds number of ingesters, set to total ring size
+	if n > len(r.ringDesc.Ingesters) {
+		n = len(r.ringDesc.Ingesters)
+	}
+
 	for i := start; len(distinctHosts) < n && iterations < len(r.ringTokens); i++ {
 		iterations++
 		// Wrap i around in the ring.
--- a/pkg/ring/ring.go
+++ b/pkg/ring/ring.go
@@ -44,6 +44,7 @@ type ReadRing interface {
 	GetAll() (ReplicationSet, error)
 	ReplicationFactor() int
 	IngesterCount() int
+	Subring(key uint32, n int) (ReadRing, error)
 }
 
 // Operation can be Read or Write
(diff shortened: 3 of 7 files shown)


pkg/util/validation/limits.go
```diff
@@ -39,6 +39,7 @@ type Limits struct {
 	RejectOldSamplesMaxAge time.Duration       `yaml:"reject_old_samples_max_age"`
 	CreationGracePeriod    time.Duration       `yaml:"creation_grace_period"`
 	EnforceMetricName      bool                `yaml:"enforce_metric_name"`
+	SubringSize            int                 `yaml:"subring_size"`
```
pracucci: I would call it `user_subring_size` like the CLI flag.

pkg/ring/ring.go
```diff
@@ -396,3 +397,56 @@ func (r *Ring) Collect(ch chan<- prometheus.Metric) {
...
+	}
+
+	if n > len(ingesters) {
+		return nil, fmt.Errorf("too few ingesters found")
+	}
+
+	d := &Desc{
+		Ingesters: ingesters,
+	}
+	d.Tokens = migrateRing(d) // TODO THOR there is likely a better way to do this
```
pracucci: Could you explain why is this required? I haven't deeply checked, but this function should already work on a `TokenDesc` which have already been migrated.

pkg/ring/ring.go
```diff
@@ -396,3 +397,56 @@ func (r *Ring) Collect(ch chan<- prometheus.Metric) {
 		r.name,
 	)
 }
+
+// Subring returns a ring of n ingesters from the given ring
+// Subrings are meant only for ingestor lookup and should have their data externalized.
```
pracucci: You could enforce it a bit returning a `ReadRing` instead of `*Ring`

pracucci:
@codesome While I believe this feature is great, do you also see any optimization we can do in TSDB to mitigate this?

pkg/ring/ring.go
thorfour: Since I have no reason for restricting behaviour I think it's better practice to return the full struct.

pkg/util/validation/limits.go
thorfour: Done.

pkg/ring/ring.go
thorfour: I've updated this piece of code, and no longer use migrate, but instead walk the tokens list and pick out the tokens for the ingesters in our subring. That what the migrate function was doing for us thought it was a little heavy handed for our use case.

codesome:
> @codesome While I believe this feature is great, do you also see any optimization we can do in TSDB to mitigate this?

Usually, such constant numbers are not exposed for users to set in case of Prometheus, and it might be the same case with `stripeSize`. As Prometheus is focused on a single installation of Prometheus on a node and performance, I don't think this constant would get reduced upstream.

pkg/ring/ring.go
pracucci: What we're returning here is not a fully functional ring (ie. created with `ring.New()`). By fully functional, I mean a ring honoring the ring's features contract, that's expected. We're just returning a sub-view of the ring, so restricting the behaviour returning with `ReadRing` looks more appropriate to me.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
 * [FEATURE] The distributor can now drop labels from samples (similar to the removal of the replica label for HA ingestion) per user via the `distributor.drop-label` flag. #1726
 * [FEATURE] Added `global` ingestion rate limiter strategy. Deprecated `-distributor.limiter-reload-period` flag. #1766
 * [FEATURE] Added support for Microsoft Azure blob storage to be used for storing chunk data. #1913
+* [FEATURE] EXPERIMENTAL: Added user sub rings for use with TSDB backend storage. #1947
```
pracucci: Technically this feature is not related to the TSDB storage (even if that's your use case). I would not mention it here, while I would mention the CLI flag controlling it.

pkg/ring/ring.go
thorfour: That's a good point. Will change.

CHANGELOG.md
thorfour: Done.

pkg/ring/ring.go
thorfour: Fixed.

pkg/util/validation/limits.go
```diff
@@ -280,6 +284,33 @@ func (o *Overrides) MinChunkLength(userID string) int {
...
+func (o *Overrides) SubringSize(userID string) int {
+	return o.getOverridesForUser(userID).SubringSize
+}
+
+// Loads overrides and returns the limits as an interface to store them in OverridesManager.
+// We need to implement it here since OverridesManager must store type Limits in an interface but
+// it doesn't know its definition to initialize it.
+// We could have used yamlv3.Node for this but there is no way to enforce strict decoding due to a bug in it
+// TODO: Use yamlv3.Node to move this to OverridesManager after https://github.com/go-yaml/yaml/issues/460 is fixed
+func loadOverrides(filename string) (map[string]interface{}, error) {
```
pracucci: What's this?

thorfour: merge error. Removed it.

bboreham:
Even with the chunk store, I have sometimes wished that users were restricted to subsets of ingesters, generally because 3+ ingesters crashed so I lost some data from everyone and I would prefer to apologise to a subset of users.

thorfour:
@codesome I definitely don't think the constant should be reduced. But is there a way to have it be user defined? So that TSDB's aren't as memory hungry but trade performance at upper bounds to do so?

pstibrany:
This PR may have implications to PR #1764.

pkg/ring/ring.go
```diff
@@ -396,3 +397,68 @@ func (r *Ring) Collect(ch chan<- prometheus.Metric) {
...
+func (r *Ring) Subring(key uint32, n int) (ReadRing, error) {
```
pstibrany: Would it make sense to use some caching here? We're going to call this on every push request, which looks like a lot of ring traversal. On the other hand, number of different keys isn't that high (= number of tenants), so perhaps we can use some of distributors memory to cache these subrings for some time.

pkg/ring/ring.go
```diff
@@ -396,3 +397,68 @@ func (r *Ring) Collect(ch chan<- prometheus.Metric) {
...
+		return nil, fmt.Errorf("subring too large")
+	}
+
+	var (
+		ingesters     = make(map[string]IngesterDesc, n)
+		distinctHosts = map[string]struct{}{}
+		start         = r.search(key)
+		iterations    = 0
+	)
+	for i := start; len(distinctHosts) < n && iterations < len(r.ringDesc.Tokens); i++ {
```
pstibrany: It seems to me that this algorithm is too complicated for what we're trying to do here (pick N ingesters based on key). That said, when I tried to come up with alternatives, they were not any better – and would cause too much ingester jumps during scale-up/down.

pkg/ring/ring.go
thorfour: I considered that, but was concerned about cache invalidation when the ring size changes. So I decided to at least punt on that for now.

pkg/ring/ring.go
thorfour: I think I would have preferred a simpler solution as well, but I didn't want to thrash when ring size changes as you've pointed out. I think if we couple this with #1958 it may not be so much of a problem? 

Tangentially related I also had the idea that if a Querier might be able to detect a ring size hasn't changed in the last N time, and could in theory stop sending queries to all ingesters and instead only send them to the subring.

pracucci:
@bboreham @gouthamve Have you had a chance to take a look at this PR? To my perspective, it's a safe feature, given doesn't change the actual Cortex logic until you enable the user sub ring.

docs/configuration/config-file-reference.md
```diff
@@ -1811,6 +1811,10 @@ The `limits_config` configures default and per-tenant limits imposed by Cortex s
 # CLI flag: -validation.enforce-metric-name
 [enforce_metric_name: <boolean> | default = true]
 
+# Per-user subring to shard metrics to ingesters. 0 is disabled.
+# CLI flag: -distributor.user-subring-size
```
pracucci: I'm wondering if we should prefix it with `-experimental.`. Thoughts?

thorfour: I think that's probably a good idea

thorfour: Done.

pkg/ring/ring.go
```diff
@@ -377,3 +378,66 @@ func (r *Ring) Collect(ch chan<- prometheus.Metric) {
...
+// Subring returns a ring of n ingesters from the given ring
+// Subrings are meant only for ingestor lookup and should have their data externalized.
+func (r *Ring) Subring(key uint32, n int) (ReadRing, error) {
+	r.mtx.RLock()
+	defer r.mtx.RUnlock()
+	if r.ringDesc == nil || len(r.ringTokens) == 0 || n <= 0 {
+		return nil, ErrEmptyRing
+	}
+
+	if n > len(r.ringDesc.Ingesters) {
```
pracucci: While doing some tests I've tried to intentionally hit this. I'm wondering if this is a real problem for which returning error is desired. If the subring is larger than the current number of ingesters, it means the returned subring will include all ingesters, which is not technically wrong. I do understand it's not desiderable (why do you use the subring, then?) but it could actually happen in case of an incident where you find yourself forgetting many ingesters while spinning up new ones.

What if we just remove this line (and cover this case with a test)?

thorfour: Good idea.

Modified that line to set n to number of ingesters, and modified test accordingly.

pracucci:
I've done another round of tests, and LGTM. It's time merge it!

PR #1952, opened 2020-01-05
Added tool to automatically generate config file documentation

pracucci:
**What this PR does**:
Last Friday, I started to manually document the YAML config file and I suddenly realized that:
1. It's huge
2. Even if we pay the effort to manually write the doc, it will be very difficult to keep it updated PR-by-PR

So I've tried a different approach: automatically generate it navigating through the entire config using the Go reflection. In order to do it, I mix both the YAML tags in the config structs and the registered CLI flags. The CLI flags are particularly important because they carry the description of each entry in the config.

This is the result of this work. In particular:
- `make doc` generates the config file reference doc. It's intended to be manually run by a contributor whenever he/she changes the config / CLI flags. In this PR the generated doc is at `docs/configuration/config-file-reference.md`.
- `make check-doc` ensure consistency in the config file reference doc. It's intended to be run in CI as part of the linting process.

Out of the scope of this PR (but welcome in subsequent PRs):
- Significantly improve the quality of CLI flags usage (which is now the config entry description)

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/cache/cache.go
+++ b/pkg/chunk/cache/cache.go
@@ -33,10 +33,10 @@ type Config struct {
 	Fifocache      FifoCacheConfig       `yaml:"fifocache,omitempty"`
 
 	// This is to name the cache metrics properly.
-	Prefix string `yaml:"prefix,omitempty"`
+	Prefix string `yaml:"prefix,omitempty" doc:"hidden"`
 
 	// For tests to inject specific implementations.
-	Cache Cache
+	Cache Cache `yaml:"-"`
 }
 
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -66,7 +66,7 @@ type Config struct {
 	Ingester       ingester.Config          `yaml:"ingester,omitempty"`
 	Storage        storage.Config           `yaml:"storage,omitempty"`
 	ChunkStore     chunk.StoreConfig        `yaml:"chunk_store,omitempty"`
-	Schema         chunk.SchemaConfig       `yaml:"schema,omitempty" doc:"hidden"`
+	Schema         chunk.SchemaConfig       `yaml:"schema,omitempty" doc:"hidden"` // Doc generation tool doesn't support it because part of the SchemaConfig doesn't support CLI flags (needs manual documentation)
 	LimitsConfig   validation.Limits        `yaml:"limits,omitempty"`
 	Prealloc       client.PreallocConfig    `yaml:"prealloc,omitempty" doc:"hidden"`
 	Worker         frontend.WorkerConfig    `yaml:"frontend_worker,omitempty"`
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -66,15 +66,15 @@ type Config struct {
 	Ingester       ingester.Config          `yaml:"ingester,omitempty"`
 	Storage        storage.Config           `yaml:"storage,omitempty"`
 	ChunkStore     chunk.StoreConfig        `yaml:"chunk_store,omitempty"`
-	Schema         chunk.SchemaConfig       `yaml:"schema,omitempty"`
+	Schema         chunk.SchemaConfig       `yaml:"schema,omitempty" doc:"hidden"`
 	LimitsConfig   validation.Limits        `yaml:"limits,omitempty"`
-	Prealloc       client.PreallocConfig    `yaml:"prealloc,omitempty"`
+	Prealloc       client.PreallocConfig    `yaml:"prealloc,omitempty" doc:"hidden"`
 	Worker         frontend.WorkerConfig    `yaml:"frontend_worker,omitempty"`
 	Frontend       frontend.Config          `yaml:"frontend,omitempty"`
 	QueryRange     queryrange.Config        `yaml:"query_range,omitempty"`
 	TableManager   chunk.TableManagerConfig `yaml:"table_manager,omitempty"`
 	Encoding       encoding.Config          `yaml:"-"` // No yaml for this, it only works with flags.
-	TSDB           tsdb.Config              `yaml:"tsdb"`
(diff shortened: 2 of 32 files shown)


pkg/cortex/cortex.go
```diff
@@ -88,7 +88,7 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	c.Server.MetricsNamespace = "cortex"
 	c.Target = All
 	c.Server.ExcludeRequestInLog = true
-	f.Var(&c.Target, "target", "target module (default All)")
+	f.Var(&c.Target, "target", "The Cortex service to run. Supported values are: all, distributor, ingester, querier, query-frontend, table-manager, ruler, alertmanager, configs.")
```
pstibrany: Default value is now missing.

tools/doc-generator/main.go
```diff
@@ -0,0 +1,239 @@
...
+
+const (
+	maxLineWidth = 80
+	tabWidth     = 2
+)
+
+var (
+	// Ordered list of root blocks. The order is the same order that will
+	// follow the markdown generation.
+	rootBlocks = []rootBlock{
```
pstibrany: It would be nicer if this information was directly in the Cortex code, if possible. One idea is to put it into cortex.go, next to the reference to config structure.

```
Server         server.Config            `yaml:"server,omitempty"` `doc:"name:server_config,configures the HTTP and gRPC server of the launched service(s)."`
```

(if tag syntax allows this at all)

Not sure how difficult it would be. But it would be very cool if we could reuse this for Loki as well, without too many modifications.

Another reason for doing it is that currently it's difficult to find these extra documentation lines. Tool itself is the last place where I would look for it.

pkg/cortex/cortex.go
pracucci: We don't need to specify the default value in the description, because it's picked from the default value set for the flag (few lines above).

The `-help` is:

```
  -target value
        The Cortex service to run. Supported values are: all, distributor, ingester, querier, query-frontend, table-manager, ruler, alertmanager, configs. (default all)
```

While the related config doc is:

```
# The Cortex service to run. Supported values are: all, distributor, ingester,
# querier, query-frontend, table-manager, ruler, alertmanager, configs.
# CLI flag: -target
[target: <string> | default = "all"]
```

pstibrany: Great, good to know. Thanks!

pkg/cortex/cortex.go
```diff
@@ -66,15 +66,15 @@ type Config struct {
 	Ingester       ingester.Config          `yaml:"ingester,omitempty"`
 	Storage        storage.Config           `yaml:"storage,omitempty"`
 	ChunkStore     chunk.StoreConfig        `yaml:"chunk_store,omitempty"`
-	Schema         chunk.SchemaConfig       `yaml:"schema,omitempty"`
+	Schema         chunk.SchemaConfig       `yaml:"schema,omitempty" doc:"hidden"`
```
gouthamve: Why are these config options hidden?

pkg/distributor/distributor_ring.go
```diff
@@ -22,10 +22,10 @@ type RingConfig struct {
 	HeartbeatTimeout time.Duration `yaml:"heartbeat_timeout,omitempty"`
 
 	// Instance details
-	InstanceID             string   `yaml:"instance_id"`
-	InstanceInterfaceNames []string `yaml:"instance_interface_names"`
-	InstancePort           int      `yaml:"instance_port"`
-	InstanceAddr           string   `yaml:"instance_addr"`
+	InstanceID             string   `yaml:"instance_id" doc:"hidden"`
```
gouthamve: Why are these hidden?

pkg/chunk/cache/cache.go
```diff
@@ -33,10 +33,10 @@ type Config struct {
 	Fifocache      FifoCacheConfig       `yaml:"fifocache,omitempty"`
 
 	// This is to name the cache metrics properly.
-	Prefix string `yaml:"prefix,omitempty"`
+	Prefix string `yaml:"prefix,omitempty" doc:"hidden"`
```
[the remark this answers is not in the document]
pracucci: This is weird. It has no related CLI flag (it's required to have a field documented by this tool and will fail if can't find any non-hidden field without a CLI flag) and in `RegisterFlagsWithPrefix()` it's set as `cfg.Prefix = prefix`. I'm missing a bit of context on the history of this config option, but the missing CLI flag is the main reason why I've hide it.

pkg/distributor/distributor_ring.go
pracucci: I marked them as hidden because I've the feeling they introduce confusion to the user. If you see them you may think you're expected to configure them, while they're here to cover very edge cases (ie. local dev env) and the default should work in most of the setups. I would re-introduce them, but I've the feeling they will just make the config file reference harder to understand when it comes to configuring the ring (same reason why they've been hidden in `ring.LifecyclerConfig`).

pkg/cortex/cortex.go
pracucci: I totally forgot to mention. TL;DR the `SchemaConfig` can't be automatically generated and my plan is to manually document it in a subsequent PR.

Why can't be generated? Because this tool is based on a mapping between the config structs and CLI flags. We have flags registered for the `LegacySchemaConfig`, but not for the `PeriodConfig`. Moreover the `PeriodConfig` is a slice and slices of struct - in the config - don't play nicely with this tool due to how CLI flags are mapped with the config (by memory address, but a default slice would be empty so there's nothing to map even if we would have the CLI flags).

gouthamve: Ack, could you leave a comment in the code and then lets get this merged!

pracucci: Commented.

> then lets get this merged!

Let's wait for the CI to pass. Please also note that there's only 1 maintainer approval.

gouthamve: > Please also note that there's only 1 maintainer approval.

I'm assuming you approve of your own PR ;) It's okay if you don't :P

PR #1957, opened 2020-01-06
pkg/distributor: shardByAllLabels didn't check for label names order correctly

pstibrany:
`shardByAllLabels` had a check to verify if labels are sorted properly, but it didn't set `lastLabelName`, so it was comparing label names against empty string only.


gouthamve:
This change affects only if `shardByAllLabels` is used right? If yes then LGTM

pstibrany:
> This change affects only if `shardByAllLabels` is used right? If yes then LGTM

Yes.

PR #1958, opened 2020-01-06
Feat/tsdb/close stale tsdb

thorfour:
**What this PR does**: Closes and removes TSDB's that haven't been written to in a given time period.

**Which issue(s) this PR fixes**: This cleans up disk and memory footprint for open TSDB's that no longer are written to. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -3,6 +3,7 @@ package ingester
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -34,9 +35,28 @@ const (
 	errTSDBCreateIncompatibleState = "cannot create a new TSDB while the ingester is not in active state (current state: %s)"
 )
 
+const (
+	active = iota
+	closing
+)
+
+// UserTSDB is a users TSDB
+type UserTSDB struct {
+	*tsdb.DB
+	sync.WaitGroup
+
+	shipperCancel context.CancelFunc
+	shipperDone   *sync.WaitGroup
+
+	firstEmpty time.Time
+
+	// state indicates if the TSDB is currently active or in the process of closing.
+	state int
+}
+
 // TSDBState holds data structures used by the TSDB storage engine
 type TSDBState struct {
-	dbs    map[string]*tsdb.DB // tsdb sharded by userID
+	dbs    map[string]*UserTSDB // tsdb sharded by userID
 	bucket objstore.Bucket
 
 	// Keeps count of in-flight requests
@@ -63,7 +83,7 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 		quit:         make(chan struct{}),
 
 		TSDBState: TSDBState{
-			dbs:    make(map[string]*tsdb.DB),
+			dbs:    make(map[string]*UserTSDB),
 			bucket: bucketClient,
 		},
 	}
@@ -85,6 +105,9 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 	// Now that user states have been created, we can start the lifecycler
 	i.lifecycler.Start()
 
+	// Start a cleanup routine that will close and remove TSDB's that are no longer active
+	i.startCleanupLoop()
+
 	return i, nil
 }
 
@@ -101,6 +124,7 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 	if err != nil {
 		return nil, wrapWithUser(err, userID)
 	}
+	defer db.Done()
 
 	// Ensure the ingester shutdown procedure hasn't started
 	i.userStatesMtx.RLock()
@@ -195,6 +219,7 @@ func (i *Ingester) v2Query(ctx old_ctx.Context, req *client.QueryRequest) (*clie
 	if db == nil {
 		return &client.QueryResponse{}, nil
 	}
+	defer db.Done()
 
 	q, err := db.Querier(int64(from), int64(through))
 	if err != nil {
@@ -243,6 +268,7 @@ func (i *Ingester) v2LabelValues(ctx old_ctx.Context, req *client.LabelValuesReq
 	if db == nil {
 		return &client.LabelValuesResponse{}, nil
 	}
+	defer db.Done()
 
 	through := time.Now()
 	from := through.Add(-i.cfg.TSDBConfig.Retention)
@@ -272,6 +298,7 @@ func (i *Ingester) v2LabelNames(ctx old_ctx.Context, req *client.LabelNamesReque
 	if db == nil {
 		return &client.LabelNamesResponse{}, nil
 	}
+	defer db.Done()
 
 	through := time.Now()
 	from := through.Add(-i.cfg.TSDBConfig.Retention)
@@ -301,6 +328,7 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
 	if db == nil {
 		return &client.MetricsForLabelMatchersResponse{}, nil
 	}
+	defer db.Done()
 
 	// Parse the request
 	from, to, matchersSet, err := client.FromMetricsForLabelMatchersRequest(req)
@@ -358,14 +386,20 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx old_ctx.Context, req *client.Me
 	return result, nil
 }
 
-func (i *Ingester) getTSDB(userID string) *tsdb.DB {
+func (i *Ingester) getTSDB(userID string) *UserTSDB {
 	i.userStatesMtx.RLock()
 	defer i.userStatesMtx.RUnlock()
-	db, _ := i.TSDBState.dbs[userID]
+	db, ok := i.TSDBState.dbs[userID]
+	if ok {
+		if db.state == closing {
+			return nil
+		}
+		db.Add(1)
+	}
 	return db
 }
 
-func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error) {
+func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*UserTSDB, error) {
 	db := i.getTSDB(userID)
 	if db != nil {
 		return db, nil
@@ -378,6 +412,10 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error) 
 	var ok bool
 	db, ok = i.TSDBState.dbs[userID]
 	if ok {
+		if db.state == closing {
+			return nil, fmt.Errorf("tsdb is closing")
+		}
+		db.Add(1)
 		return db, nil
 	}
 
@@ -391,23 +429,12 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error) 
 		return nil, fmt.Errorf(errTSDBCreateIncompatibleState, ingesterState)
 	}
 
-	// Create the database and a shipper for a user
-	db, err := i.createTSDB(userID)
-	if err != nil {
-		return nil, err
-	}
-
-	// Add the db to list of user databases
-	i.TSDBState.dbs[userID] = db
-	return db, nil
-}
-
-// createTSDB creates a TSDB for a given userID, and returns the created db.
-func (i *Ingester) createTSDB(userID string) (*tsdb.DB, error) {
+	db = &UserTSDB{}
 	udir := i.cfg.TSDBConfig.BlocksDir(userID)
 
 	// Create a new user database
-	db, err := tsdb.Open(udir, util.Logger, nil, &tsdb.Options{
+	var err error
+	db.DB, err = tsdb.Open(udir, util.Logger, nil, &tsdb.Options{
 		RetentionDuration: uint64(i.cfg.TSDBConfig.Retention / time.Millisecond),
 		BlockRanges:       i.cfg.TSDBConfig.BlockRanges.ToMillisecondRanges(),
 		NoLockfile:        true,
@@ -415,34 +442,13 @@ func (i *Ingester) createTSDB(userID string) (*tsdb.DB, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	// Thanos shipper requires at least 1 external label to be set. For this reason,
-	// we set the tenant ID as external label and we'll filter it out when reading
-	// the series from the storage.
-	l := labels.Labels{
-		{
-			Name:  cortex_tsdb.TenantIDExternalLabel,
-			Value: userID,
-		},
-	}
+	// Save the database
+	i.TSDBState.dbs[userID] = db
 
 	// Create a new shipper for this database
-	if i.cfg.TSDBConfig.ShipInterval > 0 {
-		s := shipper.New(util.Logger, nil, udir, &Bucket{userID, i.TSDBState.bucket}, func() labels.Labels { return l }, metadata.ReceiveSource)
-		i.done.Add(1)
-		go func() {
-			defer i.done.Done()
-			if err := runutil.Repeat(i.cfg.TSDBConfig.ShipInterval, i.quit, func() error {
-				if uploaded, err := s.Sync(context.Background()); err != nil {
-					level.Warn(util.Logger).Log("err", err, "uploaded", uploaded)
-				}
-				return nil
-			}); err != nil {
-				level.Warn(util.Logger).Log("err", err)
-			}
-		}()
-	}
+	i.startShipper(userID)
 
+	db.Add(1)
 	return db, nil
 }
 
@@ -456,7 +462,7 @@ func (i *Ingester) closeAllTSDB() {
 	for userID, db := range i.TSDBState.dbs {
 		userID := userID
 
-		go func(db *tsdb.DB) {
+		go func(db *UserTSDB) {
 			defer wg.Done()
 
 			if err := db.Close(); err != nil {
@@ -523,16 +529,12 @@ func (i *Ingester) openExistingTSDB(ctx context.Context) error {
 		go func(userID string) {
 			defer wg.Done()
 			defer openGate.Done()
-			db, err := i.createTSDB(userID)
+			db, err := i.getOrCreateTSDB(userID, true)
 			if err != nil {
 				level.Error(util.Logger).Log("msg", "unable to open user TSDB", "err", err, "user", userID)
 				return
 			}
-
-			// Add the database to the map of user databases
-			i.userStatesMtx.Lock()
-			i.TSDBState.dbs[userID] = db
-			i.userStatesMtx.Unlock()
+			db.Done()
 
 		}(userID)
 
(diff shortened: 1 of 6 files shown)


pkg/ingester/transfer.go
```diff
@@ -629,7 +629,10 @@ func unshippedBlocks(dir string) (map[string][]string, error) {
 
 		m, err := shipper.ReadMetaFile(filepath.Join(dir, userID))
 		if err != nil {
-			return nil, err
+			if !os.IsNotExist(err) {
```
pracucci: Could you comment explaining which is the case when it could happen and why we don't consider it an error?

pkg/ingester/ingester_v2.go
```diff
@@ -455,3 +488,93 @@ func (i *Ingester) closeAllTSDB() {
...
 	wg.Wait()
 }
+
+/* cleanup walks the TSDB directory for each user to determine if the TSDB should be closed.
+ * If
+ * - The directory is empty of blocks
+ * OR
+ * - The directory has every block marked as shipped
+ * AND
+ * - The TSDB hasn't been written to within the alloted time period
```
pracucci: `alloted` > `allowed`

pkg/ingester/ingester_v2.go
```diff
@@ -455,3 +488,93 @@ func (i *Ingester) closeAllTSDB() {
...
+ * THEN
+ * - Close the TSDB, stop the shipper routine, and delete the directory and it's contents.
+ */
+func (i *Ingester) cleanup() error {
+	ts := time.Now()
```
pracucci: Calling this variable `now` may help reading the following code.

pkg/ingester/ingester_v2.go
```diff
@@ -455,3 +488,93 @@ func (i *Ingester) closeAllTSDB() {
...
+			level.Warn(util.Logger).Log("msg", "unable to find user tsdb during cleanup", "user", user)
+			continue
+		}
+
+		udb.Lock()
+		last := udb.lastWritten
+		udb.Unlock()
+
+		// If the TSDB is inactive, close and remove it.
+		if ts.Sub(last) > i.cfg.TSDBConfig.MaxStaleAge {
```
pracucci: We could reverse with a guard, like you did above for other checks:

```
if ts.Sub(last) <= i.cfg.TSDBConfig.MaxStaleAge {
  continue
}
```

pkg/ingester/ingester_v2.go
```diff
@@ -29,9 +30,18 @@ const (
 	errTSDBCreateIncompatibleState = "cannot create a new TSDB while the ingester is not in active state (current state: %s)"
 )
 
+// UserTSDB is a users TSDB
+type UserTSDB struct {
+	sync.Mutex
+
+	*tsdb.DB
+	cancel      context.CancelFunc
```
pracucci: Maybe naming it `shipperCancel` may be more clear to the reader?

pkg/ingester/ingester_v2.go
```diff
@@ -455,3 +488,93 @@ func (i *Ingester) closeAllTSDB() {
...
+		// If the TSDB is inactive, close and remove it.
+		if ts.Sub(last) > i.cfg.TSDBConfig.MaxStaleAge {
+			wg.Add(1)
+			go func(user string) {
+				defer wg.Done()
+				i.userStatesMtx.Lock()
+				defer i.userStatesMtx.Unlock()
+
+				db, ok := i.TSDBState.dbs[user]
+				if ok {
```
pracucci: What's the case, if `!ok` we want to continue? Looks that if we don't have the TSDB opened, we don't want to continue, so a guard may be better.

pkg/ingester/ingester_v2.go
```diff
@@ -455,3 +488,93 @@ func (i *Ingester) closeAllTSDB() {
...
+
+					// Stop the shipper routine for this TSDB
+					db.cancel()
+
+					// Close the TSDB
+					if err := db.Close(); err != nil {
```
pracucci: Closing the TSDB, any samples written in the head but not yet cut into a new block will not be uploaded to the storage. We need to ensure a new block from the head is cut and a last successful run of the ship is completed before moving on (especially before deleting data on disk).

pkg/ingester/ingester_v2.go
```diff
@@ -29,9 +30,18 @@ const (
 	errTSDBCreateIncompatibleState = "cannot create a new TSDB while the ingester is not in active state (current state: %s)"
 )
 
+// UserTSDB is a users TSDB
+type UserTSDB struct {
+	sync.Mutex
```
pracucci: If this is just used to lock around `lastWritten` I would be more explicit and name it `lastWrittenLock`. If in the future we'll use it for more purposes, we'll rename it.

pkg/ingester/transfer.go
thorfour: Added
```// If the cleanup function runs before the shipper has had a chance to run there wont be a meta file written yet.
// So we can treat as an "empty" meta file and no blocks have been shipped.```

File: pkg/ingester/ingester_v2.go (hunk 2 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: I think allotted is a better verb here. Though the spelling was incorrect so I fixed that.

File: pkg/ingester/ingester_v2.go (hunk 3 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: Changed.

File: pkg/ingester/ingester_v2.go (hunk 4 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: Sure. Changed.

File: pkg/ingester/ingester_v2.go (hunk 5 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: Sure. Changed.

File: pkg/ingester/ingester_v2.go (hunk 6 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: well if we don't have a TSDB we still want to remove any on disk remnants that may exist. So I think it still makes sense to try and clean those up.

File: pkg/ingester/ingester_v2.go (hunk 7 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: Yup that's a good catch. I'll refactor some things to do this. I think it'll be a two pass operation.

File: pkg/ingester/ingester_v2.go (hunk 8 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: Agreed, changed.

File: pkg/ingester/ingester_v2.go (hunk 7 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: Okay, I've refactored some of that code, have it take a snapshot before closing. I think this does a better job of covering corner cases.

File: pkg/ingester/ingester_v2.go (hunk 6 above)
Comment (thorfour) [reply pkg/ingester/ingester_v2.go]: This no longer applies dur to recent refactor.

File: pkg/ingester/transfer.go [hunk 9]
```diff
@@ -619,33 +619,48 @@ func unshippedBlocks(dir string) (map[string][]string, error) {
 
 	blocks := make(map[string][]string, len(userIDs))
 	for _, user := range userIDs {
-		userID := user.Name()
-		blocks[userID] = []string{} // seed the map with the userID to ensure we xfer the WAL, even if all blocks are shipped
-
-		blockIDs, err := ioutil.ReadDir(filepath.Join(dir, userID))
+		blks, err := unshippedUserBlocks(filepath.Join(dir, user.Name()))
 		if err != nil {
```
pracucci: We should log this error.

pkg/ingester/ingester_v2.go
```diff
@@ -110,6 +138,12 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 	i.userStatesMtx.RUnlock()
 	defer i.TSDBState.inflightWriteReqs.Done()
 
+	// Update last written timestamp
+	now := time.Now()
+	db.lastWrittenLock.Lock()
```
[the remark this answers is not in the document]
thorfour: `lastWritten` and `lastWrittenLock` have been removed in favor of the `firstEmpty` var

pkg/ingester/ingester_v2.go
```diff
@@ -75,6 +88,21 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 	// Now that user states have been created, we can start the lifecycler
 	i.lifecycler.Start()
 
+	// Start a cleanup routine that will close and remove TSDB's that are no longer active
+	i.done.Add(1)
+	go func() {
```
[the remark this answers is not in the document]
thorfour: done.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+	}
+
+	s := i.createShipper(userID, i.cfg.TSDBConfig.BlocksDir(userID))
+	i.done.Add(1)
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer i.done.Done()
+		err := runutil.Repeat(i.cfg.TSDBConfig.ShipInterval, i.quit, func() error {
+			if uploaded, err := s.Sync(ctx); err != nil && err != context.Canceled {
+				level.Warn(util.Logger).Log("err", err, "uploaded", uploaded)
```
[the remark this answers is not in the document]
thorfour: done.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+			}
+			return nil
+		})
+		if err != nil {
+			level.Warn(util.Logger).Log("err", err, "msg", "error while shipping")
```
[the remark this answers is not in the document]
thorfour: I don't like panics in production code. I'd rather leave it as a log, especially since it's currently innocuous.

pkg/ingester/ingester_v2.go
```diff
@@ -427,9 +440,11 @@ func (i *Ingester) closeAllTSDB() {
 	for userID, db := range i.TSDBState.dbs {
 		userID := userID
 
-		go func(db *tsdb.DB) {
+		go func(db *UserTSDB) {
 			defer wg.Done()
 
+			db.shipperCancel()
```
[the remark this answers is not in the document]
thorfour: `nil` guard around it now.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+			defer wg.Done()
+			i.userStatesMtx.Lock()
+			defer i.userStatesMtx.Unlock()
+
+			blocksDir := i.cfg.TSDBConfig.BlocksDir(user)
+			snapDir := filepath.Join(blocksDir, "snapshot")
+
+			if db != nil {
+				// Stop the shipper routine for this TSDB
+				db.shipperCancel()
```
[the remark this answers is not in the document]
thorfour: `nil` guard and unit test to ensure it doesn't panic.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
 	i.userStatesMtx.Unlock()
 	wg.Wait()
 }
+
+/* cleanup walks the TSDB directory for each user to determine if the TSDB should be closed.
```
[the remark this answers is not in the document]
thorfour: done.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+ * If
+ * - The directory is empty of blocks
+ * OR
+ * - The directory has every block marked as shipped
+ * AND
+ * - The TSDB hasn't been written to within the allotted time period
+ * THEN
+ * - Close the TSDB, stop the shipper routine, and delete the directory and it's contents.
+ */
+func (i *Ingester) cleanup() error {
```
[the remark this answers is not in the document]
thorfour: done.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+			udb.lastWrittenLock.Unlock()
+
+			// TSDB has received a write within the window.
+			if now.Sub(last) <= i.cfg.TSDBConfig.MaxStaleAge {
+				continue
+			}
+		}
+
+		wg.Add(1)
+		go func(db *UserTSDB, user string) {
```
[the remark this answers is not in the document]
thorfour: nope, removed that.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+			defer wg.Done()
+			i.userStatesMtx.Lock()
```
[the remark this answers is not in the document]
thorfour: removed that and utilized a state variable to ensure no further accesses take place.

pkg/ingester/ingester_v2.go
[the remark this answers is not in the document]
thorfour: yes, added a new shipper wg for this that we wait for.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+				db.DisableCompactions()
+
+				// Snapshot the head/WAL to a block
+				if i.cfg.TSDBConfig.ShipInterval > 0 { // Snapshot only if we plan on shipping the block
+					if err := db.Snapshot(snapDir, true); err != nil {
+
+						// Restart shipper and compactor to roll-back the shutdown changes that have been made in the event this db becomes active again
+						db.EnableCompactions()
+						i.startShipper(user)
+						level.Warn(util.Logger).Log("msg", "snapshot failed", "err", err, "user", user)
```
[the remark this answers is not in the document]
thorfour: no longer applies as snapshots have been removed.

pkg/ingester/ingester_v2.go
```diff
@@ -449,3 +464,166 @@ func (i *Ingester) closeAllTSDB() {
...
+						return
+					}
+				}
+
+				// Close the TSDB
+				if err := db.Close(); err != nil {
+					// Restart shipper and compactor to roll-back the shutdown changes that have been made in the event this db becomes active again
```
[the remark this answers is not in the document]
thorfour: no longer applies as snapshots removed.

pkg/ingester/transfer.go
thorfour: N/A after refactor.

codesome:
Related to this: https://github.com/prometheus/prometheus/issues/6637

pracucci:
I've just merged the PR https://github.com/cortexproject/cortex/pull/1982. Please be aware we should now decrease the `memUsers` metric whenever a TSDB is removed.

codesome:
@thorfour: I just discussed with @pracucci offline about TSDB not compacting the last remaining data from Head after ingestion stops and possible solution in cortex (See https://github.com/prometheus/prometheus/issues/6637 as to why it won't be inbuilt in TSDB).

Here is what we have come up with as a solution (for TSDB which did not receive any sample for past X duration):
1. Reject appends on TSDB (blocked on the Cortex side)
2. Snapshot Head (I will add this to TSDB if this flow makes sense)
3. Move the new block to original data dir
4. Truncate Head (to remove data from memory) and reload the blocks.
5. Wait until the shipper has shipped all blocks to the storage (including the new one)
6. Close TSDB
7. Delete data from disk

WDYT?

pracucci:
@thorfour @codesome Another option, probably easier but less efficient, would be done like Thanos ([see code here](https://github.com/thanos-io/thanos/blob/master/pkg/receive/tsdb.go)): close TSDB, open it in read only, call `FlushWAL()`. Given having stale TSDBs shouldn't be the common case, we may trade efficiency with simplicity and follow Thanos' approach?

codesome:
In a case where user has stopped sending data but is still querying, it would lead to some downtime for the user depending on how long the WAL replay+flush takes. (Maybe we can take care of that if we actually see that happening often in real-world)

pracucci:
> In a case where user has stopped sending data but is still querying, it would lead to some downtime (EDIT: not downtime, but gaps) for the user depending on how long the WAL replay+flush takes. (Maybe we can take care of that if we actually see that happening often in real-world)

True, but re-thinking it maybe it's an edge case we could accept as far as we document it. Moreover all this logic should be able to disable setting the option to `0` (and maybe it could be a good thing having it disabled by default?).

thorfour:
> @thorfour: I just discussed with @pracucci offline about TSDB not compacting the last remaining data from Head after ingestion stops and possible solution in cortex (See [prometheus/prometheus#6637](https://github.com/prometheus/prometheus/issues/6637) as to why it won't be inbuilt in TSDB).
> 
> Here is what we have come up with as a solution (for TSDB which did not receive any sample for past X duration):
> 
>     1. Reject appends on TSDB (blocked on the Cortex side)
> 
>     2. Snapshot Head (I will add this to TSDB if this flow makes sense)
> 
>     3. Move the new block to original data dir
> 
>     4. Truncate Head (to remove data from memory) and reload the blocks.
> 
>     5. Wait until the shipper has shipped all blocks to the storage (including the new one)
> 
>     6. Close TSDB
> 
>     7. Delete data from disk
> 
> 
> WDYT?

This is similar to what I originally had implemented. I believe it works, but it's unfortunately complex code especially in the failure cases. We also have to go back to using a timestamp on writes to determine how long since a TSDB was last written which is unfortunate.

thorfour:
@codesome what if we added an optional parameter to the `Close` function in TSDB or maybe a new function like `CloseWithCompaction` that cuts a new block from the head while closing?

codesome:
@thorfour Maybe we can do the following which would not require this hack in place for upstream Prometheus: Refactor the `db.Compact()` into `CompactHead()` and `CompactBlocks()` (while `Compact()` calling these 2 methods). With that you can call `CompactHead()` in cortex and close the TSDB. Would that work for you?

pracucci:
> Maybe we can do the following which would not require this hack in place for upstream Prometheus: Refactor the db.Compact() into CompactHead() and CompactBlocks() (while Compact() calling these 2 methods). With that you can call CompactHead() in cortex and close the TSDB. Would that work for you?

@codesome As far as `CompactHead()` will check `db.head.compactable()` it will not work. The `Compact()` function is fine for us, but we would need a way to bypass `db.head.compactable()`: is there anything else we may need, that comes to your mind, to compact the entire head to a block?

codesome:
> As far as CompactHead() will check db.head.compactable() it will not work.

Yes I am aware of that. The plan is to have that check in `Compact()` while `CompactHead()` would take a `rangeHead` (while I will expose) and compact it without any checks.

pracucci:
> > As far as CompactHead() will check db.head.compactable() it will not work.
> 
> Yes I am aware of that. The plan is to have that check in `Compact()` while `CompactHead()` would take a `rangeHead` (which I will expose) and compact it without any checks.

Right. Then I think your proposed solution may work.

thorfour:
@codesome as long is it bypasses the head compatible check I think that solution would work!

codesome:
PR for breaking the `Compact()` method into parts https://github.com/prometheus/prometheus/pull/6820

thorfour:
While this feature is still needed, we need Thanos to vendor the latest Prometheus changes mentioned above, and then we need to vendor Thanos. Closing until these things are ready.

PR #1961, opened 2020-01-07
Revert "pkg/distributor: shardByAllLabels didn't check for label names order correctly"

gouthamve:
Reverts cortexproject/cortex#1957

We've deployed this into our env and are seeing 500s being thrown. Investigating.


gouthamve:
Turns out this might be due to an internal client and not Prometheus. Prometheus sends us data with labels sorted..

bboreham:
Labels are sorted once they hit the ingster, and we do use fast fingerprint to shard across ingesters.

Also note Prometheus did not sort labels before 2.0.

gouthamve:
Superseded by #1964

PR #1962, opened 2020-01-08
Website: fixed image aspect ratio for Twitter

pracucci:
**What this PR does**:
In the PR #1959 I've introduced the image for the Twitter card, but I didn't consider the required aspect ratio (which is about 1:2), so the resulting tweet is something like this:

Screen Shot 2020-01-08 at 09 13 59

In this PR I've created an image with the Cortex logo and the right aspect ratio for Twitter. It should fixes it.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


gouthamve:
Do we really need a 407KB image for it? Can you stick it through http://webresizer.com/resizer/?

pracucci:
> Do we really need a 407KB image for it? 

@gouthamve Not need, but usually not a problem being an image downloaded only from crawlers. Anyway, rescaled it.

PR #1964, opened 2020-01-08
Validate incoming labels for order and duplicate names.

pstibrany:
~Originally this PR added only check for sorted labels.~ This PR sorts unsorted labels in distributor, before using them for computing sharding token. It also adds check if there are duplicate label names, and reports them.

To properly format the input, it reimplements formatting function. Previously used function merged duplicate labels, and also sorted labels, so error message didn't include real input, but "normalized" version already. That's not very useful for debugging.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

Related to #1957 and #1961 which reverts it. This builds on #1957 and fixes it, while ignoring #1961.


gouthamve:
Given the remote_write API doesn't specify ordering and because Prometheus 1.x sends unordered labels, I think rejecting unordered labels is not the approach we want to take. Maybe we can sort them, but lets not reject them.

pstibrany:
To summarize our Slack discussion:

Reason to make this PR was to make sure that sharding token is correct when sharding using all labels. Ordering issues or duplicate label names generate different tokens (eg. `test{a="a",b="b"}`, `test{b="b",a="a"}` and `test{a="a",b="b",b="b"}` have different token and end up on different ingesters, but we currently accept them as a same timeseries.

There is no question about that duplicate labels should be rejected. In order to detect them cheaply, I added check for ordering. I assumed that since the code was already there (although not working properly, hence #1957), that is what we wanted.

If we don't want to reject unordered labels, then I suggest that we simply order them in distributor. This will help us to 1) fix above-mentioned problem with token, while not resharding the world again (by changing the sharding computation), 2) still detect duplicate label names cheaply.

I'll update my PR to implement this change instead.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
...
 * [CHANGE] Overrides mechanism has been renamed to "runtime config", and is now separate from limits. Runtime config is simply a file that is reloaded by Cortex every couple of seconds. Limits and now also multi KV use this mechanism.<br />New arguments were introduced: `-runtime-config.file` (defaults to empty) and `-runtime-config.reload-period` (defaults to 10 seconds), which replace previously used `-limits.per-user-override-config` and `-limits.per-user-override-period` options. Old options are still used if `-runtime-config.file` is not specified. This change is also reflected in YAML configuration, where old `limits.per_tenant_override_config` and `limits.per_tenant_override_period` fields are replaced with `runtime_config.file` and `runtime_config.period` respectively. #1749
+* [CHANGE] Cortex now rejects data with duplicate labels. Previously, such data was accepted, with duplicate labels removed with only one value left. #1964
```
pracucci: `duplicate` > `duplicated`

pkg/distributor/distributor.go
```diff
@@ -450,6 +452,28 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
 }
 
+func sortLabelsIfNeeded(labels []client.LabelAdapter) {
+	// no need to run sort.Slice, if labels are already sorted, which is most of the time.
+	// we can avoid extra memory allocations (mostly interface-related) this way.
+	sorted := true
+	last := ""
+	for _, l := range labels {
+		if strings.Compare(last, l.Name) > 0 {
+			sorted = false
```
pracucci: I think could be a bit simplified directly `return` from here. It will remove the `sorted` variable and the `if sorted` below.

pkg/util/validation/validate.go
```diff
@@ -14,14 +16,16 @@ import (
...
-	errTooNew            = "sample for '%s' has timestamp too new: %d"
+	errMissingMetricName  = "sample missing metric name"
+	errInvalidMetricName  = "sample invalid metric name: %.200q"
+	errInvalidLabel       = "sample invalid label: %.200q metric %.200q"
+	errLabelNameTooLong   = "label name too long: %.200q metric %.200q"
+	errLabelValueTooLong  = "label value too long: %.200q metric %.200q"
+	errTooManyLabels      = "sample for '%s' has %d label names; limit %d"
+	errTooOld             = "sample for '%s' has timestamp too old: %d"
+	errTooNew             = "sample for '%s' has timestamp too new: %d"
+	errDuplicateLabelName = "duplicate label name: %.200q metric %.200q"
```
pracucci: `duplicate` > `duplicated`

_same comment applies below_

pkg/distributor/distributor.go
pstibrany: The logic is opposite though. If not sorted, then we don't return but sort.

pracucci: Damn, sure, you're right. Forgive my comment.

pkg/util/validation/validate.go
pstibrany: I think both work fine here. Duplicate is an adjective, so duplicate label is valid expression. I believe :) ... but given that english is a second language for me, I can be convinced.

pkg/distributor/distributor.go
```diff
@@ -26,7 +25,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/limiter"
 	"github.com/cortexproject/cortex/pkg/util/validation"
-	billing "github.com/weaveworks/billing-client"
```
gouthamve: Why did this change? I prefer the package being explicit.

pkg/util/validation/validate.go
gouthamve: Duplicate and Duplicated both work and I kinda prefer Duplicate :)

CHANGELOG.md
gouthamve: I think both are fine tbh. No preference here.

pkg/distributor/distributor.go
pstibrany: My IDE flagged it as "redundant alias" and suggested to remove it and I prefer to get rid of warnings when I can. So I did and nothing broke. I can put it back, if you want (and adjust my IDE warnings setting :))

gouthamve: Please put it back :)

pstibrany: It's back.

pstibrany:
> Can you open this against the `release-0.5.0` branch?

Rebased on top of `release-0.5` and changed target branch for this PR.

PR #1965, opened 2020-01-08
Do not display the last modified timestamp in the website pages

pracucci:
**What this PR does**:
In the website, the last modified timestamp is broken and always displays `January 1, 0001` (see screenshot). I spent almost an hour trying to understand what's wrong with getting git info and I couldn't figure out, so in this PR I'm proposing to just remove it (given doesn't bring much value).

Screen Shot 2020-01-08 at 18 50 49

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
I always chucked when I saw that. Too bad it will be gone :)

PR #1971, opened 2020-01-10
Added the series to the error when a sample fail to be appended to TSDB

pracucci:
**What this PR does**:
I'm assisting an user on Slack, and I've realized we don't log the series whose sample fails to be appended to TSDB. I've switched the logic from "last partial error" to "first partial error" in order for performance reasons (in case many samples fail in a single write request).

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/validation"
 	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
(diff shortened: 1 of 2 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -135,7 +136,10 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 			// 400 error to the client. The client (Prometheus) will not retry on 400, and
 			// we actually ingested all samples which haven't failed.
 			if err == tsdb.ErrOutOfBounds || err == tsdb.ErrOutOfOrderSample || err == tsdb.ErrAmendSample {
-				lastPartialErr = err
+				if firstPartialErr == nil {
+					firstPartialErr = errors.Wrapf(err, "series=%s", lset.String())
```
gouthamve: This seems like premature optimisation. Do we know if this is indeed a problem? I would think that over the other allocs and network calls, this won't even show up.

pracucci: I see your point. Let me show you a different angle. We need to keep 1 error, either we keep the first one or last one doesn't make any difference (and practically same code). Keeping the first one saves few cycles in case there are several partial errors, so look a low hanging fruit we can get.

codesome: Could the last error also be a follow up of the first error at times? If yes, another reason to return the first error.

pracucci: Good point @codesome. I also thought about it, and my conclusion has been that these errors should not depend on a previous error, but I can't guarantee it. Generally speaking, I believe the first error is more significative than the last one.

PR #1974, opened 2020-01-10
frontend: implement cache control

GiedriusS:
:wave: from the Thanos project. Thanks, first of all, for your work on making the frontend component work with regular Prometheus end-points! We would like to extend it a bit so that it would fit better for use with Thanos. One of the things is that we would like to avoid caching the results when Thanos gets what is called a "partial response" - when one of the nodes does not respond but some end-result with partial data is still returned to the user. Obviously, such data is incomplete and might very well change after a minute or two so it must not be cached. 

Thus, let's capture the headers from the response as well. They will be used to determine if the result should be cached. For that reason two new variables are introduced: 
* `cachecontrolHeader` which is equal to `Cache-Control` - this is the header that will be checked
* `noCacheValue` which is equal to `no-store` - that is the value that we are looking for to determine if the result should be cached

More documents/ideas are available here: https://github.com/thanos-io/thanos/issues/1651

Example user: https://github.com/thanos-io/thanos/pull/1984. I actually tested that this works as expected with that PR applied on top of `master`.

Added a simple test for the function which determines if the result should be cached based on the response's headers.

All ideas/feedback/suggestions welcome as I have never contributed to Cortex thus I am unfamiliar with the codebase. Also, I wonder if you would be even open to accepting something like this. Obviously this is missing such things like documentation & E2E tests, and so on so let's discuss this before moving further. Thanks a lot!


pkg/querier/queryrange/results_cache.go
```diff
@@ -133,12 +139,37 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
...
+	if promResp, ok := r.(*PrometheusResponse); ok {
+		shouldNotCache := false
+		for _, hv := range promResp.Headers {
+			if hv == nil {
+				continue
+			}
+			if hv.Name != cachecontrolHeader {
+				continue
+			}
+			if hv.Values[0] == noCacheValue {
```
pracucci: This matches only for `Cache-Control: no-store`, while the `Cache-Control` header value could be a bit more articulated (ie. `Cache-Control: private, no-store`). Maybe we could make it a bit more robust:

1. Looking for `no-store` within the header value
2. Checking all header values (being a list)

GiedriusS: Good point, added a check for all header's values :smile:

pkg/querier/queryrange/results_cache.go
```diff
@@ -133,12 +139,41 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
 	return response, err
 }
 
+// shouldCacheResponse says whether the response should be cached or not.
+func shouldCacheResponse(r Response) bool {
+	if promResp, ok := r.(*PrometheusResponse); ok {
+		shouldNotCache := false
```
gouthamve: Just a minor nit, but I think it'll be clearer if we just do `shouldCache:= true` and setting it to false and just `return shouldCache`.

gouthamve:
Could you also add changelog entry?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] The frontend component now does not cache results if it finds a `Cache-Control` header and if one of its values is `no-store`
```
gouthamve: Full stop and add the PR number too please :)

gouthamve: And a rebase

GiedriusS: That was fast :smile: will do.

Issue #1920, opened 2019-12-17
Table-manager graceperiod bug with new schema


slim-bean:
We deployed a new schema on a periodic table config, but deployed it not on the periodic table rotation day.  When the new table was created, it was immediately deleted leading to nowhere to write index entries and OOM death of all the ingesters in the cluster.

I briefly looked at what happened and based on the logs it definitely looks related to the grace period config for table creation.  Our environment had a 3 hour grace period configured and here were the logs:

```
level=info ts=2019-12-10T20:39:20.402311965Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T20:49:20.402248128Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T20:59:20.402287276Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T21:09:20.402451685Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T21:10:42.619662501Z caller=table_manager.go:374 msg="table has exceeded the retention period" table=loki_prod_index_2605
level=info ts=2019-12-10T21:10:42.61972737Z caller=table_manager.go:379 msg="deleting table" table=loki_prod_index_2605
level=info ts=2019-12-10T21:10:42.862172893Z caller=table_manager.go:363 msg="creating table" table=loki_prod_index_2595
level=info ts=2019-12-10T21:19:20.402302368Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T21:29:20.402288783Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T21:39:20.402273464Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T21:49:20.402482189Z caller=table_manager.go:220 msg="synching tables" expected_tables=10
level=info ts=2019-12-10T21:50:53.380202176Z caller=signals.go:54 msg="=== received SIGINT/SIGTERM ===\n*** exiting"
level=info ts=2019-12-10T21:50:53.380301092Z caller=loki.go:175 msg="notifying module about stopping" module=table-manager
level=info ts=2019-12-10T21:50:53.380324053Z caller=loki.go:175 msg="notifying module about stopping" module=server
[...]

PR #1976, opened 2020-01-13
fix calculation of expected tables and create tables from upcoming schema considering grace period

sandlis:
**What this PR does**:
This PR includes following to fixes:

1. Fix calculation of expected tables:
While calculating periodic tables for a given duration, the end time of the duration was used to find what the last table number would be. To avoid creating an extra table, there was a check whether end duration was ending exactly at end of the day which was causing the whole table to be omitted. Instead of checking against a days duration, this PR uses periodic table duration to check whether we have an extra table.

2. Create tables from upcoming schema considering grace period:
We skip creating tables for configs whose start time is after `now()` which means tables would not be created until we already are past the start time of config. This can cause writes to fail since tables won't be created already for writing data to it.
This PR checks whether the config start time is before `now() + gracePeriod` when creating tables.

Fixes: https://github.com/cortexproject/cortex/issues/1920
In #1920, the start time of config was not set to the day when tables would rotate. When the time was just short by grace period before hitting next config, [this](https://github.com/cortexproject/cortex/blob/8310aa24fe176c27de9bd51fb92bdb55b524abad/pkg/chunk/table_manager.go#L287-L290) block started calculating tables for duration from `start-of-active-schema` to `start-of-next-schema`. Since the end was not a multiple of table duration but was a multiple of a days duration, issue 1 above was causing table manager to unexpectedly drop the active table.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/schema_config.go
+++ b/pkg/chunk/schema_config.go
@@ -424,7 +424,7 @@ func (cfg *PeriodicTableConfig) periodicTables(from, through model.Time, pCfg Pr
 		nowWeek        = now / periodSecs
 		result         = []TableDesc{}
 	)
-	// Make sure we don't have an extra table
+	// If interval ends exactly on a period boundary, don’t include the upcoming period
 	if through.Unix()%periodSecs == 0 {
 		lastTable--
 	}
--- a/pkg/chunk/schema_config.go
+++ b/pkg/chunk/schema_config.go
@@ -424,8 +424,8 @@ func (cfg *PeriodicTableConfig) periodicTables(from, through model.Time, pCfg Pr
 		nowWeek        = now / periodSecs
 		result         = []TableDesc{}
 	)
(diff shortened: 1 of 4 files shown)


pkg/chunk/schema_config.go
```diff
@@ -424,8 +424,8 @@ func (cfg *PeriodicTableConfig) periodicTables(from, through model.Time, pCfg Pr
 		nowWeek        = now / periodSecs
 		result         = []TableDesc{}
 	)
-	// If through ends on 00:00 of the day, don't include the upcoming day
-	if through.Unix()%secondsInDay == 0 {
+	// Make sure we don't have an extra table
```
bboreham: I think this comment is now less explanatory. Could say something like “If interval ends exactly on a period boundary, don’t include the upcoming period”.

PR #1981, opened 2020-01-15
TSDB: metrics for queries

pstibrany:
This PR updates some existing ingester metrics from blocks code. Specifically:

We already update ingestion metrics from blocks code (`cortex_ingester_ingested_samples_total`, `cortex_ingester_ingested_samples_failures_total`)

Querying:
- `cortex_ingester_queries_total` (this was already tracked before this PR)
- `cortex_ingester_queried_samples` (added by this PR)
- `cortex_ingester_queried_series` (added by this PR)
- Ignored metric: `cortex_ingester_queried_chunks` – chunks-specific

User state:
- `cortex_ingester_memory_users` (~added by this PR~ [update: now removed again, in favor of PR #1982), basically equals to number of open TSDB databases in Cortex)
- Ignored: `cortex_ingester_memory_series` – cannot see easy way to get this, not even in TSDB metrics
- Ignored: `cortex_ingester_memory_series_created_total`, `cortex_ingester_memory_series_removed_total` (these are per-user) – again, no obvious/easy way to get this from TSDB

There are some more metrics that are updated when flushing chunks. This PR doesn't update these metrics and there seems to be no obvious way how we could do that:
- `cortex_ingester_chunk_age_seconds`
- `cortex_ingester_chunk_length`
- `cortex_ingester_chunk_size_bytes`
- `cortex_ingester_chunk_utilization`
- `cortex_ingester_chunks_created_total`
- `cortex_ingester_dropped_chunks_total`
- `cortex_ingester_memory_chunks` (total chunks in memory)

Additional cortex_ingester metrics are related to transfer of data between ingesters. This PR doesn't change them, but for completeness, here they are:
- `cortex_ingester_received_bytes_total`, `cortex_ingester_received_files`, `cortex_ingester_sent_bytes_total`, `cortex_ingester_sent_files` - updated when transferring TSDB's between ingesters
- `cortex_ingester_received_chunks`, `cortex_ingester_sent_chunks` - updated when transferring chunks between ingesters

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -194,8 +194,12 @@ func (i *Ingester) v2Query(ctx old_ctx.Context, req *client.QueryRequest) (*clie
 		return nil, err
 	}
 
+	numSeries := 0
+	numSamples := 0
+
 	result := &client.QueryResponse{}
 	for ss.Next() {
+		numSeries++
 		series := ss.At()
 
 		ts := client.TimeSeries{
@@ -204,12 +208,16 @@ func (i *Ingester) v2Query(ctx old_ctx.Context, req *client.QueryRequest) (*clie
 
 		it := series.Iterator()
 		for it.Next() {
+			numSamples++
 			t, v := it.At()
 			ts.Samples = append(ts.Samples, client.Sample{Value: v, TimestampMs: t})
 		}
 
 		result.Timeseries = append(result.Timeseries, ts)
 	}
+
+	i.metrics.queriedSeries.Observe(float64(numSeries))
+	i.metrics.queriedSamples.Observe(float64(numSamples))
 
 	return result, ss.Err()
 }
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -194,12 +194,10 @@ func (i *Ingester) v2Query(ctx old_ctx.Context, req *client.QueryRequest) (*clie
 		return nil, err
 	}
 
-	numSeries := 0
 	numSamples := 0
 
 	result := &client.QueryResponse{}
 	for ss.Next() {
-		numSeries++
 		series := ss.At()
 
 		ts := client.TimeSeries{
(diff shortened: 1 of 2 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -204,13 +208,17 @@ func (i *Ingester) v2Query(ctx old_ctx.Context, req *client.QueryRequest) (*clie
 
 		it := series.Iterator()
 		for it.Next() {
+			numSamples++
```
codesome: It would be more efficient to do `numSamples += len(ts.Samples)` after the loop.

pkg/ingester/ingester_v2.go
```diff
@@ -204,13 +208,17 @@ func (i *Ingester) v2Query(ctx old_ctx.Context, req *client.QueryRequest) (*clie
...
 			t, v := it.At()
 			ts.Samples = append(ts.Samples, client.Sample{Value: v, TimestampMs: t})
 		}
 
 		result.Timeseries = append(result.Timeseries, ts)
 	}
 
+	i.metrics.queriedSeries.Observe(float64(numSeries))
```
codesome: We can directly use `len(result.Timeseries)` here instead of adding a counter.

pkg/ingester/ingester_v2.go
pstibrany: Right. Fixed both places.

pkg/ingester/ingester_v2.go
```diff
@@ -413,6 +419,7 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error)
 	}
 
 	i.TSDBState.dbs[userID] = db
+	memUsers.Inc()
```
gouthamve: This is better handled in https://github.com/cortexproject/cortex/pull/1982

pstibrany: Right. I will rebase and remove this.

pstibrany: Aah, #1982 is not yet merged. I was wondering why I don't see conflicts yet :)

pstibrany: Removed from this PR now.

pracucci: Please remind to remove mentioning `cortex_ingester_memory_users` from changelog too.

thorfour:
Thanks for doing that!

PR #1983, opened 2020-01-15
TSDB: Expose shipper metrics via ingester

pstibrany:
Since we have one TSDB shipper per tenant in memory, we need to gather all of these metrics and sum them together (fortunately they are all counters).

One shipper metric, `thanos_shipper_upload_compacted_done` ("If 1 it means shipper uploaded all compacted blocks from the filesystem.") was skipped and is not exposed, as there is no good way to group it across all tenants.

All shipper metrics now have `cortex_ingester` prefix instead of original `thanos` prefix.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/ingester/ingester_v2.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/validation"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
+	io_prometheus_client "github.com/prometheus/client_model/go"
```
pracucci: If you look at Prometheus sources, this import is named `dto`. Maybe we could keep the same naming for consistency.

pkg/ingester/ingester_v2.go
```diff
@@ -58,11 +62,16 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
...
-			dbs:    make(map[string]*tsdb.DB),
-			bucket: bucketClient,
+			dbs:               make(map[string]*tsdb.DB),
+			bucket:            bucketClient,
+			shipperRegistries: make(map[string]*prometheus.Registry),
 		},
 	}
 
+	if registerer != nil {
+		registerer.MustRegister(i)
```
pracucci: I've the feeling having the `Ingester` to be a `Collector` is not a common pattern. What do you think if we wrap the logic into a custom struct (ie. `shipperMetrics`) which is a `Collector`, while keeping the same design?

pkg/ingester/ingester_v2.go
```diff
@@ -397,7 +406,12 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*tsdb.DB, error)
 
 	// Create a new shipper for this database
 	if i.cfg.TSDBConfig.ShipInterval > 0 {
-		s := shipper.New(util.Logger, nil, udir, &Bucket{userID, i.TSDBState.bucket}, func() labels.Labels { return l }, metadata.ReceiveSource)
+		reg := prometheus.NewRegistry()
```
pracucci: If we have a custom struct for it, we could move this logic in a func like `NewRegistry(userID)` and then just call `shipper.New(..., x.NewRegistry(userID), ...)`.

pkg/ingester/ingester_v2.go
pstibrany: What do you think about simply using existing `ingesterMetrics` for this?

pracucci: Looks good. Given the quantity of code, may be worth to move it a dedicated file `pkg/ingester/metrics.go`?

pstibrany: I actually quite like `shipperMetrics`, because then it's clear that it's TSDB specific and doesn't mix with chunks code.

pstibrany: I've moved the code to a new file (incl. ingesterMetrics) and type (`shipperMetrics` now). I've also addressed other feedback. Please take a look.

pkg/ingester/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	out <- prometheus.MustNewConstMetric(sm.dirSyncFailures, prometheus.CounterValue, sumCounters(gathered["thanos_shipper_dir_sync_failures_total"]))
+	out <- prometheus.MustNewConstMetric(sm.uploads, prometheus.CounterValue, sumCounters(gathered["thanos_shipper_uploads_total"]))
+	out <- prometheus.MustNewConstMetric(sm.uploadFailures, prometheus.CounterValue, sumCounters(gathered["thanos_shipper_upload_failures_total"]))
+}
+
+func (sm *shipperMetrics) shipperRegistries() []*prometheus.Registry {
+	sm.regsMu.RLock()
+	defer sm.regsMu.RUnlock()
+
+	regs := []*prometheus.Registry(nil)
```
thorfour: `nit` this looks weird to my eyes. I'd prefer `regs := []*prometheus.Registry{}`

thorfour: Actually is there a reason they're being copied out? Couldn't we just return `sm.regs` ?

pkg/ingester/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+			m.queriedSamples,
+			m.queriedSeries,
+			m.queriedChunks,
+		)
+	}
+
+	return m
+}
+
+// TSDB shipper metrics
```
thorfour: maybe just add a further comment explaining this is a multi-tenant set of metrics that get aggregated into a single set of metrics?

pkg/ingester/metrics.go
pstibrany: > Actually is there a reason they're being copied out? Couldn't we just return `sm.regs` ?

We would need to keep the lock while doing that. I don't like the idea of metrics code blocking ingestion path (when opening new TSDB), and opted for copying here, but perhaps it's not a big deal. Not sure.

When it comes to declaration, either `nil` version or `var regs []*prometheus.Registry` are fine, but `regs := []*prometheus.Registry{}` allocates non-nil zero-length slice, which is just useless allocation.

pracucci: I think keeping the current code - which reduce the time the lock is taken - is better.

pkg/ingester/metrics.go
pstibrany: I've updated the comment. Is that what you had in mind?

pkg/ingester/metrics.go
thorfour: Okay, leave as is.

pkg/ingester/metrics.go
thorfour: Looks great, thanks!

pstibrany: Thank you for your review!

pkg/ingester/metrics.go
pstibrany: One other possibility would be to cache the slice of registries, and nil it when adding new user. We can do that if making a copy for each metrics becomes a problem, which is unlikely.

pkg/ingester/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (sm *shipperMetrics) Describe(out chan<- *prometheus.Desc) {
+	out <- sm.dirSyncs
+	out <- sm.dirSyncFailures
+	out <- sm.uploads
+	out <- sm.uploadFailures
+}
+
+func (sm *shipperMetrics) Collect(out chan<- prometheus.Metric) {
```
pracucci: Have you considered adding a test on `shipperMetrics.Collect()` to cover at least the happy path? It shouldn't be much complicated if we use `testutil.GatherAndCompare()` to test on it.

pstibrany: Added test now.

pkg/ingester/metrics.go
```diff
@@ -0,0 +1,211 @@
...
+	gathered := make(map[string][]*dto.MetricFamily)
+
+	regs := sm.shipperRegistries()
+	for userID, r := range regs {
+		m, err := r.Gather()
+		if err != nil {
+			level.Debug(util.Logger).Log("msg", "failed to gather metrics from TSDB shipper", "user", userID, "err", err)
```
pracucci: I would `Warn` instead of `Debug`. This shouldn't happen and - if happens - we get unreliable metrics.

pstibrany: I've changed it. Although user cannot really do anything about it, at least we will see it in logs easier. Hopefully it never happens.

pracucci: > at least we will see it in logs easier

Exactly, that's the point. If it happens it's likely to be a bug. Noticing would allow the user to submit us an issue.

gouthamve:
That was very cleverly done!

PR #1987, opened 2020-01-15
20190115 ruler better flags

### LINKED ISSUE #1986: Unfriendly Ruler ring flags
Since the default RegisterFlags function is used, the help text that accompanies the flags is a bit confusing. For instance:

```
  -ruler.distributor.replication-factor int
    	The number of ingesters to write to and read from. (default 3)
```

This flag above does not have a sensible description and must be defaulted to 1 by the ruler either way.

jtlisi:
**What this PR does**:

The current ruler config exposes a number of flags that should be set to default values. For instance the ruler should never be run with a replication factor >1. However, as things stand currently the flag defaults to 3 which makes configuring the sharded ruler quite difficult. This PR adds a Ruler RingConfig struct based on the Distributor RingConfig that simplifies the ring flags for the ruler.

* Adds a guide on how to set up a sharded ruler
* Add custom ruler ring config to remove unused/dangerous lifecylcer flags

Fixes #1986 

**Checklist**
- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/guides/sharded_ruler.md
+++ b/docs/guides/sharded_ruler.md
@@ -1,0 +1,26 @@
+---
+title: "Config for horizontally scaling the Ruler"
+linkTitle: "Config for horizontally scaling the Ruler"
+weight: 4
+slug: ruler-sharding
+---
+
+## Context
+
+One option to scale the ruler is by scaling it horizontally. However, with multiple ruler instances running they will need to coordinate to determine which instance will evaluate which rule. Similar to the ingesters, the rulers establish a hash ring to divide up the responsibilities of evaluating rules. 
+
+## Config
+
+In order to enable sharding in the ruler the following flag needs to be set:
+
+```
+  -ruler.enable-sharding=true
+```
+
+In addition the ruler requires it's own ring to be configured, for instance:
+
+```
+  -ruler.ring.consul.hostname=consul.dev.svc.cluster.local:8500
+```
+
+The only configuration that is required is to enable sharding and configure a key value store. From there the rulers will shard and handle the division of rules automatically.
--- a/pkg/ruler/ruler.go
+++ b/pkg/ruler/ruler.go
@@ -61,14 +61,14 @@ type Config struct {
 
 	EnableSharding   bool // Enable sharding rule groups
 	SearchPendingFor time.Duration
-	RulerRing        RingConfig
+	Ring             RingConfig
 	FlushCheckPeriod time.Duration
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.StoreConfig.RegisterFlags(f)
-	cfg.RulerRing.RegisterFlags(f)
+	cfg.Ring.RegisterFlags(f)
 
 	// Deprecated Flags that will be maintained to avoid user disruption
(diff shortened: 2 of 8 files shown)


gouthamve:
Could you open this against `release-0.5` branch?

bboreham:
I don't think it's reasonable to inject a breaking change to release 0.5.0 after we've had two release candidates out.
@gouthamve perhaps you meant just to change the docs?

docs/guides/sharded_ruler.md
```diff
@@ -0,0 +1,26 @@
...
+linkTitle: "Config for horizontally scaling the Ruler"
+weight: 4
+slug: ruler-sharding
+---
+
+## Context
+
+One option to scale the ruler is by scaling it horizontally. However, with multiple ruler instances running they will need to coordinate to determine which instance will evaluate which rule. Similar to the ingesters, the rulers establish a hash ring to divide up the responsibilities of evaluating rules. 
+
+## Config
```
pracucci: After rebasing and `make doc` you will also get the config file documentation updated. Could be worth adding a link to that doc (each root config block has an anchor, so you could link it to `#ruler-config`) mentioning it like `To see the complete set of config option, please check out...`

pkg/ruler/ruler_ring.go
```diff
@@ -0,0 +1,92 @@
...
+// many options not really required by the rulers ring. This config
+// is used to strip down the config to the minimum, and avoid confusion
+// to the user.
+type RingConfig struct {
+	KVStore          kv.Config     `yaml:"kvstore,omitempty"`
+	HeartbeatPeriod  time.Duration `yaml:"heartbeat_period,omitempty"`
+	HeartbeatTimeout time.Duration `yaml:"heartbeat_timeout,omitempty"`
+
+	// Instance details
+	InstanceID             string   `yaml:"instance_id"`
```
pracucci: Like we did in the `DistributorConfig`, I would suggest to add a `doc:"hidden"` to all `Instance*` fields.

pkg/ruler/ruler_ring.go
```diff
@@ -0,0 +1,92 @@
...
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *RingConfig) RegisterFlags(f *flag.FlagSet) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "failed to get hostname", "err", err)
+		os.Exit(1)
+	}
+
+	// Ring flags
+	cfg.KVStore.RegisterFlagsWithPrefix("ruler.ring.", "rulers/", f)
```
pracucci: This is another breaking change that we should mention in the changelog. The prefix has changed from `collectors/` to `rulers/`.

jtlisi:
@pracucci Thanks for the feedback. I addressed your comments and this should be good for a second look.

pkg/ruler/ruler.go
```diff
@@ -61,14 +61,14 @@ type Config struct {
 
 	EnableSharding   bool // Enable sharding rule groups
 	SearchPendingFor time.Duration
-	LifecyclerConfig ring.LifecyclerConfig
+	RulerRing        RingConfig
```
pracucci: The resulting yaml field name is `rulerring`, which looks a bit repetitive considering we're already inside the `ruler: {}` config. What's your take adding a `yaml:"ring"`? In case you decide to keep the current name, then `ruler_ring` may be better (easier to read for the user).

PR #1989, opened 2020-01-15
sharded ruler guide with simplified ring config/flags

jtlisi:


bboreham:
Is your intention to add a breaking change to release 0.5.0, after we've had two release candidates out?
I don't think that flies.

PR #1991, opened 2020-01-16
Improve TSDB blocks sync

pracucci:
**What this PR does**:
Currently, there's an hardcoded `30s` as blocks sync interval (TSDB storage). It's quite a lot. I believe we don't even need such an high-frequency sync interval (at least, not as default) so I'm proposing to allow to configure it and set an higher default (Thanos' default is `3m`, I've set `5m` which should be more than enough for the Cortex use case).

I've also received an error from the linter because the `rununtil.Repeat()` error wasn't checked, so I took the change to refactor it (moved to `UserStore` since it belongs to it), renamed `cortex_querier_sync_seconds` to `cortex_querier_blocks_sync_seconds` and customized the buckets: default values are too low for this operation in a medium/large cluster.

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/querier/block.go
+++ b/pkg/querier/block.go
@@ -57,7 +57,7 @@ func NewBlockQuerier(cfg tsdb.Config, logLevel logging.Level, r prometheus.Regis
 	level.Info(util.Logger).Log("msg", "successfully synchronized TSDB blocks for all users")
 
 	stopc := make(chan struct{})
-	go runutil.Repeat(30*time.Second, stopc, func() error {
+	go runutil.Repeat(cfg.BucketStore.SyncInterval, stopc, func() error {
 		ts := time.Now()
 		if err := us.SyncStores(context.Background()); err != nil && err != io.EOF {
 			level.Warn(util.Logger).Log("msg", "sync stores failed", "err", err)
--- a/pkg/querier/block.go
+++ b/pkg/querier/block.go
@@ -28,7 +28,12 @@ type BlockQuerier struct {
 
 // NewBlockQuerier returns a client to query a block store
 func NewBlockQuerier(cfg tsdb.Config, logLevel logging.Level, registerer prometheus.Registerer) (*BlockQuerier, error) {
-	us, err := NewUserStore(cfg, logLevel, util.Logger, registerer)
+	bucketClient, err := tsdb.NewBucketClient(context.Background(), cfg, "cortex-userstore", util.Logger)
+	if err != nil {
+		return nil, err
+	}
+
+	us, err := NewUserStore(cfg, bucketClient, logLevel, util.Logger, registerer)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/querier/block.go
+++ b/pkg/querier/block.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"time"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
 	"github.com/cortexproject/cortex/pkg/chunk/encoding"
@@ -17,7 +16,6 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
-	"github.com/thanos-io/thanos/pkg/runutil"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
 	"github.com/weaveworks/common/logging"
 	"google.golang.org/grpc/metadata"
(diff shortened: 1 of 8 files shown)


pkg/querier/block.go
```diff
@@ -55,7 +55,7 @@ func NewBlockQuerier(cfg tsdb.Config, logLevel logging.Level, r prometheus.Regis
 	level.Info(util.Logger).Log("msg", "successfully synchronized TSDB blocks for all users")
 
 	stopc := make(chan struct{})
-	go runutil.Repeat(30*time.Second, stopc, func() error {
+	go runutil.Repeat(cfg.BucketStore.SyncInterval, stopc, func() error {
```
pstibrany: There should probably be some validation of `SyncInterval` being >= 0.

pracucci: Right, done.

pracucci:
I've received an error from the linter because the rununtil.Repeat() error wasn't checked, so I took the change to refactor it (moved to UserStore since it belongs to it), renamed cortex_querier_sync_seconds to cortex_querier_blocks_sync_seconds and customized the buckets: default values are too low for this operation in a medium/large cluster.

@pstibrany May you re-review it, please?
@thorfour What's your take?

pracucci:
> when using target=All, initial sync of the querier will block entire startup of the single-binary Cortex. Is that right?

@pstibrany That's correct. On the other end, you can't query Cortex until the initial sync is done (but you could ingest), so not sure what's the best way to handle it.

pkg/querier/block_store.go
```diff
@@ -76,8 +103,34 @@ func (u *UserStore) InitialSync(ctx context.Context) error {
 	return nil
 }
 
-// SyncStores iterates over the storage bucket creating user bucket stores
-func (u *UserStore) SyncStores(ctx context.Context) error {
+// syncStoresLoop periodically calls syncStores() to synchronize the blocks for all tenants.
+func (u *UserStore) syncStoresLoop() {
```
thorfour: This was existent in old code, but since `runutil.Repeat` runs f right after being called, we effectively sync twice immediately. I think it might make sense to put the initial sync in this function. Then have a `time.After` where it waits to start the repeat function, so there's a little room between initial sync and the repeating sync. Thoughts?

pkg/storage/tsdb/config.go
```diff
@@ -127,3 +130,12 @@ func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 func (cfg *Config) BlocksDir(userID string) string {
 	return filepath.Join(cfg.Dir, userID)
 }
+
+// Validate the config
+func (cfg *BucketStoreConfig) Validate() error {
+	if cfg.SyncInterval <= 0 {
```
thorfour: I'd rather `syncInterval <= 0` be a disabled case. We currently rung with syncing disabled, and only send queries to ingesters in our testing, and it makes it more synonymous with Thanos where the LTS is optional.

pkg/querier/block_store.go
pracucci: I think you're right. Will work on it.

pkg/storage/tsdb/config.go
pracucci: Got it. Will do.

pracucci:
@pstibrany @thorfour I've re-iterated on this PR based on your feedback (allow to disable sync, do not synch twice at startup, fix race condition) plus a couple of other changes (max concurrency when synching multi stores, track timing for initial sync too). May you re-review it, please?

pkg/querier/block_store.go
```diff
@@ -26,25 +30,53 @@ import (
...
+		workersCancel: workersCancel,
+		syncTimes: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name:    "cortex_querier_blocks_sync_seconds",
+			Help:    "The total time it takes to perform a sync stores",
+			Buckets: []float64{0.1, 1, 10, 30, 60, 120, 300, 600, 900},
+		}),
+	}
+
+	// Configure the time range to sync all blocks.
+	if err := u.syncMint.Set("0000-01-01T00:00:00Z"); err != nil {
```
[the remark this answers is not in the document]
pracucci: It just doesn't cause any harm having it.

pkg/querier/block_store.go
```diff
@@ -88,78 +176,56 @@ func (u *UserStore) SyncStores(ctx context.Context) error {
 }
 
 func (u *UserStore) syncUserStores(ctx context.Context, f func(context.Context, *store.BucketStore) error) error {
-	mint, maxt := &model.TimeOrDurationValue{}, &model.TimeOrDurationValue{}
-	mint.Set("0000-01-01T00:00:00Z")
-	maxt.Set("9999-12-31T23:59:59Z")
+	start := time.Now()
+	defer func() {
+		u.syncTimes.Observe(time.Since(start).Seconds())
+	}()
```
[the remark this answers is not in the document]
pracucci: Good point.

pkg/querier/block_store.go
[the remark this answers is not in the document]
pstibrany: Of course, I didn't want you to change. I just don't see "resolve button" myself :(

pkg/querier/block_store.go
```diff
@@ -88,78 +176,56 @@ func (u *UserStore) SyncStores(ctx context.Context) error {
...
 		}
 
-		wg.Add(1)
-		go func(userID string, s *store.BucketStore) {
-			defer wg.Done()
-			if err := f(ctx, s); err != nil {
-				level.Warn(u.logger).Log("msg", "user sync failed", "user", userID)
-			}
-		}(user, bs)
+		jobs <- struct {
```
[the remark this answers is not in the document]
pracucci: Right.

pracucci:
@pstibrany Thanks for your review. I've addressed the comments.

PR #1990, opened 2020-01-16
TSDB: fix user metrics when using blocks storage

pstibrany:
When using TSDB, we get per-user created and removed series from TSDB metrics.

This affects `cortex_ingester_memory_series_created_total` and `cortex_ingester_memory_series_removed_total`.
    
This PR also renames `shipperMetrics` to `tsdbMetrics`, and registers the same registry to TSDB and TSDB shipper component.

Registration of prometheus registry now happens only after TSDB initialization is finished, in order to fix a panic where metrics were collected while TSDB was still initialized.

This PR builds on PR #1983, so "Files changed" also shows changes from there. I'll rebase once (if) #1983 is merged. Update: now rebased.

PR #1996, opened 2020-01-17
TSDB: add bucket store metrics in querier

pstibrany:
This PR adds export of aggegate metrics from TSDB's `store.BucketStore` and `storecache.InMemoryIndexCache`, used by Querier component.

This PR uses same approach as #1990 and #1983 before that. It builds on #1990, and moves code related to aggegating and sending metrics into pkg/util/metrics_helper.go file.

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ instructions below to upgrade your Postgres.
 * [ENHANCEMENT] Experimental TSDB: Open existing TSDB on startup to prevent ingester from becoming ready before it can accept writes. #1917
   * `--experimental.tsdb.max-tsdb-opening-concurrency-on-startup`
 * [ENHANCEMENT] Experimental TSDB: Added `cortex_ingester_shipper_dir_syncs_total`, `cortex_ingester_shipper_dir_sync_failures_total`, `cortex_ingester_shipper_uploads_total` and `cortex_ingester_shipper_upload_failures_total` metrics from TSDB shipper component. #1983
+* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos bucket store (too many to list). #1996 
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
 * [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
 * [BUGFIX] Reduce memory usage when ingester Push() errors. #1922
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,8 @@ instructions below to upgrade your Postgres.
 * [ENHANCEMENT] Experimental TSDB: Open existing TSDB on startup to prevent ingester from becoming ready before it can accept writes. #1917
   * `--experimental.tsdb.max-tsdb-opening-concurrency-on-startup`
 * [ENHANCEMENT] Experimental TSDB: Added `cortex_ingester_shipper_dir_syncs_total`, `cortex_ingester_shipper_dir_sync_failures_total`, `cortex_ingester_shipper_uploads_total` and `cortex_ingester_shipper_upload_failures_total` metrics from TSDB shipper component. #1983
-* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos bucket store (too many to list). #1996 
+* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos bucket store (many metrics to list, but all have `cortex_bucket_store_` prefix). #1996 
+* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos in memory index cache (many metrics to list, but all have `cortex_store_index_cache_` prefix). #1996 
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
 * [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
 * [BUGFIX] Reduce memory usage when ingester Push() errors. #1922
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ instructions below to upgrade your Postgres.
   * `--experimental.tsdb.max-tsdb-opening-concurrency-on-startup`
 * [ENHANCEMENT] Experimental TSDB: Added `cortex_ingester_shipper_dir_syncs_total`, `cortex_ingester_shipper_dir_sync_failures_total`, `cortex_ingester_shipper_uploads_total` and `cortex_ingester_shipper_upload_failures_total` metrics from TSDB shipper component. #1983
 * [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos bucket store (many metrics to list, but all have `cortex_querier_bucket_store_` prefix). #1996 
-* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos in memory index cache (many metrics to list, but all have `cortex_store_index_cache_` prefix). #1996 
+* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos in memory index cache (many metrics to list, but all have `cortex_querier_blocks_index_cache_` prefix). #1996 
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
 * [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
 * [BUGFIX] Reduce memory usage when ingester Push() errors. #1922
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,7 @@ instructions below to upgrade your Postgres.
 * [ENHANCEMENT] Experimental TSDB: Open existing TSDB on startup to prevent ingester from becoming ready before it can accept writes. #1917
   * `--experimental.tsdb.max-tsdb-opening-concurrency-on-startup`
 * [ENHANCEMENT] Experimental TSDB: Added `cortex_ingester_shipper_dir_syncs_total`, `cortex_ingester_shipper_dir_sync_failures_total`, `cortex_ingester_shipper_uploads_total` and `cortex_ingester_shipper_upload_failures_total` metrics from TSDB shipper component. #1983
-* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos bucket store (many metrics to list, but all have `cortex_querier_bucket_store_` prefix). #1996 
-* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos in memory index cache (many metrics to list, but all have `cortex_querier_blocks_index_cache_` prefix). #1996 
+* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos bucket store and in memory index cache (many metrics to list, but all have `cortex_querier_bucket_store_` or `cortex_querier_blocks_index_cache_` prefix). #1996 
 * [BUGFIX] Fixed unnecessary CAS operations done by the HA tracker when the jitter is enabled. #1861
 * [BUGFIX] Fixed #1904 ingesters getting stuck in a LEAVING state after coming up from an ungraceful exit. #1921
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ instructions below to upgrade your Postgres.
 * [ENHANCEMENT] Experimental TSDB: Open existing TSDB on startup to prevent ingester from becoming ready before it can accept writes. #1917
   * `--experimental.tsdb.max-tsdb-opening-concurrency-on-startup`
 * [ENHANCEMENT] Experimental TSDB: Added `cortex_ingester_shipper_dir_syncs_total`, `cortex_ingester_shipper_dir_sync_failures_total`, `cortex_ingester_shipper_uploads_total` and `cortex_ingester_shipper_upload_failures_total` metrics from TSDB shipper component. #1983
+* [ENHANCEMENT] Experimental TSDB: Querier now exports aggregate metrics from Thanos bucket store (many metrics to list, but all have `cortex_bucket_store_` prefix). #1996
```
pracucci: Duplicated changelog entry. Moreover, if you will rename the metrics prefix as suggested, please remind to update it here too.

pkg/querier/block_store_metrics.go
```diff
@@ -0,0 +1,220 @@
...
+	// thanos_store_index_cache_max_size_bytes
+	// thanos_store_index_cache_max_item_size_bytes
+}
+
+func newTSDBBucketStoreMetrics() *tsdbBucketStoreMetrics {
+	return &tsdbBucketStoreMetrics{
+		regs: map[string]*prometheus.Registry{},
+
+		blockLoads: prometheus.NewDesc(
+			"cortex_bucket_store_block_loads_total",
```
pracucci: Given the bucket store is an internal component for Cortex, I would be more keep to prefix the metrics with `cortex_querier_bucket_store` instead of `cortex_bucket_store`. What's your take?

pkg/querier/block_store_metrics.go
```diff
@@ -0,0 +1,220 @@
...
+			nil, nil),
+		seriesMergeDuration: prometheus.NewDesc(
+			"cortex_bucket_store_series_merge_duration_seconds",
+			"TSDB: Time it takes to merge sub-results from all queried blocks into a single result.",
+			nil, nil),
+		resultSeriesCount: prometheus.NewDesc(
+			"cortex_bucket_store_series_result_series",
+			"Number of series observed in the final result of a query.",
+			nil, nil),
+		chunkSizeBytes: prometheus.NewDesc(
```
pracucci: I thinks this metric is not very interesting for us, given the bucket store runs within the querier.

pkg/querier/block_store_metrics.go
```diff
@@ -0,0 +1,220 @@
...
+			"cortex_bucket_store_sent_chunk_size_bytes",
+			"TSDB: Size in bytes of the chunks for the single series, which is adequate to the gRPC message size sent to querier.",
+			nil, nil),
+
+		// Cache
+		cacheItemsEvicted: prometheus.NewDesc(
+			"cortex_store_index_cache_items_evicted_total",
```
pracucci: I would change the prefix from `cortex_store_index_cache` to `cortex_querier_blocks_index_cache` (or `cortex_querier_tsdb_index_cache`). What do you think?

pkg/querier/block.go
```diff
@@ -46,6 +46,7 @@ func NewBlockQuerier(cfg tsdb.Config, logLevel logging.Level, r prometheus.Regis
 		return nil, err
 	}
 	b.us = us
+	r.MustRegister(us.tsdbMetrics)
```
pracucci: A couple of things:
1. May be a good practice to register metrics only if `r != nil` (recently I started doing it, cause sometimes in tests we don't want to pass a registerer)
2. I would group `r.MustRegister(b.syncTimes)` here

pkg/util/metrics_helper.go
```diff
@@ -0,0 +1,311 @@
...
+func BuildMetricFamiliesPerUserFromUserRegistries(regs map[string]*prometheus.Registry) MetricFamiliesPerUser {
+	data := MetricFamiliesPerUser{}
+	for userID, r := range regs {
+		m, err := r.Gather()
+		if err == nil {
+			err = data.AddGatheredDataForUser(userID, m)
+		}
+
+		if err != nil {
+			level.Warn(Logger).Log("msg", "failed to gather metrics from TSDB shipper", "user", userID, "err", err)
```
pracucci: Given now this is a generic utility, we should remove `TSDB shipper` from here and make the log message generic too.

pkg/util/metrics_helper.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	d[userID] = perMetricName
+	return nil
+}
+
+func (d MetricFamiliesPerUser) SendSumOfCounters(out chan<- prometheus.Metric, desc *prometheus.Desc, counter string) {
+	result := float64(0)
+	for _, perMetric := range d {
+		result += sum(perMetric[counter], counterValue)
```
pracucci: If for any reason `perMetric[counter]` is `nil`, the `sum()` wil panic. Would be also great to cover this edge case with a test.

_Same comment applies to other `Send...()` functions._

pkg/util/metrics_helper.go
```diff
@@ -0,0 +1,311 @@
...
+func (d MetricFamiliesPerUser) SendSumOfCountersWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, counter string, labelNames ...string) {
+	type counterResult struct {
+		value       float64
+		labelValues []string
+	}
+
+	result := map[string]counterResult{}
+
+	for _, userMetrics := range d { // for each user
+		metricsPerLabelValue := getMetricsWithLabelNames(userMetrics[counter], labelNames)
```
pracucci: If for any reason `userMetrics[counter]` is `nil`, `getMetricsWithLabelNames()` will panic

pkg/util/metrics_helper.go
```diff
@@ -0,0 +1,311 @@
...
+func (d MetricFamiliesPerUser) SendSumOfGauges(out chan<- prometheus.Metric, desc *prometheus.Desc, gauge string) {
+	result := float64(0)
+	for _, perMetric := range d {
+		result += sum(perMetric[gauge], gaugeValue)
+	}
+	out <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, result)
+}
+
+func (d MetricFamiliesPerUser) SendSumOfGaugesWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, counter string, labelNames ...string) {
+	type gaugeResult struct {
```
pracucci: This content is equal to `SendSumOfCountersWithLabels()`, except for the different `type` name (`counterResult` vs `gaugeResult`). We may generalize it into a `sumOfCollectorsWithLabels()` and then call it from `SendSumOf(Counters|Gauge)WithLabels()` to get the sum.

pkg/util/metrics_helper.go
```diff
@@ -0,0 +1,311 @@
...
+
+func (d MetricFamiliesPerUser) SendSumOfSummaries(out chan<- prometheus.Metric, desc *prometheus.Desc, summaryName string) {
+	var (
+		sampleCount uint64
+		sampleSum   float64
+		quantiles   map[float64]float64
+	)
+
+	for _, userMetrics := range d { // for each user
+		for _, m := range userMetrics[summaryName].GetMetric() {
```
pracucci: Here `userMetrics[summaryName]` could panic as well if `nil`

pkg/util/metrics_helper.go
```diff
@@ -0,0 +1,311 @@
...
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+// MetricFamiliesPerUser is a collection of metrics gathered via calling Gatherer.Gather() method on different
+// gatherers, one per user.
+// First key = userID, second key = metric name.
+// Value = slice of gathered values with the same metric name.
+type MetricFamiliesPerUser map[string]map[string]*dto.MetricFamily
```
pracucci: Would be great to have basic unit tests on `MetricFamiliesPerUser`

pkg/querier/block_store_metrics.go
pstibrany: Fine with me. I'll change it.

pkg/querier/block_store_metrics.go
pstibrany: OK, will remove.

pkg/querier/block_store_metrics.go
pstibrany: Sounds good.

pkg/util/metrics_helper.go
pstibrany: Oops, you're right. :-)

pkg/util/metrics_helper.go
pstibrany: That's not true. range of nil slice doesn't panic.

pracucci: But `range mf.Metric {` will panic if `mf` is `nil`

pkg/util/metrics_helper.go
pstibrany: This is not true. `userMetrics[counter]` is of type `*dto.MetricFamily`, and it's `GetMetric()` is defined for `nil` as well:

```
func (m *MetricFamily) GetMetric() []*Metric {
	if m != nil {
		return m.Metric
	}
	return nil
}
```

Same is true for all protobuf-based types used here.

pkg/util/metrics_helper.go
pstibrany: Good idea, I'll do this change.

pkg/util/metrics_helper.go
pstibrany: Explained earlier, this isn't true.

pkg/util/metrics_helper.go
pstibrany: I think bucket_store_metrics_test.go already covers all code here. Do you think extra testing is needed here?

pkg/util/metrics_helper.go
pstibrany: I don't use `range mf.Metric`, but `range mf.GetMetric()`, which will not panic, even for `mf` being `nil`.

pstibrany: Ah, sorry, you're right... I use `mf.Metric` in `sum`... needs to be changed.

pstibrany: Fixed, and added test for sum function.

pstibrany:
Thank you Marco for your review. I believe (I will double check) that I've addressed all your feedback now.

pkg/querier/block.go
```diff
@@ -39,14 +39,17 @@ func NewBlockQuerier(cfg tsdb.Config, logLevel logging.Level, r prometheus.Regis
...
-	r.MustRegister(b.syncTimes)
-
 	us, err := NewUserStore(cfg, logLevel, util.Logger)
 	if err != nil {
 		return nil, err
 	}
 	b.us = us
 
+	if r != nil {
+		r.MustRegister(b.syncTimes)
```
pracucci: You can simplify it as `r.MustRegister(b.syncTimes, us.tsdbMetrics)`.

pkg/util/metrics_helper_test.go
```diff
@@ -0,0 +1,83 @@
...
+	m1 := &dto.Metric{Label: makeLabels("a", "5"), Counter: &dto.Counter{Value: proto.Float64(1)}}
+	m2 := &dto.Metric{Label: makeLabels("a", "10", "b", "20"), Counter: &dto.Counter{Value: proto.Float64(1.5)}}
+	m3 := &dto.Metric{Label: makeLabels("a", "10", "b", "20", "c", "1"), Counter: &dto.Counter{Value: proto.Float64(2)}}
+	m4 := &dto.Metric{Label: makeLabels("a", "10", "b", "20", "c", "2"), Counter: &dto.Counter{Value: proto.Float64(3)}}
+	m5 := &dto.Metric{Label: makeLabels("a", "11", "b", "21"), Counter: &dto.Counter{Value: proto.Float64(4)}}
+	m6 := &dto.Metric{Label: makeLabels("ignored", "123", "a", "12", "b", "22", "c", "30"), Counter: &dto.Counter{Value: proto.Float64(4)}}
+
+	out := getMetricsWithLabelNames(&dto.MetricFamily{Metric: []*dto.Metric{m1, m2, m3, m4, m5, m6}}, labels)
+
+	// m1 is not returned at all, as it doesn't habe both required labels.
```
pracucci: `habe` > `have`

PR #1997, opened 2020-01-17
Avoid calling flag.Parse() twice.

pstibrany:
flag.Parse() was called twice previously, first to get `-config.file` value to parse the config file, and second to parse remaining command line parameters – overwriting values from config file.

Unfortunately that confuses command line parameters that accept multiple values (eg. `-memberlist.join` or `-experimental.tsdb.block-ranges-period` in #1942). In this PR we get `-config.file` via a separate FlagSet (which doesn't report any error or usage) to avoid calling `flag.Parse()` on default FlagSet twice.

This should not be visible to end user in any way.

### DIFF
--- a/cmd/cortex/main.go
+++ b/cmd/cortex/main.go
@@ -9,6 +9,7 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -18,22 +19,25 @@ import (
 
 	"github.com/cortexproject/cortex/pkg/cortex"
 	"github.com/cortexproject/cortex/pkg/util"
-	"github.com/cortexproject/cortex/pkg/util/flagext"
 )
 
 func init() {
 	prometheus.MustRegister(version.NewCollector("cortex"))
 }
 
+const configFileOption = "config.file"
+
 func main() {
 	var (
-		cfg                  cortex.Config
-		configFile           = ""
 		eventSampleRate      int
 		ballastBytes         int
 		mutexProfileFraction int
 	)
-	flag.StringVar(&configFile, "config.file", "", "Configuration file to load.")
+
+	cfg := parseAndLoadConfigFile()
+
+	// Ignore -config.file here, since it was already parsed, but it's still present on command line.
+	flagext.IgnoredFlag(flag.CommandLine, configFileOption, "Configuration file to load.")
 	flag.IntVar(&eventSampleRate, "event.sample-rate", 0, "How often to sample observability events (0 = never).")
 	flag.IntVar(&ballastBytes, "mem-ballast-size-bytes", 0, "Size of memory ballast to allocate.")
 	flag.IntVar(&mutexProfileFraction, "debug.mutex-profile-fraction", 0, "Fraction at which mutex profile vents will be reported, 0 to disable")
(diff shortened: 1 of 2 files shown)


pstibrany:
> This PR introduces a regression.... [del]
> I've the feeling it has to do with defaults. The `flagext.RegisterFlags()` does also set default values on the config. With this PR's change, we're calling `RegisterFlags()` **after** loading the config file, so we're overriding what we loaded from the config with the default values from flags. `flagext.RegisterFlags()` should be called before parsing the config file.
> 
> Would be great if we could have a unit test on this (maybe moving all config file + CLI flags logic in a single function to unit test), but if it's too complicated just skip it and do some manual tests please 😉

Thanks for your review and my apologies! I was happy that flags work correctly that I forgot to check actual usage of config file 🤦‍♂ You're right about `RegisterFlags` -- they actually set default values from the flags to config file. I missed that completely! Now it's fixed, and I've also added unit tests for main and various combinations of config file and flags, including default values set by flags (by testing `-target=all`).

pstibrany:
Tests work by 1) introducing "test mode" to main() function, which only does config/flags parsing, dumps resulting YAML and then stop, and 2) capturing stdout/stderr and checking for expected messages.

cmd/cortex/main.go
```diff
@@ -25,34 +25,52 @@ func init() {
...
 
-	// Parse a second time, as command line flags should take precedent over the config file.
+	// Ignore -config.file here, since it was already parsed, but it's still present on command line.
+	flagext.IgnoredFlag(flag.CommandLine, configFileOption, "Configuration file to load.")
+	flag.IntVar(&eventSampleRate, "event.sample-rate", 0, "How often to sample observability events (0 = never).")
+	flag.IntVar(&ballastBytes, "mem-ballast-size-bytes", 0, "Size of memory ballast to allocate.")
+	flag.IntVar(&mutexProfileFraction, "debug.mutex-profile-fraction", 0, "Fraction at which mutex profile vents will be reported, 0 to disable")
+
+	if mutexProfileFraction > 0 {
+		runtime.SetMutexProfileFraction(mutexProfileFraction)
```
pracucci: This should be called after `flag.Parse()`, otherwise `mutexProfileFraction` is always `0`

pstibrany: Good catch, thank you!

cmd/cortex/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"io"
+	"io/ioutil"
+	"os"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestName(t *testing.T) {
```
pracucci: Did you mean `TestMain`?

pstibrany: That was default name autocompleted by Golang. I forgot to change it :) Fixed now.

cmd/cortex/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	wg                         sync.WaitGroup
+	stdoutReader, stdoutWriter *os.File
+	stderrReader, stderrWriter *os.File
+}
+
+func captureOutput(t *testing.T) *capturedOutput {
+	stdoutR, stdoutW, err := os.Pipe()
+	require.NoError(t, err)
+	os.Stdout = stdoutW
```
pracucci: Shouldn't we guarantee `os.Stdout` and `os.Stderr` are both reset to the default once the test is over? I'm thinking when tests from multiple packages run.

pstibrany: We do, in `testSingle` function in `defer` statement.

pracucci: Oh right. I haven't noticed it.

cmd/cortex/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	} {
+		t.Run(name, func(t *testing.T) {
+			testSingle(t, tc.arguments, tc.yaml, []byte(tc.stdoutMessage), []byte(tc.stderrMessage))
+		})
+	}
+}
+
+func testSingle(t *testing.T, arguments []string, yaml string, stdoutMessage, stderrMessage []byte) {
+	oldArgs, oldStdout, oldStderr := os.Args, os.Stdout, os.Stderr
+	defer func() {
```
pracucci: We may also reset `testMode = false`

pstibrany: OK, done.

Issue #1722, opened 2019-10-08
Unable to run alertmanager in single binary


jtlisi:
Currently it is not possible to enable the alertmanager module when cortex is run as a single binary. This is caused by conflicts under the `/api/prom` route.

One option I would like to suggest to rectify this change is updating the routing scheme for the alertmanager to fall under `/alertmanager`. I think this format would make more sense overall. I also think the prometheus api should exist under `/prometheus`. I think this naming scheme would make more sense for two reasons:

1. `/api/prom/api/*` routes seem clunky to me with the repetition of `api`
2. `/<service>/api/*` provides a sensible api scheme that reflects the underlying endpoints. This would also allow for endpoints with the same routes to exist in the same binary. For instance both the alertmanager and prometheus api have a `/api/v1/alerts` endpoint.

pracucci:
> One option I would like to suggest to rectify this change is updating the routing scheme for the alertmanager to fall under /alertmanager. I think this format would make more sense overall. I also think the prometheus api should exist under /prometheus.

I would agree with you if this wouldn't be a breaking change. Given this is a breaking change which doesn't just impact Cortex but also its clients (ie. Grafana config), I've some doubts about it.

> The alternative would be matching /api/prom with a /api/alert route, or something along those lines. I think either approach would work.

Following up on this alternative solution, an option may be just split the HTTP prefix config into:

- `-http.prometheus-prefix` (defaults to `/api/prom`)
- `-http.alertmanager-prefix` (defaults to be decided)

The breaking change would be in the the config option name, but not its default value, and would allow users to customize it and have the alertmanager running along with the other components in a single binary setup.

_I understand it's not ideal, but the breaking change should be softer._

bboreham:
It's not a breaking change if the default under `-target=alertmanager` retains the existing route in parallel with the new way.

pracucci:
> if the default under -target=alertmanager retains the existing route in parallel with the new way.

If we retain the existing rule in parallel with the new one, how do we solve the original problem:

> Currently it is not possible to enable the alertmanager module when cortex is run as a single binary. This is caused by conflicts under the /api/prom route.

bboreham:
By having different defaults in single-binary mode.

pracucci:
> By having different defaults in single-binary mode.

Not sure about this. Looks a bit confusing have different defaults whether you run Cortex in single binary mode or microservices.

bboreham:
This seems to be the season for breaking changes: party on.

[...]

PR #1999, opened 2020-01-17
20190117 prom rules endpoints

### LINKED ISSUE #1720: Ruler does not currently support `/api/v1/rules` or `/api/v1/alerts`
Currently there isn't support in the cortex ruler match the prometheus rules or alerts endpoing. Ideally these endpoints can be supported in the future.

jtlisi:
**What this PR does**:

This PR implements the Prometheus API alerts and rules endpoints in ruler. It adds a GRPC service to the ruler to allow for an implementation that works when rules are sharded across multiple rulers.

Fixes #1720 

**Checklist**
- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md`


pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+		level.Debug(util.Logger).Log("msg", "ruler: registering route", "name", route.name, "method", route.method, "path", route.path)
+		router.Handle(route.path, route.handler).Methods(route.method).Name(route.name)
+	}
+}
+
+// In order to reimplent the prometheus rules API, a large amount of code was copied over
+// This is required because the prometheus api implementation does not pass a context to
+// the rule retrieval function.
+// https://github.com/prometheus/prometheus/blob/2aacd807b3ec6ddd90ae55f3a42f4cffed561ea9/web/api/v1/api.go#L108
+// https://github.com/prometheus/prometheus/pull/4999
```
pracucci: As a non-blocking comment, would be great to double check with @tomwilkie if there's anything we can do to avoid copy-pasting so much code here and reconsider the PR https://github.com/prometheus/prometheus/pull/4999

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+		level.Error(logger).Log("msg", "error extracting org id from context", "err", err)
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	rgs, err := r.GetRules(ctx, userID)
+
+	if err != nil {
+		b, err := json.Marshal(&response{
```
pracucci: This block of code is duplicated between `rules()` and `alerts()`. It could make sense to have a `respondError()` utility function.

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+		{"get_rules", "GET", "/api/prom/api/v1/rules", r.rules},
+		{"get_alerts", "GET", "/api/prom/api/v1/rules", r.alerts},
+	} {
+		level.Debug(util.Logger).Log("msg", "ruler: registering route", "name", route.name, "method", route.method, "path", route.path)
+		router.Handle(route.path, route.handler).Methods(route.method).Name(route.name)
+	}
+}
+
+// In order to reimplent the prometheus rules API, a large amount of code was copied over
+// This is required because the prometheus api implementation does not pass a context to
```
pracucci: I'm a bit dubious. I was expecting the following code to be a close replica of https://github.com/cortexproject/cortex/blob/master/vendor/github.com/prometheus/prometheus/web/api/v1/api.go but it's actually quite different.

Is it because it has been copied long time ago? If so, would it make sense to redo the copy-paste? Just asking to better undestand: don't want you to do extra work.

pkg/ruler/ruler.go
```diff
@@ -444,3 +450,131 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
...
+			return nil, err
+		}
+
+		ctx, err = user.InjectIntoGRPCRequest(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("unable to inject user ID into grpc request, %v", err)
+		}
+
+		for _, rlr := range rulers.Ingesters {
+			// Don't retrieve rules using GRPC from the same ingester
```
pracucci: Have you actually considered retrieving the rules from the same ingester from gRPC too? I could simplify a bit the code because the entire function could be:

```
if r.cfg.EnableSharding {
   // Load from all ingesters via gRPC
} else {
   return r.getRules(userID)
}
```

Moreover, from a readability perspective, may be more clear if we rename `getRules()` to `getLocalRules()` and this block of code is moved to another function like `getShardedRules()` so that this function code will just be:

```
if r.cfg.EnableSharding {
   return r.getShardedRules(userID)
}

return r.getLocalRules(userID)
```

pkg/ruler/ruler.go
```diff
@@ -444,3 +450,131 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
 	}
 	return promRules.NewManager(opts), nil
 }
+
+// GetRules retrieves the running rules from this ruler and all running rulers in the ring if
+// sharding is enabled
+func (r *Ruler) GetRules(ctx context.Context, userID string) ([]*rules.RuleGroupDesc, error) {
```
pracucci: Can we add a test to at least cover the happy path for this function?

pkg/ruler/ruler.go
jtlisi: I had considered using GPRC for all of the rules but thought it was strange for the local use case. That snippet will work quite well though

pkg/ruler/ruler.go
```diff
@@ -63,6 +66,8 @@ type Config struct {
 	SearchPendingFor time.Duration
 	LifecyclerConfig ring.LifecyclerConfig
 	FlushCheckPeriod time.Duration
+
+	EnableAPI bool
```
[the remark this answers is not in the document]
jtlisi: This is called in modules.go. Sorry about that, the change got lost during a rebase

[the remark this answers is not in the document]

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/go-kit/kit/log/level"
+	"github.com/gorilla/mux"
+	"github.com/weaveworks/common/user"
+)
+
+// RegisterRoutes registers the configs API HTTP routes with the provided Router.
+func (r *Ruler) RegisterRoutes(router *mux.Router) {
```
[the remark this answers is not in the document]
jtlisi: same as above ^

pkg/ruler/api.go
jtlisi: The main things that were copied over was the structs. Most of the logic for handling the requests is simplified versions of what was in Prometheus.

pkg/ruler/ruler.go
```diff
@@ -63,6 +66,11 @@ type Config struct {
 	SearchPendingFor time.Duration
 	LifecyclerConfig ring.LifecyclerConfig
 	FlushCheckPeriod time.Duration
+
+	EnableAPI bool `yaml:"enable_api"`
+
+	// used to avoid duplicate registration in tests
+	registry prometheus.Registerer `yaml:"-"`
```
pracucci: A better approach would be passing the `prometheus.Registerer` in input to `newManager()` (and propagate it from parents if required). This way tests can pass a custom registerer. It's a common design, we already use in other parts of Cortex.

pkg/ruler/ruler.go

pkg/ruler/api.go
jtlisi: I added a generalized `respondError` function

pkg/ruler/ruler.go
```diff
@@ -444,3 +453,130 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
...
+	if exists {
+		groups = mngr.RuleGroups()
+	}
+	r.userManagerMtx.Unlock()
+
+	for _, group := range groups {
+		interval := group.Interval()
+		groupDesc := &rules.RuleGroupDesc{
+			Name:      group.Name(),
+			Namespace: group.File(),
```
pracucci: Two things:
1. Why is it called `Namespace` if it then maps the `File` in the API response (see `api.go`)?
2. Is it really a good thing exposing the path of the file inside Cortex? Looks we may disclose a sensitive information in a Cortex as a service scenario

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+						State:       a.GetState(),
+						ActiveAt:    a.ActiveAt,
+						Value:       fmt.Sprintf("%f", a.Value),
+					})
+				}
+				grp.Rules[i] = alertingRule{
+					State:       rl.GetState(),
+					Name:        rl.GetAlert(),
+					Query:       rl.GetExpr(),
+					Duration:    rl.For.Seconds(),
```
pracucci: When the duration is not set in the alert, this is exported as `0`. Is it correct? May you check Prometheus behaviour please?

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+		return
+	}
+
+	w.WriteHeader(http.StatusInternalServerError)
+	if n, err := w.Write(b); err != nil {
+		level.Error(logger).Log("msg", "error writing response", "bytesWritten", n, "err", err)
+	}
+}
+
+func (r *Ruler) rules(w http.ResponseWriter, req *http.Request) {
```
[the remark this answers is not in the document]
jtlisi: `status` is included in the Prometheus API call. You can try it on a local prom server.

pkg/ruler/api.go
jtlisi: This would go a long way in running cortex as a single binary.

pkg/ruler/api.go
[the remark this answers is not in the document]
pracucci: Oh sorry for not being clear. I didn't copy the entire response from our API.

Sure `status` is fine, but inside `groups` Prometheus API documentation mention this structure:

```
"rules":  []
```

While it looks we don't have `"rules"`. Try to take a deeper look, please.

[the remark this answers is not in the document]
pracucci: I was referring the inner `groups` structure, but I was actually wrong. I diffed it wrongly. The structure is correct.

pkg/ruler/api.go
pracucci: I double checked and Prometheus exposes it as `0`, so all good here.

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+func (r *Ruler) RegisterRoutes(router *mux.Router) {
+	if r.store == nil {
+		level.Info(util.Logger).Log("msg", "ruler configured with store that does not support api")
+		return
+	}
+	for _, route := range []struct {
+		name, method, path string
+		handler            http.HandlerFunc
+	}{
+		{"get_rules", "GET", "/api/prom/api/v1/rules", r.rules},
```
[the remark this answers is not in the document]
jtlisi: This doesn't seem to be a universal pattern throughout cortex. For instance:
https://github.com/cortexproject/cortex/blob/dd77369a2341b2ea263ecf8cd01fae2d347cbd16/pkg/cortex/modules.go#L258
https://github.com/cortexproject/cortex/blob/dd77369a2341b2ea263ecf8cd01fae2d347cbd16/pkg/cortex/modules.go#L261

I am in favor of standardizing around `cortex.Config.HTTPPrefix`. This relates to the concerns I brought up in #1722

pkg/ruler/api.go
jtlisi: Yes, in the case when `For` is not set the alert will fire immediately. From the [Prometheus Docs](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/#defining-alerting-rules):

> The optional for clause causes Prometheus to wait for a certain duration between first encountering a new expression output vector element and counting an alert as firing for this element. In this case, Prometheus will check that the alert continues to be active during each evaluation for 10 minutes before firing the alert. Elements that are active, but not firing yet, are in the pending state.

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/gorilla/mux"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/weaveworks/common/user"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+// RegisterRoutes registers the configs API HTTP routes with the provided Router.
+func (r *Ruler) RegisterRoutes(router *mux.Router) {
```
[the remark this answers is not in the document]
jtlisi: Prometheus does have a stability guarantee for it's API. It's listed here: https://github.com/prometheus/prometheus/blob/master/docs/stability.md

As far as a unit test that compares the two, I think that should be feasible. I'll investigate further.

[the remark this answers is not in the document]
pracucci: Don't get crazy if it's not straightforward. Otherwise could be an easy way to test for compatibility. Up to your judgement.

pkg/ruler/ruler_test.go
```diff
@@ -92,3 +98,26 @@ func TestNotifierSendsUserIDHeader(t *testing.T) {
 
 	wg.Wait()
 }
+
+func TestRuler_Rules(t *testing.T) {
```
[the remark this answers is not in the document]
jtlisi: This is an unintended consequence of forcing a call to `loadRules`, which was never meant to be called twice. I fixed the registry bug by changing the lock order, but a few more bugs popped up after that one was resolved. The null Queryable provided for tests was causing a seg fault and there was a race condition when updating rules that lead to empty rule groups being returned. After a bit of refactoring this should be sorted.

jtlisi:
@pracucci I'm only part way done refactoring this. I'll try to get it ready for another pass by Monday

pkg/ruler/api.go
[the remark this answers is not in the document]
pracucci: Thanks for pointing #1722. I've commented there. Given I do agree this is **not** the right PR to also address the `cortex.Config.HTTPPrefix` standardization, and given you've correctly pointed out that there are more places where it's hardcoded than based on `HTTPPrefix`, I'm going to resolve this comment, so that we can continue this discussion in #1722 before working on it in a separate PR.

pkg/ruler/mock_store.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
```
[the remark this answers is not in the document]
jtlisi: I'm going to change it to store_mock_test.go lol

pkg/ruler/ruler_test.go
```diff
@@ -92,3 +98,26 @@ func TestNotifierSendsUserIDHeader(t *testing.T) {
...
+	defer r.Stop()
+
+	r.loadRules(context.Background())
+
+	ctx := user.InjectOrgID(context.Background(), "user1")
+	rls, err := r.Rules(ctx, &RulesRequest{})
+	require.NoError(t, err)
+	require.Len(t, rls.Groups, 1)
+	for _, rule := range rls.GetGroups() {
+		fmt.Println(rule)
```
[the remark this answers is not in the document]
jtlisi: I added a fully fleshed out function for comparing and asserting the returned rules.

pkg/ruler/ruler_test.go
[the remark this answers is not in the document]
jtlisi: I added a bit more to this test. The only thing that I wasn't able to add was a firing alert. The reason being we don't have an easy way to mock an operable querier.

pkg/ruler/ruler.go
jtlisi: 2. I updated the downstream path to strip the internal cortex file path. The only thing that will be left is the namespace of the users rule group.

The reason Namespace was chosen instead of file is basically to avoid have file be an API object when interacting with Cortex.

pkg/ruler/api.go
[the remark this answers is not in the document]
jtlisi: 👍

jtlisi:
@pracucci This should be ready for another look. The scope of this PR broadened a bit while getting the tests set up. The ruler and mapper are now explicitly passed a logger upon instantiation. Also there was some updates to the fixtures to make sure the ruler loaded it's rules before the tests are run.

pkg/ruler/api.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/go-kit/kit/log/level"
+	"github.com/gorilla/mux"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/weaveworks/common/user"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+// RegisterRoutes registers the configs API HTTP routes with the provided Router.
```
pracucci: `registers the configs API` > `registers the ruler API`

pkg/cortex/modules.go
```diff
@@ -415,11 +415,16 @@ func (t *Cortex) initRuler(cfg *Config) (err error) {
 	cfg.Ruler.Ring.ListenPort = cfg.Server.GRPCListenPort
 	queryable, engine := querier.New(cfg.Querier, t.distributor, t.store)
 
-	t.ruler, err = ruler.NewRuler(cfg.Ruler, engine, queryable, t.distributor)
+	t.ruler, err = ruler.NewRuler(cfg.Ruler, engine, queryable, t.distributor, prometheus.DefaultRegisterer, util.Logger)
 	if err != nil {
 		return
```
pracucci: This should be `return err`

pkg/ruler/api.go
```diff
@@ -0,0 +1,260 @@
...
+func (r *Ruler) RegisterRoutes(router *mux.Router) {
+	for _, route := range []struct {
+		name, method, path string
+		handler            http.HandlerFunc
+	}{
+		{"get_rules", "GET", "/api/prom/api/v1/rules", r.rules},
```
pracucci: In `modules.go` we do `t.server.HTTP.PathPrefix(cfg.HTTPPrefix).Subrouter()` so now it's registered to `/api/prom/api/prom/api/v1/rules`. Here we should remove the `/api/prom` prefix.

pkg/ruler/api.go
```diff
@@ -0,0 +1,260 @@
...
+						State:       a.GetState(),
+						ActiveAt:    a.ActiveAt,
+						Value:       fmt.Sprintf("%f", a.Value),
+					})
+				}
+				grp.Rules[i] = alertingRule{
+					State:       rl.GetState(),
+					Name:        rl.GetAlert(),
+					Query:       rl.GetExpr(),
+					Duration:    rl.For.Seconds(),
```
pracucci: Can't `rl.For` be `nil`?

pkg/ruler/api.go
```diff
@@ -0,0 +1,260 @@
...
+		groups = append(groups, &grp)
+	}
+
+	b, err := json.Marshal(&response{
+		Status: "success",
+		Data:   &RuleDiscovery{RuleGroups: groups},
+	})
+	if err != nil {
+		level.Error(logger).Log("msg", "error marshaling json response", "err", err)
+		respondError(logger, w, "unable to marshal the requested data", errorBadData)
```
pracucci: Why is this a `errorBadData`? If we're able to marshal the response looks more a bug on our side. Moreover, the bad data error is mapped to a `400` response status code, while we always return 5xx, but I think it's correct 5xx here since there's nothing the user can do on it.

_Same comment applies to `alerts()`_

pkg/ruler/api.go
```diff
@@ -0,0 +1,260 @@
...
+		level.Debug(util.Logger).Log("msg", "ruler: registering route", "name", route.name, "method", route.method, "path", route.path)
+		router.Handle(route.path, route.handler).Methods(route.method).Name(route.name)
+	}
+}
+
+type errorType string
+
+const (
+	errorBadData  errorType = "bad_data"
+	errorInternal errorType = "internal"
```
pracucci: If we wanna keep full compatibility with Prometheus, in Prometheus it's:
`ErrServer      ErrorType = "server_error"`

Also, instead of redefining them here, we can directly use them from Prometheus (using its `ErrorType` instead of `errorType`).

pkg/ruler/api.go
```diff
@@ -0,0 +1,260 @@
...
+		level.Error(logger).Log("msg", "error extracting org id from context", "err", err)
+		respondError(logger, w, "no valid org id found", errorInternal)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	rgs, err := r.GetRules(ctx, userID)
+
+	if err != nil {
+		respondError(logger, w, err.Error(), "internal")
```
pracucci: `"internal"` > `errorInternal`

pkg/ruler/ruler.go
```diff
@@ -430,18 +449,147 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
...
+	}
+	r.userManagerMtx.Unlock()
+
+	prefix := filepath.Join(r.cfg.RulePath, userID) + "/"
+
+	for _, group := range groups {
+		interval := group.Interval()
+		groupDesc := &rules.RuleGroupDesc{
+			Name:      group.Name(),
+			Namespace: strings.TrimPrefix(group.File(), prefix),
```
pracucci: If we remove the path, what's left isn't just the filename? If so, I would use the safer function:

```
filepath.Base(group.File())
```

pkg/ruler/ruler.go
```diff
@@ -430,18 +449,147 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
...
+				return nil, errors.Errorf("failed to assert type of rule '%v'", rule.Name())
+			}
+			groupDesc.Rules = append(groupDesc.Rules, ruleDesc)
+		}
+		groupDescs = append(groupDescs, groupDesc)
+	}
+	return groupDescs, nil
+}
+
+func (r *Ruler) getShardedRules(ctx context.Context, userID string) ([]*rules.RuleGroupDesc, error) {
```
pracucci: I haven't manually tested it. Have you tested it with the latest version of the code? If not, may you do it, please?

pkg/ruler/ruler.go
```diff
@@ -430,18 +449,147 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
...
+
+	return r.getLocalRules(userID)
+}
+
+func (r *Ruler) getLocalRules(userID string) ([]*rules.RuleGroupDesc, error) {
+	groupDescs := []*rules.RuleGroupDesc{}
+
+	var groups []*promRules.Group
+	r.userManagerMtx.Lock()
+	mngr, exists := r.userManagers[userID]
```
pracucci: We could compact it into the following, which also reduce the scope of variables:

```
if mngr, exists := r.userManagers[userID]; exists {
    groups = mngr.RuleGroups()
}
```

pkg/ruler/ruler.go
```diff
@@ -430,18 +449,147 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
...
+func (r *Ruler) GetRules(ctx context.Context, userID string) ([]*rules.RuleGroupDesc, error) {
+	if r.cfg.EnableSharding {
+		return r.getShardedRules(ctx, userID)
+	}
+
+	return r.getLocalRules(userID)
+}
+
+func (r *Ruler) getLocalRules(userID string) ([]*rules.RuleGroupDesc, error) {
+	groupDescs := []*rules.RuleGroupDesc{}
```
pracucci: If you move this line after the locked block, you can do `groupDescs := make([]*rules.RuleGroupDesc{}, 0, len(groups))`

pkg/ruler/ruler.go
```diff
@@ -370,36 +384,7 @@ func (r *Ruler) loadRules(ctx context.Context) {
...
-			err = manager.Update(r.cfg.EvaluationInterval, files, nil)
-			if err != nil {
-				level.Error(util.Logger).Log("msg", "unable to create rule manager", "user", user, "err", err)
-				continue
-			}
-		}
+		r.syncManager(ctx, user, filteredGroups)
 	}
 
 	// Check for deleted users and remove them
```
pracucci: Note to myself: this block hasn't changed, except the logger reference.

pkg/ruler/ruler.go
```diff
@@ -409,12 +394,46 @@ func (r *Ruler) loadRules(ctx context.Context) {
...
+
+	// Map the files to disk and return the file names to be passed to the users manager
+	update, files, err := r.mapper.MapRules(user, groups.Formatted())
+	if err != nil {
+		level.Error(r.logger).Log("msg", "unable to map rule files", "user", user, "err", err)
+		return
+	}
+
+	if update {
+		level.Info(r.logger).Log("msg", "updating rules", "user", "user")
```
pracucci: Info or debug level? Up to you, just wanted to raise it, cause may be a bit verbose for an Info level.

pkg/ruler/ruler.go
```diff
@@ -409,12 +394,46 @@ func (r *Ruler) loadRules(ctx context.Context) {
...
+			level.Info(r.logger).Log("msg", "deleting rule manager", "user", user)
 		}
 	}
 
 }
 
+// syncManager maps the rule files to disk, detects any changes and will create/update the
+// the users Prometheus Rules Manager
+func (r *Ruler) syncManager(ctx native_ctx.Context, user string, groups store.RuleGroupList) {
+	r.userManagerMtx.Lock()
```
pracucci: Before we were locking only around `r.userManagers` access, while now we lock around the entire function. Could you explain why? If you confirm it's the indented behaviour, adding a comment before this `Lock()` explaining the reason may be useful for future readers (it's unclear to me).

jtlisi: As soon as `update, files, err := r.mapper.MapRules(user, groups.Formatted())` is called we need to lock until we call `manager.Update` for each user. Otherwise it there is a race condition that can make itself know in tests.

Assume there are two calls to `syncManager`:
  - `CallOne` maps files and returns true to update user managers
  - `CallTwo` maps files and returns false to update user managers
  - `CallTwo` returns and wherever it is called assumes that the running managers for each user are up to date, but they aren't because `CallOne` hasn't completed yet.

It is an issue that will only pop up in tests, since we need to force a sync instead of letting it occur in a background task. The other option would be adding a sleep to tests. However, I generally like to avoid doing that if need be. I also think it is very likely syncManagers will be called concurrently at some point. Specifically, as part of the Ruler lifecycle, we would want a new ruler joining/leaving the ring to trigger a sync in all of the other rulers.

I'll add a comment to make what is going on clear. I just wanted to write a more detailed explanation here.

pkg/ruler/ruler.go
jtlisi: I think an update would be a relatively rare event? Since we have a metric I'm ok with lowering it to debug though.

pkg/ruler/ruler.go
jtlisi: Nice

pkg/ruler/api.go
jtlisi: Good catch, makes it obvious I haven't tested this yet

pkg/cortex/modules.go
```diff
@@ -456,7 +461,7 @@ func (t *Cortex) initAlertmanager(cfg *Config) (err error) {
 
 	// TODO this clashed with the queirer and the distributor, so we cannot
 	// run them in the same process.
-	t.server.HTTP.PathPrefix("/api/prom").Handler(middleware.AuthenticateUser.Wrap(t.alertmanager))
+	t.server.HTTP.PathPrefix(cfg.HTTPPrefix).Handler(middleware.AuthenticateUser.Wrap(t.alertmanager))
```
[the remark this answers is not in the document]
jtlisi: No harm in adding an entry. I'll make a small PR separately that changes it for all of the modules that don't currently use it.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893 
   * `--store.min-chunk-age` has been removed
   * `--querier.query-store-after` has been added in it's place.
+* [FEATURE] Added flag `experimental.ruler.enable-api` to enable the ruler api which implements the Prometheus API `/api/v1/rules` and `/api/v1/alerts` endpoints. #1999
```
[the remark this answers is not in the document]
jtlisi: I'll make it clear it is prefixed with the configured `-http.prefix`.

docs/apis.md
```diff
@@ -17,6 +17,9 @@ APIs.  The encoding is Protobuf over http.
 
 Read is on `/api/prom/read` and write is on `/api/prom/push`.
 
+## Alerts & Rules API
+
+Cortex supports the Prometheus' [alerts](https://prometheus.io/docs/prometheus/latest/querying/api/#alerts) and [rules](https://prometheus.io/docs/prometheus/latest/querying/api/#rules) api endpoints. This is supported in the Ruler service and can be enabled using the `experimental.ruler.enable-api` flag.
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/api.go
jtlisi: 👍

pkg/cortex/modules.go
jtlisi: Good catch

pkg/ruler/api.go
jtlisi: Technically yes. However, it will never occur in practice. The more I look at this, I think it would be appropriate to use a nonnullable approach to Times in the rules/alerts protos. The zero values are considered unset by prometheus and cortex should take the same approach. It should be safe to make this change since, as things stand they will either be

pkg/ruler/ruler.go
jtlisi: Among the various awful things about how rules are stored in the configdb, file names with multiple `/` are allowed 😭. So we need to return the entire path of the provided file, not including our prefix.

pkg/ruler/ruler.go
jtlisi: Not yet, will do.

pkg/ruler/ruler.go

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
jtlisi: Prometheus doesn't export this type in it's API. However, I noticed this type is reimplemented in the `prometheus/client_golang` repo so I will snag it from there.

pkg/ruler/api.go
jtlisi: I updated the protos to use a non-nullable field for timestamps and durations. This allows us to avoid using any pointers for timestamps/durations except when we marshal them into JSON, similar to the prometheus API.

pkg/ruler/api.go
```diff
@@ -0,0 +1,264 @@
...
+	if r.store == nil {
+		level.Info(util.Logger).Log("msg", "ruler configured with store that does not support api")
+		return
+	}
+	for _, route := range []struct {
+		name, method, path string
+		handler            http.HandlerFunc
+	}{
+		{"get_rules", "GET", "/api/prom/api/v1/rules", r.rules},
+		{"get_alerts", "GET", "/api/prom/api/v1/rules", r.alerts},
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/api.go
```diff
@@ -0,0 +1,252 @@
...
+type recordingRule struct {
+	Name      string        `json:"name"`
+	Query     string        `json:"query"`
+	Labels    labels.Labels `json:"labels,omitempty"`
+	Health    string        `json:"health"`
+	LastError string        `json:"lastError,omitempty"`
+	Type      v1.RuleType   `json:"type"`
+}
+
+func respondError(logger log.Logger, w http.ResponseWriter, msg string, errType v1.ErrorType) {
```
pracucci: Given that now we always pass `v1.ErrServer`, we may remove `errType` and just always use `v1.ErrServer` below.

pkg/ruler/ruler.go
```diff
@@ -409,12 +394,49 @@ func (r *Ruler) loadRules(ctx context.Context) {
...
+			if err != nil {
+				level.Error(r.logger).Log("msg", "unable to create rule manager", "user", user, "err", err)
+				return
+			}
+			manager.Run()
+			r.userManagers[user] = manager
+		}
+		err = manager.Update(r.cfg.EvaluationInterval, files, nil)
+		if err != nil {
+			level.Error(r.logger).Log("dsg", "unable to update rule manager", "user", user, "err", err)
```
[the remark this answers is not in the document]

pkg/ruler/rules/compat.go
```diff
@@ -16,7 +16,7 @@ func ToProto(user string, namespace string, rl rulefmt.RuleGroup) *RuleGroupDesc
 	rg := RuleGroupDesc{
 		Name:      rl.Name,
 		Namespace: namespace,
-		Interval:  &dur,
+		Interval:  dur,
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/api.go
jtlisi: Will do

pkg/ruler/rules/compat.go
```diff
@@ -33,7 +33,7 @@ func formattedRuleToProto(rls []rulefmt.Rule) []*RuleDesc {
 			Record: rls[i].Record,
 			Alert:  rls[i].Alert,
 
-			For:         &f,
+			For:         f,
```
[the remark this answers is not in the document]
jtlisi: 👍

PR #1941, opened 2019-12-29
Bump deployment api version

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

**What this PR does**:
extensions/v1beta1 is deprecated. Usdate to apps/v1
**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2002, opened 2020-01-19
Updated k8s Cortex yaml files

VineethReddy02:
The yaml files seems to be older and as all the services are deployemnts. 
The deployments in Kubernetes have moved from **apiVersion: extensions/v1beta1 -> apiVersion: apps/v1**  and also added the selector field as it is mandatory to the deployment spec section.
Feel free to point me out if k8s yaml files need more changes which I may have missed.

Signed-off-by: vineeth <vineethpothulapati@outlook.com>

@gouthamve @pracucci

### DIFF
--- a/k8s/alertmanager-dep.yaml
+++ b/k8s/alertmanager-dep.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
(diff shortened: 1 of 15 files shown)


pracucci:
There's already an open PR for this:
https://github.com/cortexproject/cortex/pull/1941

PR #2000, opened 2020-01-18
Prevents panic using freed slice in pool

owen-d:
This defers putting a slice back into the pool (ingester push path) to prevent panicking Error() calls which reference the underlying labels.

This defers returning an allocated slice to its pool until after use. Previously we saw panics on `Error()` calls which held references to the underlying timeseries label set.

PR #2004, opened 2020-01-20
Comment unsafe memory usage in ingester push path

bboreham:
This is roughly what I was suggesting at #2000 to warn future maintainers.

~I have left (what I assume is) the bug unchanged.~

I have changed the way errors are wrapped to make it clearer how we avoid retaining references to unsafe data inside errors.


bboreham:
See #2006

PR #2006, opened 2020-01-20
Copy label values to avoid unsafe memory use

bboreham:
This is an alternative solution to #2000 (and #2004) where we do more copying on the error path but need fewer comments to avoid mistakes.

Benchmarks show this alternative is significantly more expensive on the error path:
Before #2000 
```
BenchmarkIngesterPush/encoding=DoubleDelta	   20165041 ns/op  2590736 B/op	   12441 allocs/op
BenchmarkIngesterPushErrors/encoding=DoubleDelta   26534724 ns/op  5270153 B/op	   85194 allocs/op
```

After #2000 
```
BenchmarkIngesterPush/encoding=DoubleDelta	   20305149 ns/op  2591191 B/op	   12444 allocs/op
BenchmarkIngesterPushErrors/encoding=DoubleDelta   27959445 ns/op  5276005 B/op	   85189 allocs/op
```

This branch:
```
BenchmarkIngesterPush/encoding=DoubleDelta	   20074468 ns/op  2591331 B/op	   12598 allocs/op
BenchmarkIngesterPushErrors/encoding=DoubleDelta   35415601 ns/op 11072246 B/op	   95055 allocs/op
```

### DIFF
--- a/pkg/ingester/index/index.go
+++ b/pkg/ingester/index/index.go
@@ -32,9 +32,11 @@ func New() *InvertedIndex {
 }
 
 // Add a fingerprint under the specified labels.
+// NOTE: memory for `labels` is unsafe; anything retained beyond the
+// life of this function must be copied
 func (ii *InvertedIndex) Add(labels []client.LabelAdapter, fp model.Fingerprint) labels.Labels {
 	shard := &ii.shards[util.HashFP(fp)%indexShards]
-	return shard.add(labels, fp)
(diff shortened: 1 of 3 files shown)


pracucci:
@bboreham Do you mind rebasing? This stuff has changed in the WAL PR https://github.com/cortexproject/cortex/pull/1103 which has just been merged 😞

bboreham:
I think the benchmarks show that we don't want to do this so have updated #2004.

PR #2008, opened 2020-01-20
Update code to fix almost all pre-existing lint errors

### LINKED ISSUE #1912: Pre Golang-CI Linting Errors
With Cortex moving to golang-ci (#1853) as things stand there are a large number of issues being detected by the linter. We should look into rectifying those errors. These errors will not affect CI because #1853 will not throw an issue for errors that existed before it was merged.

zendern:
**What this PR does**:
Cleanups all but one of the lint errors that were found when the hash was in place. 

Last outstanding issue is as follows
```
pkg/querier/frontend/worker.go:60:10: SA1019: naming.Watcher is deprecated: please use package resolver.  (staticcheck)
	watcher naming.Watcher
```

Not 100% clear yet on path forward on this. From what i can tell even the underlying library is still using naming.Watcher. Wondering if for now we put a //nolint:staticcheck and open an issue to fix just that issue.

Fixes #1912 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -245,7 +245,7 @@ func (a dynamoDBStorageClient) BatchWrite(ctx context.Context, input chunk.Write
 			if awsErr, ok := err.(awserr.Error); ok && ((awsErr.Code() == dynamodb.ErrCodeProvisionedThroughputExceededException) || request.Retryable()) {
 				logWriteRetry(ctx, requests)
 				unprocessed.TakeReqs(requests, -1)
-				a.writeThrottle.WaitN(ctx, len(requests))
+				_ = a.writeThrottle.WaitN(ctx, len(requests))
 				backoff.Wait()
 				continue
 			} else if ok && awsErr.Code() == validationException {
@@ -269,7 +269,7 @@ func (a dynamoDBStorageClient) BatchWrite(ctx context.Context, input chunk.Write
 		unprocessedItems := dynamoDBWriteBatch(resp.UnprocessedItems)
 		if len(unprocessedItems) > 0 {
 			logWriteRetry(ctx, unprocessedItems)
-			a.writeThrottle.WaitN(ctx, unprocessedItems.Len())
+			_ = a.writeThrottle.WaitN(ctx, unprocessedItems.Len())
 			unprocessed.TakeReqs(unprocessedItems, -1)
 		}
 
--- a/pkg/chunk/aws/dynamodb_storage_client_test.go
+++ b/pkg/chunk/aws/dynamodb_storage_client_test.go
@@ -17,10 +17,8 @@ const (
 
 func TestChunksPartialError(t *testing.T) {
 	fixture := dynamoDBFixture(0, 10, 20)
-	defer func() {
-		err := fixture.Teardown()
-		require.NoError(t, err)
-	}()
+	defer require.NoError(t, fixture.Teardown())
+
 	_, client, err := testutils.Setup(fixture, tableName)
 	require.NoError(t, err)
 
--- a/pkg/chunk/aws/dynamodb_storage_client_test.go
+++ b/pkg/chunk/aws/dynamodb_storage_client_test.go
@@ -17,7 +17,10 @@ const (
 
 func TestChunksPartialError(t *testing.T) {
 	fixture := dynamoDBFixture(0, 10, 20)
-	defer fixture.Teardown()
+	defer func() {
+		err := fixture.Teardown()
+		require.NoError(t, err)
+	}()
 	_, client, err := testutils.Setup(fixture, tableName)
 	require.NoError(t, err)
 
--- a/pkg/chunk/aws/dynamodb_storage_client_test.go
+++ b/pkg/chunk/aws/dynamodb_storage_client_test.go
@@ -17,7 +17,7 @@ const (
 
 func TestChunksPartialError(t *testing.T) {
 	fixture := dynamoDBFixture(0, 10, 20)
-	defer require.NoError(t, fixture.Teardown())
+	defer testutils.TeardownFixture(t, fixture)
 
 	_, client, err := testutils.Setup(fixture, tableName)
 	require.NoError(t, err)
--- a/pkg/chunk/cache/instrumented.go
+++ b/pkg/chunk/cache/instrumented.go
@@ -75,7 +75,7 @@ func (i *instrumentedCache) Store(ctx context.Context, keys []string, bufs [][]b
 	}
 
 	method := i.name + ".store"
-	instr.CollectedRequest(ctx, method, requestDuration, instr.ErrorCode, func(ctx context.Context) error {
+	_ = instr.CollectedRequest(ctx, method, requestDuration, instr.ErrorCode, func(ctx context.Context) error {
 		sp := ot.SpanFromContext(ctx)
 		sp.LogFields(otlog.Int("keys", len(keys)))
 		i.Cache.Store(ctx, keys, bufs)
@@ -91,7 +91,7 @@ func (i *instrumentedCache) Fetch(ctx context.Context, keys []string) ([]string,
 		method  = i.name + ".fetch"
 	)
 
-	instr.CollectedRequest(ctx, method, requestDuration, instr.ErrorCode, func(ctx context.Context) error {
+	_ = instr.CollectedRequest(ctx, method, requestDuration, instr.ErrorCode, func(ctx context.Context) error {
 		sp := ot.SpanFromContext(ctx)
 		sp.LogFields(otlog.Int("keys requested", len(keys)))
 
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -135,7 +135,7 @@ func memcacheStatusCode(err error) string {
 
 // Fetch gets keys from the cache. The keys that are found must be in the order of the keys requested.
 func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
-	instr.CollectedRequest(ctx, "Memcache.Get", c.requestDuration, memcacheStatusCode, func(ctx context.Context) error {
+	_ = instr.CollectedRequest(ctx, "Memcache.Get", c.requestDuration, memcacheStatusCode, func(ctx context.Context) error {
 		if c.cfg.BatchSize == 0 {
 			found, bufs, missed = c.fetch(ctx, keys)
 			return nil
@@ -149,7 +149,7 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 
 func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
 	var items map[string]*memcache.Item
-	instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
+	_ = instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
 		sp := opentracing.SpanFromContext(ctx)
 		sp.LogFields(otlog.Int("keys requested", len(keys)))
 
@@ -248,7 +248,7 @@ func (c *Memcached) Stop() {
 // HashKey hashes key into something you can store in memcached.
 func HashKey(key string) string {
 	hasher := fnv.New64a()
-	hasher.Write([]byte(key)) // This'll never error.
+	_, _ = hasher.Write([]byte(key)) // This'll never error.
 
 	// Hex because memcache errors for the bytes produced by the hash.
 	return hex.EncodeToString(hasher.Sum(nil))
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -36,7 +36,7 @@ type observableVecCollector struct {
 func (observableVecCollector) Register()                             {}
 func (observableVecCollector) Before(method string, start time.Time) {}
 func (o observableVecCollector) After(method, statusCode string, start time.Time) {
-	o.v.WithLabelValues(method, statusCode).Observe(time.Now().Sub(start).Seconds())
+	o.v.WithLabelValues(method, statusCode).Observe(time.Since(start).Seconds())
 }
 
 // MemcachedConfig is config to make a Memcached
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -149,7 +149,7 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 
 func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
 	var items map[string]*memcache.Item
-	_ = instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
+	err := instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
 		sp := opentracing.SpanFromContext(ctx)
 		sp.LogFields(otlog.Int("keys requested", len(keys)))
 
@@ -165,6 +165,10 @@ func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, b
 		}
 		return err
 	})
+
+	if err != nil {
+		return found, bufs, keys
+	}
 
 	for _, key := range keys {
 		item, ok := items[key]
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -125,7 +125,7 @@ func (c *store) PutOne(ctx context.Context, from, through model.Time, chunk Chun
 		return err
 	}
 
-	c.writeBackCache(ctx, chunks)
+	_ = c.writeBackCache(ctx, chunks)
 
 	writeReqs, err := c.calculateIndexEntries(chunk.UserID, from, through, chunk)
 	if err != nil {
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -255,7 +255,8 @@ func (c *store) LabelNamesForMetricName(ctx context.Context, userID string, from
 }
 
 func (c *store) validateQueryTimeRange(ctx context.Context, userID string, from *model.Time, through *model.Time) (bool, error) {
-	log, _ := spanlogger.New(ctx, "store.validateQueryTimeRange")
+	//nolint:ineffassign,staticcheck //Leaving ctx even though we don't currently use it, we want to make it available for when we might need it and hopefully will ensure us using the correct context at that time
+	log, ctx := spanlogger.New(ctx, "store.validateQueryTimeRange")
 	defer log.Span.Finish()
 
 	if *through < *from {
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -118,6 +118,7 @@ func (c *store) Put(ctx context.Context, chunks []Chunk) error {
 
 // PutOne implements ChunkStore
 func (c *store) PutOne(ctx context.Context, from, through model.Time, chunk Chunk) error {
+	log, ctx := spanlogger.New(ctx, "ChunkStore.PutOne")
 	chunks := []Chunk{chunk}
 
 	err := c.storage.PutChunks(ctx, chunks)
@@ -125,7 +126,9 @@ func (c *store) PutOne(ctx context.Context, from, through model.Time, chunk Chun
 		return err
 	}
 
-	_ = c.writeBackCache(ctx, chunks)
+	if cacheErr := c.writeBackCache(ctx, chunks); cacheErr != nil {
+		level.Warn(log).Log("msg", "could not store chunks in chunk cache", "err", cacheErr)
+	}
 
 	writeReqs, err := c.calculateIndexEntries(chunk.UserID, from, through, chunk)
 	if err != nil {
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -252,7 +252,7 @@ func (c *store) LabelNamesForMetricName(ctx context.Context, userID string, from
 }
 
 func (c *store) validateQueryTimeRange(ctx context.Context, userID string, from *model.Time, through *model.Time) (bool, error) {
-	log, ctx := spanlogger.New(ctx, "store.validateQueryTimeRange")
+	log, _ := spanlogger.New(ctx, "store.validateQueryTimeRange")
 	defer log.Span.Finish()
 
 	if *through < *from {
--- a/pkg/chunk/composite_store_test.go
+++ b/pkg/chunk/composite_store_test.go
@@ -3,9 +3,10 @@ package chunk
 import (
 	"context"
 	"fmt"
-	"github.com/stretchr/testify/require"
 	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
--- a/pkg/chunk/composite_store_test.go
+++ b/pkg/chunk/composite_store_test.go
@@ -180,7 +180,10 @@ func TestCompositeStore(t *testing.T) {
 	} {
 		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
 			have := []result{}
-			tc.cs.forStores(model.TimeFromUnix(tc.from), model.TimeFromUnix(tc.through), collect(&have))
+			err := tc.cs.forStores(model.TimeFromUnix(tc.from), model.TimeFromUnix(tc.through), collect(&have))
+			if err != nil {
+				t.Fatal(err)
+			}
 			if !reflect.DeepEqual(tc.want, have) {
 				t.Fatalf("wrong stores - %s", test.Diff(tc.want, have))
 			}
--- a/pkg/chunk/composite_store_test.go
+++ b/pkg/chunk/composite_store_test.go
@@ -3,6 +3,7 @@ package chunk
 import (
 	"context"
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"reflect"
 	"testing"
 
@@ -181,9 +182,7 @@ func TestCompositeStore(t *testing.T) {
 		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
 			have := []result{}
 			err := tc.cs.forStores(model.TimeFromUnix(tc.from), model.TimeFromUnix(tc.through), collect(&have))
-			if err != nil {
-				t.Fatal(err)
-			}
+			require.NoError(t, err)
 			if !reflect.DeepEqual(tc.want, have) {
 				t.Fatalf("wrong stores - %s", test.Diff(tc.want, have))
 			}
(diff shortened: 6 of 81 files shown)


zendern:
@jtlisi let me know if you have feeling or thoughts on how to handle the last outstanding lint issue.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -59,8 +59,6 @@ type storageClientColumnKey struct {
 	schemaCfg chunk.SchemaConfig
 	client    *bigtable.Client
 	keysFn    keysFn
-
-	distributeKeys bool
```
jtlisi: Just noting this is safe to remove because it is never called and the boolean in config is used instead.

pkg/util/grpcclient/backoff_retry.go
```diff
@@ -2,6 +2,7 @@ package grpcclient
 
 import (
 	"context"
+	"google.golang.org/grpc/status"
```
jtlisi: Can you move `"google.golang.org/grpc/status"` down with the rest of the 3rd party imports?

zendern: Fixed in 30f9c39

zendern:
Opened #2015 to deal with that last outstanding lint issue. All checks are now passing. @jtlisi should be ready for you to take a look at again.

pkg/util/http.go
```diff
@@ -129,7 +129,9 @@ func SerializeProtoResponse(w http.ResponseWriter, resp proto.Message, compressi
 	case NoCompression:
 	case FramedSnappy:
 		buf := bytes.Buffer{}
-		if _, err := snappy.NewWriter(&buf).Write(data); err != nil {
+		writer := snappy.NewBufferedWriter(&buf)
+		defer writer.Close()
```
bboreham: Don't we have to close the compressor before calling `buf.Bytes()` ?

(we try not to use FramedSnappy at all, but it's here for backwards-compatibility)

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -155,7 +155,7 @@ func (a s3ObjectClient) bucketFromKey(key string) string {
 	}
 
 	hasher := fnv.New32a()
-	hasher.Write([]byte(key))
+	_, _ = hasher.Write([]byte(key))
```
bboreham: Could add `hash.Hash32.Write` to the project `.errcheck-exclude`

pkg/chunk/cache/memcached.go
```diff
@@ -248,7 +248,7 @@ func (c *Memcached) Stop() error {
 // HashKey hashes key into something you can store in memcached.
 func HashKey(key string) string {
 	hasher := fnv.New64a()
-	hasher.Write([]byte(key)) // This'll never error.
+	_, _ = hasher.Write([]byte(key)) // This'll never error.
```
bboreham: Could add this to the project `.errcheck-exclude`

pkg/distributor/billing.go
```diff
@@ -22,7 +22,7 @@ func (d *Distributor) emitBillingRecord(ctx context.Context, buf []byte, samples
 
 	now := time.Now().UTC()
 	hasher := sha256.New()
-	hasher.Write(buf)
+	_, _ = hasher.Write(buf)
```
bboreham: Could add to the project `.errcheck-exclude`

pkg/ingester/client/timeseries.go
```diff
@@ -245,7 +245,7 @@ func ReuseSlice(slice []PreallocTimeseries) {
 	for i := range slice {
 		ReuseTimeseries(slice[i].TimeSeries)
 	}
-	slicePool.Put(slice[:0])
+	slicePool.Put(slice[:0]) //nolint:staticcheck
```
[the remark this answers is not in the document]
zendern: So this is the error from the linter.
```
pkg/ingester/client/timeseries.go:248:15: SA6002: argument should be pointer-like to avoid allocations (staticcheck)
	slicePool.Put(slice[:0])
	             ^
```

From my understanding of what we are trying to do here we want to reuse the slice but and reset it back to empty. What the linter would like is for everything that goes into a SlicePool to be a pointer. This is simply a slice of pointers so I guess my thought was that the overhead isn't a big deal. I will be completely honest here. I have not done much with SlicePools. But we can switch it to be a pointer as well. 

It'll just require more changes up the stack anywhere we pull `[]PreallocTimeseries` out of the slicePool.

pkg/chunk/composite_store_test.go
```diff
@@ -173,7 +173,10 @@ func TestCompositeStore(t *testing.T) {
 	} {
 		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
 			have := []result{}
-			tc.cs.forStores(model.TimeFromUnix(tc.from), model.TimeFromUnix(tc.through), collect(&have))
+			err := tc.cs.forStores(model.TimeFromUnix(tc.from), model.TimeFromUnix(tc.through), collect(&have))
+			if err != nil {
```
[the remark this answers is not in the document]
zendern: No reason other than to keep the test consistent. But I can do that and update the rest of the test class as well.

pkg/ingester/client/timeseries.go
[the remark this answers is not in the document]
jtlisi: A slice itself is a pointer to an underlying array.  This is super minor.

pkg/chunk/schema_util.go
```diff
@@ -112,16 +112,16 @@ func decodeBase64Value(bs []byte) (model.LabelValue, error) {
...
+	encodedThroughBytes := make([]byte, 8)
 	hex.Encode(encodedThroughBytes, throughBytes)
 	return encodedThroughBytes
 }
 
 func decodeTime(bs []byte) uint32 {
-	buf := make([]byte, 4, 4)
-	hex.Decode(buf, bs)
+	buf := make([]byte, 4)
+	_, _ = hex.Decode(buf, bs)
```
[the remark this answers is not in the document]
zendern: Fixed here 16d14b7

pkg/chunk/aws/dynamodb_storage_client_test.go
```diff
@@ -17,7 +17,10 @@ const (
 
 func TestChunksPartialError(t *testing.T) {
 	fixture := dynamoDBFixture(0, 10, 20)
-	defer fixture.Teardown()
+	defer func() {
```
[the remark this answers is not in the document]
zendern: Cleaned up here adb85c0

pkg/chunk/series_store.go
```diff
@@ -457,7 +457,7 @@ func (c *seriesStore) PutOne(ctx context.Context, from, through model.Time, chun
 			return err
 		}
 	}
-	c.writeBackCache(ctx, chunks)
+	_ = c.writeBackCache(ctx, chunks)
```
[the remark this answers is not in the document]
zendern: Added logging here ae6decc

pkg/chunk/storage/utils_test.go
```diff
@@ -34,7 +34,7 @@ func forAllFixtures(t *testing.T, storageClientTest storageClientTest) {
 		t.Run(fixture.Name(), func(t *testing.T) {
 			indexClient, objectClient, err := testutils.Setup(fixture, tableName)
 			require.NoError(t, err)
-			defer fixture.Teardown()
+			defer fixture.Teardown() //nolint:errcheck
```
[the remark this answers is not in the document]
zendern: Done 4746656

pkg/chunk/composite_store_test.go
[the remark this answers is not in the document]
zendern: Updated all instances of if err != nil to require.NoError here 18e6e59

pkg/configs/db/timed.go
```diff
@@ -56,7 +47,7 @@ func (t timed) GetAllConfigs(ctx context.Context) (map[string]configs.View, erro
 		cfgs map[string]configs.View
 		err  error
 	)
-	instrument.CollectedRequest(ctx, "DB.GetAllConfigs", databaseRequestDuration, instrument.ErrorCode, func(ctx context.Context) error {
+	_ = instrument.CollectedRequest(ctx, "DB.GetAllConfigs", databaseRequestDuration, instrument.ErrorCode, func(ctx context.Context) error {
```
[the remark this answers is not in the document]
zendern: Updated in 9202a71

pkg/chunk/chunk_store_utils.go
```diff
@@ -125,7 +125,7 @@ func NewChunkFetcher(cfg cache.Config, cacheStubs bool, storage ObjectClient) (*
 func (c *Fetcher) Stop() {
 	close(c.decodeRequests)
 	c.wait.Wait()
-	c.cache.Stop()
+	_ = c.cache.Stop()
```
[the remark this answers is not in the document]
zendern: So if we want to handle this error I would think instead of logging it here we should let it flow up the stack.

Stop() gets called here which has a return of error
https://github.com/cortexproject/cortex/blob/ab3e8360fe30150a516448e4c3bae596dc4d8a17/pkg/cortex/modules.go#L284

Which is defined in the module definition for what "stop" should do.
https://github.com/cortexproject/cortex/blob/ab3e8360fe30150a516448e4c3bae596dc4d8a17/pkg/cortex/modules.go#L511-L527

And then as modules are stopped if one errors we log it at that level
https://github.com/cortexproject/cortex/blob/ab3e8360fe30150a516448e4c3bae596dc4d8a17/pkg/cortex/cortex.go#L265-L272

But currently the hinderance on easily doing that is the Worker interface does not allow error to be returned.
https://github.com/cortexproject/cortex/blob/ab3e8360fe30150a516448e4c3bae596dc4d8a17/pkg/querier/frontend/worker.go#L49-L51

So do you have any feelings on updating that interface to allow for that?? My gut says maybe we split that out into its own issue and leave this as is since right now it's no worse than it currently is on master.

pkg/chunk/cache/instrumented.go
```diff
@@ -75,7 +75,7 @@ func (i *instrumentedCache) Store(ctx context.Context, keys []string, bufs [][]b
 	}
 
 	method := i.name + ".store"
-	instr.CollectedRequest(ctx, method, requestDuration, instr.ErrorCode, func(ctx context.Context) error {
+	_ = instr.CollectedRequest(ctx, method, requestDuration, instr.ErrorCode, func(ctx context.Context) error {
```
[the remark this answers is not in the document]
jtlisi: FWIW from `pkg/chunk/cache/cache.go`

```go
// Cache byte arrays by key.
//
// NB we intentionally do not return errors in this interface - caching is best
// effort by definition.  We found that when these methods did return errors,
// the caller would just log them - so its easier for implementation to do that.
// Whats more, we found partially successful Fetchs were often treated as failed
// when they returned an error.
type Cache interface {
	Store(ctx context.Context, key []string, buf [][]byte)
	Fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missing []string)
	Stop() error
}
```

pkg/util/http.go
jtlisi: That looks to be the case, this `Close()` call can not be deferred
```go
// The Writer returned buffers writes. Users must call Close to guarantee all
// data has been forwarded to the underlying io.Writer. They may also call
// Flush zero or more times before calling Close.
func NewBufferedWriter(w io.Writer) *Writer {
	return &Writer{
		w:    w,
		ibuf: make([]byte, 0, maxBlockSize),
		obuf: make([]byte, obufLen),
	}
}
```

pkg/chunk/chunk_store_utils.go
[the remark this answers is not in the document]
bboreham: I vote to remove the error return - #2030

pkg/ingester/client/timeseries.go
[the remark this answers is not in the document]
bboreham: We cannot pool these as pointer-to-slice because the place we use them is in `WriteRequest` which is generated from Protobuf and we don't have an option to make it a pointer.

So the warning is correct: we generate 24 bytes of garbage every time a `PreallocTimeseries` is re-used.  But since the slices are far far larger, we come out ahead.

I think there should be a comment on `slicePool` to explain this.

pkg/chunk/chunk_store_utils.go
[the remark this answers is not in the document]
pracucci: @zendern I've merged the PR https://github.com/cortexproject/cortex/pull/2030. May you rebase `master`, please?

pkg/chunk/schema_config.go
```diff
@@ -315,7 +315,7 @@ func (cfg *SchemaConfig) Load() error {
 // PrintYaml dumps the yaml to stdout, to aid in migration
 func (cfg SchemaConfig) PrintYaml() {
```
pracucci: This function is never called. We may drop it (in case will be required in the future, it's fairly easy to rewrite it).

pkg/querier/frontend/frontend.go
```diff
@@ -210,7 +210,7 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, req *ProcessRequest) (*Pro
 	tracer, span := opentracing.GlobalTracer(), opentracing.SpanFromContext(ctx)
 	if tracer != nil && span != nil {
 		carrier := (*httpgrpcHeadersCarrier)(req.HttpRequest)
-		tracer.Inject(span.Context(), opentracing.HTTPHeaders, carrier)
+		_ = tracer.Inject(span.Context(), opentracing.HTTPHeaders, carrier)
```
pracucci: @bboreham Should we add `tracer.Inject` to the exclude list?

pkg/querier/queryrange/instrumentation.go
```diff
@@ -20,7 +20,7 @@ func InstrumentMiddleware(name string) Middleware {
 	return MiddlewareFunc(func(next Handler) Handler {
 		return HandlerFunc(func(ctx context.Context, req Request) (Response, error) {
 			var resp Response
-			err := instrument.TimeRequestHistogram(ctx, name, queryRangeDuration, func(ctx context.Context) error {
+			err := instrument.CollectedRequest(ctx, name, instrument.NewHistogramCollector(queryRangeDuration), instrument.ErrorCode, func(ctx context.Context) error {
```
pracucci: Could you explain me this change, please?

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -251,7 +251,7 @@ func (a dynamoDBStorageClient) BatchWrite(ctx context.Context, input chunk.Write
 			if awsErr, ok := err.(awserr.Error); ok && ((awsErr.Code() == dynamodb.ErrCodeProvisionedThroughputExceededException) || request.Retryable()) {
 				logWriteRetry(ctx, requests)
 				unprocessed.TakeReqs(requests, -1)
-				a.writeThrottle.WaitN(ctx, len(requests))
+				_ = a.writeThrottle.WaitN(ctx, len(requests))
```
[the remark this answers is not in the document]
bboreham: 1. can't happen; it's a hard error from DynamoDB to go over the limit.
2. from #1361 "I did wonder whether to remove the backoff object, but it's still handling max retries and cancellation, so left it in place."

pkg/chunk/cache/memcached.go
```diff
@@ -148,7 +148,7 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 
 func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
 	var items map[string]*memcache.Item
-	instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
+	_ = instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
```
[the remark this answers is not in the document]
bboreham: I can't see much more we could do with the error.  Maybe the inner function should just return `nil` ?
Broadly Cortex thinks that caches are things we can do without.

pkg/querier/queryrange/instrumentation.go
bboreham: `TimeRequestHistogram` is commented as deprecated.

pkg/querier/frontend/frontend.go
bboreham: sure, whatever.

pkg/chunk/aws/dynamodb_storage_client.go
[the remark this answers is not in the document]
pracucci: Ok. Let's keep it as is then. Thanks for the details.

pkg/querier/queryrange/instrumentation.go
zendern: For even more context all I did was grab the implementation of what `TimeRequestHistogram` was doing and put it here.

pkg/chunk/cache/memcached.go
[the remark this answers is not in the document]
pracucci: If an error occurred, we should just skip the `for` loop below and return:

```
if err != nil {
  return found, bufs, keys
}

File: pkg/distributor/billing.go (hunk 6 above)
Comment (zendern) [reply pkg/distributor/billing.go]: @bboreham Not having any luck getting this method added to .errcheck-exclude. I think this is the problem 

https://github.com/golangci/golangci-lint/issues/656

Are you ok with leaving it or would you rather see a comment and I can call out that issue.

Comment (bboreham) [reply pkg/distributor/billing.go]: I think I had a lot of trouble stemming from the way it doesn't re-read the config file unless it thinks something else has changed.  So you need to touch the file or `make clean` or something like that.  I can't remember exactly.

Comment (zendern) [reply pkg/distributor/billing.go]: Ah ok... I'll give that a go and see what happens. Thanks

File: pkg/util/http.go (hunk 3 above)
Comment (zendern) [reply pkg/util/http.go]: Fixed in 1c157c6

File: pkg/util/grpcclient/ratelimit.go [hunk 21]
```diff
@@ -15,7 +15,7 @@ func NewRateLimiter(cfg *Config) grpc.UnaryClientInterceptor {
 	}
 	limiter := rate.NewLimiter(rate.Limit(cfg.RateLimit), burst)
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-		limiter.Wait(ctx)
+		_ = limiter.Wait(ctx)
```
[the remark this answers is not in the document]
zendern: Fixed in de753ec

pkg/chunk/chunk_store.go
```diff
@@ -249,7 +249,7 @@ func (c *store) LabelNamesForMetricName(ctx context.Context, userID string, from
 }
 
 func (c *store) validateQueryTimeRange(ctx context.Context, userID string, from *model.Time, through *model.Time) (bool, error) {
-	log, ctx := spanlogger.New(ctx, "store.validateQueryTimeRange")
+	log, _ := spanlogger.New(ctx, "store.validateQueryTimeRange")
```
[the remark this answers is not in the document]
zendern: Fixed in 099b7d3

pkg/ingester/client/timeseries.go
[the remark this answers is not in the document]
zendern: Comment added here 8b4be65

pkg/ring/lifecycler.go
```diff
@@ -422,7 +422,7 @@ loop:
 	}
 
 	// Mark ourselved as Leaving so no more samples are send to us.
-	i.changeState(context.Background(), LEAVING)
+	_ = i.changeState(context.Background(), LEAVING)
```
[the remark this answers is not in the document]
zendern: Added log message here 5466b4d

pkg/querier/frontend/frontend.go
```diff
@@ -167,7 +167,7 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 		hs[h] = vs
 	}
 	w.WriteHeader(resp.StatusCode)
-	io.Copy(w, resp.Body)
+	_, _ = io.Copy(w, resp.Body)
```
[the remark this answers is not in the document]
zendern: Added to global file here fcc1c31

pkg/chunk/cache/memcached.go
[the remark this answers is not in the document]
zendern: Fixed in e68ad20

pkg/chunk/schema_config.go
zendern: Removed in a90e15d

pkg/util/grpcclient/ratelimit.go
```diff
@@ -15,7 +15,10 @@ func NewRateLimiter(cfg *Config) grpc.UnaryClientInterceptor {
 	}
 	limiter := rate.NewLimiter(rate.Limit(cfg.RateLimit), burst)
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-		limiter.Wait(ctx)
+		err := limiter.Wait(ctx)
+		if err != nil {
+			return err
```
pracucci: When used in conjunction with `NewBackoffRetry()` (see `grpcclient.go`), the backoff expects a `ResourceExhausted` error code.

@bboreham what do you think returning that GRPC error here?

pkg/storage/tsdb/config.go
```diff
@@ -60,6 +60,7 @@ func (d *DurationList) String() string {
 	return strings.Join(values, ",")
 }
 
+//nolint:staticcheck
```
pracucci: What's the error here if we remove this?

pkg/ring/lifecycler.go
```diff
@@ -422,7 +422,10 @@ loop:
 	}
 
 	// Mark ourselved as Leaving so no more samples are send to us.
-	i.changeState(context.Background(), LEAVING)
+	err := i.changeState(context.Background(), LEAVING)
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "failed to set state to LEAVING", "ring", i.RingName, "err", err)
```
pracucci: `to set state` > `to change the state`

pkg/querier/frontend/frontend.go
pracucci: @zendern May you do it, please?

pkg/distributor/billing.go
pracucci: @zendern May you do it, please?

pkg/chunk/cache/memcached.go
pracucci: @zendern May you do it, please?

pkg/chunk/aws/s3_storage_client.go
pracucci: @zendern May you do it, please?

pkg/querier/frontend/frontend.go
zendern: Fixed in ccca019

zendern:
@bboreham @pracucci so looking into golangci and errcheck i dont believe I am able to add the excludes on `(hash.Hash).Write` as of now. golangci has forked errcheck and is using their own version. 

Using errcheck directly the above exclude works b/c they have built in support for embedded interfaces. Whereas the golangci fork of errcheck is out of date and does not have that code. 

https://github.com/golangci/errcheck/tree/master/internal/errcheck

https://github.com/kisielk/errcheck/tree/master/internal/errcheck

I'm working on a PR to try to bring the golangci-lint version up to the same as errcheck master but that will take a little time. So how would you like to proceed??

pkg/ring/lifecycler.go
zendern: Sorry Im not following @pracucci ..... are you suggesting I change the reported message?? I followed the lead of what we have here.

https://github.com/cortexproject/cortex/blob/b4a27d9d9732fdc118450bc8b68bfdcc55a3dd10/pkg/ring/lifecycler.go#L400-L403

pkg/storage/tsdb/config.go
zendern: This was my fault....looks like it got added accidentally and was not needed at all. Good catch. This has been cleaned up and I validated the other //nolint messages and they appear to all be valid.

pkg/ring/lifecycler.go
pracucci: Yes, I was suggesting to change to message but even the current one is fine if already used in other places. Let me resolve this comment.

pracucci:
> so looking into golangci and errcheck i dont believe I am able to add the excludes on (hash.Hash).Write as of now.
> So how would you like to proceed??

Personally I'm fine even without an override if it's not easy to do. All in all, it affect 3 places and we can always re-iterate on it once supported.

pkg/util/grpcclient/ratelimit.go
bboreham: That sounds like a bug.
"Error: I'm not going to wait because the context will expire before the wait finishes"
"Never mind, try again!"

zendern: So @bboreham  / @pracucci  if i am understanding correctly what you are saying is that making this change is the wrong thing to do. Basically since the wait would not return a ResourceExhausted that would cause the BackoffRetry, if its being used, to now exit and no longer retry like we expect it to.

So if that is correct the right course of action is to just go back to what we had in originally here
https://github.com/cortexproject/cortex/pull/2008#discussion_r369464567
and ignore the error or should the BackoffRetry code check for another error code along with `ResourceExhausted`.

pkg/chunk/aws/s3_storage_client.go
zendern: @jtlisi and @pracucci ok'd this so going to mark all the `hash.Hash32.Write to the project .errcheck-exclude` as resolved since it a limitation to golangci-lint until the errcheck version is updated. I did open this to address it https://github.com/golangci/golangci-lint/issues/959. So hopefully I or someone can get around to fixing that.

pkg/chunk/cache/instrumented.go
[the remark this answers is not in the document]
zendern: @bboreham just double checking are we good with this?? as per the comment above it was very intentional.

pkg/chunk/testutils/testutils.go
```diff
@@ -84,9 +90,13 @@ func dummyChunkFor(now model.Time, metric labels.Labels) chunk.Chunk {
...
 	// Force checksum calculation.
-	err := chunk.Encode()
+	err = chunk.Encode()
 	if err != nil {
 		panic(err)
 	}
 	return chunk
 }
+
+func TeardownFixture(t *testing.T, fixture Fixture) {
```
pracucci: Is there any specific reason why this function was introduced? It looks a bit superfluous to me, given we can directly call `require.NoError(t, fixture.Teardown())` instead of `TeardownFixture()`.

pkg/util/grpcclient/ratelimit.go
pracucci: I would not rollback but fix it. We may do:
`return status.New(codes.ResourceExhausted, err.Error())`

But I would be glad to also see a unit test on it.

pracucci:
@zendern Could you attempt another rebase and address the open comment, please? I'm hearing your pain continuously rebasing this: I'm going to stress other maintainers to review this as well, so that we can get this merged.

pkg/chunk/testutils/testutils.go
zendern: It was done in this commit. 8bf4365 It was just something i noticed we were doing 2 places so felt like a natural thing since the fixture was already in the testutils.go class. I could go either way....ill leave that up to you but that was the reasoning.

pkg/util/grpcclient/ratelimit.go
zendern: I'll look at adding a unit test for this and getting that change in place.

pracucci: Ok thanks! This looks like the very last thing left to me.

zendern: @pracucci So i wrote up this unit test last night.....which might not be enough honestly. It's fine that it can validate that its a status with the expected result but an integration test would probably drive home the fact that it's necessary if you configure it with the backoffRetry. I just ran out of time.....I'll see if i can get to it over the weekend.  

```package grpcclient_test

import (
	"context"
	"testing"

	"github.com/cortexproject/cortex/pkg/util/grpcclient"
	assert2 "github.com/stretchr/testify/assert"
	"google.golang.org/grpc"
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
)

func TestRateLimiterFailureResultsInResourceExhaustedError(t *testing.T) {
	config := grpcclient.Config{
		RateLimitBurst: 0,
		RateLimit: 0,
	}
	conn := grpc.ClientConn{}
	invoker := func(currentCtx context.Context, currentMethod string, currentReq, currentRepl interface{}, currentConn *grpc.ClientConn, currentOpts ...grpc.CallOption) error {
		return nil
	}

	limiter := grpcclient.NewRateLimiter(&config)
	err := limiter(context.Background(), "methodName", "", "expectedReply", &conn, invoker)
	
	if se, ok := err.(interface {
		GRPCStatus() *status.Status
	}); ok {
		assert2.Equal(t, se.GRPCStatus().Code(), codes.ResourceExhausted)
		assert2.Equal(t, se.GRPCStatus().Message(), "rate: Wait(n=1) exceeds limiter's burst 0")
	} else{
		assert2.Fail(t, "Could not convert error into expected Status type")
	}
}
```

PR #2016, opened 2020-01-21
[Experimental] Memberlist singleton and support for multiple codecs

pstibrany:
This PR changes experimental memberlist implementation of KV store in following ways:
- `memberlist.Client` is now only a small wrapper around `memberlist.KV`. KV is a top-level cortex component (but cannot be run on its own), and is a singleton. KV is only initialized if any component actually wants to use memberlist as its KV store.
- Instead of using custom binary protocol, memberlist KV store now exchanges protobuf messages defined in kv.proto
- memberlist.KV now supports multiple codecs. For now, this is only used in tests. Codecs must be registered during KV initialization.

These changes allow Cortex in single-binary mode to use memberlist as KV store. Previously, such attempts would fail with panic, since all components (distributor/ingester/querier) tried to create a memberlist instance, and it failed with metrics registration. Even if it didn't fail there, it wouldn't work because each instance would try to open same port for listening.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/cortex/modules.go
```diff
@@ -306,6 +311,7 @@ func (t *Cortex) stopQuerierChunkStore() error {
 
 func (t *Cortex) initIngester(cfg *Config) (err error) {
 	cfg.Ingester.LifecyclerConfig.RingConfig.KVStore.Multi.ConfigProvider = multiClientRuntimeConfigChannel(t.runtimeConfig)
+	cfg.Ingester.LifecyclerConfig.RingConfig.KVStore.MemberlistKV = t.memberlistKVState.getMemberlistKV
```
jtlisi: I think you will also have to inject this into the ruler and the distributor. The ruler runs a separate ring for rule evals and the distributor runs a separate ring for HA. Or are we going to assume people will set those rings up on separate ports/kv stores?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -673,64 +735,50 @@ func (m *Client) LocalState(join bool) []byte {
...
 
-	buf := bytes.NewBuffer(stateMsg)
+	kvPair := KeyValuePair{}
 
 	var err error
-	for buf.Len() > 0 {
-		keyLen := byte(0)
-		keyLen, err = buf.ReadByte()
-		if err != nil {
+	for len(data) > 0 {
```
jtlisi: I had a hard time understanding the slicing in this for loop. Could you add some comments explaining what `data` represents. It made more sense when I looked at some comments in other functions.

pkg/cortex/modules.go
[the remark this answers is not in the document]
pstibrany: You're correct. Such configuration will not be possible anymore, there is only one memberlist-KV configuration now, which is global.

pkg/ring/kv/memberlist/memberlist_client.go
pstibrany: Comment in `LocalState` was wrong (previous format), I've fixed it, and added comments into `MergeRemoteState` as well.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 * [BUGFIX] Experimental TSDB: fixed `/all_user_stats` and `/api/prom/user_stats` endpoints when using the experimental TSDB blocks storage. #2042
 * [BUGFIX] Experimental TSDB: fixed ruler to correctly work with the experimental TSDB blocks storage. #2101
 * [BUGFIX] Azure Blob ChunkStore: Fixed issue causing `invalid chunk checksum` errors. #2074
+* [BUGFIX] Experimental Memberlist KV store can now be used in single-binary Cortex. Attempts to use it previously would fail with panic. This change also breaks existing binary protocol used to exchange gossip messages, so this version will not be able to understand gossiped Ring when used in combination with the previous version of Cortex. #2016
```
pracucci: > This change also breaks existing binary protocol used to exchange gossip messages, so this version will not be able to understand gossiped Ring when used in combination with the previous version of Cortex.

Because of this may worth switching from `[BUGFIX]` to `[CHANGE]`. Could you also explain practically how a Cortex user using memberlist should upgrade to the next Cortex release? If it's more than a sentence, we may add a sub section after the changelog entries in `master` like we did for other changes in `0.6.0`.

pkg/cortex/cortex.go
```diff
@@ -84,6 +85,7 @@ type Config struct {
 	ConfigStore   config_client.Config                       `yaml:"config_store,omitempty"`
 	Alertmanager  alertmanager.MultitenantAlertmanagerConfig `yaml:"alertmanager,omitempty"`
 	RuntimeConfig runtimeconfig.ManagerConfig                `yaml:"runtime_config,omitempty"`
+	MemberlistKV  memberlist.KVConfig                        `yaml:"memberlist_kv"`
```
pracucci: Do you think it's more clear or confusing if we rename the YAML option from `memberlist_kv` to `memberlist`? The `kv` suffix may not me much clear to the final user.

If we keep `memberlist_kv` then I would suggest to rename the root block to `memberlist_kv_config` in `tools/doc-generator/main.go`.

pkg/ring/kv/codec/codec.go
```diff
@@ -9,16 +9,29 @@ import (
 type Codec interface {
 	Decode([]byte) (interface{}, error)
 	Encode(interface{}) ([]byte, error)
+
+	// Short ID to communicate what codec should be used to decode the value.
```
pracucci: Have you considered to begin the comment with `CodecID` to be godoc compliant?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -533,78 +582,84 @@ func (m *Client) trySingleCas(key string, f func(in interface{}) (out interface{
...
+		level.Warn(util.Logger).Log("msg", "Failed to unmarshal received KV Pair", "err", err)
 		m.numberOfInvalidReceivedMessages.Inc()
 		return
 	}
 
-	keyLen := int(msg[0])
-	if len(msg) <= 1+keyLen {
-		level.Warn(util.Logger).Log("msg", "Too short message received", "length", len(msg))
+	if len(kvPair.Key) == 0 {
+		level.Warn(util.Logger).Log("msg", "Invalid KV Pair, empty key")
```
pracucci: Maybe "Received an invalid KV Pair (empty key)" may be more clear.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -632,38 +687,45 @@ func (m *Client) GetBroadcasts(overhead, limit int) [][]byte {
...
+func (m *KV) LocalState(join bool) []byte {
 	m.numberOfPulls.Inc()
 
 	m.storeMu.Lock()
 	defer m.storeMu.Unlock()
 
-	// For each Key/Value pair in our store, we write:
-	// [1 byte key length] [key] [4-bytes value length] [value]
+	// For each Key/Value pair in our store, we write
+	// [4-bytes length of KV pair] [KV pair]
```
pracucci: `length of KV pair` > `length of marshalled KV pair` may be more clear (I didn't get it at the beginning)

pstibrany:
> Would you be up to add a integration test with single-binary Cortex and gossip to (hopefully) freeze the current implementation from future regressions?

Added integration test that verifies that Cortex in single-binary mode works using memberlist setup.

CHANGELOG.md
pstibrany: Moved it to CHANGE. Ideally to upgrade one would shutdown old Cortex, and start everything with new version... I know that we didn't do that :) We simply rolled out new version to the cluster, and since distributors are typically fastest ones to restart, they couldn't dispatch writes until new ingesters joined. But after they did, everything worked fine again.

I don't think that anybody else really runs this setup at the moment.

pkg/cortex/modules.go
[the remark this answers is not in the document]
pracucci: Sorry to bother you. I've just merged the PR #2113. May you rebase `master` and add the same change to the compactor as well, which uses the ring for sharding, please?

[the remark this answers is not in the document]
pstibrany: > add the same change to the compactor as well, which uses the ring for sharding, please

Done

PR #2018, opened 2020-01-22
Fix flaky tests in compactor

pracucci:
**What this PR does**:
In the PR #1942 I did a mistake in the compactor tests which cause some flaky tests. The compactor tests mock the bucket client **after** the compactor is started and this leads to flaky tests (unfrequent, but seen in CI). This PR fixes it.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/compactor/compactor_test.go
+++ b/pkg/compactor/compactor_test.go
@@ -69,10 +69,11 @@ func TestConfig_ShouldSupportCliFlags(t *testing.T) {
 func TestCompactor_ShouldDoNothingOnNoUserBlocks(t *testing.T) {
 	t.Parallel()
 
-	c, bucketClient, _, logs, registry := prepare(t)
-
 	// No user blocks stored in the bucket.
+	bucketClient := &cortex_tsdb.BucketClientMock{}
 	bucketClient.MockIter("", []string{}, nil)
+
+	c, _, logs, registry := prepare(t, bucketClient)
 
 	// Wait until a run has completed.
 	cortex_testutil.Poll(t, time.Second, 1.0, func() interface{} {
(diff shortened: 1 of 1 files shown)


pkg/compactor/compactor_test.go
```diff
@@ -203,14 +206,13 @@ func TestCompactor_ShouldIterateOverUsersAndRunCompaction(t *testing.T) {
...
 
-func prepare(t *testing.T) (*Compactor, *cortex_tsdb.BucketClientMock, *tsdbCompactorMock, *bytes.Buffer, prometheus.Gatherer) {
+func prepare(t *testing.T, bucketClient *cortex_tsdb.BucketClientMock) (*Compactor, *tsdbCompactorMock, *bytes.Buffer, prometheus.Gatherer) {
 	compactorCfg := Config{}
 	storageCfg := cortex_tsdb.Config{}
 	flagext.DefaultValues(&compactorCfg, &storageCfg)
 	compactorCfg.retryMinBackoff = 0
 	compactorCfg.retryMaxBackoff = 0
 
-	bucketClient := &cortex_tsdb.BucketClientMock{}
```
sandeepsukhani: Changes look good to me but I am trying to understand how this was causing tests to fail intermittently since it was still created before calling `newCompactor` down below. Anything I am missing to see here?

pracucci: It was created before `newCompactor()` but mocked later (after the `prepare()` exits).

sandeepsukhani: Right, my bad. Changes look more relevant for [this block](https://github.com/cortexproject/cortex/pull/2018/files#diff-e6297dff640b47246ada70cd6663f9c6R152-R157)

Thanks!

PR #2019, opened 2020-01-22
Fix typos in WAL arguments doc

codesome:


docs/configuration/arguments.md
```diff
@@ -308,7 +308,7 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
 #### WAL
 
 - `--ingester.wal-dir`
```
pracucci: I've just noticed that the WAL flags are documented with the `--` prefix while it should be a single `-`. May you fix it, please?

codesome: Will fix it tomorrow

codesome:
Flags fixed

pracucci:
Thanks @codesome !

Issue #2003, opened 2020-01-20
Set a max concurrency TSDB blocks compaction


pracucci:
Every two hours (by default) each ingester cuts blocks from head nearly at the same time. Each ingester does it for any tenant, because we have 1 TSDB opened for each tenant. We're observing an increased latency in the ingesters when this occur (other than an increased memory usage): the more tenants are writing to an ingester, the more this problem is apparent.

We could call `DisableCompactions()` on each TSDB right after creation and have a loop inside Cortex to periodically call `Compact()` on each TSDB (`Compact()` has been made public in the PR https://github.com/prometheus/prometheus/pull/6667). This way we can control the max concurrency.

pstibrany:
I have a code ready at https://github.com/pstibrany/cortex/tree/tsdb-blocks-compaction-concurrency, but waiting for PR #1970 to be merged first.

pracucci:
> but waiting for PR #1970 to be merged first

@pstibrany The PR #1970 has been merged.

pstibrany:
Waiting for #2088 now.

brancz:
I can't tell at a glance from your implementation, but wanted to share my thoughts: we're thinking about this on Thanos receive as well, and we're thinking about choosing a random, well-distributed value within the min block size for any TSDB instance, that way not only within one process handling multiple TSDBs has a well-distribution but also across processes.

pstibrany:
@brancz My implementation is currently very simple... idea is to disable automatic compactions, because we don't have any control over those, and simply run compaction once every defined interval on all open TSDBs (we have one per tenant). This way we can run only limited number of compactions at once.

This is going to change a bit, to tie it to the shipping process – to compact and then ship. Right now these two are separate in my implementation.

brancz:
> Just to be clear, this is not about compacting stored blocks, but HEAD compactions only.

Yes we're thinking about exactly the same.

pracucci:
> we're thinking about this on Thanos receive as well, and we're thinking about choosing a random, well-distributed value within the min block size for any TSDB instance, that way not only within one process handling multiple TSDBs has a well-distribution but also across processes.

What's the goal of spreading Head compactions across the time across different receiver replicas?

The use case we have on Cortex is different. In a multi-tenant system we end up compacting Head for multiple TSDB on the same ingester instance at the same time. We want to avoid 1K TSDB all compacting from head at the same time on the same node.

brancz:
[...]

PR #2026, opened 2020-01-23
Limit max number of concurrent tenants shipping and synchronizing TSDB blocks

pracucci:
**What this PR does**:
Currently there's no limit on the number of tenants concurrently synchronizing blocks in the querier and shipping blocks in ingester. This can lead to unpredictable high spikes in CPU and memory utilization when new blocks are compacted (remember that all blocks are compacted from the head nearly at the same time - see issue https://github.com/cortexproject/cortex/issues/2003).

In this PR I've introduced two flags to limit the concurrency:
- `-experimental.tsdb.ship-concurrency` to limit the number of tenants concurrently shipping blocks in the ingester
- `-experimental.tsdb.bucket-store.tenant-sync-concurrency` to limit the number of tenants concurrently synchronizing blocks in the querier

_I've also introduced `UserTSDB` and kept a similar design to the one proposed in #1958 in order to reduce the effort to merge the two PRs._

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ instructions below to upgrade your Postgres.
   * Renamed `cortex_querier_sync_seconds` metric to `cortex_querier_blocks_sync_seconds`
   * Track `cortex_querier_blocks_sync_seconds` metric for the initial sync too
   * Fixed race condition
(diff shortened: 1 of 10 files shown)


pracucci:
> I think that UserTSDB doesn't need to be exported, but maybe you have some plans with it

@pstibrany No real reason. I did it to simplify the merge with #1958 but shouldn't be too difficult if it's just 1 letter of difference, so I've unexported it.

PR #2027, opened 2020-01-23
Expose TSDB bucket usage metrics.

pstibrany:
This PR exposes TSDB bucket usage metrics. Each component that uses bucket store (compactor, ingester, querier) exposes metrics with its own "component" label (component="compactor", component="querier" or component="ingester").

~Metrics code comes from Thanos, objstore.go, metricBucket type. Unfortunately it wasn't possible to use `objstore.BucketWithMetrics` directly, because it tried to register same metrics for each component again and again. Hence the copy, and modifications.~

~One metric that original Thanos code tracked, and we don't, is "thanos_objstore_bucket_last_successful_upload_time" ("Second timestamp of the last successful upload to the bucket."). This seems to be quite useless in our setup with multiple tenants.~

Update: PR has been updated to use `objstore.BucketWithMetrics` directly, but pass registrerer with `cortex_<component>` prefix to it to avoid conflicts in single-binary mode. This means that all metrics written by Thanos bucket are exposed.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
PR has been updated to use `objstore.BucketWithMetrics` directly, but pass registrerer with `cortex_<component>` prefix to it to avoid conflicts in single-binary mode. This means that all metrics written by Thanos bucket are exposed.

This leads to metric like:

```
cortex_querier_thanos_objstore_bucket_operation_duration_seconds_sum{bucket="",operation="get_range"} 0.000147626
cortex_querier_thanos_objstore_bucket_operation_duration_seconds_count{bucket="",operation="get_range"} 1
cortex_ingester_thanos_objstore_bucket_operations_total{bucket="",operation="get_range"} 1
cortex_ingester_thanos_objstore_bucket_operations_total{bucket="",operation="iter"} 14
```

Thanos uses bucket name as a label, but we set this to constant empty string. It can safely be ignored.

pstibrany:
Rebased. Thanks for review Marco, and also for idea for using `WrapRegistererWithPrefix` – it's nicer than copying bunch of code from Thanos.

PR #2030, opened 2020-01-23
Remove error return from Cache.Stop()

bboreham:
All code paths return nil as the error, so we can simplify the code.

(Redis is the only one that looks as if it might, but if you go one level down it doesn't)

### DIFF
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -235,14 +235,13 @@ func (c *Memcached) Store(ctx context.Context, keys []string, bufs [][]byte) {
 }
 
 // Stop does nothing.
-func (c *Memcached) Stop() error {
+func (c *Memcached) Stop() {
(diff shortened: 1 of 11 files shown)


pkg/chunk/cache/memcached.go
```diff
@@ -235,14 +235,13 @@ func (c *Memcached) Store(ctx context.Context, keys []string, bufs [][]byte) {
 }
 
 // Stop does nothing.
-func (c *Memcached) Stop() error {
+func (c *Memcached) Stop() {
 	if c.inputCh == nil {
```
zendern: Stylistic thing but maybe simplify this and convert it to != nil and put the logic in there instead of the if and return and then the part we care about.

bboreham: The Go way is to return early and do the work outside of an `if`.

PR #2031, opened 2020-01-23
tsdb/querier: add tracing to block stores

thorfour:
**What this PR does**: Adds trace points to the user stores when running with the experimental TSDBs.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/block.go
+++ b/pkg/querier/block.go
@@ -10,6 +10,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/ingester/client"
 	"github.com/cortexproject/cortex/pkg/storage/tsdb"
 	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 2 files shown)


pkg/querier/block.go
```diff
@@ -47,6 +48,9 @@ func NewBlockQuerier(cfg tsdb.Config, logLevel logging.Level, registerer prometh
 
 // Get implements the ChunkStore interface. It makes a block query and converts the response into chunks
 func (b *BlockQuerier) Get(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]chunk.Chunk, error) {
+	log, ctx := spanlogger.New(ctx, "BlockQuerier.Get")
```
pracucci: Subsequent calls to the logger should be done like this:
`level.Error(log)`

Instead of:
`level.Error(util.Logger)`

I checked the rest of the changeset and this looks the only function also logging, so the LGTM should be just 1 line change away 😉

thorfour: Fixed!

PR #2039, opened 2020-01-27
add example to POST alertmanager config and rules

flbla:
Add an example to POST alertmanager config and rules.

### DIFF
--- a/docs/apis.md
+++ b/docs/apis.md
@@ -76,6 +76,17 @@ The following schema is used both when retrieving the current configs from the A
 - Normal Response Codes: NoContent(204)
 - Error Response Codes: Unauthorized(401), BadRequest(400)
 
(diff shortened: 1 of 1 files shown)


docs/apis.md
```diff
@@ -76,6 +76,19 @@ The following schema is used both when retrieving the current configs from the A
 - Normal Response Codes: NoContent(204)
 - Error Response Codes: Unauthorized(401), BadRequest(400)
 
+##### POST Example
```
pracucci: Using a `#####` title will break a bit the page layout when this documented will be rendered in our website. Do you mind replacing it with a text like `The POST request body is expected to be like the following example:`

flbla:
Thank you for your feedback @pracucci, I just signed it and replaced the text.

Issue #2038, opened 2020-01-27
frontend.cache-split-interval is unnecessary and adds misconfiguration pitfall.


owen-d:
## What
The [`frontend.cache-split-interval`](https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/results_cache.go#L41) is a ripe target for reducing configuration complexity in cortex. It's currently used to generate cache keys and is downstream of the SplitBy middleware. 

I believe it should instead inherit from the SplitBy configuration in order to guarantee they're aligned. Having two configurations that _should_ be equivalent only seems to leave room for configuration error. 

## Implementation path
Since this is a current configuration, we can mark it deprecated and when not specified defer to the SplitBy parameter.

PR #2040, opened 2020-01-27
Deprecate cache interval

owen-d:
## What
This PR deprecates `frontend.cache-split-interval` in favor of using `querier.split-queries-by-interval`, reducing the configuration burden. Additionally, these two configurations should be in alignment when both splitting and caching middlewares are enabled, which is now ensured. In order to not cause compatibility issues, it will continue to use the old `frontend.cache-split-interval` when specified, but log a deprecation message.

## Issue
closes https://github.com/cortexproject/cortex/issues/2038

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [CHANGE] Deprecated unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs.
+* [CHANGE] Deprecated unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs. #2040
 * [CHANGE] Removed remaining support for using denormalised tokens in the ring. If you're still running ingesters with denormalised tokens (Cortex 0.4 or earlier, with `-ingester.normalise-tokens=false`), such ingesters will now be completely invisible to distributors and need to be either switched to Cortex 0.6.0 or later, or be configured to use normalised tokens. #2034
 * [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893 
   * `--store.min-chunk-age` has been removed
@@ -19,6 +19,7 @@ Cortex 0.6.0 is the last version that can *read* denormalised tokens. Starting w
 
 Note that the ruler flags need to be changed in this upgrade. You're moving from a single node ruler to something that might need to be sharded.
 Further, if you're using the configs service, we've upgraded the migration library and this requires some manual intervention. See full instructions below to upgrade your PostgreSQL.
+
 
 * [CHANGE] The frontend component now does not cache results if it finds a `Cache-Control` header and if one of its values is `no-store`. #1974
 * [CHANGE] Flags changed with transition to upstream Prometheus rules manager:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [CHANGE] Deprecated unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs. #2040
+* [CHANGE] Removed unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs. #2040
 * [CHANGE] Removed remaining support for using denormalised tokens in the ring. If you're still running ingesters with denormalised tokens (Cortex 0.4 or earlier, with `-ingester.normalise-tokens=false`), such ingesters will now be completely invisible to distributors and need to be either switched to Cortex 0.6.0 or later, or be configured to use normalised tokens. #2034
 * [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893 
   * `--store.min-chunk-age` has been removed
@@ -19,7 +19,6 @@ Cortex 0.6.0 is the last version that can *read* denormalised tokens. Starting w
 
 Note that the ruler flags need to be changed in this upgrade. You're moving from a single node ruler to something that might need to be sharded.
 Further, if you're using the configs service, we've upgraded the migration library and this requires some manual intervention. See full instructions below to upgrade your PostgreSQL.
-
 
 * [CHANGE] The frontend component now does not cache results if it finds a `Cache-Control` header and if one of its values is `no-store`. #1974
 * [CHANGE] Flags changed with transition to upstream Prometheus rules manager:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [CHANGE] Removed unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs. #2040
+* [CHANGE] Removed unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs. Starting from now, `-querier.cache-results` may only be enabled in conjunction with `-querier.split-queries-by-interval` (previously the cache interval default was `24h` so if you want to preserve the same behaviour you should set `-querier.split-queries-by-interval=24h`). #2040
 * [CHANGE] Removed remaining support for using denormalised tokens in the ring. If you're still running ingesters with denormalised tokens (Cortex 0.4 or earlier, with `-ingester.normalise-tokens=false`), such ingesters will now be completely invisible to distributors and need to be either switched to Cortex 0.6.0 or later, or be configured to use normalised tokens. #2034
 * [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893 
   * `--store.min-chunk-age` has been removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Deprecated unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs.
 * [CHANGE] Removed remaining support for using denormalised tokens in the ring. If you're still running ingesters with denormalised tokens (Cortex 0.4 or earlier, with `-ingester.normalise-tokens=false`), such ingesters will now be completely invisible to distributors and need to be either switched to Cortex 0.6.0 or later, or be configured to use normalised tokens. #2034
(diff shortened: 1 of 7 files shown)


pkg/querier/queryrange/results_cache.go
```diff
@@ -37,8 +37,14 @@ type ResultsCacheConfig struct {
 // RegisterFlags registers flags.
 func (cfg *ResultsCacheConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.CacheConfig.RegisterFlagsWithPrefix("frontend.", "", f)
+
 	f.DurationVar(&cfg.MaxCacheFreshness, "frontend.max-cache-freshness", 1*time.Minute, "Most recent allowed cacheable result, to prevent caching very recent results that might still be in flux.")
-	f.DurationVar(&cfg.SplitInterval, "frontend.cache-split-interval", 24*time.Hour, "The maximum interval expected for each request, results will be cached per single interval.")
+	f.DurationVar(&cfg.SplitInterval, "frontend.cache-split-interval", 0, "Deprecated: The maximum interval expected for each request, results will be cached per single interval. This flag has been deprecated and this behavior is now determined by querier.split-queries-by-interval")
```
pracucci: I would suggest to remove `SplitInterval` at all. The change introduced in this PR is a breaking change anyway, and removing this config option would allow to have a cleaner `config-file-reference.md` as well as signal the change to an user which upgrades without reading the `CHANGELOG.md` (because the config parsing will fail).

What's your take?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ If you are running with a high `-ruler.num-workers` and if you're not able to ex
 Further, if you're using the configs service, we've upgraded the migration library and this requires some manual intervention. See full
 instructions below to upgrade your Postgres.
 
+* [CHANGE] Remove unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs. #2040
```
pracucci: `frontend.cache-split-interval` default was `24h`, while `querier.split-queries-by-interval` default is `0` (which means disabled).

pkg/querier/queryrange/results_cache.go
```diff
@@ -55,6 +61,18 @@ func (e ExtractorFunc) Extract(start, end int64, from Response) Response {
 	return e(start, end, from)
 }
 
+// CacheSplitter generates cache keys
+type CacheSplitter interface {
```
pracucci: The introduction of this `CacheSplitter` looks an over complication to me. If we completely get rid of `SplitInterval`, is there any other reason to keep it?

CHANGELOG.md
owen-d: That logic has actually been moved [here](https://github.com/cortexproject/cortex/pull/2040/files#diff-be5bce86bb962a94af90c800ee891d85R97-R109). This PR takes a similar approach to how the deprecation of `split-by-day` was implemented (in favor of `split-by-interval`) [here](https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/roundtrip.go#L99-L105).

pkg/querier/queryrange/results_cache.go
owen-d: I opted to not have this be a breaking change, but instead to provide a deprecation warning but still use the old config if specified.

pkg/querier/queryrange/results_cache.go
owen-d: Because this is not implemented as a breaking change, we have multiple types which fulfill this interface and are chosen based on this criteria: https://github.com/cortexproject/cortex/pull/2040/files#diff-be5bce86bb962a94af90c800ee891d85R97-R109

Furthermore it's a convenient place to have an interface for downstream consumers. I know we'd like to parameterize this based on tenant overrides in loki, for example.

CHANGELOG.md
```diff
@@ -16,6 +17,7 @@ Cortex 0.6.0 is the last version that can *read* denormalised tokens. Starting w
 Note that the ruler flags need to be changed in this upgrade. You're moving from a single node ruler to something that might need to be sharded.
 Further, if you're using the configs service, we've upgraded the migration library and this requires some manual intervention. See full instructions below to upgrade your PostgreSQL.
 
+
```
pracucci: Remove me 😉

pkg/querier/queryrange/roundtrip.go
```diff
@@ -89,6 +94,19 @@ func MergeMiddlewares(middleware ...Middleware) Middleware {
...
+	if cfg.ResultsCacheConfig.SplitInterval > 0 {
+		level.Warn(log).Log("msg", "flag frontend.cache-split-interval (or config cache_split_interval) is deprecated, use querier.split-queries-by-interval instead.")
+		return cfg.ResultsCacheConfig
+	}
+	if cfg.SplitQueriesByInterval > 0 {
+		return cfg
+	}
+
+	// by default use 24 hour segments
+	return constSplitter(24 * time.Hour)
```
pracucci: This looks somewhat problematic to me because we're hiding a default setting (24h) down here instead of the CLI flag default. Moreover, removing the default value from the CLI flag has caused the default value of 24h to disappear from `config-file-reference.md`.

I still believe that trying to not introduce a breaking change is making things over complicated. What if we:
1. Completely remove `SplitInterval`
2. Add a `Validate()` to `queryrange.Config` to ensure that if `CacheResults` is `true` then `SplitQueriesByInterval` is set a well (`> 0`)
3. Keep the default `CacheResults: false` and `SplitQueriesByInterval: 0`)
I still believe that we're getting things over complicated because

pkg/querier/queryrange/results_cache.go
cyriltovena: Yes this is not a breaking change, yet.

pkg/querier/queryrange/results_cache.go
```diff
@@ -55,6 +57,12 @@ func (e ExtractorFunc) Extract(start, end int64, from Response) Response {
 	return e(start, end, from)
 }
 
+// CacheSplitter generates cache keys. This is a useful interface for downstream
+// consumers who wish to impl their own strategies.
```
pracucci: `impl` > `implement`

pkg/querier/queryrange/roundtrip.go
```diff
@@ -46,12 +47,28 @@ type Config struct {
...
+		cfg.SplitQueriesByInterval = day
+	}
+
+	if cfg.CacheResults && cfg.SplitQueriesByInterval <= 0 {
+		return errors.New("querier.cache-results may only be enabled in conjunction with querier.split-queries-by-interval. Please set the latter")
+	}
+	return nil
+}
+
+// GenerateCacheKey impls CacheSplitter
```
pracucci: `impls` > `implements`

pkg/querier/queryrange/roundtrip.go
```diff
@@ -46,12 +47,28 @@ type Config struct {
...
+func (cfg Config) GenerateCacheKey(userID string, r Request) string {
```
pracucci: This is not a common pattern (having business logic functions to the Config). I personally did it some time ago and has been "rejected" by other maintainers, so I would get stick to it.

I would suggest to create `type cacheSplitter struct` taking in input the `cfg` and implementing `GenerateCacheKey()`. I would also remove the `generateKey()` function and move its logic into `cacheSplitter.GenerateCacheKey()` to simplify code reading.

The other place where we currently call `generateKey()` are its own tests, and will be easy to migrate that test to `cacheSplitter` test.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Removed unnecessary `frontend.cache-split-interval` in favor of `querier.split-queries-by-interval` both to reduce configuration complexity and guarantee alignment of these two configs. #2040
```
pracucci: I would also mention that ``Starting from now, `-querier.cache-results` may only be enabled in conjunction with `-querier.split-queries-by-interval` (previously the cache interval default was `24h` so if you wanna preserve the same behaviour you should set `-querier.split-queries-by-interval=24h`). ``

pkg/querier/queryrange/results_cache_test.go
```diff
@@ -307,7 +308,7 @@ func TestResultsCacheRecent(t *testing.T) {
 	var cfg ResultsCacheConfig
 	flagext.DefaultValues(&cfg)
 	cfg.CacheConfig.Cache = cache.NewMockCache()
-	rcm, _, err := NewResultsCacheMiddleware(log.NewNopLogger(), cfg, fakeLimits{}, PrometheusCodec, PrometheusResponseExtractor)
+	rcm, _, err := NewResultsCacheMiddleware(log.NewNopLogger(), cfg, constSplitter(24*time.Hour), fakeLimits{}, PrometheusCodec, PrometheusResponseExtractor)
```
jtlisi: You can replace `24*time.Hour` with the `day` constant in this instance.

pkg/querier/queryrange/results_cache_test.go
```diff
@@ -266,14 +266,15 @@ func (fakeLimits) MaxQueryParallelism(string) int {
...
 	rcm, _, err := NewResultsCacheMiddleware(
 		log.NewNopLogger(),
-		ResultsCacheConfig{
-			CacheConfig: cache.Config{
-				Cache: cache.NewMockCache(),
-			},
-			SplitInterval: 24 * time.Hour,
-		},
+		cfg,
+		constSplitter(24*time.Hour),
```
jtlisi: Use the day const?

pkg/querier/queryrange/results_cache_test.go
```diff
@@ -334,14 +335,15 @@ func TestResultsCacheRecent(t *testing.T) {
...
 	rm, _, err := NewResultsCacheMiddleware(
 		log.NewNopLogger(),
-		ResultsCacheConfig{
-			CacheConfig: cache.Config{
-				Cache: cache.NewMockCache(),
-			},
-			SplitInterval: 24 * time.Hour,
-		},
+		cfg,
+		constSplitter(24*time.Hour),
```
jtlisi: `day`

pkg/querier/queryrange/roundtrip.go
```diff
@@ -46,12 +47,23 @@ type Config struct {
...
-	f.DurationVar(&cfg.SplitQueriesByInterval, "querier.split-queries-by-interval", 0, "Split queries by an interval and execute in parallel, 0 disables it. You should use an a multiple of 24 hours (same as the storage bucketing scheme), to avoid queriers downloading and processing the same chunks.")
+	f.DurationVar(&cfg.SplitQueriesByInterval, "querier.split-queries-by-interval", 0, "Split queries by an interval and execute in parallel, 0 disables it. You should use an a multiple of 24 hours (same as the storage bucketing scheme), to avoid queriers downloading and processing the same chunks. This also determines how cache keys are chosen when result caching is enabled")
 	f.BoolVar(&cfg.AlignQueriesWithStep, "querier.align-querier-with-step", false, "Mutate incoming queries to align their start and end with their step.")
 	f.BoolVar(&cfg.CacheResults, "querier.cache-results", false, "Cache query results.")
 	cfg.ResultsCacheConfig.RegisterFlags(f)
 }
 
+func (cfg *Config) Validate() error {
+	if cfg.SplitQueriesByDay {
+		cfg.SplitQueriesByInterval = day
```
jtlisi: Can you add a log warning if `cfg.SplitQueriesByInterval` is not set to the default value? You could also suggest in the message to stop using `SplitQueriesByDay` in favor of setting `SplitQueriesByInterval` to `24h` since we are deprecating it.

pkg/querier/queryrange/results_cache_test.go
```diff
@@ -398,7 +398,7 @@ func Test_generateKey(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(fmt.Sprintf("%s - %s", tt.name, tt.interval), func(t *testing.T) {
```
pracucci: Remember to rename this test func `TestConstSplitter_GenerateCacheKey`

pkg/querier/queryrange/roundtrip.go
owen-d: This warning was already logged during initialization, I've just moved the override here. To address your feedback, I've now added a logger as an argument to our validation functions. One consequence is we're now initializing the logger based on the server config before the configs are validated [here](https://github.com/cortexproject/cortex/pull/2040/files#diff-cf2ba502f0bdf282c97f46997569748cR76-R79). It looks reasonable as the log levels are set with defaults, but let me know if I should revert/do something else.

PR #2043, opened 2020-01-28
Ensure TSDB blocks transfer work even if a spurious file is in the TSDB blocks directory

pracucci:
**What this PR does**:
In the local development environment may happen that a spurious file is in the TSDB blocks directory. This shouldn't happen in production, but if that happens for any reason I believe it's safer to just skip it instead of having the ingester transfer breaking.

In this PR:
- Ensure TSDB blocks transfer work even if a spurious file is in the TSDB blocks directory
- Wrap more errors that could occur during the transfer, in order to easily identify issues

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/transfer.go
+++ b/pkg/ingester/transfer.go
@@ -225,7 +225,7 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 
 		tmpDir, err := ioutil.TempDir("", "tsdb_xfer")
 		if err != nil {
-			return errors.Wrap(err, "unable to create temporary directory storing transferred TSDB blocks")
+			return errors.Wrap(err, "unable to create temporary directory to store transferred TSDB blocks")
 		}
 		defer os.RemoveAll(tmpDir)
 
@@ -279,7 +279,7 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 			if !ok {
 				file, err = createfile(f)
 				if err != nil {
-					return errors.Wrap(err, "unable to create file storing incoming TSDB block received from LEAVING ingester")
+					return errors.Wrap(err, fmt.Sprintf("unable to create the file %s to store incoming TSDB block", f))
 				}
 				filesXfer++
 				files[f.Filename] = file
@@ -629,10 +629,11 @@ func unshippedBlocks(dir string) (map[string][]string, error) {
 		// Ensure the user dir is actually a directory. There may be spurious files
 		// in the storage, especially when using Minio in the local development environment.
 		if stat, err := os.Stat(userDir); err == nil && !stat.IsDir() {
+			level.Warn(util.Logger).Log("msg", "skipping entry while transferring TSDB blocks because not a directory", "path", userDir)
 			continue
 		}
 
-		// Seed the map with the userID to ensure we xfer the WAL, even if all blocks are shipped.
+		// Seed the map with the userID to ensure we transfer the WAL, even if all blocks are shipped.
 		blocks[userID] = []string{}
 
 		blockIDs, err := ioutil.ReadDir(userDir)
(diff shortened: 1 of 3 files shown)


pkg/ingester/transfer.go
```diff
@@ -618,20 +618,29 @@ func (i *Ingester) findTargetIngester(ctx context.Context) (*ring.IngesterDesc,
...
 	blocks := make(map[string][]string, len(userIDs))
 	for _, user := range userIDs {
 		userID := user.Name()
-		blocks[userID] = []string{} // seed the map with the userID to ensure we xfer the WAL, even if all blocks are shipped
+		userDir := filepath.Join(dir, userID)
+
+		// Ensure the user dir is actually a directory. There may be spurious files
+		// in the storage, especially when using Minio in the local development environment.
+		if stat, err := os.Stat(userDir); err == nil && !stat.IsDir() {
+			continue
```
pstibrany: I would add a warning log message, just in case this happens in production.

pkg/ingester/transfer.go
```diff
@@ -618,20 +618,29 @@ func (i *Ingester) findTargetIngester(ctx context.Context) (*ring.IngesterDesc,
...
+		}
+
+		// Seed the map with the userID to ensure we xfer the WAL, even if all blocks are shipped.
```
pstibrany: `xfer`?

pracucci:
Thanks @pstibrany for your valuable feedback. May you check it again, please?

pkg/ingester/transfer.go
```diff
@@ -279,7 +279,7 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 			if !ok {
 				file, err = createfile(f)
 				if err != nil {
-					return err
+					return errors.Wrap(err, fmt.Sprintf("unable to create the file %s to store incoming TSDB block", f))
```
pstibrany: You can use `errors.Wrapf`, also I would remove 'the', as we name the file now.

pracucci: Done, thanks!

pkg/ingester/wal_test.go
```diff
@@ -2,19 +2,12 @@ package ingester
...
-	"os"
 	"testing"
 	"time"
 
-	"github.com/cortexproject/cortex/pkg/util"
-	"github.com/go-kit/kit/log"
 	"github.com/stretchr/testify/require"
 )
 
-func init() {
```
jtlisi: Just curious why you are removing this? Also I've noticed this a useful pattern in a lot of our tests and was curious if you had any thought's on initializing a logger during tests.

pracucci: I've re-introduced it so that we can merge this PR while discussing about the stdout logger in tests.

pkg/ingester/transfer.go
```diff
@@ -618,20 +618,30 @@ func (i *Ingester) findTargetIngester(ctx context.Context) (*ring.IngesterDesc,
...
 
 	blocks := make(map[string][]string, len(userIDs))
 	for _, user := range userIDs {
 		userID := user.Name()
-		blocks[userID] = []string{} // seed the map with the userID to ensure we xfer the WAL, even if all blocks are shipped
+		userDir := filepath.Join(dir, userID)
+
+		// Ensure the user dir is actually a directory. There may be spurious files
+		// in the storage, especially when using Minio in the local development environment.
+		if stat, err := os.Stat(userDir); err == nil && !stat.IsDir() {
```
codesome: What if `err != nil`?

pstibrany: In my opinion, if err is returned here, it will also be returned later, so I didn't flag it.

pracucci: Correct. This is just to skip non dirs, but if any error arise, the next `ioutil.ReadDir()` will fail as well.

PR #2047, opened 2020-01-28
Return error instead of os.Exit when initializing Alertmanager

jtlisi:
**What this PR does**:

Instead of calling `os.Exit` the Alertmanager will return an error when it is being initialized.

### DIFF
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
 	"github.com/prometheus/alertmanager/cluster"
 	amconfig "github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -188,8 +189,7 @@ func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, cfgCfg confi
 			cluster.DefaultProbeInterval,
 		)
 		if err != nil {
-			level.Error(util.Logger).Log("msg", "unable to initialize gossip mesh", "err", err)
-			os.Exit(1)
+			return nil, errors.Wrap(err, "unable to initialize gossip mesh")
```
pracucci: This has removed the log, which is OK as far as it's logged somewhere in the parent. However, looking at `initAlertmanager()` (`modules.go`), in case `NewMultitenantAlertmanager()` returns error we do return "success" instead of the error. I think it's a bug. May you fix it, please?

jtlisi: Wow, quite the bug haha. I’ll fix it

PR #2049, opened 2020-01-29
support for registering custom index clients, added new methods to object stores

sandeepsukhani:
**What this PR does**:
To avoid changes required for Loki storage getting into Cortex, this PR allows using custom index clients. It also adds new methods for object stores to work on objects(io.Reader) instead of just chunks.

The effort here is to avoid merging #1533 which is mostly required for Loki. For being able to reuse boltdb object client in Loki for flushing boltdb files to object stores, I have done some changes there to make some of the things public.

We can adopt a similar approach wherever Loki specific changes are required.


pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -45,71 +52,111 @@ func NewGCSObjectClient(ctx context.Context, cfg GCSConfig, schemaCfg chunk.Sche
...
+	readCloser, err := s.GetObject(ctx, input.ExternalKey())
+	if err != nil {
+		return chunk.Chunk{}, err
+	}
+
+	defer readCloser.Close()
+
+	buf, err := ioutil.ReadAll(readCloser)
+	if err != nil {
+		return chunk.Chunk{}, err
```
pracucci: Before we were doing `errors.WithStack(err)` instead of returning `err`

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -45,71 +52,111 @@ func NewGCSObjectClient(ctx context.Context, cfg GCSConfig, schemaCfg chunk.Sche
...
 
-	if err := input.Decode(decodeContext, buf); err != nil {
-		return chunk.Chunk{}, err
+	return nil
+}
+
+// List objects from the store
+func (s *GCSObjectClient) List(ctx context.Context, prefix string) (map[string]time.Time, error) {
+	objectKeysWithMtime := map[string]time.Time{}
+	prefixWithSep := prefix + "/"
```
pracucci: Don't do it. Should be the client responsability adding this prefix.

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -45,71 +52,111 @@ func NewGCSObjectClient(ctx context.Context, cfg GCSConfig, schemaCfg chunk.Sche
...
+		attr, err := iter.Next()
+		if err != nil {
+			if err == iterator.Done {
+				break
+			} else {
+				return nil, err
+			}
+		}
+
+		if attr.Name == prefixWithSep {
```
pracucci: Could you explain this, please? If you confirm it's what we want, please add a comment to give some details about why we need it.

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -45,71 +52,111 @@ func NewGCSObjectClient(ctx context.Context, cfg GCSConfig, schemaCfg chunk.Sche
...
+}
+
+// List objects from the store
+func (s *GCSObjectClient) List(ctx context.Context, prefix string) (map[string]time.Time, error) {
+	objectKeysWithMtime := map[string]time.Time{}
+	prefixWithSep := prefix + "/"
+
+	iter := s.bucket.Objects(ctx, &storage.Query{Prefix: prefix})
+	for {
+		attr, err := iter.Next()
```
pracucci: Before each iteration you should check if the context has been cancelled in the meanwhile. You can simply do:

```
if ctx.Err() != nil {
  return nil, ctx.Err()
}
```

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -45,71 +52,111 @@ func NewGCSObjectClient(ctx context.Context, cfg GCSConfig, schemaCfg chunk.Sche
...
+		if err != nil {
+			if err == iterator.Done {
+				break
+			} else {
+				return nil, err
+			}
+		}
```
pracucci: This could be simplified into:

```
if err == iterator.Done {
			break
		}
		if err != nil {
			return nil, err
		}
```

pkg/chunk/local/boltdb_index_client.go
```diff
@@ -117,41 +119,45 @@ func (b *boltIndexClient) Stop() {
...
 
-func (b *boltIndexClient) NewWriteBatch() chunk.WriteBatch {
+func (b *BoltIndexClient) NewWriteBatch() chunk.WriteBatch {
 	return &boltWriteBatch{
 		tables: map[string]map[string][]byte{},
 	}
 }
 
-func (b *boltIndexClient) getDB(name string) (*bbolt.DB, error) {
+func (b *BoltIndexClient) GetDB(name string, operation int) (*bbolt.DB, error) {
```
pracucci: The refactoring here looks going beyond exporting `BoltIndexClient`. Why do we need such changes? Can you share more details, please?

pkg/chunk/local/boltdb_index_client.go
```diff
@@ -23,6 +22,9 @@ var bucketName = []byte("index")
 const (
 	separator      = "\000"
 	dbReloadPeriod = 10 * time.Minute
+
+	DBOperationRead = iota
+	DBOperationWrite
```
pracucci: Could you explain why we may want this change into Cortex, instead of having Loki implementing its own `BoltIndexClient` with the feature it needs, please?

pkg/chunk/local/fs_object_client.go
```diff
@@ -81,6 +96,58 @@ func (f *FSObjectClient) getChunk(_ context.Context, decodeContext *chunk.Decode
...
+
+// Put object into the store
+func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object io.ReadSeeker) error {
+	fullPath := path.Join(f.cfg.Directory, objectKey)
+	err := util.EnsureDirectory(path.Dir(fullPath))
+	if err != nil {
+		return err
+	}
+
+	fl, err := os.OpenFile(fullPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
```
pracucci: Note to myself: this is a safe drop in replacement for `ioutil.WriteFile()`

pkg/chunk/local/fs_object_client.go
```diff
@@ -81,6 +96,58 @@ func (f *FSObjectClient) getChunk(_ context.Context, decodeContext *chunk.Decode
...
+
+	_, err := os.Stat(folderPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	filesInfo, err := ioutil.ReadDir(folderPath)
```
pracucci: This is not recursive. Just to confirm: is it what we want? I would document it in the function comment.

pkg/chunk/storage/factory.go
```diff
@@ -25,6 +25,8 @@ const (
 	StorageEngineTSDB   = "tsdb"
 )
 
+type IndexClientFactoryFunc func() (chunk.IndexClient, error)
```
pracucci: Don't we need to pass the config here?

pkg/chunk/storage/factory.go
```diff
@@ -73,7 +75,7 @@ func (cfg *Config) Validate() error {
 }
 
 // NewStore makes the storage clients based on the configuration.
-func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConfig, limits StoreLimits) (chunk.Store, error) {
+func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConfig, limits StoreLimits, customIndexClients map[string]IndexClientFactoryFunc) (chunk.Store, error) {
```
pracucci: I'm not a big fan of passing `customIndexClients` here. An alternative approach may be:

1. Define a global variable `customIndexClients`
2. Add a function `RegisterIndexClient(name string, factory IndexClientFactoryFunc)` which stores the index client to `customIndexClients`
3. In `NewIndexClient()`, before `switch name {` we do check if `customIndexClients` contains it. If so, it will take precedence over the defaults.

pkg/chunk/local/boltdb_index_client.go
sandeepsukhani: This is related to flushing boltdb feature. So the idea is to embed boltdb client from cortex into new client that I am creating in Loki and I will call exported QueryDb method for each db that I need to query i.e ones downloaded from the archive and the local one.

pkg/chunk/local/boltdb_index_client.go
sandeepsukhani: Without this change, boltdb client creates a new boltdb file even for just write requests which end up getting flushed to archiver. This also helps cortex not create boltdb files unnecessarily.

pkg/chunk/storage/factory.go
sandeepsukhani: no, the config and creation of index client(which means passing the config) would both be managed by the caller. They would just pass a closure basically.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -160,3 +141,64 @@ func (a s3ObjectClient) bucketFromKey(key string) string {
...
+	prefixWithSep := prefix + "/"
+
+	for i := range a.bucketNames {
+		err := instrument.CollectedRequest(ctx, "S3.List", s3RequestDuration, instrument.ErrorCode, func(ctx context.Context) error {
+			output, err := a.S3.ListObjectsV2WithContext(ctx, &s3.ListObjectsV2Input{Bucket: &a.bucketNames[i], Prefix: &prefix})
+			if err != nil {
+				return err
+			}
+
+			for i := range output.Contents {
```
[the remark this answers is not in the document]
pracucci: The Thanos implementation uses the Minio client, which internally does paginate across results. It may be a good reference to checkout to see how it works, even if in our case we don't need to work with channels like the Minio client does.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -160,3 +162,80 @@ func (a s3ObjectClient) bucketFromKey(key string) string {
...
+				Prefix: aws.String(prefix),
+			}
+
+			for {
+				output, err := a.S3.ListObjectsV2WithContext(ctx, &input)
+				if err != nil {
+					return err
+				}
+
+				for i := range output.Contents {
```
pracucci: You could avoid referencing `*output.Contents[i]` below using `for _, content := range output.Contents`, right?

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -160,3 +162,80 @@ func (a s3ObjectClient) bucketFromKey(key string) string {
...
+					storageObjects = append(storageObjects, chunk.StorageObject{
+						Key:        strings.TrimPrefix(*output.Contents[i].Key, prefix),
```
pracucci: We shouldn't trim the prefix here. The object key should be the complete key name (including the prefix). If the caller needs it without the prefix, it should be the caller responsability to remove it.

pkg/chunk/azure/blob_storage_client.go
```diff
@@ -126,18 +158,68 @@ func (b *BlobStorage) getBlobURL(blobID string) (azblob.BlockBlobURL, error) {
...
 			Policy:        azblob.RetryPolicyExponential,
 			MaxTries:      (int32)(b.cfg.MaxRetries),
 			TryTimeout:    b.cfg.RequestTimeout,
 			RetryDelay:    b.cfg.MinRetryDelay,
 			MaxRetryDelay: b.cfg.MaxRetryDelay,
 		},
-	})), nil
+	}), nil
+}
+func (b *BlobStorage) List(ctx context.Context, prefix string) ([]chunk.StorageObject, error) {
```
pracucci: Please add a newline to separate this function from the previous one

pkg/chunk/azure/blob_storage_client.go
```diff
@@ -147,18 +158,68 @@ func (b *BlobStorage) getBlobURL(blobID string) (azblob.BlockBlobURL, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		marker = listBlob.NextMarker
+
+		// Process the blobs returned in this result segment (if the segment is empty, the loop body won't execute)
+		for _, blobInfo := range listBlob.Segment.BlobItems {
+			storageObjects = append(storageObjects, chunk.StorageObject{
+				Key:        strings.TrimPrefix(blobInfo.Name, prefix),
```
pracucci: Following up my previous comment, I would not trim here as well.

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -113,3 +107,63 @@ func (s *gcsObjectClient) getChunk(ctx context.Context, decodeContext *chunk.Dec
...
+				break
+			}
+			return nil, err
+		}
+
+		if attr.Name == prefix {
+			continue
+		}
+		storageObjects = append(storageObjects, chunk.StorageObject{
+			Key:        strings.TrimPrefix(attr.Name, prefix),
```
pracucci: I would not trim here as well.

pkg/chunk/local/fs_object_client.go
```diff
@@ -81,6 +96,61 @@ func (f *FSObjectClient) getChunk(_ context.Context, decodeContext *chunk.Decode
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, fileInfo := range filesInfo {
+		if fileInfo.IsDir() {
+			continue
+		}
+		storageObjects = append(storageObjects, chunk.StorageObject{
+			Key:        fileInfo.Name(),
```
pracucci: If we do not trim in other storages, to keep compatibility here we should append the prefix.

pkg/chunk/storage/factory.go
```diff
@@ -116,6 +124,10 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 
 // NewIndexClient makes a new index client of the desired type.
 func NewIndexClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.IndexClient, error) {
+	if factory, isOK := customIndexClients[name]; isOK {
```
pracucci: It's a convention calling it `ok`, so I would rename `isOK` to `ok`.

pracucci: Please add a unit test on this. You will need a `UnregisterIndexClient` function to call it in a `defer`within the test, so that we cleanup the global state before the test exits.

pkg/chunk/local/boltdb_index_client.go
```diff
@@ -117,41 +119,45 @@ func (b *boltIndexClient) Stop() {
...
+		}
 
-	// Open the database.
-	// Set Timeout to avoid obtaining file lock wait indefinitely.
-	db, err := bbolt.Open(path.Join(b.cfg.Directory, name), 0666, &bbolt.Options{Timeout: 5 * time.Second})
-	if err != nil {
-		return nil, err
+		// Open the database.
+		// Set Timeout to avoid obtaining file lock wait indefinitely.
+		db, err := OpenBoltdbFile(path.Join(b.cfg.Directory, name))
```
pracucci: Could you explain the need of moving it to a dedicated function `OpenBoltdbFile()`? If it's required for an easier integration with Loki, could you explain how it will work, please?

pkg/chunk/local/boltdb_index_client.go
```diff
@@ -117,41 +119,45 @@ func (b *boltIndexClient) Stop() {
...
 		return db, nil
 	}
 
-	b.dbsMtx.Lock()
-	defer b.dbsMtx.Unlock()
-	db, ok = b.dbs[name]
-	if ok {
-		return db, nil
-	}
+	if _, err := os.Stat(path.Join(b.cfg.Directory, name)); err == nil || operation == DBOperationWrite {
```
pracucci: If `err != nil`, it's ignored. I also find this `if` statement quite unclear. I would suggest to improve it and add a comment to explain that if the DB is in read-only mode and the DB doesn't exist yet, then we should not create a new one. An example of an improved logic may be reverse the `if` logic:

```
if operation == DBOperationRead {
  if _, err := os.Stat(path.Join(b.cfg.Directory, name)); err != nil {
    return nil, <an error here>
  }
}
```

pkg/chunk/local/boltdb_index_client.go
```diff
@@ -117,41 +119,45 @@ func (b *boltIndexClient) Stop() {
...
+			return nil, err
+		}
+
+		b.dbs[name] = db
+		return db, nil
 	}
 
-	b.dbs[name] = db
-	return db, nil
+	return nil, nil
```
pracucci: We should never return `nil, nil`. Either a DB or an error (an error could be something like "unsupported operation by the configured mode"). The rest of the code doesn't currently handles the case both the DB and thr error are `nil`, and I belive it's correct they don't handle it.

sandeepsukhani: for read operation it does check whether db is there or not for querying https://github.com/cortexproject/cortex/pull/2049/files/6cab326d3ae40257ce2a40578b9e2cbcc0afce8c#diff-f0b676dad4198715e462b2faa06cfaa8R195

I don't feel right throwing an error because db does not exist and we don't want to create it for a read operation because it would anyways return no results.

pracucci: > I don't feel right throwing an error

My perspective: if a BoltDB client instance is not configured for read operations, no read operation should occur on it. If it occurs, it's an error, and we should return an error. This also doesn't force us to change all other functions logic to handle the case `GetDB()` returns `nil, nil`.

sandeepsukhani:
Tested GetObject/PutObject/List operation locally with all the object stores that support them, it works fine as expected.

pkg/chunk/local/boltdb_index_client_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/bbolt"
+)
+
+var (
+	testKey   = []byte("test-key")
+	testValue = []byte("test-value")
+)
+
+func setupDb(t *testing.T, boltdbIndexClient *BoltIndexClient, dbname string) {
```
pracucci: `setupDb` > `setupDB`

pkg/chunk/local/boltdb_index_client_test.go
```diff
@@ -0,0 +1,117 @@
...
+	_ = droppedDb.View(func(tx *bbolt.Tx) error {
+		b := tx.Bucket(bucketName)
+		valueFromDb = b.Get(testKey)
+		return nil
+	})
+	require.Equal(t, testValue, valueFromDb, "should match value from db")
+
+	boltdbIndexClient.reload()
+
+	require.Equal(t, 1, len(boltdbIndexClient.dbs), "There should be 1 boltdb open")
```
pracucci: May we also check if `boltdbIndexClient.GetDB(testDb1, DBOperationRead)` returns `ErrUnexistentBoltDB`?

_Same comment applies to the calls to `<bolt db>.Stop()` in the tests you've added_

sandeepsukhani: I am sorry if I didn't understand it, is this what you are looking for?
https://github.com/cortexproject/cortex/pull/2049/files/273032a393eb1a6cc815e5387a1bd2259f494dc4#diff-4f420894f40d4e2e1a9b491716cc37dfR96

pracucci: I mean trying to `GetDB(testDb1)` and then `require.Equal(t, ErrUnexistentBoltDB, err)`

sandeepsukhani:
> LGTM
> 
> However, I noticed the `GetChunks`, `getChunks`, `PutChunks`, and `putChunks` methods are basically identical for all of the object store backends (S3, GCS, Azure). I think we should consider refactoring.
> 
> For instance:
> 
> * Rename `ObjectClient` to `ChunkClient` as is.
> * Create a new `ObjectClient` interface that contains the the `PutObject`, `GetObject` and `List` functions
> * Create a generic implementation of the `ChunkClient` that would be built on top of the `ObjectClient`
> 
> @sandeepsukhani and @pracucci WDYT?

Sounds good to me but I would suggest merging this and I will do another PR for it since I need code from this PR for Delete Series feature that I am working on and would be blocked on it until this gets reviewed again and merged.

@pracucci what do you think?

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -1,23 +1,25 @@
...
 	"github.com/pkg/errors"
+	"google.golang.org/api/iterator"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
 	"github.com/cortexproject/cortex/pkg/chunk/util"
 )
 
-type gcsObjectClient struct {
-	cfg       GCSConfig
-	schemaCfg chunk.SchemaConfig
```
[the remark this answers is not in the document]
sandeepsukhani: Yes, also I do not feel schema config would be required in object stores since there is no schema there.

pracucci:
> Sounds good to me but I would suggest merging this and I will do another PR for it since I need code from this PR for Delete Series feature that I am working on and would be blocked on it until this gets reviewed again and merged.

> @pracucci what do you think?

I also suggest to NOT do it in this PR. The refactoring proposed would further complicate the diff. From my perspective the duplicated code is not that much and, personally, I don't see the effort of the refactoring valuable. In any case I would suggest to not bundle it in this PR.

pracucci:
> I also suggest to NOT do it in this PR. The refactoring proposed would further complicate the diff. From my perspective the duplicated code is not that much and, personally, I don't see the effort of the refactoring valuable. In any case I would suggest to not bundle it in this PR.

After a closer look, I still believe it's better to do it in a different PR, but actually there may be some value in refactoring.

I'm going to merge this PR, so that Sandeep can move on with Loki (this PR is necessary to allow Loki to extend the BoltDB index client and add the "archiver" support), while would be great if @sandeepsukhani considers the suggested refactoring for a not-urgent following PR.

PR #2051, opened 2020-01-29
Do not log to stdout in ingester tests

### LINKED ISSUE #2048: Log to stdout in unit tests
I've noticed that the Ingester WAL test has introduced an override to log to stdout. Given it's done in a `init()` function, this applies to any `pkg/ingester` test:

https://github.com/cortexproject/cortex/blob/032f3a83f5f057b913a2ac0c4fdfe01e5aceea57/pkg/ingester/wal_test.go#L14-L16

I personally find it both useful and noisy. While developing and running unit tests, I frequently find myself commenting these lines because add too much noise to the tests result output, while sometimes they're useful.

I'm wondering if we're OK keeping such logs always enabled by default or being able to conditionally enable them would be better.

/cc @codesome @jtlisi

pracucci:
**What this PR does**:
As per the #2048 discussion, I'm removing the override to log to stdout in `pkg/ingester` tests. If required, we'll figure out a better way to conditionally turn on the stdout logger in unit tests.

Fixes #2048

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/wal_test.go
+++ b/pkg/ingester/wal_test.go
@@ -2,18 +2,11 @@ package ingester
 
 import (
 	"io/ioutil"
-	"os"
 	"testing"
 	"time"
 
-	"github.com/cortexproject/cortex/pkg/util"
-	"github.com/go-kit/kit/log"
 	"github.com/stretchr/testify/require"
 )
-
-func init() {
-	util.Logger = log.NewLogfmtLogger(os.Stdout)
-}
 
 func TestWAL(t *testing.T) {
 	dirname, err := ioutil.TempDir("", "cortex-wal")


pstibrany:
What do you think about using `testing.Verbose()` flag to decide whether to install logger or not?

pracucci:
> What do you think about using testing.Verbose() flag to decide whether to install logger or not?

If we keep the current code and we enable it just with `testing.Verbose()` then it would just apply to `pkg/ingester` instead of having a uniform way to do it across the unit tests, right?

pstibrany:
> > What do you think about using testing.Verbose() flag to decide whether to install logger or not?
> 
> If we keep the current code and we enable it just with `testing.Verbose()` then it would just apply to `pkg/ingester` instead of having a uniform way to do it across the unit tests, right?

Right, something more global would be useful.

pstibrany:
Maybe moving this init to util package into init_test.go file. Would that work?

pstibrany:
> Maybe moving this init to util package into init_test.go file. Would that work?

No, it wouldn't. Calling `testing.Verbose()` is a no-go in the first place, since flags were not parsed at that time yet.

PR #2052, opened 2020-01-30
Adding Cassandra retries & updating the flush order of chunks/series when they use different storage layers.

### LINKED ISSUE #2009: If Index write fails or is throttled, chunk write is performed again; increasing costs on cloud providers.
The current flow in the flushing of a time series writes the chunk data first to storage, then the index data. This is probably a somewhat non-issue when the index & chunks are stored in the same storage service (DynamoDB, BigTable, etc), since the chunk write cost is close to free or is throttled along side the index write. However when the chunks are written to a blob storage service (S3, GCS or Azure) there is a per write transaction cost, which is multiplied by the number of times the index write action fails ( or is purposely throttled by the service provider)

### LINKED ISSUE #2010: Cassandra data provider has no retry policy options via config
Currently the Cassandra data package immediately fails on any error from the storage layer, even retry-able/recoverable errors such as rate throttling or network flaps.

khaines:
**What this PR does**: This PR addresses 2 items in the code base.
1. The Cassandra storage provider did not currently support retrying requests where the error is recoverable (timeout, etc). This adds settings for the number of retries to perform as well as values to control the exponential backoff details (min/max wait time).
2. Reverses the order in which series and chunks are flushed. Prior to this change the order was to write the chunk first and then the series. Since most blob storage services that can be used for chunks (S3, GCS and Azure) all charge a fee for writing data, if the series index write fails after a chunk write due to error or throttling; the cost for writing the chunk is wasted as it will be written again when the flush is retried. Since the ingesters will retry failed flushes immediately, this cost can escalate quickly if the error condition of the index store is sustained for long periods.

Fixes #2009 
Fixes #2010 

**Checklist**
- [x]  ran `make doc`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 * [ENHANCEMENT] Experimental TSDB: Export TSDB Syncer metrics from Compactor component, they are prefixed with `cortex_compactor_`. #2023
 * [ENHANCEMENT] Experimental TSDB: Added dedicated flag `-experimental.tsdb.bucket-store.tenant-sync-concurrency` to configure the maximum number of concurrent tenants for which blocks are synched. #2026
 * [ENHANCEMENT] Experimental TSDB: Expose metrics for objstore operations (prefixed with `cortex_<component>_thanos_objstore_`, component being one of `ingester`, `querier` and `compactor`). #2027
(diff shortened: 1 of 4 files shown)


gouthamve:
> Reverses the order in which series and chunks are flushed

This would cause subtle race conditions tbh. Like the queriers will be able to see the index entries but the chunk might not have been flushed so they won't be able to load it causing errors in queries.

khaines:
I'm going to split this into two PRs, because the cassandra options can stand alone aside from the flush ordering conversation.

PR #2055, opened 2020-01-30
Upgrade gogo/protobuf to 1.3.0

cyriltovena:

**What this PR does**:

This PR upgrades gogo/protobuf generation tools from 1.2.1 to 1.3. I've dig into their changelog and this is a single version jump, only bug fixes (https://github.com/gogo/protobuf/compare/v1.2.1...v1.3.0). The reason I'm upgrading is because there is an API break between 1.2.1 and 1.3.0 generated pb files, which means I can't import any proto files in the project I'm working on (Loki).

While you can't import generated 1.2 file with generated 1.3 file (compilation issue), they are although still backward compatible (two applications can communicate by using 1.2 and 1.3 respectively)

go vendoring was already using 1.3.0.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cyriltovena:
Ho no ! weaveworks needs to also be updated for the frontend.

```
import "github.com/weaveworks/common/httpgrpc/httpgrpc.proto";
```

The build error is the same problem I'm hitting in Loki. (`MarshalToSizedBuffer undefined`)

cyriltovena:
see https://github.com/gogo/protobuf#release-v130

> With this new release comes a new internal library version. This means any newly generated *pb.go files generated with the v1.3.0 library will not be compatible with the old library version (v1.2.1). However, current *pb.go files (generated with v1.2.1) should still work with the new library.

> Please make sure you manage your dependencies correctly when upgrading your project. If you are still using v1.2.1 and you update your dependencies, one of which could include a new *pb.go (generated with v1.3.0), you could get a compile time error.

CHANGELOG.md
```diff
@@ -2,8 +2,9 @@
 
 ## master / unreleased
 
+* [CHANGE] Upgraded gogo/protobuf tool chain from 1.2.1 to 1.3.0. #2055
 * [CHANGE] Removed remaining support for using denormalised tokens in the ring. If you're still running ingesters with denormalised tokens (Cortex 0.4 or earlier, with `-ingester.normalise-tokens=false`), such ingesters will now be completely invisible to distributors and need to be either switched to Cortex 0.6.0 or later, or be configured to use normalised tokens. #2034
-* [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893 
+* [CHANGE] Moved `--store.min-chunk-age` to the Querier config as `--querier.query-store-after`, allowing the store to be skipped during query time if the metrics wouldn't be found. The YAML config option `ingestermaxquerylookback` has been renamed to `query_ingesters_within` to match its CLI flag. #1893
```
pracucci: Could you rollback the changes here, except your entry, please? I think you have your editor trimming lines.

bboreham:
>  still backward compatible (two applications can communicate by using 1.2 and 1.3 respectively)

Who has tested this?  What was the process and results?

(We could add it to the "integration tests" in this repo)

CHANGELOG.md
cyriltovena: My editor is linting markdown.

https://github.com/updownpress/markdown-lint/blob/master/rules/009-no-trailing-spaces.md

cyriltovena:
@bboreham wrote a test, found a bug with the custom type, I guess I have to regen those too working on it. Everything else looks good will share more soon. I don't think there is a compatibility issue, I think I didn't regen code inside https://github.com/cortexproject/cortex/blob/master/pkg/ingester/client/timeseries.go

CHANGELOG.md
cyriltovena: Also currently the doc is inconsistent wrt spaces.

cyriltovena:
@bboreham I added a full ingesters tests and there were only issues with custom type. Once regenerated everything works fine, I can assure you now that the wire format has not changed.

Not sure if you want to keep those tests they are very long (took me a day), but at least you have the proof, I'm happy to remove them if you want.

We're now blocked by https://github.com/weaveworks/common/pull/177 which has no custom type usage so this will also be backward compatible.

CHANGELOG.md
pracucci: Yeah, but this is not the PR where it should be fixed. Let's keep things compartmented.

bboreham:
> I don't think there is a compatibility issue

With respect, what you think matters a lot less than what the tests show us.

Are the three CircleCI failures something you can fix?

bboreham:
I don't understand why we have extra data structure code for integration tests: why don't we just run the image from Cortex 0.6.0 ?

cyriltovena:
> Are the three CircleCI failures something you can fix?

The CI fails because of https://github.com/weaveworks/common/pull/177. see my comment here https://github.com/cortexproject/cortex/pull/2055#issuecomment-580409159

> With respect, what you think matters a lot less than what the tests show us.

This is the test I wrote.
https://github.com/cortexproject/cortex/blob/f60de80bce45ed7e3f1b6633356b3bbaa65954d9/integration-tests/proto/proto_test.go

It uses the two different versions of the code (1.2.1 is master and 1.3.0 is this PR). It was more simple to experiment proto services, I believe writing an full integrations test on a CI that I don't control will be way more complex and will probably cover less. This tests that I did actually covers the full ingesters proto service which the most complex we have, I  also tests marshalling of custom proto from 1.2 to 1.3 and vise versa because those were having issue (forgot to regenerate that copied code) so I put more effort into those.

I'm basically giving a proof of compatibility with that test, you can try it yourself and we can get rid of it once we all are happy.

bboreham:
Can you put an override in `go.mod` so this PR builds off your branch in `weaveworks/common`?

I still don't get what's so complicated about running an integration test with two different image versions.  E.g. 0.6.0 ingester against current distributor.  We have push tests already; why isn't it a matter of copy-pasting that with a different image tag?

cyriltovena:
I'll put the override, and give a go to integration tests.

cyriltovena:
I removed my go test (although it allowed me to discover a bug in the query path) as this was a simple proof of backward compatibility.

I merged your PR.

Let me know.

cyriltovena:
Looks like we need to update the CI with the new image, but I would need access or someone to do it for me.

bboreham:
The CI is controlled from `.circleci/config.yml`; you can change that as part of a PR.

Probably you can’t push an image under the same name, but for the purpose of testing you can use a different name.

cyriltovena:
This is ready to go, I'll rebase to upstream/master once I have the green light.

CHANGELOG.md
```diff
@@ -2,9 +2,10 @@
 
 ## master / unreleased
 
+* [CHANGE] Upgraded gogo/protobuf tool chain from 1.2.1 to 1.3.0. #2055
```
pracucci: We keep track only of changes which are visible to the final user. This change doesn't look so, so I would suggest to remove this entry (and so to rollback all `CHANGELOG.md` changes).

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 # https://circleci.com/blog/circleci-hacks-reuse-yaml-in-your-circleci-config-with-yaml/
 defaults: &defaults
   docker:
-    - image: quay.io/cortexproject/build-image:validate-k8s-specs-7c217ee7
+    - image: quay.io/cortexproject/cortex:update-gogoproto-6e528d554
```
pracucci: Note to both: rebuild the image once the PR is ready to merge and replace the tag here.

docs/configuration/config-file-reference.md
```diff
@@ -165,31 +165,6 @@ The `server_config` configures the HTTP and gRPC server of the launched service(
 # CLI flag: -server.grpc-max-concurrent-streams
 [grpc_server_max_concurrent_streams: <int> | default = 100]
 
-# The duration after which an idle connection should be closed. Default:
```
pracucci: Please rebase `master` and then `make doc` again.

go.mod
```diff
@@ -82,3 +82,5 @@ replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110
 
 // Override reference that causes an error from Go proxy - see https://github.com/golang/go/issues/33558
 replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+
+replace github.com/weaveworks/common => github.com/cyriltovena/common v0.0.0-20200130195811-b9d950b97870
```
pracucci: Need to get the `weaveworks/common` PR merged before proceeding merging this PR.

integration-tests/test-backcompat.sh
```diff
@@ -0,0 +1,34 @@
+#! /bin/sh
```
pracucci: Please remove this file, now that we have integration tests in Go (PR #1839).

pkg/ingester/client/timeseries.go
```diff
@@ -70,27 +70,36 @@ type LabelAdapter labels.Label
 
 // Marshal implements proto.Marshaller.
 func (bs *LabelAdapter) Marshal() ([]byte, error) {
-	buf := make([]byte, bs.Size())
-	_, err := bs.MarshalTo(buf)
-	return buf, err
+	size := bs.Size()
+	buf := make([]byte, size)
+	n, err := bs.MarshalToSizedBuffer(buf[:size])
+	return buf[:n], err
```
pracucci: Not sure this is correct. Looking at others generated `*.pb.go` files, I believe it should be:

```
if err != nil {
    return nil, err
}
return buf[:n], nil
```

pkg/ingester/client/timeseries.go
```diff
@@ -70,27 +70,36 @@ type LabelAdapter labels.Label
 
 // Marshal implements proto.Marshaller.
 func (bs *LabelAdapter) Marshal() ([]byte, error) {
```
pracucci: Please add a test covering `Marshal()` + `Unmarshal()`. Ideally, given a set of input `LabelAdapter`, we should marshal + unmarshal it and ensure the output is identical to the input. This will implicitly test `MarshalToSizedBuffer()` and `Size()`.

.circleci/config.yml
cyriltovena: I think we'll keep this one.

pkg/ingester/client/timeseries.go
cyriltovena: yes good catch, although not sure it matter, but will update.

.circleci/config.yml
cyriltovena: welp it seems that this is not a build image. @gouthamve can you help me here please. I have to rollback to my own for now.

cyriltovena: Rolled back to ctovena/cortex-build:update-gogoproto-2294034c if a maintainer can pull/tag/push to quay that would be great.

pracucci: I take care of building the new image. Will post the updated tag as soon as I have it.

pracucci: I built a new image from this PR and published it:
`quay.io/cortexproject/build-image:update-gogoproto-9a1b26ac6`

Could you update `.circleci/config.yml` accordingly, please?

CHANGELOG.md
gouthamve: Why was this marked as resolved?

pkg/ingester/client/timeseries_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestLabelAdapter_Marshal(t *testing.T) {
+	tests := []struct {
+		bs *LabelAdapter
+	}{
+		{&LabelAdapter{Name: "foo", Value: "bar"}},
+		{&LabelAdapter{Name: "very long label name", Value: "very long label value"}},
+		{&LabelAdapter{}},
```
bboreham: This case can't occur for real, can it?
The case of nonblank Name and blank Value is possible, if you're aiming to cover all cases.

cyriltovena: Ha yes true forgot about that one ! This all empty can't really occur but I figure it should still works ok.

cyriltovena:
Merge squash when you're ready. (Those commits are not useful)

Issue #1823, opened 2019-11-15
TSDB support for limits


tomwilkie:
We need to be able to limit number of active series etc.

thorfour:
Current design proposal. https://docs.google.com/document/d/18WJgjhBWhHqTmRPGmdG0IMfuD5qah0UAe9955gsTHwY/edit?usp=sharing

Kind of stuck on invalidating the head block limits

codesome:
^ still relevant. @pracucci: can we add `keepalive` to this?

pracucci:
Closed by https://github.com/cortexproject/cortex/pull/2676.

PR #2056, opened 2020-01-31
ingester v2: add MaxSeriesPerUser limit

kamijin-fanta:
**What this PR does**:
This is part of the Limit Support work in TSDB. Limit the number of recent series per user. Configure by `-ingester.max-series-per-user`, etc.

#1823 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -198,10 +198,17 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 		lset := client.FromLabelAdaptersToLabelsWithCopy(ts.Labels)
 
 		for _, s := range ts.Samples {
-			_, err := app.Add(lset, s.TimestampMs, s.Value)
+			// Check rate-limit
+			var err error
+			if e := i.limiter.AssertMaxSeriesPerUser(userID, int(db.currentNumSeries())); e != nil {
+				err = makeLimitError(perUserSeriesLimit, e)
+			}
 			if err == nil {
-				succeededSamplesCount++
-				continue
+				_, err = app.Add(lset, s.TimestampMs, s.Value)
+				if err == nil {
+					succeededSamplesCount++
+					continue
+				}
 			}
 
 			failedSamplesCount++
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -198,10 +198,17 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 		lset := client.FromLabelAdaptersToLabelsWithCopy(ts.Labels)
 
 		for _, s := range ts.Samples {
-			_, err := app.Add(lset, s.TimestampMs, s.Value)
+			// Check rate-limit
+			var err error
+			if e := i.limiter.AssertMaxSeriesPerUser(userID, int(db.currentNumSeries())); e != nil {
+				err = makeLimitError(perUserSeriesLimit, e)
+			}
```
codesome: This assumes that every sample creates a new series. If a user has touched the limit and is trying to add a sample to an existing series, this will error out. The check should be done only when creating a new series, which seems to be tricky as the creation is internal to TSDB.

pkg/ingester/ingester_v2.go
```diff
@@ -782,3 +792,7 @@ func (i *Ingester) shipBlocks() {
 	// Wait until all workers completed.
 	wg.Wait()
 }
+
+func (db *userTSDB) currentNumSeries() uint64 {
+	return db.Head().NumSeries()
+}
```
codesome: No need for this new method. It is fine to call `db.Head().NumSeries()` wherever needed.

pkg/ingester/ingester_v2.go
kamijin-fanta: Oh, the code seems to be wrong. When the limit is reached, the existing series cannot be pushed.

I seem to need to keep recent Series in memory or change Prometheus API...

pkg/ingester/ingester_v2.go
kamijin-fanta: ok. i'll fix it. thanks!

pkg/ingester/ingester_v2.go
codesome: I think https://github.com/cortexproject/cortex/pull/2057 will help. It can tell if a series exists or no and based on that we can apply the rate limit using `NumSeries()`. The interaction with `Purge()` method would need some attention. Maybe you can wait till that PR is merged.

pracucci:
@codesome Kind reminder to investigate what we can do on TSDB side to facilitate per-user and per-metric limits. I know this may be tricky, but would be great having a plan. If the per-metric limit is hard, we may start with the per-user limit only (which between the two is more important in a multi-tenant scenario cause protects from bad acting neighbours).

pracucci:
_Closing because @codesome is working on some TSDB changes to make this happen. We could re-open this PR if it will still relevant based on the final design (still in progress)._

PR #2057, opened 2020-01-31
Reduced memory allocations in the ingesters for the blocks storage

pracucci:
**What this PR does**:
In this PR I'm proposing a solution to reduce memory allocations in the ingesters for the blocks storage, introducing `RefCache` to hold TSDB series references.

Looking at our blocks storage test cluster, we can observe 25% of memory allocations is done by `FromLabelAdaptersToLabelsWithCopy`. Contrary to the chunks storage, whenever we call TSDB `appender.Add()` we have to pass a copy of the labels because they may be retained by TSDB (the request labels are reused to reduce memory allocations).

This PR introduces a series reference cache, so that we can use `appender.AddFast()` instead of `Add()` passing the series reference ID (which is a `uint64`) instead of the labels each time we have the reference cached.

Some notes about the `RefCache` design:
1. It's per-tenant, so we have 1 for each tenant
2. It stripes the cache into 128 buckets to reduce lock contention (128 is a trade-off to not use too much memory on clusters with many tenants but few series, like @thorfour use case)
3. It's intentional having no cap on the max number of series cached in the cache, but there's a purge mechanism to quickly remove stale series. If there would be a cap, once the cap is reached other series would have labels being copied every time, which would allocate that memory anyway until the next GC runs, so keeping all active series references in memory looks a win-win to me.

I'm running this change in our test cluster and memory allocations have been reduced by -25%. In the chart below I'm not showing memory allocation rate > 20 MB/s to hide away the noise every 2 hours when all TSDBs compact the head at the same time (it's another issue, see https://github.com/cortexproject/cortex/issues/2003).

Screen Shot 2020-01-31 at 08 39 57

I've **not** observed any difference in GC duration time or ingesters response time, neither in the ingesters `go_memstats_heap_inuse_bytes` except that it grows a bit slower because of the reduced allocations, as shown below (same time range):

Screen Shot 2020-01-31 at 08 50 02

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -17,6 +17,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/gate"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
@@ -40,6 +41,7 @@ type Shipper interface {
 
 type userTSDB struct {
 	*tsdb.DB
+	refCache *cortex_tsdb.RefCache
 
 	// Thanos shipper used to ship blocks to the storage.
 	shipper       Shipper
@@ -126,16 +128,21 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 	}
 
 	i.done.Add(1)
-	go i.rateUpdateLoop()
+	go i.updateLoop()
 
 	return i, nil
 }
 
-func (i *Ingester) rateUpdateLoop() {
+func (i *Ingester) updateLoop() {
 	defer i.done.Done()
 
 	rateUpdateTicker := time.NewTicker(i.cfg.RateUpdatePeriod)
 	defer rateUpdateTicker.Stop()
+
+	// We use an hardcoded value for this ticket because there should be no
+	// real value in customizing it.
+	refCachePurgeTicker := time.NewTicker(5 * time.Minute)
+	defer refCachePurgeTicker.Stop()
 
 	for {
 		select {
@@ -146,6 +153,13 @@ func (i *Ingester) rateUpdateLoop() {
 				db.ingestedRuleSamples.tick()
 			}
 			i.userStatesMtx.RUnlock()
+		case <-refCachePurgeTicker.C:
+			for _, userID := range i.getTSDBUsers() {
+				userDB := i.getTSDB(userID)
+				if userDB != nil {
+					userDB.refCache.Purge(time.Now().UnixNano(), cortex_tsdb.DefaultRefCacheTTL)
+				}
+			}
 		case <-i.quit:
 			return
 		}
@@ -190,18 +204,45 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 	// successfully committed
 	succeededSamplesCount := 0
 	failedSamplesCount := 0
+	now := model.Now().UnixNano()
 
 	// Walk the samples, appending them to the users database
 	app := db.Appender()
 	for _, ts := range req.Timeseries {
-		// Convert labels to the type expected by TSDB
-		lset := client.FromLabelAdaptersToLabelsWithCopy(ts.Labels)
+		// Check if we already have a cached reference for this series. Be aware
+		// that even if we have a reference it's not guaranteed to be still valid.
+		cachedRef, cachedRefExists := db.refCache.GetRef(now, client.FromLabelAdaptersToLabels(ts.Labels))
 
 		for _, s := range ts.Samples {
-			_, err := app.Add(lset, s.TimestampMs, s.Value)
-			if err == nil {
-				succeededSamplesCount++
-				continue
+			var err error
+
+			// If the cached reference exists, we try to use it.
+			if cachedRefExists {
+				if err = app.AddFast(cachedRef, s.TimestampMs, s.Value); err == nil {
+					succeededSamplesCount++
+					continue
+				}
+
+				if err == tsdb.ErrNotFound {
+					cachedRefExists = false
+				}
+			}
+
+			// If the cached reference doesn't exist, we (re)try without using the reference.
+			if !cachedRefExists {
+				var ref uint64
+
+				// Copy the label set because both TSDB and the cache may retain it.
+				copiedLabels := client.FromLabelAdaptersToLabelsWithCopy(ts.Labels)
+
+				if ref, err = app.Add(copiedLabels, s.TimestampMs, s.Value); err == nil {
+					db.refCache.SetRef(now, copiedLabels, ref)
+					cachedRef = ref
+					cachedRefExists = true
+
+					succeededSamplesCount++
+					continue
+				}
 			}
 
 			failedSamplesCount++
@@ -212,7 +253,7 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 			// we actually ingested all samples which haven't failed.
 			if err == tsdb.ErrOutOfBounds || err == tsdb.ErrOutOfOrderSample || err == tsdb.ErrAmendSample {
 				if firstPartialErr == nil {
-					firstPartialErr = errors.Wrapf(err, "series=%s", lset.String())
+					firstPartialErr = errors.Wrapf(err, "series=%s", client.FromLabelAdaptersToLabels(ts.Labels).String())
 				}
 
 				continue
@@ -552,6 +593,7 @@ func (i *Ingester) createTSDB(userID string) (*userTSDB, error) {
 
 	userDB := &userTSDB{
 		DB:                  db,
+		refCache:            cortex_tsdb.NewRefCache(),
 		ingestedAPISamples:  newEWMARate(0.2, i.cfg.RateUpdatePeriod),
 		ingestedRuleSamples: newEWMARate(0.2, i.cfg.RateUpdatePeriod),
 	}
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -156,7 +156,7 @@ func (i *Ingester) updateLoop() {
 			for _, userID := range i.getTSDBUsers() {
 				userDB := i.getTSDB(userID)
 				if userDB != nil {
-					userDB.refCache.Purge(time.Now().Add(time.Duration(-cortex_tsdb.DefaultRefCacheTTL)))
+					userDB.refCache.Purge(time.Now().Add(-cortex_tsdb.DefaultRefCacheTTL))
 				}
 			}
 		case <-i.quit:
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -138,7 +138,7 @@ func (i *Ingester) updateLoop() {
 	rateUpdateTicker := time.NewTicker(i.cfg.RateUpdatePeriod)
 	defer rateUpdateTicker.Stop()
 
-	// We use an hardcoded value for this ticket because there should be no
+	// We use an hardcoded value for this ticker because there should be no
 	// real value in customizing it.
 	refCachePurgeTicker := time.NewTicker(5 * time.Minute)
 	defer refCachePurgeTicker.Stop()
@@ -210,7 +210,7 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 	for _, ts := range req.Timeseries {
 		// Check if we already have a cached reference for this series. Be aware
 		// that even if we have a reference it's not guaranteed to be still valid.
-		cachedRef, cachedRefExists := db.refCache.GetRef(now, client.FromLabelAdaptersToLabels(ts.Labels))
+		cachedRef, cachedRefExists := db.refCache.Ref(now, client.FromLabelAdaptersToLabels(ts.Labels))
 
 		for _, s := range ts.Samples {
 			var err error
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -226,6 +226,7 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
 
 				if err == tsdb.ErrNotFound {
 					cachedRefExists = false
+					err = nil
 				}
 			}
 
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -17,7 +17,6 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/gate"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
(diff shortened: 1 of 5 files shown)


pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+
+// SetRef sets/updates the cached series reference. The input labels set IS retained.
+func (c *RefCache) SetRef(now int64, series []labels.Label, ref uint64) {
+	fp := client.FastFingerprint(client.FromLabelsToLabelAdapters(series))
+
+	// Get the related stripe
+	stripeID := uint8(util.HashFP(fp) % refCacheStripes)
+	stripe := c.stripes[stripeID]
+
+	stripe.refsMu.Lock()
```
pstibrany: The rest of this method can be moved to a method on stripe itself. Also one can use `defer refsMu.Unlock()` here.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	// The entry doesn't exist, so we have to add a new one.
+	stripe.refs[fp] = append(stripe.refs[fp], &refCacheEntry{lbs: series, ref: ref, touchedAt: now})
+	stripe.refsMu.Unlock()
+}
+
+// Purge removes expired entries from the cache. This function should be called
+// periodically to avoid memory leaks.
+func (c *RefCache) Purge(now int64, ttl int64) {
```
pstibrany: Instead of two values, one can simply pass cut-off time, everything older than that is removed.

pkg/ingester/ingester_v2.go
```diff
@@ -126,17 +128,22 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
...
 }
 
-func (i *Ingester) rateUpdateLoop() {
+func (i *Ingester) updateLoop() {
 	defer i.done.Done()
 
 	rateUpdateTicker := time.NewTicker(i.cfg.RateUpdatePeriod)
 	defer rateUpdateTicker.Stop()
 
+	// We use an hardcoded value for this ticket because there should be no
```
thorfour: `nit` spelling: ticket -> ticker

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+// is NOT retained.
+func (c *RefCache) GetRef(now int64, series []labels.Label) (uint64, bool) {
+	fp := client.FastFingerprint(client.FromLabelsToLabelAdapters(series))
+
+	// Get the related stripe
+	stripeID := uint8(util.HashFP(fp) % refCacheStripes)
+	stripe := c.stripes[stripeID]
+
+	// Look for series within the stripe
+	stripe.refsMu.Lock()
```
codesome: Given all the unlocks are near the returns, we can have it as a `defer` just after this.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+
+	refCacheStripes = 128
+)
+
+// RefCache is a single-tenant cache mapping a labels set with the reference
+// ID in TSDB, in order to be able to append samples to the TSDB head without having
+// to copy write request series labels each time (because the memory buffers used to
+// unmarshal the write request is reused).
+type RefCache struct {
+	// The cache is splitted into stripes, each one with a dedicated lock, in
```
thorfour: `nit` grammer: splitted -> split

pkg/ingester/ingester_v2.go
```diff
@@ -190,18 +204,45 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
...
+	now := model.Now().UnixNano()
 
 	// Walk the samples, appending them to the users database
 	app := db.Appender()
 	for _, ts := range req.Timeseries {
-		// Convert labels to the type expected by TSDB
-		lset := client.FromLabelAdaptersToLabelsWithCopy(ts.Labels)
+		// Check if we already have a cached reference for this series. Be aware
+		// that even if we have a reference it's not guaranteed to be still valid.
+		cachedRef, cachedRefExists := db.refCache.GetRef(now, client.FromLabelAdaptersToLabels(ts.Labels))
```
codesome: Can we also comment that `ts.Labels` should be sorted?

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+	// order to reduce lock contention.
+	stripes map[uint8]*refCacheStripe
```
thorfour: Since stripes isn't being expanded I think we should use an array (since we have a const) instead of a map.

pkg/storage/tsdb/ref_cache.go
codesome: +1 for this.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+			refs: map[model.Fingerprint][]*refCacheEntry{},
+		}
+	}
+
+	return c
+}
+
+// GetRef returns the cached series reference, and guarantees the input labels set
+// is NOT retained.
+func (c *RefCache) GetRef(now int64, series []labels.Label) (uint64, bool) {
```
thorfour: `nit` 
`GetRef` should probably just be renamed to `Ref`.
`SetRef` could also be renamed to `Set`

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+	fp := client.FastFingerprint(client.FromLabelsToLabelAdapters(series))
+
+	// Get the related stripe
+	stripeID := uint8(util.HashFP(fp) % refCacheStripes)
```
thorfour: related to above about changing from map -> array, since we already hash and modulo we should use an array instead.

pkg/storage/tsdb/ref_cache.go
thorfour: probably not worth doing without a profile, but we could likely use a reader lock here, and atomic set the touched at timestamp.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Purge removes expired entries from the cache. This function should be called
+// periodically to avoid memory leaks.
+func (c *RefCache) Purge(now int64, ttl int64) {
+	for s := uint8(0); s < refCacheStripes; s++ {
+		c.purgeStripe(now, ttl, c.stripes[s])
+	}
+}
+
+func (c *RefCache) purgeStripe(now int64, ttl int64, stripe *refCacheStripe) {
```
thorfour: could squash the `now` and `ttl` params into a single `age` param since all they're used for is to do `now - ttl`

pkg/storage/tsdb/ref_cache_test.go
```diff
@@ -0,0 +1,113 @@
...
+func BenchmarkRefCache_GetRef(b *testing.B) {
+	ls := []labels.Label{{Name: "a", Value: "1"}}
+	c := NewRefCache()
+	c.SetRef(0, ls, 1)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		c.GetRef(0, ls)
+	}
+}
```
codesome: Given you are using only 1 series, these benchmarks doesn't seem to be pushing it hard enough.

pkg/ingester/ingester_v2.go
```diff
@@ -190,18 +204,45 @@ func (i *Ingester) v2Push(ctx old_ctx.Context, req *client.WriteRequest) (*clien
...
+			// If the cached reference exists, we try to use it.
+			if cachedRefExists {
+				if err = app.AddFast(cachedRef, s.TimestampMs, s.Value); err == nil {
+					succeededSamplesCount++
+					continue
+				}
+
+				if err == tsdb.ErrNotFound {
+					cachedRefExists = false
+				}
```
codesome: Would be safe to set `err` to `nil` inside here to avoid any subtle bugs in the future.

pkg/storage/tsdb/ref_cache.go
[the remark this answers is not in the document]
pracucci: According to the benchmarks, there's a small performance penalty calling `UnixNano()` each time inside the `Get/SetRef()`.

On the other side, `time.Time` is 3x bigger than `uint64`, reason why I didn't store it before. However, for 1M series we're talking about +16MB which is actually negligible, so I've stored `time.Time`.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	stripe.refsMu.Unlock()
+	return 0, false
+}
+
+// SetRef sets/updates the cached series reference. The input labels set IS retained.
+func (c *RefCache) SetRef(now int64, series []labels.Label, ref uint64) {
+	fp := client.FastFingerprint(client.FromLabelsToLabelAdapters(series))
```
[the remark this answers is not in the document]
pracucci: I initially picked `FastFingerprint()` because it's supposed by the "fast", but I've run the benchmark with `Fingerprint()` and the penalty is negligible.

pkg/storage/tsdb/ref_cache.go
pracucci: Oh sure! I'm wondering why I've used a map here 🤔

pkg/storage/tsdb/ref_cache.go
pracucci: Agree for `GetRef` > `Ref`. I think `SetRef` is closer to Go patterns. Am I wrong?

pkg/storage/tsdb/ref_cache.go
pracucci: I did try it, but the profile didn't show any significative difference, so I've rolled back to this lock which simplify a bit the code.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,161 @@
...
+// unmarshal the write request is reused).
+type RefCache struct {
+	// The cache is splitted into stripes, each one with a dedicated lock, in
+	// order to reduce lock contention.
+	stripes map[uint8]*refCacheStripe
+}
+
+// refCacheStripe holds a subset of the series references for a single tenant.
+type refCacheStripe struct {
+	refsMu sync.Mutex
```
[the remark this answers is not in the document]
pracucci: Unfortunately not because the getter also touch the timestamp.

pkg/storage/tsdb/ref_cache.go
thorfour: yea SetRef can be left as is

pkg/storage/tsdb/ref_cache.go
thorfour: sounds good :+1:

pracucci: I didn't use `defer` because there's a performance penalty. According to the benchmark, the benefit of using an array for the stripes (already done) is lost if we introduce a `defer` here.

I've moved the logic to a function to `refCacheStripe` so we can lock outside it, keeping the code cleaner.

pstibrany: > I didn't use `defer` because there's a performance penalty.

I would prefer clearer, more idiomatic code than slightly slower code (which in Go 1.15 will be equally fast anyway).

pkg/storage/tsdb/ref_cache.go
pracucci: I didn't use `defer` cause it brings a performance penalty (measured with the benchmark). However, I've moved the logic within a function on the stripe, and locked outside it.

pkg/storage/tsdb/ref_cache.go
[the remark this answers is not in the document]
pstibrany: > According to the benchmarks, there's a small performance penalty calling `UnixNano()` each time inside the `Get/SetRef()`.

This is a weird argument... that penaly must be paid somewhere. In the original code, it was before Get/SetRef was called. I suggested to move that cost inside those method instead. Of course they will be slower, but overall (when taking code *calling* them into account), it’s the same.

pkg/storage/tsdb/ref_cache.go
thorfour: I guess I need to know what the perf penalty actually was otherwise I'd agree with @pstibrany Go 1.14 obviates the need to not defer

pkg/storage/tsdb/ref_cache.go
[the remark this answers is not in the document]
pracucci: In the original implementation it was called once for a push request. Moving it inside, it gets called once for each series in the a single push request (ie. if a push request contains 1K series, then it's called 1K times more).

Anyway, I've now stored `time.Time`.

pkg/storage/tsdb/ref_cache.go
pracucci: Ok. Used defer.

pracucci:
@codesome @pstibrany @thorfour Thanks for the feedback. I should have addressed all comments. May you check it again, please?

pkg/storage/tsdb/ref_cache.go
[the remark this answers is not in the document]
pstibrany: > I suggested to move that cost inside those method instead. Of course they will be slower, but overall (when taking code _calling_ them into account), it’s the same.

I guess the difference is that it’s only called once, when it’s outside of these methods. So what you wrote made sense. My mistake.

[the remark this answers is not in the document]
pracucci: Yes, if it's called outside it's just called once. No worries! You all left so many good feedback, that the overall change set improved a lot IMO.

pkg/storage/tsdb/ref_cache.go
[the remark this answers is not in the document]
codesome: Ah yes. If we face a lot of contention for any reason in the future, we can try our hands on atomics as a follow on work.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+const (
+	// DefaultRefCacheTTL is the default RefCache purge TTL. We use a reasonable
+	// value that should cover most use cases. The cache would be ineffective if
+	// the scrape interval of a series is greater than this TTL.
+	DefaultRefCacheTTL = int64(5 * time.Minute)
```
codesome: Why not `time.Duration(5 * time.Minute)` as it's only used that way?

pracucci: Right, done. Was incorrectly left as `int64` by the initial implementation.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+const (
+	// DefaultRefCacheTTL is the default RefCache purge TTL. We use a reasonable
+	// value that should cover most use cases. The cache would be ineffective if
+	// the scrape interval of a series is greater than this TTL.
+	DefaultRefCacheTTL = 5 * time.Minute
```
gouthamve: Can we make this `10 * time.Minute`? `5min` is still a common scrape interval on some intensive targets.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -0,0 +1,157 @@
...
+
+	refCacheStripes = 128
```
gouthamve: `refCacheStripes` --> `numRefCacheStripes`

Down in L29, it looks like an array of `refCacheStripe`.

pracucci:
Thanks @gouthamve for your review. I've addressed the comments.

> Though I must admit, I am a little worried that the purge process might block ingestion. I'd rather see it happen before we optimise for that though.

It's a legit objection. I've added a benchmark to measure it. Given 1M series in the cache and 10K expired ones, on my MacBook it takes 400ms. We have 128 stripes (the lock is individually taken for each stripe), so each lock is acquired (on average) 3.2ms. Doesn't look dramatic, but if we see performances issues because of this we can add optimizations.

Issue #2058, opened 2020-01-31
rulers are never removed from the ring if evicted


bboreham:
Basically the same problem as #1521, but rulers have no in-memory state to flush so it should be easier to ignore (or auto-forget) dead ones.  Maybe we should worry a little about a split-brain state, but surely the worst case is doubling up evaluation which isn't fatal.

The error message is:
```
level=error ts=2020-01-31T09:13:05.403429734Z caller=ruler.go:362 msg="unable to verify rule group ownership ownership, will retry on the next poll" err="at least 1 live ingesters required, could only find 0"
```

PR #2059, opened 2020-01-31
Replace 'ingesters' with 'replicas' in error message

bboreham:
This error will appear in contexts where we are not talking about ingesters, e.g. when running sharded rulers.

An example of the error message is at #2058 .

(Do we need to mention this in CHANGELOG?  Do we care about people parsing the text of error messages?)


bboreham:
So there are "people" who rely on the text matching 🤣

Issue #1195, opened 2019-01-21
Streaming queries are very inefficient


bboreham:
I noticed high resource usage in `ruler` and traced it back to a change where I turned on:

       - -querier.batch-iterators=true
       - -querier.ingester-streaming=true

Upon reverting this change, CPU went down to a third of what it was, memory down to a quarter and network traffic to a fifth.

Profiling suggests vast amounts of memory being used here:

```
github.com/cortexproject/cortex/pkg/querier/batch.newMergeIterator
/go/src/github.com/cortexproject/cortex/pkg/querier/batch/merge.go
  Total:      6.78TB     7.29TB (flat, cum) 38.52%
     20            .          .            
     21            .          .           	currErr error 
     22            .          .           } 
     23            .          .            
     24            .          .           func newMergeIterator(cs []chunk.Chunk) *mergeIterator { 
     25            .   151.63GB           	css := partitionChunks(cs) 
     26       5.39GB     5.39GB           	its := make([]*nonOverlappingIterator, 0, len(css)) 
     27            .          .           	for _, cs := range css { 
     28            .   365.93GB           		its = append(its, newNonOverlappingIterator(cs)) 
     29            .          .           	} 
     30            .          .            
     31            .          .           	c := &mergeIterator{ 
     32            .          .           		its:        its, 
     33      10.82GB    10.82GB           		h:          make(iteratorHeap, 0, len(its)), 
     34       3.36TB     3.36TB           		batches:    make(batchStream, 0, len(its)*2*promchunk.BatchSize), 
     35       3.40TB     3.40TB           		batchesBuf: make(batchStream, 0, len(its)*2*promchunk.BatchSize), 
     36            .          .           	} 
     37            .          .            
     38            .          .           	for _, iter := range c.its { 
     39            .          .           		if iter.Next(1) { 
     40            .          .           			c.h = append(c.h, iter) 
```

I am unclear why those sizes have `*promchunk.BatchSize` - they are allocating slices of `Batch` which are already sized that big.

ShuzZzle:
Is this still a thing?

ShuzZzle:
> I haven't profiled recently, but the code still looks the same to me.
> https://github.com/cortexproject/cortex/blob/20f8931014bc9f3271380884f8eadee8de83ae96/pkg/querier/batch/merge.go#L24-L36

Isn't this a major problem, when cpu/memory/bandwith explodes like you described? Especially since this is considered a good default (or even is the default)? :thinking:

[...]

PR #2060, opened 2020-01-31
Refactor Cortex querier to use Queryable interface directly

pstibrany:
This PR changes Cortex querier to use Queryable interface as its primary interface it builds on. Previously, Cortex querier was tied to Cortex chunks store, but this has problems when using it with TSDB blocks: Blocks Querier had to reencode TSDB chunks to Cortex chunks, and also ingester streaming support (PR #1845) would need to do the same.

Chunk-based queryables are still optimized in the way they were before (in mergeSeriesSets). The requirement is that they return series that implement new `SeriesWithChunks` interface, which gives new `mergeSeriesSets` method an option to use optimizations.

This PR also removes some duplicity in the code. First by merging `ingesterQueryable` (used when streaming ingester flag is set) into `distributorQuerier`, which now simply has a flag that decides whether streaming should be used or not. Second, logic from `unifiedChunkQuerier` was moved into `querier` (merging from multiple chunks-based stores), most of the code was already there anyway.

This PR has a follow-up PR that modifies BlocksQuerier to avoid TSDB -> Cortex chunks conversion.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/querier/distributor_queryable.go
```diff
@@ -37,13 +41,25 @@ type distributorQuerier struct {
...
 }
 
 func (q *distributorQuerier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
-	mint, maxt := q.mint, q.maxt
-	if sp != nil {
-		mint = sp.Start
-		maxt = sp.End
+	// Kludge: Prometheus passes nil SelectParams if it is doing a 'series' operation,
+	// which needs only metadata.
+	if sp == nil {
```
pracucci: Note to myself: this is the exact copy of `querier.metadataQuery()`, which both streaming and non-streaming `Queryable` were using before 👍

pkg/querier/querier.go
```diff
@@ -80,24 +80,26 @@ type ChunkStore interface {
...
 	var queryable storage.Queryable
-	if cfg.IngesterStreaming {
-		dq := newIngesterStreamingQueryable(distributor, iteratorFunc)
-		queryable = newUnifiedChunkQueryable(dq, chunkStore, distributor, iteratorFunc, cfg)
-	} else {
-		cq := newChunkStoreQueryable(chunkStore, iteratorFunc)
-		dq := newDistributorQueryable(distributor)
-		queryable = NewQueryable(dq, cq, distributor, cfg)
-	}
+	dq := newDistributorQueryable(distributor, cfg.IngesterStreaming, iteratorFunc)
```
pracucci: What do you thing if we rename `dq` to `distributorQueryable` for clarity, even if longer?

pkg/querier/querier.go
```diff
@@ -119,28 +121,31 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
 			maxt:        maxt,
+			chunkIterFn: chunkIterFn,
+		}
+
+		dqr, err := distribQuerier.Querier(ctx, mint, maxt)
+		if err != nil {
+			return nil, err
 		}
 
+		q.primary = dqr
```
pracucci: I find a bit unclear naming `primary` the `distributorQuerier`. I would be more explicit it, and just call it `distributorQuerier`.

pkg/querier/querier.go
```diff
@@ -153,19 +158,25 @@ func NewQueryable(dq, cq storage.Queryable, distributor Distributor, cfg Config)
...
+	chunkIterFn chunkIteratorFunc
 	ctx         context.Context
 	mint, maxt  int64
 }
 
 // Select implements storage.Querier.
 func (q querier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
 	// Kludge: Prometheus passes nil SelectParams if it is doing a 'series' operation,
-	// which needs only metadata.
+	// which needs only metadata. Here we expect that primary querier will handle that.
```
pracucci: > Here we expect that primary querier will handle that.

I would explain that, given the metadata queries have no mint/maxt and given it's not feasible to query the entire history, in Cortex we pick the trade-off to query only ingesters, so we skip querying the long-term storage.

pkg/querier/querier.go
```diff
@@ -119,28 +121,31 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
 }
 
 // NewQueryable creates a new Queryable for cortex.
-func NewQueryable(dq, cq storage.Queryable, distributor Distributor, cfg Config) storage.Queryable {
+func NewQueryable(distribQuerier, storeQuerier storage.Queryable, chunkIterFn chunkIteratorFunc, cfg Config) storage.Queryable {
 	return storage.QueryableFunc(func(ctx context.Context, mint, maxt int64) (storage.Querier, error) {
 		q := querier{
-			distributor: distributor,
 			ctx:         ctx,
```
pracucci: Is `ctx` still used?

pkg/querier/querier.go
```diff
@@ -191,28 +202,86 @@ func (q querier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (
...
+	firstNextCalled bool
+	set             storage.SeriesSet
+}
+
+func (pss *ignoreFirstNextSeriesSet) Next() bool {
+	if pss.firstNextCalled {
+		return pss.set.Next()
+	}
+	pss.firstNextCalled = true
+	return true
```
pracucci: What if the series set has only 1 item, so `Next()` is expected to return `false` on the first call? If this is a real issue, please cover this case with a unit test.

pkg/querier/querier.go
```diff
@@ -191,28 +202,86 @@ func (q querier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (
...
 
-func (q querier) metadataQuery(matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
-	ms, err := q.distributor.MetricsForLabelMatchers(q.ctx, model.Time(q.mint), model.Time(q.maxt), matchers...)
-	if err != nil {
-		return nil, nil, err
+func (querier) Close() error {
+	return nil
+}
+
+func (q querier) mergeSeriesSets(sets []storage.SeriesSet) storage.SeriesSet {
```
pracucci: Is this unit tested? I can't see its tests, but maybe I'm missing something.

pkg/querier/querier.go
```diff
@@ -191,28 +202,86 @@ func (q querier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (
...
+
+	// partitionChunks returns set with sorted series, so it can be used by NewMergeSeriesSet
+	chunksSet := partitionChunks(chunks, q.mint, q.maxt, q.chunkIterFn)
+
+	if len(otherSets) == 0 {
+		return chunksSet
 	}
-	return metricsToSeriesSet(ms), nil, nil
+
+	otherSets = append(otherSets, chunksSet)
```
pracucci: The input sets of `NewMergeSeriesSet` must be sorted. How can we guarantee that appending `otherSets ` and `chunksSet` together is still sorted?

pkg/querier/querier.go
```diff
@@ -191,28 +202,86 @@ func (q querier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (
...
+		}
+
+		s := set.At()
+		if sc, ok := s.(SeriesWithChunks); ok {
+			chunks = append(chunks, sc.Chunks()...)
+
+			// iterate over remaining series in this set, and store chunks
+			// Here we assume that all remaining series in the set are also backed-up by chunks.
+			// If not, there will be panics.
+			for set.Next() {
```
pracucci: Could you confirm me this is the way it's working right now (before this PR), and the configured `iteratorFunc` doesn't affect this, please?

pkg/querier/distributor_queryable.go
```diff
@@ -54,6 +70,45 @@ func (q *distributorQuerier) Select(sp *storage.SelectParams, matchers ...*label
...
+	for _, result := range results {
+		// Sometimes the ingester can send series that have no data.
+		if len(result.Chunks) == 0 {
+			continue
+		}
+
+		ls := client.FromLabelAdaptersToLabels(result.Labels)
+		sort.Sort(ls)
+
+		chunks, err := chunkcompat.FromChunks(userID, ls, result.Chunks)
```
pracucci: Before we were passing `nil` for the labels. Is this change intended?

pkg/querier/distributor_queryable.go
```diff
@@ -54,6 +70,45 @@ func (q *distributorQuerier) Select(sp *storage.SelectParams, matchers ...*label
 	return matrixToSeriesSet(matrix), nil, nil
 }
 
+func (q *distributorQuerier) streamingSelect(sp storage.SelectParams, matchers []*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
```
pracucci: Note to myself: except for the comment left (about labels passed to `chunkcompact.FromChunks`), this is the exact copy of `ingesterStreamingQuerier.Select()`.

pkg/querier/querier.go
pstibrany: My thinking was that it doesn't have to be distributorQueryable only, but we can also change it back to something else in the future when needed.

pkg/querier/querier.go
pstibrany: Good point. I wasn't 100% sure *why* we do it this way, only made sure to keep the functionality. I'll add your comment to the code.

Would it be possible to query long-time storage when using blocks? Queriers have that information available, don't they?

pkg/querier/querier.go
pstibrany: Good catch. It actually isn't, but I think it would make sense to use it in Select while waiting for individual results from queriers.

pkg/querier/querier.go
pracucci: > Would it be possible to query long-time storage when using blocks? Queriers have that information available, don't they?

It could be possible and Thanos does, but when I talked to Bartek about it he mentioned that Thanos is suffering the same issue as well and they may also want to introduce a limitation (Cortex limitation is quite strict because we query only active series, but still is a limitation). For now, I would stay consistent with the chunks storage.

pkg/querier/querier.go
pstibrany: > What if the series set has only 1 item, so `Next()` is expected to return `false` on the first call? If this is a real issue, please cover this case with a unit test.

I don't understand what you mean. Reason for this is that we call `set.Next()` once for every set to check if it is chunk-based or not. It if isn't, then we want to make sure that given set returns same result from `At()` call again to the query engine. Hence wrapping into this struct.

Now looking at it, I think I should cache that first `At()` result and simply return it again, instead of relying on set returning the same value twice.

pkg/querier/querier.go
pstibrany: Yes, TestQuerier function exercises all of this functionality. Merging multiple chunks is tested when streaming is on, while merging chunks and non-chunk-based sets happens with streaming off.

pkg/querier/querier.go
pstibrany: It's series in individual sets that must be sorted in order for merging to work properly.

Series in chunk-based sets are sorted (guaranteed by `partitionChunks`).

Non-chunk-based sets that we use are:
1. values from ingester when not using streaming – these are returned via `newConcreteSeriesSet`, which sorts the series by label,
2. after 2061 – series set returned by blocks storage, fortunately it also guarantees order of the series.

pkg/querier/querier.go
pstibrany: `iteratorFunc` has somewhat unfortunate name, because what it is really doing is processing different chunks that belong to the same series (this is important) into single iterator. So it's only handling single series at the time.

This code on the other hand, is iterating over the series in the set, and simply collecting chunks. They are sorted into per-series chunks by using `partitionChunks` function.

Previously this step happened in `unifiedChunkQuerier.Select` method, which dealt with chunk-based stores (chunks from all stores were returned by its `Get` method). This new method deals with both chunks-based and non-chunks-based sets.

pkg/querier/distributor_queryable.go
pstibrany: Yes, it is intended.

Code you're referring to was in `ingesterStreamingQuerier.Select` method before, and it is correct that we didn't set labels into the chunk there. However, `ingesterStreamingQuerier` was only ever used in combination with `unifiedChunkQuerier`, which does not call its `Select` method, but `Get` method instead. I believe that `Select` was only ever used from `TestIngesterStreaming`.

Reason for setting labels into chunk directly is so that `partitionChunks` can sort chunks into individual series correctly (via fingerprinting). It relies on this field to be set.

pstibrany:
Thank you Marco for your review. I believe I've addressed all your comments now.

pkg/querier/chunk_store_queryable.go
```diff
@@ -44,10 +44,11 @@ func (q *chunkStoreQuerier) Select(sp *storage.SelectParams, matchers ...*labels
 		return nil, nil, promql.ErrStorage{Err: err}
 	}
 
-	return q.partitionChunks(chunks), nil, nil
+	return partitionChunks(chunks, q.mint, q.maxt, q.chunkIteratorFunc), nil, nil
 }
 
-func (q *chunkStoreQuerier) partitionChunks(chunks []chunk.Chunk) storage.SeriesSet {
+// Series in the returned set are sorted by labels.
```
bboreham: It isn't immediately obvious to me what this means.  Sorted by ASCII ordering of `Labels.String()` ?  Sorted by fingerprint hash?

pkg/querier/distributor_queryable.go
```diff
@@ -2,14 +2,16 @@ package querier
 
 import (
 	"context"
+	"sort"
 
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/prom1/storage/metric"
+	"github.com/cortexproject/cortex/pkg/util/chunkcompat"
```
bboreham: Please leave the blank line so we get three groups of imports.

pkg/querier/distributor_queryable.go
```diff
@@ -37,13 +41,25 @@ type distributorQuerier struct {
...
+	streaming   bool
+	chunkIterFn chunkIteratorFunc
 }
 
 func (q *distributorQuerier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
-	mint, maxt := q.mint, q.maxt
-	if sp != nil {
-		mint = sp.Start
-		maxt = sp.End
+	// Kludge: Prometheus passes nil SelectParams if it is doing a 'series' operation,
```
bboreham: Can we clarify who is kludging what?  I think it's important that Cortex is relying on a behaviour of Prometheus, to make a decision that helps minimise effort at runtime.

pkg/querier/querier.go
```diff
@@ -119,58 +121,71 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
 	ctx         context.Context
 	mint, maxt  int64
 }
 
 // Select implements storage.Querier.
 func (q querier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
 	// Kludge: Prometheus passes nil SelectParams if it is doing a 'series' operation,
-	// which needs only metadata.
+	// which needs only metadata. Here we expect that distributorQuerier querier will handle that.
+	// In Cortex it is not feasible to query entire history (with no mint/maxt), so we only ingesters and skip querying
```
bboreham: s/we only/we only ask/ ?

pkg/util/merger.go
```diff
@@ -19,12 +19,8 @@ func MergeSampleSets(a, b []model.SamplePair) []model.SamplePair {
 			j++
 		}
 	}
-	for ; i < len(a); i++ {
-		result = append(result, a[i])
-	}
-	for ; j < len(b); j++ {
-		result = append(result, b[j])
-	}
+	result = append(result, a[i:]...)
```
bboreham: I think this deserves a comment like "now add the tail of either a or b; at least one of these appends is a no-op as a result of the loop condition"

pkg/querier/chunk_store_queryable_test.go
```diff
@@ -65,3 +65,23 @@ func mkChunk(t require.TestingT, mint, maxt model.Time, step time.Duration, enco
 	}
 	return chunk.NewChunk(userID, fp, metric, pc, mint, maxt)
 }
+
+func TestPartitionChunksOutputIsSortedByLabels(t *testing.T) {
```
bboreham: The relationship between the name and the code in this test isn't obvious to me. Perhaps add a comment to explain, or call `IsSorted()`.

pkg/querier/lazy_querier.go
```diff
@@ -47,16 +42,6 @@ func (l lazyQuerier) Close() error {
 	return l.next.Close()
 }
 
-// Get implements ChunkStore for the chunk tar HTTP handler.
```
bboreham: Is this implemented elsewhere or the feature has gone away?

pkg/querier/querier.go
```diff
@@ -119,58 +121,71 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
 		}
 
 		return q, nil
 	})
 }
 
 type querier struct {
-	queriers []storage.Querier
+	// used for labels and metadata queries
+	distributorQuerier storage.Querier
```
bboreham: Could we name this more like what we need from it rather than how we think it is implemented?

pkg/querier/querier.go
```diff
@@ -119,58 +121,71 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
...
 
-	distributor Distributor
+	// used for selecting series
+	selectQueriers []storage.Querier
```
bboreham: Can we do better than "selectQueriers" ?  What is the difference between a select and a query, and how does that change when we do both together?

pstibrany:
> Would like to see the benchmark results before and after, to check it hasn't introduced a performance regression. Particularly in light of #1195

Thanks Bryan for your review. Here are benchmarks of this PR against master: https://gist.github.com/pstibrany/6418ad45e9c982a9e10950c4ad8f12a8

Problem with this benchmark is that many tests don't run long enough to reasonably compare them. I will try to get a better one.

pkg/querier/querier.go
pracucci: > Now looking at it, I think I should cache that first At() result and simply return it again, instead of relying on set returning the same value twice.

What's wrong with calling `At()` twice for the same entry? If there's something wrong (or potentially wrong) would be better to add a comment to explain it. I'm missing it.

pkg/querier/chunk_store_queryable.go
pstibrany: They are sorted alphabetically by labels. This happens inside `newConcreteSeriesSet`, and is a prerequisite for efficient merging by `storage.NewMergeSeriesSet`. I will update comment to be more clear.

pkg/querier/distributor_queryable.go
pstibrany: I leave management of imports to my IDE and goimports. Sometimes they get it wrong. Sorry.

pkg/querier/distributor_queryable.go
pstibrany: I copied this comment from `querier` type, which had direct reference to distributor previously.

Any suggestion on how to improve/change this comment? It seems clear to me.

pkg/querier/lazy_querier.go
pstibrany: I didn't find any usage of this functionality.

pkg/querier/querier.go
pstibrany: What about `metadataQuerier`

pkg/querier/querier.go
pstibrany: There is only `Select` operation on `Querier` type. I wanted to express that *only* these queriers are used in `Select`, unlike `metadataQuerier`. But I think with `metadataQuerier`, I will simply rename this back.

pkg/querier/querier.go
pstibrany: Since we already have the value returned by first `At()`, there is no need to call it again. There are implementations of this function that do extra work each time `At()` is called (eg. `mergeSeriesSet` does the merging again, `dedupSeriesSet` does the deduplication again, etc.), so it seemed like a good idea.

pstibrany: I've changed the comment on `seriesSetWithFirstSeries` to mention this.

pstibrany:
I've rerun benchmarks with `-test.run '^$' -test.bench '^Bench.*' -test.v -test.benchmem -test.benchtime=10s` parameters.

https://gist.github.com/pstibrany/96b1509429714605bf59382640bc6032

pkg/querier/querier.go
pracucci: > do extra work each time At() is called

Ok. I was wondering if it was an optimization or bug fix. You confirmed it's just an optimization.

pkg/querier/querier.go
pracucci: `metadataQuerier` looks clear to me.

pkg/querier/distributor_queryable.go
pracucci: Maybe we can just remove the `Kludge:`? :) All in all, we're just describing how Prometheus works and why we handle the special case of `sp == nil`.

pkg/querier/querier.go
```diff
@@ -119,28 +121,31 @@ func New(cfg Config, distributor Distributor, chunkStore ChunkStore) (storage.Qu
 }
 
 // NewQueryable creates a new Queryable for cortex.
-func NewQueryable(dq, cq storage.Queryable, distributor Distributor, cfg Config) storage.Queryable {
+func NewQueryable(distribQuerier, storeQuerier storage.Queryable, chunkIterFn chunkIteratorFunc, cfg Config) storage.Queryable {
```
pracucci: A nit. Technically, I think `distribQuerier` and `storeQuerier` should be called `*Queryable`. Then the querier is instantiated from the queryable.

pkg/querier/querier.go
```diff
@@ -181,38 +196,106 @@ func (q querier) Select(sp *storage.SelectParams, matchers ...*labels.Matcher) (
...
+	if pss.firstNextCalled {
+		pss.firstSeries = nil
+		return pss.set.Next()
+	}
+	pss.firstNextCalled = true
+	return true
+}
+
+func (pss *seriesSetWithFirstSeries) At() storage.Series {
+	if pss.firstSeries != nil {
```
pracucci: Despite correct, it looks weird to me that this `if` is different then the one in the `Next()`. Wouldn't be more clear if we check `pss.firstNextCalled` here too? If `firstSeries` is `nil` we return `nil`. Am I wrong?

pstibrany: In situation where first `set.Next()` returns true, but at the same time Err() also returns non-nil, this would hide that error. I think I better add a comment to the code. WDYT?

pkg/querier/distributor_queryable.go
pstibrany: I don't see this documented anywhere else, so to me `Kludge:` indicates that this is possibly a hack that can break :)

pkg/querier/querier.go
pracucci: > In situation where first set.Next() returns true, but at the same time Err() also returns non-nil, this would hide that error. 

Why would the error hidden? I can't see a difference in error handling either we check `pss.firstSeries != nil` or `!pss.firstNextCalled`. Could you help me understand the scenario you're describing, please?

pstibrany: Scenario I'm talking about is this:

`mergeSeriesSets` calls `set.Next()` and then `set.At()` to get first series. `mergeSeriesSets` doesn't call `set.Err()` at all, but maybe this call would already return some error at this point if it was actually called. This is the error that would be hidden if we did the check in `seriesSetWithFirstSeries.Err()` method. (It wouldn't be lost, as it would be returned on next `Err()` call).

But looking at the actually usage, it seems that `Err()` is only checked when `Next()` returns false, so your suggestion would actually work.

I think there is a bug in `mergeSeriesSets` – we shouldn't just ignore empty sets, but return their errors if there are any. I'll fix that.

PR #2064, opened 2020-02-01
Add integration test using older image

bboreham:
An example of what I proposed at #2055 

Here we test pushing from the new distributor to an older ingester, and hand-over from the old ingester to a new ingester.

The backwards-compatibility version is pinned at v0.6.0.

### DIFF
--- a/integration-tests/test-backcompat.sh
+++ b/integration-tests/test-backcompat.sh
@@ -23,6 +23,8 @@ DIST_ADDR=$(container_ip distributor)
 wait_for "has_tokens_owned $DIST_ADDR" "distributor to see ingester in ring"
 
 docker run $RUN_ARGS --rm $AVALANCHE_IMAGE --metric-count=2 --label-count=2 --series-count=2 --remote-requests-count=1 --remote-url=http://distributor/api/prom/push
+# compute a timestamp just after the one Avalanche used, so we can query back
+TIMESTAMP=$(($(date +%s)+1))
 
 echo Start second ingester, waiting for first
 docker run $RUN_ARGS -d --name=i2 --hostname=i2 "$IMAGE_PREFIX"cortex:$IMAGE_TAG -target=ingester $INGESTER_ARGS -ingester.join-after=300s -ingester.claim-on-rollout=true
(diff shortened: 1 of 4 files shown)


integration-tests/test-backcompat.sh
```diff
@@ -0,0 +1,34 @@
...
+docker run $RUN_ARGS --rm $AVALANCHE_IMAGE --metric-count=2 --label-count=2 --series-count=2 --remote-requests-count=1 --remote-url=http://distributor/api/prom/push
+
+echo Start second ingester, waiting for first
+docker run $RUN_ARGS -d --name=i2 --hostname=i2 "$IMAGE_PREFIX"cortex:$IMAGE_TAG -target=ingester $INGESTER_ARGS -ingester.join-after=300s -ingester.claim-on-rollout=true
+
+echo Stop first ingester so it should hand over
+docker stop i1
+
+I2_ADDR=$(container_ip i2)
+wait_for "curl -s -f -m 3 $I2_ADDR/ready" "ingester ready"
```
pracucci: How does this guarantee that the hand-over succeeded? Shouldn't we query back the second ingester and ensure it has the same series pushed to the first one?

integration-tests/test-backcompat.sh
```diff
@@ -0,0 +1,34 @@
+#! /bin/sh
```
pracucci: This file should be added to `.circleci/config.yml` in order to run it as part of CI, right?

integration-tests/test-backcompat.sh
bboreham: Some failures, e.g. continuous crashing, will be detected, but you are right that we could do a lot better in checking that the test was fully successful.

integration-tests/test-backcompat.sh
bboreham: Added.

integration-tests/test-backcompat.sh
bboreham: I've extended the test to query back; it's still quite rudimentary.
I see in #1839 you did your own `Push()` rather than using Avalanche.  That would remove the problem that we can't check the random samples.

pracucci: Thanks for adding the querier. I understand the struggle of doing all this stuff with bash. The assertion you've introduced here LGTM (but please be aware tests are failing in CI).

This week I will revive the PR #1839 starting from your feedback (remove docker-compose, simplify it) and solve all TODOs: if we do a conjunct effort, me on the implementation and you on the review, we may get something solid soon, and start pushing people writing integration tests too when we do code reviews.

pracucci:
Superseded by #1839.

pracucci:
I'm closing this PR because superseeded by #1839.

PR #2066, opened 2020-02-02
Allow to configure server gRPC keepalive

pracucci:
**What this PR does**:
Following up the PR https://github.com/weaveworks/common/pull/174 merged in `weaveworks/common`, we update the dependency in order to allow to configure gRPC keepalive settings.

Other than the new gRPC flags, the `weaveworks/common` update introduce the following changes, which look safe to me:
- Added support to `AWS_REGION` env var (if not set, the default behavior hasn't changed)
- `server.http-listen-address` and `server.http-conn-limit` and the same for gRPC (if not set, the default behaviour hasn't changed)

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -109,13 +109,29 @@ runtime_config:
 The `server_config` configures the HTTP and gRPC server of the launched service(s).
 
 ```yaml
+# HTTP server listen address.
+# CLI flag: -server.http-listen-address
+[http_listen_address: <string> | default = ""]
+
 # HTTP server listen port.
 # CLI flag: -server.http-listen-port
 [http_listen_port: <int> | default = 80]
 
+# Maximum number of simultaneous http connections, <=0 to disable
+# CLI flag: -server.http-conn-limit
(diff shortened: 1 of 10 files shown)


docs/configuration/config-file-reference.md
```diff
@@ -149,6 +165,31 @@ The `server_config` configures the HTTP and gRPC server of the launched service(
 # CLI flag: -server.grpc-max-concurrent-streams
 [grpc_server_max_concurrent_streams: <int> | default = 100]
 
+# The duration after which an idle connection should be closed. Default:
+# infinity
+# CLI flag: -server.grpc.keepalive.max-connection-idle
+[grpc_server_max_connection_idle: <duration> | default = 2562047h47m16.854775807s]
```
jtlisi: `2562047h47m16.854775807s` is the default but it looks so strange here haha. Don't think there is much we can do about it.

pracucci: Agree. We should have addressed it in `weaveworks/common`, like defaulting it to `0` in the config and then internally replacing it with the max value. We could keep it for a future improvement.

PR #2070, opened 2020-02-03
Use buffered channel to unblock lazySet's goroutine

### LINKED ISSUE #1832: Lazy querier can have tens of thousands of stuck goroutines
I suspect when something goes wrong the querier stops reading responses so all the remaining ones are abandoned on an error send.

Profile shows them stuck here:

https://github.com/cortexproject/cortex/blob/6f74b9c4bcffcce6e4149a354fc4cbf2dc390a2f/pkg/querier/lazy_querier.go#L28

image

pstibrany:
**What this PR does**: This PR makes `lazyQuerier` cancellable via context.

Fixes #1832

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/lazy_querier.go
+++ b/pkg/querier/lazy_querier.go
@@ -21,7 +21,9 @@ func newLazyQuerier(next storage.Querier) storage.Querier {
 }
 
 func (l lazyQuerier) Select(params *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
-	future := make(chan storage.SeriesSet)
+	// make sure there is space in the buffer, to unblock the goroutine and let it die even if nobody is
+	// waiting for the result yet (or anymore).
+	future := make(chan storage.SeriesSet, 1)
 	go func() {
 		set, _, err := l.next.Select(params, matchers...)
 		if err != nil {


pkg/querier/lazy_querier.go
```diff
@@ -75,28 +85,43 @@ func (e errSeriesSet) Err() error {
...
+		return nil
 	}
 	return s.next.At()
 }
 
 func (s lazySeriesSet) Err() error {
-	if s.next == nil {
-		s.next = <-s.future
+	if !s.waitForSeriesSet() {
+		return nil
```
pracucci: Is it correct return `nil` here? Shouldn't we return the context error?

pkg/querier/lazy_querier.go
```diff
@@ -11,26 +11,36 @@ import (
...
 		} else {
-			future <- set
+			res = set
+		}
+		select {
+		case future <- res:
+			// ok
+		case <-l.ctx.Done():
+			// stop
 		}
```
pracucci: Could you explain why we need this and what it does? Isn't `l.next.Select()` blocking?

pstibrany: You mean entire lazy set thing? There is comment about it in commit 76cc27152feeed512434a8a8506311f0e2099bbd, where it was added.

Now the problem with cancellation is that if one such querier returns error, another one will never get its `Next` method called, so this goroutine trying to send result would not be able to do anything. Now thinking about it, much simpler solution would be to make channel with buffer size of 1. I will change this PR to do that instead. Much simpler change, and fixes the same problem.

pstibrany:
I've dropped previous version of the fix in favor of simply adding a buffer of size one to the channel.

pstibrany:
> LGTM. Is it worth a mention in the `CHANGELOG`?

Doesn't really feel like it to me.

PR #2074, opened 2020-02-03
Fixing & Simplifying Azure Blob Download

### LINKED ISSUE #2024: Azure Blob - Invalid Chunk Checksum
I'm running Cortex: `quay.io/cortexproject/cortex:v0.5.0-rc1` in Kubernetes.
[cortex-yaml-redacted.txt](https://github.com/cortexproject/cortex/files/4098145/kube-yaml-redacted.txt)
[cassandra-statefulset-redacted.txt](https://github.com/cortexproject/cortex/files/4098147/cassandra-statefulset-redacted.txt)

I've been testing out a variety of backends. I managed to get chunk and index storage running against Azure's CosmosDB. Tuns out, CosmosDB costs a fortune as it charges per Keystore per Cassandra table a min of 400 RUs.

I switched the index store to use a (temporarily) in memory Cassandra Cluster and the chunk store to use Azure Blob.

I also began running with multiple instances, and started seeing `invalid chunk checksum `. The first time was my fault, as I didn't correctly have the ingesters running against consul and was still using memory.

I shut everything down, wiped the dbs, restarted everything, force deleted a pod, and ran into #1521. After wiping the consul and verifying that everything was working, around 5 minutes into collecting metrics, I started getting "Invalid Chunk Checksum" in my Grafana Queries again.

I scaled everything down, wiped it the DB and blob, and started it all over again. It worked fine for another few minutes or so, and then I started seeing more `invalid chunk checksum`. 

Is there any information about what causes this issue? Is anyone else running into this? Is there some way of ignoring just the sections of bad data? If I change the queries to use a smaller time window, I can still pull some of the data, but if I increase the window to a section that includes bad data, the entire query fails.

I'll try to collect logs next time I run into this, and attach them to this issue.

khaines:
**What this PR does**:  Main change was to fix an annoying bug noted in #2024 where the checksum of the downloaded blob was not valid. The underlying problem was that a buffer was passed by value instead of value. The secondary change here is to use a simpler API for downloading blobs from Azure. The previous used API used, opened multiple connections to download blobs in parallel, however this is somewhat unneeded. Even with BigChunk encoding, a given chunk is still measured in KB, plus the chunk store package already manages parallel downloading of separate chunk files. 

Fixes #2024

**Checklist**
- [ x ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/pkg/chunk/azure/blob_storage_client.go
+++ b/pkg/chunk/azure/blob_storage_client.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net/url"
 	"strings"
 	"time"
@@ -85,14 +86,14 @@ func (b *BlobStorage) getChunk(ctx context.Context, decodeContext *chunk.DecodeC
 
 	// download the contents & read into a buffer
 	bodyStream := downloadResponse.Body(azblob.RetryReaderOptions{MaxRetryRequests: b.cfg.MaxRetries})
-	buf := bytes.Buffer{}
-	_, err = buf.ReadFrom(bodyStream)
+	defer bodyStream.Close()
 
+	buf, err := ioutil.ReadAll(bodyStream)
 	if err != nil {
 		return chunk.Chunk{}, err
 	}
 
-	if err := input.Decode(decodeContext, buf.Bytes()); err != nil {
+	if err := input.Decode(decodeContext, buf); err != nil {
 		return chunk.Chunk{}, err
 	}
 
--- a/pkg/chunk/azure/blob_storage_client.go
+++ b/pkg/chunk/azure/blob_storage_client.go
@@ -77,14 +77,22 @@ func (b *BlobStorage) getChunk(ctx context.Context, decodeContext *chunk.DecodeC
 		return chunk.Chunk{}, err
 	}
 
-	buf := make([]byte, 0, b.cfg.DownloadBufferSize)
-
-	err = azblob.DownloadBlobToBuffer(ctx, blockBlobURL.BlobURL, 0, 0, buf, azblob.DownloadFromBlobOptions{})
+	// Request access to the blob
+	downloadResponse, err := blockBlobURL.Download(ctx, 0, azblob.CountToEnd, azblob.BlobAccessConditions{}, false)
 	if err != nil {
 		return chunk.Chunk{}, err
 	}
 
-	if err := input.Decode(decodeContext, buf); err != nil {
+	// download the contents & read into a buffer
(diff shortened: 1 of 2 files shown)


pkg/chunk/azure/blob_storage_client.go
```diff
@@ -77,14 +77,22 @@ func (b *BlobStorage) getChunk(ctx context.Context, decodeContext *chunk.DecodeC
...
-	buf := make([]byte, 0, b.cfg.DownloadBufferSize)
+	// Request access to the blob
+	downloadResponse, err := blockBlobURL.Download(ctx, 0, azblob.CountToEnd, azblob.BlobAccessConditions{}, false)
+	if err != nil {
+		return chunk.Chunk{}, err
+	}
+
+	// download the contents & read into a buffer
+	bodyStream := downloadResponse.Body(azblob.RetryReaderOptions{MaxRetryRequests: b.cfg.MaxRetries})
+	buf := bytes.Buffer{}
```
gouthamve: I don't think I saw this usage before, maybe just use `bytes.NewBuffer(nil)`?

pracucci: If all we need is to read the entire response in a `[]byte`, then I believe `ioutil.ReadAll(bodyStream)` is preferable.

pkg/chunk/azure/blob_storage_client.go
```diff
@@ -77,14 +77,22 @@ func (b *BlobStorage) getChunk(ctx context.Context, decodeContext *chunk.DecodeC
...
 
-	buf := make([]byte, 0, b.cfg.DownloadBufferSize)
+	// Request access to the blob
+	downloadResponse, err := blockBlobURL.Download(ctx, 0, azblob.CountToEnd, azblob.BlobAccessConditions{}, false)
+	if err != nil {
+		return chunk.Chunk{}, err
+	}
+
+	// download the contents & read into a buffer
+	bodyStream := downloadResponse.Body(azblob.RetryReaderOptions{MaxRetryRequests: b.cfg.MaxRetries})
```
pracucci: I believe we need to add `defer bodyStream.Close()` right after this line, otherwise the `io.ReadCloser` is never closed. Am I missing anything?

khaines: yes, yes we do. Thank you for spotting this

pkg/chunk/azure/blob_storage_client.go
khaines: I agree with @pracucci ... ReadAll makes this a bit more tidy to read plus matches all we need to do here.

PR #2075, opened 2020-02-04
Flusher target to flush WAL

codesome:
This is part of the WAL work where during scaling down if the `/shutdown` endpoint was not hit to flush the chunks, this flusher target would be used as a job to flush those chunks.

Supersedes https://github.com/cortexproject/cortex/pull/1747

I have already tested and according to the logs I can say that it is being flushed properly. (I was not able to get the metrics, investigating it)

**Checklist**
- [ ] Tests updated
- [x] Documentation added


CHANGELOG.md
```diff
@@ -8,7 +8,10 @@
   * `--store.min-chunk-age` has been removed
   * `--querier.query-store-after` has been added in it's place.
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
-  * `--experimental.distributor.user-subring-size`
```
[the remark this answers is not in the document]
codesome: Mistake :) will add it back

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	elapsed := time.Since(start)
+
+	level.Info(util.Logger).Log("msg", "recovery from WAL completed", "time", elapsed.String())
+
+	err := f.flushAllUsers(userStates)
+
+	// Sleeping to give a chance to Prometheus
+	// to collect the metrics.
+	time.Sleep(1 * time.Minute)
```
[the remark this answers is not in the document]
codesome: That sleep comes inside lifecycler. We won't initialize lifecycler here.

codesome:
I have simplified this quite a bit to not reimplement existing stuff. I have tested it and seems to be working (from the logs). Flushing happens before starting the server (in the init phase), so, we are not able to capture any metrics. I am waiting for https://github.com/cortexproject/cortex/pull/2119 to get in so that I can start the flushing in parallel with starting the server.

gouthamve:
@codesome Can you rebase and make it a service now that #2166 is merged?

codesome:
@gouthamve rebased and now the metrics are visible with the async start from the new services.

@pstibrany I have made some changes in the services for the flusher target to work as a job.

pkg/util/services/basic_service.go
```diff
@@ -278,6 +278,9 @@ func (b *BasicService) awaitState(ctx context.Context, expectedState State, ch c
 
 		// if service has failed, include failure case in the returned error.
 		if failure := b.FailureCase(); failure != nil {
+			if failure == ErrStopCortex {
```
gouthamve: This doesn't make sense. The services package should be agnostic of Cortex, but looks like that might not be the case with this.

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,102 @@
...
+	clientConfig client.Config,
+	chunkStore ingester.ChunkStore,
+	registerer prometheus.Registerer,
+) (*Flusher, error) {
+
+	ingesterConfig.WALConfig.Dir = cfg.WALDir
+	ingesterConfig.ConcurrentFlushes = cfg.ConcurrentFlushes
+	ingesterConfig.FlushOpTimeout = cfg.FlushOpTimeout
+
+	ing, err := ingester.NewForFlusher(ingesterConfig, clientConfig, chunkStore, registerer)
```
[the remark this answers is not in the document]
codesome: WAL replay, nope, but it will be running when we are flushing, so these metrics will be scraped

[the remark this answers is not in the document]
codesome: Moved it to `starting` anyway

pkg/cortex/modules.go
```diff
@@ -366,6 +382,22 @@ func (t *Cortex) initIngester(cfg *Config) (serv services.Service, err error) {
...
+		cfg.Ingester,
+		cfg.IngesterClient,
+		t.store,
+		prometheus.DefaultRegisterer,
+	)
+	if err != nil {
+		return
+	}
+
+	t.server.HTTP.Path("/ready").Handler(http.HandlerFunc(t.flusher.ReadinessHandler))
```
pstibrany: We already set global /ready handler at cortex.Run. Can you move flusher readiness handler there as well? Similar to how ingester's readiness handler is hooked.

pkg/ingester/ingester.go
```diff
@@ -217,6 +217,55 @@ func (i *Ingester) starting(ctx context.Context) error {
...
+	return i, nil
+}
+
+func (i *Ingester) startingForFlusher(ctx context.Context) error {
+	i.flushQueuesDone.Add(i.cfg.ConcurrentFlushes)
+	for j := 0; j < i.cfg.ConcurrentFlushes; j++ {
+		i.flushQueues[j] = util.NewPriorityQueue(i.metrics.flushQueueLength)
+		go i.flushLoop(j)
+	}
+
```
pstibrany: This same code is in `starting`, would it make sense to extract into new method (eg. startFlushLoops)?

pkg/ingester/ingester.go
```diff
@@ -217,6 +217,55 @@ func (i *Ingester) starting(ctx context.Context) error {
 	return nil
 }
 
+// NewForFlusher constructs a new Ingester to be used by flusher target.
```
pstibrany: Can the comment explain what is the difference from "normal" ingester?

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (f *Flusher) running(ctx context.Context) error {
+	if err := f.ingester.StartAsync(ctx); err != nil {
+		return errors.Wrap(err, "start ingester")
+	}
+	if err := f.ingester.AwaitRunning(ctx); err != nil {
+		return errors.Wrap(err, "awaing running ingester")
+	}
+
```
pstibrany: You can use `services.StartAndAwaitRunning` here. Also, I think this can go to `Starting` too, normally service starts its own subservices in Starting state.

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,113 @@
...
+	f.ingester.Flush()
+
+	// Sleeping to give a chance to Prometheus
+	// to collect the metrics.
+	level.Info(util.Logger).Log("msg", fmt.Sprintf("sleeping for %s to give chance for collection of metrics", postFlushSleepTime.String()))
+	time.Sleep(postFlushSleepTime)
+
+	f.ingester.StopAsync()
```
pstibrany: Same for this part... can go do Stopping, incl. returning of the error. Similarly, `service.StopAndAwaitTerminated` can be used to simplify the code.

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,113 @@
...
+}
+func (f *Flusher) stopping() error {
+	// Nothing to do here.
+	return nil
+}
+
+// ReadinessHandler returns 204 always.
+func (f *Flusher) ReadinessHandler(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(http.StatusNoContent)
+}
```
pstibrany: Oh, I think readiness handler in Run would cover this already :-)

codesome: Yeah, it does, will remove this

pkg/cortex/modules.go
codesome: Didn't notice that change with services. Will remove that.

codesome:
@pstibrany addressed all your comments. Also, I put the replay of WAL in the starting function to be in sync with #2222.

pstibrany:
Please rebase on master. Cortex now returns error from `Run` method in cortex.go if any service has failed, but it should ignore `util.ErrStopCortex` while doing so.

pkg/cortex/modules.go
```diff
@@ -177,6 +184,15 @@ func (m *moduleName) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 
+func (m moduleName) IsJob() bool {
+	switch m {
+	case Flusher:
+		return true
+	}
+
+	return false
+}
+
```
pstibrany: Is this method used anywhere? If not, should it be removed?

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,104 @@
...
+func (f *Flusher) running(ctx context.Context) error {
+	if err := services.StartAndAwaitRunning(ctx, f.ingester); err != nil {
+		return errors.Wrap(err, "start and await running ingester")
+	}
+
+	f.ingester.Flush()
+
+	// Sleeping to give a chance to Prometheus
+	// to collect the metrics.
+	level.Info(util.Logger).Log("msg", fmt.Sprintf("sleeping for %s to give chance for collection of metrics", postFlushSleepTime.String()))
```
pstibrany:
```suggestion
	level.Info(util.Logger).Log("msg", "sleeping to give chance for collection of metrics", "duration", postFlushSleepTime.String())
```

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,104 @@
...
+		chunkStore:     chunkStore,
+		registerer:     registerer,
+	}
+	f.Service = services.NewBasicService(f.starting, f.running, f.stopping)
+	return f, nil
+}
+
+func (f *Flusher) starting(ctx context.Context) error {
+	// WAL replay happens here. We have it in the starting function and not New
+	// so that metrics can be collected in parallel with WAL replay.
```
pstibrany: It seems to me that entire ingester creation, its starting and stopping can be in Flusher's "running" method, and then there is no need to have `ingester` field in `Flusher`. (We can then also drop `starting` method of flusher)

If you want to keep it as is, that's fine too.

pkg/flusher/flusher.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := services.StopAndAwaitTerminated(ctx, f.ingester); err != nil {
+		return errors.Wrap(err, "stop and await terminated ingester")
+	}
+	return util.ErrStopCortex
+}
+
+func (f *Flusher) stopping(_ error) error {
+	// Nothing to do here.
+	return nil
+}
```
pstibrany: It's ok to just pass `nil` to `services.NewBasicService` as last function.

pkg/ingester/ingester.go
```diff
@@ -208,12 +208,64 @@ func (i *Ingester) starting(ctx context.Context) error {
...
+		clientConfig: clientConfig,
+		metrics:      newIngesterMetrics(registerer, true),
+		chunkStore:   chunkStore,
+		flushQueues:  make([]*util.PriorityQueue, cfg.ConcurrentFlushes),
+	}
 
+	// Should be a noop WAL.
+	cfg.WALConfig.WALEnabled = false
+	var err error
+	i.wal, err = newWAL(cfg.WALConfig, i.userStates.cp)
```
pstibrany: Is `i.userStates` set at this point?

pkg/ingester/ingester.go
```diff
@@ -208,12 +208,64 @@ func (i *Ingester) starting(ctx context.Context) error {
...
+	if err != nil {
+		return nil, err
+	}
```
pstibrany: Can't we just set `i.wal = &noopWal{}` here?

pkg/cortex/modules.go
codesome: Dead code from old iterations. Will remove it.

pkg/flusher/flusher.go
codesome: SGTM. And I didn't know I could put `nil` for starting etc methods. Now we will end up having only the running method.

pkg/ingester/ingester.go
codesome: GOOD point, it is not. I tested this already and it didn't panic. I will pass a `nil` here as we expect a noopWAL here which has no operations.

pkg/ingester/ingester.go
codesome: LGTM

PR #2082, opened 2020-02-05
Ignore expired requests in frontend scheduling

owen-d:
**What this PR does**:
This PR does two things to improve scheduling for the per tenant queues in the query frontend. This is important for two reasons:
1)  To avoid certain cases where expired requests can hamper the fairness of scheduling in the frontend across multiple tenants. 
2) To short circuit expired requests and prevent them from being pulled by the querier workers.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -256,11 +256,6 @@ func (f *Frontend) Process(server Frontend_ProcessServer) error {
 			return err
 		}
 
-		// a nil request and nil error means the chosen tenant queue is empty of unexpired requests
-		if req == nil {
-			continue
-		}
-
 		// Handle the stream sending & receiving on a goroutine so we can
 		// monitoring the contexts in a select and cancel things appropriately.
 		resps := make(chan *ProcessResponse, 1)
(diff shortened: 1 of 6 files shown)


pkg/querier/frontend/frontend.go
```diff
@@ -345,18 +350,48 @@ func (f *Frontend) getNextRequest(ctx context.Context) (*request, error) {
 			continue
 		}
 
-		request := <-queue
+		/*
```
pracucci: I believe this entire new logic is over complicated for the use case. Moreover, returning `nil, nil` looks an anti-pattern to me: the `getNextRequest()` contract is that it "block if there are no requests" but with this PR changes there are conditions under which we do "return in some cases if there are no requests".

I'm wondering if the new logic could be implemented as:
1. Add a goto label before `for len(f.queues) == 0 && ctx.Err() == nil {`
2. Within the `for {`, if `request.originalCtx.Err() != nil` we do `continue`
3. We replace `panic("should never happen")` with `goto <label we've added at point 1>`

Makes sense?

owen-d: I like to goto idea -- we can maintain the same contract.

pracucci: @owen-d Do you have any update on this?

owen-d: @pracucci It should be updated. If you have small suggestions, feel free to add them, but I think this addresses the anti pattern you described earlier and aligns the function's contract to what it was before this pr.

PR #2088, opened 2020-02-06
Update prometheus to v2.16.0

pstibrany:
**What this PR does**: This PR updates Prometheus dependency to [v2.16.0 (b90be6f32a)](https://github.com/prometheus/prometheus/releases/tag/v2.16.0) (was -rc.0 and -rc.1 before). This is required to finish work on issue #2003. 

Recent Prometheus changes introduced some incompatibilities:
- `promql.EngineOpts` struct had its [`MaxConcurrent` field removed](https://github.com/prometheus/prometheus/pull/6712), and functionality [replaced with active query tracker](https://github.com/prometheus/prometheus/pull/6701). Cortex' querier therefore enables the tracker to keep `-querier.max-concurrent` flag still working. Since tracker needs a directory to write files to, Cortex uses `./active-query-tracker` by default. **Update:** If directory is explicitly set to empty string by using `-querier.active-query-tracker-dir=""`, Active Query Tracker is disabled. (Previously, random temp directory was created for empty dir)

- [Changes in rulefmt package](https://github.com/prometheus/prometheus/pull/6533) modified `RuleNode` type, which replaced previously used `Rule`. Unfortunately, while this provides better error reporting when doing deserialization, it makes it impossible to serialize this structure back to YAML because of [problem in yaml.v3 package](https://github.com/go-yaml/yaml/issues/575). Until we find a better solution, I've made a copy of rulefmt package before the change into Cortex (as "legacy_rulefmt"). It's a very tiny package, but admittedly, this is not a good solution.

- Prometheus also added `SelectSorted` method to `storage.Querier` interface, but fortunately all our queriers already sorted their output in `Select` method anyway, so all implementations just call that.

- `promql.ParseErr` [has lost its `Line` and `Pos` fields](https://github.com/prometheus/prometheus/pull/6634) and now only has `PositionRange` with start and end location. We use this in  `/api/prom/validate_expr` API endpoint to report query syntax problems. This endpoint doesn't seem to be [documented](https://cortexmetrics.io/docs/apis/) anywhere and was added by request from Grafana Observability team. It doesn't seem to be used as of now. If it's not used anywhere else, it might be possible to remove it completely (it's not provided by Prometheus). For now I've changed endpoint to return `start` and `end` position instead of `line` and `col` fields.  /cc @davkal. **Update: this functionality has been completely removed by #2152**.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/distributor/http_server.go
```diff
@@ -92,17 +92,20 @@ func (d *Distributor) ValidateExprHandler(w http.ResponseWriter, r *http.Request
 	// and the generated error string omits the line entirely. But we
 	// want to report line numbers consistently, no matter how many
 	// lines there are (starting at 1).
-	if parseErr.Line == 0 {
-		parseErr.Line = 1
-	}
+	// Not available from Prometheus at the moment.
```
jtlisi: I think it's fine to remove the commented out code entirely here. I don't think we will bring it back any time soon in this case.

pkg/ruler/ruler_test.go
```diff
@@ -55,10 +57,16 @@ func defaultRulerConfig(store rules.RuleStore) (Config, func()) {
 }
 
 func newTestRuler(t *testing.T, cfg Config) *Ruler {
+	dir, err := ioutil.TempDir("", t.Name())
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
```
jtlisi: I think we need to call `os.RemoveAll(dir)` after the test has completed. Similar to the temp dir cleanup returned in the `defaultRulerConfig` function.

pstibrany: Fixed this. I've just realized that tracker has no way to stop and close the mmaped file. So with too many tests using the ruler or querier, we will start to run out of file descriptors :-(

pstibrany: Fortunately `go test` compiles and run tests for each package separately, so the risk is lower.

pkg/querier/querier.go
```diff
@@ -36,6 +38,13 @@ type Config struct {
 	// step if not specified.
 	DefaultEvaluationInterval time.Duration
 
+	// Directory for ActiveQueryTracker. If empty, temp dir is created, which defeats the purpose of
```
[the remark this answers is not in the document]
pstibrany: I should test that indeed. What should we do if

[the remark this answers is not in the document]
pstibrany: > level=info ts=2020-02-17T15:54:48.040599Z caller=query_logger.go:79 msg="These queries didn't finish in prometheus' last run:" queries="[{\"query\":\"label_replace(cortex_load_generator_sine_wave, \\\"dst\\\", \\\"$1\\\", \\\"wave\\\", \\\"([^0]+)\\\")\",\"timestamp_sec\":1581954860},{\"query\":\"label_replace(cortex_load_generator_sine_wave, \\\"dst\\\", \\\"$1\\\", \\\"wave\\\", \\\"([^0]+)\\\")\",\"timestamp_sec\":1581954860}]"


This was logged on startup. All the escaping makes it difficult to read, but at least it works.

pkg/querier/ingester_streaming_queryable.go
```diff
@@ -111,3 +111,8 @@ func (q *ingesterStreamingQuerier) Select(sp *storage.SelectParams, matchers ...
 
 	return newConcreteSeriesSet(serieses), nil, nil
 }
+
+func (q *ingesterStreamingQuerier) SelectSorted(sp *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
```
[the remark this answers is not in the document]
pstibrany: You're right. I'll do that everywhere.

pstibrany:
I've rebased and addressed your feedback. Now instead of creating temp dir for Active Query Tracker, we disable it, if no dir is configured. Dir defaults to `./active-query-tracker`.

pkg/querier/querier.go
```diff
@@ -36,6 +36,12 @@ type Config struct {
 	// step if not specified.
 	DefaultEvaluationInterval time.Duration
 
+	// Directory for ActiveQueryTracker. If empty, ActiveQueryTracker will be disabled and MaxConcurrent will not be applied (!).
+	// ActiveQueryTracker logs queries that were active during the last crash, but logs them on the next startup.
+	// However, we need to use active query tracker, otherwise we cannot limit Max Concurrent queries in the PromQL
+	// engine.
+	ActiveQueryTrackerDir string `yaml:"active_query_tracking_dir"`
```
pracucci: `active_query_tracking_dir` > `active_query_tracker_dir` to be consistent with the CLI flag.

pstibrany:
> Withdrawing my approval because we found out that the read path is failing with the error `missing port in address`.

There is now workaround for this, basically we fake the address in `RemoteAddr` field. Since we don't set any QueryLogger, this value is unused.

I've also sent [PR to Prometheus](https://github.com/prometheus/prometheus/pull/6840) to be less strict about missing or invalid `RemoteAddr` value.

PR #2089, opened 2020-02-06
Adds FromContext to spanlogger package.

cyriltovena:

This allows to safely logs data inside a parent span without creating a new one, and to just log to the logger if no parent span exists.

### DIFF
--- a/pkg/util/spanlogger/spanlogger.go
+++ b/pkg/util/spanlogger/spanlogger.go
@@ -31,6 +31,8 @@ func New(ctx context.Context, method string, kvps ...interface{}) (*SpanLogger, 
 	return logger, ctx
 }
 
+// FromContext returns a span logger using the current parent span.
+// If there is no parent span, the Spanlogger will only log to stdout.
 func FromContext(ctx context.Context) *SpanLogger {
 	sp := opentracing.SpanFromContext(ctx)
 	if sp == nil {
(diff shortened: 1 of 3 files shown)


pkg/util/spanlogger/spanlogger.go
```diff
@@ -31,6 +31,17 @@ func New(ctx context.Context, method string, kvps ...interface{}) (*SpanLogger,
 	return logger, ctx
 }
 
+func FromContext(ctx context.Context) *SpanLogger {
+	sp := opentracing.SpanFromContext(ctx)
+	if sp == nil {
+		return noop
```
gouthamve: `noop` makes me think that the logger will be noop as well. Can we just remove that variable and just return `&SpanLogger{Logger: util.Logger, Span: noopSpan}`

cyriltovena: sure !

pkg/util/spanlogger/spanlogger.go
```diff
@@ -31,6 +31,20 @@ func New(ctx context.Context, method string, kvps ...interface{}) (*SpanLogger,
 	return logger, ctx
 }
 
+func FromContext(ctx context.Context) *SpanLogger {
```
pracucci: May you add a comment to this exported function, please?

cyriltovena: Shouldn't the linter get catch this ?

PR #2091, opened 2020-02-07
Add additional buckets to Bigtable request duration histogram

slim-bean:
**What this PR does**:

We see bigtable requests longer than 1s and it's hard to get a good idea with the current histogram buckets just how long these requests are taking, adding a few more buckets.

Signed-off-by: Edward Welch <edward.welch@grafana.com>

### DIFF
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -52,9 +52,9 @@ var (
 		Name:      "dynamo_request_duration_seconds",
 		Help:      "Time spent doing DynamoDB requests.",
 
-		// DynamoDB latency seems to range from a few ms to a few sec and is
-		// important.  So use 8 buckets from 128us to 2s.
-		Buckets: prometheus.ExponentialBuckets(0.000128, 4, 8),
+		// DynamoDB latency seems to range from a few ms to a several seconds and is
+		// important.  So use 9 buckets from 1ms to just over 1 minute (65s).
+		Buckets: prometheus.ExponentialBuckets(0.001, 4, 9),
 	}, []string{"operation", "status_code"}))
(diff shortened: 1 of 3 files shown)


bboreham:
You might want to make it consistent with the equivalent on AWS
https://github.com/cortexproject/cortex/blob/master/pkg/chunk/aws/dynamodb_storage_client.go#L55-L57

slim-bean:
Our general problem is not getting any granularity above 1s and seeing latencies via traces up to 20s.

To use the same spacing as AWS but also get a bucket to help solve the problem, we would need to go to 10 buckets:

`ExponentialBuckets(0.000128, 4, 10)` => [0.000128 0.000512 0.002048 0.008192 0.032768 0.131072 0.524288 2.097152 8.388608 33.554432]

I could update both AWS and bigtable with this config?

WDYT @bboreham ?

bboreham:
TBH now I look at .128ms is a weirdly complicated value, and we don’t see much below 1ms so I prefer your way. 

Still, I see value in making them consistent.

If you have trace-level detail is there much benefit in counting them as well?  I don’t mind; just curious.

slim-bean:
Updated both to be the same!

> If you have trace-level detail is there much benefit in counting them as well? I don’t mind; just curious.

This was from digging down into a Jaeger trace after processing some query logs looking for slow queries.  Having the metric give us a few more buckets on the tall end should help make our dashboards give us a better idea if the latency is a couple seconds or 20+ seconds without having to find and dig through a trace.  Mostly this saves the oncall person from having to do the above steps and digging into the trace if we can easily see that bigtable is showing huge latencies.

slim-bean:
>  but I would suggest to do it for Cassandra as well (just increase from 6 to 9 buckets),

Done!

PR #2092, opened 2020-02-07
runtimeconfig: initialize overridesReloadSuccess gauge in success state

jgehrcke:
From the commit message:
```
There seem to be code paths where the overridesReloadSuccess gauge
gets initialized, but `NewRuntimeConfigManager()` never is called
(absence of both log lines: "failed to load config", "runtime config
disabled: file not specified").

promauto.NewGauge() seems to initialize the gauge with the value 0
(that seems to be common in Prometheus client libs).

In the current code, the value 0 is associated with the error
state, and the value 1 is associated with the success state.

Likewise, https://github.com/grafana/cortex-jsonnet/ defines
an alert that fires when this gauge is set to 0 for an extended
period of time.

That is, when `NewRuntimeConfigManager()` never gets called
the gauge indicates a false negative error state.

Starting in the success state is a pragmatic way to solve that
problem.

It might be better to use a counter for individual error or
-- if a gauge is preferred -- one could introduce a third
state referring to "initialized".
```
(work in progress, let's see what CI says)

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated -- note: all tests seem to have passed. I think I need some guidance if I am supposed to add a test.
- [ ] Documentation added

### DIFF
--- a/pkg/util/runtimeconfig/manager.go
+++ b/pkg/util/runtimeconfig/manager.go
@@ -17,6 +17,13 @@ var overridesReloadSuccess = promauto.NewGauge(prometheus.GaugeOpts{
 	Help: "Whether the last config reload attempt was successful.",
 })
 
+// Initialize gauge in success state (1). That is, require
+// explicit transitioning into the error state (0). That helps
+// e.g. when `NewRuntimeConfigManager()` never gets called.
+func init() {
+	overridesReloadSuccess.Set(1)
+}
+
 // Loader loads the configuration from file.
 type Loader func(filename string) (interface{}, error)
 
--- a/pkg/util/runtimeconfig/manager.go
+++ b/pkg/util/runtimeconfig/manager.go
@@ -44,18 +44,18 @@ type Manager struct {
 }
 
 // NewRuntimeConfigManager creates an instance of Manager and starts reload config loop based on config
-func NewRuntimeConfigManager(cfg ManagerConfig, metricsRegisterer prometheus.Registerer) (*Manager, error) {
+func NewRuntimeConfigManager(cfg ManagerConfig, registerer prometheus.Registerer) (*Manager, error) {
 	mgr := Manager{
 		cfg:  cfg,
 		quit: make(chan struct{}),
 		configLoadSuccess: prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: "cortex_overrides_last_reload_successful",
-			Help: "Whether the last config reload attempt was successful (0: error, 1: successful).",
+			Help: "Whether the last config reload attempt was successful.",
 		}),
 	}
 
-	if metricsRegisterer != nil {
-		metricsRegisterer.MustRegister(mgr.configLoadSuccess)
+	if registerer != nil {
+		registerer.MustRegister(mgr.configLoadSuccess)
 	}
 
 	if cfg.LoadPath != "" {
--- a/pkg/util/runtimeconfig/manager.go
+++ b/pkg/util/runtimeconfig/manager.go
@@ -40,7 +40,7 @@ type Manager struct {
 	configMtx sync.RWMutex
 	config    interface{}
 
-	overridesReloadSuccess prometheus.Gauge
+	configLoadSuccess prometheus.Gauge
 }
 
 // NewRuntimeConfigManager creates an instance of Manager and starts reload config loop based on config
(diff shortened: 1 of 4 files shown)


jgehrcke:
`make test` passed locally.

jgehrcke:
Would appreciate some initial feedback. Thank you! :)

pstibrany:
Thanks for your contribution.

`RuntimeConfigManager` is initialized by components that need it (ingester, querier, distributor, query-frontend, ruler, or when using "all"), but there are some components that don't use it currently (table-manager, alert-manager or compactor).

Does that match what you've observed?

jgehrcke:
> Does that match what you've observed?

Yes, I think so!

> We should not export the metric at all when not used. 

Agreed, that's a cleaner solution.

jgehrcke:
> My suggestion is to use the pattern we've recently started using:

The commit starting with the message `runtimeconfig: register cortex_overrides_last_reload_successful dynamically` should do that. Thanks for the suggestions!

I have added more commits on top of that proposing variable renames in code, and after all also a rename of the gauge itself. I am not sure how this affects our thinking w.r.t. backwards compatibility, but I went ahead and worded a corresponding changelog entry. If desired we can drop the last few commits again, or treat that independently in a separate discussion/PR.

pkg/util/runtimeconfig/manager_test.go
```diff
@@ -204,8 +209,8 @@ func TestOverridesManager_StopClosesListenerChannels(t *testing.T) {
 			return val, nil
 		},
 	}
-
-	overridesManager, err := NewRuntimeConfigManager(overridesManagerConfig)
+	registry := prometheus.NewRegistry()
```
pstibrany: Typically in tests we just pass nil (unless testing metrics), and make sure that New function can accept nil.

jgehrcke: thanks, done, commit pushed

jgehrcke:
Thanks for the quick feedback @pstibrany 

> Personally, I would suggest keeping previous logic and name.

Done. I have removed the more controversial commits! 

> Problem with this rename is that many people (including us at Grafana Labs) have alerts based on existing name and will need to change them. That would be fair if new new name provided improved functionality, but I’m not sure it does.

Well, it's totally ok and I respect the decision :). My thinking, though, was that `cortex_overrides_last_reload_successful` is a pretty nonexpressive name and it will only get harder to change it in the future.

About the gauge itself and its meaning:

I think in the long run we probably agree in the point of view that an error counter would be the more intuitive, simple, and complete solution. With the gauge we actually don't keep track of when exactly (and, with that, how frequently) the config load operation fails if it fails. Again, if this gauge is at 0 we actually can't even know for sure that a config load was attempted (with the current patch it's just way more likely :)).

Thanks again, looking forward to reading your feedback about the current patch.

pkg/util/runtimeconfig/manager.go
```diff
@@ -45,13 +39,23 @@ type Manager struct {
...
 
 // NewRuntimeConfigManager creates an instance of Manager and starts reload config loop based on config
-func NewRuntimeConfigManager(cfg ManagerConfig) (*Manager, error) {
+func NewRuntimeConfigManager(cfg ManagerConfig, metricsRegisterer prometheus.Registerer) (*Manager, error) {
 	mgr := Manager{
 		cfg:  cfg,
 		quit: make(chan struct{}),
+		configLoadSuccess: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_overrides_last_reload_successful",
+			Help: "Whether the last config reload attempt was successful (0: error, 1: successful).",
```
pstibrany: > `(0: error, 1: successful)`

can be removed, this is common enough

pkg/util/runtimeconfig/manager.go
```diff
@@ -45,13 +39,23 @@ type Manager struct {
 
 	configMtx sync.RWMutex
 	config    interface{}
+
+	configLoadSuccess prometheus.Gauge
 }
 
 // NewRuntimeConfigManager creates an instance of Manager and starts reload config loop based on config
-func NewRuntimeConfigManager(cfg ManagerConfig) (*Manager, error) {
+func NewRuntimeConfigManager(cfg ManagerConfig, metricsRegisterer prometheus.Registerer) (*Manager, error) {
```
pracucci: `metricsRegisterer` > `registerer`

jgehrcke: done, commit pushed

pkg/util/runtimeconfig/manager.go
jgehrcke: done, commit pushed

PR #2093, opened 2020-02-07
add option to configure approved password authenticators in Cassandra

jtlisi:
This PR optionally allows users to specify Authenticators for Cassandra that are not including in the default approved `gocql` list. The same password workflow that is used in `gocql` is used by the custom authenticator. It will just utilize a user provided set of approved authenticators.

**Checklist**
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/cassandra/storage_client.go
+++ b/pkg/chunk/cassandra/storage_client.go
@@ -29,7 +29,7 @@ type Config struct {
 	Auth                     bool                `yaml:"auth,omitempty"`
 	Username                 string              `yaml:"username,omitempty"`
 	Password                 string              `yaml:"password,omitempty"`
-	CustomAuthenticators     flagext.StringSlice `yaml:"custom_authenticator"`
+	CustomAuthenticators     flagext.StringSlice `yaml:"custom_authenticators"`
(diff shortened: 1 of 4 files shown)


pkg/chunk/cassandra/storage_client.go
```diff
@@ -12,27 +12,29 @@ import (
...
+	Consistency              string              `yaml:"consistency,omitempty"`
+	ReplicationFactor        int                 `yaml:"replication_factor,omitempty"`
+	DisableInitialHostLookup bool                `yaml:"disable_initial_host_lookup,omitempty"`
+	SSL                      bool                `yaml:"SSL,omitempty"`
+	HostVerification         bool                `yaml:"host_verification,omitempty"`
+	CAPath                   string              `yaml:"CA_path,omitempty"`
+	Auth                     bool                `yaml:"auth,omitempty"`
+	Username                 string              `yaml:"username,omitempty"`
+	Password                 string              `yaml:"password,omitempty"`
+	CustomAuthenticators     flagext.StringSlice `yaml:"custom_authenticator"`
```
pracucci: What do you think to rename `custom_authenticator` into `custom_authenticators` given it accepts a list? If so, remember to update the entry in the `CHANGELOG.md`.

PR #2095, opened 2020-02-09
Run single table-manager among multiple single binary cortex instances.

VineethReddy02:
**What this PR does**: When multiple single binary cortex instances are running every cortex instance is running it's own table-manager. This PR aims to achieve single table-manager at any point in time irrespective of any number of cortex instances running.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

@gouthamve

### DIFF
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -457,6 +457,10 @@ func (t *Cortex) initTableManager(cfg *Config) error {
 	if err != nil {
 		return err
 	}
+
+	t.tableManager.TargetModule = cfg.Target.String()
+	t.tableManager.Ring = t.ring
+	t.tableManager.Lifecycler = cfg.Ingester.LifecyclerConfig
 	t.tableManager.Start()
 	return nil
 }
@@ -638,7 +642,7 @@ var modules = map[moduleName]module{
 	},
 
 	TableManager: {
-		deps: []moduleName{Server},
+		deps: []moduleName{Server, Ring, Ingester},
 		init: (*Cortex).initTableManager,
 		stop: (*Cortex).stopTableManager,
 	},
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/rand"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -18,6 +19,8 @@ import (
 	"github.com/weaveworks/common/mtime"
 
 	"github.com/cortexproject/cortex/pkg/util"
+
+	"github.com/cortexproject/cortex/pkg/ring"
 )
 
 const (
@@ -116,13 +119,17 @@ func (cfg *ProvisionConfig) RegisterFlags(argPrefix string, f *flag.FlagSet) {
 
 // TableManager creates and manages the provisioned throughput on DynamoDB tables
 type TableManager struct {
-	client       TableClient
-	cfg          TableManagerConfig
-	schemaCfg    SchemaConfig
-	maxChunkAge  time.Duration
-	done         chan struct{}
-	wait         sync.WaitGroup
-	bucketClient BucketClient
+	Ring               *ring.Ring
+	client             TableClient
+	cfg                TableManagerConfig
+	schemaCfg          SchemaConfig
+	maxChunkAge        time.Duration
+	done               chan struct{}
+	wait               sync.WaitGroup
+	bucketClient       BucketClient
+	Lifecycler         ring.LifecyclerConfig
+	tableManagerStatus bool
+	TargetModule       string
 }
 
 // NewTableManager makes a new TableManager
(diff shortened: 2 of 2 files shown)


VineethReddy02:
The scope of this functionality is for a single binary cortex. I was thinking if I add my functionality/business logic at modules.go it is confined to single binary instance. Correct me if am wrong or feel free to let me know if there are any alternatives in doing this.

VineethReddy02:
The scope of this functionality is for a single binary cortex. I was thinking if I add my functionality/business logic at modules.go it is confined to a single binary instance. Correct me if am wrong and also feel free to let me know if there are any alternatives in doing this.

jtlisi:
I also don't think this kind of logic belongs in `modules.go`. Can you refactor to push this logic into the table manager.

You can pass the table manager a ring similar to how you would pass a ring to the distributor. Then when the table manager runs, it can check the ring to see if it has the lowest token and run. Otherwise it will be a noop. Do you think something like this would work @VineethReddy02

VineethReddy02:
Hi, @jtlisi I was about to do the same changes i.e passing ring into the table-manager struct. But I couldn't place entire business logic in table-manager.go. The goroutine I am using to sync with other cortex instances. I am unable to place it in the table-manager.go it is causing cyclic dependency issue. 
If there is an alternative place for that logic. I can do that.

Thanks @jtlisi @pracucci for the suggestions.

jtlisi:
Can you post the issue you are facing? Maybe this is something we can work on offline.

pracucci:
> Can you post the issue you are facing? Maybe this is something we can work on offline.

I agree. I'm not much happy with the current design of the changes. I would prefer to have a conversation to design it first, and then build it.

pkg/cortex/modules.go
```diff
@@ -592,7 +673,13 @@ var modules = map[moduleName]module{
 		stop: (*Cortex).stopCompactor,
 	},
 
+	SingleTableManager: {
```
jtlisi: I don't think a separate module is needed in this case. The single binary should use `TableManager`.

pkg/cortex/modules.go
```diff
@@ -421,6 +425,83 @@ func (t *Cortex) initTableManager(cfg *Config) error {
...
+
+	t.tableManager, err = chunk.NewTableManager(cfg.TableManager, cfg.Schema, cfg.Ingester.MaxChunkAge, tableClient, bucketClient)
+	if err != nil {
+		return err
+	}
+
+	//This method gets the table manager owner by default table manager
+	//owner is lowest token holding cortex instance
+	t.tableManager.Ring = t.ring
+	tableManagerOwner, err := t.tableManager.SyncTableManagerOwnership(cfg.Ingester.LifecyclerConfig)
```
jtlisi: I wouldn't run this here.

I recommend running this at the start of the `loop` function within the table manager: https://github.com/cortexproject/cortex/blob/46b99cf48071cb4674576beabfa1ec7216a2f1cd/pkg/chunk/table_manager.go#L167

Every binary should start their TableManager. The loop will only become active, if their embedded Ingester has the lowest token. This means you don't need the loop below.

VineethReddy02:
I am seeing this error when table-manager ownership passed from one cortex instance to other

`level=error ts=2020-02-13T20:56:06.809175633Z caller=table_manager.go:206 msg="error syncing tables" err="java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.apache.cassandra.exceptions.ConfigurationException: Column family ID mismatch (found a4cc43a0-4ea2-11ea-b313-6d2c86545d91; expected a4b53930-4ea2-11ea-b313-6d2c86545d91)"`

This error seems to be valid as the columns family id created by one table-manager and other table-manager will differ.
Is this okay or this should be handled?

VineethReddy02:
Hii!
With my changes at any point in time only one table manager instance exists.
But if a cortex instance is creating tables and another cortex instance is started and if it takes over the ownership of the first cortex instance. I see Cassandra is throwing a runtime exception
```level=error ts=2020-02-13T20:56:06.809175633Z caller=table_manager.go:206 msg="error syncing tables" err="java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.apache.cassandra.exceptions.ConfigurationException: Column family ID mismatch (found a4cc43a0-4ea2-11ea-b313-6d2c86545d91; expected a4b53930-4ea2-11ea-b313-6d2c86545d91)"```
Is this expected and can be ignored?

File: pkg/cortex/modules.go [hunk 3]
```diff
@@ -638,7 +642,7 @@ var modules = map[moduleName]module{
 	},
 
 	TableManager: {
-		deps: []moduleName{Server},
+		deps: []moduleName{Server, Ring, Ingester},
```
sandeepsukhani: This does not look right. It looks like `TableManager` would now also start an `Ingester`.

Maybe we could create `KVStore` client and do CAS operations to claim a key with heartbeat mechanism. One who holds the key would be allowed to do table operations. Consider key to be `tablemanager` and value to include an ID(could be hostname or some other random key) and `update-time`, which would be treated as heartbeat expected to be updated by tablemanager owning the key, every time it ticks [here](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/table_manager.go#L219).

I was thinking of doing it like:
For each tick [here](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/table_manager.go#L219) do CAS as follow: 
a. check if ID is mine, if yes update the heartbeat
b. else check if `update-time` is too old, if yes update the key with own id and the current time.
If CAS was successful do the table operations.

@pracucci @jtlisi what do you guys think about this approach?

jtlisi:
@VineethReddy02 are you still working on this?

PR #2101, opened 2020-02-10
Fixed ruler to run with blocks storage

pracucci:
**What this PR does**:
The ruler currently panics with the following trace when running with the experimental blocks storage because the querier is initialized with a `nil` store instead of blocks querier store. This PR fixes it.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2834554]

goroutine 397 [running]:
github.com/cortexproject/cortex/pkg/querier.(*chunkStoreQuerier).Select(0xc000af1600, 0xc00064a500, 0xc00017e370, 0x1, 0x1, 0x1, 0x0, 0x0, 0x0, 0xc00029d5c0, ...)
	/workspace/src/github.com/cortexproject/cortex/pkg/querier/chunk_store_queryable.go:42 +0xa4
github.com/cortexproject/cortex/pkg/querier.querier.Select.func1(0xc00064a500, 0xc00017e370, 0x1, 0x1, 0xc000088600, 0xc000088240, 0x398fca0, 0xc000af1600)
	/workspace/src/github.com/cortexproject/cortex/pkg/querier/querier.go:175 +0x8a
created by github.com/cortexproject/cortex/pkg/querier.querier.Select
	/workspace/src/github.com/cortexproject/cortex/pkg/querier/querier.go:174 +0x152
```

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
   * `-server.grpc.keepalive.time`
   * `-server.grpc.keepalive.timeout`
 * [BUGFIX] Experimental TSDB: fixed `/all_user_stats` and `/api/prom/user_stats` endpoints when using the experimental TSDB blocks storage. #2042
+* [BUGFIX] Experimental TSDB: fixed ruler to correctly work with the experimental TSDB blocks storage. #2101
 * [BUGFIX] Azure Blob ChunkStore: Fixed issue causing `invalid chunk checksum` errors. #2074
 
 Cortex 0.4.0 is the last version that can *write* denormalised tokens. Cortex 0.5.0 and above always write normalised tokens.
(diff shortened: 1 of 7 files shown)


pracucci:
> I'm just wondering if new top-level component was introduced to avoid having two copies of chunk store in single-binary Cortex? Would it be an issue?

@pstibrany Yes, it was introduced to avoid having 2 instances of chunk store in single-binary when the ruler will run in the single-binary too. It's not a real necessity today, but the risk to miss it and run 2 instances of the `BlockQuerier` once will be real is too high (they will clash because trying to use the same local directory) so I've preferred to address it now.

pstibrany:
> once will be real is too high (they will clash because trying to use the same local directory, other being inefficient)

Thanks for explaining, makes total sense!

PR #2102, opened 2020-02-10
Share blocks storage single binary local dev env

pracucci:
**What this PR does**:
Like I did with #2011, in this PR I'm proposing to share an easy way to setup a single binary local development environment based on the blocks storage.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci:
> Do you need all the .gitignores? Can you just ignore everything starting with .data?

We need them due to how Minio works (each ignored dir is a bucket) if we want a smooth experience for the final user (run `compose-up.sh` and it works, without having to create the empty dirs mapping the Cortex bucket).

PR #2104, opened 2020-02-10
Allow ingester initialization to continue in background.

pstibrany:
This PR solves problem with slow ingester initialization when doing WAL replays on block databases.

By allowing ingester initialization to continue in the background, the rest of Cortex binary can start in the meantime, especially readiness and metrics handlers. Until ingester initialization has finished, it will return errors to most (but not all, like Flush) API requests.

There may be implications for returning errors from API calls. Alternatively, we can delay registration of handlers until ingester initialization has finished. That would be easy to do, but I'm not sure about the right approach.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -293,10 +293,14 @@ func (t *Cortex) initIngester(cfg *Config) (err error) {
 	cfg.Ingester.TSDBConfig = cfg.TSDB
 	cfg.Ingester.ShardByAllLabels = cfg.Distributor.ShardByAllLabels
 
-	t.ingester, err = ingester.New(cfg.Ingester, cfg.IngesterClient, t.overrides, t.store, prometheus.DefaultRegisterer)
-	if err != nil {
-		return
+	var errfut *ingester.ErrorFuture
+	t.ingester, errfut = ingester.New(cfg.Ingester, cfg.IngesterClient, t.overrides, t.store, prometheus.DefaultRegisterer)
+	if done, err := errfut.Get(); done && err != nil {
+		return err
 	}
+
+	// if ingester initialization isn't done yet, we go ahead. Services that use ingester must check its
+	// readiness status and eventually wait for it to finish initialization.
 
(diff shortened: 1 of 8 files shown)


pkg/cortex/modules.go
```diff
@@ -293,11 +293,15 @@ func (t *Cortex) initIngester(cfg *Config) (err error) {
 	cfg.Ingester.TSDBConfig = cfg.TSDB
 	cfg.Ingester.ShardByAllLabels = cfg.Distributor.ShardByAllLabels
 
-	t.ingester, err = ingester.New(cfg.Ingester, cfg.IngesterClient, t.overrides, t.store, prometheus.DefaultRegisterer)
-	if err != nil {
-		return
+	var errfut *ingester.ErrorFuture
+	t.ingester, errfut = ingester.New(cfg.Ingester, cfg.IngesterClient, t.overrides, t.store, prometheus.DefaultRegisterer)
```
pracucci: A difference I see is that currently Cortex exits if the initialization fails. After this PR it doesn't and if the ingester initialization fails it will continue to run but not work.

pstibrany: That is a valid point, and I can address that.

pkg/cortex/modules.go
```diff
@@ -293,11 +293,15 @@ func (t *Cortex) initIngester(cfg *Config) (err error) {
...
+	if done, err := errfut.Get(); done && err != nil {
```
thorfour: Futures feel needlessly complicated here, and an anti-pattern with Go.  Concurrency should be left to the caller, so I believe the module init function should call `ingester.New` in a goroutine that waits for an error response.

pstibrany: Hmm, that is a valid point. Back to drawing board.

pracucci: However, by contract the ingester should be populated in the `Cortex` struct once the `initIngester()` terminates.

PR #2105, opened 2020-02-10
Add repair for checkpoint/WAL

codesome:
@gouthamve suggested me to add it to the release and that `v0.6.2` would be cut with this.

Things changed during checkpointing: If the old checkpoint was X and the new was Y, we were deleting the segments before Y. Now in this PR, I have changed it to delete WAL segments to delete before X so that we can recover from checkpoint X even if checkpoint Y gets corrupt.

How the replay works now with repair: 
1. Attempt replay of checkpoint Y. If the checkpoint Y is corrupt, delete the checkpoint Y and recover from checkpoint X instead.
2. If checkpoint X is also corrupt, it is a hard fail right now.
3. Depending on which checkpoint was recovered, we start replaying the WAL from either segment X or Y.
4. If WAL is corrupt, we do the usual Prometheus repair to discard everything after the corrupt record. If the repair fails its again a hard error.
 
I will be adding a test for repair now.

**Checklist**
- [x] Tests updated
- [ ] ~Documentation added~


pkg/ingester/wal.go
```diff
@@ -205,6 +205,12 @@ func (w *walWrapper) performCheckpoint() (err error) {
 		return errors.New("function to get user states not initialised")
 	}
 
+	// As the checkpoint name is taken from last WAL segment, we need to ensure
```
gouthamve: Can we only cut a new segment if the current segment and the lastCheckpoint match?

pkg/ingester/metrics.go
```diff
@@ -86,6 +87,10 @@ func newIngesterMetrics(r prometheus.Registerer, registerMetricsConflictingWithT
 			Name: "cortex_ingester_wal_replay_duration_seconds",
 			Help: "Time taken to replay the checkpoint and the WAL.",
 		}),
+		walCorruptionsTotal: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "cortex_ingester_wal_corruption_total",
```
pracucci: `cortex_ingester_wal_corruption_total` > `cortex_ingester_wal_corruptions_total` (added an `s` to `corruption`) to keep it consistent with TSDB

pkg/ingester/wal.go
```diff
@@ -215,6 +215,19 @@ func (w *walWrapper) performCheckpoint() (err error) {
 		return err
 	}
 
+	if lastCh == lastSegment {
```
pracucci: Should we check `&& lastCh != -1` to cover the case we have no checkpoints and no segments? In general, I've the feeling that `performCheckpoint()` doesn't handle the case `lastSegment == -1`: what am I missing?

pkg/ingester/wal.go
```diff
@@ -252,9 +265,10 @@ func (w *walWrapper) performCheckpoint() (err error) {
 		return errors.Wrap(err, "rename checkpoint directory")
 	}
 
-	// The last segment might still have been active during the checkpointing,
-	// hence delete only the segments before that.
-	if err := w.wal.Truncate(lastSegment - 1); err != nil {
+	// We delete the WAL segments which are before the previous checkpoint and not before the
+	// current checkpoint. This is because if the lastest checkpoint crashes for any reason, we
```
pracucci: `lastest` > `latest`

pkg/ingester/wal.go
```diff
@@ -252,9 +265,10 @@ func (w *walWrapper) performCheckpoint() (err error) {
...
+	// should be able to recover from the older checkpoint which would need the older WAL segments.
+	if err := w.wal.Truncate(lastCh - 1); err != nil {
```
pracucci: Can `lastCh` be `<= 0`? With `0` we would call `Truncate(-1)`.

pkg/ingester/wal.go
```diff
@@ -384,35 +383,89 @@ func recoverFromWAL(ingester *Ingester) (err error) {
...
+	level.Info(util.Logger).Log("msg", "recovering from checkpoint")
+	start := time.Now()
+	userStates, idx, err := processCheckpointWithRepair(walDir, ingester, nWorkers, stateCache, seriesCache)
+	if err != nil {
+		return err
 	}
+	elapsed := time.Since(start)
+	level.Info(util.Logger).Log("msg", "recovered from checkpoint", "time", elapsed.String())
+	defer func() {
+		if returnErr == nil {
```
pracucci: `returnErr` is never valued

pkg/ingester/wal.go
```diff
@@ -569,6 +622,35 @@ type samplesWithUserID struct {
...
+
+	// Attempt repair.
+	level.Info(util.Logger).Log("msg", "attemption repair of the WAL")
+	w, err := wal.New(util.Logger, nil, walDir, true)
+	if err != nil {
+		return err
+	}
+
+	err = w.Repair(corruptErr)
+	if cerr := w.Close(); cerr != nil {
```
pracucci: If both the `w.Repair()` and `w.Close()` fail, we do return the `Close()` error while I believe would be more appropriate to return the `Repair()` error (which is the first one).

pkg/ingester/wal.go
```diff
@@ -384,35 +383,89 @@ func recoverFromWAL(ingester *Ingester) (err error) {
...
+	}
+
+	// We don't call repair on checkpoint as losing even a single record is like losing the entire data of a series.
+	// We try recovering from the older checkpoint instead.
+	ingester.metrics.walCorruptionsTotal.Inc()
+	level.Error(util.Logger).Log("msg", "checkpoint recovery failed, deleting this checkpoint and trying to recover from old checkpoint", "err", err)
+
+	// Deleting this checkpoint to try the previous checkpoint.
+	if err := os.RemoveAll(lastCheckpointDir); err != nil {
+		return nil, -1, err
```
pracucci: We may wrap this `err` to add a message about what operation we were doing, something like:
`errors.Wrapf(err, "unable to delete checkpoint directory %s", lastCheckpointDir`

pkg/ingester/wal.go
```diff
@@ -384,35 +383,89 @@ func recoverFromWAL(ingester *Ingester) (err error) {
...
+	}
+
+	// If we have reached this point, it means the last checkpoint was deleted.
+	// Now the last checkpoint will be the one before the deleted checkpoint.
+	lastCheckpointDir, idx, err = lastCheckpoint(walDir)
+	if err != nil {
+		return nil, -1, err
+	}
+
```
pracucci: Shouldn't we also check `if idx < 0`? `lastCheckpoint()` may return no error, but no checkpoint is available (ie. the failing one is the first one we have).

pkg/ingester/wal.go
codesome: I think there should be at-least 1 segment when the WAL is started. But I will add the check just in case.

pkg/ingester/wal.go
codesome: Yes, it can be -1, but no problem with that. `Truncate(-1)` won't delete any segments (as expected).

pkg/ingester/wal.go
codesome: Not sure I understand. As I have `returnErr` as a return variable, any error that is returned later will be bound to it.

pkg/ingester/wal.go
codesome: Right. I will use multi error from TSDB here instead.

pkg/ingester/wal.go
codesome: Good catch! And I believe we need not error out here as for the first checkpoint all the WAL data would be present.

codesome:
I have rebased this PR and it is now pointing to `master` (following the discussion on the slack)

pkg/ingester/wal.go
codesome: Done

codesome:
Following @pracucci's comments on having only 1 or no checkpoint on disk and after fixing some code, I have added the case of having 0 or 1 checkpoint in the unit test now.

pkg/ingester/wal.go
gouthamve: Wow, you're right: https://play.golang.org/p/zeWCaztXUGh

But again it's non-obvious to me and Marco and I assume it'll be the same for others. Can we somehow refactor to use `returnErr` everywhere? 

Or not put this in a defer? I think it'll be clearer and more obvious if we manually set it and not put it in a defer.

codesome: Okay

codesome:
@gouthamve done

pkg/ingester/metrics.go
```diff
@@ -87,6 +88,10 @@ func newIngesterMetrics(r prometheus.Registerer, registerMetricsConflictingWithT
 			Name: "cortex_ingester_wal_replay_duration_seconds",
 			Help: "Time taken to replay the checkpoint and the WAL.",
 		}),
+		walCorruptionsTotal: prometheus.NewCounter(prometheus.CounterOpts{
```
pracucci: Please register the metric below, otherwise it's not exposed.

pkg/ingester/wal.go
```diff
@@ -252,9 +269,10 @@ func (w *walWrapper) performCheckpoint() (err error) {
 		return errors.Wrap(err, "rename checkpoint directory")
 	}
 
-	// The last segment might still have been active during the checkpointing,
-	// hence delete only the segments before that.
-	if err := w.wal.Truncate(lastSegment - 1); err != nil {
+	// We delete the WAL segments which are before the previous checkpoint and not before the
+	// current checkpoint. This is because if the latest checkpoint is corrupted for any reason, we
+	// should be able to recover from the older checkpoint which would need the older WAL segments.
+	if err := w.wal.Truncate(lastCh - 1); err != nil {
```
pracucci: Here we do keep until `lastCh - 1` but then below we do `w.deleteCheckpoints(lastCh)` which deletes also the checkpoint `lastCh - 1`. Is this correct?

pracucci: To delete everything < `lastCh` shouldn't we call `w.wal.Truncate(lastCh)`? The index passed to `Truncate()` looks excluded.

codesome: After some offline discussion, this should be `lastCh` here

codesome: `w.wal.Truncate(lastCh)` is the right thing here

pkg/ingester/wal.go
```diff
@@ -569,6 +633,35 @@ type samplesWithUserID struct {
 	userID  string
 }
 
+func processWALWithRepair(walDir string, startSegment int, userStates *userStates, nWorkers int,
```
pstibrany: Would it make sense to bundle all these parameters into some struct, and make these functions methods on the struct? We seem to be passing them around to different methods.

pkg/ingester/wal.go
```diff
@@ -384,35 +387,96 @@ func recoverFromWAL(ingester *Ingester) (err error) {
...
+	ingester.userStates = userStates
+	ingester.userStatesMtx.Unlock()
 	return nil
 }
 
+func processCheckpointWithRepair(walDir string, ingester *Ingester, nWorkers int,
+	stateCache []map[string]*userState, seriesCache []map[string]map[uint64]*memorySeries) (*userStates, int, error) {
+
+	// Use a local userStates, so we don't need to worry about locking.
+	userStates := newUserStates(ingester.limiter, ingester.cfg, ingester.metrics)
```
pstibrany: Isn't this going to affect real ingester metrics? Is that a problem?

codesome: I will check about that, good catch

codesome: Cross checked with how transfer handles them https://github.com/cortexproject/cortex/blob/5d7b05c36bbec3e2f478e924c3e1e1a853886abf/pkg/ingester/transfer.go#L65-L151

Looks like it's the same here too (including the memory chunks metrics). So I guess all good here.

pkg/ingester/wal.go
codesome: Done

gouthamve:
Thanks Ganesh!

PR #2106, opened 2020-02-10
goimports -local github.com/cortexproject/cortex -w pkg/ cmd/ tools/

pstibrany:
Ran this command:

```
goimports -local github.com/cortexproject/cortex -w pkg/ cmd/ tools/
```

Also modified lint check to verify order of imports.

This was prompted by some [recent feedback](https://github.com/cortexproject/cortex/pull/2060#discussion_r374023546) on PR I've authored.

### DIFF
--- a/pkg/querier/correctness/simple.go
+++ b/pkg/querier/correctness/simple.go
@@ -5,11 +5,13 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/api/prometheus/v1"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
+
+	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 )
 
 const (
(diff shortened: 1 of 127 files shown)


pkg/querier/correctness/simple.go
```diff
@@ -5,11 +5,13 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/api/prometheus/v1"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
```
pracucci: Why is this duplicated?

pstibrany: I wondered about that too. It actually caused build failure. Fixing and running test locally as I write this.

bboreham:
I think we should write down the rules, e.g. in `CONTRIBUTING.md`.

E.g. why is `weaveworks/common` in a group separate to `prometheus/common`?
Or Thanos for that matter.

pstibrany:
> I think we should write down the rules, e.g. in `CONTRIBUTING.md`.
> 
> E.g. why is `weaveworks/common` in a group separate to `prometheus/common`?
> Or Thanos for that matter.

Noticed that too. I think goimports moved it to its own group, because it was in separate group before, but it can be moved together with 3rd party libs. I'll send follow-up PR with documentation change and 3 groups only (stdlib, 3rd party, cortex).

PR #2108, opened 2020-02-10
Rename chunk.ObjectClient and implement generic object store chunk client

### LINKED ISSUE #2107: Separate ObjectClient interface
This is a followup to a recent discussion that occurred in #2049.

Between S3/GCS/Azure/BoltDB, there is a lot of duplicated code between our chunk object stores. Specifically the `GetChunks` and `PutChunks` functions. After #2049, each of these backends supports the follow functions:

```go
type ObjectClient interface {
  PutObject(ctx context.Context, objectKey string, object io.ReadSeeker) 
  GetObject(ctx context.Context, objectKey string) (io.ReadCloser, error)
  List(ctx context.Context, prefix string) ([]StorageObject, error)
}
```

For each of the Object backend stores, the current `ObjectClient` interface is implemented using the above functions.

As mentioned in the PR, I propose the following refactor:

- Rename `ObjectClient` to `Client` without changing any functionality. `Client` is chosen as opposed to `ChunkClient` since the package name is `chunk`
- Create a new `ObjectClient` interface that contains the the `PutObject`, `GetObject` and `List` functions
- Create a generic implementation of the `Client` that would be built on top of the `ObjectClient`. This would allow a single `Client` to abstract differences between S3/GCS/AzureBS/BoltDB. It would also allow the `ObjectClient` to be used to store things other than chunks. `DeletePlans` or rule configs for example.

jtlisi:
**What this PR does**:
* Rename the chunk `ObjectClient` to `Client` 
* Create a new ObjectClient interface with simple methods implemented in #2049 
* Create generic object store chunk client in `pkg/chunk/object`
* Update fixtures to detect and utilize generic chunk object client during tests
* Utilize the generic `object.ChunkClient` when S3/GCS/AzureBlobStore/LocalFS is specified as a chunk store, all of these backends already implement the `ObjectClient` interface.

Fixes #2107 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -154,12 +154,12 @@ func NewDynamoDBIndexClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (c
 	return newDynamoDBStorageClient(cfg, schemaCfg)
 }
 
-// NewDynamoDBObjectClient makes a new DynamoDB-backed ObjectClient.
-func NewDynamoDBObjectClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (chunk.Client, error) {
+// NewDynamoDBChunkClient makes a new DynamoDB-backed chunk.Client.
+func NewDynamoDBChunkClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (chunk.Client, error) {
 	return newDynamoDBStorageClient(cfg, schemaCfg)
 }
 
-// newDynamoDBStorageClient makes a new DynamoDB-backed IndexClient and ObjectClient.
+// newDynamoDBStorageClient makes a new DynamoDB-backed IndexClient and chunk.Client.
 func newDynamoDBStorageClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (*dynamoDBStorageClient, error) {
 	dynamoDB, err := dynamoClientFromURL(cfg.DynamoDB.URL)
 	if err != nil {
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -155,7 +155,7 @@ func NewDynamoDBIndexClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (c
 }
 
 // NewDynamoDBObjectClient makes a new DynamoDB-backed ObjectClient.
-func NewDynamoDBObjectClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (chunk.ObjectClient, error) {
+func NewDynamoDBObjectClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (chunk.Client, error) {
 	return newDynamoDBStorageClient(cfg, schemaCfg)
 }
 
@@ -461,7 +461,7 @@ type chunksPlusError struct {
 	err    error
 }
 
-// GetChunks implements chunk.ObjectClient.
+// GetChunks implements chunk.Client.
 func (a dynamoDBStorageClient) GetChunks(ctx context.Context, chunks []chunk.Chunk) ([]chunk.Chunk, error) {
 	log, ctx := spanlogger.New(ctx, "GetChunks.DynamoDB", ot.Tag{Key: "numChunks", Value: len(chunks)})
 	defer log.Span.Finish()
@@ -639,7 +639,7 @@ func (a dynamoDBStorageClient) PutChunkAndIndex(ctx context.Context, c chunk.Chu
 	return a.BatchWrite(ctx, dynamoDBWrites)
 }
 
-// PutChunks implements chunk.ObjectClient.
+// PutChunks implements chunk.Client.
 func (a dynamoDBStorageClient) PutChunks(ctx context.Context, chunks []chunk.Chunk) error {
 	dynamoDBWrites, err := a.writesForChunks(chunks)
 	if err != nil {
--- a/pkg/chunk/local/fs_object_client.go
+++ b/pkg/chunk/local/fs_object_client.go
@@ -1,9 +1,7 @@
 package local
 
 import (
-	"bytes"
 	"context"
-	"encoding/base64"
 	"flag"
 	"io"
 	"io/ioutil"
@@ -39,7 +37,7 @@ type FSObjectClient struct {
 	cfg FSConfig
 }
 
-// NewFSObjectClient makes a chunk.ObjectClient which stores chunks as files in the local filesystem.
+// NewFSObjectClient makes a chunk.Client which stores chunks as files in the local filesystem.
 func NewFSObjectClient(cfg FSConfig) (*FSObjectClient, error) {
 	if err := util.EnsureDirectory(cfg.Directory); err != nil {
 		return nil, err
@@ -53,55 +51,12 @@ func NewFSObjectClient(cfg FSConfig) (*FSObjectClient, error) {
 // Stop implements ObjectClient
 func (FSObjectClient) Stop() {}
 
-// PutChunks implements ObjectClient
-func (f *FSObjectClient) PutChunks(ctx context.Context, chunks []chunk.Chunk) error {
-	for i := range chunks {
-		buf, err := chunks[i].Encoded()
-		if err != nil {
-			return err
-		}
-
-		filename := base64.StdEncoding.EncodeToString([]byte(chunks[i].ExternalKey()))
-		if err := f.PutObject(ctx, filename, bytes.NewReader(buf)); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-// GetChunks implements ObjectClient
-func (f *FSObjectClient) GetChunks(ctx context.Context, chunks []chunk.Chunk) ([]chunk.Chunk, error) {
-	return util.GetParallelChunks(ctx, chunks, f.getChunk)
-}
-
-func (f *FSObjectClient) getChunk(ctx context.Context, decodeContext *chunk.DecodeContext, c chunk.Chunk) (chunk.Chunk, error) {
-	filename := base64.StdEncoding.EncodeToString([]byte(c.ExternalKey()))
-
-	readCloser, err := f.GetObject(ctx, filename)
-	if err != nil {
-		return chunk.Chunk{}, err
-	}
-
-	defer readCloser.Close()
-
-	buf, err := ioutil.ReadAll(readCloser)
-	if err != nil {
-		return chunk.Chunk{}, err
-	}
-
-	if err := c.Decode(decodeContext, buf); err != nil {
-		return c, err
-	}
-
-	return c, nil
-}
-
-// Get object from the store
+// GetObject from the store
 func (f *FSObjectClient) GetObject(ctx context.Context, objectKey string) (io.ReadCloser, error) {
 	return os.Open(path.Join(f.cfg.Directory, objectKey))
 }
 
-// Put object into the store
+// PutObject into the store
 func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object io.ReadSeeker) error {
 	fullPath := path.Join(f.cfg.Directory, objectKey)
 	err := util.EnsureDirectory(path.Dir(fullPath))
--- a/pkg/chunk/storage/factory.go
+++ b/pkg/chunk/storage/factory.go
@@ -17,6 +17,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/chunk/cassandra"
 	"github.com/cortexproject/cortex/pkg/chunk/gcp"
 	"github.com/cortexproject/cortex/pkg/chunk/local"
+	"github.com/cortexproject/cortex/pkg/chunk/objectclient"
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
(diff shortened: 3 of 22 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] The `chunk.ObjectClient` interface has been renamed `chunk.Client`. A new `chunk.ObjectClient` interface has been created to store raw byte encoded objects in an object backend. The following Object backends (S3/GCS/AzureBlobStore/LocalFS) now are implemented using a generic client in `pkg/chunk/object.ChunkClient` utilizing the new `ObjectClient` interface. #2108
```
pracucci: In the changelog we currently write down only user-facing changes. This shouldn't have any impact on the user, so I would remove this entry at all. What do you think?

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -462,7 +462,7 @@ type chunksPlusError struct {
 	err    error
 }
 
-// GetChunks implements chunk.ObjectClient.
+// GetChunks implements chunk.ChunkClient.
```
pracucci: The naming `chunk.ChunkClient` is references in comments across several files, but it should be `chunk.Client`.

pkg/chunk/storage/utils_test.go
```diff
@@ -33,6 +34,7 @@ func forAllFixtures(t *testing.T, storageClientTest storageClientTest) {
 
 	for _, fixture := range fixtures {
 		t.Run(fixture.Name(), func(t *testing.T) {
+			fmt.Println(fixture.Name())
```
pracucci: Do you need this? The test run with `t.Run(fixture.Name()` and you should already have the fixture name in the tests output.

pkg/chunk/object/chunk_object_client.go
```diff
@@ -0,0 +1,80 @@
+package object
```
pracucci: A couple of things:
1. The filename doesn't look reflecting the `type` name
2. Is it required adding a new pkg just for this file? I'm wondering if we could move it to `pkg/chunk/` and rename this `ChunkClient` to something like `ChunkStoreClient` to avoid conflicts

pkg/chunk/object/chunk_object_client.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"io/ioutil"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/chunk/util"
+)
+
+// ChunkClient is used to store chunks in object store backends
+type ChunkClient struct {
+	Store chunk.ObjectClient
```
pracucci: Is `Store` required to be exposed?

pkg/chunk/object/chunk_object_client.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Stop shuts down the object store and any underlying clients
+func (o *ChunkClient) Stop() {
+	o.Store.Stop()
+}
+
+// PutChunks stores the provided chunks in the configured backend. If multiple errors are
+// returned, the last one sequentially will be propagated up.
+func (o *ChunkClient) PutChunks(ctx context.Context, chunks []chunk.Chunk) error {
```
pracucci: This is not an exact drop-in replacement. This is the copy of the S3 client, while other clients (GCS, FS, Azure) didn't have any parallelization. What may be the implications of this change?

pkg/chunk/object/chunk_object_client.go
```diff
@@ -0,0 +1,80 @@
...
+
+// GetChunks retrieves the specified chunks from the configured backend
+func (o *ChunkClient) GetChunks(ctx context.Context, chunks []chunk.Chunk) ([]chunk.Chunk, error) {
+	return util.GetParallelChunks(ctx, chunks, o.getChunk)
+}
+
+func (o *ChunkClient) getChunk(ctx context.Context, decodeContext *chunk.DecodeContext, c chunk.Chunk) (chunk.Chunk, error) {
+	readCloser, err := o.Store.GetObject(ctx, c.ExternalKey())
+	if err != nil {
+		return chunk.Chunk{}, err
```
pracucci: In some clients we were doing `errors.WithStack()` on errors. Is it something we wanna keep?

pkg/chunk/local/fs_object_client.go
```diff
@@ -53,57 +52,16 @@ func NewFSObjectClient(cfg FSConfig) (*FSObjectClient, error) {
...
-	}
-
-	return c, nil
-}
-
-// Get object from the store
+// GetObject from the store
 func (f *FSObjectClient) GetObject(ctx context.Context, objectKey string) (io.ReadCloser, error) {
-	return os.Open(path.Join(f.cfg.Directory, objectKey))
+	filename := base64.StdEncoding.EncodeToString([]byte(objectKey))
```
pracucci: I'm not sure doing the base64 encoding here (in the `GetObject()`) is correct. When Sandeep submitted the former PR to introduce `GetObject()` the idea was to intentionally keep it generic so that `objectKey` can contain `/` to navigate through sub directories. We should keep this behavior (and also add a unit test to protect from this regression).

pkg/chunk/testutils/testutils.go
```diff
@@ -22,7 +23,7 @@ const (
 // Fixture type for per-backend testing.
 type Fixture interface {
 	Name() string
-	Clients() (chunk.IndexClient, chunk.ObjectClient, chunk.TableClient, chunk.SchemaConfig, error)
+	Clients() (chunk.IndexClient, chunk.Client, chunk.ObjectClient, chunk.TableClient, chunk.SchemaConfig, error)
```
pracucci: Having both `chunk.Client` and `chunk.ObjectClient` sounds redundant to me. Could you explain why we need `chunk.ObjectClient` too?

pkg/chunk/storage/factory.go
```diff
@@ -164,13 +165,14 @@ func NewIndexClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chun
 }
 
 // NewObjectClient makes a new ObjectClient of the desired types.
-func NewObjectClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.ObjectClient, error) {
+func NewObjectClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.Client, error) {
+	var store chunk.ObjectClient
+	var err error
 	switch name {
```
pracucci: To me, it looks a bit unclear this code now, because we have some `case` which `return` and other `case` which save `store` and `err`. Moreover, in case of error I believe we should not return an instance of `object.ChunkClient`.

A cleaner solution may be adding a function like:
```
func newChunkClientFromStore(store, err) (*object.ChunkClient) {
  if err != nil {
    return nil, err
  }

 return &object.ChunkClient{
		Store: store,
	}, nil
}
```

This way you can use it like:
`return newChunkClientFromStore(aws.NewS3ObjectClient(cfg.AWSStorageConfig.S3Config))`

pkg/chunk/object/chunk_object_client.go
jtlisi: I've thought about this a bit for an unrelated reason. The short answer is it doesn't matter if we parallelize writes here because we never pass more the one chunk at a time to the client:

From: pkg/chunk/series_store.go
```go
// Put implements ChunkStore
func (c *seriesStore) Put(ctx context.Context, chunks []Chunk) error {
	for _, chunk := range chunks {
		if err := c.PutOne(ctx, chunk.From, chunk.Through, chunk); err != nil {
			return err
		}
	}
	return nil
}
```

From: pkg/chunk/chunk_store.go
```go
// Put implements ChunkStore
func (c *store) Put(ctx context.Context, chunks []Chunk) error {
	for _, chunk := range chunks {
		if err := c.PutOne(ctx, chunk.From, chunk.Through, chunk); err != nil {
			return err
		}
	}
	return nil
}

// PutOne implements ChunkStore
func (c *store) PutOne(ctx context.Context, from, through model.Time, chunk Chunk) error {
	chunks := []Chunk{chunk}

	err := c.storage.PutChunks(ctx, chunks)
	if err != nil {
		return err
	}

	c.writeBackCache(ctx, chunks)

	writeReqs, err := c.calculateIndexEntries(chunk.UserID, from, through, chunk)
	if err != nil {
		return err
	}

	return c.index.BatchWrite(ctx, writeReqs)
}
```

So in general, writing in parallel on the client level is irrelevant. I chose the S3 implementation because I felt it was the most advanced implementation. If we want to keep a sequential implementation, the client could be updated to use a maximum number of concurrent writes, and set that value to one. I think in general we should support parallel/batch writes, which is why I chose to use the same approach as the S3 client. However, it does not matter at this point in time because of how the clients are used upstream.

pkg/chunk/local/fs_object_client.go
jtlisi: I personally don't like using the base64 encoding either. However, the current implementation uses it to encode chunks. Maybe I can maintain this behavior with a config option to allow for local storage users to keep base64 encoded chunk names. I personally think storing chunks locally using the chunk external key would work great, but I haven't thought through all of the implications. Having each users chunks stored in a directory seems conceptually sound to me:

```go
// ExternalKey returns the key you can use to fetch this chunk from external
// storage. For newer chunks, this key includes a checksum.
func (c *Chunk) ExternalKey() string {
	// Some chunks have a checksum stored in dynamodb, some do not.  We must
	// generate keys appropriately.
	if c.ChecksumSet {
		// This is the inverse of parseNewExternalKey.
		return fmt.Sprintf("%s/%x:%x:%x:%x", c.UserID, uint64(c.Fingerprint), int64(c.From), int64(c.Through), c.Checksum)
	}
	// This is the inverse of parseLegacyExternalKey, with "<user id>/" prepended.
	// Legacy chunks had the user ID prefix on s3/memcache, but not in DynamoDB.
	// See comment on parseExternalKey.
	return fmt.Sprintf("%s/%d:%d:%d", c.UserID, uint64(c.Fingerprint), int64(c.From), int64(c.Through))
}
```

pkg/chunk/object/chunk_object_client.go
pracucci: You're right. I double checked it, and we always call `PutChunks()` with 1 chunk. I'm OK keeping parallelism then. I'm not against parallelism, but I was wondering potential side effects (ie. rate limiting) of this change, but given there's no practical change LGTM.

pkg/chunk/local/fs_object_client.go
pracucci: > I personally think storing chunks locally using the chunk external key would work great, but I haven't thought through all of the implications.

If we remove the base64 we would end up with a breaking change because `base64(chunk.ExternalKey()) != chunk.ExternalKey()`.

I think this PR should be a mere refactoring PR, without altering the behaviour in any way, so `PutChunks()` and `GetChunks()` should base64 encode the key (when using the FS client), while `PutObject()` and `GetObject()` not.

An option to achieve it may be adding an extra function to the `chunk.Client` interface like `EncodeExternalKey(key string) string`. This function is a noop for all clients, except the FS client which will do the base 64. Finally, `PutChunks()` and `GetChunks` will call `EncodeExternalKey(chunks[i].ExternalKey())`.

pkg/chunk/object/chunk_object_client.go
jtlisi: This can’t live in the chunk package because of a circular dependency issue with the util package.

pracucci: I didn't notice it. Then ok.

> The filename doesn't look reflecting the type name

I believe this can still be renamed to reflect the struct name.

jtlisi: I agree I’m struggling to come up with a name that’s not to verbose but denotes a chunk client abstraction on top of an object store

pkg/chunk/object/chunk_object_client.go
jtlisi: No, I’ll make it private and add a constructor function

pkg/chunk/object/chunk_object_client.go
jtlisi: I will add this functionality

pkg/chunk/testutils/testutils.go
jtlisi: I can have the downstream fixtures wrap the object client before returning them

pkg/chunk/storage/factory.go
jtlisi: LGTM

pkg/chunk/local/fs_object_client.go
jtlisi: I think an easy way to handle this is to inject an optional `Encoder` function into the object store chunk client.

```go
// KeyEncoder is used to encode chunk keys before writing/retrieving chunks
// from the underlying ObjectClient
type KeyEncoder func(string) string

// Client is used to store chunks in object store backends
type Client struct {
	store      ObjectClient
	keyEncoder KeyEncoder
}
```

In `PutChunks`:
```go
		key := chunks[i].ExternalKey()
		if o.keyEncoder != nil {
			key = o.keyEncoder(key)
		}
```

in `getChunk`:
```go
func (o *Client) getChunk(ctx context.Context, decodeContext *DecodeContext, c Chunk) (Chunk, error) {
	key := c.ExternalKey
	if o.keyEncoder != nil {
		key = o.keyEncoder(key)
	}
```

pkg/chunk/object/chunk_object_client.go
jtlisi: I decided to call the package `objectclient` and the struct `Client`. That way it will be the `chunk/objectclient.Client` struct. This seems conceptually sound to me. Let me know what you think.

pkg/chunk/local/fs_object_client.go
pracucci: Yes, even better.

pkg/chunk/storage/utils_test.go
jtlisi: This was for debugging. I'll remove it.

CHANGELOG.md
jtlisi: I removed the entry

pkg/chunk/aws/dynamodb_storage_client.go
jtlisi: 👍

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -155,7 +155,7 @@ func NewDynamoDBIndexClient(cfg DynamoDBConfig, schemaCfg chunk.SchemaConfig) (c
 }
 
 // NewDynamoDBObjectClient makes a new DynamoDB-backed ObjectClient.
```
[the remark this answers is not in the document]
jtlisi: Will do, I will also rename this function to be `NewDynamoDBChunkClient`

pkg/chunk/storage/factory.go
```diff
@@ -164,13 +165,12 @@ func NewIndexClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chun
 }
 
 // NewObjectClient makes a new ObjectClient of the desired types.
-func NewObjectClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.ObjectClient, error) {
+func NewObjectClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.Client, error) {
```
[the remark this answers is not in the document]
jtlisi: I'll update it to reflect the changes.

PR #2013, opened 2020-01-21
Updated gocql dependency

VineethReddy02:
Signed-off-by: vineeth <vineethpothulapati@outlook.com>

@pracucci

PR #2109, opened 2020-02-11
Cassandra: Fix TLS host verification

### LINKED ISSUE #2007: Cassandra: Failing host verification without IP SANs
Hi!

I am trying to connect Cortex to Azure CosmosDB via the Cassandra API.
Using cortex 0.4.0 (target querier; `cassandra.ssl=true`) without `-cassandra.host-verification=false` yields:

```
level=error ts=2020-01-20T10:08:59.800666696Z caller=log.go:141 msg="error initializing cortex" err="gocql: unable to create session: unable to discover protocol version: x509: cannot validate certificate for <AZUREIP> because it doesn't contain any IP SANs <...>
```

This is apparently the same issue as https://github.com/gocql/gocql/issues/1232 in the used gocql library. The fix for this problem (https://github.com/gocql/gocql/pull/1310) was already merged into the gocql-master, but did not make it into the vendored dependencies of cortex yet.

Would you mind updating this dependency (or accepting a PR for it)?

joshmue:
Signed-off-by: Joshua Mühlfort <muehlfort@gonicus.de>

**What this PR does**:
Enable TLS host verification for the cassandra storage backend while handling other setups as well.

Fixes #2007 
Refs #2013 

**Checklist**
- No documentation added, as there are no changes regarding flags etc.

**Cases to cover**
- [x] Single endpoint; TLS enabled; Host verification enabled
- [x] Single endpoint; TLS enabled; Host verification disabled
- [x] Single endpoint; TLS disabled
- [x] Multiple endpoints; TLS enabled; Host verification disabled (Not tested; Should work in theory)
- [x] Multiple endpoints; TLS disabled (Not tested; Should work in theory)

According to my understanding, the following case will not work with the current gocql API design:
- Multiple endpoints; TLS enabled; Host verification enabled


joshmue:
@pracucci @VineethReddy02 

Do you think implementing and testing the above cases would be sufficient? I am a little bit concerned about the possibility of unknowingly breaking existing setups...

joshmue:
@pracucci @VineethReddy02 

I just squashed the previous commits to fewer ones and removed the draft status from the PR. Please note the updated PR description. I hadn't the time to test all cases but can't think of a way this code could break these untested cases.

Would you like to wait for implementation of #2137 or review this PR right away? Are there some other things to consider?

pkg/chunk/cassandra/storage_client.go
```diff
@@ -116,10 +117,23 @@ func (cfg *Config) session() (*gocql.Session, error) {
...
 
+	serverNames := strings.Split(cfg.Addresses, ",")
 	if cfg.SSL {
-		cluster.SslOpts = &gocql.SslOptions{
-			CaPath:                 cfg.CAPath,
-			EnableHostVerification: cfg.HostVerification,
+		if cfg.HostVerification {
+			if len(serverNames) != 1 {
+				return errors.New("host verification is only possible for a single host")
+			}
```
pracucci: Could you move this check to `cfg.Validate()`? It's based on `cfg.Addresses` and `cfg.HostVerification` so should be easy to do.

joshmue: Done!

pkg/chunk/cassandra/storage_client.go
```diff
@@ -117,9 +121,18 @@ func (cfg *Config) setClusterConfig(cluster *gocql.ClusterConfig) error {
...
-		cluster.SslOpts = &gocql.SslOptions{
-			CaPath:                 cfg.CAPath,
-			EnableHostVerification: cfg.HostVerification,
+		if cfg.HostVerification {
+			cluster.SslOpts = &gocql.SslOptions{
+				CaPath:                 cfg.CAPath,
+				EnableHostVerification: true,
+				Config: &tls.Config{
+					ServerName: strings.Split(cfg.Addresses, ",")[0],
+				},
```
pstibrany, line 130: This part is not required. When `ServerName` is set to "", then it will be inferred from hostname we're connecting to (in `tls.DialWithDialer` function). This also removes restriction that only single address can be used.

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@
 * [BUGFIX] Fixed etcd client keepalive settings. #2278
 * [BUGFIX] Fixed bug in updating last element of FIFO cache. #2270
 * [BUGFIX] Register the metrics of the WAL. #2295
+* [BUGFIX] Cassandra Storage: Fix endpoint TLS host verification. #2109
```
pstibrany, line 64: Please move this entry to new `master` section (after releasing 1.0).

pkg/chunk/cassandra/storage_client.go
woodsaj: ServerName must be set.  Gocql doesnt use tls.DialWithDialer.  It creates a TCP connection to the cassandra host, then if TLS is configured it upgrades the connection via tls.Client(conn, cfg.tslConfig).

the [crypto/tls docs](https://golang.org/pkg/crypto/tls/#Client) are clear that when using tls.Client(), "users must set either ServerName or InsecureSkipVerify in the config"

pstibrany: @woodsaj Thanks for explanation!

CHANGELOG.md
pstibrany: I've commited this change.

PR #2111, opened 2020-02-11
Remove extra newlines in imports and document Cortex formatting

pstibrany:
**What this PR does**: This is a follow-up to PR #2106, which removes extra newlines in imports (which reorders some groups) and adds section on Go files formatting into CONTRIBUTING.md guie.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


gouthamve:
Can you add this to lint step in CI? If we're going to codify it in docs, lets enforce it?

pstibrany:
> Can you add this to lint step in CI? If we're going to codify it in docs, lets enforce it?

I've added ordering check to the linter, but don't see any flags for extra new lines. Which is also why I was less strict about it in the doc ("We try to avoid extra newlines like that.")

PR #2113, opened 2020-02-11
Add support to horizontally scale blocks compactor

pracucci:
**What this PR does**:
This PR introduces the support to horizontally scale the compactor, when using the experimental blocks storage.

I've also took the change to remove `ingester` references from the ring HTTP page, given it's now used by multiple components (ingesters, ruler, compactor). The refactoring done here should be safe, given we've recently removed support for denormalised tokens.

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
   * `--experimental.distributor.user-subring-size`
 * [FEATURE] Added flag `-experimental.ruler.enable-api` to enable the ruler api which implements the Prometheus API `/api/v1/rules` and `/api/v1/alerts` endpoints under the configured `-http.prefix`. #1999
-* [FEATURE] Added sharing support to compactor when using the experimental TSDB blocks storage. #2113
+* [FEATURE] Added sharding support to compactor when using the experimental TSDB blocks storage. #2113
 * [ENHANCEMENT] Experimental TSDB: Export TSDB Syncer metrics from Compactor component, they are prefixed with `cortex_compactor_`. #2023
 * [ENHANCEMENT] Experimental TSDB: Added dedicated flag `-experimental.tsdb.bucket-store.tenant-sync-concurrency` to configure the maximum number of concurrent tenants for which blocks are synched. #2026
 * [ENHANCEMENT] Experimental TSDB: Expose metrics for objstore operations (prefixed with `cortex_<component>_thanos_objstore_`, component being one of `ingester`, `querier` and `compactor`). #2027
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
   * `--experimental.distributor.user-subring-size`
 * [FEATURE] Added flag `-experimental.ruler.enable-api` to enable the ruler api which implements the Prometheus API `/api/v1/rules` and `/api/v1/alerts` endpoints under the configured `-http.prefix`. #1999
+* [FEATURE] Added sharing support to compactor when using the experimental TSDB blocks storage. #2113
 * [ENHANCEMENT] Experimental TSDB: Export TSDB Syncer metrics from Compactor component, they are prefixed with `cortex_compactor_`. #2023
 * [ENHANCEMENT] Experimental TSDB: Added dedicated flag `-experimental.tsdb.bucket-store.tenant-sync-concurrency` to configure the maximum number of concurrent tenants for which blocks are synched. #2026
 * [ENHANCEMENT] Experimental TSDB: Expose metrics for objstore operations (prefixed with `cortex_<component>_thanos_objstore_`, component being one of `ingester`, `querier` and `compactor`). #2027
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"hash/fnv"
 	"path"
 	"strings"
 	"sync"
@@ -19,6 +20,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/compact/downsample"
 	"github.com/thanos-io/thanos/pkg/objstore"
 
+	"github.com/cortexproject/cortex/pkg/ring"
 	cortex_tsdb "github.com/cortexproject/cortex/pkg/storage/tsdb"
 	"github.com/cortexproject/cortex/pkg/util"
 )
@@ -33,6 +35,10 @@ type Config struct {
 	CompactionInterval   time.Duration            `yaml:"compaction_interval"`
 	CompactionRetries    int                      `yaml:"compaction_retries"`
 
+	// Compactors sharding.
+	ShardingEnabled bool       `yaml:"sharding_enabled"`
+	ShardingRing    RingConfig `yaml:"sharding_ring"`
+
 	// No need to add options to customize the retry backoff,
 	// given the defaults should be fine, but allow to override
 	// it in tests.
@@ -42,6 +48,8 @@ type Config struct {
 
 // RegisterFlags registers the Compactor flags.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.ShardingRing.RegisterFlags(f)
+
 	cfg.BlockRanges = cortex_tsdb.DurationList{2 * time.Hour, 12 * time.Hour, 24 * time.Hour}
 	cfg.retryMinBackoff = 10 * time.Second
 	cfg.retryMaxBackoff = time.Minute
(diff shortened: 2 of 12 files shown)


pkg/compactor/compactor.go
```diff
@@ -290,3 +361,43 @@ func (c *Compactor) discoverUsers(ctx context.Context) ([]string, error) {
...
+	_, _ = hasher.Write([]byte(userID))
+	userHash := hasher.Sum32()
+
+	// Check whether this compactor instance owns the user.
+	rs, err := c.ring.Get(userHash, ring.Read, []ring.IngesterDesc{})
+	if err != nil {
+		return false, err
+	}
+
+	if len(rs.Ingesters) != 1 {
```
thorfour: it's unfortunate it's named "ingesters" I understand this isn't your changes just commenting.

pkg/compactor/compactor.go
```diff
@@ -290,3 +361,43 @@ func (c *Compactor) discoverUsers(ctx context.Context) ([]string, error) {
...
+	}
+
+	return rs.Ingesters[0].Addr == c.ringLifecycler.Addr, nil
+}
+
+func (c *Compactor) waitRingActive() error {
+	for {
+		// Check if the ingester is ACTIVE in the ring and our ring client
+		// has detected it.
+		if rs, err := c.ring.GetAll(); err == nil {
```
thorfour: Is this check necessary? Won't `GetState` only return `ACTIVE` if it's in the ring?

pkg/compactor/compactor.go
pracucci: It is. A step at a time and we'll refactor the ring too 😉

pkg/compactor/compactor.go
pracucci: I'm not sure I'm getting your question. Here we need to check if the `ring` client (not the lifecycler) has detected the update (made by the lifecycler).

pracucci:
> I guess my only question is during a topology change is there not a way that two compactors can run on the same user? Ex. first compactor comes up, adds itself to the ring and starts compacting user 0. Next compactor starts up adds itself to the ring, and it now owns user 0 and will start compaction on that user, resulting int two compactors working on the same user.

@thorfour Definitely. This can happen "by design". After looking at the bucket compactor logic, doing some manual experiments and talking to Bartek (back in December) this looked to be not a real issue. If the same blocks for the same tenant are simultaneously compacted by two instances, the next compactor run will get rid of the extra blocks created due to simultaneous runs (at this point only 1 of the two compactors will run the next compaction).

pkg/compactor/compactor.go
thorfour: Disregard me, I need to go read the ring/lifecycler code again :)

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
   * `--experimental.distributor.user-subring-size`
 * [FEATURE] Added flag `-experimental.ruler.enable-api` to enable the ruler api which implements the Prometheus API `/api/v1/rules` and `/api/v1/alerts` endpoints under the configured `-http.prefix`. #1999
+* [FEATURE] Added sharing support to compactor when using the experimental TSDB blocks storage. #2113
```
pstibrany: sharding?

pkg/ring/http.go
```diff
@@ -59,21 +58,32 @@ const tpl = `
...
+
+			{{ if .ShowTokens }}
+				{{ range $i, $ing := .Ingesters }}
+					<h2>Instance: {{ .ID }}</h2>
+					Tokens:<br />
+					<ul>
+						{{ range $token := .Tokens }}
+							<li>{{ $token }}</li>
+						{{ end }}
+					</ul>
```
pstibrany: I think
```					
<p>
  {{ range $token := .Tokens}}
    {{ $token }}
  {{ end }}
</p>
```

would work fine, it's less markup (easier for browsers to handle), and one doesn't have to scroll huge page. They are still written one per line in HTML, so can be easily diff-ed (I did that a lot when testing memberlist code).

pkg/compactor/compactor_ring.go
```diff
@@ -0,0 +1,91 @@
...
+	cfg.InstanceInterfaceNames = []string{"eth0", "en0"}
+	f.Var((*flagext.Strings)(&cfg.InstanceInterfaceNames), "compactor.ring.instance-interface", "Name of network interface to read address from.")
+	f.StringVar(&cfg.InstanceAddr, "compactor.ring.instance-addr", "", "IP address to advertise in the ring.")
+	f.IntVar(&cfg.InstancePort, "compactor.ring.instance-port", 0, "Port to advertise in the ring (defaults to server.grpc-listen-port).")
+	f.StringVar(&cfg.InstanceID, "compactor.ring.instance-id", hostname, "Instance ID to register in the ring.")
+}
+
+// ToLifecyclerConfig returns a LifecyclerConfig based on the compactor
+// ring config.
+func (cfg *RingConfig) ToLifecyclerConfig() ring.LifecyclerConfig {
```
pstibrany: If we do refactor the ring / lifecycler code, I hope we can get rid of methods like this one and repeated config structures like `RingConfig`. (Just whining here, not actionable in this PR)

pkg/ring/http.go
```diff
@@ -118,7 +128,7 @@ func (r *Ring) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	sort.Strings(ingesterIDs)
 
 	ingesters := []interface{}{}
-	tokens, owned := countTokens(r.ringDesc, r.ringTokens)
```
pstibrany: I have found it useful in the past to be able to see original ring descriptor, as stored in KV store. Can we still expose it, similarly how we hide tokens by default?

pkg/ring/http.go
```diff
@@ -129,35 +139,30 @@ func (r *Ring) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+
+	if err := pageTemplate.Execute(w, struct {
 		Ingesters  []interface{}
 		Now        time.Time
-		Ring       string
 		ShowTokens bool
 	}{
 		Ingesters:  ingesters,
 		Now:        time.Now(),
-		Ring:       ringDescString,
```
pstibrany: I have found it useful in the past to be able to see original ring descriptor, as stored in KV store. Can we still expose it, similarly how we hide tokens by default?

[Moved comment to relevant line]

pracucci:
> Overall PR looks good to me, but I'm concerned about resharding. What if one compactor is working on a user, and at the same time, new compactor is added – and suddenly it's new compactor who owns user 1. Does first compactor detect and cancels the compaction? Or does it not upload compacted blocks in such case?

@pstibrany It doesn't detect or cancel the on-going compaction. There's no strong guarantee two replicas will not compact the same blocks. My take is this one:
https://github.com/cortexproject/cortex/pull/2113#issuecomment-584737468

pkg/ring/http.go
pracucci: The exact markup you posted will end up having all tokens in a single line, not one token per line. Is that what you want? What's the problem listing them like we're doing today?

pkg/ring/http.go
pracucci: > Can we still expose it [...] ?

I can even rollback to the previous logic, but was containing the "ingester" naming which is quite confusing to users. Which information would you like to see? I may add it without using the marshalling of the entire `RingDesc`.

pkg/ring/http.go
pstibrany: > The exact markup you posted will end up having all tokens in a single line, not one token per line. Is that what you want?

Yes

> What's the problem listing them like we're doing today?

Lot of wasted white space on the screen? :) Nothing wrong, but I don't see any benefit of doing that. [That said, it didn't bother me enough to change it before you started with changes here]

pkg/ring/http.go
pstibrany: Well, it was useful when I introduced new ingester state, which was unknown to previous versions, but I could see on this page that it was actually received correctly. In general, I want to see all fields of the protobuf message exactly as they are stored in KV.

No need to rollback, your change is easier to read when one is interested in tokens only. My use-case is probably very rare [even more than checking tokens, which is already very rare].

pkg/ring/http.go
pracucci: I agree. I've changed it to display tokens in a more compact way.

pracucci:
@pstibrany I should have addressed your comments. Do you have any other feedback?

pkg/ring/http.go
```diff
@@ -59,21 +58,30 @@ const tpl = `
...
-			<input type="button" value="Show Ingester Tokens" onclick="window.location.href = '?tokens=true'" />
+			<input type="button" value="Show Tokens" onclick="window.location.href = '?tokens=true'" />
+			{{ end }}
+
+			{{ if .ShowTokens }}
+				{{ range $i, $ing := .Ingesters }}
+					<h2>Instance: {{ .ID }}</h2>
+					<p>
+						Tokens:<br />
+						{{ range $token := .Tokens }}{{ $token }} {{ end }}
```
pstibrany: Can we please include each token on its own line in HTML? It's easier to diff different versions of HTML page that way.

pracucci: Sure, done.

PR #2116, opened 2020-02-12
add per-tenant alertmanager metrics

### LINKED ISSUE #1631: Tenanted Alertmanagers are not currently instrumented
Currently a non-active registry is passed to each alertmanager tenant:

```go
// New creates a new Alertmanager.
func New(cfg *Config) (*Alertmanager, error) {
	am := &Alertmanager{
		cfg:    cfg,
		logger: log.With(cfg.Logger, "user", cfg.UserID),
		stop:   make(chan struct{}),
	}

	// TODO(cortex): Build a registry that can merge metrics from multiple users.
	// For now, these metrics are ignored, as we can't register the same
	// metric twice with a single registry.
	localRegistry := prometheus.NewRegistry()
```

We should investigate ways to instrument each alertmanager since it would provide significant improvement in the observability of the alertmanager.

jtlisi:
**What this PR does**:

This PR takes advantage of the `util.MetricFamiliesPerUser` struct to provide per-tenant Alertmanager metrics.

Fixes #1631 

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/alertmanager/alertmanager_metrics.go
+++ b/pkg/alertmanager/alertmanager_metrics.go
@@ -53,23 +53,23 @@ func newAlertmanagerMetrics() *alertmanagerMetrics {
 		alertsReceived: prometheus.NewDesc(
 			"cortex_alertmanager_alerts_received_total",
 			"The total number of received alerts.",
-			[]string{"user", "version", "status"}, nil),
+			nil, nil),
 		alertsInvalid: prometheus.NewDesc(
 			"cortex_alertmanager_alerts_invalid_total",
 			"The total number of received alerts that were invalid.",
-			[]string{"user", "version"}, nil),
+			nil, nil),
 		numNotifications: prometheus.NewDesc(
 			"cortex_alertmanager_notifications_total",
 			"The total number of attempted notifications.",
-			[]string{"user", "integration"}, nil),
+			[]string{"user"}, nil),
 		numFailedNotifications: prometheus.NewDesc(
 			"cortex_alertmanager_notifications_failed_total",
 			"The total number of failed notifications.",
-			[]string{"user", "integration"}, nil),
+			[]string{"user"}, nil),
 		notificationLatencySeconds: prometheus.NewDesc(
 			"cortex_alertmanager_notification_latency_seconds",
 			"The latency of notifications in seconds.",
-			[]string{"integration"}, nil),
+			nil, nil),
 		nflogGCDuration: prometheus.NewDesc(
 			"cortex_alertmanager_nflog_gc_duration_seconds",
 			"Duration of the last notification log garbage collection cycle.",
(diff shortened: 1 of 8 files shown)


pkg/util/metrics_helper.go
```diff
@@ -150,6 +150,21 @@ func (d MetricFamiliesPerUser) SendSumOfCountersPerUser(out chan<- prometheus.Me
 	}
 }
 
+// SendSumOfCountersPerUserWithLabels provides metrics with the provided label names on a per-user basis. This function assumes that `user` is the
+// first label on the provided metric Desc
+func (d MetricFamiliesPerUser) SendSumOfCountersPerUserWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
```
pracucci: Please add unit tests on this function.

pkg/util/metrics_helper.go
```diff
@@ -162,6 +177,21 @@ func (d MetricFamiliesPerUser) SendSumOfGaugesWithLabels(out chan<- prometheus.M
 	d.sumOfSingleValuesWithLabels(gauge, gaugeValue, labelNames).WriteToMetricChannel(out, desc, prometheus.GaugeValue)
 }
 
+// SendSumOfGaugesPerUserWithLabels provides metrics with the provided label names on a per-user basis. This function assumes that `user` is the
+// first label on the provided metric Desc
+func (d MetricFamiliesPerUser) SendSumOfGaugesPerUserWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
```
pracucci: Please add unit tests on this function.

pkg/util/metrics_helper.go
```diff
@@ -150,6 +150,21 @@ func (d MetricFamiliesPerUser) SendSumOfCountersPerUser(out chan<- prometheus.Me
 	}
 }
 
+// SendSumOfCountersPerUserWithLabels provides metrics with the provided label names on a per-user basis. This function assumes that `user` is the
+// first label on the provided metric Desc
+func (d MetricFamiliesPerUser) SendSumOfCountersPerUserWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
+	for user, userMetrics := range d {
+		metricsPerLabelValue := getMetricsWithLabelNames(userMetrics[metric], labelNames)
+		for _, mlv := range metricsPerLabelValue {
+			for _, m := range mlv.metrics {
```
pracucci: The function name is `SendSumOfCountersPerUserWithLabels()` so if you have more than one `mlv.metrics` I would expect it sum them and the writes to `out` only 1 metric. On the contrary, I think with the current logic we will have clashing series (same exact series reported multiple times) and should be fixed. This case should be covered with a unit test too.

_ SendSumOfGaugesPerUserWithLabels_

pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	silencesSnapshotDuration        *prometheus.Desc
+	silencesSnapshotSize            *prometheus.Desc
+	silencesQueriesTotal            *prometheus.Desc
+	silencesQueryErrorsTotal        *prometheus.Desc
+	silencesQueryDuration           *prometheus.Desc
+	silences                        *prometheus.Desc
+	silencesPropagatedMessagesTotal *prometheus.Desc
+}
+
+func newAlertmanagerMetrics() *alertmanagerMetrics {
```
pracucci: What's the impact, in terms of cardinality, for all the metrics with the `user` label? I'm wondering if this change may potentially lead to a series explosion.

For example, the metrics with `integration` can have 9 integrations, so the cardinality is 10x the number of tenants for each of such series (x the number of alert manager instances).

pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	out <- m.silencesQueryDuration
+	out <- m.silences
+	out <- m.silencesPropagatedMessagesTotal
+}
+
+func (m *alertmanagerMetrics) Collect(out chan<- prometheus.Metric) {
+	data := util.BuildMetricFamiliesPerUserFromUserRegistries(m.registries())
+
+	data.SendSumOfCountersPerUserWithLabels(out, m.alertsReceived, "alertmanager_alerts_received_total", "version", "status")
+	data.SendSumOfCountersPerUserWithLabels(out, m.alertsInvalid, "alertmanager_alerts_invalid_total", "version")
```
pracucci: Do we need the version for these two metrics?

pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	data.SendSumOfCountersPerUserWithLabels(out, m.numNotifications, "alertmanager_notifications_total", "integration")
+	data.SendSumOfCountersPerUserWithLabels(out, m.numFailedNotifications, "alertmanager_notifications_failed_total", "integration")
+	data.SendSumOfCountersWithLabels(out, m.notificationLatencySeconds, "alertmanager_notification_latency_seconds", "integration")
```
pracucci: `alertmanager_notification_latency_seconds ` is an histogram. How could `SendSumOfCountersWithLabels` even work? Please update the unit test to see it working.

jtlisi: My mistake, I overlooked this and used the wrong function. I'll use the correct function and drop the labels entirely.

pkg/util/metrics_helper.go
jtlisi: 👍

pkg/util/metrics_helper.go
jtlisi: 👍

pkg/util/metrics_helper.go
jtlisi: 👍

pkg/alertmanager/alertmanager_metrics.go
jtlisi: Nice to have but not required.

pkg/alertmanager/alertmanager_metrics.go
jtlisi: I paired down the number of metrics that are reporting a `user` label and removed the `integration` entirely.

jtlisi:
Closing to switch to #2124 which is not source in the Grafana Org which is unable to run integration tests in circleci due to env variables set at the org level.

PR #2119, opened 2020-02-12
Modules can now start additional startup tasks, while server is starting

pstibrany:
Modules can now have start function for starting additional tasks on startup.

Start functions are called concurrently, and only after all modules are initialized. Server's HTTP service starts while modules are still starting, so modules must be able to handle that. Both ingester and queriers will return error.

This is currently used by blocks-based ingester (to replay WAL for locally stored TSDBs) and querier (to do initial fetch of blocks).

This is alternative to https://github.com/cortexproject/cortex/pull/2104.

Note: this is not ideal, but is relatively simple change. I think we should overhaul the lifecycle of our modules, and have separate NEW (instantiated, but nothing started yet), STARTING (replaying WALs, fetching blocks, whatever needs to be done for service to be ready) and RUNNING (service is running) states and make those states observable by other modules (for dependencies and for global view on the server, plus make those states observable via metrics / simple HTML page), but this PR doesn't go that far. I'll write a design document for that first.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,6 +67,7 @@ require (
 	github.com/weaveworks/common v0.0.0-20200201141823-27e183090ab1
 	go.etcd.io/bbolt v1.3.3
 	go.etcd.io/etcd v0.0.0-20190709142735-eb7dd97135a5
+	go.uber.org/atomic v1.5.0
 	golang.org/x/net v0.0.0-20191112182307-2180aed22343
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 	google.golang.org/api v0.14.0
--- a/go.mod
+++ b/go.mod
@@ -69,6 +69,7 @@ require (
 	go.etcd.io/etcd v0.0.0-20190709142735-eb7dd97135a5
 	go.uber.org/atomic v1.5.0
 	golang.org/x/net v0.0.0-20191112182307-2180aed22343
+	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 	google.golang.org/api v0.14.0
 	google.golang.org/grpc v1.25.1
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -1,9 +1,11 @@
 package cortex
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
+	"sync"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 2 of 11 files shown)


pkg/cortex/cortex.go
```diff
@@ -247,6 +249,42 @@ func (t *Cortex) init(cfg *Config, m moduleName) error {
 	return t.initModule(cfg, m)
 }
 
+// calls start functions of modules. Functions are called concurrently.
+func (t *Cortex) start(target moduleName) error {
+	deps := orderedDeps(target)
+
+	wg := sync.WaitGroup{}
```
jtlisi: Since you are only returning errors you should consider using a [`sync.errgroup`](https://godoc.org/golang.org/x/sync/errgroup#Group.Go) instead of the waitgroup and channel.

jtlisi:
@pstibrany Can you rebase, and check the integration test?

pkg/cortex/cortex.go
pstibrany: Thanks, looks exactly like what I need.

pstibrany:
Before approving this, I will check return values from gRPC methods to make sure we send the correct errors and status codes back.

pstibrany:
> Withdrawing approval til integration tests are sorted

Integration tests found a real bug (start was only called on dependencies, but not target module itself), which is now fixed.

pstibrany:
> Before approving this, I will check return values from gRPC methods to make sure we send the correct errors and status codes back.

After some reading, I believe returning `status.Error(codes.Unavailable, message)` is the right thing to do from gRPC handlers, when ingester or querier are not ready yet. When client receives [Unavailable](https://godoc.org/google.golang.org/grpc/codes#Code), it is expected to retry (with some backoff), which is exactly what we want to express here.

I've updated the code accordingly.

pstibrany:
> After some reading, I believe returning `status.Error(codes.Unavailable, message)` is the right thing to do from gRPC handlers, when ingester or querier are not ready yet. When client receives [Unavailable](https://godoc.org/google.golang.org/grpc/codes#Code), it is expected to retry (with some backoff), which is exactly what we want to express here.
> 
> I've updated the code accordingly.

Of course, ingester doesn't even register itself into the Ring, until it is ready.

With blocks store situation is more complicated... both querier and ruler that use it are started immediately, and don't wait for "block store" to be ready first. All requests to the block store will however fail until it is. This may cause troubles and I will investigate more how to solve it. Ideally they should observe it's component for readiness before using it.

go.mod
```diff
@@ -67,7 +67,9 @@ require (
 	github.com/weaveworks/common v0.0.0-20200201141823-27e183090ab1
 	go.etcd.io/bbolt v1.3.3
 	go.etcd.io/etcd v0.0.0-20190709142735-eb7dd97135a5
+	go.uber.org/atomic v1.5.0
```
codesome: Instead of this additional dependency, can we use https://golang.org/pkg/sync/atomic/ for atomic loads and stores?

gouthamve:
@pstibrany Is the querier and ruler using the ingester only a problem in single binary mode? Can we not register the ingester in the ring (or mark it pending) until it starts up?

go.mod
pstibrany: We already have this dependency, although under old name (github.com/uber-go/atomic). I've removed old version in some PR #2127.

pstibrany:
> @pstibrany Is the querier and ruler using the ingester only a problem in single binary mode? Can we not register the ingester in the ring (or mark it pending) until it starts up?

Ruler isn't using ingester. Querier is, but only indirectly, via gRPC calls.

Problem that this PR is solving for QuerierChunkStore is that blocks-based QuerierChunkStore needs to fetch blocks metadata on startup, which can take a while. However, Ruler is using QuerierChunkStore and has no way to see whether QuerierChunkStore has finished its fetching and is now ready to use, or not.

I'd like to solve this in a general way, and will send new PR showing how to do that. I'd suggest to  keep this PR open for now, so that we can compare both approaches and see which one works better. (I don't really like approach in this PR, to be honest)

pstibrany:
I have sent alternative PR to this one, #2166. It revamps services module and supports everything that we need to correctly implement in this PR, in a much better way. If you have time, please take a look.

gouthamve:
Closing this in favour of #2166

PR #2121, opened 2020-02-12
Stop table-manager loop early, if it is done

pstibrany:
This is trivial fix that helps integration tests to run slightly faster, as discussed at https://github.com/cortexproject/cortex/pull/2016#pullrequestreview-357570026.

If table-manager is stopped during initial wait, it now exits early.

### DIFF
--- a/integration/framework/service.go
+++ b/integration/framework/service.go
@@ -221,6 +221,8 @@ func (s *Service) WaitReady() error {
 }
 
 func (s *Service) WaitMetric(port int, metric string, value float64) error {
+	lastFoundValue := 0.0
+
 	for s.retryBackoff.Reset(); s.retryBackoff.Ongoing(); {
 		metrics, err := s.Metrics(80)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pstibrany:
> Thanks @pstibrany. This PR looks based on a fairly old version of `master`. May you rebase it, please?

Done. How did you even find out? You had to manually check my source branch I guess :)

pracucci:
> How did you even find out?

I checked out the PR to test it. I can see a failed integration test in CI. May you take a look, please?

pstibrany:
> Thanks!

Worth noting: I haven't fixed the failing integration test. I think single-binary test with memberlist is flaky. I just added observed value into the error message. That said, I think it's unrelated to this change.

pracucci:
> Worth noting: I haven't fixed the failing integration test. I think single-binary test with memberlist is flaky. I just added observed value into the error message. That said, I think it's unrelated to this change.

Sure. But the log you've added may help to investigate it, as soon as we're able to reproduce it.

PR #2123, opened 2020-02-12
Generalise framework to be reusable for Thanos as well.

bwplotka:
We love this functionality! Especially asserting on metric, amazing! Super clean.

Thanks for this @pracucci!

NOTE: We are in progress in trying this out to [replace Thanos e2e engine](https://github.com/thanos-io/thanos/tree/master/test/e2e), which is schedules Linux processes, so it's not cross-platform compatible really and quite messy. Also, docker-engine @pracucci built allows better visibility from outside (e.g memory utilization metrics). Anyway, work in progress, we might tweak this PR a bit until we are done (: Review welcome anyway, if we are going right direction!

## Changes

* Rename framework to e2e. `framework` to me does not mean for what framework. Happy to discuss naming here :D
* Split cortex stuff to e2ecortex
* Split DBs services to e2edb I think it's useful to share those.
* Added various TODOs

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


GiedriusS:
This framework looks awesome! We could definitely use it over at Thanos. Gave it a quick cursory look and it could probably be improved a lot performance-wise if we'd talk with the Docker socket directly. So that could be another TODO in the list.

bwplotka:
Sure, curious why `framework`? It does not tell anything. I am not complaining, the package is awesome but it's like naming a package `code` (:

pracucci:
> Sure, curious why framework? It does not tell anything. I am not complaining, the package is awesome but it's like naming a package code (:

Well, within the `integration/` tests it's the `framework` used to run tests. I'm not against renaming it, but not in this PR to ease the diff. If you prefer, you can submit a dedicated PR just to rename the package.

bwplotka:
Anyway, done.

integration/integration_memberlist_single_binary_test.go
```diff
@@ -68,19 +71,19 @@ func startSingleBinary(s *framework.Scenario, name string, join string) error {
...
 		name,
-		framework.GetDefaultCortexImage(),
-		framework.NetworkName,
+		e2ecortex.GetDefaultImage(),
+		e2e.NetworkName,
 		[]int{80, 8000},
 		nil,
-		framework.NewCommandWithoutEntrypoint("cortex", framework.BuildArgs(framework.MergeFlags(ChunksStorage, flags))...),
-		framework.NewReadinessProbe(80, "/ready", 204),
+		e2e.NewCommandWithoutEntrypoint("cortex", e2e.BuildArgs(e2e.MergeFlags(ChunksStorage, flags))...),
```
pracucci: - `e2e.MergeFlags` > `mergeFlags`
- What do you think to map `buildArgs = e2e.BuildArgs` in `integration/util.go` as well?

integration/framework/scenario.go
```diff
@@ -258,7 +140,23 @@ func (s *Scenario) WaitReady(services ...string) error {
...
+func (s *Scenario) Close() {
+	if s == nil {
+		return
+	}
+
+	s.shutdown()
+	s.clean()
+}
+
+// TODO(bwplotka): Add comments. Return err?
```
pracucci: Previously, the `Shutdown()` didn't intentionally return err (but was just logging it) because even if it returns err there's no much the caller can do.

integration/framework/scenario.go
```diff
@@ -56,15 +83,16 @@ func (s *Scenario) Endpoint(name string, port int) string {
 }
 
 func (s *Scenario) StartService(service *Service) error {
+	// TODO(bwplotka): Some basic logger would be nice.
```
pracucci: Yes, in a separate PR 😄

integration/framework/scenario.go
```diff
@@ -1,41 +1,68 @@
...
-	MinioSecretKey = "supersecret"
-	NetworkName    = "cortex-integration-test"
+	NetworkName        = "e2e-test"
+	ContainerSharedDir = "/shared"
 )
 
 type Scenario struct {
 	services []*Service
+
+	dir string
```
pracucci: `dir` > `localSharedDir`

integration/framework/scenario.go
```diff
@@ -1,41 +1,68 @@
...
 	}
 
+	if os.Getenv("E2E_INTEGRATION_TEST_DIR") != "" {
+		s.dir = os.Getenv("E2E_INTEGRATION_TEST_DIR")
+		if err := os.MkdirAll(s.dir, os.ModePerm); err != nil {
+			return nil, err
+		}
+
+	} else {
+		tmpDir, err := ioutil.TempDir(".", "e2e_integration_test")
```
pracucci: What do you think about using `os.TempDir()` instead of `.`? Is there a good reason behind `.`?

integration/framework/scenario.go
```diff
@@ -1,41 +1,68 @@
...
+
 	// Force a shutdown in order to cleanup from a spurious situation in case
 	// the previous tests run didn't cleanup correctly
-	s.Shutdown()
+	s.shutdown()
 
 	// Setup the docker network
 	if out, err := RunCommandAndGetOutput("docker", "network", "create", NetworkName); err != nil {
 		fmt.Println(string(out))
+		_ = os.RemoveAll(s.dir)
```
pracucci: Is this require, given we now call `s.Close()` even in case `NewScenario()` returns error?

integration/configs.go
```diff
@@ -20,7 +42,11 @@ var (
 	ChunksStorage = map[string]string{
 		"-dynamodb.url":                   "dynamodb://u:p@dynamodb.:8000",
 		"-dynamodb.poll-interval":         "1m",
-		"-config-yaml":                    "/integration/chunks-storage-schema-dynamodb.yaml",
+		"-config-yaml":                    filepath.Join(e2e.RelativeContainerSharedDir, "shared", cortexConfigFile),
 		"-table-manager.retention-period": "168h",
 	}
 )
+
+func writeCortexConfigFile(s *e2e.Scenario) error {
```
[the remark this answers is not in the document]
bwplotka: Then there is no point in this function, will just inline.

integration/framework/db/kvstores.go
```diff
@@ -0,0 +1,20 @@
+package e2edb
```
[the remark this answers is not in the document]
bwplotka: Hm, don't have strong opinion either. Happy to change, does not matter now. I would maybe even do `db.go` as the functions are super small, WDYT?

integration/framework/db/objstores.go
```diff
@@ -0,0 +1,29 @@
...
+		e2e.NetworkName,
+		[]int{9000},
+		map[string]string{
+			"MINIO_ACCESS_KEY": MinioAccessKey,
+			"MINIO_SECRET_KEY": MinioSecretKey,
+			"MINIO_BROWSER":    "off",
+			"ENABLE_HTTPS":     "0",
+		},
+		// Create the "cortex" bucket before starting minio
+		e2e.NewCommandWithoutEntrypoint("sh", "-c", "mkdir -p /data/e2e-minio && minio server --quiet /data"),
```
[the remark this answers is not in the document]
bwplotka: Why this matters?  The storage is ephemeral anyway.

integration/framework/scenario.go
```diff
@@ -1,41 +1,68 @@
...
+
+	dir string
 }
 
 func NewScenario() (*Scenario, error) {
 	s := &Scenario{
 		services: []*Service{},
 	}
 
+	if os.Getenv("E2E_INTEGRATION_TEST_DIR") != "" {
```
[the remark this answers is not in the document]
bwplotka: no, I thought you need this.

integration/framework/db/objstores.go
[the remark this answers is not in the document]
pracucci: Cortex expects the bucket to be already created (even if emphemeral and empty). We need `/data/<bucket-name>` directory to be there, otherwise Cortex will fail.

integration/integration_memberlist_single_binary_test.go
bwplotka: If I can be honest this practice is quite surprising to me. But it's your e2e test, so happy to add those awkward vars. This is for... shorter name?

integration/configs.go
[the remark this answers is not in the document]
pracucci: Inlining it will be a bit more ugly. We can keep the function with params in input.

integration/framework/scenario.go
bwplotka: fair

integration/framework/db/kvstores.go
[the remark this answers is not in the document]
pracucci: Then having the package `integration/framework/db` just containing `db.go` looks another overengineering at this stage. Why not moving it back into a single file inside `integration/framework/`?

integration/integration_memberlist_single_binary_test.go
pracucci: It was for shorter names. With the `e2e` package naming it's less a problem. You can remove the mapping at all.

integration/framework/scenario.go
bwplotka: I was fighting with permission issues that's it ;p  `/tmp` would do.

integration/framework/scenario.go
bwplotka: > ow call s.Close() even in case NewScenario() returns error?

We don't as scenario is nil.

pracucci: Right, then we can call `s.clean()` here, right?

bwplotka:
In separate PR I would consider actually tying those scenarios to `t *testing.T` so we don't need to have those repetitions for `require.NoError`

I might start with just separate struct I guess that wraps scenario as a start (:

pracucci:
> In separate PR I would consider actually tying those scenarios to t *testing.T so we don't need to have those repetitions for require.NoError

Agree. I was actually thinking the same.

pracucci:
> I might start with just separate struct I guess that wraps scenario as a start (:

Yes. Wrapping the scenario would also allow to have custom logic directly in the Scenario, not tied to the framework.

bwplotka:
Another improvement would be to avoid strings in referencing services (again, for further PR). Either use vars or constants.

integration/README.md
```diff
@@ -3,8 +3,9 @@
 ## Supported environment variables
 
 - `CORTEX_IMAGE`: Docker image used to run Cortex in integration tests (defaults to `quay.io/cortexproject/cortex:latest`)
-- `CORTEX_INTEGRATION_DIR`: Absolute path to the integration/ directory on the host (defaults to `$GOPATH/src/github.com/cortexproject/cortex/integration`)
 
 ## Mounting files to Cortex container
 
-The `integration/` directory in the repository is mounted to `/integration` within the container. If you need to add fixture files (ie. config file) you can add the files to this repository (in `integration/`) and the files will be available within the Cortex container at `/integration`.
+The `integration/` directory in the repository is mounted to `/integration` within the container.
```
pracucci: This is no more correct. We can remove this entire section.

integration/framework/scenario.go
bwplotka: :+1:

integration/framework/db/objstores.go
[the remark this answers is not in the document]
bwplotka: Wowow, I did not know just creating directory means creating bucket 0.o fair :+1:

PR #2124, opened 2020-02-12
add per-tenant alertmanager metrics

### LINKED ISSUE #1631: Tenanted Alertmanagers are not currently instrumented

jtlisi:
Moved from #2116 because integration tests don't work for Grafana org repo PRs due to `NOQUAY` being set as an environment variable.

**What this PR does**:

This PR takes advantage of the `util.MetricFamiliesPerUser` struct to provide per-tenant Alertmanager metrics.

Fixes #1631 

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -244,7 +244,7 @@ func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
 	return nil
 }
 
-func (am *Alertmanager) isActive() bool {
+func (am *Alertmanager) IsActive() bool {
 	am.activeMtx.Lock()
 	defer am.activeMtx.Unlock()
 	return am.active
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -176,11 +176,6 @@ func clusterWait(p *cluster.Peer, timeout time.Duration) func() time.Duration {
 
 // ApplyConfig applies a new configuration to an Alertmanager.
 func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
-	// Ensure the alertmanager is set to active
-	am.activeMtx.Lock()
-	am.active = true
-	am.activeMtx.Unlock()
-
 	templateFiles := make([]string, len(conf.Templates))
 	if len(conf.Templates) > 0 {
 		for i, t := range conf.Templates {
@@ -241,25 +236,37 @@ func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
 	go am.dispatcher.Run()
 	go am.inhibitor.Run()
 
+	// Ensure the alertmanager is set to active
+	am.activeMtx.Lock()
+	am.active = true
+	am.activeMtx.Unlock()
+
 	return nil
 }
 
+func (am *Alertmanager) isActive() bool {
+	am.activeMtx.Lock()
+	defer am.activeMtx.Unlock()
+	return am.active
+}
+
 // Pause running jobs in the alertmanager that are able to be restarted and sets
-// to inactive
+// to inactives
 func (am *Alertmanager) Pause() {
 	// Set to inactive
 	am.activeMtx.Lock()
 	am.active = false
 	am.activeMtx.Unlock()
 
-	// Ensure inhibitor is set before being called
+	// Stop the inhibitor and dispatcher which will be recreated when
+	// a new config is applied
 	if am.inhibitor != nil {
 		am.inhibitor.Stop()
+		am.inhibitor = nil
 	}
-
-	// Ensure dispatcher is set before being called
 	if am.dispatcher != nil {
 		am.dispatcher.Stop()
+		am.dispatcher = nil
 	}
 
 	// Remove all of the active silences from the alertmanager
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -244,6 +244,8 @@ func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
 	return nil
 }
 
+// IsActive returns if the alertmanager is currently running
+// or is paused
 func (am *Alertmanager) IsActive() bool {
 	am.activeMtx.Lock()
 	defer am.activeMtx.Unlock()
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/alertmanager/api"
 	"github.com/prometheus/alertmanager/cluster"
 	"github.com/prometheus/alertmanager/config"
@@ -66,6 +67,9 @@ type Alertmanager struct {
 	wg              sync.WaitGroup
 	mux             *http.ServeMux
 	registry        *prometheus.Registry
+
+	activeMtx sync.Mutex
+	active    bool
 }
 
 var webReload = make(chan chan error)
@@ -81,17 +85,16 @@ func init() {
 }
 
 // New creates a new Alertmanager.
-func New(cfg *Config) (*Alertmanager, error) {
+func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 	am := &Alertmanager{
-		cfg:    cfg,
-		logger: log.With(cfg.Logger, "user", cfg.UserID),
-		stop:   make(chan struct{}),
+		cfg:       cfg,
+		logger:    log.With(cfg.Logger, "user", cfg.UserID),
+		stop:      make(chan struct{}),
+		active:    false,
+		activeMtx: sync.Mutex{},
 	}
 
-	// TODO(cortex): Build a registry that can merge metrics from multiple users.
-	// For now, these metrics are ignored, as we can't register the same
-	// metric twice with a single registry.
-	am.registry = prometheus.NewRegistry()
+	am.registry = reg
 
 	am.wg.Add(1)
 	nflogID := fmt.Sprintf("nflog:%s", cfg.UserID)
@@ -173,7 +176,12 @@ func clusterWait(p *cluster.Peer, timeout time.Duration) func() time.Duration {
 
 // ApplyConfig applies a new configuration to an Alertmanager.
 func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
-	templateFiles := make([]string, len(conf.Templates))
+	// Ensure the alertmanager is set to active
+	am.activeMtx.Lock()
+	am.active = true
+	am.activeMtx.Unlock()
+
+	templateFiles := make([]string, len(conf.Templates), len(conf.Templates))
 	if len(conf.Templates) > 0 {
 		for i, t := range conf.Templates {
 			templateFiles[i] = filepath.Join(am.cfg.DataDir, "templates", userID, t)
@@ -236,9 +244,47 @@ func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
 	return nil
 }
 
+// Pause running jobs in the alertmanager that are able to be restarted and sets
+// to inactive
+func (am *Alertmanager) Pause() {
+	// Set to inactive
+	am.activeMtx.Lock()
+	am.active = false
+	am.activeMtx.Unlock()
+
+	// Ensure inhibitor is set before being called
+	if am.inhibitor != nil {
+		am.inhibitor.Stop()
+	}
+
+	// Ensure dispatcher is set before being called
+	if am.dispatcher != nil {
+		am.dispatcher.Stop()
+	}
+
+	// Remove all of the active silences from the alertmanager
+	silences, _, err := am.silences.Query()
+	if err != nil {
+		level.Warn(am.logger).Log("msg", "unable to retrieve silences for removal", "err", err)
+	}
+	for _, si := range silences {
+		err = am.silences.Expire(si.Id)
+		if err != nil {
+			level.Warn(am.logger).Log("msg", "unable to remove silence", "err", err, "silence", si.Id)
+		}
+	}
+}
+
 // Stop stops the Alertmanager.
 func (am *Alertmanager) Stop() {
-	am.dispatcher.Stop()
+	if am.inhibitor != nil {
+		am.inhibitor.Stop()
+	}
+
+	if am.dispatcher != nil {
+		am.dispatcher.Stop()
+	}
+
 	am.alerts.Close()
 	close(am.stop)
 	am.wg.Wait()
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -181,7 +181,7 @@ func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
 	am.active = true
 	am.activeMtx.Unlock()
 
-	templateFiles := make([]string, len(conf.Templates), len(conf.Templates))
+	templateFiles := make([]string, len(conf.Templates))
 	if len(conf.Templates) > 0 {
 		for i, t := range conf.Templates {
 			templateFiles[i] = filepath.Join(am.cfg.DataDir, "templates", userID, t)
--- a/pkg/alertmanager/alertmanager_metrics.go
+++ b/pkg/alertmanager/alertmanager_metrics.go
@@ -54,11 +54,11 @@ func newAlertmanagerMetrics() *alertmanagerMetrics {
 		alertsReceived: prometheus.NewDesc(
 			"cortex_alertmanager_alerts_received_total",
 			"The total number of received alerts.",
-			nil, nil),
+			[]string{"user"}, nil),
 		alertsInvalid: prometheus.NewDesc(
 			"cortex_alertmanager_alerts_invalid_total",
 			"The total number of received alerts that were invalid.",
-			nil, nil),
+			[]string{"user"}, nil),
 		numNotifications: prometheus.NewDesc(
 			"cortex_alertmanager_notifications_total",
 			"The total number of attempted notifications.",
@@ -183,8 +183,8 @@ func (m *alertmanagerMetrics) Describe(out chan<- *prometheus.Desc) {
 func (m *alertmanagerMetrics) Collect(out chan<- prometheus.Metric) {
 	data := util.BuildMetricFamiliesPerUserFromUserRegistries(m.registries())
 
-	data.SendSumOfCounters(out, m.alertsReceived, "alertmanager_alerts_received_total")
-	data.SendSumOfCounters(out, m.alertsInvalid, "alertmanager_alerts_invalid_total")
+	data.SendSumOfCountersPerUser(out, m.alertsReceived, "alertmanager_alerts_received_total")
+	data.SendSumOfCountersPerUser(out, m.alertsInvalid, "alertmanager_alerts_invalid_total")
 
 	data.SendSumOfCountersPerUser(out, m.numNotifications, "alertmanager_notifications_total")
 	data.SendSumOfCountersPerUser(out, m.numFailedNotifications, "alertmanager_notifications_failed_total")
--- a/pkg/alertmanager/alertmanager_metrics.go
+++ b/pkg/alertmanager/alertmanager_metrics.go
@@ -1,0 +1,210 @@
+package alertmanager
+
+import (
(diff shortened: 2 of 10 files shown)


jtlisi:
@pracucci I paired down the number of user metrics considerably. The ones that remain convey either essential basic information about the number of alerts/silences or help users avoid silent failures.

`cortex_alertmanager_alerts` -- Conveys the number of alerts by state on a user basis. Basic information about the performance of cortex on a per-user level.

`cortex_alertmanager_silences` -- Same as above

`cortex_alertmanager_notifications_total` & `cortex_alertmanager_notifications_failed_total` -- These metrics are vital for detecting a silent failure in a users config. A user could have a valid Alertmanager config with a misconfigured integration. This could easily lead to a silent failure where a user is not receiving alerts. Having this metric on a per-user basis makes it easy to configure redundant alerts in either a separate Prometheus/Alertmanager system or using something like Prometheus/(Grafana Alerts).

pkg/util/metrics_helper.go
```diff
@@ -162,6 +162,22 @@ func (d MetricFamiliesPerUser) SendSumOfGaugesWithLabels(out chan<- prometheus.M
 	d.sumOfSingleValuesWithLabels(gauge, gaugeValue, labelNames).WriteToMetricChannel(out, desc, prometheus.GaugeValue)
 }
 
+// SendSumOfGaugesPerUserWithLabels provides metrics with the provided label names on a per-user basis. This function assumes that `user` is the
+// first label on the provided metric Desc
+func (d MetricFamiliesPerUser) SendSumOfGaugesPerUserWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
```
pracucci: @pstibrany Could you review this function, please?

pkg/util/metrics_helper_test.go
```diff
@@ -81,3 +82,110 @@ func makeLabels(namesAndValues ...string) []*dto.LabelPair {
 
 	return out
 }
+
+// TestSendSumOfGaugesPerUserWithLabels tests to ensure multiple metrics for the same user with a matching label are
+// summed correctly
+func TestSendSumOfGaugesPerUserWithLabels(t *testing.T) {
```
pracucci: I think this test is too much complicated and quite difficult to follow. I've tried to rewritten it in an easier form. Take a look at the following, please (pay attention that `require.ElementsMatch()` is intentional to avoid flaky tests, because output ordering is not guaranteed to the use of a map).

```
func TestSendSumOfGaugesPerUserWithLabels(t *testing.T) {
	user1Metric := prometheus.NewGaugeVec(prometheus.GaugeOpts{Name: "test_metric"}, []string{"label_one", "label_two"})
	user2Metric := prometheus.NewGaugeVec(prometheus.GaugeOpts{Name: "test_metric"}, []string{"label_one", "label_two"})
	user1Metric.WithLabelValues("a", "b").Set(100)
	user1Metric.WithLabelValues("a", "c").Set(80)
	user2Metric.WithLabelValues("a", "b").Set(60)
	user2Metric.WithLabelValues("a", "c").Set(40)

	user1Reg := prometheus.NewRegistry()
	user2Reg := prometheus.NewRegistry()
	user1Reg.MustRegister(user1Metric)
	user2Reg.MustRegister(user2Metric)

	mf := BuildMetricFamiliesPerUserFromUserRegistries(map[string]*prometheus.Registry{
		"user-1": user1Reg,
		"user-2": user2Reg,
	})

	desc := prometheus.NewDesc("test_metric", "", []string{"user", "label_one"}, nil)
	actual := collectMetrics(func(out chan prometheus.Metric) {
		mf.SendSumOfGaugesPerUserWithLabels(out, desc, "test_metric", "label_one")
	})
	expected := []*dto.Metric{
		{Label: makeLabels("label_one", "a", "user", "user-1"), Gauge: &dto.Gauge{Value: proto.Float64(180)}},
		{Label: makeLabels("label_one", "a", "user", "user-2"), Gauge: &dto.Gauge{Value: proto.Float64(100)}},
	}
	require.ElementsMatch(t, expected, actual)

	desc = prometheus.NewDesc("test_metric", "", []string{"user", "label_two"}, nil)
	actual = collectMetrics(func(out chan prometheus.Metric) {
		mf.SendSumOfGaugesPerUserWithLabels(out, desc, "test_metric", "label_two")
	})
	expected = []*dto.Metric{
		{Label: makeLabels("label_two", "b", "user", "user-1"), Gauge: &dto.Gauge{Value: proto.Float64(100)}},
		{Label: makeLabels("label_two", "c", "user", "user-1"), Gauge: &dto.Gauge{Value: proto.Float64(80)}},
		{Label: makeLabels("label_two", "b", "user", "user-2"), Gauge: &dto.Gauge{Value: proto.Float64(60)}},
		{Label: makeLabels("label_two", "c", "user", "user-2"), Gauge: &dto.Gauge{Value: proto.Float64(40)}},
	}
	require.ElementsMatch(t, expected, actual)

	desc = prometheus.NewDesc("test_metric", "", []string{"user", "label_one", "label_two"}, nil)
	actual = collectMetrics(func(out chan prometheus.Metric) {
		mf.SendSumOfGaugesPerUserWithLabels(out, desc, "test_metric", "label_one", "label_two")
	})
	expected = []*dto.Metric{
		{Label: makeLabels("label_one", "a", "label_two", "b", "user", "user-1"), Gauge: &dto.Gauge{Value: proto.Float64(100)}},
		{Label: makeLabels("label_one", "a", "label_two", "c", "user", "user-1"), Gauge: &dto.Gauge{Value: proto.Float64(80)}},
		{Label: makeLabels("label_one", "a", "label_two", "b", "user", "user-2"), Gauge: &dto.Gauge{Value: proto.Float64(60)}},
		{Label: makeLabels("label_one", "a", "label_two", "c", "user", "user-2"), Gauge: &dto.Gauge{Value: proto.Float64(40)}},
	}
	require.ElementsMatch(t, expected, actual)
}

func collectMetrics(send func(out chan prometheus.Metric)) []*dto.Metric {
	out := make(chan prometheus.Metric)

	go func() {
		send(out)
		close(out)
	}()

	metrics := []*dto.Metric{}
	for m := range out {
		collected := &dto.Metric{}
		m.Write(collected)

		metrics = append(metrics, collected)
	}

	return metrics
}
```

pkg/util/metrics_helper.go
pstibrany: To fit the rest of the code, here is my suggestion:

```suggestion
func (d MetricFamiliesPerUser) SendSumOfGaugesPerUserWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
	for user, userMetrics := range d {
		result := singleValueWithLabelsMap{}
		userMetrics.sumOfSingleValuesWithLabels(metric, labelNames, gaugeValue, result.aggregateFn)

		for key, mlv := range result {
			mlv.LabelValues = append([]string{user}, mlv.LabelValues...)
			result[key] = mlv
		}

		result.WriteToMetricChannel(out, desc, prometheus.GaugeValue)
	}
}
```

where the inner `for` for adding user label could even be moved to `singleValueWithLabelsMap` type as a new method.

pkg/util/metrics_helper_test.go
```diff
@@ -81,3 +82,110 @@ func makeLabels(namesAndValues ...string) []*dto.LabelPair {
...
+		"",
+		[]string{"user", labelOneName}, nil)
+
+	baseFamily := MetricFamilyMap{
+		"test_metric": &dto.MetricFamily{
+			Name: &metricName,
+			Type: &metricType,
+			Metric: []*dto.Metric{
+				{
+					Label: []*dto.LabelPair{
```
pstibrany: you can use `makeLabels` function here to save some extra lines.

pkg/util/metrics_helper_test.go
```diff
@@ -81,3 +82,110 @@ func makeLabels(namesAndValues ...string) []*dto.LabelPair {
...
+	for _, l := range userTwoBuf.GetLabel() {
+		switch l.GetName() {
+		case "user":
+			require.Equal(t, "user2", l.GetValue())
+		case labelOneName:
+			require.Equal(t, labelValueOne, l.GetValue())
+		default:
+			require.FailNow(t, "unexpected label="+l.GetName())
+		}
+	}
```
pstibrany: This is somewhat hard to read, and order of series is not really guaranteed, so the test can fail.

Test for `alertmanagerMetrics` may be enough, if it exercises this functionality properly.

Alternatively, it may be more readable by collecting the result and comparing against full expected value, but I'm not too sure about that.

pkg/alertmanager/multitenant.go
```diff
@@ -431,6 +439,8 @@ func (am *MultitenantAlertmanager) deleteUser(userID string) {
 }
 
 func (am *MultitenantAlertmanager) newAlertmanager(userID string, amConfig *amconfig.Config) (*Alertmanager, error) {
+	reg := prometheus.NewRegistry()
+	am.metrics.addUserRegistry(userID, reg)
```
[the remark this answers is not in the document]
pstibrany: This is tricky actually... by removing the registry, you can break some metrics. It's ok if per-user counter disappears, and then reappears from 0. But it's not OK for global counters (`alertmanager_alerts_received_total`, `alertmanager_alerts_invalid_total`), summaries (`alertmanager_nflog_gc_duration_seconds`) or histograms (`alertmanager_notification_latency_seconds`) to go down.

For global counters like that, you need to preserve related values from the registry to some local counters in `alertmanagerMetrics` and then add those stored values when sending metrics in `Collect` method. metrics_helper.go has `SummaryData` and `HistogramData` that can help with that.

pkg/util/metrics_helper_test.go
pracucci: @pstibrany What do you think about my proposal https://github.com/cortexproject/cortex/pull/2124#discussion_r378803769 ?

pkg/util/metrics_helper_test.go
pstibrany: 👍 Yes, this is much easier to follow. I was confused for a second about the labels order, but it turns out that Prometheus will sort them by name in `prometheus.MustNewConstMetric`, who knew? :) 

Personally I'd put in some extra `{ ... }` blocks, to separate variables between individual tests, but that's really just nitpicking.

pkg/alertmanager/multitenant.go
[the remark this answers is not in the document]
pstibrany: If user will not be re-added later, it might actually be easier just to leave it in the map.

However, if same user can reappear later, that will not work, as previous registry cannot be passed to new AlertManager – registering metrics would panic :(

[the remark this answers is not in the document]
jtlisi: In that case the best course of action would be to deactivate the users Alertmanager instead of delete it. It is totally conceivable for a user to delete then recreate an Alertmanager. It is much easier to reactivate an Alertmanager with a new config instead of going through the rigamarole of handling this in the registry.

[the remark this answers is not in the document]
jtlisi: I'm going to wait for #2125 to be merged before handling solving this issue.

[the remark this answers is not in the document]
pracucci: @jtlisi #2125 is merged done. I think we can now address this issue (and BTW, I do agree that de-activating/re-activating looks a reasonable solution *if* doesn't have side effects).

pkg/util/metrics_helper.go
jtlisi: I will use your suggested snippet and move that `for` loop to a method called `appendUserLabelValue` for the `singleValueWithLabelsMap` struct

pkg/util/metrics_helper_test.go
jtlisi: I will update to use this test.

pkg/alertmanager/multitenant.go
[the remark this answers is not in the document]
jtlisi: I have updated this PR to only pause a users alertmanager and restart it when a new config is applied.

pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	silences                        *prometheus.Desc
+	silencesPropagatedMessagesTotal *prometheus.Desc
+}
+
+func newAlertmanagerMetrics() *alertmanagerMetrics {
+	return &alertmanagerMetrics{
+		regs:   map[string]*prometheus.Registry{},
+		regsMu: sync.Mutex{},
+		alertsReceived: prometheus.NewDesc(
+			"cortex_alertmanager_alerts_received_total",
```
gouthamve: Shouldn't we track these per user too? Seems like a useful metric to track.

pkg/util/metrics_helper.go
```diff
@@ -31,6 +31,13 @@ func (m singleValueWithLabelsMap) aggregateFn(labelsKey string, labelValues []st
 	m[labelsKey] = r
 }
 
+func (m singleValueWithLabelsMap) appendUserLabelValue(user string) {
```
pracucci: Technically `prependUserLabelValue` would be a more accurate name, cause we add at the beginning.

pkg/alertmanager/multitenant.go
```diff
@@ -300,15 +307,14 @@ func (am *MultitenantAlertmanager) syncConfigs(cfgs map[string]alerts.AlertConfi
 	defer am.alertmanagersMtx.Unlock()
 	for user, userAM := range am.alertmanagers {
 		if _, exists := cfgs[user]; !exists {
-			level.Info(am.logger).Log("msg", "deleting per-tenant alertmanager", "user", user)
-			userAM.Stop()
-			delete(am.alertmanagers, user)
+			level.Info(am.logger).Log("msg", "deactivating per-tenant alertmanager", "user", user)
+			userAM.Pause()
```
pracucci: Could you add a comment to explain why we `Pause()` it instead of `Stop()` + delete, please?

pkg/alertmanager/multitenant_test.go
```diff
@@ -112,10 +113,36 @@ func TestLoadAllConfigs(t *testing.T) {
...
 	currentConfig, exists = am.cfgs["user3"]
 	require.False(t, exists)
-	require.Equal(t, alerts.AlertConfigDesc{}, currentConfig)
+	require.Equal(t, "", currentConfig.RawConfig)
+
+	userAM, exists := am.alertmanagers["user3"]
+	require.True(t, exists)
+	userAM.activeMtx.Lock()
+	require.False(t, userAM.active)
+	userAM.activeMtx.Unlock()
```
pracucci: I would suggest to add a `IsActive()` function to the alert manager, so that we can use it here and below.

pkg/alertmanager/multitenant.go
```diff
@@ -452,7 +460,11 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	am.alertmanagersMtx.Lock()
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
-	if !ok {
+
+	userAM.activeMtx.Lock()
```
pracucci: Have you considered moving this logic (starting from this line until the end of the function) to a `ServeHTTP()` function in the alertmanager? From a design perspective, I think it would be better if we keep the `activeMtx` usage isolated within the alertmanager.

pkg/alertmanager/alertmanager.go
```diff
@@ -236,9 +244,47 @@ func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
...
+
+	// Ensure inhibitor is set before being called
+	if am.inhibitor != nil {
+		am.inhibitor.Stop()
+	}
+
+	// Ensure dispatcher is set before being called
+	if am.dispatcher != nil {
+		am.dispatcher.Stop()
+	}
```
pracucci: Where are they restarted when we "unpause"? If this is confirmed to be a bug, I would suggest to:
1. Add an `Unpause()` function moving all the "unpause" logic there
2. Enrich the tests to cover this issue

pkg/alertmanager/multitenant.go
```diff
@@ -423,6 +429,8 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
 }
 
 func (am *MultitenantAlertmanager) newAlertmanager(userID string, amConfig *amconfig.Config) (*Alertmanager, error) {
+	reg := prometheus.NewRegistry()
+	am.metrics.addUserRegistry(userID, reg)
```
pstibrany: Nit: I'd suggest adding registry to `am.metrics` only if creation of alert manager doesn't fail with error.

pkg/alertmanager/alertmanager.go
pstibrany: My understanding is that it is the `ApplyConfig` method that will "unpause" alert manager. It will also start new inhibitor and dispatcher.

This is not really obvious from the code though. Some comment at least would be nice.

pkg/alertmanager/multitenant.go
pstibrany: I think it's fair to have check for active and return early here, but accessing Active flag definitely does belong to alertmanager, to keep locking in one place only. Also `mux` should not be accessed directly.

```go
    if !ok || !userAM.isActive() { // isActive should do the locking
        ...
    }
    userAM.ServeHTTP(w, req)
```

pkg/alertmanager/alertmanager.go
```diff
@@ -173,6 +176,11 @@ func clusterWait(p *cluster.Peer, timeout time.Duration) func() time.Duration {
 
 // ApplyConfig applies a new configuration to an Alertmanager.
 func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
+	// Ensure the alertmanager is set to active
+	am.activeMtx.Lock()
+	am.active = true
+	am.activeMtx.Unlock()
```
pracucci: Shouldn't this done at the end of the method, once we're sure we'll not return error and everything has been setup correctly?

pkg/alertmanager/alertmanager.go
pracucci: You're right Peter. Not obvious tho.

pkg/alertmanager/multitenant.go
jtlisi: 👍

pkg/alertmanager/alertmanager.go
jtlisi: I will add a comment explaining that.

pkg/util/metrics_helper.go
jtlisi: 👍

pkg/alertmanager/multitenant_test.go
jtlisi: 👍

pkg/alertmanager/multitenant.go
jtlisi: 👍

pkg/alertmanager/alertmanager.go
jtlisi: Yes, I think that would be better.

pkg/alertmanager/multitenant.go
jtlisi: 👍

pkg/alertmanager/alertmanager_metrics.go
jtlisi: I agree, these metrics are pretty trivial to make per-user so I'll update it

pkg/alertmanager/alertmanager.go
```diff
@@ -233,12 +236,62 @@ func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
...
 
+	// Ensure the alertmanager is set to active
+	am.activeMtx.Lock()
+	am.active = true
+	am.activeMtx.Unlock()
+
 	return nil
 }
 
+func (am *Alertmanager) isActive() bool {
```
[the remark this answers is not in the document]
jtlisi: Makes sense

PR #2127, opened 2020-02-13
Memberlist integration test fix

pstibrany:
**What this PR does**:
This PR does 2+1 things:
- It makes Memberlist KV store to wait a little bit on shutdown to increase the chance of sending out messages caused by local CAS updates.
- It modifies memberlist integration test to use 3s for push/pull mechanism. Default is 30s. Push/pull transfers full state between two random nodes. It makes cluster convergence faster, which speeds up the test and avoids flakiness.
- It also changes usage of older github.com/uber-go/atomic 1.4.0 to go.uber.org/atomic 1.5.0. Both refer to the same package, but "go.uber.org/atomic" is a new module name.


pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -294,7 +299,23 @@ func (m *KV) JoinMembers(members []string) (int, error) {
...
+	// generated on this node via CAS, and those are disabled now (via casBroadcastsEnabled), and should be able
+	// to get out in this timeout.
+
+	waitTimeout := time.Now().Add(10 * time.Second)
+	for m.broadcasts.NumQueued() > 0 && m.memberlist.NumMembers() > 1 && time.Now().Before(waitTimeout) {
+		time.Sleep(250 * time.Millisecond)
+	}
+
+	if cnt := m.broadcasts.NumQueued(); cnt > 0 {
+		level.Warn(util.Logger).Log("msg", "broadcast message left in queue", "count", cnt, "nodes", m.memberlist.NumMembers())
```
pracucci: `broadcast message` > `broadcast messages`

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -524,7 +545,12 @@ outer:
 		if change != nil {
 			m.casSuccesses.Inc()
 			m.notifyWatchers(key)
-			m.broadcastNewValue(key, change, newver, codec)
+
+			if m.casBroadcastsEnabled.Load() {
+				m.broadcastNewValue(key, change, newver, codec)
+			} else {
+				level.Error(util.Logger).Log("msg", "CAS with broadcasts disabled", "key", key)
```
pracucci: - Is this an error? Doesn't look actionable to me. Maybe `Warn` is better?
- What do you think if we change the log message into something like `skipped broadcasting CAS message because broadcasting is disabled` for clarity?

pstibrany: > * Is this an error? Doesn't look actionable to me. Maybe `Warn` is better?

It's an error in the code, in a sense that it should never happen. We stop writers before KV. But from user's point of view, it's not something they can do anything with, so I'll make it a warning.

> * What do you think if we change the log message into something like `skipped broadcasting CAS message because broadcasting is disabled` for clarity?

Sounds good

pstibrany: What about: `skipped broadcasting CAS update because memberlist KV is shutting down` ?

pracucci: > What about: `skipped broadcasting CAS update because memberlist KV is shutting down` ?

Even better. But then what if we rename `casBroadcastsEnabled` to `shuttingDown`? May be more clear to the code reader.

pstibrany: I like `casBroadcastsEnabled` more, because it's specific. `shuttingDown` may imply more than it really does.

Issue #2130, opened 2020-02-13
When using (experimental) memberlist KV, ingesters reusing same names may not be able to join memberlist cluster for some time


pstibrany:
Memberlist library uses two options to deal with nodes that don't reply back:

- GossipToTheDeadTime – this is interval (defaults to 30s) saying for how long should memberlist node try to gossip to node which doesn't reply (in a hope that node comes back). After this time expires and node is still dead, memberlist library removes the node from memory, and loses information about it.

- DeadNodeReclaimTime – controls the time before a dead node's name can be reclaimed by one with a different address or port. By default, this is 0, meaning nodes cannot be reclaimed this way. (That is, until node entry is removed after GossipToTheDeadTime expires).

For ingesters using stateful set, it's useful to be able to configure these values, and enable DeadNodeReclaimTime. Otherwise new ingester cannot join memberlist cluster for up to 30s, since it uses same node name, but has different IP, so other members will ignore it.

PR #2131, opened 2020-02-13
Memberlist: Options to control handling of dead nodes and name reuse

pstibrany:
**What this PR does**: This PR exposes `-memberlist.gossip-to-dead-nodes-time` and `-memberlist.dead-node-reclaim-time` options to control how memberlist library handles dead nodes and name reuse.

Fixes https://github.com/cortexproject/cortex/issues/2130

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/configuration/arguments.md
+++ b/docs/configuration/arguments.md
@@ -186,7 +186,8 @@ Flags for configuring KV store based on memberlist library. This feature is expe
    How long to keep gossiping to the nodes that seem to be dead. After this time, dead node is removed from list of nodes.
 - `memberlist.dead-node-reclaim-time`
    How soon can node name be reused by a new node (using different IP). Disabled by default, name reclaim is not allowed until `gossip-to-dead-nodes-time` expires. This can be useful to set to low numbers when reusing node names, eg. in stateful sets.
-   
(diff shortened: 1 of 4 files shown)


docs/configuration/arguments.md
```diff
@@ -182,7 +182,12 @@ Flags for configuring KV store based on memberlist library. This feature is expe
    Timeout for writing 'packet' data.
 - `memberlist.transport-debug`
    Log debug transport messages. Note: global log.level must be at debug level as well.
-   
+- `memberlist.gossip-to-dead-nodes-time`
+   How long to keep gossiping to the nodes that seem to be dead. After this time, dead node is removed from list of nodes.
```
pracucci: Could you also explain what happens if a dead node reappers after `gossip-to-dead-nodes-time` (ie. in the case of a network partition).

docs/configuration/arguments.md
```diff
@@ -182,7 +182,12 @@ Flags for configuring KV store based on memberlist library. This feature is expe
...
+- `memberlist.dead-node-reclaim-time`
+   How soon can node name be reused by a new node (using different IP). Disabled by default, name reclaim is not allowed until `gossip-to-dead-nodes-time` expires. This can be useful to set to low numbers when reusing node names, eg. in stateful sets.
```
pracucci: Could you also explain that this may occurs during ingesters rollout when using stateful sets (thus name is reused) and in such case it's suggested to enable this config option to a value of X?

pracucci:
@pstibrany Please also remember to rebase `master` to fix the failing `integration-configs-db` tests.

docs/configuration/arguments.md
pstibrany: Isn't that what 

> This can be useful to set to low numbers when reusing node names, eg. in stateful sets.

already says?

PR #2135, opened 2020-02-14
Break background cache writes into batches of 100

### LINKED ISSUE #2134: Background cache batching is too coarse
PR #1011 changed `backgroundCache` from storing one key at a time with a queue of max size 10,000 to storing a batch at a time, but a single batch can have any number of keys in it.  For instance a query which loaded 100,000 chunks will create one batch of 100,000 keys.

This has the following downsides:
 * 100,000 keys will be processed serially while the other `backgroundCache` goroutines sit idle.
 * The backlog is much larger than its configured limit.
 * This bump in unprocessed keys is likely not visible in `cortex_cache_background_queue_length` because the `Sub()` comes before doing the writes, so the queue length is only at its maxiumum for a very brief time.
 * If Jaeger tracing is in use, a span with 100,000 children exceeds Jaeger's limit to display.

I suggest breaking down the write into 1000 batches of 100 so that they can be parallelised, which would also improve the observability.

bboreham:
**What this PR does**:

Improves parallelism and observability of the background cache writes.

Fixes #2134

**Checklist**
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/chunk/cache/background.go
+++ b/pkg/chunk/cache/background.go
@@ -33,7 +33,7 @@ type BackgroundConfig struct {
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
 func (cfg *BackgroundConfig) RegisterFlagsWithPrefix(prefix string, description string, f *flag.FlagSet) {
 	f.IntVar(&cfg.WriteBackGoroutines, prefix+"memcache.write-back-goroutines", 10, description+"How many goroutines to use to write back to memcache.")
-	f.IntVar(&cfg.WriteBackBuffer, prefix+"memcache.write-back-buffer", 10000, description+"How many chunks to buffer for background write back.")
+	f.IntVar(&cfg.WriteBackBuffer, prefix+"memcache.write-back-buffer", 10000, description+"How many key batches to buffer for background write-back.")
 }
 
 type backgroundCache struct {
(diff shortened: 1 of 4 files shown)


pkg/chunk/cache/background.go
```diff
@@ -80,21 +80,33 @@ func (c *backgroundCache) Stop() {
...
+			num = len(keys)
+		}
+
+		bgWrite := backgroundWrite{
+			keys: keys[:num],
+			bufs: bufs[:num],
+		}
+		select {
+		case c.bgWrites <- bgWrite:
+			c.queueLength.Add(float64(len(keys)))
```
jtlisi:
```suggestion
			c.queueLength.Add(float64(len(num)))
```

I think you should utilize `num` here. If I'm not mistaken `keys` would include all of the unwritten keys.

bboreham: Yeah it's a bug, thanks.

pkg/chunk/cache/background.go
```diff
@@ -80,21 +80,33 @@ func (c *backgroundCache) Stop() {
...
+		select {
+		case c.bgWrites <- bgWrite:
+			c.queueLength.Add(float64(num))
+		default:
+			c.droppedWriteBack.Add(float64(len(keys)))
+			sp := opentracing.SpanFromContext(ctx)
+			if sp != nil {
+				sp.LogFields(otlog.Int("dropped", len(keys)))
+			}
+			break // queue is full; give up
```
pracucci: Should we give up this batch only or all subsequent ones as well? It's likely the subsequent ones will hit the full queue as well, so I'm wondering if we should break the `for` / `return` here.

pkg/chunk/cache/background.go
```diff
@@ -80,21 +80,33 @@ func (c *backgroundCache) Stop() {
 	c.Cache.Stop()
 }
 
+const keysPerBatch = 100
+
 // Store writes keys for the cache in the background.
 func (c *backgroundCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
```
pracucci: Do we have a test convering the case the input keys are > `keysPerBatch`?

pkg/chunk/cache/background.go
bboreham: Yes, sorry, that's a bug.  I meant to break the loop.

pkg/chunk/cache/background.go
```diff
@@ -33,7 +33,7 @@ type BackgroundConfig struct {
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
 func (cfg *BackgroundConfig) RegisterFlagsWithPrefix(prefix string, description string, f *flag.FlagSet) {
 	f.IntVar(&cfg.WriteBackGoroutines, prefix+"memcache.write-back-goroutines", 10, description+"How many goroutines to use to write back to memcache.")
-	f.IntVar(&cfg.WriteBackBuffer, prefix+"memcache.write-back-buffer", 10000, description+"How many chunks to buffer for background write back.")
+	f.IntVar(&cfg.WriteBackBuffer, prefix+"memcache.write-back-buffer", 10000, description+"How many key batches to buffer for background write-back.")
```
gouthamve: Currently this means we can buffer 10000 * 100 = 1Mil chunks which is not something we want? Can we reduce it to `1000/100` = 10 batches at once? But this further means that if there are 10 writes each with a single key, then we'll only buffer 10 keys.

Not sure how to handle this simply.

bboreham: This is not a new problem; previously we could buffer 10000*(any number) chunks.
Suggest we deal with your point separately.

pkg/chunk/cache/background.go
bboreham: I modified the test to ensure this.

PR #2139, opened 2020-02-16
Fix website documentation edit links

cyriltovena:

The `edit this page` link as shown below

image

was broken since I've started the website.

 This fixes it and also introduces a way to override the page original path `originalpath` using front matter variable. This is useful for files that are not using the same destination path.

### DIFF
--- a/tools/website/web-pre.sh
+++ b/tools/website/web-pre.sh
@@ -48,7 +48,7 @@ echo "$(cat <<EOT
 ---
 title: Contributing
 type: docs
-original_path: CONTRIBUTING.md
+originalpath: CONTRIBUTING.md
 menu:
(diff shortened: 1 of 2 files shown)


tools/website/web-pre.sh
```diff
@@ -46,6 +48,7 @@ echo "$(cat <<EOT
 ---
 title: Contributing
 type: docs
+original_path: CONTRIBUTING.md
```
pracucci: `original_path` > `originalpath` to quick fix it, or rename all other `originalpath` references to `original_path` to make it clearer.

cyriltovena: yeah my bad will do

PR #2145, opened 2020-02-17
Fixed execution of integration tests framework's unit tests in CI

pracucci:
**What this PR does**:
In this PR:
- Run integration tests framework's unit tests along with integration tests (cause some of them require Docker)
- Remove `grep -v` from unit tests discovery (may be risky)
- Fixed `WaitSumMetric()` flaky tests

Here is the proof that the list of unit tests hasn't changed:

```
$ git ls-files -- '*_test.go' | grep -vE '^(vendor|experimental|integration)/' > before.txt
$ git ls-files -- 'cmd/*_test.go' 'pkg/*_test.go' 'tools/*_test.go' > after.txt
$ diff before.txt after.txt
No diff
```

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,7 @@ jobs:
           export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:${CIRCLE_TAG:-$(./tools/image-tag)}"
           export CORTEX_CHECKOUT_DIR="/home/circleci/.go_workspace/src/github.com/cortexproject/cortex"
           echo "Running integration tests with image: $CORTEX_IMAGE"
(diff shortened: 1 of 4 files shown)


pracucci:
@bwplotka I should have fixed flaky tests for `WaitSumMetric()`. May you take a look please?

pracucci:
> What was flaky, exactly?

If the call to `WaitSumMetric()` does the HTTP request before the goroutine calling `ListenAndServe()` did listen on the port, the `WaitSumMetric()` was immediately returning with error. In this PR I do listen first and then just call `Serve()` in the goroutine.

bwplotka:
yea, make sense. THanks!

PR #2146, opened 2020-02-17
e2e: Reduced number times the networkName variable is needed; Also allowed injecting different comparisons to WaitSumMetrics.

bwplotka:
Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

### DIFF
--- a/integration/e2e/service.go
+++ b/integration/e2e/service.go
@@ -26,6 +26,9 @@ var (
 
 // ConcreteService represents microservice with optional ports which will be discoverable from docker
 // with <name>:<port>. For connecting from test, use `Endpoint` method.
+//
+// ConcreteService can be reused (started and stopped many time), but it can represent only one running container
+// at the time.
 type ConcreteService struct {
 	name         string
 	image        string
@@ -41,7 +44,9 @@ type ConcreteService struct {
 	// Generic retry backoff.
 	retryBackoff *util.Backoff
 
-	stopped bool
+	// docker NetworkName used to start this container.
+	// If empty it means service is stopped.
+	usedNetworkName string
 }
 
 func NewConcreteService(
@@ -63,8 +68,11 @@ func NewConcreteService(
 			MaxBackoff: 600 * time.Millisecond,
 			MaxRetries: 50, // Sometimes the CI is slow ¯\_(ツ)_/¯
 		}),
-		stopped: true,
 	}
+}
+
+func (s *ConcreteService) isExpectedRunning() bool {
+	return s.usedNetworkName != ""
 }
 
 func (s *ConcreteService) Name() string { return s.name }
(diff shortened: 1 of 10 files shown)


integration/e2e/service.go
```diff
@@ -183,13 +192,17 @@ func (s *ConcreteService) Endpoint(port int) string {
 	return fmt.Sprintf("127.0.0.1:%d", localPort)
 }
 
-func (s *ConcreteService) NetworkEndpoint(networkName string, port int) string {
-	return fmt.Sprintf("%s:%d", s.containerName(networkName), port)
+func (s *ConcreteService) NetworkEndpoint(port int) string {
+	return fmt.Sprintf("%s:%d", s.containerName(), port)
```
pracucci: I think this could be:
`return s.NetworkEndpointFor(s.usedNetworkName, port)`

integration/e2e/service.go
```diff
@@ -413,28 +489,22 @@ func (s *HTTPService) WaitSumMetric(metric string, value float64) error {
...
+		if isExpected(sums...) {
 			return nil
 		}
-		lastValue = sum
 
 		s.retryBackoff.Wait()
 	}
 
-	return fmt.Errorf("unable to find a metric %s with value %v. LastValue: %v", metric, value, lastValue)
+	return fmt.Errorf("unable to find a metrics %s with expected values. LastValues: %v", metricNames, sums)
```
pracucci: `a metrics` > `metrics`

integration/e2e/service.go
```diff
@@ -394,12 +411,71 @@ func (s *HTTPService) HTTPEndpoint() string {
...
+
+func sumValues(family *io_prometheus_client.MetricFamily) float64 {
+	sum := 0.0
+	for _, m := range family.Metric {
+		if m.GetGauge() != nil {
+			sum += m.GetGauge().GetValue()
+		} else if m.GetCounter() != nil {
+			sum += m.GetCounter().GetValue()
+		} else if m.GetHistogram() != nil {
+			sum += m.GetHistogram().GetSampleSum()
```
pracucci: Before we were using `GetSampleCount()`. This is the reason why integration tests are breaking. The `GetSampleSum()` is more correct and I'm OK keeping it.

To fix the tests, an option is to replace the `e2e.Equals()` on the metric `cortex_dynamo_sync_tables_seconds` with a `e2e.Greater(0)` (adding `Greater()` should be straightforward).

bwplotka: :+1:

integration/e2e/service.go
bwplotka: :+1:

bwplotka:
Thanks @pracucci 

I see you already started to add more cases nice!

Our Thanos e2e is ready for review as well: https://github.com/thanos-io/thanos/pull/2138 

It looks 1000x cleaner and it's a lot faster. :heart: 

PLUS it tests our metrics *heavily*

bwplotka:
Thanks for fixing this. (:

bwplotka:
Ping (:

PR #2147, opened 2020-02-18
Add ability to override configuration settings using environment variables

Wing924:
**What this PR does**:

You can use environment variable references in the config file to set values that need to be configurable during deployment.
To do this, use:

```
${VAR}
```

Where VAR is the name of the environment variable.

Each variable reference is replaced at startup by the value of the environment variable.
The replacement is case-sensitive and occurs before the YAML file is parsed.
References to undefined variables are replaced by empty strings unless you specify a default value or custom error text.

To specify a default value, use:

```
${VAR:default_value}
```

Where default_value is the value to use if the environment variable is undefined.

**Checklist**
- [x] Tests updated
- [x] Documentation added


Wing924:
@pracucci 
Now I use kustomize to manage k8s manifest. As I asked on the slack, it's better to put all configs on one config file (as a configmap) compared to set them by flags. The problem is that we can't change/override values on configmap per overlays.
Some values like db-url are different from each overlay. Some values like password are secrets, which don't want to put on plaintext.

As you say, put them on flags with environment variables is an option, but I don't like this because many components need to set the same flags:
```yml
# cassandra settings
- -cassandra.addresses=$(CASSANDRA_ADDRESSES)
- -cassandra.username=$(CASSANDRA_USERNAME)
- -cassandra.password=$(CASSANDRA_PASSWORD)
```
it looks very redundant to me.

tomwilkie:
I'm not averse to this myself; having multiple ways of overriding config settings (flags and environment variables) is a oft requested feature in the Prometheus world, and the "config management is someone else's problem" approach has gone down particularly badly.  Plus this PR includes docs and tests, so I can't find anything to nit pick!

That being said I can see this being a little divisive, so I want to get a feeling from the other maintainers.  But I'm +1 on this.

cmd/cortex/main.go
```diff
@@ -155,3 +158,19 @@ func DumpYaml(cfg *cortex.Config) {
 		fmt.Printf("%s\n", out)
 	}
 }
+
+// expandEnv replaces ${var} or $var in config according to the values of the current environment variables.
+// The replacement is case-sensitive. References to undefined variables are replaced by the empty string.
+// A default value can be given by using the form ${var:default value}.
+func expandEnv(config []byte) []byte {
```
pracucci: The config can include passwords as well. What happens if a password, specified in the config, contains the `$` symbol? I think it will end up in a breaking change because after the `expandEnv()` the `$` will be removed even if the environment variable name is not set.

Maybe could be safer enabling the env var expansion via a CLI flag (disabled by default)?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
   * `--experimental.distributor.user-subring-size`
 * [FEATURE] Added flag `-experimental.ruler.enable-api` to enable the ruler api which implements the Prometheus API `/api/v1/rules` and `/api/v1/alerts` endpoints under the configured `-http.prefix`. #1999
 * [FEATURE] Added sharding support to compactor when using the experimental TSDB blocks storage. #2113
+* [FEATURE] Add ability to override configuration settings using environment variables. #2147
```
pracucci: I would replace `configuration settings` with `YAML config file settings` to be more clear.

cmd/cortex/main.go
Wing924: I added `config.expand-env` (defaults to false) options.

PR #2156, opened 2020-02-18
Frontend: do not return 500 if the user request cancellation.

cyriltovena:

As discussed with @pracucci , This change how the frontend handle cancellation/timeout. 

If the user request/upstream cancellation, then we don't throw a 5xx but a 499 instead (I'm happy to choose another code).

However in case of deadline exceeded/timeout, or unknown error we're sending back a 500.

The reason for this change is because if a user quickly cancel a request (page refresh) before it has time to answer the frontend will register this as an error (500) which I don't think we should.

### DIFF
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -38,8 +38,9 @@ var (
 		Help:      "Number of queries in the queue.",
 	})
 
-	errTooManyRequest     = httpgrpc.Errorf(http.StatusTooManyRequests, "too many outstanding requests")
-	statusRequestCanceled = 499
+	errTooManyRequest   = httpgrpc.Errorf(http.StatusTooManyRequests, "too many outstanding requests")
+	errCanceled         = httpgrpc.Errorf(499, context.Canceled.Error())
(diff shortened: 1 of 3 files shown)


pkg/querier/frontend/frontend.go
```diff
@@ -38,8 +38,9 @@ var (
 		Help:      "Number of queries in the queue.",
 	})
 
-	errTooManyRequest = httpgrpc.Errorf(http.StatusTooManyRequests, "too many outstanding requests")
-	errCanceled       = httpgrpc.Errorf(http.StatusInternalServerError, "context cancelled")
+	errTooManyRequest   = httpgrpc.Errorf(http.StatusTooManyRequests, "too many outstanding requests")
+	errCanceled         = httpgrpc.Errorf(499, context.Canceled.Error())
```
tomwilkie: 499 seems appropriate here (https://httpstatuses.com/499)

Nit: I would create a constant for it, `StatusClientClosedRequest`, and use said constant in the test.

cyriltovena: Will do!

PR #2157, opened 2020-02-19
chunk slice support

sandeepsukhani:
**What this PR does**:
Adds support for slicing chunks for the given time range.
`Chunk.Slice` method iterates through the data to build a new chunk out of it with the same encoding.
This PR lays foundation for #2103 and helps reduce its size.

**Checklist**
- [x] Tests added

### DIFF
--- a/pkg/chunk/chunk.go
+++ b/pkg/chunk/chunk.go
@@ -22,10 +22,12 @@ import (
 
 // Errors that decode can return
 const (
-	ErrInvalidChecksum = errs.Error("invalid chunk checksum")
-	ErrWrongMetadata   = errs.Error("wrong chunk metadata")
-	ErrMetadataLength  = errs.Error("chunk metadata wrong length")
-	ErrDataLength      = errs.Error("chunk data wrong length")
+	ErrInvalidChecksum    = errs.Error("invalid chunk checksum")
+	ErrWrongMetadata      = errs.Error("wrong chunk metadata")
+	ErrMetadataLength     = errs.Error("chunk metadata wrong length")
+	ErrDataLength         = errs.Error("chunk data wrong length")
+	ErrSliceOutOfRange    = errs.Error("chunk can't be sliced out of its data range")
+	ErrSliceNoDataInRange = errs.Error("chunk has no data for given range to slice")
 )
 
 var castagnoliTable = crc32.MakeTable(crc32.Castagnoli)
@@ -326,4 +328,48 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
 	it := c.Data.NewIterator(nil)
 	interval := metric.Interval{OldestInclusive: from, NewestInclusive: through}
 	return prom_chunk.RangeValues(it, interval)
+}
+
+// Slice builds a new smaller chunk with data only from given time range
+func (c *Chunk) Slice(from, through model.Time) (*Chunk, error) {
+	if from > c.Through || through < c.From {
+		return nil, ErrSliceOutOfRange
+	}
+
+	itr := c.Data.NewIterator(nil)
+	if !itr.FindAtOrAfter(from) {
+		return nil, ErrSliceNoDataInRange
+	}
+
+	pc, err := prom_chunk.NewForEncoding(c.Data.Encoding())
+	if err != nil {
+		return nil, err
+	}
+
+	for !itr.Value().Timestamp.After(through) {
+		oc, err := pc.Add(itr.Value())
+		if err != nil {
+			return nil, err
+		}
+
+		if oc != nil {
+			// ToDo: revisit this
+			return nil, errors.New("slicing should not overflow a chunk")
+		}
+		if !itr.Scan() {
+			break
+		}
+	}
+
+	err = itr.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	if pc.Len() == 0 {
+		return nil, ErrSliceNoDataInRange
+	}
+
+	nc := NewChunk(c.UserID, c.Fingerprint, c.Metric, pc, from, through)
+	return &nc, nil
 }
--- a/pkg/chunk/chunk.go
+++ b/pkg/chunk/chunk.go
@@ -28,6 +28,7 @@ const (
 	ErrDataLength         = errs.Error("chunk data wrong length")
 	ErrSliceOutOfRange    = errs.Error("chunk can't be sliced out of its data range")
 	ErrSliceNoDataInRange = errs.Error("chunk has no data for given range to slice")
+	ErrSliceChunkOverflow = errs.Error("slicing should not overflow a chunk")
 )
 
 var castagnoliTable = crc32.MakeTable(crc32.Castagnoli)
@@ -332,6 +333,7 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
 
 // Slice builds a new smaller chunk with data only from given time range (inclusive)
 func (c *Chunk) Slice(from, through model.Time) (*Chunk, error) {
+	// there should be atleast some overlap between chunk interval and slice interval
 	if from > c.Through || through < c.From {
 		return nil, ErrSliceOutOfRange
 	}
@@ -353,8 +355,7 @@ func (c *Chunk) Slice(from, through model.Time) (*Chunk, error) {
 		}
 
 		if oc != nil {
-			// ToDo: revisit this
-			return nil, errors.New("slicing should not overflow a chunk")
+			return nil, ErrSliceChunkOverflow
 		}
 		if !itr.Scan() {
 			break
--- a/pkg/chunk/chunk.go
+++ b/pkg/chunk/chunk.go
@@ -330,7 +330,7 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
 	return prom_chunk.RangeValues(it, interval)
 }
 
-// Slice builds a new smaller chunk with data only from given time range
+// Slice builds a new smaller chunk with data only from given time range (inclusive)
 func (c *Chunk) Slice(from, through model.Time) (*Chunk, error) {
 	if from > c.Through || through < c.From {
 		return nil, ErrSliceOutOfRange
--- a/pkg/chunk/chunk_test.go
+++ b/pkg/chunk/chunk_test.go
@@ -135,13 +135,30 @@ func TestChunkCodec(t *testing.T) {
 const fixedTimestamp = model.Time(1557654321000)
 
 func TestChunkDecodeBackwardsCompatibility(t *testing.T) {
+	// lets build a new chunk same as what was built using code at commit b1777a50ab19
+	c, _ := encoding.NewForEncoding(encoding.Bigchunk)
+	nc, err := c.Add(model.SamplePair{Timestamp: fixedTimestamp, Value: 0})
+	require.NoError(t, err)
+	require.Equal(t, nil, nc, "returned chunk should be nil")
+
+	chunk := NewChunk(
+		userID,
+		client.Fingerprint(labelsForDummyChunks),
+		labelsForDummyChunks,
+		c,
+		fixedTimestamp.Add(-time.Hour),
+		fixedTimestamp,
+	)
+	// Force checksum calculation.
+	require.NoError(t, chunk.Encode())
+
 	// Chunk encoded using code at commit b1777a50ab19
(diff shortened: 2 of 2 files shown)


pkg/chunk/chunk_test.go
```diff
@@ -133,9 +136,9 @@ const fixedTimestamp = model.Time(1557654321000)
 
 func TestChunkDecodeBackwardsCompatibility(t *testing.T) {
 	// Chunk encoded using code at commit b1777a50ab19
-	rawData := []byte("\x00\x00\x00\xb7\xff\x06\x00\x00sNaPpY\x01\xa5\x00\x00\x04\xc7a\xba{\"fingerprint\":18245339272195143978,\"userID\":\"userID\",\"from\":1557650721,\"through\":1557654321,\"metric\":{\"bar\":\"baz\",\"toms\":\"code\",\"__name__\":\"foo\"},\"encoding\":3}\n\x00\x00\x00\x15\x01\x00\x11\x00\x00\x01\xd0\xdd\xf5\xb6\xd5Z\x00\x00\x00\x00\x00\x00\x00\x00\x00")
+	rawData := []byte("\x00\x00\x00\xb7\xff\x06\x00\x00sNaPpY\x01\xa5\x00\x00\x04\xc7a\xba{\"fingerprint\":18245339272195143978,\"userID\":\"userID\",\"from\":1557650721,\"through\":1557654321,\"metric\":{\"bar\":\"baz\",\"toms\":\"code\",\"__name__\":\"foo\"},\"encoding\":3}\n\x00\x00\x00\x15\x01\x00\x11\x00\x00\x01У\xbe\xb3\xd5Z\x00\x00\x00\x00\x00\x00\x00\x00\x00")
 	decodeContext := NewDecodeContext()
-	have, err := ParseExternalKey(userID, "userID/fd3477666dacf92a:16aab37c8e8:16aab6eb768:38eb373c")
+	have, err := ParseExternalKey(userID, "userID/fd3477666dacf92a:16aab37c8e8:16aab6eb768:ff216d5b")
```
pracucci: Could you explain why this has changed? Given this is a backward compatibility test, we shouldn't touch it.

pkg/chunk/chunk.go
```diff
@@ -327,3 +329,47 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
 	interval := metric.Interval{OldestInclusive: from, NewestInclusive: through}
 	return prom_chunk.RangeValues(it, interval)
 }
+
+// Slice builds a new smaller chunk with data only from given time range
+func (c *Chunk) Slice(from, through model.Time) (*Chunk, error) {
```
pracucci: If I'm not wrong, the `Chunk.From` and `Chunk.Through` are both inclusive. If confirmed, it may make sense having the input `from` and `through` inclusive as well for consistency (and whatever decision we pick about it we should document it in the function).

pkg/chunk/chunk.go
```diff
@@ -327,3 +329,47 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
...
+	if from > c.Through || through < c.From {
+		return nil, ErrSliceOutOfRange
+	}
+
+	itr := c.Data.NewIterator(nil)
```
pracucci: Sorry for the very dumb question. Can't we use `c.Data.Slice()`?

pkg/chunk/chunk_test.go
sandeepsukhani: > func dummyChunkForEncoding(now model.Time, metric labels.Labels, enc encoding.Encoding, samples int)
This is the function which is used for generating chunks for testing. It creates a chunk from `now-1 hour` to `now` but builds the samples starting from now, which means it would have samples with higher timestamp than chunks end time.

While most of the tests just add 1 sample, it works because it adds a sample for the last second of the chunks time range. I have fixed that behaviour which means it now adds samples for chunks from the start time of chunk, which also means it would create a different chunk.

I have built that rawData and signature from the same commit that is mentioned in the comments. Another option is to add a new function which would be just used in legacy tests building the same chunk and use the modified function everywhere else.

pkg/chunk/chunk.go
sandeepsukhani: Yes, the time ranges are inclusive in both cases. I will add a comment.
Thanks!

pkg/chunk/chunk.go
sandeepsukhani: It is certainly not a dumb question because I had discussed the same with Goutham :smile: 
First of all, it is not implemented for all encodings and secondly, for `BigChunk` It may overestimate. Also, Goutham mentioned that it is mainly used as an optimisation in the read path. See: https://github.com/cortexproject/cortex/blob/master/pkg/chunk/encoding/chunk.go#L53-L55

pkg/chunk/chunk_test.go
jtlisi: I can second this. I ran into this issue previously when working on unit tests for a migration tool that was never merged.

sandeepsukhani: I have added code to build chunk to the same test function instead of sharing code with other tests to avoid such future encounters.

pkg/chunk/chunk_test.go
```diff
@@ -35,22 +35,25 @@ func dummyChunk(now model.Time) Chunk {
 
 func dummyChunkForEncoding(now model.Time, metric labels.Labels, enc encoding.Encoding, samples int) Chunk {
 	c, _ := encoding.NewForEncoding(enc)
+	chunkStart := now.Add(-time.Hour)
```
pracucci: We pick now in input and then we manipulate it here. It's a bit hidden this manipulation. I would suggest to:
1. Rename the input `now` in `chunkStart`
2. Remove this line `chunkStart := now.Add(-time.Hour)`
3. In the places where you need a different start time, you can just call `dummyChunkForEncoding()` with the correct start time (ie. `now.Add(-time.Hour)`)

This way it will be more explicit.

pkg/chunk/chunk_test.go
```diff
@@ -282,3 +302,76 @@ func benchmarkDecode(b *testing.B, batchSize int) {
 		}
 	}
 }
+
+func TestChunk_Slice(t *testing.T) {
+	now := model.Now()
+
+	// dummy chunk is created with time range now-1hour to now
+	chunkStartTime := now.Add(-time.Hour)
```
pracucci: I personally find these tests a bit hard to read due to the usage of `now`. Have you considered the following:
1. Here you calculate `chunkStartTime` and `chunkEndTime` (based on the number of samples of the chunk and their interval)
2. In the test cases below you never use `now` but always use `chunkStartTime` and `chunkEndTime` to define the time ranges

I believe this way tests would be more clear and less error prone. What's your take?

pkg/chunk/chunk_test.go
```diff
@@ -282,3 +302,76 @@ func benchmarkDecode(b *testing.B, batchSize int) {
...
+
+	for _, tc := range []struct {
+		name       string
+		sliceRange model.Interval
+		err        string
```
pracucci: This can be `err error`

pkg/chunk/chunk.go
```diff
@@ -327,3 +329,47 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
...
+
+	for !itr.Value().Timestamp.After(through) {
+		oc, err := pc.Add(itr.Value())
+		if err != nil {
+			return nil, err
+		}
+
+		if oc != nil {
+			// ToDo: revisit this
+			return nil, errors.New("slicing should not overflow a chunk")
```
pracucci: Please define a `const` for this error too (ie. similar to `ErrSliceNoDataInRange`).

pkg/chunk/chunk.go
```diff
@@ -327,3 +329,47 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
...
+	}
+
+	for !itr.Value().Timestamp.After(through) {
+		oc, err := pc.Add(itr.Value())
+		if err != nil {
+			return nil, err
+		}
+
+		if oc != nil {
+			// ToDo: revisit this
```
pracucci: What should we revisit here? Is it worth to keep this `TODO`?

pkg/chunk/chunk_test.go
```diff
@@ -132,13 +135,30 @@ func TestChunkCodec(t *testing.T) {
...
+	chunk := NewChunk(
+		userID,
+		client.Fingerprint(labelsForDummyChunks),
+		labelsForDummyChunks,
+		c,
+		fixedTimestamp.Add(-time.Hour),
+		fixedTimestamp,
+	)
+	// Force checksum calculation.
+	require.NoError(t, chunk.Encode())
```
pracucci: Thanks for changing this based on my previous comment. I was wondering why of the previous comment but I'm not against changing the fixture if you build the fixtures based on the commit `b1777a50ab19`. I think the initial version of the change was clearer instead of copy-pasting code from `b1777a50ab19`.

pkg/chunk/chunk_test.go
```diff
@@ -282,3 +302,76 @@ func benchmarkDecode(b *testing.B, batchSize int) {
...
 	}
 }
+
+func TestChunk_Slice(t *testing.T) {
+	now := model.Now()
+
+	// dummy chunk is created with time range now-1hour to now
+	chunkStartTime := now.Add(-time.Hour)
+
+	for _, tc := range []struct {
```
pracucci: All these tests have the `Start` and `End` exactly matching the timestamp of a sample in the chunk, because the whole logic is based on `now` and the samples interval is 15s. I would suggest add 1 test where the start and end doesn't exactly match the timestamp of a sample to have an higher confidence.

pkg/chunk/chunk.go
```diff
@@ -327,3 +329,47 @@ func (c *Chunk) Samples(from, through model.Time) ([]model.SamplePair, error) {
 	interval := metric.Interval{OldestInclusive: from, NewestInclusive: through}
 	return prom_chunk.RangeValues(it, interval)
 }
+
+// Slice builds a new smaller chunk with data only from given time range (inclusive)
+func (c *Chunk) Slice(from, through model.Time) (*Chunk, error) {
+	if from > c.Through || through < c.From {
+		return nil, ErrSliceOutOfRange
+	}
```
pracucci: This check allows for out of ranges as well. For example, it would allow to pass a range where `through > c.Through`. It's not necessarily a problem (it just depends on the contract of the function). I wanna be sure you're aware of that and it's by design.

pkg/chunk/chunk_test.go
sandeepsukhani: You are right that the implementation is not clean. Changing this function would mean changing its usage in about 30 places where it is used. I wanted to avoid doing too many changes in delete series PR by avoiding doing irrelevant changes. But since this is smaller PR, I think I can change it.

pkg/chunk/chunk.go
sandeepsukhani: I think we can remove it. I had added it before being confident about my code.

pkg/chunk/chunk_test.go
pracucci: > Changing this function would mean changing its usage in about 30 places where it is used.

Why do you need to change it across all existing tests as well, instead of just your new test?

pkg/chunk/chunk_test.go
sandeepsukhani: I didn't copy any code from `b1777a50ab19`. This code just includes relevant parts from `dummyChunkForEncoding`. I think this change makes it easy for us to freely change that code in future and avoid breaking backwards compatibility accidentally. Someone could err in building a chunk from that old commit and tests might still work but we realize something is wrong only in future. Please let me know what do you think.

pkg/chunk/chunk.go
sandeepsukhani: The check is to not allow passing time range which has no overlap. With some overlap, a new chunk can still be created but for no overlap, there is no chance to build a new chunk.

pracucci: Sounds good. I just wanted to make sure it was by design.

pkg/chunk/chunk_test.go
pracucci: _Sorry, now I got it. It's unfortunate we've to carry on the legacy behaviour but I do understand it._

pkg/chunk/chunk_test.go
pracucci: You're right. LGTM.

Issue #2149, opened 2020-02-18
Cannot run integration tests on macOS X with Docker Desktop version 2.2.0.3


pstibrany:
After removing `CORTEX_INTEGRATION_DIR` env variable and upgrading Docker Desktop to version 2.2.0.3, I can no longer run integration tests on macOS X.

I was advised to add `/var/folders` to the "File Sharing" directories, but I'm unable to do so, since Docker always resolves it to `/private/var/folders` instead, and I cannot edit this afterwards.

<img width="1152" alt="Screenshot 2020-02-18 at 08 15 41" src="https://user-images.githubusercontent.com/895919/74712615-ff14a780-5226-11ea-87b5-0e31a3d2cbe5.png">

pracucci:
Fixed in #2143.

PR #2161, opened 2020-02-19
Use current directory for integration test shared dir.

tomwilkie:
Otherwise, you get the following error when using docker for mac:

```
alertmanager: docker: Error response from daemon: Mounts denied:
alertmanager: The path /var/folders/g5/3pb5_td927l61zm9xfmmtcp80000gn/T/e2e_integration_test388308663
alertmanager: is not shared from OS X and is not known to Docker.
alertmanager: You can configure shared paths from Docker -> Preferences... -> File Sharing.
alertmanager: See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
alertmanager: .
alertmanager: time="2020-02-19T10:32:42Z" level=error msg="error waiting for container: context canceled"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ docker-images/
 website/public
 website/resources
 website/content/en/docs
+e2e_integration_test*
--- a/integration/e2e/scenario.go
+++ b/integration/e2e/scenario.go
@@ -34,7 +34,11 @@ type Scenario struct {
 func NewScenario(networkName string) (*Scenario, error) {
 	s := &Scenario{networkName: networkName}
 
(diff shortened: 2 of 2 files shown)


pracucci:
Fixed #2149

pracucci:
@bwplotka Does this cause any troubles to Thanos?

bwplotka:
Lgtm to merge for now.

pracucci:
> It should be fine, we just prefer not to clutter repo. We can add some envvar support at some point

@bwplotka I definitely agree. We should revive the initial env variable we dropped at some point cause I thought it wasn't required anymore, but apparently is.

PR #2165, opened 2020-02-20
Add /config handler which exposes the current YAML config of cortex.

### LINKED ISSUE #1886: Expose HTTP endpoint to export live config in yaml format
Given the intention to deprecate CLI flags, I'm wondering if it's useful / required to give users a way to convert CLI flags to yaml. An option - for users already running Cortex clusters - may be introduce an HTTP endpoint to export the live config (which is the result of yaml + CLI flags + defaults) as yaml.

_I'm not super sure this is useful, so I would like to gather some feedback and real use cases first._

tomwilkie:

Fixes #1886

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 * [FEATURE] Added sharding support to compactor when using the experimental TSDB blocks storage. #2113
 * [FEATURE] Add ability to override YAML config file settings using environment variables. #2147
   * `-config.expand-env`
+* [FEATURE] Add /config HTTP endpoint which exposes the current Cortex configuration as YAML.
 * [ENHANCEMENT] Add `status` label to `cortex_alertmanager_configs` metric to gauge the number of valid and invalid configs. #2125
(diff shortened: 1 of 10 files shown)


cyriltovena:
> I think we need to mask passwords.

Good point, although I think the usage is meant to be internal.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 * [FEATURE] Added sharding support to compactor when using the experimental TSDB blocks storage. #2113
 * [FEATURE] Add ability to override YAML config file settings using environment variables. #2147
   * `-config.expand-env`
+* [FEATURE] Add /config HTTP endpoint which exposes the current Cortex configuration as YAML.
```
```suggestion
* [FEATURE] Add /config HTTP endpoint which exposes the current Cortex configuration as YAML. #2165
```

We now a script to check the changelog when we do a release, and it's based on the PR number in the changelog entry, reason why it's important to always add it.

tomwilkie: Good to know, thanks!

tomwilkie:
Thanks @pracucci!  Your commits LGTM.  Good catch re:omitempty, yes I think these should all be removed - and there are a lot of them.

PR #2167, opened 2020-02-20
Add config file support to integration tests

pracucci:
**What this PR does**:
While reviewing the PR #1878 it popped up that the PR may change the `query-frontend` config requirements (discussion still on-going). This made me realise that we don't have any integration test running with the config instead of CLI flags.

This PR is a first step to ease writing tests where Cortex gets started with a config file (CLI flags are still used for some overrides).

In particular:
- Added `query-frontend` support to local `tsdb-blocks-storage-s3` dev env
- Introduced `CortexService` to have cleaner integration tests 
- Added to `writeFileToSharedDir()` the ability to create the path of directories
- Enhanced `query-frontend` tests to run with config file too

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/development/tsdb-blocks-storage-s3/config/cortex.yaml
+++ b/development/tsdb-blocks-storage-s3/config/cortex.yaml
@@ -66,3 +66,6 @@ compactor:
       store: consul
       consul:
         host: consul:8500
+
+frontend_worker:
+  address: "query-frontend:9007"
--- a/development/tsdb-blocks-storage-s3/docker-compose.yml
+++ b/development/tsdb-blocks-storage-s3/docker-compose.yml
@@ -108,3 +108,17 @@ services:
       - 8006:8006
     volumes:
       - ./config:/cortex/config
+
+  query-frontend:
+    build:
+      context:    .
+      dockerfile: dev.dockerfile
+    image: cortex
+    command: ["sh", "-c", "sleep 3 && exec ./cortex -config.file=./config/cortex.yaml -target=query-frontend -server.http-listen-port=8007 -server.grpc-listen-port=9007"]
+    depends_on:
+      - consul
+      - minio
+    ports:
+      - 8007:8007
+    volumes:
+      - ./config:/cortex/config
--- a/integration/alertmanager_test.go
+++ b/integration/alertmanager_test.go
@@ -17,7 +17,7 @@ func TestAlertmanager(t *testing.T) {
 
 	require.NoError(t, writeFileToSharedDir(s, "alertmanager_configs/user-1.yaml", []byte(cortexAlertmanagerUserConfigYaml)))
 
-	alertmanager := e2ecortex.NewAlertmanager("alertmanager", AlertmanagerConfigs, "")
+	alertmanager := e2ecortex.NewAlertmanager("alertmanager", AlertmanagerFlags, "")
 	require.NoError(t, s.StartAndWaitReady(alertmanager))
 	require.NoError(t, alertmanager.WaitSumMetrics(e2e.Equals(1), "cortex_alertmanager_configs"))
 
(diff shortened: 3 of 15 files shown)


pracucci:
> I'm little worried how many combinations we will be testing in query_frontend_test.go. Perhaps it would be worth doing some refactoring there, but it can also lead to big test obfuscation.

Me too. What're your thoughts about the refactoring? What would refactor?

tomwilkie:
> I'm little worried how many combinations we will be testing in query_frontend_test.go. Perhaps it would be worth doing some refactoring there, but it can also lead to big test obfuscation.

I used Table testing pattern but with a generated table here: https://github.com/cortexproject/cortex/blob/master/pkg/distributor/distributor_test.go#L284

Not sure its super useful in this case, but it does allow you to ensure you hit all the combinations of different parameters.

tomwilkie:
> I'm little worried how many combinations we will be testing in query_frontend_test.go. Perhaps it would be worth doing some refactoring there, but it can also lead to big test obfuscation.

I used Table testing pattern but with a generated table here: https://github.com/cortexproject/cortex/blob/master/pkg/distributor/distributor_test.go#L284

Not sure its super useful in this case, but it does allow you to ensure you hit all the combinations of different parameters.

pracucci:
> > I'm little worried how many combinations we will be testing in query_frontend_test.go. Perhaps it would be worth doing some refactoring there, but it can also lead to big test obfuscation.
> 
> I used Table testing pattern but with a generated table here: https://github.com/cortexproject/cortex/blob/master/pkg/distributor/distributor_test.go#L284
> 
> Not sure its super useful in this case, but it does allow you to ensure you hit all the combinations of different parameters.

Not sure it would help much (will end up writing the same logic but in a single test). I've an idea to introduce fixtures support to the e2e framework, in order to be able to write a single test and then run it against different set of fixtures (different configs). I will give it a try when I've some spare time.

pstibrany:
> > I'm little worried how many combinations we will be testing in query_frontend_test.go. Perhaps it would be worth doing some refactoring there, but it can also lead to big test obfuscation.
> 
> Me too. What're your thoughts about the refactoring? What would refactor?

Oh, I've never replied to this. Something like what Tom linked to or what [querier_test.go in Cortex does](https://github.com/cortexproject/cortex/blob/b81d808df52ff61d6603ca739b3957664f2d1280/pkg/querier/chunk_store_queryable_test.go#L18) to test various combinations of encodings and chunk merging algorithms.

pracucci:
> > > I'm little worried how many combinations we will be testing in query_frontend_test.go. Perhaps it would be worth doing some refactoring there, but it can also lead to big test obfuscation.
> > 
> > 
> > Me too. What're your thoughts about the refactoring? What would refactor?
> 
> Oh, I've never replied to this. Something like what Tom linked to or what [querier_test.go in Cortex does](https://github.com/cortexproject/cortex/blob/b81d808df52ff61d6603ca739b3957664f2d1280/pkg/querier/chunk_store_queryable_test.go#L18) to test various combinations of encodings and chunk merging algorithms.

Yes, but we need custom setup for each test (reason why there's the setup function) so I don't see a real benefit (to me it's looks like testing style: single function, like table testing, or multiple functions). However, if we introduce a generic fixture support (moving the custom setup logic to re-usable fixtures), we may actually simplify these kinds of tests and make fixtures reusable across tests, but to me it's a work decoupled from this PR.

pstibrany:
> However, if we introduce a generic fixture support (moving the custom setup logic to re-usable fixtures), we may actually simplify these kinds of tests and make fixtures reusable across tests,

👍 

> but to me it's a work decoupled from this PR.

👍

PR #2168, opened 2020-02-20
delete chunk and seriesIDs

sandeepsukhani:
**What this PR does**:
This is another smaller PR from #2103 which includes just the functionality to allow deletion of chunks and seriesIDs(series store only) for the given time range.

This includes the implementation for just `filesystem` object store and `boltdb` index store. Separate PRs will follow for other stores. 

**Checklist**
- [X] Tests added/updated

### DIFF
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -599,6 +599,7 @@ func (a dynamoDBStorageClient) getDynamoDBChunks(ctx context.Context, chunks []c
 }
 
 func (a dynamoDBStorageClient) DeleteChunk(ctx context.Context, chunkID string) error {
+	// ToDo: implement this to support deleting chunks from DynamoDB
 	return chunk.ErrMethodNotImplemented
 }
 
@@ -754,7 +755,8 @@ func (b dynamoDBWriteBatch) Add(tableName, hashValue string, rangeValue []byte, 
 }
 
 func (b dynamoDBWriteBatch) Delete(tableName, hashValue string, rangeValue []byte) {
-	panic("implement me")
+	// ToDo: implement this to support deleting index entries from DynamoDB
+	panic("DynamoDB does not support Deleting index entries yet")
 }
 
 // Fill 'b' with WriteRequests from 'from' until 'b' has at most max requests. Remove those requests from 'from'.
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -598,6 +598,10 @@ func (a dynamoDBStorageClient) getDynamoDBChunks(ctx context.Context, chunks []c
 	return result, nil
 }
 
+func (a dynamoDBStorageClient) DeleteChunk(ctx context.Context, chunkID string) error {
+	return chunk.ErrMethodNotImplemented
+}
+
 func processChunkResponse(response *dynamodb.BatchGetItemOutput, chunksByKey map[string]chunk.Chunk) ([]chunk.Chunk, error) {
 	result := []chunk.Chunk{}
 	decodeContext := chunk.NewDecodeContext()
@@ -747,6 +751,10 @@ func (b dynamoDBWriteBatch) Add(tableName, hashValue string, rangeValue []byte, 
 			Item: item,
 		},
 	})
+}
+
+func (b dynamoDBWriteBatch) Delete(tableName, hashValue string, rangeValue []byte) {
+	panic("implement me")
 }
 
 // Fill 'b' with WriteRequests from 'from' until 'b' has at most max requests. Remove those requests from 'from'.
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -485,3 +485,109 @@ func (c *store) convertChunkIDsToChunks(ctx context.Context, userID string, chun
 
 	return chunkSet, nil
 }
+
+func (c *store) DeleteChunk(ctx context.Context, from, through model.Time, userID, chunkID string, metric labels.Labels, partiallyDeletedInterval *model.Interval) error {
+	metricName := metric.Get(model.MetricNameLabel)
+	if metricName == "" {
+		return errors.New("No metric name label")
+	}
+
+	chunkWriteEntries, err := c.schema.GetWriteEntries(from, through, userID, string(metricName), metric, chunkID)
+	if err != nil {
+		return err
+	}
+
+	return c.deleteChunk(ctx, userID, chunkID, metric, chunkWriteEntries, partiallyDeletedInterval, func(chunk Chunk) error {
+		return c.PutOne(ctx, chunk.From, chunk.Through, chunk)
+	})
+}
+
+func (c *store) deleteChunk(ctx context.Context, userID, chunkID string, metric labels.Labels,
+	chunkWriteEntries []IndexEntry, partiallyDeletedInterval *model.Interval, putChunkFunc func(chunk Chunk) error) error {
+	metricName := metric.Get(model.MetricNameLabel)
+	if metricName == "" {
+		return errors.New("No metric name label")
+	}
+
+	// if chunk is partially deleted, fetch it, slice non-deleted portion and put it to store before deleting original chunk
+	if partiallyDeletedInterval != nil {
+		err := c.deleteChunkPartially(ctx, userID, chunkID, *partiallyDeletedInterval, putChunkFunc)
+		if err != nil {
+			return err
+		}
+	}
+
+	batch := c.index.NewWriteBatch()
+	for i := range chunkWriteEntries {
+		batch.Delete(chunkWriteEntries[i].TableName, chunkWriteEntries[i].HashValue, chunkWriteEntries[i].RangeValue)
+	}
+
+	err := c.index.BatchWrite(ctx, batch)
+	if err != nil {
+		return err
+	}
+
+	return c.chunks.DeleteChunk(ctx, chunkID)
+}
+
+func (c *store) deleteChunkPartially(ctx context.Context, userID, chunkID string, partiallyDeletedInterval model.Interval, putChunkFunc func(chunk Chunk) error) error {
+	chunk, err := ParseExternalKey(userID, chunkID)
+	if err != nil {
+		return err
+	}
+
+	if !intervalsOverlap(model.Interval{Start: chunk.From, End: chunk.Through}, partiallyDeletedInterval) {
+		return nil
+	}
+
+	chunks, err := c.Fetcher.FetchChunks(ctx, []Chunk{chunk}, []string{chunkID})
+	if err != nil {
+		return err
+	}
+
+	chunk = chunks[0]
+	var newChunks []*Chunk
+	if partiallyDeletedInterval.Start > chunk.From {
+		newChunk, err := chunk.Slice(chunk.From, partiallyDeletedInterval.Start-1)
+		if err != nil && err != ErrSliceNoDataInRange {
+			return err
+		}
+
+		if newChunk != nil {
+			newChunks = append(newChunks, newChunk)
+		}
+	}
+
+	if partiallyDeletedInterval.End < chunk.Through {
+		newChunk, err := chunk.Slice(partiallyDeletedInterval.End+1, chunk.Through)
+		if err != nil && err != ErrSliceNoDataInRange {
+			return err
+		}
+
+		if newChunk != nil {
+			newChunks = append(newChunks, newChunk)
+		}
+	}
+
+	for _, newChunk := range newChunks {
+		// we don't want to create chunks with no samples in it
+		if newChunk.Data.Len() == 0 {
+			continue
+		}
+
+		if err := newChunk.Encode(); err != nil {
+			return err
+		}
+
+		err = putChunkFunc(*newChunk)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *store) DeleteSeriesIDs(ctx context.Context, from, through model.Time, userID string, metric labels.Labels) error {
+	return nil
+}
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -2,7 +2,6 @@ package chunk
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"net/http"
@@ -11,6 +10,7 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
@@ -24,6 +24,11 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 	"github.com/cortexproject/cortex/pkg/util/validation"
+)
+
+var (
+	ErrMetricNameLabelMissing     = errors.New("metric name label missing")
+	ErrParialDeleteChunkNoOverlap = errors.New("interval for partial deletion has not overlap with chunk interval")
 )
 
 var (
@@ -136,7 +141,7 @@ func (c *store) calculateIndexEntries(userID string, from, through model.Time, c
 
 	metricName := chunk.Metric.Get(labels.MetricName)
 	if metricName == "" {
-		return nil, fmt.Errorf("no MetricNameLabel for chunk")
+		return nil, ErrMetricNameLabelMissing
 	}
 
 	entries, err := c.schema.GetWriteEntries(from, through, userID, metricName, chunk.Metric, chunk.ExternalKey())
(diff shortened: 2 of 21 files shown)


pkg/chunk/chunk_store.go
```diff
@@ -482,3 +482,109 @@ func (c *store) convertChunkIDsToChunks(ctx context.Context, userID string, chun
...
+	for i := range chunkWriteEntries {
+		batch.Delete(chunkWriteEntries[i].TableName, chunkWriteEntries[i].HashValue, chunkWriteEntries[i].RangeValue)
+	}
+
+	err := c.index.BatchWrite(ctx, batch)
+	if err != nil {
+		return err
+	}
+
+	return c.chunks.DeleteChunk(ctx, chunkID)
```
pracucci: We should check the error and ignore it (do not propagate it) if it's `chunk.ErrStorageObjectNotFound`. It's better to do the check here because the error could be wrapped to add more contextual information in parent callers.

pkg/chunk/inmemory_storage_client.go
```diff
@@ -311,14 +332,82 @@ func (m *MockStorage) GetChunks(ctx context.Context, chunkSet []Chunk) ([]Chunk,
 	return result, nil
 }
 
-type mockWriteBatch []struct {
-	tableName, hashValue string
-	rangeValue           []byte
-	value                []byte
+// DeleteChunk implements StorageClient.
+func (m *MockStorage) DeleteChunk(ctx context.Context, chunkID string) error {
```
pracucci: I'm a bit confused. Could you explain me why we need to add these functions (`DeleteChunk()`, `GetObject()`, etc) to the `MockStorage` please?

pkg/chunk/inmemory_storage_client.go
```diff
@@ -162,15 +164,9 @@ func (m *MockStorage) BatchWrite(ctx context.Context, batch WriteBatch) error {
...
-			keyType := itemComponents[3][0]
-			if keyType != metricNameRangeKeyV1 &&
-				keyType != seriesRangeKeyV1 &&
-				keyType != labelNamesRangeKeyV1 &&
-				keyType != labelSeriesRangeKeyV1 {
-				return fmt.Errorf("Dupe write")
-			}
+			// if duplicate write then just update the value
+			items[i].value = req.value
+			continue
```
pracucci: Could you explain why we need this change, please?

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -749,6 +753,10 @@ func (b dynamoDBWriteBatch) Add(tableName, hashValue string, rangeValue []byte,
 	})
 }
 
+func (b dynamoDBWriteBatch) Delete(tableName, hashValue string, rangeValue []byte) {
+	panic("implement me")
```
jtlisi: Instead of implement me can you make this panic be a bit more informative

jtlisi: If a backend isn't implemented yet be sure to check for it when the purger starts in your other PR.

jtlisi: Can you add a `// TODO:` statement to each of the unimplemented backends?

pkg/chunk/chunk_store.go
```diff
@@ -482,3 +482,109 @@ func (c *store) convertChunkIDsToChunks(ctx context.Context, userID string, chun
...
+}
+
+func (c *store) deleteChunkPartially(ctx context.Context, userID, chunkID string, partiallyDeletedInterval model.Interval, putChunkFunc func(chunk Chunk) error) error {
```
jtlisi: I think this function should be renamed. It doesn't partially delete a chunk, instead it creates a chunk outside of the deletion boundary. Consider calling it `sliceChunk` or `reboundChunk`.

pkg/chunk/chunk_store.go
```diff
@@ -482,3 +482,109 @@ func (c *store) convertChunkIDsToChunks(ctx context.Context, userID string, chun
...
+	if err != nil {
+		return err
+	}
+
+	return c.deleteChunk(ctx, userID, chunkID, metric, chunkWriteEntries, partiallyDeletedInterval, func(chunk Chunk) error {
+		return c.PutOne(ctx, chunk.From, chunk.Through, chunk)
+	})
+}
+
+func (c *store) deleteChunk(ctx context.Context, userID, chunkID string, metric labels.Labels,
```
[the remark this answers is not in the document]
sandeepsukhani: I think 1 per line would be too less. Maybe split it into 2 or three lines with multiple params per line?

pkg/chunk/chunk_store.go
```diff
@@ -482,3 +482,109 @@ func (c *store) convertChunkIDsToChunks(ctx context.Context, userID string, chun
...
+		if newChunk != nil {
+			newChunks = append(newChunks, newChunk)
+		}
+	}
+
+	for _, newChunk := range newChunks {
+		// we don't want to create chunks with no samples in it
+		if newChunk.Data.Len() == 0 {
+			continue
+		}
```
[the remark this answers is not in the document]
sandeepsukhani: I had it before adding returning error for empty chunks and missed removing this. Will remove it. Thanks!

pkg/chunk/chunk_store.go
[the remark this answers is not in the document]
sandeepsukhani: Sounds good to me. Will do. Thanks!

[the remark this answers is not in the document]
sandeepsukhani: Maybe the caller could do it at single place. What do you think?

pkg/chunk/composite_store.go
```diff
@@ -142,6 +145,20 @@ func (c compositeStore) GetChunkRefs(ctx context.Context, userID string, from, t
...
 }
 
+// DeleteSeriesIDs deletes series IDs from index in series store
+func (c CompositeStore) DeleteSeriesIDs(ctx context.Context, from, through model.Time, userID string, metric labels.Labels) error {
+	return c.forStores(from, through, func(from, through model.Time, store Store) error {
+		return store.DeleteSeriesIDs(ctx, from, through, userID, metric)
+	})
+}
+
+// DeleteChunk removes a chunk and its reference from index
```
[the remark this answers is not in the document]
sandeepsukhani: It deletes it both from the index table and chunk store. You want me to change the comment?

pkg/chunk/chunk_store.go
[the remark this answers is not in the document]
pracucci: It's pretty much a standard in go. Either all params in 1 line or 1 param per line.

pkg/chunk/composite_store.go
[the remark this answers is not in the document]
pracucci: > You want me to change the comment?

Yes, please. Let's try to be as much accurate as possible in comments. Reading a comment is 10 seconds. Reading the function code is at least 10x 😉

pkg/chunk/composite_store.go
```diff
@@ -25,6 +25,9 @@ type Store interface {
 	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([][]Chunk, []*Fetcher, error)
 	LabelValuesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string, labelName string) ([]string, error)
 	LabelNamesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string) ([]string, error)
+
+	DeleteChunk(ctx context.Context, from, through model.Time, userID, chunkID string, metric labels.Labels, partiallyDeletedInterval *model.Interval) error
+	DeleteSeriesIDs(ctx context.Context, from, through model.Time, userID string, metric labels.Labels) error
```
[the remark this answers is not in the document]
sandeepsukhani: I don't have a strong opinion. It literally deletes `SeriesIDs` so I named it that way. I think `DeleteSeriesIndex` sounds good too. @jtlisi what do you think?

pkg/chunk/schema_config.go
```diff
@@ -320,10 +320,11 @@ func (cfg SchemaConfig) PrintYaml() {
...
 type Bucket struct {
-	from      uint32
-	through   uint32
-	tableName string
-	hashKey   string
+	from       uint32
+	through    uint32
+	tableName  string
+	hashKey    string
+	bucketSize uint32 // helps with deletion of series ids in series store
```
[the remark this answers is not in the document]
sandeepsukhani: I think people usually go to the definition of the struct to find more details about it. I will add it to the unit as well then. Thanks!

pkg/chunk/schema.go
```diff
@@ -204,6 +208,67 @@ func (s schema) GetChunksForSeries(from, through model.Time, userID string, seri
...
+	if len(buckets) == 0 {
+		return nil, nil
+	}
+
+	seriesID := string(labelsSeriesID(metric))
+
+	// Only first and last buckets needs to be checked for in-use series ids.
+	// Only partially deleted first/last deleted bucket needs to be checked otherwise
+	// not since whole bucket is anyways considered for deletion.
+	if buckets[0].from != 0 {
```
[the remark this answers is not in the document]
sandeepsukhani: Yes you are right, will add a comment. Thanks!

pkg/chunk/schema.go
```diff
@@ -204,6 +208,67 @@ func (s schema) GetChunksForSeries(from, through model.Time, userID string, seri
...
+		bucketStartTime := from - model.Time(buckets[0].from)
```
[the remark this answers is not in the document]
sandeepsukhani: from would always be bigger because `bucket.from` is relative i.e for a per-day bucket it would be the number of seconds elapsed since the start of that day. Sounds good?

pkg/chunk/schema.go
```diff
@@ -204,6 +208,67 @@ func (s schema) GetChunksForSeries(from, through model.Time, userID string, seri
...
+			buckets = buckets[1:]
+			if len(buckets) == 0 {
+				return nil, nil
+			}
+		}
+	}
+
+	lastBucket := buckets[len(buckets)-1]
+
+	if lastBucket.through != lastBucket.bucketSize {
```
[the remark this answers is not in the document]
sandeepsukhani: Yes you are right, will add a comment. Thanks!

pkg/chunk/schema.go
```diff
@@ -204,6 +208,67 @@ func (s schema) GetChunksForSeries(from, through model.Time, userID string, seri
 	return result, nil
 }
 
+// returns IndexEntrys for deletion. It makes sure that we don't include series entries which are in use by verifying using hasChunksForIntervalFunc
+func (s schema) GetSeriesDeleteEntries(from, through model.Time, userID string, metric labels.Labels, hasChunksForIntervalFunc hasChunksForIntervalFunc) ([]IndexEntry, error) {
```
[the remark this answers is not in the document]
sandeepsukhani: `SeriesID` is stored once per bucket so for partial deletion, we don't want to delete those entries otherwise rest of the chunks won't be accessible. Will add a comment. Thanks!

pkg/chunk/chunk_store.go
sandeepsukhani: I wanted to leave it up to the caller to decide whether its a problem or not. I will give it a thought.

pkg/chunk/inmemory_storage_client.go
sandeepsukhani: `MockStorage` is used at various places for doing store tests. I think for this PR I don't need them. I will need it for my next PR so I can either leave it here for now or add it in the next PR. What do you think?

pkg/chunk/inmemory_storage_client.go
sandeepsukhani: The Add method in all the other stores does the updation of value if the record already exists. I wanted the behaviour of `MockStorage` to be the same.

pkg/chunk/aws/dynamodb_storage_client.go
sandeepsukhani: Sure will do. Thanks!

pkg/chunk/chunk_store.go
sandeepsukhani: `sliceChunk` sounds good to me. Thanks!

pkg/chunk/series_store.go
```diff
@@ -509,3 +510,58 @@ func (c *seriesStore) calculateIndexEntries(ctx context.Context, from, through m
...
+	}
+
+	chunks, err := c.convertChunkIDsToChunks(ctx, userID, chunkIDs)
+	if err != nil {
+		return false, err
+	}
+
+	seriesInUse := false
+	for i := range chunks {
+		if from <= chunks[i].From && chunks[i].From <= through || from <= chunks[i].Through && chunks[i].Through <= through {
```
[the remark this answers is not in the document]
sandeepsukhani: I think not. I have a better function here: https://github.com/cortexproject/cortex/pull/2168/files/3bea3320f83e0edd296c452f1771fa6c17ef489f#diff-7865f904ff5d528ee5d205326baf011eR378
Will add test for it.

pkg/chunk/local/fs_object_client.go
```diff
@@ -106,6 +111,19 @@ func (f *FSObjectClient) List(ctx context.Context, prefix string) ([]chunk.Stora
 	return storageObjects, nil
 }
 
+func (f *FSObjectClient) DeleteChunk(ctx context.Context, chunkID string) error {
```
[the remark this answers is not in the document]
sandeepsukhani: Right, will change it. Thanks!

pkg/chunk/chunk_store.go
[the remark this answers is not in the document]
pracucci: Doing it in the caller is an alternative. The main advantage of doing it here, in my opinion, is that you can be more specific, with a different message for each single error.

pkg/chunk/inmemory_storage_client.go
pracucci: Ok then.

pkg/chunk/inmemory_storage_client.go
pracucci: I'm OK keeping it if necessary (whether in this or subsequent PR). The only thing which sounds as a bad smell to me (but I may be wrong cause I didn't check all usages of `MockStorage`) is that after Jacob refactoring we now have two interfaces: one for the chunk client and one for the object store client. Here we're mixing both interfaces in the same implementation and I was wondering if this is required or not. If yes, I'm good.

pkg/chunk/chunk_store.go
sandeepsukhani: I think I will go with `reboundChunk` because it creates a confusion with `Chunk.Slice` method which creates a new chunk for the given interval while this removes the given interval from the chunk.

pkg/chunk/inmemory_storage_client.go
sandeepsukhani: `MockStorage` is jack of all trades for tests that need stores. Which means it needs to satisfy both the interfaces.
Regarding reusing code, there is a package called `objectclient` which takes in `chunks.ObjectClient` and converts it to `objectclient.Client`. `chunks.ObjectClient` works on raw objects while the methods in `objectclient.Client` reuse that code to allow working on chunks. The package `objectclient` imports `chunks` package while the implementation of `MockStorage` lives in `chunks` package which means I can't import `objectclient` back in `chunks` package to reuse the code which does the conversion of interfaces.

pkg/chunk/schema_config.go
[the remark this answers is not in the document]
pracucci: Have you considered adding the unit?

pkg/chunk/chunk_store.go
```diff
@@ -482,3 +487,126 @@ func (c *store) convertChunkIDsToChunks(ctx context.Context, userID string, chun
...
+	putChunkFunc func(chunk Chunk) error) error {
+
+	metricName := metric.Get(model.MetricNameLabel)
+	if metricName == "" {
+		return ErrMetricNameLabelMissing
+	}
+
+	// if chunk is partially deleted, fetch it, slice non-deleted portion and put it to store before deleting original chunk
+	if partiallyDeletedInterval != nil {
+		err := c.reboundChunk(ctx, userID, chunkID, *partiallyDeletedInterval, putChunkFunc)
```
pracucci: In case the error `ErrParialDeleteChunkNoOverlap`, shouldn't we ignore it and `return nil`? It means there's nothing to do but it's also OK.

pkg/chunk/chunk_store.go
```diff
@@ -482,3 +487,126 @@ func (c *store) convertChunkIDsToChunks(ctx context.Context, userID string, chun
 
 	return chunkSet, nil
 }
+
+func (c *store) DeleteChunk(ctx context.Context, from, through model.Time, userID, chunkID string, metric labels.Labels, partiallyDeletedInterval *model.Interval) error {
```
pracucci: Now that I'm getting a better understand of the logic, I think that passing `partiallyDeletedInterval` is a bit confusing. Have you considered splitting the interface into:
- `DeleteChunk(ctx context.Context, from, through model.Time, userID, chunkID string, metric labels.Labels)`
- `ReboundChunk(ctx context.Context, from, through model.Time, userID, chunkID string, metric labels.Labels, reboundInterval model.Interval)`

Internally, it's perfectly fine to have the same shared implementation, but splitting the interface would make the intention more clear in my opinion:
- `DeleteChunk()` deletes an entire chunk
- `ReboundChunk()` deletes a portion of the chunk, rebounding it

I would like to know what's your take on this.

pkg/chunk/chunk_store.go
sandeepsukhani: I wanted to return an error to tell the caller that it is doing something wrong. 
A caller should never try to delete chunk partially with an interval which is out of its range.
What do you think?

pracucci: Good point. Sounds good as is then.

pkg/chunk/chunk_store.go
pracucci: _Let's keep this as food for thought for the next PR. We can eventually iteratively address it._

PR #2169, opened 2020-02-21
fix docs related to rulers

Wing924:

**What this PR does**:

fix docs related to rulers
added configs API docs

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added


docs/architecture.md
```diff
@@ -221,10 +222,20 @@ The query frontend supports caching query results and reuses them on subsequent
 
 The **ruler** is an **optional service** executing PromQL queries for recording rules and alerts. The ruler requires a database storing the recording rules and alerts for each tenant.
 
-Ruler can be scaled horizontally.
+Ruler are **stateless** and can be scaled horizontally.
```
pracucci:
```suggestion
Ruler is **stateless** and can be scaled horizontally.
```

docs/architecture.md
```diff
@@ -221,10 +222,20 @@ The query frontend supports caching query results and reuses them on subsequent
...
 
 ### Alertmanager
 
 The **alertmanager** is an **optional service** responsible for accepting alert notifications from the [ruler](#ruler), deduplicating and grouping them, and routing them to the correct notification channel, such as email, PagerDuty or OpsGenie.
 
 The Cortex alertmanager is built on top of the [Prometheus Alertmanager](https://prometheus.io/docs/alerting/alertmanager/), adding multi-tenancy support. Like the [ruler](#ruler), the alertmanager requires a database storing the per-tenant configuration.
+
+Alertmanager are **stateful**.
```
pracucci: Could you elaborate why you consider the Alertmanager stateful? It's very likely I'm missing something obvious.

@jtlisi Do you think it's correct stating that Alertmanager is stateful while Ruler is not?

docs/architecture.md
```diff
@@ -221,10 +222,20 @@ The query frontend supports caching query results and reuses them on subsequent
...
+
+### Configs API
+
+The **configs API** is an **optional service** managing the configuration of Rulers and Alertmanagers.
+It provides APIs to get/set/update the rules and alertmanager configurations and store them into backend.
+Current supported backend are PostgreSQL, static files and in-memory.
+
+Configs API are **stateless** and can be scaled up and down as needed.
```
pracucci:
```suggestion
Configs API is **stateless**.
```

Not sure it's worth to mention the "scaled up and down as needed" cause you don't really need to scale it up/ down (we don't expect an huge traffic on these endpoints). What you usually want is just having a couple of replicas for HA purposes.

docs/architecture.md
Wing924: The Peometheus Alertmanager need to store alert state and silence rule.
I think Cortex one also need it. And cortex alertmanager have this config:
```
# Base path for data storage.
# CLI flag: -alertmanager.storage.path
[datadir: <string> | default = "data/"]
```

docs/architecture.md
```diff
@@ -221,10 +222,20 @@ The query frontend supports caching query results and reuses them on subsequent
...
-Ruler can be scaled horizontally.
+Ruler is **stateless** and can be scaled horizontally.
 
 ### Alertmanager
 
 The **alertmanager** is an **optional service** responsible for accepting alert notifications from the [ruler](#ruler), deduplicating and grouping them, and routing them to the correct notification channel, such as email, PagerDuty or OpsGenie.
 
 The Cortex alertmanager is built on top of the [Prometheus Alertmanager](https://prometheus.io/docs/alerting/alertmanager/), adding multi-tenancy support. Like the [ruler](#ruler), the alertmanager requires a database storing the per-tenant configuration.
+
+Alertmanager are **stateful**.
```
pracucci:
```suggestion
Alertmanager is **stateful**.
```

docs/architecture.md
pracucci: Yes, true.

I'm not an expert of it, but if clustering is used it's somewhat half stateful, in the sense that you don't necessarily need data persistency cause data is replicated across alertmanager instances (and you end up with state/silence loss only if you lose all instances at once).

@jtlisi Could you share more details to better phrase this stateful thing in the doc, please?

Wing924:
@pracucci @jtlisi Could you review this?

docs/architecture.md
```diff
@@ -221,10 +222,20 @@ The query frontend supports caching query results and reuses them on subsequent
...
 
 The Cortex alertmanager is built on top of the [Prometheus Alertmanager](https://prometheus.io/docs/alerting/alertmanager/), adding multi-tenancy support. Like the [ruler](#ruler), the alertmanager requires a database storing the per-tenant configuration.
+
+Alertmanager is **stateful**.
+
+### Configs API
+
+The **configs API** is an **optional service** managing the configuration of Rulers and Alertmanagers.
+It provides APIs to get/set/update the rules and alertmanager configurations and store them into backend.
+Current supported backend are PostgreSQL, static files and in-memory.
```
jtlisi: I don't think static files are supported in the configs api. However, the alertmanager itself has a support for static files as an alternative to the config API.

docs/architecture.md
jtlisi: Both of these services can be considered stateful to different degrees. The Alertmanager persists information about silences and active alerts to its disk. If all of the alertmanager nodes failed simultaneously there would be a loss of data.

The ruler is stateful as well. Running rules internally have state, as well as the ring the rulers initiate. However, if the rulers all fail and restart, Prometheus alert rules have a feature where an alert is restored and returned to a firing state if it would have been active in its `for` period. However, there would be gaps in the series generated by the recording rules.

I don't think the term "stateful" is particularly helpful in this context. It might be better to explain how the Alertmanager persists it's the state of alerts and active silences to a local disk and that it is recommended they are run using a stateful set.

Wing924:
@jtlisi Could you review it again?

PR #2172, opened 2020-02-21
TSDB head compactor concurrency

### LINKED ISSUE #2003: Set a max concurrency TSDB blocks compaction

pstibrany:
**What this PR does**: This PR makes TSDB head compaction interval and concurrency configurable (defaults to 2h). Shipping now happens right after compaction, and previous options to control shipping interval and concurrency have been removed. Shipping can still be disabled.

Fixes #2003

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 * [CHANGE] Renamed the cache configuration setting `defaul_validity` to `default_validity`. #2140
 * [CHANGE] Removed unused /validate_expr endpoint. #2152
 * [CHANGE] Updated Prometheus dependency to v2.16.0. This Prometheus version uses Active Query Tracker to limit concurrent queries. In order to keep `-querier.max-concurrent` working, Active Query Tracker is enabled by default, and is configured to store its data to `active-query-tracker` directory (relative to current directory when Cortex started). This can be changed by using `-querier.active-query-tracker-dir` option. Purpose of Active Query Tracker is to log queries that were running when Cortex crashes. This logging happens on next Cortex start. #2088
-* [CHANGE] Experimental TSDB: TSDB head compaction interval is now configurable (defaults to 2h). Shipping now happens right after compaction, and shipping interval and concurrency options were removed. Shipping can still be disabled. #2172
+* [CHANGE] Experimental TSDB: TSDB head compaction interval and concurrency is now configurable (defaults to 5m interval and 5 concurrent head compactions). New options: `-experimental.tsdb.head-compaction-interval` and `-experimental.tsdb.head-compaction-concurrency`. #2172
 * [FEATURE] Added a read-only local alertmanager config store using files named corresponding to their tenant id. #2125
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
   * `--experimental.distributor.user-subring-size`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 * [CHANGE] Renamed the cache configuration setting `defaul_validity` to `default_validity`. #2140
 * [CHANGE] Removed unused /validate_expr endpoint. #2152
 * [CHANGE] Updated Prometheus dependency to v2.16.0. This Prometheus version uses Active Query Tracker to limit concurrent queries. In order to keep `-querier.max-concurrent` working, Active Query Tracker is enabled by default, and is configured to store its data to `active-query-tracker` directory (relative to current directory when Cortex started). This can be changed by using `-querier.active-query-tracker-dir` option. Purpose of Active Query Tracker is to log queries that were running when Cortex crashes. This logging happens on next Cortex start. #2088
-* [CHANGE] Experimental TSDB: TSDB head compaction interval and concurrency is now configurable (defaults to 5m interval and 5 concurrent head compactions). New options: `-experimental.tsdb.head-compaction-interval` and `-experimental.tsdb.head-compaction-concurrency`. #2172
+* [CHANGE] Experimental TSDB: TSDB head compaction interval and concurrency is now configurable (defaults to 1 min interval and 5 concurrent head compactions). New options: `-experimental.tsdb.head-compaction-interval` and `-experimental.tsdb.head-compaction-concurrency`. #2172
 * [FEATURE] Added a read-only local alertmanager config store using files named corresponding to their tenant id. #2125
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
   * `--experimental.distributor.user-subring-size`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 * [CHANGE] Renamed the cache configuration setting `defaul_validity` to `default_validity`. #2140
 * [CHANGE] Removed unused /validate_expr endpoint. #2152
 * [CHANGE] Updated Prometheus dependency to v2.16.0. This Prometheus version uses Active Query Tracker to limit concurrent queries. In order to keep `-querier.max-concurrent` working, Active Query Tracker is enabled by default, and is configured to store its data to `active-query-tracker` directory (relative to current directory when Cortex started). This can be changed by using `-querier.active-query-tracker-dir` option. Purpose of Active Query Tracker is to log queries that were running when Cortex crashes. This logging happens on next Cortex start. #2088
-* [CHANGE] Experimental TSDB: TSDB head compaction interval is now configurable (defaults to 2h). Shipping now happens right after compaction, and shipping interval and concurrency options were removed. Shipping can still be disabled. 
+* [CHANGE] Experimental TSDB: TSDB head compaction interval is now configurable (defaults to 2h). Shipping now happens right after compaction, and shipping interval and concurrency options were removed. Shipping can still be disabled. #2172
 * [FEATURE] Added a read-only local alertmanager config store using files named corresponding to their tenant id. #2125
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
   * `--experimental.distributor.user-subring-size`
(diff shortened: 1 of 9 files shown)


pkg/storage/tsdb/config.go
```diff
@@ -104,10 +106,11 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Dir, "experimental.tsdb.dir", "tsdb", "directory to place all TSDB's into")
 	f.Var(&cfg.BlockRanges, "experimental.tsdb.block-ranges-period", "comma separated list of TSDB block ranges in time.Duration format")
 	f.DurationVar(&cfg.Retention, "experimental.tsdb.retention-period", 6*time.Hour, "TSDB block retention")
-	f.DurationVar(&cfg.ShipInterval, "experimental.tsdb.ship-interval", 1*time.Minute, "How frequently the TSDB blocks are scanned and new ones are shipped to the storage. 0 means shipping is disabled.")
-	f.IntVar(&cfg.ShipConcurrency, "experimental.tsdb.ship-concurrency", 10, "Maximum number of tenants concurrently shipping blocks to the storage.")
 	f.StringVar(&cfg.Backend, "experimental.tsdb.backend", "s3", "TSDB storage backend to use")
 	f.IntVar(&cfg.MaxTSDBOpeningConcurrencyOnStartup, "experimental.tsdb.max-tsdb-opening-concurrency-on-startup", 10, "limit the number of concurrently opening TSDB's on startup")
+	f.DurationVar(&cfg.CompactionInterval, "experimental.tsdb.compaction-interval", 2*time.Hour, "How frequently should locally-stored TSDB blocks be compacted (and shipped)")
```
gouthamve: This is misleading. The blocks are compacted every 2hrs irrespective of this setting. I'd set it to something like 10mins by default. I know that most of the checks will be no-ops, but I'd rather check it more frequently than less frequently. It is likely that the blocks will become `5hrs` long with this setting (3hrs + 2hrs).

This is because when you call `db.Compact()`, it'll compact only if the `maxt - mint > 1.5 * blockRanges[0]`, else it'll shortcut and I think the default `blockRanges[0]` is `2h` anyways.

pstibrany: > This is misleading. The blocks are compacted every 2hrs irrespective of this setting. I'd set it to something like 10mins by default. I know that most of the checks will be no-ops, but I'd rather check it more frequently than less frequently. It is likely that the blocks will become `5hrs` long with this setting (3hrs + 2hrs).
> 
> This is because when you call `db.Compact()`, it'll compact only if the `maxt - mint > 1.5 * blockRanges[0]`, else it'll shortcut and I think the default `blockRanges[0]` is `2h` anyways.

You're absolutely right... I was just looking at this, while adding metrics, and realized that 2 hours doesn't make sense. Prometheus itself tries to run this every minute (!) I've now set default to 5 mins. I should update the argument help too. I feel like this is such a low-level setting, that exposing it to user may cause more harm than good.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 * [CHANGE] Renamed the cache configuration setting `defaul_validity` to `default_validity`. #2140
 * [CHANGE] Removed unused /validate_expr endpoint. #2152
 * [CHANGE] Updated Prometheus dependency to v2.16.0. This Prometheus version uses Active Query Tracker to limit concurrent queries. In order to keep `-querier.max-concurrent` working, Active Query Tracker is enabled by default, and is configured to store its data to `active-query-tracker` directory (relative to current directory when Cortex started). This can be changed by using `-querier.active-query-tracker-dir` option. Purpose of Active Query Tracker is to log queries that were running when Cortex crashes. This logging happens on next Cortex start. #2088
+* [CHANGE] Experimental TSDB: TSDB head compaction interval is now configurable (defaults to 2h). Shipping now happens right after compaction, and shipping interval and concurrency options were removed. Shipping can still be disabled. #2172
```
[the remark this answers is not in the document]
pstibrany: This comment is now obsolete. Shipping and head compaction are separate again.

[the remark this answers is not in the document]
pstibrany: Done.

pkg/storage/tsdb/config.go
pstibrany: Updated the help too.

docs/operations/blocks-storage.md
```diff
@@ -123,6 +123,14 @@ tsdb:
   # CLI flag: -experimental.tsdb.ship-concurrency
   [ship_concurrency: <int> | default = 10]
 
+  # How frequently does Cortex try to compact TSDB head. Block is only created if data covers smallest block range.
+  # CLI flag: -experimental.tsdb.head-compaction-interval
+  [head_compaction_interval: <duration> | default = 2h]
```
gouthamve: Hrm, this should be `5m` and the tests should fail. Can you check why its not failing?

pstibrany: I have edited this file manually. Somehow it isn't covered by Marco's tool, but haven't looked into more detail why yet. I'll fix the default.

pstibrany: I think the reason is simply because block-storage is a separate file, while doc tool only generates config-file-reference.md. @pracucci surely knows better, I'm just guessing based on 5min looking around.

pracucci:
Fixes #2174

pkg/ingester/ingester_v2.go
```diff
@@ -129,6 +145,11 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 	i.done.Add(1)
 	go i.updateLoop()
 
+	if i.cfg.TSDBConfig.HeadCompactionConcurrency > 0 && i.cfg.TSDBConfig.HeadCompactionInterval > 0 {
```
pracucci: Why is it possible to disable compactions? We should never allow it. Even if you run without shipper you definitely wants compactions otherwise your memory will soon explode.

I would suggest:
1. Add a config validation to ensure that `HeadCompactionConcurrency > 0`
2. Always run `compactionLoop()`

What's your take?

pkg/storage/tsdb/config.go
```diff
@@ -108,6 +112,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.ShipConcurrency, "experimental.tsdb.ship-concurrency", 10, "Maximum number of tenants concurrently shipping blocks to the storage.")
 	f.StringVar(&cfg.Backend, "experimental.tsdb.backend", "s3", "TSDB storage backend to use")
 	f.IntVar(&cfg.MaxTSDBOpeningConcurrencyOnStartup, "experimental.tsdb.max-tsdb-opening-concurrency-on-startup", 10, "limit the number of concurrently opening TSDB's on startup")
+	f.DurationVar(&cfg.HeadCompactionInterval, "experimental.tsdb.head-compaction-interval", 5*time.Minute, "How frequently does Cortex try to compact TSDB head. Block is only created if data covers smallest block range.")
```
pracucci: 5m looks a bit unfrequent to me. This is a very inexpensive operation and I would suggest to run it more frequently: 30s or 1m.

pkg/storage/tsdb/config.go
```diff
@@ -108,6 +112,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
+	f.IntVar(&cfg.HeadCompactionConcurrency, "experimental.tsdb.head-compaction-concurrency", 5, "Maximum number of tenants concurrently compacting locally-stored TSDB blocks")
```
pracucci: > compacting locally-stored TSDB blocks

The description of `head-compaction-interval` specifically mentions the head, so maybe it's better to keep it consistent here as well and mention something like "Maximum number of tenants concurrently compacting TSDB head into a new block". What's your take?

pkg/ingester/ingester_v2.go
pstibrany: I left an option to disable it when running tests. But now I think it doesn't make sense even in that scenario. I'll modify it.

pkg/storage/tsdb/config.go
pstibrany: Do we even need to make it configurable? I'd just hardcode 1m into the code.

pkg/storage/tsdb/config.go
pstibrany: Sounds good.

pkg/storage/tsdb/config.go
pracucci: I thought about hardcoding to 1m as well, and that was my initial comment. I then changed my mind because I think being able to configure it could help future integration tests (ie. compact every 5s to have integration tests running faster). What's your take?

P.S. I'm talking about integration tests cause I would like to add at least 1 integration test testing the compaction and shipping soon.

pstibrany: We need a secret option then ;-) I'd suggest adding a check to avoid setting it too high (> 5 mins).

pracucci: > We need a secret option then ;-)

We can have a secret YAML config option with `doc:"hidden"` but I'm not sure it's possible to have an hidden CLI flag (I haven't investigated).

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 * [CHANGE] Renamed the cache configuration setting `defaul_validity` to `default_validity`. #2140
 * [CHANGE] Removed unused /validate_expr endpoint. #2152
 * [CHANGE] Updated Prometheus dependency to v2.16.0. This Prometheus version uses Active Query Tracker to limit concurrent queries. In order to keep `-querier.max-concurrent` working, Active Query Tracker is enabled by default, and is configured to store its data to `active-query-tracker` directory (relative to current directory when Cortex started). This can be changed by using `-querier.active-query-tracker-dir` option. Purpose of Active Query Tracker is to log queries that were running when Cortex crashes. This logging happens on next Cortex start. #2088
+* [CHANGE] Experimental TSDB: TSDB head compaction interval and concurrency is now configurable (defaults to 5m interval and 5 concurrent head compactions). New options: `-experimental.tsdb.head-compaction-interval` and `-experimental.tsdb.head-compaction-concurrency`. #2172
```
pracucci: `defaults to 5m` > `defaults to 1m`

pstibrany: 🤦‍♂ Thanks, fixed! :)

PR #2176, opened 2020-02-24
Hide passwords in URLs on the /config endpoint.

tomwilkie:

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/util/flagext/secret_test.go
+++ b/pkg/util/flagext/secret_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func TestSecretdYAML(t *testing.T) {
-	// Test embedding of Password.
+	// Test embedding of Secret.
 	{
 		type TestStruct struct {
 			Secret Secret `yaml:"secret"`
@@ -25,14 +25,14 @@ func TestSecretdYAML(t *testing.T) {
 		assert.Equal(t, expected, actual)
 
 		var actualStruct TestStruct
-		yamlPassword := []byte(`secret: pa55w0rd
+		yamlSecret := []byte(`secret: pa55w0rd
 `)
-		err = yaml.Unmarshal(yamlPassword, &actualStruct)
+		err = yaml.Unmarshal(yamlSecret, &actualStruct)
 		require.NoError(t, err)
 		assert.Equal(t, testStruct, actualStruct)
 	}
 
-	// Test pointers of Password.
+	// Test pointers of Secret.
 	{
 		type TestStruct struct {
 			Secret *Secret `yaml:"secret"`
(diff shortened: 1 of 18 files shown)


Wing924:
I think we also need to mask db and cassandra password.

Wing924:
How about create `flagext.Password` for those secret?

pracucci:
> How about create flagext.Password for those secret?

Makes sense. Maybe `flagext.Secret` is more appropriate, cause it's not just passwords but sensitive data in general.

tomwilkie:
> S3 secret access key, GCS service_account

AFAICT these are part of a URL, so handled as part of the original `URLValue` changes.

> Possibly others, I haven't fully investigated the whole config

Also Cassandra, didn't find anymore.

pracucci:
> AFAICT these are part of a URL, so handled as part of the original URLValue changes.

@tomwilkie We're both right. You're right about the chunks storage, while I had the blocks storage in mind. Please take a look here:

https://github.com/cortexproject/cortex/blob/9cec079a434079e3989f924e323c99231df226a2/pkg/storage/tsdb/backend/s3/config.go#L8-L14

https://github.com/cortexproject/cortex/blob/9cec079a434079e3989f924e323c99231df226a2/pkg/storage/tsdb/backend/gcs/config.go#L8-L11

tomwilkie:
> cortex/pkg/storage/tsdb/backend/gcs/config.go

I didn't think ServiceAccount is secret, is it?  I thought GCS credentials were always read from the env...

pkg/util/flagext/secret_test.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
+)
+
+func TestSecretdYAML(t *testing.T) {
+	// Test embedding of Password.
```
pracucci: Please replace `Password` with `Secret` in the comments, for correctness.

pkg/util/flagext/secret_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+		var actualStruct TestStruct
+		yamlSecret := []byte(`secret: pa55w0rd
+`)
+		err = yaml.Unmarshal(yamlSecret, &actualStruct)
+		require.NoError(t, err)
+		assert.Equal(t, testStruct, actualStruct)
+	}
+
+	// Test no url set in Password.
```
pracucci: > no url set

I guess you mean "no value set"

tomwilkie:
Thank you for your review @pracucci!

PR #2181, opened 2020-02-25
Configs API: Allow POST configs in YAML format.

### LINKED ISSUE #2170: feat: support YAML format for config API
### WHAT

Support posting YAML body to configs API: `/api/prom/configs/rules`.

### WHY

Current format example:
```json
{
    "id": 4,
    "rule_format_version": "2",
    "rules_files": {
        "rule1.yml": "groups:\n- name: container_spec_memory_limit_bytes:container\n  rules:\n  - record: container_spec_memory_limit_bytes:container\n    expr: max by (namespace,container)(container_spec_memory_limit_bytes{container!=\"POD\",container!=\"\"})\n"
    }
}
```

It's hard to read and write without some custom tool.

If we support YAML, the example become:
```yml
id: 1
rule_format_version: 2
rules_files:
  rule1.yml: |
    groups:
    - name: container_spec_memory_limit_bytes:container
      rules:
      - record: container_spec_memory_limit_bytes:container
        expr: max by (namespace,container)(container_spec_memory_limit_bytes{container!=\"POD\",container!=\"\"})
```

It's much easier for reading and writing.

### Backward Compatibility

Since YAML is superset for JSON, there is no backward compatibility problem.

Wing924:
**What this PR does**:

allow post configs API in YAML format
allow get configs API in YAML format

Fixes #2170

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci:
_Removed myself and Goutham as reviewers cause this is @jtlisi area of expertise. I can jump in as second reviewer, if necessary._

Wing924:
@jtlisi Could you review this?

pkg/configs/api/api.go
```diff
@@ -111,7 +113,7 @@ func (a *API) setConfig(w http.ResponseWriter, r *http.Request) {
 	logger := util.WithContext(r.Context(), util.Logger)
 
 	var cfg configs.Config
-	if err := json.NewDecoder(r.Body).Decode(&cfg); err != nil {
+	if err := yaml.NewDecoder(r.Body).Decode(&cfg); err != nil {
```
tomwilkie: Instead of assuming JSON is a subset of YAML for this to work (works most of the time, but there are corner cases), can you use a content-type head to detect YAML?  And have JSON be the default.

pkg/configs/api/api.go
```diff
@@ -276,3 +306,29 @@ func (a *API) restoreConfig(w http.ResponseWriter, r *http.Request) {
...
 	w.WriteHeader(http.StatusOK)
 }
+
+const (
+	FormatInvalid = "invalid"
+	FormatJSON    = "json"
+	FormatYAML    = "yaml"
+)
+
+func ParseConfigFormat(v string, defaultFormat string) string {
```
pracucci: No need to expose this function. Can be `parseConfigFormat()`

pkg/configs/api/api.go
```diff
@@ -93,12 +97,22 @@ func (a *API) getConfig(w http.ResponseWriter, r *http.Request) {
...
+	switch ParseConfigFormat(r.Header.Get("Accept"), FormatJSON) {
+	case FormatJSON:
+		w.Header().Set("Content-Type", "application/json")
+		err = json.NewEncoder(w).Encode(cfg)
+	case FormatYAML:
+		w.Header().Set("Content-Type", "application/yaml")
+		err = yaml.NewEncoder(w).Encode(cfg)
+	default:
+		// should not be here
+		level.Error(logger).Log("msg", "unknown Format")
```
pracucci: Not sure it's worth to log it as error. This case depends on the client input and we generally don't log as error any `4xx` error. We could omit this log at all.

pkg/configs/api/api.go
```diff
@@ -93,12 +97,22 @@ func (a *API) getConfig(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
pracucci: This should be a `StatusBadRequest` error, not `StatusInternalServerError `, because it's caused by an invalid input from client.

pkg/configs/api/api.go
```diff
@@ -111,12 +125,28 @@ func (a *API) setConfig(w http.ResponseWriter, r *http.Request) {
...
+	case FormatYAML:
+		if err := yaml.NewDecoder(r.Body).Decode(&cfg); err != nil {
+			// XXX: Untested
+			level.Error(logger).Log("msg", "error decoding yaml body", "err", err)
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	default:
+		// should not be here
+		level.Error(logger).Log("msg", "unknown Format")
```
pracucci: No need to log this error (actually, not an error, just invalid input from the client).

pkg/configs/api/api.go
Wing924: If I don't expose this function, I can't write test against this function.

pkg/configs/api/api.go
Wing924: actually, this block is unreachable because parseConfigFormat has a default value.

pkg/configs/api/api.go
Wing924: actually, this block is unreachable because parseConfigFormat has a default value.

pkg/configs/api/api.go
pracucci: This function is tested within the same package `pkg/configs/api/api_test.go` so you shouldn't have any problem testing it even if unexported (because unexported functions are still visibile within the same package).

pkg/configs/api/api.go
pracucci:
```suggestion
		// should never reach this point
		level.Error(logger).Log("msg", "unexpected error detecting the config format")
		http.Error(w, err.Error(), http.StatusInternalServerError)
```

pkg/configs/api/api.go
```diff
@@ -111,12 +125,28 @@ func (a *API) setConfig(w http.ResponseWriter, r *http.Request) {
...
 		http.Error(w, err.Error(), http.StatusBadRequest)
```
pracucci:
```suggestion
		// should never reach this point
		level.Error(logger).Log("msg", "unexpected error detecting the config format")
		http.Error(w, err.Error(), http.StatusInternalServerError)
```

Wing924:
@pracucci since `api_test.go` use `api_test` package while `api.go` use `api` package, I created new `api2_test.go` to do test against private function.

pracucci:
> @pracucci since api_test.go use api_test package while api.go use api package, I created new api2_test.go to do test against private function.

Hold on. Why "api_test.go use api_test package"? I think that's a mistake. It should just use the same package of api.go

Wing924:
@pracucci You're right. There is no reason to use `api_test` package.

PR #2185, opened 2020-02-25
tsdb: expose stripe size option to reduce tsdb memory footprint

thorfour:

**What this PR does**:

Exposes the new TSDB StripeSize option for reducing memory footprint.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/storage/tsdb/config.go
```diff
@@ -114,6 +115,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxTSDBOpeningConcurrencyOnStartup, "experimental.tsdb.max-tsdb-opening-concurrency-on-startup", 10, "limit the number of concurrently opening TSDB's on startup")
 	f.DurationVar(&cfg.HeadCompactionInterval, "experimental.tsdb.head-compaction-interval", 1*time.Minute, "How frequently does Cortex try to compact TSDB head. Block is only created if data covers smallest block range. Must be greater than 0 and max 5 minutes.")
 	f.IntVar(&cfg.HeadCompactionConcurrency, "experimental.tsdb.head-compaction-concurrency", 5, "Maximum number of tenants concurrently compacting TSDB head into a new block")
+	f.IntVar(&cfg.StripeSize, "experimental.tsdb.stripe-size", 14, "Power of 2 to use for the tsdb stripe size. Reducing this will decrease memory footprint, but can negatively impact performance.")
```
pracucci: You mention it's a power of 2 and the default is 14, but then when you configure the `tsdb.Options` you're not actually applying the "power of 2" so the effective size is 14 and not 2^14.

pstibrany: Oh, I believe you're right. I have assumed it will be used as a power of 2 internally, but we actually need to do it ourselves (```StripeSize:        1 << i.cfg.TSDBConfig.StripeSize,```).

thorfour: Lol you're right. I shouldn't have tried to make this PR, too tired :joy: 

Added the shift, into the open options.

pracucci:
> LGTM but I think asking people to set the whole thing is better than asking to pass the power of 2. Curious what others thing.

I thought the same and I would tend to agree. The only thing is that the value must be a power of 2, but we can validate it in `Validate()`. Let's do this change @thorfour and then merge it.

pkg/storage/tsdb/config.go
codesome: Maybe we can add a sentence saying it represents number of shards of series in TSDB?

pkg/ingester/ingester_v2.go
```diff
@@ -607,6 +607,7 @@ func (i *Ingester) createTSDB(userID string) (*userTSDB, error) {
 		RetentionDuration: uint64(i.cfg.TSDBConfig.Retention / time.Millisecond),
 		BlockRanges:       i.cfg.TSDBConfig.BlockRanges.ToMilliseconds(),
 		NoLockfile:        true,
+		StripeSize:        int(1 << uint(i.cfg.TSDBConfig.StripeSize)),
```
codesome: In addition to validation that Marco suggested, should we also check for `<=0` and either error out or use the default `1 << 14` in that case?

pkg/storage/tsdb/config.go
thorfour: Done.

pkg/ingester/ingester_v2.go
thorfour: Done.

thorfour:
Added validation that the given value is a power of 2. Thanks for the feedback!

PR #2186, opened 2020-02-25
Generate blocks storage config file doc

pracucci:
**What this PR does**:
Currently, the blocks storage config documented in `docs/operations/blocks-storage.md` was manually generated. This is error prone and drifted a bit.

In this PR I'm introducing the templating support to the doc generator tool, switching the config file reference specified in `docs/operations/blocks-storage.md` to the auto-generated one as well.

Notes:
- Blocks storage config has been added to the config file reference doc too
- The changes to the doc generator tool have been made to try to keep the changeset small, but some refactoring would be probably required at some point

_Please enable "hide whitespace changes" when reviewing this._

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -105,12 +105,12 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 		cfg.BlockRanges = []time.Duration{2 * time.Hour} // Default 2h block
 	}
 
-	f.StringVar(&cfg.Dir, "experimental.tsdb.dir", "tsdb", "directory to place all TSDB's into")
-	f.Var(&cfg.BlockRanges, "experimental.tsdb.block-ranges-period", "comma separated list of TSDB block ranges in time.Duration format")
-	f.DurationVar(&cfg.Retention, "experimental.tsdb.retention-period", 6*time.Hour, "TSDB block retention")
+	f.StringVar(&cfg.Dir, "experimental.tsdb.dir", "tsdb", "Local directory to store TSDBs in the ingesters.")
+	f.Var(&cfg.BlockRanges, "experimental.tsdb.block-ranges-period", "TSDB blocks range period.")
+	f.DurationVar(&cfg.Retention, "experimental.tsdb.retention-period", 6*time.Hour, "TSDB blocks retention in the ingester before a block is removed. This should be larger than the block_ranges_period and large enough to give ingesters enough time to discover newly uploaded blocks.")
 	f.DurationVar(&cfg.ShipInterval, "experimental.tsdb.ship-interval", 1*time.Minute, "How frequently the TSDB blocks are scanned and new ones are shipped to the storage. 0 means shipping is disabled.")
 	f.IntVar(&cfg.ShipConcurrency, "experimental.tsdb.ship-concurrency", 10, "Maximum number of tenants concurrently shipping blocks to the storage.")
-	f.StringVar(&cfg.Backend, "experimental.tsdb.backend", "s3", "TSDB storage backend to use")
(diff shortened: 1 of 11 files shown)


pkg/storage/tsdb/config.go
```diff
@@ -152,12 +152,12 @@ type BucketStoreConfig struct {
 
 // RegisterFlags registers the BucketStore flags
 func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
-	f.StringVar(&cfg.SyncDir, "experimental.tsdb.bucket-store.sync-dir", "tsdb-sync", "Directory to place synced tsdb indicies.")
+	f.StringVar(&cfg.SyncDir, "experimental.tsdb.bucket-store.sync-dir", "tsdb-sync", "Directory to store synched TSDB index headers.")
```
pstibrany: Can we just use `synchronized` or `fetched`? I don't remember seeing `synched` before, but apparently it's a common variant.

pkg/storage/tsdb/config.go
```diff
@@ -152,12 +152,12 @@ type BucketStoreConfig struct {
...
 	f.DurationVar(&cfg.SyncInterval, "experimental.tsdb.bucket-store.sync-interval", 5*time.Minute, "How frequently scan the bucket to look for changes (new blocks shipped by ingesters and blocks removed by retention or compaction). 0 disables it.")
-	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(250*units.Mebibyte), "Size of index cache in bytes per tenant.")
-	f.Uint64Var(&cfg.MaxChunkPoolBytes, "experimental.tsdb.bucket-store.max-chunk-pool-bytes", uint64(2*units.Gibibyte), "Max size of chunk pool in bytes per tenant.")
-	f.Uint64Var(&cfg.MaxSampleCount, "experimental.tsdb.bucket-store.max-sample-count", 0, "Max number of samples (0 is no limit) per query when loading series from storage.")
-	f.IntVar(&cfg.MaxConcurrent, "experimental.tsdb.bucket-store.max-concurrent", 20, "Max number of concurrent queries to the storage per tenant.")
+	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(250*units.Mebibyte), "Size - in bytes - of a per-tenant in-memory index cache used to speed up blocks index lookups.")
```
pstibrany: Would it make sense to use megabytes as a unit here?

```suggestion
	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(250*units.Mebibyte), "Size of a per-tenant in-memory index cache used to speed up blocks index lookups (bytes).")
```

Same below. I think it reads better, but it's not a strong preference.

pkg/storage/tsdb/config.go
pracucci: Ok. Picked `synchronized` (cause it's more then just fetching, given it also offload removed blocks).

pkg/storage/tsdb/config.go
pracucci: Yes, but I would like to use a marshallable unit like I did in Thanos ([`units.go`](https://github.com/thanos-io/thanos/blob/master/pkg/store/cache/units.go)). What do you think if we keep it for a separate PR?

pstibrany: > Yes, but I would like to use a marshallable unit like I did in Thanos ([`units.go`](https://github.com/thanos-io/thanos/blob/master/pkg/store/cache/units.go)). What do you think if we keep it for a separate PR?

Sure, this PR is not about changing units.

PR #2187, opened 2020-02-26
Change config flags, add flags to disable Alertmanager notifications methods: email & webhoook

Wing924:
**What this PR does**:

Added flags to disable Alertmanager notifications methods.

* [CHANGE] Renamed Configs configuration options.
  * configuration options
    * `-database.*` -> `-configs.database.*`
    * `-database.migrations` -> `-configs.database.migrations-dir`
  * config file
    * `configdb.uri:` -> `configs.database.uri:`
    * `configdb.migrationsdir:` -> `configs.database.migrations_dir:`
    * `configdb.passwordfile:` -> `configs.database.password_file:`

The Alertmanager notifications via email is disabled in cortex.
I tested on my cluster and found email works without any problems.

I asked on Slack and found the reason is email may attack SaaS.
I other words, we can enable email on on-premises.

**Checklist**
- [x] Tests updated
- [x] Documentation added


bboreham:
So many flags seems like overkill to me.
I could see an argument to disable (anything with an email address) and (anything with a URL); I don't really see the case to selectively disable Wechat over OpsGenie.

Wing924:
you’re right. we don’t need so many flags.
what flags should we keep?

docs/configuration/config-file-reference.md
```diff
@@ -108,6 +108,10 @@ Where default_value is the value to use if the environment variable is undefined
 # The ruler_config configures the Cortex ruler.
 [ruler: <ruler_config>]
 
+# The configs_api_config configures the configs API used by the 'configs'
+# service to expose APIs to manage them.
+[configs_api: <configs_api_config>]
```
pracucci: From the configuration perspective, I'm a bit concerned having another root block config. Ideally we should have 1 root config block for each service (even if in the past some weird things have been done), so think it would be better adding the `NotificationsConfig` as part of the `ConfigDB` config.

Wing924:
@bboreham 
I reduced the options to only email and webhook.

@pracucci 
Thank you for your advice.
I changed the configurations structure.

pkg/configs/api/api.go
```diff
@@ -164,15 +188,18 @@ func (a *API) validateAlertmanagerConfig(w http.ResponseWriter, r *http.Request)
...
 	amCfg, err := amconfig.Load(cfg)
 	if err != nil {
 		return err
 	}
 
 	for _, recv := range amCfg.Receivers {
-		if len(recv.EmailConfigs) != 0 {
-			return fmt.Errorf("email notifications are not supported in Cortex yet")
+		if noCfg.DisableEmail && len(recv.EmailConfigs) > 0 {
+			return fmt.Errorf("email notifications are disabled in Cortex yet")
```
gouthamve: The `in Cortex yet` can be removed now.

pkg/configs/api/api.go
```diff
@@ -164,15 +188,18 @@ func (a *API) validateAlertmanagerConfig(w http.ResponseWriter, r *http.Request)
...
+		}
+		if noCfg.DisableWebHook && len(recv.WebhookConfigs) > 0 {
+			return fmt.Errorf("web-hook notifications are disabled in Cortex yet")
```
gouthamve: `yet` can be removed. It can just be: `web-hook notifications are disabled`

docs/configuration/config-file-reference.md
```diff
@@ -2058,22 +2057,32 @@ The `fifo_cache_config` configures the local in-memory cache.
...
+  # URI where the database can be found (for dev you can use memory://)
+  # CLI flag: -configs.database.uri
+  [uri: <string> | default = "postgres://postgres@configs-db.weave.local/configs?sslmode=disable"]
 
-# Path where the database migration files can be found
-# CLI flag: -database.migrations
-[migrationsdir: <string> | default = ""]
+  # Path where the database migration files can be found
+  # CLI flag: -configs.database.migrations
+  [migrations_dir: <string> | default = ""]
```
pracucci: Please mention this breaking change in the CHANGELOG.

docs/configuration/config-file-reference.md
```diff
@@ -2058,22 +2057,32 @@ The `fifo_cache_config` configures the local in-memory cache.
...
 
-# File containing password (username goes in URI)
-# CLI flag: -database.password-file
-[passwordfile: <string> | default = ""]
+  # File containing password (username goes in URI)
+  # CLI flag: -configs.database.password-file
+  [password_file: <string> | default = ""]
```
pracucci: Please mention this breaking change in the CHANGELOG.

pkg/configs/api/api.go
```diff
@@ -19,15 +20,38 @@ import (
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
+// Config configures Configs API
+type Config struct {
```
pracucci: Is this the right place where it should stay? It both allow to configure DB and API, so given it's the root block I would see it more correctly places in `pkg/configs`. The clashing `type Config` in `pkg/configs` could be more correctly renamed `UserConfig`, given it contains the configuration submitted by a single user? What's your take?

pkg/configs/api/api.go
pracucci: `web-hook` > `webhook`

Wing924:
@pracucci 
I fixed it as you described.
I renamed `configs.Config` to `userconfig.Config` due to circular dependency.
Please check it again.

CHANGELOG.md
```diff
@@ -25,14 +25,24 @@
...
 * [CHANGE] Experimental TSDB: TSDB head compaction interval and concurrency is now configurable (defaults to 1 min interval and 5 concurrent head compactions). New options: `-experimental.tsdb.head-compaction-interval` and `-experimental.tsdb.head-compaction-concurrency`. #2172
 * [CHANGE] Remove fluentd-based billing infrastructure and flags such as `-distributor.enable-billing`. #1491
+* [CHNAGE] Renamed Configs configuration options. #2187
+  * configuration options
+    * `-database.*` -> `-configs.database.*`
+  * config file
+    * `configdb:` -> `configs:`
+    * `configdb.migrationsdir:` -> `configs.migrations_dir:`
+    * `configdb.passwordfile:` -> `configs.password_file:`
```
```suggestion
    * `configdb.uri:` -> `configs.database.uri:`
    * `configdb.migrationsdir:` -> `configs.database.migrations_dir:`
    * `configdb.passwordfile:` -> `configs.database.password_file:`
```

pkg/configs/db/db.go
```diff
@@ -7,50 +7,50 @@ import (
...
 	Mock DB `yaml:"-"`
 }
 
 // RegisterFlags adds the flags required to configure this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
-	f.StringVar(&cfg.URI, "database.uri", "postgres://postgres@configs-db.weave.local/configs?sslmode=disable", "URI where the database can be found (for dev you can use memory://)")
-	f.StringVar(&cfg.MigrationsDir, "database.migrations", "", "Path where the database migration files can be found")
-	f.StringVar(&cfg.PasswordFile, "database.password-file", "", "File containing password (username goes in URI)")
+	f.StringVar(&cfg.URI, "configs.database.uri", "postgres://postgres@configs-db.weave.local/configs?sslmode=disable", "URI where the database can be found (for dev you can use memory://)")
+	f.StringVar(&cfg.MigrationsDir, "configs.database.migrations", "", "Path where the database migration files can be found")
```
pracucci: Given we're introducing breaking changes anyway, I would suggest to rename this flag to `configs.database.migrations-dir` to consistency.

pkg/configs/api/api_test.go
```diff
@@ -244,7 +245,20 @@ var amCfgValidationTests = []struct {
           email_configs:
           - to: myteam@foobar.org`,
 		shouldFail:  true,
-		errContains: "email notifications are not supported in Cortex yet",
+		errContains: "email notifications are disabled",
```
pracucci: Would it be possible to add a test covering the case email notifications are enabled?

pkg/configs/api/api.go
```diff
@@ -164,27 +186,30 @@ func (a *API) validateAlertmanagerConfig(w http.ResponseWriter, r *http.Request)
...
 	amCfg, err := amconfig.Load(cfg)
 	if err != nil {
 		return err
 	}
 
 	for _, recv := range amCfg.Receivers {
-		if len(recv.EmailConfigs) != 0 {
-			return fmt.Errorf("email notifications are not supported in Cortex yet")
+		if noCfg.DisableEmail && len(recv.EmailConfigs) > 0 {
+			return fmt.Errorf("email notifications are disabled")
```
pracucci: If you define a global var for this err and the next one, then you could reference that error in tests (which makes tests more robusts).

pkg/configs/api/api_test.go
```diff
@@ -244,11 +245,27 @@ var amCfgValidationTests = []struct {
...
```
pracucci: Now you can reference `ErrEmailNotificationsAreDisabled`

go.sum
```diff
@@ -983,6 +983,7 @@ golang.org/x/tools v0.0.0-20191111182352-50fa39b762bc/go.mod h1:b+2E5dAYhXwXZwtn
 golang.org/x/tools v0.0.0-20191113191852-77e3bb0ad9e7/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191115202509-3a792d9c32b2 h1:EtTFh6h4SAKemS+CURDMTDIANuduG5zKEXShyy18bGA=
 golang.org/x/tools v0.0.0-20191115202509-3a792d9c32b2/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
+golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7 h1:9zdDQZ7Thm29KFXgAX/+yaf3eVbP7djjWp/dXAppNCc=
```
pracucci: Please cleanup this.

Wing924: I don't know why but after I ran `go mod tidy`, that line is appeared.

Wing924: I've fixed it.

bboreham:
Please edit the PR title and description to cover all that it now changes.

PR #2189, opened 2020-02-27
Shared in-memory index cache for queriers with blocks storage

### LINKED ISSUE #2069: Shift from a per-tenant to a per-querier in-memory index cache in the blocks storage
In the blocks storage implementation, we currently instance a new Thanos in-memory index cache for each tenant:
https://github.com/cortexproject/cortex/blob/master/pkg/querier/block_store.go#L355-L358

This causes the in-memory index cache max size to be unpredictable, given each tenant index cache has a max size, but the number of tenants may grow over the time.

We should shift from a per-tenant to a per-querier in-memory index cache. The only downside is that we do loose the strong guarantee that caches are isolated by tenant, because the cache key is based on the block ID and, theoretically, we may have two tenants with a clashing block ID. For this reason, we should:

1. Investigate how the block ID generation entropy works, to better understand how likely is we get clashing IDs (two blocks with a clashing ID must be generated the same exact nanosecond and have the same random generated value, so we should verify how the entropy is implemented)
2. If it's unrealistic having clashing IDs, we should go on with a single instance of the in-memory index cache, otherwise we should open an issue in Thanos and investigate the feasibility to add multi-tenancy support to the index cache

pracucci:
**What this PR does**:
In this PR I propose to shift to a shared in-memory index cache for queriers with blocks storage. The problem with per-tenant caches is that the total max cache size linearly increase with the number of tenants, while with a single cache it's easier to keep the max memory used under control.

Fixes #2069

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -154,7 +154,7 @@ type BucketStoreConfig struct {
 func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.SyncDir, "experimental.tsdb.bucket-store.sync-dir", "tsdb-sync", "Directory to store synchronized TSDB index headers.")
 	f.DurationVar(&cfg.SyncInterval, "experimental.tsdb.bucket-store.sync-interval", 5*time.Minute, "How frequently scan the bucket to look for changes (new blocks shipped by ingesters and blocks removed by retention or compaction). 0 disables it.")
-	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(250*units.Mebibyte), "Size - in bytes - of a per-tenant in-memory index cache used to speed up blocks index lookups.")
+	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(1*units.Gibibyte), "Size in bytes of an in-memory index cache used to speed up blocks index lookups (shared across multiple tenants).")
 	f.Uint64Var(&cfg.MaxChunkPoolBytes, "experimental.tsdb.bucket-store.max-chunk-pool-bytes", uint64(2*units.Gibibyte), "Max size - in bytes - of a per-tenant chunk pool, used to reduce memory allocations.")
 	f.Uint64Var(&cfg.MaxSampleCount, "experimental.tsdb.bucket-store.max-sample-count", 0, "Max number of samples per query when loading series from the long-term storage. 0 disables the limit.")
 	f.IntVar(&cfg.MaxConcurrent, "experimental.tsdb.bucket-store.max-concurrent", 20, "Max number of concurrent queries to execute against the long-term storage on a per-tenant basis.")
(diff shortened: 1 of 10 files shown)


pracucci:
About this TODO:

```
// TODO: apparently Thanos has a bug which cause a block to not be considered if the
//       query timetamp matches the block max timestamp
```

This issue is fixed in Thanos `master` but will upgrade it in a separate PR.

pracucci:
About this TODO:

```
// TODO: apparently Thanos has a bug which cause a block to not be considered if the
//       query timetamp matches the block max timestamp
```

This issue is fixed in Thanos `master` but will upgrade it in a separate PR.

pkg/storage/tsdb/config.go
```diff
@@ -154,7 +154,7 @@ type BucketStoreConfig struct {
 func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.SyncDir, "experimental.tsdb.bucket-store.sync-dir", "tsdb-sync", "Directory to store synchronized TSDB index headers.")
 	f.DurationVar(&cfg.SyncInterval, "experimental.tsdb.bucket-store.sync-interval", 5*time.Minute, "How frequently scan the bucket to look for changes (new blocks shipped by ingesters and blocks removed by retention or compaction). 0 disables it.")
-	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(250*units.Mebibyte), "Size - in bytes - of a per-tenant in-memory index cache used to speed up blocks index lookups.")
+	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(1*units.Gibibyte), "Size in bytes of an in-memory index cache used to speed up blocks index lookups (shared across multiple tenants).")
```
pstibrany:
```suggestion
	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(1*units.Gibibyte), "Size in bytes of in-memory index cache used to speed up blocks index lookups (shared across multiple tenants).")
```

I'd remove `an`

pkg/storage/tsdb/config.go
```diff
@@ -154,7 +154,7 @@ type BucketStoreConfig struct {
 func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.SyncDir, "experimental.tsdb.bucket-store.sync-dir", "tsdb-sync", "Directory to store synchronized TSDB index headers.")
 	f.DurationVar(&cfg.SyncInterval, "experimental.tsdb.bucket-store.sync-interval", 5*time.Minute, "How frequently scan the bucket to look for changes (new blocks shipped by ingesters and blocks removed by retention or compaction). 0 disables it.")
-	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(250*units.Mebibyte), "Size - in bytes - of a per-tenant in-memory index cache used to speed up blocks index lookups.")
+	f.Uint64Var(&cfg.IndexCacheSizeBytes, "experimental.tsdb.bucket-store.index-cache-size-bytes", uint64(1*units.Gibibyte), "Size in bytes of in-memory index cache used to speed up blocks index lookups (shared across multiple tenants).")
```
gouthamve: `shared across all tenants` is better, but only slightly :)

pstibrany: shared between ?

pracucci: I've updated it to "shared between all tenants". I hope to have correctly merged both feedback 😉

pstibrany: Merges can be difficult :)

pracucci:
> Though for me there is a bigger concern that we are not prefixing anything with the tenantID. I know its extremely rare for ULIDs to match across tenants, I'm afraid that if it did, we might be leaking some data.

> Can we add a TODO to make sure that we propagate the tenantID into the cache somehow?

This is a very good point. After some discussion in #2069 we decided to not propagate the  tenantID but make the block ID entropy safer, which is what I did here:
https://github.com/prometheus/prometheus/pull/6867

Feel free share further feedback. I'm willing to re-consider this decision if you think it's risky.

PR #2190, opened 2020-02-27
Pass logger with userID to per-user TSDB components.

pstibrany:
**What this PR does**: this PR adds "org_id" information to logger passed to per-user TSDB components.


pstibrany:
> Thanks @pstibrany, LGTM! May you fix the tests, please? We have some tests where we also asserts on logs 🎉

OK, that was unexpected :) Fixed now.

PR #2196, opened 2020-03-02
Support `lastEvaluation` and `evaluationTime` in `/api/v1/rules` endpoints and make order of groups stable.

Wing924:
**What this PR does**:

Support `lastEvaluation` and `evaluationTime` in `/api/v1/rules` endpoints and make order of groups stable.

Without `lastEvaluation` and `evaluationTime`, `/api/prom/new/rules` show NaN in prometheus UI.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -65,7 +66,7 @@
 * [BUGFIX] The gauge `cortex_overrides_last_reload_successful` is now only exported by components that use a `RuntimeConfigManager`. Previously, for components that do not initialize a `RuntimeConfigManager` (such as the compactor) the gauge was initialized with 0 (indicating error state) and then never updated, resulting in a false-negative permanent error state. #2092
 * [BUGFIX] Fixed WAL metric names, added the `cortex_` prefix.
 * [BUGFIX] Restored histogram `cortex_configs_request_duration_seconds` #2138
-* [BUGFIX] Fix wrong syntax for `url` in config-file-reference. #2148
+* [ENHANCEMENT] Fix wrong syntax for `url` in config-file-reference. #2148
```
jtlisi: This doesn't seem related to this PR?

Wing924: I've fixed it.

pkg/ruler/ruler.go
```diff
@@ -560,7 +566,7 @@ func (r *Ruler) getShardedRules(ctx context.Context, userID string) ([]*rules.Ru
 		return nil, fmt.Errorf("unable to inject user ID into grpc request, %v", err)
 	}
 
-	rgs := []*rules.RuleGroupDesc{}
+	rgs := make([]*rules.RuleGroupDesc, 0, len(rulers.Ingesters))
```
jtlisi: I don't think you need this change. The number of Rulers in the ring does not correlate to the number of rule groups being evaluated.

Wing924: > The number of Rulers in the ring does not correlate to the number of rule groups being evaluated.

Yes, but it should be no more than `len(rulers.Ingesters)`. This fix is for avoid realloc slice.
if the slice is empty (before my fix), the len and cap is `0`. After each append, the len increase 1 but cap will be 0, 2, 2, 4, 4..., which make malloc many times.

gouthamve: Can you add a comment for the same? I'll merge it after!

Wing924: I've added the comment.

PR #2197, opened 2020-03-02
Add comment and check for chunkSeries implementation of SeriesWithChunks

pstibrany:
This trivial PR just adds a comment to `chunkSeries` type about implementing SeriesWithChunks interface. It also adds a check.

### DIFF
--- a/pkg/querier/chunk_store_queryable.go
+++ b/pkg/querier/chunk_store_queryable.go
@@ -87,6 +87,7 @@ func (q *chunkStoreQuerier) Close() error {
 	return nil
 }
 
+// Implements SeriesWithChunks
 type chunkSeries struct {
 	labels            labels.Labels
 	chunks            []chunk.Chunk
@@ -103,7 +104,7 @@ func (s *chunkSeries) Iterator() storage.SeriesIterator {
 	return s.chunkIteratorFunc(s.chunks, model.Time(s.mint), model.Time(s.maxt))
 }
 
-// implements SeriesWithChunks
+// method from SeriesWithChunks interface
 func (s *chunkSeries) Chunks() []chunk.Chunk {
 	return s.chunks
 }
--- a/pkg/querier/chunk_store_queryable.go
+++ b/pkg/querier/chunk_store_queryable.go
@@ -103,6 +103,7 @@ func (s *chunkSeries) Iterator() storage.SeriesIterator {
 	return s.chunkIteratorFunc(s.chunks, model.Time(s.mint), model.Time(s.maxt))
 }
 
+// implements SeriesWithChunks
 func (s *chunkSeries) Chunks() []chunk.Chunk {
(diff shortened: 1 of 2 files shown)


pkg/querier/chunk_store_queryable.go
```diff
@@ -103,6 +103,7 @@ func (s *chunkSeries) Iterator() storage.SeriesIterator {
 	return s.chunkIteratorFunc(s.chunks, model.Time(s.mint), model.Time(s.maxt))
 }
 
+// implements SeriesWithChunks
```
codesome: Should this comment be with the declaration of `chunkSeries`?

pkg/querier/chunk_store_queryable_test.go
```diff
@@ -15,6 +15,9 @@ import (
 	promchunk "github.com/cortexproject/cortex/pkg/chunk/encoding"
 )
 
+// Make sure that chunkSeries implements SeriesWithChunks
```
codesome: If we are using `chunkSeries` as `SeriesWithChunks`, and if it is not compatible, it would be caught during build time itself. Do we need this additional test?

codesome: Btw, never thought it could be tested like this :) TIL

pstibrany: We aren't. We only check if series implements this interface, and if it does, then we use chunk-based code.

https://github.com/cortexproject/cortex/blob/fbe1c63a1f9f42963b93af865c35065f8ba30bde/pkg/querier/querier.go#L267

codesome: Makes sense then

pkg/querier/chunk_store_queryable.go
pstibrany: That is also a possibility. I don't have a strong preference where to put it, as long as it is documented.

codesome: The declaration seems like a natural place to know all about the struct, which in this case the fact that `chunkSeries` implements `SeriesWithChunks`

pstibrany: 👍

pkg/querier/chunk_store_queryable.go
```diff
@@ -103,6 +104,7 @@ func (s *chunkSeries) Iterator() storage.SeriesIterator {
 	return s.chunkIteratorFunc(s.chunks, model.Time(s.mint), model.Time(s.maxt))
 }
 
+// method from SeriesWithChunks interface
```
codesome: Comments for exported methods usually start with the method names :) (just a Go convention)
```suggestion
// Chunks implements SeriesWithChunks interface.
```

pstibrany: :-) Done!

PR #2201, opened 2020-03-03
Documented how to run the website locally

pracucci:
**What this PR does**:
In this PR I've attempted to document how to run the website locally, in order to allow everyone to have a quick feedback loop when writing the documentation.

It's worth to note that I've moved `CONTRIBUTING.md` into a folder but the URL in the website has **not** changed (https://cortexmetrics.io/docs/contributing/).

Preview:
Screen Shot 2020-03-03 at 11 32 12

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/contributing/how-to-run-website-locally.md
+++ b/docs/contributing/how-to-run-website-locally.md
@@ -14,7 +14,7 @@ The following initial setup is required only once:
 
 1. Install [Hugo](https://gohugo.io/) `v0.59.1`
 2. Install [Node.js](https://nodejs.org/en/) (or via [`nvm`](https://github.com/nvm-sh/nvm))
-3. Install required Node modules
+3. Install required Node modules with:
    ```
    cd website && npm install -g postcss-cli autoprefixer
    ```
--- a/docs/contributing/how-to-run-website-locally.md
+++ b/docs/contributing/how-to-run-website-locally.md
@@ -5,7 +5,7 @@ weight: 2
 slug: how-to-run-the-website-locally
 ---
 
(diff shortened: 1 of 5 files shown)


docs/contributing/how-to-run-website-locally.md
```diff
@@ -0,0 +1,43 @@
...
+
+The Cortex documentation is compiled into a website published at [cortexmetrics.io](https://cortexmetrics.io/). This document explains how to run the website locally, in order to have a quick feedback loop while contributing to the documentation or website styling.
+
+
+## Initial setup
+
+The following initial setup is required only once:
+
+1. Install [Hugo](https://gohugo.io/) `v0.59.1`
+2. Install [Node.js](https://nodejs.org/en/) (or via [`nvm`](https://github.com/nvm-sh/nvm))
```
gotjosh: I think it might worth specifying a minimum node version as guidance, I had to switch from system (10.x) to one of the lastest table (12.x) for the instructions to work.

docs/contributing/how-to-run-website-locally.md
```diff
@@ -0,0 +1,43 @@
...
+3. Install required Node modules
+   ```
+   cd website && npm install -g postcss-cli autoprefixer
```
gotjosh: Do these need to be global? I don't understand the full process but I think these should be locally installed to be picked up by the build process (as you might not have access via the global install path). Since you already have a `package.json` on the root of `website` I would suggest using `npm install` as the comand to run. 


The error I received was:

```
Error: Loading PostCSS Plugin failed: Cannot find module 'autoprefixer'
```

Which has you can see means it fails to find the dependency.

pracucci:
Thanks @gotjosh for your review! I should have addressed all comments. May you check it out again, please?

gotjosh:
Looks good to me!

PR #2203, opened 2020-03-03
Introduce query-tee tool

pracucci:
**What this PR does**:
While working on the experimental blocks storage, we started having the need to compare chunks vs blocks storage in the real world. To do it, we setup a test cluster (with blocks storage) ingesting the same exact series of another cluster with chunks storage. We now have two parallel clusters with the same series and we have the need to query both clusters and compare performances and results.

Given this need, the idea we got is to add a new tool called `query-tee` that exposes the Prometheus API read endpoints and, for each request, send the same request to 2+ backends and make them compete.

In this PR:
- Added a new tool `query-tee`
- Track a metric with response time for each backend

What's **out of the scope of this PR**:
- Compare results from backends

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


cmd/query-tee/main.go
```diff
@@ -0,0 +1,61 @@
...
+	BackendReadTimeout time.Duration
+	LogLevel           logging.Level
+}
+
+func main() {
+	// Parse CLI flags.
+	cfg := Config{}
+	flag.IntVar(&cfg.ServerServicePort, "server.service-port", 80, "The port where the query-tee service listens to.")
+	flag.IntVar(&cfg.ServerMetricsPort, "server.metrics-port", 9900, "The port where metrics are exposed.")
+	flag.StringVar(&cfg.BackendEndpoints, "backend.endpoints", "", "Comma separated list of backend endpoints to query.")
```
pstibrany: Nit: flag package supports multiple values for single parameter (`-backend.endpoint <url1> -backend.endpoint <url2>`), I find that nicer than extra splitting on commas.

cmd/query-tee/main.go
```diff
@@ -0,0 +1,61 @@
...
+	PreferredBackend   string
+	BackendReadTimeout time.Duration
+	LogLevel           logging.Level
+}
+
+func main() {
+	// Parse CLI flags.
+	cfg := Config{}
+	flag.IntVar(&cfg.ServerServicePort, "server.service-port", 80, "The port where the query-tee service listens to.")
+	flag.IntVar(&cfg.ServerMetricsPort, "server.metrics-port", 9900, "The port where metrics are exposed.")
```
pstibrany: Why two servers? Main server can map some subtree (`/api/*` to backend endpoints, and keep the rest, including `/metrics` for itself).

cmd/query-tee/proxy.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		// The backend name is hardcoded as the backend hostname.
+		name := u.Hostname()
+
+		p.backends = append(p.backends, NewProxyBackend(name, u, cfg.BackendReadTimeout, name == cfg.PreferredBackend))
+	}
+
+	// At least 2 backends are required
+	if len(p.backends) < 2 {
```
pstibrany: Proxy can also run with single backend, right? I mean, it's pointless, but it's not a hard requirement to have at least two, or is it?

cmd/query-tee/proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return p, nil
+}
+
+func (p *Proxy) Run() error {
+	router := mux.NewRouter()
+
+	// Read endpoints.
+	router.Path("/api/prom/api/v1/query").Methods("GET").Handler(NewProxyEndpoint(p.backends, "api_v1_query", p.metrics, p.logger))
```
pstibrany: This seems very Cortex-specific. Prometheus listens on `/api/v1/query`. Personally I think we should handle some `<prefix>/*`, where `/*` part would be appended to backend URLs, so that backend URL for Prometheus can be `http://prometheus/` and for Cortex it would be `http://cortex/api/prom/`. Query-tee datasource URL in Grafana would be `http://query-tee/prefix`. What do you think?

For route names and result comparison, I would recognize some well-known suffixes (like `/api/v1/query_range` and `/api/v1/label/[^/]+/values`) and ignore the rest.

cmd/query-tee/proxy_endpoint.go
```diff
@@ -0,0 +1,121 @@
...
+		if _, err := w.Write(downstreamRes.body); err != nil {
+			level.Warn(p.logger).Log("msg", "Unable to write response", "err", err)
+		}
+	}
+}
+
+func (p *ProxyEndpoint) pickResponseForDownstream(responses []*backendResponse) *backendResponse {
+	// Look for a successfull response (or 4xx) from the preferred backend.
+	for _, res := range responses {
+		if res.backend.preferred && res.err == nil && res.status < 500 {
```
pstibrany: `< 500` will also match redirect responses (3xx). Comment only talks about `4xx`.

cmd/query-tee/proxy_backend.go
```diff
@@ -0,0 +1,112 @@
...
+		client: &http.Client{
+			CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
+				return errors.New("The query-tee proxy does not follow redirects.")
+			},
+			Transport: &http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout:   30 * time.Second,
+					KeepAlive: 30 * time.Second,
+					DualStack: true,
```
pstibrany: This field is now deprecated and enabled by default.

cmd/query-tee/proxy_backend.go
```diff
@@ -0,0 +1,112 @@
...
+				}).DialContext,
+				MaxIdleConns:          10000,
+				MaxIdleConnsPerHost:   1000, // see https://github.com/golang/go/issues/13801
+				IdleConnTimeout:       90 * time.Second,
+				DisableKeepAlives:     true,
```
pstibrany: Why? On one hand, we enable 10000 idle connections, on the other hand we disable idle connections completely here.

cmd/query-tee/main.go
pracucci: Yes, I did it. Then Grafana jsonnet doesn't support it (args are a map), so I rolled back to comma separated to be able to easily deploy it in our infra.

cmd/query-tee/main.go
pracucci: This tool can be publicly exposed given it's the very entrypoint of the system (in front of any "cortex auth gateway"). For this reason, we may want to expose `/metrics` only privately while the read API endpoint will be public.

cmd/query-tee/proxy.go
pracucci: It can, but I'm missing the use case. If you configured it with less than 2 backends, it's probably a misconfiguration, reason why I'm reporting it as an error. Thoughts?

cmd/query-tee/proxy.go
pracucci: > Query-tee datasource URL in Grafana would be http://query-tee/prefix. What do you think?

I think your comment makes much sense. The `query-tee` now exposes Prometheus compatible APIs (ie. `/api/v1/query`). If the backend URL contains a path, that path is prepended to the original request. What's your take?

cmd/query-tee/proxy_endpoint.go
pracucci: Agree. I've added a `succeeded()` func which returns true only for 2xx and 4xx.

cmd/query-tee/proxy.go
pstibrany: > I think your comment makes much sense. The `query-tee` now exposes Prometheus compatible APIs (ie. `/api/v1/query`). If the backend URL contains a path, that path is prepended to the original request. What's your take?

This looks good. I'm little concerned about joining paths like that – I think it can lead to various anomalies [or not, haven't tested it] – but since this is a test tool, this will do.

cmd/query-tee/proxy_backend.go
pracucci: Agree. It was non-sense. Was blindly copied from a "gateway" 🤦‍♂ . I've kept keep-alive enabled (default), which I believe for this use case makes sense.

cmd/query-tee/proxy.go
pracucci: I've now used `path.Join()`. Better safe than sorry.

cmd/query-tee/proxy.go
pstibrany: Possible use case, starting with one backend, making sure it works, then adding another one? Agree it's not a great one, I just feel that warning here would be enough.

pracucci: Ok. Don't have strong opinions. Lowered the requirement to 1 and warning of it's just 1.

cmd/query-tee/proxy.go
pstibrany: > I've now used `path.Join()`. Better safe than sorry.

👍 Although I think it would still be possible to smuggle `../../../../../metrics` into request path :)

cmd/query-tee/main.go
pstibrany: Makes sense, thanks for explanation.

cmd/query-tee/main.go
pstibrany: Right. We could also combine the two approaches :-) But let's not bother with that.

pracucci:
Thanks for your review @pstibrany!

> I haven't noticed any test for different backend path prefixes (say one backend using /prefix1/api/v1/query and other one different prefix), would it be worth having one?

Makes sense. Done!

PR #2207, opened 2020-03-04
Update the default configs for spread flush and bigchunks

### LINKED ISSUE #2202: RFC: Make BigChunk the default encoding
WDYT?

https://github.com/cortexproject/cortex/blob/master/pkg/chunk/encoding/factory.go#L18

tomwilkie:
**What this PR does**:
- Make spread flush the default using the example single binary config (not the flag)
- Make bigchunk the default encoding in the code

Fixes #2202 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,19 +2,9 @@
 
 ## master / unreleased
 
-* [FEATURE] Fan out parallelizable queries to backend queriers concurrently. #1878
-  * `querier.parallelise-shardable-queries` (bool)
-  * Requires a shard-compatible schema (v10+)
-  * This causes the number of traces to increase accordingly.
-  * The query-frontend now requires a schema config to determine how/when to shard queries, either from a file or from flags (i.e. by the `config-yaml` CLI flag). This is the same schema config the queriers consume. The schema is only required to use this option.
-  * It's also advised to increase downstream concurrency controls as well:
-    * `querier.max-outstanding-requests-per-tenant`
-    * `querier.max-query-parallelism`
-    * `querier.max-concurrent`
-    * `server.grpc-max-concurrent-streams` (for both query-frontends and queriers)
 * [CHANGE] The frontend http server will now send 502 in case of deadline exceeded and 499 if the user requested cancellation. #2156
 * [CHANGE] Config file changed to remove top level `config_store` field in favor of a nested `configdb` field. #2125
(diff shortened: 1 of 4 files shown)


pkg/chunk/encoding/factory.go
```diff
@@ -15,7 +15,7 @@ type Config struct{}
 
 var (
 	// DefaultEncoding exported for use in unit tests elsewhere
-	DefaultEncoding             = DoubleDelta
+	DefaultEncoding             = Bigchunk
```
gouthamve: This requires calling out in the changelog. Maybe that the storage size would increase slightly if there are lots of series with constant value

gouthamve:
Hrm, tests are failing :(

tomwilkie:
> Hrm, tests are failing :(

Will fix this up at the weekend.

CHANGELOG.md
```diff
@@ -29,6 +19,7 @@
 * [CHANGE] Experimental TSDB: the querier in-memory index cache used by the experimental blocks storage shifted from per-tenant to per-querier. The `-experimental.tsdb.bucket-store.index-cache-size-bytes` now configures the per-querier index cache max size instead of a per-tenant cache and its default has been increased to 1GB. #2189
 * [CHANGE] If you are vendoring Cortex and use its components in your project, be aware that many Cortex components no longer start automatically when they are created. You may want to review PR and attached document. #2166
 * [CHANGE] Cortex now has /ready probe for all services, not just ingester and querier as before. In single-binary mode, /ready reports 204 only if all components are running properly. #2166
+* [CHANGE] Default to BigChunk encoding; may results in slightly high disk usage if many of you series have a constant value, but should generally result in fewer, bigger chunks. #2207
```
bboreham:
```suggestion
* [CHANGE] Default to BigChunk encoding; may result in slightly higher disk usage if many timeseries have a constant value, but should generally result in fewer, bigger chunks. #2207
```

pracucci:
_Rebased `master` to fix conflicts_

pracucci:
In tests we do compare the chunk object. Switching to Bigchunk the comparison is not clean because of the appender pointer. In the commit https://github.com/cortexproject/cortex/pull/2207/commits/44ada476fc4c633528a24e0b9019fb46eb287488 I've used a trick to cleanup the expected chunk (the clean chunk has no appender internally). I'm not much happy with the trick I've done, but haven't come up with a better idea.

Issue #2204, opened 2020-03-03
Enforce all metrics to be registered to a registry


pracucci:
While doing code reviews I've realized a common mistake is to create metrics but not register them to the registry. Functions like `promauto.NewXXX()` are now discouraged due to #1512 and this makes even easier to forget.

In order to overcome this problem, Prometheus client `1.5.0` has introduced a new factory `promauth.With(registerer).NewXXX()` ([713](https://github.com/prometheus/client_golang/pull/713)) which allows to use `promauto` with a custom registerer (thus honoring the design we picked in #1512).

The proposal to enforce all metrics to be registered is:
1. Upgrade Prometheus client to `1.5.0`
2. Switch all metrics to `promauto.With().NewXXX()` taking the opportunity to pass the `registerer` everywhere so we will also solve #1512 
3. Add a linter rule to forbid the import of `github.com/prometheus/client_golang/prometheus` (except few cases where it's OK, like the metrics mapping we do for the blocks storage and alertmanager)

pracucci:
_It's a very low priority on-going work I continue every while and then._

pracucci:
_Help wanted!_

roidelapluie:
I will take this one.

pracucci:
_Still valid, even if over the time we've significantly improved it_

pracucci:
_Still valid_

PR #2210, opened 2020-03-04
No globally registered metrics in pkg/querier

pracucci:
**What this PR does**:
As next step to solve #2204 and #1512 I've iterated over all metrics registered in `pkg/querier` and refactored the code to:

1. Pick `prometheus.Registerer` in input in each component
2. Register metrics using `promauto.With()`

I've also introduced a new integration test to make sure that now `cortex_frontend_*` and `cortex_query_frontend_*` metrics are exposed only by the query-frontend service (same test will be enriched in subsequent PRs once more packages will be refactored).

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


integration/metrics_test.go
```diff
@@ -0,0 +1,110 @@
+package main
```
pstibrany: Missing integration build tag.

pracucci: Right, done.

PR #2213, opened 2020-03-05
add metrics for number of failures in creation and deletion of tables

sandeepsukhani:
**What this PR does**:
Expose metrics with count of the number of errors in the creation and deletion of tables.
It is the same code from PR #1409 which got closed due to accidental deletion of its branch.


pstibrany:
Oh guys, you're too fast.

PR #2214, opened 2020-03-05
query time filtering of chunks for pending delete requests

sandeepsukhani:
**What this PR does**:
Add support for query time filtering when there are pending delete requests.
Delete requests with status `Received`, `BuildingPlan`, `Deleting` are considered to be pending.
A new entity is added called `TombstonesLoader` which keeps delete requests in memory and keeps checking for updates periodically. It helps with doing query time filtering of data.

This is a smaller part of a bigger PR #1906.

I have added some basic tests. Will add more tests and a benchmark test for filtering of chunks with and without tombstones once someone does an initial review of the code.

**Checklist**
- [ ] Tests updated


pkg/chunk/encoding/chunk.go
```diff
@@ -246,3 +280,114 @@ func (it *indexAccessingChunkIterator) Batch(size int) Batch {
...
+	return &deletedChunkIterator{
+		itr:                       itr,
+		deletedIntervals:          deletedIntervals,
+		nextDeletedIntervalsIndex: 0,
+	}
+}
+
+// Scan checks if we ran into next deleted interval, if yes then it jumps ahead and fetches data right after deleted interval that we ran into.
+func (dci *deletedChunkIterator) Scan() bool {
+	if dci.nextDeletedIntervalsIndex == -1 {
```
gouthamve: Instead of tracking the index, we can reduce the array size everytime we cross an interval. I like this logic better: https://github.com/prometheus/prometheus/blob/52025bd7a9446c3178bf01dd2949d4874dd45f24/tsdb/querier.go#L1190-L1208

gouthamve: Can we simplify the logic by reducing the array size by one everytime?

pkg/chunk/encoding/chunk.go
```diff
@@ -101,6 +101,40 @@ type Batch struct {
 	Length     int
 }
 
+func (b *Batch) dropValuesForInterval(interval model.Interval) {
+	if b.Length == 0 || model.Time(b.Timestamps[0]) > interval.End || model.Time(b.Timestamps[b.Length-1]) < interval.Start {
+		return
+	}
+
+	removedIdxStart := 0
```
gouthamve: I am not sure being super performant is important here. Can we simplify this by making a new array and iterating through the samples?

like 
```
reducedTs := [BatchSize]int64
reducedVals := [BatchSize]float64
length := 0

for i := range b.Timestamps {
  if inbound(model.Time(b.Timestamps[i], interval) {
    continue
  }
  reducedTs[length] = b.Timestamps[i]
  reducedValues[length] = b.Values[i]
}

b.Timestamps = reducedTs
b.Values = reducedVals
b.Length = length
```

pkg/ingester/series.go
```diff
@@ -158,7 +158,7 @@ func (s *memorySeries) head() *desc {
 	return s.chunkDescs[len(s.chunkDescs)-1]
 }
 
-func (s *memorySeries) samplesForRange(from, through model.Time) ([]model.SamplePair, error) {
+func (s *memorySeries) samplesForRange(from, through model.Time, deletedInterval []model.Interval) ([]model.SamplePair, error) {
```
gouthamve: Why do we have any changes in the ingester? As far as I understand, the tombstones are loaded in the querier only.

gouthamve: Like can we simplify this by only have the tombstones in the querier layer? I don't expect to see any deletes on the recent 6hrs of data and if any happens, we can filter it out in the querier. We shouldn't ideally change the proto files for this.

pkg/chunk/inmemory_storage_client.go
```diff
@@ -442,3 +444,19 @@ func (b *mockReadBatchIter) RangeValue() []byte {
 func (b *mockReadBatchIter) Value() []byte {
 	return b.items[b.index].value
 }
+
+func SetupTestDeleteStore() (*DeleteStore, error) {
```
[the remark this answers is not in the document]
sandeepsukhani: It is being used for tests in purger package. Moved it to another file called `testutils.go`.

pkg/chunk/purger/tombstones.go
```diff
@@ -0,0 +1,293 @@
...
+			return
+		}
+	}
+}
+
+func (tl *TombstonesLoader) reloadTombstones() error {
+	// check for updates in loaded gen numbers
+	tl.tombstonesMtx.Lock()
+
+	userIDs := make([]string, 0, len(tl.tombstones))
```
gouthamve: From what I can see, we're only reloading or loading tombstones for userIDs that are already there. How are we getting new userIDs in, in the first place?

sandeepsukhani: Consumers always call [GetPendingTombstones](https://github.com/cortexproject/cortex/pull/2214/files/4b75b8e685a90acdf52b7ace02e91ac288bfbbd5#diff-c0f1a864b800713f20d644cfca5d25c0R89) which checks whether it has loaded delete requests already, if not it calls [loadPendingTombstones](https://github.com/cortexproject/cortex/pull/2214/files/4b75b8e685a90acdf52b7ace02e91ac288bfbbd5#diff-c0f1a864b800713f20d644cfca5d25c0R99) which loads it.
Does it make sense?

pkg/cortex/modules.go
```diff
@@ -568,7 +597,7 @@ var modules = map[moduleName]module{
 	},
 
 	StoreQueryable: {
-		deps:           []moduleName{Store},
+		deps:           []moduleName{Store, DeleteRequestsStore},
```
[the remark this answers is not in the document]
sandeepsukhani: `Ruler` also creates a `Querier` which does not have `Querier` in dependencies. Don't we need it there? See https://github.com/cortexproject/cortex/pull/2214/files#diff-596d5408728e4813da55bfbc5a171d3dR432

PR #2216, opened 2020-03-05
fix broken tests in wal

sandeepsukhani:
**What this PR does**:
#2105 got merged recently which was using `Shutdown` method from `Ingester` in tests. The `Shutdown` method was removed recently by Services PR merged before that. This PR fixes the broken tests.

**Checklist**
- [X] Tests updated

### DIFF
--- a/pkg/ingester/wal_test.go
+++ b/pkg/ingester/wal_test.go
@@ -92,7 +92,7 @@ func TestCheckpointRepair(t *testing.T) {
 			userIDs, _ = pushTestSamples(t, ing, numSeries, numSamplesPerSeriesPerPush, (i+1)*numSamplesPerSeriesPerPush)
 			if i == numCheckpoints-1 {
 				// Shutdown creates a checkpoint. This is only for the last checkpoint.
-				ing.Shutdown()
+				require.NoError(t, services.StopAndAwaitTerminated(context.Background(), ing))
 			} else {
 				require.NoError(t, w.performCheckpoint())
 			}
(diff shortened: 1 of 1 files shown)


pkg/ingester/wal_test.go
```diff
@@ -136,7 +136,9 @@ func TestCheckpointRepair(t *testing.T) {
 		w, ok = ing.wal.(*walWrapper)
 		require.True(t, ok)
 		// defer in case we hit an error though we explicitly close it later.
-		defer ing.Shutdown()
+		defer func() {
+			require.NoError(t, services.StopAndAwaitTerminated(context.Background(), ing))
+		}()
```
pstibrany:
```suggestion
		defer services.StopAndAwaitTerminated(context.Background(), ing)
```

I think this would suffice here, since we already check for NoError later in the test. This can only return error if test didn't get that far, and failed earlier.

sandeepsukhani: `StopAndAwaitTerminated` returns an error which would cause linter to fail. To make linter happy we would need a function there to throw away the error or add a `nolint` there. I feel whatever is there now looks more natural compared to `nolint` or throwing away the error.
What do you think?

pracucci: I'm OK keeping the current version. Let's merge it to fix `master`.

pstibrany: Btw, we use `//nolint:errcheck` for these defers all over the tests.

PR #2219, opened 2020-03-05
Move call of stopIncomingRequests before lifecycler shutdown.

pstibrany:
Lifecycler was recently refactored (by #2166), which unfortunately breaks some expectations in Loki. This PR makes fixing Loki possible.

Before #2166, [Lifecycler's `Shutdown` method](https://github.com/cortexproject/cortex/blob/f0fb4b7eb50a60b848b6b685c29b445be15134ac/pkg/ring/lifecycler.go#L330) called `flushTransferer.StopIncomingRequests()` and only then stopped lifecycler's loop.

After #2166, this has been modified. Lifecycler is now stopped via `StopAsync` method (through embedded BasicService), which cancels context used by `loop()` method, which makes `loop()` exit. [`stopping()` method](https://github.com/cortexproject/cortex/blob/9c1f44d0900649f9a78de1b796fdd65df76d250e/pkg/ring/lifecycler.go#L428) is called afterwards. This method still calls `flushTransferer.StopIncomingRequests()`, but by this time, `loop` has already finished.

In Cortex, this doesn't make much difference – only Ingester implements `StopIncomingRequests` method with non-empty body, and it simply sets the readonly flag. In Loki, `StopIncomingRequests` blocks until data transfer has finished, but it expects loop to be still running.

Solution to the issue is removing `StopIncomingRequests` from `FlushTransferer` interface – logically, it doesn't belong there anyway – and let Ingester call it explicitely *before* stopping the lifecycler. This reverts the method call order back to original.

Why doesn't `StopIncomingRequests` belong to `FlushTransferer`? All other `FlushTransferer` methods are called by lifecycler after main lifecycler loop has exited, during lifecycler's shutdown process. `StopIncomingRequests` was the only method that was called *before* triggering lifecycler's shutdown procedure.

(Fix in Loki is the same – call StopIncomingRequests from Ingester's shutdown method, before stopping the lifecycler. I have patch, incl. Cortex update ready.)

cc @rfratto

### DIFF
--- a/pkg/ring/flush.go
+++ b/pkg/ring/flush.go
@@ -9,8 +9,12 @@ import (
 var ErrTransferDisabled = errors.New("transfers disabled")
 
 // FlushTransferer controls the shutdown of an instance in the ring.
+// Methods on this interface are called when lifecycler is stopping.
+// At that point, it no longer runs the "actor loop", but it keeps updating heartbeat in the ring.
+// Ring entry is in LEAVING state.
+// After calling TransferOut and then Flush, lifecycler stops.
 type FlushTransferer interface {
-	StopIncomingRequests()
+	// StopIncomingRequests()
 	Flush()
 	TransferOut(ctx context.Context) error
 }
@@ -23,9 +27,6 @@ type NoopFlushTransferer struct{}
 func NewNoopFlushTransferer() *NoopFlushTransferer {
 	return &NoopFlushTransferer{}
 }
-
-// StopIncomingRequests is a noop
-func (t *NoopFlushTransferer) StopIncomingRequests() {}
 
 // Flush is a noop
 func (t *NoopFlushTransferer) Flush() {}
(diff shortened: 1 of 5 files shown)


pkg/ring/flush.go
```diff
@@ -9,8 +9,12 @@ import (
 var ErrTransferDisabled = errors.New("transfers disabled")
 
 // FlushTransferer controls the shutdown of an instance in the ring.
+// Methods on this interface are called when lifecycler is stopping.
+// At that point, it no longer runs the "actor loop", but it keeps updating heartbeat in the ring.
+// Ring entry is in LEAVING state.
+// After calling TransferOut and then Flush, lifecycler stops.
 type FlushTransferer interface {
-	StopIncomingRequests()
+	// StopIncomingRequests()
```
codesome: I think this can be removed altogether

pstibrany: Correct. Removed.

PR #2220, opened 2020-03-05
gRPC Storage Service

### LINKED ISSUE #1681: Explore storage plugins
So there are 4 issues open to support new storage types:
1. [Elastic](https://github.com/cortexproject/cortex/issues/1477)
2. [SQL](https://github.com/cortexproject/cortex/issues/1419)
3. [TiKV](https://github.com/cortexproject/cortex/issues/1110)
4. [Azure](https://github.com/cortexproject/cortex/issues/1235)

As we increase usage I can only imagine new requests coming up. Further, we're re-using cortex in Loki a fair bit and are finding that the storage trade-offs in Loki and Cortex are different and I'm seeing some implementation details for Loki leaking into Cortex.

I think we need a good way to come up with a plugin system that lets the new storage codes live outside the cortex tree and I think the first obvious candidate is the [grpc-plugin](https://github.com/hashicorp/go-plugin) system from Hashicorp, that has been used to do similar storage plugins for [Jaeger](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/grpc/README.md).

VineethReddy02:
**What this PR does**:
This PR contains gRPC service client i.e
1. Index Client
2. Storage Client
3. Table Client

This is still in work in progress PR this needs testing with gRPC storage server implementation

Fixes #1681

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,121 @@
+syntax = "proto3";
+
+package grpc;
+
+service grpc {
```
gouthamve: grpc-store

docs/configuration/config-file-reference.md
```diff
@@ -1551,6 +1551,11 @@ delete_store:
   # Name of the table which stores delete requests
   # CLI flag: -deletes.requests-table-name
   [requests_table_name: <string> | default = "delete_requests"]
+
+grpc:
```
gouthamve: grpc-store

pkg/chunk/grpc/index_client.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"context"
+
+	"github.com/pkg/errors"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/chunk/util"
+)
+
+type writeBatch struct {
```
gouthamve: Why do we have this. Why can't we use `WriteBatch` directly?

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,121 @@
...
+    //index-client
+    rpc BatchWrite(WriteBatch) returns (Nothing);
+    rpc QueryPages(IndexQuery) returns (ReadBatch);
```
gouthamve: We shouldn't put the Cassandra limitation of a single query at once into the grpc layer. Should be `rpc QueryPages([]IndexQuery) returns (ReadBatch);`, which should return a streaming response, with each read on the stream returning a `ReadBatch`.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,121 @@
...
+
+message TableName {
+    string tableName = 1;
+}
+
+message WriteBatch {
+    repeated IndexEntry indexEntry = 1;
+}
+
+message ReadBatch {
```
gouthamve: Why is the ReadBatch only returning a single row? We should accept multiple rows in the ReadBatch.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,121 @@
...
+
+package grpc;
+
+service grpc {
+    //index-client
+    rpc BatchWrite(WriteBatch) returns (Nothing);
+    rpc QueryPages(IndexQuery) returns (ReadBatch);
+
+    //storage-client
+    rpc PutChunks(Chunk) returns (Nothing);
```
gouthamve: PutChunks() should accept multiple chunks not a single chunk.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,121 @@
...
+}
+
+message TableDesc {
+    string name                  = 1;
+    bool useOnDemandIOMode       = 2;
+    int64 provisionedRead        = 3;
+    int64 provisionedWrite       = 4;
+    map<string, string> tags     = 5;
+    AutoScalingConfig writeScale = 6;
+    AutoScalingConfig readScale  = 7;
```
gouthamve: All these are dynamo specific, we can remove these.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,121 @@
...
+
+message AutoScalingConfig {
+    bool enabled        = 1;
+    string roleARN      = 2;
+    int64 minCapacity   = 3;
+    int64 maxCapacity   = 4;
+    int64 outCoolDown   = 5;
+    int64 inCoolDown    = 6;
+    double targetValue   = 7;
+}
```
gouthamve: Remove this

VineethReddy02: Removed this.

pkg/chunk/grpc/gRPC.proto
VineethReddy02: Yeah. Removed them.

pkg/chunk/grpc/gRPC.proto
VineethReddy02: Made the necessary changes to do PutChunks multiple chunks in a single RPC call.

pkg/chunk/grpc/gRPC.proto
VineethReddy02: Changes this to stream. So now ReadBatch sends multiple rows in the single RPC call

docs/configuration/config-file-reference.md
VineethReddy02: Renamed it.

pkg/chunk/grpc/gRPC.proto
VineethReddy02: .protoc file doesn't accept **grpc-store** So named it as grpc_store

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,122 @@
+syntax = "proto3";
+
+package grpc;
+
+service grpc {
+    //index-client
+    rpc NewWriteBatch(Nothing) returns (WriteBatch);
```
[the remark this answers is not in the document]
VineethReddy02: Removed it.

pkg/chunk/grpc/index_client.go
VineethReddy02: I see Cassandra is maintaining the writeBatch implicitly within itself. So, I also added writeBatch within grpc client. I see dynamoDBstorageclient maintaining one within itself.

gouthamve: Yes, but we have `grpc.WriteBatch` already. Why are we using another one? Specially when we're just copying this in `BatchWrite()`

pkg/chunk/grpc/gRPC.proto
VineethReddy02: @gouthamve Made required changes to support multiple rows per stream.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,127 @@
...
+message TableName {
+    string tableName = 1;
+}
+
+message WriteBatch {
+    repeated IndexEntry indexEntry = 1;
+}
+
+message ReadBatch {
+   repeated Row rows = 1;
```
VineethReddy02: @gouthamve added as you mentioned.

docs/configuration/config-file-reference.md
```diff
@@ -1574,6 +1574,11 @@ filesystem:
 # CLI flag: -store.index-cache-validity
 [index_cache_validity: <duration> | default = 5m0s]
 
+grpc-store:
+  # Hostname or IP of gRPC instance.
+  # CLI flag: -grpc.server-addresses
```
gouthamve, line 1579: `-grpc-store.server-address`?

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,127 @@
...
+message ChunksData {
+    repeated Chunk chunks = 1;
+}
+
+message ChunksResponse {
+    repeated ChunkResponse chunks = 1;
+}
+
+message ChunkResponse {
+    uint32 checksum = 1;
```
gouthamve, line 35: Why are we returning the checksum here? Further, I checked the Cassandra grpc implementation code and you're just returning the same checksum as the caller?

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,127 @@
...
+package grpc;
+
+service grpc_store {
+    //index-client
+    rpc BatchWrite(WriteBatch) returns (Nothing);
+    rpc QueryPages(IndexQuery) returns (stream ReadBatch);
+
+    //storage-client
+    rpc PutChunks(ChunksData) returns (Nothing);
+    rpc GetChunks(Chunks) returns (ChunksResponse);
```
gouthamve, line 12: This doesn't work yet. This requires that the upstream service has knowledge about the schema (to figure out which table to read from), but that shouldn't be the case. We should send the table info as well.

pkg/chunk/grpc/gRPC.proto
VineethReddy02: @gouthamve I am returning the same checksum. Shall I drop it from the response?

pkg/chunk/grpc/gRPC.proto
VineethReddy02: Yeah fixed it. gRPC server will not be aware of `schemaConfig`. `Tablename `will be sent in `GetChunks `request field.

pkg/chunk/grpc/gRPC.proto
VineethReddy02: Dropped the checksum here.

docs/configuration/config-file-reference.md
VineethReddy02: Changed the flag name as suggested.

pkg/chunk/grpc/gRPC.proto
gouthamve, line 12: Needs to be a streaming response as we can return GBs of data

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,127 @@
...
+
+    //table-client
+    rpc ListTables(Nothing) returns (ListTablesResponse);
+    rpc CreateTable(TableDesc) returns (Nothing);
+    rpc DeleteTable(TableName) returns (Nothing);
+    rpc DescribeTable(TableName) returns (DescribeTableResponse);
+    rpc UpdateTable(UpdateTableRequest) returns (Nothing);
+
+    //used by storage-client & index-client
+    rpc Stop(Nothing) returns (Nothing);
```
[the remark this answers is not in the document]
VineethReddy02: I personally feel let it be in proto. I also added a gRPC server call at gRPC client Stop() invocation.
In server-level am closing the DB connection.

Having this in spec indicates we have added all the calls existing in cortex spec.

pkg/chunk/grpc/index_client.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (s *StorageClient) NewWriteBatch() chunk.WriteBatch {
+	return &WriteBatch{}
+}
+
+func (s *StorageClient) BatchWrite(c context.Context, batch chunk.WriteBatch) error {
+	b := batch.(*WriteBatch)
+	writeBatch := &WriteBatch{}
+	writeBatch.IndexEntry = []*IndexEntry{}
+	for _, entry := range b.IndexEntry {
```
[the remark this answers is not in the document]
VineethReddy02: I missed this removing the copying part of code.

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,121 @@
...
+		receivedChunks, err := streamer.Recv()
+		if err == io.EOF {
+			break
+		}
+		if receivedChunks != nil {
+			for _, chunkResponse := range receivedChunks.Chunks {
+				var c chunk.Chunk
+				if chunkResponse != nil {
+					decodeContext := chunk.NewDecodeContext()
+					// if error ignore as partial results are still useful
```
gouthamve, line 111: We should still return the error.

Can we have
```
err := c.Decode(decodeContext, chunkResponse.Buf)
if err != nil {
  lastErr = err
}
```

And return `lastErr`

We can also reuse `decodeContext` by creating at the top and not have to create it again and again.

pkg/chunk/grpc/index_client.go
```diff
@@ -59,8 +59,10 @@ func (s *StorageClient) query(ctx context.Context, query chunk.IndexQuery, callb
 		if err == io.EOF {
 			break
 		}
-		if !callback(readBatch) {
-			return nil
+		if readBatch != nil {
```
gouthamve, line 62: We should check the error here tbh.

pracucci:
_I've pushed a commit to rebase `master`._

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,111 @@
+syntax = "proto3";
```
pracucci, line 1: I would keep consistency with filenames and rename `gRPC.proto` to the lower case `grpc.proto`. Then remember to change it in `Makefile` too.

pkg/chunk/storage/factory.go
```diff
@@ -57,20 +59,18 @@ type StoreLimits interface {
...
+	Engine                  string                   `yaml:"engine"`
+	AWSStorageConfig        aws.StorageConfig        `yaml:"aws"`
+	AzureStorageConfig      azure.BlobStorageConfig  `yaml:"azure"`
+	GCPStorageConfig        gcp.Config               `yaml:"bigtable"`
+	GCSConfig               gcp.GCSConfig            `yaml:"gcs"`
+	CassandraStorageConfig  cassandra.Config         `yaml:"cassandra"`
+	BoltDBConfig            local.BoltDBConfig       `yaml:"boltdb"`
+	FSConfig                local.FSConfig           `yaml:"filesystem"`
+	IndexCacheValidity      time.Duration            `yaml:"index_cache_validity"`
+	GrpcConfig              grpc.Config              `yaml:"grpc-store"`
```
pracucci, line 71: The convention is `_` in YAML, `-`  in CLI flags:

```suggestion
	GrpcConfig              grpc.Config              `yaml:"grpc_store"`
```

pkg/chunk/grpc/grpc_client.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// Config for a StorageClient
+type Config struct {
+	Address string `yaml:"address,omitempty"`
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.Address, "grpc-store.server-address", "", "Hostname or IP of gRPC instance.")
```
pracucci, line 16:
```suggestion
	f.StringVar(&cfg.Address, "grpc-store.server-address", "", "Hostname or IP of the gRPC store instance.")
```

pkg/chunk/grpc/grpc_client.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"flag"
+
+	"google.golang.org/grpc"
+)
+
+// Config for a StorageClient
+type Config struct {
+	Address string `yaml:"address,omitempty"`
```
pracucci, line 11: We should keep naming consistency between YAML and CLI flags. Either we rename this to `server_address` or the CLI flag to `grpc-store.address`.

pkg/chunk/grpc/grpc_client.go
```diff
@@ -0,0 +1,25 @@
...
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.Address, "grpc-store.server-address", "", "Hostname or IP of gRPC instance.")
+}
+
+func connectToGrpcServer(serverAddress string) (GrpcStoreClient, *grpc.ClientConn, error) {
+	cc, err := grpc.Dial(serverAddress, grpc.WithInsecure())
+	if err != nil {
+		return nil, nil, err
```
pracucci, line 22:
```suggestion
		return nil, nil, errors.Wrapf(err, "failed to dial grpc-store %s", serverAddress)
```

pkg/chunk/grpc/grpc_client.go
```diff
@@ -0,0 +1,25 @@
...
+type Config struct {
+	Address string `yaml:"address,omitempty"`
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.Address, "grpc-store.server-address", "", "Hostname or IP of gRPC instance.")
+}
+
+func connectToGrpcServer(serverAddress string) (GrpcStoreClient, *grpc.ClientConn, error) {
```
pracucci, line 19: Note: I think using the gRPC client pool with the Thanos DNS service discovery (like we did [here](https://github.com/cortexproject/cortex/blob/master/pkg/querier/blocks_store_balanced_set.go)) would have been preferrable here, but **this can be left for a future iteration**, so no need to reiterate on it in this PR.

pkg/chunk/grpc/grpc_client.go
```diff
@@ -0,0 +1,25 @@
...
+	cc, err := grpc.Dial(serverAddress, grpc.WithInsecure())
```
pracucci, line 20: Given there's no health checking mechanism (ie. we're not using the client pool) we should setup the gRPC keepalive (we had few outages with etcd due to this). See `grpc.WithKeepaliveParams()` and the [etcd implementation](https://github.com/cortexproject/cortex/blob/c52331d3cbd4d8aa36106f0e2aeb3c1ecf8e0016/vendor/go.etcd.io/etcd/clientv3/client.go#L235-L240) as a reference.

pkg/chunk/grpc/index_client.go
```diff
@@ -0,0 +1,98 @@
...
+	w.IndexEntry = append(w.IndexEntry, &IndexEntry{
+		TableName:  tableName,
+		HashValue:  hashValue,
+		RangeValue: rangeValue,
+		Value:      value,
+	})
+}
+
+func (w *WriteBatch) Delete(tableName, hashValue string, rangeValue []byte) {
+	// TODO: implement this to support deleting index entries for gRPC spec
```
pracucci, line 23: We should resolve this TODO.

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,124 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+)
+
+type StorageClient struct {
+	cfg        Config
+	schemaCfg  chunk.SchemaConfig
+	client     GrpcStoreClient
+	connection *grpc.ClientConn
```
pracucci, line 17: Not used. Can be removed.

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,124 @@
...
+		client:     grpcClient,
+		connection: conn,
+	}
+	return client, nil
+}
+
+func (s *StorageClient) Stop() {
+	_, err := s.client.Stop(context.Background(), &Nothing{})
+	if err != nil {
+		return
```
pracucci, line 38: Never suppress errors. Please log it.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,111 @@
...
+
+package grpc;
+
+service grpc_store {
+    //index-client
+    rpc BatchWrite(WriteBatch) returns (Nothing);
```
pracucci, line 7: What's this `Nothing`? Why do we need it?

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,111 @@
...
+    //used by storage-client & index-client
+    rpc Stop(Nothing) returns (Nothing);
+}
+
+message Chunks {
+    repeated Chunk chunks = 1;
+}
+
+message Chunk {
+    bytes buf         = 1;
```
pracucci, line 32: I would name it more clearly. This is the encoded chunk, so why not calling it `encoded`?

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,111 @@
...
+}
+
+message ChunkResponse {
+    bytes buf       = 2;
+}
+
+message Nothing {}
+
+message ChunkID {
+    string chunkId = 1;
```
pracucci, line 48:
```suggestion
    string chunkID = 1;
```

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,111 @@
...
+    repeated ChunkResponse chunks = 1;
+}
+
+message ChunkResponse {
+    bytes buf       = 2;
+}
+
+message Nothing {}
+
+message ChunkID {
```
pracucci, line 47: Why do we need a message for this? Looks overkilled to me. A string would have been enough IMO.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,111 @@
...
+
+service grpc_store {
+    //index-client
+    rpc BatchWrite(WriteBatch) returns (Nothing);
+    rpc QueryPages(IndexQuery) returns (stream ReadBatch);
+
+    //storage-client
+    rpc PutChunks(Chunks) returns (Nothing);
+    //In GetChunks rpc request send buf as nil
+    rpc GetChunks(Chunks) returns (stream Chunks);
```
pracucci, line 13: I'm not sure `Chunks` is the right data structure to use here. Design wise, it would be cleaner if we would have a different data structure for read and write. In the read path (this rpc) what you actually need is a list of chunk keys (table name and key). In the write path you need the chunk key + data.

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,124 @@
...
+	var result []chunk.Chunk
+	for {
+		receivedChunks, err := streamer.Recv()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		if receivedChunks != nil {
```
pracucci, line 109: You can remove this if and use `range receivedChunks.GetChunks()` in the next line.

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,124 @@
...
+			for _, chunkResponse := range receivedChunks.Chunks {
+				var c chunk.Chunk
+				if chunkResponse != nil {
```
pracucci, line 112: When this could happen?

pkg/chunk/grpc/storage_client.go
pracucci: My only comment is that `firstErr` could be more useful than `lastErr`. It's generally preferable seeing the first error not the last error. The last error could be a consequence of first error, but no viceversa.

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/cortexproject/cortex/pkg/chunk"
+)
+
+type TableClient struct {
+	cfg        Config
+	client     GrpcStoreClient
+	connection *grpc.ClientConn
+}
+
+// NewStorageClient returns a new StorageClient.
```
pracucci, line 19:
```suggestion
// NewTableClient returns a new StorageClient.
```

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,113 @@
...
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+)
+
+type TableClient struct {
+	cfg        Config
+	client     GrpcStoreClient
+	connection *grpc.ClientConn
```
pracucci, line 16: Unused. Please remove it.

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// NewStorageClient returns a new StorageClient.
+func NewTableClient(ctx context.Context, cfg Config) (*TableClient, error) {
```
pracucci, line 20: `ctx` is unused. Please remove it.

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,113 @@
+package grpc
+
+import (
+	stdContext "context"
+
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
```
pracucci, line 7: I don't think this should be used. Please use only `"context"`.

pkg/chunk/grpc/gRPC.proto
```diff
@@ -0,0 +1,111 @@
...
+message TableDesc {
+    string name                  = 1;
+    bool useOnDemandIOMode       = 2;
+    int64 provisionedRead        = 3;
+    int64 provisionedWrite       = 4;
+    map<string, string> tags     = 5;
+}
+
+message ListTablesResponse {
+    repeated string tables = 1;
```
pracucci, line 103:
```suggestion
    repeated string tableNames = 1;
```

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,113 @@
...
+	tableDesc, err := c.client.DescribeTable(ctx, tableName)
+	if err != nil {
+		return desc, true, errors.WithStack(err)
+	}
+	desc.Name = tableDesc.Desc.Name
+	desc.ProvisionedRead = tableDesc.Desc.ProvisionedRead
+	desc.ProvisionedWrite = tableDesc.Desc.ProvisionedWrite
+	desc.UseOnDemandIOMode = tableDesc.Desc.UseOnDemandIOMode
+	desc.Tags = tableDesc.Desc.Tags
+	return desc, true, nil
```
pracucci, line 61: Why we always return `true` as `isActive`? The response contains `IsActive` which we ignore.

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,113 @@
...
+		return errors.WithStack(err)
+	}
+	return nil
+}
+
+func (c *TableClient) DescribeTable(ctx context.Context, name string) (desc chunk.TableDesc, isActive bool, err error) {
+	tableName := &TableName{TableName: name}
+	tableDesc, err := c.client.DescribeTable(ctx, tableName)
+	if err != nil {
+		return desc, true, errors.WithStack(err)
```
pracucci, line 54: A minor thing, but a good practice when returning errors:

```suggestion
		return desc, false, errors.WithStack(err)
```

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,113 @@
...
+
+	currentTable.Name = current.Name
+	currentTable.UseOnDemandIOMode = current.UseOnDemandIOMode
+	currentTable.ProvisionedWrite = current.ProvisionedWrite
+	currentTable.ProvisionedRead = current.ProvisionedRead
+	if current.Tags != nil {
+		currentTable.Tags = current.Tags
+	} else {
+		currentTable.Tags = map[string]string{}
+	}
```
pracucci, line 76: Why don't just allow to set `nil` to `currentTable`? To give you an example, in `CreateTable()` you're not doing this and you allow to set `Tags` to `nil`.

pkg/chunk/grpc/index_client.go
pracucci: What do you mean @gouthamve? The error is checked above. However, I'm not sure why we need this `if` at all.

pkg/chunk/grpc/index_client.go
VineethReddy02: @pracucci I didn't notice any delete implementation for Cassandra and other stores. As it was TODO in respective store implementations. I reflected the same here. Do you want me to drop such comment with TODO?

pracucci: I think it has been added after you initiated this work. Please rebase `master`.

pkg/chunk/grpc/index_client.go
VineethReddy02: @pracucci Though we check the error. If the response is nil the callback func fails. I am just trying to make sure if grpc server doesn't handle all cases. cortex shouldn't crash in any scenario.

pracucci: If the gRPC server is buggy and returns `nil` batch and `nil` error (which is a case we never expect to happen) you will end up silently ignoring it and you would potentially end up with wrong query results. If you would have the same exact bug in Cortex today, Cortex would panic. I don't see why we should do it differently with the gRPC store.

VineethReddy02: Okay. In that case, I can remove the additional ```nil``` value validation.

pkg/chunk/grpc/grpc_client.go
VineethReddy02: @pracucci Do you have any suggestions on these params. the keepalive check interval and the timeout duration when we don't have active rpcs?
I am thinking to put a 1 min interval to check is alive and 2 mins timeout interval to close the connection if no activity is noticed.
```
	params := keepalive.ClientParameters{
		Time:                time.Minute*1,
		Timeout:           time.Minute*2,
		PermitWithoutStream: true,
	}
```

pkg/chunk/grpc/gRPC.proto
VineethReddy02: So in proto files. When we have nothing to return back. We maintain an empty message i.e
`message Nothing {};`
If there any workaround to avoid this. I can consider using them.

pkg/chunk/grpc/gRPC.proto
VineethReddy02: So we use ChunkID message only in DeleteChunk rpc as
` rpc DeleteChunks(ChunkID) returns (Nothing); `
I will replace this into  
` rpc DeleteChunks(string) returns (Nothing); `

pkg/chunk/grpc/gRPC.proto
VineethReddy02: Initially, we maintained different structures but @gouthamve, and I felt to use the same data structure for read & write path to make it simple. Maybe @gouthamve can here on the context behind this decision.

pkg/chunk/grpc/grpc_client.go
pracucci: I would be even more aggressive, like: 
- Time: 20s
- Timeout: 10s
- PermitWithoutStream: true

pkg/chunk/grpc/gRPC.proto
pracucci: You can use the built-in `google.protobuf.Empty` ([see this](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/empty.proto))

pkg/chunk/grpc/gRPC.proto
pracucci: 👍

pkg/chunk/grpc/gRPC.proto
pracucci: Design wise I think it's a bit risky for future evolutions. As I commented in the review message, I think that having a dedicated Request and Response message for each function is more future proof, ie.

```
rpc PutChunks(PutChunksRequest) returns (google.protobuf.Empty);
rpc GetChunks(GetChunksRequest) returns (stream GetChunksResponse);
```

VineethReddy02:
@pracucci @gouthamve 
Addressed the review comments & added unit tests by writing the gRPC mock server.
Feel free to suggest changes around unit tests. I am writing unit tests with a mock gRPC server for the first time. :) Would love to iterate it if needed.
I will also extend that with integration tests for this feature in the future PR's.

pkg/chunk/grpc/grpc_client_test.go
```diff
@@ -0,0 +1,183 @@
+package grpc
+
+import (
+	"testing"
+
+	"context"
+
+	"github.com/prometheus/prometheus/pkg/labels"
```
gouthamve, line 8: Why are we spacing the imports out?

pkg/chunk/grpc/grpc_server_mock.go
```diff
@@ -0,0 +1,207 @@
+package grpc
+
+import (
+	"context"
+	"log"
+	"net"
+	"time"
+
+	"github.com/pkg/errors"
```
gouthamve, line 9: Same here

pkg/chunk/grpc/table_client.go
```diff
@@ -0,0 +1,100 @@
+package grpc
+
+import (
+	"context"
+
```
gouthamve, line 5: Same here

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,128 @@
...
+		for _, chunkResponse := range receivedChunks.GetChunks() {
+			var c chunk.Chunk
+			if chunkResponse != nil {
+				decodeContext = chunk.NewDecodeContext()
+				err = c.Decode(decodeContext, chunkResponse.Encoded)
+				if err != nil {
+					return result, err
+				}
+			}
+			if c.Data != nil {
```
gouthamve, line 121: I think we can remove this check here.

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,128 @@
+package grpc
+
+import (
+	"io"
+
+	"context"
+
+	"github.com/pkg/errors"
```
gouthamve, line 8: Don't space imports here too

pkg/chunk/grpc/storage_client.go
VineethReddy02: @gouthamve for some reason if the gRPC server sends `c.Data `stream as nil cortex will crash.
So it's an additional check to ensure we append data which is not nil. If it's okay I can remove it.

pkg/chunk/grpc/grpc_server_mock.go
VineethReddy02: In all the go files in the cortex code base, I see this grouping of imports I mean standards packages, other Github imports, and then the cortex project reference imports.

pkg/chunk/grpc/storage_client.go
```diff
@@ -0,0 +1,124 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		for _, chunkResponse := range receivedChunks.GetChunks() {
+			var c chunk.Chunk
+			if chunkResponse != nil {
+				decodeContext = chunk.NewDecodeContext()
```
pracucci, line 111: We've recently learned this is an expensive operation. You can use the same `decodeContext` to decode all chunks, because `c.Decode()` will call `Reset()` on the decode context before decoding the chunk.

pkg/chunk/grpc/storage_client.go
pracucci: I would remove it as well. Silently ignoring this condition may lead to very subdle bugs. Having Cortex panicking sounds actually better in this edge case.

pkg/chunk/grpc/grpc.proto
```diff
@@ -0,0 +1,121 @@
...
+    bytes encoded     = 1;
+    string key        = 2;
+    string tableName  = 3;
+}
+
+message ChunkID {
+    string chunkID = 1;
+}
+
+message TableName {
```
pracucci, line 52: A message `TableName` containing `tableName` looks a weird design to me. If you wanna keep it extensible, then I would split it into:

```
message DeleteTableRequest {
  string tableName = 1;
}

message DescribeTableRequest {
  string tableName = 1;
}
```

pkg/chunk/grpc/grpc.proto
```diff
@@ -0,0 +1,121 @@
+syntax = "proto3";
+
+package grpc;
+
+import "google/protobuf/empty.proto";
+
+service grpc_store {
+    //index-client
+    rpc BatchWrite(BatchWrites) returns (google.protobuf.Empty);
+    rpc QueryPages(IndexQuery) returns (stream ReadBatch);
```
pracucci, line 10: Should we call it `QueryIndex(QueryIndexRequest) returns (stream QueryIndexResponse)` for clarity/consistency?

pkg/chunk/grpc/grpc_server_mock.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	param := grpc.WithKeepaliveParams(params)
+	cc, err := grpc.Dial(serverAddress, param, grpc.WithInsecure())
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to dial grpc-store %s", serverAddress)
+	}
+	return NewGrpcStoreClient(cc), cc, nil
+}
+
+func createTestGrpcServer() {
```
pracucci, line 183: You can pass `t *testing.T` and then assert on no error using `require.NoError(t, err)`. That's a pattern we frequently use. For example:

```
lis, err := net.Listen("tcp", "localhost:6688")
require.NoError(t, err)
```

pkg/chunk/grpc/grpc_server_mock.go
```diff
@@ -0,0 +1,203 @@
...
+
+// NewStorageClient returns a new StorageClient.
+func newTestStorageServer(cfg Config) (*server, error) {
+	client := &server{
+		Cfg: cfg,
+	}
+	return client, nil
+}
+
+func connectToTestGrpcServer(serverAddress string) (GrpcStoreClient, *grpc.ClientConn, error) {
```
pracucci, line 169: How is this different than `connectToGrpcServer()`? Why not using `connectToGrpcServer()`?

pkg/chunk/grpc/grpc_server_mock.go
```diff
@@ -0,0 +1,203 @@
...
+	cc, err := grpc.Dial(serverAddress, param, grpc.WithInsecure())
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to dial grpc-store %s", serverAddress)
+	}
+	return NewGrpcStoreClient(cc), cc, nil
+}
+
+func createTestGrpcServer() {
+	var cfg server
+	lis, err := net.Listen("tcp", "localhost:6688")
```
pracucci, line 185: Don't hardcode ports. It makes tests more flaky. You can listen on `localhost:0`, and then having `createTestGrpcServer()` returning the address where the server is listening to.

pkg/chunk/grpc/grpc_server_mock.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
+	s := grpc.NewServer()
+	// Enable reflection to expose endpoints this server offers.
+	reflection.Register(s)
```
pracucci, line 191: Why do you need this?

pkg/chunk/grpc/grpc_server_mock.go
```diff
@@ -0,0 +1,203 @@
...
+	return NewGrpcStoreClient(cc), cc, nil
+}
+
+func createTestGrpcServer() {
+	var cfg server
+	lis, err := net.Listen("tcp", "localhost:6688")
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
+	s := grpc.NewServer()
```
pracucci, line 189: We should call `GracefulStop()` at the end of the test. A tecnique you can use is returning a cleanup function from `createTestGrpcServer()` and then having the test calling the cleanup in a defer. For example:

```
func createTestGrpcServer() (func()) {
  s := grpc.NewServer()
  cleanup := func() {
    s.GracefulStop()
  }

  return cleanup
```

_This is pattern we frequently use._

pkg/chunk/grpc/grpc.proto
```diff
@@ -0,0 +1,121 @@
...
+
+    //table-client
+    rpc ListTables(google.protobuf.Empty) returns (ListTablesResponse);
+    rpc CreateTable(TableDesc) returns (google.protobuf.Empty);
+    rpc DeleteTable(TableName) returns (google.protobuf.Empty);
+    rpc DescribeTable(TableName) returns (DescribeTableResponse);
+    rpc UpdateTable(UpdateTableRequest) returns (google.protobuf.Empty);
+
+    //used by storage-client & index-client
+    rpc Stop(google.protobuf.Empty) returns (google.protobuf.Empty);
```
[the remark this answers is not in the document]
VineethReddy02: So, Initially I was thinking there is no use of Stop(). But I was thinking in next iteration when we add support for multiple grpc-stores. If for some reason we want to terminate the connection from the client-side with one of the grpc-store maybe this can help, just a thought. We can definitely drop this. If we have no use.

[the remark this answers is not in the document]
pracucci: I would suggest to remove `Stop()` and add it if/when required.

pkg/chunk/grpc/grpc_server_mock.go
VineethReddy02: Yeah, we don't need this. Didn't notice.

pkg/chunk/grpc/grpc_server_mock.go
VineethReddy02: yeah. We can use it.

VineethReddy02:
Thanks @pracucci for taking the time & reviewing. I made all the suggested changes. 
Looking forward in getting this merged. :)

pkg/chunk/grpc/grpc.proto
```diff
@@ -0,0 +1,122 @@
...
+
+    //storage-client
+    rpc PutChunks(PutChunksRequest) returns (google.protobuf.Empty);
+    //In GetChunks rpc request send buf as nil
+    rpc GetChunks(GetChunksRequest) returns (stream GetChunksResponse);
+    rpc DeleteChunks(ChunkID) returns (google.protobuf.Empty);
+
+    //table-client
+    rpc ListTables(google.protobuf.Empty) returns (ListTablesResponse);
+    rpc CreateTable(TableDesc) returns (google.protobuf.Empty);
```
pracucci, line 21: I would define `CreateTableRequest` for extensibility, like we did for any other rpc. The `CreateTableRequest` can just contain `TableDesc` for now.

pkg/chunk/grpc/storage_client.go
pracucci: This comment is still valid. `connection` is unused, but I think we should use it to close the connection in `Stop()`.

pkg/chunk/grpc/grpc.proto
VineethReddy02: Rather nesting the rpc messages. I think it's better to rename it as suggested.

pracucci: Yes, but `TableDesc` is also used in other places, no?

VineethReddy02: Ahh... I didn't notice. TableDesc being used in many other places. I will nest the message as suggested.

pkg/chunk/grpc/table_client.go
```diff
@@ -84,12 +84,13 @@ func (c *TableClient) UpdateTable(ctx context.Context, current, expected chunk.T
 }
 
 func (c *TableClient) CreateTable(ctx context.Context, desc chunk.TableDesc) error {
-	tableDesc := &TableDesc{}
-	tableDesc.Name = desc.Name
-	tableDesc.ProvisionedRead = desc.ProvisionedRead
-	tableDesc.ProvisionedWrite = desc.ProvisionedWrite
-	tableDesc.Tags = desc.Tags
-	tableDesc.UseOnDemandIOMode = desc.UseOnDemandIOMode
+	tableDesc := &CreateTableRequest{}
```
pracucci, line 87: Please rename `tableDesc` to `req`, otherwise below it's a bit weird reading `tableDesc.Desc.*`.

pstibrany:
Since this API is supposedly to be used for writing 3rd party storage implementations, I think it lacks documentation (at least inside grpc.proto file, but more generic documentation about how Cortex uses the storage and what should the user of this API reasonably expect).

Is there any implementation using this API? How do we know it satisfies the needs of external storage client?

(I am also skeptical about including this in Cortex in the first place. Many failure modes related to gRPC (grpc server not running, big responses, bad config, ...) would be automatically removed if client was included in the Cortex)

gouthamve:
> I think it lacks documentation

Absolutely. We still have follow up work to document and give a reference in-memory implementation. And to use the in-mem implementation to write an integration test.

> would be automatically removed if client was included in the Cortex

What do you mean?

pstibrany:
> Absolutely. We still have follow up work to document and give a reference in-memory implementation. And to use the in-mem implementation to write an integration test.

I think we need real implementation or two, not just in-mem one.

> > would be automatically removed if client was included in the Cortex
> 
> What do you mean?

I was talking about failure modes related to gRPC -- which wouldn't be a problem if client was in Cortex directly, and not using gRPC (well, obviously). Just saying that using gRPC adds more complexity.

gouthamve:
I kinda disagree :) I see gRPC having the same amount of failure modes as a remote Bigtable/Cassandra. They are both endpoints.

> grpc server not running, big responses, bad config, ...

We faced these with Bigtable tbh, specially the big responses. We had to up our gRPC message limit with a previous schema, hence came up with the v9/v10 schema. I would say gRPC makes the failure modes more likely, but doesn't introduce new ones.

And while I agree that having the client directly in Cortex would be the best, I am not super enthused about a ton of storage clients in the tree.

gouthamve:
> I think we need real implementation or two, not just in-mem one.

The first one was going to be Postgres/MySQL so Loki could use it as the index. Not sure what the next one would be tbh.

pstibrany:
> And while I agree that having the client directly in Cortex would be the best, I am not super enthused about a ton of storage clients in the tree.

I personally think that biggest issue in people contributing more clients is lack of good documentation, and pointers about how to start.

Anyway, please feel free to merge when PR is ready, my comments are not merge-blockers.

VineethReddy02:
@gouthamve @pracucci 
I see CI fails with quay error.

`Error response from daemon: Get https://quay.io/v2/: socks connect tcp 127.0.0.1:1080->quay.io:443: unknown error address type not supported`

pracucci:
Let's gooo 🚀 ! _I will merge as soon as the CI passes_

PR #2221, opened 2020-03-06
Remove support schema flags, only use config file.

### LINKED ISSUE #2208: Remove the schema config from flags and use only config file
Right now for setting the schema config, there are a ton of flags like:

`dynamodb.daily-buckets-from`, `dynamodb.base64-buckets-from`, `dynamodb.v4-schema-from`,....... `dynamodb.use-periodic-tables`, `dynamodb.periodic-table.from`, etc.

This is confusing and can be better served by the config file which is cleaner and clearer:
```yaml
      - from: "2019-02-13"
        chunks:
            period: 168h
            prefix: dev_chunks_
        index:
            period: 168h
            prefix: dev_index_
        object_store: gcs
        schema: v9
        store: gcp-columnkey
      - from: "2019-02-24"
        chunks:
            period: 168h
            prefix: dev_chunks_
        index:
            period: 168h
            prefix: dev_index_
        object_store: gcs
        schema: v9
        store: bigtable-hashed
      - from: "2019-03-05"
        chunks:
            period: 168h
            prefix: dev_chunks_
        index:
            period: 168h
            prefix: dev_index_
        object_store: gcs
        schema: v10
        store: bigtable-hashed
```

Further, you can set more things via the config file than flags. I initially set out to rename the dynamodb prefixes, but then realised its better to totally remove these flags and only allow it to be set via config file.

gouthamve:
Also rename the schema config file flag to something sane.

Fixes #2208 

**Checklist**
- [ ] ~Tests updated~
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,13 @@
 ## master / unreleased
 
 <<<<<<< HEAD
+<<<<<<< HEAD
 =======
 * [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file has been renamed from `-config-yaml` to `schema-config-file`. See https://cortexmetrics.io/docs/configuration/schema-configuration/ for more details on how to configure the schema using the YAML file. 
 >>>>>>> Remove support schema flags, only use config file.
+=======
+* [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file (`-config-yaml`) has been deprecated. Please use `schema-config-file`. See https://cortexmetrics.io/docs/configuration/schema-configuration/ for more details on how to configure the schema using the YAML file. 
+>>>>>>> Deprecate not remove the config flag.
 * [CHANGE] The frontend http server will now send 502 in case of deadline exceeded and 499 if the user requested cancellation. #2156
 * [CHANGE] Config file changed to remove top level `config_store` field in favor of a nested `configdb` field. #2125
 * [CHANGE] We now enforce queries to be up to `-querier.max-query-into-future` into the future (defaults to 10m). #1929
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,16 @@
 * [CHANGE] Experimental TSDB: switched the blocks storage index header to the binary format. This change is expected to have no visible impact, except lower startup times and memory usage in the queriers. It's possible to switch back to the old JSON format via the flag `-experimental.tsdb.bucket-store.binary-index-header-enabled=false`. #2223
 * [CHANGE] WAL replays are now done while the rest of Cortex is starting, and more specifically, when HTTP server is running. This makes it possible to scrape metrics during WAL replays. Applies to both chunks and experimental blocks storage. #2222
 * [CHANGE] Default to BigChunk encoding; may result in slightly higher disk usage if many timeseries have a constant value, but should generally result in fewer, bigger chunks. #2207
+* [FEATURE] Fan out parallelizable queries to backend queriers concurrently. #1878
+  * `querier.parallelise-shardable-queries` (bool)
+  * Requires a shard-compatible schema (v10+)
+  * This causes the number of traces to increase accordingly.
+  * The query-frontend now requires a schema config to determine how/when to shard queries, either from a file or from flags (i.e. by the `-schema-config-file` CLI flag). This is the same schema config the queriers consume. The schema is only required to use this option.
+  * It's also advised to increase downstream concurrency controls as well:
+    * `querier.max-outstanding-requests-per-tenant`
+    * `querier.max-query-parallelism`
+    * `querier.max-concurrent`
+    * `server.grpc-max-concurrent-streams` (for both query-frontends and queriers)
 * [FEATURE] Added a read-only local alertmanager config store using files named corresponding to their tenant id. #2125
 * [FEATURE] Added user sub rings to distribute users to a subset of ingesters. #1947
   * `--experimental.distributor.user-subring-size`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,7 @@
 
 ## master / unreleased
 
-<<<<<<< HEAD
-<<<<<<< HEAD
-=======
-* [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file has been renamed from `-config-yaml` to `schema-config-file`. See https://cortexmetrics.io/docs/configuration/schema-configuration/ for more details on how to configure the schema using the YAML file. 
->>>>>>> Remove support schema flags, only use config file.
-=======
 * [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file (`-config-yaml`) has been deprecated. Please use `schema-config-file`. See https://cortexmetrics.io/docs/configuration/schema-configuration/ for more details on how to configure the schema using the YAML file. 
->>>>>>> Deprecate not remove the config flag.
 * [CHANGE] The frontend http server will now send 502 in case of deadline exceeded and 499 if the user requested cancellation. #2156
 * [CHANGE] Config file changed to remove top level `config_store` field in favor of a nested `configdb` field. #2125
 * [CHANGE] We now enforce queries to be up to `-querier.max-query-into-future` into the future (defaults to 10m). #1929
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file (`-config-yaml`) has been deprecated. Please use `schema-config-file`. See https://cortexmetrics.io/docs/configuration/schema-configuration/ for more details on how to configure the schema using the YAML file. 
+* [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file (`-config-yaml`) has been deprecated. Please use `schema-config-file`. See https://cortexmetrics.io/docs/configuration/schema-configuration/ for more details on how to configure the schema using the YAML file. #2221
 * [CHANGE] The frontend http server will now send 502 in case of deadline exceeded and 499 if the user requested cancellation. #2156
 * [CHANGE] Config file changed to remove top level `config_store` field in favor of a nested `configdb` field. #2125
 * [CHANGE] We now enforce queries to be up to `-querier.max-query-into-future` into the future (defaults to 10m). #1929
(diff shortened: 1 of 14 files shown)


docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
...
+```yaml
+configs: []<period_config>
+```
+
+Where `period_config` is
+```
+from: <Date>
+store: <string> // The index client to use.
+object_store: <string> // The object client to use. If none is specified, `store` is used for storing chunks as well.
+schema: <string> // The schema version to use. Valid ones are v1, v2, v3,..., v11
```
sandeepsukhani: We don't have `v7` and `v8` schemas.

CHANGELOG.md
```diff
@@ -2,16 +2,7 @@
...
-  * `querier.parallelise-shardable-queries` (bool)
-  * Requires a shard-compatible schema (v10+)
-  * This causes the number of traces to increase accordingly.
-  * The query-frontend now requires a schema config to determine how/when to shard queries, either from a file or from flags (i.e. by the `config-yaml` CLI flag). This is the same schema config the queriers consume. The schema is only required to use this option.
-  * It's also advised to increase downstream concurrency controls as well:
-    * `querier.max-outstanding-requests-per-tenant`
-    * `querier.max-query-parallelism`
-    * `querier.max-concurrent`
-    * `server.grpc-max-concurrent-streams` (for both query-frontends and queriers)
+* [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file has been renamed from `-config-yaml` to `schema-config-file`. See <URL> for more details on how to configure the schema using the YAML file.
```
sandeepsukhani: I think you missed adding the URL here.

docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
+---
+title: "Schema Configuration"
+linkTitle: "Schema Configuration"
+weight: 1
+slug: schema-configuration
+---
+
+Cortex uses a NoSQL Store to store its index and optionally an Object store to store its chunks. Cortex has overtime evolved its schema to be more optimal and better fit the use-cases that arised. Currently there are 10 schemas that are used in production. You can move from one schema to another if a new schema fits your purpose better, but you still need to configure Cortex to make sure it can read the old data in the old schemas.
```
sandeepsukhani: We actually have 9 schemas now.

docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
...
+object_store: <string> // The object client to use. If none is specified, `store` is used for storing chunks as well.
+schema: <string> // The schema version to use. Valid ones are v1, v2, v3,..., v11
+index: <periodic_table_config>
+chunks: <periodic_table_config>
+```
+
+Where `periodic_table_config` is
+```
+prefix: <string> // The prefix to use for the tables.
+period: <time.Duration> // The period at which new tables are created and used. Typically 168h
```
sandeepsukhani: `The period at which new tables are created and used.` sounds a little confusing for a new user. Any other way to describe it? Maybe let us not correlate with when tables are created and used?

gouthamve: Is it better now?

pkg/chunk/schema_config.go
```diff
@@ -70,111 +69,11 @@ type SchemaConfig struct {
...
-			config.ChunkTables = cfg.legacy.ChunkTables
-		})
-	}
-	if cfg.legacy.BigtableColumnKeyFrom.IsSet() {
-		cfg.ForEachAfter(cfg.legacy.BigtableColumnKeyFrom.Time, func(config *PeriodConfig) {
-			config.IndexType = "gcp-columnkey"
-		})
-	}
-	return nil
+	flag.StringVar(&cfg.fileName, "schema-config-file", "", "The path to the schema config file.")
```
bboreham: Suggest supporting the old name as a deprecated option.

docs/configuration/schema-config-reference.md
sandeepsukhani: Much better, just add a space in `theperiod`. Thanks!

pkg/chunk/schema_config.go
gouthamve: Done, also tested that it works.

docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
+---
+title: "Schema Configuration"
+linkTitle: "Schema Configuration"
+weight: 1
+slug: schema-configuration
+---
+
+Cortex uses a NoSQL Store to store its index and optionally an Object store to store its chunks. Cortex has overtime evolved its schema to be more optimal and better fit the use-cases that arose. Currently there are 9 schemas that are used in production but we recommend running with `v9` schema when possible. You can move from one schema to another if a new schema fits your purpose better, but you still need to configure Cortex to make sure it can read the old data in the old schemas.
```
pracucci: I would make it clear this applies to the chunks storage engine only.

pracucci:
```suggestion
Cortex uses a NoSQL Store to store its index and optionally an Object store to store its chunks. Cortex has overtime evolved its schema to be more optimal and better fit the use cases that arose.

Currently there are 9 schema versions that are used in production but we recommend running with `v9` schema when possible. You can move from one schema to another if a new schema fits your purpose better, but you still need to configure Cortex to make sure it can read the old data stored with the old schemas.
```

docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
...
+
+You can configure the schemas using a YAML config file, that you can point to using the `-schema-config-file` flag. It has the following YAML spec:
+
+```yaml
+configs: []<period_config>
+```
+
+Where `period_config` is
+```
+from: <Date>
```
pracucci: I would suggest to use the same config syntax of the config file reference, to keep consistency.

docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
...
+    store: bigtable-hashed
+```
+
+Note how we started out with v9 and just Bigtable, but later added GCS as the object store, finally moving to v10. This is a complex schema file from out dev cluster that saw a lot of changes dating all the way back to 2018, a typical schema config file usually has one or two schema versions.
+
+### Migrating from flags to schema file
+
+We've recently removed the ability to configure schema via flags. If you were using flags, you need to port your config from flags to the config file.
+
+If you're using:
```
pracucci: Please add a new line after this, in order to render the markdown list correctly.

docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
...
+        prefix: dev_index_
+    object_store: gcs
+    store: bigtable-hashed
+```
+
+Note how we started out with v9 and just Bigtable, but later added GCS as the object store, finally moving to v10. This is a complex schema file from out dev cluster that saw a lot of changes dating all the way back to 2018, a typical schema config file usually has one or two schema versions.
+
+### Migrating from flags to schema file
+
+We've recently removed the ability to configure schema via flags. If you were using flags, you need to port your config from flags to the config file.
```
pracucci: I would suggest to never use time-sensitive words like "recently" in the doc, because you never know when this doc will be updated again and the "recently" may easily become months or years.

```suggestion
Legacy versions of Cortex did support the ability to configure schema via flags. If you are still using flags, you need to migrate your configuration from flags to the config file.
```

docs/configuration/schema-config-reference.md
```diff
@@ -0,0 +1,104 @@
...
+      period: 168h
+      prefix: cortex_chunks  
+    store: aws-dynamo/bigtable-hashed/cassandra/boltdb
+    object_store: <above options>/s3/gcs/azure/filesystem
+```
+
+An example of an advanced schema file with a lot of changes:
+```
+configs:
+  - from: "2018-08-23"
```
pracucci: For each config block I would add a `#` comment to explain what it does. For example, I'm adding a suggestion for the first block:

```suggestion
  # Starting from 2018-08-23 Cortex should store chunks and indexes
  # on Google BigTable using weekly periodic tables. The chunks table
  # names will be prefixed with "dev_chunks_", while index tables will be
  # prefixed with "dev_index_".
  - from: "2018-08-23"
```

integration/query_frontend_test.go
```diff
@@ -89,6 +89,7 @@ func runQueryFrontendTest(t *testing.T, setup queryFrontendSetup) {
 	require.NoError(t, s.StartAndWaitReady(consul))
 
 	configFile, flags := setup(t, s)
+	fmt.Println(flags, "SHAKAKA")
```
pracucci: Please remove this.

pkg/chunk/schema_config.go
```diff
@@ -69,116 +69,27 @@ func (d *DayTime) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-
-	f.StringVar(&cfg.OriginalTableName, "dynamodb.original-table-name", "cortex", "The name of the DynamoDB table used before versioned schemas were introduced.")
-	f.BoolVar(&cfg.UsePeriodicTables, "dynamodb.use-periodic-tables", false, "Should we use periodic tables.")
-
-	f.Var(&cfg.IndexTablesFrom, "dynamodb.periodic-table.from", "Date after which to use periodic tables.")
-	cfg.IndexTables.RegisterFlags("dynamodb.periodic-table", "cortex_", f)
-	f.Var(&cfg.ChunkTablesFrom, "dynamodb.chunk-table.from", "Date after which to write chunks to DynamoDB.")
-	cfg.ChunkTables.RegisterFlags("dynamodb.chunk-table", "cortex_chunks_", f)
+	flag.StringVar(&cfg.fileName, "schema-config-file", "", "The path to the schema config file.")
+	// TODO(gouthamve): Add a metric and log line for this.
```
pracucci: Not sure the metric is that important, but we could already add a log line. We can do it in the `loadFromFile()`.

khaines: I agree that a log line is appropriate, to note the use of a deprecated flag. I don't think a metric is useful.

pkg/chunk/schema_config.go
```diff
@@ -69,116 +69,31 @@ func (d *DayTime) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-
-	f.StringVar(&cfg.OriginalTableName, "dynamodb.original-table-name", "cortex", "The name of the DynamoDB table used before versioned schemas were introduced.")
-	f.BoolVar(&cfg.UsePeriodicTables, "dynamodb.use-periodic-tables", false, "Should we use periodic tables.")
-
-	f.Var(&cfg.IndexTablesFrom, "dynamodb.periodic-table.from", "Date after which to use periodic tables.")
-	cfg.IndexTables.RegisterFlags("dynamodb.periodic-table", "cortex_", f)
-	f.Var(&cfg.ChunkTablesFrom, "dynamodb.chunk-table.from", "Date after which to write chunks to DynamoDB.")
-	cfg.ChunkTables.RegisterFlags("dynamodb.chunk-table", "cortex_chunks_", f)
+	flag.StringVar(&cfg.fileName, "schema-config-file", "", "The path to the schema config file.")
+	// TODO(gouthamve): Add a metric for this.
```
pracucci: Do we wanna keep this TODO? Is it really useful a metric for this?

PR #2222, opened 2020-03-06
Move WAL replays (chunks / tsdb) into ingester's Starting state.

pstibrany:
**What this PR does**: This PR moves WAL replays into Ingester's Starting method. Ingester cannot operate until these are finished, so all public methods now check if Ingester is in Running or Stopping state. Why Stopping? Because querying methods still work while it is stopping.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

/cc @codesome

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -196,7 +196,7 @@ func (i *Ingester) starting(ctx context.Context) error {
 	var err error
 	i.wal, err = newWAL(i.cfg.WALConfig, i.userStates.cp)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "starting WAL")
 	}
 
 	// Now that user states have been created, we can start the lifecycler.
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester.go
```diff
@@ -689,6 +732,10 @@ func (i *Ingester) AllUserStats(ctx context.Context, req *client.UserStatsReques
 
 // Check implements the grpc healthcheck
 func (i *Ingester) Check(ctx context.Context, req *grpc_health_v1.HealthCheckRequest) (*grpc_health_v1.HealthCheckResponse, error) {
+	if err := i.checkRunningOrStopping(); err != nil {
```
pracucci: The readiness probe below (`CheckReady()`) only consider `Running` valid, while here we consider `Stopping` too. Which one is right? Shouldn't they be aligned?

pkg/ingester/ingester.go
```diff
@@ -689,6 +732,10 @@ func (i *Ingester) AllUserStats(ctx context.Context, req *client.UserStatsReques
...
+		return nil, err
```
pracucci: Instead of returning error, have you considered returning a gRPC `HealthCheckResponse` with `HealthCheckResponse_NOT_SERVING` status? I'm not sure which is better, so I'm asking :)

pstibrany: I think it doesn't matter too much, as [our code](https://github.com/cortexproject/cortex/blob/9a40de2cafbc20cb324b1f4bb1313c9f50cba0af/pkg/ingester/client/pool.go#L193) only checks for `HealthCheckResponse_SERVING`, but `HealthCheckResponse_NOT_SERVING` may still be nicer, so I'll use that.

pkg/ingester/ingester.go
pstibrany: I don't see any issue for `CheckReady` using `checkRunningOrStopping` check as well.

PR #2223, opened 2020-03-06
Upgraded Thanos to 0.11.0 and switched to binary index header

pracucci:
**What this PR does**:
In this PR:
- Upgraded Thanos to 0.11.0
- Switched to binary index header with an option to go back if necessary (considered experimental on Thanos side, but the whole blocks storage is experimental in Cortex so I would suggest to use the binary header as default)

Upgraded dependencies:
- Thanos
- Minio (only used by Thanos)
- `go-openapi` (no breaking changes)
- Prometheus alertmanager from `v0.19.0` to `v0.20.0`: looking at the [changelog](https://github.com/prometheus/alertmanager/blob/master/CHANGELOG.md) and code changes, looks safe (except a bug in dispatcher metrics which I've fixed in https://github.com/prometheus/alertmanager/pull/2200)

Preliminary results in a small test cluster:

Screen Shot 2020-03-06 at 15 54 26

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci:
_Integration tests in CI are failing because they run on a older version of Go. I will upgrade the Go version used by CI (integration tests) in a separate PR (which will need to get merged before this)._

docs/configuration/config-file-reference.md
```diff
@@ -2198,6 +2198,17 @@ bucket_store:
   # CLI flag: -experimental.tsdb.bucket-store.meta-sync-concurrency
   [meta_sync_concurrency: <int> | default = 20]
 
+  # Whether the bucket store should use the binary index header. If false, it
+  # uses the legacy JSON index header.
```
pstibrany: Calling old one "legacy" may be little premature, given that binary index header is still experimental.

pkg/alertmanager/alertmanager.go
```diff
@@ -72,7 +72,15 @@ type Alertmanager struct {
...
 
-var webReload = make(chan chan error)
+var (
+	webReload = make(chan chan error)
+
+	// Workaround a bug in the alertmanager which doesn't register the
+	// metrics in the input registry but to the global default one.
+	// TODO change once the vendored alertmanager will have this PR merged into:
+	//      https://github.com/prometheus/alertmanager/pull/2200
+	dispatcherMetrics = dispatch.NewDispatcherMetrics(prometheus.NewRegistry())
```
pstibrany: I don't get how this is a workaround for that bug? `dispatch.NewDispatcherMetrics` still registers into default registry.

pracucci: Yes, but we need to pass something otherwise it panics.

docs/configuration/config-file-reference.md
pracucci: Fair enough. I've been too much optimistic. Fixed!

pkg/alertmanager/alertmanager.go
pstibrany: Makes sense now, thanks!

PR #2225, opened 2020-03-06
Added Noop service implementation

pstibrany:
**What this PR does**: Added service implementation that does nothing. It is useful when some "noop" type also needs to be a service. There is no need to keep extra goroutine running in such case. (I've come across this need when trying to convert WAL component into a Service. There is "noop" wal implementation that could use this).

This PR also extracts services listener type, shared between basicService and noopService.

### DIFF
--- a/pkg/util/services/basic_service.go
+++ b/pkg/util/services/basic_service.go
@@ -56,7 +56,6 @@ type BasicService struct {
 	stateMu     sync.RWMutex
 	state       State
 	failureCase error
-	listeners   []chan func(l Listener)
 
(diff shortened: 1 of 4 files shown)


pstibrany:
> LGTM! The `serviceListeners` looks a mere copy-paste of the previous logic in `BasicService`, so good 😉

It is! :)

pstibrany:
Let's not merge this. I think "Idle" service will be a better idea (with start/stop functions only), and Noop will be simply "Idle" with no functions. We already have "Idle" service, but with blocked goroutine, which isn't nice.

Apologies for wasting your time with review.

PR #2235, opened 2020-03-09
Updated CHANGELOG with 0.7.0 RC

pracucci:
**What this PR does**:
I'm going to cut Cortex `0.7.0-rc.0`. To avoid a pain with the changelog once Cortex `0.7.0` stable will be released, I would prefer if we could have all changelog entries added below `## master / unreleased` from now on.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+## 0.7.0-rc.0 / 2020-03-09
+
 * [CHANGE] Removed support for flags to configure schema. Further, the flag for specifying the config file (`-config-yaml`) has been deprecated. Please use `schema-config-file`. See https://cortexmetrics.io/docs/configuration/schema-configuration/ for more details on how to configure the schema using the YAML file. #2221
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+## 0.7.0-rc.0 / 2020-03-09
+
```
gouthamve: Can we make it clear that this release has a bunch of breaking changes, and that we are making those changes in anticipation of a 1.0 release soon which will offer much more stability when upgrading?

pracucci: Uhm, sure but this is not the Changelog for the Cortex 0.7 release. This is just to add a new marker in the changelog and having new PRs merged that will add the changelog log entry under `master / unreleased`. I will submit a different PR for the Cortex 0.7 release (and that PR will be merged once Cortex `0.7.0` release will be out). Makes sense?

pracucci:
Closing, since we have decided to go through the path of a pre-release for `v0.7.0-rc.0`

PR #2236, opened 2020-03-09
Release Cortex 0.7.0-rc.0

pracucci:
**What this PR does**:
In this PR I've prepared for the Cortex `0.7.0` release. Few notes:

- The changelog has been re-ordered to group breaking changes at the top
- Added a section at the beginning of the changelog outlining the primary breaking changes (the ones that will affect **any** user)
- The `VERSION` will be switched to `0.7.0` before merging (once `0.7.0-rc.0` will be confirmed to be stable)

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


docs/guides/cortex-with-cassandra.md
```diff
@@ -110,12 +110,12 @@ storage:
 ```
 
 The latest tag is not published for the Cortex docker image. Visit quay.io/repository/cortexproject/cortex 
-to find the latest stable version tag and use it in the command bellow. At the time of writing it was v0.6.1. 
+to find the latest stable version tag and use it in the command bellow (currently it is `v0.7.0`).
```
gouthamve: We're only doing the RC now. This will break

k8s/alertmanager-dep.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: alertmanager
-        image: quay.io/cortexproject/cortex:v0.6.1
+        image: quay.io/cortexproject/cortex:v0.7.0
```
gouthamve: All these will break because we're only doing the RC

PR #2238, opened 2020-03-10
Fixed cortex_distributor_ingester_query_failures_total metric if an error occurr in the stream.Recv()

pracucci:
**What this PR does**:
Fixed `cortex_distributor_ingester_query_failures_total` metric if an error occurr in the `stream.Recv()`.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -145,6 +145,7 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 			if err == io.EOF {
 				break
 			} else if err != nil {
+				ingesterQueryFailures.WithLabelValues(ing.Addr).Inc()
 				return nil, err
 			}
 


pkg/distributor/query.go
```diff
@@ -145,6 +145,7 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 			if err == io.EOF {
 				break
 			} else if err != nil {
+				ingesterQueryFailures.WithLabelValues(ing.Addr).Inc()
```
gouthamve: Can we have this check at a higher level, like after `replication.Do` returns?

pracucci: The metric tracks the failures per ingester. If we do after `replication.Do` returns, we don't have the detail about which ingester failed. Moreover, this metric should be specular to `ingesterQueries` metric, which is tracked within the same function.

Issue #2191, opened 2020-02-27
Add a prefix to chunk cache flags


tomwilkie:
Its the only one without a prefix, and thats confusing.

Issue #553, opened 2017-08-30
memcache option flags are inconsistent


bboreham:
We have `memcached.hostname`, `memcached.service` and `memcached.expiration`, but `memcache.write-back-goroutines` and `memcache.write-back-buffer`

Unless I'm missing some subtlety in the naming, Cortex should pick one version of the name (`d` or no `d`) and stick to it.

PR #2241, opened 2020-03-10
Prefix chunks cache flags.

gouthamve:
Fixes https://github.com/cortexproject/cortex/issues/2191 #553 

**Checklist**
- [x] Documentation added

### DIFF
--- a/docs/guides/caching.md
+++ b/docs/guides/caching.md
@@ -121,7 +121,7 @@ These are typically a few KB in size, and depend mostly on the duration and enco
 The chunk cache is a write-through cache - chunks are written to the cache as they are flushed to the chunk store.  This ensures the cache always contains the most recent chunks.
 Items stay in the cache indefinitely.
 
-The chunk cache should be configured on the **ingester**, **querier** and **ruler** using the flags with a prefix: `-store.chunks-cache`.
+The chunk cache should be configured on the **ingester**, **querier** and **ruler** using the flags with the prefix `-store.chunks-cache`.
 
 It is best practice to ensure the chunk cache is big enough to accommodate at least 24 hours of chunk data.
 You can use the following query (from the [cortex-mixin](https://github.com/grafana/cortex-jsonnet)) to estimate the required number of memcached replicas:
--- a/docs/guides/caching.md
+++ b/docs/guides/caching.md
@@ -41,9 +41,9 @@ If deploying Cortex on Kubernetes, Cortex should be pointed at a memcached [head
 The flags used to configure memcached are common for each caching caching opportunity, differentiated by a prefix:
 
 ```
--<prefix>.memcache.write-back-buffer int
+-<prefix>.cache.write-back-buffer int
     How many chunks to buffer for background write back. (default 10000)
--<prefix>.memcache.write-back-goroutines int
+-<prefix>.cache.write-back-goroutines int
     How many goroutines to use to write back to memcache. (default 10)
 -<prefix>.memcached.batchsize int
     How many keys to fetch in each batch.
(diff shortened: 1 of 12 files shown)


gouthamve:
Thanks @pracucci, I thought I got them all, but somehow missed 2. Fixed those two as well!

pkg/chunk/chunk_store.go
```diff
@@ -61,8 +61,8 @@ type StoreConfig struct {
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *StoreConfig) RegisterFlags(f *flag.FlagSet) {
-	cfg.ChunkCacheConfig.RegisterFlagsWithPrefix("", "Cache config for chunks. ", f)
-	f.BoolVar(&cfg.chunkCacheStubs, "store.chunk-cache-stubs", false, "If true, don't write the full chunk to cache, just a stub entry.")
+	cfg.ChunkCacheConfig.RegisterFlagsWithPrefix("store.chunks-cache.", "Cache config for chunks. ", f)
```
pracucci: This setting is disabled in YAML with the comment:
```
// Not visible in yaml because the setting shouldn't be common between ingesters and queriers
```

Could you explain me why? I can't understand the issue from the comment.

pkg/chunk/cache/background.go
```diff
@@ -32,8 +32,8 @@ type BackgroundConfig struct {
 
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
 func (cfg *BackgroundConfig) RegisterFlagsWithPrefix(prefix string, description string, f *flag.FlagSet) {
-	f.IntVar(&cfg.WriteBackGoroutines, prefix+"memcache.write-back-goroutines", 10, description+"How many goroutines to use to write back to memcache.")
-	f.IntVar(&cfg.WriteBackBuffer, prefix+"memcache.write-back-buffer", 10000, description+"How many key batches to buffer for background write-back.")
+	f.IntVar(&cfg.WriteBackGoroutines, prefix+"cache.write-back-concurrency", 10, description+"At what concurrency to write back to cache.")
```
pracucci: Have considered changing the prefix from `cache.` into `background.`? Two possible reasons:
1. In the YAML it's called `background:`
2. The resulting flags are something like `-store.chunks-cache.cache.write-back-concurrency` where we have `-cache.cache.` which looks a bit redundant

It's an option question, not a request for change.

k8s/memcached-dep.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: memcached
-        image: memcached:1.4.25
+        image: memcached:1.6.0
```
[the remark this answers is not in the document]
pracucci: I would advocate the same. Moving it to another PR, just to keep things separated.

docs/guides/caching.md
```diff
@@ -121,7 +121,7 @@ These are typically a few KB in size, and depend mostly on the duration and enco
 The chunk cache is a write-through cache - chunks are written to the cache as they are flushed to the chunk store.  This ensures the cache always contains the most recent chunks.
 Items stay in the cache indefinitely.
 
-The chunk cache should be configured on the **ingester**, **querier** and **ruler** using the flags without a prefix.
+The chunk cache should be configured on the **ingester**, **querier** and **ruler** using the flags with a prefix: `-store.chunks-cache`.
```
pracucci:
```suggestion
The chunk cache should be configured on the **ingester**, **querier** and **ruler** using the flags with the prefix `-store.chunks-cache`.
```

pkg/chunk/chunk_store.go
gouthamve: Added comments explaining this

PR #2242, opened 2020-03-10
Ingester gRPC servers should return context error if any

pracucci:
**What this PR does**:
Today I investigated the following warning log we see quite frequently on high-traffic clusters:
```
level=warn ts=2020-03-10T09:19:42.839335323Z caller=grpc_logging.go:54 method=/cortex.Ingester/QueryStream duration=981.401µs err="rpc error: code = Internal desc = transport: transport: the stream is done or WriteHeader was already called" msg="gRPC\n"
```

After some investigating, I found out this happens whenever a gRPC server calls `Send()` on a closed stream because the context has been closed (canceled / deadline exceeded). In this PR I propose to wrap the `Send()` in order to check whether the context has an error and, if so, returning that error instead of the cryptic `Send()` error.

After this PR changes the log is the following:
```
level=debug ts=2020-03-11T05:12:21.1804904Z caller=grpc_logging.go:53 method=/cortex.Ingester/QueryStream duration=301µs err="rpc error: code = Canceled desc = context canceled" msg="gRPC\n"
```

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/grpc/util.go
+++ b/pkg/util/grpc/util.go
@@ -1,0 +1,17 @@
+package grpc
+
+import (
+	"github.com/gogo/status"
+	"google.golang.org/grpc/codes"
+)
+
+// IsGRPCContextCanceled returns whether the input error is a GRPC error wrapping
+// the context.Canceled error.
+func IsGRPCContextCanceled(err error) bool {
+	s, ok := status.FromError(err)
+	if !ok {
+		return false
+	}
+
+	return s.Code() == codes.Canceled
+}
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -5,19 +5,18 @@ import (
 	"io"
 	"sort"
 
-	"github.com/gogo/status"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/weaveworks/common/instrument"
 	"github.com/weaveworks/common/user"
-	"google.golang.org/grpc/codes"
 
 	"github.com/cortexproject/cortex/pkg/ingester/client"
 	ingester_client "github.com/cortexproject/cortex/pkg/ingester/client"
 	"github.com/cortexproject/cortex/pkg/ring"
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/extract"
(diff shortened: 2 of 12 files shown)


pkg/util/grpc/util.go
```diff
@@ -0,0 +1,17 @@
+package grpc
+
+import (
+	"github.com/gogo/status"
+	"google.golang.org/grpc/codes"
+)
+
+// IsGRPCContextCanceled returns whether the input error is a GRPC error wrapping
+// the context.Canceled error.
+func IsGRPCContextCanceled(err error) bool {
```
pstibrany: Is this used anywhere?

pracucci: So far it's used here:
https://github.com/cortexproject/cortex/pull/2242/files#diff-5c63c6fec32b3d7499095ed68c1ea1caR150

pracucci:
> Is it worth bit of extra complexity? I’m not sure.

A couple of things:
1. The `the stream is done or WriteHeader was already called`  may also be caused by bugs and it's difficult (if not impossible) distinguish if it's a bug in our code or the context done
2. Once we upgrade `weaveworks/common` the `context canceled` log will be a debug so that we don't log it as warning (while we couldn't safely switch all `the stream is done or WriteHeader was already called` to debug level

pkg/util/grpc/util.go
pstibrany: Thanks. I clicked “viewed” checkbox and then couldn’t find it. 🤦‍♂

pstibrany:
> 1. The `the stream is done or WriteHeader was already called`  may also be caused by bugs and it's difficult (if not impossible) distinguish if it's a bug in our code or the context done

👍 

> 2. Once we upgrade `weaveworks/common` the `context canceled` log will be a debug so that we don't log it as warning (while we couldn't safely switch all `the stream is done or WriteHeader was already called` to debug level

Looking forward to that!

pracucci:
@pstibrany I've merged PR #2243 (upgrade `weaveworks/common`) and rebased `master` and I can confirm the log level is now `debug` (being a `context canceled` error). I've also added a unit test on one of the helpers. Do you have any further comment or concern about this PR

pstibrany:
> @pstibrany I've merged PR #2243 (upgrade `weaveworks/common`) and rebased `master` and I can confirm the log level is now `debug` (being a `context canceled` error). I've also added a unit test on one of the helpers. Do you have any further comment or concern about this PR?

No, thanks!

PR #2083, opened 2020-02-05
Experimental TSDB: Enable Azure Storage Backend

khaines:
**What this PR does**: Adds the needed cortex structs and configuration to enable azure storage usage with the experimental tsdb engine. 

I've got a small cluster using this currently as it's storage backend, but it's still relatively new so I would say testing for this over the long term is still ongoing.

**Checklist**
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2245, opened 2020-03-10
Experimental TSDB: Add support for local filesystem backend

slim-bean:
**What this PR does**:

Add support for TSDB blocks to be saved to the local filesystem via the implementation which already exists for this in Thanos.

Used #2083 as a template

**Checklist**
- [ ] Tests updated
- [x] Documentation added

Signed-off-by: Edward Welch <edward.welch@grafana.com>

### DIFF
--- a/pkg/storage/tsdb/backend/filesystem/bucket_client.go
+++ b/pkg/storage/tsdb/backend/filesystem/bucket_client.go
@@ -1,0 +1,12 @@
+package filesystem
+
+import (
+	"github.com/go-kit/kit/log"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/objstore/filesystem"
+)
+
+// NewBucketClient creates a new filesystem bucket client
+func NewBucketClient(cfg Config, name string, logger log.Logger) (objstore.Bucket, error) {
+	return filesystem.NewBucket(cfg.Directory)
+}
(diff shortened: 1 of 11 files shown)


pkg/storage/tsdb/backend/filesystem/bucket_client.go
```diff
@@ -0,0 +1,12 @@
+package filesystem
+
+import (
+	"github.com/go-kit/kit/log"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/objstore/filesystem"
+)
+
+// NewBucketClient creates a new filesystem bucket client
+func NewBucketClient(cfg Config, name string, logger log.Logger) (objstore.Bucket, error) {
```
pstibrany: I would leave out unused parameters completely. Or name them `_` to indicate that they are unused.

pstibrany:
> What's the use case for having this in Cortex?

As Marco points out, one can simply disable shipping to the external storage, which is effectively the same.

pracucci:
I talked to Ed offline. Ed's use case is running a single-node Cortex offline and then having a periodic rsync to synch blocks to secondary Cortex cluster. Despite this can be done disabling the shipper and having an rsync of blocks (not WAL), from a UX perspective looks more cleaner using the `filesystem` storage backend and having an `rsync` with the filesystem `dir` as source. Moreover the shipper manipulates the `meta.json` (ie. adds the external labels) and looks the safest approach in this case.

pracucci:
> I think the use case is for single node cortex, because cortex can run on arm arch and not prometheus.

My take is that for a single node Cortex with blocks storage you don't need the `filesystem` support. You can just run Cortex single binary disabling the shipper.

cyriltovena:
Yeah I don't have enough knowledge on this !

PR #2246, opened 2020-03-11
Converted frontend worker to a service.

pstibrany:
Issue: The query frontend worker in the querier was previously (before this PR) started in the init function. The side effect was that it started pulling jobs from the queue before the queryable is ready. This means queries fail during the rollout, until the queryable is ready.

This PR changes that. Frontend worker is now a service, and it is only started after all Querier module dependencies are started. This includes queryable.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master / unreleased
 
 * [CHANGE] Utilize separate protos for rule state and storage. Experimental ruler API will not be functional until the rollout is complete. #2226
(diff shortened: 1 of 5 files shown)


pstibrany:
> LGTM, thanks! I just think this fix is worth a CHANGELOG entry.

Thanks for review! Added changelog entry now.

pstibrany:
cc @cyriltovena this also affects Loki, which uses frontend worker directly. Post 0.7 as well.

(I'll start working on Loki integration soon, but just so you know about it)

PR #2249, opened 2020-03-11
add support for bigtable as index store for integration tests

sandeepsukhani:
**What this PR does**:
Add support for Bigtable in integration tests using https://github.com/Shopify/bigtable-emulator
It also adds a simple test for all the supported index stores in integration tests while keeping `dynamo-db` as default for rest of the tests.

### DIFF
--- a/integration/all_stores_test.go
+++ b/integration/all_stores_test.go
@@ -62,7 +62,9 @@ func TestAllIndexStores(t *testing.T) {
 	// Start rest of the Cortex components.
 	require.NoError(t, writeFileToSharedDir(s, cortexSchemaConfigFile, []byte(buildSchemaConfigWith(storeConfigs))))
 
-	ingester := e2ecortex.NewIngester("ingester", consul.NetworkHTTPEndpoint(), ChunksStorageFlags, "")
+	ingester := e2ecortex.NewIngester("ingester", consul.NetworkHTTPEndpoint(), mergeFlags(ChunksStorageFlags, map[string]string{
+		"-ingester.retain-period": "0s", // we want to make ingester not retain any chunks in memory after they are flushed so that queries get data only from the store
+	}), "")
 	ingester.HTTPService.SetEnvVars(bigtableFlag)
 
 	distributor := e2ecortex.NewDistributor("distributor", consul.NetworkHTTPEndpoint(), ChunksStorageFlags, "")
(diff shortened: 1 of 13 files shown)


sandeepsukhani:
> LGTM. I think number of parameters to various `New<Service>` functions is so high now that we should refactor it to use [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) (in new PR)

Yeah, right. I will add it to my todo list.
Thanks for the review!

integration/configs.go
```diff
@@ -99,6 +105,9 @@ storage:
   aws:
     dynamodbconfig:
       dynamodb: {{.DynamoDBURL}}
+  bigtable:
```
pracucci: By contract, `ChunksStorageConfig` should be specular to `ChunksStorageFlags`, so if you do a change to `ChunksStorageConfig` you should do the same change to `ChunksStorageFlags` and viceversa.

However, these configs where designed to be ready to use and should not have mixed config (either the storage is AWS or BigTable). I would suggest to:
1. Rename `ChunksStorageConfig` to `ChunksStorageDynamoDBConfig`
2. Rename `ChunksStorageFlags` to `ChunksStorageDynamoDBFlags`
3. Create `ChunksStorageBigtableConfig` and `ChunksStorageBigtableFlags` with the config you want

integration/e2ecortex/services.go
```diff
@@ -51,14 +51,15 @@ func NewDistributorWithConfigFile(name, consulAddress, configFile string, flags
 		e2e.NewReadinessProbe(httpPort, "/ready", 204),
 		httpPort,
 		grpcPort,
+		nil,
 	)
 }
 
-func NewQuerier(name string, consulAddress string, flags map[string]string, image string) *CortexService {
-	return NewQuerierWithConfigFile(name, consulAddress, "", flags, image)
+func NewQuerier(name string, consulAddress string, flags map[string]string, image string, envVars map[string]string) *CortexService {
```
pracucci: I'm not sure it's worth adding `envVars` to all such functions, considering it's not a commong use case and you could just call `service.SetEnvVars()` in the single integration test where you need it. What's your take?

integration/all_stores_test.go
```diff
@@ -0,0 +1,94 @@
...
+	for i, store := range stores {
+		storeConfigs[i] = storeConfig{From: oldestStoreStartTime.Add(time.Duration(i) * perStoreDuration).Format("2006-01-02"), IndexStore: store}
+	}
+
+	// bigtable client needs to set an environment variable when connecting to an emulator
+	bigtableFlag := map[string]string{"BIGTABLE_EMULATOR_HOST": bigtable.NetworkHTTPEndpoint()}
+
+	// here we are starting and stopping table manager for each index store
+	// this is a workaround to make table manager create tables for each config since it considers only latest schema config while creating tables
+	for i := range storeConfigs {
```
pracucci: This looks a bit hacky. Why don't we run the table manager only once, but with a schema config containing both storages (two entries in the same schema config)? Doing it with `buildSchemaConfigWith()` may be engineered (would require extra refactoring I'm not sure it's worth). You could just rollback `buildSchemaConfigWith()` and have a static schema config defined in this file.

integration/e2ecortex/services.go
sandeepsukhani: I had started with that approach first but then it made me feel someone could miss setting that env var, while setting all the params in the function is explicitly done. But I think we would mostly continue using dynamo for all the other test so I can use that function for now. We can refactor it again later if we see more needs for setting environment variables.
Thanks!

pracucci: > then it made me feel someone could miss setting that env var

It would miss anyway if that person is not aware of the required env var. If you want to guaraantee it you should take `bigtableAddress` in input, like we do for `consulAddress`, but to me looks a bit over-engineered adding `bigtableAddress` to each of such functions given it's only required by 1 single scenario (contrary to consul).

integration/all_stores_test.go
sandeepsukhani: This is done because `TableManager` only uses latest config to create tables i.e it does not consider older configs while creating tables. There is a PR open to fix this behaviour in Cortex. See https://github.com/cortexproject/cortex/pull/1446

We have not seen someone complain about it because we add only one schema at a time to cortex as and when needed but here I am setting multiple schema configs at a time.
Does it make sense?

integration/configs.go
sandeepsukhani: It seems I can get rid of this change completely because it is not used in the new test that I added.

integration/all_stores_test.go
pracucci: Yes. I wasn't aware of the issue. Could you add a `// TODO` comment referencing the PR and saying that we can merge two two schema configs into 1 and just run the table-manager once once that PR is merged, please?

integration/all_stores_test.go
```diff
@@ -0,0 +1,98 @@
...
+		series, expectedVector := generateSeries("series_1", ts)
+
+		c, err := e2ecortex.NewClient(distributor.HTTPEndpoint(), "", "", "user-1")
+		require.NoError(t, err)
+
+		res, err := c.Push(series)
+		require.NoError(t, err)
+		require.Equal(t, 200, res.StatusCode)
+
+		// Query the series both from the querier and query-frontend (to hit the read path).
```
pracucci:
```suggestion
		// Query back the series.
```

integration/all_stores_test.go
```diff
@@ -0,0 +1,98 @@
...
+		c, err = e2ecortex.NewClient("", querier.HTTPEndpoint(), "", "user-1")
```
pracucci: This test should hit the storage, but how is it guaranteed that the series have been flushed to the storage and offloaded from the ingesters memory at this point?

I've the feeling we're not hitting the backend store neither on the write and read path, because the chunks are just kept in the ingesters memory.

PR #2250, opened 2020-03-11
Make module service wrapper reusable from other projects

pstibrany:
This PR moves `moduleServiceWrapper` from Cortex main package to util package. It is now available via `NewModuleService` function, which is also bit more generic, to make it reusable between projects. (Going to use it in Loki)

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/module_service.go
+++ b/pkg/util/module_service.go
@@ -75,6 +75,10 @@ func (w *moduleService) stop(_ error) error {
 	// wait until all stopDeps have stopped
 	stopDeps := w.stopDeps(w.name)
 	for _, s := range stopDeps {
+		if s == nil {
+			continue
+		}
+
 		// Passed context isn't canceled, so we can only get error here, if service
 		// fails. But we don't care *how* service stops, as long as it is done.
 		_ = s.AwaitTerminated(context.Background())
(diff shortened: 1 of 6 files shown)


pkg/util/module_service.go
```diff
@@ -0,0 +1,92 @@
...
+	// wait until service stops, or context is canceled, whatever happens first.
+	// We don't care about exact error here
+	_ = w.service.AwaitTerminated(serviceContext)
+	return w.service.FailureCase()
+}
+
+func (w *moduleService) stop(_ error) error {
+	// wait until all stopDeps have stopped
+	stopDeps := w.stopDeps(w.name)
+	for _, s := range stopDeps {
```
pracucci: Shouldn't we check if `s == nil` here as well? To me it should be the same of `start()`. In Cortex, it's not an issue because in `module_service_wrapper.go` you filter out `nil` deps, but being a generic utility I would consider safer to add the check here as well.

Am I missing anything?

pstibrany: You're right. I think I planned to *remove* nil check from both places originally, but having it is a bit safer.

pstibrany:
Thanks for review Marco, I've addressed both of your comments.

PR #2251, opened 2020-03-11
Lifecycler errors don't make process exit immediately, but let other modules shutdown properly

pstibrany:
**What this PR does**: This PR does two things:
- removes `os.Exit` calls from lifecycler. Lifecycler now enters Failed state, and reports problem via failure case.
- components that use `Lifecycler` now watch it for failures, and if Lifecycler fails, these components enter Failed state as well. To do that, new `ServiceFailureWatcher` was introduced. It  can watch either a service, or service manager.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
cc @rfratto @cyriltovena – you may want to know about this. If merged, it will be post-0.7.

pkg/compactor/compactor.go
```diff
@@ -87,7 +87,8 @@ type Compactor struct {
 	ring           *ring.Ring
 
 	// Subservices manager (ring, lifecycler)
-	subservices *services.Manager
+	subservices    *services.Manager
+	serviceWatcher *util.ServiceFailureWatcher
```
pracucci: Have you considered naming it `subservicesWatcher`?

pkg/util/service_failure_watcher.go
```diff
@@ -0,0 +1,37 @@
+package util
```
pracucci: Shouldn't be part of the `pkg/util/services` framework? Looks generic enough to be put there.

pkg/util/service_failure_watcher.go
```diff
@@ -0,0 +1,37 @@
...
+func (w *ServiceFailureWatcher) Chan() <-chan error {
+	if w == nil {
+		return nil
+	}
+	return w.ch
+}
+
+func (w *ServiceFailureWatcher) WatchService(service services.Service) {
+	service.AddListener(services.NewListener(nil, nil, nil, nil, func(from services.State, failure error) {
+		w.ch <- errors.Wrapf(failure, "service %v failed", service)
```
pracucci: What if, due to misuse or bug, no gorouting is reading on the chan? Given each listener has its own goroutine, this shouldn't be a problem. May you just confirm it, please?

pstibrany: That is correct. Goroutine handling the listener events will be blocked forever though. We can use some small buffer to help with that [although it can still occur, if more errors happen], or ignore it, as we expect to server to shutdown soon after anyway.

pkg/util/service_failure_watcher.go
pstibrany: You're right. I moved it and renamed to FailureWatcher (`services.FailureWatcher` in full).

pstibrany:
Thanks for review, I accepted all your suggestions.

PR #2244, opened 2020-03-10
WIP: Support 1w, 1y for retention and table period

gouthamve:
This will let us `1w` in the schema config YAML and also lets us set the retention as `2y` etc. If this looks good, I'll go ahead and change docs, etc.

PR #2252, opened 2020-03-11
Support 1w, 1y for retention and table period take 2

gouthamve:
**What this PR does**:
Support 1w and 1y for retention and table period. Will change docs and add changelog entries if this approach looks good as opposed to #2244 

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/integration/configs.go
+++ b/integration/configs.go
@@ -22,10 +22,10 @@ const (
   schema: v9
   index:
     prefix: cortex_
-    period: 168h0m0s
+    period: 1w 
   chunks:
     prefix: cortex_chunks_
-    period: 168h0m0s
+    period: 1w 
 `
 
 	cortexAlertmanagerUserConfigYaml = `route:
@@ -91,7 +91,7 @@ tsdb:
 		"-dynamodb.url":                   fmt.Sprintf("dynamodb://u:p@%s-dynamodb.:8000", networkName),
 		"-dynamodb.poll-interval":         "1m",
 		"-schema-config-file":             filepath.Join(e2e.ContainerSharedDir, cortexSchemaConfigFile),
-		"-table-manager.retention-period": "168h",
+		"-table-manager.retention-period": "1w",
 	}
 
 	ChunksStorageConfig = buildConfigFromTemplate(`
@@ -102,7 +102,7 @@ storage:
 
 table_manager:
   dynamodb_poll_interval: 1m
-  retention_period:       168h
+  retention_period:       1w 
 
 schema:
 {{.SchemaConfig}}
--- a/integration/configs.go
+++ b/integration/configs.go
@@ -22,10 +22,10 @@ const (
   schema: v9
   index:
     prefix: cortex_
-    period: 1w 
+    period: 168h 
   chunks:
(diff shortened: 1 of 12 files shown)


pkg/chunk/schema_config.go
```diff
@@ -295,6 +295,39 @@ type PeriodicTableConfig struct {
 	Tags   Tags          `yaml:"tags,omitempty"`
```
pracucci: I think you can remove all `yaml` tags here because you have a custom marshalling. If I'm not missing anything, leaving them here could lead to misunderstanding to future readers.

pkg/chunk/table_manager.go
```diff
@@ -94,6 +96,34 @@ type TableManagerConfig struct {
...
+func (cfg *TableManagerConfig) MarshalYAML() (interface{}, error) {
+	cfg.RetentionPeriodModel = model.Duration(cfg.RetentionPeriod)
+	return cfg, nil
+}
+
+// Validate validates the config.
+func (cfg *TableManagerConfig) Validate() error {
+	if cfg.RetentionPeriodModel > 0 {
+		cfg.RetentionPeriod = time.Duration(cfg.RetentionPeriodModel)
+	}
```
pracucci: Why do you need this? Aren't you already covered by the custom marshalling? I'm asking because having business logic here breaks the `Validate()` contract which is expected to not manipulate the config and just return an error if it's valid.

pkg/chunk/schema_config.go
```diff
@@ -295,6 +295,39 @@ type PeriodicTableConfig struct {
...
 }
 
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (cfg *PeriodicTableConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
pracucci: Could you also add a unit test covering the unmarshalling with a non standard duration (ie. `1y`)?

pkg/chunk/table_manager.go
gouthamve: This is for flags. I don't know of a good way to set the right flag.

pracucci: Right. The only idea I have is to add a custom data type for the flag and having the `Set()` writing to both `cfg.RetentionPeriodModel` and `cfg.RetentionPeriod` (this one copying the value from the model) but it adds extra complexity and it's not worth to me.

So, I would just suggest you to add a comment before this `if` to explain why we do it (I didn't get it at first).

integration/configs.go
```diff
@@ -22,10 +22,10 @@ const (
   schema: v9
   index:
     prefix: cortex_
-    period: 168h0m0s
```
bboreham: Why was this changed?

gouthamve: Missed this when I changed `1w` to `168h` when I wanted older versions to run in integration tests: https://github.com/cortexproject/cortex/pull/2252/commits/d5f8f81a7a767149d22694bbe50f95ea58208081

PR to update changelog here: https://github.com/cortexproject/cortex/pull/2265

PR #2263, opened 2020-03-12
Deduplicate streamed samples from ingesters using blocks storage

pstibrany:
**What this PR does**: this PR adds deduplication of streamed samples coming from different ingesters that use blocks storage.

Merge function stores result into first slice to avoid extra allocations if not needed. We can use more straightforward merge implementation.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -3,7 +3,6 @@ package distributor
 import (
 	"context"
 	"io"
-	"sort"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
@@ -183,7 +182,11 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 			hash := client.FastFingerprint(series.Labels)
 			existing := hashToTimeSeries[hash]
 			existing.Labels = series.Labels
-			existing.Samples = append(existing.Samples, series.Samples...)
+			if existing.Samples == nil {
+				existing.Samples = series.Samples
+			} else {
+				existing.Samples = mergeSamplesIntoFirst(existing.Samples, series.Samples)
+			}
 			hashToTimeSeries[hash] = existing
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/distributor/query.go
```diff
@@ -196,15 +199,33 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
...
-		sort.Sort(byTimestamp(series.Samples))
 		resp.Timeseries = append(resp.Timeseries, series)
 	}
 
 	return resp, nil
 }
 
-type byTimestamp []client.Sample
+// Merges and dedupes two sorted slices with samples together, storing result in the first slice.
+// If a and b overlap, weird things will happen.
```
gouthamve: What does `if a and b overlap` mean? If we're deduping, then they're overlapping?

I had to read the code to understand that you meant the slice addresses overlapping, but best to make it clear in the comment?

pkg/distributor/query.go
```diff
@@ -196,15 +199,33 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
...
+func mergeSamplesIntoFirst(a, b []ingester_client.Sample) []ingester_client.Sample {
+	i, j := 0, 0
+	for i < len(a) && j < len(b) {
+		if a[i].TimestampMs < b[j].TimestampMs {
+			i++
+		} else if a[i].TimestampMs > b[j].TimestampMs {
+			a = append(a, ingester_client.Sample{}) // make sure 'a' has enough space
+			copy(a[i+1:], a[i:])                    // shift elements in 'a'
```
gouthamve: If `b` in entirely smaller than `a`, then there is a lot of copying going on. How do we want to deal with that? How expensive is this copying?

pstibrany: Primary reason to use existing slice was that most typical scenario is merging two slices with same content – an extra allocation seemed pointless. We can also be smart about it, and simply check if they are really the same, and if not, do the standard merge. With no extra copying inside the slice [which can arguably make some elements to be copied too many times].

pstibrany:
> LGTM overall, just a little concerned about the copying. Can we reuse the heap merge iterator stuff?

Can you be specific? We do mostly the same work here as we do elsewhere, but they operate on different structures :-(

pkg/distributor/query.go
pstibrany: I've modified the code, and removed the comment.

pstibrany:
I've changed the code to use standard merging, but it also checks if samples are all the same, and skips merging them if they are.

PR #2265, opened 2020-03-12
Update breaking change around mixed time units

gouthamve:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Don't support mixed time units anymore for duration. For example, 168h5m0s doesn't work anymore, please use just one unit (s|m|h|d|w|y). #2252
```
bboreham: This doesn't seem to be uniformly true - for instance I have `-ingester.retain-period=10m10s`.
I guess it's ok to simplify the message.

gouthamve: Yeah, I can mention its only for table period and retention, but I think it'll get confusing.

PR #2266, opened 2020-03-12
Compile code in vendor dir; don't download modules again

### LINKED ISSUE #1710: Cortex is not using vendored dependencies in CI
During builds Cortex is not using its vendored dependencies.  This probably changed in the migration to 1.13 due to changes in the way it handled go modules:

https://github.com/golang/go/wiki/Modules#recent-changes

> GO111MODULE=auto enables module-mode if any go.mod is found, even inside GOPATH. 

Example circleci output from a recent PR:
```
CGO_ENABLED=0 go build -ldflags "-extldflags \"-static\" -s -w" -tags netgo -o cmd/cortex/cortex ./cmd/cortex
go: downloading gopkg.in/yaml.v2 v2.2.2
go: downloading github.com/julienschmidt/httprouter v1.2.0
go: downloading github.com/jonboulle/clockwork v0.1.0
go: downloading github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f
go: downloading github.com/hashicorp/consul/api v1.1.0
```

### LINKED ISSUE #1795: metadataInIndex flag is broken
Since @tomwilkie asked about this from the stage in PromCon...

As noted in #873:

> The big casualty here is removing the support for metadataInIndex chunks 

So we cannot read data written in 2016.

Given `Chunk.metadataInIndex` is never set we could remove all mention of it from the code.

bboreham:
**What this PR does**:
Tells Go to use the code we have checked in under `vendor`.

Fixes #1710 
Fixes #1795

**Checklist**
N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,8 +118,6 @@ jobs:
     machine:
       image: ubuntu-1604:201903-01
 
-    # Make sure the working directory is within the GOPATH, otherwise
-    # "go test" will not download module dependencies.
     working_directory: ~/.go_workspace/src/github.com/cortexproject/cortex
     steps:
     - run:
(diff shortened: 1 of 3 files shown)


bboreham:
Turns out it is a bad idea to make a branch with `=` in the name.

PR #2268, opened 2020-03-12
Allow more connection reuse for DynamoDB and S3 calls

bboreham:
**What this PR does**:
We will connect many times in parallel to the same DynamoDB server, and with default settings Go will close and re-open connections; see https://github.com/golang/go/issues/13801

Raise MaxIdleConnsPerHost to avoid this.

**Checklist**
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/chunk/aws/dynamodb_storage_client.go
+++ b/pkg/chunk/aws/dynamodb_storage_client.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net"
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -865,5 +867,24 @@ func awsSessionFromURL(awsURL *url.URL) (client.ConfigProvider, error) {
...
 	return session.NewSession(config)
 }
+
+// Copy-pasted http.DefaultTransport
+var defaultTransport http.RoundTripper = &http.Transport{
+	Proxy: http.ProxyFromEnvironment,
+	DialContext: (&net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+		DualStack: true,
```
pracucci: This is deprecated. Given now it's enabled by default, this option can be removed.

bboreham: Interesting - why doesn't the linter show it as deprecated?

pracucci: I have no idea. I spent an hour on it, without any result. I can't get the deprecated check working (which should be done by `staticcheck`).

bboreham: I removed the line.

PR #2269, opened 2020-03-12
Ruler API and ObjectStore Backend

jtlisi:
### API Changes

This PR adds an API to the ruler service based on this design document (https://docs.google.com/document/d/1gNKiC5L_zWQ4GGgtdhJJ6FnaZXyJJBvdGTUmF0wgEFg/edit?usp=sharing)

### Object Storage Changes

* It creates a generic storage client for rules based on our `chunk.ObjectClient` interface
* The delimiter was made a configurable option for the object storage clients to make this possible. The delimiter is set at the construction of the object storage client

### Integration Tests

* A ruler integration test was added that ensures a rule group can be set, is running in a Prometheus manager, and can be deleted
* A ruler address was added to the integration test cortex client

Related #1547 #1244

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md`

### DIFF
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -291,6 +291,8 @@ var (
 	ErrNoRuleGroups = errors.New("no rule groups found")
 	// ErrNoUserID is returned when no user ID is provided
 	ErrNoUserID = errors.New("no id provided")
+	// ErrBadRuleGroup is returned when the provided rule group can not be unmarshalled
+	ErrBadRuleGroup = errors.New("unable to decoded rule group")
 )
 
 // ValidateRuleGroup validates a rulegroup
@@ -316,29 +318,58 @@ func ValidateRuleGroup(g rulefmt.RuleGroup) []error {
 	return errs
 }
 
-func (r *Ruler) listRules(w http.ResponseWriter, req *http.Request) {
-	logger := util.WithContext(req.Context(), util.Logger)
-	userID, _, err := user.ExtractOrgIDFromHTTPRequest(req)
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusUnauthorized)
-		return
+// parseNamespace parses the namespace from the provided set of params, in this
+// api these params are derived from the url path
+func parseNamespace(params map[string]string) (string, error) {
+	namespace, exists := params["namespace"]
+	if !exists {
+		return "", ErrNoNamespace
 	}
 
-	if userID == "" {
-		http.Error(w, ErrNoUserID.Error(), http.StatusUnauthorized)
+	namespace, err := url.PathUnescape(namespace) // namespaces needs to be unescaped if in the URL
+	if err != nil {
+		return "", err
+	}
+
+	return namespace, nil
+}
+
+// parseGroupName parses the group name from the provided set of params, in this
+// api these params are derived from the url path
+func parseGroupName(params map[string]string) (string, error) {
+	groupName, exists := params["groupName"]
+	if !exists {
+		return "", ErrNoGroupName
+	}
+
+	groupName, err := url.PathUnescape(groupName) // groupName needs to be unescaped if in the URL
+	if err != nil {
+		return "", err
+	}
+
+	return groupName, nil
+}
+
+func (r *Ruler) listRules(w http.ResponseWriter, req *http.Request) {
+	logger := util.WithContext(req.Context(), util.Logger)
+	userID, err := user.ExtractOrgID(req.Context())
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
 	vars := mux.Vars(req)
 
-	namespace := vars["namespace"]
-	if namespace != "" {
-		namespace, err = url.PathUnescape(namespace) // namespaces need to be unescaped if in the URL
-		if err != nil {
+	// Parse the namespace from the url path parameters and return
+	// a 400 if it is invalid and return a full set of rules if no
+	// namespace is provided
+	namespace, err := parseNamespace(vars)
+	if err != nil {
+		// If a namespace is not provided continue as usual and return a full set of rules
+		if err != ErrNoNamespace {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
-
 		level.Debug(logger).Log("msg", "retrieving rule groups with namespace", "userID", userID, "namespace", namespace)
 	}
 
@@ -376,38 +407,25 @@ func (r *Ruler) listRules(w http.ResponseWriter, req *http.Request) {
 
 func (r *Ruler) getRuleGroup(w http.ResponseWriter, req *http.Request) {
 	logger := util.WithContext(req.Context(), util.Logger)
-
-	userID, _, err := user.ExtractOrgIDFromHTTPRequest(req)
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusUnauthorized)
-		return
-	}
-
-	if userID == "" {
-		http.Error(w, ErrNoUserID.Error(), http.StatusUnauthorized)
-		return
-	}
-
-	vars := mux.Vars(req)
-	namespace, exists := vars["namespace"]
-	if !exists {
-		http.Error(w, ErrNoNamespace.Error(), http.StatusUnauthorized)
-		return
-	}
-
-	namespace, err = url.PathUnescape(namespace) // namespaces need to be unescaped if in the URL
+	userID, err := user.ExtractOrgID(req.Context())
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
-	groupName, exists := vars["groupName"]
-	if !exists {
-		http.Error(w, ErrNoGroupName.Error(), http.StatusUnauthorized)
+	vars := mux.Vars(req)
+
+	// Parse the namespace from the url path parameters and return
+	// a 400 if it is invalid
+	namespace, err := parseNamespace(vars)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
-	groupName, err = url.PathUnescape(groupName) // groupName need to be unescaped if in the URL
+	// Parse the rule group name from the url path parameters and return
+	// a 400 if it is invalid
+	groupName, err := parseGroupName(vars)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
@@ -442,27 +460,15 @@ func (r *Ruler) getRuleGroup(w http.ResponseWriter, req *http.Request) {
 
 func (r *Ruler) createRuleGroup(w http.ResponseWriter, req *http.Request) {
 	logger := util.WithContext(req.Context(), util.Logger)
-	userID, _, err := user.ExtractOrgIDFromHTTPRequest(req)
+	userID, err := user.ExtractOrgID(req.Context())
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusUnauthorized)
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
-	if userID == "" {
-		http.Error(w, ErrNoUserID.Error(), http.StatusUnauthorized)
-		return
-	}
-
-	vars := mux.Vars(req)
-
-	namespace := vars["namespace"]
-	if namespace == "" {
-		level.Error(logger).Log("err", "no namespace provided with rule group")
-		http.Error(w, ErrNoNamespace.Error(), http.StatusBadRequest)
-		return
-	}
-
-	namespace, err = url.PathUnescape(namespace) // namespaces need to be unescaped if in the URL
+	// Parse the namespace from the url path parameters and return
+	// a 400 if it is invalid
+	namespace, err := parseNamespace(mux.Vars(req))
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
@@ -481,7 +487,7 @@ func (r *Ruler) createRuleGroup(w http.ResponseWriter, req *http.Request) {
 	err = yaml.Unmarshal(payload, &rg)
 	if err != nil {
 		level.Error(logger).Log("err", err.Error())
-		http.Error(w, err.Error(), http.StatusBadRequest)
+		http.Error(w, ErrBadRuleGroup.Error(), http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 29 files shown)


pkg/ruler/api.go
```diff
@@ -2,33 +2,47 @@ package ruler
...
+	rulefmt "github.com/cortexproject/cortex/pkg/ruler/legacy_rulefmt"
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
+	store "github.com/cortexproject/cortex/pkg/ruler/rules"
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
 // RegisterRoutes registers the ruler API HTTP routes with the provided Router.
-func (r *Ruler) RegisterRoutes(router *mux.Router) {
+func (r *Ruler) RegisterRoutes(router *mux.Router, middleware middleware.Interface) {
+	router = router.UseEncodedPath()
```
pstibrany: This needs a comment, because it's quite hidden here, and also because later in the code we use `url.PathUnescape`. If someone accidentaly removes this line, he will break `url.PathUnescape` later in the code possibly without realizing the connection.

Handling paths with escaped characters in namespace and group also needs a unit test.

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+// parseNamespace parses the namespace from the provided set of params, in this
+// api these params are derived from the url path
+func parseNamespace(params map[string]string) (string, error) {
+	namespace, exists := params["namespace"]
+	if !exists {
+		return "", ErrNoNamespace
+	}
+
+	namespace, err := url.PathUnescape(namespace) // namespaces needs to be unescaped if in the URL
```
pstibrany: I would instead mention using `router.UseEncodedPath()` in the routes setup, same in `parseGroupName`.

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		// If a namespace is not provided continue as usual and return a full set of rules
+		if err != ErrNoNamespace {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		level.Debug(logger).Log("msg", "retrieving rule groups with namespace", "userID", userID, "namespace", namespace)
+	}
+
+	level.Debug(logger).Log("msg", "retrieving rule groups from rule store", "userID", userID)
```
pstibrany: These two log messages are the same, except first one tries to log *missing* namespace, while second one doesn't log namespace at all (even though it actually can exist there).

I suggest replacing second log with first one, and remove first one.

```suggestion
	}

		level.Debug(logger).Log("msg", "retrieving rule groups from rule store", "userID", userID, "namespace", namespace)
```

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		level.Error(logger).Log("msg", "error marshalling yaml rule groups", "err", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/yaml")
+	if _, err := w.Write(d); err != nil {
+		level.Error(logger).Log("msg", "error writing yaml response", "err", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
pstibrany: If write failed, there is no need to try sending another status code and error message to the client.

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
```
pstibrany: Can we extract `marshalAndSend` method, and reuse it?

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+	// Parse the rule group name from the url path parameters and return
+	// a 400 if it is invalid
+	groupName, err := parseGroupName(vars)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
```
pstibrany: Can we extract all of this into a method that takes request, response, and boolean flags if namespace and group are required, and returns bool (true if required parameters were set, false if error has already been sent to client), userID, optional namespace and group? This same code is repeated over and over.

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+	errs := ValidateRuleGroup(rg)
+	if len(errs) > 0 {
+		level.Error(logger).Log("err", err.Error())
+		http.Error(w, errs[0].Error(), http.StatusBadRequest)
+		return
+	}
+
+	rgProto := store.ToProto(userID, namespace, rg)
+
+	level.Debug(logger).Log("msg", "attempting to store rulegroup", "userID", userID, "group", rgProto.String())
```
pstibrany: `rgProto.String()` can be huge here, do we really want to log it? Same for logging payload above.

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -0,0 +1,164 @@
+package objectclient
```
pstibrany: Why extra package?

pkg/ruler/api.go
jtlisi: Since this log is debug level, I think it should be ok?

pkg/ruler/rules/objectclient/rule_store.go
jtlisi: This isn't consistent with the other store in this package. However, this is consistent with how the `pkg/alertmanager/alerts` package has it's storage clients. It's also similar to `pkg/chunks/` does. In a future PR I will nest the configdb rulestore client into a separate package as well.

pkg/ruler/api.go
jtlisi: 👍 makes sense

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+var (
+	// ErrNoNamespace signals the requested namespace does not exist
+	ErrNoNamespace = errors.New("a namespace must be provided in the url")
+	// ErrNoGroupName signals a group name url parameter was not found
+	ErrNoGroupName = errors.New("a matching group name must be provided in the url")
+	// ErrNoRuleGroups signals the rule group requested does not exist
+	ErrNoRuleGroups = errors.New("no rule groups found")
+	// ErrNoUserID is returned when no user ID is provided
+	ErrNoUserID = errors.New("no id provided")
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
 		level.Error(logger).Log("msg", "error writing response", "bytesWritten", n, "err", err)
 	}
 }
+
+var (
+	// ErrNoNamespace signals the requested namespace does not exist
+	ErrNoNamespace = errors.New("a namespace must be provided in the url")
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/api.go
```diff
@@ -268,3 +282,274 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
 		level.Error(logger).Log("msg", "error writing response", "bytesWritten", n, "err", err)
 	}
 }
+
+var (
+	// ErrNoNamespace signals the requested namespace does not exist
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/rules/store.go
```diff
@@ -106,3 +110,23 @@ func getLatestConfigID(cfgs map[string]userconfig.VersionedRulesConfig, latest u
 	}
 	return ret
 }
+
+// ListRuleGroups is not implemented
```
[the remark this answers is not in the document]
jtlisi: I don't plan on implementing it in this PR. Due to the complexity of the configdb api, the implementation will be a bit tricky and it's best to do in a separate PR.

pkg/ruler/api.go
jtlisi: will do

pkg/ruler/api.go
jtlisi: 👍 `

pkg/ruler/api.go
jtlisi: I will add coverage for this in the integration tests. It's difficult to replicate this in a unit test since it requires the router to parse the variables.

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/rules/store.go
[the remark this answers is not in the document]
pstibrany: Sounds good!

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+				RuleName: ruleName,
+				Err:      err,
+			})
+		}
+	}
+
+	return errs
+}
+
+func marshalAndSend(formatted interface{}, w http.ResponseWriter, logger log.Logger) {
```
pstibrany: nit: I wouldn't call it `formatted` here, perhaps just output, or result.

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+	groupName, err := url.PathUnescape(groupName)
+	if err != nil {
+		return "", err
+	}
+
+	return groupName, nil
+}
+
+func parseRequest(req *http.Request, requireNamespace, requireGroup bool) (string, string, string, error) {
+	id, err := user.ExtractOrgID(req.Context())
```
pstibrany: `orgId` or `userId`

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+	groupName, err := url.PathUnescape(groupName)
+	if err != nil {
+		return "", err
+	}
+
+	return groupName, nil
+}
+
+func parseRequest(req *http.Request, requireNamespace, requireGroup bool) (string, string, string, error) {
```
pstibrany: With 4 return values, some comment would be nice about what they are.

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+	return id, namespace, group, nil
+}
+
+func (r *Ruler) listRules(w http.ResponseWriter, req *http.Request) {
+	logger := util.WithContext(req.Context(), util.Logger)
+
+	userID, namespace, _, err := parseRequest(req, false, false)
+	if err != nil {
+		respondError(logger, w, err.Error())
```
pstibrany: missing `return` after `respondError`

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+	formatted := rgs.Formatted()
+	marshalAndSend(formatted, w, logger)
+}
+
+func (r *Ruler) getRuleGroup(w http.ResponseWriter, req *http.Request) {
+	logger := util.WithContext(req.Context(), util.Logger)
+	userID, namespace, groupName, err := parseRequest(req, true, true)
+	if err != nil {
+		respondError(logger, w, err.Error())
```
pstibrany: missing `return` after `respondError`

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+	formatted := store.FromProto(rg)
+	marshalAndSend(formatted, w, logger)
+}
+
+func (r *Ruler) createRuleGroup(w http.ResponseWriter, req *http.Request) {
+	logger := util.WithContext(req.Context(), util.Logger)
+	userID, namespace, _, err := parseRequest(req, true, false)
+	if err != nil {
+		respondError(logger, w, err.Error())
```
pstibrany: missing `return` after `respondError`

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+
+	respondAccepted(w, logger)
+}
+
+func (r *Ruler) deleteRuleGroup(w http.ResponseWriter, req *http.Request) {
+	logger := util.WithContext(req.Context(), util.Logger)
+
+	userID, namespace, groupName, err := parseRequest(req, true, true)
+	if err != nil {
+		respondError(logger, w, err.Error())
```
pstibrany: missing `return` after `respondError`

pkg/ruler/api.go
```diff
@@ -268,3 +284,241 @@ func (r *Ruler) alerts(w http.ResponseWriter, req *http.Request) {
...
+func (r *Ruler) createRuleGroup(w http.ResponseWriter, req *http.Request) {
+	logger := util.WithContext(req.Context(), util.Logger)
+	userID, namespace, _, err := parseRequest(req, true, false)
+	if err != nil {
+		respondError(logger, w, err.Error())
+	}
+
+	payload, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		level.Error(logger).Log("err", err.Error())
```
pstibrany: There are multiple instances of log message like this, without providing any other message or context. I think some extra details could be useful for debugging.

Also, why call `err.Error()`, instead of passing `err` directly?

pkg/ruler/api.go
pstibrany: One possible way to test it would be to have `muxVars` global variable, that defaults to `mux.Vars` but could be set to your own function  in tests.

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
jtlisi: Technically we use `x-orgid-scope` across are stack. Weave has support for users but it's not used in cortex. In general userID and orgID are the same across are stack. I'll add an issue to make the nomenclature consistent.

pstibrany: I meant to use one of those instead of just `id`, that seems little too short here.

jtlisi: Of course, I decided to use userID. Just haven’t pushed the to the branch yet 😉

pkg/ruler/api.go
jtlisi: I'll add a clarifying comment to the function

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
jtlisi: 👍

pkg/ruler/api.go
jtlisi: I'll update these log lines to be more descriptive

jtlisi:
> One possible way to test it would be to have muxVars global variable, that defaults to mux.Vars but could be set to your own function in tests.

One way would be injecting them into the ctx of the request that is passed to the function. But I feel like it's more important to have integration tests for this behavior since the interplay with the gorilla router is so important.

pkg/ruler/api.go
jtlisi: One way would be injecting them into the ctx of the request that is passed to the function. But I feel like it's more important to have integration tests for this behavior since the interplay with the gorilla router is so important.

I updated the integration tests to use encoded names for the rule groups and namespaces that I think should be sufficient

PR #2270, opened 2020-03-13
added FIFO cache metrics for current number of entries and memory usage

### LINKED ISSUE #2258: FIFO cache does not report number of entries and memory in use
FIFO cache has multiple useful metrics, but it lacks metrics for the current number of entries and used memory.

dmitsh:
fixed bug in updating last element of FIFO
updated unit tests

**What this PR does**:
Adding metrics for current number of entries and memory usage in FIFO cache
Fixing corner case bug in FIFO (updating last element in FIFO)

Fixes #2258 

**Checklist**
- [x ] Tests updated
- [ ] Documentation added


pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -72,18 +75,55 @@ func TestFifoCache(t *testing.T) {
...
 
-func TestFifoCacheExpiry(t *testing.T) {
-	c := NewFifoCache("test", FifoCacheConfig{Size: size, Validity: 5 * time.Millisecond})
+func TestFifoCacheEvictionExpiry(t *testing.T) {
+	c := NewFifoCache("test2", FifoCacheConfig{Size: 3, Validity: 5 * time.Millisecond})
 	ctx := context.Background()
+	memorySz := int(unsafe.Sizeof(cacheEntry{})) * 3
 
-	c.Put(ctx, []string{"0"}, []interface{}{0})
+	assert.Equal(t, testutil.ToFloat64(c.entriesAdded), float64(0))
```
gouthamve: Can we use `require.Equal` here not to add one more dependency? Or use this asset package: https://godoc.org/github.com/stretchr/testify/assert

pracucci: `assert` is more appropriate, but as Goutham outlined the wrong package was imported. The correct one is:
`github.com/stretchr/testify/assert`

Please cleanup `vendor/` and `go.mod` once done.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -213,6 +244,9 @@ func (c *FifoCache) put(ctx context.Context, key string, value interface{}) {
 	c.entries[c.last].next = index
 	c.first = index
 	c.index[key] = index
+
+	c.memoryBytes.Add(float64(sizeOf(key) + sizeOf(value)))
+	c.entriesCurrent.Set(float64(len(c.entries)))
```
pracucci: [nit, no strong opinions] Is it equal to do `c.entriesCurrent.Inc()` or am I wrong? If I'm correct, `.Inc()` may be just clearer.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -240,3 +274,32 @@ func (c *FifoCache) Get(ctx context.Context, key string) (interface{}, bool) {
...
+	case []int32:
+		return len(v) * 4
+	case []uint32:
+		return len(v) * 4
+	case []float32:
+		return len(v) * 4
+	case []int:
+		return len(v) * 8
+	case []uint:
+		return len(v) * 8
```
pracucci: FYI, these are machine dependent. It's not safe to assume it's `8` bytes, even if `8` is the most common case (ie. we don't release 32bit binaries and probably any production system runs on 64 bit). That being said, if we keep it like this I would add a comment to explain it.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -289,16 +289,23 @@ func sizeOf(i interface{}) int {
...
-		return len(v) * 8
-	case []uint:
-		return len(v) * 8
 	case []int64:
 		return len(v) * 8
 	case []uint64:
 		return len(v) * 8
 	case []float64:
 		return len(v) * 8
+	// next 2 cases are machine dependen
```
pracucci: `dependen` > `dependent`

pkg/chunk/cache/fifo_cache.go
```diff
@@ -33,7 +33,7 @@ var (
 		Help:      "The total number of evicted entries",
 	}, []string{"cache"})
 
-	cacheEntriesCurrent = promauto.NewGaugeVec(prometheus.GaugeOpts{
+	cacheEntriesCurrent = promauto.NewCounterVec(prometheus.CounterOpts{
```
pracucci: The fact that currently the number of items in the cache can't decrease is just an implementation detail. Being an LRU cache and having the concept of eviction, in the future we may actually have a routine deleting expired items. For this reason, I would suggest to revert this back to a gague.

pkg/chunk/cache/fifo_cache.go
dmitsh: thanks :)

pkg/chunk/cache/fifo_cache.go
dmitsh: agree!

PR #2273, opened 2020-03-13
Saner and consistent YAML fields in config

gouthamve:

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -546,45 +546,45 @@ lifecycler:
 
 # Period with which to attempt to flush chunks.
 # CLI flag: -ingester.flush-period
-[flushcheckperiod: <duration> | default = 1m0s]
+[flush_period: <duration> | default = 1m0s]
 
 # Period chunks will remain in memory after flushing.
 # CLI flag: -ingester.retain-period
-[retainperiod: <duration> | default = 5m0s]
+[retain_period: <duration> | default = 5m0s]
 
 # Maximum chunk idle time before flushing.
 # CLI flag: -ingester.max-chunk-idle
-[maxchunkidle: <duration> | default = 5m0s]
+[max_chunk_idle: <duration> | default = 5m0s]
 
 # Maximum chunk idle time for chunks terminating in stale markers before
 # flushing. 0 disables it and a stale series is not flushed until the
 # max-chunk-idle timeout is reached.
 # CLI flag: -ingester.max-stale-chunk-idle
-[maxstalechunkidle: <duration> | default = 0s]
+[max_stale_chunk_idle: <duration> | default = 0s]
 
 # Timeout for individual flush operations.
 # CLI flag: -ingester.flush-op-timeout
-[flushoptimeout: <duration> | default = 1m0s]
+[flush_op_timeout: <duration> | default = 1m0s]
 
 # Maximum chunk age before flushing.
 # CLI flag: -ingester.max-chunk-age
-[maxchunkage: <duration> | default = 12h0m0s]
+[max_chunk_age: <duration> | default = 12h0m0s]
 
 # Range of time to subtract from MaxChunkAge to spread out flushes
 # CLI flag: -ingester.chunk-age-jitter
-[chunkagejitter: <duration> | default = 20m0s]
+[chunk_age_jitter: <duration> | default = 20m0s]
 
 # Number of concurrent goroutines flushing to dynamodb.
 # CLI flag: -ingester.concurrent-flushes
-[concurrentflushes: <int> | default = 50]
+[concurrent_flushes: <int> | default = 50]
 
 # If true, spread series flushes across the whole period of MaxChunkAge
 # CLI flag: -ingester.spread-flushes
-[spreadflushes: <boolean> | default = false]
+[spread_flushes: <boolean> | default = false]
 
 # Period with which to update the per-user ingestion rates.
 # CLI flag: -ingester.rate-update-period
-[rateupdateperiod: <duration> | default = 15s]
+[rate_update_period: <duration> | default = 15s]
 ```
 
 ### `querier_config`
@@ -594,7 +594,7 @@ The `querier_config` configures the Cortex querier.
 ```yaml
 # The maximum number of concurrent queries.
 # CLI flag: -querier.max-concurrent
-[maxconcurrent: <int> | default = 20]
+[max_concurrent: <int> | default = 20]
 
 # The timeout for a query.
 # CLI flag: -querier.timeout
@@ -608,15 +608,15 @@ The `querier_config` configures the Cortex querier.
 # Use batch iterators to execute query, as opposed to fully materialising the
 # series in memory.  Takes precedent over the -querier.iterators flag.
 # CLI flag: -querier.batch-iterators
-[batchiterators: <boolean> | default = false]
+[batch_iterators: <boolean> | default = false]
 
 # Use streaming RPCs to query ingester.
 # CLI flag: -querier.ingester-streaming
-[ingesterstreaming: <boolean> | default = false]
+[ingester_streaming: <boolean> | default = false]
 
 # Maximum number of samples a single query can load into memory.
 # CLI flag: -querier.max-samples
-[maxsamples: <int> | default = 50000000]
+[max_samples: <int> | default = 50000000]
 
 # Maximum lookback beyond which queries are not sent to ingester. 0 means all
 # queries are sent to ingester.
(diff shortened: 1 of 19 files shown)


pkg/ingester/ingester.go
```diff
@@ -48,17 +48,17 @@ type Config struct {
...
-	MaxChunkAge       time.Duration
-	ChunkAgeJitter    time.Duration
-	ConcurrentFlushes int
-	SpreadFlushes     bool
-
-	RateUpdatePeriod time.Duration
+	FlushCheckPeriod  time.Duration `yaml:"flush_period"`
+	RetainPeriod      time.Duration `yaml:"retain_period"`
+	MaxChunkIdle      time.Duration `yaml:"max_chunk_idle"`
+	MaxStaleChunkIdle time.Duration `yaml:"max_stale_chunk_idle"`
```
pstibrany:
```suggestion
	MaxChunkIdle      time.Duration `yaml:"max_chunk_idle_time"`
	MaxStaleChunkIdle time.Duration `yaml:"max_stale_chunk_idle_time"`
```

pkg/ingester/ingester.go
```diff
@@ -803,9 +803,9 @@ func (i *Ingester) Watch(in *grpc_health_v1.HealthCheckRequest, stream grpc_heal
 	return status.Error(codes.Unimplemented, "Watching is not supported")
 }
 
-// ReadinessHandler is used to indicate to k8s when the ingesters are ready for
-// the addition removal of another ingester. Returns 204 when the ingester is
-// ready, 500 otherwise.
+// CheckReady is the readiness handler is used to indicate to k8s when the ingesters
```
pstibrany:
```suggestion
// CheckReady is the readiness handler used to indicate to k8s when the ingesters
```

pkg/ingester/ingester.go
```diff
@@ -803,9 +803,9 @@ func (i *Ingester) Watch(in *grpc_health_v1.HealthCheckRequest, stream grpc_heal
...
+// are ready for the addition removal of another ingester. Returns 204 when the
+// ingester is ready, 500 otherwise.
```
pstibrany:
```suggestion
// are ready for the addition removal of another ingester.
```

HTTP status codes are not returned from here anymore. (Fixing my code change via your PR :))

pkg/querier/frontend/worker.go
```diff
@@ -31,9 +31,9 @@ var (
 
 // WorkerConfig is config for a worker.
 type WorkerConfig struct {
-	Address           string
-	Parallelism       int
-	DNSLookupDuration time.Duration
+	Address           string        `yaml:"address"`
```
pstibrany: What is address? URL? Can be `frontend_url` instead

pkg/querier/querier.go
```diff
@@ -21,12 +21,12 @@ import (
...
 type Config struct {
-	MaxConcurrent        int
-	Timeout              time.Duration
-	Iterators            bool
-	BatchIterators       bool
-	IngesterStreaming    bool
-	MaxSamples           int
+	MaxConcurrent        int           `yaml:"max_concurrent"`
+	Timeout              time.Duration `yaml:"timeout"`
+	Iterators            bool          `yaml:"iterators"`
```
pstibrany: This sounds like "number of iterators" to me, not as a boolean flag.

pkg/ring/kv/consul/client.go
```diff
@@ -35,12 +35,12 @@ var (
 
 // Config to create a ConsulClient
 type Config struct {
-	Host              string
-	ACLToken          string
-	HTTPClientTimeout time.Duration
-	ConsistentReads   bool
-	WatchKeyRateLimit float64 // Zero disables rate limit
-	WatchKeyBurstSize int     // Burst when doing rate-limit, defaults to 1
+	Host              string        `yaml:"host"`
```
pstibrany: `url`? I think it would be nice to be consistent to use `url` in the names for fields which are actually URLs.

jtlisi: I agree, I think this should be consistent where relevant

pkg/ruler/ruler.go
```diff
@@ -52,23 +52,35 @@ var (
...
-	Ring             RingConfig
-	FlushCheckPeriod time.Duration
+	// This is used for template expansion in alerts; must be a valid URL.
+	ExternalURL flagext.URLValue `yaml:"external_url"`
+	// How frequently to evaluate rules by default.
+	EvaluationInterval time.Duration `yaml:"evaluation_interval"`
+	// How frequently to poll for updated rules.
+	PollInterval time.Duration `yaml:"poll_interval"`
+	// Rule Storage and Polling configuration.
+	StoreConfig RuleStoreConfig `yaml:"store_config"`
```
jtlisi: Can you call this yaml field `storage` to match the flag prefix it adds?

pkg/ingester/ingester.go
jtlisi: @pstibrany would you suggest we suffix all duration values like this with `_time`. Similar to how counters in Prometheus are suffixed with `_total`. I think that might be a little clunky but does provide value.

pkg/alertmanager/multitenant.go
```diff
@@ -96,20 +96,20 @@ func init() {
...
+	Peers                flagext.StringSlice `yaml:"peers"`
+	PeerTimeout          time.Duration       `yaml:"peer_timeout"`
 
-	FallbackConfigFile string
-	AutoWebhookRoot    string
+	FallbackConfigFile string `yaml:"fallback_config_file"`
+	AutoWebhookRoot    string `yaml:"auto_webhook_root"`
 
-	Store AlertStoreConfig
+	Store AlertStoreConfig `yaml:"store"`
```
jtlisi: Can this also be a `storage` field? Similar to the ruler?

pkg/ring/kv/consul/client.go
pstibrany: Actually, this is not a URL :-( Expected format here is host:port.

pkg/querier/frontend/worker.go
pstibrany: We should specify expected format of address / host fields, especially if it is not just hostname.

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -97,13 +97,13 @@ func init() {
...
-	ApplicationAutoScaling flagext.URLValue
-	Metrics                MetricsAutoScalingConfig
-	ChunkGangSize          int
-	ChunkGetMaxParallelism int
+	DynamoDB               flagext.URLValue         `yaml:"dynamodb_url"`
+	APILimit               float64                  `yaml:"api_limit"`
+	ThrottleLimit          float64                  `yaml:"throttle_limit"`
+	ApplicationAutoScaling flagext.URLValue         `yaml:"application_autoscaling_url"`
+	Metrics                MetricsAutoScalingConfig `yaml:"metrics"`
+	ChunkGangSize          int                      `yaml:"chunk_gang_size"`
```
pracucci: I believe its CLI flag name is not correct. It should end with `.chunk-gang-size`

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -97,13 +97,13 @@ func init() {
...
+	ChunkGetMaxParallelism int                      `yaml:"chunk_get_max_parallelism"`
```
pracucci: I believe its CLI flag name is not correct. It should end with `.chunk-get-max-parallelism`

pkg/ring/kv/consul/client.go
```diff
@@ -35,12 +35,12 @@ var (
...
+	ACLToken          string        `yaml:"acl_token"`
```
pracucci: The CLI flag should end with `.acl-token`. Then remember to update `arguments.md`

bboreham:
>  booleans names like enable_ (not disable)

Historically these were chosen so the zero value (`false`) is the default.
At some point we went to mandatory initialization of config structs, so maybe that reason has gone, but I don't think there is a strong argument just on consistency.

For many flags the default is "what Cortex did earlier" but should be changed to "what most people would use".  I think this type of change is much more important.

docs/configuration/config-file-reference.md
```diff
@@ -1345,24 +1346,24 @@ The `storage_config` configures where Cortex stores the data (chunks storage eng
 [engine: <string> | default = "chunks"]
 
 aws:
-  dynamodbconfig:
+  dynamodb_config:
```
pracucci: Why not simply `dynamodb`?

PR #2275, opened 2020-03-16
support for cassandra in integration tests

sandeepsukhani:
**What this PR does**:
Adds support of cassandra in integration tests

I have used an optimized docker image for integration tests. Here is the repo for it https://github.com/saidbouras/cassandra-docker-unit.
There is also a community maintained docker image https://hub.docker.com/_/cassandra, which is slow to start and takes about 10-15 additional seconds.
Cassandra version in image built by saidbouras is 3.11 while community build docker has 3.11.6 which I feel is okay. We can use community maintained docker image as well if we want if that is not okay.

Since Cassandra is slow to start and does not have an HTTP readiness probe, I have added support for command based probe. An example of readiness probe for Cassandra can be found here: https://github.com/kubernetes/examples/blob/b86c9d50be45eaf5ce74dee7159ce38b0e149d38/cassandra/image/files/ready-probe.sh

### DIFF
--- a/integration/e2e/service.go
+++ b/integration/e2e/service.go
@@ -354,22 +354,22 @@ type ReadinessProbe interface {
 	Ready(service *ConcreteService) (err error)
 }
 
-// ReadinessProbeHTTP checks readiness by making HTTP call and checking for expected HTTP status code
-type ReadinessProbeHTTP struct {
+// HTTPReadinessProbe checks readiness by making HTTP call and checking for expected HTTP status code
+type HTTPReadinessProbe struct {
 	port           int
 	path           string
 	expectedStatus int
 }
 
-func NewReadinessProbeHTTP(port int, path string, expectedStatus int) *ReadinessProbeHTTP {
-	return &ReadinessProbeHTTP{
+func NewHTTPReadinessProbe(port int, path string, expectedStatus int) *HTTPReadinessProbe {
+	return &HTTPReadinessProbe{
 		port:           port,
 		path:           path,
 		expectedStatus: expectedStatus,
 	}
 }
 
-func (p *ReadinessProbeHTTP) Ready(service *ConcreteService) (err error) {
+func (p *HTTPReadinessProbe) Ready(service *ConcreteService) (err error) {
 	// Map the container port to the local port
 	localPort, ok := service.networkPortsContainerToLocal[p.port]
 	if !ok {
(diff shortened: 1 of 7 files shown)


integration/e2e/service.go
```diff
@@ -334,21 +350,32 @@ func NewCommandWithoutEntrypoint(cmd string, args ...string) *Command {
...
+// ReadinessProbeHTTP checks readiness by making HTTP call and checking for expected HTTP status code
+type ReadinessProbeHTTP struct {
 	port           int
 	path           string
 	expectedStatus int
 }
 
-func NewReadinessProbe(port int, path string, expectedStatus int) *ReadinessProbe {
-	return &ReadinessProbe{
+func NewReadinessProbeHTTP(port int, path string, expectedStatus int) *ReadinessProbeHTTP {
```
pstibrany: I would use `HTTPReadinessProbe` and `CmdReadinesProbe` names.

integration/e2e/service.go
```diff
@@ -363,6 +390,20 @@ func (p *ReadinessProbe) Ready(localPort int) (err error) {
 	return fmt.Errorf("got no expected status code: %v, expected: %v", res.StatusCode, p.expectedStatus)
 }
 
+// ReadinessProbeCmd checks readiness by `Exec`ing a command which returns 0 to consider status being ready
```
pstibrany: executing or running a command (within container)

integration/e2ecortex/service.go
```diff
@@ -13,7 +13,7 @@ func NewCortexService(
 	name string,
 	image string,
 	command *e2e.Command,
-	readiness *e2e.ReadinessProbe,
+	readiness *e2e.ReadinessProbeHTTP,
```
pstibrany:
```suggestion
	readiness readiness e2e.ReadinessProbe,
```

sandeepsukhani:
> LGTM, with minor nits.

Thanks Peter for the review! I have pushed the changes.

integration/e2e/service.go
```diff
@@ -334,21 +350,32 @@ func NewCommandWithoutEntrypoint(cmd string, args ...string) *Command {
...
 		port:           port,
 		path:           path,
 		expectedStatus: expectedStatus,
 	}
 }
 
-func (p *ReadinessProbe) Ready(localPort int) (err error) {
+func (p *HTTPReadinessProbe) Ready(service *ConcreteService) (err error) {
+	// Map the container port to the local port
+	localPort, ok := service.networkPortsContainerToLocal[p.port]
```
pracucci: This should be readiness probe's responsability. What do you think to add a function `ConcreteService.EndpointPort()` which is like `.Endpoint()` but just returns the port? We could return `0` if the port doesn't exist and convert the following `if` from `if !ok` to `if port == 0`.

What's your take?

pracucci: Errata corrige. You need the endpoint here, so you can directly use `Endpoint()`.

Issue #2037, opened 2020-01-27
Cortex should tell remote write clients to slow down on rate-limiting


bboreham:
Currently, when the ingest rate limit is hit for a tenant, Cortex goes immediately from accepting all data to dropping all data.  The dropped data is lost permanently because the mechanism is to return a 500 error. 

I propose Cortex should have an inbetween state where the return value indicates that the sender should slow down, but does not cause the data to be dropped. For instance it could return 429 - current Prometheus will retry immediately, but it could be enhanced to slow down.  Further, Cortex could return an indication of how close to the limit the tenant is.

This idea is not original - I have seen similar in DynamoDB, CosmosDB and GitHub.

I think this would solve #837, probably better than #879.

csmarchbanks:
I believe current Prometheus only retries on 5XX errors, though perhaps it should also retry 429.

ranton256:
I have been doing some thinking about this and asked around for some other ideas, then after looking at the RFC that covers 429, I realize it actually has an optional header for “Retry-After” so that seems like at least one avenue we should consider.
 
The code to decide what errors are retryable was added in https://github.com/prometheus/prometheus/pull/2552/files and now is at https://github.com/prometheus/prometheus/blob/4e5b1722b342948a55b3d7753f6539040db0e5f0/storage/remote/client.go#L200 .
It only treats 5xx as retryable. I think we should consider changing it to look for the "Retry-After" header and set that from Cortex or other remote storage providers of remote write so that we could give it a hint when to retry or not without having to send some less appropriate status code than 429.

We could also consider something like X-RateLimit, https://tools.ietf.org/id/draft-polli-ratelimit-headers-00.html

bboreham:
Agreed.  Couple more points at https://github.com/cortexproject/cortex/pull/3654#issuecomment-763006285

I couldn't see an existing issue in the Prometheus repo so opened https://github.com/prometheus/prometheus/issues/8418

csmarchbanks:
I left the link in the issue Bryan opened as well, but there is an open PR for adding the Retry-After functionality that @Harkishen-Singh has been working on: https://github.com/prometheus/prometheus/pull/8237.

bboreham:
The necessary part is now implemented in Prometheus https://github.com/prometheus/prometheus/pull/8237
(and made optional in https://github.com/prometheus/prometheus/pull/8477).

I'll leave this open until Cortex sends the `retry-after` header to make it slow down an appropriate amount.

[...]

PR #2276, opened 2020-03-16
Wait for rate-limit instead of bailing out immediately

### LINKED ISSUE #879: Distributor rate-limiter should call Wait() not Allow()
I think this will work better under very sudden bursts of input.

Under certain circumstances I have seen a scraping Prometheus (v1.8) expand to 1,000 shards, and then have all 1,000 shards upload at the same time, creating an input burst of x00,000 samples.

Our rate-limit defaults to 25,000 samples per second, per distributor.

The behaviour of `Allow()` under a large burst is to count up to 25,000 saying 'yes', then say 'no' to the rest.

The behaviour of `Wait()` is to say 'yes' to the first 25,000, then start a sleep for the next one, then the next one, and so on.  If the `Context` has a timeout, let's say 10 seconds, then it figures out it can allow 250,000 samples to wait, and it starts saying 'no' immediately to subsequent callers.

Note we don't currently have a timeout on the input side of distributor, but we have one defaulted to 2 seconds downstream to ingesters.

bboreham:
This allows for temporary peaks to be smoothed out without dropping data.

Fixes #879 - see description there for more detail on the idea.
Probably needs further discussion - I suspect that, as coded here, if the sender just has too much data we will never catch up.  EDIT and the tests don't match the new behaviour hence fail.

I deployed this to my production in the catch-up phase after an outage, and subjectively the behaviour is vastly preferrable.  Smooth catch-up at the maximum allowed rate, no data dropped due to rate-limiting.

See also #2037  

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -423,9 +423,13 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	}
 
 	now := time.Now()
-	if !d.ingestionRateLimiter.AllowN(now, userID, validatedSamples) {
+	if err := d.ingestionRateLimiter.WaitN(ctx, now, userID, validatedSamples); err != nil {
 		// Ensure the request slice is reused if the request is rate limited.
 		client.ReuseSlice(req.Timeseries)
+
+		if ctxErr := ctx.Err(); ctxErr != nil { // bail out if caller cancelled
+			return nil, ctxErr
+		}
 
 		// Return a 4xx here to have the client discard the data and not retry. If a client
 		// is sending too much data consistently we will unlikely ever catch up otherwise.
--- a/pkg/util/limiter/rate_limiter.go
+++ b/pkg/util/limiter/rate_limiter.go
@@ -1,6 +1,7 @@
 package limiter
 
 import (
+	"context"
 	"sync"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/distributor/distributor.go
```diff
@@ -423,10 +423,14 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
 	now := time.Now()
-	if !d.ingestionRateLimiter.AllowN(now, userID, validatedSamples) {
+	if err := d.ingestionRateLimiter.WaitN(ctx, now, userID, validatedSamples); err != nil {
 		// Ensure the request slice is reused if the request is rate limited.
 		client.ReuseSlice(req.Timeseries)
 
+		if ctxErr := ctx.Err(); ctxErr != nil { // bail out if caller cancelled
+			return nil, ctxErr
+		}
+
```
pstibrany: wouldn't `WaitN` have returned error already in such case?

bboreham: The point of this bit is not to count a cancellation by the caller as a discard.

pracucci:
_Still valid. Would be interesting hearing your thoughts about my two questions above, cause I'm intrigued by this change._

bboreham:
On the community call @tomwilkie noted that adding latency inside Cortex is likely to trigger alerts.
He suggested returning a 500 error for Prometheus to retry.

PR #2277, opened 2020-03-16
Updated RELEASE instructions

pracucci:
**What this PR does**:
Updated the release instructions to ensure the release tag is merged (without squash) into `master` to make sure `go` versioning works as expected.

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -85,7 +85,7 @@ To publish a stable release:
    - Merge to `master` in the local checkout
    - Fix any conflict and `git commit -s`
    - Temporarily disable "Include administrators" in the [`master` branch protection rule](https://github.com/cortexproject/cortex/settings/branch_protection_rules)
-   - Push changes to upstream
+   - Push changes to upstream (please double check before pushing!)
    - Re-enable "Include administrators" in the [`master` branch protection rule](https://github.com/cortexproject/cortex/settings/branch_protection_rules)
 
 ### How to tag a release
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -80,7 +80,13 @@ To publish a stable release:
 4. Wait until CI pipeline succeeded (once a tag is created, the release process through CircleCI will be triggered for this tag)
 5. Create a release in GitHub
    - Write the release notes (including a copy-paste of the changelog)
-   - Build binaries with `make disk` and attach them to the release
+   - Build binaries with `make dist` and attach them to the release
+6. Merge the release branch `release-x.y` to `master`
+   - Merge to `master` in the local checkout
+   - Fix any conflict and `git commit -s`
+   - Temporarily disable "Include administrators" in the [`master` branch protection rule](https://github.com/cortexproject/cortex/settings/branch_protection_rules)
+   - `git push`
```
pstibrany: This is correct, if you know what you're doing. In my setup, it would push to incorrect fork though. (My `master` branch tracks my own fork).

I would perhaps be less specific with commands, and just say "push merge commit to Cortex master" – make sure to not push anything else.

Maybe I would instead say that it's best to do this work on a separate branch, not "master". Eg. `git checkout -b upstream_master upstream/master` (provided that your `upstream` remote is original fork, as suggested by GitHub own CLI tools), then do the merge, and then push back to `upstream`.

pracucci: I see your point. I guess we all have a different workflow and we can probably give `git` knowledge for granted and just mention "Push changes to upstream".

What's your take?

pstibrany: I think that's fair assumption.

pstibrany: I'd add "Double check before pushing!" :)

pracucci: Makes sense. What's about now?

pstibrany: Looks good.

PR #2279, opened 2020-03-17
cache invalidation using cache gen numbers to invalidate only specific users data

sandeepsukhani:
**What this PR does**:
When a delete request is received, we do query time filtering of data requested for deletion until it actually gets deleted. Since we cache query results, index and chunks, we need to invalidate the cache. Since we can selectively invalidate the cache for data requested for deletion, we can at best limit the scope to the cache entries for the user performing the deletion.

We are doing cache invalidating for specific users by prefixing a Generation Number to cache keys. To invalidate cache for a user we just update its Generation Number which is stored in an Index Table. Generation Number here is just current timestamp in seconds. A counter would be preferred here for Generation Number but some of our Index stores do not support atomic increment of counters.

To simplify the process of invalidating the cache using Generation Numbers, I have added a new middleware called `GenNumMiddleware` which would work on a reference of another `Cache` instance. `Store` and `Fetch` method would get the Generation Number from `ctx` and simply add it to the cache keys.

Cache Invalidation Process:
In Cortex we cache 2 types of data, one is data from stores(index, chunk) and another is the query results that the frontend receives. I am treating cache invalidation for both the results differently because of how deletion works. Let us call cache for stores a `StoreCache` and for query results a `ResultsCache`.

* Cache Invalidation for `ResultsCache`: When a delete request is received we do query time filtering for it until data is actually deleted, which means the data that is cached in `ResultsCache` might still have results built from data requested to be deleted. To avoid returning results built from data requested for deletion, we will invalidate the `ResultsCache` for a user whenever they send a new delete request. We do not have to invalidate `ResultsCache` again after delete request is processed because we are anyways building cache from results built from filtered out data.

* Cache Invalidation for `StoreCache`: When a delete request is received we just do query time filtering for it until data is actually deleted, which means there is no change in stores yet so we do not have to invalidate `StoreCache` until the delete requests get processed. We will invalidate `StoreCache` for a user whenever one of its delete requests gets processed i.e all the data gets deleted from stores.

Cache Invalidation in Query Frontend:
Since Query Frontend does not have access to `Store` to get Cache Generation numbers for users doing the query, I have added a new HTTP header called `results-cache-gen-number` to the query response that they get from `Queriers`. This new header will hold the Generation Number that the `Querier` which is sending the response is seeing. Query Frontend would keep Generation Numbers for each user with it and use it during caching the data and lookups.

Since Cache Generation numbers are loaded asynchronously, different `Querier` could see different  Cache Generation numbers at a different point in time. The difference in Generation Numbers means the results could change and is handled by Query Frontend by caching the response only if all the responses that are being considered for a query have the same cache generation number.

**Checklist**
- [ ] Tests updated


pkg/chunk/purger/delete_requests_store.go
```diff
@@ -47,6 +55,11 @@ type DeleteRequest struct {
 	CreatedAt model.Time          `json:"created_at"`
 }
 
+// CacheGenNumbers holds store and results cache gen numbers for a user
```
gouthamve: End with a fullstop?

pkg/querier/queryrange/results_cache.go
```diff
@@ -202,6 +223,34 @@ func (s resultsCache) handleMiss(ctx context.Context, r Request) (Response, []Ex
 		return response, []Extent{}, nil
 	}
 
+	cacheGenNumber := s.extractor.ExtractCacheGenNumber(response)
+
+	if cacheGenNumber == "-1" {
```
gouthamve: When will this be true in `handleMiss`? From what I can see this is only going to have downstream requests which will never return `-1` as the gen number?

pkg/querier/queryrange/results_cache.go
```diff
@@ -153,7 +168,13 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
 	if ok {
 		response, extents, err = s.handleHit(ctx, r, cached)
 	} else {
-		response, extents, err = s.handleMiss(ctx, r)
+		response, extents, err = s.handleMiss(ctx, r, userID)
+
+		// check if there is change in gen number to cache data with new gen number
```
gouthamve: What is happening here?

pkg/querier/queryrange/results_cache.go
```diff
@@ -202,6 +223,34 @@ func (s resultsCache) handleMiss(ctx context.Context, r Request) (Response, []Ex
...
+	if cacheGenNumber == "-1" {
+		level.Debug(s.logger).Log("msg", "there is an inconsistency in cache generation numbers, not caching the response")
+		return response, []Extent{}, nil
+	}
+
+	if cacheGenNumber != "" {
+		if s.cacheGenNumbers[userID] == "" {
+			s.cacheGenNumbers[userID] = cacheGenNumber
+		} else {
+			// see if we have received higher gen number and start using it for caching new results using that
```
gouthamve: Why are we doing this? Lets avoid this optimisation and let the query-frontend depend only on the periodic reload?

pkg/chunk/storage/factory.go
```diff
@@ -118,11 +118,16 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 	chunksCache = cache.StopOnce(chunksCache)
 	writeDedupeCache = cache.StopOnce(writeDedupeCache)
 
+	// lets wrap all caches with CacheGenMiddleware to facilitate cache invalidation using cache generation numbers
+	indexReadCache = cache.NewCacheGenNumMiddleware(indexReadCache)
+	chunksCache = cache.NewCacheGenNumMiddleware(chunksCache)
+	writeDedupeCache = cache.NewCacheGenNumMiddleware(writeDedupeCache)
```
jtlisi: I don't think we need gen numbers on the writeDedupCache.

pkg/querier/queryrange/results_cache.go
```diff
@@ -49,43 +54,51 @@ func (cfg *ResultsCacheConfig) RegisterFlags(f *flag.FlagSet) {
 type Extractor interface {
 	// Extract extracts a subset of a response from the `start` and `end` timestamps in milliseconds in the `from` response.
 	Extract(start, end int64, from Response) Response
+	ExtractCacheGenNumber(from Response) string
```
jtlisi: If we are going to get the Cache gen number here, can we propagate it to the downstream calls using a context value, instead of retrieving for each call to the store/cache?

pkg/chunk/composite_store.go
```diff
@@ -171,11 +177,17 @@ func (c compositeStore) Stop() {
...
 }
 
-func (c compositeStore) forStores(from, through model.Time, callback func(from, through model.Time, store Store) error) error {
+func (c compositeStore) forStores(ctx context.Context, userID string, from, through model.Time, callback func(ctx context.Context, from, through model.Time, store Store) error) error {
 	if len(c.stores) == 0 {
 		return nil
 	}
 
+	var err error
+	ctx, err = c.injectCacheGen(ctx, userID)
```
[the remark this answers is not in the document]
jtlisi: Since we add the cache gen to the context in the query frontend we shouldn't have to inject it here.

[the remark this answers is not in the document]
sandeepsukhani: We would still need it because we add cache gen to keys during writes as well and not everyone would be using query frontend. Did I miss something?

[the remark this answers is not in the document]
jtlisi: When I say `writes` I mean when the ingesters flush chunks to the store. Writes to the chunks-cache and index-cache only occur during query operations. The only cache that is accessed during `Put`/`PutOne`/`PutChunks`/`BatchWrite` operations is the write-dedup cache. Which really has no need for cachegen anyway. I'm proposing we add middleware to the query api for both the querier and the query-frontend that ensures a cache gen is provided and if not it adds one to the context. We could add a value for both the results cache and store here. Then we have no need to load a cache gen number downstream. The only code that needs to be aware of it's existence is the cache clients themselves.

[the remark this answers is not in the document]
jtlisi: Actually I was incorrect, the chunkstore does write it's chunks using a writeBackCache. However, that code intentionally does not return an error. 

```
	if cacheErr := c.writeBackCache(ctx, chunks); cacheErr != nil {
		level.Warn(log).Log("msg", "could not store chunks in chunk cache", "err", cacheErr)
	}
```

The primary issue I have is the blanket error return on all calls to `forStore` if the cachegen number can't be retrieved and injected. Caching is by definition best effort. We should not be failing writes or reads because of a caching error.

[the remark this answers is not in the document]
sandeepsukhani: The error is returned only when it fails to load the gen number from the store the first time and once it loaded it should never throw an error because we serve them from memory next time. If we do not return any error we would never know something is wrong unless we go and check logs or have an alert for it. I understand that caching is best effort but the errors here are thrown while loading something from db not while reading or writing the cache entries.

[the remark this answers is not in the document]
jtlisi: If that is the case, why return an error here at all. Why not have the cachegen fail on startup if it can't load from the DB? If we are returning an error here when the ingester is trying to flush chunks it's already too late and we will drop data.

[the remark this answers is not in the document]
gouthamve: Further, even with we just return error, this is on the flush path, which would mean we need to have alerts to know about the flushes alert anyways.

pkg/chunk/purger/tombstones.go
```diff
@@ -178,9 +243,53 @@ func (tl *TombstonesLoader) loadPendingTombstones(userID string) error {
 	defer tl.tombstonesMtx.Unlock()
 	tl.tombstones[userID] = &tombstoneSet
 
+	tl.metrics.deleteRequestsLoadFailures.Set(0)
 	return nil
 }
 
+// GetStoreCacheGenNumber returns store cache gen number for a user
+func (tl *TombstonesLoader) GetStoreCacheGenNumber(userID string) string {
```
jtlisi, line 251: Can we pass a context here and propagate it, to ensure if the delete store is accessed we have context for the trace?

pkg/querier/queryrange/results_cache.go
```diff
@@ -169,23 +201,32 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
...
+			headerValues = append(headerValues, hv.GetValues()...)
 		}
 	}
-	return true
+
+	return
 }
 
-func (s resultsCache) handleMiss(ctx context.Context, r Request) (Response, []Extent, error) {
+func (s resultsCache) handleMiss(ctx context.Context, r Request, userID string) (Response, []Extent, error) {
```
jtlisi, line 229:
```suggestion
func (s resultsCache) handleMiss(ctx context.Context, r Request) (Response, []Extent, error) {
```

The userID seems to be unused.

pkg/querier/queryrange/results_cache.go
```diff
@@ -169,23 +201,32 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
...
+		}
+	}
+
+	return true
+}
+
+func getHeaderValuesWithName(r Response, headerName string) (headerValues []string) {
 	if promResp, ok := r.(*PrometheusResponse); ok {
 		for _, hv := range promResp.Headers {
 			if hv.GetName() != cachecontrolHeader {
```
jtlisi, line 218:
```suggestion
			if hv.GetName() != headerName {
```

pkg/cortex/modules.go
```diff
@@ -614,3 +637,25 @@ var modules = map[ModuleName]module{
 		deps: []ModuleName{Querier, Ingester, Distributor, TableManager, DataPurger, StoreGateway},
 	},
 }
+
+// middleware for setting cache gen header to let consumer of response know all previous responses could be invalid due to delete operation
+func getHTTPCacheGenNumberHeaderSetterMiddleware(cacheGenNumbersLoader *purger.TombstonesLoader) middleware.Interface {
```
jtlisi, line 642: If this is going to be an http middleware, it should live in either the `api` package or a function specific package.

jtlisi, line 642: I don't think the best way to do this is in an HTTP middleware. The routing for the queriers and query-frontend is already relatively complex. I think it would make more sense to create a new `queryable` that wraps the sharded querier.

pkg/querier/queryrange/results_cache.go
```diff
@@ -49,14 +56,41 @@ func (cfg *ResultsCacheConfig) RegisterFlags(f *flag.FlagSet) {
...
-type ExtractorFunc func(start, end int64, from Response) Response
+// ExtractCacheGenNumbersFromHeaders extracts gen numbers from a response.
+func (PrometheusResponseExtractor) ExtractCacheGenNumbersFromHeaders(from Response) []string {
+	return getHeaderValuesWithName(from, ResultsCacheGenNumberHeaderName)
+}
 
-// Extract calls the `ExtractorFunc` to implements an `Extractor`.
-func (e ExtractorFunc) Extract(start, end int64, from Response) Response {
-	return e(start, end, from)
+// ResponseWithoutHeaders is useful in caching data without headers.
```
jtlisi, line 84: Can you add a comment explaining why?

pkg/querier/queryrange/queryable.go
```diff
@@ -19,18 +20,27 @@ const (
...
+func (q *ShardedQueryable) getResponseHeaders() map[string][]string {
+	return q.sharededQuerier.ResponseHeaders
 }
 
 // ShardedQuerier is a an implementor of the Querier interface.
 type ShardedQuerier struct {
-	Ctx     context.Context
-	Req     Request
-	Handler Handler
+	Ctx                context.Context
```
jtlisi, line 39: Instead of adding the cache generation implementation into the ShardedQuerier can you create a new querier that wraps this functionality?

pkg/querier/queryrange/results_cache.go
jtlisi: This variable name seems to have changed when it was moved into a separate function.

pkg/cortex/modules.go
sandeepsukhani: I added it to modules to keep `api` package purely for defining apis. Will see where I can move it. Thanks!

pkg/chunk/purger/tombstones.go
```diff
@@ -16,6 +20,35 @@ import (
 
 const tombstonesReloadDuration = 5 * time.Minute
 
+type tombstonesLoaderMetrics struct {
+	cacheGenLoadFailures       prometheus.Gauge
```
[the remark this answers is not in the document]
sandeepsukhani: I did it this way to help with alerting. If we keep incrementing it then we won't be able to figure out the state of the metric and decide whether the issue has resolved or not. I might be wrong though here. Do you think it would be easy to determine the alert state if I do not reset it and keep incrementing whenever we fail to load the cache gen?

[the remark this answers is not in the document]
jtlisi: In our alerts we can use a rate function which will let us see the state at any given time. There should be no need to reset it. Just increment it every time a failure occurs.

If we keep it as a gauge, that leaves open the potential to miss failures if the gauge resets before we scrape the instance.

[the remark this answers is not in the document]
sandeepsukhani: That makes sense. Thanks!

pkg/cortex/modules.go
jtlisi: This is fine as a middleware and probably the only way to accomplish this right now.

jtlisi: The api package near the RegisterQuerier code would be good,  and then you can pass the tombstone loader directly into the registration function.

The other option would be adding a `CacheGenNumLoader` interface as a field in the API struct and when the tombstone loader is first instantiated, have it register with the API struct.

pkg/querier/queryrange/results_cache.go
```diff
@@ -195,7 +237,19 @@ func (s resultsCache) handleMiss(ctx context.Context, r Request) (Response, []Ex
 		return response, []Extent{}, nil
 	}
 
-	extent, err := toExtent(ctx, r, response)
+	if s.cacheGenNumberLoader != nil {
```
gouthamve, line 240: We shouldn't be doing it at this level imo. We also cache in `handleHit` case. Can we move this logic into `Do()`?

pkg/chunk/purger/tombstones.go
```diff
@@ -178,9 +243,53 @@ func (tl *TombstonesLoader) loadPendingTombstones(userID string) error {
...
+		level.Error(util.Logger).Log("msg", "error loading cache generation numbers", "err", err)
+		tl.metrics.cacheGenLoadFailures.Add(1)
+		return &cacheGenNumbers{}
+	}
+
+	tl.cacheGenNumbersMtx.Lock()
+	defer tl.cacheGenNumbersMtx.Unlock()
+
+	tl.cacheGenNumbers[userID] = genNumbers
+	tl.metrics.cacheGenLoadFailures.Set(0)
```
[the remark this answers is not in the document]
sandeepsukhani: This is already taken care of in the last commit.

pkg/chunk/purger/tombstones.go
```diff
@@ -6,6 +6,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/pkg/errors"
+
+	"github.com/prometheus/client_golang/prometheus"
```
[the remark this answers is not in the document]
sandeepsukhani: Sorry my bad, will fix it.

bboreham:
Hi, looking over this change I am wondering whether it is really necessary to drop the cache for chunks?  We cannot change the contents of a chunk without changing its key.

sandeepsukhani:
> Hi, looking over this change I am wondering whether it is really necessary to drop the cache for chunks? We cannot change the contents of a chunk without changing its key.

You are right! Will create a PR. Thanks for the catch!

PR #2280, opened 2020-03-17
Shortcut DoParallelQueries for single-query case

bboreham:
The single query case is very common, e.g. for series-to-chunks lookup in schema v9.
[EDIT: is it? I may be confused about that]

Shortcutting it avoids the creation of two goroutines and associated data structures, and removes an uninteresting tracing span.

Before that, small refactor to make the code overall a bit simpler.

**Checklist**
- NA Tests updated
- NA Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
   * `-flusher.concurrent-flushes` for number of concurrent flushes.
   * `-flusher.flush-op-timeout` is duration after which a flush should timeout.
 * [ENHANCEMENT] Better re-use of connections to DynamoDB and S3. #2268
-* [ENHANCEMENT] Fewer goroutines for queries. #2280
+* [ENHANCEMENT] Reduce number of goroutines used while executing a single index query. #2280
 * [ENHANCEMENT] Experimental TSDB: Add support for local `filesystem` backend. #2245
 * [ENHANCEMENT] Allow 1w (where w denotes week) and 1y (where y denotes year) when setting table period and retention. #2252 
 * [ENHANCEMENT] Added FIFO cache metrics for current number of entries and memory usage. #2270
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
   * `-flusher.concurrent-flushes` for number of concurrent flushes.
   * `-flusher.flush-op-timeout` is duration after which a flush should timeout.
 * [ENHANCEMENT] Better re-use of connections to DynamoDB and S3. #2268
+* [ENHANCEMENT] Fewer goroutines for queries. #2280
```
pracucci: Could you me more explicit, please? Maybe something like `Reduce number of goroutines used while executing a single index query` ?

pkg/chunk/util/util.go
```diff
@@ -15,7 +15,7 @@ import (
 // DoSingleQuery is the interface for indexes that don't support batching yet.
 type DoSingleQuery func(
 	ctx context.Context, query chunk.IndexQuery,
-	callback func(chunk.ReadBatch) bool,
+	callback func(chunk.IndexQuery, chunk.ReadBatch) bool,
```
pracucci: What do you think to replace `callback func(chunk.IndexQuery, chunk.ReadBatch) bool` with `callback Callback` given we have a type for that?

bboreham: Good spot: I missed that we already have that type.

CHANGELOG.md
bboreham: Sure.  TBH I'm unclear exactly where this helps; I was chasing a situation where one query creates 7,000 goroutines but now I'm fairly sure this isn't the fix.  But yes, your wording is clearer.

pracucci:
@bboreham This PR did look fine at a first review. I did left few minor comments. Would you mind addressing them and merging?

pstibrany:
> LGTM. Is this ready to merge?

Nah, tests are failing :(

pracucci:
> Nah, tests are failing :(

They were actually cancelled. I've triggered another run, let's see.

PR #2281, opened 2020-03-17
Allow usage of host lookups for memcache discovery

### LINKED ISSUE #1505: Can't statically (w/o SRV lookup) set memcached servers
Hello, sorry for all the issues I'm creating today :D

I have the ingesters working but not being able to push to memcached (in aws) I tried pushing an item with another pod in the same network and namespace and I was able to do it:

```
root@ubuntu-649f64d646-482db:/# telnet test-cortex.xxxx.cfg.use1.cache.amazonaws.com 11211
Trying 10.80.17.123...
Connected to test-cortex.xxxx.cfg.use1.cache.amazonaws.com.
Escape character is '^]'.
get greeting
END
set greeting 1 0 11
Hello world
STORED
get greeting
VALUE greeting 1 11
Hello world
END
```

But ingester is complaining:

```
ingester-84c4b85c4d-sfdsm ingester level=error ts=2019-07-10T16:49:07.529243532Z caller=memcached.go:230 msg="failed to put to memcached" name= err="memcache: no servers configured or available"
ingester-84c4b85c4d-sfdsm ingester level=error ts=2019-07-10T16:49:07.535683451Z caller=memcached.go:230 msg="failed to put to memcached" name= err="memcache: no servers configured or available"
ingester-84c4b85c4d-sfdsm ingester level=error ts=2019-07-10T16:49:07.539257288Z caller=memcached.go:230 msg="failed to put to memcached" name= err="memcache: no servers configured or available"
ingester-84c4b85c4d-sfdsm ingester level=error ts=2019-07-10T16:49:07.541753931Z caller=memcached.go:230 msg="failed to put to memcached" name= err="memcache: no servers configured or available"
```

The relevant part in the config:

```
      -memcached.hostname=test-cortex.xxxxx.cfg.use1.cache.amazonaws.com:11211
      -memcached.timeout=100ms
      -memcached.service=memcached
```

I wonder if it's related to https://github.com/cortexproject/cortex/issues/1501

gouthamve:
Fixes #1505


gouthamve:
> and infer in code whether we need to do srv lookup or not

How?

annanay25:
Default -> srv lookup.
If `port` flag is supplied -> no srv lookup.

pstibrany:
Why don’t we always use standard hostname lookup?

pstibrany:
Perhaps we can do something like Memberlist does when joining the cluster, https://github.com/hashicorp/memberlist/blob/842cc1703d4e141a9fe096c37149652f1e1bb875/memberlist.go#L249, which works just find with headless k8s service. It translates single hostname to list of IP addresses. It uses TCP first to avoid truncating the response.

pstibrany:
> Perhaps we can do something like Memberlist does when joining the cluster, https://github.com/hashicorp/memberlist/blob/842cc1703d4e141a9fe096c37149652f1e1bb875/memberlist.go#L249, which works just find with headless k8s service. It translates single hostname to list of IP addresses — endpoints in the service. It uses TCP first to avoid truncating the response.

OK, I admit that doing the same would require bit more code :-(

As for testing, integration test looks like a good option to me, for hostname based config.

docs/configuration/config-file-reference.md
```diff
@@ -2114,6 +2114,15 @@ The `memcached_client_config` configures the client used to connect to Memcached
 # Use consistent hashing to distribute to memcache servers.
 # CLI flag: -<prefix>.memcached.consistent-hash
 [consistent_hash: <boolean> | default = false]
+
+# If we should do an SRV lookup. Else we do an host lookup. Port must be
+# provided if this is set to false.
+# CLI flag: -<prefix>.memcached.use-srv
+[use_srv: <boolean> | default = true]
```
pracucci: I've the feeling this may be difficult to extend over the time (ie. what when we'll add a 3rd service discovery mode? another boolean flag?).

From a config perspective, I believe we should have 1 single config option to specify the endpoints. An idea is to:
- Deprecate `hostname`
- Introduce `endpoints` which accepts a comma separated list of endpoints. Endpoints can be:
  - `hostname:port` (so no need for a dedicated `port`). This will be a straight DNS resolution (from record name to A/AAAA record, which is the IP address)
  - `dnssrvnoa+` prefix (ie. `dnssrvnoa+<hostname>`) which is the SRV NOA address resolution (which is that we're supporting right now). This is one of the prefixes supported by Thanos service discovery that will also be supported by the memcached client used by the blocks index cache (PR will land soon). Re-using the same Thanos' strategy would allow us to uniform service discovery across Cortex.

If you don't want to reinvent the wheel, you can just reuse the Thanos service discovery:
https://thanos.io/service-discovery.md/

gouthamve:
Hi everyone, based on the feedback, I think the Thanos DNS SD makes the most sense here. But I want to hold off on deprecating the existing setup but rather I want to mark the new Thanos based system as experimental. Curious what you think. I'll work on tests now.

pkg/chunk/cache/memcached_client.go
```diff
@@ -30,27 +43,35 @@ type serverSelector interface {
...
 
 // MemcachedClientConfig defines how a MemcachedClient should be constructed.
 type MemcachedClientConfig struct {
-	Host           string        `yaml:"host"`
-	Service        string        `yaml:"service"`
-	Timeout        time.Duration `yaml:"timeout"`
-	MaxIdleConns   int           `yaml:"max_idle_conns"`
-	UpdateInterval time.Duration `yaml:"update_interval"`
-	ConsistentHash bool          `yaml:"consistent_hash"`
+	Host           string        `yaml:"host,omitempty"`
```
pracucci: We've just get rid of all `omitempty`. We shouldn't re-introduce them, or am I missing anything?

pkg/chunk/cache/memcached_client.go
```diff
@@ -114,17 +140,25 @@ func (c *memcachedClient) updateLoop(updateInterval time.Duration) {
...
-	_, addrs, err := net.LookupSRV(c.service, "tcp", c.hostname)
-	if err != nil {
-		return err
-	}
 	var servers []string
-	for _, srv := range addrs {
-		servers = append(servers, fmt.Sprintf("%s:%d", srv.Target, srv.Port))
+
+	if len(c.addresses) > 0 {
+		c.provider.Resolve(context.Background(), c.addresses)
```
pracucci: Maybe worth to add a sane hardcoded timeout to the context, to avoid the  request stuck for any reason. Even an high timeout (10s) will do the job.

PR #2290, opened 2020-03-18
Added memcached support to the blocks storage index cache

pracucci:
**What this PR does**:
In this PR I've added memcached support to the blocks storage index cache.

The PR is quite large due to a refactoring I did in the index cache metrics re-mapping, moving it to the generic storage package. The index cache metrics mapping hasn't changed, while I've added the memcached one.

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/util/metrics_helper_test.go
```diff
@@ -197,6 +197,78 @@ func TestSendMaxOfGauges(t *testing.T) {
 	require.ElementsMatch(t, expected, actual)
 }
 
+func TestSendSumOfHistogramsWithLabels(t *testing.T) {
```
pstibrany: As this test function contains multiple tests, have you considered using blocks to separate them more visibly in the code, and to avoid mixing variables between them?

docs/operations/query-auditor.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Query Auditor (tool)"
 linkTitle: "Query Auditor (tool)"
-weight: 2
+weight: 3
```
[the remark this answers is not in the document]
pracucci: It's the order of the pages in the menu. Lower weight, higher in the menu.

pkg/util/metrics_helper_test.go
pracucci: Good point, done.

PR #2293, opened 2020-03-19
Introduced a basic logger to the integration tests framework

pracucci:
**What this PR does**:
We have some flaky integration tests in `query_frontend_test.go`. I can't understand the root cause but I've the feeling it's a timing issue. In this PR I'm adding a very simple logger to prefix each log with the time, so that we can have a better idea where the time is being spent when running in CI.

An example of the tests output is:
```
=== RUN   TestQueryFrontendWithChunksStorageViaConfigFile
10:00:47 Starting consul
10:00:48 consul: ==> Starting Consul agent...
10:00:48 consul: ==> Consul agent running!
10:00:48 consul: Version: 'v0.9.4'
10:00:48 consul: Node ID: 'b67d8cc8-029c-d764-dc65-a7205203b3cd'
10:00:48 consul: Node name: 'consul'
```

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,7 +156,7 @@ jobs:
           export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:${CIRCLE_TAG:-$(./tools/image-tag)}"
           export CORTEX_CHECKOUT_DIR="/home/circleci/.go_workspace/src/github.com/cortexproject/cortex"
           echo "Running integration tests with image: $CORTEX_IMAGE"
-          go test -tags=requires_docker -timeout 300s -v -count=1 ./integration/...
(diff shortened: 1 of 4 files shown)


pracucci:
And thanks to the timing displayed in logs I've also found and fixed the root cause, which is trivial. We currently have a 300s timeout for the integration tests, but due to the new tests we're adding, the whole suite sometimes takes more than 5 minutes to run in CI and thus it breaks due to the max execution time reached.

I've increased the timeout to 10 minutes ([commit](https://github.com/cortexproject/cortex/pull/2293/commits/329b4a94cd3dd1f087591c493b55e98d402c7f93)). In the future we may experiment with parallel integration tests execution.

PR #2298, opened 2020-03-19
Add ability for test-exporter to only query a specified duration in the past

joe-elliott:
**What this PR does**:
Adds a new flag to test-exporter to restrict the range to a certain time in the past: `-test-query-since`

There is no test-exporter documentation so I made no changes.  Additionally I did not feel this was worth a changelog entry, but I would be glad to add one if we'd like it.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/testexporter/correctness/runner.go
+++ b/pkg/testexporter/correctness/runner.go
@@ -62,17 +62,18 @@ func init() {
 
 // RunnerConfig is config, for the runner.
 type RunnerConfig struct {
-	testRate         float64
-	testQueryMinSize time.Duration
-	testQueryMaxSize time.Duration
-	testTimeEpsilon  time.Duration
-	testEpsilon      float64
-	prometheusAddr   string
-	userID           string
-	MinTime          TimeValue
-	extraSelectors   string
-	ScrapeInterval   time.Duration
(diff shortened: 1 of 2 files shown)


pkg/testexporter/correctness/runner.go
```diff
@@ -86,14 +87,27 @@ func (cfg *RunnerConfig) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.userID, "user-id", "", "UserID to send to Cortex.")
 
 	// By default, we only query for values from when this process started
-	cfg.MinTime = NewTimeValue(time.Now())
-	f.Var(&cfg.MinTime, "test-query-start", "Minimum start date for queries")
+	f.Var(&cfg.timeQueryStart, "test-query-start", "Minimum start date for queries")
+	f.DurationVar(&cfg.durationQuerySince, "test-query-since", 0, "Duration in the past to test.  Overrides --test-query-start")
```
gouthamve: Only a single `-`, `-test-query-start` iirc.

joe-elliott: good catch.  fixed.

PR #2299, opened 2020-03-19
Switch pkg/ingester to promauto.With(reg)

pracucci:
**What this PR does**:
Following the work related to https://github.com/cortexproject/cortex/issues/2204 to get rid of global metrics and registerer, in this PR I've migrated `pkg/ingester` metrics to `promauto.With(registerer)`.

Few notes:
- There's still a global metric in `pkg/ingester/client` which I will address in a separate PR
- Not much happy with the path followed for the WAL and `memorySeries`. The problem is that the initial design doesn't look much clean in regards of global variables and I had to deal with that in order to keep the changeset small (but I admit there's much room for improvement design-wise there)

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/metrics.go
+++ b/pkg/ingester/metrics.go
@@ -54,7 +54,7 @@ type ingesterMetrics struct {
 	oldestUnflushedChunkTimestamp prometheus.Gauge
 }
 
-func newIngesterMetrics(r prometheus.Registerer, registerMetricsConflictingWithTSDB bool) *ingesterMetrics {
+func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSDB bool) *ingesterMetrics {
 	m := &ingesterMetrics{
 		flushQueueLength: promauto.With(r).NewGauge(prometheus.GaugeOpts{
 			Name: "cortex_ingester_flush_queue_length",
@@ -186,7 +186,7 @@ func newIngesterMetrics(r prometheus.Registerer, registerMetricsConflictingWithT
 		}),
 	}
 
-	if registerMetricsConflictingWithTSDB {
+	if createMetricsConflictingWithTSDB {
 		m.memSeriesCreatedTotal = promauto.With(r).NewCounterVec(prometheus.CounterOpts{
 			Name: memSeriesCreatedTotalName,
 			Help: memSeriesCreatedTotalHelp,
--- a/pkg/ingester/metrics.go
+++ b/pkg/ingester/metrics.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/cortexproject/cortex/pkg/util"
 )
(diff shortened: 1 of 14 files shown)


pkg/ingester/transfer.go
```diff
@@ -695,15 +661,18 @@ func transferUser(ctx context.Context, stream client.Ingester_TransferTSDBClient
 				return err
 			}
 
-			if err := batchSend(1024*1024, b, stream, &client.TimeSeriesFile{
+			bytesSent, err := batchSend(1024*1024, b, stream, &client.TimeSeriesFile{
```
codesome: Not sure if it is relevant, but if a single file transfer takes a long time, then the metric increment wont be reflected till we have sent the entire file. Do you think we should pass the metric in the function instead so that the gradual increment is visible?

pkg/ingester/metrics.go
```diff
@@ -28,93 +29,173 @@ type ingesterMetrics struct {
...
-			m.walCorruptionsTotal,
-		)
-
-		if registerMetricsConflictingWithTSDB {
-			r.MustRegister(
-				m.memSeriesCreatedTotal,
-				m.memSeriesRemovedTotal,
-			)
-		}
+	if registerMetricsConflictingWithTSDB {
```
pstibrany: Previously these metrics were always created but not registered, just for safety. Now they are not created at all if `registerMetricsConflictingWithTSDB` is false. I've checked access to these two metrics, and this change looks safe to me... that is, these two metrics are not accessed, if metrics were created without them.

Should we rename `registerMetricsConflictingWithTSDB` now?

pkg/ingester/transfer.go
pracucci: @gouthamve @pstibrany Opinion on this? I see a value in what Ganesh is mentioning, while returning the sent bytes looks cleaner on the code side.

pstibrany: No strong opinions from my side, but at previous jobs our admins preferred to see ongoing transfers in metrics, not just big "jump" at the end of transfer. If it requires some non-trivial code changes, I'd suggest doing it in separate PR.

pracucci: Right. Let's get stick to the previous behaviour. I restored it, tracking the metric directly in `batchSend()`.

pkg/ingester/metrics.go
pracucci: Wow, very deep though. I agree renaming it. Done.

gouthamve:
> Not much happy with the path followed for the WAL and memorySeries.

Can you clarify this?

pracucci:
> Can you clarify this? Just curious as I didn't find anything obviously off when I glanced through the code. Is this about the integration tests or something about the WAL metrics themselves?

@gouthamve I haven't been clear, sorry. My comment is about the change I did to pass the metric instances to some functions (`batchSend()`, `newMemorySeries()`, `processCheckpointRecord()`) because it looks a bad design to me. I'm not happy with my changes but I picked that path to avoid to do a larger refactoring and keeping the change set easier to diff and understand.

PR #2303, opened 2020-03-19
Make each DNS provider register a diff metric

### LINKED ISSUE #2302: Crash on startup with duplicate metrics collector registration
As of commit 59bbc46c, in ingester, querier and ruler.
Beginning of log:

```
level=info ts=2020-03-19T17:56:09.556287259Z caller=server.go:147 http=[::]:80 grpc=[::]:9095 msg="server listening on addresses"

goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*Registry).MustRegister(0xc0000ce550, 0xc000838670, 0x1, 0x1)
	/go/pkg/mod/github.com/prometheus/client_golang@v1.5.0/prometheus/registry.go:400 +0xad
github.com/thanos-io/thanos/pkg/discovery/dns.NewProvider(0x3a17800, 0xc00013b650, 0x3a70400, 0xc0000ce550, 0x3388a7a, 0x6, 0xc0008385f0)
	/go/pkg/mod/github.com/thanos-io/thanos@v0.11.0/pkg/discovery/dns/provider.go:74 +0x36a
github.com/cortexproject/cortex/pkg/chunk/cache.NewMemcachedClient(0x7ffd811bb6a1, 0x26, 0x338f0c5, 0x9, 0x0, 0x0, 0x5f5e100, 0x40, 0xdf8475800, 0x0, ...)
	/go/src/github.com/cortexproject/cortex/pkg/chunk/cache/memcached_client.go:101 +0x19c
```

I guess we need integration tests that include memcached.

gouthamve:

Fixes #2302 

**Checklist**
- [x] Tests updated

### DIFF
--- a/integration/querier_test.go
+++ b/integration/querier_test.go
@@ -3,10 +3,13 @@
 package main
 
 import (
+	"fmt"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/prompb"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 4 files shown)


pkg/chunk/cache/memcached_client.go
```diff
@@ -98,8 +98,10 @@ func NewMemcachedClient(cfg MemcachedClientConfig, name string) MemcachedClient
 		hostname:   cfg.Host,
 		service:    cfg.Service,
 		addresses:  strings.Split(cfg.Addresses, ","),
-		provider:   dns.NewProvider(util.Logger, prometheus.DefaultRegisterer, dns.GolangResolverType),
-		quit:       make(chan struct{}),
+		provider: dns.NewProvider(util.Logger, prometheus.WrapRegistererWith(prometheus.Labels{
```
pracucci: This PR made me realise that the Thanos DNS provider intentionally (and correctly) register metrics without prefix. We should also `WrapRegistererWithPrefix()` with the prefix `cortex_`.

pkg/chunk/cache/memcached_client.go
```diff
@@ -98,8 +98,10 @@ func NewMemcachedClient(cfg MemcachedClientConfig, name string) MemcachedClient
...
+			"name": name,
```
pracucci: Is `name` guaranteed to be unique across the codebase? I've seen we're using `cfg.Prefix` as name (which looks another bad smell to me because it's not a name) so I would like to make sure uniqueness is guaranteed.

pkg/chunk/cache/memcached_client.go
pstibrany: Can we supply registrerer to `NewMemcachedClient` function instead of using `prometheus.DefaultRegisterer`? We're moving to that approach across the codebase, this PR seems like a nice opportunity to do it here.

pkg/chunk/cache/memcached_client.go
gouthamve: Yes it should be, as multiple registrations of the prefix would be caught by flag registration.

pkg/chunk/cache/memcached_client.go
```diff
@@ -92,13 +92,17 @@ func NewMemcachedClient(cfg MemcachedClientConfig, name string) MemcachedClient
 	client.Timeout = cfg.Timeout
 	client.MaxIdleConns = cfg.MaxIdleConns
 
+	registerer := prometheus.WrapRegistererWithPrefix("cortex_", prometheus.WrapRegistererWith(prometheus.Labels{
```
pracucci: I would just suggest to rename `registerer` to `dnsProviderRegisterer` to avoid in the future people will erroneously use the replaced `registerer` to register other metrics.

integration/querier_test.go
```diff
@@ -251,3 +254,88 @@ func TestQuerierWithBlocksStorageOnMissingBlocksFromStorage(t *testing.T) {
...
+
+		var series []prompb.TimeSeries
+		series, expectedVectors[u] = generateSeries("series_1", now)
+
+		res, err := c.Push(series)
+		require.NoError(t, err)
+		require.Equal(t, 200, res.StatusCode)
+	}
+
+	// Add 2 memcache instances to test for: https://github.com/cortexproject/cortex/issues/2302
```
bboreham: Are these memcached instances actually running, or this is just a fake to trigger the behaviour?

integration/querier_test.go
```diff
@@ -251,3 +254,88 @@ func TestQuerierWithBlocksStorageOnMissingBlocksFromStorage(t *testing.T) {
...
+	querierFlags := mergeFlags(flags, map[string]string{
+		"-store.index-cache-read.memcached.addresses":  "dns+memcached0:11211",
+		"-store.index-cache-write.memcached.addresses": "dns+memcached1:11211",
+	})
+	// Start the querier and configure it to not frequently sync blocks.
```
bboreham: What does "not frequently sync blocks" mean here?

gouthamve:
Sorry for the stale comments (copied the test from the blocks storage test). Clarified them now.

PR #2306, opened 2020-03-20
Add Prometheus to local dev env setup

pracucci:
**What this PR does**:
The easiest way to locally run Cortex with the experimental blocks storage is using the development setup and running `./development/tsdb-blocks-storage-s3/compose-up.sh`. This is also the same way I use to locally run Cortex while developing it.

A missing thing is that the local dev setup doesn't scrape Cortex metrics itself, so it's not easy to manually test exported metrics. In this PR I'm setting up Prometheus to scrape Cortex metrics from containers and push them to Cortex via remote write.

Notes:
- I've added some `.dockerignore` to speed up building the Cortex container on OSX if the volumes are big (because the entire content of shared dirs need to be transferred to the Docker context if not ignored)

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/development/tsdb-blocks-storage-s3/docker-compose.yml
+++ b/development/tsdb-blocks-storage-s3/docker-compose.yml
@@ -32,7 +32,7 @@ services:
     volumes:
       - ./config:/etc/prometheus
     ports:
-      - 8010:8010
+      - 9090:9090
 
   distributor:
     build:
--- a/development/tsdb-blocks-storage-s3/docker-compose.yml
+++ b/development/tsdb-blocks-storage-s3/docker-compose.yml
@@ -26,6 +26,14 @@ services:
     volumes:
       - .data-configstore:/usr/share/nginx/html/private:ro
 
+  prometheus:
+    image: prom/prometheus:v2.16.0
+    command: ["--config.file=/etc/prometheus/prometheus.yaml"]
+    volumes:
+      - ./config:/etc/prometheus
+    ports:
+      - 8010:8010
+
   distributor:
     build:
       context:    .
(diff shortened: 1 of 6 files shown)


pracucci:
@rfratto Thanks for the agent!

gouthamve:
Why not use Prometheus itself? What does the agent give us over Prometheus?

pracucci:
> Why not use Prometheus itself? What does the agent give us over Prometheus?

I actually wanted to give a try to the Grafana agent to learn about it. Prometheus would work as well.

gouthamve:
Haha, lets stick to Prometheus here rather than introduce another component for people to be familiar with :) While the agent is close to Prometheus, it might confuse new contributors (who would already be familiar with Prometheus mostly).

development/tsdb-blocks-storage-s3/docker-compose.yml
```diff
@@ -26,6 +26,14 @@ services:
     volumes:
       - .data-configstore:/usr/share/nginx/html/private:ro
 
+  prometheus:
+    image: prom/prometheus:v2.16.0
+    command: ["--config.file=/etc/prometheus/prometheus.yaml"]
+    volumes:
+      - ./config:/etc/prometheus
+    ports:
+      - 8010:8010
```
gouthamve: Prometheus runs on 9090 right?

pracucci: Damn, yes! Thanks for spotting it! I was left there from the previous Grafana agent setup.

PR #2307, opened 2020-03-20
Renamed table-manager metrics to remove cortex_dynamo prefix

pracucci:
**What this PR does**:
I've noticed that a couple of table-manager metrics have the `cortex_dynamo` prefix. I'm proposing to rename them to the `cortex_table_manager` prefix for clarity.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -47,7 +47,7 @@ func newTableManagerMetrics(r prometheus.Registerer) *tableManagerMetrics {
 
 	m.tableCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
-		Name:      "table_manager_table_capacity_units",
+		Name:      "table_manager_table_capacity",
 		Help:      "Per-table provisioned read and write capacity units.",
 	}, []string{"op", "table"})
 
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -47,8 +47,8 @@ func newTableManagerMetrics(r prometheus.Registerer) *tableManagerMetrics {
 
 	m.tableCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
-		Name:      "table_manager_table_capacity",
-		Help:      "Per-table provisioned read and write capacity units.",
+		Name:      "dynamo_table_capacity_units",
+		Help:      "Per-table DynamoDB capacity, measured in DynamoDB capacity units.",
 	}, []string{"op", "table"})
 
 	m.createFailures = prometheus.NewGauge(prometheus.GaugeOpts{
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -41,7 +41,7 @@ func newTableManagerMetrics(r prometheus.Registerer) *tableManagerMetrics {
 	m.syncTableDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: "cortex",
 		Name:      "table_manager_sync_duration_seconds",
-		Help:      "Time spent doing synching tables.",
+		Help:      "Time spent synching tables.",
 		Buckets:   prometheus.DefBuckets,
 	}, []string{"operation", "status_code"})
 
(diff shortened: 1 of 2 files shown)


pracucci:
@bboreham What's your take on this renaming?

pkg/chunk/table_manager.go
```diff
@@ -40,15 +40,15 @@ func newTableManagerMetrics(r prometheus.Registerer) *tableManagerMetrics {
 	m := tableManagerMetrics{}
 	m.syncTableDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: "cortex",
-		Name:      "dynamo_sync_tables_seconds",
-		Help:      "Time spent doing SyncTables.",
+		Name:      "table_manager_sync_duration_seconds",
+		Help:      "Time spent doing synching tables.",
```
bboreham:
```suggestion
		Help:      "Time spent synching tables.",
```

pkg/chunk/table_manager.go
```diff
@@ -40,15 +40,15 @@ func newTableManagerMetrics(r prometheus.Registerer) *tableManagerMetrics {
...
 		Buckets:   prometheus.DefBuckets,
 	}, []string{"operation", "status_code"})
 
 	m.tableCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
-		Name:      "dynamo_table_capacity_units",
-		Help:      "Per-table DynamoDB capacity, measured in DynamoDB capacity units.",
+		Name:      "table_manager_table_capacity_units",
```
bboreham:
```suggestion
		Name:      "table_provisioned_capacity",
```

pracucci:
> Maybe I'm out of the loop re prefixing metrics with component name, but table_manager doesn't add much, and units is opaque.

I 100% agree on removing the `_units` since it's opaque (done). The reason I added the prefix `cortex_table_manager` (even if a bit redundant for the capacity units specific metric) is because if we have a single prefix per service it comes a bit easier to identify metrics exposed by a single Cortex service. Maybe it's just my paranoia, and I'm open to reevaluate it but considering it shouldn't do any harm (not violating any best practice) we may consider to keep it.

pracucci:
> Do you want to do cortex_dynamo_consumed_capacity_total at the same time?

I'm not sure it's correct to remove the `dynamo` wording from it, because it looks to be DynamoDB specific. Am I missing anything? Maybe you were referring to something different?

bboreham:
I cannot see that the consumed capacity is any more or less DynamoDB-specific than the provisioned capacity.

> single prefix per service [...] shouldn't do any harm

Until we move the place a metric is generated, split one service into two, etc.

pracucci:
> I cannot see that the consumed capacity is any more or less DynamoDB-specific than the provisioned capacity.

To my understanding, the consumed capacity metrics are increased only when you the DynamoDB client, while the provisioned capacity metric value is set regardless of backend index store used.

>> single prefix per service [...] shouldn't do any harm
> Until we move the place a metric is generated, split one service into two, etc.

Given moving a piece of logic from a service to another doesn't look a frequent operation, we can always rename them, like we're renaming them in this PR. As of today, we have decided to not guarantee any non breaking changes in metrics even after Cortex 1.0, so until this policy won't change doesn't look a big deal to me (given it's not my/our intention to abuse of it, but it's option we'll have).

bboreham:
> the provisioned capacity metric value is set regardless of backend index store used

It's set to zero for everything except DynamoDB. This is not a good basis to argue from.

Any store _could_ have a concept of provisioned and consumed capacity.  The ones that I know _do_ are DynamoDB and CosmosDB, but Cortex doesn't yet have any code to set and fetch CosmosDB capacity.

bboreham:
No wait, it won't be zero, it will be the default flag values of 1000 and 300.
Still, it has no meaning for every store except DynamoDB.

pkg/chunk/table_manager.go
```diff
@@ -40,15 +40,15 @@ func newTableManagerMetrics(r prometheus.Registerer) *tableManagerMetrics {
...
+		Help:      "Time spent doing synching tables.",
 		Buckets:   prometheus.DefBuckets,
 	}, []string{"operation", "status_code"})
 
 	m.tableCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
-		Name:      "dynamo_table_capacity_units",
-		Help:      "Per-table DynamoDB capacity, measured in DynamoDB capacity units.",
+		Name:      "table_manager_table_capacity",
+		Help:      "Per-table provisioned read and write capacity units.",
```
tomwilkie: I don't think we should rename this one, its dynamodb specific.

pracucci:
> "doing synching tables" doesn't sound right to me, but I guess that text is unlikely to be read by many people.

@bboreham I think you already gave me this feedback previously and I lost the fix at some point. Sorry for that. Accepted your suggestion to fix it 🙏

PR #2308, opened 2020-03-20
Improve metrics integration tests

pracucci:
**What this PR does**:
While working on the metrics registration refactoring I've realized the current `integration/metrics_test.go` is not that good because doesn't exercise much Cortex code.

In this PR I've removed the metrics-specific test and I've added the same assertions to all other integration tests, so that more Cortex code is exercised (ie. chunks storage, blocks storage, caching, ...).

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


integration/asserts.go
```diff
@@ -0,0 +1,76 @@
...
+		Ingester:      []string{},
+		Querier:       []string{},
+		QueryFrontend: []string{"cortex_frontend", "cortex_query_frontend"},
+		TableManager:  []string{},
+		AlertManager:  []string{},
+	}
+
+	// Blacklisted metrics prefixes across any Cortex service.
+	blacklistedMetricsPrefixes = []string{
+		"cortex_alert_manager", // It should be "cortex_alertmanager"
```
jtlisi: I'm a bit confused where this is coming from?

pracucci: There's no such metric prefix. I just wanna avoid we'll add it in the future. I think it's a trivial mistake to do.

For example in the past we did a similar mistake with the query frontend and now we have mixed prefixes `cortex_query_frontend` and `cortex_frontend`.

Makes sense to you? If it's just my paranoia I can remove this.

jtlisi: Fine to leave it IMO

PR #2309, opened 2020-03-20
Fix multiple registrations with multi schema

### LINKED ISSUE #2302: Crash on startup with duplicate metrics collector registration

gouthamve:
Fixes #2302 

@bboreham Can you test if this actually fixes the issue?

### DIFF
--- a/pkg/chunk/cache/memcached_client.go
+++ b/pkg/chunk/cache/memcached_client.go
@@ -101,11 +101,14 @@ func NewMemcachedClient(cfg MemcachedClientConfig, name string, r prometheus.Reg
 		serverList: selector,
 		hostname:   cfg.Host,
 		service:    cfg.Service,
-		addresses:  strings.Split(cfg.Addresses, ","),
 		provider:   dns.NewProvider(util.Logger, dnsProviderRegisterer, dns.GolangResolverType),
 		quit:       make(chan struct{}),
 
 		numServers: memcacheServersDiscovered.WithLabelValues(name),
+	}
+
+	if len(cfg.Addresses) > 0 {
+		newClient.addresses = strings.Split(cfg.Addresses, ",")
 	}
 
 	err := newClient.updateMemcacheServers()
--- a/pkg/chunk/storage/factory.go
+++ b/pkg/chunk/storage/factory.go
@@ -95,14 +95,28 @@ func (cfg *Config) Validate() error {
 
 // NewStore makes the storage clients based on the configuration.
 func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConfig, limits StoreLimits) (chunk.Store, error) {
-	tieredCache, err := cache.New(cfg.IndexQueriesCacheConfig)
+	indexReadCache, err := cache.New(cfg.IndexQueriesCacheConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	writeDedupeCache, err := cache.New(storeCfg.WriteDedupeCacheConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	chunkCacheCfg := storeCfg.ChunkCacheConfig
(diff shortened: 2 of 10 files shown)


pkg/chunk/cache/memcached_client.go
```diff
@@ -92,17 +94,28 @@ func NewMemcachedClient(cfg MemcachedClientConfig, name string, r prometheus.Reg
...
+	provider, ok := providersForName[name]
+	if !ok {
+		dnsProviderRegisterer := prometheus.WrapRegistererWithPrefix("cortex_", prometheus.WrapRegistererWith(prometheus.Labels{
+			"name": name,
+		}, r))
+
+		provider = dns.NewProvider(util.Logger, dnsProviderRegisterer, dns.GolangResolverType)
+
+		providersForName[name] = provider
+	}
```
pstibrany: I think this belongs outside of this function. Perhaps caller can maintain the map of dns providers, and pass the correct one to the new memcached client?

bboreham:
It stops the crashing, but caches no longer work - just get `caller=memcached.go:164 msg="Failed to get keys from memcached" err="dial tcp :0: connect: connection refused"`

integration/configs.go
```diff
@@ -21,7 +21,17 @@ const (
 	bucketName             = "cortex"
 	cortexConfigFile       = "config.yaml"
 	cortexSchemaConfigFile = "schema.yaml"
-	storeConfigTemplate    = `
+	// Adding an old schema so we test multi schema issues.
```
pracucci: I think this may break the usage in `integration/chunks_storage_backends_test.go`. Given the `chunks_storage_backends_test.go` already builds a multi schema config, I'm not really sure you need to add this. I think we're already covered there. Am I missing anything?

pkg/chunk/storage/factory.go
```diff
@@ -95,14 +95,28 @@ func (cfg *Config) Validate() error {
...
 	if err != nil {
 		return nil, err
 	}
 
 	// Cache is shared by multiple stores, which means they will try and Stop
 	// it more than once.  Wrap in a StopOnce to prevent this.
-	tieredCache = cache.StopOnce(tieredCache)
+	indexReadCache = cache.StopOnce(indexReadCache)
+	chunksCache = cache.StopOnce(chunksCache)
+	writeDedupeCache = cache.StopOnce(writeDedupeCache)
```
pstibrany: This is a strange design... if store isn't initializing the cache, it should not stop it. Cache should only be stopped after the *last* store stops using it, not by first store. Something to look at in the future I guess.

PR #2314, opened 2020-03-23
Gettings started with gossip ring.

pstibrany:
**What this PR does**: this PR adds quick-start guide for gossiped ring, with two example configuration files.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/configuration/single-process-config-blocks-gossip-1.yaml
+++ b/docs/configuration/single-process-config-blocks-gossip-1.yaml
@@ -1,0 +1,83 @@
+
+# Configuration for running Cortex in single-process mode.
+# This should not be used in production.  It is only for getting started
+# and development.
+
+# Disable the requirement that every request to Cortex has a
+# X-Scope-OrgID header. `fake` will be substituted in instead.
+auth_enabled: false
+
+server:
+  http_listen_port: 9109
+  grpc_listen_port: 9195
+
+  # Configure the server to allow messages up to 100MB.
+  grpc_server_max_recv_msg_size: 104857600
+  grpc_server_max_send_msg_size: 104857600
+  grpc_server_max_concurrent_streams: 1000
+
+distributor:
+  shard_by_all_labels: true
+  pool:
+    health_check_ingesters: true
+
+ingester_client:
+  grpc_client_config:
+    # Configure the client to allow messages up to 100MB.
+    max_recv_msg_size: 104857600
+    max_send_msg_size: 104857600
+    use_gzip_compression: true
+
+ingester:
+  # Disable blocks transfers on ingesters shutdown or rollout.
+  max_transfer_retries: 0
+
+  lifecycler:
+    # The address to advertise for this ingester.  Will be autodiscovered by
+    # looking up address on eth0 or en0; can be specified if this fails.
+    address: 127.0.0.1
+    # Defaults to hostname, but we run both ingesters in this demonstration on the same machine.
+    id: "Ingester 1"
+
+    # We don't want to join immediately, but wait a bit to see other ingesters and their tokens first.
+    # It can take a while to have the full picture when using gossip
+    join_after: 10s
+
+    # To avoid generating same tokens by multiple ingesters, they can "observe" the ring for a while,
+    # after putting their own tokens into it. This is only useful when using gossip, since multiple
+    # ingesters joining at the same time can have conflicting tokens if they don't see each other yet.
+    observe_period: 10s
+    min_ready_duration: 0s
+    claim_on_rollout: false
+    final_sleep: 5s
+    num_tokens: 512
+
+    # Use an in memory ring store, so we don't need to launch a Consul.
+    ring:
+      kvstore:
+        store: memberlist
+
+      replication_factor: 1
+
+memberlist:
+  # defaults to hostname
+  node_name: "Ingester 1"
+  bind_port: 7946
+  join_members:
+    - localhost:7947
+  abort_if_cluster_join_fails: false
+
+storage:
+  engine: tsdb
+
+tsdb:
+  dir: /tmp/cortex/tsdb-ing1
+  bucket_store:
+    sync_dir: /tmp/cortex/tsdb-sync-querier1
+
+  # This is where Cortex uploads generated blocks. Queriers will fetch blocks from here as well.
+  # Cortex of course supports multiple options (S3, GCS, Azure), but for demonstration purposes
+  # we only use shared directory.
+  backend: filesystem # s3, gcs, azure or filesystem are valid options
+  filesystem:
+    dir: /tmp/cortex/storage
--- a/docs/getting-started/getting-started-with-gossip-ring.md
+++ b/docs/getting-started/getting-started-with-gossip-ring.md
@@ -1,0 +1,98 @@
+---
+title: "Getting Started with Gossiped Ring (experimental)"
+linkTitle: "Gossip Ring (experimental)"
+weight: 2
+slug: getting-started-with-gossiped-ring
+---
+
+Cortex requires Key-Value store to store the ring. It can use traditional KV stores like Consul or Etcd,
+but now it can also build its own KV store on top of memberlist library using a gossip algorithm.
+
+This short guide shows how to start Cortex in [single-binary mode](../architecture.md) with memberlist-based ring.
+To reduce number of required dependencies in this guide, it will use [blocks storage](../operations/blocks-storage.md) with no shipping to external stores.
+Storage engine and external storage configuration are not dependant on the ring configuration.
+
+## Single-binary, two Cortexes
+
+
+
+For simplicity and to get started, we'll run it as a two instances of Cortex on local computer.
+We will use prepared configuration files ([file 1](../configuration/single-process-config-blocks-gossip-1.yaml), [file 2](../configuration/single-process-config-blocks-gossip-2.yaml)), with no external
+dependencies.
+
+Build Cortex first:
+```sh
+$ go build ./cmd/cortex
+```
+
+First instance:
+```
+$ ./cortex -config.file docs/configuration/single-process-config-blocks-gossip-1.yaml
+```
+
+Second instance is similar, but uses `...-2.yaml` file.
+```
+$ ./cortex -config.file docs/configuration/single-process-config-blocks-gossip-2.yaml
+```
+
+The two instances should be able to find each other via memberlist configuration:
+
+```yaml
+memberlist:
+  # defaults to hostname
+  node_name: "Ingester 1"
+  bind_port: 7946
+  join_members:
+    - localhost:7947
+  abort_if_cluster_join_fails: false
+```
+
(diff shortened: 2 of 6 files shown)


docs/configuration/single-process-config-blocks-gossip-1.yaml
```diff
@@ -0,0 +1,83 @@
+
```
pracucci: We're used to add an integration test for each getting started config we do provide. May you add a test with this config to `integration/getting_started_single_process_config_test.go`, please?

docs/configuration/single-process-config-blocks-gossip-1.yaml
```diff
@@ -0,0 +1,83 @@
...
+    use_gzip_compression: true
+
+ingester:
+  # Disable blocks transfers on ingesters shutdown or rollout.
+  max_transfer_retries: 0
+
+  lifecycler:
+    # The address to advertise for this ingester.  Will be autodiscovered by
+    # looking up address on eth0 or en0; can be specified if this fails.
+    address: 127.0.0.1
```
pracucci: Is it intentionally uncommented? If so, isn't a problem the fact it's the same of the `-2.yaml` config?

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
...
+but now it can also build its own KV store on top of memberlist library using a gossip algorithm.
+
+This short guide shows how to start Cortex in [single-binary mode](../architecture.md) with memberlist-based ring.
+To reduce number of required dependencies in this guide, it will use [blocks storage](../operations/blocks-storage.md) with no shipping to external stores.
+Storage engine and external storage configuration are not dependant on the ring configuration.
+
+## Single-binary, two Cortexes
+
+
+
```
pracucci: Extra white spaces

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
+---
+title: "Getting Started with Gossiped Ring (experimental)"
+linkTitle: "Gossip Ring (experimental)"
+weight: 2
+slug: getting-started-with-gossiped-ring
+---
+
+Cortex requires Key-Value store to store the ring. It can use traditional KV stores like Consul or Etcd,
+but now it can also build its own KV store on top of memberlist library using a gossip algorithm.
```
pracucci: I would remove any temporal reference (ie. "now") from the doc, given it's likely this doc will survive a long time.

```suggestion
but it can also build its own KV store on top of memberlist library using a gossip algorithm.
```

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
+---
+title: "Getting Started with Gossiped Ring (experimental)"
+linkTitle: "Gossip Ring (experimental)"
+weight: 2
+slug: getting-started-with-gossiped-ring
+---
+
+Cortex requires Key-Value store to store the ring. It can use traditional KV stores like Consul or Etcd,
```
pracucci:
```suggestion
Cortex requires Key-Value (KV) store to store the ring. It can use traditional KV stores like Consul or Etcd,
```

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
...
+but now it can also build its own KV store on top of memberlist library using a gossip algorithm.
+
+This short guide shows how to start Cortex in [single-binary mode](../architecture.md) with memberlist-based ring.
+To reduce number of required dependencies in this guide, it will use [blocks storage](../operations/blocks-storage.md) with no shipping to external stores.
+Storage engine and external storage configuration are not dependant on the ring configuration.
+
+## Single-binary, two Cortexes
```
pracucci:
```suggestion
## Single-binary, two Cortex instances
```

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+
+For simplicity and to get started, we'll run it as a two instances of Cortex on local computer.
+We will use prepared configuration files ([file 1](../configuration/single-process-config-blocks-gossip-1.yaml), [file 2](../configuration/single-process-config-blocks-gossip-2.yaml)), with no external
```
pracucci: These links don't work. I think it's due to the fact that the current page ends with `/`, so here you would need `../../`. However, I believe it would be wise to also manage `.yaml` files in the `convertLinks()` in `tools/website/website.go`

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
...
+dependencies.
+
+Build Cortex first:
+```sh
+$ go build ./cmd/cortex
+```
+
+First instance:
```
pracucci: Could be simplified as:

Run two instances of Cortex, each one with its own dedicated config file:

```
$ ./cortex -config.file docs/configuration/single-process-config-blocks-gossip-1.yaml
$ ./cortex -config.file docs/configuration/single-process-config-blocks-gossip-2.yaml
```

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
...
+```
+$ ./cortex -config.file docs/configuration/single-process-config-blocks-gossip-1.yaml
+```
+
+Second instance is similar, but uses `...-2.yaml` file.
+```
+$ ./cortex -config.file docs/configuration/single-process-config-blocks-gossip-2.yaml
+```
+
+The two instances should be able to find each other via memberlist configuration:
```
pracucci: To a first time reader, it may not be clear if the following config is a change they should do or it's just you explaining how things work under the hood. An alternative may be explaining how to run Cortex in this section and then adding another section below (ie. "How it works") explaining how things work under the hood. In this way we keep the two things separated.

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,98 @@
...
+distribute incoming samples between the two instances.
+
+To query that data, you can configure your Grafana instance to use http://localhost:9109/api/prom (first Cortex) as a Prometheus data source.
+
+## How to add another instance?
+
+To add another Cortex to the small cluster, copy `docs/configuration/single-process-config-blocks-gossip-1.yaml` to a new file,
+and make following modifications. We assume that third Cortex will run on the same machine again, so we change node name and ingester ID as well:
+
+- change `http_listen_port` and `grpc_listen_port` to unique port numbers (eg. 9309 and 9395 respectively)
```
pracucci: Instead of listing the changes this way, have you thought about providing an annotated `diff` showing the changes required, similarly to what we did in http://localhost:1313/docs/changelog/#annotated-config-file-breaking-changes ? I'm just wondering if may be clearer to follow.

docs/configuration/single-process-config-blocks-gossip-1.yaml
pstibrany: No, since Addr field in the ring contains both address and grpc port.

docs/getting-started/getting-started-with-gossip-ring.md
pstibrany: From my understanding (mind you, I'm Hugo noob), relref only works when used with another page. I think we might be able to use static files feature instead -- move config files under /static, and include them into markdown pages. But in this PR I have only fixed the links by adding another `../` prefix.

docs/configuration/single-process-config-blocks-gossip-1.yaml
pstibrany: Added (although in a different filename).

docs/getting-started/getting-started-with-gossip-ring.md
pstibrany: Good suggestion, thanks!

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,186 @@
...
+distribute incoming samples between the two instances.
+
+To query that data, you can configure your Grafana instance to use http://localhost:9109/api/prom (first Cortex) as a Prometheus data source.
+
+## How to add another instance?
+
+To add another Cortex to the small cluster, copy `docs/configuration/single-process-config-blocks-gossip-1.yaml` to a new file,
+and make following modifications. We assume that third Cortex will run on the same machine again, so we change node name and ingester ID as well:
+
+```diff
```
pracucci: What do you think if we diff only the config that needs to be changed instead of copy-pasting the entire config? The main risk is that if we copy-paste the entire config, this will drift the first time we'll change the `.yaml` (and will happen).

docs/getting-started/getting-started-with-gossip-ring.md
```diff
@@ -0,0 +1,186 @@
...
+
+```yaml
+remote_write:
+- url: http://localhost:9109/api/prom/push
+```
+
+After starting Prometheus, it will now start pushing data to Cortex. Distributor component in Cortex will
+distribute incoming samples between the two instances.
+
+To query that data, you can configure your Grafana instance to use http://localhost:9109/api/prom (first Cortex) as a Prometheus data source.
```
pracucci: This should be moved under `Single-binary, two Cortex instances`, because it's part of the getting started to have a working setup.

docs/getting-started/getting-started-with-gossip-ring.md
pstibrany: Good idea.

pstibrany:
Thanks for review Marco, I've addressed your feedback. Can you please recheck?

tomwilkie:
Is it worth having two getting started guides?  I would remove the old one and just have getting started (with gossip).

pstibrany:
> Is it worth having two getting started guides? I would remove the old one and just have getting started (with gossip).

We already have two (getting started with chunks, blocks), this is third (gossip, but still using blocks). Getting started with blocks goes into more details and examples how to do HA with replication using docker compose and Consul, perhaps we can replace it with gossip example instead? WDYT @pracucci?

pracucci:
> We already have two (getting started with chunks, blocks), this is third (gossip, but still using blocks). Getting started with blocks goes into more details and examples how to do HA with replication using docker compose and Consul, perhaps we can replace it with gossip example instead? WDYT @pracucci?

I personally don't see anything bad having different getting started guides, based on the kind of setup you're interested into (chunks or blocks? gossip?).

pstibrany:
> I personally don't see anything bad having different getting started guides, based on the kind of setup you're interested into (chunks or blocks? gossip?).

The thing is, getting started with gossip is basically based on getting started with blocks, but adds multiple single-binary Cortex instances. Perhaps we can split "getting started with blocks" into "basic setup" and "high-availability part" and have alternative "HA part" based on gossip.

I can do that in another PR.

pstibrany:
> I can do that in another PR.

If you agree

PR #2315, opened 2020-03-23
Proposal: blocks storage sharding

pracucci:
**What this PR does**:
Let's start a discussion how to reliably shard blocks to horizontally scale the read path.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Trade-offs
+
+The proposed solution comes with the following trade-offs:
+
+- A querier is not ready until it has completed an initial full scan of the bucket, downloading the `meta.json` file of every block
+- A store-gateway is not ready until it has completed an initial full scan of the bucket, downloading the `meta.json` and index header of each block matching its shard
+- If a querier hits 2+ store-gateways it may receive duplicated series if the 2+ store-gateways share some blocks due to the replication factor
```
pstibrany: All of these look fine to me. Querier is already fetching meta.json and index headers, so it will do less work now.

Querier will also do deduplication, as it does today when handling results from multiple streaming ingesters.

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,77 @@
...
+
+1. Shard blocks (index headers) across a pool of nodes
+2. Do not compromise HA on the read path (if a node fails, queries should continue to work)
+3. Do not compromise correctness (either the query result is correct or it fails)
+
+## Proposed solution
+
+The idea is to introduce a new Cortex service - `store-gateway` - internally running the Thanos [`BucketStore`](https://github.com/thanos-io/thanos/blob/master/pkg/store/bucket.go). At query time, a querier will run a query fetching the matching series both from ingesters and the subset of gateways holding the related blocks (based on the query time range). Blocks are replicated across the gateways in order to guarantee query results consistency and HA even in the case of a gateway instance failure.
+
+### Ring-based sharding and replication
```
pstibrany: Are store-gateways going to transfer blocks within them when joining / leaving the ring? Would it make sense to do so?

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,77 @@
...
+4. Merge and deduplicate received series
+   - Optimization: can be skipped if the querier hits only 1 store-gateway
+
+### Query results consistency
+
+When a querier executes a query, it should guarantee that either all blocks matching the time range are queried or the query fails.
+
+However, due to the (intentional) lack of a strong coordination between queriers and store-gateways, and the ring topology which can change any time, there's no guarantee that the blocks assigned to a store-gateway shard are effectively loaded on the store-gateway itself at any given point in time.
+
+The idea is introduce a **consistency check in the querier**. When a store-gateway receives a request from the querier, the store-gateway includes in the response the list of block IDs currently loaded on the store-gateway itself. The querier can then merge the list of block IDs received from all store-gateway hit, and match it against the list of block IDs computed at the beginning of the query execution.
```
pstibrany: > the store-gateway includes in the response the list of block IDs currently loaded on the store-gateway itself

Isn't this going to be big? Could store-gateway return block IDs that were used to handle the query?

Alternatively, could querier tell store-gateway, which block it expects that store-gateway should use? (List of blocks it "must" use, and list of blocks it "could" use -- for new blocks. This can also trigger store-gateway to do a new scan for those blocks, if they are missing).

docs/proposals/blocks-storage-sharding.md
pracucci: Looks an unnecessary complexity. From a performances perspective, I don't expect any significant difference transferring them between store-gateways instead of re-downloading them from the storage.

docs/proposals/blocks-storage-sharding.md
pracucci: > Isn't this going to be big?

Given a query is bounded to 1 tenant, 1 block per day = 365 blocks per 1 year. Considering a block ID is 16 bytes, 365 blocks = 5.7KB.

> Could store-gateway return block IDs that were used to handle the query?

The store-gateway should just need to return the blocks matching the query time range. It's an optimization we can/should put in place.

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+2. **On-demand**<br/>
+   when the ring topology changes (the tokens ranges assigned to the gateway instance have changed)
+
+It's important to outline that the sync takes time (typically will have to re-scan the bucket and download new blocks index headers) and Cortex needs to guarantee query results consistency at any given time (_see below_).
+
+### Query execution
+
+When a querier executes a query, it will need to fetch series both from ingesters and the store-gateway instances.
+
+For a given query, the number of blocks to query is expected to be low, especially if the Cortex cluster is running the `query-frontend` with a `24h` query split interval. In this scenario, whatever is the client's query time range, the `query-frontend` will split the client's query into sub-queries each with up to `24h` time range and the querier will likely hit not more than 1 block per sub-query (except for the last 24h for which blocks may have not been compacted yet).
```
bwplotka: Let's not mix Prometheus `subqueries` with this (: I think we used other terms like `separate-queries` or something

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+Given this assumption, we want to avoid sending every query to every store-gateway instance. The querier should be able to take an informed decision about the minimum subset of store-gateway instances which needs to query given a time range.
+
+The idea is to run the [`MetaFetcher`](https://github.com/thanos-io/thanos/blob/master/pkg/block/fetcher.go#L127) also within the querier, but without any sharding filter (contrary to the store-gateway). At any given point in time, the querier knows the entire list of blocks in the storage. When the querier executes the `Select()` (or `SelectSorted()`) it does:
```
bwplotka: Nice! Interesting idea :+1: 

cc @brancz as we already talked about block-aware querying for Querier. I wish we could do it on Thanos, but there it's not that easy - not all StoreAPI are block based, so our Querier does not have this knowledge.

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+1. Compute the list of blocks by the query time range
```
bwplotka:
```suggestion
1. Compute the list of blocks by the query time range.
```

and below :hugs:

docs/proposals/blocks-storage-sharding.md
bwplotka: This is actually a good fit for `Info` (something we have in StoreAPI) endpoint rather... Have you considered this?

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+However, due to the (intentional) lack of a strong coordination between queriers and store-gateways, and the ring topology which can change any time, there's no guarantee that the blocks assigned to a store-gateway shard are effectively loaded on the store-gateway itself at any given point in time.
+
+The idea is introduce a **consistency check in the querier**. When a store-gateway receives a request from the querier, the store-gateway includes in the response the list of block IDs currently loaded on the store-gateway itself. The querier can then merge the list of block IDs received from all store-gateway hit, and match it against the list of block IDs computed at the beginning of the query execution.
+
+There are three possible scenarios:
+
+1. The list match: all good
+2. All the blocks known by the querier are within the list of blocks returned by store-gateway, but the store-gateway also included blocks unknown to the querier: all good (it means the store-gateways have discovered and loaded new blocks before the querier discovered them)
+3. Some blocks known by the querier are **not** within the list of blocks returned by store-gateway: potential consistency issue
```
bwplotka: There is also 4th case: Since you look on response result only you need to first request things from it to tell... That's why `Info` method  might be a better fit

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+We want to protect from a partial results response which may occur in the case #3. However, there are some legit cases which, if not handled, would lead to frequent false positives. Given the querier and store-gateway instances independently scan the bucket at a regular interval (to find new blocks or deleted blocks), we may be in one of the following cases:
+
+a. The querier has discovered new blocks before the store-gateway successfully discovered and loaded them
+b. The store-gateway has offloaded blocks "marked for deletion" before the querier
```
bwplotka: It's rather, before other store gateway loaded new ones, right?

Also note as our current logic offloads those blocks UNTIL replacement is visible OR delay happens. (half of delete delay), so you might be fine here. (no `b` case)

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+There are three possible scenarios:
+
+1. The list match: all good
+2. All the blocks known by the querier are within the list of blocks returned by store-gateway, but the store-gateway also included blocks unknown to the querier: all good (it means the store-gateways have discovered and loaded new blocks before the querier discovered them)
+3. Some blocks known by the querier are **not** within the list of blocks returned by store-gateway: potential consistency issue
+
+We want to protect from a partial results response which may occur in the case #3. However, there are some legit cases which, if not handled, would lead to frequent false positives. Given the querier and store-gateway instances independently scan the bucket at a regular interval (to find new blocks or deleted blocks), we may be in one of the following cases:
+
+a. The querier has discovered new blocks before the store-gateway successfully discovered and loaded them
```
bwplotka: Info/response will tell us, right?

docs/proposals/blocks-storage-sharding.md
bwplotka: With such Info endpoint you really don't need querier consistency check at all ,right? :thinking:

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+b. The store-gateway has offloaded blocks "marked for deletion" before the querier
+
+To protect from case (a), we can exclude the blocks which have been uploaded in the last `X` time from the consistency check (same technique already used in other Thanos components). This `X` delay time is used to give the store-gateway enough time to discover and load new blocks, before the querier consider them for the consistency check. This value `X` should be greater than the `-experimental.tsdb.bucket-store.consistency-delay`, because we do expect the querier to consider a block for consistency check once it's reasonably safe to assume that its store-gateway already loaded it.
```
bwplotka: :+1:

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+To protect from case (b) we need to understand how blocks are offloaded. The `BucketStore` (running within the store-gateway) offloads a block as soon as it's not returned by the `MetaFetcher`. This means we can configure the `MetaFetcher` with a [`IgnoreDeletionMarkFilter`](https://github.com/thanos-io/thanos/blob/4bd19b16a752e9ceb1836c21d4156bdeb517fe50/pkg/block/fetcher.go#L648) with a delay of `X` (could be the same value used for case (a)) and in the querier exclude the blocks which have been marked for deletion more than `X` time ago from the consistency check.
```
bwplotka: IMO there is no case b, so skipping reading this part ;p

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Trade-offs
+
+The proposed solution comes with the following trade-offs:
+
+- A querier is not ready until it has completed an initial full scan of the bucket, downloading the `meta.json` file of every block
```
bwplotka: All of those we can fix in later steps with our old proposal of different block ID right @pracucci ?

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+- A store-gateway is not ready until it has completed an initial full scan of the bucket, downloading the `meta.json` and index header of each block matching its shard
+- If a querier hits 2+ store-gateways it may receive duplicated series if the 2+ store-gateways share some blocks due to the replication factor
```
bwplotka: That's fine - During the work on chunk iterator I touch merge series Set implementation and we plan lot's of optimizations ther for **exactly** the same series duplicates etc. :hugs:

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+slug: blocks-storage-sharding
+---
+
+- Author: [Marco Pracucci](https://github.com/pracucci)
+- Date: March 2020
+- Status: proposal
+
+## Problem
+
+In Cortex, when using the experimental blocks storage, each querier internally runs the Thanos [`BucketStore`](https://github.com/thanos-io/thanos/blob/master/pkg/store/bucket.go). This means that each querier has a full view over all blocks in the long-term storage and all blocks index headers are loaded in each querier memory. The querier memory usage linearly increase with number and size of all blocks in the storage, imposing a scalability limit to the blocks storage.
```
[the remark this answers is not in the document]
pracucci: We also observe a quite high memory utilization in our cluster. We have a cluster with 6M active series, 1d blocks, 45 days data so far (13 months is the retention) and the Go heap inuse by `newFileBinaryReader` is 3.5GB. Projecting this to 13 months retention will be 30GB (and 6M active series is quite small).

The 1d block range period is definitely a factor here, but it's a choice made because it couples well with the 1d interval splitting by the query-frontend (once we'll have the query sharding which is the 2nd purpose of this work).

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+
+In this proposal we want to solve this. In particular, we want to:
+
+1. Shard blocks (index headers) across a pool of nodes
+2. Do not compromise HA on the read path (if a node fails, queries should continue to work)
+3. Do not compromise correctness (either the query result is correct or it fails)
+
+## Proposed solution
+
+The idea is to introduce a new Cortex service - `store-gateway` - internally running the Thanos [`BucketStore`](https://github.com/thanos-io/thanos/blob/master/pkg/store/bucket.go). At query time, a querier will run a query fetching the matching series both from ingesters and the subset of gateways holding the related blocks (based on the query time range). Blocks are replicated across the gateways in order to guarantee query results consistency and HA even in the case of a gateway instance failure.
```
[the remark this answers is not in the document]
pracucci: At this stage we would need some glue on top of it (ie. Cortex ring, multi-tenancy, etc), reason why we plan to use the `BucketStore` (wrapped to add multi-tenancy support) and a Cortex store-gateway service.

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
+---
+title: "Blocks storage sharding"
```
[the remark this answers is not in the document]
pracucci: Why is it time sharding? We'll not shard by time.

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
pracucci: The querier will have knowledge about the full list of blocks (and metadata for each block). The querier computes the list of necessary blocks given the query time range and uses it to compute the minimum set of store-gateway nodes to visit.

docs/proposals/blocks-storage-sharding.md
```diff
@@ -0,0 +1,84 @@
...
+2. Do not compromise HA on the read path (if a node fails, queries should continue to work)
+3. Do not compromise correctness (either the query result is correct or it fails)
+
+## Proposed solution
+
+The idea is to introduce a new Cortex service - `store-gateway` - internally running the Thanos [`BucketStore`](https://github.com/thanos-io/thanos/blob/master/pkg/store/bucket.go). At query time, a querier will run a query fetching the matching series both from ingesters and the subset of gateways holding the related blocks (based on the query time range). Blocks are replicated across the gateways in order to guarantee query results consistency and HA even in the case of a gateway instance failure.
+
+### Ring-based sharding and replication
+
+In order to build blocks sharding and replication, the `store-gateway` instances form a [ring](../architecture.md#the-hash-ring). Each gateway instance uses a custom [`MetaFetcherFilter`](https://github.com/thanos-io/thanos/blob/master/pkg/block/fetcher.go#L108) to filter blocks loaded on the instance itself, keeping only blocks whose `hash(block-id)` is within the tokens range assigned to the gateway instance within the ring.
```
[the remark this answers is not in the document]
pracucci: Isn't the `MetaFetcherFilter` more appropriate?

docs/proposals/blocks-storage-sharding.md
pracucci: @owen-d What's the right terminology for the sub-interval queries done by the query-frontend?

docs/proposals/blocks-storage-sharding.md
pracucci: Not sure. According to this proposal we would need the "marked for deletion" information as well, so even if the min/max timestamp is in the storage prefix, it may not be enough.

docs/proposals/blocks-storage-sharding.md
pracucci: We're also already doing this de-duplication in the Cortex querier (at query time) because we need to de-duplicate series merged between ingesters and storage.

docs/proposals/blocks-storage-sharding.md
pracucci: > That's why Info method might be a better fit

We considered using the `Info()` which is basically an API exposed by the store-gateway. However, even if you query all store-gateways to fetch all their `Info()` there's no guarantee that all blocks are loaded across by the running gateways at any given point in time (because of how the re-sharding works whenever the ring topology changes, which hasn't any sort of "distributed lock" around it). For this reason, the proposal is about having the querier keeping an independent list of blocks (without any filter) and to run a consistency check on the querier side.

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
bwplotka: Does the 1/32 symbols, 1/32 posting offset knowledge holds here? 6 millions in each block?

`3758096384` bytes for 6M series in 1d blocks for 45d: `83513252.977777778` per block, `13.918875496` per single series. I think this is fair 1/32 symbols 1/32 posting offset is fair assumption assuming single symbol is like usually ~30B + `uint32` and single posting offset for  binary header is

```
map[string]*postingValueOffsets

type postingValueOffsets struct {
	offsets       []postingOffset
	lastValOffset int64
}

type postingOffset struct {
	// label value.
	value string
	// offset of this entry in posting offset table in index-header file.
	tableOff int
}
```
... so like 100B or something. 

Still 3.5GB is not much for such amount of series and almost 2month of data :man_shrugging:

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
bwplotka: Sad, but you might know better. Curious how much work would be to add missing parts. Ring libs we already depend on as we import Cortex.

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
bwplotka: True, wanted to remove this comment, forgot. Ignore

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
bwplotka: Yes, with the consistent check or some info improved endpoint it makes sense :+1:

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
bwplotka: Relabelling filter is already such filter. So it's just the questions extending or creating new. Up to you whatever is easier to maintain

docs/proposals/blocks-storage-sharding.md
bwplotka: No one knows, but subqueries are wrong one. `Partitioned Query?`

docs/proposals/blocks-storage-sharding.md
bwplotka: But that is not the design. You also propose something like Info from result, which is weird.

docs/proposals/blocks-storage-sharding.md
bwplotka: good point. But at least it will improve a lot.

docs/proposals/blocks-storage-sharding.md
owen-d: I think we call them `split` internally, but I'm not opinionated. `partitioned` sound good.

owen-d:
@pracucci great job :). Gave it a read and have a few questions. Please correct my misunderstandings.

1) I see a fourth case for queriers contacting gateways: needed blocks for the query are not yet registered by either service. This would impede our correctness guarantees.
2) Consistency checks are unbounded right now, yes? Can we limit them by time (i.e. do one consistency check for each `<time_interval>` in parallel?) Then we'd worry less about the unbounded nature of our _data_ because our _queries_ are not unbounded.
3) If a querier does not have a complete blocks list internally, it may not hit a store gateway that contains blocks it _does_ need for the query, compromising correctness. In other words, the querier does not guarantee it checks all the right gateways.
4)  Are startup times going to be a problem b/c queriers/gateways need to download the whole store’s metadata?

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
gouthamve: Heh, but we need to scale to 120Mil active series with 13months of retention ;) If not more.

docs/proposals/blocks-storage-sharding.md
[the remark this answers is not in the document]
gouthamve: Yep, the idea is to upstream it once we have things working. But we want to iterate quickly here and see if this actually makes sense and what issues we'll hit. Once we have this working at scale, we'll send a proposal to Thanos and basically upstream the entire thing :)

docs/proposals/blocks-storage-sharding.md
gouthamve: Yep the problem is when do we query the info endpoint? Periodically, before we send the query, or after we send the query?

The problem with periodically and after we send the query is that the list is not consistent, but maybe we could do it right before we send the query to the -gateway. But do we really want to do 2 RPC calls when we could do one?

bwplotka: periodically.

bwplotka: if you do it before, after or even in the middle! There will be inconsistency **anyway** ... This how distributed systems work ;p

gouthamve: Not if its in the same call ;)

Again the problem is we don't want to return a partial result. The problem is suppose, we're doing it every 10secs. And we synced at `t0`, and the topology changed at `t0+8s`, and we sent a query at `t0+9s`. We might have missed a block because the gateway doesn't have it anymore but the querier thinks the gateway has it and might respond with a gap in the result.

It is not OK in the case of Cortex to return inconsistent result. We'd rather fail the query.

docs/proposals/blocks-storage-sharding.md
thorfour: If each `store-gateway` is holding the blocks assigned to it, would that mean that we'd be storing all of S3, in these gateways? Or does the `MetaFectherFilter` only store the metadata for it's assigned blocks, and then caches recently read blocks? If so that could be made more clear here.

docs/proposals/blocks-storage-sharding.md
thorfour: Is there a reason why it needs to be a new service? Could we not simply have each querier maintain it's own shard of blocks, and then ask other queriers when necessary?

docs/proposals/blocks-storage-sharding.md
bwplotka: Good point, missed that. With the same call and addting that to response, can help to give more consistent insight :+1: 
Overhead does not problematic.

docs/proposals/blocks-storage-sharding.md
pstibrany: Perhaps. Having it as a separate component allows independent scaling of queriers and store gateways. If we later realize that they can run together, we can simply use multiple targets when starting them (support for that is coming soon).

khaines: +1 for a separate component. I can see this running in-process with the querier with multi-targeting on smaller setups but completely separate on larger configurations.

docs/proposals/blocks-storage-sharding.md
khaines: My 2cents is go with `partitioned` here.

pracucci: Updated the doc using "partitioned query". Thanks for the feedback.

docs/proposals/blocks-storage-sharding.md
pracucci: The `MetaFetcherFilter` filters out blocks not assigned to a specific `store-gateway` shard. For the matching blocks, only the binary block index header is downloaded. The full index and chunks never hit the `store-gateway` disk and they're directly streamed from the storage (S3/GCS) using byte-range requests (which is how it works right now).

docs/proposals/blocks-storage-sharding.md
pracucci: In order to run the consistency check on the querier side, we need to know (on the querier side) which block have been hit for a given `Series()` API call in the `store-gateway`. This is the reason why the `Series()` response is expected to also list the block IDs hit by the request, so that the querier can merge all block IDs contained in `store-gateway` responses (in case 2+ gateway have been hit) and match against the list of blocks computed on the querier side (based on the known blocks by the `MetaFetcher` running within the querier itself).

pracucci:
Thanks @owen-d for your feedback. Please see my reply below.

> I see a fourth case for queriers contacting gateways: needed blocks for the query are not yet registered by either service. This would impede our correctness guarantees.

It's definitely a case, but doesn't break correctness guarantees due to how the system work. New blocks can be uploaded to the storage by two entities:

- Ingesters
- Compactor

The ingesters, after uploading blocks to the storage, do not immediately delete blocks locally. So there's a time window (in our cluster 1h) during which the querier will both query shipped blocks both from ingesters and querier and then merge (+ deduplicate) matching series.

The compactor does download some blocks from the storage, locally "compact" these blocks and upload the compacted block to the storage. There's an improvement in Thanos (which we still have to integrate) that avoids the source blocks to be immediately deleted by the compactor, leaving a large time window during which both source and compacted blocks are in storage.

> Consistency checks are unbounded right now, yes? Can we limit them by time (i.e. do one consistency check for each <time_interval> in parallel?) Then we'd worry less about the unbounded nature of our data because our queries are not unbounded.

The consistency check is based on the query time range.

> If a querier does not have a complete blocks list internally, it may not hit a store gateway that contains blocks it does need for the query, compromising correctness. In other words, the querier does not guarantee it checks all the right gateways.

The querier always have the full list of blocks seen during the periodic last bucket scan. The case of "new blocks" uploaded since the last bucket scan is covered by the first question/answer above.

> Are startup times going to be a problem b/c queriers/gateways need to download the whole store’s metadata?

Today, the querier startup time is made of: scan + download meta and index header of each block. After this proposal:

- Querier: scan bucket (1K objects per API request) + download metadata of every block (few KB for each block)
- Gateway: scan bucket + download meta of blocks matching the shard + download index header of blocks matching the shard

For this reason, the startup time is primarily dependent on the number of blocks (currently we have 1 block per day per tenant). It's definitely something we can re-iterate on (ie. caching in front of metadata) and we're currently mitigating it using stateful sets with persistent disks (because both metadata and index header are cached on disk), so that the startup time is still significative for scale ups, but not that much for rolling updates.

bwplotka:
All, before we merge, can we quickly mention one alternative to this, which is a centralized controller that would coordinate this, and what would be trade-off here?

I am asking because what we see here is really similar to what we are doing in Thanos Receiver with help of https://github.com/observatorium/thanos-replicate, but we do that for writes. Curious what you think. (: 

cc @brancz

pracucci:
> All, before we merge, can we quickly mention one alternative to this, which is a centralized controller that would coordinate this, and what would be trade-off here?

@bwplotka Without looking into all details, I think an external controller would work as well. Building the sharding on the ring looks a more natural fit for Cortex: other forms of sharding we have for ingesters, rulers, compactor work on the ring, and this proposal is just an extension of that. From a UX perspective, avoiding to introduce a controller (unless strictly required) would allow to keep the overall archicture more consistent with how the sharding works for other Cortex services.

bwplotka:
Cool, I just wanted to make sure we considered that alternative. SGTM (:

pracucci:
> Cool, I just wanted to make sure we considered that alternative. SGTM (:

Thanks to you @bwplotka for all the ideas and feedback you continuously share ❤️

pracucci:
I've changed the status of the proposal to "approved". Will keep it updated if, while building it, we'll find any blocker that would lead us to change the design.

PR #2317, opened 2020-03-23
Enabling availability zone awareness in metric R/W with ingesters.

### LINKED ISSUE #612: Make distributors/ingesters aware of availability zones

khaines:
**What this PR does**:  This PR adds support for defining an availability zone for ingesters so that metric writes & reads will be assigned across zone boundaries. This is added as a CLI flag or config file field, as this feature is agnostic to how that particular is set/discovered; as it varies greatly based on the given environment configuration.

Fixes #612

**Checklist**
- [x] Tests updated
- [x] Documentation added


khaines:
The lint failure is because of the doc check. The default for the zone setting at the moment is the hostname, which makes the document verification expectations dynamic. I'm working on a solution to this.

khaines:
A great suggestion @tomwilkie. I'll work on that right away and get a PR for it out fast.

bboreham:
Looks like this broke my staging system:
```
time="2020-03-29T22:43:33Z" level=warning msg="POST /api/prom/push (500) 2.254634ms Response: \"at least 2 live replicas required, could only find 1\\n\" ws: false; Content-Encoding: snappy; Content-Length: 16461; Content-Type: application/x-protobuf; User-Agent: Prometheus/2.12.0; X-Prometheus-Remote-Write-Version: 0.1.0; " traceID=77042d0abcee28fd
```

docs/guides/zone-replication.md
```diff
@@ -0,0 +1,30 @@
+---
+title: "Ingester Hand-over"
+linkTitle: "Ingester Hand-over"
+weight: 5
+slug: ingester-handover
```
pracucci: @khaines This comes from a bad copy-paste. May you submit a PR to fix it, please?

docs/guides/zone-replication.md
```diff
@@ -0,0 +1,30 @@
...
+---
+
+In a default configuration, time-series written to ingesters are replicated based on the container/pod name of the ingester instances. It is completely possible that all the replicas for the given time-series are held with in the same availability zone, even if the cortex infrastructure spans multiple zones within the region. Storing multiple replicas for a given time-series poses a risk for data loss if there is an outage affecting various nodes within a zone or a total outage.
+
+## Configuration
+
+Cortex can be configured to consider an availability zone value in its replication system. Doing so mitigates risks associated with losing multiple nodes with in the same availability zone. The availability zone for an ingester can be defined on the command line of the ingester using the `ingester.availability-zone` flag or using the yaml configuration:
```
pracucci: @khaines `with in` > `within`

pkg/ring/lifecycler.go
```diff
@@ -103,6 +104,7 @@ func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.Flag
 	f.StringVar(&cfg.Addr, prefix+"lifecycler.addr", "", "IP address to advertise in consul.")
 	f.IntVar(&cfg.Port, prefix+"lifecycler.port", 0, "port to advertise in consul (defaults to server.grpc-listen-port).")
 	f.StringVar(&cfg.ID, prefix+"lifecycler.ID", hostname, "ID to register into consul.")
+	f.StringVar(&cfg.Zone, prefix+"availability-zone", "", "The availability zone of the host, this instance is running on. Default is the lifecycler ID.")
```
pracucci: @khaines `Default is the lifecycler ID.` This looks more an internal implementation detail than something we should let the final user know. For the final user the default is an empty string, which means the zone-awareness is disabled.

docs/guides/zone-replication.md
khaines: yes it did... looked at it a few times and this didn't catch my eye.

pkg/ring/lifecycler.go
khaines: Made an edit to this with https://github.com/cortexproject/cortex/pull/2384

PR #2319, opened 2020-03-24
FIFO cache to support eviction based on memory usage

### LINKED ISSUE #2316: FIFO cache is limited by the number of entries, but not by their total size
FIFO cache limits number of entries, regardless of their size. I would suggest using the total memory size of the cache as the threshold to trigger eviction. That way we will have better control over memory usage, and it will be easier to configure k8s resource requests/limits.
I can put together a PR, if this proposal will be accepted.

dmitsh:

**What this PR does**:
Change FIFO cache to evict entries based on memory usage.

Fixes #2316 

**Checklist**
- [ x] Tests updated
- [ x] Documentation added


docs/configuration/config-file-reference.md
```diff
@@ -2206,9 +2206,9 @@ The `fifo_cache_config` configures the local in-memory cache. The supported CLI
 &nbsp;
 
 ```yaml
-# The number of entries to cache.
-# CLI flag: -<prefix>.fifocache.size
-[size: <int> | default = 0]
+# Maximum memory size of the cache.
+# CLI flag: -<prefix>.fifocache.max-size-bytes
```
khaines: Thank you for taking the time to propose this change. I'm still going over the details here, but something that has stuck out to me is that currently, this is a breaking change due to the fact that this is replacing the previous implementation based on the number of items. I feel we should be adding this size based management in addition to the count based for the time being.

pracucci: > I feel we should be adding this size based management in addition to the count based for the time being.

Should we? Technically, this (before Cortex 1.0) may be the right time to do this kind of improvements which imply having some breaking changes. This looks quite a visible breaking change to the user (Cortex will not start if the old config is passed) and doesn't look much different then the huge list of config breaking changes @gouthamve is introducing.

dmitsh: I can make it not a breaking change.
For example, we can keep "size" parameter and introduce "eviction_method" string parameter, which might be either "entries" (by default) or "memory".
If set to "entries", I will follow the old logic. Otherwise new logic will be involved.

khaines: @pracucci Even if we can because we're before 1.0, I would prefer to keep that for things we just can't get around. To me, this particular change seems like something we can phase-in but I'm open to counterpoints and if I'm the sole one thinking this way, I won't hold things up because of it.

@dmitsh I would go here with keeping both `-<prefix>.fifocache.size` and `-<prefix>.fifocache.max-size-bytes` , with the bytes flag taking precedence. This makes tracking of deprecated flag use more apparent with users.

dmitsh: I updated the code per @khaines recommendations

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [CHANGE] Utilize separate protos for rule state and storage. Experimental ruler API will not be functional until the rollout is complete. #2226
 * [CHANGE] Frontend worker in querier now starts after all Querier module dependencies are started. This fixes issue where frontend worker started to send queries to querier before it was ready to serve them (mostly visible when using experimental blocks storage). #2246
 * [CHANGE] Lifecycler component now enters Failed state on errors, and doesn't exit the process. (Important if you're vendoring Cortex and use Lifecycler) #2251
+* [CHANGE] FIFO cache to evict entries based on used memory size #2319
```
pracucci: Given it's now backward compatible, please switch this entry to an `ENHANCEMENT` and re-phrase the message.

CHANGELOG.md
```diff
@@ -20,7 +21,6 @@
 * [ENHANCEMENT] Added FIFO cache metrics for current number of entries and memory usage. #2270
 * [ENHANCEMENT] Output all config fields to /config API, including those with empty value. #2209
 * [BUGFIX] Fixed etcd client keepalive settings. #2278
-* [BUGFIX] Fixed bug in updating last element of FIFO cache. #2270
```
pracucci: Why this entry has been removed? Looks it was fixed in #2270, no?

pkg/chunk/cache/fifo_cache.go
```diff
@@ -71,27 +71,30 @@ var (
...
+	MaxSize  int           `yaml:"max_size_bytes"`
+	MaxCount int           `yaml:"size"`
 	Validity time.Duration `yaml:"validity"`
 }
 
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
 func (cfg *FifoCacheConfig) RegisterFlagsWithPrefix(prefix, description string, f *flag.FlagSet) {
-	f.IntVar(&cfg.Size, prefix+"fifocache.size", 0, description+"The number of entries to cache.")
+	f.IntVar(&cfg.MaxSize, prefix+"fifocache.max-size-bytes", 0, description+"Maximum memory size of the cache.")
+	f.IntVar(&cfg.MaxCount, prefix+"fifocache.size", 0, description+"The number of entries to cache. (Deprecating, use 'fifocache.max-size-bytes' instead)")
```
pracucci:
```suggestion
	f.IntVar(&cfg.MaxCount, prefix+"fifocache.size", 0, description+"The number of entries to cache. Deprecated: use fifocache.max-size-bytes instead.")
```

pkg/chunk/cache/fifo_cache.go
```diff
@@ -71,27 +71,30 @@ var (
 
 // FifoCacheConfig holds config for the FifoCache.
 type FifoCacheConfig struct {
-	Size     int           `yaml:"size"`
+	MaxSize  int           `yaml:"max_size_bytes"`
```
pracucci: `MaxSize` > `MaxSizeBytes`

pkg/chunk/cache/fifo_cache.go
```diff
@@ -71,27 +71,30 @@ var (
...
+	MaxCount int           `yaml:"size"`
```
pracucci: We may call it `MaxSizeItems` given it's the max cache size measured as max number of items

pkg/chunk/cache/fifo_cache.go
```diff
@@ -71,27 +71,30 @@ var (
...
 type FifoCache struct {
 	lock     sync.RWMutex
-	size     int
+	maxCount int
+	maxSize  int
+	currSize int
 	validity time.Duration
-	entries  []cacheEntry
-	index    map[string]int
+	entries  map[string]*cacheEntry
```
pracucci: Could you explain why we need to change the data structure used for entries?

dmitsh:
> Thanks @dmitsh. Honestly the change set is larger than what I would have expected. I'm not sure I do understand why we need to change how the cache internally store data, instead of just keeping the current cache size (in bytes) and trigger the eviction based on that. I would suggest to see if there's an easier way to implement the bytes-size-based eviction without having to change the data structures.

@pracucci The reason I cannot keep the internal cache storage implementation is because it is tailored for count-based eviction: we allocate the exact number of entries in the array, and play with indices. With the memory usage based eviction we cannot pre-allocate such an array, because we don't know up-front how many entries might fit in the cache.

pkg/chunk/cache/fifo_cache.go
dmitsh: explained below

CHANGELOG.md
dmitsh: Since I changed LRU implementation, this BUGFIX does not apply anymore.

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,144 @@ package cache
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-const size = 10
-const overwrite = 5
-
-func TestFifoCache(t *testing.T) {
-	c := NewFifoCache("test1", FifoCacheConfig{Size: size, Validity: 1 * time.Minute})
+func TestFifoCacheEviction(t *testing.T) {
```
khaines: Are you planning to add an eviction test for the memory limit configuration?

khaines:
Thank you @dmitsh for making this work both with the item count and memory limits. The only item I have outstanding for myself is that one test question.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -69,29 +69,37 @@ var (
...
+	MaxSizeBytes int           `yaml:"max_size_bytes"`
+	MaxSizeItems int           `yaml:"size"`
+	Validity     time.Duration `yaml:"validity"`
 }
 
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
 func (cfg *FifoCacheConfig) RegisterFlagsWithPrefix(prefix, description string, f *flag.FlagSet) {
-	f.IntVar(&cfg.Size, prefix+"fifocache.size", 0, description+"The number of entries to cache.")
+	f.IntVar(&cfg.MaxSizeBytes, prefix+"fifocache.max-size-bytes", 0, description+"Maximum memory size of the cache.")
+	f.IntVar(&cfg.MaxSizeItems, prefix+"fifocache.size", 0, description+"The number of entries to cache. Deprecated, use fifocache.max-size-bytes instead.")
```
khaines: Please update this flag to match the other deprecated flags. Examples can be found if you search for `flagext.DeprecatedFlag` This enables some tracking and warnings to users and the text should start with `DEPRECATED`.

pkg/chunk/cache/fifo_cache_test.go
dmitsh: The second test in this file tests both eviction and expiration for memory-based eviction, but it is a good idea to run these tests for both cases. I'll make a change.

dmitsh:
@khaines @pracucci  What do you think about not deprecating `fifocache.size`? 
We can support both for the same price, and folks will have a choice what to use.
The count-based eviction is more intuitive, and the memory-based eviction can be used in k8s deployments, where memory usage is monitored.
What do you guys think?

dmitsh:
@khaines @pracucci I went ahead and renamed `fifocache.size` to `fifocache.max-size-items`.
It seems to be the right thing to leave both eviction options available.
I also deprecated `fifocache.size`, so this is now a breaking change again.
Hopefully it could make it to the 1.0 release (cc @gouthamve)

pkg/chunk/cache/fifo_cache.go
```diff
@@ -69,29 +71,39 @@ var (
...
 }
 
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
 func (cfg *FifoCacheConfig) RegisterFlagsWithPrefix(prefix, description string, f *flag.FlagSet) {
-	f.IntVar(&cfg.Size, prefix+"fifocache.size", 0, description+"The number of entries to cache.")
+	f.IntVar(&cfg.MaxSizeBytes, prefix+"fifocache.max-size-bytes", 0, description+"Maximum memory size of the cache.")
+	f.IntVar(&cfg.MaxSizeItems, prefix+"fifocache.max-size-items", 0, description+"Maximum number of entries in the cache.")
 	f.DurationVar(&cfg.Validity, prefix+"fifocache.duration", 0, description+"The expiry duration for the cache.")
+
+	flagext.DeprecatedFlag(f, prefix+"fifocache.size", "This flag has been renamed to "+prefix+"fifocache.max-size-items")
```
pracucci: @gouthamve we need this, if we merge the PR before Cortex `1.0`?

pkg/chunk/cache/fifo_cache.go
```diff
@@ -179,91 +191,130 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
-			c.last = entry.prev
+		c.deleteFromList(entry)
+		entrySz := sizeOf(entry)
+		c.currSize -= entrySz
+
+		// New entry size
+		entrySz += (sizeOfInterface(value) - sizeOfInterface(entry.value))
+
+		if c.maxSizeBytes > 0 && c.currSize+entrySz > c.maxSizeBytes {
+			// New entry does not fit in the cache. Delete the entry.
```
pracucci: This is not how an LRU should work. We should evicted the last-recently used entries in the cache until we've freed enough space for this entry.

An easier way to implement this `put()` function is just keeping the "delete the existing item from list" in this `if ok` branch, and then just let re-add it to the data structure below, so that the eviction logic below is triggered also for the case the entry already exists.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -179,91 +191,130 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
+		updated: time.Now(),
+		key:     key,
+		value:   value,
+	}
+	entrySz := sizeOf(entry)
 
-		c.entries[index] = entry
-		c.memoryBytes.Add(float64(deltaSize))
+	// Check is the entry fits in the cache
+	if c.maxSizeBytes > 0 && entrySz > c.maxSizeBytes {
```
pracucci: This check should be moved to the top of the function and done as first thing.

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,161 @@ package cache
...
+		key:   "00",
+		value: 0,
+	}
+	testFifoCacheEviction(t, "test-memory-eviction", FifoCacheConfig{MaxSizeBytes: cnt * sizeOf(itemTemplate), Validity: 1 * time.Minute})
+	testFifoCacheEviction(t, "test-items-eviction", FifoCacheConfig{MaxSizeItems: cnt, Validity: 1 * time.Minute})
+}
 
-func TestFifoCache(t *testing.T) {
-	c := NewFifoCache("test1", FifoCacheConfig{Size: size, Validity: 1 * time.Minute})
+func testFifoCacheEviction(t *testing.T, testname string, cfg FifoCacheConfig) {
```
pracucci: The `cfg` param is unused. I guess you wanted to use it in the `NewFifoCache()`

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,161 @@ package cache
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-const size = 10
-const overwrite = 5
+func TestFifoCacheEviction(t *testing.T) {
+	const (
+		cnt       = 10
+		overwrite = 5
```
pracucci: Unused

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,161 @@ package cache
...
 	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-const size = 10
-const overwrite = 5
+func TestFifoCacheEviction(t *testing.T) {
+	const (
+		cnt       = 10
```
pracucci: This is duplicate between this function and `testFifoCacheEviction()`

pkg/chunk/cache/fifo_cache.go
```diff
@@ -69,29 +71,39 @@ var (
 	}, []string{"cache"})
 )
 
+// This FIFO cache implementation supports two eviction methods - based on number of items in the cache, and based on memory usage.
+// For the memory-based eviction, set FifoCacheConfig.MaxSizeBytes to a positive integer, indicating upper limit of memory allocated by items in the cache.
+// Alternatively, set FifoCacheConfig.MaxSizeItems to a positive integer, indicating maximum number of items in the cache.
+// If both parameters are set, the memory-based eviction method takes precedence.
```
pracucci: What do you mean by this? Looking at the code, if both are set they're both enforced.

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,161 @@ package cache
...
+		overwrite = 5
+	)
+	itemTemplate := &cacheEntry{
```
pracucci: This is duplicated in `testFifoCacheEviction`

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,161 @@ package cache
...
+	testFifoCacheEviction(t, "test-memory-eviction", FifoCacheConfig{MaxSizeBytes: cnt * sizeOf(itemTemplate), Validity: 1 * time.Minute})
+	testFifoCacheEviction(t, "test-items-eviction", FifoCacheConfig{MaxSizeItems: cnt, Validity: 1 * time.Minute})
+}
 
-func TestFifoCache(t *testing.T) {
-	c := NewFifoCache("test1", FifoCacheConfig{Size: size, Validity: 1 * time.Minute})
+func testFifoCacheEviction(t *testing.T, testname string, cfg FifoCacheConfig) {
+	const (
+		cnt       = 10
+		overwrite = 5
```
pracucci: The name `overwrite` makes thinking it's the number of keys which get `put()` again, while it's actually the number of evictions we want to trigger.

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,161 @@ package cache
...
+	assert.Equal(t, testutil.ToFloat64(c.totalMisses), float64(0))
+	assert.Equal(t, testutil.ToFloat64(c.staleGets), float64(0))
+	assert.Equal(t, testutil.ToFloat64(c.memoryBytes), float64(cnt*sizeOf(itemTemplate)))
+
 	// Check evictions
 	keys = []string{}
 	values = []interface{}{}
-	for i := size; i < size+overwrite; i++ {
-		keys = append(keys, strconv.Itoa(i))
+	for i := cnt; i < cnt+overwrite; i++ {
```
pracucci: We should start at `cnt-overwrite` otherwise we miss to check some times. Then the check below on `c.totalGets` should adapted.

```suggestion
	for i := cnt-overwrite; i < cnt+overwrite; i++ {
```

pkg/chunk/cache/fifo_cache.go
```diff
@@ -162,12 +170,16 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
 
 // Stop implements Cache.
 func (c *FifoCache) Stop() {
```
pracucci: You should update metrics here

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,100 +2,161 @@ package cache
...
+	assert.Equal(t, testutil.ToFloat64(c.memoryBytes), float64(cnt*sizeOf(itemTemplate)))
+
+	c.Stop()
 }
 
-func TestFifoCacheEvictionExpiry(t *testing.T) {
-	c := NewFifoCache("test2", FifoCacheConfig{Size: 3, Validity: 5 * time.Millisecond})
-	ctx := context.Background()
-	memorySz := int(unsafe.Sizeof(cacheEntry{})) * 3
+func TestFifoCacheExpiry(t *testing.T) {
```
pracucci: If you change this test to the table pattern you will not have to pass `keys` and `data` to the other function.

pkg/chunk/cache/fifo_cache.go
dmitsh: Disagree. If we already have an entry with that key, and the new entry doesn't fit the cache, we need to remove the old entry first, and then and ignore the new one.

pkg/chunk/cache/fifo_cache.go
dmitsh: At this point I have already deleted the item from the LRU. This code handles a special (unlikely) case where we already have an item with this key in the cache, but the new item is too big to fit the cache.

pkg/chunk/cache/fifo_cache.go
pracucci: You're right. Resolving this comment.

pkg/chunk/cache/fifo_cache.go
pracucci: I may argue this is not an edge case. After some time you run the cache, the cache will run always close to the limit (because we start evicting only once the cache is full), and so hitting this case doesn't look unlikely to me.

dmitsh: You are right, thank you for catching this!

pkg/chunk/cache/fifo_cache.go
gouthamve: Lets remove that flag and mark it as a breaking change.

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@
   - `-<prefix>.memcached.consistent-hash`: Old default: false, new default: true. This allows for better cache hits when the memcaches are scaled up and down.
   - `-querier.batch-iterators`: Old default: false, new default: true.
   - `-querier.ingester-streaming`: Old default: false, new default: true.
+* [CHANGE] FIFO cache to support eviction based on memory usage. #2319
```
pracucci: Please mention the CLI flag breaking change

```suggestion
* [CHANGE] FIFO cache to support eviction based on memory usage. The `-<prefix>.fifocache.size` CLI flag has been renamed to `-<prefix>.fifocache.max-size-items` as well as its YAML config option `size` renamed to `max_size_items `.  #2319
```

pkg/chunk/cache/fifo_cache.go
```diff
@@ -162,12 +173,22 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// update metrics
+	c.entriesEvicted.Add(float64(len(c.entries)))
+	c.entriesCurrent.Set(float64(0))
+	c.memoryBytes.Set(float64(0))
+
+	c.entries = nil
+	c.currSizeBytes = 0
```
pracucci: You should also reset `c.first` and `c.last` to `nil`

pkg/chunk/cache/fifo_cache.go
```diff
@@ -179,91 +200,118 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
-	c.entriesAddedNew.Inc()
+
+	if !ok {
+		c.entriesAddedNew.Inc()
+	}
 
 	// Otherwise, see if we need to evict an entry.
-	if len(c.entries) >= c.size {
+	for (c.maxSizeBytes > 0 && c.currSizeBytes+entrySz > c.maxSizeBytes) || (c.maxSizeItems > 0 && len(c.entries) >= c.maxSizeItems) {
 		c.entriesEvicted.Inc()
```
pracucci: This should be moved within the first branch of the `if` below

pkg/chunk/cache/fifo_cache.go
```diff
@@ -179,91 +200,118 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
-		c.memoryBytes.Add(float64(deltaSize))
-		return
+		if evicted := c.deleteFromTail(); evicted != nil {
+			delete(c.entries, evicted.key)
+			c.currSizeBytes -= sizeOf(evicted)
+			c.entriesCurrent.Dec()
+		} else {
+			c.entries = nil
+			c.currSizeBytes = 0
+			c.entriesCurrent.Set(float64(0))
```
pracucci: In case we keep this `else` break for a good reason, then I would suggest to add a `break` to avoid any infinite loop in case of any bug (there shouldn't be a bug, but better safe than sorry)

pkg/chunk/cache/fifo_cache.go
```diff
@@ -179,91 +200,118 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
-		entry.key = key
-		c.entries[index] = entry
-		c.memoryBytes.Add(float64(deltaSize))
-		return
+		if evicted := c.deleteFromTail(); evicted != nil {
+			delete(c.entries, evicted.key)
+			c.currSizeBytes -= sizeOf(evicted)
+			c.entriesCurrent.Dec()
+		} else {
+			c.entries = nil
```
pracucci: In case we keep this `else` break for a good reason, then if you set it to `nil` it will panic
```suggestion
			c.entries = make(map[string]*cacheEntry)
```

pkg/chunk/cache/fifo_cache.go
```diff
@@ -162,12 +173,22 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
 func (c *FifoCache) Stop() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// update metrics
+	c.entriesEvicted.Add(float64(len(c.entries)))
+	c.entriesCurrent.Set(float64(0))
+	c.memoryBytes.Set(float64(0))
+
+	c.entries = nil
```
pracucci: To cover the case someone accidentally calls the cache functions after `Stop()`
```suggestion
	c.entries = make(map[string]*cacheEntry)
```

pkg/chunk/cache/fifo_cache.go
```diff
@@ -179,91 +200,118 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
-		delete(c.index, entry.key)
-		c.index[key] = index
-
-		entry.updated = time.Now()
-		entry.value = value
-		entry.key = key
-		c.entries[index] = entry
-		c.memoryBytes.Add(float64(deltaSize))
-		return
+		if evicted := c.deleteFromTail(); evicted != nil {
```
pracucci: Why don't we simply do the following? I think the `else` branch is superfluous

```
		if evicted := c.deleteFromTail(); evicted != nil {
		  break
		}
```

pkg/chunk/cache/fifo_cache.go
```diff
@@ -179,91 +200,118 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
+			c.entriesEvicted.Inc()
+		}
+		c.memoryBytes.Set(float64(c.currSizeBytes))
 		return
 	}
-	c.entriesAddedNew.Inc()
+
+	if !ok {
+		c.entriesAddedNew.Inc()
+	}
```
pracucci: I would move this at the end of the function for clarity, once we add the entry to the cache

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,128 +2,211 @@ package cache
...
+		cfg  FifoCacheConfig
+	}{
+		{
+			name: "test-memory-eviction",
+			cfg:  FifoCacheConfig{MaxSizeBytes: cnt * sizeOf(itemTemplate), Validity: 1 * time.Minute},
+		},
+		/*{
+			name: "test-items-eviction",
+			cfg:  FifoCacheConfig{MaxSizeItems: cnt, Validity: 1 * time.Minute},
+		},*/
```
pracucci: Uncomment this, please

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -2,128 +2,211 @@ package cache
...
+		assert.Equal(t, testutil.ToFloat64(c.totalGets), float64(cnt))
+		assert.Equal(t, testutil.ToFloat64(c.totalMisses), float64(0))
+		assert.Equal(t, testutil.ToFloat64(c.staleGets), float64(0))
+		assert.Equal(t, testutil.ToFloat64(c.memoryBytes), float64(cnt*sizeOf(itemTemplate)))
+
+		for i := 0; i < cnt-evicted; i++ {
+			_, ok := c.Get(ctx, fmt.Sprintf("%02d", i))
+			require.False(t, ok)
+		}
+		for i := cnt; i < cnt+evicted; i++ {
```
pracucci:
```suggestion
		for i := cnt-evicted; i < cnt+evicted; i++ {
```

pracucci:
We're cutting Cortex 1.0 RC, so this PR will not make to merge it in time. This means we need to keep backward compatibility in the config. Could you adopt the same approach we adopted in https://github.com/cortexproject/cortex/blob/master/pkg/chunk/schema_config.go#L78-L93 to deprecated the old flag/config while still supporting it?

pkg/chunk/cache/fifo_cache.go
```diff
@@ -166,12 +185,23 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
+	c.memoryBytes.Set(float64(0))
 }
 
 // Put stores the value against the key.
 func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}) {
 	c.entriesAdded.Inc()
-	if c.size == 0 {
+	if c.maxSizeBytes == 0 && c.maxSizeItems == 0 {
 		return
 	}
```
pstibrany, line 206: Normal situation is to have one of these non-zero, so check will always fail. I suggest we remove it.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -166,12 +185,23 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
 func (c *FifoCache) Stop() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	c.entries = make(map[string]*cacheEntry)
+	c.first = nil
+	c.last = nil
+	c.currSizeBytes = 0
+
+	c.entriesEvicted.Add(float64(len(c.entries)))
```
pstibrany, line 196: `c.entries` is always empty at this point

pkg/chunk/cache/fifo_cache.go
```diff
@@ -183,91 +213,116 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
+	} else {
+		entry.next.prev = entry.prev
+	}
+}
 
-		// Corner case: updating last element
-		if c.last == index {
-			c.last = entry.prev
-		}
+func (c *FifoCache) put(ctx context.Context, key string, value interface{}) {
```
pstibrany, line 260: we can remove unused `ctx` parameter from here, it's not part of interface.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -166,12 +185,23 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
+	c.last = nil
+	c.currSizeBytes = 0
+
+	c.entriesEvicted.Add(float64(len(c.entries)))
+	c.entriesCurrent.Set(float64(0))
+	c.memoryBytes.Set(float64(0))
 }
 
 // Put stores the value against the key.
 func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}) {
```
pstibrany, line 202: Given that this cache is only used with byte slices, why don't we modify `Put` method to accept byte slices only, and simplify the size calculation code? It's a lot of code that's not used.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -183,91 +213,116 @@ func (c *FifoCache) Put(ctx context.Context, keys []string, values []interface{}
...
-func (c *FifoCache) put(ctx context.Context, key string, value interface{}) {
-	// See if we already have the entry
-	index, ok := c.index[key]
-	if ok {
-		entry := c.entries[index]
-		deltaSize := sizeOf(value) - sizeOf(entry.value)
+func (c *FifoCache) addToHead(entry *cacheEntry) {
+	if entry == nil {
+		return
+	}
```
pstibrany, line 219: Remove this check please. If entry is `nil`, let it crash so that we fix it quickly. It should never be called with `nil` pointer.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -71,29 +73,41 @@ var (
...
+	maxSizeItems  int
+	maxSizeBytes  int
+	currSizeBytes int
+	validity      time.Duration
+	entries       map[string]*cacheEntry
 
-	// indexes into entries to identify the most recent and least recent entry.
-	first, last int
+	// pointers to the first and last elements of FIFO doubly linked list.
+	first, last *cacheEntry
```
pstibrany, line 110: Can we use https://golang.org/pkg/container/list/?

pkg/chunk/cache/fifo_cache.go
```diff
@@ -81,13 +83,17 @@ type FifoCacheConfig struct {
...
+	DeprecatedSize int `yaml:"size"`
 }
 
 // RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
 func (cfg *FifoCacheConfig) RegisterFlagsWithPrefix(prefix, description string, f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxSizeBytes, prefix+"fifocache.max-size-bytes", 0, description+"Maximum memory size of the cache.")
 	f.IntVar(&cfg.MaxSizeItems, prefix+"fifocache.max-size-items", 0, description+"Maximum number of entries in the cache.")
 	f.DurationVar(&cfg.Validity, prefix+"fifocache.duration", 0, description+"The expiry duration for the cache.")
+
+	f.IntVar(&cfg.DeprecatedSize, prefix+"fifocache.size", 0, "DEPRECATED(use "+prefix+"fifocache.max-size-{items|bytes}) "+description+"The number of entries to cache.")
```
pracucci: Unfortunately adding the prefix to the help doesn't work with the config generator. You can simplify it as:

```suggestion
	f.IntVar(&cfg.DeprecatedSize, prefix+"fifocache.size", 0, "Deprecated (use max-size-items or max-size-bytes instead): "+description+"The number of entries to cache. ")
```

pkg/chunk/cache/fifo_cache.go
```diff
@@ -125,6 +131,15 @@ type cacheEntry struct {
 func NewFifoCache(name string, cfg FifoCacheConfig) *FifoCache {
 	util.WarnExperimentalUse("In-memory (FIFO) cache")
 
+	if cfg.DeprecatedSize > 0 {
+		flagext.DeprecatedFlagsUsed.Inc()
+		level.Warn(util.Logger).Log("msg", "running with DEPRECATED flag fifocache.size, use fifocache.max-size-{items|bytes} instead", "cache", name)
```
pracucci, line 136:
```suggestion
		level.Warn(util.Logger).Log("msg", "running with DEPRECATED flag fifocache.size, use fifocache.max-size-items or fifocache.max-size-bytes instead", "cache", name)
```

pkg/chunk/cache/fifo_cache.go
pracucci: Why should we remove it? The check here is used to avoid storing items in the cache if the cache is disabled (which is when the max items/bytes are set to 0).

pstibrany: Fifo cache can be disabled via `enable_fifocache` flag at pkg/chunk/cache/cache.go:66. I would suggest that `NewFifoCache` returns nil, if both max size and items are zero, and then cache is not created. Handling this specially in `Put` function is design smell... in my opinion, created cache should never have both parameters set to zero.

pracucci: Design wise I do agree with you. I'm just wondering if what you propose may be considered a breaking change (which we can't afford at this stage).

pstibrany: You mean breaking change in terms of API? Are we going to adhere to API versioning rules now?

I believe we consider all Cortex packages as internal packages (with the exception of integration tests framework). (Btw, I checked that Loki doesn't use this)

Functionality-wise, it's not a breaking change, although it may affect some metrics.

pracucci: I did mean breaking change in terms of config, but you're right: there's no config breaking change in your proposal:

> I would suggest that NewFifoCache returns nil, if both max size and items are zero

Your proposal LGTM 👍

pkg/chunk/cache/fifo_cache.go
dmitsh: Let's address that in a separate PR

pkg/chunk/cache/fifo_cache.go
dmitsh: Given the history of this PR, I would rather do that in a separate PR

pkg/chunk/cache/fifo_cache.go
dmitsh: Actually, I went ahead and changed it in this PR. Thank you for suggesting. The code became more clear.

pstibrany:
Thank you for your work on this PR. There is one place where NewFifoCache is called, which now needs to check if returned value is ‘nil’: https://github.com/cortexproject/cortex/blob/a8003f05065eaf8b88927defa2c317d412e5937e/pkg/chunk/cache/cache.go#L71

dmitsh:
@pstibrany Thank you for catching this. I've updated the PR.

pkg/chunk/cache/fifo_cache.go
pstibrany: Thanks. This was a different suggestion :) I'd like to see it done, but not necessarily in this PR.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -71,29 +74,40 @@ var (
 	}, []string{"cache"})
 )
 
+// This FIFO cache implementation supports two eviction methods - based on number of items in the cache, and based on memory usage.
+// For the memory-based eviction, set FifoCacheConfig.MaxSizeBytes to a positive integer, indicating upper limit of memory allocated by items in the cache.
+// Alternatively, set FifoCacheConfig.MaxSizeItems to a positive integer, indicating maximum number of items in the cache.
+// If both parameters are set, the memory-based eviction method takes precedence.
```
pstibrany, line 80: If both parameters are set, cannot we enforce both limits, whichever is hit first? Why are they mutually exclusive?

pkg/chunk/cache/fifo_cache.go
```diff
@@ -166,108 +190,93 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
+	entrySz := sizeOf(entry)
 
-		// Corner case: updating last element
-		if c.last == index {
-			c.last = entry.prev
+	if c.maxSizeBytes > 0 && entrySz > c.maxSizeBytes {
+		// Cannot keep this item in the cache.
+		if ok {
+			// We do not replace this item.
+			c.entriesEvicted.Inc()
```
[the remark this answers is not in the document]
pstibrany: `c.entriesCurrent` was already decreased if `ok` is set.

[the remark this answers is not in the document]
pracucci: Oh right, sorry!

pkg/chunk/cache/fifo_cache.go
```diff
@@ -166,108 +190,93 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
 	if ok {
-		entry := c.entries[index]
-		deltaSize := sizeOf(value) - sizeOf(entry.value)
-
-		entry.updated = time.Now()
-		entry.value = value
+		// Remove the item from the cache.
+		entry := c.lru.Remove(element).(*cacheEntry)
+		delete(c.entries, key)
+		c.currSizeBytes -= sizeOf(entry)
```
pstibrany, line 225: I think we need to cache entry size from moment when entry is inserted to cache. Entry size may have changed since then, and then this computation will be wrong.

pkg/chunk/cache/fifo_cache.go
dmitsh: Indeed, we can support both.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -166,108 +190,93 @@ func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
...
+		c.memoryBytes.Set(float64(c.currSizeBytes))
 		return
 	}
-	c.entriesAddedNew.Inc()
 
-	// Otherwise, see if we need to evict an entry.
-	if len(c.entries) >= c.size {
+	// Otherwise, see if we need to evict item(s).
+	for (c.maxSizeBytes > 0 && c.currSizeBytes+entrySz > c.maxSizeBytes) || (c.maxSizeItems > 0 && len(c.entries) >= c.maxSizeItems) {
+		evicted := c.lru.Remove(c.lru.Back()).(*cacheEntry)
```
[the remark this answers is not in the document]
dmitsh: Thanks, good catch. I was under impression that nil can be casted to any pointer. I stand corrected :)

pkg/chunk/cache/fifo_cache.go
dmitsh: Not sure I understand. We have three types of items: the old cache item for the existing key, the evicted item(s), and the new cache item for the existing key. All three types are tracked correctly IMO. The `c.currSizeBytes` is updated the moment the cache content is changed.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -281,6 +287,11 @@ func (c *FifoCache) Get(ctx context.Context, key string) (interface{}, bool) {
 
 func sizeOf(i interface{}) int {
```
pracucci, line 288: I think that for slices we should measure the size with `cap()` and not `len()`

pkg/chunk/cache/fifo_cache.go
```diff
@@ -149,125 +170,106 @@ func (c *FifoCache) Fetch(ctx context.Context, keys []string) (found []string, b
...
 }
 
 // Store implements Cache.
 func (c *FifoCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
-	values := make([]interface{}, 0, len(bufs))
-	for _, buf := range bufs {
-		values = append(values, buf)
-	}
-	c.Put(ctx, keys, values)
+	c.Put(ctx, keys, bufs)
```
pstibrany, line 180: Having both `Store` and `Put` methods seems redundant. We need `Store`, since it's part of `cache.Cache` interface, but maybe we can remove `Put`?

dmitsh: agree

pkg/chunk/cache/fifo_cache.go
```diff
@@ -279,38 +281,9 @@ func (c *FifoCache) Get(ctx context.Context, key string) (interface{}, bool) {
...
-		}
-		return 0
-	default:
-		return int(unsafe.Sizeof(i))
-	}
+func sizeOf(item *cacheEntry) int {
+	return int(unsafe.Sizeof(*item)) + // size of cacheEntry
+		cap(item.value) + // size of value
+		(2 * len(item.key)) + // counting key twice: in the cacheEntry and in the map
+		int(unsafe.Sizeof(&list.Element{})) // size of the pointer to an element in the map
```
pstibrany, line 283: `sizeOf` is little more tricky that I'd like.

For example, I don't really think string length should be counted twice, because it's only string descriptor that's copied, but not the actual string data.

Also `cap(item.value)` is not the full story, as subslice of large slice will report smaller `cap`, but still keep entire large slice referenced.

`int(unsafe.Sizeof(&list.Element{}))` – why counting just a pointer and not full Element struct?

Personally I would simplify it to `len(item.key) + cap(item.value)`, since it's just a best-effort guess, ignoring maintainance structs.

dmitsh: Thanks, I updated `sizeOf`.

PR #2321, opened 2020-03-24
Spread checkpoint writing

codesome:
Due to a burst of disk writes during checkpoint, we were facing some heavy disk I/O throttling. This has a big effect on ingestion latency as we write to WAL (north of 60s in worst cases for around 500k series). Hence spread the writing of checkpoint to the entire checkpoint duration.

### DIFF
--- a/pkg/ingester/wal.go
+++ b/pkg/ingester/wal.go
@@ -161,7 +161,7 @@ func (w *walWrapper) run() {
 		case <-ticker.C:
 			start := time.Now()
 			level.Info(util.Logger).Log("msg", "starting checkpoint")
-			if err := w.performCheckpoint(); err != nil {
+			if err := w.performCheckpoint(false); err != nil {
 				level.Error(util.Logger).Log("msg", "error checkpointing series", "err", err)
 				continue
 			}
@@ -170,7 +170,7 @@ func (w *walWrapper) run() {
 			w.checkpointDuration.Observe(elapsed.Seconds())
 		case <-w.quit:
 			level.Info(util.Logger).Log("msg", "creating checkpoint before shutdown")
-			if err := w.performCheckpoint(); err != nil {
+			if err := w.performCheckpoint(true); err != nil {
 				level.Error(util.Logger).Log("msg", "error checkpointing series during shutdown", "err", err)
 			}
 			return
@@ -180,7 +180,7 @@ func (w *walWrapper) run() {
 
 const checkpointPrefix = "checkpoint."
 
-func (w *walWrapper) performCheckpoint() (err error) {
+func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
 	if !w.cfg.CheckpointEnabled {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/ingester/wal.go
```diff
@@ -241,6 +241,24 @@ func (w *walWrapper) performCheckpoint() (err error) {
 		os.RemoveAll(checkpointDirTemp)
 	}()
 
+	// Count number of series - we'll use this to rate limit checkpoints.
+	numSeries := 0
+	for _, state := range w.getUserStates() {
```
pracucci: `w.getUserStates()` does a copy of the map and it's also called below within the same file. You may just call it once and then reuse it.

pkg/ingester/wal.go
```diff
@@ -241,6 +241,24 @@ func (w *walWrapper) performCheckpoint() (err error) {
...
+	}
+	if numSeries == 0 {
+		return nil
+	}
+
+	perSeriesDuration := w.cfg.CheckpointDuration / time.Duration(numSeries)
+	ticker := time.NewTicker(perSeriesDuration)
+
+	if immediate {
+		ticker.Stop()
```
pracucci: What's the point of creating a `ticker` we immediately `Stop()`?

pkg/ingester/wal.go
```diff
@@ -180,7 +180,7 @@ func (w *walWrapper) run() {
 
 const checkpointPrefix = "checkpoint."
 
-func (w *walWrapper) performCheckpoint() (err error) {
+func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
```
pracucci: With this change, there's no guarantee that the next `performCheckpoint()` (called by the priodically `CheckpointDuration` ticker) will be performed once the previous one has completed. Shouldn't we protect from this case?

pkg/ingester/wal.go
codesome: Right, modified

pkg/ingester/wal.go
codesome: I guess I was in a hurry :) I have modified it now to only initialize when needed

pkg/ingester/wal.go
codesome: Good catch! I thought it was protected by a mutex, but apparently now. I have added one now.

pkg/ingester/wal.go
```diff
@@ -241,15 +245,40 @@ func (w *walWrapper) performCheckpoint() (err error) {
...
+	numSeries := 0
+	us := w.getUserStates()
+	for _, state := range us {
+		numSeries += state.fpToSeries.length()
+	}
+	if numSeries == 0 {
+		return nil
+	}
+
+	perSeriesDuration := w.cfg.CheckpointDuration / time.Duration(numSeries)
```
pracucci: This can be moved inside the `if !immediate` block

pkg/ingester/wal.go
```diff
@@ -180,11 +181,14 @@ func (w *walWrapper) run() {
 
 const checkpointPrefix = "checkpoint."
 
-func (w *walWrapper) performCheckpoint() (err error) {
+func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
 	if !w.cfg.CheckpointEnabled {
 		return nil
 	}
 
+	w.checkpointMtx.Lock()
```
pracucci: Please add a comment to explain why we do it. May not be much clear why we lock the entire function.

PR #2322, opened 2020-03-24
[WIP] Batch up the WAL records before writing

codesome:
As the WAL logging uses a mutex, due to disk throttling or in general latency from disk, writes get stuck at the mutex waiting for other writes to finish.

Updated Design: Batch up WAL records and write in batches to reduce disk IO, and an ingester returns only when the batch is written.

Old Design:

In this PR, we hold the records in a channel and log it to the WAL in parallel by not locking the ingestion path.

One con of the proposed solution in this PR is that we are no longer returning the WAL log error to the ingestion call.

### DIFF
--- a/pkg/ingester/wal.go
+++ b/pkg/ingester/wal.go
@@ -66,6 +66,7 @@ type walWrapper struct {
 
 	wal           *wal.WAL
 	getUserStates func() map[string]*userState
+	checkpointMtx sync.Mutex
 
 	// Checkpoint metrics.
 	checkpointDeleteFail    prometheus.Counter
@@ -161,7 +162,7 @@ func (w *walWrapper) run() {
 		case <-ticker.C:
 			start := time.Now()
 			level.Info(util.Logger).Log("msg", "starting checkpoint")
-			if err := w.performCheckpoint(); err != nil {
+			if err := w.performCheckpoint(false); err != nil {
 				level.Error(util.Logger).Log("msg", "error checkpointing series", "err", err)
 				continue
 			}
@@ -170,7 +171,7 @@ func (w *walWrapper) run() {
 			w.checkpointDuration.Observe(elapsed.Seconds())
 		case <-w.quit:
 			level.Info(util.Logger).Log("msg", "creating checkpoint before shutdown")
-			if err := w.performCheckpoint(); err != nil {
+			if err := w.performCheckpoint(true); err != nil {
 				level.Error(util.Logger).Log("msg", "error checkpointing series during shutdown", "err", err)
 			}
 			return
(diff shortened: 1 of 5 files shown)


gouthamve:
This is strictly not a WAL behaviour. We have a chance to lose data here. But we can't scale without this. The reason being that in Prometheus we do a `Log()` call per scrape (in our env that is 2000 samples on average) while here we're doing it for less than 10 samples. Our batch size at the distributor is 100, and that is split into smaller number of samples per ingester.

This means we're trying to call `Log()` more than 2000 per second, and because calls to Log() are serialised inside there is a lot of contention. Not sure how to proceed here tbh.

codesome:
> This is strictly not a WAL behaviour. We have a chance to lose data here.

Yes :/

Can we have `X` number of WALs where a `series` goes to `series.Fingerprint % X`th WAL?

gouthamve:
That would be even more syscalls, I'm not sure.

codesome:
More syscalls for creating files. But if we look at the pages flushes to the files, it would be same because the data size remains the same. Though I personally would not like to have a lot of WAL floating around, I see this as an option (like a stripped-down ingester with TSDB blocks but without blocks :))

gouthamve:
I was meaning `Write()` syscalls :)

codesome:
Ah, we flush the page not only when it's full, but also when it's the final record! That calls for another optimization here which again is not strictly WAL, i.e. batching up records and writing them to reduce the syscalls.

codesome:
Syscalls would remain the same btw with WAL sharding as number of syscalls = number of WAL Log = remains same after sharding

codesome:
Forgot that a record will have samples from more than 1 series :) So yeah, more syscalls [offline discussion]

pkg/ingester/wal.go
```diff
@@ -91,12 +96,22 @@ func newWAL(cfg WALConfig, userStatesFunc func() map[string]*userState, register
...
-		cfg:           cfg,
-		quit:          make(chan struct{}),
-		wal:           tsdbWAL,
-		getUserStates: userStatesFunc,
+		cfg:                   cfg,
+		quit:                  make(chan struct{}),
+		wal:                   tsdbWAL,
+		getUserStates:         userStatesFunc,
+		recordBuffer:          make([][]byte, 100),
+		recordBufferErrorChan: make([]chan error, 100),
```
pstibrany: Why 100? It should be a constant too and reused.

codesome: Still experimenting :) nothing is fixed here, not even the implementation. Latencies are still high.

codesome:
> Can we have a WAL per tenant? Would that help?

Might help when there are many tenants, but not useful when there is only 1 tenant. So we need some other solution.

codesome:
No longer needed as the latency that we saw was because of CPU throttling.

PR #2326, opened 2020-03-24
Store gateway service scaffolding

pracucci:
**What this PR does**:
As a first step to build the blocks sharding for the read path (#2315), in this PR I'm introducing the scaffolding for the `store-gateway` service.

Notes:
- In order to keep PRs easier to review and quicker to iterate on, I propose to build the sharding in incremental PRs
- The `store-gateway` runs within the `all` target (only when the blocks storage is used) because it will be essential for the querier

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/ring.go
+++ b/pkg/ring/ring.go
@@ -34,6 +34,9 @@ const (
 
 	// CompactorRingKey is the key under which we store the compactors ring in the KVStore.
 	CompactorRingKey = "compactor"
+
+	// StoreGatewayRingKey is the key under which we store the store gateways ring in the KVStore.
+	StoreGatewayRingKey = "store-gateway"
 )
 
 // ReadRing represents the read interface to the ring.
--- a/pkg/storegateway/gateway_http.go
+++ b/pkg/storegateway/gateway_http.go
@@ -1,0 +1,53 @@
+package storegateway
+
+import (
+	"net/http"
+	"text/template"
+
+	"github.com/go-kit/kit/log/level"
+
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/services"
+)
+
+var (
+	statusPageTemplate = template.Must(template.New("main").Parse(`
+	<!DOCTYPE html>
+	<html>
+		<head>
+			<meta charset="UTF-8">
+			<title>Cortex Store Gateway Ring</title>
+		</head>
+		<body>
+			<h1>Cortex Store Gateway Ring</h1>
+			<p>{{ .Message }}</p>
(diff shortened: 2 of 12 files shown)


pkg/ring/ring.go
```diff
@@ -34,6 +34,9 @@ const (
 
 	// CompactorRingKey is the key under which we store the compactors ring in the KVStore.
 	CompactorRingKey = "compactor"
+
+	// StoreGatewayRingKey is the key under which we store the store gateways ring in the KVStore.
+	StoreGatewayRingKey = "store-gateway"
```
pstibrany: None of these keys belong here, but that's a separate PR.

pkg/storegateway/gateway_http.go
```diff
@@ -0,0 +1,53 @@
...
+	err := statusPageTemplate.Execute(w, struct {
+		Message string
+	}{Message: message})
+
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "unable to serve store gateway ring page", "err", err)
+	}
+}
+
+func (c *StoreGateway) RingHandler(w http.ResponseWriter, req *http.Request) {
```
pstibrany: We have multiple copies of this code around Cortex now, we should unify it into single "ring handler" (again, in separate PR)

pkg/storegateway/gateway_ring.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/cortexproject/cortex/pkg/ring/kv"
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
+)
+
+// RingConfig masks the ring lifecycler config which contains
+// many options not really required by the store gateways ring. This config
+// is used to strip down the config to the minimum, and avoid confusion
+// to the user.
+type RingConfig struct {
```
pstibrany: Would it be possible to reduce repetition with these RingConfig structures too at some point? We have 4 of them now 😞

pkg/ring/ring.go
pracucci: Agree. Will take care in a separate PR.

pkg/storegateway/gateway_http.go
pracucci: Agree. Will take care in a separate PR.

pkg/storegateway/gateway_ring.go
pracucci: Yes. There're some slightly differences (ie. replication factor and tokens) but we may workaround it in a refactoring to avoid duplication.

PR #2327, opened 2020-03-24
API Design Proposal

jtlisi:
**What this PR does**:
This PR adds an initial draft of a proposal to restructure and version the Cortex API.

### DIFF
--- a/docs/proposals/api_design.md
+++ b/docs/proposals/api_design.md
@@ -5,7 +5,10 @@ weight: 1
 slug: http-api-design
 ---
 
-# HTTP API Design
+- Author: @jtlisi
+- Reviewers: @pracucci, @pstibrany, @khaines, @gouthamve
+- Date: March 2020
+- Status: Accepted
 
 ## Overview
 
--- a/docs/proposals/api_design.md
+++ b/docs/proposals/api_design.md
@@ -1,0 +1,47 @@
+---
+title: "HTTP API Design"
+linkTitle: "HTTP API Design"
+weight: 1
+slug: http-api-design
+---
+
+# HTTP API Design
+
+## Overview
+
+The purpose of this design document is to propose a set of standards that should be the basis of the Cortex HTTP API. This document will outline the current state of the Cortex http api and describe limitations that result from the current approach. It will also outline a set of paradigms on how http routes should be created within Cortex.
+
+## Current Design
+
+As things currently stand, the majority of HTTP API calls exist under the `/api/prom` path prefix. This prefix is configurable. However, since this prefix is shared between all the modules which leads to conflicts if the Alertmanager is attempted to be run as as part of the single binary.
+
+## Proposed Design
+
+### Module-Based Routing
+
+Cortex incorporates three separate APIs: Alertmanager, Prometheus, and Cortex. Each of these APIs should use a separate route prefix that accurately describes the API. Currently, all of the api calls in Cortex reside under the configured http prefix. Instead the following routing tree is proposed:
+
+- `/prometheus/*` -- Under this path prefix, Cortex will act as a Prometheus web server. It will host all of the required Prometheus api endpoints. For example to query cortex the endpoint `/prometheus/api/v1/query_range` will be used.
+- `/alertmanager/*` -- Under this path prefix, Cortex will act as a Alertmanager web server. In this case, it will forward requests to the alertmanager and support the alertmanager API
+- `/api/v1/*` -- The cortex API will exist under this path prefix.
+  - `/push` & `/read` -- The Prometheus remote write endpoints will be fulfilled directly under the `/api/v1/*` path
+  - `/user_stats` -- The per user stats endpoint will exist at this endpoint
+  - `/chunks` 
+  - `/rules/*`
+- `/` -- Endpoints directly exposed under the root `/` path will be maintained as is.
+  - `/config`
+  - `/ring`
+  - `/ruler_ring`
+  - `/compactor_ring`
+  - `/ha-tracker`
+  - `/all_user_stats`
+  - `/flush`
+  - `/shutdown`
+
+### Path Versioning
+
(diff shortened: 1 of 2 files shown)


docs/proposals/api_design.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Overview
+
+The purpose of this design document is to propose a set of standards that should be the basis of the Cortex HTTP API. This document will outline the current state of the Cortex http api and describe limitations that result from the current approach. It will also outline a set of paradigms on how http routes should be created within Cortex.
+
+## Current Design
+
+As things currently stand, the majority of HTTP API calls exist under the `/api/prom` path prefix. This prefix is configurable. However, since this prefix is shared between all the modules which leads to conflicts if the Alertmanager is attempted to be run as as part of the single binary.
+
+## Proposed Design
```
pracucci: Does your proposed design keeps the `-http.prefix` support? I think there's some value in being able to configure a prefix which applies to all routes, but the default should be empty string.

docs/proposals/api_design.md
```diff
@@ -0,0 +1,62 @@
...
+Under this path prefix, Cortex will act as a Alertmanager web server. In this case, it will forward requests to the alertmanager and support the alertmanager API
+
+#### `/api/v1/*` -- The cortex API will exist under this path prefix.
+
+- `/push` & `/read` -- The Prometheus remote write endpoints will be fulfilled directly under the `/api/v1/*` path
+- `/user_stats` -- The per user stats endpoint will exist at this endpoint
+- `/chunks` 
+- `/rules/*`
+
+#### `/` -- Endpoints directly exposed under the root path. This is identical to current behaviour.
```
pracucci: For service-specific paths, what if we add a `/<service>/` prefix? Ie. `/compactor/ring`, `/ingester/flush`, etc.

docs/proposals/api_design.md
pracucci: In the Cortex API endpoints there's also a mixed usage of `_` and `-`. I would suggest to uniform it (`-` looks more natural IMO).

docs/proposals/api_design.md
```diff
@@ -0,0 +1,62 @@
...
+- `/ruler_ring`
+- `/compactor_ring`
+- `/ha-tracker`
+- `/all_user_stats`
+- `/flush`
+- `/shutdown`
+
+### Path Versioning
+
+Cortex will utilize path based versioning similar to both Prometheus and Alertmanager. This will allow future versions of the API to be released with changes over time.
```
pracucci: Will `v1` be introduced at this stage or later? If at this stage, would be great listing the new versioned endpoints.

docs/proposals/api_design.md
jtlisi: I agree! I'll update the design document to reflect this.

docs/proposals/api_design.md
jtlisi: `v1` will be introduced at this stage. I'll add a list of what it will look like.

docs/proposals/api_design.md
jtlisi: I'm not sure what you mean? I'm not aware on any endpoints with `-` in the path? Also since Prometheus uses `_` shouldn't we follow suit?

jtlisi: I agree. However, I would like to keep that flag as is and add a deprecated warning to it's usage and still route to the endpoints as they exists under that prefix. I think a new set of flags can be added to prefix the new API.

pracucci: Sure. My point was just about being consistent (and here being consistent with Prometheus `_`  looks perfect). I left the comment without double checking and I thought there was a mixed usage.

jtlisi: 👍

docs/proposals/api_design.md
```diff
@@ -0,0 +1,74 @@
...
+
+# HTTP API Design
+
+## Overview
+
+The purpose of this design document is to propose a set of standards that should be the basis of the Cortex HTTP API. This document will outline the current state of the Cortex http api and describe limitations that result from the current approach. It will also outline a set of paradigms on how http routes should be created within Cortex.
+
+## Current Design
+
+As things currently stand, the majority of HTTP API calls exist under the `/api/prom` path prefix. This prefix is configurable. However, since this prefix is shared between all the modules which leads to conflicts if the Alertmanager is attempted to be run as as part of the single binary.
```
gouthamve: Can you link to the github issue as well?

docs/proposals/api_design.md
```diff
@@ -0,0 +1,74 @@
...
+
+Under this path prefix, Cortex will act as a Prometheus web server. It will host all of the required Prometheus api endpoints. For example to query cortex the endpoint `/prometheus/api/v1/query_range` will be used.
+
+#### `/alertmanager/*`
+
+Under this path prefix, Cortex will act as a Alertmanager web server. In this case, it will forward requests to the alertmanager and support the alertmanager API. This means for a user to access their Alertmanager UI, they will use the `/alertmanager` path of cortex.
+
+#### `/api/v1/*` -- The cortex API will exist under this path prefix.
+
+- `/push` & `/read` -- The Prometheus remote write endpoints will be fulfilled directly under the `/api/v1/*` path
```
gouthamve: `/read` is a Prometheus API endpoint.

jtlisi: @gouthamve I noticed. It's a bit strange since we currently host the read endpoint under `/api/prom/read`. It should be `/api/prom/api/v1/read`. In this case it make sense to move both of these endpoints under the Prometheus API:

`/prometheus/api/v1/read` & `/prometheus/api/v1/write` WDYT?

pracucci:
Can we consider the proposal accepted and merge it?

jtlisi:
I'm fine with merging it now.

pracucci:
@jtlisi I rebased `master` and fixed the doc page header to have a nice formatting in the website. I will merge once CI passes.

pracucci:
The failed `lint` is unrelated and will be fixed in a separate PR.

PR #2330, opened 2020-03-25
/ready now returns 200, not 204

pstibrany:
**What this PR does**: This PR changes /ready endpoint to return status code 200 instead of 204. 200 is more common, and in some environments it's difficult to change the check from 200 to 204. In Kubernetes, [any code greater than or equal to 200 and less than 400 indicates success](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request).

~This PR also fixes `TestBackwardCompatibilityWithChunksStorage` to start first ingester with old image. This same fix has now been posted as #2329, and I will rebase this PR when #2329 is merged.~ (Update: Rebased)

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 * [CHANGE] Utilize separate protos for rule state and storage. Experimental ruler API will not be functional until the rollout is complete. #2226
 * [CHANGE] Frontend worker in querier now starts after all Querier module dependencies are started. This fixes issue where frontend worker started to send queries to querier before it was ready to serve them (mostly visible when using experimental blocks storage). #2246
 * [CHANGE] Lifecycler component now enters Failed state on errors, and doesn't exit the process. (Important if you're vendoring Cortex and use Lifecycler) #2251
+* [CHANGE] /ready handler now returns 200 instead of 204.
 * [FEATURE] Added experimental storage API to the ruler service that is enabled when the `-experimental.ruler.enable-api` is set to true #2269
   * `-ruler.storage.type` flag now allows `s3`,`gcs`, and `azure` values
   * `-ruler.storage.(s3|gcs|azure)` flags exist to allow the configuration of object clients set for rule storage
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 * [CHANGE] Utilize separate protos for rule state and storage. Experimental ruler API will not be functional until the rollout is complete. #2226
 * [CHANGE] Frontend worker in querier now starts after all Querier module dependencies are started. This fixes issue where frontend worker started to send queries to querier before it was ready to serve them (mostly visible when using experimental blocks storage). #2246
 * [CHANGE] Lifecycler component now enters Failed state on errors, and doesn't exit the process. (Important if you're vendoring Cortex and use Lifecycler) #2251
+* [CHANGE] /ready handler now returns 200 instead of 204.
```
```suggestion
* [CHANGE] `/ready` handler now returns 200 instead of 204. #2330
```

integration/backward_compatibility_test.go
```diff
@@ -16,24 +16,34 @@ import (
...
+type serviceSetupFn func(name string, s *e2ecortex.CortexService)
+
 var (
 	// If you change the image tag, remember to update it in the preloading done
 	// by CircleCI too (see .circleci/config.yml).
-	previousVersionImages = []string{
-		"quay.io/cortexproject/cortex:v0.6.0",
-		"quay.io/cortexproject/cortex:v0.7.0",
+	previousVersionImages = map[string]serviceSetupFn{
+		"quay.io/cortexproject/cortex:v0.6.0": func(name string, s *e2ecortex.CortexService) {
```
pracucci: What do you think if we change the `NewHTTPReadinessProbe` instead to accept a response status code range, so that we can just configure Cortex services to use the range `200-299` as success?

pstibrany: Great idea!

PR #2332, opened 2020-03-25
enforce sharding defaults in period configs

owen-d:
**What this PR does**:
This fixes a bug where the default shard factors were not set properly on `PeriodConfigs`. It caused us to never update the period config’s shard factor (although the derived Schema, which is used in the store, has the default shard values applied — 16 starting in v10+).

I introduced this bug while addressing the following:
https://github.com/cortexproject/cortex/pull/1878#discussion_r363884225

**Discovery**

We discovered this bug when we tried to enable `querier.parallelise-shardable-queries`. It effectively turned that code path into a no-op because it didn't register any shard-able `PeriodConfigs` (we don't specify overrides and use the derived default).

**The Fix**

- Introduces a `func (cfg *PeriodConfig) applyDefaults()` function which mutates the associated `PeriodConfig` by applying default values (currently only shard factor for appropriate schemas).
- Errors in the case of an invalid sharding configuration in each `PeriodConfig.Validate()`
- Adds a number of test cases to ensure these guarantees

This should ensure that once validated , configs are safe for further use. 

/cc @gouthamve @cyriltovena 

Signed-off-by: Owen Diehl <ow.diehl@gmail.com>

### DIFF
--- a/pkg/chunk/schema_config.go
+++ b/pkg/chunk/schema_config.go
@@ -107,7 +107,9 @@ func (cfg *SchemaConfig) loadFromFile() error {
 // Validate the schema config and returns an error if the validation
 // doesn't pass
 func (cfg *SchemaConfig) Validate() error {
-	for _, periodCfg := range cfg.Configs {
+	for i, _ := range cfg.Configs {
+		periodCfg := &cfg.Configs[i]
+		periodCfg.applyDefaults()
 		if err := periodCfg.validate(); err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


pkg/chunk/schema_config.go
```diff
@@ -210,6 +214,17 @@ func (cfg PeriodConfig) validate() error {
 		return errInvalidTablePeriod
 	}
 
+	switch cfg.Schema {
+	case "v1", "v2", "v3", "v4", "v5", "v6", "v9":
```
pracucci: Even if I guess it doesn't cause any harm, to avoid any misunderstanding/misconfiguration from the user side, shouldn't we ensure the `cfg.RowShards` for these schemas is `0`? Am I missing anything?

PR #2338, opened 2020-03-26
Update metrics if transfer fails

### LINKED ISSUE #2337: UserState metrics are not updated if a transfer fails
If a transfer fails the userState struct is discarded. None of the metrics it reports are updated to reflect this. Most notably the following:

```
cortex_ingester_memory_series_removed_total
cortex_ingester_memory_series
cortex_ingester_memory_users
```

jtlisi:
**What this PR does**:

This PR ensures the UserState metrics are updated if a transfer fails. A `teardown` function is added to the `userState` struct and it is called if a transfer fails.

Fixes #2337 

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ingester/transfer.go
+++ b/pkg/ingester/transfer.go
@@ -30,7 +30,20 @@ var (
 )
 
 // returns source ingesterID, number of received series, added chunks and error
-func (i *Ingester) fillUserStatesFromStream(userStates *userStates, stream client.Ingester_TransferChunksServer) (fromIngesterID string, seriesReceived int, chunksAdded float64, err error) {
+func (i *Ingester) fillUserStatesFromStream(userStates *userStates, stream client.Ingester_TransferChunksServer) (fromIngesterID string, seriesReceived int, err error) {
+	chunksAdded := 0.0
+
+	defer func() {
+		if err != nil {
+			// Ensure the in memory chunks are updated to reflect the number of dropped chunks from the transfer
+			i.metrics.memoryChunks.Sub(chunksAdded)
+
+			// If an error occurs during the transfer and the user state is to be discarded,
+			// ensure the metrics it exports reflect this.
+			userStates.teardown()
+		}
+	}()
+
 	for {
 		var wireSeries *client.TimeSeriesChunk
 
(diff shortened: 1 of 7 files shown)


pkg/ingester/user_state_test.go
```diff
@@ -71,3 +75,65 @@ func TestForSeriesMatchingBatching(t *testing.T) {
 		})
 	}
 }
+
+// TestTeardDown ensures metrics are updated correctly if the userState is discarded
+func TestTeardDown(t *testing.T) {
```
gouthamve: `TestTeardown`

pkg/ingester/transfer.go
```diff
@@ -36,62 +36,74 @@ func (i *Ingester) TransferChunks(stream client.Ingester_TransferChunksServer) e
...
-			if err != nil {
-				return errors.Wrapf(err, "TransferChunks: getOrCreateSeries: user %s series %s", wireSeries.UserId, wireSeries.Labels)
+				err = series.setChunks(descs)
+				state.fpLocker.Unlock(fp) // acquired in getOrCreateSeries
+				if err != nil {
+					return errors.Wrapf(err, "TransferChunks: setChunks: user %s series %s", wireSeries.UserId, wireSeries.Labels)
+				}
+
+				seriesReceived++
+				i.metrics.memoryChunks.Add(float64(len(series.chunkDescs) - prevNumChunks))
```
gouthamve: This metric needs to be reset too.

pkg/ingester/user_state_test.go
jtlisi: 👍

pkg/ingester/transfer.go
jtlisi: I don't think so, since the chunks were technically created. We just removed them. This is the same reason we don't reset `cortex_ingester_memory_series_created_total`. Instead, we just increment `cortex_ingester_memory_series_removed_total`.

codesome: `memoryChunks` is a Gauge that tells active chunks in memory and there is no `_removed_` metric for chunks. So I guess it needs to be reset.

pracucci: I'm also convinced `memoryChunks` should be decremented.

jtlisi: I'll update it to decrement if the transfer fails

pkg/ingester/transfer.go
```diff
@@ -35,63 +35,79 @@ func (i *Ingester) TransferChunks(stream client.Ingester_TransferChunksServer) e
 	seriesReceived := 0
 	xfer := func() error {
 		userStates := newUserStates(i.limiter, i.cfg, i.metrics)
+		var chunksAdded float64
+		// run as an anonymous function so only a single error needs handling if returned
+		err := func() error {
+			for {
+				wireSeries, err := stream.Recv()
+				if err == io.EOF {
+					break
```
pstibrany: Can we use `defer` here to handle failure instead of putting everything into new anonymous function?

gouthamve: I personally find error handling in defer a little too magical. I'd rather be explicit.

pstibrany: Yes, it is. But having two nested anonymous functions is just too ugly. Can we move it to named function at least, or perhaps a method?

Issue #1425, opened 2019-05-29
Add metric for use of deprecated flags


bboreham:
As requested at https://github.com/cortexproject/cortex/pull/1361#discussion_r280377989:

> if the code that looks for usage of deprecated flags also had a metric for how many deprecated flags are used, we could have an alert for it that lets us know that we need to look at the logs and see which param has been deprecated so we get time to update our CI/CD pipeline before the change actually break stuff

PR #2341, opened 2020-03-27
Add metric for deprecated flags set.

gouthamve:
Also, I couldn't find a better way to do this.

Fixes: https://github.com/cortexproject/cortex/issues/1425

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/flagext/deprecated.go
+++ b/pkg/util/flagext/deprecated.go
@@ -4,9 +4,19 @@ import (
 	"flag"
 
 	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/cortexproject/cortex/pkg/util"
 )
+
+// DeprecatedFlagsUsed is the metric that counts deprecated flags set.
+var DeprecatedFlagsUsed = promauto.NewCounter(
+	prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "deprecated_flags_inuse_total",
+		Help:      "The number of deprecated flags currently set.",
+	})
 
 type deprecatedFlag struct {
 	name string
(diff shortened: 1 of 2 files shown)


gouthamve:
This unfortunately doesn't work with config options. Not sure how to proceed there.

tomwilkie:
> This unfortunately doesn't work with config options. Not sure how to proceed there.

I think this is a step in the right direction at least.  We can figure out config files later. But do file an issue for it!

pkg/util/flagext/deprecated.go
```diff
@@ -4,10 +4,20 @@ import (
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
+// DeprecatedFlagsUsed is the metric that counts deprecated flags set.
+var DeprecatedFlagsUsed = promauto.NewCounter(
+	prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "deprecated_flags_inuse_total",
```
pracucci: @gouthamve What's the purpose of this metric? I think I'm missing the real use case. When we switch a flag to `flagext.DeprecatedFlag`, that flag is ignored (its value is unused). I understand the policy we want to have for Cortex 1.0 but I think that using `flagext.DeprecatedFlag` for the purpose is not correct, because Cortex will run anyway and deprecated flag ignored.

gouthamve: Yep, but we'll have an alert on this and users can then check if they have any deprecated flags set. The user can then check their logs to see which flag has been deprecated.

In the past we've changed flags or features, and or removed flags, by setting the old one to deprecated. And also, this metric is exported, so we can increment in other places where `DeprecatedFlag` is not used.

PR #2342, opened 2020-03-27
TSDB soft errors: check original error

pstibrany:
This PR changes error checks to unwrap the error first. It also adds timestamp to the log message to provide more details when debugging problems.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany:
I think we need a unit test here.

PR #2346, opened 2020-03-27
Increment discarded samples for invalid metric names

gotjosh:
**What this PR does**:

When enabled, in the distributors, we consider metrics with an invalid
metric name a failure and discard the corresponding sample. When it
happens for this reason, the discarded samples metric appears to not be
incremented.

This commit fixes that by incrementing the counter.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,10 @@
 * [ENHANCEMENT] Better re-use of connections to DynamoDB and S3. #2268
 * [ENHANCEMENT] Experimental TSDB: Add support for local `filesystem` backend. #2245
 * [ENHANCEMENT] Experimental TSDB: Added memcached support for the TSDB index cache. #2290
-* [ENHANCEMENT] Allow 1w (where w denotes week) and 1y (where y denotes year) when setting table period and retention. #2252 
+* [ENHANCEMENT] Allow 1w (where w denotes week) and 1y (where y denotes year) when setting table period and retention. #2252
 * [ENHANCEMENT] Added FIFO cache metrics for current number of entries and memory usage. #2270
 * [ENHANCEMENT] Output all config fields to /config API, including those with empty value. #2209
+* [ENHANCEMENT] Add "missing_metric_name" and "metric_name_invalid" reasons to cortex_discarded_samples_total metric. #2338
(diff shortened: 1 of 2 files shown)


pkg/util/validation/validate.go
```diff
@@ -93,10 +94,12 @@ func ValidateLabels(cfg LabelValidationConfig, userID string, ls []client.LabelA
 	metricName, err := extract.MetricNameFromLabelAdapters(ls)
 	if cfg.EnforceMetricName(userID) {
 		if err != nil {
+			DiscardedSamples.WithLabelValues(invalidMetricName, userID).Inc()
```
tomwilkie: This case isn't an invalid metrics name, but rather a missing metric name.  Can you make a new reason string for this?

tomwilkie:
(Need LGTM from someone else as I put in the feedback fix)

pkg/util/validation/validate.go
gotjosh: Thanks for adding it Tom.

CHANGELOG.md
```diff
@@ -16,9 +16,10 @@
 * [ENHANCEMENT] Better re-use of connections to DynamoDB and S3. #2268
 * [ENHANCEMENT] Experimental TSDB: Add support for local `filesystem` backend. #2245
 * [ENHANCEMENT] Experimental TSDB: Added memcached support for the TSDB index cache. #2290
-* [ENHANCEMENT] Allow 1w (where w denotes week) and 1y (where y denotes year) when setting table period and retention. #2252 
+* [ENHANCEMENT] Allow 1w (where w denotes week) and 1y (where y denotes year) when setting table period and retention. #2252
 * [ENHANCEMENT] Added FIFO cache metrics for current number of entries and memory usage. #2270
 * [ENHANCEMENT] Output all config fields to /config API, including those with empty value. #2209
+* [ENHANCEMENT] Add "missing_metric_name" and "metric_name_invalid" reasons to cortex_discarded_samples_total metric. #2338
```
pracucci: @gotjosh May you submit a PR to fix the PR number. This should be `#2346` and not `#2338`.

PR #2357, opened 2020-03-30
Fix pushes when ingesters don't have a zone.

### LINKED ISSUE #2356: Completely broken after #2317
Following the deployment of #2317 into my staging environment, distributors, queriers and ruler all started erroring like this:

time="2020-03-29T22:43:33Z" level=warning msg="POST /api/prom/push (500) 2.254634ms Response: \"at least 2 live replicas required, could only find 1\\n\" ws: false; Content-Encoding: snappy; Content-Length: 16461; Content-Type: application/x-protobuf; User-Agent: Prometheus/2.12.0; X-Prometheus-Remote-Write-Version: 0.1.0; " traceID=77042d0abcee28fd

Ingester was on `master-0473dd88` and all ingesters are in the same AZ.

gouthamve:
Fixes #2356

### DIFF
--- a/integration/backward_compatibility_test.go
+++ b/integration/backward_compatibility_test.go
@@ -37,10 +37,10 @@ func TestBackwardCompatibilityWithChunksStorage(t *testing.T) {
 	}
 }
 
-func TestNewDistributorsCanPushToOldIngesters(t *testing.T) {
+func TestNewDistributorsCanPushToOldIngestersWithReplication(t *testing.T) {
 	for _, previousImage := range previousVersionImages {
 		t.Run(fmt.Sprintf("Backward compatibility upgrading from %s", previousImage), func(t *testing.T) {
-			runNewDistributorsCanPushToOldIngesters(t, previousImage)
+			runNewDistributorsCanPushToOldIngestersWithReplication(t, previousImage)
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


bboreham:
Thanks; I expected the backwards-compatibility test to catch this, but it runs with replication factor 1.
Do you think we should change it to use replication?

pkg/ring/ring.go
```diff
@@ -199,7 +199,7 @@ func (r *Ring) Get(key uint32, op Operation, buf []IngesterDesc) (ReplicationSet
 		if _, ok := distinctHosts[token.Ingester]; ok {
 			continue
 		}
-		if _, ok := distinctZones[token.Zone]; ok {
+		if _, ok := distinctZones[token.Zone]; ok && token.Zone != "" { // Ignore if the ingesters don't have a zone set.
```
bboreham: I think I would put the check on `token.Zone` first, because it is more efficient.

bboreham: put the check around assigning to `distinctHosts` too, save garbage.

integration/backward_compatibility_test.go
```diff
@@ -120,3 +128,85 @@ func runBackwardCompatibilityTestWithChunksStorage(t *testing.T, previousImage s
 		require.NoError(t, s.Stop(querier))
 	}
 }
+
+// Check for issues like https://github.com/cortexproject/cortex/issues/2356
+func runNewDistributorsCanPushToOldIngesters(t *testing.T, previousImage string) {
```
pracucci: Note to myself: the case "distributors can push to old ingesters" is actually covered by `runBackwardCompatibilityTestWithChunksStorage()`. The real difference in the setup is the replication factor.

khaines:
Indeed thanks for addressing this scenario!

PR #2362, opened 2020-03-30
tsdb: record a histogram of replay tsdb times

thorfour:

**What this PR does**:
Adds a histogram to record tsdb replay timings.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -75,6 +75,7 @@ type TSDBState struct {
 	// Head compactions metrics.
 	compactionsTriggered prometheus.Counter
 	compactionsFailed    prometheus.Counter
+	walReplayTime        prometheus.Histogram
 }
 
 // NewV2 returns a new Ingester that uses prometheus block storage instead of chunk storage
@@ -108,6 +109,11 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 			compactionsFailed: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
 				Name: "cortex_ingester_tsdb_compactions_failed_total",
 				Help: "Total number of compactions that failed.",
+			}),
+			walReplayTime: promauto.With(registerer).NewHistogram(prometheus.HistogramOpts{
+				Name:    "cortex_ingester_tsdb_wal_replay_duration_seconds",
+				Help:    "The total time it takes to open and replay a TSDB WAL.",
+				Buckets: prometheus.DefBuckets,
 			}),
 		},
 	}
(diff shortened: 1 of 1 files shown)


pkg/ingester/metrics.go
```diff
@@ -25,6 +25,7 @@ type ingesterMetrics struct {
 	queriedSamples        prometheus.Histogram
 	queriedSeries         prometheus.Histogram
 	queriedChunks         prometheus.Histogram
+	tsdbReplayTime        prometheus.Histogram
```
pstibrany: I think this should go to `TSDBState` struct in ingester_v2.go. Also technically, it's "open" time for existing DBs, so perhaps "tsdbStartupOpenTime" (?)

thorfour: Yea that's fair, I forgot we had metrics in that struct.

thorfour: Well, actually, should we have metrics in there? Or should we move all metrics into the metrics struct?

pstibrany: I think blocks-specific metrics should go there. Chunks and shared metrics between chunks and blocks go to metrics.go

thorfour: I get that delineation, but from another perspective it feels strange to have a metrics file that doesn't contain all the metrics.

pracucci: My feedback on the topic:

1. `tsdbMetrics struct` is a different thing and it's used for the Thanos mapped metrics. I believe this metric should go to `TSDBState struct` where we already have other blocks-storage related metrics tracked in Cortex (ie. `compactionsTriggered`).
2. The `tsdb` prefix in the `tsdbReplayTime` is redundant and should be removed. We could call it `walReplayDuration` which is that it is (I understand Peter's feedback that technically it's TSDB open, but practically the time there is taken by WAL replay and we track this metric only for existing TSDBs and not new ones we just open)

Then, we can discuss whether it's a good or bad things having ingester metrics defined in multiple places, but this shouldn't be address by this PR (we can discuss it offline and eventually come up with a dedicated PR to address it).

pkg/ingester/metrics.go
```diff
@@ -184,6 +185,11 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
 			Name: "cortex_oldest_unflushed_chunk_timestamp_seconds",
 			Help: "Unix timestamp of the oldest unflushed chunk in the memory",
 		}),
+		tsdbReplayTime: promauto.With(r).NewHistogram(prometheus.HistogramOpts{
+			Name:    "cortex_ingester_tsdb_replay_time_seconds",
```
pracucci: `cortex_ingester_tsdb_replay_time_seconds` > `cortex_ingester_tsdb_wal_replay_duration_seconds`

pkg/ingester/metrics.go
```diff
@@ -184,6 +185,11 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
...
+			Help:    "The total time it takes to open and replay a TSDB.",
```
pracucci: `The total time it takes to open and replay a TSDB WAL.`

pkg/ingester/metrics.go
pracucci: Other thoughts. We already have `cortex_ingester_wal_replay_duration_seconds` metric (chunks storage). Why don't we reuse the same?

thorfour: Only problem with that is that's a gauge and not a histogram.

PR #2363, opened 2020-03-31
add delete index support for dynamodb, cassandra and bigtable

sandeepsukhani:
**What this PR does**:
Implement the `Delete` methods for all the `IndexClient`s. As of now `Delete` methods are used by Purger which does the deletion of data requested by the users.

### DIFF
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -151,6 +151,18 @@ type bigtableWriteBatch struct {
 }
 
 func (b bigtableWriteBatch) Add(tableName, hashValue string, rangeValue []byte, value []byte) {
+	b.setMutation(tableName, hashValue, rangeValue, func(mutation *bigtable.Mutation, columnKey string) {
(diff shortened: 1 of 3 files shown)


pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -151,6 +151,18 @@ type bigtableWriteBatch struct {
...
+	})
+}
+
+func (b bigtableWriteBatch) Delete(tableName, hashValue string, rangeValue []byte) {
+	b.setMutation(tableName, hashValue, rangeValue, func(mutation *bigtable.Mutation, columnKey string) {
+		mutation.DeleteCellsInColumn(columnFamily, columnKey)
+	})
+}
+
+func (b bigtableWriteBatch) setMutation(tableName, hashValue string, rangeValue []byte, callback func(mutation *bigtable.Mutation, columnKey string)) {
```
pracucci: I think `addMutation()` would be a more correct name, because it adds a mutation to the batch.

Issue #2366, opened 2020-03-31
Blocks storage unable to ingest samples older than 1h after an outage


pracucci:
TSDB doesn't allow to append samples whose timestamp is older than the last block cut from the head. Given a block is cut from the head up until -50% of the max timestamp within the head and given the default block range period is 2h, this means that the blocks storage doesn't allow to append a sample whose timestamp is older than 1h compared to the most recent timestamp in the head.

Let's consider this scenario:
- Multiple Prometheus servers remote writing to the same Cortex tenant
- Some Prometheus servers stop remote writing to Cortex (for any reason, ie. networking issue) and they fall behind more than 1h
- When the Prometheus servers will be back online, Cortex will discard any sample whose timestamp is older than 1h because the max timestamp in the TSDB head is close to "now" (due to the working Prometheus servers which never stopped to write series) while the failing ones are trying to catch up writing samples older than 1h

We recently had an outage in our staging environment which triggered this condition and we should find a way to solve it.

_@bwplotka You may be interested, given I think this issue affects Thanos receive too._

pracucci:
@codesome Given we have vertical compaction in TSDB and we can have overlapping blocks, what would be the implications to allow to write samples "out of bounds" in the TSDB head?

pstibrany:
Noting other possible options (not necessarily good options):
- extend block range (would lead to higher ingester memory usage, and longer WAL replays)
- change the cutting formula

bwplotka:
> blocks storage doesn't allow to append a sample whose timestamp is older than 1h compared to the most recent timestamp in the head.

Really? Can we find code path in Prometheus which does it?

bwplotka:
Also, clock skew can cause it.

pstibrany:
> > blocks storage doesn't allow to append a sample whose timestamp is older than 1h compared to the most recent timestamp in the head.
> 
> Really? Can we find code path in Prometheus which does it?

New block is cut from the head, when head (in-memory data) covers more than 1.5x of the block range. For 2 hours block range, it means that head needs to have 3h of data to cut a block. Block "start time" is always the minimum sample time in the head, while block "end time" is aligned on block range boundary. New block always covers single "block range" period. Data stored into a block is then removed from the Head. That means that after cutting the block, head will already have at least 1h of data, but possibly even more.

[...]

PR #2367, opened 2020-03-31
Documented blocks storage limitation

pracucci:
**What this PR does**:
I've documented the known issue related to the issue #2366.

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/operations/blocks-storage.md
+++ b/docs/operations/blocks-storage.md
@@ -455,6 +455,20 @@ compactor:
 
 ## Known issues
 
+### Can't ingest samples older than 1h compared to the latest received sample of a tenant
+
+The blocks storage opens a TSDB for each tenant in each ingester receiving samples for that tenant. The received series are kept in the TSDB head and then a new block is cut from the head once the time range period covered by the series in the head reaches the block range period (defaults to 2h).
+
+When a new block is cut from head, the new block contains samples up until -50% of the configured block range period. This means that if you use the default 2h block range period, the head will contain up until 3h series (2h + 50%) and when a new block is cut from head it will contain series up until 1h ago (2h - 50%); technically it's 1h less than the max timestamp in the head.
+
+Given TSDB doesn't allow to append samples out of head bounds, the Cortex blocks storage can't ingest samples older than 1h compared to the latest received sample of a tenant.
+
+The typical case where this issue triggers is after a long outage. Let's consider this scenario:
+
+- Multiple Prometheus servers remote writing to the same Cortex tenant
+- Some Prometheus servers stop remote writing to Cortex (ie. networking issue) and they fall behind more than 1h
+- When the failing Prometheus servers will be back online, Cortex blocks storage will discard any sample whose timestamp is older than 1h because the max timestamp in the TSDB head is close to "now" (due to the working Prometheus servers which never stopped to write samples) while the failing ones are trying to catch up writing samples older than 1h
+
 ### Migrating from the chunks to the blocks storage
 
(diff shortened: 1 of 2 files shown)


docs/operations/blocks-storage.md
```diff
@@ -455,6 +455,20 @@ compactor:
 
 ## Known issues
 
+### Can't ingest samples older than 1h compared to the latest received sample of a tenant
+
+The blocks storage opens a TSDB for each tenant in each ingester receiving samples for that tenant. The received series are kept in the TSDB head and then a new block is cut from the head once the time range period covered by the series in the head reaches the block range period (defaults to 2h).
+
+When a new block is cut from head, the new block contains samples up until -50% of the configured block range period. This means that if you use the default 2h block range period, the head will contain up until 3h series (2h + 50%) and when a new block is cut from head it will contain series up until 1h ago (2h - 50%); technically it's 1h less than the max timestamp in the head.
+
+Given TSDB doesn't allow to append samples out of head bounds, the Cortex blocks storage can't ingest samples older than 1h compared to the latest received sample of a tenant.
```
pstibrany: Limit is computed for each push as Max of (minT, maxT-0.5*block range), where minT and maxT are minimum/maximum sample times in the head.

When head covers less than 0.5 of block range (<1h for 2h block range), samples cannot be older than min time in the head.

When head covers more than 0.5 of block range (>1h for 2h block range), samples cannot be older than half block range since max time in the head.

In other words, it can be less than 1h :-(

docs/operations/blocks-storage.md
```diff
@@ -455,6 +455,20 @@ compactor:
 
 ## Known issues
 
+### Can't ingest samples older than 1h compared to the latest received sample of a tenant
+
+The blocks storage opens a TSDB for each tenant in each ingester receiving samples for that tenant. The received series are kept in the TSDB head and then a new block is cut from the head once the time range period covered by the series in the head reaches the block range period (defaults to 2h).
+
+When a new block is cut from head, the new block contains samples up until -50% of the configured block range period. This means that if you use the default 2h block range period, the head will contain up until 3h series (2h + 50%) and when a new block is cut from head it will contain series up until 1h ago (2h - 50%); technically it's 1h less than the max timestamp in the head.
```
pstibrany: Different description of same idea...

New block is cut from the head, when head (in-memory data) covers more than 1.5x of the block range. For 2 hours block range, it means that head needs to have 3h of data to cut a block. Block "start time" is always the minimum sample time in the head, while block "end time" is aligned on block range boundary. New block always covers single "block range" period. Data stored into a block is then removed from the Head. That means that after cutting the block, head will already have at least 1h of data, but possibly even more.

pstibrany: > and when a new block is cut from head it will contain series up until 1h ago (2h - 50%)

When I read this few several times, I understood "it" as "block", but in fact it refers to the head.

docs/operations/blocks-storage.md
pstibrany: > In other words, it can be less than 1h :-(

This is true for completely new tenants. For normal operation, it is usually 1h.

pracucci: > When head covers less than 0.5 of block range (<1h for 2h block range), samples cannot be older than min time in the head.

This is true, but looks more an edge case compared to the general case I'm trying to describe here.

pstibrany: > This is true, but looks more an edge case compared to the general case I'm trying to describe here.

Right. I think I was mostly writing note to better understand this.

docs/operations/blocks-storage.md
pracucci: I picked your suggestion and I've done few changes. May you take a look now, please?

PR #2372, opened 2020-03-31
API Implementation

jtlisi:
**What this PR does**:
This is the implementation associated with the design document in #2327.

In this implementation the API struct mainly just handles routing for Cortex. In the future this package will centralize non-component based handlers and create a set of functions and structs that can be reused by Cortex handlers to make the API consistent.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+		false,                              // Disable admin APIs.
+		a.logger,
+		querier.DummyRulesRetriever{},
+		0, 0, 0, // Remote read samples and concurrency limit.
+		regexp.MustCompile(".*"),
+		func() (v1.RuntimeInfo, error) { return v1.RuntimeInfo{}, errors.New("not implemented") },
+		&v1.PrometheusVersion{},
+	)
+
+	promRouter := route.New().WithPrefix(a.cfg.ServerPrefix + a.cfg.PrometheusHTTPPrefix + "/api/v1")
```
pracucci, line 240: Why prefixing with `ServerPrefix`? Looks something we're just doing here.

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/metadata", promHandler, true, "GET")
+
+	a.registerRoute("/api/v1/user_stats", http.HandlerFunc(distributor.UserStatsHandler), true)
+	a.registerRoute("/api/v1/chunks", querier.ChunksHandler(queryable), true)
+
+	// Legacy Routes
+	a.registerRoute("/user_stats", http.HandlerFunc(distributor.UserStatsHandler), true)
+	a.registerRoute("/chunks", querier.ChunksHandler(queryable), true)
+
+	legacyPromRouter := route.New().WithPrefix(a.cfg.ServerPrefix + a.cfg.LegacyHTTPPrefix + "/api/v1")
```
pracucci, line 259: Why do we need `ServerPrefix`? Before we were not adding it. Also the bext block of `registerRoute()` calls wasn't done before because already covered by `api.Register()` without `ServerPrefix`. Am I missing anything?

pkg/api/handlers.go
```diff
@@ -0,0 +1,64 @@
...
+	<body>
+		<h1>Cortex</h1>
+		<p>Admin Endpoints:</p>
+		<ul>
+			<li><a href="/compactor/ring">Compactor Ring Status</a></li>
+			<li><a href="/config">Current Config</a></li>
+			<li><a href="/distributor/all_user_stats">Usage Statistics</a></li>
+			<li><a href="/distributor/ha-tracker">HA Tracking Status</a></li>
+			<li><a href="/multitenant-alertmanager/status">Alertmanager Status</a></li>
+			<li><a href="/ring">Ring Status</a></li>
```
pracucci, line 29:
```suggestion
			<li><a href="/ring">Ingester Ring Status</a></li>
```

pkg/api/handlers.go
```diff
@@ -0,0 +1,64 @@
...
+			<li><a href="/ruler/ring">Ruler Ring Status</a></li>
+			<li><a href="/services">Service Status</a></li>
+			<li><a href="/store-gateway/ring">Ruler Ring Status</a></li>
```
pracucci, line 32:
```suggestion
			<li><a href="/store-gateway/ring">Ruler Store Gateway Ring (experimental blocks storage)</a></li>
```

pkg/api/handlers.go
```diff
@@ -0,0 +1,64 @@
...
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<title>Cortex</title>
+	</head>
+	<body>
+		<h1>Cortex</h1>
+		<p>Admin Endpoints:</p>
+		<ul>
+			<li><a href="/compactor/ring">Compactor Ring Status</a></li>
```
pracucci, line 24:
```suggestion
			<li><a href="/compactor/ring">Compactor Ring Status (experimental blocks storage)</a></li>
```

pkg/api/handlers.go
```diff
@@ -0,0 +1,64 @@
...
+const indexPageContent = `
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<title>Cortex</title>
+	</head>
+	<body>
+		<h1>Cortex</h1>
+		<p>Admin Endpoints:</p>
```
pracucci, line 22: Currently endpoints are alphabetically sorted by URL. Maybe it's better to group the compactor and store-gateway endpoints at the end (since related to the experimental blocks storage).

pkg/api/api.go
jtlisi: This is a bugfix. Currently the `server.path-prefix` flag will break all requests routed to the promRouter. I have wrote up why in this issue: https://github.com/cortexproject/cortex/issues/2411.

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+// component/
+func (a *API) RegisterPurger(store *purger.DeleteStore) {
+	deleteRequestHandler := purger.NewDeleteRequestHandler(store)
+
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/admin/tsdb/delete_series", http.HandlerFunc(deleteRequestHandler.AddDeleteRequestHandler), true, "PUT", "POST")
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/admin/tsdb/delete_series", http.HandlerFunc(deleteRequestHandler.GetAllDeleteRequestsHandler), true, "GET")
+
+	// Legacy Routes
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/admin/tsdb/delete_series", http.HandlerFunc(deleteRequestHandler.AddDeleteRequestHandler), true, "PUT", "POST")
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/admin/tsdb/delete_series", http.HandlerFunc(deleteRequestHandler.GetAllDeleteRequestsHandler), true, "GET")
```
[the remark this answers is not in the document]
jtlisi: 👍 Good catch

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// RegisterAlertmanager registers endpoints associated with the alertmanager. It will only
+// serve endpoints using the legacy http-prefix if it is not run as a single binary.
+func (a *API) RegisterAlertmanager(am *alertmanager.MultitenantAlertmanager, target bool) {
+	// Ensure this route is registered before the prefixed AM route
+	a.registerRoute("/multitenant-alertmanager/status", am.GetStatusHandler(), false)
+
+	// UI components lead to a large number of routes to support, utilize a path prefix instead
+	a.server.HTTP.PathPrefix(a.cfg.AlertmanagerHTTPPrefix).Handler(a.authMiddleware.Wrap(am))
```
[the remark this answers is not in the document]
jtlisi: Makes sense, I'll add that as a convenience function.

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/label/{name}/values", promHandler, true, "GET")
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/series", promHandler, true, "GET", "POST", "DELETE")
+	a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/metadata", promHandler, true, "GET")
+
+	a.registerRoute("/api/v1/user_stats", http.HandlerFunc(distributor.UserStatsHandler), true)
+	a.registerRoute("/api/v1/chunks", querier.ChunksHandler(queryable), true)
+
+	// Legacy Routes
+	a.registerRoute("/user_stats", http.HandlerFunc(distributor.UserStatsHandler), true)
+	a.registerRoute("/chunks", querier.ChunksHandler(queryable), true)
```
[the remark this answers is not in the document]
jtlisi: Yes, I'll add a prefix.

pkg/api/api.go
jtlisi: I wrote it up in an issue here: #2411 

TLDR: the router used here handles the path separately from the weaveworks server router. If a prefix is set in the weaveworks server router, the promrouter will always return a 404 because it is not expecting the prefix. Anything directly registered with the weaveworks server router will not have this problem.

docs/configuration/config-file-reference.md
```diff
@@ -62,6 +62,15 @@ Where default_value is the value to use if the environment variable is undefined
 # CLI flag: -http.prefix
 [http_prefix: <string> | default = "/api/prom"]
 
+api:
+  # Base path for data storage.
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+	if apiEnabled {
+		a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/rules", http.HandlerFunc(r.PrometheusRules), true, "GET")
+		a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/alerts", http.HandlerFunc(r.PrometheusAlerts), true, "GET")
+
+		ruler.RegisterRulerServer(a.server.GRPC, r)
+
+		a.registerRoute("/api/v1/rules", http.HandlerFunc(r.ListRules), true, "GET")
+		a.registerRoute("/api/v1/rules/{namespace}", http.HandlerFunc(r.ListRules), true, "GET")
+		a.registerRoute("/api/v1/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
+		a.registerRoute("/api/v1/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+		a.registerRoute("/api/v1/rules/{namespace}", http.HandlerFunc(r.CreateRuleGroup), true, "POST")
+		a.registerRoute("/api/v1/rules/{namespace}/{groupName}", http.HandlerFunc(r.DeleteRuleGroup), true, "DELETE")
+
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/api/v1/rules", http.HandlerFunc(r.PrometheusRules), true, "GET")
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/api/v1/alerts", http.HandlerFunc(r.PrometheusAlerts), true, "GET")
+
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules", http.HandlerFunc(r.ListRules), true, "GET")
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules/{namespace}", http.HandlerFunc(r.ListRules), true, "GET")
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+type API struct {
+	cfg            Config
+	authMiddleware middleware.Func
+	server         *server.Server
+	logger         log.Logger
+}
+
+func New(cfg Config, s *server.Server, logger log.Logger) (*API, error) {
+	// Ensure the encoded path is used. Required for the rules API
+	s.HTTP.UseEncodedPath()
```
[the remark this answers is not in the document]
jtlisi: I took a look. There isn't any routes in cortex or in the embedded prom/alertmanager API that will be affected.

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// RegisterIngester registers the ingesters HTTP and GRPC service
+func (a *API) RegisterIngester(i *ingester.Ingester, pushConfig distributor.Config) {
+	client.RegisterIngesterServer(a.server.GRPC, i)
+	grpc_health_v1.RegisterHealthServer(a.server.GRPC, i)
+
+	a.registerRoute("/ingester/flush", http.HandlerFunc(i.FlushHandler), false)
+	a.registerRoute("/ingester/shutdown", http.HandlerFunc(i.ShutdownHandler), false)
+	a.registerRoute("/ingester/push", push.Handler(pushConfig, i.Push), true) // For testing and debugging.
```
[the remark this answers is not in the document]
jtlisi: That would work for me. I was also thinking it could live under `/ingester/debug/push`. Since it's only for debugging/testing. It might be nice to make that explicit in the path so it's clear to new users not to use it.

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+		a.registerRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/alerts", http.HandlerFunc(r.PrometheusAlerts), true, "GET")
+
+		ruler.RegisterRulerServer(a.server.GRPC, r)
+
+		a.registerRoute("/api/v1/rules", http.HandlerFunc(r.ListRules), true, "GET")
+		a.registerRoute("/api/v1/rules/{namespace}", http.HandlerFunc(r.ListRules), true, "GET")
+		a.registerRoute("/api/v1/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
+		a.registerRoute("/api/v1/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
+		a.registerRoute("/api/v1/rules/{namespace}", http.HandlerFunc(r.CreateRuleGroup), true, "POST")
+		a.registerRoute("/api/v1/rules/{namespace}/{groupName}", http.HandlerFunc(r.DeleteRuleGroup), true, "DELETE")
```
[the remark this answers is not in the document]
jtlisi: Yes, these are the update experimental ruler api routes.

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules/{namespace}/{groupName}", http.HandlerFunc(r.GetRuleGroup), true, "GET")
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules/{namespace}", http.HandlerFunc(r.CreateRuleGroup), true, "POST")
+		a.registerRoute(a.cfg.LegacyHTTPPrefix+"/rules/{namespace}/{groupName}", http.HandlerFunc(r.DeleteRuleGroup), true, "DELETE")
+	}
+}
+
+// // RegisterRing registers the ring UI page associated with the distributor for writes.
+func (a *API) RegisterRing(r *ring.Ring) {
+	a.registerRoute("/ring", r, false)
```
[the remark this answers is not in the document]
jtlisi: Makes sense

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+// RegisterAPI registers the standard endpoints associated with a running Cortex.
+func (a *API) RegisterAPI(cfg interface{}) {
+	a.registerRoute("/config", configHandler(cfg), false)
+	a.registerRoute("/", http.HandlerFunc(indexHandler), false)
+}
+
+// RegisterDistributor registers the endpoints associated with the distributor.
+func (a *API) RegisterDistributor(d *distributor.Distributor, pushConfig distributor.Config) {
+	a.registerRoute("/api/v1/push", push.Handler(pushConfig, d.Push), true)
+	a.registerRoute("/distributor/all_user_stats", http.HandlerFunc(d.AllUserStatsHandler), false)
```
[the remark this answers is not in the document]
jtlisi: Personally, I prefer dashes for URLs. I can update accordingly accross the board for all the new routes if that works:

`/user_stats` -> `/api/v1/user-stats`
`all_user_stats` -> `/distributor/all-user-stats`

pkg/cortex/modules.go
```diff
@@ -98,6 +90,22 @@ func (m *ModuleName) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return m.Set(s)
 }
 
+func (t *Cortex) initAPI(cfg *Config) (services.Service, error) {
+	cfg.API.ServerPrefix = cfg.Server.PathPrefix
+	cfg.API.LegacyHTTPPrefix = cfg.HTTPPrefix
```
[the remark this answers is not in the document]
jtlisi: I'm not sure either. I always felt it was an issue if not a bug that services did not respect the configured prefix. I will add a line in the changelog explaining the change in behaviour. However, I don't think this goes against any of our 1.0 guarantees:
```
Additional API endpoints for management of Cortex itself, such as the ring. These APIs are not part of the any compatibility guarantees.
```

pkg/api/handlers.go
jtlisi: :+

pkg/api/handlers.go
jtlisi: 👍

pkg/api/handlers.go
jtlisi: 👍

pkg/api/handlers.go
jtlisi: Makes sense, I'll put them at the end and add `(experimental blocks storage)`

pkg/api/api.go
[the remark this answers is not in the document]
jtlisi: On second though, since Prometheus uses underscores Cortex should as well. `/query_range` for instance.

pkg/api/api.go
```diff
@@ -0,0 +1,322 @@
...
+	a.server.HTTP.Path(path).Methods(methods...).Handler(handler)
+}
+
+func (a *API) registerRoutesWithPrefix(path string, handler http.Handler, auth bool, methods ...string) {
+	level.Debug(a.logger).Log("msg", "api: registering route", "methods", strings.Join(methods, ","), "path", path, "auth", auth)
+	if auth {
+		handler = a.authMiddleware.Wrap(handler)
+	}
+	if len(methods) == 0 {
+		a.server.HTTP.Path(path).Handler(handler)
```
pracucci, line 99: Prefix?

pkg/api/api.go
```diff
@@ -0,0 +1,322 @@
...
+		return
+	}
+	a.server.HTTP.Path(path).Methods(methods...).Handler(handler)
+}
+
+func (a *API) registerRoutesWithPrefix(path string, handler http.Handler, auth bool, methods ...string) {
```
pracucci, line 93: I'm lost. We pick only the path which is set to be the prefix (in the code) while we should pick both the prefix and path. Am I missing anythin?

pkg/api/api.go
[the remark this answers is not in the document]
pracucci: Then I think all new routes should use underscores (I also prefer dashes, but my personal preference doesn't matter when it breaks consistency, so agree on underscores).

pkg/api/api.go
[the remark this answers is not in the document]
pracucci: Making it explicit could be a good idea 👍

pkg/api/api.go
pracucci, line 93: I would rename `path` to `prefix` to make it clearer

docs/configuration/config-file-reference.md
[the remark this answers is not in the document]
pracucci: Remember to address this.

pkg/api/api.go
jtlisi: Sorry about that, I was just looking into the failed integration tests and that looked to be the issue. Also the primary reason for not explicitly registering the alertmanager routes is because the Alertmanager UI routing is relatively complex and I didn't want to handle it in this PR. I also think it is fine to serve the Alertmanager directly if it is routed under the `/alertmanager/` prefix.

pkg/api/api.go
```diff
@@ -0,0 +1,322 @@
...
+
+// RegisterFlags adds the flags required to config this to the given FlagSet.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix("", f)
+}
+
+// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet with the set prefix.
+func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.StringVar(&cfg.AlertmanagerHTTPPrefix, prefix+"http.alertmanager-http-prefix", "/alertmanager", "HTTP URL path under which the Alertmanager ui and api will be served.")
+	f.StringVar(&cfg.PrometheusHTTPPrefix, prefix+"http.prometheus-http-prefix", "/prometheus", "TTP URL path under which the Prometheus api will be served.")
```
pracucci, line 52:
```suggestion
	f.StringVar(&cfg.PrometheusHTTPPrefix, prefix+"http.prometheus-http-prefix", "/prometheus", "HTTP URL path under which the Prometheus api will be served.")
```

pkg/api/api.go
```diff
@@ -0,0 +1,304 @@
...
+	HTTPAuthMiddleware middleware.Func `yaml:"-"`
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix("", f)
+}
+
+// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet with the set prefix.
+func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
```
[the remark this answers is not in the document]
jtlisi: We don't utilize this. I just added it out of habit to allow for easier reuse or embedding if that is something that is needed later.

pkg/cortex/modules.go
```diff
@@ -577,7 +527,7 @@ var modules = map[ModuleName]module{
 	},
 
 	Ring: {
-		deps:           []ModuleName{Server, RuntimeConfig, MemberlistKV},
```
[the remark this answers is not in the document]
jtlisi: I was under the understanding that modules inherit the dependencies of the their dependencies. I assumed since the API is dependent on the Server, all modules dependent on the API would be dependent on the server?

jtlisi:
I just wanted to circle back and address:
> Alertmanager: `/status` is available only when running in microservices mode and the target is the alertmanager

I don't know if that is worth addressing since there is no current way to run the alertmanager in the single binary. Even after this PR that functionality has not been enabled.

pkg/api/api.go
jtlisi: 👍

pkg/api/api.go
jtlisi: 👍

pkg/cortex/modules.go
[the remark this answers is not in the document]
pracucci: Sure you're right. My feedback was more about making it explicit. Anyway, it's a minor thing and very subjective. I'm resolving the comment.

pracucci:
> I just wanted to circle back and address:
> 
> > Alertmanager: `/status` is available only when running in microservices mode and the target is the alertmanager
> 
> I don't know if that is worth addressing since there is no current way to run the alertmanager in the single binary. Even after this PR that functionality has not been enabled.

🤦‍♂ You're definitely right. All good on this 👍

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master / unreleased
 
+* [CHANGE] Added v1 API routes documented in #2327. #2372
+  * Added `-http.alertmanager-http-prefix` and `http.prometheus-http-prefix` flags.
+  * Updated the index hosted at the root prefix to point to the updated routes.
+  * Legacy routes hardcoded with the `/api/prom` prefix now respect the `http.prefix` flag.
```
pracucci, line 8:
```suggestion
  * Legacy routes hardcoded with the `/api/prom` prefix now respect the `-http.prefix` flag.
```

PR #2374, opened 2020-04-01
improve config file doc

Wing924:

**What this PR does**:

improve config file doc by move some big blocks to separated sections

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/tools/doc-generator/main.go
+++ b/tools/doc-generator/main.go
@@ -98,7 +98,7 @@ var (
 		{
 			name:       "flusher_config",
 			structType: reflect.TypeOf(flusher.Config{}),
-			desc:       "The flusher_config configures how flusher to flush all the chunks.",
+			desc:       "The flusher_config configures the WAL flusher target, used to manually run one-time flushes when scaling down ingesters.",
 		},
 		{
 			name:       "chunk_store_config",
@@ -178,7 +178,7 @@ var (
 		{
 			name:       "store_gateway_config",
 			structType: reflect.TypeOf(storegateway.Config{}),
-			desc:       "The store_gateway_config configures the the read path for blocks sharding.",
+			desc:       "The store_gateway_config configures the store-gateway service used by the experimental blocks storage.",
 		},
 		{
 			name:       "purger_config",
--- a/tools/doc-generator/main.go
+++ b/tools/doc-generator/main.go
@@ -9,8 +9,6 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/weaveworks/common/server"
-
 	"github.com/cortexproject/cortex/pkg/alertmanager"
 	"github.com/cortexproject/cortex/pkg/chunk"
 	"github.com/cortexproject/cortex/pkg/chunk/cache"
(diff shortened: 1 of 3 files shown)


tools/doc-generator/main.go
```diff
@@ -93,6 +95,11 @@ var (
 			structType: reflect.TypeOf(storage.Config{}),
 			desc:       "The storage_config configures where Cortex stores the data (chunks storage engine).",
 		},
+		{
+			name:       "flusher_config",
+			structType: reflect.TypeOf(flusher.Config{}),
+			desc:       "The flusher_config configures how flusher to flush all the chunks.",
```
pracucci, line 101: @codesome Is my suggested change correct?

```suggestion
			desc:       "The flusher_config configures the WAL flusher target, used to manually run one-time flushes when scaling down ingesters.",
```

tools/doc-generator/main.go
```diff
@@ -168,6 +175,11 @@ var (
 			structType: reflect.TypeOf(compactor.Config{}),
 			desc:       "The compactor_config configures the compactor for the experimental blocks storage.",
 		},
+		{
+			name:       "store_gateway_config",
+			structType: reflect.TypeOf(storegateway.Config{}),
+			desc:       "The store_gateway_config configures the the read path for blocks sharding.",
```
pracucci, line 181:
```suggestion
			desc:       "The store_gateway_config configures the store-gateway service used by the experimental blocks storage.",
```

tools/doc-generator/main.go
```diff
@@ -31,7 +30,10 @@ import (
 	"github.com/cortexproject/cortex/pkg/ring/kv/memberlist"
 	"github.com/cortexproject/cortex/pkg/ruler"
 	"github.com/cortexproject/cortex/pkg/storage/tsdb"
+	"github.com/cortexproject/cortex/pkg/storegateway"
 	"github.com/cortexproject/cortex/pkg/util/validation"
+
+	"github.com/weaveworks/common/server"
```
pstibrany, line 36: 3rd party imports should be kept above cortex internal imports.

tools/doc-generator/main.go
```diff
@@ -168,6 +175,11 @@ var (
 			structType: reflect.TypeOf(compactor.Config{}),
 			desc:       "The compactor_config configures the compactor for the experimental blocks storage.",
 		},
+		{
+			name:       "store_gateway_config",
+			structType: reflect.TypeOf(storegateway.Config{}),
+			desc:       "The store_gateway_config configures the store-gateway service used by the experimental blocks storage.",
+		},
```
pstibrany, line 182: Exposing this may be little premature, given that this component is not yet finished. /cc @pracucci

tools/doc-generator/main.go
```diff
@@ -93,6 +95,11 @@ var (
 			structType: reflect.TypeOf(storage.Config{}),
 			desc:       "The storage_config configures where Cortex stores the data (chunks storage engine).",
 		},
+		{
+			name:       "flusher_config",
+			structType: reflect.TypeOf(flusher.Config{}),
+			desc:       "The flusher_config configures the WAL flusher target, used to manually run one-time flushes when scaling down ingesters.",
+		},
```
pstibrany, line 102: @codesome Are we ok with making this more public now?

tools/doc-generator/main.go
pracucci: You're actually right. Let's hide it 👍

Wing924: There is no way to hide this by doc-generator.
The only way to do that is add `doc:"hidden"` tag on Config struct.

You can check current config file docs looks not so good.
https://cortexmetrics.io/docs/configuration/configuration-file/#supported-contents-and-default-values-of-the-config-file

pracucci: Yeah, definitely. We did mean to use `doc:"hidden"` on the store gateway config.

Wing924: I've added doc:hidden

PR #2375, opened 2020-04-01
Decouple blocks meta fetcher metrics mapping from bucket store metrics

pracucci:
**What this PR does**:
In this PR I'm decoupling blocks meta fetcher metrics mapping from bucket store metrics, as preliminary refactoring required by the on-going work to introduce the `store-gateway` service (because in the near future the querier will not run the bucket store anymore).

I've also took the chance to rename metrics to remove the `_bucket_store` portion from meta fetcher metrics, for the same reason.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany:
Nice job!

PR #2376, opened 2020-04-01
integration: Moved image references to single place.

pstibrany:
**What this PR does**: This trivial PR moves all docker image references used in integration tests to single place, to keep them together and remove repeated comment.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

/cc @bwplotka

### DIFF
--- a/integration/backward_compatibility_test.go
+++ b/integration/backward_compatibility_test.go
@@ -14,18 +14,17 @@ import (
 	"github.com/cortexproject/cortex/integration/e2e"
 	e2edb "github.com/cortexproject/cortex/integration/e2e/db"
 	"github.com/cortexproject/cortex/integration/e2ecortex"
+	"github.com/cortexproject/cortex/integration/images"
 )
 
 var (
-	// If you change the image tag, remember to update it in the preloading done
-	// by CircleCI too (see .circleci/config.yml).
 	previousVersionImages = []string{
 		// 0.6.0 used 204 status code for querier and ingester
 		// distributor didn't have /ready page, and we used check on the /ring page instead
-		"quay.io/cortexproject/cortex:v0.6.0",
+		images.Cortex_0_6_0,
 
 		// 0.7.0 used 204 status code for all components
-		"quay.io/cortexproject/cortex:v0.7.0",
+		images.Cortex_0_7_0,
 	}
 )
 
--- a/integration/backward_compatibility_test.go
+++ b/integration/backward_compatibility_test.go
@@ -21,10 +21,10 @@ var (
 	previousVersionImages = []string{
 		// 0.6.0 used 204 status code for querier and ingester
 		// distributor didn't have /ready page, and we used check on the /ring page instead
-		images.Cortex_0_6_0,
+		images.Cortex0_6_0,
 
 		// 0.7.0 used 204 status code for all components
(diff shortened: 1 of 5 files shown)


pstibrany:
> If we consider this package as a library, I think we might want something more flexible than globals ):
> 
> I know it's more work, but I would vote for parameter instead, so Cortex can have it's own globals and Thanos can choose to use different version etc... WDYT? 🤔

What about making a global map with images?

```
var Images map[string]string = {
    "memcached": "memcached:1.6.1",
    "minio": "minio/minio:RELEASE.2019-12-30T05-45-39Z",
    ...
}
```

This way, we would keep image references at one place, and they can still by updated by client code via `init` method? Passing image name as a parameter would affect lot of code, for little benefit.

bwplotka:
I mean for now it's hardcoded anyway, so global map or var is better than nothings (: Would do for now (: In terms of map I think just global `vars` would do for now.

Once we will need parameters we will think what to do (:

pstibrany:
> Once we will need parameters we will think what to do (:

I'd like to introduce functional parameters here. So that we can write things like: `e2e.NewHTTPService("name", Image("..."), ReadinessProbe(...))` etc.

That would also allow us to introduce new options without touching half of the tests each time.

bwplotka:
In this case maybe struct would do better, no?

pstibrany:
> In this case maybe struct would do better, no?

I agree structs would work too. How would be better though?

bwplotka:
Because it's always better to refer to things by variable name that is typed than random string (:

pstibrany:
> Because it's always better to refer to things by variable name that is typed and hast to exists on compile time than random string (:

Using functional parameters are not random strings, but functions that have to have correct signature. So they are checked at compile time. Similar to go-kit logger, or many other examples. (https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)

bwplotka:
Yea, I know what functional arguments works, this will work perfectly for sure.

But this does not looks like functional argument, so that's why I was confused.

```
var Images = map[string]string{
    "memcached": "memcached:1.6.1",
    "minio": "minio/minio:RELEASE.2019-12-30T05-45-39Z",
    ...
}
```

As long as this map does not exists, and some functional args are used it would be perfect yes (:

pstibrany:
> Yea, I know what functional arguments works, this will work perfectly for sure.
> 
> But this does not looks like functional argument, so that's why I was confused.
> 
> ```
> var Images = map[string]string{
>     "memcached": "memcached:1.6.1",
>     "minio": "minio/minio:RELEASE.2019-12-30T05-45-39Z",
>     ...
> }
> ```
> 
> As long as this map does not exists, and some functional args are used it would be perfect yes (:

Ah, sorry, I misunderstood you. I've changed opinion on this, and used `var` instead of `const`. That way they can still be overwritten, but also compile-time checked.

(I wasn't going to use raw strings like that anyway, but would provide key constants.)

pstibrany:
> > Yea, I know what functional arguments works, this will work perfectly for sure.
> > But this does not looks like functional argument, so that's why I was confused.

Functional arguments suggestion was proposal for further changes in this library, but not in this PR. It was reaction to your comment:

> Once we will need parameters we will think what to do (:

PR #2379, opened 2020-04-01
Remove the ticker from WAL.run()

codesome:
As we are already spreading the checkpoint over the entire checkpoint duration, we no longer need this ticker here.

### DIFF
--- a/pkg/ingester/wal.go
+++ b/pkg/ingester/wal.go
@@ -156,12 +156,21 @@ func (w *walWrapper) run() {
 		return
 	}
 
-	ticker := time.NewTicker(w.cfg.CheckpointDuration)
-	defer ticker.Stop()
-
+	lastCheckpoint := time.Now().Add(-w.cfg.CheckpointDuration)
 	for {
 		select {
-		case <-ticker.C:
+		case <-w.quit:
+			level.Info(util.Logger).Log("msg", "creating checkpoint before shutdown")
+			if err := w.performCheckpoint(true); err != nil {
+				level.Error(util.Logger).Log("msg", "error checkpointing series during shutdown", "err", err)
+			}
+			return
+		default:
+			timeElapsedSinceLastCheckpoint := time.Since(lastCheckpoint)
+			if timeElapsedSinceLastCheckpoint < w.cfg.CheckpointDuration {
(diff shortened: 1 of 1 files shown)


pkg/ingester/wal.go
```diff
@@ -156,12 +156,21 @@ func (w *walWrapper) run() {
...
+			level.Info(util.Logger).Log("msg", "creating checkpoint before shutdown")
+			if err := w.performCheckpoint(true); err != nil {
+				level.Error(util.Logger).Log("msg", "error checkpointing series during shutdown", "err", err)
+			}
+			return
+		default:
+			timeElapsedSinceLastCheckpoint := time.Now().Sub(lastCheckpoint)
+			if timeElapsedSinceLastCheckpoint < w.cfg.CheckpointDuration {
+				time.Sleep(w.cfg.CheckpointDuration - timeElapsedSinceLastCheckpoint)
+			}
```
codesome, line 172: This is better than the ticker because: we do not take into account the time taken to write the checkpoint and new series added in between. So checkpoint usually takes more than the checkpoint duration to write. This can cause a tick to be buffered up eventually if many checkpoints take more time. Hence if suddenly there is a drop in series and the checkpoint gets over before the checkpoint duration, the next checkpoint will be started before the checkpoint duration completition. This makes sure it never happens. (Microoptimization? Yes :P but checkpoints cause some disk throttling so lets avoid it)

pkg/ingester/wal.go
```diff
@@ -156,12 +156,21 @@ func (w *walWrapper) run() {
...
 	for {
 		select {
-		case <-ticker.C:
+		case <-w.quit:
+			level.Info(util.Logger).Log("msg", "creating checkpoint before shutdown")
+			if err := w.performCheckpoint(true); err != nil {
+				level.Error(util.Logger).Log("msg", "error checkpointing series during shutdown", "err", err)
+			}
+			return
+		default:
```
gouthamve, line 168: The problem with this is, we don't run a checkpoint on exit until the sleep + checkpoint completes

codesome: You are right. I can have the sleep in a `select {}` with `<-time.After()`.

codesome: (or instead of `default:`, replace it with `<-time.After()`)

codesome:
No longer relevant

PR #2383, opened 2020-04-01
Return first validation error on sample ingestion

gotjosh:
**What this PR does**:

This commit changes which error we return on validation back to the
clients (Prometheus or Distributors).

The samples that do not pass validation at distributors and/or ingesters
now return the first validation error as opposed to the last.

Abstract from a conversation in #2336 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -357,7 +357,7 @@ func (i *Ingester) Push(ctx context.Context, req *client.WriteRequest) (*client.
 		return nil, fmt.Errorf("no user id")
 	}
 
-	var lastPartialErr *validationError
+	var firstPartialErr *validationError
 	var record *Record
 	if i.cfg.WALConfig.WALEnabled {
 		record = recordPool.Get().(*Record)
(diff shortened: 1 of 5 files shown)


pkg/ingester/ingester.go
```diff
@@ -357,7 +357,7 @@ func (i *Ingester) Push(ctx context.Context, req *client.WriteRequest) (*client.
 		return nil, fmt.Errorf("no user id")
 	}
 
-	var lastPartialErr *validationError
```
gotjosh, line 360: TSDB blocks ingester already does this.

gotjosh:
Hi everyone 👋,

A question and a statement from me.

1. It feels like this behaviour should be documented somewhere. My vote goes to the [Remote API section](https://cortexmetrics.io/docs/apis/#remote-api) but please let me know if there's a better place for that.

2. One of the reasons behind this change is that with the work being done in https://github.com/cortexproject/cortex/pull/2336 we'll also introduce metadata failure errors. Given I propose processing series/samples first, returning the last error (which would be metadata related) doesn't really make sense.

pracucci:
> 1. It feels like this behaviour should be documented somewhere. My vote goes to the Remote API section but please let me know if there's a better place for that.

Yes, please. To me there's no strong need to do it within this PR. You can take a bit more time and work on a overall improvement of the "Cortex API" doc which include mentioning this behaviour.

PR #2390, opened 2020-04-02
add support for creating more ObjectClient types and flush file immediately for filesystem type

sandeepsukhani:
**What this PR does**:
We already implement `ObjectClient` for various storages but `NewObjectClient` was allowing them all. This PR adds support for them.
It also adds a call to `File.Flush` method in `PutObject` method of filesystem type to make it flush the file to disk immediately.

### DIFF
--- a/pkg/chunk/local/fs_object_client.go
+++ b/pkg/chunk/local/fs_object_client.go
@@ -62,22 +62,31 @@ func (f *FSObjectClient) GetObject(ctx context.Context, objectKey string) (io.Re
 }
 
 // PutObject into the store
-func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object io.ReadSeeker) error {
+func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object io.ReadSeeker) (err error) {
 	fullPath := path.Join(f.cfg.Directory, objectKey)
-	err := util.EnsureDirectory(path.Dir(fullPath))
+	err = util.EnsureDirectory(path.Dir(fullPath))
 	if err != nil {
-		return err
+		return
 	}
 
 	fl, err := os.OpenFile(fullPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
 	if err != nil {
-		return err
+		return
 	}
 
-	defer fl.Close()
+	defer func() {
+		if fcErr := fl.Close(); fcErr != nil {
+			if err == nil {
+				err = fcErr
+				return
+			}
+			level.Error(pkgUtil.Logger).Log("msg", "error closing file", "filepath", fullPath, "err", fcErr)
+		}
+	}()
 
-	if _, err := io.Copy(fl, object); err != nil {
-		return err
+	_, err = io.Copy(fl, object)
+	if err != nil {
+		return
 	}
 
 	return fl.Sync()
--- a/pkg/chunk/local/fs_object_client.go
+++ b/pkg/chunk/local/fs_object_client.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log/level"
+	"github.com/thanos-io/thanos/pkg/runutil"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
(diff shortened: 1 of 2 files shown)


pkg/chunk/storage/factory.go
```diff
@@ -265,11 +265,17 @@ func NewBucketClient(storageConfig Config) (chunk.BucketClient, error) {
 // NewObjectClient makes a new StorageClient of the desired types.
 func NewObjectClient(name string, cfg Config) (chunk.ObjectClient, error) {
 	switch name {
+	case "aws", "s3":
```
pracucci, line 268: Given we're adding the support now, I would suggest to not support the ambiguous name `aws`, unless this results into any issue (if so, which one?):

```suggestion
	case "s3":
```

pkg/chunk/storage/factory.go
```diff
@@ -265,11 +265,17 @@ func NewBucketClient(storageConfig Config) (chunk.BucketClient, error) {
...
+		return gcp.NewGCSObjectClient(context.Background(), cfg.GCSConfig, chunk.DirDelim)
+	case "azure":
+		return azure.NewBlobStorage(&cfg.AzureStorageConfig, chunk.DirDelim)
 	case "inmemory":
 		return chunk.NewMockStorage(), nil
 	case "filesystem":
 		return local.NewFSObjectClient(cfg.FSConfig)
 	default:
-		return nil, fmt.Errorf("Unrecognized storage client %v, choose one of: filesystem", name)
+		return nil, fmt.Errorf("Unrecognized storage client %v, choose one of: aws, s3, gcs, azure, filesystem", name)
```
pracucci, line 279:
```suggestion
		return nil, fmt.Errorf("Unrecognized storage client %v, choose one of: s3, gcs, azure, filesystem", name)
```

pkg/chunk/storage/factory.go
sandeepsukhani: The reason for keeping it like this is because we support both `aws` and `s3` for creating object client for chunks in S3 and I want to keep it consistent because for boltdb shipper we are going to pick up the type of storage that is configured for storing chunks. Does that make sense?

pracucci: Yes, consistency is a good argument in favor of it.

pkg/chunk/local/fs_object_client.go
```diff
@@ -76,8 +76,11 @@ func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object
 
 	defer fl.Close()
 
-	_, err = io.Copy(fl, object)
-	return err
+	if _, err := io.Copy(fl, object); err != nil {
+		return err
+	}
+
+	return fl.Sync()
```
pstibrany, line 83: `Close` may also return error that should be reported to the caller.

pkg/chunk/local/fs_object_client.go
```diff
@@ -62,22 +62,34 @@ func (f *FSObjectClient) GetObject(ctx context.Context, objectKey string) (io.Re
...
+		}
+	}()
 
 	_, err = io.Copy(fl, object)
-	return err
+	if err != nil {
+		return
+	}
+
+	return fl.Sync()
```
pstibrany, line 92: Can we step back to previous design without named params, and error handling in main code and not defer?

```suggestion
func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object io.ReadSeeker) error {
	fullPath := path.Join(f.cfg.Directory, objectKey)
	err := util.EnsureDirectory(path.Dir(fullPath))
	if err != nil {
		return err
	}

	fl, err := os.OpenFile(fullPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
	if err != nil {
		return err
	}

	defer runutil.CloseWithLogOnErr(pkgUtil.Logger, fl, "fullPath: %s", fullPath)

	_, err = io.Copy(fl, object)
	if err != nil {
		return err
	}

	err = fl.Sync()
	if err != nil {
		return err
	}

	return fl.Close()
}
```

PR #2391, opened 2020-04-02
Decouple gRPC health check service from ingesters

pracucci:
**What this PR does**:
In order to build the `store-gateway` service ([proposal](https://cortexmetrics.io/docs/proposals/blocks-storage-sharding/)), I would need to be able to do gRPC health checks against the `store-gateway` when running in microservices mode. However, the gRPC health check service is currently registered by the ingester so can't be also registered in other places (otherwise it clashes).

Given the `Ingester.Check()` implementation is not ingester-specific (it just checks if the ingester service is running/stopping), in this PR I'm proposing to generalise the gRPC health check service registration and mark the Cortex instance as healthy once all services are running, making it consistent with the `/ready` endpoint.

Notes:
- ~~`services.Manager.IsHealthy()` returns true after all services are running and until all of them are stopped (so stopping is considered as healthy, @pstibrany correct me if I'm wrong)~~
- Loki is not affected by this change

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"os"
 
+	"google.golang.org/grpc/health/grpc_health_v1"
+
 	"github.com/cortexproject/cortex/pkg/configs"
 	"github.com/cortexproject/cortex/pkg/storegateway"
 
@@ -316,8 +318,10 @@ func (t *Cortex) Run() error {
 		return err
 	}
 
-	// before starting servers, register /ready handler. It should reflect entire Cortex.
+	// before starting servers, register /ready handler and gRPC health check service.
+	// It should reflect entire Cortex.
 	t.server.HTTP.Path("/ready").Handler(t.readyHandler(sm))
+	grpc_health_v1.RegisterHealthServer(t.server.GRPC, newHealthCheck(sm))
 
 	// Let's listen for events from this manager, and log them.
 	healthy := func() { level.Info(util.Logger).Log("msg", "Cortex started") }
--- a/pkg/cortex/health_check.go
+++ b/pkg/cortex/health_check.go
@@ -1,0 +1,35 @@
+package cortex
+
+import (
+	"context"
+
+	"github.com/gogo/status"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/health/grpc_health_v1"
+
+	"github.com/cortexproject/cortex/pkg/util/services"
+)
+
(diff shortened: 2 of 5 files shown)


pkg/cortex/cortex.go
```diff
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"os"
 
+	"google.golang.org/grpc/health/grpc_health_v1"
+
```
pstibrany, line 12: Can we please fix imports in this file? Our check apparently failed here. :-(

pstibrany:
> * `services.Manager.IsHealthy()` returns true after all services are running and until all of them are stopped (so stopping is considered as healthy

`services.Manager.IsHealthy()` returns true only if ALL services are running. If any service is in different than `Running` state, `services.Manager.IsHealthy()` will return false.

pracucci:
> > * `services.Manager.IsHealthy()` returns true after all services are running and until all of them are stopped (so stopping is considered as healthy
> 
> `services.Manager.IsHealthy()` returns true only if ALL services are running. If any service is in different than `Running` state, `services.Manager.IsHealthy()` will return false.

You're right (I initially misread the `IsHealthy()` implementation). Then I need to change this PR logic because the gRPC health check should succeed while stopping (it's OK for the ingesters, which need to transfer chunks).

pracucci:
@pstibrany I've changed the gRPC health check logic to keep it closer to the previous one: it's healthy once all services are running and until at least of them is still stopping. Not healthy until there's still someone starting or anyone has failed.

1. Does it make sense to you?
2. I've the feeling the `mockService` in the unit tests is over complicated, but haven't found a smarter way to mock it (you're the guru of the services framework, maybe you have the 🔮 )

pkg/cortex/health_check.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// isHealthy returns whether the Cortex instance should be considered healthy.
+func (h *healthCheck) isHealthy() bool {
+	states := h.sm.ServicesByState()
+
+	// Given this is an health check endpoint for the whole instance, we should consider
+	// it healthy after all services have been started (running) and until there's
+	// still a service stopping (because some services, like ingesters, are still
+	// fully functioning while stopping).
```
pstibrany, line 44:
```suggestion
	// Given this is an health check endpoint for the whole instance, we should consider
	// it healthy after all services have been started (running) and until all
	// services are terminated. Some services, like ingesters, are still
	// fully functioning while stopping.
```

PR #2394, opened 2020-04-02
Generalize client pool

pracucci:
**What this PR does**:
In order to build the `store-gateway` service ([proposal](https://cortexmetrics.io/docs/proposals/blocks-storage-sharding/)), I would need to use a pool of store-gateway clients (because the store-gateway join a ring when sharding is enabled).

The current `Pool` implementation in `pkg/ingester/client` is almost generalised except:
- The config references `ingester` 
- Metric
- `ingester` references in logs/errors

In this PR I'm proposing to generalise it. In particular:
- Introduced `PoolClient` interface (given `io.Closer` is also required by the pool)
- Take the clients count metric in input
- I've kept `pkg/ingester/client` to guarantee config backward-compatibility and also easy the integration (Loki too)

Notes to reviewers:
- To simplify the diff I've submitted two commits: (1) [files moving only](https://github.com/cortexproject/cortex/commit/402f5d6e024f5da92d2c6b1f8db9d665c8ff3694) (2) [all others changes](https://github.com/cortexproject/cortex/commit/d30ed506818fbe8a78e1aeb20b7001bd0ebc2860). Reviewing (2) will be enough and easier.

@cyriltovena May you also review it, given it will sligthly affect Loki too.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cyriltovena:
I’ll take a look on Monday!

pkg/ingester/client/pool.go
```diff
@@ -42,156 +31,12 @@ func (cfg *PoolConfig) RegisterFlags(f *flag.FlagSet) {
...
-// NewPool creates a new Pool.
-func NewPool(cfg PoolConfig, ring ring.ReadRing, factory Factory, logger log.Logger) *Pool {
-	p := &Pool{
-		cfg:     cfg,
-		ring:    ring,
-		factory: factory,
-		logger:  logger,
-
-		clients: map[string]grpc_health_v1.HealthClient{},
+func NewPool(cfg PoolConfig, ring ring.ReadRing, factory ring_client.PoolFactory, logger log.Logger) *ring_client.Pool {
```
pstibrany, line 34: Would it make sense to pass `clients` gauge as a parameter? It's name is `cortex_distributor_ingester_clients` after all, so it should be in distributor. 

Actually, looking at it, I think this entire file should go to distributor package now, since it's completely used from there. What do you think?

pkg/ring/client/pool.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Pool holds a cache of grpc_health_v1 clients.
+type Pool struct {
+	services.Service
+
+	cfg         PoolConfig
+	ring        ring.ReadRing
+	factory     PoolFactory
+	logger      log.Logger
+	serviceName string
```
pstibrany, line 46: I would call this `clientName` or `clientType`, to avoid confusion with this being a pool's name (since it's a service).

pkg/ring/client/pool.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+		}
+	}
+}
+
+// healthCheck will check if the client is still healthy, returning an error if it is not
+func healthCheck(client PoolClient, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	ctx = user.InjectOrgID(ctx, "0")
```
pstibrany, line 188: I wonder why we need to do this (unrelated to this PR).

pracucci: It's required by the `ServerUserHeaderInterceptor` middleware for any gRPC call.

pkg/ingester/client/pool.go
pracucci: Agree. Moved to `pkg/distributor`.

pkg/ring/client/pool.go
```diff
@@ -1,8 +1,7 @@
-package client
+package grpcclient
```
cyriltovena, line 1: I'm wondering is that ok to have a package name different than the folder name ?

cyriltovena: The only include I've seen so far is renaming I'm wondering if that has an effect we you don't rename ?

pracucci: Thanks for noticing it! It wasn't intended and I've fixed it.

PR #2403, opened 2020-04-03
Handle cache-control headers also during a partial cache hit.

cyriltovena:
I was reading this code and realize that #1974 missed a place where we also query downstream for partial result in case of partial cache hit.

I've chatted about this with on Thanos dev and suggested to send a fix for them.

/cc @GiedriusS 

### DIFF
--- a/pkg/querier/queryrange/results_cache.go
+++ b/pkg/querier/queryrange/results_cache.go
@@ -168,7 +168,7 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
 }
 
 // shouldCacheResponse says whether the response should be cached or not.
-func shouldCacheResponse(r Response) bool {
+func (s resultsCache) shouldCacheResponse(r Response) bool {
 	if promResp, ok := r.(*PrometheusResponse); ok {
 		shouldCache := true
 	outer:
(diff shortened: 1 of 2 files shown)


pkg/querier/queryrange/results_cache.go
```diff
@@ -182,6 +182,7 @@ func shouldCacheResponse(r Response) bool {
 			for _, v := range hv.Values {
 				if v == noCacheValue {
 					shouldCache = false
+					level.Debug(s.logger).Log("msg", fmt.Sprintf("%s header in response is equal to %s, not caching the response", cachecontrolHeader, noCacheValue))
 					break outer
 				}
 			}
```
pstibrany:
```suggestion
				if v == noCacheValue {
					level.Debug(s.logger).Log("msg", fmt.Sprintf("%s header in response is equal to %s, not caching the response", cachecontrolHeader, noCacheValue))
					return false
				}
			}
```
(also get rid of `shouldCache`)

cyriltovena:
> LGTM.
> 
> I've removed some extra code from `shouldCacheResponse` to avoid extra review cycle, hope that's ok.

Of course @pstibrany  !

PR #2404, opened 2020-04-04
BugFix: 1.0 Ingesters having availability zone value erased by pre-1.0 ingesters during rolling upgrade.

### LINKED ISSUE #2401: Cortex 1.0.0 ingesters rollout breaks zone set in the ring
While upgrading to Cortex 1.0.0 and doing a rolling update of ingesters, I've noticed only the last rolled out ingester has a zone set in the ring, while all previous one (already rolled out) haven't.

The issue is caused by the co-existence of old and new ingesters. Old ingesters don't know the zone in the ring data structure so when they do deserialise the ring, the zone is lost. This issue disappear once the last ingester rollout is completed but at this point the zone for all ingesters (except the last one) is already lost.

khaines:
**What this PR does**: Corrects a bug reported by @pracucci when upgrading ingesters from pre-1.0, which causes the new ingesters to have their availability zone string overwritten by a blank value. This PR addresses the issue by having the ingesters update the ring data with their zone on every heartbeat, similar to what they do with address information.

Fixes #2401 

**Checklist**
- [ x ] Tests updated
- [ X ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [BUGFIX] Experimental TSDB: fixed chunk data corruption when querying back series using the experimental blocks storage. #2400
 * [BUGFIX] Cassandra Storage: Fix endpoint TLS host verification. #2109
 * [BUGFIX] Experimental TSDB: fixed response status code from `422` to `500` when an error occurs while iterating chunks with the experimental blocks storage. #2402
+* [BUGFIX] Ring: Fixed a situation where upgrading from pre-1.0 cortex with a rolling strategy caused new 1.0 ingesters to lose their zone value in the ring until manually forced to re-register. #2404
 
 ## 1.0.0 / 2020-04-02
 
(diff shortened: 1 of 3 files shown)


pstibrany:
This is a step in the right direction. Unfortunately, it doesn't fix the fact that old ingesters (not supporting the zones) will clear the zone information for *all* ingesters on each heartbeat.

I think we should also make "no zone" to be empty string (and not ingester ID as it is today), to avoid seeing artifacts like this when watching the /ring page. Also visual check of the /ring would be faster -- it would be easy to see which ingesters have zone set. What do you think?

khaines:
@pstibrany this is only an issue during a rolling upgrade from pre1.0, once the old ingesters are rolled off the zones will be no longer overwritten each heartbeat. 

since it was added to ignore zoning for a blank string, we should make that the default "no zone"

PR #2408, opened 2020-04-06
support for registering custom table client with index client

sandeepsukhani:
**What this PR does**:
We had added support for registering custom index client. This PR also adds support for registering custom table client with it.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/storage/factory.go
+++ b/pkg/chunk/storage/factory.go
@@ -28,7 +28,7 @@ const (
 	StorageEngineTSDB   = "tsdb"
 )
 
-type indexClientFactories struct {
+type indexStoreFactories struct {
 	indexClientFactoryFunc IndexClientFactoryFunc
 	tableClientFactoryFunc TableClientFactoryFunc
 }
@@ -39,17 +39,17 @@ type IndexClientFactoryFunc func() (chunk.IndexClient, error)
 // TableClientFactoryFunc defines signature of function which creates chunk.TableClient for managing tables in index store
 type TableClientFactoryFunc func() (chunk.TableClient, error)
 
-var customIndexClients = map[string]indexClientFactories{}
+var customIndexStores = map[string]indexStoreFactories{}
 
-// RegisterIndexType is used for registering a custom index type.
+// RegisterIndexStore is used for registering a custom index type.
 // When an index type is registered here with same name as existing types, the registered one takes the precedence.
-func RegisterIndexType(name string, indexClientFactory IndexClientFactoryFunc, tableClientFactory TableClientFactoryFunc) {
-	customIndexClients[name] = indexClientFactories{indexClientFactory, tableClientFactory}
+func RegisterIndexStore(name string, indexClientFactory IndexClientFactoryFunc, tableClientFactory TableClientFactoryFunc) {
+	customIndexStores[name] = indexStoreFactories{indexClientFactory, tableClientFactory}
 }
 
 // useful for cleaning up state after tests
-func unregisterAllCustomIndexClients() {
-	customIndexClients = map[string]indexClientFactories{}
+func unregisterAllCustomIndexStores() {
+	customIndexStores = map[string]indexStoreFactories{}
 }
 
 // StoreLimits helps get Limits specific to Queries for Stores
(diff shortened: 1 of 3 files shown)


pkg/chunk/storage/factory.go
```diff
@@ -28,17 +28,28 @@ const (
...
+// TableClientFactoryFunc defines signature of function which creates chunk.TableClient for managing tables in index store
+type TableClientFactoryFunc func() (chunk.TableClient, error)
 
-func RegisterIndexClient(name string, factory IndexClientFactoryFunc) {
-	customIndexClients[name] = factory
+var customIndexClients = map[string]indexClientFactories{}
+
+// RegisterIndexType is used for registering a custom index type.
+// When an index type is registered here with same name as existing types, the registered one takes the precedence.
+func RegisterIndexType(name string, indexClientFactory IndexClientFactoryFunc, tableClientFactory TableClientFactoryFunc) {
```
pracucci, line 46: As discussed offline, we agreed on naming this `RegisterIndexStore` and rename "index client" into "index store" across the file (as far as concerning the custom index store registration).

pkg/chunk/storage/factory.go
```diff
@@ -28,12 +28,23 @@ const (
 	StorageEngineTSDB   = "tsdb"
 )
 
+type indexStoreFactories struct {
+	indexClientFactoryFunc IndexClientFactoryFunc
+	tableClientFactoryFunc TableClientFactoryFunc
+}
```
pstibrany, line 34: Since these functions don't seem to be used together, they could simply be stored in two different maps.

pkg/chunk/storage/factory.go
```diff
@@ -28,12 +28,23 @@ const (
...
+type TableClientFactoryFunc func() (chunk.TableClient, error)
 
-func RegisterIndexClient(name string, factory IndexClientFactoryFunc) {
-	customIndexClients[name] = factory
+var customIndexStores = map[string]indexStoreFactories{}
+
+// RegisterIndexStore is used for registering a custom index type.
+// When an index type is registered here with same name as existing types, the registered one takes the precedence.
+func RegisterIndexStore(name string, indexClientFactory IndexClientFactoryFunc, tableClientFactory TableClientFactoryFunc) {
+	customIndexStores[name] = indexStoreFactories{indexClientFactory, tableClientFactory}
```
pstibrany, line 47: Should we panic, if both functions are nil? (Alternatively, we can have two separate Register functions, each for one type, and panic if supplied function is nil).

pstibrany: Since `RegisterIndexStore` is not used anywhere (except in the tests), it's hard to see if this is a problem or not, but in general, there is synchronization missing on access to `customIndexStores` map. It may be worth creating a small struct to keep this map (or two maps), and use a mutex to protect access to it.

pracucci: > Should we panic, if both functions are nil? 

I don't think so. Passing `nil` could be a legit way to unregister an index store, unless we offer an `UnregisterIndexStore` function.

pkg/chunk/storage/factory.go
sandeepsukhani: They could be used together and in most cases, they would be. TableClient manages the tables while IndexClient manages the index within those tables. If you are defining a custom IndexStore you most likely would need both of them but it would not be the case always. It depends on the use case. The current requirement for boltdb shipper implementation in loki needs both. Does it make sense?

pstibrany: What I meant is that only one of two functions is used each time in `factory.go` (in `NewTableClient` and `NewIndexClient`). But this is just nitpicking, if you prefer this version, let's keep it.

pkg/chunk/storage/factory.go
sandeepsukhani: > Should we panic, if both functions are nil?
It is up to the user of this to decide whether they need both IndexClient or TableClient or one of them to be custom. Most of the cases should have both of them custom-defined but it would not be the case always. I do not see any value in keeping two different maps and functions to register them separately. It might be just me though.

> It may be worth creating a small struct to keep this map (or two maps), and use a mutex to protect access to it.
I do not think reads and writes would ever be simultaneous. `NewStore` call in Cortex initializes all stores at once and it is expected to have all the custom clients registered by then which means all the writes would be done before `NewStore` is called. Also, the custom client registration and `NewStore` would never run async since they both run in the same goroutine most likely main. I do not see a usecase of ever running them async.
What do you think?

pstibrany: > I do not think reads and writes would ever be simultaneous.

It's not about simultaneous reads/writes, but about reading and writing from different goroutines (at different times). To make sure that writes are visible to readers, you need *some* kind of happens-before relationship. Mutex is one possibility. Creating all reading goroutines only after registering these functions is another one. Since it's not used in real code currently, it's not wrong (yet), but lack of synchronization is a red flag.

> NewStore call in Cortex initializes all stores at once and it is expected to have all the custom clients registered by then which means all the writes would be done before NewStore is called.

This would indeed establish happens-before relationship.

PR #2410, opened 2020-04-06
Refactor: Unify metadata and samples sends to ingesters

gotjosh:
**What this PR does**:

Previously, we had samples and metadata using the ring on their own to
send data to the ingesters. Since the `WriteRequest` supports metadata
alongside samples there's no reason to split the calls so we unify them. There's an
additional tiny refactor to inline the metadata validation as it is just
a single function call.

A follow up from #2336 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/go-kit/kit/log/level"
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -354,10 +353,6 @@ func (d *Distributor) validateSeries(ts ingester_client.PreallocTimeseries, user
 		nil
 }
 
-func (d *Distributor) validateMetadata(m *ingester_client.MetricMetadata, userID string) error {
-	return validation.ValidateMetadata(d.limits, userID, m)
-}
-
 // Push implements client.IngesterServer
 func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*client.WriteResponse, error) {
 	userID, err := user.ExtractOrgID(ctx)
@@ -470,7 +465,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	}
 
 	for _, m := range req.Metadata {
-		err := d.validateMetadata(m, userID)
+		err := validation.ValidateMetadata(d.limits, userID, m)
 
 		if err != nil {
 			if firstPartialErr == nil {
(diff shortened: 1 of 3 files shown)


pkg/distributor/distributor.go
```diff
@@ -519,49 +514,32 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
-		}, func() {})
-		if err != nil {
-			// Metadata is a best-effort approach so we consider failures non-fatal, log them, and move on.
-			logger := util.WithContext(ctx, util.Logger)
-			level.Error(logger).Log("msg", "Failed to send metadata to ingesters", "err", err)
-		}
-	}
+	err = ring.DoBatch(ctx, subRing, keys, func(ingester ring.IngesterDesc, indexes []int) error {
+		timeseries := make([]client.PreallocTimeseries, 0, len(indexes))
+		metadata := make([]*client.MetricMetadata, 0, len(indexes))
```
gotjosh, line 522: Is this the most efficient way to allocate these spaces? Given we don't really know how many of the indexes are actually for metadata or samples (w/o changing the ring function - which I tried to avoid) we at least know that both of them are less the total number of indexes.

pkg/distributor/distributor.go
```diff
@@ -470,7 +465,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 	}
 
 	for _, m := range req.Metadata {
-		err := d.validateMetadata(m, userID)
```
gotjosh, line 473: This was suggested by @pstibrany.

pkg/ingester/ingester.go
```diff
@@ -373,10 +373,8 @@ func (i *Ingester) Push(ctx context.Context, req *client.WriteRequest) (*client.
 	}
 
 	if len(req.Metadata) > 0 {
-		// Given requests can only contain either metadata or samples, no-op if there is metadata for now.
```
gotjosh, line 376: Technically, this is "no longer true" from an implementation perspective. The remote write protocol only sends metadata or samples (for now) but IMO that comment/responsibility belongs in the distributor. The ingester should operate under no assumption that you'd receive one or the other if we can (e.g. no increase of complexity). 

Obviously, this is my own personal opinion with what I know about the system, but others feel free to suggest any alternatives.

gotjosh:
Should I be updating any tests? I didn't see anything that felt worth updating but feel free to suggest otherwise.

pkg/distributor/distributor.go
```diff
@@ -519,49 +514,32 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
...
-			metadata := make([]*client.MetricMetadata, 0, len(indexes))
-			for _, i := range indexes {
-				metadata = append(metadata, validatedMetadata[i])
-			}
+	keys := append(seriesKeys, metadataKeys...)
+	initialMetadataIndex := len(seriesKeys)
 
-			localCtx, cancel := context.WithTimeout(context.Background(), d.cfg.RemoteTimeout)
-			defer cancel()
-			localCtx = user.InjectUserID(localCtx, userID)
```
gotjosh, line 531: This was a bug 😞. It should have been `InjectOrgID` - This would have been caught by an e2e test hopefully.

pkg/distributor/distributor.go
```diff
@@ -597,32 +575,23 @@ func (d *Distributor) sendSamples(ctx context.Context, ingester ring.IngesterDes
...
-	h, err := d.ingesterPool.GetClientFor(ingester.Addr)
-	if err != nil {
-		return err
-	}
-	c := h.(ingester_client.IngesterClient)
-
-	req := client.WriteRequest{
-		Metadata: metadata,
+	if len(metadata) > 0 {
+		ingesterAppends.WithLabelValues(ingester.Addr, typeSamples).Inc()
```
pracucci, line 583: In the metrics tracking you've inverted `typeSamples` with `typeMetadata`. Please enrich unit tests to assert on tracked metrics. This should be spot by unit tests.

pkg/ingester/ingester.go
pstibrany: If we're going to handle samples and metadata in the same push, then I would suggest that we return errors for them separately. As it is now, errors reported by ingesters will be counted for both samples and metadata (provided both were present in the request).

pkg/distributor/distributor.go
gotjosh: Good shout, done.

pkg/ingester/ingester.go
gotjosh: Sorry Peter, I'm not following your comment. With this change, we're no longer making assumptions that a `WriteRequest` can come with one or the other.

The error handling in `send` is purely for metrics reporting. In the ingesters, we simply log when we receive metadata, so erroring out on that is not an option at the moment. 

The comment I originally left in the distributors is an operational concern, but for the code itself, it wouldn't make a difference - that is indented.

pkg/distributor/distributor.go
gotjosh: Good catch, have added the tests.

PR #2412, opened 2020-04-06
pkg/ring/kv: add List and Delete operations

### LINKED ISSUE #2382: Add ability to list and delete keys in `pkg/ring/kv`
I'm planning on using `pkg/ring/kv` to implement an API for [grafana/agent](https://github.com/grafana/agent). Two of the things I need to do that currently aren't covered by the `kv.Client` interface are:

- List keys: The list of stored objects will be available through the API, so I need to retrieve the list of keys under a given prefix.
- Delete key: Users will be able to delete stored objects through the API, so I need `kv.Client` to provide the ability to delete a key. 

I believe the only way to handle a list operation is just to add `ListKeys(ctx context.context, prefix string) ([]string, error)` to the Client interface. 

For deletion, there's two ways I can see implementing it:

1. Return a special `kv.ErrDeleteKey` error when doing a CAS, signalling to the implementation that the underlying key should be deleted.
2. Add a `Delete` method to the Client interface. 

I'm happy to implement these myself, but I wanted to open an issue first to collect feedback and thoughts on how to implement it. 

/cc @tomwilkie @gouthamve

rfratto:
This commit extends the kv interface to add List and Delete operations. These methods are not used within Cortex, but instead are meant to be utilized by downstream consumers (e.g., grafana/agent).

Note that the memberlist KV implementation does not support deletes as of this commit, but List has been implemented.

This PR also slightly tweaks the implementation of ETCD's Get operation to return no error when the key is not found. This matches the behavior of Consul and Memberlist and adds predictability to the result. A previous commit added a shared error instead, but I backpedaled on this. 

Signed-off-by: Robert Fratto <robert.fratto@grafana.com>

Closes #2382.

/cc @pracucci @pstibrany

### DIFF
--- a/integration/e2e/images/images.go
+++ b/integration/e2e/images/images.go
@@ -9,6 +9,7 @@ var (
 	Memcached        = "memcached:1.6.1"
 	Minio            = "minio/minio:RELEASE.2019-12-30T05-45-39Z"
 	Consul           = "consul:0.9"
+	ETCD             = "gcr.io/etcd-development/etcd:v3.4.7"
 	DynamoDB         = "amazon/dynamodb-local:1.11.477"
 	BigtableEmulator = "shopify/bigtable-emulator:0.1.0"
 	Cassandra        = "rinscy/cassandra:3.11.0"
--- a/pkg/ring/kv/client.go
+++ b/pkg/ring/kv/client.go
@@ -67,16 +67,17 @@ func (cfg *Config) RegisterFlagsWithPrefix(flagsPrefix, defaultPrefix string, f 
 // It also deals with serialisation by using a Codec and having a instance of
 // the the desired type passed in to methods ala json.Unmarshal.
 type Client interface {
-	// List returns a list of keys under the given prefix. The key names will
-	// include the prefix name.
+	// List returns a list of keys under the given prefix. Returned keys will
+	// include the prefix.
 	List(ctx context.Context, prefix string) ([]string, error)
 
 	// Get a specific key.  Will use a codec to deserialise key to appropriate type.
 	// If the key does not exist, Get will return nil and no error.
 	Get(ctx context.Context, key string) (interface{}, error)
 
-	// Delete a specific key. Returns true if an object was deleted.
-	Delete(ctx context.Context, key string) (bool, error)
+	// Delete a specific key. Deletions are best-effort and no error will
+	// be returned if the key does not exist.
+	Delete(ctx context.Context, key string) error
 
 	// CAS stands for Compare-And-Swap.  Will call provided callback f with the
 	// current value of the key and allow callback to return a different value.
--- a/pkg/ring/kv/client.go
+++ b/pkg/ring/kv/client.go
@@ -72,10 +72,11 @@ type Client interface {
 	List(ctx context.Context, prefix string) ([]string, error)
 
 	// Get a specific key.  Will use a codec to deserialise key to appropriate type.
+	// If the key does not exist, Get will return nil and no error.
 	Get(ctx context.Context, key string) (interface{}, error)
 
-	// Delete a specific key.
-	Delete(ctx context.Context, key string) error
+	// Delete a specific key. Returns true if an object was deleted.
+	Delete(ctx context.Context, key string) (bool, error)
 
 	// CAS stands for Compare-And-Swap.  Will call provided callback f with the
 	// current value of the key and allow callback to return a different value.
--- a/pkg/ring/kv/client.go
+++ b/pkg/ring/kv/client.go
@@ -67,8 +67,15 @@ func (cfg *Config) RegisterFlagsWithPrefix(flagsPrefix, defaultPrefix string, f 
 // It also deals with serialisation by using a Codec and having a instance of
 // the the desired type passed in to methods ala json.Unmarshal.
 type Client interface {
+	// List returns a list of keys under the given prefix. The key names will
+	// include the prefix name.
+	List(ctx context.Context, prefix string) ([]string, error)
+
 	// Get a specific key.  Will use a codec to deserialise key to appropriate type.
 	Get(ctx context.Context, key string) (interface{}, error)
+
+	// Delete a specific key.
+	Delete(ctx context.Context, key string) error
 
 	// CAS stands for Compare-And-Swap.  Will call provided callback f with the
 	// current value of the key and allow callback to return a different value.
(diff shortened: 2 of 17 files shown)


rfratto:
I'm going to fix the test error and lint error in the morning.

pstibrany:
> This commit also slightly tweaks the implementation of consul and etcd's Get operation to return the same error when a key is not found. This is useful for consumers to be able to detect a nonexistent key rather than having to check for the specific error case of each individual implementation. ETCD always returned an error before when Getting a non-existent key, but Consul silently returned nothing.
> 
> Note that I'm not confident about where the shared error variable has been added; I just implemented it to a common package that didn't cause an import loop, but I'm happy to go for a better alternative.

Here are two other possibilities how to handle problem of cyclic dependancy:

- Add `IsNotFoundError(error) bool` method to `kv.Client` interface. (Example: https://github.com/thanos-io/thanos/blob/eedbaa27b27fae803c0668ed54af29fe28b8d06e/pkg/objstore/objstore.go#L78)
- Add marker interface. Example: https://golang.org/pkg/net/#Error

pstibrany:
> * Add marker interface. Example: https://golang.org/pkg/net/#Error (this actually has methods, but empty marker interface without methods would work in your scenario)

Ah, in Go, empty marker interface is implemented by any type ;-) This isn’t Java...

rfratto:
> Add IsNotFoundError(error) bool method to kv.Client interface.

Hmm, that's not a bad idea, but I'm not sure it's an improvement in this case. IIUC error test methods are used when the error type is wrapped or one of many instances of a struct type. Since the error here is always the same string, we wouldn't be gaining much. I'm only concerned about it going in the `codec` package, which is kind of a bad fit.

rfratto:
Another way of doing this is to make sure all the KV implementations return `nil` and no error when the key isn't found. WDYT?

pstibrany:
> Another way of doing this is to make sure all the KV implementations return `nil` and no error when the key isn't found. WDYT?

Nice and simple! 👍 Please put a comment into Client interface about it too.

rfratto:
> Nice and simple! +1 Please put a comment into Client interface about it too.

Done :slightly_smiling_face: I also changed `Delete` to do something similar: it'll return `true` if the key was deleted so we don't need to use `ErrNotFound` in that case either.

integration/e2e/images/images.go
```diff
@@ -9,6 +9,7 @@ var (
 	Memcached        = "memcached:1.6.1"
 	Minio            = "minio/minio:RELEASE.2019-12-30T05-45-39Z"
 	Consul           = "consul:0.9"
+	ETCD             = "gcr.io/etcd-development/etcd:v3.4.7"
```
pracucci, line 12: As mentioned on the top of this file, please add this image to the preloading done in `.circleci/config.yml`

pkg/ring/kv/client.go
```diff
@@ -67,9 +67,17 @@ func (cfg *Config) RegisterFlagsWithPrefix(flagsPrefix, defaultPrefix string, f
 // It also deals with serialisation by using a Codec and having a instance of
 // the the desired type passed in to methods ala json.Unmarshal.
 type Client interface {
+	// List returns a list of keys under the given prefix. The key names will
+	// include the prefix name.
```
pracucci, line 71:
```suggestion
	// include the prefix.
```

pkg/ring/kv/client.go
```diff
@@ -67,9 +67,17 @@ func (cfg *Config) RegisterFlagsWithPrefix(flagsPrefix, defaultPrefix string, f
...
+	List(ctx context.Context, prefix string) ([]string, error)
+
 	// Get a specific key.  Will use a codec to deserialise key to appropriate type.
+	// If the key does not exist, Get will return nil and no error.
 	Get(ctx context.Context, key string) (interface{}, error)
 
+	// Delete a specific key. Returns true if an object was deleted.
```
pracucci, line 78: Could you also document the case the key doesn't exist (what is expected to return?) just to write the expected behaviour on stone for future client implementations :)

pkg/ring/kv/client.go
```diff
@@ -67,9 +67,17 @@ func (cfg *Config) RegisterFlagsWithPrefix(flagsPrefix, defaultPrefix string, f
 // It also deals with serialisation by using a Codec and having a instance of
 // the the desired type passed in to methods ala json.Unmarshal.
 type Client interface {
+	// List returns a list of keys under the given prefix. The key names will
```
pstibrany, line 70:
```suggestion
	// List returns a list of keys under the given prefix. Returned keys will
```

pkg/ring/kv/client.go
pstibrany: Consul API doesn't provide this feedback... it returns 200 OK even if key didn't exist before DELETE. Consul client tries to handle that by first Getting the value and then using DeleteCAS, which can fail if value has changed in the meantime -- in a totally unexpected way, based on this method signature.

I would suggest to remove `bool` return value, and make Delete best-effort. If it returns no error, client can assume that key was deleted, but it cannot assume that key is no longer in the KV store, since another client might have written it again. I think it would be better than returning errors in such case (as consul client does can do right now).

rfratto: That's a good point. I'll change it to do this and do a `Get` before `Delete` from the agent side (even though it's subject to race conditions, it's unlikely to be a problem for the Agent's use case)

pkg/ring/kv/consul/client.go
```diff
@@ -47,6 +47,8 @@ type kv interface {
 	CAS(p *consul.KVPair, q *consul.WriteOptions) (bool, *consul.WriteMeta, error)
 	Get(key string, q *consul.QueryOptions) (*consul.KVPair, *consul.QueryMeta, error)
 	List(path string, q *consul.QueryOptions) (consul.KVPairs, *consul.QueryMeta, error)
+	Delete(key string, q *consul.WriteOptions) (*consul.WriteMeta, error)
+	DeleteCAS(key *consul.KVPair, q *consul.WriteOptions) (bool, *consul.WriteMeta, error)
```
pstibrany, line 51: We don't need `DeleteCAS` anymore, do we? Shall we get rid of it?

rfratto: hah, I forgot I added that. Sure.

rfratto:
@pracucci @pstibrany Thank you both for the review!

pstibrany:
> @pracucci @pstibrany Thank you both for the review!

Thank you for your work!

PR #2413, opened 2020-04-07
Introduced BasicLifecycler

pracucci:
**What this PR does**:
The upcoming `store-gateway` ([proposal](https://cortexmetrics.io/docs/proposals/blocks-storage-sharding/)) will use the ring as the foundation for the sharding.

However, the `store-gateway` ring lifecycle will be different than the ingesters one and unfortunately the current `ring.Lifecycler` logic doesn't fit (ie. we need to inject `store-gateway` custom logic between `JOINING` and `ACTIVE` state while continuing doing heatbeats, or we need a new `store-gateway` instance to always reset its state to `JOINING` even if already present within the ring).

I initially tried to change `ring.Lifecycler` to also add `store-gateway` use cases but I ended up with many hacks I wasn't comfortable with. This lead me to approach a different solution: build a ring lifecycler foundation (`BasicLifecycler`) which we can use to build higher level lifecyclers.

Currently, this change doesn't touch the existing `ring.Lifecycler` at all and `BasicLifecycler` is not a drop-in replacement for it, but `BasicLifecycler` is a first step in this direction. On the other side, `BasicLifecycler` allows to easily build a custom lifecycler for the `store-gateway` (few lines of code) and may be used to build other ring lifecyclers too (ie. compactor).

The `BasicLifecycler` contract is simple:
- Delegate any decision to a set of delegated functions (provided in input)
- Never automatically change the instance state (it's the delegate responsability deciding when the state should change or which tokens should be used to join the ring)

I've also provided a couple of built-in delegates which can be chained together:
- Load/store tokens to disk
- Switch state to `LEAVING` on lifecycler stop

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/basic_lifecycler.go
+++ b/pkg/ring/basic_lifecycler.go
@@ -1,0 +1,474 @@
+package ring
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/cortexproject/cortex/pkg/ring/kv"
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/services"
+)
+
+type BasicLifecyclerDelegate interface {
+	// OnRingInstanceRegister is called while the lifecycler is registering the
+	// instance within the ring and should return the state and set of tokens to
+	// use for the instance itself.
+	OnRingInstanceRegister(lifecycler *BasicLifecycler, ringDesc Desc, instanceExists bool, instanceID string, instanceDesc IngesterDesc) (IngesterState, Tokens)
+
+	// OnRingInstanceTokens is called once the instance tokens are set and are
+	// stable within the ring (honoring the observe period, if set).
+	OnRingInstanceTokens(lifecycler *BasicLifecycler, tokens Tokens)
+
+	// OnRingInstanceStopping is called while the lifecycler is stopping. The lifecycler
+	// will continue to hearbeat the ring the this function is executing and will proceed
+	// to unregister the instance from the ring only after this function has returned.
+	OnRingInstanceStopping(lifecycler *BasicLifecycler)
+}
+
+type OnRingInstanceRegister func(lifecycler *BasicLifecycler, ringDesc Desc, instanceExists bool, instanceID string, instanceDesc IngesterDesc) (IngesterState, Tokens)
+type OnRingInstanceTokensChanged func(lifecycler *BasicLifecycler, tokens Tokens)
+type OnRingInstanceStopping func(lifecycler *BasicLifecycler)
+
+type BasicLifecyclerConfig struct {
+	InstanceID          string
+	InstanceAddr        string
+	InstancePort        int
+	InstanceZone        string
+	KVStore             kv.Config
+	HeartbeatPeriod     time.Duration
+	TokensObservePeriod time.Duration
+	NumTokens           int
+}
+
+// BasicLifecycler is a basic ring lifecycler which allows to hook custom
+// logic at different stages of the lifecycle. This lifecycler should be
+// used to build higher level lifecyclers.
+//
+// Contract:
+// 1. This lifecycler never change the instance state. It's the delegate
+//    responsibility to ChangeState().
+type BasicLifecycler struct {
+	*services.BasicService
+
+	cfg      BasicLifecyclerConfig
+	logger   log.Logger
+	store    kv.Client
+	delegate BasicLifecyclerDelegate
+	metrics  *BasicLifecyclerMetrics
+
+	// Channel used to execute logic within the lifecycler loop.
+	actorChan chan func()
+
+	// These values are initialised at startup, and never change
+	instanceID   string
+	instanceAddr string
+	instanceZone string
+	ringName     string
+	ringKey      string
+
+	// The current instance state.
+	currState        sync.RWMutex
+	currInstanceDesc *IngesterDesc
+}
+
+// NewBasicLifecycler makes a new BasicLifecycler.
+func NewBasicLifecycler(cfg BasicLifecyclerConfig, ringName, ringKey string, delegate BasicLifecyclerDelegate, logger log.Logger, reg prometheus.Registerer) (*BasicLifecycler, error) {
+	store, err := kv.NewClient(cfg.KVStore, GetCodec())
+	if err != nil {
+		return nil, errors.Wrap(err, "create KV store client")
+	}
+
+	return NewBasicLifecyclerWithStoreClient(cfg, ringName, ringKey, store, delegate, logger, reg)
+}
+
+// NewBasicLifecycler makes a new BasicLifecycler.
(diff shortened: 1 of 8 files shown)


pkg/ring/basic_lifecycler.go
```diff
@@ -0,0 +1,474 @@
...
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/cortexproject/cortex/pkg/ring/kv"
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/services"
+)
+
+type BasicLifecyclerDelegate interface {
```
pstibrany, line 20: Some naming nitpicking... `OnRingInstance` prefix can perhaps be reduced to `On`?

pkg/ring/basic_lifecycler.go
```diff
@@ -0,0 +1,474 @@
...
+
+	// OnRingInstanceStopping is called while the lifecycler is stopping. The lifecycler
+	// will continue to hearbeat the ring the this function is executing and will proceed
+	// to unregister the instance from the ring only after this function has returned.
+	OnRingInstanceStopping(lifecycler *BasicLifecycler)
+}
+
+type OnRingInstanceRegister func(lifecycler *BasicLifecycler, ringDesc Desc, instanceExists bool, instanceID string, instanceDesc IngesterDesc) (IngesterState, Tokens)
+type OnRingInstanceTokensChanged func(lifecycler *BasicLifecycler, tokens Tokens)
+type OnRingInstanceStopping func(lifecycler *BasicLifecycler)
```
pstibrany, line 38: This is only used in `mockDelegate`. Is it needed? (Also `OnRingInstanceTokens` vs `OnRingInstanceTokensChanged` is inconsistent)

pkg/ring/basic_lifecycler.go
```diff
@@ -0,0 +1,474 @@
...
+
+type BasicLifecyclerConfig struct {
+	InstanceID          string
+	InstanceAddr        string
+	InstancePort        int
```
pstibrany, line 43: Would it make sense to put addr and port together into a field? Lifecycler doesn't care about port specifically, and only writes this to ring.

Also `Instance` prefix seems excessive.

pkg/ring/basic_lifecycler.go
```diff
@@ -0,0 +1,474 @@
...
+	NumTokens           int
+}
+
+// BasicLifecycler is a basic ring lifecycler which allows to hook custom
+// logic at different stages of the lifecycle. This lifecycler should be
+// used to build higher level lifecyclers.
+//
+// Contract:
+// 1. This lifecycler never change the instance state. It's the delegate
+//    responsibility to ChangeState().
```
pstibrany, line 57: There is no 2.?

pkg/ring/basic_lifecycler.go
```diff
@@ -0,0 +1,474 @@
...
+	delegate BasicLifecyclerDelegate
+	metrics  *BasicLifecyclerMetrics
+
+	// Channel used to execute logic within the lifecycler loop.
+	actorChan chan func()
+
+	// These values are initialised at startup, and never change
+	instanceID   string
+	instanceAddr string
+	instanceZone string
```
pstibrany, line 73: Since this lifecycler has a copy of cfg, it doesn't need to make a copy into these fields.

pkg/ring/basic_lifecycler.go
```diff
@@ -0,0 +1,474 @@
...
+type OnRingInstanceRegister func(lifecycler *BasicLifecycler, ringDesc Desc, instanceExists bool, instanceID string, instanceDesc IngesterDesc) (IngesterState, Tokens)
+type OnRingInstanceTokensChanged func(lifecycler *BasicLifecycler, tokens Tokens)
+type OnRingInstanceStopping func(lifecycler *BasicLifecycler)
+
+type BasicLifecyclerConfig struct {
+	InstanceID          string
+	InstanceAddr        string
+	InstancePort        int
+	InstanceZone        string
+	KVStore             kv.Config
```
pstibrany, line 45: Do you think KVStore can be removed from lifecycler config? I think Lifecycler should be passed a KV store, and not create one on its own, but maybe it would complicate planned usage. (Right now, only `NewBasicLifecyclerWithStoreClient` is used though)

pkg/ring/basic_lifecycler.go
```diff
@@ -0,0 +1,474 @@
...
+		// or the instance has been forgotten. In this case, we do re-insert it.
+		if !ok {
+			level.Warn(l.logger).Log("msg", "instance missing in the ring, adding it back", "ring", l.ringName)
+			instanceDesc = ringDesc.AddIngester(l.instanceID, l.instanceAddr, l.instanceZone, l.GetTokens(), l.GetState())
+		}
+
+		changed := update(*ringDesc, &instanceDesc)
+		if !changed {
+			return nil, false, nil
+		}
```
pstibrany, line 403: If `update` didn't do a change, but there was no entry before, we should still proceed and write updated ring to the store.

pkg/ring/basic_lifecycler_delegates.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (d *TokensPersistencyDelegate) OnRingInstanceRegister(lifecycler *BasicLifecycler, ringDesc Desc, instanceExists bool, instanceID string, instanceDesc IngesterDesc) (IngesterState, Tokens) {
+	// Skip if no path has been configured.
+	if d.tokensPath == "" {
+		level.Info(d.logger).Log("msg", "not loading tokens from file, tokens file path is empty")
+		return d.next.OnRingInstanceRegister(lifecycler, ringDesc, instanceExists, instanceID, instanceDesc)
+	}
+
+	// Do not load tokens from disk if the instance is already in the ring.
+	if instanceExists {
```
pstibrany, line 63: It may exist, but not have tokens.

pkg/ring/basic_lifecycler.go
pracucci: It's quite verbose right now. The reason why I added the prefix `OnRingInstance` is because then you can add it directly to a service struct (ie. `StoreGateway`) with no ambiguity. I'm open to reconsider it. Thoughts?

pkg/ring/basic_lifecycler.go
pracucci: You're right. It's a relict from a previous version.

pkg/ring/basic_lifecycler.go
pracucci: Agree that addr and port can be merged together. I've also removed the `Instance` prefix even if I usually prefer clarity vs compacted options.

pkg/ring/basic_lifecycler.go
pracucci: Creating a KVStore client is very easy (few lines of code) and I agree we should always pick it in input to make code that uses `BasicLifecycler` easier to test.

pkg/ring/basic_lifecycler.go
pracucci: Thanks for spotting it! I fixed it and added a test to cover it.

pkg/ring/basic_lifecycler_delegates.go
pracucci: Right! Fixed and covered with a test. The previous implementation was inherited by `ring.Lifecycler` (which doesn't handle this case) but I think what you suggest is better.

pkg/ring/basic_lifecycler.go
pstibrany: I don't see problem with shorter names, but this works too.

PR #2416, opened 2020-04-07
Enable WAL checkpoint by default

codesome:
There is no mechanism to delete the old WAL yet if checkpoint is disabled. It should be clearly stated for `-ingester.checkpoint-enabled` so that it is not set to `false` for production or extended dev usage.

### DIFF
--- a/docs/configuration/arguments.md
+++ b/docs/configuration/arguments.md
@@ -343,9 +343,6 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
 
    Setting this to `true` enables writing to WAL during ingestion.
 
-- `-ingester.checkpoint-enabled`
-   Set this to `true` to enable checkpointing of in-memory chunks to disk. This is optional which helps in speeding up the replay process.
-
 - `-ingester.checkpoint-duration` 
    This is the interval at which checkpoints should be created.
 
--- a/docs/production/ingesters-with-wal.md
+++ b/docs/production/ingesters-with-wal.md
@@ -16,9 +16,8 @@ _The WAL is currently considered experimental._
 1. Since ingesters need to have the same persistent volume across restarts/rollout, all the ingesters should be run on [statefulset](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) with fixed volumes.
 
 2. Following flags needs to be set
-    * `--ingester.wal-dir` to the directory where the WAL data should be stores and/or recovered from. Note that this should be on the mounted volume.
     * `--ingester.wal-enabled` to `true` which enables writing to WAL during ingestion.
-    * `--ingester.checkpoint-enabled` to `true` to enable checkpointing of in-memory chunks to disk. This is optional which helps in speeding up the replay process.
+    * `--ingester.wal-dir` to the directory where the WAL data should be stores and/or recovered from. Note that this should be on the mounted volume.
     * `--ingester.checkpoint-duration` to the interval at which checkpoints should be created. Default is `30m`, and depending on the number of series, it can be brought down to `15m` if there are less series per ingester (say 1M).
     * `--ingester.recover-from-wal` to `true` to recover data from an existing WAL. The data is recovered even if WAL is disabled and this is set to `true`. The WAL dir needs to be set for this.
         * If you are going to enable WAL, it is advisable to always set this to `true`.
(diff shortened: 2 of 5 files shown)


docs/configuration/arguments.md
```diff
@@ -343,15 +343,16 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
...
-
 - `-ingester.checkpoint-duration` 
    This is the interval at which checkpoints should be created.
 
 - `-ingester.recover-from-wal`
    Set this to `true` to recover data from an existing WAL. The data is recovered even if WAL is disabled and this is set to `true`. The WAL dir needs to be set for this.
 
+- `-ingester.checkpoint-enabled`
+   Set this to `true` to enable checkpointing of in-memory chunks to disk. This is optional which helps in speeding up the replay process.
+   (**Note:** Don't set it to `false` for anything other than small tests. There is no mechanism to delete the old WAL yet if checkpoint is disabled.).
```
pracucci, line 354: If setting it to `false` is discouraged, why is the default `false`? Shouldn't we change the default to `true`?

docs/production/ingesters-with-wal.md
```diff
@@ -16,13 +16,14 @@ _The WAL is currently considered experimental._
...
     * `--ingester.wal-enabled` to `true` which enables writing to WAL during ingestion.
-    * `--ingester.checkpoint-enabled` to `true` to enable checkpointing of in-memory chunks to disk. This is optional which helps in speeding up the replay process.
+    * `--ingester.wal-dir` to the directory where the WAL data should be stores and/or recovered from. Note that this should be on the mounted volume.
     * `--ingester.checkpoint-duration` to the interval at which checkpoints should be created. Default is `30m`, and depending on the number of series, it can be brought down to `15m` if there are less series per ingester (say 1M).
     * `--ingester.recover-from-wal` to `true` to recover data from an existing WAL. The data is recovered even if WAL is disabled and this is set to `true`. The WAL dir needs to be set for this.
         * If you are going to enable WAL, it is advisable to always set this to `true`.
     * `--ingester.tokens-file-path` should be set to the filepath where the tokens should be stored. Note that this should be on the mounted volume. Why this is required is described below.
+    Not to be changed for production but only for small testing:
```
pracucci, line 25: Adding a sentence before the next list item doesn't correctly render in the website. Please move this information to the `* --ingester.checkpoint-enabled` list item.

docs/configuration/arguments.md
codesome: Good point! Checkpoint should be enabled. I will make that change. But, does that count as a breaking change? :( (well, it's experimental, and not a big breaking change)

codesome: Updated the default.

docs/production/ingesters-with-wal.md
codesome: Remove this line as it was not required, and modified the below line too.

docs/production/ingesters-with-wal.md
```diff
@@ -16,13 +16,13 @@ _The WAL is currently considered experimental._
...
+    * `--ingester.wal-dir` to the directory where the WAL data should be stores and/or recovered from. Note that this should be on the mounted volume.
     * `--ingester.checkpoint-duration` to the interval at which checkpoints should be created. Default is `30m`, and depending on the number of series, it can be brought down to `15m` if there are less series per ingester (say 1M).
     * `--ingester.recover-from-wal` to `true` to recover data from an existing WAL. The data is recovered even if WAL is disabled and this is set to `true`. The WAL dir needs to be set for this.
         * If you are going to enable WAL, it is advisable to always set this to `true`.
     * `--ingester.tokens-file-path` should be set to the filepath where the tokens should be stored. Note that this should be on the mounted volume. Why this is required is described below.
+    * `--ingester.checkpoint-enabled` to `true` to enable checkpointing of in-memory chunks to disk (`true` by default). There is no mechanism to delete the old WAL yet if checkpoint is disabled, so it should always be enabled unless for some (small) testing.
```
pracucci, line 25: Given now it's the default, can't we remove this list entry at all?

docs/configuration/arguments.md
```diff
@@ -343,15 +343,16 @@ It also talks to a KVStore and has it's own copies of the same flags used by the
...
-   Set this to `true` to enable checkpointing of in-memory chunks to disk. This is optional which helps in speeding up the replay process.
-
 - `-ingester.checkpoint-duration` 
    This is the interval at which checkpoints should be created.
 
 - `-ingester.recover-from-wal`
    Set this to `true` to recover data from an existing WAL. The data is recovered even if WAL is disabled and this is set to `true`. The WAL dir needs to be set for this.
 
+- `-ingester.checkpoint-enabled`
+   Set this to `true` to enable checkpointing of in-memory chunks to disk. This is optional which helps in speeding up the replay process.
```
pracucci, line 353: Now that `true` is the default, reading the doc looks a bit weird. Maybe we can rephrase it explaining when should be set to false? If there's no good reason to set it to false, why don't we remove the `-ingester.checkpoint-enabled` from this page at all?

docs/production/ingesters-with-wal.md
codesome: Makes sense for the production doc.

docs/configuration/arguments.md
codesome: I don't mind removing it from the docs. I will update the flag description to communicate when to set it to false.

PR #2418, opened 2020-04-07
Small logic simplification on the query-frontend

joe-elliott:
**What this PR does**:
Small simplification to the logic in `getNextRequest` on the query frontend.  Maps are already returned in a random order when ranged over.  Removed code that attempted to pick a random spot in the map of queues to reduce code complexity.

It should be noted that the current code already relies upon the map range behavior in go.  Otherwise the first queue in the map would be underserviced.

### DIFF
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
 	"net/http"
 	"net/url"
 	"path"
@@ -359,13 +358,7 @@ FindQueue:
 		return nil, err
 	}
 
-	i, n := 0, rand.Intn(len(f.queues))
 	for userID, queue := range f.queues {
-		if i < n {
-			i++
-			continue
-		}
-
 		/*
 		  We want to dequeue the next unexpired request from the chosen tenant queue.
 		  The chance of choosing a particular tenant for dequeueing is (1/active_tenants).
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"path"
@@ -358,7 +359,17 @@ FindQueue:
 		return nil, err
 	}
 
-	for userID, queue := range f.queues {
+	keys := make([]string, 0, len(f.queues))
+	for k := range f.queues {
+		keys = append(keys, k)
+	}
+	rand.Shuffle(len(keys), func(i, j int) { keys[i], keys[j] = keys[j], keys[i] })
+
+	for _, userID := range keys {
+		queue, ok := f.queues[userID]
+		if !ok {
+			continue
+		}
 		/*
 		  We want to dequeue the next unexpired request from the chosen tenant queue.
 		  The chance of choosing a particular tenant for dequeueing is (1/active_tenants).
--- a/pkg/querier/frontend/queue_test.go
+++ b/pkg/querier/frontend/queue_test.go
@@ -100,14 +100,20 @@ func BenchmarkGetNextRequest(b *testing.B) {
 	frontends := make([]*Frontend, 0, b.N)
 
 	for n := 0; n < b.N; n++ {
-		f, _ := setupFrontend(config)
+		f, err := setupFrontend(config)
+		if err != nil {
+			b.Fatal(err)
+		}
 
 		for i := 0; i < config.MaxOutstandingPerTenant; i++ {
 			for j := 0; j < numTenants; j++ {
 				userID := strconv.Itoa(j)
 				ctx := user.InjectOrgID(context.Background(), userID)
 
-				_ = f.queueRequest(ctx, testReq(ctx))
+				err = f.queueRequest(ctx, testReq(ctx))
+				if err != nil {
+					b.Fatal(err)
+				}
 			}
 		}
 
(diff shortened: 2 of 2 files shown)


joe-elliott:
If that's true then we should probably explicitly randomly order the map and then iterate through it.  The current code already has an expectation that the map is random.  Will adjust my PR.

pracucci:
> The current code already has an expectation that the map is random

Could you point me to the code where this expectation is? I didn't get it at first glance.

joe-elliott:
The existing code chooses a random spot to begin iterating through the map and then proceeds to the end of the map until it finds a valid request to forward to a querier.  If the order of a map is not random then those tenants that tend to be toward the beginning of the map will be underserved.  Consider a tenant that always fell on index 0 it would only have a chance of being served if the random value generated were 0.  Whereas a tenant that always fell on the last index would have a chance of being served regardless of the random value generated (if no other queues had valid requests).

However, upon further review, I recognize that the impact of this will be less than I had initially considered.  Primarily b/c empty channels are removed:
```
			if len(queue) == 0 {
				delete(f.queues, userID)
			}
```

I _do_ think explicit randomization of the keys would be a better approach.

pracucci:
> However, upon further review, I recognize that the impact of this will be less than I had initially considered. Primarily b/c empty channels are removed:

Exactly. I think this is the reason why what you actually get with the current code is basically a real randomisation.

> I do think explicit randomization of the keys would be a better approach.

I may agree could be better from a code readability (and I'm always in favour of it), but I'm wondering if it could also be better from a correctness stand point. I still can't see the real difference of the randomness.

joe-elliott:
If you have two users "user0" and "user1".  "user0" has 1 expired request and "user1" has 1 valid request.

If the map ranges in this order: user0 -> user1
Then user1 would always get their request serviced, regardless of the random value generated.

If the map ranges in this order:  user1 -> user0
Then user1 would only get their request serviced if the random value generated were 0.  If the random value generated were 1 then `getNextRequest()` would complete both loops, hit `goto FindQueue` and start over.  

So in a situation with a large number of cancelled requests (perhaps due to a bogged down backend) the level of service given to clients will be dependent on the randomized order of the map range.  Those tenants toward the end of the map will receive better service than those at the beginning.

When I first looked at this I thought the issue were more pronounced, but I recognize now it only is an issue if there happen to be a large number of cancelled requests.

joe-elliott:
Added a benchmark and a straightforward approach to key randomization.  I do not think this performance hit is worth it.  There is probably a better approach though. 

Original:
```
$ go test -bench=. -benchmem
BenchmarkGetNextRequest-8   	   15484	     78696 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/querier/frontend	5.470s

$ go test -bench=. -benchmem
BenchmarkGetNextRequest-8   	   15493	     76862 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/querier/frontend	5.431s

$ go test -bench=. -benchmem
BenchmarkGetNextRequest-8   	   15121	     76683 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/querier/frontend	5.116s
```

Branch With Key Randomization:
```
$ go test -bench=. -benchmem
BenchmarkGetNextRequest-8   	    6982	    177956 ns/op	   53155 B/op	     100 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/querier/frontend	3.749s

$ go test -bench=. -benchmem
BenchmarkGetNextRequest-8   	    6488	    177739 ns/op	   53177 B/op	     100 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/querier/frontend	3.283s

$ go test -bench=. -benchmem
pkg: github.com/cortexproject/cortex/pkg/querier/frontend
BenchmarkGetNextRequest-8   	    6513	    177573 ns/op	   53203 B/op	     100 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/querier/frontend	3.397s
```

pracucci:
> If you have two users "user0" and "user1". "user0" has 1 expired request and "user1" has 1 valid request.

Yes, but having expired requests looks more an edge case (in my experience it's not that frequent, even if we correctly have to handle it).

That being said the current implementation is tricky, so I'm in favour of simplifying it for readability.

pkg/querier/frontend/queue_test.go
```diff
@@ -84,3 +89,36 @@ func TestDequeuesExpiredRequests(t *testing.T) {
...
+	var config Config
+	flagext.DefaultValues(&config)
+	config.MaxOutstandingPerTenant = 2
+
+	const numTenants = 50
+
+	frontends := make([]*Frontend, 0, b.N)
+
+	for n := 0; n < b.N; n++ {
+		f, _ := setupFrontend(config)
```
codesome, line 103: Maybe lets not ignore any errors here (or anywhere in benchmark)?

joe-elliott: Good point.  Resolved.

pkg/querier/frontend/frontend.go
```diff
@@ -359,13 +359,17 @@ FindQueue:
...
 
-	i, n := 0, rand.Intn(len(f.queues))
-	for userID, queue := range f.queues {
-		if i < n {
-			i++
+	keys := make([]string, 0, len(f.queues))
+	for k := range f.queues {
+		keys = append(keys, k)
+	}
+	rand.Shuffle(len(keys), func(i, j int) { keys[i], keys[j] = keys[j], keys[i] })
```
jtlisi, line 366: Can you add a quick comment explaining why we do a random sort here.

joe-elliott: This comment already exists at the top of the function:

https://github.com/cortexproject/cortex/blob/d18a3b6b1e4cd3dc99edd6793265380698b3418c/pkg/querier/frontend/frontend.go#L347

This feels redundant, but I can add a comment if you'd like.

jtlisi: In that case all good!

PR #2419, opened 2020-04-07
Use snappy compression in caches (#2380)

### LINKED ISSUE #2380: Make all the caching components use snappy
We have an easy way to add [snappy](https://en.wikipedia.org/wiki/Snappy_(compression)) compression to caches, and [we do it for the index-read-cache](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/storage/caching_index_client.go#L58). But we should add it to other caches, specially the frontend-results cache.

We had it in the frontend-cache, but we lost it in this refactoring by mistake: https://github.com/cortexproject/cortex/pull/1437

RajatVaryani:
**What this PR does**:
It makes all the cache components to use snappy compression.

Fixes #2380 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/chunk_store_utils.go
+++ b/pkg/chunk/chunk_store_utils.go
@@ -102,7 +102,7 @@ type decodeResponse struct {
 func NewChunkFetcher(cacher cache.Cache, cacheStubs bool, storage Client) (*Fetcher, error) {
 	c := &Fetcher{
 		storage:        storage,
-		cache:          cacher,
+		cache:          cache.NewSnappy(cacher),
 		cacheStubs:     cacheStubs,
 		decodeRequests: make(chan decodeRequest),
 	}
--- a/pkg/querier/queryrange/results_cache.go
+++ b/pkg/querier/queryrange/results_cache.go
@@ -123,13 +123,13 @@ func NewResultsCacheMiddleware(
 			logger:    logger,
 			cfg:       cfg,
 			next:      next,
-			cache:     c,
+			cache:     cache.NewSnappy(c),
 			limits:    limits,
 			merger:    merger,
 			extractor: extractor,
 			splitter:  splitter,
 		}
-	}), c, nil
+	}), cache.NewSnappy(c), nil
 }
 
 func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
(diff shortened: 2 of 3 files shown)


pstibrany:
Have you measured any benefits from this? What about costs in terms of increased CPU usage and increased latency on read path?

RajatVaryani:
@pstibrany I am a new contributor and unfortunately, I did not. The front-end results already used snappy compression but it got removed in refactoring. I think we should not benchmark frontend results cache assuming it was an informed decision. @gouthamve please correct me. I am not sure how to benchmark increased CPU usage and latency.

pstibrany:
@RajatVaryani Thank you for your contribution! I didn't mean my comment to be harsh (and it was), I was just curious if this is indeed useful improvement.

@gouthamve Especially in case of chunks, does it really help them to snappy-compress? And if so, why don't we snappy-compress them before uploading to the storage?

pkg/chunk/chunk_store_utils.go
```diff
@@ -102,7 +102,7 @@ type decodeResponse struct {
 func NewChunkFetcher(cacher cache.Cache, cacheStubs bool, storage Client) (*Fetcher, error) {
 	c := &Fetcher{
 		storage:        storage,
-		cache:          cacher,
+		cache:          cache.NewSnappy(cacher),
```
gouthamve, line 105: This likely doesn't require compression.

pkg/querier/queryrange/results_cache.go
```diff
@@ -123,13 +123,13 @@ func NewResultsCacheMiddleware(
 			logger:    logger,
 			cfg:       cfg,
 			next:      next,
-			cache:     c,
+			cache:     cache.NewSnappy(c),
```
gouthamve, line 126: Can we do `c = cache.NewSnappy(c)` right before the return?

pkg/chunk/chunk_store_utils.go
RajatVaryani: @gouthamve What are the parameters to take into consideration when a cache requires compression or not?

gouthamve: Well the idea is that the chunks are already compressed using our own custom encoding format and compression on top of that will give us little gains.

jtlisi: IMO only the results cache would really benefit from this.

PR #2421, opened 2020-04-07
Decoupled BucketStores from UserStore

pracucci:
**What this PR does**:
The upcoming `store-gateway` ([proposal](https://cortexmetrics.io/docs/proposals/blocks-storage-sharding/)) will run the Thanos `BucketStore` instead of the querier.

Currently, we do have a multi-tenant `BucketStore` wrapper which is `UserStore`. The `UserStore` is both a wrapper over `BucketStore` and a service which does an initial blocks synching + a periodic sync.

The `store-gateway` will need to trigger the sync differently due to the sharding. In particular, the `store-gateway` needs to coordinate the initial sync while the instance is in the `JOINING` state within the ring (because it needs to have the tokens first) and to trigger a sync both periodically and when the token ranges assigned to instance itself change (ring topology changed).

Instead of duplicating code between the querier and store-gateway for the co-existence period, in this PR I'm decoupling the multi-tenant wrapper of `BucketStore` from `UserStore`:
- `BucketStores`: multi-tenant wrapper over `BucketStore` (will be directly used by the store-gateway). The `Series()` API here matches the exact gRPC service signature of the upcoming store-gateway so that we can do a pass-through to the Thanos `BucketStore`
- `UserStore`: `BucketStores` + service. I haven't renamed it in this PR to keep the diff easier. Also `UserStore` will disappear once the store-gateway will be complete.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/querier/block_store.go
```diff
@@ -268,14 +217,15 @@ func (u *UserStore) getOrCreateStore(userID string) (*store.BucketStore, error)
...
 		fetcherReg,
-		[]block.MetadataFilter{
-			// List of filters to apply (order matters).
+		// The input filters MUST be before the ones we create here (order matters).
+		append(u.filters, []block.MetadataFilter{
 			block.NewConsistencyDelayMetaFilter(userLogger, u.cfg.BucketStore.ConsistencyDelay, fetcherReg),
 			block.NewIgnoreDeletionMarkFilter(userLogger, userBkt, u.cfg.BucketStore.IgnoreDeletionMarksDelay),
 			// Filters out duplicate blocks that can be formed from two or more overlapping
 			// blocks that fully submatches the source blocks of the older blocks.
+			// TODO(pracucci) can this cause troubles with the upcoming blocks sharding in the store-gateway?
```
pstibrany, line 226: I don't think it will. I think deduplicate filter removes old blocks if new blocks cover it completely, but we better verify it.

pkg/querier/blocks_user_store.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/thanos-io/thanos/pkg/runutil"
+	"github.com/thanos-io/thanos/pkg/store/storepb"
+	"github.com/weaveworks/common/logging"
+	"google.golang.org/grpc/metadata"
+
+	"github.com/cortexproject/cortex/pkg/storage/tsdb"
+	"github.com/cortexproject/cortex/pkg/util/services"
+)
+
+type UserStore struct {
```
pstibrany, line 23: Can we add short comment with description here?

pkg/querier/block_store.go
pracucci: I think we may have troubles with the consistency check, at least as I did design it. I will write down different scenarios in tests and let's see (once we'll reach the point to build the consistency check).

pracucci:
> Unrelated to this PR, but perhaps worth fixing... should function used in Iter call react on context being done?

Definitely.

pracucci:
> Feel free to rename, it's in its own file now anyway, diff will not be much bigger I think.

As agreed offline, renamed `UserStore` to `BucketStoresService`.

PR #2423, opened 2020-04-07
Configurable delay on rule evaluation

### LINKED ISSUE #1723: Ruler doesn't take account of delays in incoming data
With the move to WAL-watching, it became possible for a sending Prometheus to get several minutes behind then catch up.  However ruler always runs its queries as-of "now" so you get a gap and/or corrupted results until the sender catches up.

Conceptually, Ruler should fire each time a set of samples is available, as-of that time.  However I believe Prometheus remote write code doesn't attempt to send breadth-first so that won't be easy.

jtlisi:
**What this PR does**:

This PR gives the operator the option to add an evaluation delay interval to the Ruler. This can be configured to ensure the ruler does not evaluate rules before the data required for the evaluation of those rules is written to Cortex.

Fixes #1723

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -700,10 +700,10 @@ The `ruler_config` configures the Cortex ruler.
 # CLI flag: -ruler.evaluation-interval
 [evaluation_interval: <duration> | default = 1m0s]
 
-# Set interval to delay the evaluation of rules to ensure they underlying
-# metrics have been pushed to cortex. Default
-# CLI flag: -ruler.evaluation-delay
-[evaluation_delay: <duration> | default = 0s]
+# Duration to delay the evaluation of rules to ensure they underlying metrics
+# have been pushed to cortex.
+# CLI flag: -ruler.evaluation-delay-duration
+[evaluation_delay_duration: <duration> | default = 0s]
 
 # How frequently to poll for rule changes
 # CLI flag: -ruler.poll-interval
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -700,6 +700,11 @@ The `ruler_config` configures the Cortex ruler.
 # CLI flag: -ruler.evaluation-interval
 [evaluation_interval: <duration> | default = 1m0s]
 
+# Set interval to delay the evaluation of rules to ensure they underlying
+# metrics have been pushed to cortex. Default
+# CLI flag: -ruler.evaluation-delay
+[evaluation_delay: <duration> | default = 0s]
+
 # How frequently to poll for rule changes
 # CLI flag: -ruler.poll-interval
 [poll_interval: <duration> | default = 1m0s]
--- a/pkg/ruler/compat.go
+++ b/pkg/ruler/compat.go
@@ -2,8 +2,12 @@ package ruler
 
 import (
 	"context"
+	"time"
 
+	"github.com/pkg/errors"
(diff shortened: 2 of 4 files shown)


pkg/ruler/ruler.go
```diff
@@ -103,6 +106,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.ExternalURL.URL, _ = url.Parse("") // Must be non-nil
 	f.Var(&cfg.ExternalURL, "ruler.external.url", "URL of alerts return path.")
 	f.DurationVar(&cfg.EvaluationInterval, "ruler.evaluation-interval", 1*time.Minute, "How frequently to evaluate rules")
+	f.DurationVar(&cfg.EvaluationDelay, "ruler.evaluation-delay", 0, "Set interval to delay the evaluation of rules to ensure they underlying metrics have been pushed to cortex. Default ")
```
pracucci, line 109:
```suggestion
	f.DurationVar(&cfg.EvaluationDelay, "ruler.evaluation-delay", 0, "Set interval to delay the evaluation of rules to ensure they underlying metrics have been pushed to cortex. ")
```

pkg/ruler/compat.go
```diff
@@ -78,3 +82,32 @@ func (t *tsdb) StartTime() (int64, error) {
 func (t *tsdb) Close() error {
 	return nil
 }
+
+// engineQueryFunc returns a new query function that executes instant queries against
+// the given engine, after subtracting the provided delay from the instant query timestamp.
+// It converts scalar into vector results.
+// Based on https://github.com/prometheus/prometheus/blob/ecda6013edf58bf645c6661b9f78ccce03b1f315/rules/manager.go#L162-L187
+func engineQueryFunc(engine *promql.Engine, q storage.Queryable, delay time.Duration) rules.QueryFunc {
```
pracucci, line 90: This whole function is a copy of `promRules.EngineQueryFunc()` just to add `t = t.Add(-delay)` as first thing. Why don't we simply wrap `promRules.EngineQueryFunc()` instead of duplicating it? Something like (pseudo-code):

```
func engineQueryFunc(engine *promql.Engine, q storage.Queryable, delay time.Duration) rules.QueryFunc {
orig := promRules.EngineQueryFunc(engine, queryable)

return func(ctx context.Context, qs string, t time.Time) (promql.Vector, error) {
  return orig(ctx, qs, t.Add(-delay))
}
}
```

docs/configuration/config-file-reference.md
```diff
@@ -700,6 +700,11 @@ The `ruler_config` configures the Cortex ruler.
 # CLI flag: -ruler.evaluation-interval
 [evaluation_interval: <duration> | default = 1m0s]
 
+# Set interval to delay the evaluation of rules to ensure they underlying
+# metrics have been pushed to cortex. Default
+# CLI flag: -ruler.evaluation-delay
+[evaluation_delay: <duration> | default = 0s]
```
gouthamve, line 706: `evaluation_delay_time`/`evaluation_delay_duration`?

pkg/ruler/compat.go
gouthamve: +1

jtlisi: Will do.

docs/configuration/config-file-reference.md
jtlisi: `evaluation_delay_duration` makes sense

PR #2425, opened 2020-04-07
New Helm Chart (cortexproject#2424)

### LINKED ISSUE #2424: New production-grade Helm Chart for Cortex
Since 1.0 version of Cortex was released, I've planned to deploy Cortex to our Kubernetes on-prem clusters in order to collect and store Prometheus metrics in the same fashion we collect and store logs in Cassandra with Grafana Loki & Promtail. However, the chart from the other cortexproject repo (https://github.com/cortexproject/cortex-helm-chart) looks more like a proof of concept than flexible production-grade chart we can see in Helm Hub. On top of that, it wasn't updated for 10 months.

I've created the Cortex Helm chart from scratch, based on the provided Cortex documentation and production practices, including running all Cortex components as a separate Kubernetes Pods, managed by the separate Deployments and StatefulSets (for ingesters). The chart from cortexproject repo and Grafana Loki's chart also proved to be helpful - Loki uses Cortex internally and hence has very similar configuration. 

I will create a pull request and add the chart to the Cortex repo - I've tested the chart on our testing environment and it seems to be working properly.

Konpon96:
**What this PR does**:
Minimalist but fully functioning and extensible Cortex Helm Chart, created from scratch and utilizing the production-grade requirements like splitting individual Cortex components to separate instances that communicate with each other as described in the [architecture section of the documentation](https://cortexmetrics.io/docs/architecture/). 

I've checked it on my test cluster with Cassandra for index and chunk storage and memberlist for exchanging ingester ring. I know memberlist is considered experimental, but in case of this chart it seems to work better than the external Consul instance - during my tests, after simulating cluster crash, the leftovers in Consul made it impossible for ingesters to restart. Thanks to Kubernetes' StatefulSet and the fact each ingester has their own FQDN, passed to each instance through `-memberlist.join` CLI flag, their reconciliation through gossip is fast and reliable. What's more, the memberlist allowed me to create Chart without unnecessary dependencies. If somebody has external Consul instance deployed, it can be easily utilized by changing the chart values, without modifying the templates.

Feedback would be much appreciated.

Fixes #2424 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


gouthamve:
Hi @Konpon96, thanks for doing this but unfortunately we have a helm chart here: https://github.com/cortexproject/cortex-helm-chart. Its very outdated, but we'd like only one helm chart to exist. I am going to kickstart a conversation in the community regarding the helm charts again.

There is also a possibility of generating helm-charts from jsonnet that needs to be explored. Given several options, I am not going to merge this right away but will kickstart the discussion around the helm chart sometime this week and will keep you looped in that discussion.

kubicgruenfeld:
This would be a big improvement to the current helm chart in my opinion. Although I see some issues arising, depending on the configuration used you might want a statefulset for the queriers for example.

Konpon96:
@gouthamve I thought about making a commit to the Helm chart repo, but because I basically recreated it from scratch, I thought of committing the new chart in the same fashion it is placed in Grafana Loki's repository. Also, the second repository is basically dead so it's easier to kickstart the discussion about the Helm chart in the core Cortex repository. I can make a pull request into the chart repository instead and close that one here. However, it would mean replacing everything there - the question I have if something is planned to be done with the chart which is already there or replacing it with the one I made is okay.

@kubicgruenfeld You mean the use case where the experimental blocks storage is used and queriers [store the index headers in the local disk](https://cortexmetrics.io/docs/operations/blocks-storage/#the-read-path)? I could refactor the chart to use StatefulSet instead of Deployment - in choosing if to use former or the latter I used the [services description](https://cortexmetrics.io/docs/architecture/#services) which says which components are stateful or stateless. It won't do anything bad and will extend the chart use cases - thank you for the suggestion!

hairyhenderson:
(see also https://github.com/cortexproject/cortex-helm-chart/issues/22 and https://github.com/cortexproject/cortex-helm-chart/issues/25)

khaines:
This PR should be moved to https://github.com/cortexproject/cortex-helm-chart.

PR #2428, opened 2020-04-08
Don't checkpoint when shutting down

codesome:
As checkpointing is spread over the checkpoint duration, we don't need to start another checkpointing at the end as one checkpoint would be just over.

Opened as draft as I have to see if the additional replay time that it incurs is less than checkpointing itself (if not, checkpointing at the end is better).

### DIFF
--- a/pkg/ingester/wal.go
+++ b/pkg/ingester/wal.go
@@ -172,10 +172,6 @@ func (w *walWrapper) run() {
 			level.Info(util.Logger).Log("msg", "checkpoint done", "time", elapsed.String())
 			w.checkpointDuration.Observe(elapsed.Seconds())
 		case <-w.quit:
-			level.Info(util.Logger).Log("msg", "creating checkpoint before shutdown")
-			if err := w.performCheckpoint(true); err != nil {
-				level.Error(util.Logger).Log("msg", "error checkpointing series during shutdown", "err", err)
-			}
 			return
 		}
 	}


gouthamve:
This looks like a great idea mainly because, we're constantly checkpointing anyways.

codesome:
The main reason we are checkpointing is to reduce the replay time. When we do a checkpoint we advance the WAL segments, so checkpointing at the end makes the amount of WAL to be replayed to 0. If we do not checkpoint at the end, it makes up for non-0 WAL to replay as the previous checkpoint started a little earlier. I am testing it out now and will update on the restart timings.

codesome:
I have some numbers.

For checkpointing during exit: ~10s for 600-700k series (should have been ~500k, but because of some deployment there was a spillover). And the replay took less than 10s in general (~4-7s for checkpoint replay, <1-2s for WAL segments). Hence in total, < 20s for the entire process.

Without checkpointing during exit: 0s for the additional checkpoint when exiting. ~26s for replay, where ~5-6s was for checkpoint replay and 20s for the WAL replay. In this case, I tried to pick the near to worst case of exiting when the on-going checkpointing was near to over (12m elapsed of the 15-16m checkpointing).

With the above number, I guess we should have checkpointing at the end. Writing whole chunks is faster than replaying individual samples :)

gouthamve:
Hrm, how long does the checkpoint itself take?

codesome:
I guess I was not clear :)
`~10s for 600-700k` --> ~10s for checkpointing

pracucci:
@codesome Is this still relevant?

codesome:
I think we can close this as from the tests checkpont+replay takes less time on average than not checkpointing at the end (that's how slow WAL replay is).

PR #2432, opened 2020-04-08
Fix List in mock consul

rfratto:
Using List on the inmemory KV store always silently failed because the Consul WaitTime is always zero, which the mock implementation treated as already past the deadline.

Signed-off-by: Robert Fratto <robert.fratto@grafana.com>

/cc @pstibrany


pstibrany:
Unit test for `List` should probably go to `pkg/ring/kv/kv_test.go`.

PR #2433, opened 2020-04-09
Define store-gateway protobuf and client

pracucci:
**What this PR does**:
In this PR I'm doing a step forward towards integrating the first version of the blocks storage `store-gateway` service ([proposal](https://cortexmetrics.io/docs/proposals/blocks-storage-sharding/)). In particular:

1. Defined the protobuf for the `store-gateway` service, keeping it 100% compatible with the Thanos store (at the current stage won't be possible to build the consistency check because we lack the list of queried blocks in the response, but the consistency check will be added in a second phase and I would like to propose to expose blocks info in Thanos too)
2. Implemented a `store-gateway` client (that will be used by the querier) following the same approach we took for the ingester client (since the `store-gateway` client will use the gRPC clients pool too)

Notes to the reviewer:
- The `store-gateway` protobuf definition is in a separate package to avoid circular dependencies between `querier` and `store-gateway` (the querier needs the `store-gateway` protobuf and, for a while, the `store-gateway` needs the querier's `BucketStores`)

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/querier/store_gateway_client.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health/grpc_health_v1"
+
+	"github.com/cortexproject/cortex/pkg/ring/client"
+	"github.com/cortexproject/cortex/pkg/storegateway/storegatewaypb"
+	"github.com/cortexproject/cortex/pkg/util/grpcclient"
+)
+
+func NewStoreGatewayClientFactory(cfg grpcclient.Config, reg prometheus.Registerer) client.PoolFactory {
```
jtlisi, line 15: This doesn't look to be called anywhere? I assume it will be added in a consecutive PR.

pracucci: Exactly. I'm still in the process of submitting partial PRs in preparation of the final PR where I will submit the store-gateway and blocks sharding logic, otherwise the final one will be too huge to review.

PR #2435, opened 2020-04-09
Remove additional step, the issue is closed?

gouthamve:
Was going through the circle config and realised that we're running lint because of an issue that has since been closed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,10 +72,6 @@ jobs:
     - run:
         name: Lint
         command: make BUILD_IN_CONTAINER=false lint
-    # fails to run everything first time - see https://github.com/golangci/golangci-lint/issues/866
-    - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -72,10 +72,6 @@ jobs:
     - run:
         name: Lint
         command: make BUILD_IN_CONTAINER=false lint
-    # fails to run everything first time - see https://github.com/golangci/golangci-lint/issues/866
```
pracucci, line 75: We're currently running `golangci-lint` `v1.21.0` (see `build-image/Dockerfile`) which doesn't include the fix. You should upgrade to the latest `v1.24.0` first, build and publish a new `build-image` (replacing it across `.circleci/config.yml` - there are multiple references) before merging this PR.

PR #2436, opened 2020-04-09
Add type header to WAL/Checkpoint records and use TSDB records

### LINKED ISSUE #2426: Ingester WAL records need a type header
Currently, the records are written directly to the WAL without any headers this makes it difficult to change the record safely for any future extensions.

I propose to add a single byte header to the records to determine it's type just like Prometheus (before we remove the experimental tag on the WAL). This could be a breaking change, but we can try to support header + no header for 2 releases and ideas are welcome for that.

codesome:
**What this PR does**:

Add a single-byte type header to WAL and Checkpoint records and way to handle old records concurrently. Additionally, replace existing proto records in WAL with Prometheus TSDB records. 

Fixes #2426

**Checklist**
- [x] Tests updated
- [ ] Documentation added


codesome:
PR is ready for review

pkg/ingester/wal.go
```diff
@@ -563,12 +585,15 @@ func processCheckpoint(name string, userStates *userStates, params walRecoveryPa
 Loop:
 	for reader.Next() {
 		s := seriesPool.Get().(*Series)
-		if err := proto.Unmarshal(reader.Record(), s); err != nil {
+		m, err := decodeRecord(reader.Record(), CheckpointRecordType1, s)
```
gouthamve, line 588: So what would happen if the checkpoint is w/o a record? From what I can see it returns an error.

pkg/ingester/wal.go
```diff
@@ -745,12 +770,14 @@ Loop:
 			break Loop
 		default:
 		}
-		if err := proto.Unmarshal(reader.Record(), record); err != nil {
+		m, err := decodeRecord(reader.Record(), WALRecordType1, record)
```
gouthamve, line 773: Same here, an error is returned so basically ignore this?

codesome:
After some offline discussion with Goutham, we decided to add the Prometheus TSDB records in this PR itself to not have many versions or records, as the current record would be obsolete immediately after TSDB records are in.

codesome:
I wrote a benchmark for WAL replay (will open PR after https://github.com/cortexproject/cortex/pull/2465 is merged as it depends on that) to compare the new records, and the results look good. Largely depends on Checkpoint/WAL ratio. The benchmark had equal number of samples in WAL and Checkpoint.
```
benchmark                    old ns/op      new ns/op      delta
BenchmarkWALReplay/#00-8     2027607295     1656442045     -18.31%

benchmark                    old allocs     new allocs     delta
BenchmarkWALReplay/#00-8     3050719        1142134        -62.56%

benchmark                    old bytes      new bytes      delta
BenchmarkWALReplay/#00-8     5943738312     5638734152     -5.13%
```

pkg/ingester/wal.go
```diff
@@ -121,6 +152,14 @@ func newWAL(cfg WALConfig, userStatesFunc func() map[string]*userState, register
 		Help:       "Time taken to create a checkpoint.",
 		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 	})
+	w.walLoggedBytes = promauto.With(registerer).NewCounter(prometheus.CounterOpts{
+		Name: "cortex_ingester_wal_logged_bytes_total",
```
pstibrany, line 156: Name doesn't suggest any failure, like Help does. Same for next metric. In fact, it's not used for failures.

pkg/ingester/wal.go
```diff
@@ -49,15 +51,36 @@ func (cfg *WALConfig) RegisterFlags(f *flag.FlagSet) {
...
+	// will produce an error if the first byte is less than 7 (thus we know its not the old record).
+	// The old record will be removed in the future releases, hence the record type should not cross
+	// '7' till then.
+	// Relevant code to verify this is https://github.com/cortexproject/cortex/blob/90516410b9e1fe2fe3ae9041260c02d2a6f0468d/pkg/ingester/wal.pb.go#L894-L917
+	// (with error at line 916).
+
+	// WALRecordSeriesType1 is the type for the WAL record on Prometheus TSDB record for series.
+	WALRecordSeriesType1 RecordType = 1
+	// WALRecordSamplesType1 is the type for the WAL record based on Prometheus TSDB record for samples.
+	WALRecordSamplesType1 RecordType = 2
```
pstibrany, line 74: What is `Type1` suffix?

pkg/ingester/wal.go
```diff
@@ -240,7 +298,7 @@ func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
 	if err := os.MkdirAll(checkpointDirTemp, 0777); err != nil {
 		return errors.Wrap(err, "create checkpoint dir")
 	}
-	checkpoint, err := wal.New(nil, nil, checkpointDirTemp, true)
+	checkpoint, err := wal.New(nil, nil, checkpointDirTemp, false)
```
pstibrany, line 301: No need for compression anymore?

pkg/ingester/wal.go
```diff
@@ -267,10 +325,11 @@ func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
 	}
 
 	var wireChunkBuf []client.Chunk
+	b := make([]byte, 0, 1024)
 	for userID, state := range us {
 		for pair := range state.fpToSeries.iter() {
 			state.fpLocker.Lock(pair.fp)
-			wireChunkBuf, err = w.checkpointSeries(checkpoint, userID, pair.fp, pair.series, wireChunkBuf)
+			wireChunkBuf, b, err = w.checkpointSeries(checkpoint, userID, pair.fp, pair.series, wireChunkBuf, b)
```
pstibrany, line 332: `b` is unused after being returned. Is the idea to use bytes pool here?

pkg/ingester/wal.go
```diff
@@ -936,3 +1060,109 @@ func SegmentRange(dir string) (int, int, error) {
...
+	var enc tsdb_record.Encoder
+	// The 'encoded' already has the type header and userID here, hence re-using
+	// the remaining part of the slice (i.e. encoded[len(encoded):]))to encode the samples.
+	encoded := buf.Get()
+	encoded = append(encoded, enc.Samples(record.Samples, encoded[len(encoded):])...)
+
+	return encoded
+}
+
+func decodeWALRecord(b []byte, rec *Record, walRec *WALRecord) (_ *Record, _ *WALRecord, err error) {
```
pstibrany, line 1128: Why does this function return `*Record` and `*WALRecord` when it simply returns input parameters for these return values?

pkg/ingester/wal.go
```diff
@@ -774,39 +868,70 @@ Loop:
...
-					buf.userID = record.UserId
+					buf.userID = userID
 					shards[i] = buf
 				default:
 					shards[i] = &samplesWithUserID{
-						userID: record.UserId,
+						userID: userID,
 					}
 				}
 			}
```
pstibrany, line 904: This seems like a lot of work for what's basically reusing of `samples` slice. It seems to me that pool would work equally well.

pkg/ingester/wal.go
```diff
@@ -49,15 +51,36 @@ func (cfg *WALConfig) RegisterFlags(f *flag.FlagSet) {
...
+type RecordType byte
+
+const (
+	// Currently we also support the old records without a type header.
+	// For that, we assume the record type does not cross 7 as the proto unmarshalling
+	// will produce an error if the first byte is less than 7 (thus we know its not the old record).
+	// The old record will be removed in the future releases, hence the record type should not cross
+	// '7' till then.
+	// Relevant code to verify this is https://github.com/cortexproject/cortex/blob/90516410b9e1fe2fe3ae9041260c02d2a6f0468d/pkg/ingester/wal.pb.go#L894-L917
+	// (with error at line 916).
```
pstibrany, line 69: > Protobuf messages never start with a byte less than 8 (see [wire format](https://developers.google.com/protocol-buffers/docs/encoding#structure)).

That's all explanation that needs to be here. Linking to single specific decoder seems weird.

pkg/ingester/wal.go
codesome: Bad copy-paste! That said, I made a separate PR for that https://github.com/cortexproject/cortex/pull/2497 as I was not sure when would this be merged.

pkg/ingester/wal.go
codesome: I just meant to say first one in the series of records for samples. Similarly, if we plan to change that record, it would be `Type2` (with the description in comments).  What other name do you suggest?

pkg/ingester/wal.go
codesome: We could, but that causes additional CPU usage. We can turn it on later if very necessary (TSDB supports co-existence of compressed and uncompressed records, so it is fine to move around).

pkg/ingester/wal.go
codesome: You are right, this is not needed

pkg/ingester/wal.go
codesome: I copied this logic as it is from Prometheus TSDB, and would like to keep any improvements to another PR. One reason I could think is that having a channel per worker reduces contention, else effectively a common pool would act as a bottleneck.

pkg/ingester/wal.go
codesome: Will remove it :P

pkg/ingester/wal.go
pstibrany: I would just remove `Type1` until we need `Type2` :)

pstibrany: (I mean the suffix in the name)

pkg/ingester/wal.go
pstibrany: Is this worth mentioning in changelog?

pkg/ingester/wal.go
pstibrany: Ah, of course. 🤦

pstibrany:
Tests are failing here now.

pkg/ingester/wal.go
codesome: Will add

pkg/ingester/wal_test.go
```diff
@@ -1,4 +1,4 @@
-package ingester
+ package ingester
```
pstibrany, line 1: extra space here? :)

codesome: Fixed!

Issue #2262, opened 2020-03-12
Migrate object storage clients to use thanos objstore.Bucket implementations


jtlisi:
Currently we have generic `ObjectClient` interface defined here: 
https://github.com/cortexproject/cortex/blob/58c081a8eab77b70ce4d9f425fd2400d1155eee8/pkg/chunk/storage_client.go#L64-L71

This interface is used by the `pkg/chunk/objectclient` to store chunks. It will also be used by @sandeepsukhani works on delete. I also utilize this interface for my rules/alerts branches. These clients are pretty generic and I think we might be able to utilize some off the shelf libraries instead of rolling our own object client. I found two potential libraries that might work:

1. Go Cloud Development Kit

https://github.com/google/go-cloud

Keeps things very generic and easy for reuse.

2. Reuse Thanos
The Thanos libary is pretty full featured and would be able to do most of what we want. It would also be nice for unifying the two ecosystems.
https://github.com/thanos-io/thanos/blob/master/pkg/objstore/objstore.go

pracucci:
> Reuse Thanos

Unless we have a good argument to not use the Thanos objstore I would use it. We're already using it for the blocks storage anyway, so it's already vendored.

jtlisi:
@pracucci  I agree. I added some tags to signal to anyone who wants to take up this issue.

jaybatra26:
@jtlisi and @pracucci Can I take this up as my next work item?

jtlisi:
Go for it!

jaybatra26:
Thanks @jtlisi . I am assuming that I would have to write something like this [code snippet](https://github.com/cortexproject/cortex/blob/82eae45bc7f71e89c8fcf5a4551683b842b3e6bb/pkg/storage/tsdb/backend/gcs/bucket_client.go#L13-L17) for each of the object store i.e(s3/gcs/azure etc). 
Also, looks like local file system is in testing mode.
![image](https://user-images.githubusercontent.com/24604293/76728544-03c47100-677d-11ea-81b9-f7054757eedd.png)

Is that correct?

jtlisi:
[...]

PR #2440, opened 2020-04-10
Adds object storage client for Swift OpenStack

cyriltovena:
This adds a new storage client to support swift OpenStack. This is a popular client for companies that runs their own internal cloud on their hardware.

Right now there is no tests, I'm looking at the e2e package and add some tests for it using this docker container https://github.com/jeantil/openstack-swift-keystone-docker WDYT ? I have no other way of testing this.

This seems to be a popular request for Loki.

I still need to update the changelog too, but I'm tired going to bed.

**Checklist**
- [ ] Tests updated
- [x ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * [CHANGE] Experimental TSDB: Modified default values for `compactor.deletion-delay` option from 48h to 12h and `-experimental.tsdb.bucket-store.ignore-deletion-marks-delay` from 24h to 6h. #2414
 * [CHANGE] Experimental WAL: Default value of `-ingester.checkpoint-enabled` changed to `true`. #2416
 * [FEATURE] Ruler: The `-ruler.evaluation-delay` flag was added to allow users to configure a default evaluation delay for all rules in cortex. The default value is 0 which is the current behavior. #2423
+* [FEATURE] Experimental: Added a new object storage client for OpenStack Swift. #2440
 * [ENHANCEMENT] Experimental TSDB: sample ingestion errors are now reported via existing `cortex_discarded_samples_total` metric. #2370
 * [ENHANCEMENT] Failures on samples at distributors and ingesters return the first validation error as opposed to the last. #2383
 * [ENHANCEMENT] Experimental TSDB: Added `cortex_querier_blocks_meta_synced`, which reflects current state of synced blocks over all tenants. #2392
--- a/integration/chunks_storage_backends_test.go
+++ b/integration/chunks_storage_backends_test.go
@@ -3,16 +3,35 @@
 package main
 
 import (
+	"context"
+	"fmt"
+	"sort"
 	"testing"
 	"time"
 
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	thanos "github.com/thanos-io/thanos/pkg/objstore/swift"
+	"github.com/weaveworks/common/user"
 
 	"github.com/cortexproject/cortex/integration/e2e"
 	e2edb "github.com/cortexproject/cortex/integration/e2e/db"
 	"github.com/cortexproject/cortex/integration/e2ecortex"
+	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/chunk/encoding"
+	"github.com/cortexproject/cortex/pkg/chunk/openstack"
+	"github.com/cortexproject/cortex/pkg/chunk/storage"
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/ruler"
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
+	"github.com/cortexproject/cortex/pkg/util/validation"
+)
+
+var (
+	userID = "e2e-user"
 )
 
 func TestChunksStorageAllIndexBackends(t *testing.T) {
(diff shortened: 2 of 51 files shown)


pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,207 @@
...
+	cfg.RegisterFlagsWithPrefix("", f)
+}
+
+// RegisterFlagsWithPrefix registers flags with prefix.
+func (cfg *SwitfConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.StringVar(&cfg.ContainerName, prefix+"swift.container-name", "cortex", "Name of the Swift container to put chunks in.")
+	f.StringVar(&cfg.Domain, prefix+"swift.domain", "", "Openstack user's domain name.")
+	f.StringVar(&cfg.DomainId, prefix+"swift.domain-id", "", "Openstack user's domain id.")
+	f.StringVar(&cfg.UserName, prefix+"swift.username", "", "Openstack username for the api.")
+	f.StringVar(&cfg.UserId, prefix+"swift.userid", "", "Openstack userid for the api.")
```
pracucci, line 60: Just to keep it consistent with the YAML:

```suggestion
	f.StringVar(&cfg.UserName, prefix+"swift.user-name", "", "Openstack username for the api.")
	f.StringVar(&cfg.UserId, prefix+"swift.user-id", "", "Openstack userid for the api.")
```

pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,207 @@
...
+	ApiKey                      string        `yaml:"api_key"`               // Key for api access
+	ApplicationCredentialId     string        `yaml:"app_credential_id"`     // Application Credential ID
+	ApplicationCredentialName   string        `yaml:"app_crendential_name"`  // Application Credential Name
+	ApplicationCredentialSecret string        `yaml:"app_credential_secret"` // Application Credential Secret
+	AuthUrl                     string        `yaml:"auth_url"`              // Auth URL
+	Retries                     int           `yaml:"retries"`               // Retries on error (default is 3)
+	ConnectTimeout              time.Duration `yaml:"connect_timeout"`       // Connect channel timeout (default 10s)
+	Timeout                     time.Duration `yaml:"timeout"`               // Data channel timeout (default 60s)
+	Region                      string        `yaml:"region"`                // Region to use eg "LON", "ORD" - default is use first region (v2,v3 auth only)
+	AuthVersion                 int           `yaml:"auth"`                  // Set to 1, 2 or 3 or leave at 0 for autodetect
```
pracucci, line 39: To keep it consistent with the CLI flag:

```suggestion
	AuthVersion                 int           `yaml:"auth_version"`          // Set to 1, 2 or 3 or leave at 0 for autodetect
```

pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,207 @@
...
+	util.WarnExperimentalUse("OpenStack Swift Storage")
+
+	if cfg.EndpointType != "" &&
+		cfg.EndpointType != string(swift.EndpointTypePublic) &&
+		cfg.EndpointType != string(swift.EndpointTypeAdmin) &&
+		cfg.EndpointType != string(swift.EndpointTypeInternal) {
+		return nil, fmt.Errorf(
+			"unexpected endpoint type %s must be one of %s, %s or %s",
+			cfg.EndpointType, swift.EndpointTypePublic, swift.EndpointTypeAdmin, swift.EndpointTypeInternal)
+	}
```
pracucci, line 91: As pattern, we're currently transitioning towards validating the config in a `cfg.Validate() error` function called by the parent config. I navigate back the (2) places where `SwiftConfig` is referenced and should be easy to call. As a reference you can see this:
https://github.com/cortexproject/cortex/blob/9451bc694c572212f3eefaa535427f30d2ac5468/pkg/chunk/storage/factory.go#L91-L100

cyriltovena: The rule store doesn't have a validate. I'll add it.

pracucci: Yes exactly. Just add the missing `Validate()` going back the parent, but should be trivial.

CHANGELOG.md
```diff
@@ -17,7 +18,7 @@
 * [CHANGE] Experimental WAL: Default value of `-ingester.checkpoint-enabled` changed to `true`. #2416
 * [FEATURE] Ruler: The `-ruler.evaluation-delay` flag was added to allow users to configure a default evaluation delay for all rules in cortex. The default value is 0 which is the current behavior. #2423
 * [ENHANCEMENT] Experimental TSDB: sample ingestion errors are now reported via existing `cortex_discarded_samples_total` metric. #2370
-* [ENHANCEMENT] Failures on samples at distributors and ingesters return the first validation error as opposed to the last. #2383 
+* [ENHANCEMENT] Failures on samples at distributors and ingesters return the first validation error as opposed to the last. #2383
```
cyriltovena, line 21: Sorry I have a linter for markdown, Hopefully this is ok.

cyriltovena:
Alright I'll tackle the E2E later this weekend.

CHANGELOG.md
pracucci: Yes, no worries. I think it's time for us to lint spurious whitespaces too in markdown so we solve the problem at the root.

cyriltovena:
The integration tests timeout on go get, I might need help to understand how to overcome this.

gouthamve:
Before we move forward with this, I think @jtlisi has some ideas for reusing the Thanos codebase here: https://github.com/cortexproject/cortex/issues/2262

If we move there we will automatically add support for Swift as well. But I am not too familiar with it, @jtlisi could you comment?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [FEATURE] Experimental: Added a new object storage client for OpenStack Swift.
```
pracucci, line 5: A couple of minor things:
- Please move the `[FEATURE]` to the group of `[FEATURE]` below
- Remember to add the PR number at the end of the sentence (ie. `#2440`) cause we have a script trying to match PRs with CHANGELOG entries when we release a new Cortex version

integration/e2e/util.go
```diff
@@ -53,11 +53,22 @@ func BuildArgs(flags map[string]string) []string {
 	return args
 }
 
-func GetRequest(url string) (*http.Response, error) {
+func GetRequest(url string, h http.Header) (*http.Response, error) {
```
pracucci, line 56: Is the input header used? I can't find where we use it.

integration/e2e/db/db.go
```diff
@@ -109,3 +109,14 @@ func NewCassandra() *e2e.HTTPService {
 		9042,
 	)
 }
+
+func NewSwiftStorage() *e2e.HTTPService {
+	return e2e.NewHTTPService(
+		"swift",
+		images.SwiftEmulator,
+		nil,
+		// readiness probe inspired from https://github.com/kubernetes/examples/blob/b86c9d50be45eaf5ce74dee7159ce38b0e149d38/cassandra/image/files/ready-probe.sh
```
pracucci, line 118: Please remove this comment, since doesn't apply.

integration/e2e/db/db.go
```diff
@@ -109,3 +109,14 @@ func NewCassandra() *e2e.HTTPService {
...
+		e2e.NewHTTPReadinessProbe(8080, "/", 404, 404),
```
pracucci, line 119: Do we have any better way to test for readiness? We also support command probes.

integration/chunks_storage_backends_test.go
```diff
@@ -114,3 +131,187 @@ func TestChunksStorageAllIndexBackends(t *testing.T) {
 	assertServiceMetricsPrefixes(t, Ingester, ingester)
 	assertServiceMetricsPrefixes(t, Querier, querier)
 }
+
+func TestSwiftChunkStorage(t *testing.T) {
```
pracucci, line 135: This test looks flaky. Sometimes I get this:

```
--- FAIL: TestSwiftChunkStorage (16.57s)
    chunks_storage_backends_test.go:206:
        	Error Trace:	chunks_storage_backends_test.go:206
        	Error:      	Not equal:
        	            	expected: 0x1561c881
        	            	actual  : 0xb4b81a0f
        	Test:       	TestSwiftChunkStorage
```

pracucci, line 135: Any reason why you didn't add switf to `TestChunksStorageAllIndexBackends`?

jtlisi:
@gouthamve I think it would make sense to move in that direction but no one is currently working on that. @jaybatra26 started some refactoring work to make it possible but I would say we are at a standstill at this point.

@cyriltovena My recommendation at this point would be to reuse https://github.com/thanos-io/thanos/blob/91f5960a3becdbd13c296c1e57bdbf57c329eeed/pkg/objstore/swift/swift.go#L29-L44 to ensure the switchover will be seemless when it occurs.

cyriltovena:
> Before we move forward with this, I think @jtlisi has some ideas for reusing the Thanos codebase here: #2262
> 
> If we move there we will automatically add support for Swift as well. But I am not too familiar with it, @jtlisi could you comment?

That would be great if we can have this, having Swift without having to wait for a big refactoring.

integration/e2e/util.go
cyriltovena: Yeah this is a leftover, I needed this at some point to get credentials. But I 've found another way around. I'll revert that back.

integration/chunks_storage_backends_test.go
cyriltovena: Yes like I mentioned offline. This test is mostly for index and refactoring it to test multiple object storage was more difficult than creating a new one more focused and simpler.

pracucci: Sure. As already agreed offline, I'm resolving this comment.

integration/e2e/db/db.go
cyriltovena: I feel like this is a good one for the use case, waiting for the HTTP api to be responsive. That's all we need for the e2e.

jaybatra26:
@jtlisi I am working on it. I will try to wind it up by the end of this week(hopefully).

cyriltovena:
@jtlisi I've tried to use Thanos config but it's not a one to one mapping. It seems that we use a different client. I didn't checked if Thanos had swift and choose a newer version with more identification options.

If you guys wants to pause this and wait for the Thanos change I'm fine with it. We can also get that into Loki itself /cc @slim-bean if we don't want to wait. Not sure 🤷

cyriltovena:
I can try to retrofit Thanos object store, give me some time I'll try to give more effort so that's we don't have to make a breaking change.

I have a e2e working I can rely on, don't merge yet I'll give at least 2 hour.

cyriltovena:
So I tried again and there's something missing with the Thanos Client it doesn't allow application credentials. I was able to find my way around the config by reading the long long long boring openstack documentation.

I still pushed the commit so we don't introduce API break.

We should add that to Thanos though. I think you most likely want to use application credentials.

cyriltovena:
Ready !

cyriltovena:
Ready !

pracucci:
@cyriltovena I've fixed a couple of CLI flags, rebased `master`, fixed `List()` (signature has changed in the meanwhile in `master`) and squashed all commits because rebasing was a bit a pain due to the `config-file-reference.md`.

pracucci:
> So I tried again and there's something missing with the Thanos Client it doesn't allow application credentials. I was able to find my way around the config by reading the long long long boring openstack documentation.
> 
> I still pushed the commit so we don't introduce API break.
> 
> We should add that to Thanos though. I think you most likely want to use application credentials.

I think this is not a blocker and can be addressed separately. @cyriltovena could you open an issue to Thanos, please?

pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,178 @@
...
+	f.StringVar(&cfg.UserDomainID, prefix+"swift.user-domain-id", "", "Openstack user's domain id.")
+	f.StringVar(&cfg.Username, prefix+"swift.username", "", "Openstack username for the api.")
+	f.StringVar(&cfg.UserId, prefix+"swift.user-id", "", "Openstack userid for the api.")
+	f.StringVar(&cfg.Password, prefix+"swift.password", "", "Openstack api key.")
+	f.StringVar(&cfg.AuthUrl, prefix+"swift.auth-url", "", "Openstack authentication URL.")
+	f.StringVar(&cfg.RegionName, prefix+"swift.region-name", "", "Openstack Region to use eg LON, ORD - default is use first region (v2,v3 auth only)")
+	f.StringVar(&cfg.ProjectName, prefix+"swift.project-name", "", "Openstack project name (v2,v3 auth only).")
+	f.StringVar(&cfg.ProjectID, prefix+"swift.project-id", "", "Openstack project id (v2,v3 auth only).")
+	f.StringVar(&cfg.ProjectDomainName, prefix+"swift.project-domain-name", "", "Name of the project's domain (v3 auth only), only needed if it differs from the user domain.")
+	f.StringVar(&cfg.ProjectDomainID, prefix+"swift.project-domain-id", "", "Id of the project's domain (v3 auth only), only needed if it differs the from user domain.")
```
pstibrany, line 54: I would suggest to keep order consistent with order of fields in `thanos.SwiftConfig` struct. I would also prefer to *use* it in the same order in `NewSwiftObjectClient` method, but that's trickier.

```suggestion
	f.StringVar(&cfg.AuthUrl, prefix+"swift.auth-url", "", "Openstack authentication URL.")
	f.StringVar(&cfg.Username, prefix+"swift.username", "", "Openstack username for the api.")
	f.StringVar(&cfg.UserDomainName, prefix+"swift.user-domain-name", "", "Openstack user's domain name.")
	f.StringVar(&cfg.UserDomainID, prefix+"swift.user-domain-id", "", "Openstack user's domain id.")
	f.StringVar(&cfg.UserId, prefix+"swift.user-id", "", "Openstack userid for the api.")
	f.StringVar(&cfg.Password, prefix+"swift.password", "", "Openstack api key.")
	f.StringVar(&cfg.DomainId, prefix+"swift.domain-id", "", "Openstack user's domain id.")
	f.StringVar(&cfg.DomainName, prefix+"swift.domain-name", "", "Openstack user's domain name.")
	f.StringVar(&cfg.ProjectID, prefix+"swift.project-id", "", "Openstack project id (v2,v3 auth only).")
	f.StringVar(&cfg.ProjectName, prefix+"swift.project-name", "", "Openstack project name (v2,v3 auth only).")
	f.StringVar(&cfg.ProjectDomainID, prefix+"swift.project-domain-id", "", "Id of the project's domain (v3 auth only), only needed if it differs the from user domain.")
	f.StringVar(&cfg.ProjectDomainName, prefix+"swift.project-domain-name", "", "Name of the project's domain (v3 auth only), only needed if it differs from the user domain.")
	f.StringVar(&cfg.RegionName, prefix+"swift.region-name", "", "Openstack Region to use eg LON, ORD - default is use first region (v2,v3 auth only)")
	f.StringVar(&cfg.ContainerName, prefix+"swift.container-name", "cortex", "Name of the Swift container to put chunks in.")
```

pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,178 @@
...
+
+		Region: cfg.RegionName,
+	}
+
+	switch {
+	case cfg.UserDomainName != "":
+		c.Domain = cfg.UserDomainName
+	case cfg.UserDomainID != "":
+		c.DomainId = cfg.UserDomainID
+	}
```
pstibrany, line 84: Should we reuse Thanos comment to explain what's happening here?

```
	// Support for cross-domain scoping (user in different domain than project).
	// If a userDomainName or userDomainID is given, the user is scoped to this domain.
```

pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,178 @@
...
+	c := &swift.Connection{
+		AuthUrl:  cfg.AuthUrl,
+		ApiKey:   cfg.Password,
+		UserName: cfg.Username,
+		UserId:   cfg.UserId,
+
+		TenantId:       cfg.ProjectID,
+		Tenant:         cfg.ProjectName,
+		TenantDomain:   cfg.ProjectDomainName,
+		TenantDomainId: cfg.ProjectDomainID,
```
pstibrany, line 71: Thanos seems to be putting `ProjectDomainName` and `ProjectDomainID` to `Domain` and `DomainId` fields.
Why the difference?

pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,178 @@
...
+	}, nil
+}
+
+func (s *SwiftObjectClient) Stop() {
+	s.conn.UnAuthenticate()
+}
+
+// GetObject returns a reader for the specified object key from the configured swift container. If the
+// key does not exist a generic chunk.ErrStorageObjectNotFound error is returned.
+func (s *SwiftObjectClient) GetObject(ctx context.Context, objectKey string) (io.ReadCloser, error) {
```
pstibrany, line 118: The fact that `ctx` parameter is not used in any of `GetObject`, `PutObject`, `List` or `DeleteObject` methods is concerning.

pkg/chunk/openstack/swift_object_client.go
```diff
@@ -0,0 +1,178 @@
...
+
+// RegisterFlags registers flags.
+func (cfg *SwiftConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix("", f)
+}
+
+// Validate config and returns error on failure
+func (cfg *SwiftConfig) Validate() error {
+	return nil
+}
```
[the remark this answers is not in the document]
pracucci: It was me suggesting it before the offline conversation we had (between me and you Peter). We already use `Validate()` in many places so I don't see it a problem. However, I invite you to open an issue where we can take a final decision about it and then we can submit a PR to refactor it.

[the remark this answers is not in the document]
pstibrany: My (snarky) comment here is about Validate method being empty, and not validating anything. That seems weird.

I will open issue for making decision on validation.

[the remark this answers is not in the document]
pracucci: You're right. I think it was supposed to contain the delimiter validation, which has been probably lost at some point while working on the PR.

cyriltovena:
> There are some eyebrow raising concerns about context, or lack of validation (why have it then?). > Approving since it's experimental feature and there are no guarantees that it works.

The client I choose is having a timeout by default which Thanos doesn't have. Both client doesn't support context propagation.

Validation was there before because the original config was in my opinion better than Thanos one, more granular, and I did my best to retrofit this for the like of yours. Since you wanted to have Thanos config in case one day (I'm sure it will be soon) we want to move to Thanos storage.

Now regarding the `no guarantees that it works` I did write 2 integration tests ensuring this works for the ruler and the storage where this again doesn't exist on Thanos side.

😓

cyriltovena:
Also I suggest to not rebase PR until they get merge (e.g. merge commit to get update from master) so that you can have the full history of commits which might help understand how we get there.

pstibrany:
> Now regarding the `no guarantees that it works` I did write 2 integration tests ensuring this works for the ruler and the storage where this again doesn't exist on Thanos side.

Don't get me wrong, I'm not saying it doesn't work, I'm just worried about not using context properly in those methods, so it can have different behaviour than other stores.

I think you did a great job here!

Let's merge this, we can improve it when we get feedback about it.

PR #2441, opened 2020-04-10
Create index table from local table client for boltdb 

adityacs:
**What this PR does**:
When using boltdb for index, calling create and update table doesn't have any action and this makes logs too much verbose.
Reference: https://github.com/grafana/loki/issues/1786

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/local/boltdb_table_client.go
+++ b/pkg/chunk/local/boltdb_table_client.go
@@ -36,7 +36,12 @@ func (c *TableClient) ListTables(ctx context.Context) ([]string, error) {
 }
 
 func (c *TableClient) CreateTable(ctx context.Context, desc chunk.TableDesc) error {
-	return nil
+	file, err := os.OpenFile(filepath.Join(c.directory, desc.Name), os.O_CREATE|os.O_RDONLY, 0666)
+	if err != nil {
+		return err
+	}
+
+	return file.Close()
 }
 
 func (c *TableClient) DeleteTable(ctx context.Context, name string) error {
--- a/pkg/chunk/local/boltdb_index_client_test.go
+++ b/pkg/chunk/local/boltdb_index_client_test.go
@@ -1,6 +1,8 @@
 package local
 
 import (
+	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -8,6 +10,8 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"go.etcd.io/bbolt"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
 )
 
 var (
@@ -113,4 +117,56 @@ func TestBoltDB_GetDB(t *testing.T) {
 
 	_, err = boltdbIndexClient.GetDB(testDb1, DBOperationRead)
 	require.NoError(t, err)
+}
+
+func Test_CreateTable_BoltdbRW(t *testing.T) {
+	tableName := "test"
+	dirname, err := ioutil.TempDir(os.TempDir(), "boltdb")
+	require.NoError(t, err)
+
+	indexClient, err := NewBoltDBIndexClient(BoltDBConfig{
+		Directory: dirname,
+	})
+	require.NoError(t, err)
+
+	tableClient, err := NewTableClient(dirname)
+	require.NoError(t, err)
+
+	err = tableClient.CreateTable(context.Background(), chunk.TableDesc{
+		Name: tableName,
+	})
+	require.NoError(t, err)
+
+	batch := indexClient.NewWriteBatch()
+	batch.Add(tableName, fmt.Sprintf("hash%s", "test"), []byte(fmt.Sprintf("range%s", "value")), nil)
+
+	err = indexClient.BatchWrite(context.Background(), batch)
+	require.NoError(t, err)
+
+	// try to create the same file which is already existing
+	err = tableClient.CreateTable(context.Background(), chunk.TableDesc{
+		Name: tableName,
+	})
+	require.NoError(t, err)
+
+	// make sure file content is not modified
+	entry := chunk.IndexQuery{
+		TableName: tableName,
+		HashValue: fmt.Sprintf("hash%s", "test"),
+	}
+	var have []chunk.IndexEntry
+	err = indexClient.query(context.Background(), entry, func(read chunk.ReadBatch) bool {
+		iter := read.Iterator()
+		for iter.Next() {
+			have = append(have, chunk.IndexEntry{
+				RangeValue: iter.RangeValue(),
+			})
+		}
+		return true
+	})
+	require.NoError(t, err)
+	require.Equal(t, []chunk.IndexEntry{
+		{RangeValue: []byte(fmt.Sprintf("range%s", "value"))},
+	}, have)
+
 }
--- a/pkg/chunk/local/fixtures.go
+++ b/pkg/chunk/local/fixtures.go
@@ -59,6 +59,10 @@ func (f *fixture) Clients() (
 				Prefix: "chunks",
 				Period: 10 * time.Minute,
 			},
+			IndexTables: chunk.PeriodicTableConfig{
+				Prefix: "index",
+				Period: 10 * time.Minute,
+			},
 		}},
 	}
 


pkg/chunk/table_manager.go
```diff
@@ -300,12 +300,14 @@ func (m *TableManager) SyncTables(ctx context.Context) error {
 		return err
 	}
 
-	if err := m.createTables(ctx, toCreate); err != nil {
-		return err
-	}
+	if reflect.TypeOf(m.client).String() != "*local.tableClient" {
```
adityacs, line 303: Let me know if any other better ways to handle this

pkg/chunk/table_manager.go
```diff
@@ -540,3 +543,10 @@ func ExpectTables(ctx context.Context, client TableClient, expected []TableDesc)
 
 	return nil
 }
+
+func (m *TableManager) validateClient() bool {
+	if reflect.TypeOf(m.client).String() == "*local.tableClient" {
+		return false
+	}
+	return true
+}
```
adityacs, line 552: If this implementation is OK, will add a small test for this. Let me know if we can do it in any other better way.

adityacs:
@pracucci Thanks for the review. Totally agree this is a hacky solution. We can create the directory in `CreateTable`. However, we configure the expected number of tables here https://github.com/cortexproject/cortex/blob/master/pkg/chunk/table_manager.go#L371 and we don't need all of this when we are using boltdb. 

I haven't gone through the code flow completely. I may be wrong here. When I run the code I see that boltdb directory is created with just the last entry from the list of expected tables. So, we simply loop in here https://github.com/cortexproject/cortex/blob/master/pkg/chunk/table_manager.go#L371 for tables which is not required to be created.

adityacs:
@pracucci One solution I could think of is can we change this interface to support an iterator?
If we add an iterator, we can create and update tables by checking if it has `nextValue`. If it has `nextValue`, continue, otherwise we break the loop here https://github.com/cortexproject/cortex/blob/26fa43286ddc42b3a98c89243c6ad2cba96f2f92/pkg/chunk/table_manager.go#L444

In case of boltdb we will have only one value(in case of using schema v11) to create and iterator will not have next value to create and we can break the loop.

OR

Somewhere we have to check if client is `local.tableClient` and skip including in `expectedTables`. May be even here we can skip https://github.com/cortexproject/cortex/blob/26fa43286ddc42b3a98c89243c6ad2cba96f2f92/pkg/chunk/schema_config.go#L364. However, this would still be just work around solution

WDYT?

pracucci:
I still believe this whole change should be relegated to BoltDB only. The BoltDB `CreateTable()` and `ListTables()` should be made to play nicely together. @sandeepsukhani do you have any feedback on this?

adityacs:
@sandeepsukhani Thanks for the feedback. Will make the required changes. 
I still have one query. I may be missing something here

When we calculate the expected tables from here https://github.com/cortexproject/cortex/blob/55a4fcfbc07903fd722e9c72659fd8ff9a873f84/pkg/chunk/table_manager.go#L315
won't we get more tables to be created than required for boltdb? Even if we create a file from `createTable`, we would still be looping in `createTables` for the files not required to be created

pkg/chunk/local/boltdb_table_client.go
```diff
@@ -36,6 +36,9 @@ func (c *TableClient) ListTables(ctx context.Context) ([]string, error) {
 }
 
 func (c *TableClient) CreateTable(ctx context.Context, desc chunk.TableDesc) error {
+	if _, err := os.OpenFile(filepath.Join(c.directory, desc.Name), os.O_RDWR|os.O_CREATE, 0666); err != nil {
+		return err
+	}
```
sandeepsukhani, line 41: I think this would keep the file reference open. Let us do this instead.
```suggestion
	return ioutil.WriteFile(filepath.Join(c.directory, desc.Name), nil, 0666)
```

adityacs: Perfect. Thanks @sandeepsukhani 
Other way we can call `Close()` on the file. However, let's stick to your code. Less number of lines 😄

adityacs:
@sandeepsukhani One more thing. Should we change this to `Debug` ?
https://github.com/cortexproject/cortex/blob/989553066a6868158e4d93ddd0bf86f14be7a6fa/pkg/chunk/table_manager.go#L499

sandeepsukhani:
> @sandeepsukhani One more thing. Should we change this to `Debug` ?
> https://github.com/cortexproject/cortex/blob/989553066a6868158e4d93ddd0bf86f14be7a6fa/pkg/chunk/table_manager.go#L499

Are you seeing that in the logs with the change in `CreateTable` method of boltdb table client?
I feel it should not but I might be wrong though.

adityacs:
@sandeepsukhani Yeah, I still see in the logs

sandeepsukhani:
> @sandeepsukhani Yeah, I still see in the logs

@adityacs that sounds strange. I tried it locally and I do not see that in my logs. Can you please push the changes to see what you have?
Also please share the table manager config that you tried with here in the comments.

adityacs:
I am just using the default configs

```
table_manager:
  chunk_tables_provisioning:
    inactive_read_throughput: 0
    inactive_write_throughput: 0
    provisioned_read_throughput: 0
    provisioned_write_throughput: 0
  index_tables_provisioning:
    inactive_read_throughput: 0
    inactive_write_throughput: 0
    provisioned_read_throughput: 0
    provisioned_write_throughput: 0
  retention_deletes_enabled: false
  retention_period: 0s
```

adityacs:
@sandeepsukhani I have not disabled `ThroughputUpdates` so updateTable is not getting skipped. Is that correct?

sandeepsukhani:
> @sandeepsukhani I have not disabled `ThroughputUpdates` so updateTable is not getting skipped. Is that correct?

The throughput settings are only relevant for dynamodb, for all the other clients it is a noop so I guess you need not worry about it in this PR. I did not have any of those throughput settings set while trying it and I was not seeing that line being logged for me. Also, those throughput settings are removed from all the default configs to let only dynamodb users set them when need be.

sandeepsukhani:
Let us not merge this, we should have a check in `CreateTable` to see if the file already exists. Thanks to @pracucci for pointing this out.

sandeepsukhani:
I am checking to see if we can do it atomically.

pkg/chunk/local/boltdb_table_client.go
```diff
@@ -36,7 +37,7 @@ func (c *TableClient) ListTables(ctx context.Context) ([]string, error) {
 }
 
 func (c *TableClient) CreateTable(ctx context.Context, desc chunk.TableDesc) error {
-	return nil
+	return ioutil.WriteFile(filepath.Join(c.directory, desc.Name), nil, 0666)
```
sandeepsukhani, line 40: Sorry @adityacs for all the back and forth. To avoid any data loss during race conditions let us change it to the following which should create the file if it does not exist and if it is there already, it should open the file in read-only mode.
```suggestion
	fl, err := os.OpenFile(filepath.Join(c.directory, desc.Name), os.O_CREATE|os.O_RDONLY, 0666)
	if err != nil {
		return err
	}
	
	return fl.Close()
```

adityacs: @sandeepsukhani Just one question before making this change. 

Since, we don't write any data while we create file using `ioutil.Writefile` and `ioutil.Writefile` opens the file in `WRONLY` mode and also we have the  `bbolt.Open()` in `boltdb_index_client` which in turn does a file lock when doing any Read/Write operation https://github.com/etcd-io/bbolt/blob/a8af23b57f672fef05637de531bba5aa00013364/db.go#L223

Would there be data race still?

sandeepsukhani: I think only way to tell about it for sure is to write some tests with that scenario. If we get an error here when file is already locked then we can avoid returning an error for it because it means file is already there.

adityacs: Sure. Will try to add the tests

adityacs: @sandeepsukhani You will be definitely aware of this. Still, this is what I found out.
We could have not used `ioutil.Writefile` since it truncates the file if already existing. 

Also, regarding boltdb file lock. It only does the `advisory` lock so any other process can be non cooperative and override the lock. I don't think we would run in to that situation here as the binary runs as single process. So, adding test for it is not required. 

So, we are good with the recent change you mentioned.  Have added a test to verify "trying to create an already existing file doesn't modify any content". This is not so much required to test. We can remove it.

adityacs:
@sandeepsukhani Yeah `os.OpenFile` doesn't check file locks.

PR #2443, opened 2020-04-10
Allow custom replication strategy in the Ring

pracucci:
**What this PR does**:
The upcoming `store-gateway` will need a replication strategy different than the ingesters one. In this PR I've refactored the ring, extracting the replication strategy into an interface which can be plugged via `ring.NewWithStoreClientAndStrategy()` (this function is still unused, but will be used in a subsequent PR).

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/replication_strategy.go
+++ b/pkg/ring/replication_strategy.go
@@ -2,22 +2,37 @@ package ring
 
 import (
 	"fmt"
+	"time"
 )
 
-// replicationStrategy decides, given the set of ingesters eligible for a key,
+type ReplicationStrategy interface {
+	// Filter out unhealthy instances and checks if there're enough instances
+	// for an operation to succeed. Returns an error if there are not enough
+	// instances.
+	Filter(instances []IngesterDesc, op Operation, replicationFactor int, heartbeatTimeout time.Duration) (healthy []IngesterDesc, maxFailures int, err error)
+
+	// ShouldExtendReplicaSet returns true if given an instance that's going to be
+	// added to the replica set, the replica set size should be extended by 1
+	// more instance for the given operation.
+	ShouldExtendReplicaSet(instance IngesterDesc, op Operation) bool
+}
+
+type DefaultReplicationStrategy struct{}
+
+// Filter decides, given the set of ingesters eligible for a key,
 // which ingesters you will try and write to and how many failures you will
 // tolerate.
 // - Filters out dead ingesters so the one doesn't even try to write to them.
 // - Checks there is enough ingesters for an operation to succeed.
 // The ingesters argument may be overwritten.
-func (r *Ring) replicationStrategy(ingesters []IngesterDesc, op Operation) ([]IngesterDesc, int, error) {
+func (s *DefaultReplicationStrategy) Filter(ingesters []IngesterDesc, op Operation, replicationFactor int, heartbeatTimeout time.Duration) ([]IngesterDesc, int, error) {
 	// We need a response from a quorum of ingesters, which is n/2 + 1.  In the
 	// case of a node joining/leaving, the actual replica set might be bigger
 	// than the replication factor, so use the bigger or the two.
-	replicationFactor := r.cfg.ReplicationFactor
 	if len(ingesters) > replicationFactor {
 		replicationFactor = len(ingesters)
 	}
+
 	minSuccess := (replicationFactor / 2) + 1
 	maxFailure := replicationFactor - minSuccess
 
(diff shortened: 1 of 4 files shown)


pkg/ring/replication_strategy.go
```diff
@@ -4,28 +4,43 @@ import (
 	"fmt"
 )
 
-// replicationStrategy decides, given the set of ingesters eligible for a key,
+type ReplicationStrategy interface {
+	// Filter out unhealthy instances and checks if there're enough instances
+	// for an operation to succeed. Returns an error if there are not enough
+	// instances.
+	Filter(cfg Config, instances []IngesterDesc, op Operation) (healthy []IngesterDesc, maxFailures int, err error)
```
pstibrany, line 11: Would it make sense move avoid using Config here? Filter needs two parameters from Config, not entire Config.

pkg/ring/replication_strategy.go
```diff
@@ -4,28 +4,43 @@ import (
...
+
+	// ExtendReplicaSet returns true if given an instance that's going to be
+	// added to the replica set, the replica set size should be extended by 1
+	// more instance for the given operation.
+	ExtendReplicaSet(instance IngesterDesc, op Operation) bool
```
pstibrany, line 16: This method is not extending anything, and is a check instead. Can we name it such?

pracucci: It returns whether the replica set should be extended. Could you suggest a better name, please?

pkg/ring/replication_strategy.go
pracucci: We may. Would you pass `replicationFactor` and `heartbeatTimeout` to the function?

pkg/ring/replication_strategy.go
pstibrany: `CanUseIngesterForOperation` or `IsOperationAllowedForInstance`.

pracucci: I thought about a similar naming, but it's not really what it does. We still keep the instance, but we do extend the replica set size.

pkg/ring/replication_strategy.go
pstibrany: Passing RF and heartbeat timeout sounds good to me. WDYT?

pkg/ring/replication_strategy.go
pstibrany: But not in this method. This method doesn't even take replica set as a parameter, nor does it operate on or extend one, so current name is confusing.

pkg/ring/replication_strategy.go
```diff
@@ -44,6 +59,22 @@ func (r *Ring) replicationStrategy(ingesters []IngesterDesc, op Operation) ([]In
 	return ingesters, maxFailure, nil
 }
 
+func (s *DefaultReplicationStrategy) ExtendReplicaSet(ingester IngesterDesc, op Operation) bool {
+	// We do not want to Write to Ingesters that are not ACTIVE, but we do want
+	// to write the extra replica somewhere.  So we increase the size of the set
+	// of replicas for the key. This means we have to also increase the
+	// size of the replica set for read, but we can read from Leaving ingesters,
+	// so don't skip it in this case.
+	// NB dead ingester will be filtered later (by replication_strategy.go).
```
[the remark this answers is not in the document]
pracucci: I think the entire comment is still valid. I just changed `replication_strategy.go` to `DefaultReplicationStrategy.Filter()`.

pkg/ring/replication_strategy.go
pracucci: I see your point, but I still believe something like `CanUseIngesterForOperation()` (or similar) is not correct either, because what this function tells is whether the ring should extend or not the replica set. Maybe `ShouldExtendReplicaSet()` sounds better?

pkg/ring/replication_strategy.go
pracucci: Done

pkg/ring/replication_strategy.go
pstibrany: What about `InstanceAllowedInReplicaSet` ?

pstibrany: No, that's still wrong...

pstibrany: I think `ShouldExtendReplicaSet` describes this best. After looking into the code, it actually does something different than I thought.

pstibrany: It's a weird thing to put into this interface  :-(

pracucci: > It's a weird thing to put into this interface :-(

Why? It's part of the replication strategy.

pracucci: > I think ShouldExtendReplicaSet describes this best.

Renamed in the meanwhile.

pstibrany: > Renamed in the meanwhile.
Thanks.

> Why? It's part of the replication strategy.

You're probably right. I still find our replication strategy code quite confusing, but that's just me.

Thanks for addressing my feedback, feel free to merge.

PR #2444, opened 2020-04-10
Refactor Schema to make code more explicit about interface usage

pstibrany:
While studying how schemas work, I've found some oddities, that this PR tries to address.
- some entries implementations having many methods that only returned "not supported" errors
- `seriesStore` embedding `store` to reuse some of the methods
- `Schema` having methods for supporting different stores

This PR splits `Schema` interface (and then also `entries`) into `StoreSchema` and `SeriesStoreSchema` plus shared `BaseSchema` supertype.

Similarly, `baseStore` was extracted from `store`, and `seriesStore` now uses that, instead of store directly.

This makes code more explicit about what is used where, or why some methods are not implemented in various schema / entries versions.

(I haven't reordered methods for to keep `baseStore` and `store` methods together, as that would increase diff size. It would be nice thing to do at some point. Same for `baseSchema` vs `storeSchema` and `seriesStoreSchema`)

**Checklist**
- [x] Tests updated
- [ ] Documentation added

/cc @owen-d @cyriltovena

### DIFF
--- a/pkg/chunk/schema.go
+++ b/pkg/chunk/schema.go
@@ -42,20 +42,29 @@ type hasChunksForIntervalFunc func(userID, seriesID string, from, through model.
 
 // Schema interface defines methods to calculate the hash and range keys needed
 // to write or read chunks from the external index.
-type Schema interface {
-	// When doing a write, use this method to return the list of entries you should write to.
-	GetWriteEntries(from, through model.Time, userID string, metricName string, labels labels.Labels, chunkID string) ([]IndexEntry, error)
-
-	// Should only be used with the seriesStore. TODO: Make seriesStore implement a different interface altogether.
-	// returns cache key string and []IndexEntry per bucket, matched in order
-	GetCacheKeysAndLabelWriteEntries(from, through model.Time, userID string, metricName string, labels labels.Labels, chunkID string) ([]string, [][]IndexEntry, error)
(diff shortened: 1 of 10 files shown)


pkg/chunk/schema.go
```diff
@@ -42,21 +42,30 @@ type hasChunksForIntervalFunc func(userID, seriesID string, from, through model.
 
 // Schema interface defines methods to calculate the hash and range keys needed
```
pracucci, line 43: Could you update this comment, please?

pkg/chunk/schema.go
```diff
@@ -42,21 +42,30 @@ type hasChunksForIntervalFunc func(userID, seriesID string, from, through model.
...
-type Schema interface {
+type BaseSchema interface {
+	// When doing a read, use these methods to return the list of entries you should query
+	GetReadQueriesForMetric(from, through model.Time, userID string, metricName string) ([]IndexQuery, error)
+	GetReadQueriesForMetricLabel(from, through model.Time, userID string, metricName string, labelName string) ([]IndexQuery, error)
+	GetReadQueriesForMetricLabelValue(from, through model.Time, userID string, metricName string, labelName string, labelValue string) ([]IndexQuery, error)
+	FilterReadQueries(queries []IndexQuery, shard *astmapper.ShardAnnotation) []IndexQuery
+}
+
+// Schema used by store
```
pracucci, line 53: Should be something like the following? If so, please remember to update `SeriesStoreSchema` doc too:
`StoreSchema  is the schema used by store.`

PR #2446, opened 2020-04-10
Regex set optimization when querying for chunks in index

pstibrany:
This PR implements "regex set" optimization when querying chunks in index. If label matcher is in form label=~"a|b", then we can simply query index for label=a, label=b. This reduces number of fetched chunk IDs and matching we need to do in memory. (This is similar optimization to what Prometheus has in TSDB layer)

**Checklist**
- [x] Tests updated
- [ ] Documentation added

/cc @codesome


pstibrany:
Thanks for reviews!

PR #2456, opened 2020-04-13
Total Worker Parallelism in the Querier

### LINKED ISSUE #1883: querier.worker-parallelism should be derivable
Currently, `querier.worker-parallelism` is the number of parallel queries that can be processed per querier per frontend. In an effort to simplify operating cortex, this could be changed to a max parallelism configuration per querier _irrespective_ of the number of frontends. Changing this would alleviate the common practice of determining querier parallelism -> dividing by n_frontends -> setting `querier.worker-parallelism`.

Additionally, the queriers already have a DNS watch loop running in order to add/drop ephemeral frontends, which can be used to readjust the derived per-frontend limits.

To maintain backwards compatibility, this change could be effected by adding a new flag and marking `querier.worker-parallelism` deprecated if necessary.

joe-elliott:
**What this PR does**:
- Adds a parameter `-querier.worker-total-parallelism` that controls a total number of concurrent requests allowed spread across workers.
- Adds workerFrontendManager to help encapsulate functionality related to maintaining a set of concurrent workers per frontend.
- Related tests/docs
- Extends existing frontend tests to cover both the existing and new parallelism modes

Fixes #1883 

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott:
So the primary negative to this approach is that the workers all block on this line:

joe-elliott:
Via side discussions with @tomwilkie and @pracucci the `-querier.max-concurrent` parameter has come to my attention.

This PR adds a new option `-querier.worker-total-parallelism`.  Instead of adding this new option should `-querier.max-concurrent` be used both to configure promql concurrency and total concurrent workers?  

Perhaps the option I should have added is `-querier.force-max-worker-concurrency` or something like that?

tomwilkie:
> Perhaps the option I should have added is -querier.worker-match-max-concurrency or something like that?

Yes please!

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +111,64 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
 				level.Debug(w.log).Log("msg", "adding connection", "addr", update.Addr)
-				ctx, cancel := context.WithCancel(servCtx)
-				cancels[update.Addr] = cancel
-				w.runMany(ctx, update.Addr)
+				client, err := w.connect(update.Addr)
+				if err != nil {
+					level.Error(w.log).Log("msg", "error connecting", "addr", update.Addr, "err", err)
+				}
+
+				w.managers[update.Addr] = newFrontendManager(servCtx, w.log, w.server, client, 0, w.cfg.GRPCClientConfig.MaxRecvMsgSize)
```
tomwilkie, line 119: Too many arguments to a function is a bit a smell to me.

Can you pass `w.cfg.GRPCClientConfig` and have the client create by `newFrontendManager`?

Also, probably don't need to pass 0, it can be assumed.

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +111,64 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
 	}
 }
 
-// runMany starts N runOne loops for a given address.
-func (w *worker) runMany(ctx context.Context, address string) {
-	client, err := w.connect(address)
+func (w *worker) connect(address string) (FrontendClient, error) {
+	opts := []grpc.DialOption{grpc.WithInsecure()}
+	opts = append(opts, w.cfg.GRPCClientConfig.DialOption([]grpc.UnaryClientInterceptor{middleware.ClientUserHeaderInterceptor}, nil)...)
+	conn, err := grpc.Dial(address, opts...)
```
tomwilkie, line 139: Whilst we're refactoring, can we use `grpc.DialContext`?

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +111,64 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
-		return
+		return nil, err
 	}
+	return NewFrontendClient(conn), nil
+}
 
-	w.wg.Add(w.cfg.Parallelism)
-	for i := 0; i < w.cfg.Parallelism; i++ {
-		go w.runOne(ctx, client)
+func (w *worker) resetParallelism() {
```
tomwilkie, line 146: If I read this right:
- When #frontend doesn't divide cleanly into the parallelism limit, we give the remainder as +1s to the last few.
- When #frontend > parallelism limit, we give each worker 1 - potentially exceeding the parallism limit and queueing in the queriers.

I don't see any other alternative at this stage, but I'd like to see comments and warning logs to this effect please.

pkg/querier/frontend/worker_frontend_manager.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+type frontendManager struct {
+	client FrontendClient
+	server *server.Server
+
+	log            log.Logger
+	maxSendMsgSize int
+
+	gracefulQuit      []chan struct{}
```
tomwilkie, line 36: My gut feeling here is we need a map of cancel funcs, not close channels - one per worker goroutine.  That will allow us to interrupt workers to scale down and stop more gracefully.

pkg/querier/frontend/worker.go
joe-elliott: Will definitely remove the 0.  I had intended that for something else, but obviously it fell by the wayside.

I would prefer continuing to pass in the client
- it would not reduce the number of required parameters.  I would have to pass in client address and client config instead of client and client config.
- it makes testing easier.  Currently I'm passing in a mock client.  If I pass in a config then I need to find a way to have the frontend manager create a real client that works for testing.

pkg/querier/frontend/worker_frontend_manager.go
joe-elliott: Swapped to context.CancelFunc's.  the frontend will now force cancel queries on scaling.  There's pros/cons here, but ultimately I am agreeing with this approach.

joe-elliott:
I believe all concerns have been addressed.  I did prefer to pass the FrontendClient into the manager for reasons detailed above.

Please let me know if any other changes are desired!

pkg/querier/frontend/worker.go
tomwilkie: Yeah thats a good reason to, I agree.

pkg/querier/frontend/worker.go
```diff
@@ -4,62 +4,56 @@ import (
...
-)
-
 // WorkerConfig is config for a worker.
 type WorkerConfig struct {
-	Address           string        `yaml:"frontend_address"`
-	Parallelism       int           `yaml:"parallelism"`
-	DNSLookupDuration time.Duration `yaml:"dns_lookup_duration"`
+	Address             string        `yaml:"frontend_address"`
+	Parallelism         int           `yaml:"parallelism"`
+	MatchMaxConcurrency bool          `yaml:"match_max_concurrency"`
```
tomwilkie, line 27: Can you set this to true in the single-process-config.yaml please?

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +114,80 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
+
+	totalConcurrency := 0
+	for i, addr := range addresses {
+		concurrentRequests := 0
+		if w.cfg.MatchMaxConcurrency {
+			concurrentRequests = w.querierCfg.MaxConcurrent / len(w.managers)
+
+			// if max concurrency does not evenly divide into our frontends we will choose some
+			//  to receive an extra connection.  addresses were randomized above so this will be a
+			//  random selection of frontends
```
tomwilkie, line 168: Nit: sentences (capitalisation, full stops) please.

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +114,80 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
+			concurrentRequests = w.cfg.Parallelism
 		}
 
-		if err := w.process(c); err != nil {
-			level.Error(w.log).Log("msg", "error processing requests", "err", err)
-			backoff.Wait()
-			continue
+		// max concurrency is less than the total number of query frontends.  to prevent accidentally
+		// starving a frontend we are just going to always connect once to every frontend
+		//  this is dangerous b/c we may start exceeding promql max concurrency
```
tomwilkie, line 179: Nit: sentences (capitalisation, full stops) please.

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +114,80 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
+		if concurrentRequests == 0 {
+			concurrentRequests = 1
 		}
```
tomwilkie, line 182: Nit: I feel like the logic for the parallelisation per worker would make a nice self-contained function.

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +114,80 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
-	// Build a child context so we can cancel querie when the stream is closed.
-	ctx, cancel := context.WithCancel(c.Context())
-	defer cancel()
-
-	for {
-		request, err := c.Recv()
-		if err != nil {
-			return err
+		totalConcurrency += concurrentRequests
+		if mgr, ok := w.managers[addr]; ok {
```
tomwilkie, line 185: `!ok` would be bug, right?  if so lets not hide it.

pkg/querier/frontend/worker_frontend_manager.go
```diff
@@ -0,0 +1,163 @@
...
+type frontendManager struct {
+	server    *server.Server
+	client    FrontendClient
+	clientCfg grpcclient.Config
+
+	log log.Logger
+
+	workerCancels     []context.CancelFunc
+	managerCtx        context.Context
+	managerCancel     context.CancelFunc
```
tomwilkie, line 36: Given we have all the worker cancellations, do we need a manager level one?  Seems superfluous.

pkg/querier/frontend/worker_frontend_manager.go
```diff
@@ -0,0 +1,163 @@
...
+			continue
+		}
+
+		if len(f.workerCancels) > n {
+			// remove from slice and shutdown
+			var cancel context.CancelFunc
+			cancel, f.workerCancels = f.workerCancels[0], f.workerCancels[1:]
+			cancel()
+		}
+	}
```
tomwilkie, line 84: Would this be less nested with two for loops?

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +114,90 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
-// runMany starts N runOne loops for a given address.
-func (w *worker) runMany(ctx context.Context, address string) {
-	client, err := w.connect(address)
-	if err != nil {
-		level.Error(w.log).Log("msg", "error connecting", "addr", address, "err", err)
-		return
+func (w *worker) connect(ctx context.Context, address string) (FrontendClient, error) {
+	opts := []grpc.DialOption{
+		grpc.WithInsecure(),
+		grpc.WithBlock(),
```
tomwilkie, line 142: Oh you snuck a `WithBlock` in there - why?

joe-elliott: @tomwilkie 
`DialContext()` is non blocking by default.  Adding the `WithBlock()` option makes it behave like `Dial()` in this regard.

tomwilkie: Do we want it to block?  I think we're assuming the DNS will remove unhealthy clients, which isn't necessarily the case in non-k8s envs.

joe-elliott: Reread the docs and you're right.

pkg/querier/frontend/worker.go
```diff
@@ -118,116 +114,89 @@ func (w *worker) watchDNSLoop(servCtx context.Context) error {
...
+				}
+
+				w.managers[update.Addr] = newFrontendManager(servCtx, w.log, w.server, client, w.cfg.GRPCClientConfig)
 
 			case naming.Delete:
 				level.Debug(w.log).Log("msg", "removing connection", "addr", update.Addr)
-				if cancel, ok := cancels[update.Addr]; ok {
-					cancel()
+				if mgr, ok := w.managers[update.Addr]; ok {
+					mgr.stop()
```
jtlisi, line 127: Since we only stop the manager here, it's still in the map. We use the length of the map to calculate the currency in the functions below this. Wouldn't we need to account for stopped managers?

joe-elliott: Dang, good catch.  Pushing a fix that removes the stopped manager from the map.

PR #2457, opened 2020-04-14
Distributor: /all_user_stats now show API and Rule Ingest Rate correctly

### LINKED ISSUE #2171: [bug] /all_user_stats don't show correct Rule Ingest Rate
### WHAT

/all_user_stats show Rule Ingest Rate `0` even if rules are defined.

### Version

cortex master-c63a4f2e
image: quay.io/cortexproject/cortex:master-c63a4f2e

### How to reproduce

* setup ruler
* check /all_user_stats

Wing924:

**What this PR does**:
* Distributor: `/all_user_stats` now show API and Rule Ingest Rate correctly.

Fixes #2171

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -360,6 +360,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		return nil, err
 	}
 
+	source := req.Source
 	var firstPartialErr error
 	removeReplica := false
 
@@ -536,7 +537,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		if sp := opentracing.SpanFromContext(ctx); sp != nil {
 			localCtx = opentracing.ContextWithSpan(localCtx, sp)
 		}
-		return d.send(localCtx, ingester, timeseries, metadata)
+		return d.send(localCtx, ingester, timeseries, metadata, source)
 	}, func() { client.ReuseSlice(req.Timeseries) })
 	if err != nil {
 		return nil, err
@@ -566,7 +567,7 @@ func sortLabelsIfNeeded(labels []client.LabelAdapter) {
 	})
 }
 
-func (d *Distributor) send(ctx context.Context, ingester ring.IngesterDesc, timeseries []client.PreallocTimeseries, metadata []*client.MetricMetadata) error {
+func (d *Distributor) send(ctx context.Context, ingester ring.IngesterDesc, timeseries []client.PreallocTimeseries, metadata []*client.MetricMetadata, source client.WriteRequest_SourceEnum) error {
 	h, err := d.ingesterPool.GetClientFor(ingester.Addr)
 	if err != nil {
 		return err
@@ -576,6 +577,7 @@ func (d *Distributor) send(ctx context.Context, ingester ring.IngesterDesc, time
 	req := client.WriteRequest{
 		Timeseries: timeseries,
 		Metadata:   metadata,
+		Source:     source,
 	}
 	_, err = c.Push(ctx, &req)
 
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -360,7 +360,6 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		return nil, err
 	}
 
-	source := req.Source
 	var firstPartialErr error
 	removeReplica := false
 
@@ -537,7 +536,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		if sp := opentracing.SpanFromContext(ctx); sp != nil {
 			localCtx = opentracing.ContextWithSpan(localCtx, sp)
 		}
-		return d.send(localCtx, ingester, timeseries, metadata, source)
+		return d.send(localCtx, ingester, timeseries, metadata, req.Source)
 	}, func() { client.ReuseSlice(req.Timeseries) })
 	if err != nil {
 		return nil, err
--- a/pkg/ruler/compat.go
+++ b/pkg/ruler/compat.go
@@ -39,9 +39,9 @@ func (a *appender) AddFast(l labels.Labels, ref uint64, t int64, v float64) erro
 }
 
 func (a *appender) Commit() error {
-	wq := client.ToWriteRequest(a.labels, a.samples, nil, client.RULE)
-	_, err := a.pusher.Push(user.InjectOrgID(context.Background(), a.userID), wq)
-	client.ReuseSlice(wq.Timeseries)
+	// Since a.pusher is distributor, client.ReuseSlice will be called in a.pusher.Push.
+	// We shouldn't call client.ReuseSlice here.
+	_, err := a.pusher.Push(user.InjectOrgID(context.Background(), a.userID), client.ToWriteRequest(a.labels, a.samples, nil, client.RULE))
 	a.labels = nil
 	a.samples = nil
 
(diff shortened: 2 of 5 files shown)


pkg/ruler/compat.go
```diff
@@ -39,9 +39,12 @@ func (a *appender) AddFast(l labels.Labels, ref uint64, t int64, v float64) erro
 }
 
 func (a *appender) Commit() error {
-	_, err := a.pusher.Push(user.InjectOrgID(context.Background(), a.userID), client.ToWriteRequest(a.labels, a.samples, nil, client.RULE))
+	wq := client.ToWriteRequest(a.labels, a.samples, nil, client.RULE)
+	_, err := a.pusher.Push(user.InjectOrgID(context.Background(), a.userID), wq)
+	client.ReuseSlice(wq.Timeseries)
```
bboreham, line 44: This doesn't seem to be part of the PR description?

Wing924: it's another bugfix.
I'll update changelog and PR title.

pkg/util/push/push.go
```diff
@@ -17,14 +17,14 @@ func Handler(cfg distributor.Config, push func(context.Context, *client.WriteReq
...
 		var req client.PreallocWriteRequest
-		req.Source = client.API
 		_, err := util.ParseProtoReader(r.Context(), r.Body, int(r.ContentLength), cfg.MaxRecvMsgSize, &req, compressionType)
 		logger := util.WithContext(r.Context(), util.Logger)
 		if err != nil {
 			level.Error(logger).Log("err", err.Error())
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
+		req.Source = client.API
```
bboreham, line 27: why did this move?

Wing924: `_, err := util.ParseProtoReader(r.Context(), r.Body, int(r.ContentLength), cfg.MaxRecvMsgSize, &req, compressionType)` overwrite that value.

pkg/ruler/compat.go
pracucci: I don't think this is correct. `a.pusher` is an instance of `Distributor` and the slice reusing is already done there. I think we should rollback this change.

pkg/util/push/push.go
bboreham: I believe that was deliberate: if the caller supplies a value then we use that, otherwise we fall back to `API`.

pkg/distributor/distributor.go
```diff
@@ -360,6 +360,7 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 		return nil, err
 	}
 
+	source := req.Source
```
bboreham, line 363: This doesn't seem to be adding much; just use `req.Source` when you need it.

pkg/util/push/push.go
Wing924: I tested here and I'm sure `util.ParseProtoReader` will overwrite `req.Source` to `0`.
Because `client.API` equals to `0`, you may not notice it.

**reproduce**
```diff
diff --git a/pkg/ingester/client/cortex.proto b/pkg/ingester/client/cortex.proto
index 6a341b59..a9dfcfee 100644
--- a/pkg/ingester/client/cortex.proto
+++ b/pkg/ingester/client/cortex.proto
@@ -30,7 +30,8 @@ service Ingester {
 message WriteRequest {
   repeated TimeSeries timeseries = 1 [(gogoproto.nullable) = false, (gogoproto.customtype) = "PreallocTimeseries"];
   enum SourceEnum {
-    API = 0;
+    UNKNOWN_SOURCE = 0;
+    API = 2;
     RULE = 1;
   }
   SourceEnum Source = 2;
```
and print `req.Source` before and after`util.ParseProtoReader`.

result:
* before is `API`
* after is `UNKNOWN_SOURCE`

pkg/distributor/distributor.go
Wing924: I've fixed it.

pkg/ruler/compat.go
Wing924: I've rolled back this change.

pkg/util/push/push.go
bboreham: Did you test it with and without the caller supplying a value in the protobuf?

Wing924: I just test it with real remote write from distributor.
I don't think remote write protobuf have `Source` field.

Wing924: remote write protobuf:
https://github.com/prometheus/prometheus/blob/master/prompb/remote.proto#L22-L24

pracucci: I agree on @Wing924 on this. The source is not part of the remote write protocol, so we need to inject it after the `WriteRequest` has been deseralized, otherwise the deserialization will overwrite it with the default value.

Wing924: I added test to cover both prometheus remote write and cortex writerequest.

pkg/util/push/push.go
```diff
@@ -17,14 +17,16 @@ func Handler(cfg distributor.Config, push func(context.Context, *client.WriteReq
...
 		_, err := util.ParseProtoReader(r.Context(), r.Body, int(r.ContentLength), cfg.MaxRecvMsgSize, &req, compressionType)
 		logger := util.WithContext(r.Context(), util.Logger)
 		if err != nil {
 			level.Error(logger).Log("err", err.Error())
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
+		if req.Source == 0 {
+			req.Source = client.API
+		}
```
bboreham, line 29: As per our [instructions](https://cortexmetrics.io/docs/contributing/#workflow), every commit must explain the changes made.  This is an extraordinary change, so needs a very good explanation.

pracucci, line 29: Getting back to this change, @bboreham as right and I was wrong. The `req.Source` is not overridden by the unmarshalling if not included in the request, so we should have fixed the issue to the root: ensuring `Source` is always set within Cortex.

https://github.com/cortexproject/cortex/blob/master/pkg/ingester/client/cortex.pb.go#L5502-L5520

@Wing924 May you re-iterate on this?

Wing924: > The req.Source is not overridden by the unmarshalling if not included in the request
It's not true.

Let's change the code like this:
```diff
+ req.Source = 2
		_, err := util.ParseProtoReader(r.Context(), r.Body, int(r.ContentLength), cfg.MaxRecvMsgSize, &req, compressionType)
		logger := util.WithContext(r.Context(), util.Logger)
		if err != nil {
			level.Error(logger).Log("err", err.Error())
			http.Error(w, err.Error(), http.StatusBadRequest)
			return
		}
- if req.Source == 0 {
- 	req.Source = client.API
- }
```
if `Source` is not included in the request, we expect `req.Source` is `2`, right?
But actually it's `0`.
You can check my test.

Wing924: I'll send another PR to add comment here.

Wing924: https://github.com/cortexproject/cortex/blob/master/pkg/util/http.go#L107
this line reset req.Source to `0`

bboreham: How about you explain your change, where I have substituted one const value with its definition for clarity:
```
if req.Source == 0 {
	req.Source = 0
}
```

Wing924: I know `client.API` is equals to `0` so that you think these code is useless.
This code is explicitly set default value if req.Source is not set.

Wing924: Physically, `client.API` equals to `0`, but logically, they are different things.

bboreham: OK, the `Reset()` is indeed the important line, thanks.
So the original code didn't do anything (it set to 0 which was then overridden to 0) and the new code doesn't do anything.

PR #2459, opened 2020-04-14
Handle metric metadata in the ingesters

gotjosh:
This change allows us to handle metric metadata sent by distributors within the Ingesters.

We use an in-memory map, where keys are the metric name and the values are a set of metadata for that metric name. We only ever keep a single copy of each metadata alongside a timestamp of when was the last time we saw it. This is later used to purge the set of metadata we have not seen in a while and avoid the cache from ever-growing.

A follow up from #2336 and a part of #2311 

**TODOs**

I still have some TODOs which I indent to resolve as part of this PR, but figured I would open this up for feedback sooner rather than later. I've described them here as for the reader's benefit - anything I have missed here feel free to point it out:

- [x] Ensure limits across # of metrics to and # of metadata per metric in the ingesters.
    - [x] implementation
    - [x] tests
- [x] Ensure limits of the number of metadata we can output. Ideally, this matches ^ but we need to ensure we don't OOM the queriers.
    - [x] implementation
    - [x] tests
- [ ] Ensure we don't look across all the ingesters while retrieving metadata if `shardByAllLabels` is not enabled.
    - [ ] implementation
    - [ ] tests
- [x] Create something (probably part of the flusher?) to periodically purge metadata from the memory.
    - [x] implementation
    - [x] tests

Prometheus-related
- [x] Fix https://github.com/prometheus/prometheus/issues/7103
- [ ] vendor the updated Prometheus into Cortex
- [ ] then finish the querier implementation

The last item should be part of another PR, but noted here for the reader. 

**Checklist**
- [ ] Tests updated
- [x] Documentation added


pstibrany:
> vendor the updated Prometheus into Cortex

Cortex depends on Thanos, and Thanos needs to be updated to work with latest Prometheus first. Required changes don't seem to be big.

gotjosh:
Thanks for the head's up peter. Luckily, I don't plan to introduce that as part of this PR as it is not necessary. Once this is in, I'll start looking into the whole Cortex/Thanos/Prometheus dependency.

pkg/ingester/limiter_test.go
```diff
@@ -102,6 +102,96 @@ func TestSeriesLimit_maxSeriesPerMetric(t *testing.T) {
 	}
 }
 
+func TestSeriesLimit_maxMetadataPerMetric(t *testing.T) {
```
gotjosh, line 105: Initially, I did a complete refactor of the tests to make them accept different types of entities to test (metadata or samples) after the increasing complexity I reverted back and opted to just copy/paste. I'm generally of the opinion that tests should be dumb - I don't expect this to change quick enough to justify making It more complicated for the sake of DRYing up.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+)
+
+type metadataMap struct {
```
gotjosh, line 12: This is as simple as it can be. There's no sharding and no concurrent iteration - the volume of metadata is very low and IMHO does not justify the increased complexity. Please let me know if you differ and I'll be happy to refactor to a more performant version.

pkg/ingester/metadata_purger.go
```diff
@@ -0,0 +1,7 @@
+package ingester
+
+func (i *Ingester) purgeUsersMetadata() {
+	for _, state := range i.userStates.cp() {
+		state.purgeMetadata(i.cfg.MetadataRetainPeriod)
```
gotjosh, line 5: The purge process went through my decision-making similarly to the map. There's no priority queue and no concurrency per user. This is deliberate to get it into a reviewable state ASAP - On this one, I'm pretty sure we at least need the per-user concurrency but I'll let you tell me that instead of just assuming it, given I don't have data to work with.

pkg/ingester/ingester.go
```diff
@@ -77,15 +81,20 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 	f.DurationVar(&cfg.MaxStaleChunkIdle, "ingester.max-stale-chunk-idle", 2*time.Minute, "Maximum chunk idle time for chunks terminating in stale markers before flushing. 0 disables it and a stale series is not flushed until the max-chunk-idle timeout is reached.")
+	f.DurationVar(&cfg.FlushOpTimeout, "ingester.flush-op-timeout", 1*time.Minute, "Timeout for individual flush operations.")
 	f.DurationVar(&cfg.MaxChunkAge, "ingester.max-chunk-age", 12*time.Hour, "Maximum chunk age before flushing.")
 	f.DurationVar(&cfg.ChunkAgeJitter, "ingester.chunk-age-jitter", 0, "Range of time to subtract from -ingester.max-chunk-age to spread out flushes")
-	f.BoolVar(&cfg.SpreadFlushes, "ingester.spread-flushes", true, "If true, spread series flushes across the whole period of -ingester.max-chunk-age.")
 	f.IntVar(&cfg.ConcurrentFlushes, "ingester.concurrent-flushes", 50, "Number of concurrent goroutines flushing to dynamodb.")
+	f.BoolVar(&cfg.SpreadFlushes, "ingester.spread-flushes", true, "If true, spread series flushes across the whole period of -ingester.max-chunk-age.")
```
gotjosh, line 93: No changes here, I just moved these to match the were in which the attributes were defined in the struct itself.

pkg/ingester/ingester.go
```diff
@@ -77,15 +81,20 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
+
+	f.DurationVar(&cfg.MetadataPurgePeriod, "ingester.metadata-purge-period", 2*time.Minute, "Period at which to attempt purging metadata from memory.")
+	f.DurationVar(&cfg.MetadataRetainPeriod, "ingester.metadata-retain-period", 3*time.Minute, "Period at which metadata we have not seen will remain in memory before being deleted.")
```
gotjosh, line 96: 3x and 2x the default scrape interval. In theory, this allows at least 1 scrape for the new ones before the delete the old ones.

pkg/util/validation/limits.go
```diff
@@ -88,6 +96,13 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.MaxGlobalSeriesPerMetric, "ingester.max-global-series-per-metric", 0, "The maximum number of active series per metric name, across the cluster. 0 to disable.")
 	f.IntVar(&l.MinChunkLength, "ingester.min-chunk-length", 0, "Minimum number of samples in an idle chunk to flush it to the store. Use with care, if chunks are less than this size they will be discarded.")
 
+	f.IntVar(&l.MaxMetricsPerQuery, "ingester.max-metrics-per-query", 5000, "The maximum number of metrics with metadata that a query can return.")
+	f.IntVar(&l.MaxMetadataPerQuery, "ingester.max-metadata-per-query", 10, "The maximum number of metadata per metric that a query can return.")
+	f.IntVar(&l.MaxLocalMetadataPerUser, "ingester.max-metadata-per-user", 10000, "The maximum number of active metrics with metadata per user, per ingester. 0 to disable.")
```
gotjosh, line 101: At an average of ~55 bytes per metadata (a combination of HELP, TYPE, UNIT, METRIC NAME) we're talking of a maximum of about 55 x 10,000 x 50 which is ~27mb per tenant (plus the replication factor). 

String interning (which I have yet to do) seems like a great optimisation for this work because metadata will have a high overlap across tenants. Furthermore, metadata in typical scenarios should not go wide (what we call per metric). I have set per metric to 50 but in almost every case it should be 1 - so maybe I should lower than to 10? It would put us to around ~5mb per tenant (plus the replication factor)

gotjosh: Querying values are much lower to protect queries but I think I will tweak them a bit more.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (mms *memoryMetadataSet) all() map[memoryMetadata]model.Time {
+	mms.mtx.Lock()
+	defer mms.mtx.Unlock()
+	s := mms.set
+
+	return s
+}
+
+// add It returns false if the item was already in the set or true if it wasn't.
```
pracucci, line 98:
```suggestion
// add returns false if the item was already in the set or true if it wasn't.
```

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+func (mms *memoryMetadataSet) add(m memoryMetadata) bool {
+	mms.mtx.Lock()
+	defer mms.mtx.Unlock()
+
+	_, ok := mms.set[m]
+	mms.set[m] = model.Now()
+	return !ok
+}
+
+func (mms *memoryMetadataSet) purge(now model.Time, max time.Duration) int {
```
pracucci, line 108: Any reason why you're working with `model.Time` here instead of `time.Time`?

pkg/ingester/user_state.go
```diff
@@ -36,26 +37,36 @@ type userState struct {
 	userID              string
 	fpLocker            *fingerprintLocker
 	fpToSeries          *seriesMap
+	metricToMetadata    *metadataMap
```
pracucci, line 40: Design question. Is there the strong need to couple the metadata ingestion implementation within `userState`? I'm asking because the blocks-storage doesn't use `userState` (only its lock for other purposes) and my plan would be get rid completely of `userState` in the blocks-storage implementation (before this PR `userState` was strictly related to chunks storage).

Would be more difficult to implement the metadata ingestion outside of `userState`?

pkg/ingester/ingester.go
```diff
@@ -277,8 +286,14 @@ func (i *Ingester) loop(ctx context.Context) error {
 	rateUpdateTicker := time.NewTicker(i.cfg.RateUpdatePeriod)
 	defer rateUpdateTicker.Stop()
 
+	metadataPurgeTicker := time.NewTicker(i.cfg.MetadataPurgePeriod)
```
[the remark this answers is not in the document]
pstibrany: Alternatively: spawn a separate goroutine in `starting` and `startingV2` (for blocks), pass it the context from those starting methods, and make sure to react on context being done.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (mms *memoryMetadataSet) length() int {
+	mms.mtx.Lock()
+	l := len(mms.set)
+	mms.mtx.Unlock()
+
+	return l
+}
+
+type memoryMetadata struct {
```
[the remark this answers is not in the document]
gotjosh: I agree, I mainly added it for consistencies sake with the `memorySeries` it signals "its the same concept but for metadata". I'm happy to remove it if its a pattern we want to support.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	return deleted
+}
+
+func (mm *metadataMap) all() map[string]*memoryMetadataSet {
+	mm.mtx.Lock()
+	defer mm.mtx.Unlock()
+	metadata := mm.metricToMetadata
+	return metadata
```
pstibrany, line 43: This doesn't create a copy of map (if that's what you want), and is the same as returning `mm.metricToMetadata`. Unfortunately this is broken, since now the client calling `all()` does access the same map without any lock.

It may be easier to implement forEach method (with callback function) that keeps the lock for the entire iteration.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func (mm *metadataMap) get(metric string) (*memoryMetadataSet, bool) {
+	mm.mtx.Lock()
+	defer mm.mtx.Unlock()
+	set, ok := mm.metricToMetadata[metric]
+	if !ok {
+		return nil, false
+	}
+	return set, true
```
pstibrany, line 53:
```suggestion
	return set, ok
```

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+		set: map[memoryMetadata]model.Time{},
+	}
+}
+
+type memoryMetadataSet struct {
+	mtx sync.Mutex
+	set map[memoryMetadata]model.Time
+}
+
+func (mms *memoryMetadataSet) all() map[memoryMetadata]model.Time {
```
pstibrany, line 90: Same comment as for `(*metadataMap).all()` applies here.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+	metricToMetadata map[string]*memoryMetadataSet
+}
+
+func newMetadataMap() *metadataMap {
+	return &metadataMap{
+		metricToMetadata: map[string]*memoryMetadataSet{},
+	}
+}
+
+func (mm *metadataMap) purge(max time.Duration) int {
```
pstibrany, line 23: Functions calling `model.Now()` are hard to test. This specific function can instead be passed "cut-off" time parameter, and entries older than cut-off time are removed. (Same for `set.purge`)

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+	return s
+}
+
+// add returns false if the item was already in the set or true if it wasn't.
+func (mms *memoryMetadataSet) add(m memoryMetadata) bool {
+	mms.mtx.Lock()
+	defer mms.mtx.Unlock()
+
+	_, ok := mms.set[m]
+	mms.set[m] = model.Now()
```
pstibrany, line 104: [nit]: I would suggest to pass the time as parameter, and make all metadata with single push have same time.

pkg/ingester/ingester.go
```diff
@@ -374,9 +389,25 @@ func (i *Ingester) Push(ctx context.Context, req *client.WriteRequest) (*client.
 		}
 	}
 
-	if len(req.Metadata) > 0 {
+	var firstMetadataErr error
+	for _, metadata := range req.Metadata {
+		err := i.appendMetadata(ctx, userID, metadata)
```
pstibrany, line 394: Since this is user-supplied data, can `metadata` be nil at this point? If so, `appendMetadata` will panic.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+	"time"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+)
+
+type metadataMap struct {
+	mtx              sync.Mutex
+	metricToMetadata map[string]*memoryMetadataSet
```
pstibrany, line 14: Could `map[string]map[memoryMetadata]model.Time` here simplify the design a bit?

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+	"sync"
+	"time"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/cortexproject/cortex/pkg/ingester/client"
+)
+
+type metadataMap struct {
+	mtx              sync.Mutex
```
pstibrany, line 13: Since this map is always accessed while holding `i.userStatesMtx` (not 100% sure, but it looks like it is to me), does it really need another two (!) levels of locking?

pkg/ingester/ingester.go
gotjosh: I'm gonna go with 10, the problem I foresaw with longer periods is turning off sending metadata and having to wait 15 minutes before we would free up those limits. I'd rather keep that on the lower end if that makes any sense.

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,153 @@
...
+	metricType client.MetricMetadata_MetricType
+}
+
+func newMemoryMetadata(m *client.MetricMetadata) memoryMetadata {
+	return memoryMetadata{
+		metricName: m.MetricName,
+		unit:       m.Unit,
+		metricType: m.Type,
+		help:       m.Help,
+	}
```
pstibrany, line 143:
```suggestion
	return memoryMetadata{
		metricName: m.GetMetricName(),
		unit:       m.GetUnit(),
		metricType: m.GetType(),
		help:       m.GetHelp(),
	}
```

This will also work if `m` is `nil`.

pkg/ingester/ingester.go
```diff
@@ -77,15 +81,20 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
+	f.DurationVar(&cfg.FlushOpTimeout, "ingester.flush-op-timeout", 1*time.Minute, "Timeout for individual flush operations.")
 	f.DurationVar(&cfg.MaxChunkAge, "ingester.max-chunk-age", 12*time.Hour, "Maximum chunk age before flushing.")
 	f.DurationVar(&cfg.ChunkAgeJitter, "ingester.chunk-age-jitter", 0, "Range of time to subtract from -ingester.max-chunk-age to spread out flushes")
-	f.BoolVar(&cfg.SpreadFlushes, "ingester.spread-flushes", true, "If true, spread series flushes across the whole period of -ingester.max-chunk-age.")
 	f.IntVar(&cfg.ConcurrentFlushes, "ingester.concurrent-flushes", 50, "Number of concurrent goroutines flushing to dynamodb.")
+	f.BoolVar(&cfg.SpreadFlushes, "ingester.spread-flushes", true, "If true, spread series flushes across the whole period of -ingester.max-chunk-age.")
+
+	f.DurationVar(&cfg.MetadataPurgePeriod, "ingester.metadata-purge-period", 2*time.Minute, "Period at which to attempt purging metadata from memory.")
```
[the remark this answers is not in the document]
gotjosh: My thinking here was that implementations with high churn rate would want to be aggressive with purging - so I wanted to give an option on that. But I can see your point, a sensible default could work for a long time before we even have to worry about that - Would 5m (1/2 of the default purge period) work as a hardcoded value?

pkg/ingester/ingester.go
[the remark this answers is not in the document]
gotjosh: I had a similar thought process but decide that retain period was closer based on the discarded design. `MaxChunkIdle ` refers to the time before we commit the chunk (as in saving it) whereas `RetainPeriod` refers to the time before we remove it from memory. Given metadata (as you mentioned) is discarded by design this felt a bit closer to that. 

But more importantly, should we later decide to save metadata somewhere we would have to change the semantics of the flags at that point - I would like to avoid it if possible. 

tl;dr: In my head, "idle" refers to "time before saving" and "retain period" refers to "time before discarding" - that's why I chose the latter.

pkg/distributor/distributor.go
```diff
@@ -700,6 +702,34 @@ func (d *Distributor) MetricsForLabelMatchers(ctx context.Context, from, through
...
+
+	result := []scrape.MetricMetadata{}
+	for _, resp := range resps {
+		r := resp.(*client.MetricMetadataResponse)
+		for _, m := range r.Metadata {
+			result = append(result, scrape.MetricMetadata{
+				Metric: m.MetricName,
+				Help:   m.Help,
+				Unit:   m.Unit,
+				Type:   textparse.MetricTypeCounter,
```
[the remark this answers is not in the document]
gotjosh: Great shout, that's why TODOs are useful 😆.

pkg/distributor/distributor_test.go
```diff
@@ -1005,6 +1005,7 @@ type mockIngester struct {
 	happy      bool
 	stats      client.UsersStatsResponse
 	timeseries map[uint32]*client.PreallocTimeseries
+	metadata   map[uint32]map[*client.MetricMetadata]struct{}
```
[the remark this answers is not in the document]
gotjosh: I quickly learned that pointers as key were a bad idea. This was just a remanent of that. Thanks for catching it.

pkg/ingester/ingester.go
```diff
@@ -77,15 +81,20 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.WALConfig.RegisterFlags(f)
 
 	f.IntVar(&cfg.MaxTransferRetries, "ingester.max-transfer-retries", 10, "Number of times to try and transfer chunks before falling back to flushing. Negative value or zero disables hand-over.")
+
```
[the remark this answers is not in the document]
gotjosh: This is indented - it is expressed that way in the attributes section.

docs/configuration/config-file-reference.md
```diff
@@ -2100,6 +2109,33 @@ The `limits_config` configures default and per-tenant limits imposed by Cortex s
 # CLI flag: -ingester.min-chunk-length
 [min_chunk_length: <int> | default = 0]
 
+# The maximum number of metrics with metadata that a query can return.
+# CLI flag: -ingester.max-metrics-per-query
+[max_metrics_per_query: <int> | default = 5000]
```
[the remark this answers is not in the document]
gotjosh: I think it is, queriers might have a very spec from ingesters and without them, users risk OOMing the queriers.

pkg/ingester/user_state.go
gotjosh: Let me know what you think here, I've tried to keep everything in one place - we can worry about extracting methods for re-using after we get the coupling right.

pkg/distributor/distributor.go
```diff
@@ -700,6 +702,34 @@ func (d *Distributor) MetricsForLabelMatchers(ctx context.Context, from, through
 	return result, nil
 }
 
+// MetricMetadata returns all metric metadata of a user.
+func (d *Distributor) MetricMetadata(ctx context.Context) ([]scrape.MetricMetadata, error) {
```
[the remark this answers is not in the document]
gotjosh: I have added new tests that use this, shouldn't be the case anymore.

pkg/ingester/metadata_map.go
gotjosh: I removed all locks and left the locking quirks to the caller.

pkg/ingester/metadata_map.go
gotjosh: Ditto from above.

pkg/ingester/metadata_map.go
gotjosh: I agree that mocking time in Go is a hard thing. However, these are never tested directly, just indirectly through ingesters so that wasn't really a factor. That being said, I think the deadline concept makes it easier to reason about so I have changed to that. Thanks!

pkg/ingester/ingester.go
gotjosh: I have started using the `Get` prefixed functions all around, thanks for the hint in that direction 😄

pkg/ingester/metadata_map.go
gotjosh: Ditto from above about the locking.

pkg/ingester/ingester.go
```diff
@@ -495,6 +536,58 @@ func (i *Ingester) append(ctx context.Context, userID string, labels labelPairs,
...
+func (i *Ingester) appendMetadata(ctx context.Context, userID string, metadata *client.MetricMetadata) error {
+	i.userStatesMtx.RLock()
+	if i.stopped {
+		i.userStatesMtx.RUnlock()
+		return fmt.Errorf("ingester stopping")
+	}
+	i.userStatesMtx.RUnlock()
+
+	i.usersMetadataMtx.Lock()
+	defer i.usersMetadataMtx.Unlock()
```
pracucci, line 548: In my idea (the designed I shared with you and Peter offline) I suggested to use this lock only for the `getOrCreateUserMetadata()` and then having a lock inside `metadataMap struct` basically making `metadataMap` and `memoryMetadataSet` functions goroutines safe. This way we would have a fine-grained lock instead of a global lock.

The global lock here guarantees perfect consistency in the code below (ie. between `i.getOrCreateUserMetadata()` and `userMetadata.put()`) but I think we don't need a perfect limit while a global lock is quite scary.

@pstibrany How do you feel with a global lock here?

pkg/ingester/ingester.go
```diff
@@ -495,6 +536,58 @@ func (i *Ingester) append(ctx context.Context, userID string, labels labelPairs,
...
+
+	// As we get the set, we also validate two things:
+	// 1. The user is allowed to create new metrics to add metadata to.
+	// 2. If the metadata set is already present, it hasn't reached the limit of metadata we can append.
+	userMetadata := i.getOrCreateUserMetadata(userID)
+	set, ok := userMetadata.get(metadata.GetMetricName())
+	if !ok {
+		// Verify that the user can create more metric metadata given we don't have a set for that metric name.
+		if err := i.limiter.AssertMaxMetricsWithMetadataPerUser(userID, userMetadata.length()); err != nil {
+			validation.DiscardedMetadata.MustCurryWith(prometheus.Labels{"user": userID}).WithLabelValues(perUserMetadataLimit).Inc()
```
pracucci, line 558: Why not simply `validation.DiscardedMetadata.WithLabelValues(perUserMetadataLimit, userID).Inc()` ?

pkg/util/validation/limits.go
```diff
@@ -88,6 +96,13 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.MaxGlobalSeriesPerMetric, "ingester.max-global-series-per-metric", 0, "The maximum number of active series per metric name, across the cluster. 0 to disable.")
 	f.IntVar(&l.MinChunkLength, "ingester.min-chunk-length", 0, "Minimum number of samples in an idle chunk to flush it to the store. Use with care, if chunks are less than this size they will be discarded.")
 
+	f.IntVar(&l.MaxMetricsWithMetadataPerQuery, "ingester.max-metrics-with-metadata-per-query", 4000, "The maximum number of metrics with metadata that a query can return.")
+	f.IntVar(&l.MaxMetadataPerQuery, "ingester.max-metadata-per-query", 5, "The maximum number of metadata per metric that a query can return.")
```
pracucci, line 100: As discussed offline, I think limits in the read path are not necessary because the read metadata API endpoint allows to read all metadata only, so we never have write-path limits != read-path limits, thus write-path limits should be enough.

pkg/util/validation/limits.go
```diff
@@ -88,6 +96,13 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
...
+	f.IntVar(&l.MaxLocalMetricsWithMetadataPerUser, "ingester.max-metadata-per-user", 8000, "The maximum number of active metrics with metadata per user, per ingester. 0 to disable.")
+	f.IntVar(&l.MaxLocalMetadataPerMetric, "ingester.max-metadata-per-metric", 10, "The maximum number of metadata per metric, per ingester. 0 to disable.")
+	f.IntVar(&l.MaxGlobalMetricsWithMetadataPerUser, "ingester.max-global-metadata-per-user", 0, "The maximum number of active metrics with metadata per user, across the cluster. 0 to disable. Supported only if -distributor.shard-by-all-labels is true.")
+	f.IntVar(&l.MaxGlobalMetadataPerMetric, "ingester.max-global-metadata-per-metric", 0, "The maximum number of metadata per metric, across the cluster. 0 to disable.")
```
pracucci, line 104: Local limits are scary like they're for series. There's no real good reason to use local limits, except that we're not able to enforce global limits if `shard-by-all-labels` is disabled. I'm wondering if could makes sense to always use `shard-by-all-labels` for metadata and remove local limits at all (supporting only global limits). @gouthamve @pstibrany ?

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (mm *metadataMap) all() map[string]*memoryMetadataSet {
+	metadata := mm.metricToMetadata
+	return metadata
+}
+
+func (mm *metadataMap) get(metric string) (*memoryMetadataSet, bool) {
+	set, ok := mm.metricToMetadata[metric]
+	return set, ok
```
pracucci, line 43:
```suggestion
	return mm.metricToMetadata[metric]
```

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (mm *metadataMap) put(metric string, set *memoryMetadataSet) {
+	mm.metricToMetadata[metric] = set
+}
+
+func (mm *metadataMap) length() int {
+	l := len(mm.metricToMetadata)
+	return l
```
pracucci, line 52:
```suggestion
	return len(mm.metricToMetadata)
```

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,118 @@
...
+			delete(mm.metricToMetadata, m)
+		}
+	}
+
+	return deleted
+}
+
+func (mm *metadataMap) all() map[string]*memoryMetadataSet {
+	metadata := mm.metricToMetadata
+	return metadata
```
pracucci, line 38:
```suggestion
	return mm.metricToMetadata
```

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,118 @@
...
+
+func newMemoryMetadataSet() *memoryMetadataSet {
+	return &memoryMetadataSet{
+		set: map[memoryMetadata]time.Time{},
+	}
+}
+
+func (mms *memoryMetadataSet) all() map[memoryMetadata]time.Time {
+	s := mms.set
+	return s
```
pracucci, line 67:
```suggestion
	return mms.set
```

pkg/ingester/metadata_map.go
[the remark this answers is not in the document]
pracucci: @pstibrany ?

pkg/ingester/metadata_map.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	return deleted
+}
+
+func (mms *memoryMetadataSet) length() int {
+	l := len(mms.set)
+
+	return l
```
pracucci, line 92:
```suggestion
	return len(mms.set)
```

pkg/ingester/limiter_test.go
pracucci: Different tests are preferable to me too.

pkg/ingester/ingester.go
```diff
@@ -739,6 +832,47 @@ func (i *Ingester) MetricsForLabelMatchers(ctx context.Context, req *client.Metr
...
+		// If this is a "high-cardinality" metadata set. Let's skip it to at least return
+		// part of the metadata.
+		// TODO: Should we create a metric to keep track of metadata skipped?
+		if set.length() > i.limiter.MaxMetadataPerQuery(userID) {
+			continue
+		}
+		for mm := range set.all() {
+			metadata = append(metadata, mm.toClientMetadata())
+		}
+	}
```
pracucci, line 871: If we remove the limit on the read path, we can move all this logic to `metadataMap.toClientMetadata()`. This would have the advantage to remove the `all()` functions which are problematic if we move away from the global lock.

pkg/distributor/distributor.go
```diff
@@ -700,6 +701,42 @@ func (d *Distributor) MetricsForLabelMatchers(ctx context.Context, from, through
...
+			result = append(result, scrape.MetricMetadata{
+				Metric: m.MetricName,
+				Help:   m.Help,
+				Unit:   m.Unit,
+				Type:   client.MetricMetadataToMetricTypetoMetricType(m.GetType()),
+			})
+		}
+	}
+
+	return result, nil
```
pracucci, line 737: I don't know: does the Prometheus API guarantees any ordering on the result? Here results are not guaranteed to be ordered by metric name, so I'm wondering if we should to keep 100% Prometheus compatibility or not.

pkg/ingester/ingester.go
```diff
@@ -739,6 +832,47 @@ func (i *Ingester) MetricsForLabelMatchers(ctx context.Context, req *client.Metr
...
+	i.usersMetadataMtx.RLock()
+	defer i.usersMetadataMtx.RUnlock()
+	userMetadata, ok := i.usersMetadata[userID]
+	if !ok {
+		return &client.MetricMetadataResponse{}, nil
+	}
+
+	if userMetadata.length() > i.limiter.MaxMetricsWithMetadataPerQuery(userID) {
+		return nil, httpgrpc.Errorf(http.StatusRequestEntityTooLarge, "exceeded maximum number of metadata metrics in a query")
+	}
```
pstibrany, line 858: Per private discussion between you and Marco, I expect that limits on read paths will be removed (?)

gotjosh: Limits have been removed.

pkg/ingester/ingester.go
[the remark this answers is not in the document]
gotjosh: Done.

pkg/ingester/ingester.go
gotjosh: Done.

docs/configuration/config-file-reference.md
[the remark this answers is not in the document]
gotjosh: Limits have been removed.

pkg/util/validation/limits.go
gotjosh: Limits have been removed.

pkg/distributor/distributor.go
gotjosh: It is. Prometheus does not guarantee any order.

pkg/ingester/user_metrics_metadata_purge.go
```diff
@@ -0,0 +1,30 @@
...
+import "time"
+
+const (
+	metadataPurgePeriod = 5 * time.Minute
+)
+
+func (i *Ingester) purgeUserMetricsMetadata() {
+	i.usersMetadataMtx.Lock()
+	usersMetadata := i.usersMetadata
+	i.usersMetadataMtx.Unlock()
```
pstibrany, line 12: This does nothing.

Unfortunately lock will need to be held during entire `purgeUserMetricsMetadata` method. Iterating the `i.usersMetadata` map while it's possibly being modified by other goroutines is not safe.

pkg/ingester/user_metrics_metadata_purge.go
```diff
@@ -0,0 +1,30 @@
+package ingester
+
+import "time"
+
+const (
+	metadataPurgePeriod = 5 * time.Minute
+)
+
+func (i *Ingester) purgeUserMetricsMetadata() {
```
pstibrany, line 9: I don't see why this function needs its own file :)

pkg/ingester/user_metrics_metadata_purge.go
gotjosh: Good catch, you're completely right - at the end of the day it's just a reference.

gotjosh:
Have addressed your lastest on https://github.com/cortexproject/cortex/pull/2459/commits/bd9df6e7a259a98c75ebfe552cbd0766a9a31c12 @pstibrany.

pkg/ingester/ingester.go
```diff
@@ -374,9 +389,25 @@ func (i *Ingester) Push(ctx context.Context, req *client.WriteRequest) (*client.
 		}
 	}
 
-	if len(req.Metadata) > 0 {
+	var firstMetadataErr error
```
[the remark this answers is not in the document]
gotjosh: We're close to 100 comments in this PR - is it OK if we tackle v2 as part of a separate one? I'm happy to open it immediately after this one.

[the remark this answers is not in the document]
gotjosh: I think the current design should make it effortless to include - the biggest changes would come from the tests.

pkg/ingester/ingester.go
```diff
@@ -502,6 +545,55 @@ func (i *Ingester) append(ctx context.Context, userID string, labels labelPairs,
...
+	i.userStatesMtx.RLock()
+	if i.stopped {
+		i.userStatesMtx.RUnlock()
+		return fmt.Errorf("ingester stopping")
+	}
+	i.userStatesMtx.RUnlock()
+
+	i.usersMetadataMtx.Lock()
+	userMetadata := i.getOrCreateUserMetadata(userID)
+	i.usersMetadataMtx.Unlock()
```
pracucci, line 558: This locking should be moved inside `getOrCreateUserMetadata()` so that we can implement it taking a read lock in most of the cases:

```
func (i *Ingester) getOrCreateUserMetadata(userID string) *userMetricsMetadata {
	i.usersMetadataMtx.RLock()
	userMetadata, ok := i.usersMetadata[userID]
	i.usersMetadataMtx.RUnlock()
	
	if ok {
		return userMetadata
	}
	
	i.usersMetadataMtx.Lock()
	defer i.usersMetadataMtx.Unlock()
	
	// Ensure it has not been created in the meanwhile.
	userMetadata, ok = i.usersMetadata[userID]
	if !ok {
		userMetadata = newMetadataMap(i.limiter, i.metrics, userID)
		i.usersMetadata[userID] = userMetadata
	}

	return userMetadata
}
```

pkg/ingester/ingester.go
```diff
@@ -502,6 +545,55 @@ func (i *Ingester) append(ctx context.Context, userID string, labels labelPairs,
...
+
+	if err := userMetadata.add(m.GetMetricName(), m); err != nil {
+		return err
+	}
+
+	return nil
```
pracucci, line 564:
```suggestion
	return userMetadata.add(m.GetMetricName(), m)
```

pkg/ingester/ingester.go
```diff
@@ -502,6 +545,55 @@ func (i *Ingester) append(ctx context.Context, userID string, labels labelPairs,
...
+	userMetadata, ok := i.usersMetadata[userID]
+	if !ok {
+		userMetadata = newMetadataMap(i.limiter, i.metrics, userID)
+		i.usersMetadata[userID] = userMetadata
+	}
+
+	return userMetadata
+}
+
+func (i *Ingester) purgeUserMetricsMetadata() {
```
pracucci, line 577: We shouldn't take the global lock for the entire purge duration. It can be re-implemented like this:

```

func (i *Ingester) getOrCreateUserMetadata(userID string) *userMetricsMetadata {
	userMetadata := i.getUserMetadata(userID)
	if userMetadata != nil {
		return userMetadata
	}

	i.usersMetadataMtx.Lock()
	defer i.usersMetadataMtx.Unlock()

	// Ensure it has not been created in the meanwhile.
	userMetadata, ok := i.usersMetadata[userID]
	if !ok {
		userMetadata = newMetadataMap(i.limiter, i.metrics, userID)
		i.usersMetadata[userID] = userMetadata
	}

	return userMetadata
}

func (i *Ingester) getUserMetadata(userID string) *userMetricsMetadata {
	i.usersMetadataMtx.RLock()
	defer i.usersMetadataMtx.RUnlock()

	return i.usersMetadata[userID]
}

func (i *Ingester) getUsersWithMetadata() []string {
	i.usersMetadataMtx.RLock()
	defer i.usersMetadataMtx.RUnlock()

	userIDs := make([]string, 0, len(i.usersMetadata))
	for userID, _ := range i.usersMetadata {
		userIDs = append(userIDs, userID)
	}

	return userIDs
}

func (i *Ingester) purgeUserMetricsMetadata() {
	deadline := time.Now().Add(-i.cfg.MetadataRetainPeriod)

	for _, userID := range i.getUsersWithMetadata() {
		metadata := i.getUserMetadata(userID)
		if metadata == nil {
			continue
		}

		// Remove all metadata that we no longer need to retain.
		deletedCount := metadata.purge(deadline)

		i.metrics.memMetadata.Sub(float64(deletedCount))
		i.metrics.memMetadataRemovedTotal.WithLabelValues(userID).Add(float64(deletedCount))
	}
}
```

**SUPER IMPORTANT**: with the lock optimizations I'm suggesting, there's no safe way to do `delete(i.usersMetadata, userID)` once the user metadata is empty (because another one outside the global lock could retain a reference to it and writing). It's perfectly OK keeping the data structure. It's super small and will not go over the max number of tenants in the cluster.

pkg/ingester/ingester.go
```diff
@@ -746,6 +838,31 @@ func (i *Ingester) MetricsForLabelMatchers(ctx context.Context, req *client.Metr
...
+		return nil, err
+	}
+	i.userStatesMtx.RUnlock()
+
+	userID, err := user.ExtractOrgID(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("no user id")
+	}
+
+	i.usersMetadataMtx.RLock()
```
pracucci, line 855: Following my suggested code you can now use `getUserMetadata()`

pkg/ingester/user_metrics_metadata.go
```diff
@@ -0,0 +1,111 @@
...
+	if !ok {
+		mm.metrics.memMetadata.Inc()
+		mm.metrics.memMetadataCreatedTotal.WithLabelValues(mm.userID).Inc()
+	}
+
+	mm.metricToMetadata[metric][*metadata] = time.Now()
+	return nil
+}
+
+func (mm *userMetricsMetadata) purge(deadline time.Time) int {
```
pracucci, line 66: Given you have metrics in this data structure, why not moving the following here?

```
i.metrics.memMetadata.Sub(float64(deletedCount))
i.metrics.memMetadataRemovedTotal.WithLabelValues(userID).Add(float64(deletedCount))
```

pkg/ingester/user_metrics_metadata.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	return deleted
+}
+
+func (mm *userMetricsMetadata) toClientMetadata() []*client.MetricMetadata {
+	mm.mtx.RLock()
+	defer mm.mtx.RUnlock()
+	r := []*client.MetricMetadata{}
```
pracucci, line 84: We could make the assumption that in most cases each set contains only 1 metadata (which means that even if you're running multiple Prometheus servers remote writing, given a metric name the metadata is the exact same). It's just a guess, but better that starting with a 0-length slice and doubling its size each time.

```suggestion
	r := make([]*client.MetricMetadata, 0, len(mm.metricToMetadata))
```

pkg/ingester/ingester.go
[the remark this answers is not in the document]
gotjosh: Agreed offline to include it in this PR.

gotjosh:
@pracucci @pstibrany I think the last three commits should address the latest feedback now.

pkg/ingester/ingester.go
```diff
@@ -463,6 +441,8 @@ func (i *Ingester) Push(ctx context.Context, req *client.WriteRequest) (*client.
 		recordPool.Put(record)
 	}
 
+	i.pushMetadata(ctx, userID, req.GetMetadata())
```
pracucci, line 444: Given we're implementing it to both support series and metadata (you told so), this line should be moved before series, otherwise in case of a partial error in the series metadata will not be ingested. Same comment applies to the blocks storage.

PR #2463, opened 2020-04-15
Add initial per tenant query and chunk metrics AND RENAME SOME EXISTING METRICS

gouthamve:
More will be added once the following is merged:
https://github.com/prometheus/prometheus/pull/6890

The idea is that we now have a ton of QPS and we don't know which tenants are using queries more than others. We want to add the samples/chunks loaded metrics as well so that we can come up with fair usage policy on the read path.

Ideally we'd add latency metrics as well, but the cardinality is just too much.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,8 +27,9 @@
 * [ENHANCEMENT] Single Binary: Added query-frontend to the single binary.  Single binary users will now benefit from various query-frontend features.  Primarily: sharding, parallelization, load shedding, additional caching (if configured), and query retries. #2437
 * [ENHANCEMENT] Allow 1w (where w denotes week) and 1y (where y denotes year) when setting `-store.cache-lookups-older-than` and `-store.max-look-back-period`. #2454
 * [ENHANCEMENT] Optimize index queries for matchers using "a|b|c"-type regex. #2446 #2475
-* [ENHANCEMENT] Added per tenant metrics for chunks and bytes read from chunk store: #2463
+* [ENHANCEMENT] Added per tenant metrics for queries and chunks and bytes read from chunk store: #2463
   * `cortex_chunk_store_fetched_chunks_total` and `cortex_chunk_store_fetched_chunk_bytes_total`
+  * `cortex_query_frontend_queries_total` (per tenant queries counted by the frontend)
 * [BUGFIX] Fixes #2411, Ensure requests are properly routed to the prometheus api embedded in the query if `-server.path-prefix` is set. #2372
 * [BUGFIX] Experimental TSDB: fixed chunk data corruption when querying back series using the experimental blocks storage. #2400
 * [BUGFIX] Cassandra Storage: Fix endpoint TLS host verification. #2109
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
   * Updated the index hosted at the root prefix to point to the updated routes.
   * Legacy routes hardcoded with the `/api/prom` prefix now respect the `-http.prefix` flag.
 * [CHANGE] The metrics `cortex_distributor_ingester_appends_total` and `distributor_ingester_append_failures_total` now includes a `type` label to differentiate between `samples` and `metadata`. #2336
+* [CHANGE] The metrics for number of chunks and bytes flushed to the chunk store are renamed from: #2463
+  * `cortex_ingester_chunks_stored_total` > `cortex_chunk_store_stored_chunks_total`
+  * `cortex_ingester_chunk_stored_bytes_total` > `cortex_chunk_store_stored_chunk_bytes_total`
 * [CHANGE] Experimental TSDB: renamed blocks meta fetcher metrics: #2375
   * `cortex_querier_bucket_store_blocks_meta_syncs_total` > `cortex_querier_blocks_meta_syncs_total`
   * `cortex_querier_bucket_store_blocks_meta_sync_failures_total` > `cortex_querier_blocks_meta_sync_failures_total`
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
   * Updated the index hosted at the root prefix to point to the updated routes.
   * Legacy routes hardcoded with the `/api/prom` prefix now respect the `-http.prefix` flag.
 * [CHANGE] The metrics `cortex_distributor_ingester_appends_total` and `distributor_ingester_append_failures_total` now includes a `type` label to differentiate between `samples` and `metadata`. #2336
+* [CHANGE] The metrics for number of chunks and bytes flushed to the chunk store are renamed from: #2463
+  * `cortex_ingester_chunks_stored_total` > `cortex_chunk_store_stored_chunks_total`
```
pstibrany, line 12: Would it make sense to keep original names, and name new metrics with `cortex_querier_` prefix, instead of using `cortex_chunk_store` prefix?

pkg/chunk/chunk_store.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/validation"
 )
 
+// Exported errors.
```
pstibrany, line 28: Lint complaining again? Otherwise this comment is just noise.

pkg/chunk/storage/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+	chks, err := c.client.GetChunks(ctx, chunks)
+	if err != nil {
+		return chks, err
+	}
+
+	user, err := user.ExtractOrgID(ctx)
+	// Should never happen.
+	if err != nil {
+		return nil, err
+	}
```
pstibrany, line 101: Should we use userID from chunks again? Does `GetChunk` method contract say that chunks can only be for single user?

pkg/querier/queryrange/roundtrip.go
```diff
@@ -130,6 +132,13 @@ func NewTripperware(
 	minShardingLookback time.Duration,
 	registerer prometheus.Registerer,
 ) (frontend.Tripperware, cache.Cache, error) {
+	// Per tenant query metrics.
+	queriesPerTenant := promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "query_frontend_queries_total",
+		Help:      "Total queries sent per tenant.",
+	}, []string{"method", "user"})
```
pstibrany, line 140: "method" is easily confused with HTTP method. What about "op"?

CHANGELOG.md
gouthamve: We'll be using the deletes package to track this and soon the backfilling / migration tool to track costs as well. So ingester is not the only thing that will read and write chunks.

pkg/chunk/chunk_store.go
gouthamve: My local editor complains, I'll remove this.

CHANGELOG.md
codesome: It would be nice to know which component is fetching how many chunks in that case, a combined metric sounds less useful. Maybe an additional label?

pkg/chunk/storage/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/weaveworks/common/user"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+)
+
+type metricsChunkClient struct {
```
codesome, line 12: Sounds confusing. How about `chunkClientWithMetricsWrapper`? A bit verbose though.

pkg/chunk/storage/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+	client chunk.Client
+
```
codesome, line 14:
```suggestion
```

CHANGELOG.md
gouthamve: It's currently not possible to track which component is doing the request without significant plumbing. And given our deletes and migrations are one offs, I'd rather go down that route just yet.

pkg/chunk/storage/metrics.go
```diff
@@ -0,0 +1,119 @@
...
+func (c metricsChunkClient) Stop() {
+	c.client.Stop()
+}
+
+func (c metricsChunkClient) PutChunks(ctx context.Context, chunks []chunk.Chunk) error {
+	if err := c.client.PutChunks(ctx, chunks); err != nil {
+		return err
+	}
+
+	if len(chunks) == 0 {
```
codesome, line 66: Can we have this at the top instead?

pkg/chunk/storage/metrics.go
```diff
@@ -0,0 +1,119 @@
...
+		userCounts[c.UserID]++
+	}
+
+	for user, size := range userSizes {
+		c.metrics.chunksSizePutPerUser.WithLabelValues(user).Add(float64(size))
+	}
+
+	for user, num := range userCounts {
+		c.metrics.chunksPutPerUser.WithLabelValues(user).Add(float64(num))
+	}
```
codesome, line 85: Nit: Looks very spacious for small loops :P we can avoid newlines between the loops (even for `GetChunks`).

gouthamve:
@pstibrany Thanks for the feedback, I've incorporated most of it. It's ready for another review

pkg/querier/queryrange/roundtrip.go
```diff
@@ -130,6 +132,13 @@ func NewTripperware(
 	minShardingLookback time.Duration,
 	registerer prometheus.Registerer,
 ) (frontend.Tripperware, cache.Cache, error) {
+	// Per tenant query metrics.
+	queriesPerTenant := promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "query_frontend_queries_total",
```
pracucci, line 138: I would mention it in the CHANGELOG.

bboreham:
Please be really loud when you make breaking changes.

pkg/chunk/storage/metrics.go
pstibrany: Given its limited scope, I think this works. Short comment would be nice :)

PR #2468, opened 2020-04-15
include Cortex version in cortex_build_info

### LINKED ISSUE #1709: Startup message can have blank version

thorfour:
**What this PR does**: Adds the version,branch,revision to the build args, so that `cortex_build_info` correctly populates the version number.

Fixes #1709 

`cortex_build_info{branch="build-info-version",goversion="go1.13.3",revision="ca93c1b87",version="1.0.0"} 1`

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


bboreham:
Thanks for this PR, definitely needed!  Should add "fixes #1709".

We had some discussion in #1751 why `./tools/image-tag` might be a better choice.

thorfour:
I've update the Makefile to use the short revision format, but I don't think there's any value in using `./tools/image-tage` since that combines the fields we want.

pstibrany:
Shall we have CHANGELOG entry for this?

PR #2470, opened 2020-04-16
Undo a spurious change in push handler

bboreham:
The meaning of the code before #2457 is that we initialise to `API` but allow the unmarshal code to override.

Tests were added in #2457 to confirm this works as expected.


bboreham:
Closing; the original code was ineffective because `Unmarshal()` calls `Reset()`.

The new code is also ineffective but fixing needs more thought.

PR #2267, opened 2020-03-12
Compile code in vendor dir; don't download modules again

bboreham:
**What this PR does**:
Tells Go to use the code we have checked in under `vendor`.

**Which issue(s) this PR fixes**:
Fixes #1710 
Fixes #1795

**Checklist**
N/A

PR #2480, opened 2020-04-17
Update golang to 1.14.4 and other build deps

gouthamve:
Checked to see that there are no breaking changes


gouthamve:
Will merge this and open a new PR with the updated build-image

pracucci:
> Will merge this and open a new PR with the updated build-image

@gouthamve Why? What's the problem of updating the build-image in this PR?

tomwilkie:
> @gouthamve Why? What's the problem of updating the build-image in this PR?

I think Goutham prefers to use the `master-xyz` tag for the build image.

pracucci:
> > @gouthamve Why? What's the problem of updating the build-image in this PR?
> 
> I think Goutham prefers to use the `master-xyz` tag for the build image.

I see it, but the problem is that we're going to merge a change to the `build-image` before seeing CI pass with it. That's why, so far, we've always updated the build-image tag along side the PR, even if the branch used for it is not `master-xyz` (which to me looks a secondary problem, compared to merging changes in the build-image without having the CI green light).

pracucci:
Closes #2267.

gouthamve:
Currently blocked by https://github.com/etcd-io/bbolt/issues/187and https://github.com/etcd-io/bbolt/issues/214

Hitting us both in our etcd tests (which spins up a mock etcd, hence an internal bolt instance) and our local boltdb index storage tests.

pracucci:
Luckily we tested before merging.

pstibrany:
Can we revisit this? Running tests locally with Go 1.14.3 works just fine for me.

build-image/Dockerfile
```diff
@@ -1,26 +1,27 @@
...
 RUN apt-get install -y nodejs npm && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 RUN npm install -g postcss-cli autoprefixer
-ENV HUGO_VERSION=v0.59.1
+ENV HUGO_VERSION=v0.72.0
 RUN git clone https://github.com/gohugoio/hugo.git --branch ${HUGO_VERSION} --depth 1 && \
 	cd hugo  && go install --tags extended && cd ../ && \
 	rm -rf hugo/ && rm -rf /go/pkg /go/src
-RUN curl -fsSLo shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt_v1.3.0_linux_amd64 && \
-	echo "b1925c2c405458811f0c227266402cf1868b4de529f114722c2e3a5af4ac7bb2  shfmt" | sha256sum -c && \
+ENV SHFMT_VERSION=3.1.0
```
pracucci, line 11: Who does use/need `shfmt`?

build-image/Dockerfile
```diff
@@ -1,26 +1,27 @@
-FROM golang:1.13.3-stretch
+FROM golang:1.14.4-stretch
 RUN apt-get update && apt-get install -y curl python-requests python-yaml file jq unzip protobuf-compiler libprotobuf-dev && \
 	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-RUN curl -sL https://deb.nodesource.com/setup_6.x | sh -
+RUN curl -sL https://deb.nodesource.com/setup_13.x | sh -
```
pracucci, line 4: Node.js 13.x is no longer actively supported. We should switch to 14.x which is LTS.

pracucci, line 4: While installing Node 13.x while building image I'm also seeing some errors:
```
Step 3/23 : RUN curl -sL https://deb.nodesource.com/setup_13.x | sh -
 ---> Running in 068ea5b2bde2
sh: 129: [: X_13.x: unexpected operator
sh: 85: [[: not found
```

These errors look going away if you change `| sh` to `| bash`

build-image/Dockerfile
gouthamve: Not sure

gouthamve: Will test and send another PR. Don't want to block this.

pstibrany:
We now build binaries using Go 1.14, which is a change that should be noted in the CHANGELOG in my opinion.

pracucci:
> We now build binaries using Go 1.14, which is a change that should be noted in the CHANGELOG in my opinion.

Mentioned in https://github.com/cortexproject/cortex/pull/2753.

PR #2481, opened 2020-04-17
frontend: also log the POST data of long queries

GiedriusS:
The feature of query-frontend to log slower queries is already amazing
but we can make it even better by logging the POST body as well. It is
not uncommon nowadays to use POST with the Prometheus API. All of the
data that is passed via POST is not visible via the URL. Thus, let's
also print the POST data if it is available.

Small testing shows that the message now looks like the following:

```
level=info ts=2020-04-17T16:34:51.200200445Z caller=frontend.go:174 msg="slow query" org_id=fake url=http://localhost:9009/api/prom/api/v1/query_range time_taken=1.638201ms body="end=1587141285&query=prometheus_http_requests_total&start=1587140385&step=15"
```

Signed-off-by: Giedrius Statkevičius <giedriuswork@gmail.com>


GiedriusS:
> LGTM

Thanks, I have added the PR number to the changelog as well which I have missed before.

PR #2483, opened 2020-04-17
QueryFrontend sets incorrect status code on HTTP timeout

### LINKED ISSUE #2482: QueryFrontend sets incorrect status code on HTTP timeout
QueryFrontend does not check for the error from io.Copy, and sets incorrect status code if the error occurred.

dmitsh:

**What this PR does**:
It fixes bug where query frontend ignores an error in HTTP handler and reports incorrect status code

Fixes #2482

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -171,16 +171,16 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if _, err = io.Copy(w, resp.Body); err != nil {
-		server.WriteError(w, err)
-		return
-	}
-
 	hs := w.Header()
 	for h, vs := range resp.Header {
 		hs[h] = vs
 	}
 	w.WriteHeader(resp.StatusCode)
+
+	if _, err = io.Copy(w, resp.Body); err != nil {
+		server.WriteError(w, err)
+		return
+	}
 }
 
 func writeError(w http.ResponseWriter, err error) {
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -171,12 +171,16 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if _, err = io.Copy(w, resp.Body); err != nil {
+		server.WriteError(w, err)
+		return
+	}
+
 	hs := w.Header()
 	for h, vs := range resp.Header {
 		hs[h] = vs
 	}
 	w.WriteHeader(resp.StatusCode)
-	io.Copy(w, resp.Body)
 }
 
 func writeError(w http.ResponseWriter, err error) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 * [BUGFIX] Ring: Fixed a situation where upgrading from pre-1.0 cortex with a rolling strategy caused new 1.0 ingesters to lose their zone value in the ring until manually forced to re-register. #2404
 * [BUGFIX] Distributor: `/all_user_stats` now show API and Rule Ingest Rate correctly. #2457
 * [BUGFIX] Fixed `version`, `revision` and `branch` labels exported by the `cortex_build_info` metric. #2468
+* [BUGFIX] QueryFrontend: fixed a situation where HTTP error is ignored and an incorrect status code is set. #2483
 
 ## 1.0.0 / 2020-04-02
 


bboreham:
Thanks for this; it's in a good direction but I don't think you can move `WriteHeader` to after the response body is written - see https://golang.org/pkg/net/http/#ResponseWriter.

dmitsh:
> Thanks for this; it's in a good direction but I don't think you can move `WriteHeader` to after the response body is written - see https://golang.org/pkg/net/http/#ResponseWriter.

@bboreham thank you for reviewing, I updated the PR.

bboreham:
I don't get this.  If we already set a good status then setting a bad one will fail.

What exactly is the failure scenario that this PR works in?
Can we have a test for it?

pkg/querier/frontend/frontend.go
```diff
@@ -176,7 +176,11 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 		hs[h] = vs
 	}
 	w.WriteHeader(resp.StatusCode)
-	io.Copy(w, resp.Body)
+
+	if _, err = io.Copy(w, resp.Body); err != nil {
+		server.WriteError(w, err)
```
pracucci, line 181: I can't understand how this could could work. It will write the HTTP header but it has already been written above. Am I missing anything @dmitsh?

gouthamve: `server.WriteError` calls `w.WriteHeader` as well https://github.com/weaveworks/common/blob/8f725fc52d8defd2b0f7eecab55a86f33421c08f/httpgrpc/server/server.go#L163-L175 which I thought overrides the previous header set.

But I was wrong:
```
	// WriteHeader sends an HTTP response header with the provided
	// status code.
	//
	// If WriteHeader is not called explicitly, the first call to Write
	// will trigger an implicit WriteHeader(http.StatusOK).
	// Thus explicit calls to WriteHeader are mainly used to
	// send error codes.
	//
	// The provided code must be a valid HTTP 1xx-5xx status code.
	// Only one header may be written. Go does not currently
	// support sending user-defined 1xx informational headers,
	// with the exception of 100-continue response header that the
	// Server sends automatically when the Request.Body is read.

	WriteHeader(statusCode int)
```

We can only call `WriteHeader` once.

dmitsh: This is a chicken and egg situation. We have to set the status before calling `Write`, but the `Write` can fail.
With this change, `server.WriteError` will override the status, and we are logging the correct information:
```
level=warn ts=2020-04-20T14:37:20.70899Z caller=logging.go:49 traceID=29b2da71f9d67c9b msg="GET /api/v1/query_range?&query=custom_metric_2000&start=2020-04-20T06%3A00%3A00-07%3A00&end=2020-04-20T06%3A30%3A00-07%3A00&step=10 (500) 32.354122946s Response: \"write tcp [::1]:9090->[::1]:56587: i/o timeout\\n\" ws: false; Accept: */*; User-Agent: curl/7.64.1; "
```

dmitsh: Just to clarify:
Before this PR, if `io.Copy()` has failed, cortex was returning an empty response (probably because the write was aborted in the middle). But the log was indicating successful status code (200).
With my PR, cortex still returns an empty response if `io.Copy()` fails (for the same reason), but we have the correct status code (500) and the error message in the log and metrics

gouthamve: Yep, just went through the code again. The problem is `io.Copy` will fail if `w.Write` fails. And we shouldn't try to attempt a `w.Write` again (which is what `server.WriteError(w, err)` does internally). I am not sure why the log is printing 200, I would say this is a logging middleware issue and this is not the right fix. But I haven't looked at the logging middleware yet, will do now.

gouthamve: https://github.com/weaveworks/common/blob/master/middleware/logging.go#L40-L42

This is only because we had a bad implementation of the logging middleware. https://github.com/weaveworks/common/blob/master/middleware/response.go#L42-L54 <-- We should ideally capture the error here.

bboreham: I imagine the most likely scenario for `Write()` to fail is that the client has disconnected.
In that case the context should be cancelled; I tried to detect and log those in https://github.com/weaveworks/common/pull/180 but perhaps ineffectively.

Note it is somewhat unwarranted to count a user action as a Cortex failure - if you're trying to apply an SLI you don't want them lumped in together.  We could do like Nginx, and count them as code 499.

dmitsh: An easy way to reproduce this problem is to set `http_server_write_timeout` to a short duration (1 sec?), and then invoke HTTP request. If it takes more than 1 second to get the data, the request will time out.

Issue #2353, opened 2020-03-30
Metric for chunks stored should take acount of de-duplication


bboreham:
Currently `cortex_ingester_chunks_stored_total` reports all chunks that were sent to the store; below that level the store may avoid the write and return success, but this isn't visible.

Similarly `cortex_ingester_chunk_stored_bytes_total`

(Note these metrics were renamed in April 2020)

(Note that even when the Cortex code doesn't detect duplication the underlying store may do so; there isn't much we can do about that in real-time)

gouthamve:
`chunk_store_stored_chunks_total` and `cortex_chunk_store_stored_chunk_bytes_total` now report per user metric. And this is after deduping :)


See: #2463

bboreham:
I see nothing in the description or CHANGELOG for #2463 about deduping.

pstibrany:
> I see nothing in the description or CHANGELOG for #2463 about deduping.

I added this in https://github.com/cortexproject/cortex/pull/2579, which will eventually be merged into master.

pracucci:
@bboreham Is the issue solved?

bboreham:
Yes I believe it was solved by #2463

PR #2485, opened 2020-04-19
Add a counter to track chunks which were de-duplicated at store time.

slim-bean:
Signed-off-by: Ed Welch <edward.welch@grafana.com>

**What this PR does**:

At the moment it's a bit tricky to understand how effectively chunks are being de-duplicated at the store with the cache based de-dupe code.

This PR adds a counter which explicitly increments when a chunk was not stored because it was already in the cache.  

This can be combined with `cortex_ingester_flush_reasons` to get a ratio of chunks flushed vs deduped to better tell how well the de-duplication is working.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/series_store.go
+++ b/pkg/chunk/series_store.go
@@ -61,6 +61,11 @@ var (
 		// For 100k series for 7 week, could be 1.2m - 10*(8^(7-1)) = 2.6m.
 		Buckets: prometheus.ExponentialBuckets(10, 8, 7),
 	})
+	dedupedChunksTotal = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "chunk_store_deduped_chunks_total",
+		Help:      "Count of chunks which were not stored because they have already been stored by another replica.",
+	})
 )
 
 // seriesStore implements Store
@@ -414,6 +419,7 @@ func (c *seriesStore) PutOne(ctx context.Context, from, through model.Time, chun
 	// If this chunk is in cache it must already be in the database so we don't need to write it again
 	found, _, _ := c.cache.Fetch(ctx, []string{chunk.ExternalKey()})
 	if len(found) > 0 {
+		dedupedChunksTotal.Inc()
 		return nil
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
 * [ENHANCEMENT] Added per tenant metrics for queries and chunks and bytes read from chunk store: #2463
   * `cortex_chunk_store_fetched_chunks_total` and `cortex_chunk_store_fetched_chunk_bytes_total`
   * `cortex_query_frontend_queries_total` (per tenant queries counted by the frontend)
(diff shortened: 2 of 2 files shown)


pracucci:
I think we could assume this fixes #2353.

pkg/chunk/series_store.go
```diff
@@ -61,6 +61,11 @@ var (
 		// For 100k series for 7 week, could be 1.2m - 10*(8^(7-1)) = 2.6m.
 		Buckets: prometheus.ExponentialBuckets(10, 8, 7),
 	})
+	dedupedChunksTotal = promauto.NewCounter(prometheus.CounterOpts{
```
pracucci, line 64: This correlate with `cortex_chunk_store_stored_chunks_total` which is per user. @gouthamve are we OK having this metric non per-user?

slim-bean: I don't know of a case where I would want to query this metric with a user dimension, that is, I'm not sure it's ever meaningful to know if one user has more de-duped chunks than another.

However if you were already querying `cortex_chunk_store_stored_chunks_total` with a user dimension and wanted to compare accurately with this metric then it would also need a user dimension.

My general feeling is there isn't a use case for checking de-duplication with a user and you would query both `cortex_chunk_store_stored_chunks_total` and this metric without a user label.

bboreham: You need it per-user if you're trying to charge users based on how much of the store they used.

bboreham:
Is this different to the dedupe cache hit rate?

pkg/chunk/series_store.go
slim-bean: Is a count metric useful for this? Is there a good way to convert number of chunks flushed to the size of those chunks?

slim-bean:
> Is this different to the dedupe cache hit rate?

If I followed all this correctly, the cache used for checking de-dupes is the chunks cache which is also used when `FetchChunks` is called, so the hit rate on this cache is affected by queries (if you have a lot of queries for really old data they would drive up the miss rate on this cache)

The index `writeDedupeCache` is only checked after the chunks cache is checked, so if the chunks cache returned a result, the writeDedupeCache is never checked and therefore the stats for it won't reflect successful de-dupes

bboreham:
OK, is this different to the dedupe cache hit rate exposed by ingesters?
(ingesters never call FetchChunks)

pkg/chunk/series_store.go
bboreham: It's all a little fuzzy since we don't include metadata in `cortex_ingester_chunk_stored_bytes_total` and don't have any user-specific counters for labels.

I'll modify #2353 to mention the bytes metric too.

slim-bean:
> OK, is this different to the dedupe cache hit rate exposed by ingesters?
(ingesters never call FetchChunks)

Interesting I hadn't though about the runtime separation.

The problem here would be anyone running a singlebinary, which is possible to do with caching and distributed stores like dynamo/s3.  I know this probably wouldn't be advised but we see this a lot with Loki Helm users.

Also a recent change in Loki will allow ingesters to query the store directly (part of how the persisting BoltDB to the object store implementation works), which would result in ingesters calling FetchChunks and impacting using cache stats for this purpose.

For these reasons I still see value in a separate metric for this.

PR #2486, opened 2020-04-20
Enable WAL by default in single process config

codesome:
Signed-off-by: Ganesh Vernekar <cs15btech11018@iith.ac.in>

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/configuration/single-process-config.yaml
+++ b/docs/configuration/single-process-config.yaml
@@ -33,6 +33,11 @@ ingester:
   spread_flushes: true
   chunk_age_jitter: 0
 
+  walconfig:
+    wal_enabled: true
+    recover_from_wal: true
+    wal_dir: /tmp/cortex/wal
+
   lifecycler:
     # The address to advertise for this ingester.  Will be autodiscovered by
     # looking up address on eth0 or en0; can be specified if this fails.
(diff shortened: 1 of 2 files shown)


docs/configuration/single-process-config.yaml
```diff
@@ -33,6 +33,11 @@ ingester:
   spread_flushes: true
   chunk_age_jitter: 0
 
+  walconfig:
+    wal_enabled: true
+    recover_from_wal: true
+    wal_dir: /tmp/cortex/wal
```
pracucci, line 39: Running Cortex with the default image fails with the following error:
`failed to recover from WAL: open /tmp/cortex/wal: no such file or directory`

This is not just an error you get when running integration tests but also running it locally with (the following command picked from the doc):
`./cortex -config.file=./docs/configuration/single-process-config.yaml`

codesome: Yes, I found that out before seeing this review and have already pushed a fix for it. Now a directory is created when WAL or recovery from WAL is enabled.

codesome: Integration tests are passing now :)

pkg/ingester/ingester.go
```diff
@@ -157,6 +158,12 @@ func New(cfg Config, clientConfig client.Config, limits *validation.Overrides, c
 		}
 	}
 
+	if cfg.WALConfig.WALEnabled || cfg.WALConfig.Recover {
+		if err := os.MkdirAll(cfg.WALConfig.Dir, 0777); err != nil {
```
pracucci, line 162:
```suggestion
		if err := os.MkdirAll(cfg.WALConfig.Dir, os.ModePerm); err != nil {
```

PR #2487, opened 2020-04-20
Additional notes on running WAL in production

codesome:
The changes are mostly geared towards non-kubernetes scenerio.

/cc @tomwilkie

### DIFF
--- a/docs/production/ingesters-with-wal.md
+++ b/docs/production/ingesters-with-wal.md
@@ -74,3 +74,14 @@ You can run it as a Kubernetes job which will:
 3. And flush all the chunks.
 
 This job is to be run for all the PVCs linked to the ingesters that you missed hitting the shutdown endpoint as a first option.
+
+## Additional notes
+
+* If you have lots of ingestion with the WAL replay taking a longer time, you can try reducing the checkpoint duration to `10-15m`. This would require slightly higher disk bandwidth for writes (still less in absolute terms), but it will reduce the WAL replay time overall.
+
+### Non-Kubernetes or baremetal deployments
+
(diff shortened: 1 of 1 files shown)


docs/production/ingesters-with-wal.md
```diff
@@ -74,3 +74,14 @@ You can run it as a Kubernetes job which will:
 3. And flush all the chunks.
 
 This job is to be run for all the PVCs linked to the ingesters that you missed hitting the shutdown endpoint as a first option.
+
+## Additional notes
+
+* If you have lots of ingestion with the WAL replay taking a longer time, you can try reducing the checkpoint duration to `10-15m`. This would require slightly higher disk bandwidth for writes (still less in absolute terms), but it will reduce the WAL replay time overall.
```
pracucci, line 80: > reducing the checkpoint duration

I would mention the CLI flag to do it. Also I would keep it simple a give a value to set (not a range), so I would pick either `10m` or `15m`.

docs/production/ingesters-with-wal.md
```diff
@@ -74,3 +74,14 @@ You can run it as a Kubernetes job which will:
...
+
+### Non-Kubernetes or baremetal deployments
+
+* When the ingester restarts for any reason (upgrade, crash, etc), it should be able to attach to the same volume in order to recover back the WAL and tokens.
+    * If it fails to attach to the same volume for any reason, use the flusher as described above to flush that data.
```
pracucci, line 85:
```suggestion
    * If it fails to attach to the same volume for any reason, use the [flusher](#scale-down) to flush that data.
```

docs/production/ingesters-with-wal.md
codesome: I will mention a specific value. And about CLI flag, I will add that too. As the flags were mentioned at the start of the doc, I felt it was not necessary.

PR #2491, opened 2020-04-21
Change LifecyclerConfig.ListenPort from pointer to int

pracucci:
**What this PR does**:
Looking at the current state of the code (including Loki) I can't see a good reason why `LifecyclerConfig.ListenPort` should be a pointer instead of `int`. In this PR I'm proposing to switch it to `int` to cleanup the config.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci:
@gouthamve Do you see any problem with this change?

PR #2502, opened 2020-04-22
Add TLS support to HTTP/GRPC clients

### LINKED ISSUE #2350: Support TLS
Cortex as a project does not have many mechanisms in place to secure its messages. As things stand,  no security mechanisms exist with cortex and that responsibility is off-loaded to the system that proxies requests to Cortex. As a distributed system with significant traffic between services, I think we should consider adding support for TLS within Cortex. Initially this support could be focuses on GRPC services within Cortex and then expanded to cover its HTTP services.

annanay25:
**What this PR does**:
- `go.mod` updated to vendor latest commit of `weaveworks/common` which adds TLS support to the server.
- Adds TLS support to all HTTP and GRPC clients in cortex.
  - GRPC: 
    - `pkg/util/grpcclient` updated to include TLS info.
    - All GRPC clients updated to use `grpcclient.DialOptions()` which provides TLS options as well.
  - HTTP:
    - Util package `pkg/util/httpclient` created with a `Config` struct that stores necessary endpoint, timeout and tls config parameters.
    - All HTTP clients updated to use this config struct.

Fixes #2350 

**Checklist**
- [ ] Tests updated
- [x] Documentation added


cmd/query-tee/main.go
```diff
@@ -3,23 +3,23 @@ package main
...
-	ServerMetricsPort  int
-	BackendEndpoints   string
-	PreferredBackend   string
-	BackendReadTimeout time.Duration
-	LogLevel           logging.Level
+	ServerServicePort int
+	ServerMetricsPort int
+	BackendEndpoints  string
+	PreferredBackend  string
+	ClientConfig      httpclient.Config
```
annanay25, line 21: This is going to be slightly confusing because now there's `Config.BackendEndpoints` as well as `Config.ClientConfig.Endpoint`

cmd/query-tee/proxy_backend.go
```diff
@@ -5,46 +5,50 @@ import (
...
+		clientConfig: config,
+		preferred:    preferred,
 		client: &http.Client{
 			CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
 				return errors.New("the query-tee proxy does not follow redirects")
 			},
-			Transport: &http.Transport{
-				Proxy: http.ProxyFromEnvironment,
-				DialContext: (&net.Dialer{
-					Timeout:   30 * time.Second,
```
annanay25, line 41: Warning: This is changed to `5 * time.Second`

pkg/querier/querier.go
```diff
@@ -60,6 +62,7 @@ var (
 
 // RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.StoreGatewayGRPCClientConfig.RegisterFlagsWithPrefix("querier.store-gateway-client", f)
```
pracucci, line 65: The store-gateway is experimental so any of its flags should be prefixed with "experimental".

```suggestion
	cfg.StoreGatewayGRPCClientConfig.RegisterFlagsWithPrefix("experimental.querier.store-gateway-client", f)
```

pkg/querier/store_gateway_client.go
```diff
@@ -58,17 +57,7 @@ func (c *storeGatewayClient) String() string {
 	return c.conn.Target()
 }
 
-func newStoreGatewayClientPool(discovery client.PoolServiceDiscovery, logger log.Logger, reg prometheus.Registerer) *client.Pool {
-	// We prefer sane defaults instead of exposing further config options.
-	clientCfg := grpcclient.Config{
```
pracucci, line 63: Unfortunately having used `grpcclient.Config` now we have added more complexity to the config for something we don't really want to let configure at this stage.

pkg/ruler/ruler.go
```diff
@@ -59,6 +60,8 @@ var (
 type Config struct {
 	// This is used for template expansion in alerts; must be a valid URL.
 	ExternalURL flagext.URLValue `yaml:"external_url"`
+	// Config parameters for the GRPC Client
+	GRPCClientConfig grpcclient.Config `yaml:"grpc_client_config"`
```
pracucci, line 64: From the UX perspective, we're adding all `grpcclient.Config` to the ruler client just for TLS.

docs/configuration/config-file-reference.md
```diff
@@ -1064,7 +1164,7 @@ storage:
 
   # The configstore_config configures the config database storing rules and
   # alerts, and is used by the Cortex alertmanager.
-  # The CLI flags prefix for this block config is: alertmanager
+  # The CLI flags prefix for this block config is: alertmanager.client
```
pracucci, line 1167: This is a config breaking change we can't affort. Alertmanager is not among the list of experimental features.

docs/configuration/config-file-reference.md
```diff
@@ -2513,19 +2649,35 @@ api:
 
 The `configstore_config` configures the config database storing rules and alerts, and is used by the Cortex alertmanager. The supported CLI flags `<prefix>` used to reference this config block are:
 
-- `alertmanager`
-- `ruler`
+- `alertmanager.client`
+- `ruler.`
```
pracucci, line 2653: The ending `.` here means there's a misconfiguration somewhere in the `RegisterFlagsWithPrefix()` which causes a double `..` in the final generated CLI flag.

docs/configuration/config-file-reference.md
```diff
@@ -2513,19 +2649,35 @@ api:
...
 
 &nbsp;
 
 ```yaml
-# URL of configs API server.
-# CLI flag: -<prefix>.configs.url
-[configs_api_url: <url> | default = ]
+# Endpoint to connect to
+# CLI flag: -<prefix>.http-endpoint
```
pracucci, line 2659: This is a breaking change.

docs/configuration/config-file-reference.md
```diff
@@ -2513,19 +2649,35 @@ api:
...
+[endpoint: <url> | default = ]
 
-# Timeout for requests to Weave Cloud configs service.
-# CLI flag: -<prefix>.configs.client-timeout
+# Timeout for connecting to the endpoint
+# CLI flag: -<prefix>.http-client-timeout
```
pracucci, line 2663: This is a breaking change.

Also, CLI flags and YAML option should be specular. If the YAML is `client_timeout` then the CLI flag should be `client-timeout`. _Same comment applies to all other flags._

pkg/util/httpclient/httpclient.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/prometheus/common/log"
+
+	"github.com/cortexproject/cortex/pkg/util/flagext"
+)
+
+// Config is the config for the HTTP client
+type Config struct {
+	Endpoint           flagext.URLValue `yaml:"endpoint"`
+	ClientTimeout      time.Duration    `yaml:"client_timeout"`
+	BackendReadTimeout time.Duration    `yaml:"backend_read_timeout"`
```
pracucci, line 20: This is used only by the query-tee. If there's a config option in a generic `Config` then it must be supported by any client using this config. In this specific case, I would suggest to rollback any change in the `query-tee` (not supporting TLS there is fine, the query-tee is just used for testing right now and we could add TLS support anytime in the future) and I would remove this config option at all.

pkg/querier/store_gateway_client.go
annanay25: These can practically be left un-configured because they will continue to use the defaults.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -69,7 +69,11 @@ type storageClientV1 struct {
 
 // NewStorageClientV1 returns a new v1 StorageClient.
 func NewStorageClientV1(ctx context.Context, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.IndexClient, error) {
-	opts := toOptions(cfg.GRPCClientConfig.DialOption(bigtableInstrumentation()))
+	dialOpts, err := cfg.GRPCClientConfig.DialOption(bigtableInstrumentation())
```
jtlisi, line 72: Do we need this for Bigtable? Google seems to have their own way of doing this: https://godoc.org/google.golang.org/api/option#WithCredentials. It's based around GCP service accounts. I think we should avoid adding this for GCP services for now.

integration/query_frontend_test.go
```diff
@@ -92,16 +97,25 @@ func runQueryFrontendTest(t *testing.T, setup queryFrontendSetup) {
...
 		"-querier.cache-results":             "true",
 		"-querier.split-queries-by-interval": "24h",
 		"-frontend.memcached.addresses":      "dns+" + memcached.NetworkEndpoint(e2ecache.MemcachedPort),
 	})
 
 	// Start Cortex components.
-	queryFrontend := e2ecortex.NewQueryFrontendWithConfigFile("query-frontend", configFile, flags, "")
-	ingester := e2ecortex.NewIngesterWithConfigFile("ingester", consul.NetworkHTTPEndpoint(), configFile, flags, "")
-	distributor := e2ecortex.NewDistributorWithConfigFile("distributor", consul.NetworkHTTPEndpoint(), configFile, flags, "")
+	queryFrontend := e2ecortex.NewQueryFrontendWithConfigFile("query-frontend", configFile, mergeFlags(flags, GetServerTLSFlags()), "")
```
jtlisi, line 116: When you create the Service you need to ensure that the built in ReadinessCheck supports TLS. Same with any external client that accesses the service.

pkg/ruler/ruler.go
```diff
@@ -59,6 +60,8 @@ var (
 type Config struct {
 	// This is used for template expansion in alerts; must be a valid URL.
 	ExternalURL flagext.URLValue `yaml:"external_url"`
+	// TLS parameters for the GRPC Client
+	ClientTLSConfig tls_cfg.ClientConfig `yaml:",inline"`
```
pracucci, line 64: The YAML config is inline, while the CLI flag is prefixed by `-ruler.client` (ie. `-ruler.client.tls-cert-path`). This is a config inconsistency that we should fix. I think having the struct inline here is quite confusing (for which specific client are we configuring TLS?), so I would suggest to not inline the YAML struct here, so:

```suggestion
	ClientTLSConfig tls_cfg.ClientConfig `yaml:"ruler_client"`
```

I've named it `ruler_client` (and CLI flag prefix should be reflected as well) because it's the client that a ruler uses to connect to other rulers. The ruler can also connect to other endpoints (ie. alert manager) so it's quite important to avoid ambiguities here.

pkg/querier/querier.go
```diff
@@ -51,7 +52,8 @@ type Config struct {
 	ActiveQueryTrackerDir string `yaml:"active_query_tracker_dir"`
 
 	// Blocks storage only.
-	StoreGatewayAddresses string `yaml:"store_gateway_addresses"`
+	StoreGatewayAddresses        string            `yaml:"store_gateway_addresses"`
+	StoreGatewayGRPCClientConfig grpcclient.Config `yaml:"store_gateway_client_config"`
```
pracucci, line 56:
```suggestion
	StoreGatewayGRPCClientConfig grpcclient.Config `yaml:"store_gateway_client"`
```

pkg/querier/store_gateway_client.go
pracucci: In the ruler you just imported the TLS config. Can we do the same thing here, please?

pkg/configs/client/client.go
```diff
@@ -2,34 +2,37 @@ package client
...
+	tls_cfg "github.com/cortexproject/cortex/pkg/util/tls"
 )
 
 // Config says where we can find the ruler userconfig.
 type Config struct {
-	ConfigsAPIURL flagext.URLValue `yaml:"configs_api_url"`
-	ClientTimeout time.Duration    `yaml:"client_timeout"` // HTTP timeout duration for requests made to the Weave Cloud configs service.
+	ConfigsAPIURL flagext.URLValue     `yaml:"configs_api_url"`
+	ClientTimeout time.Duration        `yaml:"client_timeout"` // HTTP timeout duration for requests made to the Weave Cloud configs service.
+	TLSStruct     tls_cfg.ClientConfig `yaml:",inline"`
```
pracucci, line 28: Remove the `Struct` suffix please.

integration/util.go
```diff
@@ -51,3 +51,26 @@ func copyFileToSharedDir(s *e2e.Scenario, src, dst string) error {
 
 	return writeFileToSharedDir(s, dst, content)
 }
+
+// GetServerTLSFlags generates generic TLS flags for a server
+func GetServerTLSFlags() map[string]string {
```
pracucci, line 56: No need to expose it.

integration/util.go
```diff
@@ -51,3 +51,26 @@ func copyFileToSharedDir(s *e2e.Scenario, src, dst string) error {
...
+		"-server.http-tls-ca-path":     filepath.Join(e2e.ContainerSharedDir, rootCertFile),
+		"-server.grpc-tls-cert-path":   filepath.Join(e2e.ContainerSharedDir, serverCertFile),
+		"-server.grpc-tls-key-path":    filepath.Join(e2e.ContainerSharedDir, serverKeyFile),
+		"-server.grpc-tls-client-auth": "RequireAndVerifyClientCert",
+		"-server.grpc-tls-ca-path":     filepath.Join(e2e.ContainerSharedDir, rootCertFile),
+	}
+}
+
+// GetClientTLSFlagsWithPrefix generates generic TLS flags for a client
+func GetClientTLSFlagsWithPrefix(prefix string) map[string]string {
```
pracucci, line 70: No need to expose it.

integration/query_frontend_test.go
```diff
@@ -92,16 +97,25 @@ func runQueryFrontendTest(t *testing.T, setup queryFrontendSetup) {
 
 	configFile, flags := setup(t, s)
 
+	// setup tls
+	cmd := exec.Command("bash", "certs/genCerts.sh", "certs", "1")
```
pracucci, line 101: I would prefer if we move the script to `docs/` as a reference to generate certificates and we commit the certificates used in integration tests directly into `integration/certs` so that we don't have to generate them each time.

integration/configs.go
```diff
@@ -22,16 +22,21 @@ const (
 	cortexConfigFile       = "config.yaml"
 	cortexSchemaConfigFile = "schema.yaml"
 	blocksStorageEngine    = "tsdb"
+	clientCertFile         = "certs/client.crt"
+	clientKeyFile          = "certs/client.key"
+	rootCertFile           = "certs/root.crt"
```
pracucci, line 27: I would call it `caCertFile` to keep the naming consistent with the Cortex config.

docs/configuration/single-process-config-blocks-tls.yaml
```diff
@@ -0,0 +1,100 @@
+
```
pracucci, line 1: For each pre-cooked config we provide, we have an integration test in `integration/getting_started_single_process_config_test.go`. Would be great if you could add a test there for this file (a new test, which can be an existing one you copy, paste and modify as needed).

docs/production/tls.md
```diff
@@ -0,0 +1,108 @@
+---
+title: "Securing communication between cortex components with TLS"
```
pracucci, line 2: Please always `Cortex` with the upper case `C`.

_This comment applies across the whole file._

docs/production/tls.md
```diff
@@ -0,0 +1,108 @@
...
+
+The first step to securing inter-service communication in Cortex with TLS is
+generating certificates. A Certifying Authority (CA) will be used for this
+purpose which should be private to the organization, as any certificates signed
+by this CA will have permissions to communicate with the cluster.
+
+We will use the following script to generate self signed certs for the cluster:
+
+```
+# Refer: https://github.com/joe-elliott/cert-exporter/blob/69d3d7230378325a1de4fa313432d3d6ced4a518/test/files/genCerts.sh
```
pracucci, line 22: Can we refer `genCerts.sh` committed in this PR?

docs/production/tls.md
```diff
@@ -0,0 +1,108 @@
...
+    -server.grpc-tls-key-path=/path/to/server.key
+
+    # Type of Client Auth for the GRPC Server
+    -server.grpc-tls-client-auth="RequireAndVerifyClientCert"
+
+    # Path to the Client CA Cert for the GRPC Server
+    -server.grpc-tls-ca-path=/path/to/root.crt
+```
+
+#### Client flags
```
pracucci, line 82: You've referenced alertmanager and querier. What's about other components?

integration/query_frontend_test.go
```diff
@@ -92,16 +97,25 @@ func runQueryFrontendTest(t *testing.T, setup queryFrontendSetup) {
...
+	require.NoError(t, cmd.Run())
+	require.NoError(t, copyFileToSharedDir(s, integrationHomeFolder+clientCertFile, clientCertFile))
+	require.NoError(t, copyFileToSharedDir(s, integrationHomeFolder+clientKeyFile, clientKeyFile))
+	require.NoError(t, copyFileToSharedDir(s, integrationHomeFolder+rootCertFile, rootCertFile))
+	require.NoError(t, copyFileToSharedDir(s, integrationHomeFolder+serverCertFile, serverCertFile))
+	require.NoError(t, copyFileToSharedDir(s, integrationHomeFolder+serverKeyFile, serverKeyFile))
+
```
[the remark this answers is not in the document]
pracucci: I would actually suggest to commit certs in the `integration/certs` and then copy all of them to the shared dir.

pkg/util/tls/tls.go
```diff
@@ -0,0 +1,69 @@
...
+type ClientConfig struct {
+	CertPath string `yaml:"tls_cert_path"`
+	KeyPath  string `yaml:"tls_key_path"`
+	CAPath   string `yaml:"tls_ca_path"`
+}
+
+// RegisterFlagsWithPrefix registers flags with prefix.
+func (cfg *ClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.StringVar(&cfg.CertPath, prefix+".tls-cert-path", "", "TLS cert path for the client")
+	f.StringVar(&cfg.KeyPath, prefix+".tls-key-path", "", "TLS key path for the client")
```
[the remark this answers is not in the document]
annanay25: _As discussed online this is needed for client authentication._

pkg/util/tls/tls.go
```diff
@@ -0,0 +1,69 @@
...
+func (cfg *ClientConfig) GetGRPCDialOptions() ([]grpc.DialOption, error) {
+	var opts []grpc.DialOption
+	if tlsConfig, err := cfg.GetTLSConfig(); err != nil {
+		return nil, err
+	} else if tlsConfig != nil {
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+	} else {
+		opts = append(opts, grpc.WithInsecure())
+	}
+	return opts, nil
```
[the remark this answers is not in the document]
annanay25: Thanks.

pkg/util/grpcclient/grpcclient.go
```diff
@@ -7,6 +7,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/cortexproject/cortex/pkg/util"
+	tls_cfg "github.com/cortexproject/cortex/pkg/util/tls"
```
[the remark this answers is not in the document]
annanay25: Some files require a rename because they are also importing `crypto/tls`. But not needed here. Will make the change.

pkg/util/grpcclient/grpcclient.go
```diff
@@ -65,3 +69,24 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 		grpc.WithStreamInterceptor(grpc_middleware.ChainStreamClient(streamClientInterceptors...)),
 	}
 }
+
+// DialOptionWithTLS returns the config as a grpc.DialOptions
+func (cfg *Config) DialOptionWithTLS(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) ([]grpc.DialOption, error) {
```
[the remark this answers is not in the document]
annanay25: Thanks, there is just one other problem. The function signatures for `DialOption()` and `DialOptionWithTLS()` are different (one of them returns an error while the other does not):

```
DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) []grpc.DialOption

DialOptionWithTLS(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) ([]grpc.DialOption, error)
```

Will is be okay for two structs `Config` & `ConfigWithTLS` to have the same function name `DialOption()` but different return parameters? If we would like to fix this, we can either - Change function signature of `Config.DialOption()` to match that of `ConfigWithTLS.DialOption()` or rename the latter one to `DialOptionWithTLS` wherever used.

docs/production/tls.md
annanay25: Complete list of flags is available in `config-file-reference.md`. This is just an example setup.

pkg/util/tls/tls.go
```diff
@@ -32,15 +30,13 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
 	if cfg.CertPath != "" && cfg.KeyPath != "" && cfg.CAPath != "" {
 		clientCert, err := tls.LoadX509KeyPair(cfg.CertPath, cfg.KeyPath)
 		if err != nil {
-			level.Error(util.Logger).Log("msg", "error loading certs", "error", err)
-			return nil, err
+			return nil, fmt.Errorf("failed to load TLS certs: %v", err)
```
pracucci, line 33: Please use `errors.Wrapf()` (from `github.com/pkg/errors`) and also add the path of the files (following my previous comment).

pkg/util/tls/tls.go
```diff
@@ -32,15 +30,13 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
...
 		}
 
 		var caCertPool *x509.CertPool
 		caCert, err := ioutil.ReadFile(cfg.CAPath)
 		if err != nil {
-			level.Error(util.Logger).Log("msg", "error loading ca cert", "error", err)
-			return nil, err
+			return nil, fmt.Errorf("error loading ca cert: %v", err)
```
pracucci, line 39: Please use `errors.Wrapf()` (from `github.com/pkg/errors`) and also add the path of the files (following my previous comment).

pkg/util/grpcclient/grpcclient.go
```diff
@@ -7,6 +7,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/cortexproject/cortex/pkg/util"
+	tls "github.com/cortexproject/cortex/pkg/util/tls"
```
pracucci, line 10:
```suggestion
	"github.com/cortexproject/cortex/pkg/util/tls"
```

pkg/ruler/ruler.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/services"
+	tls_cfg "github.com/cortexproject/cortex/pkg/util/tls"
```
pracucci, line 38: No need for `tls_cfg`

pkg/querier/store_gateway_client.go
```diff
@@ -58,8 +62,7 @@ func (c *storeGatewayClient) String() string {
 	return c.conn.Target()
 }
 
-func newStoreGatewayClientPool(discovery client.PoolServiceDiscovery, logger log.Logger, reg prometheus.Registerer) *client.Pool {
-	// We prefer sane defaults instead of exposing further config options.
```
pracucci, line 62: Do not remove this comment, please.

docs/production/tls.md
pracucci: Then please mention it in this doc, because reading this doc is not clear that other clients support it.

integration/query_frontend_test.go
pracucci: Discussed offline. I've been told the general consensus is not to commit certs/keys (even if they're self signed).

integration/e2ecortex/services.go
```diff
@@ -72,7 +72,7 @@ func NewQuerierWithConfigFile(name, consulAddress, configFile string, flags map[
 		image,
 		e2e.NewCommandWithoutEntrypoint("cortex", e2e.BuildArgs(e2e.MergeFlags(map[string]string{
 			"-target":                         "querier",
-			"-log.level":                      "warn",
+			"-log.level":                      "debug",
```
pracucci, line 75: Could you rollback to `debug` please?

integration/query_frontend_test.go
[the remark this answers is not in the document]
jtlisi: sounds good, better than generating them for every run.

PR #2505, opened 2020-04-22
Re-use buffers when serialising chunks.

bboreham:
Improves performance of ingester handover and WAL checkpoint by
reducing garbage.

Triggered by this memory allocation profile, taken mid-transfer:
```
  Total:      3.81GB     4.07GB (flat, cum) 20.56%
    349            .          .           } 
    350            .          .            
    351            .          .           // The passed wireChunks slice is for re-use. 
    352            .          .           func toWireChunks(descs []*desc, wireChunks []client.Chunk) ([]client.Chunk, error) { 
    353            .          .           	if cap(wireChunks) < len(descs) { 
    354         30MB       30MB           		wireChunks = make([]client.Chunk, 0, len(descs)) 
    355            .          .           	} 
    356            .          .           	wireChunks = wireChunks[:0] 
    357            .          .           	for _, d := range descs { 
    358            .          .           		wireChunk := client.Chunk{ 
    359            .          .           			StartTimestampMs: int64(d.FirstTime), 
    360            .          .           			EndTimestampMs:   int64(d.LastTime), 
    361            .          .           			Encoding:         int32(d.C.Encoding()), 
    362            .          .           		} 
    363            .          .            
    364       3.78GB     3.98GB           		buf := bytes.NewBuffer(make([]byte, 0, d.C.Size())) 
    365            .       64MB           		if err := d.C.Marshal(buf); err != nil { 
    366            .          .           			return nil, err 
    367            .          .           		} 
```

It looks like `QueryStream()` could also reuse its memory, but I didn't look too far into that.

**Checklist**
- NA Tests updated
- NA Documentation added
- NA `CHANGELOG.md` updated

### DIFF
--- a/pkg/ingester/transfer.go
+++ b/pkg/ingester/transfer.go
@@ -354,23 +354,29 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
 // The passed wireChunks slice is for re-use.
 func toWireChunks(descs []*desc, wireChunks []client.Chunk) ([]client.Chunk, error) {
 	if cap(wireChunks) < len(descs) {
-		wireChunks = make([]client.Chunk, 0, len(descs))
+		wireChunks = make([]client.Chunk, len(descs))
+	} else {
+		wireChunks = wireChunks[:len(descs)]
 	}
-	wireChunks = wireChunks[:0]
-	for _, d := range descs {
+	for i, d := range descs {
 		wireChunk := client.Chunk{
 			StartTimestampMs: int64(d.FirstTime),
 			EndTimestampMs:   int64(d.LastTime),
 			Encoding:         int32(d.C.Encoding()),
 		}
 
-		buf := bytes.NewBuffer(make([]byte, 0, d.C.Size()))
+		slice := wireChunks[i].Data[:0] // try to re-use the memory from last time
+		if cap(slice) < d.C.Size() {
+			slice = make([]byte, 0, d.C.Size())
+		}
+		buf := bytes.NewBuffer(slice)
+
 		if err := d.C.Marshal(buf); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/ingester/transfer.go
```diff
@@ -351,28 +351,36 @@ func (i *Ingester) TransferTSDB(stream client.Ingester_TransferTSDBServer) error
...
 }
 
-// The passed wireChunks slice is for re-use.
-func toWireChunks(descs []*desc, wireChunks []client.Chunk) ([]client.Chunk, error) {
+// The passed wireChunks and bufs slices are for re-use.
+func toWireChunks(descs []*desc, wireChunks []client.Chunk, bufs []bytes.Buffer) ([]client.Chunk, []bytes.Buffer, error) {
 	if cap(wireChunks) < len(descs) {
 		wireChunks = make([]client.Chunk, 0, len(descs))
 	}
+	if len(bufs) < len(descs) {
```
codesome, line 359: We can be consistent by having only one of `len` or `cap` to check `bufs` and `wireChunks` above.

bboreham: No, they are used in different ways.

codesome: Fair enough

bboreham:
I could simplify this by creating the `Buffer` from `wireChunk.Data`.

bboreham:
> It looks like `QueryStream()` could also reuse its memory

This would be much more complicated. There are two loops: one calling `toWireChunks` and building up a batch, and another sending out batches.  The memory can only be reused after the batch is sent.

I'll leave `QueryStream()` for another day.

bboreham:
I've re-done this so the diffs are a lot smaller, but it's a bit easier to mis-use.

PR #2506, opened 2020-04-22
Handle forward slashes in ruler mapper/manager

jtlisi:
**What this PR does**:

- Ensure the Ruler can properly handle rule Namespaces with special characters. E.G. `/+<>?`

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## master / unreleased
 
-* [CHANGE] Ruler: Better handling of special characters in filenames when mapping them onto the filesystem. #2506
 * [CHANGE] Added v1 API routes documented in #2327. #2372
   * Added `-http.alertmanager-http-prefix` flag which allows the configuration of the path where the Alertmanager API and UI can be reached. The default is set to `/alertmanager`.
   * Added `-http.prometheus-http-prefix` flag which allows the configuration of the path where the Prometheus API and UI can be reached. The default is set to `/prometheus`.
@@ -45,6 +44,7 @@
 * [ENHANCEMENT] Add de-duplicated chunks counter `cortex_chunk_store_deduped_chunks_total` which counts every chunk not sent to the store because it was already sent by another replica. #2485
 * [ENHANCEMENT] query-frontend now also logs the POST data of long queries. #2481
 * [ENHANCEMENT] Experimental WAL: Ingester WAL records now have type header and the custom WAL records have been replaced by Prometheus TSDB's WAL records. Old records will not be supported from 1.3 onwards. Note: once this is deployed, you cannot downgrade without data loss. #2436
+* [BUGFIX] Ruler: Ensure temporary rule files with special characters are properly mapped and cleaned up. #2506
 * [BUGFIX] Fixes #2411, Ensure requests are properly routed to the prometheus api embedded in the query if `-server.path-prefix` is set. #2372
 * [BUGFIX] Experimental TSDB: fixed chunk data corruption when querying back series using the experimental blocks storage. #2400
 * [BUGFIX] Cassandra Storage: Fix endpoint TLS host verification. #2109
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Ruler: Better handling of special characters in filenames when mapping them onto the filesystem. #2506
 * [CHANGE] Added v1 API routes documented in #2327. #2372
   * Added `-http.alertmanager-http-prefix` flag which allows the configuration of the path where the Alertmanager API and UI can be reached. The default is set to `/alertmanager`.
   * Added `-http.prometheus-http-prefix` flag which allows the configuration of the path where the Prometheus API and UI can be reached. The default is set to `/prometheus`.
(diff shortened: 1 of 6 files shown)


pkg/ruler/mapper.go
```diff
@@ -59,8 +63,16 @@ func (m *mapper) MapRules(user string, ruleConfigs map[string][]legacy_rulefmt.R
...
 
 	for _, existingFile := range existingFiles {
-		fullFileName := path + existingFile.Name()
-		ruleGroups := ruleConfigs[existingFile.Name()]
+		fullFileName := filepath.Join(path, existingFile.Name())
+
+		// Ensure the namespace is decoded from base64 to see if it is still required
+		decodedNamespace, err := base64.URLEncoding.DecodeString(existingFile.Name())
+		if err != nil {
+			level.Warn(m.logger).Log("msg", "unable to remove rule file on disk", "file", existingFile.Name(), "err", err)
```
pstibrany, line 71:
```suggestion
			level.Warn(m.logger).Log("msg", "unable to remove rule file on disk", "file", fullFileName.Name(), "err", err)
```

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 * [CHANGE] Experimental TSDB: Modified default values for `compactor.deletion-delay` option from 48h to 12h and `-experimental.tsdb.bucket-store.ignore-deletion-marks-delay` from 24h to 6h. #2414
 * [CHANGE] Experimental WAL: Default value of `-ingester.checkpoint-enabled` changed to `true`. #2416
 * [FEATURE] Ruler: The `-ruler.evaluation-delay` flag was added to allow users to configure a default evaluation delay for all rules in cortex. The default value is 0 which is the current behavior. #2423
+* [ENHANCEMENT] Ruler: Better handling of special characters in filenames when mapping them onto the filesystem. #2506
```
pracucci, line 22: Shouldn't this be considered a CHANGE, mentioning that when we store rules in the filesystem we now base64 encode the filename?

pkg/ruler/mapper.go
```diff
@@ -34,15 +36,17 @@ func (m *mapper) MapRules(user string, ruleConfigs map[string][]legacy_rulefmt.R
 	filenames := []string{}
 
 	// user rule files will be stored as `/<path>/<userid>/filename`
```
pracucci, line 38:
```suggestion
	// user rule files will be stored as `/<path>/<userid>/<encoded filename>`
```

pkg/ruler/mapper.go
```diff
@@ -34,15 +36,17 @@ func (m *mapper) MapRules(user string, ruleConfigs map[string][]legacy_rulefmt.R
...
 	err := m.FS.MkdirAll(path, 0777)
 	if err != nil {
 		return false, nil, err
 	}
 
 	// write all rule configs to disk
 	for filename, groups := range ruleConfigs {
-		fullFileName := path + filename
+		// Store the encoded file name to better handle `/` characters
+		encodedFileName := base64.URLEncoding.EncodeToString([]byte(filename))
```
pracucci, line 48: The main downside of this approach is that the encoded filename is unreadable and makes debugging harder. An alternative may be using `url.PathEscape()` (as far as `%` character is not a problem). Same examples:

- `Hello` > `Hello`
- `Hello world` > `Hello%20world`
- `a/b` > `a%2Fb`

pkg/ruler/mapper.go
jtlisi: These aren't rule files meant for the storage and persistence of rules. They are temporary files that are loaded by the upstream prometheus manager code for evaluation. They are essentially stateless so no upgrade path is required.

pkg/ruler/mapper.go
gouthamve: I like this better. This means the existing file names are same as well.

jtlisi: This would only work for POSIX, no windows. Will that be an issue?

pkg/ruler/mapper.go
jtlisi: 👍

CHANGELOG.md
jtlisi: Makes sense

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Ruler: Better handling of special characters in filenames when mapping them onto the filesystem. #2506
```
pracucci, line 5: Sorry to bother you on this, but I think that reading the changelog entry it's not clear why it's a CHANGE. Could you clarify it?

jtlisi: I need to give it more thought. I'm starting to think a changelog entry is not needed or if anything it should be labeled a BUGFIX since it aligns Cortex with expected behavior.

pkg/ruler/ruler.go
```diff
@@ -530,10 +530,17 @@ func (r *Ruler) getLocalRules(userID string) ([]*GroupStateDesc, error) {
 
 	for _, group := range groups {
 		interval := group.Interval()
+
+		// The mapped filename is base64 encoded to make handling `/` characters easier
```
pracucci, line 534: Please update this comment (it's not base64 anymore)

PR #2511, opened 2020-04-23
Fix querier HTTP middlewares

### LINKED ISSUE #2507: Querier HTTP metrics are not working with query frontend
After #2437, the querier no longer passes requests from the worker to the weaveworks/common server. This means none of the instrumentation/logging middleware is used on the request.

https://github.com/cortexproject/cortex/blob/fa0fc64235063d770d6aac223b11e016cf64fcf2/pkg/cortex/modules.go#L213

The two possible solutions I see are as follows:

1. Update the weaveworks common server to allow for more control over the instrumentation middleware it configures.

2. Use a separate set of metrics instrumented specifically for the queries the worker services.

pracucci:
**What this PR does**:
After #2437, the querier no longer passes requests from the worker to the weaveworks/common server. This means none of the instrumentation/logging middleware is used on the request.

This PR tries to fix it (I manually tested it both running Cortex both in microservices and single-binary mode). It's a **draft PR** because it needs:
1. Weaveworks common change https://github.com/weaveworks/common/pull/189 to be approved and merged, and `vendor/` updated
2. Integration tests to be updated

/cc @joe-elliott 

Fixes #2507

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -210,7 +210,7 @@ func (t *Cortex) initQuerier(cfg *Config) (serv services.Service, err error) {
 
 	// Query frontend worker will only be started after all its dependencies are started, not here.
 	// Worker may also be nil, if not configured, which is OK.
-	worker, err := frontend.NewWorker(cfg.Worker, httpgrpc_server.NewServer(handler), util.Logger)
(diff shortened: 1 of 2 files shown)


pstibrany:
Change looks good, but needs to be done upstream :(

pracucci:
> Change looks good, but needs to be done upstream :(

Yes. I opened a draft PR to give everything the context of the fix to better understand the changes here:
https://github.com/weaveworks/common/pull/189

PR #2515, opened 2020-04-23
Querier Metrics Fix

### LINKED ISSUE #2507: Querier HTTP metrics are not working with query frontend

joe-elliott:
**What this PR does**:
Returns the weaveworks/common server instead of the router. This restores the usage of weaveworks standard middleware/instrumentation in the querier.

Fixes #2507

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/development/tsdb-blocks-storage-s3-single-binary/docker-compose.yml
+++ b/development/tsdb-blocks-storage-s3-single-binary/docker-compose.yml
@@ -30,6 +30,7 @@ services:
     image: jaegertracing/all-in-one
     ports:
       - 16686:16686
+      - "14268"
 
   cortex-1:
     build:
@@ -41,8 +42,9 @@ services:
       - consul
       - minio
     environment:
-      - JAEGER_COLLECTOR_URL=http://jaeger:14268
-      - JAEGER_TAGS=cortex=1
+      - JAEGER_AGENT_HOST=jaeger
+      - JAEGER_AGENT_PORT=6831
+      - JAEGER_TAGS=app=cortex-1
       - JAEGER_SAMPLER_TYPE=const
       - JAEGER_SAMPLER_PARAM=1
     ports:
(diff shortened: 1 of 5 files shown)


joe-elliott:
Note that in single binary mode an internal router is being used.  Currently researching impact on single binary metrics/traces.

development/tsdb-blocks-storage-s3-single-binary/docker-compose.yml
```diff
@@ -50,6 +60,11 @@ services:
     depends_on:
       - consul
       - minio
+    environment:
+      - JAEGER_COLLECTOR_URL=http://jaeger:14268
+      - JAEGER_TAGS=cortex=1
```
pracucci, line 65:
```suggestion
      - JAEGER_TAGS=cortex=2
```

joe-elliott:
After a quick review in single binary mode:

- Metrics are fine
  - `cortex_request_duration_seconds_count` is present for read paths
- Trace context is broken as suspected by @jtlisi.  Will submit a fix.

joe-elliott:
Ok, I consider this ready for review.

*Single Binary*
- Metrics and Logs are still handled by the primary http server.  We don't want to replicate this on the internal router or we'd log/metric every call twice.
- Appropriate middleware was added to wrap the internal router.  This will extract the trace context from the http headers and inject it into the Go context.  The span name was hardcoded as `internalQuerier`.

Trace:
image

Metrics:
image

*Microservices*
- Original HTTPServer (instead of just router) is returned to restore metrics on the querier.

Trace:
image

Metrics:
image

Note that this branch has the same commits as #2516.  It was merged into this branch to allow for local tracing.

pstibrany:
Would it be easy to have a test for this?

pstibrany:
Does the same problem exist for user ID in the context?

pstibrany:
> Does the same problem exist for user ID in the context?

Since user ID is re-extracted from headers by auth middleware, this seems to not be a problem.

PR #2518, opened 2020-04-23
Add traceID if possible when using context logger.

cyriltovena:
This change the behaviour of the context logger, there is situation where we don't have an org_id and we would still want to log with the spanID.

I'm also renaming `span_id` to `spanID` because we have inconsistency  with weaveworks server which uses `spanID`. Wasn't sure if this would be welcomed change in weaveworks but I'm can send the PR there.

### DIFF
--- a/pkg/util/log.go
+++ b/pkg/util/log.go
@@ -103,10 +103,12 @@ func WithContext(ctx context.Context, l log.Logger) log.Logger {
 	// Weaveworks uses "orgs" and "orgID" to represent Cortex users,
 	// even though the code-base generally uses `userID` to refer to the same thing.
 	userID, err := user.ExtractOrgID(ctx)
-	if err != nil {
+	if err != nil && err != user.ErrNoOrgID {
 		return l
 	}
-	l = WithUserID(userID, l)
+	if err == nil {
+		l = WithUserID(userID, l)
+	}
 
 	traceID, ok := middleware.ExtractTraceID(ctx)
 	if !ok {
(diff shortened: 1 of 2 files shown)


cyriltovena:
Need rebase on https://github.com/cortexproject/cortex/pull/2521 once merged.

pracucci:
I've pushed a commit to rebase `master`.

pkg/util/log.go
```diff
@@ -127,7 +129,7 @@ func WithUserID(userID string, l log.Logger) log.Logger {
 // its details.
 func WithTraceID(traceID string, l log.Logger) log.Logger {
 	// See note in WithContext.
-	return log.With(l, "trace_id", traceID)
+	return log.With(l, "traceID", traceID)
```
pracucci, line 132: > I'm also renaming span_id to spanID because we have inconsistency with weaveworks server which uses spanID.

Note to myself: I guess you meant `traceID` and not `spanID`.

pkg/util/log.go
```diff
@@ -103,10 +103,12 @@ func WithContext(ctx context.Context, l log.Logger) log.Logger {
...
 	// even though the code-base generally uses `userID` to refer to the same thing.
 	userID, err := user.ExtractOrgID(ctx)
-	if err != nil {
+	if err != nil && err != user.ErrNoOrgID {
 		return l
 	}
-	l = WithUserID(userID, l)
+	if err == nil {
+		l = WithUserID(userID, l)
+	}
```
pracucci, line 111: Can't we simplify it as the following, so that the trace ID is injected in any case?

```
if err == nil {
    l = WithUserID(userID, l)
}
```

pstibrany:
> I'm also renaming `span_id` to `spanID` because we have inconsistency with weaveworks server which uses `spanID`. Wasn't sure if this would be welcomed change in weaveworks but I'm can send the PR there.

For reference: https://github.com/cortexproject/cortex/blob/183b65a81d85d3e7cd1d31135ac1a105c4544211/vendor/github.com/weaveworks/common/middleware/logging.go#L19

pkg/util/log.go
pstibrany: Agree, let's do that.

PR #2519, opened 2020-04-23
Fix logger context when creating a spanLogger from context.

cyriltovena:
I made this function a while ago to avoid creating a span but still keep logging to a span.

However I just realized this won't log the org id nor the trace id as the original SpanLogger.New


cyriltovena:
Need rebase on https://github.com/cortexproject/cortex/pull/2521 once merged.

PR #2521, opened 2020-04-23
Fix a failing test because of a merge conflict.

cyriltovena:
Introduced by https://github.com/cortexproject/cortex/pull/2459 via a conflict on master.


pracucci:
Thanks! Closing just because we had another PR #2517 fixing the same but wasn't merge earlier because missing an approval. Thanks again!

PR #2522, opened 2020-04-24
Basic functional options support to metrics assertion in integration tests

### LINKED ISSUE #2975: Integration Tests: Add Assertion for waiting on a metric to appear
As part of our integration tests framework, we have some form of metric-based assertion. At the moment, is it not capable of waiting on a metric.  When the metric is not there as it fetches `/metrics` it will bail out saying that the metric is not present even though the metric could eventually be there after a certain async action completes. 

In https://github.com/cortexproject/cortex/pull/2960/files#r464597389 we discussed about adding an option to the `WaitMetric` method that allows us to wait on a metric as a solution to this problem.

https://github.com/cortexproject/cortex/pull/2522 provides a foundation for this work given the ability to wait on a condition could be added as an option.

pracucci:
**What this PR does**:
In #2515 (by Joe) I wanted to assert on `cortex_request_duration_seconds` in the integration tests, but I wasn't able because unable to assert on a histogram's count value.

In this PR I'm doing **baby steps** do add functional options support to metrics assection in integration tests. Could be improved in many ways (and I will do in subsequent PRs if this one is approved), but in this PR I would prefer to keep the change set as smallest as possible to validate the basic design.

/cc @bwplotka 

Fixes #2975.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/alertmanager_test.go
+++ b/integration/alertmanager_test.go
@@ -5,8 +5,8 @@ package main
 import (
 	"context"
 	"testing"
-	"time"
 
+	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/stretchr/testify/require"
 
 	"github.com/cortexproject/cortex/integration/e2e"
@@ -80,7 +80,9 @@ func TestAlertmanagerStoreAPI(t *testing.T) {
 	err = c.SetAlertmanagerConfig(context.Background(), cortexAlertmanagerUserConfigYaml, map[string]string{})
 	require.NoError(t, err)
 
-	require.NoError(t, am.WaitSumMetricsWithOptions(e2e.Equals(0), []string{"cortex_alertmanager_config_invalid"}, e2e.WaitMissingMetrics))
+	require.NoError(t, am.WaitSumMetricsWithOptions(e2e.Equals(0), []string{"cortex_alertmanager_config_invalid"},
+		e2e.WithLabelMatchers(labels.MustNewMatcher(labels.MatchEqual, "user", "user-1")),
+		e2e.WaitMissingMetrics))
 
 	cfg, err := c.GetAlertmanagerConfig(context.Background())
 	require.NoError(t, err)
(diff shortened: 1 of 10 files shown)


pstibrany:
This feels bit complicated, but if we don't find a simpler way, I'm fine with it. Each time I need to write new metric-based condition, I need to find out how.

As an alternative, what do you think about a possibility to use PromQL expressions for checking these conditions?

integration/e2e/composite_service.go
```diff
@@ -61,11 +61,11 @@ func (s *CompositeHTTPService) WaitSumMetrics(isExpected func(sums ...float64) b
 }
 
 // SumMetrics returns the sum of the values of each given metric names.
-func (s *CompositeHTTPService) SumMetrics(metricNames ...string) ([]float64, error) {
+func (s *CompositeHTTPService) SumMetrics(metricNames []string, opts ...MetricsOption) ([]float64, error) {
```
bwplotka, line 64: I get the use case, plus, sometimes you really want only certain label, so we could have `WithLabelEquals(name, value string)` 

However what do you think about simplifying interfaces and having just single metric name? We can save on `[]string{` boilerplate :thinking:

pracucci:
_We decided to move forward with this, but haven't got the time to do it yet._

integration/e2e/composite_service.go
pracucci: > However what do you think about simplifying interfaces and having just single metric name? We can save on []string{ boilerplate

We could, but then the testers `*AmongTwo` wouldn't work anymore. Cortex is not using it, but Thanos does. Thoughts?

pracucci: > so we could have WithLabelEquals(name, value string)

Agree, done.

integration/alertmanager_test.go
```diff
@@ -80,8 +80,9 @@ func TestAlertmanagerStoreAPI(t *testing.T) {
 	err = c.SetAlertmanagerConfig(context.Background(), cortexAlertmanagerUserConfigYaml, map[string]string{})
 	require.NoError(t, err)
 
-	time.Sleep(2 * time.Second)
-	require.NoError(t, am.WaitSumMetrics(e2e.Equals(0), "cortex_alertmanager_config_invalid"))
+	require.NoError(t, am.WaitSumMetricsWithOptions(e2e.Equals(0), []string{"cortex_alertmanager_config_invalid"},
```
gotjosh, line 83: Thanks!

PR #2523, opened 2020-04-24
Clarify store-gateway ring config

pracucci:
**What this PR does**:
My initial plan was to move the store-gateway ring config to the generic `tsdb.Config`. However, after I did it, I did sound weird too, because most of the ring config is actually required only by the store-gateway.

For this reason, I've picked a different path. The querier still needs a bit of store-gateway config but I've tried to clarify it in the config file documentation.

What do you think?

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/tools/doc-generator/parser.go
+++ b/tools/doc-generator/parser.go
@@ -150,7 +150,7 @@ func parseConfig(block *configBlock, cfg interface{}, flags map[uintptr]*flag.Fl
 					blockDesc = rootDesc
 				} else {
 					blockName = fieldName
-					blockDesc = ""
+					blockDesc = getFieldDescription(field, "")
 				}
 
 				subBlock = &configBlock{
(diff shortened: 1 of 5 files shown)


tools/doc-generator/parser.go
```diff
@@ -391,3 +399,36 @@ func isRootBlock(t reflect.Type) (string, string, bool) {
...
+
+func parseDocTag(f reflect.StructField) map[string]string {
+	cfg := map[string]string{}
+	tag := f.Tag.Get("doc")
+
+	if tag == "" {
+		return cfg
+	}
+
+	for _, entry := range strings.Split(tag, "|") {
```
pracucci, line 422: The idea is to use the same pattern as `flagarize` from Bartek, so:

```
field type `doc:"required|description=Put whatever you want here including commas , and so on."`
```

PR #2524, opened 2020-04-24
Added PromQL evaluation on top of returned metrics.

pstibrany:
This can simplify testing for metrics as shown by alertmanager and query frontend test.

This is an alternative to proposal in #2522. 

/cc @bwplotka 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/metrics_promql.go
+++ b/integration/e2e/metrics_promql.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -172,7 +173,7 @@ func (sq *samplesQueryable) Querier(ctx context.Context, mint, maxt int64) (stor
 	return sq, nil
 }
 
-func (sq samplesQueryable) SelectSorted(params *storage.SelectParams, matcher ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
+func (sq samplesQueryable) SelectSorted(_ *storage.SelectParams, matcher ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
 	var series []storage.Series
 
 	for _, s := range sq {
@@ -181,7 +182,8 @@ func (sq samplesQueryable) SelectSorted(params *storage.SelectParams, matcher ..
 		}
 	}
 
-	return &seriesSet{series: series}, nil, nil
+	sort.Sort(byLabels(series))
+	return &seriesSet{cur: -1, series: series}, nil, nil
 }
 
 func (sq samplesQueryable) Select(params *storage.SelectParams, matcher ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
@@ -208,11 +210,11 @@ type seriesSet struct {
 
 func (c *seriesSet) Next() bool {
 	c.cur++
-	return c.cur-1 < len(c.series)
+	return c.cur < len(c.series)
 }
 
 func (c *seriesSet) At() storage.Series {
-	return c.series[c.cur-1]
+	return c.series[c.cur]
 }
 
 func (c *seriesSet) Err() error {
@@ -281,3 +283,9 @@ func (c *singleSampleIterator) Next() bool {
 func (c *singleSampleIterator) Err() error {
 	return nil
 }
+
+type byLabels []storage.Series
+
+func (b byLabels) Len() int           { return len(b) }
+func (b byLabels) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
+func (b byLabels) Less(i, j int) bool { return labels.Compare(b[i].Labels(), b[j].Labels()) < 0 }
--- a/integration/e2e/metrics_promql.go
+++ b/integration/e2e/metrics_promql.go
@@ -14,6 +14,7 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb/chunkenc"
 )
 
 var (
@@ -171,21 +172,17 @@ func (sq samplesQueryable) Querier(ctx context.Context, mint, maxt int64) (stora
 	return sq, nil
 }
 
-func (sq samplesQueryable) SelectSorted(_ *storage.SelectParams, matcher ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
+func (sq samplesQueryable) Select(_ bool, _ *storage.SelectHints, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
 	var series []storage.Series
 
 	for _, s := range sq {
-		if s.matches(matcher...) {
+		if s.matches(matchers...) {
 			series = append(series, s)
 		}
 	}
 
 	sort.Sort(byLabels(series))
 	return &seriesSet{cur: -1, series: series}, nil, nil
-}
-
-func (sq samplesQueryable) Select(params *storage.SelectParams, matcher ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
-	return sq.SelectSorted(params, matcher...)
 }
 
 func (sq samplesQueryable) LabelValues(name string) ([]string, storage.Warnings, error) {
@@ -236,7 +233,7 @@ func (c sample) Labels() labels.Labels {
 	return b.Labels()
 }
 
-func (c sample) Iterator() storage.SeriesIterator {
+func (c sample) Iterator() chunkenc.Iterator {
 	return newSampleIterator(c.ts, c.val)
 }
 
@@ -250,7 +247,7 @@ func (c sample) matches(matcher ...*labels.Matcher) bool {
 	return true
 }
 
-var _ storage.SeriesIterator = &singleSampleIterator{}
+var _ chunkenc.Iterator = &singleSampleIterator{}
 
 type singleSampleIterator struct {
 	nextCalled bool
@@ -258,7 +255,7 @@ type singleSampleIterator struct {
 	val        float64
 }
 
-func newSampleIterator(ts int64, val float64) storage.SeriesIterator {
+func newSampleIterator(ts int64, val float64) chunkenc.Iterator {
 	return &singleSampleIterator{
 		ts:  ts,
 		val: val,
--- a/integration/e2e/metrics_promql.go
+++ b/integration/e2e/metrics_promql.go
@@ -258,10 +258,9 @@ func newSampleIterator(ts int64, val float64) storage.SeriesIterator {
 
 // Seek implements storage.SeriesIterator.
 func (c *singleSampleIterator) Seek(t int64) bool {
-	if t <= c.ts {
-		return true
-	}
-	return false
+	// If requested timestamp is before what we have, we can "advance".
+	// If it is past it, we cannot.
+	return t <= c.ts
 }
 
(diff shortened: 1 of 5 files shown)


integration/query_frontend_test.go
```diff
@@ -161,6 +161,9 @@ func runQueryFrontendTest(t *testing.T, setup queryFrontendSetup) {
 
 	require.NoError(t, queryFrontend.WaitSumMetrics(e2e.Equals(numUsers*numQueriesPerUser), "cortex_query_frontend_queries_total"))
 
+	require.NoError(t, queryFrontend.WaitForPromQL(fmt.Sprintf("sum(cortex_request_duration_seconds_count) > %d", numUsers*numQueriesPerUser)))
+	require.NoError(t, querier.WaitForPromQL(fmt.Sprintf("sum(cortex_request_duration_seconds_count) > %d", numUsers*numQueriesPerUser)))
```
pracucci, line 165: Now that is way more powerful you can check for equality (`==` instead of `>`) filtering by the label `{route=''<query route>"}` where `<query route>` is the API v1 query route (don't remember the exact value in the label).

integration/e2e/service.go
```diff
@@ -571,6 +571,32 @@ func (s *HTTPService) SumMetrics(metricNames ...string) ([]float64, error) {
...
+
+		if val > 0 {
+			return nil
+		}
+
+		lastVal = val
+		s.retryBackoff.Wait()
+	}
+
+	return fmt.Errorf("expression %q did not evaluate to positive number, last value: %g", expr, lastVal)
```
pracucci, line 597: Last value is always zero right?

integration/e2e/metrics_promql.go
```diff
@@ -0,0 +1,283 @@
...
+
+	now := time.Now()
+
+	samples := convertMetricFamiliesToSamples(int64(model.TimeFromUnixNano(now.UnixNano())), families)
+
+	eng := promql.NewEngine(promql.EngineOpts{
+		Logger:             nil,
+		Reg:                nil,
+		MaxSamples:         1e6,
+		Timeout:            time.Hour, // time.Second,
```
pracucci, line 38: There's some commented code here. I guess few seconds timeout is enough.

integration/e2e/metrics_promql.go
```diff
@@ -0,0 +1,283 @@
...
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/storage"
```
pracucci, line 15: Have you measured if there's any significative negative impact in integration tests compilation time due to these imports?

integration/e2e/metrics_promql.go
```diff
@@ -0,0 +1,283 @@
...
+	series []storage.Series
+}
+
+func (c *seriesSet) Next() bool {
+	c.cur++
+	return c.cur-1 < len(c.series)
+}
+
+func (c *seriesSet) At() storage.Series {
+	return c.series[c.cur-1]
```
pracucci, line 215: You could remove the `-1` which is tricky to read just initializing the `seriesSet{cur : -1}`

integration/e2e/metrics_promql.go
pstibrany: Leftover from manually stepping through the code and hitting timeouts 😄

integration/e2e/metrics_promql.go
pstibrany: I expect that compilation time is small compared to actually running these tests, but I haven't timed it.

bwplotka: We import those anyway.

integration/e2e/metrics_promql.go
```diff
@@ -0,0 +1,283 @@
...
+		result[labels.BucketLabel] = fmt.Sprintf("%g", upperBound)
+	}
+
+	for _, p := range pairs {
+		result[p.GetName()] = p.GetValue()
+	}
+	return result
+}
+
+// implements Queryable and Querier
```
bwplotka, line 169: Instead of this comment (which is not proper full sentence), maybe it's better document by code?

```
var (
    _ Queryable = samplesQueryable{}
    _ Querier = samplesQueryable{}
)
```

integration/e2e/metrics_promql.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (sq *samplesQueryable) LabelNames() ([]string, storage.Warnings, error) {
+	panic("implement me")
+}
+
+func (sq *samplesQueryable) Close() error {
+	return nil
+}
+
+// seriesSet implements storage.SeriesSet.
```
bwplotka, line 205: ditto

integration/e2e/metrics_promql.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (c *seriesSet) At() storage.Series {
+	return c.series[c.cur-1]
+}
+
+func (c *seriesSet) Err() error {
+	return nil
+}
+
+// sample implements storage.Series
```
bwplotka, line 224: same

integration/e2e/metrics_promql.go
```diff
@@ -0,0 +1,283 @@
...
+	for _, m := range matcher {
+		if !m.Matches(c.labels[m.Name]) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// concreteSeriesIterator implements storage.SeriesIterator.
```
bwplotka, line 247: wrong comment (should fixed and either full sentence or removed and you can add validating that)

integration/e2e/metrics_promql_test.go
```diff
@@ -0,0 +1,106 @@
...
+		expr string
+		val  float64
+		err  error
+	}{
+		{
+			expr: "sum(blocks_index_cache_items_evicted_total)",
+			val:  15984,
+		},
+		{
+			expr: "sum({__name__=~'blocks_index_cache_items_evicted_.*'})",
```
bwplotka, line 60: How `rate` `increase` and 20 others will behave? 

https://prometheus.io/docs/prometheus/latest/querying/functions/#functions

Should we define some behaviour for those (maybe in next PR)

integration/e2e/metrics_promql_test.go
```diff
@@ -0,0 +1,106 @@
...
+			expr: "blocks_index_cache_requests_total{item_type!='Series'}",
+			val:  15984,
+		},
+		{
+			expr: "count(blocks_index_cache_hits_total)",
+			val:  2,
+		},
+		{
+			expr: "blocks_index_cache_items > 50000",
+			val:  53280, // returns blocks_index_cache_items{item_type="Series"} 53280
```
bwplotka, line 77: not a full sentence

integration/e2e/service.go
```diff
@@ -571,6 +571,32 @@ func (s *HTTPService) SumMetrics(metricNames ...string) ([]float64, error) {
 	return sums, nil
 }
 
+// Waits until promQL expression returns scalar (or vector with single value) with positive value.
+func (s *HTTPService) WaitForPromQL(matcher func(sums ...float64) bool, expr string) error {
```
bwplotka, line 575: Not sure if the name of method is readable .. to me it reads like we wait for some evaluation only or somethig. 

What about `WaitForPromQLNonEmptyResult` :thinking: 
Not blocking, I get that we want short name as well.

integration/e2e/metrics_promql.go
pstibrany: I copied this code from somewhere, without much thinking. Turns out, I actually forgot to sort series!

We could also use `github.com/cortexproject/cortex/pkg/querier/series.NewConcreteSeriesSet` method, but that introduces dependency to cortex package.

integration/e2e/service.go
pstibrany: We now use matchers to compare returned value, so lastValue is useful again.

pstibrany:
> The user experience is much nicer though, just I would love to whitelist functions we support or something.. Otherwise LGTM 🤗

I'm not sure how I would do that :) I'm also sure it's necessary. Range vectors obviously don't work and functions requiring them don't work either.

> Is there a plan to remove old way of doing stuff? Because we rely on some behavior not supported by the function here. E.g GetValueAndCompare like logic: https://github.com/thanos-io/thanos/blob/master/test/e2e/rule_test.go#L349

I don't think we will remove `WaitSumMetrics` anytime soon, and surely not without your approval.

integration/e2e/metrics_promql_test.go
pstibrany: I will add them to unit test.

bwplotka:
LGTM :+1: 

Awesome work!

> I don't think we will remove WaitSumMetrics anytime soon, and surely not without your approval.

We should remove, otherwise, we have too wide API.. (too many things doing same thing), but maybe we can enable my use case with separate smaller API)

pracucci:
Since Prometheus upgrade path is Thanos > Cortex, and Thanos depends on Cortex for the e2e tests, and Cortex can't upgrade Prometheus before Thanos, I think that now that Prometheus is imported in e2e tests we may end up in a deadlock situation when Thanos needs to upgrade Prometheus and there are some code breaking changes affecting Cortex e2e tests.

@bwplotka @pstibrany WDYT?

bwplotka:
Hah, wow... might be true... Quite insane indeed. We plan to depend on query range bits so we might be not able to avoid that...

I think for now.. until at some point we will start shared module or common repo we will need to live with it / create some tooling and docs. We can totally have tmp fork of Cortex and point to that which breaks ant circular dep because it's now different module etc.. (:

The value is to great to be afraid of those, let's try and see the challenges. We can always revert this if needed.

pstibrany:
> Since Prometheus upgrade path is Thanos > Cortex, and Thanos depends on Cortex for the e2e tests, and Cortex can't upgrade Prometheus before Thanos, I think that now that Prometheus is imported in e2e tests we may end up in a deadlock situation when Thanos needs to upgrade Prometheus and there are some code breaking changes affecting Cortex e2e tests.

This is a good point, and actually quite worrisome. There may be a way out of this by using copy of `promql` package in Cortex (there is already such copy), but I need to take a closer look if it would be possible to use it.

bwplotka:
Let's make sure we don't copy anything, let's merge and let's experience possible problems first, then we can start mitigating (:

pstibrany:
Unfortunately, this cannot be part of `integration/e2e` package. We have seen just recently how breaking change in Prometheus would make upgrade of Thanos and Cortex impossible.

I think we need to move this to internal Cortex package, and use this package from Cortex tests only.
Thanos can make a copy of the code, and use it for own integration tests.

This is not great, but I'm not sure we have another option.

pstibrany:
Alternatively, it could be included directly in Prometheus, if there was interest.

pracucci:
> I think we need to move this to internal Cortex package, and use this package from Cortex tests only.

I'm not a big fan of it. Do we really need this PR? I know, it's very nice, but until we don't have a nice way to handle the vendored Prometheus maybe we can just avoid to merge it? Doesn't look essential and we also have the less power option https://github.com/cortexproject/cortex/pull/2522.

pstibrany:
> I'm not a big fan of it. Do we really need this PR? I know, it's very nice, but until we don't have a nice way to handle the vendored Prometheus maybe we can just avoid to merge it? Doesn't look essential and we also have the less power option #2522.

That is probably the best way forward for now.

PR #2526, opened 2020-04-24
Added auto-forget to store-gateway ring instances

pracucci:
**What this PR does**:
In this PR I've introduced the support to auto-forget a store-gateway instance in the ring, 10x after the heartbeat timeout period is expired.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/basic_lifecycler_delegates.go
+++ b/pkg/ring/basic_lifecycler_delegates.go
@@ -138,7 +138,7 @@ func (d *AutoForgetDelegate) OnRingInstanceHeartbeat(lifecycler *BasicLifecycler
 		lastHeartbeat := time.Unix(instance.GetTimestamp(), 0)
 
 		if time.Since(lastHeartbeat) > d.forgetPeriod {
-			level.Warn(d.logger).Log("msg", "auto-forgetting instance from the ring because unhealthy since a long time", "instance", id, "last_heartbeat", lastHeartbeat.String(), "forget_period", d.forgetPeriod)
+			level.Warn(d.logger).Log("msg", "auto-forgetting instance from the ring because it is unhealthy for a long time", "instance", id, "last_heartbeat", lastHeartbeat.String(), "forget_period", d.forgetPeriod)
 			ringDesc.RemoveIngester(id)
 		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/ring/basic_lifecycler_delegates.go
```diff
@@ -96,3 +100,48 @@ func (d *TokensPersistencyDelegate) OnRingInstanceTokens(lifecycler *BasicLifecy
...
+	d.next.OnRingInstanceStopping(lifecycler)
+}
+
+func (d *AutoForgetDelegate) OnRingInstanceHeartbeat(lifecycler *BasicLifecycler, ringDesc *Desc, instanceDesc *IngesterDesc) {
+	for id, instance := range ringDesc.Ingesters {
+		lastHeartbeat := time.Unix(instance.GetTimestamp(), 0)
+
+		if time.Since(lastHeartbeat) > d.forgetPeriod {
+			level.Warn(d.logger).Log("msg", "auto-forgetting instance from the ring because unhealthy since a long time", "instance", id, "last_heartbeat", lastHeartbeat.String(), "forget_period", d.forgetPeriod)
+			ringDesc.RemoveIngester(id)
```
pstibrany, line 142:
```suggestion
			level.Warn(d.logger).Log("msg", "auto-forgetting instance from the ring because it is unhealthy for a long time", "instance", id, "last_heartbeat", lastHeartbeat.String(), "forget_period", d.forgetPeriod)
			ringDesc.RemoveIngester(id)
```

pracucci:
> Is it worth a mention in changelog or some documentation?

Yes, but I haven't added any changelog entry or doc for the whole store-gateway yet. My plan is to submit a dedicated PR to update the whole doc (and add a changelog entry too).

pracucci:
> I'm slightly uneasy as it is a departure from our usual workflow of investigating things if they leave the ring uncleanly. But as this is only on the read path, I am okay with it.

I agree on the write path. The main rationale here is that given a store-gateway is not really stateful (if you loose a store-gateway you don't loose any data, but you may get some queries failing) it's preferable a auto-recovery in case multiple nodes have failed instead of waiting for an operator to take manual actions.

PR #2531, opened 2020-04-27
Better spread of checkpointing

codesome:
**What this PR does**:

* Make the spacing between 2 series writes to be 95% of what it was before. I have seen checkpoints taking 3-5% more time than the checkpoint duration, so reducing the sleep time between 2 series to have checkpoint duration around the configured one.
* Using `time.After` instead of a ticker to ensure spacing between 2 series writes, else the ticker could be holding up an additional tick (micro-opt, but it also makes the code simpler).
* Batching up the checkpoint records upto 1 MiB to reduce the disk I/Ops (main aim of this PR).

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/wal.go
+++ b/pkg/ingester/wal.go
@@ -324,31 +324,51 @@ func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
 		return nil
 	}
 
-	var ticker *time.Ticker
-	if !immediate {
-		perSeriesDuration := w.cfg.CheckpointDuration / time.Duration(numSeries)
-		ticker = time.NewTicker(perSeriesDuration)
-		defer ticker.Stop()
-	}
+	perSeriesDuration := (95 * w.cfg.CheckpointDuration) / (100 * time.Duration(numSeries))
 
 	var wireChunkBuf []client.Chunk
-	b := make([]byte, 0, 1024)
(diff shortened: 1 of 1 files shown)


pkg/ingester/wal.go
```diff
@@ -324,33 +324,53 @@ func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
...
 				select {
-				case <-ticker.C:
+				case <-time.After(perSeriesDuration):
 				case <-w.quit: // When we're trying to shutdown, finish the checkpoint as fast as possible.
 				}
 			}
 		}
 	}
 
+	if err := checkpoint.Log(records...); err != nil {
```
pstibrany, line 370: There may be no records at this point.

codesome:
> I am not sure I understand why sleeping is there in the first place.

To make the checkpoint writing spread over the entire checkpoint duration. If otherwise, there is a burst in disk IO when checkpoint happens which affects WAL writes too, hence causing latency spikes.

pkg/ingester/wal.go
codesome: Yes, but there maybe be records too. By doing like this we skip a `if` check as 0 records would make it a noop.

PR #2532, opened 2020-04-27
Avoid sharding sub-aggregations in `querier.parallelise-shardable-queries`

owen-d:
## What
This PR fixes and issue where the groupings of sub-aggregations can cause query correctness issues. This only occurs  within a sharded sum aggregation (via the `querier.parallelise-shardable-queries` flag). This is because aggregations expose the `by (<labels...>)` syntax. Since the sharding optimizations [alter](https://github.com/cortexproject/cortex/blob/master/pkg/querier/astmapper/shard_summer.go#L123-L171) `sum` groupings,  we must be careful not to shard in such a way that will cause non-associative series merging due to `by` groupings. For an example, please see these [two new test cases](https://github.com/cortexproject/cortex/compare/master...owen-d:sharding-subaggregations?expand=1#diff-bef1fd1e4cc0b9ab1f9c4f205d7f2ca3R242-R281) which illustrate this.

## Further work
There may be further optimizations that can be applied ([example](https://github.com/cortexproject/cortex/compare/master...owen-d:sharding-subaggregations?expand=1#diff-bef1fd1e4cc0b9ab1f9c4f205d7f2ca3R282-R311)), but this PR addresses the current faulty behavior by avoiding trying to shard sum queries which wrap other aggregations.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/querier/astmapper/parallel.go
+++ b/pkg/querier/astmapper/parallel.go
@@ -42,7 +42,17 @@ func CanParallelize(node promql.Node) bool {
 
 	case *promql.AggregateExpr:
 		_, ok := summableAggregates[n.Op]
-		return ok && CanParallelize(n.Expr)
+		if !ok {
+			return ok
+		}
+
+		// Ensure there are no nested aggregations
+		nestedAggs, err := Predicate(n.Expr, func(node promql.Node) (bool, error) {
+			_, ok := node.(*promql.AggregateExpr)
+			return ok, nil
+		})
+
+		return err == nil && !nestedAggs && CanParallelize(n.Expr)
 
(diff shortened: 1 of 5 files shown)


pkg/querier/astmapper/parallel.go
```diff
@@ -42,7 +42,17 @@ func CanParallelize(node promql.Node) bool {
 
 	case *promql.AggregateExpr:
 		_, ok := summableAggregates[n.Op]
-		return ok && CanParallelize(n.Expr)
+		if !ok {
+			return ok
```
pracucci, line 46: Just for the sake of clarity
```suggestion
			return false
```

PR #2535, opened 2020-04-28
Update weaveworks/common dependency version

annanay25:

**What this PR does**:
This PR is a pre-req for #2502, since it got too big for a review.
The only changes in this PR are a vendor dependency update:

```
github.com/weaveworks/common v0.0.0-20200310113808-2708ba4e60a4 -> github.com/weaveworks/common v0.0.0-20200422083114-5790e7482ff8
```

Part of #2350 

**Checklist**
- NA - Tests updated
- [x] Documentation added

**What changed**:
- `weaveworks/common`: TLS config options now available for the server.
- `aws-sdk-go`: No breaking changes to components used by cortex.
- `go-kit/log`: Turn off HTML escaping in json logger.
- `golang/protobuf`: Interesting change to allow for client-side load-balancing for grpc.
- `gorilla/mux`: Improve CORS Method Middleware. We must now specify an OPTIONS method matcher for the middleware to set CORS headers.
- `go.etcd.io/etcd`: 
  - clientv3: Upgrade to round robin balancer based on gRPC 1.12 balancer API. We could use this in the clients to resolved DNS addresses. 
- `golang.google.org/grpc`: No major breaking changes.
- `gopkg.in/yaml.v2`: No major changes.


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * [CHANGE] `trace_id` field in log files has been renamed to `traceID`. #2518
 * [CHANGE] Slow query log has a different output now. Previously used `url` field has been replaced with `host` and `path`, and query parameters are logged as individual log fields with `qs_` prefix. #2520
 * [CHANGE] Experimental WAL: WAL and checkpoint compression is now disabled. #2436
+* [CHANGE] Updated weaveworks/common dependency to `v0.0.0-20200422083114-5790e7482ff8`. The latest changes in weaveworks/common/server allow to speficy TLS options for the server. #2535
```
pracucci, line 24: We usually don't add any entry in the changelog for any change which is not user-facing. I would suggest to remove this.

annanay25:
Thanks for pointing that out @pracucci. The [docs pointed out](https://github.com/etcd-io/etcd/blob/master/Documentation/upgrades/upgrade_3_4.md#require-grpcwithblock-for-client-dial) the need for a new `DiapOption` for the grpc client connecting to etcd: `grpc.WithBlock()`

I've updated the PR with this new option.

docs/configuration/config-file-reference.md
```diff
@@ -183,6 +183,40 @@ The `server_config` configures the HTTP and gRPC server of the launched service(
 # CLI flag: -server.grpc-conn-limit
 [grpc_listen_conn_limit: <int> | default = 0]
 
+http_tls_config:
```
jtlisi, line 186: Is this file supposed to be updated as part of this PR

pracucci: Yes, because of the CI linter.

jtlisi: Ahh, this was added by the weaveworks common server. Not a remnant of the other https client PR.

annanay25:
I suggest we hold merging this until https://github.com/weaveworks/common/pull/190 is merged

I'll re-update the dependencies :)

pkg/ring/kv/etcd/etcd.go
```diff
@@ -41,6 +42,7 @@ func New(cfg Config, codec codec.Codec) (*Client, error) {
 	cli, err := clientv3.New(clientv3.Config{
 		Endpoints:   cfg.Endpoints,
 		DialTimeout: cfg.DialTimeout,
+		DialOptions: []grpc.DialOption{grpc.WithBlock()},
```
pracucci, line 45: I'm not sure adding this is correct. The etcd client we have in master already has the new client balancer merged into (https://github.com/etcd-io/etcd/pull/9860). To preserve the current logic we should **not** add this. Have you found out anything different?

bboreham:
I think your list of "what changed" should highlight the addition of TLS options.

Also I like to know there are no subtle changes, as well as no breaking changes.
I read through all the diffs, apart from etcd, and the only thing that caught my eye was [turning off html escaping in logs](https://github.com/cortexproject/cortex/pull/2535/files#diff-d99918994387d9bc34564bc64cc9437fR35).

pkg/ring/kv/etcd/etcd.go
annanay25: My bad, you're right, this change was made in June 2018, I mixed it up.

annanay25:
> I read through all the diffs, apart from etcd, and the only thing that caught my eye was turning off html escaping in logs.

@bboreham - I could not find instances of unescaped HTML logging within the repo, so I do not think this should affect us. Is there something specific that you are concerned about?

bboreham:
I think you misunderstand (or maybe I do).  Any time the code was logging `&`, that would come out as `&amp` before, and it will be `&` now.  So that is a user-visible change; conceivably someone is parsing logs and they will need to react.

No, it's not something I am concerned about.  Just a fact, one that would go in the CHANGELOG.

annanay25:
Thanks for the explanation, I'll update the `CHANGELOG` per your comments.

PR #2537, opened 2020-04-28
handling of empty directories in filesystem object client

sandeepsukhani:
**What this PR does**:
When using`filesystem` as ObjectClient we will not remove the directory as well when it is the last object that is being deleted.
This PR also excludes empty directories from the listing with `List` calls.
Both these changes make `filesystem` as ObjectClient consistent with hosted object stores like S3, GCS and Azure.

Changes inspired from filesystem as object store in Thanos.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/local/fs_object_client.go
+++ b/pkg/chunk/local/fs_object_client.go
@@ -40,6 +40,10 @@ type FSObjectClient struct {
 
 // NewFSObjectClient makes a chunk.Client which stores chunks as files in the local filesystem.
 func NewFSObjectClient(cfg FSConfig) (*FSObjectClient, error) {
+	// filepath.Clean cleans up the path by removing unwanted duplicate slashes, dots etc.
+	// This is needed because DeleteObject works on paths which are already cleaned up and it
+	// checks whether it is about to delete the configured directory when it becomes empty
+	cfg.Directory = filepath.Clean(cfg.Directory)
 	if err := util.EnsureDirectory(cfg.Directory); err != nil {
 		return nil, err
 	}
@@ -135,30 +139,23 @@ func (f *FSObjectClient) List(ctx context.Context, prefix string) ([]chunk.Stora
 }
 
 func (f *FSObjectClient) DeleteObject(ctx context.Context, objectKey string) error {
-	filePath := filepath.Join(f.cfg.Directory, objectKey)
+	// inspired from copied from https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/objstore/filesystem/filesystem.go#L195
+	file := filepath.Join(f.cfg.Directory, objectKey)
 
-	err := os.Remove(filePath)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return chunk.ErrStorageObjectNotFound
+	for file != f.cfg.Directory {
+		if err := os.RemoveAll(file); err != nil {
+			return err
 		}
-		return err
-	}
 
-	// remove the parent directory when it is empty
-	parentDir := filepath.Dir(filePath)
-	ok, err := isDirEmpty(parentDir)
-	if err != nil {
-		return err
-	}
+		file = filepath.Dir(file)
+		empty, err := isDirEmpty(file)
+		if err != nil {
+			return err
+		}
 
-	if !ok {
-		return nil
-	}
-
-	err = os.Remove(parentDir)
-	if err != nil && !isNotEmptyErr(err) {
-		return err
+		if !empty {
+			break
+		}
 	}
 
 	return nil
--- a/pkg/chunk/local/fs_object_client.go
+++ b/pkg/chunk/local/fs_object_client.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"syscall"
 	"time"
 
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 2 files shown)


pkg/chunk/local/fs_object_client.go
```diff
@@ -127,12 +134,28 @@ func (f *FSObjectClient) List(ctx context.Context, prefix string) ([]chunk.Stora
...
+	if err != nil {
+		return err
+	}
+
+	if !ok {
+		return nil
 	}
 
-	return err
+	return os.RemoveAll(parentDir)
```
pracucci, line 158: Using `RemoveAll()` may suffer race conditions. If another process has created an object in the meanwhile (after your call to `isDirEmpty()`, this `RemoveAll()` call will erroneously delete it too. You should use `Remove()` here and check if the returned error is "directory not empty" and, if so, ignore it (because it's OK).

pkg/chunk/local/fs_object_client.go
```diff
@@ -127,12 +135,33 @@ func (f *FSObjectClient) List(ctx context.Context, prefix string) ([]chunk.Stora
...
+	ok, err := isDirEmpty(parentDir)
+	if err != nil {
+		return err
+	}
+
+	if !ok {
+		return nil
+	}
+
+	err = os.Remove(parentDir)
```
pstibrany, line 159: If parentDir was the only dir in its parent directory, we should recurse, until we reach the root of filesystem object store.

pkg/chunk/local/fs_object_client.go
```diff
@@ -64,8 +64,8 @@ func (f *FSObjectClient) GetObject(ctx context.Context, objectKey string) (io.Re
 
 // PutObject into the store
 func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object io.ReadSeeker) error {
-	fullPath := path.Join(f.cfg.Directory, objectKey)
-	err := util.EnsureDirectory(path.Dir(fullPath))
+	fullPath := filepath.Join(f.cfg.Directory, objectKey)
+	err := util.EnsureDirectory(filepath.Dir(fullPath))
```
pstibrany, line 68: Logic in `EnsureDirectory` seems to copy logic in `os.MkdirAll`, I think it can be simplified to 

```
// EnsureDirectory makes sure directory is there, if not creates it if not
func EnsureDirectory(dir string) error {
	return os.MkdirAll(dir, 0777)
}
```

at which point, I am not sure it's really needed.

pkg/chunk/local/fs_object_client.go
```diff
@@ -127,12 +135,33 @@ func (f *FSObjectClient) List(ctx context.Context, prefix string) ([]chunk.Stora
...
+	// remove the parent directory when it is empty
+	parentDir := filepath.Dir(filePath)
+	ok, err := isDirEmpty(parentDir)
+	if err != nil {
+		return err
+	}
+
+	if !ok {
+		return nil
+	}
```
pstibrany, line 157: This is not necessary. `os.Remove` removes `parentDir` only if it's empty.

Also, there should be a check to avoid removing `f.cfg.Directory`  -- while not wrong, it would seem weird.

pkg/chunk/local/fs_object_client.go
sandeepsukhani: That code was already there added by Tom in the initial commit to add boltdb support. I had just moved that code to util so not sure why it was added in the first place. See https://github.com/cortexproject/cortex/commit/78e83530e03c0fcc729a3e7c640b39e36ce787c7#diff-f0b676dad4198715e462b2faa06cfaa8R220
It is being used at a couple of places so I think it needs a separate PR to address that.

sandeepsukhani:
@pstibrany It seems you were asking for all the things that filesystem object client in Thanos was doing i.e recursively removing empty parent directories and not removing the configured directory so I have actually copied the code from Thanos in my last commit and did some changes relevant to our code.
Thanks for the review!

pkg/chunk/local/fs_object_client.go
```diff
@@ -147,3 +173,26 @@ func (f *FSObjectClient) DeleteChunksBefore(ctx context.Context, ts time.Time) e
...
+}
+
+func isNotEmptyErr(err error) bool {
+	pathErr, ok := err.(*os.PathError)
+	if ok && pathErr.Err == syscall.ENOTEMPTY {
+		return true
+	}
+
+	return false
+}
```
pstibrany, line 198: Only used in tests now.

pkg/chunk/local/fs_object_client.go
```diff
@@ -127,12 +139,26 @@ func (f *FSObjectClient) List(ctx context.Context, prefix string) ([]chunk.Stora
...
 
 func (f *FSObjectClient) DeleteObject(ctx context.Context, objectKey string) error {
-	err := os.Remove(path.Join(f.cfg.Directory, objectKey))
-	if err != nil && os.IsNotExist(err) {
-		return chunk.ErrStorageObjectNotFound
+	// inspired from https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/objstore/filesystem/filesystem.go#L195
+	file := filepath.Join(f.cfg.Directory, objectKey)
+
+	for file != f.cfg.Directory {
+		if err := os.RemoveAll(file); err != nil {
```
pstibrany, line 146: We want `os.Remove` only, not `All`, as that will clean non-empty directories too. It will also remove entire sub-tree, if passed objectKey points to a directory!

Issue #2291, opened 2020-03-18
Generalize Module Service


jtlisi:
Cortex initializes using a series of modules which are an abstraction on top of the util.services package. A log of this log could be abstracted into a dedicated package for modules.

This would also allow any projects to reuse the modules logic if they want to. For instance, Loki copied over most of the module code.

PR #2542, opened 2020-04-29
Design doc - Generalize Module Service

annanay25:

**What this PR does**:
This PR adds an initial draft for a design document proposing generalizing the module service. 

Part of #2291 

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/proposals/generalize-modules.md
+++ b/docs/proposals/generalize-modules.md
@@ -6,15 +6,15 @@ slug: generalize-modules
 ---
 
 - Author: @annanay25
-- Reviewers: 
+- Reviewers: @jtlisi, @pstibrany
 - Date: April 2020
-- Status: Draft
+- Status: Review
 
 ## Overview
 
-Cortex uses modules to start and operate services with dependencies. Inter-service dependencies are specified in a map and passed to a module manager which ensures that they are started in the right order of dependencies. While this works really well, the implementation is tied in specifically to the Cortex struct and is not flexible for use with other projects like Loki, which also require similar forms of dependency management. 
- 
-We would like to extend modules in cortex to a generic dependency management framework, that can be used by any project with no ties to cortex. 
+Cortex uses modules to start and operate services with dependencies. Inter-service dependencies are specified in a map and passed to a module manager which ensures that they are started in the right order of dependencies. While this works really well, the implementation is tied in specifically to the Cortex struct and is not flexible for use with other projects like Loki, which also require similar forms of dependency management.
+
+We would like to extend modules in cortex to a generic dependency management framework, that can be used by any project with no ties to cortex.
 
 ## Specific goals
 
@@ -25,39 +25,33 @@ We would like to extend modules in cortex to a generic dependency management fra
 - Framework should allow for building an application from scratch using the `modules` package, with no dependencies on Cortex. For ex: Remove code from Loki that was copied from `pkg/cortex/cortex.go`.
 
 
-
 ## Proposed Design
+
+### Modules package
 
 To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to -
 
-- Make a new component `moduleManager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. The following is the interface for interacting with the `moduleManager`:
+- Make a new component `Manager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. `Manager` has the following methods for interaction:
 ```
-// Manager defines the interface for interaction with moduleManager.
-type Manager interface {
-   RegisterModule(name string, deps []string, svc service, options ...Option)
-   AddDependency(fromModule string, toModule string) error
-   ModuleServiceWrapper(name string, modServ services.Service)
-   InitModuleServices(target string) (map[string]services.Service, error)
-}
+   func (m *Manager) RegisterModule(name string, deps []string, initFn func() (Service, error))
+   func (m *Manager) AddDependency(fromModule string, toModule string) error
+   func (m *Manager) InitModuleServices(target string) (map[string]services.Service, error)
 ```
 
-- Modules can be created by the application and registered with the `ModuleManager` using `RegisterModule`. The parameters are:
+- Modules can be created by the application and registered with `modules.Manager` using `RegisterModule`. The parameters are:
   - `name`: Name of the module
   - `deps`: A list of modules that this module depends on to run.
-  - `svc`: A function that will be used to start the module.
-  - `options`: Options are a variadic function parameter that can be used by the `moduleManager` to wrap the service function. In Cortex, we pass `Manager.ModuleServiceWrapper`, which wraps service to work with dependencies. 
+  - `initFn`: A function that will be used to start the module. If it returns nil, and other modules depend on it `InitModuleServices` will return an error.
 
-- Dynamic dependencies between modules can be added using `AddDependency`. However, these dependencies need to be added before the call to `InitModuleServices`.
+- Dynamic dependencies between modules can be added using `AddDependency`. These need to be added before the call to `InitModuleServices`.
 
 - The application can be initialized by running all the modules in the right order of dependencies by invoking `InitModuleServices`.
 
-- `WrappedService` present in the current `module` design has been deprecated. To distinguish between `Service` and `WrappedService` in `RegisterModule`, the following options are available:
-  - The simplest option is to pass a `boolean` parameter in that can default to `true` for a `WrappedService` since most services in Cortex are `WrappedServices`.
-  - The second option is to wrap every `Service` into `WrappedService` if `options == nil` (default case), and for the special case where a service need not be wrapped, a dummy function can be passed.
-  - A special function can be passed to the `ModuleManager` to wrap a `Service` into a `WrappedService` which waits on dependencies to start before it and stop after it. This parameter is `ModuleManager.ModuleServiceWrapper` referenced in the `Manager` interface.
 
-- While the process of loading modules into `modules.Manager` should be remain as part of the `Cortex.New()` function, `InitModuleServices` should be part of `Cortex.Run()` and to enable this, `modules.Manager` would be made a member of the `Cortex` struct. 
+### Changes to `pkg/cortex`:
+- `WrappedService` present in the current `module` design will be deprecated. All `initFn`'s will be wrapped into `WrappedService` by default.
 
+- While the process of loading modules into `modules.Manager` should be remain as part of the `Cortex.New()` function, `InitModuleServices` should be part of `Cortex.Run()` and to enable this, `modules.Manager` would be made a member of the `Cortex` struct.
 
 
 ## Usage
@@ -66,7 +60,7 @@ Following these changes, the Modules package will be a generic dependency manage
 
 #### To use the modules framework:
 - Import the `pkg/util/modules` package, and initialize a new instance of the `Manager` using `modules.NewManager()`
-- Create components in the system that implement the services interface (present in `pkg/util/services`).  
-- Register each of these components as a module using `Manager.RegisterModule()` by passing name of the module, dependencies, initFn and any extra options that can be used to wrap the initFn.
+- Create components in the system that implement the services interface (present in `pkg/util/services`).
+- Register each of these components as a module using `Manager.RegisterModule()` by passing name of the module, dependencies, and `initFn`.
 - To add dynamic dependencies between modules, use `Manager.AddDependency()`
 - Once all modules are added into `modules.Manager`, initialize the application by calling `Manager.InitModuleServices()` which initializes modules in the right order of dependencies.
--- a/docs/proposals/generalize-modules.md
+++ b/docs/proposals/generalize-modules.md
@@ -6,13 +6,13 @@ slug: generalize-modules
 ---
 
 - Author: @annanay25
-- Reviewers: @jtlisi, @pstibrany
+- Reviewers: @jtlisi, @pstibrany, @cyriltovena, @pracucci
 - Date: April 2020
-- Status: Review
+- Status: Accepted
 
 ## Overview
 
-Cortex uses modules to start and operate services with dependencies. Inter-service dependencies are specified in a map and passed to a module manager which ensures that they are started in the right order of dependencies. While this works really well, the implementation is tied in specifically to the Cortex struct and is not flexible for use with other projects like Loki, which also require similar forms of dependency management.
+Cortex uses modules to start and operate services with dependencies. Inter-service dependencies are specified in a map and passed to a module manager which ensures that they are initialised in the right order of dependencies. While this works really well, the implementation is tied in specifically to the Cortex struct and is not flexible for use with other projects like Loki, which also require similar forms of dependency management.
 
 We would like to extend modules in cortex to a generic dependency management framework, that can be used by any project with no ties to cortex.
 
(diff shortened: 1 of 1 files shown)


docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,72 @@
...
+
+To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to -
+
+- Make a new component `moduleManager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. The following is the interface for interacting with the `moduleManager`:
+```
+// Manager defines the interface for interaction with moduleManager.
+type Manager interface {
+   RegisterModule(name string, deps []string, svc service, options ...Option)
+   AddDependency(fromModule string, toModule string) error
+   ModuleServiceWrapper(name string, modServ services.Service)
```
jtlisi, line 39: I don't think this function should be part of the interface.

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,72 @@
...
+  - `svc`: A function that will be used to start the module.
+  - `options`: Options are a variadic function parameter that can be used by the `moduleManager` to wrap the service function. In Cortex, we pass `Manager.ModuleServiceWrapper`, which wraps service to work with dependencies. 
+
+- Dynamic dependencies between modules can be added using `AddDependency`. However, these dependencies need to be added before the call to `InitModuleServices`.
+
+- The application can be initialized by running all the modules in the right order of dependencies by invoking `InitModuleServices`.
+
+- `WrappedService` present in the current `module` design has been deprecated. To distinguish between `Service` and `WrappedService` in `RegisterModule`, the following options are available:
+  - The simplest option is to pass a `boolean` parameter in that can default to `true` for a `WrappedService` since most services in Cortex are `WrappedServices`.
+  - The second option is to wrap every `Service` into `WrappedService` if `options == nil` (default case), and for the special case where a service need not be wrapped, a dummy function can be passed.
```
jtlisi, line 56: I strongly prefer the second option here.

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,72 @@
...
+To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to -
+
+- Make a new component `moduleManager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. The following is the interface for interacting with the `moduleManager`:
+```
+// Manager defines the interface for interaction with moduleManager.
+type Manager interface {
+   RegisterModule(name string, deps []string, svc service, options ...Option)
+   AddDependency(fromModule string, toModule string) error
+   ModuleServiceWrapper(name string, modServ services.Service)
+   InitModuleServices(target string) (map[string]services.Service, error)
```
jtlisi, line 40: It would be nice if this returned a services.Manager since we immediately convert it into one in `Cortex.Run`. As far as I'm aware the only other use for this map is to support the `/services` endpoint. Could the services.Manager be refactored to act as a handler for this endpoint?

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,72 @@
...
+## Proposed Design
+
+To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to -
+
+- Make a new component `moduleManager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. The following is the interface for interacting with the `moduleManager`:
+```
+// Manager defines the interface for interaction with moduleManager.
+type Manager interface {
+   RegisterModule(name string, deps []string, svc service, options ...Option)
+   AddDependency(fromModule string, toModule string) error
```
jtlisi, line 38: I think we could remove this from the interface as well by adding a `SetDependant` option. When provided it can update the target module with the one currently being registered as a dependency.

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,72 @@
...
+
+
+## Proposed Design
+
+To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to -
+
+- Make a new component `moduleManager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. The following is the interface for interacting with the `moduleManager`:
+```
+// Manager defines the interface for interaction with moduleManager.
+type Manager interface {
```
pstibrany, line 36: No need to have this as interface, let's just expose `ModuleManager`. (Interfaces are best defined at place where they are consumed)

Let's just say "ModuleManager" has following methods.

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,72 @@
...
+   RegisterModule(name string, deps []string, svc service, options ...Option)
```
pstibrany, line 37: What is `service` here? Module needs to pass a function that creates a service (like various `init` functions in Cortex).

docs/proposals/generalize-modules.md
pstibrany: It could be useful to define dependencies on modules that are not yet registered. Possible errors should be detected later.

docs/proposals/generalize-modules.md
pstibrany: `InitModuleServices` contract is not clear. What state are returned services in?

> Could the services.Manager be refactored to act as a handler for this endpoint?

`services.Manager` is not a place to put HTTP handlers into. One has many service managers in a running program. Service manager doesn't use a service names, and only needs list of services, not their names.

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,72 @@
...
+   ModuleServiceWrapper(name string, modServ services.Service)
+   InitModuleServices(target string) (map[string]services.Service, error)
+}
+```
+
+- Modules can be created by the application and registered with the `ModuleManager` using `RegisterModule`. The parameters are:
+  - `name`: Name of the module
+  - `deps`: A list of modules that this module depends on to run.
+  - `svc`: A function that will be used to start the module.
+  - `options`: Options are a variadic function parameter that can be used by the `moduleManager` to wrap the service function. In Cortex, we pass `Manager.ModuleServiceWrapper`, which wraps service to work with dependencies.
```
pstibrany, line 48: The reason why `Server` service is special in Cortex is because underlying "Server" from weaveworks package handles SIGINT/SIGTERM, and reacts on it by exiting from its `Run` method, while keeping HTTP server running. This makes `Server` service transition to Stopping state, which is a signal for Cortex to begin shutdown procedure.

Notice that Server service still waits for other services to stop first (!) before stopping HTTP server.

If "Server" service was wrapped in "module wrapper", Cortex would only see outer (wrapping) service -- which waits until underlying service is fully stopped before moving to Stopping state.

If we 1) disable weaveworks' Server handling of SIGTERM/SIGHUP (not currently possible), and 2) make Cortex wait for SIGHUP/SIGTERM on it's own, then we can also use module wrapper for Server.

docs/proposals/generalize-modules.md
pstibrany: I would prefer to make all services returned by module manager to be use "wrapping" with module dependencies, and fix the root cause why we're doing this in the first place, as explained above.

docs/proposals/generalize-modules.md
annanay25: > Interfaces are best defined at place where they are consumed

I agree. Thanks!

> Let's just say "ModuleManager" has following methods.

Just a nit, would `modules.Manager` be cleaner?

docs/proposals/generalize-modules.md
annanay25: `service` is a `func()` type, would it be better renamed as `initFn`?

docs/proposals/generalize-modules.md
annanay25: Since @pstibrany suggested we do away with this interface, it will now be a function of `modules.Manager` struct.

docs/proposals/generalize-modules.md
annanay25: Thanks for the detailed explanation @pstibrany! This design doc addresses this problem by not wrapping the `Server` service in module wrapper (by passing `nil` options) like so - 
```
mm := modules.NewManager()
.
.
mm.RegisterModule(Server, nil, t.initServer)
```

.. and by wrapping other services in module wrapper like so -

```
mm.RegisterModule(Ring, []string{API, RuntimeConfig, MemberlistKV}, t.initRing, mm.ModuleServiceWrapper)
```

docs/proposals/generalize-modules.md
pstibrany: `service` type was unclear to me. Yes, I think `initFn func() (Service, error)` would be more descriptive. It should be documented that it can return nil (right? that's what current init functions do), in which case no service is returned for this module ... but if any other service depends on it, how should that behave?

docs/proposals/generalize-modules.md
pstibrany: `modules.Manager` sgtm.

docs/proposals/generalize-modules.md
pstibrany: > It would be nice if this returned a services.Manager since we immediately convert it into one in Cortex.Run

I find current proposal (returning map) better – more flexible. Client may be interested in modifying (eg. extra wrapping) some specific services before using them.

docs/proposals/generalize-modules.md
pstibrany: I'm not sure we even need `Options`. What other options do you expect to have? We can always introduce variadic options in the future, when needed.

annanay25: Hrm, I'm not sure but this could enable extra-wrapping of services, as you mentioned in the other comment https://github.com/cortexproject/cortex/pull/2542/files#r417929921

pstibrany: I think stuff that can be done by caller should be done by caller. When we see what other options are interesting, we can always introduce them. But I would start with simpler design (unless we already have some options now)

docs/proposals/generalize-modules.md
annanay25: > It should be documented that it can return nil

Yes. I'll make sure to add that.

> how should that behave?

In `InitModuleServices` we can error out if `initFn()` returns `nil` and the module has "inverse dependencies".

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+  - Overwrite the implementation of a current module
+  - Manage dependencies
+- Framework should allow for building an application from scratch using the `modules` package, with no dependencies on Cortex. For ex: Remove code from Loki that was copied from `pkg/cortex/cortex.go`.
+
+
+## Proposed Design
+
+### Modules package
+
+To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to -
```
pracucci, line 32:
```suggestion
To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to:
```

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Make a new component `Manager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. `Manager` has the following methods for interaction:
+```
+   func (m *Manager) RegisterModule(name string, deps []string, initFn func() (Service, error))
+   func (m *Manager) AddDependency(fromModule string, toModule string) error
+   func (m *Manager) InitModuleServices(target string) (map[string]services.Service, error)
```
pracucci, line 38: Have you considered calling it `Start()` given I assume it will start module services?

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+## Proposed Design
+
+### Modules package
+
+To make the modules package extensible, we need to abstract away any Cortex specific details from the module manager. The proposed design is to -
+
+- Make a new component `Manager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. `Manager` has the following methods for interaction:
+```
+   func (m *Manager) RegisterModule(name string, deps []string, initFn func() (Service, error))
+   func (m *Manager) AddDependency(fromModule string, toModule string) error
```
pracucci, line 37: Just looking at the function signature it's not clear to me if `fromModule` depends on `toModule` or viceversa. Variable naming wise, you may consider:

```
func (m *Manager) AddDependency(name string, dependsOn string) error
```

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+  - `name`: Name of the module
+  - `deps`: A list of modules that this module depends on to run.
+  - `initFn`: A function that will be used to start the module. If it returns nil, and other modules depend on it `InitModuleServices` will return an error.
+
+- Dynamic dependencies between modules can be added using `AddDependency`. These need to be added before the call to `InitModuleServices`.
+
+- The application can be initialized by running all the modules in the right order of dependencies by invoking `InitModuleServices`.
+
+
+### Changes to `pkg/cortex`:
```
pracucci, line 51: (just to fix markdown compilation when building the website)

```suggestion
### Changes to `pkg/cortex`:

```

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+- `WrappedService` present in the current `module` design will be deprecated. All `initFn`'s will be wrapped into `WrappedService` by default.
+
+- While the process of loading modules into `modules.Manager` should be remain as part of the `Cortex.New()` function, `InitModuleServices` should be part of `Cortex.Run()` and to enable this, `modules.Manager` would be made a member of the `Cortex` struct.
+
+
+## Usage
+
+Following these changes, the Modules package will be a generic dependency management framework that can be used by any project.
+
+#### To use the modules framework:
```
pracucci, line 61: (just to fix markdown compilation when building the website)
```suggestion
#### To use the modules framework:

```

docs/proposals/generalize-modules.md
pstibrany: We should mention ability to pass multiple targets as a future extension (unless you want to put it into first implementation, but that's not necessary).

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+- Make a new component `Manager`, which is envisioned to be a central manager for all modules of the application. It stores modules & dependencies, and will be housed under a new package `pkg/util/modules`. `Manager` has the following methods for interaction:
+```
+   func (m *Manager) RegisterModule(name string, deps []string, initFn func() (Service, error))
+   func (m *Manager) AddDependency(fromModule string, toModule string) error
+   func (m *Manager) InitModuleServices(target string) (map[string]services.Service, error)
+```
+
+- Modules can be created by the application and registered with `modules.Manager` using `RegisterModule`. The parameters are:
+  - `name`: Name of the module
+  - `deps`: A list of modules that this module depends on to run.
```
pstibrany, line 43: There are two ways of defining dependencies now... via `deps` in `RegisterModule`, and via `AddDependency`. I would prefer to have single way of doing it. Personally I think configuring dependencies via `AddDependency` with variadic `dependsOn` parameter would be enough.

docs/proposals/generalize-modules.md
pstibrany: Extending Marco's suggestion:

```
func (m *Manager) AddDependency(name string, dependsOn... string) error
```

(making `dependsOn` variadic)

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+  - `initFn`: A function that will be used to start the module. If it returns nil, and other modules depend on it `InitModuleServices` will return an error.
+
+- Dynamic dependencies between modules can be added using `AddDependency`. These need to be added before the call to `InitModuleServices`.
+
+- The application can be initialized by running all the modules in the right order of dependencies by invoking `InitModuleServices`.
```
pstibrany, line 48: Nit: "running all the modules" -- it's running "initialization functions" of those modules.

docs/proposals/generalize-modules.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+### Changes to `pkg/cortex`:
+- `WrappedService` present in the current `module` design will be deprecated. All `initFn`'s will be wrapped into `WrappedService` by default.
+
+- While the process of loading modules into `modules.Manager` should be remain as part of the `Cortex.New()` function, `InitModuleServices` should be part of `Cortex.Run()` and to enable this, `modules.Manager` would be made a member of the `Cortex` struct.
```
pstibrany, line 54: > `modules.Manager` would be made a member of the `Cortex` struct.

Just curious, is this required? It feels like there should be no need to put Manager into Cortex structure, as we don't really need it beyond the initialization of the Cortex.

docs/proposals/generalize-modules.md
annanay25: It initialises all the module services, but does not start them. I would prefer calling it `InitModuleServices`.

> We should mention ability to pass multiple targets as a future extension

Definitely. I will implement it as a follow up after the first implementation is merged.

pstibrany: Agree with Annanay, this naming is clearer. Method documentation should be clear about returning services in New state. Also requirement of init function is that returned service (if not nil) is in New state -- module.Manager can test that and return error if that's not true.

docs/proposals/generalize-modules.md
annanay25: I agree, this is cleaner. Having a single way of adding dependencies as per your other comment also LGTM. :)

docs/proposals/generalize-modules.md
pracucci: Agree too. I initially misunderstood it. I thought it was also starting them, not just initialising. All good.

PR #2546, opened 2020-04-30
provisioning of table for delete requests using table manager

sandeepsukhani:
**What this PR does**:
Adds support for provisioning tables for delete store using table manager.
Throughput settings for tables created for delete requests store are part of the store configuration itself.
I have refactored the code used for provisioning throughput for index and chunks tables in table manager to be able to reuse most of the code.
This also adds support for `map` type in doc generator.

**NOTE**: Since the refactoring of provisioning code in table manager changed the order of elements in configuration struct the order of the entries in configuration doc as well changed. There are no breaking changes in configuration.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/purger/table_provisioning.go
+++ b/pkg/chunk/purger/table_provisioning.go
@@ -1,0 +1,30 @@
+package purger
+
+import (
+	"flag"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+)
+
+// TableProvisioningConfig holds config for table throuput and autoscaling. Currently only used by DynamoDB.
+type TableProvisioningConfig struct {
+	chunk.ActiveTableProvisionConfig `yaml:",inline"`
+	TableTags                        chunk.Tags `yaml:"tags"`
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet.
+// Adding a separate RegisterFlags here instead of using it from embedded chunk.ActiveTableProvisionConfig to be able to manage defaults separately.
+// Defaults for WriteScale and ReadScale are shared for now to avoid adding further complexity since autoscaling is disabled anyways by default.
+func (cfg *TableProvisioningConfig) RegisterFlags(argPrefix string, f *flag.FlagSet) {
+	f.Int64Var(&cfg.ProvisionedWriteThroughput, argPrefix+".write-throughput", 1, "Table default write throughput. Supported by DynamoDB")
+	f.Int64Var(&cfg.ProvisionedReadThroughput, argPrefix+".read-throughput", 300, "Table default read throughput. Supported by DynamoDB")
+	f.BoolVar(&cfg.ProvisionedThroughputOnDemandMode, argPrefix+".enable-ondemand-throughput-mode", false, "Enables on demand throughput provisioning for the storage provider (if supported). Applies only to tables which are not autoscaled. Supported by DynamoDB")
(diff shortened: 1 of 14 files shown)


pkg/chunk/table_provisioning.go
```diff
@@ -0,0 +1,105 @@
+package chunk
+
+import "flag"
+
+// ProvisionConfig holds config for provisioning capacity for index and chunk tables (on DynamoDB for now)
+type ProvisionConfig struct {
+	InactiveWriteScaleLastN int64 `yaml:"inactive_write_scale_lastn"`
```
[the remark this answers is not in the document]
sandeepsukhani: I wanted to keep `InactiveTableProvisionConfig` reusable but I think we do not have a use-case for it now so I can move them to `InactiveTableProvisionConfig`. Will do the changes.

pkg/chunk/purger/table_provisioning.go
```diff
@@ -0,0 +1,30 @@
...
+// Adding a separate RegisterFlags here instead of using it from embedded chunk.ActiveTableProvisionConfig to be able to manage defaults separately.
+// Defaults for WriteScale and ReadScale are shared for now to avoid adding further complexity since autoscaling is disabled anyways by default.
+func (cfg *TableProvisioningConfig) RegisterFlags(argPrefix string, f *flag.FlagSet) {
+	f.Int64Var(&cfg.ProvisionedWriteThroughput, argPrefix+".write-throughput", 1, "Table default write throughput. Supported by DynamoDB")
+	f.Int64Var(&cfg.ProvisionedReadThroughput, argPrefix+".read-throughput", 300, "Table default read throughput. Supported by DynamoDB")
+	f.BoolVar(&cfg.ProvisionedThroughputOnDemandMode, argPrefix+".enable-ondemand-throughput-mode", false, "Enables on demand throughput provisioning for the storage provider (if supported). Applies only to tables which are not autoscaled. Supported by DynamoDB")
+	f.Var(&cfg.TableTags, argPrefix+".tags", "Tag (of the form key=value) to be added to the tables. Supported by DynamoDB")
+
+	cfg.WriteScale.RegisterFlags(argPrefix+".write-throughput.scale", f)
+	cfg.ReadScale.RegisterFlags(argPrefix+".read-throughput.scale", f)
```
pstibrany, line 25:
```suggestion
	cfg.ActiveTableProvisionConfig.RegisterFlags(argPrefix, f)
	f.Var(&cfg.TableTags, argPrefix+".tags", "Tag (of the form key=value) to be added to the tables. Supported by DynamoDB")
```

sandeepsukhani: @pstibrany thanks for the review. The defaults in `ActiveTableProvisionConfig` are way too high considering the amount of traffic expected for tables storing delete request, which is why I am doing it separately here with lower defaults. Does it make sense?

pstibrany: In that case, I would suggest that `cfg.ActiveTableProvisionConfig.RegisterFlags` takes defaults from `cfg.ActiveTableProvisionConfig`, and we set correct defaults in this method, and other place where `RegisterFlags` is called. I'd prefer if we can avoid duplicating all config option names and help strings. WDYT?

pstibrany:
Apologies for letting you wait for so long! :-(

pstibrany:
There are conflicts in this PR that need resolving before merge.

sandeepsukhani:
@pstibrany all the conflicts are resolved. This is good to go.

PR #2549, opened 2020-04-30
Implement /api/v1/metadata

gotjosh:
**What this PR does**:

Properly Implements the `/api/v1/metadata` endpoint. Up until now, we were using the default behaviour introduced by the Prometheus API in:

https://github.com/cortexproject/cortex/blob/d94cdce0aa46b69fff7cb40d1ad57e77b11b4ab0/pkg/api/api.go#L253-L268

To fetch this information the dummy struct `dummy. DummyTargetRetriever` was used which meant no records where ever returned. 

The reason for this implementation is two-fold:

- Given an implemented `dummy. DummyTargetRetriever`  does not support `context` passing, we're not able to extract the tenant ID. This was fixed in upstream Prometheus by https://github.com/prometheus/prometheus/pull/7125
- We vendor Prometheus through Thanos. By the looks of it is going to take a while before we upgrade based on https://github.com/thanos-io/thanos/pull/2535

Once Prometheus is upgraded, this can be removed in favour of the implementation of `dummy. DummyTargetRetriever`.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [] Documentation added (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * [FEATURE] Ruler: The `-ruler.evaluation-delay` flag was added to allow users to configure a default evaluation delay for all rules in cortex. The default value is 0 which is the current behavior. #2423
 * [FEATURE] Experimental: Added a new object storage client for OpenStack Swift. #2440
 * [FEATURE] Update in dependency `weaveworks/common`. TLS config options added to the Server. #2535
+* [FEATURE] Experimental: Added support for `/api/v1/metadata` Prometheus-based endpoint. #XXXX
 * [ENHANCEMENT] Experimental TSDB: sample ingestion errors are now reported via existing `cortex_discarded_samples_total` metric. #2370
 * [ENHANCEMENT] Failures on samples at distributors and ingesters return the first validation error as opposed to the last. #2383
 * [ENHANCEMENT] Experimental TSDB: Added `cortex_querier_blocks_meta_synced`, which reflects current state of synced blocks over all tenants. #2392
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 * [CHANGE] Experimental WAL: WAL and checkpoint compression is now disabled. #2436
 * [FEATURE] Ruler: The `-ruler.evaluation-delay` flag was added to allow users to configure a default evaluation delay for all rules in cortex. The default value is 0 which is the current behavior. #2423
 * [FEATURE] Experimental: Added a new object storage client for OpenStack Swift. #2440
+* [FEATURE] Experimental: Added support for `/api/v1/metadata` Prometheus-based endpoint. #XXXX
```
gotjosh, line 26:
```suggestion
* [FEATURE] Experimental: Added support for `/api/v1/metadata` Prometheus-based endpoint. #2549
```

gotjosh:
I can't assign reviewers just I'll just give you a gentle nudge @pracucci @pstibrany

pkg/api/api.go
```diff
@@ -294,7 +294,7 @@ func (a *API) RegisterQuerier(queryable storage.Queryable, engine *promql.Engine
 	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/labels", promHandler, true, "GET", "POST")
 	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/label/{name}/values", promHandler, true, "GET")
 	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/series", promHandler, true, "GET", "POST", "DELETE")
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/metadata", promHandler, true, "GET")
+	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/metadata", querier.MetadataHandler(distributor), true, "GET")
```
gouthamve, line 297: Can we add a TODO here explaining that we have this special handler because of the issues you mentioned in the PR description?

gotjosh: Will do.

PR #2559, opened 2020-05-05
Generalise Modules Service to make it extensible

### LINKED ISSUE #2291: Generalize Module Service

annanay25:

**What this PR does**:
Implementation of the [generalise module service](https://github.com/annanay25/cortex/blob/7cce7ba50d3ea1dfa42d42bd01226a23870e8aab/docs/proposals/generalize-modules.md) design doc.

Fixes #2291 

**Checklist**
- [x] Tests updated
- NA Documentation added


pkg/cortex/cortex.go
```diff
@@ -185,38 +185,39 @@ func (c *Config) Validate(log log.Logger) error {
...
+	Ingester         *ingester.Ingester
+	Flusher          *flusher.Flusher
+	Store            chunk.Store
+	DeletesStore     *purger.DeleteStore
+	Frontend         *frontend.Frontend
+	TableManager     *chunk.TableManager
+	Cache            cache.Cache
+	RuntimeConfig    *runtimeconfig.Manager
+	DataPurger       *purger.DataPurger
+	TombstonesLoader *purger.TombstonesLoader
```
pstibrany, line 208: Are renames important for this PR? Let's keep them as a separate PR to minimize amount of changes in this one.

pkg/cortex/cortex.go
```diff
@@ -229,18 +230,13 @@ func New(cfg Config) (*Cortex, error) {
...
 
-	serviceMap, err := cortex.initModuleServices(&cfg, cfg.Target)
-	if err != nil {
-		return nil, err
-	}
-
-	cortex.serviceMap = serviceMap
-	cortex.api.RegisterServiceMapHandler(http.HandlerFunc(cortex.servicesHandler))
+	mm := cortex.createModuleManager()
+	cortex.ModuleManager = mm
```
pstibrany, line 239: It's very strange to call `createModuleManager` on cortex instance, only to put the result back into this instance.

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,157 @@
+package modules
+
+import (
+	"fmt"
+
+	"github.com/cortexproject/cortex/pkg/util/services"
+	"github.com/pkg/errors"
+)
+
+type service func() (services.Service, error)
```
pstibrany, line 10: `service` is not a good name. I'm not sure extra type is necessary, it's used at exactly two places.

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,157 @@
...
+// AddDependency adds a dependency from name(source) to dependsOn(targets)
+func (m *Manager) AddDependency(name string, dependsOn ...string) error {
+	if mod, ok := m.modules[name]; ok {
+		for _, dep := range dependsOn {
+			if _, ok := m.modules[dep]; ok {
+				mod.deps = append(mod.deps, dep)
+			} else {
+				return fmt.Errorf("no such module: %s", dep)
+			}
+		}
```
pstibrany, line 51: I expected that dependencies will be only checked during `InitModuleServices`, not when adding them. Otherwise modules need to be added in the correct order.

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,157 @@
...
+	} else {
+		return fmt.Errorf("no such module: %s", name)
+	}
+	return nil
+}
+
+// InitModuleServices starts the target module
```
pstibrany, line 58: This comment is underspecified. What does "starts" mean here? No details on what is being returned, what are the error conditions. This was part of the proposal.

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,157 @@
...
+		if mod.initFn != nil {
+			s, err := mod.initFn()
+			if err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("error initialising module: %s", n))
+			}
+
+			invDeps := m.findInverseDependencies(n, deps[ix+1:])
+			if s == nil {
+				if invDeps != nil {
+					return nil, fmt.Errorf("module %s returned nil service but has other modules dependent on it", n)
```
pstibrany, line 80: Previously it was easier to find "inverse dependencies" in the source code, but in this dynamic version, it may be tricky. I think we should include them in the error message.

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,157 @@
...
+
+// RegisterModule registers a new module with ModuleManager
+func (m *Manager) RegisterModule(name string, initFn service) {
+	m.modules[name] = module{
+		initFn: initFn,
+	}
+	return
+}
+
+// AddDependency adds a dependency from name(source) to dependsOn(targets)
```
pstibrany, line 42: Under which conditions can error be returned?

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// NewManager creates a new Manager
+func NewManager() *Manager {
+	return &Manager{
+		modules: make(map[string]module),
+	}
+}
+
+// RegisterModule registers a new module with ModuleManager
```
pstibrany, line 34:
```suggestion
// RegisterModule registers a new module with name and init function.
```

Can init function be nil? Must name be unique or not? Please specify.

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,157 @@
...
+func (m *Manager) RegisterModule(name string, initFn service) {
+	m.modules[name] = module{
+		initFn: initFn,
+	}
+	return
```
pstibrany, line 39:
```suggestion
```

pkg/cortex/modules.go
```diff
@@ -125,512 +97,447 @@ func (t *Cortex) initServer(cfg *Config) (services.Service, error) {
...
+		StoreQueryable: []string{Store},
+		QueryFrontend:  []string{API, Overrides, DeleteRequestsStore},
+		TableManager:   []string{API},
+		Ruler:          []string{Distributor, Store, StoreQueryable},
+		Configs:        []string{API},
+		AlertManager:   []string{API},
+		Compactor:      []string{API},
+		StoreGateway:   []string{API},
+		DataPurger:     []string{Store, DeleteRequestsStore, API},
+		All:            []string{QueryFrontend, Querier, Ingester, Distributor, TableManager, DataPurger, StoreGateway},
```
pstibrany, line 536: Please remove `[]string` from this code, it's not needed. 

```go
		API:  {Server},
		Ring: {API, RuntimeConfig, MemberlistKV},
...
```
is fine.

pkg/cortex/modules.go
```diff
@@ -125,512 +97,447 @@ func (t *Cortex) initServer(cfg *Config) (services.Service, error) {
...
+	}
+	for mod, targets := range deps {
+		mm.AddDependency(mod, targets...)
```
pstibrany, line 539: Returned error is ignored. I guess linter will point it out as well.

pkg/cortex/cortex.go
jtlisi: @annanay25 Can you open a separate PR that just handles renaming. Since there is no functional change it should be quick to review and lower the surface area of this PR.

pkg/util/modules/modules.go
jtlisi: I agree we shouldn't bother checking for this here.

pkg/util/modules/modules.go
annanay25: I'm not sure I fully understand what you mean here.

pstibrany: I mean to include names of "other modules"  in the error message. Basically.

```suggestion
					return nil, fmt.Errorf("module %s returned nil service but has other modules dependent on it: %v", n, invDeps)
```

Also it's better to check slice for length than `nil` (`if len(invDeps) > 0 ...`).

pkg/util/modules/modules.go
```diff
@@ -0,0 +1,153 @@
...
+
+		if mod.initFn != nil {
+			s, err := mod.initFn()
+			if err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("error initialising module: %s", n))
+			}
+
+			invDeps := m.findInverseDependencies(n, deps[ix+1:])
+			if s == nil {
+				if len(invDeps) > 0 {
```
jtlisi, line 75: If a service returns `nil` like this, but without an error that means we will probably need a stub service instead of throwing an error here. There are some modules that are really only meant for orchestration and don't need a running service.

pstibrany: I think such modules should not be used as a dependency for other modules. In that case, error is valid outcome here. WDYT?

jtlisi: In that case we would need to rearchitect other modules to return a service. For instance the API module only exists to register routes with the server. It doesn't need service to manage it since it only is required during initialization. Since every service that serves web traffic depends on it, it will need to return a service that does nothing.

pstibrany: OK, I see. Let's not treat nil service for module with other modules depending on it as an error then. (It's would be a change from what Cortex is currently doing anyway).

pstibrany:
> This looks close to ready, just needs some cleanup to make sure CICD is passing.

This PR also changes wrapping part (all modules are wrapped), so it cannot be merged until Server module can work with that.

annanay25:
@pstibrany - Looks like your changes to signal handling on `weaveworks/common` were merged. I will push the final changes here now.

pstibrany:
> This PR also changes wrapping part (all modules are wrapped), so it cannot be merged until Server module can work with that.

#2576 implements this.

pkg/util/modules/modules_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	invDeps := mm.findInverseDependencies("serviceA", []string{"serviceB", "serviceC"})
+	require.Len(t, invDeps, 1)
+	assert.Equal(t, invDeps[0], "serviceB")
+
+	svcs, err := mm.InitModuleServices("serviceC")
+	assert.NotNil(t, svcs)
+	assert.NoError(t, err)
+
+	svcs, err = mm.InitModuleServices("service_unknown")
```
pstibrany, line 46: This test is interesting because it calls `InitModuleServices` twice... it's not clear from documentation what should actually happen in that case!

Would you mind adding a test for missing dependency too?

annanay25:
Thanks! I'll send a follow up PR for the test.

PR #2543, opened 2020-04-29
Transition ruler to use Thanos bucket clients

jtlisi:
**What this PR does**:

As a next step for #2262 this PR transitions the ruler to using the Thanos bucket client instead of the chunk object client.

The swift integration tests were removed since the client library Thanos uses no longer supports the Auth protocol of the image we are using for integration tests:
[Thanos Upstream Client](https://github.com/gophercloud/gophercloud/blob/105a9694c07417728e1c8616497aeef7e192d2b0/openstack/client.go#L57-L112)
[Docker Image](https://github.com/swiftstack/docker-swift/issues/50)

**Which issue(s) this PR fixes**:
Related: #2262 

**Checklist**
- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2568, opened 2020-05-07
Do not prefix Thanos objstore client metrics and fix instrumentation in querier

pracucci:
**What this PR does**:
Currently Thanos objstore metrics are prefixed by `cortex_<service>_`. This leads to two issues:
1. There's no future-proof way to build dashboards including objstore metrics across all Cortex services (because we may change prefixes, add new services, etc). May look a remote possibility, but to give a concrete example @jtlisi is working to use it in the ruler (https://github.com/cortexproject/cortex/pull/2543).
2. It's against how Prometheus was designed ([see whole discussion here](https://github.com/thanos-io/thanos/issues/2444))

In this PR I propose to:
1. Remove the metrics prefixing
2. Adding a `{component="name"}` label
3. Instrument the bucket directly in the factory so that we don't forget about it (I forgot in `pkg/querier/blocks_store_queryable.go` - the bug fix is not mentioned in the CHANGELOG because the whole store-gateway work doesn't appear in the CHANGELOG yet)

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
I would prefer to add instrumentation in `tsdb.NewBucketClient` instead of individual object-store specific factories. That way it's in the single place, and we always create our bucket clients through that function anyway. WDYT?

pracucci:
> I would prefer to add instrumentation in tsdb.NewBucketClient instead of individual object-store specific factories. That way it's in the single place, and we always create our bucket clients through that function anyway. WDYT?

I'm OK with it. The reason why I did it for every single backend storage is because the `NewBucketClient` for a specific backend (ie. `s3.NewBucketClient`) are exposed and wrapping the instrumentation there would make sure that none forgets about instrumentation either when we use the factory or the single backend. WDYT?

pstibrany:
> I'm OK with it. The reason why I did it for every single backend storage is because the `NewBucketClient` for a specific backend (ie. `s3.NewBucketClient`) are exposed and wrapping the instrumentation there would make sure that none forgets about instrumentation either when we use the factory or the single backend. WDYT?

I see your point, but one can also use factory methods from Thanos and then get unwrapped bucket clients. I think we should be consistent. Furthermore our own code inside Cortex doesn't do that (apart from tests). But this isn't a big deal.

Thanks for fixing this, it will make working with these metrics easier.

pracucci:
> > I'm OK with it. The reason why I did it for every single backend storage is because the `NewBucketClient` for a specific backend (ie. `s3.NewBucketClient`) are exposed and wrapping the instrumentation there would make sure that none forgets about instrumentation either when we use the factory or the single backend. WDYT?
> 
> I see your point, but one can also use factory methods from Thanos and then get unwrapped bucket clients. I think we should be consistent. Furthermore our own code inside Cortex doesn't do that (apart from tests). But this isn't a big deal.
> 
> Thanks for fixing this, it will make working with these metrics easier.

I've applied your suggestion. Thanks for the review!

PR #2572, opened 2020-05-08
TSDB: chunks caching

pstibrany:
**What this PR does**: this PR integrates latest Thanos master and allows bucket stores (used by querier and gateway) to be configured with "chunks caching bucket".

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/operations/blocks-storage.md
+++ b/docs/operations/blocks-storage.md
@@ -262,6 +262,70 @@ tsdb:
       # CLI flag: -experimental.tsdb.bucket-store.index-cache.postings-compression-enabled
       [postings_compression_enabled: <boolean> | default = false]
 
+    caching_bucket:
+      # Caching bucket backend type. Supported values: memcached.
+      # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.backend
+      [backend: <string> | default = ""]
+
+      memcached:
+        # Comma separated list of memcached addresses. Supported prefixes are:
+        # dns+ (looked up as an A/AAAA query), dnssrv+ (looked up as a SRV
+        # query, dnssrvnoa+ (looked up as a SRV query, with no A/AAAA lookup
+        # made after that).
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.addresses
+        [addresses: <string> | default = ""]
+
+        # The socket read/write timeout.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.timeout
+        [timeout: <duration> | default = 100ms]
+
+        # The maximum number of idle connections that will be maintained per
+        # address.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.max-idle-connections
+        [max_idle_connections: <int> | default = 16]
+
+        # The maximum number of concurrent asynchronous operations can occur.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.max-async-concurrency
+        [max_async_concurrency: <int> | default = 50]
+
+        # The maximum number of enqueued asynchronous operations allowed.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.max-async-buffer-size
+        [max_async_buffer_size: <int> | default = 10000]
+
+        # The maximum number of concurrent connections running get operations.
+        # If set to 0, concurrency is unlimited.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.max-get-multi-concurrency
+        [max_get_multi_concurrency: <int> | default = 100]
+
+        # The maximum number of keys a single underlying get operation should
+        # run. If more keys are specified, internally keys are splitted into
+        # multiple batches and fetched concurrently, honoring the max
+        # concurrency. If set to 0, the max batch size is unlimited.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.max-get-multi-batch-size
+        [max_get_multi_batch_size: <int> | default = 0]
+
+        # The maximum size of an item stored in memcached. Bigger items are not
+        # stored. If set to 0, no maximum size is enforced.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.memcached.max-item-size
+        [max_item_size: <int> | default = 1048576]
+
+      caching:
+        # Subrange size used for caching chunks.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.chunks.subrange-size
+        [chunk_subrange_size: <int> | default = 16000]
+
+        # Max GetRange requests for chunks.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.chunks.max-getrange-requests
+        [max_chunks_get_range_requests: <int> | default = 3]
+
+        # TTL for caching object size for chunks.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.chunks.object-size-ttl
+        [chunk_object_size_ttl: <duration> | default = 24h]
+
+        # TTL for caching individual chunks subranges.
+        # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.chunks.subrange-ttl
+        [chunk_subrange_ttl: <duration> | default = 24h]
+
     # Duration after which the blocks marked for deletion will be filtered out
     # while fetching blocks. The idea of ignore-deletion-marks-delay is to
     # ignore blocks that are marked for deletion with some delay. This ensures
--- a/docs/operations/blocks-storage.md
+++ b/docs/operations/blocks-storage.md
@@ -74,7 +74,7 @@ Whenever the pool of compactors increase or decrease (ie. following up a scale u
 
 ## Index cache
 
-The querier supports a cache to speed up postings and series lookups from TSDB blocks indexes. Two backends are supported:
+The querier and store-gateway now supports a cache to speed up postings and series lookups from TSDB blocks indexes. Two backends are supported:
 
 - `inmemory`
 - `memcached`
@@ -102,6 +102,14 @@ For example, if you're running Memcached in Kubernetes, you may:
 1. Deploy your Memcached cluster using a [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/)
 2. Create an [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) for Memcached StatefulSet
 3. Configure the Cortex's Memcached client address using the `dnssrvnoa+` [service discovery](../configuration/arguments.md#dns-service-discovery)
+
+## Chunks cache
+
+Store-gateway and querier now also support cache for storing chunks fetched from storage. Chunks contain actual samples, and can be reused if user query needs same data.
+
+To enable chunks cache, please set `experimental.tsdb.bucket-store.chunks-cache.backend`. Chunks can currently only be stored into Memcached cache. Memcached client can be configured via flags with `experimental.tsdb.bucket-store.chunks-cache.memcached` prefix.
+
+There are additional low-level options for configuring chunks cache. Please refer to other flags with `experimental.tsdb.bucket-store.chunks-cache` prefix.
 
 ## Configuration
 
@@ -262,9 +270,9 @@ tsdb:
       # CLI flag: -experimental.tsdb.bucket-store.index-cache.postings-compression-enabled
       [postings_compression_enabled: <boolean> | default = false]
 
-    caching_bucket:
-      # Caching bucket backend type. Supported values: memcached.
-      # CLI flag: -experimental.tsdb.bucket-store.caching-bucket.backend
+    chunks_cache:
+      # Backend for chunks cache. Supported values: memcached.
+      # CLI flag: -experimental.tsdb.bucket-store.chunks-cache.backend
       [backend: <string> | default = ""]
 
       memcached:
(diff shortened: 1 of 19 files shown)


pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,73 @@
...
+	cfg.RegisterFlagsWithPrefix(f, "experimental.tsdb.bucket-store.index-cache.")
+}
+
+func (cfg *CachingBucketConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
+	f.StringVar(&cfg.Backend, prefix+"backend", "", fmt.Sprintf("Caching bucket backend type. Supported values: %s.", storecache.MemcachedBucketCacheProvider))
+
+	cfg.Memcached.RegisterFlagsWithPrefix(f, prefix+"memcached.")
+	cfg.CachingConfig = storecache.DefaultCachingBucketConfig()
+	f.IntVar(&cfg.CachingConfig.MaxChunksGetRangeRequests, prefix+"max-chunks-get-range-requests", cfg.CachingConfig.MaxChunksGetRangeRequests, "Max GetRange requests for chunks.")
+	f.Int64Var(&cfg.CachingConfig.ChunkSubrangeSize, prefix+"chunk-subrange-size", cfg.CachingConfig.ChunkSubrangeSize, "Subrange size used for caching chunks.")
```
pracucci, line 33:
```suggestion
	f.Int64Var(&cfg.CachingConfig.ChunkSubrangeSize, prefix+"chunk-subrange-size", cfg.CachingConfig.ChunkSubrangeSize, "Size of each subrange a bucket object is splitted into for better caching.")
```

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,73 @@
...
+func (cfg *CachingBucketConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix(f, "experimental.tsdb.bucket-store.index-cache.")
+}
+
+func (cfg *CachingBucketConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
+	f.StringVar(&cfg.Backend, prefix+"backend", "", fmt.Sprintf("Caching bucket backend type. Supported values: %s.", storecache.MemcachedBucketCacheProvider))
+
+	cfg.Memcached.RegisterFlagsWithPrefix(f, prefix+"memcached.")
+	cfg.CachingConfig = storecache.DefaultCachingBucketConfig()
+	f.IntVar(&cfg.CachingConfig.MaxChunksGetRangeRequests, prefix+"max-chunks-get-range-requests", cfg.CachingConfig.MaxChunksGetRangeRequests, "Max GetRange requests for chunks.")
```
pracucci, line 32:
```suggestion
	f.IntVar(&cfg.CachingConfig.MaxChunksGetRangeRequests, prefix+"max-chunks-get-range-requests", cfg.CachingConfig.MaxChunksGetRangeRequests, "Max number of sub-GetRange requests a single GetRange request can be splitted into when fetching chunks. This is the max multiplier of the number of GetRange requests that can be issued in the worst case scenario when chunks caching is enabled.")
```

docs/operations/blocks-storage.md
```diff
@@ -262,6 +262,70 @@ tsdb:
       # CLI flag: -experimental.tsdb.bucket-store.index-cache.postings-compression-enabled
       [postings_compression_enabled: <boolean> | default = false]
 
+    caching_bucket:
```
pracucci, line 265: In `docs/operations/blocks-storage.template` we do have a paragraph explaining the index cache. Would be great add a similar one for the chunks caching.

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/cache"
+	"github.com/thanos-io/thanos/pkg/cacheutil"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	storecache "github.com/thanos-io/thanos/pkg/store/cache"
+)
+
+type CachingBucketConfig struct {
+	Backend   string                    `yaml:"backend"`
+	Memcached MemcachedIndexCacheConfig `yaml:"memcached"`
```
pracucci, line 18: Design wise looks a bit weird reusing `MemcachedIndexCacheConfig`. What do you think if we rename `MemcachedIndexCacheConfig` to `MemcachedClientConfig` (or something generic like that) and we move it away from `pkg/storage/tsdb/index_cache.go`?

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,73 @@
...
+
+	CachingConfig storecache.CachingBucketConfig `yaml:"caching"`
+}
+
+func (cfg *CachingBucketConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix(f, "experimental.tsdb.bucket-store.index-cache.")
```
pracucci, line 24: Is `index-cache` intended here? Do we need this `RegisterFlags()` function at all?

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,73 @@
...
+	cfg.Memcached.RegisterFlagsWithPrefix(f, prefix+"memcached.")
+	cfg.CachingConfig = storecache.DefaultCachingBucketConfig()
+	f.IntVar(&cfg.CachingConfig.MaxChunksGetRangeRequests, prefix+"max-chunks-get-range-requests", cfg.CachingConfig.MaxChunksGetRangeRequests, "Max GetRange requests for chunks.")
+	f.Int64Var(&cfg.CachingConfig.ChunkSubrangeSize, prefix+"chunk-subrange-size", cfg.CachingConfig.ChunkSubrangeSize, "Subrange size used for caching chunks.")
+	f.DurationVar(&cfg.CachingConfig.ChunkObjectSizeTTL, prefix+"chunk-object-size-ttl", cfg.CachingConfig.ChunkObjectSizeTTL, "TTL for caching object size for chunks.")
+	f.DurationVar(&cfg.CachingConfig.ChunkSubrangeTTL, prefix+"chunk-subrange-ttl", cfg.CachingConfig.ChunkSubrangeTTL, "TTL for caching individual chunks subranges.")
+}
+
+// Validate the config.
+func (cfg *CachingBucketConfig) Validate() error {
```
pracucci, line 39: Who calls it?

pkg/storage/tsdb/config.go
```diff
@@ -183,6 +184,7 @@ type BucketStoreConfig struct {
 // RegisterFlags registers the BucketStore flags
 func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.IndexCache.RegisterFlagsWithPrefix(f, "experimental.tsdb.bucket-store.index-cache.")
+	cfg.CachingBucket.RegisterFlagsWithPrefix(f, "experimental.tsdb.bucket-store.caching-bucket.")
```
pracucci, line 187: [not a blocker] This is a discussion we already had. I think that for the use it's a bit inconsistent seeing `index-cache` and `caching-bucket`. Technically we're currently having only chunks caching and even when we'll cache metadata we already agreed that we want to store metadata on different memcached server (at least giving the possibility to do it) so I'm wondering if in the Cortex implementation we could go more specific with `chunks-cache`.

_Let's talk about this offline_

pstibrany: I agree with your naming. I think we also need to rethink usage of Thanos’ config object, and introduce our own config object here –– one for chunks, one for metadata. (As a consequence of https://github.com/thanos-io/thanos/pull/2579, even if not merged yet, I already think it will break stuff).

pracucci: Agree on using our own structs. We shouldn't directly use Thanos config structs, while we should use our own which wraps Thanos ones, in order to keep control of our config (ie. avoid config to change whenever Thanos renames stuff, or keep naming more consistent within Cortex).

pstibrany:
@pracucci I've addressed your feedback now, please take a look again as time permits.

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,89 @@
...
+func (cfg *ChunksCacheConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
+	f.StringVar(&cfg.Backend, prefix+"backend", "", fmt.Sprintf("Backend for chunks cache. Supported values: %s.", storecache.MemcachedBucketCacheProvider))
+
+	cfg.Memcached.RegisterFlagsWithPrefix(f, prefix+"memcached.")
+
+	defaultCfg := storecache.DefaultCachingBucketConfig()
+	cfg.SubrangeSize = defaultCfg.ChunkSubrangeSize
+	cfg.ObjectSizeTTL = defaultCfg.ChunkObjectSizeTTL
+	cfg.SubrangeTTL = defaultCfg.ChunkSubrangeTTL
+	cfg.MaxGetRangeRequests = defaultCfg.MaxChunksGetRangeRequests
```
pracucci, line 39: Looks a weird pattern to me updating the `cfg` here. We should just use `defaultCfg` below when we define the flags (ie. use `defaultCfg.MaxChunksGetRangeRequests` instead of `cfg.MaxGetRangeRequests`)

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,89 @@
...
+
+	f.IntVar(&cfg.MaxGetRangeRequests, prefix+"max-chunks-get-range-requests", cfg.MaxGetRangeRequests, "Maximum number of sub-GetRange requests that a single GetRange request can be split into when fetching chunks. Zero or negative value = unlimited number of sub-requests.")
+	f.Int64Var(&cfg.SubrangeSize, prefix+"chunk-subrange-size", cfg.SubrangeSize, "Size of each subrange that bucket object is split into for better caching.")
+	f.DurationVar(&cfg.ObjectSizeTTL, prefix+"chunk-object-size-ttl", cfg.ObjectSizeTTL, "TTL for caching object size for chunks.")
+	f.DurationVar(&cfg.SubrangeTTL, prefix+"chunk-subrange-ttl", cfg.SubrangeTTL, "TTL for caching individual chunks subranges.")
```
pracucci, line 44: These CLI flags are inconsistent with the YAML (ie. `subrange_size` vs `chunk-subrange-size`). We should be consistent with naming.

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,89 @@
...
+	ObjectSizeTTL time.Duration `yaml:"object_size_ttl"`
+	SubrangeTTL   time.Duration `yaml:"subrange_ttl"`
+}
+
+func (cfg *ChunksCacheConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
+	f.StringVar(&cfg.Backend, prefix+"backend", "", fmt.Sprintf("Backend for chunks cache. Supported values: %s.", storecache.MemcachedBucketCacheProvider))
+
+	cfg.Memcached.RegisterFlagsWithPrefix(f, prefix+"memcached.")
+
+	defaultCfg := storecache.DefaultCachingBucketConfig()
```
pracucci, line 35: [not a blocker] Is it a good think dynamically picking defaults from Thanos instead of hard-coding them? I'm asking it because once we'll move blocks storage from experimental to stable, we'll have to avoid any breaking change both in config and defaults: being these defaults dynamic, a change in Thanos will mean a change to Cortex defaults too.

pkg/storage/tsdb/config.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/alecthomas/units"
+	errors2 "github.com/pkg/errors"
```
pracucci, line 12: Can we remove the import of `"errors"` in this file so that we don't have to name it `errors2`?

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,89 @@
+package tsdb
```
pracucci, line 1: Should we name this file `chunks_cache.go` now?

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -0,0 +1,89 @@
...
+	switch chunksConfig.Backend {
+	case "":
+		// No caching.
+		return bkt, nil
+
+	case string(storecache.MemcachedBucketCacheProvider):
+		var memcached cacheutil.MemcachedClient
+		memcached, err := cacheutil.NewMemcachedClientWithConfig(logger, "chunks-cache", chunksConfig.Memcached.ToMemcachedClientConfig(), reg)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create memcached client")
```
pracucci, line 74:
```suggestion
			return nil, errors.Wrapf(err, "failed to create memcached client for chunks-cache")
```

docs/operations/blocks-storage.template
```diff
@@ -74,7 +74,7 @@ Whenever the pool of compactors increase or decrease (ie. following up a scale u
 
 ## Index cache
 
-The querier supports a cache to speed up postings and series lookups from TSDB blocks indexes. Two backends are supported:
+The querier and store-gateway now supports a cache to speed up postings and series lookups from TSDB blocks indexes. Two backends are supported:
```
pracucci, line 77: In the documentation we shouldn't use temporarl references like "now". I would just suggest to remove it:
```suggestion
The querier and store-gateway support a cache to speed up postings and series lookups from TSDB blocks indexes. Two backends are supported:
```

docs/operations/blocks-storage.template
```diff
@@ -103,6 +103,14 @@ For example, if you're running Memcached in Kubernetes, you may:
 2. Create an [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) for Memcached StatefulSet
 3. Configure the Cortex's Memcached client address using the `dnssrvnoa+` [service discovery](../configuration/arguments.md#dns-service-discovery)
 
+## Chunks cache
+
+Store-gateway and querier now also support cache for storing chunks fetched from storage. Chunks contain actual samples, and can be reused if user query needs same data.
```
pracucci, line 108:
```suggestion
Store-gateway and querier also support cache for storing chunks fetched from storage. Chunks contain actual samples, and can be reused if user query hits the same series for the same time range.
```

docs/operations/blocks-storage.template
```diff
@@ -103,6 +103,14 @@ For example, if you're running Memcached in Kubernetes, you may:
...
+
+To enable chunks cache, please set `experimental.tsdb.bucket-store.chunks-cache.backend`. Chunks can currently only be stored into Memcached cache. Memcached client can be configured via flags with `experimental.tsdb.bucket-store.chunks-cache.memcached` prefix.
```
pracucci, line 110: CLI flags should be always prefixed by `-` for clarity.

```suggestion
To enable chunks cache, please set `-experimental.tsdb.bucket-store.chunks-cache.backend`. Chunks can currently only be stored into Memcached cache. Memcached client can be configured via flags with `-experimental.tsdb.bucket-store.chunks-cache.memcached` prefix.
```

docs/operations/blocks-storage.template
```diff
@@ -103,6 +103,14 @@ For example, if you're running Memcached in Kubernetes, you may:
...
+
+There are additional low-level options for configuring chunks cache. Please refer to other flags with `experimental.tsdb.bucket-store.chunks-cache` prefix.
```
pracucci, line 112:
```suggestion
There are additional low-level options for configuring chunks cache. Please refer to other flags with `-experimental.tsdb.bucket-store.chunks-cache` prefix.
```

pkg/storage/tsdb/caching_bucket.go
pstibrany: I've hardcoded the values now (using defaults from Thanos). Following default values Thanos by reusing their "default" function could surprise us, so let's not do that.

pkg/storage/tsdb/caching_bucket.go
pstibrany: I plan to extend this file with metadata caching config soon (separate config struct, with separate cache config, but handled by the same instance of `CachingBucket`), so I would keep it as is for now. WDYT?

pracucci: Sure, it's OK.

PR #2573, opened 2020-05-08
Track metric with the last successful blocks sync, scan and compaction

pracucci:
**What this PR does**:
While working on critical Cortex alerts for the blocks storage I've realised there's no reliable way to alert on few critical conditions: blocks are not successfully scanned, synched and compacted.

In this PR I propose the following metrics which allows to alert if a successful scan/sync/compact hasn't occurred in the last X time:
  * `cortex_compactor_last_successful_run_time`
  * `cortex_querier_blocks_last_successful_sync_time` (when store-gateway is disabled)
  * `cortex_querier_blocks_last_successful_scan_time` (when store-gateway is enabled)
  * `cortex_storegateway_blocks_last_successful_sync_time`

I've also removed the check on `io.EOF` because I can't see a good reason of having it. We can add it anytime if we figure out there's a good reason of having it, but right now I'm wondering if it's a just misprint from the past.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,7 @@
 * [ENHANCEMENT] Experimental TSDB: added the following metrics to the ingester: #2580
   * `cortex_ingester_tsdb_appender_add_duration_seconds`
   * `cortex_ingester_tsdb_appender_commit_duration_seconds`
-* [ENHANCEMENT] Experimental TSDB: added metrics useful to alert on critical conditions of the blocks storage:
+* [ENHANCEMENT] Experimental TSDB: added metrics useful to alert on critical conditions of the blocks storage: #2573
   * `cortex_compactor_last_successful_run_time`
   * `cortex_querier_blocks_last_successful_sync_time` (when store-gateway is disabled)
   * `cortex_querier_blocks_last_successful_scan_time` (when store-gateway is enabled)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,10 +60,10 @@
   * `cortex_ingester_tsdb_appender_add_duration_seconds`
   * `cortex_ingester_tsdb_appender_commit_duration_seconds`
 * [ENHANCEMENT] Experimental TSDB: added metrics useful to alert on critical conditions of the blocks storage: #2573
-  * `cortex_compactor_last_successful_run_time`
-  * `cortex_querier_blocks_last_successful_sync_time` (when store-gateway is disabled)
-  * `cortex_querier_blocks_last_successful_scan_time` (when store-gateway is enabled)
-  * `cortex_storegateway_blocks_last_successful_sync_time`
+  * `cortex_compactor_last_successful_run_timestamp_seconds`
+  * `cortex_querier_blocks_last_successful_sync_timestamp_seconds` (when store-gateway is disabled)
+  * `cortex_querier_blocks_last_successful_scan_timestamp_seconds` (when store-gateway is enabled)
+  * `cortex_storegateway_blocks_last_successful_sync_timestamp_seconds`
 * [BUGFIX] Ruler: Ensure temporary rule files with special characters are properly mapped and cleaned up. #2506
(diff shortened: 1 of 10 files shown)


pkg/querier/blocks_scanner.go
```diff
@@ -72,6 +73,10 @@ func NewBlocksScanner(cfg BlocksScannerConfig, bucketClient objstore.Bucket, log
 			Help:    "The total time it takes to run a full blocks scan across the storage.",
 			Buckets: []float64{1, 10, 20, 30, 60, 120, 180, 240, 300, 600},
 		}),
+		scanLastSuccess: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_querier_blocks_last_successful_scan_time",
+			Help: "The time of the last successful blocks scan.",
```
pstibrany, line 78:
```suggestion
			Help: "Unix timestamp of the last successful blocks scan.",
```

pkg/querier/blocks_scanner.go
```diff
@@ -135,9 +140,12 @@ func (d *BlocksScanner) scan(ctx context.Context) error {
 	return nil
 }
 
-func (d *BlocksScanner) scanBucket(ctx context.Context) error {
+func (d *BlocksScanner) scanBucket(ctx context.Context) (err error) {
```
pstibrany, line 143: Nit: Calling this e.g. “returnErr” would minimize changes and also make it clearer.

pkg/storegateway/bucket_stores.go
```diff
@@ -130,16 +137,18 @@ func (u *BucketStores) syncUsersBlocks(ctx context.Context, f func(context.Conte
 			defer wg.Done()
 
 			for job := range jobs {
-				if err := f(ctx, job.store); err != nil {
-					level.Warn(u.logger).Log("msg", "failed to synchronize TSDB blocks for user", "user", job.userID, "err", err)
+				if jobErr := f(ctx, job.store); jobErr != nil {
+					errsMx.Lock()
+					errs.Add(errors.Wrap(jobErr, fmt.Sprintf("failed to synchronize TSDB blocks for user %s", job.userID)))
```
pstibrany, line 142: We can use `Wrapf` directly.

pkg/storegateway/bucket_stores.go
```diff
@@ -67,6 +69,10 @@ func NewBucketStores(cfg tsdb.Config, filters []block.MetadataFilter, bucketClie
 			Help:    "The total time it takes to perform a sync stores",
 			Buckets: []float64{0.1, 1, 10, 30, 60, 120, 300, 600, 900},
 		}),
+		syncLastSuccess: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+			Name: "blocks_last_successful_sync_time",
+			Help: "The time of the last successful blocks sync.",
```
pstibrany, line 74: Same as before.

CHANGELOG.md
```diff
@@ -59,6 +59,11 @@
 * [ENHANCEMENT] Experimental TSDB: added the following metrics to the ingester: #2580
   * `cortex_ingester_tsdb_appender_add_duration_seconds`
   * `cortex_ingester_tsdb_appender_commit_duration_seconds`
+* [ENHANCEMENT] Experimental TSDB: added metrics useful to alert on critical conditions of the blocks storage: #2573
+  * `cortex_compactor_last_successful_run_time`
+  * `cortex_querier_blocks_last_successful_sync_time` (when store-gateway is disabled)
+  * `cortex_querier_blocks_last_successful_scan_time` (when store-gateway is enabled)
+  * `cortex_storegateway_blocks_last_successful_sync_time`
```
pstibrany, line 66: For consistency with some other Cortex metrics, should these end with `_timestamp_seconds` suffix?

pracucci: Done

PR #2575, opened 2020-05-10
Add Cassandra table options flag

### LINKED ISSUE #2453: Cassandra CompactionStrategy
The compaction strategy of the tables created for Cassandra should be configurable. Currently it defaults to the Cassanda default: "SizeTieredCompactionStrategy".

Some users may want to experiment with other compaction strategies, like "TimeWindowCompactionStrategy", which is not currently possible since Cortex creates tables on its own. 

EDIT: Other table settings, like "default_time_to_live", would also be nice. Perhaps it would be a good idea to allow the user to configure all table options in a config map?

See: https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlCreateTable.html#table_options

kwSeo:
**What this PR does**:
This PR adds `cassandra.table-options` flag and `storage.cassandra.table_options` configuration option to customize Cassandra table options when creating index or chunk table by table-manager.

**Configuration Example**
```yaml
storage:
  engine: "chunks"
  cassandra:
    addresses: "127.0.0.1"
    port: 9042
    keyspace: "cortex"
    consistency: "ONE"
    auth: false
    table_options: "CLUSTERING ORDER BY (range DESC)
      AND comment = 'Important records'
      AND compaction = { 'class' : 'LeveledCompactionStrategy' }"
```

EDIT: rename `table_with` -> `table_options`

Fixes #2453 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -1738,6 +1738,11 @@ cassandra:
   # CLI flag: -cassandra.convict-hosts-on-failure
   [convict_hosts_on_failure: <boolean> | default = true]
 
+  # Table options used to create index or chunk tables.(Default = "": use
+  # default table options of Cassandra
+  # CLI flag: -cassandra.table-with
+  [table_with: <string> | default = ""]
+
 boltdb:
   # Location of BoltDB index files.
   # CLI flag: -boltdb.dir
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -1738,10 +1738,13 @@ cassandra:
   # CLI flag: -cassandra.convict-hosts-on-failure
   [convict_hosts_on_failure: <boolean> | default = true]
 
-  # Table options used to create index or chunk tables.(Default = "": use
-  # default table options of Cassandra
-  # CLI flag: -cassandra.table-with
-  [table_with: <string> | default = ""]
+  # Table options used to create index or chunk tables. This value is used as
+  # plain text in the table `WITH` like this, "CREATE TABLE
+  # <generated_by_cortex> (...) WITH <cassandra.table-options>". For details,
+  # see https://cortexmetrics.io/docs/production/cassandra. (Default = "": use
+  # default table options of your Cassandra)
+  # CLI flag: -cassandra.table-options
+  [table_options: <string> | default = ""]
 
 boltdb:
   # Location of BoltDB index files.
--- a/pkg/chunk/cassandra/storage_client.go
+++ b/pkg/chunk/cassandra/storage_client.go
@@ -48,6 +48,8 @@ type Config struct {
 	QueryConcurrency         int                 `yaml:"query_concurrency"`
 	NumConnections           int                 `yaml:"num_connections"`
 	ConvictHosts             bool                `yaml:"convict_hosts_on_failure"`
+	TableWith                string              `yaml:"table_with"`
+>>>>>>> add the cassandra.table-with flag
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet
@@ -75,6 +77,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.QueryConcurrency, "cassandra.query-concurrency", 0, "Limit number of concurrent queries to Cassandra. (Default is 0: no limit)")
 	f.IntVar(&cfg.NumConnections, "cassandra.num-connections", 2, "Number of TCP connections per host.")
 	f.BoolVar(&cfg.ConvictHosts, "cassandra.convict-hosts-on-failure", true, "Convict hosts of being down on failure.")
+	f.StringVar(&cfg.TableWith, "cassandra.table-with", "", "Table options used to create index or chunk tables.(Default = \"\": use default table options of Cassandra")
 }
 
 func (cfg *Config) Validate() error {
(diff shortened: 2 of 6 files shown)


pkg/chunk/cassandra/table_client.go
```diff
@@ -65,3 +60,17 @@ func (c *tableClient) DescribeTable(ctx context.Context, name string) (desc chun
...
+func (c *tableClient) getCreateTableQuery(desc *chunk.TableDesc) (query string) {
+	query = fmt.Sprintf(`
+		CREATE TABLE IF NOT EXISTS %s (
+			hash text,
+			range blob,
+			value blob,
+			PRIMARY KEY (hash, range)
+		)`, desc.Name)
+	if c.cfg.TableWith != "" {
+		query = fmt.Sprintf("%s WITH %s", query, c.cfg.TableWith)
```
jtlisi, line 73: IMO, joining a string like this for a query to Cassandra is dangerous. If the end goal is to change the compaction strategy, I think that would be better handled with a `CompactionConfig` struct that a parameterizes all of the possible compactions strategies into a config and generates the appropriate query string.

kwSeo: @jtlisi @pracucci 

Thank you for review! I agree your opinion.
And there are some considerations.
Some of table options for Cassandra are related to duration, like this:

```
gc_grace_seconds = 864000
default_time_to_live = 0
tombstone_compaction_interval = 864000
max_sstable_age_days = 100
...
```

I thought table options like above are suitable for `time.Duration` type.
But some of names for table options indicate time unit.
I am not sure if `time.Duration` is suitable.
There is a way to rename the table options but I'm concerned about user confusion. 
I am curious about your opinion.

jtlisi: Since this is specifically for Cassandra,I recommend using types that will make input correspond to Cassandra. That way users already familiar with the system will be able to directly copy config values over.

So in the cases you mentioned above I would use `int` values. Just make sure you document each option in the flag help text.

coelho: In our use case, we are using ScyllaDB, which is compatible but contains many options that are not default in Cassandra. 

https://docs.scylladb.com/getting-started/ddl/
CTRL-F scylla_encryption_options

It is a bad idea to assume what flags the Cassandra server will support. In my opinion this should be made extensible and the current PR does that nicely. 

This is a niche flag that will only be used by those who fully understand the implications of changing the flag, better solved by documentation, and already have authorization on the Cassandra server through the configuration.

kwSeo: @jtlisi 
In addition to @coelho, there may be another use cases, such as DataStax products.
Also, table options depend on the Cassandra version.
It is hard to find all options and consider all differences.
My goals have been just caching and compaction strategy.
But this issue seems to have more considerations than I think.
If you still think joining string is dangerous, it may need another way.

jtlisi: I see your point, and I agree in some cases a separate string makes sense. @pracucci WDYT?

pkg/chunk/cassandra/storage_client.go
```diff
@@ -66,6 +67,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.MinBackoff, "cassandra.retry-min-backoff", 100*time.Millisecond, "Minimum time to wait before retrying a failed request. (Default = 100ms)")
 	f.DurationVar(&cfg.MaxBackoff, "cassandra.retry-max-backoff", 10*time.Second, "Maximum time to wait before retrying a failed request. (Default = 10s)")
 	f.IntVar(&cfg.QueryConcurrency, "cassandra.query-concurrency", 0, "Limit number of concurrent queries to Cassandra. (Default is 0: no limit)")
+	f.StringVar(&cfg.TableWith, "cassandra.table-with", "", "Table options used to create index or chunk tables.(Default = \"\": use default table options of Cassandra")
```
jtlisi, line 70: I would recommend naming this variable TableOptions and the flag `table-options`.

Also the help text should explain that the provided string will be concatenated to all table creation operations.

docs/configuration/config-file-reference.md
```diff
@@ -1689,6 +1689,11 @@ cassandra:
   # CLI flag: -cassandra.query-concurrency
   [query_concurrency: <int> | default = 0]
 
+  # Table options used to create index or chunk tables.(Default = "": use
+  # default table options of Cassandra
+  # CLI flag: -cassandra.table-with
+  [table_with: <string> | default = ""]
```
jtlisi, line 1695: Be sure to update https://github.com/cortexproject/cortex/blob/master/docs/production/storage-cassandra.md with information about this flag. As well as warnings about the potential issues a user could face by misconfiguring it.

pkg/chunk/cassandra/storage_client.go
kwSeo: @jtlisi 
I have renamed `table-with` to `table-options` and improved the help text.

docs/configuration/config-file-reference.md
kwSeo: @jtlisi @coelho
I have written the description about `table_options` to the `storage-cassandra.md`.
I think I am not good at English...
Please check if there is anything strange about the contents.

coelho: LGTM

kwSeo:
@jtlisi 
Thank you for your review.
I have done `rebase`.

PR #2576, opened 2020-05-11
Move signal handling out of Server module.

pstibrany:
This PR integrates latest version of weaveworks/common library, which allows Cortex to [disable signal handling](https://github.com/weaveworks/common/pull/191) in `Server` module.

This simplifies module service setup, as we no longer need to make a distinction between [services when using the wrapper or not](https://github.com/cortexproject/cortex/pull/2542#discussion_r417884587). This also simplifies PR https://github.com/cortexproject/cortex/pull/2559.

**Checklist**
- [ ] Tests updated (tested manually, I'm not sure how to write unit tests for this)
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/server_service.go
+++ b/pkg/cortex/server_service.go
@@ -55,15 +55,15 @@ func NewServerService(serv *server.Server, servicesToWaitFor func() []services.S
 }
 
 func DisableSignalHandling(config *server.Config) {
-	config.SignalHandler = make(dummySignalHandler)
+	config.SignalHandler = make(ignoreSignalHandler)
 }
 
-type dummySignalHandler chan struct{}
+type ignoreSignalHandler chan struct{}
 
-func (dh dummySignalHandler) Loop() {
+func (dh ignoreSignalHandler) Loop() {
 	<-dh
 }
 
-func (dh dummySignalHandler) Stop() {
+func (dh ignoreSignalHandler) Stop() {
 	close(dh)
 }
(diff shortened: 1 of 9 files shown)


pkg/util/module_service.go
```diff
@@ -95,11 +95,12 @@ func (w *moduleService) stop(_ error) error {
 func (w *moduleService) waitForModulesToStop() {
 	// wait until all stopDeps have stopped
 	stopDeps := w.stopDeps(w.name)
-	for _, s := range stopDeps {
+	for n, s := range stopDeps {
 		if s == nil {
 			continue
 		}
 
+		level.Debug(Logger).Log("msg", "module waiting for", "module", w.name, "other_module", n)
```
pracucci, line 103:
```suggestion
		level.Debug(Logger).Log("msg", "module waiting for", "module", w.name, "waiting_for", n)
```

pkg/cortex/server_service.go
```diff
@@ -51,3 +53,17 @@ func NewServerService(serv *server.Server, servicesToWaitFor func() []services.S
 
 	return services.NewBasicService(nil, runFn, stoppingFn)
 }
+
+func DisableSignalHandling(config *server.Config) {
+	config.SignalHandler = make(dummySignalHandler)
+}
+
+type dummySignalHandler chan struct{}
```
pracucci, line 61: Maybe `ignoreSignalHandler` is a better name?

PR #2586, opened 2020-05-12
Fix memory leak in blocks storage ingesters

pracucci:
**What this PR does**:
In a production cluster running the blocks storage we've seen a large portion of the inuse heap retaining memory allocate by `grpc.recvMsg()`. This is caused by the fact that `FromLabelAdaptersToLabelsWithCopy()` doesn't do a deep copy but reuse input strings (which are read-only slice pointers).

In this PR we're fixing it doing a deep copy.

**Credits: the root cause has been spotted and fixed by @pstibrany**. I've just done double checks and opening the PR for it.

Given the same exact test scenario, this is the `recvMsg()` allocations **before the fix**:
Screen Shot 2020-05-12 at 17 29 12

And this is **after the fix**:
Screen Shot 2020-05-12 at 17 24 46

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/client/compat.go
+++ b/pkg/ingester/client/compat.go
@@ -229,15 +229,19 @@ func FromLabelAdaptersToLabels(ls []LabelAdapter) labels.Labels {
 // get in input labels whose data structure is reused.
 func FromLabelAdaptersToLabelsWithCopy(input []LabelAdapter) labels.Labels {
 	result := make(labels.Labels, len(input))
-
 	for i, l := range input {
 		result[i] = labels.Label{
-			Name:  l.Name,
-			Value: l.Value,
+			Name:  copyString(l.Name),
+			Value: copyString(l.Value),
 		}
 	}
-
 	return result
+}
+
+func copyString(in string) string {
+	out := make([]byte, len(in))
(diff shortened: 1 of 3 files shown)


pkg/ingester/client/compat.go
```diff
@@ -229,17 +229,21 @@ func FromLabelAdaptersToLabels(ls []LabelAdapter) labels.Labels {
...
-			Value: l.Value,
+			Name:  copyString(l.Name),
+			Value: copyString(l.Value),
 		}
 	}
-
 	return result
 }
 
+func copyString(in string) string {
```
bboreham, line 241: In pkg/util/extract/extract.go we do ` string([]byte(label.Value))`

pstibrany: I know, I was just afraid that Go compiler can easily optimize that away as no-op.

bboreham: Also in `pkg/ingester/index/index.go` yay DRY.

bboreham:
Very similar to #1897. 

In the 'v1' ingester, `index.add()` does the `copyString()`.

PR #2587, opened 2020-05-12
Automatically forget unhealthy Rulers

### LINKED ISSUE #2058: rulers are never removed from the ring if evicted

jtlisi:
**What this PR does**:
- Reuse pattern used by the store-gateway ring to enable the ruler to autoforget instances

Fixes #2058 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ruler/lifecycle.go
+++ b/pkg/ruler/lifecycle.go
@@ -5,9 +5,9 @@ import (
 )
 
 func (r *Ruler) OnRingInstanceRegister(_ *ring.BasicLifecycler, ringDesc ring.Desc, instanceExists bool, instanceID string, instanceDesc ring.IngesterDesc) (ring.IngesterState, ring.Tokens) {
-	// When we initialize the store-gateway instance in the ring we want to start from
-	// a clean situation, so whatever is the state we set it JOINING, while we keep existing
-	// tokens (if any) or the ones loaded from file.
+	// When we initialize the ruler instance in the ring we want to start from
+	// a clean situation, so whatever is the state we set it ACTIVE, while we keep existing
+	// tokens (if any).
 	var tokens []uint32
 	if instanceExists {
 		tokens = instanceDesc.GetTokens()
@@ -19,7 +19,7 @@ func (r *Ruler) OnRingInstanceRegister(_ *ring.BasicLifecycler, ringDesc ring.De
 	// Tokens sorting will be enforced by the parent caller.
 	tokens = append(tokens, newTokens...)
 
-	return ring.JOINING, tokens
+	return ring.ACTIVE, tokens
 }
 
 func (r *Ruler) OnRingInstanceTokens(_ *ring.BasicLifecycler, _ ring.Tokens) {}
--- a/pkg/ruler/lifecycle.go
+++ b/pkg/ruler/lifecycle.go
@@ -1,16 +1,28 @@
 package ruler
 
 import (
-	"context"
+	"github.com/cortexproject/cortex/pkg/ring"
 )
 
-// TransferOut is a noop for the ruler
-func (r *Ruler) TransferOut(ctx context.Context) error {
-	return nil
+func (r *Ruler) OnRingInstanceRegister(_ *ring.BasicLifecycler, ringDesc ring.Desc, instanceExists bool, instanceID string, instanceDesc ring.IngesterDesc) (ring.IngesterState, ring.Tokens) {
+	// When we initialize the store-gateway instance in the ring we want to start from
+	// a clean situation, so whatever is the state we set it JOINING, while we keep existing
+	// tokens (if any) or the ones loaded from file.
+	var tokens []uint32
+	if instanceExists {
+		tokens = instanceDesc.GetTokens()
+	}
+
+	_, takenTokens := ringDesc.TokensFor(instanceID)
+	newTokens := ring.GenerateTokens(r.cfg.Ring.NumTokens-len(tokens), takenTokens)
+
+	// Tokens sorting will be enforced by the parent caller.
+	tokens = append(tokens, newTokens...)
+
+	return ring.JOINING, tokens
 }
 
-// Flush triggers a flush of all the work items currently
-// scheduled by the ruler, currently every ruler will
-// query a backend rule store for it's rules so no
-// flush is required.
-func (r *Ruler) Flush() {}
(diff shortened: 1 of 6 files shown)


pkg/ruler/ruler.go
```diff
@@ -187,26 +188,50 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable promStorage.Queryable
 		logger:       logger,
 	}
 
+	if cfg.EnableSharding {
+		ringStore, err := kv.NewClient(cfg.Ring.KVStore, ring.GetCodec())
+		if err != nil {
+			return nil, errors.Wrap(err, "create KV store client")
+		}
+
+		if enableSharding(ruler, ringStore) != nil {
```
pracucci, line 197: You're not catching the correct error here:

```suggestion
		if err = enableSharding(ruler, ringStore); err != nil {
```

pkg/ruler/ruler.go
```diff
@@ -187,26 +188,50 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable promStorage.Queryable
...
+	lifecyclerCfg, err := r.cfg.Ring.ToLifecyclerConfig()
+	if err != nil {
+		return errors.Wrap(err, "failed to initialize ruler's lifecycler config")
+	}
+
+	// Define lifecycler delegates in reverse order (last to be called defined first because they're
+	// chained via "next delegate").
+	delegate := ring.BasicLifecyclerDelegate(r)
+	delegate = ring.NewLeaveOnStoppingDelegate(delegate, r.logger)
+	delegate = ring.NewAutoForgetDelegate(r.cfg.Ring.HeartbeatTimeout, delegate, r.logger)
```
pracucci, line 216: I think this is not correct. I guess you need to apply `ringAutoForgetUnhealthyPeriods` too.

```suggestion
	delegate = ring.NewAutoForgetDelegate(r.cfg.Ring.HeartbeatTimeout * ringAutoForgetUnhealthyPeriods, delegate, r.logger)
```

pkg/ruler/lifecycle.go
```diff
@@ -1,16 +1,28 @@
...
+	"github.com/cortexproject/cortex/pkg/ring"
 )
 
-// TransferOut is a noop for the ruler
-func (r *Ruler) TransferOut(ctx context.Context) error {
-	return nil
+func (r *Ruler) OnRingInstanceRegister(_ *ring.BasicLifecycler, ringDesc ring.Desc, instanceExists bool, instanceID string, instanceDesc ring.IngesterDesc) (ring.IngesterState, ring.Tokens) {
+	// When we initialize the store-gateway instance in the ring we want to start from
+	// a clean situation, so whatever is the state we set it JOINING, while we keep existing
+	// tokens (if any) or the ones loaded from file.
```
pracucci, line 10: The `NewTokensPersistencyDelegate()` is not used, so you should remove "or the ones loaded from file" because loading tokens from file is not supported by the ruler.

pkg/ruler/lifecycle.go
```diff
@@ -1,16 +1,28 @@
...
+		tokens = instanceDesc.GetTokens()
+	}
+
+	_, takenTokens := ringDesc.TokensFor(instanceID)
+	newTokens := ring.GenerateTokens(r.cfg.Ring.NumTokens-len(tokens), takenTokens)
+
+	// Tokens sorting will be enforced by the parent caller.
+	tokens = append(tokens, newTokens...)
+
+	return ring.JOINING, tokens
```
pracucci, line 22: The basic lifecycler, by design, never change the state of an instance in the ring, so the ruler instance will never switch to `ACTIVE`. Considering how the ruler works, I think there's no need to transition to `JOINING` so the easiest thing to do here is to just switch it to `ACTIVE` (if you do so, remember to update the comment at the beginning of the function).

pkg/ruler/lifecycle_test.go
```diff
@@ -2,73 +2,102 @@ package ruler
...
+	require.NoError(t, services.StartAndAwaitRunning(ctx, r))
+	defer services.StopAndAwaitTerminated(ctx, r) //nolint:errcheck
+
 	// Wait until the tokens are registered in the ring
 	test.Poll(t, 100*time.Millisecond, config.Ring.NumTokens, func() interface{} {
-		return testutils.NumTokens(config.Ring.KVStore.Mock, "localhost", ring.RulerRingKey)
+		return testutils.NumTokens(ringStore, "localhost", ring.RulerRingKey)
 	})
 
 	require.NoError(t, services.StopAndAwaitTerminated(context.Background(), r))
```
pracucci, line 42: Could you also test the instance is `ACTIVE` in the ring?

pkg/ruler/ruler.go
jtlisi: Good catch!

pkg/ruler/ruler.go
jtlisi: 👍 Makes sense. Shouldn't have omitted it

pkg/ruler/lifecycle.go
jtlisi: 👍 Gotcha, I just copied this block and didn't read the comments very closely. It actually still references the store-gateway

pkg/ruler/lifecycle_test.go
jtlisi: I'll update the test to do so

jtlisi:
I update the PR per the review comments. I will ping for another review after I manually test it.

jtlisi:
Rolled out some rulers with a very low memory limit and they started crash looping: I then deleted the pod and verified the unhealthy instance was in the ring.
### Before
Screen Shot 2020-05-14 at 1 51 06 AM
### Moments Later
Screen Shot 2020-05-14 at 1 51 54 AM

This should be ready for review. The basiclifecycler made this a breeze. Really great feature.

jtlisi:
@gouthamve The ruler keeps track of all the failed attempts to the ring to verify the ownership of a ruler group and stores it in this metric: `cortex_ruler_ring_check_errors_total`

PR #2590, opened 2020-05-13
Properly fix the wrong logging of 200 when write to client fails

gouthamve:
This pulls in weaveworks/common#187 and reverts #2483 

Before:
```
2020-05-12 16:14:31.381381 I | http: superfluous response.WriteHeader call from github.com/opentracing-contrib/go-stdlib/nethttp.(*statusCodeTracker).WriteHeader (status-code-tracker.go:19)
level=warn ts=2020-05-12T15:14:31.381571716Z caller=logging.go:49 traceID=1bce8fc8c1a57be7 msg="GET /api/v1/query?query=up[10000m]&start=1557759455&step=126144 (500) 729.261961ms Response: \"write tcp 127.0.0.1:9091->127.0.0.1:54276: write: broken pipe\\n\" ws: false; Accept: */*; User-Agent: curl/7.70.0; "
```
After:
```
level=warn ts=2020-05-12T15:13:17.161683485Z caller=logging.go:51 traceID=604b1b0b5218a0e7 msg="GET /api/v1/query?query=up[10000m]&start=1557759455&step=126144 728.757586ms, error: write tcp 127.0.0.1:9091->127.0.0.1:54270: write: broken pipe ws: false; Accept: */*; User-Agent: curl/7.70.0; "
```
The new log is more accurate (the status code in this case is not 500), and also doesn't log the duplicate header write golang error.


pkg/querier/frontend/frontend.go
```diff
@@ -192,11 +192,7 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 		hs[h] = vs
 	}
 	w.WriteHeader(resp.StatusCode)
-
-	if _, err = io.Copy(w, resp.Body); err != nil {
-		server.WriteError(w, err)
-		return
-	}
+	io.Copy(w, resp.Body)
```
pstibrany, line 195: We cannot do much about errors returned here, but doing nothing seems like too little. We could log it, which may have benefit of being logged with traceID so it can be tracked back to specific clients (in case someone is complaining).

pstibrany: Ah, sorry I see it's logged by middleware. Ignore.

vendor/github.com/weaveworks/common/server/server.go
```diff
@@ -31,6 +31,16 @@ import (
 	"github.com/weaveworks/common/signals"
 )
 
+// SignalHandler used by Server.
+type SignalHandler interface {
```
pstibrany, line 35: This is going to be merged by PR https://github.com/cortexproject/cortex/pull/2576. I suggest merge them in order.

PR #2592, opened 2020-05-14
Add metric to monitor cert expiry time

annanay25:
**What this PR does**:
Adds metric to monitor TLS client cert expiry time.

Follow up from #2502 

**Checklist**
- NA Tests updated
- NA Documentation added

### DIFF
--- a/pkg/util/tls/tls.go
+++ b/pkg/util/tls/tls.go
@@ -7,8 +7,17 @@ import (
 	"io/ioutil"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+)
+
+var (
+	tlsCertNotAfterTimestamp = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "cortex",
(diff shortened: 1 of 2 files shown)


pstibrany:
There are tools for this already, why have it in Cortex?

pkg/util/tls/tls.go
```diff
@@ -41,6 +50,13 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
 		caCertPool = x509.NewCertPool()
 		caCertPool.AppendCertsFromPEM(caCert)
 		if len(clientCert.Certificate) > 0 && caCertPool != nil {
+			prometheus.MustRegister(tlsCertNotAfterTimestamp)
```
pstibrany, line 53: Please use new `promauto` package and give it a registrerer.

pracucci: It's not even correct register it here, because if the `GetTLSConfig()` is called twice, it will panic because of a double registration. You can just use `promauto.NewGaugeVec()` above, even if we should stop using the global prometheus registerer, but if you really want this metric then using the global registry is the easiest option given the current design.

jtlisi:
I'm closing this for now since the general consensus is an external tool should be used to monitor this.

PR #2593, opened 2020-05-14
Add blocks storage consistency check in the querier

pracucci:
**What this PR does**:
This is the first PR to introduce a consistency check run by the querier to ensure all expected blocks have been queried through the store-gateways.

The main problem of the current implementation (reason why I've introduced a temporarily flag to disable it) is that there's no way to know the timestamp of when a block has been uploaded to the storage and so there's no reliable way to exclude from the check "recently uploaded blocks". The current implementation uses the time when the block ID has been generated which could be way different than when the block has completed the upload (a.k.a. the metafile has been uploaded). I will work to add this capability to Thanos first so that we can improve this logic.

A couple of notes:
- Added `BlocksStoreClient` interface to have a clean way to get the store-gateway remote address (both with the real client implementation and in tests with the mocked one)
- `BlocksConsistencyChecker` picks queried blocks splitted by store-gateway address in preparation for the next PR where we'll try to re-run a second fetch for the missing blocks (if replication is enabled)

Out of the scope of this PR (will be done in separate PRs):
- Re-run the query a second time to query missing blocks on different store-gateway instances (if replication factor > 1)
- Update documentation

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci:
This is an example of the trace span logging some information about the queried blocks:
Screen Shot 2020-05-14 at 12 06 25

pkg/querier/blocks_consistency_checker.go
```diff
@@ -0,0 +1,90 @@
...
+
+	for _, blockID := range expectedBlocks {
+		// Some recently uploaded blocks, already discovered by the querier, may not have been discovered
+		// and loaded by the store-gateway yet. In order to avoid false positives, we grant some time
+		// to the store-gateway to discover them. It's safe to exclude recently uploaded blocks because:
+		// - Blocks uploaded by ingesters: we will continue querying them from ingesters for a while (depends
+		//   on the configured retention period).
+		// - Blocks uploaded by compactor: the source blocks are marked for deletion but will continue to be
+		//   queried by store-gateways for a while (depends on the configured deletion marks delay).
+		if ulid.Now()-blockID.Time() < uint64(c.uploadGracePeriod/time.Millisecond) {
```
pstibrany, line 63:
```suggestion
		if ulid.Now()-blockID.Time() < uint64(c.uploadGracePeriod.Milliseconds()) {
```

pkg/querier/blocks_consistency_checker.go
```diff
@@ -0,0 +1,90 @@
...
+			continue
+		}
+
+		// The store-gateway may offload blocks before the querier. If that happen, the querier will run a consistency check
+		// on blocks that can't be queried because offloaded. For this reason, we don't run the consistency check on any block
+		// which has been marked for deletion more then "grace period" time ago. Basically, the grace period is the time
+		// we still expect a block marked for deletion to be still queried.
+		if mark := knownDeletionMarks[blockID]; mark != nil {
+			if time.Since(time.Unix(mark.DeletionTime, 0)).Seconds() > c.deletionGracePeriod.Seconds() {
```
pstibrany, line 72:
```suggestion
			if time.Since(time.Unix(mark.DeletionTime, 0)) > c.deletionGracePeriod {
```

pkg/querier/blocks_consistency_checker.go
```diff
@@ -0,0 +1,90 @@
...
+		// - Blocks uploaded by ingesters: we will continue querying them from ingesters for a while (depends
+		//   on the configured retention period).
+		// - Blocks uploaded by compactor: the source blocks are marked for deletion but will continue to be
+		//   queried by store-gateways for a while (depends on the configured deletion marks delay).
+		if ulid.Now()-blockID.Time() < uint64(c.uploadGracePeriod/time.Millisecond) {
+			continue
+		}
+
+		// The store-gateway may offload blocks before the querier. If that happen, the querier will run a consistency check
+		// on blocks that can't be queried because offloaded. For this reason, we don't run the consistency check on any block
```
pstibrany, line 68:
```suggestion
		// The store-gateway may offload blocks before the querier. If that happens, the querier will run a consistency check
		// on blocks that can't be queried because they were offloaded. For this reason, we don't run the consistency check on any block
```

pkg/querier/blocks_consistency_checker.go
```diff
@@ -0,0 +1,90 @@
...
+			missingBlockIDs = append(missingBlockIDs, id)
+		}
+	}
+
+	if len(missingBlocks) == 0 {
+		return nil
+	}
+
+	c.checksFailed.Inc()
+	return fmt.Errorf("consistency check failed because of non-queried blocks: %s", strings.Join(missingBlockIDs, " "))
```
pstibrany, line 89:
```suggestion
	return fmt.Errorf("consistency check failed because some blocks were not queried: %s", strings.Join(missingBlockIDs, " "))
```

docs/configuration/config-file-reference.md
```diff
@@ -659,6 +659,19 @@ store_gateway_client:
...
+blocks_consistency_check:
+  # Whether the querier should run a consistency check to ensure all expected
+  # blocks have been queried.
+  # CLI flag: -experimental.querier.blocks-consistency-check.enabled
+  [enabled: <boolean> | default = true]
+
+  # The grace period allowed before a new block is included in the consistency
+  # check.
+  # CLI flag: -experimental.querier.blocks-consistency-check.upload-grace-period
+  [upload_grace_period: <duration> | default = 1h]
```
gouthamve, line 674: The default seems excessively high?

pkg/util/timeutil/time.go
```diff
@@ -0,0 +1,16 @@
+package timeutil
+
+import "time"
+
+const (
+	nanosecondsInMillisecond = int64(time.Millisecond / time.Nanosecond)
+)
+
+// TimeFromMillis is a helper to turn milliseconds -> time.Time
```
gouthamve, line 9: We already have similar functions in `pkg/util/time.go`, can we use that package instead of a new file?

docs/configuration/config-file-reference.md
pracucci: Unfortunately we have to protect from the compactor as well, and the block ID is generated at the beginning of the compaction so it could easily pass 1h. However, this `upload_grace_period` will soon disappear (it's a temporarily hack) as soon as we have this done (I'm going to work on it right now): https://github.com/thanos-io/thanos/issues/2585

pkg/util/timeutil/time.go
pracucci: Definitely! I totally missed it.

PR #2600, opened 2020-05-15
query-tee: Support metadata, rules and alerts routes

gotjosh:
**What this PR does**:

Allows `query-tee` to support `/metadata`, `/alerts`, and `/rules` endpoints.

# The problem

Instead of opening an issue, I decided to just go ahead and establish the discussion on this PR. This is a bit tricky given the `/metadata` and `/rules` endpoints don't really query Prometheus but instead are served from memory or a file, no "queries" are involved. Which makes this semantically incorrect.

However, `query-tee` is typically deployed as a shim - which means that if you're using something like Grafana, the endpoint of your data source will be the `query-tee` URL. Grafana will fail to access these endpoints given `query-tee` only proxies certain routes. 

An alternative to this might be to allow `query-tee` to operate in a whitelist or blacklist mode where URLs not supported can either be blocked or forwarded. Happy to take it in whichever direction maintainers decide. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@
 * [FEATURE] Experimental: Added support for `/api/v1/metadata` Prometheus-based endpoint. #2549
 * [FEATURE] Add ability to limit concurrent queries to Cassandra with `-cassandra.query-concurrency` flag. #2562
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
-* [ENHANCEMENT] `query-tee` supports `/metadata`, `/alerts`, and `/rules` #XXXX
+* [ENHANCEMENT] `query-tee` supports `/metadata`, `/alerts`, and `/rules` #2600
 * [ENHANCEMENT] Ruler: Automatically remove unhealthy rulers from the ring. #2587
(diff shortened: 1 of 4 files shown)


pkg/ruler/storage.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/ruler/rules/objectclient"
 )
 
-// RuleStoreConfig conigures a rule store
```
gotjosh, line 20: This is unrelated, but found it whilst browsing and thought it would be small enough to just pass-by within this PR.

gotjosh:
cc: @pracucci

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 * [FEATURE] Experimental: Added support for `/api/v1/metadata` Prometheus-based endpoint. #2549
 * [FEATURE] Add ability to limit concurrent queries to Cassandra with `-cassandra.query-concurrency` flag. #2562
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
+* [ENHANCEMENT] `query-tee` supports `/metadata`, `/alerts`, and `/rules` #XXXX
```
pracucci, line 37:
```suggestion
* [ENHANCEMENT] `query-tee` supports `/metadata`, `/alerts`, and `/rules` #2600
```

PR #2609, opened 2020-05-19
Add per-tenant max cache freshness to query-frontend

annanay25:
**What this PR does**:
Add `frontend.per-user-max-cache-freshness` flag to query-frontend, which allows users to configure per-tenant max-cache-freshness values.
- By default the globally configured `frontend.max-cache-freshness` value will be applied and hence there will be **no breaking changes**.
- If however, a per-tenant value is specified for a specific tenant, the globally configured value **will** be overridden.

Fixes  NA

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/querier/queryrange/results_cache.go
```diff
@@ -184,7 +189,7 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
 	}
 
 	if err == nil && len(extents) > 0 {
-		extents, err := s.filterRecentExtents(r, extents)
+		extents, err := s.filterRecentExtents(ctx, r, extents)
```
cyriltovena, line 192: Can you pass the duration instead. ? to avoid calling this twice `s.limits.MaxCacheFreshness(userID)` ?

docs/configuration/config-file-reference.md
```diff
@@ -2366,6 +2366,11 @@ The `limits_config` configures default and per-tenant limits imposed by Cortex s
 # CLI flag: -store.cardinality-limit
 [cardinality_limit: <int> | default = 100000]
 
+# Most recent allowed cacheable result per-tenant, to prevent caching very
+# recent results that might still be in flux.
+# CLI flag: -frontend.per-user-max-cache-freshness
+[max_cache_freshness: <duration> | default = 0s]
```
joe-elliott, line 2372: struggling with this name.  i would not have guessed the functionality based on the name.

cache_lockout_duration?  dunno.

it also kind of looks like it was already named from the diff so it may not be changeable.

annanay25: I can change the name where I'm defining the config parameter [here](https://github.com/cortexproject/cortex/pull/2609/files#diff-16c9f0ec82b1590ddc2eb9f5b5df9f51R63), but the same flag exists in results cache config as well, so the naming seems to be approved there!

pkg/querier/queryrange/results_cache.go
annanay25: Sure, fixing this.

docs/configuration/config-file-reference.md
pracucci: I agree this naming should match the global one we have, so I agree keeping it as is.

pkg/util/validation/limits.go
```diff
@@ -103,6 +104,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&l.MaxQueryLength, "store.max-query-length", 0, "Limit to length of chunk store queries, 0 to disable.")
 	f.IntVar(&l.MaxQueryParallelism, "querier.max-query-parallelism", 14, "Maximum number of queries will be scheduled in parallel by the frontend.")
 	f.IntVar(&l.CardinalityLimit, "store.cardinality-limit", 1e5, "Cardinality limit for index queries.")
+	f.DurationVar(&l.MaxCacheFreshness, "frontend.per-user-max-cache-freshness", 0, "Most recent allowed cacheable result per-tenant, to prevent caching very recent results that might still be in flux.")
```
pracucci, line 107: I think this CLI flag should replace the existing one `frontend.max-cache-freshness`, instead of introducing a new one (at this, this is how all other flags in this struct work). What do you think?

pkg/querier/queryrange/results_cache.go
```diff
@@ -171,7 +171,12 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
 		response Response
 	)
 
-	maxCacheTime := int64(model.Now().Add(-s.cfg.MaxCacheFreshness))
+	// check if per-tenant cache freshness value is provided
+	maxCacheFreshness := s.limits.MaxCacheFreshness(userID)
+	if maxCacheFreshness == time.Duration(0) {
+		maxCacheFreshness = s.cfg.MaxCacheFreshness
+	}
```
pracucci, line 178: You have to do this because now there are two flags doing the same thing:
- `-frontend.per-user-max-cache-freshness` (new one, which allows overrides)
- `-frontend.max-cache-freshness` (old one)

I think this is not correct from a design perspective. We shouldn't introduce a new flag, but making the old one supporting overrides.

pkg/util/validation/limits.go
annanay25: That would involve a breaking change to the config file, as the flag would now be a part of the `Limits` struct. If I understand correctly, we should support the old flag for a few releases (I can add a "deprecated" in the flag description).

pracucci: Moving the flag (without renaming it) doesn't introduce any breaking change to the flag itself, but would introduce a breaking change in the YAML config. To overcome the breaking change in the YAML config, you can keep the old config in YAML (without registering its CLI flag), mark it as deprecated in the CHANGELOG and add a `// TODO` comment to remove it in Cortex 1.4.0.

pkg/querier/queryrange/results_cache.go
```diff
@@ -40,16 +40,14 @@ type CacheGenNumberLoader interface {
 // ResultsCacheConfig is the config for the results cache.
 type ResultsCacheConfig struct {
 	CacheConfig       cache.Config  `yaml:"cache"`
-	MaxCacheFreshness time.Duration `yaml:"max_freshness"`
+	MaxCacheFreshness time.Duration `yaml:"max_freshness" doc:"hidden"` // TODO: (deprecated) remove in Cortex v1.4.0
```
pracucci, line 43: Could we name the variable `LegacyMaxCacheFreshness` for clarity?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * [CHANGE] Query Frontend now uses Round Robin to choose a tenant queue to service next. #2553
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
+* [FEATURE] The flag `frontend.max-cache-freshness` can now be used to specify per-tenant max cache freshness values. The corresponding YAML config parameter has been changed from `results_cache.max_freshness` to `limits_config.max_cache_freshness`. The legacy YAML config parameter (`results_cache.max_freshness`) will continue to be supported till Cortex release `v1.4.0`. #2609
```
pracucci, line 7: Please always prefix flags with `-` for clarity. I've also reworded it a bit, because overrides are possible only via YAML.

```suggestion
* [FEATURE] The flag `-frontend.max-cache-freshness` is now supported within the limits overrides, to specify per-tenant max cache freshness values. The corresponding YAML config parameter has been changed from `results_cache.max_freshness` to `limits_config.max_cache_freshness`. The legacy YAML config parameter (`results_cache.max_freshness`) will continue to be supported till Cortex release `v1.4.0`. #2609
```

docs/configuration/config-file-reference.md
```diff
@@ -741,11 +741,6 @@ results_cache:
     # The CLI flags prefix for this block config is: frontend
     [fifocache: <fifo_cache_config>]
 
-  # Most recent allowed cacheable result, to prevent caching very recent results
```
gouthamve, line 744: Should we outright remove it from docs or should we mark it DEPRECATED, but keep it in docs?

pracucci: I would keep it simple and keep in the doc only the current version of the flags (the ones expected to be used).

PR #2610, opened 2020-05-19
Use Ticker for WAL series spreading

csmarchbanks:
Right now series are spread out across the WAL checkpoint duration using
a time.After() call, however that does not account for the amount of
time it may take to checkpoint the series in question. I am consistently
seeing a checkpoint take longer than the goal, and this will help
account for that difference.

**Checklist**
- [N/A] Tests updated
- [N/A] Documentation added
  * Is this needed? Possibly ENHANCEMENT but I don't know if it needs mentioning. WDYT?

### DIFF
--- a/pkg/ingester/wal.go
+++ b/pkg/ingester/wal.go
@@ -333,6 +333,8 @@ func (w *walWrapper) performCheckpoint(immediate bool) (err error) {
 	}
 	records := [][]byte{}
 	totalSize := 0
+	ticker := time.NewTicker(perSeriesDuration)
+	defer ticker.Stop()
 	for userID, state := range us {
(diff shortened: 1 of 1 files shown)


csmarchbanks:
@codesome Is the goal to take exactly 100% of the period? In which case I could change the perSeriesDuration logic. I was thinking it would still be nice to take a little bit under the full duration just in case.

codesome:
> Is the goal to take exactly 100% of the period

The goal was to spread the checkpoint writes mainly and should not matter if it is slightly above or below the specified duration I believe. But having it under the duration seems fine over it crossing the duration. LGTM to me.

PR #2611, opened 2020-05-20
support for comparison of response in querytee

sandeepsukhani:
**What this PR does**:
Adds support for comparison of response in query-tee. It as of now only does a comparison of `matrix`, `vector`, `scalar` kind of responses. It also allows defining route based response comparison since different routes could have a different kind of response.

The code is also moved to a separate package from cmd so that it can be reused in other projects like Loki.

**Checklist**
- [x] Tests updated


tools/querytee/response_comparator.go
```diff
@@ -0,0 +1,208 @@
...
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/pkg/labels"
+
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+// ResponsesComparatorFunc helps with comparing different responses from various routes.
+type ResponsesComparatorFunc func(expected, actual []byte) error
```
pstibrany, line 15: No need to add `Func` to the name.

tools/querytee/response_comparator.go
```diff
@@ -0,0 +1,208 @@
...
+
+	return comparator(expected.Data.Result, actual.Data.Result)
+}
+
+type matrixResult struct {
+	Result []struct {
+		Metric map[string]string
+		Values [][]interface{}
+	}
+}
```
pstibrany, line 73: Would it make sense to reuse existing data types from `prometheus/common/model` package here?

tools/querytee/response_comparator.go
```diff
@@ -0,0 +1,208 @@
...
+
+type SamplesResponse struct {
+	Status string
+	Data   struct {
+		ResultType string
+		Result     json.RawMessage
+	}
+}
+
+var samplesComparator = map[string]SamplesComparatorFunc{
```
pstibrany, line 28: Can we avoid using package-scoped global variables, and introduce a type for this map? Both `RegisterSamplesComparator` and `CompareSamplesResponse` will be methods on the type.

tools/querytee/response_comparator.go
pstibrany: This type should be defined in `proxy_endpoint.go` where it is actually used.

I would (weakly) prefer to use interface for this, it's typically easier to find implementations for interface than for func type.

tools/querytee/response_comparator.go
```diff
@@ -0,0 +1,208 @@
...
+	if len(actual) != 2 {
+		return fmt.Errorf("expected 2 values in samplePair but got %d", len(actual))
+	}
+
+	if expected[0] != actual[0] {
+		return fmt.Errorf("expected timestamp %d but got %d", int64(expected[0].(float64)), int64(actual[0].(float64)))
+	}
+	if expected[1] != actual[1] {
+		return fmt.Errorf("expected value %s for timestamp %d but got %s", expected[1].(string), int64(expected[0].(float64)), actual[1].(string))
+	}
```
pstibrany, line 205: There is lot of type casting going on here, but if server replies with something unexpected, these casts will panic and crash. I'd suggest using checked type casts here.

tools/querytee/proxy_endpoint.go
```diff
@@ -108,6 +132,27 @@ func (p *ProxyEndpoint) pickResponseForDownstream(responses []*backendResponse)
 	return responses[0]
 }
 
+func (p *ProxyEndpoint) compareResponses(expectedResponse, actualResponse *backendResponse) error {
+	if p.comparator == nil {
+		return nil
+	}
+
```
pstibrany, line 139: This is only called if `p.comparator` is not nil.

tools/querytee/response_comparator.go
```diff
@@ -0,0 +1,208 @@
...
+
+		actualMetric := actual.Result[actualMetricIndex]
+		expectedMetricLen := len(expectedMetric.Values)
+		actualMetricLen := len(actualMetric.Values)
+
+		if expectedMetricLen != actualMetricLen {
+			err := fmt.Errorf("expected %d samples for metric %s but got %d", expectedMetricLen,
+				expectedMetric.Metric, actualMetricLen)
+			if expectedMetricLen > 0 && actualMetricLen > 0 {
+				level.Error(util.Logger).Log("msg", err.Error(), "oldest-expected-ts", int64(expectedMetric.Values[0][0].(float64)),
```
pstibrany, line 111: Can we introduce a function to do interface{} -> float64 -> int64 conversion, and also do proper type-checks? Repeating conversion all over the place makes code harder to read.

sandeepsukhani:
Thanks @pstibrany for the review! The code looks neat now with all the improvements that you have suggested.

Issue #2603, opened 2020-05-19
runtime error: integer divide by zero


pstibrany:
Found in logs while investigating customer query error:

```
level=error ts=2020-05-19T07:16:14.487217809Z caller=engine.go:810 msg="runtime panic in parser" err="runtime error: integer divide by zero" stacktrace="goroutine 1700 [running]:
github.com/prometheus/prometheus/promql.(*evaluator).recover(0xc001e26800, 0xc001e26590)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200213233353-b90be6f32a33/promql/engine.go:808 +0xd8
panic(0x2e7e120, 0x5612dd0)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/prometheus/prometheus/promql.(*evaluator).eval(0xc001e26800, 0x3b66900, 0xc001cd9e90, 0x0, 0xc001e250f8)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200213233353-b90be6f32a33/promql/engine.go:1330 +0x35d3
github.com/prometheus/prometheus/promql.(*evaluator).evalSubquery(0xc001e26800, 0xc001cd9e90, 0x345b36b)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200213233353-b90be6f32a33/promql/engine.go:1005 +0x59
github.com/prometheus/prometheus/promql.(*evaluator).eval(0xc001e26800, 0x3b667c0, 0xc001cd9ec0, 0xc001e761b0, 0x0)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200213233353-b90be6f32a33/promql/engine.go:1074 +0x222c
github.com/prometheus/prometheus/promql.(*evaluator).rangeEval(0xc001e26800, 0xc001e26158, 0xc001e26220, 0x2, 0x2, 0x0, 0x564a1a0, 0x0)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200213233353-b90be6f32a33/promql/engine.go:892 +0x1408
github.com/prometheus/prometheus/promql.(*evaluator).eval(0xc001e26800, 0x3b66780, 0xc001e0a500, 0x3b65b80, 0xc001e76120)
[...]

PR #2614, opened 2020-05-20
Ensures promql's default evaluation interval is set for the query frontend

owen-d:
Ensure the default evaluation interval is set (promql uses a package-scoped mutable variable). This is important when `querier.parallelise-shardable-queries` is enabled because the frontend aggregates the sharded queries.

Closes https://github.com/cortexproject/cortex/issues/2603


owen-d:
Thanks @pracucci!

PR #2617, opened 2020-05-21
Ignore time range in /series request for the blocks storage

### LINKED ISSUE #2612: Unexpected behaviour if using nested variables in grafana templates with block storage
We are using the experimental block storage and have an issue with nested variables in Grafana.

If using template variables in Grafana like `label_values(node_load1{hardwarenode='$hardwarenode'}, instance)` and try to query data which is not present in the ingesters anymore, there are no labels found.

If templates are using no nested variables like `label_values(node_load1, instance)`, there is no problem, even querying old data as long as you use `Refresh: 'On Dashboard Load'` in Grafana.

There are some screenshots attached to show the behaviour.

Querying last 7 days with nested variables is working:
Bildschirmfoto 2020-05-20 um 13 02 24

Querying a timerange not present anymore in ingesters there are no labels found for this timerange:
Bildschirmfoto 2020-05-20 um 13 02 08

It looks like the labels are tried to be retrieved from ingesters for the selected timerange. Since labels are only queried from ingesters, there are no labels found for this timerange and therefore empty. I am not sure if this is expected behaviour or an issue.

pracucci:
**What this PR does**:
The `/series` request is currently served only fetching series from the ingesters (both chunks and blocks storage). This means that when the query has a time range older then the ingester's data time range, the `/series` always return an empty result set if the request time range is honored.

In the chunks storage we picked the path to ignore the request time range right now (until efficiently querying the long-term storage for the `/series` will be a thing) and in this PR I'm doing the same for the blocks storage as well, keeping the two storage engines aligned in this regard.

Fixes #2612

**Checklist**
- [x] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -99,6 +99,7 @@
 * [BUGFIX] Experimental TSDB: fixed memory leak in ingesters. #2586
 * [BUGFIX] Ingester: Fix an ingester starting up in the JOINING state and staying there forever. #2565
 * [BUGFIX] QueryFrontend: fixed a panic (`integer divide by zero`) in the query-frontend. The query-frontend now requires the `-querier.default-evaluation-interval` config to be set to the same value of the querier. #2603
+* [BUGFIX] Experimental TSDB: when the querier receives a `/series` request with a time range older than the data stored in the ingester, it now returns ignores the request time range and returns known series anyway instead of returning an empty response. This aligns the behaviour with the chunks storage. #2617
```
pstibrany, line 102:
```suggestion
* [BUGFIX] Experimental TSDB: when the querier receives a `/series` request with a time range older than the data stored in the ingester, it now ignores the requested time range and returns known series anyway instead of returning an empty response. This aligns the behaviour with the chunks storage. #2617
```

pkg/ingester/ingester_v2.go
```diff
@@ -527,13 +527,16 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx context.Context, req *client.Me
...
 	// Parse the request
-	from, to, matchersSet, err := client.FromMetricsForLabelMatchersRequest(req)
+	_, _, matchersSet, err := client.FromMetricsForLabelMatchersRequest(req)
 	if err != nil {
 		return nil, err
 	}
 
-	// Create a new instance of the TSDB querier
-	q, err := db.Querier(int64(from), int64(to))
+	// Since we ingester runs with a very limited TSDB retention, we can (and should) query
```
pstibrany, line 535:
```suggestion
	// Since ingester runs with a very limited TSDB retention, we can (and should) query
```

pracucci:
Thanks @pstibrany for fixing my typos!

PR #2620, opened 2020-05-21
Fix remote read route registration

### LINKED ISSUE #2619: Remote Reads are Broken
Prometheus remote-read requests require a POST method with an encoded `prompb.ReadRequest`. See [here for more context](https://github.com/prometheus/prometheus/blob/1c99adb9fdc79446534e55edba35c676d5e89d57/storage/remote/client.go#L149)

Cortex currently only registers the read endpoint for `GET` requests

jtlisi:
**What this PR does**:

- Ensure the `POST` method is set for the`/read` route registration to match the endpoint in Prometheus. 

Fixes #2619

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -291,7 +291,7 @@ func (a *API) RegisterQuerier(queryable storage.Queryable, engine *promql.Engine
 	cacheGenHeaderMiddleware := getHTTPCacheGenNumberHeaderSetterMiddleware(tombstonesLoader)
 	promHandler := fakeRemoteAddr(cacheGenHeaderMiddleware.Wrap(promRouter))
 
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/read", querier.RemoteReadHandler(queryable), true, "GET")
+	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/read", querier.RemoteReadHandler(queryable), true, "POST")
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -290,7 +290,7 @@ func (a *API) RegisterQuerier(queryable storage.Queryable, engine *promql.Engine
 	cacheGenHeaderMiddleware := getHTTPCacheGenNumberHeaderSetterMiddleware(tombstonesLoader)
 	promHandler := fakeRemoteAddr(cacheGenHeaderMiddleware.Wrap(promRouter))
 
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/read", querier.RemoteReadHandler(queryable), true, "GET")
+	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/read", querier.RemoteReadHandler(queryable), true, "POST")
```
slim-bean, line 293: is there any harm in supporting both `GET` and `POST`?

jtlisi: `GET` will always return an error since the handler will always try to decode an empty response body. If we only specify `POST`, `GET` will return a method no valid which is what we would want in this case.

pracucci: How has this ever worked?

jtlisi: We used to not do any form of method validation.

pracucci:
@jtlisi Could you rebase `master`, please?

PR #2626, opened 2020-05-25
Metadata caching

pstibrany:
**What this PR does**: This PR adds support for caching metadata (meta files, iteration results) to blocks storage, using caching bucket from Thanos. Caching bucket is configured with Cortex usage of blocks in mind (caching top level directory with users, list of blocks per user, content of meta.json and deletion mark files, list of chunks directory).

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -2807,7 +2807,7 @@ bucket_store:
     # CLI flag: -experimental.tsdb.bucket-store.chunks-cache.max-get-range-requests
     [max_get_range_requests: <int> | default = 3]
 
-    # TTL for caching object size for chunks.
+    # TTL for caching object attributes for chunks.
     # CLI flag: -experimental.tsdb.bucket-store.chunks-cache.attributes-ttl
     [attributes_ttl: <duration> | default = 24h]
 
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -2815,6 +2815,82 @@ bucket_store:
     # CLI flag: -experimental.tsdb.bucket-store.chunks-cache.subrange-ttl
     [subrange_ttl: <duration> | default = 24h]
 
+  metadata_cache:
+    cachebackend:
+      # Backend for metadata cache, if not empty. Supported values: memcached.
+      # CLI flag: -experimental.tsdb.bucket-store.metadata-cache.backend
+      [backend: <string> | default = ""]
+
+      memcached:
+        # Comma separated list of memcached addresses. Supported prefixes are:
+        # dns+ (looked up as an A/AAAA query), dnssrv+ (looked up as a SRV
+        # query, dnssrvnoa+ (looked up as a SRV query, with no A/AAAA lookup
+        # made after that).
+        # CLI flag: -experimental.tsdb.bucket-store.metadata-cache.memcached.addresses
+        [addresses: <string> | default = ""]
+
+        # The socket read/write timeout.
+        # CLI flag: -experimental.tsdb.bucket-store.metadata-cache.memcached.timeout
+        [timeout: <duration> | default = 100ms]
+
+        # The maximum number of idle connections that will be maintained per
+        # address.
+        # CLI flag: -experimental.tsdb.bucket-store.metadata-cache.memcached.max-idle-connections
+        [max_idle_connections: <int> | default = 16]
+
+        # The maximum number of concurrent asynchronous operations can occur.
+        # CLI flag: -experimental.tsdb.bucket-store.metadata-cache.memcached.max-async-concurrency
+        [max_async_concurrency: <int> | default = 50]
+
+        # The maximum number of enqueued asynchronous operations allowed.
+        # CLI flag: -experimental.tsdb.bucket-store.metadata-cache.memcached.max-async-buffer-size
+        [max_async_buffer_size: <int> | default = 10000]
+
+        # The maximum number of concurrent connections running get operations.
+        # If set to 0, concurrency is unlimited.
+        # CLI flag: -experimental.tsdb.bucket-store.metadata-cache.memcached.max-get-multi-concurrency
+        [max_get_multi_concurrency: <int> | default = 100]
+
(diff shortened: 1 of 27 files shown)


pkg/storage/tsdb/caching_bucket.go
```diff
@@ -36,47 +62,133 @@ func (cfg *ChunksCacheConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix st
...
+type MetadataCacheConfig struct {
+	CacheBackend
+
+	TenantsListTTL         time.Duration `json:"tenants_list_ttl"`
+	TenantBlocksListTTL    time.Duration `json:"tenant_blocks_list_ttl"`
+	ChunksListTTL          time.Duration `json:"chunks_list_ttl"`
+	MetafileExistsTTL      time.Duration `json:"metafile_exists_ttl"`
+	MetafileDoesntExistTTL time.Duration `json:"metafile_doesnt_exist_ttl"`
+	MetafileContentTTL     time.Duration `json:"metafile_content_ttl"`
+	MetafileMaxSize        int           `json:"metafile_max_size"`
```
pracucci, line 78: yaml, not json 😉

pstibrany: Lol, I need to setup my IDE for adding `yaml`, instead of `json` :) Thanks for catching this.

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -36,47 +62,133 @@ func (cfg *ChunksCacheConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix st
...
+type MetadataCacheConfig struct {
+	CacheBackend `yaml:",inline"`
+
+	TenantsListTTL         time.Duration `yaml:"tenants_list_ttl"`
+	TenantBlocksListTTL    time.Duration `yaml:"tenant_blocks_list_ttl"`
+	ChunksListTTL          time.Duration `yaml:"chunks_list_ttl"`
+	MetafileExistsTTL      time.Duration `yaml:"metafile_exists_ttl"`
+	MetafileDoesntExistTTL time.Duration `yaml:"metafile_doesnt_exist_ttl"`
+	MetafileContentTTL     time.Duration `yaml:"metafile_content_ttl"`
+	MetafileMaxSize        int           `yaml:"metafile_max_size"`
```
pracucci, line 78: What do you think about adding `_bytes` suffix to clarity (and CLI too)?

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -36,47 +62,133 @@ func (cfg *ChunksCacheConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix st
...
+	cfg := storecache.NewCachingBucketConfig()
+	cachingConfigured := false
+
+	chunksCache, err := createCache("chunks-cache", chunksConfig.Backend, chunksConfig.Memcached, logger, reg)
+	if err != nil {
+		return nil, errors.Wrapf(err, "chunks-cache")
+	}
+	if chunksCache != nil {
+		cachingConfigured = true
+		cfg.CacheGetRange("chunks", chunksCache, isTSDBChunkFile, chunksConfig.SubrangeSize, chunksConfig.ObjectSizeTTL, chunksConfig.SubrangeTTL, chunksConfig.MaxGetRangeRequests)
```
pracucci, line 109: We should rename `ObjectSizeTTL` to `AttributesTTL` (or something like this) and rename the YAML/CLI config too.

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -36,47 +62,133 @@ func (cfg *ChunksCacheConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix st
...
+		cachingConfigured = true
 
-	return nil
-}
+		cfg.CacheExists("metafile", metadataCache, isMetaFile, metadataConfig.MetafileExistsTTL, metadataConfig.MetafileDoesntExistTTL)
+		cfg.CacheGet("metafile", metadataCache, isMetaFile, metadataConfig.MetafileMaxSize, metadataConfig.MetafileContentTTL, metadataConfig.MetafileExistsTTL, metadataConfig.MetafileDoesntExistTTL)
 
-func CreateCachingBucket(chunksConfig ChunksCacheConfig, bkt objstore.Bucket, logger log.Logger, reg prometheus.Registerer) (objstore.Bucket, error) {
-	var chunksCache cache.Cache
+		cfg.CacheIter("tenants-iter", metadataCache, isTenantsDir, metadataConfig.TenantsListTTL, storecache.JSONIterCodec{})
```
pracucci, line 122: Why not compress this too with snappy? If you have 10K tenants, we could save some bytes. Given it's snappy, if you only have 1 tenant there's no bytes saving but also the impact on performances would be not noticeable.

pstibrany: We can do that. While I don't expect compression ratio to be great for root directory, as you note, performance impact is minimal.

docs/configuration/config-file-reference.md
```diff
@@ -2807,14 +2807,89 @@ bucket_store:
     # CLI flag: -experimental.tsdb.bucket-store.chunks-cache.max-get-range-requests
     [max_get_range_requests: <int> | default = 3]
 
-    # TTL for caching object size for chunks.
-    # CLI flag: -experimental.tsdb.bucket-store.chunks-cache.object-size-ttl
-    [object_size_ttl: <duration> | default = 24h]
+    # TTL for caching object attributes for chunks.
+    # CLI flag: -experimental.tsdb.bucket-store.chunks-cache.attributes-ttl
+    [attributes_ttl: <duration> | default = 24h]
```
codesome, line 2812: I see this (flag and config change) did not make into the CHANGELOG, is it intentional?

pstibrany: No, it's an omission and changelog entry would be nice. On the other hand, old attribute name was not yet `released` (it's only on master), so probably not worth it?

PR #2629, opened 2020-05-25
Added Cortex 1.0 and 1.1 to integration test.

pstibrany:
**What this PR does**: This PR adds Cortex 1.0 and 1.1 to the integration test .


pracucci:
@pstibrany Pay attention that integration tests fail due to some changes in CLI flags.

pstibrany:
I've also increased timeout for running integration tests in this PR, as we have many tests and they are becoming quite slow. On my machine:

```
$ time go test -tags=requires_docker ./integration/ 
ok  	github.com/cortexproject/cortex/integration	514.022s

real	8m40.335s
user	1m11.831s
sys	0m25.904s
```

pracucci:
> I've also increased timeout for running integration tests in this PR, as we have many tests and they are becoming quite slow

That's fine (increasing the timeout). About slow integration tests, I think there are two things we could work on:
1. Tests parallelisation (doable with some changes)
2. Migrate integration tests CI to GitHub actions (even if we parallelise, we have limited resources in CircleCI, while we could have more in GitHub actions thanks to CNCF)

pstibrany:
Thanks for review!

PR #2630, opened 2020-05-25
Remove upload grace period

pracucci:
**What this PR does**:
In the PR #2593 I've introduced a basic version of the consistency check for the read path, introducing the "upload grace period" based on the "block creation time" instead of the "block upload completed time" because we didn't have a way to know when a block was shipped to the storage.

In this PR I'm fixing it.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/blocks_consistency_checker.go
+++ b/pkg/querier/blocks_consistency_checker.go
@@ -77,7 +77,7 @@ func (c *BlocksConsistencyChecker) Check(expectedBlocks []*BlockMeta, knownDelet
 			deletionTime := time.Unix(mark.DeletionTime, 0)
 
 			if c.deletionGracePeriod > 0 && time.Since(deletionTime) > c.deletionGracePeriod {
-				level.Debug(c.logger).Log("msg", "block skipped from consistency check because marked for deletion", "block", meta.ULID.String(), "deletionTime", deletionTime.String())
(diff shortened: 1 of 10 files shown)


pkg/querier/blocks_consistency_checker.go
```diff
@@ -52,29 +56,33 @@ func (c *BlocksConsistencyChecker) Check(expectedBlocks []ulid.ULID, knownDeleti
...
 		// on blocks that can't be queried because they were offloaded. For this reason, we don't run the consistency check on any block
 		// which has been marked for deletion more then "grace period" time ago. Basically, the grace period is the time
 		// we still expect a block marked for deletion to be still queried.
-		if mark := knownDeletionMarks[blockID]; mark != nil {
-			if time.Since(time.Unix(mark.DeletionTime, 0)) > c.deletionGracePeriod {
+		if mark := knownDeletionMarks[meta.ULID]; mark != nil {
+			deletionTime := time.Unix(mark.DeletionTime, 0)
+
+			if c.deletionGracePeriod > 0 && time.Since(deletionTime) > c.deletionGracePeriod {
+				level.Debug(c.logger).Log("msg", "block skipped from consistency check because marked for deletion", "block", meta.ULID.String(), "deletionTime", deletionTime.String())
```
pstibrany, line 80:
```suggestion
				level.Debug(c.logger).Log("msg", "block skipped from consistency check because it is marked for deletion", "block", meta.ULID.String(), "deletionTime", deletionTime.String())
```

PR #2633, opened 2020-05-26
Micro optimization in newBlockQuerierSeries()

pracucci:
**What this PR does**:
Since #2599 we don't need to use labels builder anymore. An unsafe conversion should just be fine.

```
BEFORE: Benchmark_newBlockQuerierSeries-4   	  921447	      1135 ns/op
AFTER:  Benchmark_newBlockQuerierSeries-4   	 9260770	       108 ns/op
```

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/block.go
+++ b/pkg/querier/block.go
@@ -204,12 +204,7 @@ func newBlockQuerierSeries(lbls []storepb.Label, chunks []storepb.AggrChunk) *bl
 		return chunks[i].MinTime < chunks[j].MinTime
 	})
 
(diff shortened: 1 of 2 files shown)


pracucci:
> LGTM (failing test can be removed, as it's not real-case scenario, as discussed at https://cloud-native.slack.com/archives/CL25937SP/p1590491672334400)

I double checked and looks safe. I've also opened a 1-line PR in Thanos to document it https://github.com/thanos-io/thanos/pull/2662

PR #2635, opened 2020-05-26
Removed auth from internal http server

joe-elliott:
**What this PR does**:
Removes auth middleware from the querier when operating in single binary mode.  In single binary mode the querier does not register any externally accessible endpoints.  All endpoints are registered to an internal router that is then used by the worker to push requests it retrieves from the query-frontend.  

Removing auth will improve performance (minor) and also allow for custom auth to work in single binary mode.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/api/api.go
```diff
@@ -285,48 +285,53 @@ func (a *API) RegisterQuerier(queryable storage.Queryable, engine *promql.Engine
...
 	}
 
-	// Since we have a new router and the request will not go trough the default server
+	// Since we have a new router and the request will not go through the default server
 	// HTTP middleware stack, we need to add a middleware to extract the trace context
 	// from the HTTP headers and inject it into the Go context.
-	return nethttp.MiddlewareFunc(opentracing.GlobalTracer(), router.ServeHTTP, nethttp.OperationNameFunc(func(r *http.Request) string {
+	// Additionally add use the weaveworks AuthenticateUser middleware to copy the OrgID header
+	// into the request context.
+	return middleware.AuthenticateUser.Wrap(nethttp.MiddlewareFunc(opentracing.GlobalTracer(), router.ServeHTTP, nethttp.OperationNameFunc(func(r *http.Request) string {
```
pracucci, line 332: I don't quite understand this whole change. We have introduced `needsAuth`, set it to `false` and then here we do wrap `middleware.AuthenticateUser` anyway (which is the default auth middleware, used in any case except when auth is disabled).

joe-elliott: Basically we want to override the auth using `cfg.HTTPAuthMiddleware`, but not have it apply to the internal querier.  However, we still need the "normal" auth to still apply to translate `X-Scope-OrgID` => org id in the go context.  The querier (and almost every Cortex component) relies on this.

This change is confusing.  Marking as draft while we discuss internally.

PR #2636, opened 2020-05-26
Add extra details to query spans.

pstibrany:
**What this PR does**: ~This PR tries to add a span for iterating series set used in blocks storage.~ (Update: didn't work) Plus it adds little extra details to existing spans.


pkg/querier/blocks_store_queryable.go
```diff
@@ -392,7 +395,7 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
 		}
 	}
 
-	return storage.NewMergeSeriesSet(seriesSets, storage.ChainedSeriesMerge), warnings, nil
+	return newTracingSet(q.ctx, "blocksStoreQuerier.seriesSet", storage.NewMergeSeriesSet(seriesSets, storage.ChainedSeriesMerge)), warnings, nil
```
pracucci, line 398: We should pass `spanCtx` instead of `q.ctx` here.

pkg/querier/block.go
```diff
@@ -108,9 +112,9 @@ func (b *blocksQuerier) Select(_ bool, sp *storage.SelectHints, matchers ...*lab
 		return nil, nil, promql.ErrStorage{Err: err}
 	}
 
-	return &blockQuerierSeriesSet{
-		series: series,
-	}, warnings, nil
+	level.Debug(log).Log("series", len(series), "warnings", len(warnings))
+
+	return newTracingSet(b.ctx, "blocksQuerier.seriesSet", &blockQuerierSeriesSet{series: series}), warnings, nil
```
pracucci, line 117: We should pass `ctx` instead of `b.ctx` here.

_Unrelated to this PR: renaming `ctx` to `spanCtx` could help making it more clear._

pkg/querier/tracing_set.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"context"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb/chunkenc"
+)
+
+func newTracingSet(ctx context.Context, op string, set storage.SeriesSet) storage.SeriesSet {
```
pracucci, line 12: Renaming "tracing set" into "tracing series set" could be more clear.

pkg/querier/tracing_set.go
```diff
@@ -0,0 +1,95 @@
...
+		t.span.Finish()
+		// Nil to avoid double closing.
+		t.span = nil
+	}
+	return false
+}
+
+func (t *tracingSet) At() storage.Series {
+	s := t.s.At()
+	if t.span != nil && t.series%100 == 0 {
```
pracucci, line 45: A bit scared about this % 100. 100K series (which is still reasonable) will end up into 1K logs. A %1000 could be enough.

pkg/querier/blocks_store_queryable.go
pstibrany: I think `q.ctx` is correct, because `spanCtx` will finish when function returns. WDYT?

pracucci: Right, makes sense.

pkg/querier/block.go
pracucci: > We should pass ctx instead of b.ctx here.

Ignore this comment too, for the same reason.

PR #2637, opened 2020-05-26
Add more tracing spans to read path

pracucci:
**What this PR does**:
Add a couple of more tracing spans to better understand where the time is being spent while executing a query.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/distributor_queryable.go
+++ b/pkg/querier/distributor_queryable.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/prom1/storage/metric"
 	"github.com/cortexproject/cortex/pkg/querier/series"
 	"github.com/cortexproject/cortex/pkg/util/chunkcompat"
+	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 )
 
 // Distributor is the read interface to the distributor, made an interface here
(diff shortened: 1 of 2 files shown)


pkg/querier/distributor_queryable.go
```diff
@@ -52,10 +53,13 @@ type distributorQuerier struct {
 // Select implements storage.Querier interface.
 // The bool passed is ignored because the series is always sorted.
 func (q *distributorQuerier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
+	log, ctx := spanlogger.New(q.ctx, "distributorQuerier.Select")
```
pstibrany, line 56: This is almost the same as `Distributor.Query` (just saying, we can have both).

pkg/querier/querier.go
```diff
@@ -245,6 +245,7 @@ type querier struct {
 func (q querier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
 	log, ctx := spanlogger.New(q.ctx, "querier.Select")
 	defer log.Span.Finish()
+	level.Debug(log).Log("start", sp.Start, "end", sp.End, "step", sp.Step)
```
pstibrany, line 248: This will panic if `sp` is nil (as checked on next line).

pstibrany: It would also be nice to format those times.

PR #2639, opened 2020-05-27
cassandra-cql-fixup

storyicon:
**What this PR does**:
Fix query statement in cassandra storage


pracucci:
@storyicon Could you add a CHANGELOG entry like this one, please?

```
* [BUGFIX] Cassandra: fixed an edge case leading to an invalid CQL query when querying the index on a Cassandra store. #2639
```

storyicon:
@pracucci Hello, I followed your advice. At the same time, I have some other ideas about cassandra storage, I will open an issue to explain it later 😄

PR #2645, opened 2020-05-28
improve query speed when does not enable caching

### LINKED ISSUE #2644: cachingIndexClient brings negative optimization
Hello, the maintainer of Cortex, when I tested Cortex, I found:
```
query_range?query=node_cpu{instance_name="prometheus-23"}&start=1590662430&end=1590663330&step=15
```
The above query took me up to 12.56s. I printed the log of the CQL statement executed by Querier and found that the query statement of Cortex Querier was like this:
```
SELECT range, value FROM index_18409 WHERE hash = 'fake:d18409:node_cpu:instance_name'
```
It took me 12.111s and returned 1,610,998 records.

I think the query here should like this:
```
SELECT range, value FROM index_18409 WHERE hash = 'fake:d18409:node_cpu:instance_name' and value = textasblob('prometheus-23')
```
It takes less time(I execute it in cassandra and it only takes 4s to return. After I add an index to the value column, it only takes 3ms.), uses less resources(only 3 records returned), and generates less calculations in Cortex.

I found the following code:
https://github.com/cortexproject/cortex/blob/8cfa79c9265492fbd78a3d89d29e526d35e3db9b/pkg/chunk/storage/caching_index_client.go#L118-L121

The conversion here makes the `RangeValuePrefix`, `RangeValueStart` and `ValueEqual` parsed by the parser lost, I understand that this is done to get the "cacheable result set", but it has a negative impact.

I have three suggestions for improvement:
1. When caching is not enabled, `RangeValuePrefix`, `RangeValueStart` and `ValueEqual` should not be removed by `cacheIndexClient.QueryPages`.
2. table Manager should add indexes to the `value` column of cassandra's index table.
3. When the cache is turned on, two requests can be issued, one request contains the `value` and `range` conditions, used to quickly return the query results to the user, and one request is used to obtain the "cacheable result set" and cache it.

Do you have any other ideas? 👋

storyicon:
Signed-off-by: storyicon <storyicon@foxmail.com>

**What this PR does**:
When the cache is not enabled, more query conditions are used to reduce query costs and increase query speed

Fixes #2644

### DIFF
--- a/pkg/chunk/cache/tiered.go
+++ b/pkg/chunk/cache/tiered.go
@@ -15,10 +15,8 @@ func NewTiered(caches []Cache) Cache {
 
 // IsEmptyTieredCache is used to determine whether the current Cache is implemented by an empty tiered.
 func IsEmptyTieredCache(cache Cache) bool {
-	if c, ok := cache.(tiered); ok && len(c) == 0 {
-		return true
-	}
-	return false
+	c, ok := cache.(tiered)
+	return ok && len(c) == 0
 }
 
 func (t tiered) Store(ctx context.Context, keys []string, bufs [][]byte) {
--- a/pkg/chunk/cache/tiered.go
+++ b/pkg/chunk/cache/tiered.go
@@ -13,6 +13,14 @@ func NewTiered(caches []Cache) Cache {
 	return tiered(caches)
 }
 
+// IsEmptyTieredCache is used to determine whether the current Cache is implemented by an empty tiered.
+func IsEmptyTieredCache(cache Cache) bool {
+	if c, ok := cache.(tiered); ok && len(c) == 0 {
+		return true
+	}
+	return false
+}
+
 func (t tiered) Store(ctx context.Context, keys []string, bufs [][]byte) {
 	for _, c := range []Cache(t) {
 		c.Store(ctx, keys, bufs)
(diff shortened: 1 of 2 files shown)


pkg/chunk/storage/caching_index_client.go
```diff
@@ -115,10 +115,18 @@ func (s *cachingIndexClient) QueryPages(ctx context.Context, queries []chunk.Ind
...
 		// Only need to consider one of the queries as they have the same table & hash.
 		queries := queriesByKey[key]
-		cacheableMissed = append(cacheableMissed, chunk.IndexQuery{
+		query := chunk.IndexQuery{
 			TableName: queries[0].TableName,
 			HashValue: queries[0].HashValue,
-		})
+		}
+
+		if cache.IsEmptyTieredCache(s.cache) {
```
gouthamve, line 123: Can we move this check to `newCachingIndexClient`?

Basically do:
```
func newCachingIndexClient(client chunk.IndexClient, c cache.Cache, validity time.Duration, limits StoreLimits) chunk.IndexClient {
	if c == nil || cache.IsEmptyTieredCache(c) {
		return client
	}
        
	return &cachingIndexClient{
		IndexClient: client,
		cache:       cache.NewSnappy(c),
		validity:    validity,
		limits:      limits,
	}
}
```

pkg/chunk/cache/tiered.go
```diff
@@ -13,6 +13,14 @@ func NewTiered(caches []Cache) Cache {
 	return tiered(caches)
 }
 
+// IsEmptyTieredCache is used to determine whether the current Cache is implemented by an empty tiered.
+func IsEmptyTieredCache(cache Cache) bool {
+	if c, ok := cache.(tiered); ok && len(c) == 0 {
+		return true
+	}
+	return false
```
codesome, line 21: Nit
```suggestion
	c, ok := cache.(tiered)
	return ok && len(c) == 0
```

storyicon:
Hello, I took your comments and spent some time reorganizing my submissions, thanks!

PR #2648, opened 2020-05-28
Generic KV Client Instrumentation

### LINKED ISSUE #2484: Need a way to differentiate between KV metrics for different rings and dudupers in single binary mode.
Current its all in one histogram - `cortex_kv_request_duration_seconds_count`.  Need to add a name label or something.

jtlisi:
**What this PR does**:

- Add a name and backend type label to the prometheus metrics that wrap KV clients
- Remove dedicated instrumentation for the consul client in favor of a generic instrumentation for all clients

Fixes #2484 

**Checklist**
- [x] Tests updated


pkg/ring/kv/client.go
```diff
@@ -97,19 +99,19 @@ type Client interface {
 
 // NewClient creates a new Client (consul, etcd or inmemory) based on the config,
 // encodes and decodes data for storage using the codec.
-func NewClient(cfg Config, codec codec.Codec) (Client, error) {
+func NewClient(name string, cfg Config, codec codec.Codec, reg prometheus.Registerer) (Client, error) {
```
pstibrany, line 102: I think passing `name` to components that is used just for metrics registration is an anti-pattern. Caller of this call can easily wrap `reg` into `prometheus.WrapRegistererWith(prometheus.Labels{"name": "some name here"}, reg)` if it wishes to do so. I would prefer to use this pattern instead. KV store client doesn't need a name for anything else, and it also doesn't need it if registrerer is `nil`.

pkg/ring/kv/client.go
```diff
@@ -149,10 +151,10 @@ func createClient(name string, prefix string, cfg StoreConfig, codec codec.Codec
 		client = PrefixClient(client, prefix)
 	}
 
-	return metrics{client}, nil
+	return newMetricsClient(name, backend, client, reg), nil
```
pstibrany, line 154: If `reg` is nil (which is a case we should support and is typical for tests), we can skip creation of metrics client completely.

pkg/ring/lifecycler.go
```diff
@@ -147,7 +147,8 @@ func NewLifecycler(cfg LifecyclerConfig, flushTransferer FlushTransferer, ringNa
 	}
 	port := GetInstancePort(cfg.Port, cfg.ListenPort)
 	codec := GetCodec()
-	store, err := kv.NewClient(cfg.RingConfig.KVStore, codec)
+	// Suffix all client names with "-lifecycler" to denote this kv client is used by the lifecycler
+	store, err := kv.NewClient(ringName+"-lifecycler", cfg.RingConfig.KVStore, codec, nil)
```
pstibrany, line 151: This is a great opportunity to pass registrerer to lifecycler and move lifecycler metrics away from global variables.

pkg/distributor/ha_tracker_test.go
```diff
@@ -135,7 +136,7 @@ func TestWatchPrefixAssignment(t *testing.T) {
 		UpdateTimeout:          time.Millisecond,
 		UpdateTimeoutJitterMax: 0,
 		FailoverTimeout:        time.Millisecond * 2,
-	})
+	}, prometheus.NewRegistry())
```
pstibrany, line 139: `nil`, unless metrics are checked.

pkg/ring/kv/consul/client.go
```diff
@@ -85,7 +84,7 @@ func NewClient(cfg Config, codec codec.Codec) (*Client, error) {
 		return nil, err
 	}
 	c := &Client{
-		kv:    consulMetrics{client.KV()},
```
pstibrany, line 88: Previously we could see individual `CAS` calls to consul. Now we will only see "CAS loop".

pracucci:
> Better way of handling multiple components with same metrics is to wrap registrerer to add constant label to all metrics registered by component

Agree. It's more powerful.

pkg/distributor/distributor.go
```diff
@@ -184,7 +184,7 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 	replicationFactor.Set(float64(ingestersRing.ReplicationFactor()))
 	cfg.PoolConfig.RemoteTimeout = cfg.RemoteTimeout
 
-	replicas, err := newClusterTracker(cfg.HATrackerConfig)
+	replicas, err := newClusterTracker(cfg.HATrackerConfig, nil)
```
pracucci, line 187: Why don't we pass the registerer here?

pkg/distributor/ha_tracker.go
```diff
@@ -157,7 +157,7 @@ func newClusterTracker(cfg HATrackerConfig) (*haTracker, error) {
 	}
 
 	if cfg.EnableHATracker {
-		client, err := kv.NewClient(cfg.KVStore, GetReplicaDescCodec())
+		client, err := kv.NewClient(RingNameForClient, cfg.KVStore, GetReplicaDescCodec(), reg)
```
pracucci, line 160: I think using the same name here would clash with the distributor ring name. We could just pass `distributor-hatracker` as ring client name. What do you think? /cc @pstibrany

pkg/ring/kv/metrics.go
```diff
@@ -32,12 +22,36 @@ func errorCode(err error) string {
...
-	c Client
+	c               Client
+	requestDuration *instrument.HistogramCollector
+}
+
+func newMetricsClient(name string, backend string, c Client, reg prometheus.Registerer) Client {
+	// TODO: Ensure all clients are passed a valid registry and remove the following check
+	if reg == nil {
+		reg = prometheus.DefaultRegisterer
+	}
```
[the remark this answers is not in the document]
pracucci: Agree. This TODO should be removed (and the default register never used).

pkg/querier/blocks_store_queryable.go
```diff
@@ -139,7 +139,7 @@ func NewBlocksStoreQueryableFromConfig(querierCfg Config, gatewayCfg storegatewa
 
 	if gatewayCfg.ShardingEnabled {
 		storesRingCfg := gatewayCfg.ShardingRing.ToRingConfig()
-		storesRingBackend, err := kv.NewClient(storesRingCfg.KVStore, ring.GetCodec())
+		storesRingBackend, err := kv.NewClient(storegateway.RingNameForClient+"-block-store", storesRingCfg.KVStore, ring.GetCodec(), reg)
```
pracucci, line 142: The generated name here looks a bit weird to me. Maybe `querier-store-gateway` may be more clear?

pkg/ring/kv/consul/client.go
```diff
@@ -99,21 +98,17 @@ func (c *Client) Put(ctx context.Context, key string, value interface{}) error {
 		return err
 	}
 
-	return instrument.CollectedRequest(ctx, "Put", consulRequestDuration, instrument.ErrorCode, func(ctx context.Context) error {
```
pracucci, line 102: Now that I notice, I think previous version was even wrong. We did track "Put" twice for 1 Put call. Once here, and once in `consul/metrics.go`. Am I missing anything?

pkg/ring/kv/metrics.go
```diff
@@ -32,12 +22,36 @@ func errorCode(err error) string {
...
+		requestDuration: instrument.NewHistogramCollector(
+			promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+				Namespace: "cortex",
+				Name:      "kv_request_duration_seconds",
+				Help:      "Time spent on kv store requests.",
+				Buckets:   prometheus.DefBuckets,
+				ConstLabels: prometheus.Labels{
+					"name": name,
+					"type": backend,
+				},
```
pracucci, line 46: Please mention this change in the CHANGELOG. You should also check if it has an impact on the cortex mixin's dashboards (depends on how we group series, I haven't checked).

pkg/ring/lifecycler.go
pracucci: But maybe in another PR?

pkg/ring/kv/consul/metrics.go
```diff
@@ -1,82 +0,0 @@
...
-
-import (
-	"context"
-
-	consul "github.com/hashicorp/consul/api"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/weaveworks/common/instrument"
-)
-
-var consulRequestDuration = instrument.NewHistogramCollector(prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
pracucci, line 11: Please remember to mention in the CHANGELOG that the metric `cortex_consul_request_duration_seconds` has been removed.

pkg/ring/kv/client.go
jtlisi: 👍 Make sense

pkg/ring/lifecycler.go
jtlisi: I will pass the Registerer but keep the lifecycler metrics global for now.

pkg/distributor/ha_tracker_test.go
jtlisi: 👍

pkg/ring/kv/client.go
jtlisi: Makes sense

pkg/distributor/ha_tracker.go
jtlisi: Makes sense

pkg/ring/kv/consul/client.go
jtlisi: After investigating further, the consul metrics probably should remain as is. They provide lower level metrics for the client interacting with consul itself: https://github.com/hashicorp/consul/blob/master/api/kv.go

In our KV implementation CAS is retried multiple times and by using a generic metrics client we are missing out on instrumenting some of these calls. One option that may be used to solve this problem is moving retries to there own KV implementation and wrap clients with it.

jtlisi: I am going to re-add the separate consul instrumentation and will add an issue to add a generic retry implementation for the KV store interface. WDYT?

pkg/querier/blocks_store_queryable.go
jtlisi: Sounds good, I wasn't sure how best to name some of these clients.

pkg/ring/kv/consul/metrics.go
jtlisi: I re-added this metric since it fits a different use case.

pkg/distributor/distributor.go
jtlisi: The distributor didn't pass a registerer. I added one.

jtlisi:
@pstibrany & @pracucci 

I updated this PR per your comments. Changes include:

- Keeping the `cortex_consul_request_duration_seconds` metric and consul instrumentation w/ note about a possible way of removing it in a future PR
- Registerers are now plumbed through to each call of the kv.NewClient
- Updated ChangeLog

I will test this PR w/ the cortex-mixin this week and ensure it operates correctly with the new metrics.

If you have anymore feedback about the naming of KV clients let me know. I wasn't sure if some of the names/suffixes I chose made sense.

pkg/distributor/distributor_test.go
```diff
@@ -369,7 +369,7 @@ func TestDistributor_PushHAInstances(t *testing.T) {
 						KVStore:         kv.Config{Mock: mock},
 						UpdateTimeout:   100 * time.Millisecond,
 						FailoverTimeout: time.Second,
-					})
+					}, prometheus.NewRegistry())
```
pstibrany, line 372:
```suggestion
					}, nil)
```

Since we don't verify metrics.

pkg/ring/kv/metrics.go
```diff
@@ -32,12 +37,35 @@ func errorCode(err error) string {
...
-	c Client
+	c               Client
+	requestDuration *instrument.HistogramCollector
+}
+
+func newMetricsClient(backend string, c Client, reg prometheus.Registerer) Client {
+	// If no Registerer is provided return the raw client
+	if reg == nil {
+		return c
+	}
```
pstibrany, line 48: Nit: I'd move this check into `createClient` function.

pkg/ring/kv/metrics.go
```diff
@@ -5,19 +5,24 @@ import (
...
-func init() {
-	requestDuration.Register()
+// RegistererWithKVName wraps the provided Registerer with the KV name label. If a nil reg
+// is provided, a nil registry is returned
+func RegistererWithKVName(reg prometheus.Registerer, name string) prometheus.Registerer {
+	if reg == nil {
+		return nil
+	}
+
+	return prometheus.WrapRegistererWith(prometheus.Labels{"name": name}, reg)
```
pstibrany, line 25: Is it worth using something more unique, e.g. `kv-name` here?

pkg/ring/kv/metrics.go
```diff
@@ -5,19 +5,24 @@ import (
...
 
-var requestDuration = instrument.NewHistogramCollector(prometheus.NewHistogramVec(prometheus.HistogramOpts{
-	Namespace: "cortex",
-	Name:      "kv_request_duration_seconds",
-	Help:      "Time spent on consul requests.",
-	Buckets:   prometheus.DefBuckets,
-}, []string{"operation", "status_code"}))
+var (
+	primaryLabel   = prometheus.Labels{"role": "primary"}
+	secondaryLabel = prometheus.Labels{"role": "secondary"}
```
pstibrany, line 15: Only used from client.go, I would move it into that file.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Metric `cortex_kv_request_duration_seconds` now includes `name` label to denote which client is being used as well as the `backend` label to denote the KV backend implemnetation in use. #2648
```
pstibrany, line 5:
```suggestion
* [CHANGE] Metric `cortex_kv_request_duration_seconds` now includes `name` label to denote which client is being used as well as the `backend` label to denote the KV backend implementation in use. #2648
```

pkg/distributor/distributor_ring.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 )
 
+var (
+	RingNameForClient = "distributor"
+)
+
```
pstibrany, line 19:
```suggestion
const (
	ringNameForClient = "distributor"
)
```

And move to `distributor_ring.go` where it is used.

(Alternatively, just inline it)

pkg/ruler/ruler.go
```diff
@@ -190,7 +190,11 @@ func NewRuler(cfg Config, engine *promql.Engine, queryable promStorage.Queryable
 	}
 
 	if cfg.EnableSharding {
-		ringStore, err := kv.NewClient(cfg.Ring.KVStore, ring.GetCodec())
+		ringStore, err := kv.NewClient(
+			cfg.Ring.KVStore,
+			ring.GetCodec(),
+			kv.RegistererWithKVName(reg, ring.RulerRingKey),
```
pstibrany, line 196:
```suggestion
			kv.RegistererWithKVName(reg, "ruler"),
```

`ring.RulerRingKey` is actually `ring`.

pkg/storegateway/gateway.go
```diff
@@ -81,7 +81,11 @@ func NewStoreGateway(gatewayCfg Config, storageCfg cortex_tsdb.Config, logLevel
 	}
 
 	if gatewayCfg.ShardingEnabled {
-		ringStore, err = kv.NewClient(gatewayCfg.ShardingRing.KVStore, ring.GetCodec())
+		ringStore, err = kv.NewClient(
+			gatewayCfg.ShardingRing.KVStore,
+			ring.GetCodec(),
+			kv.RegistererWithKVName(reg, RingNameForClient),
```
pstibrany, line 87:
```suggestion
			kv.RegistererWithKVName(reg, "store-gateway"),
```

pkg/ring/kv/metrics.go
jtlisi: That makes sense. I was just basing it off the label we use for the ring.

PR #2651, opened 2020-05-29
Refcache

pstibrany:
**What this PR does**: this PR does some changes to `RefCache` to improve its throughput in new `RefCacheConcurrency` benchmark. Unfortunately it also makes individual `Ref` and `SetRef` calls slower. Sending for discussion, not necessarily merging.

```
name                           old time/op  new time/op  delta
RefCacheConcurrency_1m_50-4     1.29s ± 4%   1.10s ± 5%  -14.84%  (p=0.000 n=9+10)
RefCacheConcurrency_1m_250-4    2.35s ± 3%   1.85s ± 6%  -21.46%  (p=0.000 n=10+10)
RefCacheConcurrency_1m_1000-4   10.6s ± 2%    7.8s ± 3%  -27.00%  (p=0.000 n=10+10)
RefCache_SetRef-4               414µs ± 2%   651µs ± 2%  +57.43%  (p=0.000 n=10+10)
RefCache_Ref-4                  815µs ± 2%   894µs ± 3%   +9.69%  (p=0.000 n=10+10)
RefCache_purge-4                102ms ± 2%    54ms ± 2%  -47.13%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/storage/tsdb/ref_cache.go
+++ b/pkg/storage/tsdb/ref_cache.go
@@ -27,20 +27,20 @@ const (
 type RefCache struct {
 	// The cache is split into stripes, each one with a dedicated lock, in
 	// order to reduce lock contention.
-	stripes [numRefCacheStripes]*refCacheStripe
+	stripes [numRefCacheStripes]refCacheStripe
 }
 
 // refCacheStripe holds a subset of the series references for a single tenant.
 type refCacheStripe struct {
 	refsMu sync.Mutex
-	refs   map[model.Fingerprint][]*refCacheEntry
+	refs   map[model.Fingerprint][]refCacheEntry
 }
 
 // refCacheEntry holds a single series reference.
 type refCacheEntry struct {
 	lbs       labels.Labels
 	ref       uint64
-	touchedAt time.Time
+	touchedAt int64 // Unix nano time.
 }
 
 // NewRefCache makes a new RefCache.
@@ -49,9 +49,7 @@ func NewRefCache() *RefCache {
 
 	// Stripes are pre-allocated so that we only read on them and no lock is required.
 	for i := 0; i < numRefCacheStripes; i++ {
-		c.stripes[i] = &refCacheStripe{
-			refs: map[model.Fingerprint][]*refCacheEntry{},
-		}
+		c.stripes[i].refs = map[model.Fingerprint][]refCacheEntry{}
 	}
 
 	return c
@@ -61,7 +59,7 @@ func NewRefCache() *RefCache {
 // is NOT retained.
 func (c *RefCache) Ref(now time.Time, series labels.Labels) (uint64, bool) {
 	fp := client.Fingerprint(series)
-	stripeID := uint8(util.HashFP(fp) % numRefCacheStripes)
+	stripeID := util.HashFP(fp) % numRefCacheStripes
 
 	return c.stripes[stripeID].ref(now, series, fp)
 }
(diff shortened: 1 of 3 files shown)


pkg/storage/tsdb/ref_cache.go
```diff
@@ -131,7 +129,7 @@ func (s *refCacheStripe) purge(keepUntil time.Time) {
 		// Since we do expect very few fingerprint collisions, we
 		// have an optimized implementation for the common case.
 		if len(entries) == 1 {
-			if entries[0].touchedAt.Before(keepUntil) {
+			if entries[0].touchedAt < keepUntil.UnixNano() {
```
pracucci, line 132: There's a little bit of math involved each time we call `.UnixNano()`. Since we do it once for each entry and we know we have 1M+ entries per ingester, we could store the result of `.UnixNano()` at the beginning of the function.

pkg/storage/tsdb/ref_cache.go
```diff
@@ -27,31 +28,29 @@ const (
 type RefCache struct {
 	// The cache is split into stripes, each one with a dedicated lock, in
 	// order to reduce lock contention.
-	stripes [numRefCacheStripes]*refCacheStripe
```
pracucci, line 30: Does the pointer change bring any benefit in terms of performances?

pracucci: Talked offline. Answer is yes.

pstibrany:
> * Would you mind measuring mem allocations diff too, please?

Benchmark with allocations:

```
$ benchstat before_mem.txt after_mem.txt 
name                           old time/op    new time/op    delta
RefCacheConcurrency_1m_50-4       1.28s ± 9%     1.05s ±13%   -17.55%  (p=0.000 n=10+9)
RefCacheConcurrency_1m_250-4      2.24s ± 7%     1.67s ± 2%   -25.52%  (p=0.000 n=9+10)
RefCacheConcurrency_1m_1000-4     10.7s ±15%      7.2s ± 2%   -32.15%  (p=0.000 n=9+10)
RefCache_SetRef-4                 393µs ± 3%     657µs ±12%   +67.16%  (p=0.000 n=10+10)
RefCache_Ref-4                    783µs ± 2%     826µs ± 4%    +5.42%  (p=0.000 n=10+10)
RefCache_purge-4                 95.7ms ± 3%    49.9ms ± 2%   -47.87%  (p=0.000 n=9+10)

name                           old alloc/op   new alloc/op   delta
RefCacheConcurrency_1m_50-4       235MB ± 0%     214MB ± 0%    -8.85%  (p=0.000 n=10+8)
RefCacheConcurrency_1m_250-4      235MB ± 0%     214MB ± 0%    -8.87%  (p=0.000 n=10+10)
RefCacheConcurrency_1m_1000-4     235MB ± 0%     214MB ± 0%    -8.87%  (p=0.000 n=10+10)
RefCache_SetRef-4                 0.00B          0.00B           ~     (all equal)
RefCache_Ref-4                    0.00B          0.00B           ~     (all equal)
RefCache_purge-4                   768B ± 0%     1706B ± 0%  +122.14%  (p=0.000 n=10+8)

name                           old allocs/op  new allocs/op  delta
RefCacheConcurrency_1m_50-4       2.03M ± 0%     1.02M ± 0%   -49.63%  (p=0.000 n=10+8)
RefCacheConcurrency_1m_250-4      2.03M ± 0%     1.02M ± 0%   -49.64%  (p=0.000 n=10+10)
RefCacheConcurrency_1m_1000-4     2.03M ± 0%     1.02M ± 0%   -49.63%  (p=0.000 n=10+10)
RefCache_SetRef-4                  0.00           0.00           ~     (all equal)
RefCache_Ref-4                     0.00           0.00           ~     (all equal)
RefCache_purge-4                   21.0 ± 0%      21.0 ± 0%      ~     (all equal)
```

> * Why is the single get/set slower? What makes it slower?

`SetRef` is slower because locking unlocked `RWMutex` is bit slower than `Mutex`. `Ref` is slower because of `RWMutex` and atomic update of timestamp.

pracucci:
> SetRef is slower because locking unlocked RWMutex is bit slower than Mutex. Ref is slower because of RWMutex and atomic update of timestamp.

That's fine. Supporting higher concurrency here is better.

PR #2655, opened 2020-05-29
Update thanos and renable tracing

pracucci:
**What this PR does**:
In Thanos we removed tracing spans from MemcachedClient (was flooding us) and thus we can now re-enable the Thanos tracing instrumentation we removed in #2605.

Moreover, the good @pstibrany added bucket client and cache instrumentation support into Thanos and here we're enabling it.

Tested locally:

Screen Shot 2020-05-29 at 12 51 36

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci:
> There is one small bug visible in the screenshot, related to "BucketStores.Series" and children spans -- they are not children spans, but should be. Fixing that would be possible eg. by wrapping storepb.Store_SeriesServer in (*BucketStores).Series and using context from new span in the wrapper. We can live with the current state though.

Good point. I noticed it as well some time ago. I will address it in a separate PR.

PR #2656, opened 2020-05-29
Remove -store.fullsize-chunks option

### LINKED ISSUE #1163: -store.fullsize-chunks=false breaks ingester hand-overs
The feature implemented in #1044 results in handed-over chunks lacking the "footer" to add more values.

A simple way to work round the problem would be to reset the global flag before beginning a hand-over.

Maybe cleaner to add a "closed" flag to each chunk so it knows whether further writing is expected?

bboreham:
It broke ingester hand-overs, and is not needed now we use BigChunk.

Fixes #1163

**Checklist**
- [x] Tests updated
- [x] Documentation added N/A
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 * [CHANGE] Experimental Memberlist ring: randomize gossip node names to avoid conflicts when running multiple clients on the same host, or reusing host names (eg. pods in statefulset). Node name randomization can be disabled by using `-memberlist.randomize-node-name=false`. #2715
 * [CHANGE] Memberlist KV client is no longer considered experimental. #2725
 * [CHANGE] Change target flag for purger from `data-purger` to `purger` and make delete request cancellation duration configurable. #2760
+* [CHANGE] Removed `-store.fullsize-chunks` option which was undocumented and unused (it broke ingester hand-overs). #2656
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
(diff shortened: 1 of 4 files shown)


pkg/chunk/encoding/factory.go
```diff
@@ -15,15 +15,13 @@ type Config struct{}
 
 var (
 	// DefaultEncoding exported for use in unit tests elsewhere
-	DefaultEncoding             = Bigchunk
-	alwaysMarshalFullsizeChunks = true
```
pracucci, line 19: Why are you saying it was unused? Being the default `true` doesn't it mean it's always used unless explicitly disabled via CLI flag?

bboreham: Yes, it was written this way to avoid changing things for existing users.
I mean the option to disable was never used.

bboreham: However I got the change wrong.

pracucci: Yes, then we need to remove the flag while always keeping it enabled ;)

bboreham: Try now.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * [CHANGE] Query Frontend now uses Round Robin to choose a tenant queue to service next. #2553
 * [CHANGE] `-promql.lookback-delta` is now deprecated and has been replaced by `-querier.lookback-delta` along with `lookback_delta` entry under `querier` in the config file. `-promql.lookback-delta` will be removed in v1.4.0. #2604
+* [CHANGE] Removed `-store.fullsize-chunks` option which was undocumented and unused (it broke ingester hand-overs).
```
pracucci, line 7: Would you mind adding the PR number at the end of the line? The release script we use look for PRs missing in the CHANGELOG (for manual check) and does the check based on the PR number here.

PR #2668, opened 2020-06-01
do not invalidate chunks cache when series are deleted

sandeepsukhani:
**What this PR does**:
PR #2279 introduced invalidation of cache when the series deletion is requested. It invalidates results cache when delete request is received and all store caches(chunks, index and write-dedupe) when delete request is processed. It seems we need not invalidate chunks cache since we refer chunks by their ID and changing the content of the chunk requires changing its ID as well.

Also, a chunk is accessible only if it has an index entry for it which also means even though deleted chunks stay in cache until they are evicted, there should not be any harm since its index is already deleted and we would avoid clearing the whole cache containing non-deleted chunks as well.

Thanks to Bryan for pointing this out [here](https://github.com/cortexproject/cortex/pull/2279#issuecomment-636485265).

### DIFF
--- a/pkg/chunk/storage/factory.go
+++ b/pkg/chunk/storage/factory.go
@@ -142,7 +142,6 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 
 	// lets wrap all caches with CacheGenMiddleware to facilitate cache invalidation using cache generation numbers
 	indexReadCache = cache.NewCacheGenNumMiddleware(indexReadCache)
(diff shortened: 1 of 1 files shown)


pkg/chunk/storage/factory.go
```diff
@@ -142,7 +142,6 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 
 	// lets wrap all caches with CacheGenMiddleware to facilitate cache invalidation using cache generation numbers
```
gouthamve, line 143: Can we add in the comment why we're excluding chunk cache?

PR #2673, opened 2020-06-01
Support Accept: application/json on common HTTP endpoints

joe-elliott:
**What this PR does**:
- Adds support for the `Accept: application/json` header to the following endpoints:
```
  /distributor/all_user_stats
  /distributor/ha_tracker
  /ingester/ring
  /store-gateway/ring
  /compactor/ring
  /ruler/ring
  /services
```
- Renders the `/services` endpoint in the standard Cortex html style

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
Parsing `Accept` correctly is non-trivial:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

and there is some discussion about adding it to the standard library as well as alternatives here:
https://github.com/golang/go/issues/19307

For our purposes I think a simple `string.Contains` is sufficient since there is no standard implementation, but I would be willing to research and use an existing implementation of Accept parsing.

PR #2674, opened 2020-06-02
Dedicated struct and methods for metric counter

codesome:
**What this PR does**:
This refactoring will help in adding series limits to block based ingester. It also fixes a bug where the [series per metric is not incremented when recovering from WAL](https://github.com/cortexproject/cortex/blob/85f93ff4abd7af81a5eeeae0e07379dd6ff2071d/pkg/ingester/user_state.go#L225-L231).

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/user_state.go
+++ b/pkg/ingester/user_state.go
@@ -42,7 +42,7 @@ type userState struct {
 	ingestedAPISamples  *ewmaRate
 	ingestedRuleSamples *ewmaRate
 
-	seriesInMetric []metricCounterShard
+	seriesInMetric *metricCounter
 
 	// Series metrics.
 	memSeries             prometheus.Gauge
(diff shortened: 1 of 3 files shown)


pkg/ingester/user_state.go
```diff
@@ -353,3 +319,65 @@ outer:
...
+	shard := m.getShard(metric)
+	shard.mtx.Lock()
+	defer shard.mtx.Unlock()
+
+	err := m.limiter.AssertMaxSeriesPerMetric(userID, shard.m[metric])
+	if err != nil {
+		return err
+	}
+
+	shard.m[metric]++
```
pracucci, line 374: Why do we increment it here, if then we also call `seriesAddedFor()`? This means we're double tracking it. Don't we have any test covering the in-memory series limit? If not, we should add it.

codesome: This actually should not be here, a copy paste error

codesome:
I have updated the tests to catch this bug.

`TestIngesterUserLimitExceeded` passes even on master as the bug is not for per user limit. 

`TestIngesterMetricLimitExceeded` fails on master as expected as per metric counter is not incremented during recover.

codesome:
One thing to note about this change: the limit is now a soft guarantee for the per metric limit compared to a strong guarantee before as the check and the increment does not happen inside same method. Is that fine?

pracucci:
> One thing to note about this change: the limit is now a soft guarantee for the per metric limit compared to a strong guarantee before as the check and the increment does not happen inside same method. Is that fine?

Definitely yes. We don't need "pixel accuracy". If we over limit by few units, it's perfectly fine.

PR #2676, opened 2020-06-02
Series limit for TSDB blocks based ingester

codesome:
**What this PR does**:

Adds series limit for TSDB blocks based ingester. Requires https://github.com/cortexproject/cortex/pull/2674

**TODO**
- [x] Unit tests
- [x] Handle these error where `app.Add` and `app.AddFast` is called - consider them to be soft errors
- [x] What to do when there is no metric name?
- [x] Update `CHANGELOG.md`


codesome:
This PR is now ready for review. I was able to re-use the existing tests to add case for the ingester v2. I need help in deciding the desired action for metric name limit when there is no metric name in a series.

pkg/ingester/ingester_v2.go
```diff
@@ -58,6 +62,52 @@ type userTSDB struct {
...
+
+	// Total series limit.
+	if err := u.limiter.AssertMaxSeriesPerUser(u.userID, int(u.DB.Head().NumSeries())); err != nil {
+		return makeLimitError(perUserSeriesLimit, err)
+	}
+
+	// Series per metric name limit.
+	metricName, err := extract.MetricNameFromLabels(metric)
+	if err != nil {
+		// TODO(codesome): what should be the action?
```
pstibrany, line 79: This should never happen. Given that, I suggest we just return the error.

pkg/ingester/ingester_v2.go
```diff
@@ -363,6 +413,15 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 				continue
 			}
 
+			if ve, ok := cause.(*validationError); ok {
```
pstibrany, line 416:
```suggestion
			var ve *validationError
			if errors.As(cause, &ve) {
```

This also works if errors are wrapped.

pkg/ingester/ingester_test.go
```diff
@@ -503,190 +505,239 @@ func TestIngesterAppendBlankLabel(t *testing.T) {
...
-	testLimits := func() {
-		// Append to two series, expect series-exceeded error.
-		_, err = ing.Push(ctx, client.ToWriteRequest([]labels.Labels{labels1, labels3}, []client.Sample{sample2, sample3}, nil, client.API))
-		if resp, ok := httpgrpc.HTTPResponseFromError(err); !ok || resp.Code != http.StatusTooManyRequests {
-			t.Fatalf("expected error about exceeding metrics per user, got %v", err)
+		return ing
+	}
+
+	for _, ingGenerator := range []func() *Ingester{chunksIngesterGenerator, blocksIngesterGenerator} {
+		ing := ingGenerator()
```
pstibrany, line 547: Run in subtest?

pkg/ingester/ingester_v2.go
```diff
@@ -58,6 +62,52 @@ type userTSDB struct {
...
+	}
+
+	return nil
+}
+
+// PostCreation implements SeriesLifecycleCallback interface.
+func (u *userTSDB) PostCreation(metric labels.Labels) {
+	metricName, err := extract.MetricNameFromLabels(metric)
+	if err != nil {
+		// TODO(codesome): what should be the action?
```
pracucci: I would just replace this TODO with a comment saying:
`This should never happen because it has already been checked in PreCreation().`

pkg/ingester/ingester_v2.go
```diff
@@ -58,6 +62,52 @@ type userTSDB struct {
...
+	}
+	u.seriesInMetric.seriesAddedFor(metricName)
+}
+
+// PostDeletion implements SeriesLifecycleCallback interface.
+func (u *userTSDB) PostDeletion(metrics ...labels.Labels) {
+	for _, metric := range metrics {
+		metricName, err := extract.MetricNameFromLabels(metric)
+		if err != nil {
+			// TODO(codesome): what should be the action?
```
pracucci: I would just replace this TODO with a comment saying:
`This should never happen because metrics have already been checked in PreCreation().`

pkg/ingester/ingester_v2.go
```diff
@@ -397,7 +456,11 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 	}
 
 	if firstPartialErr != nil {
-		return &client.WriteResponse{}, httpgrpc.Errorf(http.StatusBadRequest, wrapWithUser(firstPartialErr, userID).Error())
+		code := http.StatusBadRequest
+		if ve, ok := errors.Cause(firstPartialErr).(*validationError); ok {
```
pracucci, line 456: I would use `errors.As()` here as well instead of `.(*validationError)` for the same reason suggested by peter.

pkg/util/extract/extract.go
```diff
@@ -51,3 +51,12 @@ func MetricNameMatcherFromMatchers(matchers []*labels.Matcher) (*labels.Matcher,
 	// Return all matchers if none are metric name matchers
 	return nil, matchers, false
 }
+
+// MetricNameFromLabels extracts the metric name from a list of Prometheus Labels.
+func MetricNameFromLabels(lbls labels.Labels) (metricName string, err error) {
+	metricName = lbls.Get(model.MetricNameLabel)
+	if metricName == "" {
+		err = fmt.Errorf("No metric name label")
```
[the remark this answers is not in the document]
codesome: Race :) I just pushed it

pkg/ingester/ingester_test.go
```diff
@@ -503,190 +505,245 @@ func TestIngesterAppendBlankLabel(t *testing.T) {
...
-	sample3 := client.Sample{
-		TimestampMs: 1,
-		Value:       3,
+
+	blocksIngesterGenerator := func() *Ingester {
+		ing, err := newIngesterMockWithTSDBStorageAndLimits(defaultIngesterTestConfig(), limits, blocksDir, nil)
+		require.NoError(t, err)
+		require.NoError(t, services.StartAndAwaitRunning(context.Background(), ing))
+		// Wait until it's ACTIVE
+		test.Poll(t, 10*time.Millisecond, ring.ACTIVE, func() interface{} {
```
pracucci, line 660: Let's be more generous to avoid flaky tests.

```suggestion
		test.Poll(t, time.Second, ring.ACTIVE, func() interface{} {
```

pkg/ingester/ingester_test.go
```diff
@@ -503,190 +505,245 @@ func TestIngesterAppendBlankLabel(t *testing.T) {
...
-			t.Fatalf("expected error about exceeding metrics per user, got %v", err)
-		}
-		// Append two metadata, expect no error since metadata is a best effort approach.
-		_, err = ing.Push(ctx, client.ToWriteRequest(nil, nil, []*client.MetricMetadata{metadata1, metadata2}, client.API))
+	blocksIngesterGenerator := func() *Ingester {
+		ing, err := newIngesterMockWithTSDBStorageAndLimits(defaultIngesterTestConfig(), limits, blocksDir, nil)
 		require.NoError(t, err)
+		require.NoError(t, services.StartAndAwaitRunning(context.Background(), ing))
+		// Wait until it's ACTIVE
+		test.Poll(t, 10*time.Millisecond, ring.ACTIVE, func() interface{} {
```
pracucci, line 539: Let's be more generous to avoid flaky tests.

```suggestion
		test.Poll(t, time.Second, ring.ACTIVE, func() interface{} {
```

pracucci:
> Also, if I remember correctly the changes in TSDB, callback doesn't guarantee atomicity. If this is correct, we should improve a bit removeMetricName() and handle the case shard.m[metricName] does not exist?

Discussed offline. Not an issue.

PR #2679, opened 2020-06-02
Use vendor deps when building.

### LINKED ISSUE #1710: Cortex is not using vendored dependencies in CI

tomwilkie:

**What this PR does**:

Use `-mod=vendor` when building, to prevent us downloading all the modules and make sure we build with the vendored files.

Fixes #1710

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ RM := --rm
 # as it currently disallows TTY devices. This value needs to be overridden
 # in any custom cloudbuild.yaml files
 TTY := --tty
-GO_FLAGS := -ldflags "-X main.Branch=$(GIT_BRANCH) -X main.Revision=$(GIT_REVISION) -X main.Version=$(VERSION) -extldflags \"-static\" -s -w" -tags netgo
+GO_FLAGS := -ldflags "-X main.Branch=$(GIT_BRANCH) -X main.Revision=$(GIT_REVISION) -X main.Version=$(VERSION) -extldflags \"-static\" -s -w" -tags netgo -mod vendor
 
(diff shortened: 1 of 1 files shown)


tomwilkie:
(don't merge, trying to figure out why this works and Bryan's earlier PR doesnt)

pracucci:
> (don't merge, trying to figure out why this works and Bryan's earlier PR doesnt)

I don't. I think we also need to add it to `go test` in `.circleci/config.yml` (for integration tests) and `tools/test` (for unit tests).

pstibrany:
This will be default once we fully migrate to 1.14 (incl. change in go.mod). That was the reason we didn't bothered last time.

tomwilkie:
I can't figure out why Bryan's PR didn't work, but I suspect its as I'm only changing the build, not the tests.  Either way I think we should move forward with this now.

PR #2684, opened 2020-06-03
Various Cassandra client tweaks

tomwilkie:
- Add option to not convict host of being down when a request fails.
- Split the object client from the index client.
- Use gocql fork with metrics and logs.


tomwilkie:
Looks like they're not interested in merging the gocql changes: https://github.com/gocql/gocql/pull/1453

So we need to decide if we will maintain this fork forever or find another way of doing this.

pkg/chunk/cassandra/storage_client.go
```diff
@@ -98,13 +103,19 @@ func (cfg *Config) session() (*gocql.Session, error) {
...
+	cluster.Registerer = prometheus.WrapRegistererWith(
+		prometheus.Labels{"client": name}, prometheus.DefaultRegisterer)
 	if cfg.Retries > 0 {
 		cluster.RetryPolicy = &gocql.ExponentialBackoffRetryPolicy{
 			NumRetries: cfg.Retries,
 			Min:        cfg.MinBackoff,
 			Max:        cfg.MaxBackoff,
 		}
 	}
+	if cfg.ConvictHosts {
```
joe-elliott, line 116: Should this be `!cfg.ConvictHosts`.  i.e. if the user does not want to convict hosts then we replace the `cluster.ConvictionPolicy` with noop.

This would also maintain current behavior with default values.

pkg/chunk/cassandra/storage_client.go
```diff
@@ -435,7 +484,26 @@ func (s *StorageClient) getChunk(ctx context.Context, decodeContext *chunk.Decod
...
+	s.readSession.Close()
+	s.writeSession.Close()
+}
+
+type noopConvictionPolicy struct{}
+
+// AddFailure should return `true` if the host should be convicted, `false` otherwise.
+// Convicted means connections are removed - we don't want that.
+// Implementats gocql.ConvictionPolicy.
+func (noopConvictionPolicy) AddFailure(err error, host *gocql.HostInfo) bool {
```
joe-elliott, line 503: Should we add a metric here?

Also, what happens if a cassandra node _really_ goes down?  I think we would want gocql to remove it from the pool.  I feel like failure rate would be the best way to determine this, but unsure if that is possible to calculate.

This will also prevent Cortex from ever backing off a struggling Cassandra.  Have we implemented a way for Cassandra to exert backpressure on Cortex?

go.mod
```diff
@@ -77,3 +77,5 @@ replace k8s.io/client-go => k8s.io/client-go v0.18.3
 
 // >v1.2.0 has some conflict with prometheus/alertmanager. Hence prevent the upgrade till it's fixed.
 replace github.com/satori/go.uuid => github.com/satori/go.uuid v1.2.0
+
+replace github.com/gocql/gocql => github.com/grafana/gocql v0.0.0-20200604163347-a8642710dbb1
```
pracucci, line 81: Could you add a comment to explain why? Will be easier to remember the rationale over the time.

pkg/chunk/cassandra/storage_client.go
```diff
@@ -44,6 +47,7 @@ type Config struct {
 	MinBackoff               time.Duration       `yaml:"retry_min_backoff"`
 	QueryConcurrency         int                 `yaml:"query_concurrency"`
 	NumConnections           int                 `yaml:"num_connections"`
+	ConvictHosts             bool                `yaml:"convict_hosts"`
```
pracucci, line 50: For clarity, I would call the YAML option/CLI flag `convict_hosts_on_failure`/`convict-hosts-on-failure`

pkg/chunk/cassandra/storage_client.go
```diff
@@ -70,6 +74,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.MaxBackoff, "cassandra.retry-max-backoff", 10*time.Second, "Maximum time to wait before retrying a failed request. (Default = 10s)")
 	f.IntVar(&cfg.QueryConcurrency, "cassandra.query-concurrency", 0, "Limit number of concurrent queries to Cassandra. (Default is 0: no limit)")
 	f.IntVar(&cfg.NumConnections, "cassandra.num-connections", 2, "Number of TCP connections per host.")
+	f.BoolVar(&cfg.ConvictHosts, "cassandra.convict-hosts", true, "Convict hosts of being down on failure. (Default is true)")
```
pracucci, line 77: No need to specify "(Default is true)" in the flag description because it's automatically built based on the actual default value (both in the generated doc and `--help` output).

pkg/chunk/cassandra/table_client.go
```diff
@@ -17,7 +17,7 @@ type tableClient struct {
 
 // NewTableClient returns a new TableClient.
 func NewTableClient(ctx context.Context, cfg Config) (chunk.TableClient, error) {
-	session, err := cfg.session()
+	session, err := cfg.session("table_manager")
```
pracucci, line 20: Just for consistency. Looks other client names are all dash separated.

```suggestion
	session, err := cfg.session("table-manager")
```

pkg/chunk/cassandra/storage_client.go
tomwilkie: If the host is really failed, connections won't succeed and connection pool will be empty - and the host will be skipped by the query executor.  So I don't think this is a real problem.

As for backing off from a struggling Cassandra, we have timeouts and retries which will pick another host.  Honestly trying to do "clever" load balancing almost always leads to problems like the ones this PR is trying to fix, so I prefer not to do it.

PR #2688, opened 2020-06-04
Upgraded Thanos to master

pracucci:
**What this PR does**:
In this PR I've upgraded Thanos to master as first step to introduce changes to selectively query requested blocks via the store-gateway.

**Which issue(s) this PR fixes**:
N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2689, opened 2020-06-04
Selectively query blocks from each store-gateway

pracucci:
**What this PR does**:
What we learned running Cortex blocks storage at a medium scale is that store-gateway blocks sharding may lead to some inefficiencies at query time, having the same block unintentionally queried from multiple store-gateway (replicas) at the same time.

To improve this, we introduced in Thanos the ability to selectively pick which blocks should be queried (via block matchers) and in this PR I'm leveraging on this to selectively query blocks from each store-gateway instance.

~~Draft PR until~~:
- ~~https://github.com/cortexproject/cortex/pull/2688 gets merged and rebased here~~
- ~~Tests are done in our dev clusters~~

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/blocks_store_replicated_set.go
+++ b/pkg/querier/blocks_store_replicated_set.go
@@ -70,8 +70,8 @@ func (s *blocksStoreReplicationSet) stopping(_ error) error {
 	return services.StopManagerAndAwaitStopped(context.Background(), s.subservices)
 }
 
-func (s *blocksStoreReplicationSet) GetClientsFor(blockIDs []ulid.ULID) ([]BlocksStoreClient, error) {
-	var sets []ring.ReplicationSet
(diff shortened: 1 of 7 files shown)


pkg/querier/blocks_store_replicated_set.go
```diff
@@ -85,66 +85,27 @@ func (s *blocksStoreReplicationSet) GetClientsFor(blockIDs []ulid.ULID) ([]Block
...
 
-		sets = append(sets, set)
+		// Pick the first store-gateway instance.
+		addr := set.Ingesters[0].Addr
+
+		if _, ok := shards[addr]; ok {
+			shards[addr] = append(shards[addr], blockID)
+		} else {
+			shards[addr] = []ulid.ULID{blockID}
+		}
```
pstibrany, line 95:
```suggestion
		shards[addr] = append(shards[addr], blockID)
```

PR #2695, opened 2020-06-05
Refetch series for missing blocks

pracucci:
**What this PR does**:
_Take a deep breath..._

This should be the last missing piece in my original design when I've started working on introducing the store-gateway (doc excluded, will be done in a separate PR).

Due to the consistency check, a query may fail because some expected blocks are missing from the specific selected store-gateway. This is not a edge condition and could actually occur relatively frequently due to the resharding (ie. during a rolling update or scale up/down).

In this PR I'm introducing the ability to refetch series for missing blocks from different store-gateway replicas which are supposed to have these blocks. I've done some manual local tests and looks working as expected; further tests will be done on staging next Monday.

Draft until:
- Further tests are done in our staging environment

Notes to reviewers:
- The [first commit](https://github.com/cortexproject/cortex/commit/d14667d2fbd2c5eb673f2f4650a6aa5d297a892a) I've pushed contains a simple refactoring without any logic change. This commit has been pushed separately to help review the next commit, which is where the changes occurred.
- Tests changes could be easier to review enabling "hide whitespace changes"
- The consistency check is now always enabled

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/blocks_store_queryable.go
+++ b/pkg/querier/blocks_store_queryable.go
@@ -95,7 +95,7 @@ func newBlocksStoreQueryableMetrics(reg prometheus.Registerer) *blocksStoreQuery
 			Namespace: "cortex",
 			Name:      "querier_storegateway_refetches_per_query",
 			Help:      "Number of re-fetches attempted while querying store-gateway instances due to missing blocks.",
-			Buckets:   []float64{0, 1, 2, 3},
+			Buckets:   []float64{0, 1, 2},
 		}),
 	}
 }
--- a/pkg/querier/blocks_store_queryable.go
+++ b/pkg/querier/blocks_store_queryable.go
@@ -411,7 +411,7 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
 
 	// We've not been able to query all expected blocks after all retries.
 	err = fmt.Errorf("consistency check failed because some blocks were not queried: %s", strings.Join(convertULIDsToString(remainingBlocks), " "))
-	level.Warn(util.WithContext(q.ctx, q.logger)).Log("msg", "failed consistency check", "err", err)
+	level.Warn(util.WithContext(spanCtx, q.logger)).Log("msg", "failed consistency check", "err", err)
 
 	return nil, nil, err
 }
--- a/pkg/querier/blocks_store_queryable.go
+++ b/pkg/querier/blocks_store_queryable.go
@@ -335,6 +335,9 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
 
 	level.Debug(spanLog).Log("msg", "found blocks to query", "expected", BlockMetas(metas).String())
 
+	// Convert matchers.
+	convertedMatchers := convertMatchersToLabelMatcher(matchers)
+
 	// Find the set of store-gateway instances having the blocks.
 	blockIDs := getULIDsFromBlockMetas(metas)
 	clients, err := q.stores.GetClientsFor(blockIDs)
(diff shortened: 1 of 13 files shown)


pracucci:
@pstibrany I've updated the CHANGELOG, taking the opportunity to backfill CHANGELOG entries for all the work done on the store-gateway (was missing so far).

pkg/querier/blocks_store_queryable.go
```diff
@@ -48,7 +55,7 @@ type BlocksStoreSet interface {
 
 	// GetClientsFor returns the store gateway clients that should be used to
 	// query the set of blocks in input.
-	GetClientsFor(blockIDs []ulid.ULID) (map[BlocksStoreClient][]ulid.ULID, error)
+	GetClientsFor(blockIDs []ulid.ULID, blacklist map[ulid.ULID][]string) (map[BlocksStoreClient][]ulid.ULID, error)
```
pstibrany, line 58: Can we extend comment to explain usage of blacklist? (Comment from https://github.com/cortexproject/cortex/pull/2695/files#diff-d7623e8d715cf64e2bad62001eac549bR362 is fine)

pkg/querier/blocks_store_queryable.go
```diff
@@ -38,6 +38,13 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 )
 
+const (
+	// The maximum number of times we retry fetching missing blocks from different
+	// store-gateways. If no more store-gateways are left (ie. due to lower replication
+	// factor) than we'll end the retries earlier.
+	maxFetchSeriesTries = 3
```
pstibrany, line 45: I'd use word attempt here (and elsewhere in the code), but perhaps better to consult with native english speaker.

pkg/querier/blocks_store_queryable.go
```diff
@@ -315,42 +326,111 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
...
+		// Ensure all expected blocks have been queried (during all tries done so far).
+		missingBlocks := q.consistency.Check(knownMetas, knownDeletionMarks, resQueriedBlocks)
+		if len(missingBlocks) == 0 {
+			q.metrics.storesHit.Observe(float64(len(touchedStores)))
+			q.metrics.refetches.Observe(float64(retry))
+
+			return storage.NewMergeSeriesSet(resSeriesSets, storage.ChainedSeriesMerge), resWarnings, nil
+		}
+
+		level.Debug(spanLog).Log("msg", "consistency check failed", "missing blocks", strings.Join(convertULIDsToString(missingBlocks), " "))
```
pstibrany, line 406: Shall we include attempt number in the log message?

pkg/querier/blocks_store_queryable.go
```diff
@@ -315,42 +326,111 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
...
+
+		// The next retry should just query the missing blocks.
+		remainingBlocks = missingBlocks
 	}
-	level.Debug(spanLog).Log("msg", "found store-gateway instances to query", "num instances", len(clients))
 
+	// We've not been able to query all expected blocks after all retries.
+	err = fmt.Errorf("consistency check failed because some blocks were not queried: %s", strings.Join(convertULIDsToString(remainingBlocks), " "))
+	level.Warn(util.WithContext(spanCtx, q.logger)).Log("msg", "failed consistency check", "err", err)
```
pstibrany, line 414: Use `spanLog` here?

pkg/querier/blocks_store_replicated_set.go
```diff
@@ -70,9 +72,14 @@ func (s *blocksStoreReplicationSet) stopping(_ error) error {
...
 
-func (s *blocksStoreReplicationSet) GetClientsFor(blockIDs []ulid.ULID) (map[BlocksStoreClient][]ulid.ULID, error) {
+func (s *blocksStoreReplicationSet) GetClientsFor(blockIDs []ulid.ULID, blacklist map[ulid.ULID][]string) (map[BlocksStoreClient][]ulid.ULID, error) {
 	shards := map[string][]ulid.ULID{}
 
+	// Easy way to handle the case of a missing blacklist.
+	if blacklist == nil {
+		blacklist = map[ulid.ULID][]string{}
+	}
+
```
pstibrany, line 82: This should not be necessary. Querying or iterating `nil` map is safe.

pracucci:
Thanks @pstibrany for your useful review. I should have addressed all comments and I've rebased too.

pkg/querier/blocks_store_queryable.go
```diff
@@ -358,20 +359,20 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
...
-		// Find the set of store-gateway instances having the blocks. The exclude list passed is the
+	for attempt := 1; attempt <= maxFetchSeriesAttempts; attempt++ {
+		// Find the set of store-gateway instances having the blocks. The exclude parameter is the
 		// map of blocks queried so far, with the list of store-gateway addresses for each block.
 		clients, err := q.stores.GetClientsFor(remainingBlocks, attemptedBlocks)
 		if err != nil {
 			// If it's a retry and we get an error, it means there are no more store-gateways left
 			// from which running another attempt, so we're just stopping retrying.
-			if retry > 0 {
+			if attempt > 1 {
```
pstibrany, line 369: There can be different errors returned by `GetClientsFor`. Should we at least log it, just in case we need to investigate it in the future?

pracucci: You're right. What's about now?

pstibrany: Thanks!

pracucci:
_I've done extensive tests in our dev environment and I haven't found any issue. I will merge once CI is green._

PR #2701, opened 2020-06-09
check response comparison flag in query tee to compare responses

sandeepsukhani:
**What this PR does**:
`query-tee` was not checking `proxy.compare-responses` and doing the comparison of responses always. This PR fixes that behaviours. It also adds the `RawQuery` in logs when comparison fails.

### DIFF
--- a/tools/querytee/proxy.go
+++ b/tools/querytee/proxy.go
@@ -134,8 +134,8 @@ func (p *Proxy) Start() error {
 	}))
 
 	// register routes
+	var comparator ResponsesComparator
 	for _, route := range p.routes {
-		var comparator ResponsesComparator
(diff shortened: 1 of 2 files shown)


tools/querytee/proxy.go
```diff
@@ -135,7 +135,11 @@ func (p *Proxy) Start() error {
 
 	// register routes
 	for _, route := range p.routes {
-		router.Path(route.Path).Methods(route.Methods).Handler(NewProxyEndpoint(p.backends, route.RouteName, p.metrics, p.logger, route.ResponseComparator))
+		var comparator ResponsesComparator
```
pracucci, line 138: Let's move this before the `for` loop

PR #2702, opened 2020-06-09
Do not wait for the slowest response in the query-tee

pracucci:
**What this PR does**:
Currently, the `query-tee` waits all backend responses before sending back the response to the client. This means that the query performances, as perceived by the client, are always influenced by the slowest backend.

In this PR I'm fixing this. The `query-tee` now sends back to the client the first viable response received, where "viable" depends on whether the preferred backend is configured. If the preferred backend is configured we need to honor it, so we have to wait for the preferred backend response before sending back the response to the client, while if it's not configured then the first successful response will be picked.

/cc @sandeepsukhani 

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added


pracucci:
@sandeepsukhani I've changed `cortex_querytee_responses_compared_total` `route_name` to `route` to keep it consistent with the already existing `cortex_querytee_request_duration_seconds` metric. I've also added a change log entry and doc for your PR #2611.

pracucci:
> What do you think about adding metric to see how often proxy replied with response from each backend?

@pstibrany Yes, makes sense. I've added the metric `cortex_querytee_responses_total`. What's your take?

sandeepsukhani:
Thanks Marco for taking care of response comparison docs as well!

PR #2705, opened 2020-06-10
Add retry with backoff on memberlist join other members

### LINKED ISSUE #2638: Clean up gossip and remove `experimental` flag from gossiped ring method in docs and config
Gossiped Ring works well, and so we should remove the `experimental` flag from the docs and config comments. In addition, we should do a few polishing steps:

* investigate error messages about being unable to ping, but tcp working. They seem harmless. Document their harmlessness or silence them if possible.
* Randomize node names used in gossip. This fixes problems with stateful sets reusing node names. "Problems" are that new nodes with old names are ignored for short period of time, which produces log messages, but they will eventually (after 30s in default config) connect.
* Periodic reconnect to cluster -- may help in situations after networking problems, or in misconfigured rollout.

periklis:
**What this PR does**:
This PR adds background retries with backoff on joining a memberlist cluster. 

Fixes a part of #2638 (Part of regarding joining a memberlist cluster)

**Checklist**
- [x] Tests updated
- [x] Documentation added

/cc @pstibrany


pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -305,13 +312,55 @@ func (m *KV) GetCodec(codecID string) codec.Codec {
...
+	if reached > 0 {
+		level.Debug(util.Logger).Log("msg", "joined memberlist cluster", "node", m.cfg.NodeName, "reached", reached)
+		return reached, nil
+	}
+
+	if m.cfg.MaxJoinRetries > 0 {
+		go m.retryJoinMembersWithBackoff(members)
+	}
+
+	return 0, nil
```
pstibrany, line 337: I suggest we move this logic to `NewKV` for now. When/if https://github.com/cortexproject/cortex/pull/2713 is merged, we can move it to `starting method. WDYT?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -308,12 +318,39 @@ func (m *KV) GetListeningPort() int {
 	return int(m.memberlist.LocalNode().Port)
 }
 
-// JoinMembers joins the cluster with given members.
+// JoinMembers joins the cluster with given members. If the first attempt fails (e.g. no members reached) and
+// immediate abort is not set JoinMembers will retry with backoff up to max retries configured.
```
pstibrany, line 322: This comment should be reverted back.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -288,11 +298,11 @@ func NewKV(cfg KVConfig) (*KV, error) {
...
 
-		if err != nil {
-			level.Error(util.Logger).Log("msg", "failed to join memberlist cluster", "err", err)
-		} else {
-			level.Info(util.Logger).Log("msg", "joined memberlist cluster", "reached_nodes", reached)
+		if reached == 0 && cfg.MaxJoinRetries > 0 {
+			go mlkv.retryJoinWithBackoff(cfg.JoinMembers)
 		}
+
+		level.Info(util.Logger).Log("msg", "joined memberlist cluster", "reached_nodes", reached)
```
pstibrany, line 305: I think this should be in `else` branch, otherwise it's not correct (It will log "joined memberlist cluster, reached_nodes=0").

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -288,11 +298,11 @@ func NewKV(cfg KVConfig) (*KV, error) {
 			return nil, err
 		}
 
-		if err != nil {
-			level.Error(util.Logger).Log("msg", "failed to join memberlist cluster", "err", err)
-		} else {
-			level.Info(util.Logger).Log("msg", "joined memberlist cluster", "reached_nodes", reached)
+		if reached == 0 && cfg.MaxJoinRetries > 0 {
+			go mlkv.retryJoinWithBackoff(cfg.JoinMembers)
```
pstibrany, line 302: Should we add log message saying something "failed to join memberlist cluster, will keep trying in background"? (And also please log at least the first error, it's useful for debugging problems)

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -308,12 +318,39 @@ func (m *KV) GetListeningPort() int {
...
 }
 
+func (m *KV) retryJoinWithBackoff(members []string) {
+	cfg := util.BackoffConfig{
+		MinBackoff: m.cfg.MinJoinBackoff,
+		MaxBackoff: m.cfg.MaxJoinBackoff,
+		MaxRetries: m.cfg.MaxJoinRetries,
+	}
+
+	backoff := util.NewBackoff(context.TODO(), cfg)
```
pstibrany, line 335:
```suggestion
	backoff := util.NewBackoff(context.Background(), cfg)
```

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -308,12 +318,39 @@ func (m *KV) GetListeningPort() int {
...
+
+	for backoff.Ongoing() {
+		reached, _ := m.memberlist.Join(members)
+		if reached == 0 {
+			level.Debug(util.Logger).Log("msg", "retry to join memberlist cluster", "node", m.cfg.NodeName, "attempts", backoff.NumRetries())
```
pstibrany, line 340: Perhaps we should include error in the log? I know they can be quite verbose, but it's better than hiding it, especially when you need to troubleshoot this.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -259,8 +267,16 @@ func NewKV(cfg KVConfig) (*KV, error) {
...
+	mlkv.createAndRegisterMetrics()
+
+	for _, c := range cfg.Codecs {
+		mlkv.codecs[c.CodecID()] = c
+	}
+
 	mlCfg.Delegate = mlkv
 
+	// Wait for memberlist and broadcasts
+	mlkv.initWG.Add(1)
```
pracucci, line 279: What is this `unitWG` used for? The `Done()` is called below before returning so I can't see how any function on this receiver may actually hit the case there's something to wait.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -288,8 +298,9 @@ func NewKV(cfg KVConfig) (*KV, error) {
 			return nil, err
 		}
 
-		if err != nil {
-			level.Error(util.Logger).Log("msg", "failed to join memberlist cluster", "err", err)
+		if reached == 0 && cfg.MaxJoinRetries > 0 {
+			level.Debug(util.Logger).Log("msg", "failed to join memberlist cluster, keep trying in the background", "node", cfg.NodeName, "err", err)
+			go mlkv.retryJoinWithBackoff(cfg.JoinMembers)
```
pracucci, line 303: Before this PR there was no retry and, in case of failure joining the cluster, the creation of the ring client was failing causing a cascading failure. This PR adds background retries, but in case it fails to join the cluster after all retries, the error is just logged so the system will keep running but not functioning correctly. Am I missing anything?

pkg/ring/kv/memberlist/memberlist_client.go
pstibrany: This actually solves real race... as soon as `memberlist.Create` is called, memberlist may start calling [delegate methods](https://github.com/cortexproject/cortex/blob/33944e9bfa25855605b62f0083129285a1b2736b/vendor/github.com/hashicorp/memberlist/delegate.go) if it receives traffic (those that call `initWG.Wait()`), but delegate initialization (`mlkv` here) is not finished yet.

pracucci: Thanks for the explaination. What do you think @periklis to enhance the comment to explain it? What Peter said is quite clear to me (which I'm not much expert of this gossip integration).

pkg/ring/kv/memberlist/memberlist_client.go
pstibrany: You're not missing anything. Me and Periklis agreed on this design until we have memberlist in service mode ready.

For now, we don't want to block `NewKV` method, so we only check the abort flag on first failure.

Once memberlist KV is converted to a service, it will do joining with backoff in `Starting` state, and if that fails and Abort flag is set, it will report error.

pracucci:
@periklis Could you also add a CHANGELOG entry, please? 🙏

pkg/ring/kv/memberlist/memberlist_client.go
pracucci: Ok. Sorry for having missed the conversation (I may have checked). Then LGTM.

pstibrany: It was private conversation, sorry for not including it here.

periklis:
> @periklis Could you also add a CHANGELOG entry, please? pray

Sure, is this is an Enhancement or feature for you?

pracucci:
I would say enhancement.

PR #2708, opened 2020-06-10
Separate HTTP Instrumentation for Querier

### LINKED ISSUE #2707: Querier API Metrics are not reported when running in Single-Binary mode
The querier registers a separate router when running in a single binary. This router is not currently instrumented with metrics.

jtlisi:
**What this PR does**:

When running cortex as a single binary the requests processed by the querier are not instrumented. This PR adds a separate histogram metric `cortex_querier_request_duration_seconds` that allows the operator to make this distinction.

The metric is registered in both single binary and microservice mode. This will allow the new metric to be used for dashboards/panels in both modes. 

Fixes #2707 

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -71,10 +71,10 @@ func New(cfg Config, s *server.Server, logger log.Logger) (*API, error) {
 	s.HTTP.UseEncodedPath()
 
 	api := &API{
-		cfg:                    cfg,
-		authMiddleware:         cfg.HTTPAuthMiddleware,
-		server:                 s,
-		logger:                 logger,
+		cfg:            cfg,
+		authMiddleware: cfg.HTTPAuthMiddleware,
+		server:         s,
+		logger:         logger,
 	}
 
 	// If no authentication middleware is present in the config, use the default authentication middleware.
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -14,15 +14,15 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/gorilla/mux"
 	"github.com/prometheus/common/route"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/storage"
 	v1 "github.com/prometheus/prometheus/web/api/v1"
+	"github.com/weaveworks/common/instrument"
 	"github.com/weaveworks/common/middleware"
 	"github.com/weaveworks/common/server"
-
-	"github.com/gorilla/mux"
 
 	"github.com/cortexproject/cortex/pkg/alertmanager"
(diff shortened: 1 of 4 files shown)


pkg/api/api.go
```diff
@@ -65,17 +65,29 @@ type API struct {
...
+	querierRequestDuration *prometheus.HistogramVec
 }
 
-func New(cfg Config, s *server.Server, logger log.Logger) (*API, error) {
+func New(cfg Config, s *server.Server, logger log.Logger, reg prometheus.Registerer) (*API, error) {
 	// Ensure the encoded path is used. Required for the rules API
 	s.HTTP.UseEncodedPath()
 
+	// Prometheus histograms for requests.
+	querierRequestDuration := prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
pracucci, line 77: You can simply use `promauto.With(reg).NewHistogramVec()`.

pkg/api/api.go
```diff
@@ -331,31 +350,34 @@ func (a *API) RegisterQuerier(queryable storage.Queryable, engine *promql.Engine
 	}))
 }
 
+// RegisterQueryAPI registers the Prometheus routes supported by the
+// Cortex querier service. Currently this can not be registered simultaneously
+// with the Querier.
+func (a *API) RegisterQueryAPI(handler http.Handler) {
```
pracucci, line 356: Is this refactoring related to the scope of the PR? I'm not against it if you need it in a project vendoring Cortex but, if it's not required, probably the previous version was more clear.

pkg/api/api.go
pracucci, line 77: Have you considered moving this to `RegisterQuerier()`? The metric would be registered only if `RegisterQuerier()` is used and the change isolated there.

jtlisi: I wanted to avoid passing a `prometheus.Registerer` to that function. However, maybe it would make more sense to store the Registerer in the API struct and and initiate the metric is in the `RegisterQuerier` function.

pracucci: Yes, or you can pass the metric itself to the `RegisterQuerier()` (ie. in Thanos it's quite a common design pattern).

jtlisi: 👍 Cool I updated it to pass a metric as a parameter.

pkg/api/api.go
pracucci: Could you comment on this? I'm fine if you need it in a project vendoring Cortex, otherwise the previous version looks more explicit to me.

jtlisi:
> https://github.com/cortexproject/cortex/pull/2708#discussion_r440727442

I don't have any particular use case for that function. It came about when I was expirementing with alternative ways of solving the "separate querier instrumentation" problem. I kept it because I thought it looked cleaner. However, I can return it to it's previous form if you fell otherwise.

pracucci:
Up to your judgement, but if we keep it we could make it unexported so that
we signal it's just used internally.

PR #2709, opened 2020-06-10
add . to chunk cache name for clearer instrumentation

jtlisi:
**What this PR does**:

Renames the label used for chunk cache metrics to be prefixed with `chunks.` instead of chunks.

Currently the chunks cache name label is inconsistent with the reset of the metrics. It is currently formatted as `chunksmemcached.`. The rest of the labels have a dot separation between the cache name and cache type.

Screen Shot 2020-06-10 at 1 54 01 PM

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/storage/factory.go
+++ b/pkg/chunk/storage/factory.go
@@ -128,7 +128,7 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 	}
 
 	chunkCacheCfg := storeCfg.ChunkCacheConfig
-	chunkCacheCfg.Prefix = "chunks"
+	chunkCacheCfg.Prefix = "chunks."
 	chunksCache, err := cache.New(chunkCacheCfg)
 	if err != nil {
 		return nil, err


pkg/chunk/storage/factory.go
```diff
@@ -128,7 +128,7 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 	}
 
 	chunkCacheCfg := storeCfg.ChunkCacheConfig
-	chunkCacheCfg.Prefix = "chunks"
+	chunkCacheCfg.Prefix = "chunks."
```
pracucci, line 131: There few implications in this change and I'm not sure the change here is correct. For example, this prefix is also used by fifocache `Instrument(cfg.Prefix+"fifocache", cache)` which then leads to `method := i.name + ".store"`. Adding the dot will cause the method name to be `chunks..store`.

Could you take a deeper look at the effect of this change, please?

jtlisi:
I'm closing this for now. I'll reopen at some point when I have more time.

PR #2711, opened 2020-06-11
load delete requests frequently if there are more pending requests for a user

sandeepsukhani:
**What this PR does**:
Purger checks for pending delete requests hourly and we only process only 1 delete requests for a user at a time to avoid some problems. This means we process only 1 delete request per user per hour. This should not be a problem for most of the time but it can for some cases. One problem I noticed is with test exporter which creates a delete request every 5 mins which means purger would never be able to catch up.

This PR loads more delete request as soon as a delete request gets processed which belonged to a user who has more pending requests.

### DIFF
--- a/pkg/chunk/purger/purger.go
+++ b/pkg/chunk/purger/purger.go
@@ -125,7 +125,7 @@ func NewDataPurger(cfg Config, deleteStore *DeleteStore, chunkStore chunk.Store,
 		chunkStore:               chunkStore,
 		objectClient:             storageClient,
 		metrics:                  newPurgerMetrics(registerer),
-		pullNewRequestsChan:      make(chan struct{}, 2),
+		pullNewRequestsChan:      make(chan struct{}, 1),
 		executePlansChan:         make(chan deleteRequestWithLogger, 50),
 		workerJobChan:            make(chan workerJob, 50),
 		inProcessRequestIDs:      map[string]string{},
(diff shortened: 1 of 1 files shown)


pkg/chunk/purger/purger.go
```diff
@@ -151,6 +150,24 @@ func (dp *DataPurger) init(ctx context.Context) error {
 	return dp.loadInprocessDeleteRequests()
 }
 
+func (dp *DataPurger) loop(ctx context.Context) error {
+	loadRequestsTicker := time.NewTicker(time.Hour)
+	defer loadRequestsTicker.Stop()
+
+	for {
+		select {
+		case <-loadRequestsTicker.C:
```
pstibrany, line 159: Nothing currently happens on tick, which I assume is not what you want. Unfortunately, [`fallthrough` or multi-select is not an option](https://github.com/golang/go/issues/23196) :-(

pkg/chunk/purger/purger.go
```diff
@@ -183,6 +200,18 @@ func (dp *DataPurger) workerJobCleanup(job workerJob) {
...
+		// request loading of more delete request if
+		// - user has more pending requests and
+		// - we do not have a pending request to load more requests
+		dp.usersWithPendingRequestsMtx.Lock()
+		defer dp.usersWithPendingRequestsMtx.Unlock()
+		if _, ok := dp.usersWithPendingRequests[job.userID]; ok {
+			delete(dp.usersWithPendingRequests, job.userID)
+			if len(dp.pullNewRequestsChan) == 0 {
+				dp.pullNewRequestsChan <- struct{}{}
+			}
```
pstibrany, line 213: Standard idiom for non-blocking send is:

```
select {
case dp.pullNewRequestsChan <- struct{}{}:
    // sent
default:
    // chan is full, not sent
}
```

pkg/chunk/purger/purger.go
sandeepsukhani: hmm, sorry my bad. I thought it would work the same as fallthrough in switch statement. Will fix it.

pstibrany: Unit test my be a good idea here :)

sandeepsukhani:
Unit test can be written but for that I would have to make [deleteRequestCancellationDeadline](https://github.com/cortexproject/cortex/pull/2711/files#diff-9bbc5791559c0580c57ce196ee8cc0d6L30) which is what decides the minimum age of delete requests to be picked up for processing.
I anyways have to make it configurable for integration tests. I can do it now if you guys think that is the only way out. If we do allow it to be configured by the user, we need to put a strong warning that anything less than 24h(or max chunk age at least) could result in an unexpected outcome.

pstibrany:
I am fine if it is in next PR.

PR #2713, opened 2020-06-11
Make memberlist KV a service.

pstibrany:
**What this PR does**: This PR modifies `memberlist.KV` into a Cortex service:
- it needs to be started to start gossiping part
- if configured to join the cluster and abort if join fails, Starting can result in Failed state
- KV clients now wait for service to enter Running or Stopping state. (Stopping because we still receive updates until cluster is left)

This is first step in adding periodic rejoins (https://github.com/cortexproject/cortex/issues/2638), and should also make https://github.com/cortexproject/cortex/pull/2705 bit easier (we can use backoff in Starting state, and not block NewKV)

Update:
- Now that #2705 is merged, joining of the cluster with backoff is now part of Starting state, and Abort flag is honored.

**Checklist**
- [x] Tests updated


integration/getting_started_with_gossiped_ring_test.go
```diff
@@ -28,6 +28,7 @@ func TestGettingStartedWithGossipedRing(t *testing.T) {
 		// decrease timeouts to make test faster. should still be fine with two instances only
 		"-ingester.join-after":     "0s", // join quickly
 		"-ingester.observe-period": "5s", // to avoid conflicts in tokens
+		"-log.level":               "debug",
```
pracucci, line 31: Just for testing or intended to keep it?

pkg/ring/kv/memberlist/kv_init_service.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+}
+
+func (kvs *KVInitService) stopping(_ error) error {
+	if kvs.kv == nil {
+		return nil
+	}
+
+	kvs.kv.StopAsync()
+	return kvs.kv.AwaitTerminated(context.Background())
```
pracucci, line 63: Can we use `services.StopAndAwaitTerminated()` for convenience?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -335,49 +396,65 @@ func (m *KV) GetCodec(codecID string) codec.Codec {
...
 
-		level.Debug(util.Logger).Log("msg", "joined memberlist cluster", "node", m.cfg.NodeName, "reached", reached)
+		level.Info(util.Logger).Log("msg", "joined memberlist cluster", "reached_nodes", reached)
 		break
 	}
 
-	if backoff.Err() != nil {
-		level.Error(util.Logger).Log("msg", "failed to join memberlist cluster", "node", m.cfg.NodeName, "err", backoff.Err())
-	}
+	return lastErr
```
pracucci, line 450: I think there's a bug here: `lastErr` is not reset to `nil` before the `break` in the loop. I think we can simplify the logic switching `break` into `return nil`.

pstibrany: Correct. Added test to catch this. Also added test for scenario when memberlist fails to join the cluster and abort flag is true.

PR #2717, opened 2020-06-11
Added proposal for ingester+querier migration from chunks to blocks and back.

pstibrany:
**What this PR does**: This is a proposal for adding some options for making switch of ingesters and querier between chunks and blocks possible.

### DIFF
--- a/docs/proposals/ingesters-migration.md
+++ b/docs/proposals/ingesters-migration.md
@@ -29,6 +29,7 @@ How do we handle ingesters with WAL? There are several possibilities, but the si
 2. Turn off chunks, enable TSDB (second rollout). During the second rollout, as the ingester shuts down, it will flush all chunks in memory, and when it restarts, it will start using TSDB.
 
 Benefit of this approach is that it is trivial to add the flag, and then rollout in both steps can be fully automated.
+In this scenario, we will reconfigure existing statefulset of ingesters to use blocks in step 2 – we are not creating new statefulset.
 
 Notice that querier can ask only ingesters for most recent data and not consult the store, but during the rollout (and some time after), ingesters that are already using blocks will **not** have the most recent chunks in memory. To make sure queries work correctly, `-querier.query-store-after` needs to be set to 0, in order for queriers to not rely on ingesters only for most recent data. After couple of hours after rollout, this value can be increased again, depending on how much data ingesters keep. (`-experimental.tsdb.retention-period` for blocks, `-ingester.retain-period` for chunks)
 
--- a/docs/proposals/ingesters-migration.md
+++ b/docs/proposals/ingesters-migration.md
@@ -39,11 +39,17 @@ Other alternatives considered for flushing chunks / handling WAL:
 
 After all ingesters are converted to blocks, we can set cut-off time for querying chunks storage on queriers.
 
-For rollback from blocks to chunks, we need to be able to flush data from ingesters to the blocks storage, and then switch ingesters back to chunks. Ingesters are currently not able to flush blocks to storage, but adding flush-on-shutdown option, support for `/shutdown` endpoint and support in flusher component similar to chunks is doable, and should be part of this work.
+For rollback from blocks to chunks, we need to be able to flush data from ingesters to the blocks storage, and then switch ingesters back to chunks.
+Ingesters are currently not able to flush blocks to storage, but adding flush-on-shutdown option, support for `/shutdown` endpoint and support in flusher component similar to chunks is doable, and should be part of this work.
 
 With this ability, rollback would follow the same process, just in reverse: 1) redeploy with flush flag enabled, 2) redeploy with config change from blocks to chunks.
+Note that this isn't a *full* rollback to chunks-only solution, as generated blocks still need to be queried after the rollback, otherwise samples pushed to blocks would be missing.
+This means running store-gateways and queriers that can query both chunks and blocks store.
 
-Alternative plan could be to use a separate Cortex cluster configured to use blocks, and redirect incoming traffic to both chunks and blocks cluster. When one is confident about the blocks cluster running correctly, old chunks cluster can be shutdown. In this plan, there is an overlap where both clusters are ingesting same data. Blocks cluster needs to be configured to be able to query chunks storage as well, with cut-off time based on when clusters were configured (at latest, to minimize amount of duplicated samples that need to be processed during queries.)
+Alternative plan could be to use a separate Cortex cluster configured to use blocks, and redirect incoming traffic to both chunks and blocks cluster.
+When one is confident about the blocks cluster running correctly, old chunks cluster can be shutdown.
+In this plan, there is an overlap where both clusters are ingesting same data.
+Blocks cluster needs to be configured to be able to query chunks storage as well, with cut-off time based on when clusters were configured (at latest, to minimize amount of duplicated samples that need to be processed during queries.)
 
 ## Querying
 
@@ -54,7 +60,7 @@ For querying chunks storage, we have two options:
 - Always query the chunks store – useful during ingesters switch, or after rollback from blocks to chunks.
 - Query chunk store only for queries that ask for data after specific cut-off time. This is useful after all ingesters have switched, and we know the timestamp since ingesters are only writing blocks.
 
-Querier needs to support both. Switching between them via config option seems fine. (We could use runtime-config for on-the-fly switch without restarts)
+Querier needs to support both modes of running. While we could use runtime-config for on-the-fly switch without restarts, queriers restart quickly and so switching via configuration or command line option seems enough.
 
 ## Work to do
 
--- a/docs/proposals/ingesters-migration.md
+++ b/docs/proposals/ingesters-migration.md
@@ -47,7 +47,7 @@ After all ingesters are converted to blocks, we can set cut-off time for queryin
 For rollback from blocks to chunks, we need to be able to flush data from ingesters to the blocks storage, and then switch ingesters back to chunks.
 Ingesters are currently not able to flush blocks to storage, but adding flush-on-shutdown option, support for `/shutdown` endpoint and support in flusher component similar to chunks is doable, and should be part of this work.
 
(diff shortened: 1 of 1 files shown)


docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,60 @@
...
+- Status: Proposed
+
+# Migrating ingesters from chunks to blocks
+
+This short document describes the first step in full migration of the Cortex cluster from using chunks storage to using blocks storage, specifically switching ingesters to using blocks, and modification of queriers to query both chunks and blocks storage.
+
+## Ingesters
+
+When switching ingesters from chunks to blocks, we need to consider the following:
+- Ingesting of new data, and querying should work during the switch.
```
pracucci, line 20: Just to fix the website markdown compilation.

```suggestion

- Ingesting of new data, and querying should work during the switch.
```

docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,60 @@
...
+- Ingesters are rolled out with new configuration over time. There is overlap: ingesters of both kinds (chunks, blocks) are running at the same time.
+- Ingesters using WAL (all environments in GL) don’t flush in-memory chunks to storage on shutdown.
```
pracucci, line 22: I wouldn't mention GL specific setup in a generic proposal.

```suggestion
- Ingesters using WAL don’t flush in-memory chunks to storage on shutdown.
```

docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,60 @@
...
+- Rollout should be as automated as possible.
+
+How do we handle ingesters with WAL? There are several possibilities, but the simplest option seems to be adding a new flag to ingesters to flush chunks on shutdown. This is trivial change to ingester, and allows us to do automated migration by:
+1) enabling this flag on each ingester (first rollout).
+2) turn off chunks, enable TSDB (second rollout). During the second rollout, as the ingester shuts down, it will flush all chunks in memory, and when it restarts, it will start using TSDB.
```
pracucci, line 27: Just to fix the website markdown compilation.
```suggestion

1. Enabling this flag on each ingester (first rollout).
2. Turn off chunks, enable TSDB (second rollout). During the second rollout, as the ingester shuts down, it will flush all chunks in memory, and when it restarts, it will start using TSDB.
```

docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,60 @@
...
+* Shutdown single chunks-ingester, run flusher in its place, and when done start new blocks ingester. This is similar to the procedure we did during the introduction of WAL. Flusher can be run via initContainer support in pods. This still requires two-step deployment: 1) enable flusher and reconfigure ingesters to use blocks, 2) remove flusher.
+
+After all ingesters are converted to blocks, we can set cut-off time for querying chunks storage on queriers.
+
+For rollback from blocks to chunks, we need to be able to flush data from ingesters to the blocks storage, and then switch ingesters back to chunks. Ingesters are currently not able to flush blocks to storage, but adding flush-on-shutdown option, support for `/shutdown` endpoint and support in flusher component similar to chunks is doable, and should be part of this work.
+
+With this ability, rollback would follow the same process, just in reverse: 1) redeploy with flush flag enabled, 2) redeploy with config change from blocks to chunks.
```
pracucci, line 42: I would mention that it's not a pure rollback because rolledback queriers will have to query blocks anyway, otherwise series pushed while ingesters were running the blocks storage will be missing.

docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Querying
+
+To be able to query both blocks and chunks, querier needs to be extended with this ability, and merge results from both. Queriers can always query blocks – each querier knows about existing blocks and their timeranges, so it can quickly determine whether there are any blocks with relevant data. Always querying blocks is also useful when there is some background process converting chunks to blocks. As new blocks appear on the store, they get queried even if they are old.
+
+For querying chunks storage, we have two options:
+- Always query the chunks store – useful during ingesters switch, or after rollback from blocks to chunks.
```
pracucci, line 51:
```suggestion

- Always query the chunks store – useful during ingesters switch, or after rollback from blocks to chunks.
```

docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,60 @@
...
+- Query chunk store only for queries that ask for data after specific cut-off time. This is useful after all ingesters have switched, and we know the timestamp since ingesters are only writing blocks.
+
+Querier needs to support both. Switching between them via config option seems fine. (We could use runtime-config for on-the-fly switch without restarts)
```
pracucci, line 54: Queriers are stateless and restarting them is quick. I don't think we need to use the runtime-config for that.

docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Work to do
+
+- Ingester: Add flags for always flushing on shutdown, even when using WAL or blocks.
+- Querier: Add support for querying both chunk store and blocks at the same time
```
pracucci, line 59:
```suggestion
- Querier: Add support for querying both chunk store and blocks at the same time and test the support for querying both chunks and blocks from ingesters works correctly
```

docs/proposals/ingesters-migration.md
```diff
@@ -0,0 +1,69 @@
...
+## Ingesters
+
+When switching ingesters from chunks to blocks, we need to consider the following:
+
+- Ingesting of new data, and querying should work during the switch.
+- Ingesters are rolled out with new configuration over time. There is overlap: ingesters of both kinds (chunks, blocks) are running at the same time.
+- Ingesters using WAL don’t flush in-memory chunks to storage on shutdown.
+- Rollout should be as automated as possible.
+
+How do we handle ingesters with WAL? There are several possibilities, but the simplest option seems to be adding a new flag to ingesters to flush chunks on shutdown. This is trivial change to ingester, and allows us to do automated migration by:
```
gouthamve, line 26: Or something simpler, just don't send any data for 30mins (idle-timeout) before shutting down :)

gouthamve: Oh wait, both the ingester deployments will share the same ring? That makes sense. Can we call out that we will be using the same statefulset and not going to create a new one for blocks? 

We also have a shutdown endpoint on the chunks ingester which just flushes and shuts down.

pstibrany: > We also have a shutdown endpoint on the chunks ingester which just flushes and shuts down.

Yes we do, but I don't quite see how to use it for automation. On calling /shutdown, ingester flushes everything and stops. Kubernetes will just restart it. But we need restart with new configuration.

gouthamve: I think it idles out, and doesn't stop iirc. @codesome ?

codesome: @gouthamve you are right. Kubernetes won't restart it. /shutdown endpoint will flush all the chunks and remove itself from the ring and stay idle.

gouthamve:
So somethings that are missing here but would provide more context would be:

1. Mention that we are only tackling WAL (or statefulsets). Maybe a section on how it would work with deployments?
2. Mention that we share the same ring and queriers will only see a single set of ingesters (chunks + blocks) at times.
3. Mention that the queriers will be configured to read from blocks on object store + chunks on NoSQL store at the same time.
4. Mention how the potential schema config will look like, specially in the case of a rollback. Currently we support splitting it by the UTC day boundary, but how will that work because the rollout won't align to that. I see that for a particular time-range we need to query both.

These are some concerns I see and we can iterate on the proposal again once these are addressed?

codesome:
> Maybe a section on how it would work with deployments?

I think it will look similar to migration from deployments to WAL.

pstibrany:
> I think it will look similar to migration from deployments to WAL.

I think it's actually bit easier, since we cannot transfer data from chunks ingesters to blocks ingesters, so they don't need to be scaled up and down in lock-step. But chunks ingesters must be reconfigured to avoid transfers.

docs/proposals/ingesters-migration.md
pstibrany: > 4. Mention how the potential schema config will look like, specially in the case of a rollback. Currently we support splitting it by the UTC day boundary, but how will that work because the rollout won't align to that. I see that for a particular time-range we need to query both.

Schemas are low-level thing for configuring chunks store. We will not use schema for anything in this proposal.

Querier needs single timestamp to decide whether to query chunks store or not. Blocks store can always be queried, because querier knows whether to hit any block based on block metadata.

See my PR #2747 which implements this already.

I've updated "Querying" section of the document to better explain this.

pstibrany:
> 4. Mention how the potential schema config will look like, specially in the case of a rollback. Currently we support splitting it by the UTC day boundary, but how will that work because the rollout won't align to that. I see that for a particular time-range we need to query both.

Schemas are low-level thing for configuring chunks store. We will not use schema for anything in this proposal.

Querier needs single timestamp to decide whether to query chunks store or not. Blocks store can always be queried, because querier knows whether to hit any block based on block metadata.

See my PR #2747 which implements this already.

I've updated "Querying" section of the document to better explain this.

pstibrany:
@gouthamve I've updated the document with your feedback. Please take a look again when you have time.

docs/proposals/ingesters-migration.md
```diff
@@ -58,14 +58,23 @@ Blocks cluster needs to be configured to be able to query chunks storage as well
...
 For querying chunks storage, we have two options:
 
 - Always query the chunks store – useful during ingesters switch, or after rollback from blocks to chunks.
 - Query chunk store only for queries that ask for data after specific cut-off time. This is useful after all ingesters have switched, and we know the timestamp since ingesters are only writing blocks.
 
-Querier needs to support both modes of running. While we could use runtime-config for on-the-fly switch without restarts, queriers restart quickly and so switching via configuration or command line option seems enough.
+Querier needs to support both modes of querying chunks store.
+Which one of these two modes is used depends on single timestamp flag passed to the querier.
+If timestamp is configured, chunks store is only used for queries that ask for data older than timestamp.
+If timestamp is not configured, chunks store is always querierd.
```
pracucci, line 71:
```suggestion
If timestamp is not configured, chunks store is always queried.
```

pracucci:
_Let's merge!_

PR #2722, opened 2020-06-13
more detailed slow query logging

jtlisi:
**What this PR does**:

This PR updates the query logging functionality in the Frontend to log more details about the query. It now includes the `tenant id` and `method` of the request. It also consolidates form from the query parameters and post form into a single loop. This makes it easier to analyze queries the use different methods, POST or GET.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -51,7 +51,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxOutstandingPerTenant, "querier.max-outstanding-requests-per-tenant", 100, "Maximum number of outstanding requests per tenant per frontend; requests beyond this error with HTTP 429.")
 	f.BoolVar(&cfg.CompressResponses, "querier.compress-http-responses", false, "Compress HTTP responses.")
 	f.StringVar(&cfg.DownstreamURL, "frontend.downstream-url", "", "URL of downstream Prometheus.")
-	f.DurationVar(&cfg.LogQueriesLongerThan, "frontend.log-queries-longer-than", 0, "Log queries that are slower than the specified duration. 0 to disable.")
+	f.DurationVar(&cfg.LogQueriesLongerThan, "frontend.log-queries-longer-than", 0, "Log queries that are slower than the specified duration. Set to 0 to disable. Set to < 0 to enable on all queries.")
 }
 
 // Frontend queues HTTP requests, dispatches them to backends, and handles retries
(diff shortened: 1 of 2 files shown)


pkg/querier/frontend/frontend.go
```diff
@@ -164,34 +164,45 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
...
+	} else {
+		hs := w.Header()
+		for h, vs := range resp.Header {
+			hs[h] = vs
+		}
+		w.WriteHeader(resp.StatusCode)
+		io.Copy(w, resp.Body)
+	}
+
+	if f.cfg.LogQueriesLongerThan == 0 && queryResponseTime > f.cfg.LogQueriesLongerThan {
```
pstibrany, line 178: This condition seems wrong. If `f.cfg.LogQueriesLongerThan` is zero, then all `queryResponseTimes` will be higher than that.

Changelog says `0` is disabled.

jtlisi: Sorry about that. That statement should `!= 0`

pkg/querier/frontend/frontend.go
```diff
@@ -164,34 +164,45 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
...
+	} else {
+		hs := w.Header()
+		for h, vs := range resp.Header {
+			hs[h] = vs
+		}
+		w.WriteHeader(resp.StatusCode)
+		io.Copy(w, resp.Body)
+	}
+
+	if f.cfg.LogQueriesLongerThan != 0 && queryResponseTime > f.cfg.LogQueriesLongerThan {
```
pstibrany, line 178: I'd add comment here that negative value of `f.cfg.LogQueriesLongerThan` is explicitly allowed. Otherwise it looks hacky and someone may remove it by accident.

pkg/querier/frontend/frontend.go
```diff
@@ -164,34 +164,46 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
...
+			logMessage = append(logMessage, "tenant_id", id)
 		}
-		pf := r.PostForm.Encode()
-		if pf != "" {
-			logMessage = append(logMessage, "body", pf)
+
+		// Ensure the form has been parsed so all the parameters are present
+		err = r.ParseForm()
+		if err != nil {
+			level.Warn(util.WithContext(r.Context(), f.log)).Log("unable to parse form for request")
```
pracucci, line 197: Could you add the error to the log here?

pkg/querier/frontend/frontend.go
```diff
@@ -164,34 +164,46 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
...
 			"path", r.URL.Path,
 			"time_taken", queryResponseTime.String(),
 		}
-		for k, v := range r.URL.Query() {
-			logMessage = append(logMessage, fmt.Sprintf("qs_%s", k), strings.Join(v, ","))
+
+		// Extract the ID from the header of the request
+		id, _, err := user.ExtractOrgIDFromHTTPRequest(r)
+		if err != nil {
+			logMessage = append(logMessage, "tenant_id", id)
```
pracucci, line 191: Logging with `util.WithContext()` (done below) already injects the `org_id`. Why do we need to add it here too?

pkg/querier/frontend/frontend.go
```diff
@@ -164,34 +164,46 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
...
-		level.Info(util.WithContext(r.Context(), f.log)).Log(logMessage...)
-	}
 
-	if err != nil {
-		writeError(w, err)
-		return
-	}
+		// Attempt to iterate through the Form to log any filled in values
+		for k, v := range r.Form {
+			logMessage = append(logMessage, fmt.Sprintf("qs_%s", k), strings.Join(v, ","))
```
pracucci, line 202: The field prefix is `qs_` which was originally done because we parsed only the query string. I'm wondering if we should rename it for clarity, ie. `param_`.

pkg/querier/frontend/frontend.go
jtlisi: It does, sorry I was confused since the weaveworks common server logs do not include the org_id

pkg/querier/frontend/frontend.go
jtlisi: 👍

pkg/querier/frontend/frontend.go
jtlisi: I agree, I only kept `qs` since that is what was used previously.

pkg/querier/frontend/frontend.go
```diff
@@ -185,21 +186,15 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
...
-		id, _, err := user.ExtractOrgIDFromHTTPRequest(r)
-		if err != nil {
-			logMessage = append(logMessage, "tenant_id", id)
-		}
-
 		// Ensure the form has been parsed so all the parameters are present
 		err = r.ParseForm()
 		if err != nil {
-			level.Warn(util.WithContext(r.Context(), f.log)).Log("unable to parse form for request")
+			level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
```
pracucci, line 192: Super small thing, just to be consistent across the codebase.

```suggestion
			level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "error", err)
```

PR #2725, opened 2020-06-15
Unexperimentalize memberlist

### LINKED ISSUE #2638: Clean up gossip and remove `experimental` flag from gossiped ring method in docs and config

pstibrany:
**What this PR does**: This PR adds missing flags from documentation, and marks Memberlist as no longer experimental. Note that this PR assumes #2713 and #2724 to be merged first.

Fixes #2638

**Checklist**
- [ ] Tests updated
- [x] Documentation added


pstibrany:
Marking as draft to prevent merging, until #2713 and #2724 are merged.

PR #2726, opened 2020-06-15
Remove old temporary checkpoints

csmarchbanks:
If an ingester crashes while doing a checkpoint, which is likely as
checkpoints are almost always happening, then a checkpoint like
checkpoint.123456.tmp will be left on disk. When the next checkpoint
succeeds, the old .tmp checkpoints are not deleted even though it will
never be used. After enough crashes it is possible to fill the entire
disk with .tmp checkpoints, bringing down an ingester.

**Checklist**
- [x] Tests updated
- [x] Documentation added (Not needed)

### DIFF
--- a/pkg/ingester/wal.go
+++ b/pkg/ingester/wal.go
@@ -8,9 +8,9 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/ingester/wal.go
```diff
@@ -413,16 +413,13 @@ func lastCheckpoint(dir string) (string, int, error) {
 	for i := 0; i < len(dirs); i++ {
 		di := dirs[i]
 
-		if !strings.HasPrefix(di.Name(), checkpointPrefix) {
+		idx, err := checkpointIndex(di.Name())
```
codesome, line 416: This would consider a temporary checkpoint as a valid checkpoint for the last checkpoint, which is wrong. We should be skipping them here.

csmarchbanks: :+1: Good catch, thanks! I made `checkpointIndex` take another arg as to whether temporary checkpoints should be included or not.

pkg/ingester/wal.go
```diff
@@ -464,6 +458,23 @@ func (w *walWrapper) deleteCheckpoints(maxIndex int) (err error) {
 	return errs.Err()
 }
 
+var checkpointRe = regexp.MustCompile("^" + checkpointPrefix + "(\\d+)(\\.tmp)?$")
```
codesome, line 461: While this works, `checkpointPrefix` is `checkpoint.`, while it should be `checkpoint\\.` for the regex. Else it also matches something like `checkpoints12345`.

csmarchbanks:
@pstibrany Thanks for pointing out `regexp.QuoteMeta`, I didn't know that function existed! Addressed the comment, added a test case, and will merge soon unless someone opposes.

PR #2728, opened 2020-06-16
change flag DisableChunksDeduplication to DisableIndexDeduplication for disabling just index deduplication

sandeepsukhani:
**What this PR does**:
`DisableChunksDeduplication` was added for disabling chunks deduplication for boltdb shipper in Loki but the problem is it still makes us write an already written chunk again to the storage while the requirement is to not dedupe just the index.
This PR changes flag `DisableChunksDeduplication` to `DisableIndexDeduplication` to disable deduping of just the index and not both chunk+index. It also still writes the chunk to cache for write performance.

### DIFF
--- a/pkg/chunk/series_store.go
+++ b/pkg/chunk/series_store.go
@@ -419,13 +419,18 @@ func (c *seriesStore) Put(ctx context.Context, chunks []Chunk) error {
 // PutOne implements ChunkStore
 func (c *seriesStore) PutOne(ctx context.Context, from, through model.Time, chunk Chunk) error {
 	log, ctx := spanlogger.New(ctx, "SeriesStore.PutOne")
-	if !c.cfg.DisableChunksDeduplication {
-		// If this chunk is in cache it must already be in the database so we don't need to write it again
-		found, _, _ := c.cache.Fetch(ctx, []string{chunk.ExternalKey()})
-		if len(found) > 0 {
-			dedupedChunksTotal.Inc()
-			return nil
-		}
+
+	// If this chunk is in cache it must already be in the database so we don't need to write it again
+	found, _, _ := c.cache.Fetch(ctx, []string{chunk.ExternalKey()})
+	if len(found) > 0 {
+		dedupedChunksTotal.Inc()
+	}
+
+	// If chunk is in the cache and DisableIndexDeduplication is false, we do not have to do anything.
+	// If chunk is in the cache and DisableIndexDeduplication is true, we have to write index and not chunk.
+	// Otherwise write both index and chunk.
+	if len(found) > 0 && !c.cfg.DisableIndexDeduplication {
+		return nil
 	}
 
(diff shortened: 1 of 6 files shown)


pstibrany:
Would it make sense to have a unit test for this change?

pkg/chunk/series_store.go
```diff
@@ -440,16 +445,23 @@ func (c *seriesStore) PutOne(ctx context.Context, from, through model.Time, chun
 			return err
 		}
 	} else {
-		err := c.storage.PutChunks(ctx, chunks)
-		if err != nil {
-			return err
+		// chunk not found, write it.
+		if len(found) == 0 {
```
pstibrany, line 449: Can we introduce new bool variables `writeChunk` and `writeIndex`, and set them at the beginning based on `len(found)` and `c.cfg.DisableIndexDeduplication`? I think it would make code slightly easier to follow. WDYT?

pkg/chunk/series_store.go
```diff
@@ -440,16 +445,23 @@ func (c *seriesStore) PutOne(ctx context.Context, from, through model.Time, chun
...
+			}
 		}
 		if err := c.index.BatchWrite(ctx, writeReqs); err != nil {
 			return err
 		}
 	}
-	if cacheErr := c.writeBackCache(ctx, chunks); cacheErr != nil {
-		level.Warn(log).Log("msg", "could not store chunks in chunk cache", "err", cacheErr)
+
+	// we already have the chunk in the cache so don't write it.
```
pstibrany, line 460: I'm not sure this comment belongs here... here we write the chunk to the cache.

pkg/chunk/series_store.go
```diff
@@ -440,16 +445,23 @@ func (c *seriesStore) PutOne(ctx context.Context, from, through model.Time, chun
 			return err
```
pstibrany, line 445: What about this branch? If storage supports writing chunk and index at the same time, both are written, even if chunk is already stored.

(GitHub won't let me comment on line 438, but that's where it's supposed to be)

pkg/chunk/series_store.go
sandeepsukhani: Sorry for the confusion, I meant not writing the chunk to the cache here, will update the comment.

pkg/chunk/series_store.go
pracucci: Agree on this.

sandeepsukhani: I have added just the `writeChunk` variable because writing chunk is optional while the index is not.

pkg/chunk/series_store.go
sandeepsukhani: While this change is not relevant for dynamodb which is the only store which implements `PutChunksAndIndex` with the added condition of using dynamodb for storing both index and chunks, I have changed the code to not break if there are any changes in future.

PR #2730, opened 2020-06-16
more metrics in purger for monitoring delete requests processing progress

sandeepsukhani:
**What this PR does**:
Adds the following 3 new metrics for monitoring the progress of the processing of delete request:
* `purger_load_pending_requests_attempts_total`: Number of attempts that were made to load pending requests with status.
* `purger_oldest_pending_delete_request_age_seconds`: Age of oldest pending delete request in seconds.
* `purger_pending_delete_requests_count`: Count of requests which are in process or are ready to be processed.

This PR also loads new delete requests for processing on startup instead of waiting for the first tick, which is an hour now.

### DIFF
--- a/pkg/chunk/purger/purger.go
+++ b/pkg/chunk/purger/purger.go
@@ -119,7 +119,7 @@ type DataPurger struct {
 
 	// we would only allow processing of singe delete request at a time since delete requests touching same chunks could change the chunk IDs of partially deleted chunks
 	// and break the purge plan for other requests
-	inProcessRequestIDs    map[string]DeleteRequest
+	inProcessRequests      map[string]DeleteRequest
 	inProcessRequestIDsMtx sync.RWMutex
 
 	// We do not want to limit pulling new delete requests to a fixed interval which otherwise would limit number of delete requests we process per user.
@@ -148,7 +148,7 @@ func NewDataPurger(cfg Config, deleteStore *DeleteStore, chunkStore chunk.Store,
 		pullNewRequestsChan:      make(chan struct{}, 1),
 		executePlansChan:         make(chan deleteRequestWithLogger, 50),
 		workerJobChan:            make(chan workerJob, 50),
-		inProcessRequestIDs:      map[string]DeleteRequest{},
+		inProcessRequests:        map[string]DeleteRequest{},
 		usersWithPendingRequests: map[string]struct{}{},
 		pendingPlansCount:        map[string]int{},
 	}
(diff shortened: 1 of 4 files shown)


pkg/chunk/purger/purger.go
```diff
@@ -99,7 +119,7 @@ type DataPurger struct {
 
 	// we would only allow processing of singe delete request at a time since delete requests touching same chunks could change the chunk IDs of partially deleted chunks
 	// and break the purge plan for other requests
-	inProcessRequestIDs    map[string]string
+	inProcessRequestIDs    map[string]DeleteRequest
```
pstibrany, line 122: Rename to `inProcessRequests`? :)

pkg/chunk/purger/purger.go
```diff
@@ -424,6 +472,9 @@ func (dp *DataPurger) pullDeleteRequestsToPlanDeletes() error {
 		dp.executePlansChan <- req
 	}
 
+	dp.metrics.oldestPendingDeleteRequestAgeSeconds.Set(float64(now.Sub(oldestPendingRequestCreatedAt) / time.Second))
+	dp.metrics.pendingDeleteRequestsCount.Set(float64(pendingDeleteRequestsCount))
+
```
pstibrany, line 477: This will only update metrics after all delete requests have plans for them built. Is that OK?

sandeepsukhani: Yes, that is the only option right now because if we change the count as we go through each delete request the partially updated value could get scraped and we would see unexpected jumps in the number of pending requests.

pkg/chunk/purger/purger.go
sandeepsukhani: sorry I missed that, fixed it.

pkg/chunk/purger/purger.go
sandeepsukhani: This could be done by refactoring the code to find the requests to process as well as do a count of pending requests and build the plans after updating the metric. What do you think?

pkg/chunk/purger/purger_test.go
```diff
@@ -377,6 +387,87 @@ func TestDataPurger_Restarts(t *testing.T) {
...
+
+	// load new delete requests for processing
+	require.NoError(t, purger.pullDeleteRequestsToPlanDeletes())
+
+	// there must be 2 pending delete requests, oldest being 3 days old
+	require.InDelta(t, float64(3*86400), testutil.ToFloat64(purger.metrics.oldestPendingDeleteRequestAgeSeconds), 1)
+	require.Equal(t, float64(2), testutil.ToFloat64(purger.metrics.pendingDeleteRequestsCount))
+
+	// wait until purger_delete_requests_processed_total starts to show up.
+	for {
```
pracucci, line 448: Instead of this `for` loops, could you use `test.Poll()`? The advantage of `test.Pool()` is that it has a timeout too.

_Same comment applies to the next `for`_

sandeepsukhani: nice, I didn't know it existed. Changed it, thanks!

PR #2732, opened 2020-06-16
 Add validation check for chunks and index in schema config 

### LINKED ISSUE #2731: Missing chunks and  index config causing silent failure
**If the following configuration is missing, Cortex fails silently without a proper explanation. 
In my opinion, this should be validated when Cortex loads the config file.**
For chunks
```
chunks:
      prefix: chunks_
      period: 1w 
```

For Index
```
index:
      prefix: index_
      period: 1w 
```

jaybatra26:
**What this PR does**:

This change would add a check for the presence of  chunks and index 
config in schema
Signed-off-by: Jay Batra jaybatra73@gmail.com

Fixes #2731 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


bboreham:
This relates to #2731, right?

jaybatra26:
@bboreham Yes, I realized it's a schema config check, so I have added a check in `pkg/chunk/schema_config.go`

bboreham:
Please re-read the issue description template and put the information in the places indicated.
In particular, the text "Fixes #2731" will tie the PR and issue together automatically.

pkg/chunk/schema_config.go
```diff
@@ -148,6 +151,16 @@ func (cfg *SchemaConfig) ForEachAfter(t model.Time, f func(config *PeriodConfig)
 func (cfg PeriodConfig) CreateSchema() (BaseSchema, error) {
 	buckets, bucketsPeriod := cfg.createBucketsFunc()
 
+	// Ensure Index table prefix exists
+	if cfg.IndexTables.Prefix == "" {
+		return nil, errConfigIndexNotSet
+	}
+
+	//Ensure Chunks table prefix exists
+	if cfg.ChunkTables.Prefix == "" {
```
pracucci, line 160: I think this is not correct. Even if obscure, as of today having an empty prefix for the chunk tables means that the table manager should not create such tables ([see code here](https://github.com/cortexproject/cortex/blob/311ab3eb147773ba52a07150cdf5511a3797e06b/pkg/chunk/table_manager.go#L415-L419)). This is a valid setting used by people storing chunks on a object storage.

pkg/chunk/schema_config.go
```diff
@@ -28,6 +28,8 @@ var (
 	errInvalidSchemaVersion = errors.New("invalid schema version")
 	errInvalidTablePeriod   = errors.New("the table period must be a multiple of 24h (1h for schema v1)")
 	errConfigFileNotSet     = errors.New("schema config file needs to be set")
+	errConfigChunkNotSet    = errors.New("schema config for chunks is missing")
+	errConfigIndexNotSet    = errors.New("schema config for index is missing")
```
pracucci, line 32: I would be more clear.

```suggestion
	errConfigIndexPrefixNotSet    = errors.New("schema config for index is missing the 'prefix' setting")
```

pkg/chunk/schema_config.go
```diff
@@ -109,6 +111,7 @@ func (cfg *SchemaConfig) loadFromFile() error {
 // Validate the schema config and returns an error if the validation
 // doesn't pass
 func (cfg *SchemaConfig) Validate() error {
+
```
pracucci, line 114: This is nit. Generally speaking, we try to preserve the rest of the codebase as is, without adding non-functional changes to blocks of that haven't changed.

pkg/chunk/schema_config.go
```diff
@@ -148,6 +151,16 @@ func (cfg *SchemaConfig) ForEachAfter(t model.Time, f func(config *PeriodConfig)
 func (cfg PeriodConfig) CreateSchema() (BaseSchema, error) {
 	buckets, bucketsPeriod := cfg.createBucketsFunc()
 
+	// Ensure Index table prefix exists
```
pracucci, line 154: We have `periodCfg.validate()` that should have the validation logic for the `PeriodConfig`. Is there any blocker to move this logic there?

pkg/chunk/schema_config.go
jaybatra26: @pracucci  I agree, I learned it the hard way. My configuration was missing `chunks` section and all I was getting was `Table name empty error` which was most likely coming from table manager. After comparing  config file, provided by @bboreham I realized that I was missing this piece of information in my config file. So, I thought there should be a check for that as well.

pkg/chunk/schema_config.go
jaybatra26: I will keep in mind about preserving the rest of the codebase.

pkg/chunk/schema_config.go
jaybatra26: The `periodCfg.validate()` ultimately calls `[createSchema](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/schema_config.go#L148)` function which validates the Period config for index and chunk. It validates by creating schema and fails if it fails. That's why I have put the check statement here.

pkg/chunk/schema_config.go
bboreham: So you need to narrow the check. There should only be a check if the chunk store intends to use tables.
Chunks can be stored in non-table store, like S3 or Swift.
This may simplify the changes to tests that you made; I don't think so many should be necessary.

jaybatra26: Ohh! Okay.

jaybatra26: ohh! okay. Looking deeper into the code.

CHANGELOG.md
```diff
@@ -926,6 +926,7 @@ Please be aware that Cortex `0.7.0` introduces some **breaking changes**. You're
 * [BUGFIX] Fixed leaked goroutines in the querier. #2070
 * [BUGFIX] Experimental TSDB: fixed `/all_user_stats` and `/api/prom/user_stats` endpoints when using the experimental TSDB blocks storage. #2042
 * [BUGFIX] Experimental TSDB: fixed ruler to correctly work with the experimental TSDB blocks storage. #2101
+ [BUGFIX] Fixed `Missing chunks and index config causing silent failure` Absence of chunks and index from schema config is not validated. #2731
```
bboreham, line 929: Please move this to the top of the file.

pkg/chunk/schema_config.go
```diff
@@ -144,6 +145,22 @@ func (cfg *SchemaConfig) ForEachAfter(t model.Time, f func(config *PeriodConfig)
 	}
 }
 
+func validateChunks(cfg PeriodConfig) error {
```
pracucci, line 148: I still believe this logic should belong to `PeriodConfig.Validate()`. Currently it does:

```
func (cfg PeriodConfig) validate() error {
	_, err := cfg.CreateSchema()
	return err
}
```

What you could do is:

```
func (cfg PeriodConfig) validate() error {
	if _, err := cfg.CreateSchema(); err != nil {
	    return err
       }

      // Then here you move the extra validation you're currently doing in validateChunks()
}
```

pkg/chunk/schema_config.go
```diff
@@ -144,6 +145,22 @@ func (cfg *SchemaConfig) ForEachAfter(t model.Time, f func(config *PeriodConfig)
...
+	objectStore := cfg.IndexType
+	if cfg.ObjectType != "" {
+		objectStore = cfg.ObjectType
+	}
+	switch objectStore {
+	case "cassandra", "aws-dynamo", "bigtable-hashed", "gcp", "gcp-columnkey", "bigtable", "boltdb", "grpc-store":
```
pracucci, line 154: `"boltdb"`  is not an option for the chunk store. Should be removed from this.

CHANGELOG.md
jaybatra26: @bboreham Sure! Thanks

pkg/chunk/schema_config.go
jaybatra26: @pracucci Thanks for the review. I will move this.

jaybatra26:
Hi @pracucci I have update the PR, I hope this is now mergeable :)
@Bryan Please take a look

jaybatra26:
@bboreham Please take a look.

CHANGELOG.md
```diff
@@ -946,6 +946,7 @@ Please be aware that Cortex `0.7.0` introduces some **breaking changes**. You're
 * [ENHANCEMENT] Experimental TSDB: Added support for Azure Storage to be used for block storage, in addition to S3 and GCS. #2083
 * [ENHANCEMENT] Experimental TSDB: Reduced memory allocations in the ingesters when using the experimental blocks storage. #2057
 * [ENHANCEMENT] Experimental Memberlist KV: expose `-memberlist.gossip-to-dead-nodes-time` and `-memberlist.dead-node-reclaim-time` options to control how memberlist library handles dead nodes and name reuse. #2131
+* [BUGFIX] Fixed `Missing chunks and index config causing silent failure` Absence of chunks and index from schema config is not validated. #2731
```
pracucci, line 949: Please move this above, under the "master / unreleased" section.

pracucci, line 949:
```suggestion
* [BUGFIX] Fixed `Missing chunks and index config causing silent failure` Absence of chunks and index from schema config is not validated. #2732
```

pkg/chunk/schema_config.go
pracucci: Could you remove `boltdb`, please?

CHANGELOG.md
jaybatra26: Done

pkg/chunk/schema_config.go
jaybatra26: My apologies, I removed the tests for BoltDb but forgot to remove it from the check list.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
   * `s3.http.insecure-skip-verify`
 * [ENHANCEMENT] Prometheus upgraded. #2798 #2849
   * Optimized labels regex matchers for patterns containing literals (eg. `foo.*`, `.*foo`, `.*foo.*`)
+* [BUGFIX] Fixed `Missing chunks and index config causing silent failure` Absence of chunks and index from schema config is not validated. #2731
```
pracucci, line 52: We're used to reference the PR number instead of the issue number:

```suggestion
* [BUGFIX] Fixed `Missing chunks and index config causing silent failure` Absence of chunks and index from schema config is not validated. #2732
```

jaybatra26:
@pracucci and @bboreham  Done.

CHANGELOG.md
jaybatra26: Done :)

pracucci:
@bboreham Do you have any other feedback on this PR?

PR #2733, opened 2020-06-16
Query frontend readiness

joe-elliott:
**What this PR does**:
- Adjusts the behavior of the frontend readiness handler at `/ready` to return 200 only if the frontend is ready to receive requests
  - Note that this also impacts the single binary
- Adds appropriate unit/integration tests
- Adds documentation on scaling the query frontend
- Minor updates/fixes to scalable frontend proposal

Fixes #None

Addresses this accepted proposal: https://github.com/cortexproject/cortex/blob/master/docs/proposals/scalable-query-frontend.md#querier-discovery-lag

**Checklist**
- [x] Tests updated
- [x] Documentation added


pstibrany:
All Cortex componets have /ready handler. I don’t quite understand why does frontend need its own. Would it make sense to extend /ready handler with frontend check, similar to ingester?

joe-elliott:
The current readiness handler returns 200 if all of the Cortex modules have successfully started and are running.  This is normally fine, but in the case of the query frontend it is not actually ready to receive requests until a querier has established a GRPC connection to it.  Depending on settings this could be 10s of seconds.

In discussions surrounding the original proposal it was decided an additional handler was more appropriate then modifying the behavior of the original, but we can revisit this conversation.

pstibrany:
> In discussions surrounding the original proposal it was decided an additional handler was more appropriate then modifying the behavior of the original, but we can revisit this conversation.

By using separate handler, we risk sending requests to Cortex instance which already has querier connected to it, but is otherwise still initializing. If initialization of other modules fails and Cortex exits, queued requests will be dropped.

Benefit of using separate handler is that query-frontend may start serving requests faster, while e.g. WAL is being replayed or blocks being downloaded (both of which can take minutes).

I am not sure what is better. I will review previous discussion to educate myself.

gouthamve:
I would be in favour of re-using the existing handler and adding the check to it.

> Benefit of using separate handler is that query-frontend may start serving requests faster, while e.g. WAL is being replayed or blocks being downloaded (both of which can take minutes).

If this is the case for single-binary, I am not sure how much benefit we'll have by getting the frontend ready but not the other components.

joe-elliott:
> Thanks @joe-elliott for working on it! LGTM. I just have few little requests:

Done!

> If this is the case for single-binary, I am not sure how much benefit we'll have by getting the frontend ready but not the other components.

In the single binary scenario the current readiness endpoint works well for insertion into a gossip ring or readiness for ingestion, but not for reads.  The idea behind this endpoint is that a query path load balancer or k8s service could use this endpoint to delay querying the new shard until the query-frontend has attached queriers.

pstibrany:
> In the single binary scenario the current readiness endpoint works well for insertion into a gossip ring or readiness for ingestion, but not for reads. The idea behind this endpoint is that a query path load balancer or k8s service could use this endpoint to delay querying the new shard until the query-frontend has attached queriers.

Discussion is whether to extend existing `/ready` handler with check for query-frontend, or create new handler. I'm not convinced that new handler is necessary, but don't have strong opinion on it.

pstibrany:
Also in single-binary mode, I assume that querier running in the same binary "connects" to the query-frontend quickly.

joe-elliott:
> Also in single-binary mode, I assume that querier running in the same binary "connects" to the query-frontend quickly.

You could configure the querier to use localhost.  In this case each query-frontend would have a single querier attached.  

However, I believe that the recommended configuration would be exactly as we have it now:  The querier component of each single binary will be configured to use a DNS entry for discovery of the frontend component of all the other single binaries.  In this case the connection lag would be no different than microservices mode.

docs/apis.md
```diff
@@ -376,4 +376,14 @@ Note that setting a new config will effectively "re-enable" the Rules and Alertm
...
+`POST /push` - Push samples directly to ingesters.  Accepts requests in Prometheus remote write format.  Indended for performance testing and debugging.
+
+## Query Frontend
+
+### Readiness
+
+This special readiness handler is designed to indicate when the frontend component is ready to receive queries.  This differs depending on configuration but is usually the time it takes for a querier to discover the frontend and attach to it.
+
+```
+GET /query-frontend/readiness
```
pracucci, line 388:
```suggestion
GET /query-frontend/ready
```

joe-elliott: sry, silly mistake.  good eyes!

pkg/api/api.go
```diff
@@ -373,6 +373,9 @@ func (a *API) registerQueryAPI(handler http.Handler) {
 func (a *API) RegisterQueryFrontend(f *frontend.Frontend) {
 	frontend.RegisterFrontendServer(a.server.GRPC, f)
 	a.registerQueryAPI(f.Handler())
+
+	// Readiness
+	a.RegisterRoute("/query-frontend/ready", http.HandlerFunc(f.ReadinessHandler), false, "GET")
```
pracucci, line 378: Should we keep this or remove it? I'm lost now ;)

CHANGELOG.md
```diff
@@ -115,8 +115,10 @@
 * [ENHANCEMENT] Use separate Cassandra clients and connections for reads and writes. #2666
 * [ENHANCEMENT] Add `-cassandra.reconnect-interval` to allow specifying the reconnect interval to a Cassandra server that has been marked `DOWN` by the gocql driver. Also change the default value of the reconnect interval from `60s` to `1s`. #2687
 * [ENHANCEMENT] Experimental TSDB: Applied a jitter to the period bucket scans in order to better distribute bucket operations over the time and increase the probability of hitting the shared cache (if configured). #2693
+
```
pracucci, line 118:
```suggestion
```

pkg/api/api.go
joe-elliott: Yes, this will be removed.  Apologies!  I'm getting back to this this week.  I'm going to make this a Draft PR until I feel like it's ready again.

pracucci: No need to apologies! 🤝

docs/operations/scalable-query-frontend.md
```diff
@@ -0,0 +1,29 @@
...
+linkTitle: "Scaling the Query Frontend"
+weight: 5
+slug: scaling-query-frontend
+---
+
+Historically scaling the Cortex query frontend has [posed some challenges](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/).  This document aims to detail how to use some of the added configuration parameters to correctly scale the frontend.  Note that these instructions apply in both the HA single binary scenario or microservices mode.
+
+## DNS Configuration / Readiness
+
+When a new frontend is first created on scale up it will not immediately have queriers attached to it.  The existing endpoint `/ready` was updated to only return http 200 when the query frontend was ready to serve queries.  Make sure to configure this endpoint as a healthcheck in your load balancer.  Otherwise a query frontend scale up event might result in failed queries for a bit while queriers attach.
```
pracucci, line 12:
```suggestion
When a new frontend is first created on scale up it will not immediately have queriers attached to it.  The existing endpoint `/ready` was updated to only return http 200 when the query frontend was ready to serve queries.  Make sure to configure this endpoint as a healthcheck in your load balancer.  Otherwise a query frontend scale up event might result in failed queries or high latency for a bit while queriers attach.
```

docs/operations/scalable-query-frontend.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Querier Max Concurrency
+
+Make sure to configure the querier frontend worker to match max concurrency.  This will allow the operator to freely scale the frontend up and down without impacting the amount of work an individual querier is attempting to perform.  More details [here](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/#dynamic-querier-concurrency).
+
+### Example Configuration
+
+**CLI**
+```
+-querier.worker-match-max-concurrent
```
pracucci, line 22:
```suggestion
-querier.worker-match-max-concurrent=true
```

pkg/querier/frontend/frontend.go
```diff
@@ -426,3 +432,34 @@ FindQueue:
 	// and wait for more requests.
 	goto FindQueue
 }
+
+// ReadinessHandler is a HandlerFunc that is designed to indicate if this frontend is ready to
+//  receive requests.  It will return a positive status code (200) if there are any attached queriers.
+func (f *Frontend) ReadinessHandler(w http.ResponseWriter, r *http.Request) {
```
pracucci, line 438: No one calls it.

integration/query_frontend_test.go
```diff
@@ -137,6 +138,11 @@ func runQueryFrontendTest(t *testing.T, testMissingMetricName bool, setup queryF
 	require.NoError(t, queryFrontend.WaitSumMetrics(e2e.Equals(1), "cortex_memcache_client_servers"))
 	require.NoError(t, queryFrontend.WaitSumMetrics(e2e.Greater(0), "cortex_dns_lookups_total"))
 
+	// Confirm that the query-frontend is not ready b/c the querier is no connected
+	resp, err := http.Get("http://" + queryFrontend.HTTPEndpoint() + "/ready")
```
pracucci, line 142: The previous call `s.StartAndWaitReady(queryFrontend, distributor, ingester)` waits until the readiness probe succeed, so this test is not expected to work.

joe-elliott:
Ok, I think these are in a good place now.  As discussed we settled on modifying the behavior of the `/ready` endpoint.

I will also be submitting a PR to this repo: https://github.com/grafana/cortex-jsonnet to handle the k8s configuration changes.  Deploying these changes will need to be coordinated.

pkg/querier/frontend/frontend.go
```diff
@@ -12,6 +13,7 @@ import (
 	"path"
 	"strings"
 	"sync"
+	"sync/atomic"
```
pstibrany, line 16: Let's use `connectedClients *atomic.Int32` instead. It makes it clear that this is accessed atomically, and it also avoids possible alignment issues (although that's not relevant for 32-bit values I think).

pkg/querier/frontend/frontend_test.go
```diff
@@ -161,7 +162,43 @@ func TestFrontendCancelStatusCode(t *testing.T) {
...
+			require.Equal(t, tt.msg, errMsg)
+		})
+	}
+}
+
 func testFrontend(t *testing.T, handler http.Handler, test func(addr string), matchMaxConcurrency bool) {
+	testFrontendRunWorker(t, handler, test, matchMaxConcurrency, true)
+}
+
+func testFrontendRunWorker(t *testing.T, handler http.Handler, test func(addr string), matchMaxConcurrency bool, runWorker bool) {
```
pstibrany, line 201: Can we inline this? runWorker is always true, should be removed?

pkg/querier/frontend/frontend.go
```diff
@@ -426,3 +433,23 @@ FindQueue:
...
+		return nil
+	}
+
+	// if we have more than one querier connected we will consider ourselves ready
+	connectedClients := atomic.LoadInt32(&f.connectedClients)
+	if connectedClients > 0 {
+		return nil
+	}
+
+	msg := fmt.Sprintf("not ready: connected clients %d", connectedClients)
```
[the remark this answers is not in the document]
joe-elliott: right, i included this detail in case we were unexpected receiving service unavailable.  i don't believe it's possible for connectedClients to be negative, but you know how that goes in complex systems :)

pkg/querier/frontend/frontend_test.go
joe-elliott: Good catch.  Was using both methods in a previous iteration of the PR.  removing.

pkg/querier/frontend/frontend.go
joe-elliott: good call.  updating.

pkg/querier/frontend/frontend.go
```diff
@@ -426,3 +434,23 @@ FindQueue:
...
+		return nil
+	}
+
+	// if we have more than one querier connected we will consider ourselves ready
+	connectedClients := f.connectedClients.Load()
+	if connectedClients > 0 {
+		return nil
+	}
+
+	msg := fmt.Sprintf("not ready: connected clients %d", connectedClients)
```
pracucci, line 453: I would be more specific:
```suggestion
	msg := fmt.Sprintf("not ready: number of queriers connected to query-frontend is %d", connectedClients)
```

pkg/querier/frontend/frontend.go
```diff
@@ -426,3 +434,23 @@ FindQueue:
...
+	level.Warn(f.log).Log("msg", msg)
```
pracucci, line 454: Warn or Info? I would prefer to avoid having warnings logged on each startup for a legit case (considering we'll have health checks done before queriers will be attached).

joe-elliott: agreed.

PR #2735, opened 2020-06-17
Add Schema and Storage config for DynamoDb in doc.

jaybatra26:
This PR adds sample schema and storage configuration
to make DynamoDb work locally.

Signed-off-by: Jay Batra <jaybatra73@gmail.com>

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/production/storage-aws.md
+++ b/docs/production/storage-aws.md
@@ -89,3 +89,34 @@ Several things to note here:
 -  If you want to add AWS tags to the created DynamoDB tables you
    can do it by adding a `tags` map to your schema definition. See
    [`schema configuration`](../configuration/schema-config-reference.md)
+
+## Running DynamoD in local
+
+For development, DynamoDb can be used locally. For this it's docker image can be used which is available on [`docker hub`](https://hub.docker.com/r/amazon/dynamodb-local/).
+
+Make sure the schema config is set to:
+
+```
+schema:
+  configs:
+  - from: <yyyy-mm-dd>
+    store: aws-dynamo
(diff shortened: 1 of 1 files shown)


docs/production/storage-aws.md
```diff
@@ -89,3 +89,34 @@ Several things to note here:
 -  If you want to add AWS tags to the created DynamoDB tables you
    can do it by adding a `tags` map to your schema definition. See
    [`schema configuration`](../configuration/schema-config-reference.md)
+
+## Running DynamoD in local
```
pracucci, line 93: This file covers a Cortex production setup on AWS. In a production setup we shouldn't mention the local DynamoDB application. I'm not 100% sure where this documentation should go: ideally we should provide an easy way to startup a local environment instead of documenting it. For example, if you look at the `development/` you can find a couple of docker-compose config to run a local env (running the blocks storage) with a single command. Maybe we can do something similar for the chunks storage?

jaybatra26: @pracucci  Docker compose sounds good, maybe I can try setting up all the three services(prometheus, cortex and dynamoDb) with docker compose). Also, I think the errors that I get are mostly related to validation check which I am covering in this PR. https://github.com/cortexproject/cortex/pull/2732

pracucci: Hi @jaybatra26. Have you had a chance to look at it?

pracucci:
@jaybatra26 Are you still interested into improving this doc?

PR #2739, opened 2020-06-17
Fix website

gouthamve:
I broke website when I upgraded hugo, config change makes everything shiny again :)

Also see google/docsy#174 and https://github.com/gohugoio/hugo/releases/tag/v0.60.0


gouthamve:
Yes it does!

PR #2741, opened 2020-06-18
 Propgate GOPROXY value to build-image/Dockerfile

### LINKED ISSUE #2740: Propagate GOPROXY value to build-image docker build
My company's network disallow connection to the default Go Proxy; I had to set GOPROXY to `direct`, but my configured value is not propagated when building `build-image`, which caused the build to fail.

alvinlin123:
**What this PR does**:

Propagate configured GOPROXY value to build-image/Dockerfile so that we can build the code inside network that disallow access to default Go proxy server.

Fixes #2740

**Checklist**
- [Not applicable ] Tests updated
   - I have tested the build still works on a vanilla Ubuntu box.
   - I have tested my change works behind my company's network.
- [ Not applicable] Documentation added
- [ Done] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
 * [FEATURE] The flag `-frontend.max-cache-freshness` is now supported within the limits overrides, to specify per-tenant max cache freshness values. The corresponding YAML config parameter has been changed from `results_cache.max_freshness` to `limits_config.max_cache_freshness`. The legacy YAML config parameter (`results_cache.max_freshness`) will continue to be supported till Cortex release `v1.4.0`. #2609
 * [FEATURE] Experimental gRPC Store: Added support to 3rd parties index and chunk stores using gRPC client/server plugin mechanism. #2220
-* [ENHANCEMENT] Propagate GOPROXY value when building `build-image`. This is to help the builders building the code in a Network where default Go proxy is not accessible (e.g. when behind some corporate VPN).
+* [ENHANCEMENT] Propagate GOPROXY value when building `build-image`. This is to help the builders building the code in a Network where default Go proxy is not accessible (e.g. when behind some corporate VPN). #2741
 * [ENHANCEMENT] Querier: Added metric `cortex_querier_request_duration_seconds` for all requests to the querier. #2708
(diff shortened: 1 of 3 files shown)


pstibrany:
Thank you. Alternatively, since we have updated build image with go1.14, we can also update go.mod file, which will then automatically use vendor directory instead of using modules. This is the direction we plan to go. Do you want to give it a try and update PR with this change?

alvinlin123:
HI @pstibrany thanks for the comment. I just tried to change go.mod to 1.14, but still got the GOPROXY problem. I guess this is because the problem is still there because we run some  `go get` commands inside `build-image\Dockerfile`, and `go.mod` is not used during building that Dockerfile? Do I have to update the Dockerfile too? Sorry I am new to the Go ecosystem :)

pstibrany:
I see, yes you're right that building the build-images needs to access the proxy as well. Propagating GOPROXY setting seems like best option then.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
 * [FEATURE] The flag `-frontend.max-cache-freshness` is now supported within the limits overrides, to specify per-tenant max cache freshness values. The corresponding YAML config parameter has been changed from `results_cache.max_freshness` to `limits_config.max_cache_freshness`. The legacy YAML config parameter (`results_cache.max_freshness`) will continue to be supported till Cortex release `v1.4.0`. #2609
 * [FEATURE] Experimental gRPC Store: Added support to 3rd parties index and chunk stores using gRPC client/server plugin mechanism. #2220
+* [ENHANCEMENT] Propagate GOPROXY value when building `build-image`. This is to help the builders building the code in a Network where default Go proxy is not accessible (e.g. when behind some corporate VPN).
```
pstibrany, line 56: Can you please add PR number here?

alvinlin123: Ah, missed that. Added.

alvinlin123:
What is the process of getting the PR merged? I don't believe I can, nor should merge it myself right?

pstibrany:
> What is the process of getting the PR merged? I don't believe I can, nor should merge it myself right?

No, you cannot merge it yourself. In Cortex, we have a rule that 2 maintainers need to approve the PR before merging. One more approval is needed, I'm sure someone will take a look at it very soon.

PR #2745, opened 2020-06-18
Give example of default label keys for HA pair

wardbekker:
**What this PR does**:

I found it confusing that the example doesn't use the default label names, and says the non-default (replica vs __replica__) is ideal. Made the documentation less confusing by giving example of the sensible default.

**Checklist**
- [ ] Tests updated
- [X] Documentation added


pstibrany:
Thanks and congrats to your first Cortex PR! 🎉

PR #2749, opened 2020-06-18
Set go version to 1.14, to enable vendor builds by default.

pstibrany:
**What this PR does**: After #2480, we can enable go 1.14 in go.mod. That enables -mod=vendor by default for all builds.

Requiring go 1.14 for builds also allows us to use new features from this version.

Pinging Loki team: @slim-bean @cyriltovena @owen-d.

**Checklist**
- [NA] Tests updated
- [NA] Documentation added

### DIFF
--- a/tools/test
+++ b/tools/test
@@ -4,7 +4,6 @@ set -e
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SLOW=
-NO_GO_GET=true
 TAGS=
 PARALLEL=
 RACE="-race -covermode=atomic"
@@ -12,7 +11,7 @@ TIMEOUT=5m
 VERBOSE=
 
 usage() {
-    echo "$0 [-slow] [-in-container foo] [-netgo] [-(no-)go-get] [-timeout 1m]"
+    echo "$0 [-slow] [-in-container foo] [-netgo] [-timeout 1m]"
 }
 
 while [ $# -gt 0 ]; do
@@ -27,14 +26,6 @@ while [ $# -gt 0 ]; do
             ;;
         "-no-race")
             RACE=
-            shift 1
-            ;;
-        "-no-go-get")
-            NO_GO_GET=true
-            shift 1
-            ;;
-        "-go-get")
-            NO_GO_GET=
             shift 1
             ;;
         "-netgo")
@@ -102,10 +93,6 @@ PACKAGE_BASE=$(go list -e ./)
 
 run_test() {
     local dir=$1
-    if [ -z "$NO_GO_GET" ]; then
-        go get -t -tags "${TAGS[@]}" "$dir"
-        go mod vendor # re-vendor everything
-    fi
 
     local GO_TEST_ARGS_RUN=("${GO_TEST_ARGS[@]}")
     if [ -n "$SLOW" ]; then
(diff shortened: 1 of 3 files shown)


pracucci:
Minimum version in Thanos is still 1.13 (even if built with 1.14). Could this change be a problem? @bwplotka

pstibrany:
> Minimum version in Thanos is still 1.13 (even if built with 1.14). Could this change be a problem? @bwplotka

This should not be a problem for dependencies. This sets expected language version for the module, but go 1.14 doesn’t change the language. It changes how vendoring works, but that is no concern when used as a dependency. Usage of some go new methods in 1.14 (eg. testing.T.Cleanup) is also not a problem, unless used in integration tests.

tools/test
```diff
@@ -104,6 +104,7 @@ run_test() {
     local dir=$1
     if [ -z "$NO_GO_GET" ]; then
         go get -t -tags "${TAGS[@]}" "$dir"
+        go mod vendor # re-vendor everything
```
pracucci, line 107: Could you explain why we need to run `go mod vendor` in this file?

pstibrany: Reason is to avoid error when go get fetches additional depedendency, but doesn’t put it into vendor. In go 1.14, if you have vendor directory, it is used by default, but vendor/modules.txt is verified against go.mod. Without calling go mod vendor here, that verification fails.

pstibrany: I would prefer if our test didn’t do additional go get, but I’m not sure why it’s here, so left it for now.
In case of cover tool later in the file, I think that should be part of the build image instead.

pracucci: What if we remove the `-go-get` and `-no-go-get` support at all, always enforcing the behaviour of `NO_GO_GET=true` (current default)? `-go-get` is not used by any script we have, so the only way to use it is to manually call `tools/test -go-get`.

pstibrany: > What if we remove the `-go-get` and `-no-go-get` support at all, always enforcing the behaviour of `NO_GO_GET=true` (current default)? `-go-get` is not used by any script we have, so the only way to use it is to manually call `tools/test -go-get`.

I will remove it. This doesn't fix problem with `cover` tool, used when run under CIRCLECI.

bwplotka:
https://github.com/cortexproject/cortex/pull/2749#issuecomment-646634077

Thanos is basing on 1.14 as minimum, we will fix the docs. So all good ! (:

pracucci:
_If we merge it soon we'll get it into next Cortex release._

tools/test
```diff
@@ -147,6 +135,8 @@ fi
 
 if [ -n "$SLOW" ] && [ -z "$COVERDIR" ]; then
     go get github.com/weaveworks/tools/cover
+    go mod vendor # re-vendor everything
```
bboreham, line 138: What about putting the `cover` tool into the build image, like we do with various other tools?
https://github.com/cortexproject/cortex/blob/b64dabe749023da882c9739b216de683194b9bfb/build-image/Dockerfile#L19-L23

pstibrany:
Yes, I think that makes sense, and would like to do that.

tools/test
pstibrany: I will send PR to include cover tool in build image.

Issue #2596, opened 2020-05-15
Make it possible to mark module as "visible" for users


pstibrany:
Not all modules should be selectable as "targets", only user-visible ones. For example, selecting "api" or "memberlist-kv" as targets makes no sense (and actually selecting "memberlist-kv" leads to panic, since this module doesn't depend on anything, but Cortex.Run method still tries to register some routes to API -- which is not initialized).

Once we have ability to set some modules as user-visible, `InitModuleServices` should report available modules if incorrect module name is passed to it.

/cc @annanay25

pstibrany:
Implemented in https://github.com/cortexproject/cortex/pull/2752.

PR #2752, opened 2020-06-19
Dynamically generate available options for -target commandline arg

### LINKED ISSUE #2710: Document for the `-target` parameter does not align with the ALL target defined in module manager
Currently the command line document for `-target` it says applicable values are `all, distributor, ingester, querier, query-frontend, table-manager, ruler, alertmanager, configs`. 

However, reading [coretex/modules.go](https://github.com/cortexproject/cortex/blob/master/pkg/cortex/modules.go#L554) it looks like there are more possible values for `-target`; take `data-purger` for example. 

I am wondering if this intentionally so, or it was just the document was not updated after adding new module? If it's the later case, I would love to work on this and make the document dynamic, so we don't have to remember to update the document when adding new module.

alvinlin123:
**What this PR does**:

Add missing options to the command line help text for `-target`. The options are dynamically generated so we don't have to manually modify the text when we add new module in the future. 

**Testing**:
```
➜  cortex git:(issue-2710-commandline-doc) go run cmd/cortex/main.go -modules
alertmanager
all
compactor
configs
data-purger
distributor
flusher
ingester
querier
query-frontend
ruler
store-gateway
table-manager
exit status 2

➜  cortex git:(issue-2710-commandline-doc) go run cmd/cortex/main.go -target api
...
level=warn ts=2020-06-23T08:25:27.092681Z caller=cortex.go:298 msg="'api' is not a public module, is this intended?" public-modules="alertmanager, all, compactor, configs, data-purger, distributor, flusher, ingester, querier, query-frontend, ruler, store-gateway, table-manager"
```

Fixes #2710, #2596

**Checklist**
- [x] Tests updated
- [x] Documentation added


pstibrany:
Thank you for your work on this. Unfortunately this goes against the idea why module manager was introduced in the first place: to allow projects that extend Cortex to dynamically add new modules (and set new depedendies). In this light, it’s not easily possible to enumerate all modules before they are actually added to the module manager.

Given that, one way to make this work may be to define special target like “modules” that would simply list all modules. (Or only modules that should be public in the first place, which can be defined by a new flag passed to RegisterModule function). It’s not as elegant as having it in the -help output, but should work.

pstibrany:
Related: https://github.com/cortexproject/cortex/issues/2596

alvinlin123:
Hi @pstibrany  thank you for your comment. Can you elaborate more on why "this goes against the idea why module manager was introduced in the first place: to allow projects that extend Cortex to dynamically add new modules"? 

I checked [the commit that added module manager](https://github.com/cortexproject/cortex/pull/2559) and it seems like to "allow other projects to dynamically add new modules" was achieved by adding the module manager. My PR doesn't change anything regarding interaction with the module manager. My idea was, instead of using constant strings to describe a module, use a struct so that we can mark a module as public. The logic of registering and adding dependencies still remain the same. so even with this PR, other packages should still be able to dynamically add module through module manager. 

I guess if there are other packages depending on the constant strings of module names then this PR would be problematic.

pstibrany:
> Hi @pstibrany thank you for your comment. Can you elaborate more on why "this goes against the idea why module manager was introduced in the first place: to allow projects that extend Cortex to dynamically add new modules"?

There is at least one (not open source) project extending Cortex by reusing the main Cortex structure, its flags and registering additional modules. I would like help for -target option to reflect such added modules. Since they are added to the module manager only *after* help string for this option has been constructed, I think the best way to do that is by saying “To get list of available modules, use -target=modules”, and to introduce such “modules” module that would only print list of registered user-visible modules, and then stop.

> 
> I checked [the commit that added module manager](https://github.com/cortexproject/cortex/pull/2559) and it seems like to "allow other projects to dynamically add new modules" was achieved by adding the module manager. My PR doesn't change anything regarding interaction with the module manager. My idea was, instead of using constant strings to describe a module, use a struct so that we can mark a module as public. The logic of registering and adding dependencies still remain the same. so even with this PR, other packages should still be able to dynamically add module through module manager.

I don’t think struct you have introduced is necessary, and I would like to avoid changing module type from string. 
Instead, I suggest that we add flag to module manager register method to mark module as “public” (visible to user). To do that, I suggest that we change RegisterModule method to take variable number of flags of type “ModuleOption”, and implement one such option for this.

> 
> I guess if there are other packages depending on the constant strings of module names then this PR would be problematic.

Correct, it is a breaking change. We could do that if needed, but I don’t think we are there yet.

alvinlin123:
@pstibrany  I see, thanks for the explanation, I get the point now. Let me ponder a bit, and I will update this PR. 

Thanks again!

alvinlin123:
I have push new changes that display list of possible target values as well as throw error if given non-public target. 

However, I didn't go with the `-target module` approach, it felt akward that you can use "module" in yaml config. Also, all the modules needs to depend on API, which is really unnecessary for just printing list of public modules. 

So, I decided to introduce a `-modules` command line only flag: 
```bash
➜  cortex git:(issue-2710-commandline-doc) go run cmd/cortex/main.go -modules
alertmanager
data-purger
flusher
query-frontend
all
distributor
store-gateway
configs
ingester
compactor
table-manager
ruler
querier
exit status 2
```

Below is an example when user pass in non-public target (I list out the possible values in the error message): 

```bash
➜  cortex git:(issue-2710-commandline-doc) go run cmd/cortex/main.go -target ring
level=info ts=2020-06-20T06:42:36.235822Z caller=main.go:140 msg="Starting Cortex" version="(version=, branch=, revision=)"
level=error ts=2020-06-20T06:42:36.235893Z caller=log.go:140 msg="error running cortex" err="invalid target: ring; valid targets are: store-gateway, data-purger, compactor, flusher, query-frontend, ruler, configs, ingester, table-manager, distributor, alertmanager, all, querier"
exit status 1
```

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
 * [FEATURE] The flag `-frontend.max-cache-freshness` is now supported within the limits overrides, to specify per-tenant max cache freshness values. The corresponding YAML config parameter has been changed from `results_cache.max_freshness` to `limits_config.max_cache_freshness`. The legacy YAML config parameter (`results_cache.max_freshness`) will continue to be supported till Cortex release `v1.4.0`. #2609
 * [FEATURE] Experimental gRPC Store: Added support to 3rd parties index and chunk stores using gRPC client/server plugin mechanism. #2220
+* [ENHANCEMENT] Add command line flag to list possible values for -target. Also, throw error if the given target is not public. #2710
```
pstibrany, line 56: I think "internal" modules should be selectable, on some occasions it is useful. We can perhaps just log the warning instead?

cmd/cortex/main.go
```diff
@@ -130,6 +130,13 @@ func main() {
 	t, err := cortex.New(cfg)
 	util.CheckFatal("initializing cortex", err)
 
+	if t.Cfg.ListModules {
```
pstibrany, line 133: Can we add test for this?

cmd/cortex/main.go
```diff
@@ -130,6 +130,13 @@ func main() {
...
+		for _, m := range t.ModuleManager.PublicModuleNames() {
+			fmt.Fprintln(flag.CommandLine.Output(), m)
```
pstibrany, line 135: This should go stdout. Also, can we make the output sorted?

pkg/util/modules/modules.go
```diff
@@ -23,19 +26,33 @@ type Manager struct {
 	modules map[string]*module
 }
 
+// ModuleOption contains options of a module.
+// public option indicates if a module can be use as top level module (i.e. pass in as command line argument).
+type ModuleOption struct {
```
pstibrany, line 31: Let's put all flags into `module` structure. Individual module options can modify `module` structure. (See my other comment about functional options).

pkg/util/modules/modules.go
```diff
@@ -57,6 +74,11 @@ func (m *Manager) InitModuleServices(target string) (map[string]services.Service
 	if _, ok := m.modules[target]; !ok {
 		return nil, fmt.Errorf("unrecognised module name: %s", target)
 	}
+
+	if !m.modules[target].option.Public {
```
pstibrany, line 78: I think these are two separate concerns: whether module is public (visible in -modules) and whether it is runnable (can be selected for -target). There can be modules that are runnable, but not public. (for testing, during development)

I suggest we remove this check for now. We can introduce second option to make module "runnable" later.

pkg/util/modules/modules.go
```diff
@@ -23,19 +26,33 @@ type Manager struct {
...
 // name must be unique to avoid overwriting modules
 // if initFn is nil, the module will not initialise
 func (m *Manager) RegisterModule(name string, initFn func() (services.Service, error)) {
+	m.RegisterModuleWithOption(name, initFn, defaultModuleOption())
+}
+
+// RegisterModule registers a new module with name, init function, and option
+// name must be unique to avoid overwriting modules
+// if initFn is nil, the module will not initialise
+func (m *Manager) RegisterModuleWithOption(name string, initFn func() (services.Service, error), option ModuleOption) {
```
pstibrany, line 52: I would prefer to have single `RegisterModule` method, with variable arguments with individual options.

We can use [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) pattern for this.

CHANGELOG.md
alvinlin123: > I think "internal" modules should be selectable, on some occasions it is useful. We can perhaps just log the warning instead?

Yea, if there are cases where we want to use internal modules, then we definitely should just log an error instead of throwing error. I will make this change. 

> I'd prefer to have single RegisterModule method with options, to reduce number of exported methods and structs. What do you think?

I would prefer this too, but I was worry about backward compatibility since RegisterModule is an exported method. Is there anyway to have an optional parameter in Go? Or should I not worry about backward compatibility?

cmd/cortex/main.go
alvinlin123: Ah, sorry, didn't realize default is stderr. Will make the change. Will sort the output too.

pkg/util/modules/modules.go
alvinlin123: Make sense. I will update :)

pkg/util/modules/modules.go
alvinlin123: I would prefer single `RegisterModule` too, I created `RegisterModuleWithOptions` just because Go doesn't allow method overloading (I came from Java :)) and I did not want to break backward compatibility. Thanks for pointing me to Functional options, I learned something new. 

But [this StackOverflow post](https://stackoverflow.com/questions/55163005/could-adding-variadic-parameters-to-a-function-break-existing-code) gave some examples where adding variadic parameter may break existing code; should we worry about this?

pstibrany: Issues pointed out by StackOverflow post are real, but not something to be worried about in this case.

alvinlin123:
> Let's put all flags into module structure. Individual module options can modify module structure. (See my other comment about functional options).

since the module struct is not exported, I can't move the module option to the struct because doing so the `cortex` package will not be able to define an option setting function with argument of type `*module`.

pstibrany:
> > Let's put all flags into module structure. Individual module options can modify module structure. (See my other comment about functional options).
> 
> since the module struct is not exported, I can't move the module option to the struct because doing so the `cortex` package will not be able to define an option setting function with argument of type `*module`.

All options passable to `RegisterModule` should be defined in the package with module manager. Other packages should not be able to define options.

pkg/util/modules/modules.go
```diff
@@ -23,19 +27,31 @@ type Manager struct {
 	modules map[string]*module
 }
 
+// PrivateModule is a functional option that can be used
+// when registering a module.
```
pstibrany, line 31:
```suggestion
// PrivateModule is an option for `RegisterModule` that marks module as private. Modules are public by default.
```

pkg/util/modules/modules_test.go
```diff
@@ -47,3 +47,109 @@ func TestDependencies(t *testing.T) {
...
+	sut := NewManager()
+	sut.RegisterModule("module1", mockInitFunc)
+
+	m := sut.modules["module1"]
+
+	assert.NotNil(t, mockInitFunc, m.initFn, "initFn not assigned")
+	assert.True(t, m.public, "module should be public")
+}
+
+func TestRegisterModuleSetsDefaultOption(t *testing.T) {
```
pstibrany, line 61: This test and `TestRegisterModuleWithOptions` are almost the same. We can remove one of them.

pkg/util/modules/modules_test.go
```diff
@@ -47,3 +47,109 @@ func TestDependencies(t *testing.T) {
...
+	sut.RegisterModule("public3", mockInitFunc)
+	sut.RegisterModule("private1", mockInitFunc, PrivateModule)
+	sut.RegisterModule("private2", mockInitFunc, PrivateModule)
+
+	pm := sut.PublicModuleNames()
+
+	assert.Len(t, pm, 3, "wrong result slice size")
+	assert.Contains(t, pm, "public1", "missing public module")
+	assert.Contains(t, pm, "public2", "missing public module")
+	assert.Contains(t, pm, "public3", "missing public module")
```
pstibrany, line 96: I would assert on entire slice at once, which would also check sorting. Something like:

```suggestion
      assert.Equals(t, []string{"public1", "public2", "public3"}, sut.PublicModuleNames())
```

Same in the next test.

pkg/util/modules/modules_test.go
```diff
@@ -47,3 +47,109 @@ func TestDependencies(t *testing.T) {
...
+	sut.RegisterModule("private2", mockInitFunc, PrivateModule)
+	sut.RegisterModule("private3", mockInitFunc, PrivateModule)
+	sut.RegisterModule("private4", mockInitFunc, PrivateModule)
+
+	pm := sut.PublicModuleNames()
+
+	assert.Len(t, pm, 0, "wrong result slice size")
+}
+
+func TestPublicModuleNamesReturnsSortedList(t *testing.T) {
```
pstibrany, line 128: If suggestion in `TestGetAllPublicModulesNames` is accepted, this test can be dropped.

pkg/util/modules/modules_test.go
```diff
@@ -47,3 +47,109 @@ func TestDependencies(t *testing.T) {
...
+	sut.RegisterModule("b", mockInitFunc)
+	sut.RegisterModule("a", mockInitFunc)
+
+	pm := sut.PublicModuleNames()
+
+	assert.Len(t, pm, 3, "wrong result slice size")
+	assert.Equal(t, []string{"a", "b", "c"}, pm, "module names list is not sorted in ascending order")
+}
+
+func TestIsPublicModule(t *testing.T) {
```
pstibrany, line 140: It feels like this could easily be merged into one of the previous tests (`TestRegisterModuleWithOptions` or `TestRegisterModuleSetsDefaultOption`). But that's just nit-picking.

pkg/util/modules/modules.go
```diff
@@ -23,19 +27,31 @@ type Manager struct {
...
+
 // NewManager creates a new Manager
 func NewManager() *Manager {
 	return &Manager{
 		modules: make(map[string]*module),
 	}
 }
 
-// RegisterModule registers a new module with name and init function
+// RegisterModule registers a new module with name, init function, and option functions.
```
pstibrany, line 43:
```suggestion
// RegisterModule registers a new module with name, init function, and options.
```

pkg/util/modules/modules.go
```diff
@@ -23,19 +27,31 @@ type Manager struct {
...
 // name must be unique to avoid overwriting modules
 // if initFn is nil, the module will not initialise
-func (m *Manager) RegisterModule(name string, initFn func() (services.Service, error)) {
+// if options is not given, then by default module is public
```
pstibrany, line 46: Can you please reformat this entire comment to have sentences starting with capital letters, and ending with period. Last sentence can be just: "Modules are public by default".

pkg/util/modules/modules_test.go
alvinlin123: I would think that having separate test for `IsPublicModule` is better because it adheres to single responsibility principle more. Also, when this test fails, it's much easer to tell which method is at fault.

pkg/util/modules/modules_test.go
alvinlin123: Thanks for catching this! I was too focused on refactoring and forgot to re-validate the tests.

pkg/util/modules/modules_test.go
pstibrany: When tests fail, we get a line number where they failed, so it's easy to find out. If you want to keep the test, I'm fine with that. It just felt like lot of repeated boilerplate to me.

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
 * [FEATURE] The flag `-frontend.max-cache-freshness` is now supported within the limits overrides, to specify per-tenant max cache freshness values. The corresponding YAML config parameter has been changed from `results_cache.max_freshness` to `limits_config.max_cache_freshness`. The legacy YAML config parameter (`results_cache.max_freshness`) will continue to be supported till Cortex release `v1.4.0`. #2609
 * [FEATURE] Experimental gRPC Store: Added support to 3rd parties index and chunk stores using gRPC client/server plugin mechanism. #2220
+* [ENHANCEMENT] Add command line flag to list possible values for -target. Also, log warning if given target is not public. #2710
```
pstibrany, line 59: We refer to PR numbers in Changelog entries. Let's also mention new command line flag (`-modules`).

```suggestion
* [ENHANCEMENT] Add `-modules` command line flag to list possible values for `-target`. Also, log warning if given target is internal component. #2752
```

pkg/cortex/modules.go
```diff
@@ -514,29 +514,28 @@ func (t *Cortex) setupModuleManager() error {
...
 	mm.RegisterModule(TableManager, t.initTableManager)
 	mm.RegisterModule(Ruler, t.initRuler)
 	mm.RegisterModule(Configs, t.initConfig)
 	mm.RegisterModule(AlertManager, t.initAlertManager)
-	mm.RegisterModule(Compactor, t.initCompactor)
-	mm.RegisterModule(StoreGateway, t.initStoreGateway)
-	mm.RegisterModule(Purger, t.initPurger)
+	mm.RegisterModule(Querier, t.initQuerier)
+	mm.RegisterModule(Ingester, t.initIngester)
+	mm.RegisterModule(Distributor, t.initDistributor)
```
pstibrany, line 537: It would be nice to minimize diff here to make changed parts visible. (Thanks for fixing double-registration of StoreGateway!)

alvinlin123:
> Thanks for your work on this PR!

Thanks for your patience and mentorship :)

pkg/util/modules/modules.go
```diff
@@ -23,19 +27,29 @@ type Manager struct {
 	modules map[string]*module
 }
 
+// PrivateModule is an option for `RegisterModule` that marks module as private. Modules are public by default.
+func PrivateModule(m *module) {
+	m.public = false
+}
```
pstibrany, line 33: `PrivateModule` vs `public` flag seems inconsistent. I would not use "private"/"public" terminology at all, just "internal" or "not internal": `InternalModule` option, `m.internal`, `IsInternalModule` method. WDYT?

It's a nit pick, and we can do this in separate PR if you prefer.

pkg/util/modules/modules.go
```diff
@@ -9,6 +9,16 @@ import (
...
+type moduleVisibility int
+
+const (
+	// A public module is intended to be passed into `InitModuleServices`.
+	public moduleVisibility = iota
+
+	// A private module is not intended to be passed into `InitModuleServices`.
+	private
+)
+
```
pstibrany, line 21: I don't see a need for this. Let's please use simple `bool`, like before. My suggestion was to modify terminology to move away from public/private to internal, since that's what these modules are.

pkg/util/modules/modules.go
alvinlin123: While trying to rename public -> internal, I got stuck on what should I change method name `PublicModuleNames` method to; `NotInternalModuleName` looks wrong, `ExternalModuleName` sounds ok, but we will have to have code like `! module.internal`, which is not much different than `func PrivateModule(m *module) {m.public = false}`. Then I realized the problem is probably not the terms public/private v.s. internal/not internal, the problem is using boolean to represent the public private concept. Seeing `public = false`, `internal = false`, and `!internal` incurs mental cost for the reader. 

So I came up with using the name "visibility" (I think I borrowed this term from OOP) and the value of it would be descriptive enum values. After changing the name I do feel the reading the code flows better. Let me know what you think, I am curious.

pkg/util/modules/modules.go
alvinlin123: oh, you commented faster than I replied :) 

In this case, what would you suggest for the name for `PublicModuleNames`?

pstibrany: Hmm, your other comment explains the thinking, and I can see why you went this route. Still, I feel like this goes too far for what is quite simple concept. Would `ModuleNames(skipInternal bool)` work for you?

alvinlin123: By the way, what if instead of `type moduleVisibility int` I use `type moduleVisibility bool`?

pstibrany: We can use term "visible", or "userVisible", but let's just use that as a bool field in module struct, without introducing new type.

pstibrany: `UserVisibleModuleNames()` is also pretty clear I think.

alvinlin123: > Hmm, your other comment explains the thinking, and I can see why you went this route. Still, I feel like this goes too far for what is quite simple concept. Would `ModuleNames(skipInternal bool)` work for you?

Don't you think it adds some mental power to the reader and the coder? When I read `ModuleNames(true)` it's very unclear to me what I get in return, I will probably have to have an alias variable so I can write `ModuleName(skipInternal)`, which makes the intend clearer, but I will have to define alias variable. For simple things like getting list of public modules, I feel a method with no arg is friendlier to the user.

alvinlin123: > We can use term "visible", or "userVisible", but let's just use that as a bool field in module struct, without introducing new type.

I think this will work, let me give it a try.

alvinlin123: Ok, done. Have a look and let me know. 

Please do let me know if you think it's still off; I happen think variable/method/class/whatever names are very important and I am more than welling to spend time to get the terms right to increase readability, so please don't run out of patience with me :)

pkg/cortex/cortex.go
```diff
@@ -291,6 +294,12 @@ func (t *Cortex) setupThanosTracing() {
 
 // Run starts Cortex running, and blocks until a Cortex stops.
 func (t *Cortex) Run() error {
+	if !t.ModuleManager.IsUserVisibleModule(t.Cfg.Target) {
+		level.Warn(util.Logger).Log(
+			"msg", fmt.Sprintf("'%v' is not a user visible module, is this intended?", t.Cfg.Target),
+			"user-visible-modules", strings.Join(t.ModuleManager.UserVisibleModuleNames(), ", "))
```
pstibrany, line 300: I would rephrase the message slightly, and drop listing of modules -- it is long and obscures main message.

```suggestion
		level.Warn(util.Logger).Log("msg", "selected target is an internal module, is this intended?", "target", t.Cfg.Target)
```

alvinlin123: Bike-shedding can be fun and educational when it does not involve money ;) 

Regarding the warning message, I can agree to that since whoever see the warning message either 1) knows what they are doing or 2) will check documentation to find the correct target name.

PR #2756, opened 2020-06-19
Upgrade prometheus to 2.19.1

pracucci:
**What this PR does**:
In this PR I'm upgrading Prometheus to `master` (and Thanos too) to primarily add two changes:
- Fix unsequential m-map files (https://github.com/prometheus/prometheus/pull/7414)
- Optimized bstream reader used by XORChunk iterator (https://github.com/prometheus/prometheus/pull/7390)

This is a **draft PR** because:
- Blocked by the Prometheus fix https://github.com/prometheus/prometheus/pull/7435

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/querier/lazyquery/lazyquery.go
```diff
@@ -48,22 +48,17 @@ func (l LazyQuerier) createSeriesSet(selectSorted bool, params *storage.SelectHi
...
-			future <- set
-		}
+		future <- l.next.Select(selectSorted, params, matchers...)
 	}()
 	return future
 }
 
 // Select implements Storage.Querier
-func (l LazyQuerier) Select(selectSorted bool, params *storage.SelectHints, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
+func (l LazyQuerier) Select(selectSorted bool, params *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
```
pstibrany, line 57: We can get rid of `createSeriesSet` now. It was introduced when there were Select and SelectSorted methods.

pkg/querier/querier.go
```diff
@@ -258,49 +258,38 @@ func (q querier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Mat
...
-		return nil, nil, promql.ErrStorage{Err: err}
+		return storage.ErrSeriesSet(promql.ErrStorage{Err: err})
 	}
 
 	if len(q.queriers) == 1 {
-		seriesSet, warning, err := q.queriers[0].Select(true, sp, matchers...)
-		if err != nil {
-			return nil, warning, err
-		}
+		seriesSet := q.queriers[0].Select(true, sp, matchers...)
```
pstibrany, line 270: This changes the logic slightly, but end result looks to be the same.

pracucci: I checked the code and practically there should be no difference. The `seriesSet.Err()` had to be handled even before.

pkg/querier/lazyquery/lazyquery.go
pracucci: Good point, done.

PR #2759, opened 2020-06-22
Add -querier.max-step-alignment to disable align querier with step if the step is greater than given value.

Wing924:
**What this PR does**:

Add `-querier.max-step-alignment` to disable align querier with step if the step is greater than given value.

If the step is very large (ex. 6h or 1d), the bar graph looks bad in Grafana because cortex changed the query start/end.
If the step is small enough, we can ignore it.

Without align:
スクリーンショット 2020-06-22 12 05 09

With align:
スクリーンショット 2020-06-22 12 06 14

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/querier/queryrange/step_align.go
```diff
@@ -6,17 +6,24 @@ import (
...
+}
 
 type stepAlign struct {
-	next Handler
+	next             Handler
+	maxStepAlignment int64
 }
 
 func (s stepAlign) Do(ctx context.Context, r Request) (Response, error) {
+	if s.maxStepAlignment > 0 && r.GetStep() > s.maxStepAlignment {
```
sandeepsukhani, line 24: The `maxStepAlignment` is configured in seconds while `GetStep` returns milliseconds which would make it not work as expected.

Wing924: I've fixed it

CHANGELOG.md
```diff
@@ -146,7 +147,7 @@
   * `cortex_compactor_block_cleanup_completed_total`
   * `cortex_compactor_block_cleanup_failed_total`
   * `cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds`
-* [ENHANCEMENT] Experimental TSDB: Use shared cache for metadata. This is especially useful when running multiple querier and store-gateway components to reduce number of object store API calls. #2626 #2640
+* [ENHANCEMENT] Experimental TSDB: Use shared cache for metadata. This is especially usHEADful when running multiple querier and store-gateway components to reduce number of object store API calls. #2626 #2640
```
owen-d, line 150: Small spelling mistake here.

PR #2760, opened 2020-06-22
make delete request cancellation duration configurable and rename data-purger to purger

sandeepsukhani:
**What this PR does**:
Duration until which delete requests are allowed to be cancelled after creation is made configurable. This would help with integration tests and run this initially with a higher value until we have more confidence in its stability.
This PR also renames flags and references from datapurger to purger for simplicity and consistency. 

**Checklist**
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
 * [CHANGE] Available command-line flags are printed to stdout, and only when requested via `-help`. Using invalid flag no longer causes printing of all available flags. #2691
 * [CHANGE] Experimental Memberlist ring: randomize gossip node names to avoid conflicts when running multiple clients on the same host, or reusing host names (eg. pods in statefulset). Node name randomization can be disabled by using `-memberlist.randomize-node-name=false`. #2715
 * [CHANGE] Memberlist KV client is no longer considered experimental. #2725
+* [CHANGE] Change target flag for purger from `data-purger` to `purger` and make delete request cancellation duration configurable. #2760
 * [FEATURE] TLS config options added for GRPC clients in Querier (Query-frontend client & Ingester client), Ruler, Store Gateway, as well as HTTP client in Config store client. #2502
 * [FEATURE] The flag `-frontend.max-cache-freshness` is now supported within the limits overrides, to specify per-tenant max cache freshness values. The corresponding YAML config parameter has been changed from `results_cache.max_freshness` to `limits_config.max_cache_freshness`. The legacy YAML config parameter (`results_cache.max_freshness`) will continue to be supported till Cortex release `v1.4.0`. #2609
 * [FEATURE] Experimental gRPC Store: Added support to 3rd parties index and chunk stores using gRPC client/server plugin mechanism. #2220
--- a/pkg/chunk/purger/purger.go
+++ b/pkg/chunk/purger/purger.go
@@ -26,10 +26,9 @@ import (
 )
 
 const (
-	millisecondPerDay                 = int64(24 * time.Hour / time.Millisecond)
-	deleteRequestCancellationDeadline = 24 * time.Hour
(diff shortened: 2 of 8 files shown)


pkg/chunk/purger/purger.go
```diff
@@ -83,18 +82,20 @@ type deleteRequestWithLogger struct {
...
-// Config holds config for DataPurger
+// Config holds config for Purger
 type Config struct {
-	Enable          bool   `yaml:"enable"`
-	NumWorkers      int    `yaml:"num_workers"`
-	ObjectStoreType string `yaml:"object_store_type"`
+	Enable                              bool          `yaml:"enable"`
+	NumWorkers                          int           `yaml:"num_workers"`
+	ObjectStoreType                     string        `yaml:"object_store_type"`
+	AllowDeleteRequestCancellationUntil time.Duration `yaml:"allow_delete_request_cancellation_until"`
```
pstibrany, line 90: What about `CancelPeriodDuration`? It is shorter and I think it describes the intent well.

pstibrany: Or `RequestCancellationPeriod`.

sandeepsukhani: Well, naming is hard. How about `DeleteRequestCancellableUntil`? I wanted to add `Delete` word to the flag to make it clearer. I can drop it if you think that is too much for a flag and people should infer from it or refer the docs to understand what type of request we are talking about.
What do you think?

pstibrany: > Well, naming is hard. How about `DeleteRequestCancellableUntil`? I wanted to add `Delete` word to the flag to make it clearer. I can drop it if you think that is too much for a flag and people should infer from it or refer the docs to understand what type of request we are talking about.
> What do you think?

I'd like to get rid of "Until" part as that implies static timestamp to me, and prefer to use Period or Duration instead. Also in Go world, people spell Cancelled as "Canceled", we can be consistent.

So: `DeleteRequestCancelationPeriod` ?

pstibrany: Or slightly shorter: `DeleteRequestCancelPeriod`

sandeepsukhani: Changed it to `DeleteRequestCancelPeriod`. Thanks!

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
 * [CHANGE] Available command-line flags are printed to stdout, and only when requested via `-help`. Using invalid flag no longer causes printing of all available flags. #2691
 * [CHANGE] Experimental Memberlist ring: randomize gossip node names to avoid conflicts when running multiple clients on the same host, or reusing host names (eg. pods in statefulset). Node name randomization can be disabled by using `-memberlist.randomize-node-name=false`. #2715
 * [CHANGE] Memberlist KV client is no longer considered experimental. #2725
+* [CHANGE] Change target flag for purger from `data-purger` to `purger` and make delete request cancellation duration configurable. #2760
```
bboreham, line 53: Can you clarify where this happened?
Not, for instance, here: https://github.com/cortexproject/cortex/pull/2760/files#diff-596d5408728e4813da55bfbc5a171d3dR63

bboreham:
Did we mark the purger as experimental?  If not, changing the name would trigger a major release.

CHANGELOG.md
sandeepsukhani: Sorry, my bad. I just renamed the components and missed changing the flag. Will open a PR to fix this.

sandeepsukhani:
> Did we mark the purger as experimental? If not, changing the name would trigger a major release.

Yes, it is marked experimental. See https://github.com/cortexproject/cortex/blob/master/pkg/chunk/purger/purger.go#L141

PR #2761, opened 2020-06-22
Added query-tee to the list of released binaries

pracucci:
**What this PR does**:
We're currently releasing only the cortex binaries. Since we've got few users running the query-tee too, I would like to also build and release query-tee binaries along with Cortex releases.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci:
> Is it worth adding a changelog entry?

I'm not sure, because it's an internal tooling used for releases and people will just see the binary in the next Cortex release.

PR #2762, opened 2020-06-22
fix race in processing of headers in sharded queries

sandeepsukhani:
**What this PR does**:
We saw query frontend panicking with `fatal error: concurrent map iteration and map write` [here](https://github.com/cortexproject/cortex/blob/a92b613bc6a3ad8ce1080ae96b403fb2eb955968/pkg/querier/queryrange/querysharding.go#L327). It is possibly due to a race with finishing all the querying and processing of headers while the response headers from sharded queries are still not merged [here](https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/queryable.go#L99)

This PR takes care of that issue. 

Fixes: #2763


joe-elliott:
Fixes #2762 ?

pstibrany:
Can you please add CHANGELOG entry?

Issue #2765, opened 2020-06-22
Timestamp Skew Across -querier.query-ingesters-within


joe-elliott:
Queries across the `-querier.query-ingesters-within` boundary don't come back with regular intervals.

With this value

`-querier.query-ingesters-within=12h`

And these `query_range` parameters:
```
  query=1+1
  start=1592769600  (now-24h)
  end=1592856000   (now)
  step=3600
```

These results are retrieved:
```
{"status":"success","data":{"resultType":"matrix","result":[{"metric":{},"values":[[1592769600,"2"],[1592773200,"2"],[1592776800,"2"],[1592780400,"2"],[1592784000,"2"],[1592787600,"2"],[1592791200,"2"],[1592794800,"2"],[1592798400,"2"],[1592802000,"2"],[1592805600,"2"],[1592809200,"2"],[1592812800,"2"],[1592814998.564,"2"],[1592818598.564,"2"],[1592822198.564,"2"],[1592825798.564,"2"],[1592829398.564,"2"],[1592832998.564,"2"],[1592836598.564,"2"],[1592840198.564,"2"],[1592843798.564,"2"],[1592847398.564,"2"],[1592850998.564,"2"],[1592854598.564,"2"]]}]}}
```

Note the transition from ts `1592812800` => `1592814998.564`.  These are the only consecutive timestamps that don't match the requested step of `3600`.  Also note that: `(1592856000 - 1592812800) / 3600 = 12`.  This is the interval that crosses the `-querier.query-ingesters-within` boundary.  

cc @owen-d

owen-d:
Thanks for writing this up @joe-elliott. We should be able to circumvent this problem by not trying to be clever via splitting across the `-querier.query-ingesters-within` boundary.

Instead, we should split as normally dictated by the `split_by_interval` middleware and disable the sharding optimizations for partitions that touch the ingester side of that boundary.

PR #2766, opened 2020-06-22
Do not split requests around ingester boundaries

owen-d:
This PR avoids trying to optimize requests that cross the `querier.query-ingesters-within` boundary because it mangles regular step values when splitting queries in the query-frontend.

Now, only queries which would only hit the store are eligible for sharding and the sharding middleware no longer splits queries by time.

Closes https://github.com/cortexproject/cortex/issues/2765


pstibrany:
Can you please add CHANGELOG entry?

PR #2771, opened 2020-06-24
refactor processing of writes in boltdb index client

sandeepsukhani:
**What this PR does**:
This PR refactors batch writes in BoltDB by doing both inserts and deletes in the same transaction.
It also makes write batch and a method to write on passed db reference public which helps with boltdb shipper implementation in Loki.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/local/boltdb_index_client.go
+++ b/pkg/chunk/local/boltdb_index_client.go
@@ -291,7 +291,7 @@ type BoltWriteBatch struct {
 	Writes map[string]TableWrites
 }
 
-func (b *BoltWriteBatch) getOrCreateTableWritesBatch(tableName string) TableWrites {
+func (b *BoltWriteBatch) getOrCreateTableWrites(tableName string) TableWrites {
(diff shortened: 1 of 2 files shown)


pkg/chunk/local/boltdb_index_client.go
```diff
@@ -285,31 +282,40 @@ func (b *BoltIndexClient) QueryDB(ctx context.Context, db *bbolt.DB, query chunk
...
+	deletes map[string]struct{}
 }
 
-func (b *boltWriteBatch) Delete(tableName, hashValue string, rangeValue []byte) {
-	table, ok := b.deletes[tableName]
+type BoltWriteBatch struct {
+	Writes map[string]TableWrites
+}
+
+func (b *BoltWriteBatch) getOrCreateTableWritesBatch(tableName string) TableWrites {
```
pracucci, line 294: I would name this `getOrCreateTableWrites` because the struct is called `TableWrites`, unless you intended to call the struct `TableWritesBatch` (in that case I would rename the struct).

PR #2774, opened 2020-06-24
make all clients for cassandra singleton

### LINKED ISSUE #2772: Panics when using Purger with Cassandra
```
level=info ts=2020-06-24T06:30:45.832053361Z caller=main.go:133 msg="Starting Cortex" version="(version=1.1.0, branch=r93, revision=ecad1573f)"
level=warn ts=2020-06-24T06:30:45.832128263Z caller=experimental.go:19 msg="experimental feature in use" feature="Cassandra Backend"
level=info ts=2020-06-24T06:30:45.869170931Z caller=events.go:247 module=gocql client=index-read msg=Session.handleNodeUp ip=52.176.155.127 port=10350
level=info ts=2020-06-24T06:30:45.920053787Z caller=events.go:247 module=gocql client=index-write msg=Session.handleNodeUp ip=52.176.155.127 port=10350
level=info ts=2020-06-24T06:30:45.933321962Z caller=server.go:192 http=[::]:80 grpc=[::]:9095 msg="server listening on addresses"
level=warn ts=2020-06-24T06:30:45.933795672Z caller=experimental.go:19 msg="experimental feature in use" feature="Cassandra Backend"
goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*wrappingRegisterer).MustRegister(0xc000b6a3f0, 0xc00068f7f0, 0x1, 0x1)
	/go/src/github.com/cortexproject/cortex/vendor/github.com/prometheus/client_golang/prometheus/wrap.go:102 +0x195
github.com/prometheus/client_golang/prometheus/promauto.Factory.NewGaugeFunc(0x3dc1a80, 0xc000b6a3f0, 0x0, 0x0, 0x0, 0x0, 0x367085e, 0x1b, 0x36c024b, 0x2f, ...)
	/go/src/github.com/cortexproject/cortex/vendor/github.com/prometheus/client_golang/prometheus/promauto/auto.go:319 +0x13c
github.com/gocql/gocql.newPolicyConnPool(0x3d661c0, 0xc000b6a390, 0x3dc1a80, 0xc000b6a3f0, 0xc000b2b000, 0xc000b57560)
	/go/src/github.com/cortexproject/cortex/vendor/github.com/gocql/gocql/connectionpool.go:132 +0x22f
github.com/gocql/gocql.PoolConfig.buildPool(...)
	/go/src/github.com/cortexproject/cortex/vendor/github.com/gocql/gocql/cluster.go:28
github.com/gocql/gocql.NewSession(0xc00068f760, 0x1, 0x1, 0x362f7c1, 0x5, 0x0, 0x77359400, 0x12a05f200, 0x286e, 0x7ffe8900bcb0, ...)
	/go/src/github.com/cortexproject/cortex/vendor/github.com/gocql/gocql/session.go:136 +0x55f
github.com/gocql/gocql.(*ClusterConfig).CreateSession(...)
	/go/src/github.com/cortexproject/cortex/vendor/github.com/gocql/gocql/cluster.go:207
github.com/cortexproject/cortex/pkg/chunk/cassandra.(*Config).session(0xc000c18fa0, 0x363ae17, 0xa, 0x0, 0x0, 0xc000b58ac0)
	/go/src/github.com/cortexproject/cortex/pkg/chunk/cassandra/storage_client.go:125 +0x6da
github.com/cortexproject/cortex/pkg/chunk/cassandra.NewStorageClient(0x7ffe8900bc08, 0x2f, 0x286e, 0x7ffe8900bcb0, 0x6, 0x3631695, 0x6, 0x3, 0x101, 0x0, ...)
	/go/src/github.com/cortexproject/cortex/pkg/chunk/cassandra/storage_client.go:228 +0x6e
github.com/cortexproject/cortex/pkg/chunk/storage.NewIndexClient(0xc000315990, 0x9, 0x3631c71, 0x6, 0x0, 0x4000000000000000, 0x4024000000000000, 0x0, 0x0, 0x186a0, ...)
	/go/src/github.com/cortexproject/cortex/pkg/chunk/storage/factory.go:211 +0x909
github.com/cortexproject/cortex/pkg/chunk/storage.NewStore(0x3631c71, 0x6, 0x0, 0x4000000000000000, 0x4024000000000000, 0x0, 0x0, 0x186a0, 0x3ff4cccccccccccd, 0x3ff0000000000000, ...)
	/go/src/github.com/cortexproject/cortex/pkg/chunk/storage/factory.go:156 +0x53e
github.com/cortexproject/cortex/pkg/cortex.(*Cortex).initStore(0xc0008f2800, 0x0, 0x0, 0x0, 0xc000100a80)
	/go/src/github.com/cortexproject/cortex/pkg/cortex/modules.go:272 +0x201
github.com/cortexproject/cortex/pkg/util/modules.(*Manager).InitModuleServices(0xc00011e800, 0x7ffe8900be35, 0xb, 0x3d98c00, 0xc000a8b5c0, 0x3e109e0)
	/go/src/github.com/cortexproject/cortex/pkg/util/modules/modules.go:72 +0x26f
github.com/cortexproject/cortex/pkg/cortex.(*Cortex).Run(0xc0008f2800, 0xc000599c40, 0x4)
	/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:294 +0x5b
main.main()
	/go/src/github.com/cortexproject/cortex/cmd/cortex/main.go:135 +0xa18
```

Config options:

```
      -azure.account-key=<hidden>
      -azure.account-name=<hidden>
      -azure.container-name=<hidden>
      -cassandra.addresses=<hidden>
      -cassandra.auth=true
      -cassandra.disable-initial-host-lookup=true
      -cassandra.host-verification=false
      -cassandra.keyspace=cortex
      -cassandra.password=<hidden>
      -cassandra.port=10350
      -cassandra.replication-factor=3
      -cassandra.ssl=true
      -cassandra.username=<hidden>
      -deletes.store=cassandra
      -purger.enable=true
      -purger.num-workers=2
      -purger.object-store-type=azure
      -schema-config-file=/etc/cortex/schema/config.yaml
      -target=data-purger
```

sandeepsukhani:
**What this PR does**:
We create Cassandra clients for object store, index store and table manager. We sometimes need multiple instances of each of them at different places based on how Cortex is configured. While creating a single instance for each type of usage is not a problem, creating multiple of same types of client instances causes the code to panic due to duplicate metric registration. See #2772. The issue can be reproduced easily by using Cassandra as index store with multiple periodic schemas.

This code makes each type of client singleton. The underlying client for each of the type is anyways pooling the connections so there is no need to create multiple instances of them.

Fixes #2772 

**Checklist**


tomwilkie:
Globals are bad (especially like this - I can forgive them for metrics).

Can we have the name of the schema (perhaps based on its start and end date) passed to the client so it can name it metrics appropriately and prevent duplicates?  I should be enough to add this as an extra label in the registererer.

tomwilkie:
Also, lets add a test that reproduces this issue first (ie starts Cortex with two Cassandra schemas)

pkg/chunk/cassandra/sessions.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"sync"
+
+	"github.com/gocql/gocql"
+)
+
+var (
+	sharedSessions map[string]*sharedSession
+	sessionsMtx    sync.Mutex
```
tomwilkie, line 11: These are still globals.  If we give easy session its own name, we won't get registrations error, and we won't need globals.

jtlisi:
@sandeepsukhani One way you could solve this problem is by making use of the https://github.com/cortexproject/cortex/blob/master/pkg/util/metrics_helper.go library. We use this in quite a few packages in Cortex: 
- https://github.com/cortexproject/cortex/blob/aa0d48adf6309a406e34b2d9e467787328312efb/pkg/ingester/metrics.go
- https://github.com/cortexproject/cortex/blob/aa0d48adf6309a406e34b2d9e467787328312efb/pkg/storegateway/bucket_store_metrics.go
- https://github.com/cortexproject/cortex/blob/3c6875dd433a34faaed604cca6b73ac5b63faea8/pkg/alertmanager/alertmanager_metrics.go

It allows you to create a new Prometheus registry when creating a client. Then you can instantiate a metrics struct that reports the aggregate metrics from each clients injected registry. Look at how Marco configured the bucket_store or the alertmanager_metrics link above and you should get a sense of what I mean.

pkg/chunk/cassandra/sessions.go
sandeepsukhani: Sorry, I saw your review after pushing the code. I thought we didn't want to create multiple sessions per schema config where Cassandra is use so I went this way which I accept is ugly.
I didn't get a chance to work on this yet. Will fix this.

sandeepsukhani:
@jtlisi Thanks! Will check it out.

sandeepsukhani:
> Globals are bad (especially like this - I can forgive them for metrics).
> 
> Can we have the name of the schema (perhaps based on its start and end date) passed to the client so it can name it metrics appropriately and prevent duplicates? I should be enough to add this as an extra label in the registererer.

@tomwilkie I have just added the start date of the schema to the label to keep things simple.
Also please note that when using the same date for multiple periodic configs with Cassandra can still cause the code to panic since it would try to register the metric with the same label values. Considering the chances of someone having a need for such config is negligible we should be good here. We can always add an end date as well to the label if someone faces an issue.
Please let me know if you think we should have it now.

pkg/chunk/cassandra/storage_client.go
```diff
@@ -89,7 +89,7 @@ func (cfg *Config) Validate() error {
 	return nil
 }
 
-func (cfg *Config) session(name string) (*gocql.Session, error) {
+func (cfg *Config) session(name, purpose string, registerer prometheus.Registerer) (*gocql.Session, error) {
```
tomwilkie, line 92: Lets not add purpose here, but rather pass in `prometheus.DefaultRegisterer(prometheus.Labels{"schema_start": ...})` to `NewStorageClient`

pkg/chunk/storage/factory.go
```diff
@@ -180,7 +180,7 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 }
 
 // NewIndexClient makes a new index client of the desired type.
-func NewIndexClient(name string, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.IndexClient, error) {
+func NewIndexClient(name string, cfg Config, schemaCfg chunk.SchemaConfig, purpose string, registerer prometheus.Registerer) (chunk.IndexClient, error) {
```
tomwilkie, line 183: Lets not pass a string here but the DayTime.  And lets not pass the registerer here, but wrap a new registerer next to the instantiation of the Cassandra client.  Otherwise I feel it sets expectation that the registerer will be used for all client.s

sandeepsukhani: The problem is `NewIndexClient` is also used to create `IndexClient` for storing delete requests [here](https://github.com/cortexproject/cortex/blob/92622a86800c5f5d12eda1cc74db030413f866d3/pkg/cortex/modules.go#L336), which is not tied to start date.
I added the `purpose` label to track the sessions based on the purpose of usage which could be for storing series index(using Daytime for this since there are multiple schemas) or for indexing delete requests or any other use case that we may have in future.

tomwilkie: In that case leave the registerer plumbed through but don't plumb reason though - just wrap the registerer before each instantiation of NewIndexClient.

sandeepsukhani: We can create the registerer with the `purpose` or something label for all the clients if you think that is more appropriate and clearer. I think there is no harm in having an extra-label.

tomwilkie: I don't mind having the extra label (its needed) but there is no point in plumbing it through like that.  We should add it close to where it semantically makes sense.

sandeepsukhani: @tomwilkie I have done these changes and also made changes to use the same registerer for metrics exported by dynamodb. Thanks for the review!

pkg/chunk/storage/factory.go
```diff
@@ -267,7 +270,7 @@ func newChunkClientFromStore(store chunk.ObjectClient, err error) (chunk.Client,
 }
 
 // NewTableClient makes a new table client based on the configuration.
-func NewTableClient(name string, cfg Config) (chunk.TableClient, error) {
+func NewTableClient(name string, cfg Config, registerer prometheus.Registerer) (chunk.TableClient, error) {
```
pracucci, line 273: This is called in a couple of places in `modules.go`. Shouldn't we wrap the registered there as well to add the `component` label?

pkg/chunk/storage/factory.go
```diff
@@ -153,7 +153,10 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 	stores := chunk.NewCompositeStore(cacheGenNumLoader)
 
 	for _, s := range schemaCfg.Configs {
-		index, err := NewIndexClient(s.IndexType, cfg, schemaCfg)
+		reg := prometheus.WrapRegistererWith(
```
pracucci, line 156: To avoid any misunderstanding with the outer `reg` variable, I would name it `indexReg`.

pkg/chunk/storage/factory.go
```diff
@@ -163,7 +166,7 @@ func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConf
 		if objectStoreType == "" {
 			objectStoreType = s.IndexType
 		}
-		chunks, err := NewChunkClient(objectStoreType, cfg, schemaCfg)
+		chunks, err := NewChunkClient(objectStoreType, cfg, schemaCfg, reg)
```
pracucci, line 169: Here you should wrap the reg with a different label value, like `chunks-client-` + `s.From.String()`.

pkg/chunk/storage/factory_test.go
```diff
@@ -154,6 +157,45 @@ func TestCustomIndexClient(t *testing.T) {
 	}
 }
 
+func TestCassandraInMultipleSchemas(t *testing.T) {
```
pracucci, line 160: What are you testing specifically here? That the registered doesn't panic?

pkg/cortex/modules.go
```diff
@@ -439,9 +441,14 @@ func (t *Cortex) initTableManager() (services.Service, error) {
 
 	var extraTables []chunk.ExtraTables
 	if t.Cfg.PurgerConfig.Enable {
-		deleteStoreTableClient, err := storage.NewTableClient(t.Cfg.Storage.DeleteStoreConfig.Store, t.Cfg.Storage)
-		if err != nil {
-			return nil, err
+		var deleteStoreTableClient chunk.TableClient
```
pracucci, line 444: I don't think you should do this. I believe you should wrap the registerer here as well, to keep the delete store fully decoupled from the table manager.

pkg/chunk/aws/metrics.go
```diff
@@ -0,0 +1,60 @@
+package aws
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/weaveworks/common/instrument"
+)
+
+type metrics struct {
```
pracucci, line 9: I would call it `dynamoDBMetrics` (and filename + `newMetric()` renamed accordingly) considering in this package we also have the S3 client and these metrics are DynamoDB specific.

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -792,17 +746,17 @@ func (b dynamoDBReadRequest) TakeReqs(from dynamoDBReadRequest, max int) {
...
 			if req.Error != nil {
-				recordDynamoError(tableName, req.Error, operation)
+				recordDynamoError(tableName, req.Error, operation, dynamoFailures)
 			}
 		})
 	}
 }
 
-func recordDynamoError(tableName string, err error, operation string) {
+func recordDynamoError(tableName string, err error, operation string, dynamoFailures *prometheus.CounterVec) {
```
pracucci, line 759: You may consider to pass the entire `*metrics`. I would be less error prone, otherwise you may do a mistake and pass the wrong metric.

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -792,17 +746,17 @@ func (b dynamoDBReadRequest) TakeReqs(from dynamoDBReadRequest, max int) {
 	}
 }
 
-func withErrorHandler(tableName, operation string) func(req *request.Request) {
+func withErrorHandler(tableName, operation string, dynamoFailures *prometheus.CounterVec) func(req *request.Request) {
```
pracucci, line 749: You may consider to pass the entire `*metrics`. I would be less error prone, otherwise you may do a mistake and pass the wrong metric.

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -187,7 +141,7 @@ func (a dynamoDBStorageClient) NewWriteBatch() chunk.WriteBatch {
 	return dynamoDBWriteBatch(map[string][]*dynamodb.WriteRequest{})
 }
 
-func logWriteRetry(ctx context.Context, unprocessed dynamoDBWriteBatch) {
+func logWriteRetry(unprocessed dynamoDBWriteBatch, dynamoThrottled *prometheus.CounterVec) {
```
pracucci, line 144: You may consider to pass the entire `*metrics`. I would be less error prone, otherwise you may do a mistake and pass the wrong metric.

pkg/chunk/storage/factory_test.go
sandeepsukhani: yes, do you see any problem with it or the way it is tested?

pkg/chunk/storage/factory.go
sandeepsukhani: We don't need to because we currently only manage tables for current periodic config but I have changed it anyways.

pkg/cortex/modules.go
```diff
@@ -333,7 +333,9 @@ func (t *Cortex) initDeleteRequestsStore() (serv services.Service, err error) {
 	}
 
 	var indexClient chunk.IndexClient
-	indexClient, err = storage.NewIndexClient(t.Cfg.Storage.DeleteStoreConfig.Store, t.Cfg.Storage, t.Cfg.Schema)
+	reg := prometheus.WrapRegistererWith(
+		prometheus.Labels{"purpose": "delete-requests"}, prometheus.DefaultRegisterer)
```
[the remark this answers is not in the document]
sandeepsukhani: Sounds good, I have changed it to `component` label.

sandeepsukhani:
> Thanks @sandeepsukhani ! I appreciate the refactoring you did. I left few comments, minor things, but I would be glad if you could take a look. Thanks!

Thanks for the detailed feedback @pracucci!

> Also, is there any existing unit test we can enrich passing the registerer and asserting on the collected metrics?

Unfortunately, I don't think we have any such test. None of the index clients was accepting a registerer before.
Also, different index clients have different metrics(if any) so we would only have tests specific to the index clients.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 * [BUGFIX] Ingester: Flushing chunks via `/flush` endpoint could previously lead to panic, if chunks were already flushed before and then removed from memory during the flush caused by `/flush` handler. Immediate flush now doesn't cause chunks to be flushed again. Samples received during flush triggered via `/flush` handler are no longer discarded. #2778
 * [BUGFIX] Prometheus upgraded. #2849
   * Fixed unknown symbol error during head compaction
+* [BUGFIX] Fix panic when using cassandra as store for both index and delete requests. #2774
```
pracucci, line 50: Please also add a `[CHANGE]` entry to explain `component` label and which metric names affect.

pkg/cortex/modules.go
```diff
@@ -425,7 +427,10 @@ func (t *Cortex) initTableManager() (services.Service, error) {
 		os.Exit(1)
 	}
 
-	tableClient, err := storage.NewTableClient(lastConfig.IndexType, t.Cfg.Storage)
+	reg := prometheus.WrapRegistererWith(
+		prometheus.Labels{"component": "table-manager-index-chunk-" + lastConfig.From.String()}, prometheus.DefaultRegisterer)
```
pracucci, line 431: I would say `table-manager-store` as label value would be enough, being the table manager used to manage the index/chunks store, differently then the `table-manager-delete-requests-store` below.

pkg/chunk/cassandra/storage_client.go
```diff
@@ -89,7 +89,7 @@ func (cfg *Config) Validate() error {
 	return nil
 }
 
-func (cfg *Config) session(name string) (*gocql.Session, error) {
+func (cfg *Config) session(name string, registerer prometheus.Registerer) (*gocql.Session, error) {
```
jtlisi, line 92:
```suggestion
func (cfg *Config) session(name string, reg prometheus.Registerer) (*gocql.Session, error) {
```

I noticed you used `reg` for the dynamodb package and `registerer` here. I don't have a preference but we should try to be consistent if we can.

pkg/cortex/modules.go
sandeepsukhani: We want to change table manager to manage table for older schemas as well so I think we would have to add back the start time of the schema then which we might miss so I added it now itself. I can change just the prefix to `table-manager-store-` if you don't mind.

pracucci: I'm always in favour of doing a change when required (right now, it's not and if in the future you change the value of `component` it's not even considered a CHANGE, contrary to adding a new label like in this case) but **I'm not opposed** if you prefer to already do it. Up to you.

sandeepsukhani: Fair enough changed it. Thanks!

pracucci:
@sandeepsukhani Sorry for the veeeeeery long wait. I just learned how to dismiss a review and unblock the merge. Thanks for your work here and... 🚀  let's go!

Issue #2744, opened 2020-06-18
/flush handler doesn't wait for flush to finish


pstibrany:
When `/flush` handler on ingester is called, it triggers flush of all chunks. It logs following lines:

```
level=info ts=2020-06-18T08:23:53.474269984Z caller=flush.go:42 msg="starting to flush all the chunks"
level=info ts=2020-06-18T08:24:03.286485064Z caller=flush.go:44 msg="flushing of chunks complete"
```

But this is misleading. Chunks flush is not complete at this point, all unflushed chunks are only enqueued for flushing, and are waiting in the queue when this log message is written and handler returns.

We should make `/flush` handler to wait for flush queues to be empty, if possible.

pstibrany:
On the other hand, waiting for empty queues can take a long time. We already have `cortex_ingester_flush_queue_length` metric to monitor queue lengths, so maybe we should just fix the logged message, and write something like "123456 chunks enqueued for flushing" to the HTTP response (right now response is empty).

pstibrany:
`cortex_ingester_flush_queue_length` alone is not enough -- there are still ongoing flushes even when `cortex_ingester_flush_queue_length` is 0. PR #2778 introduces new `cortex_ingester_flush_series_in_progress` metric that can help to monitor that. Only when `cortex_ingester_flush_queue_length + cortex_ingester_flush_series_in_progress`  reaches 0, all flushes are done.

Issue #2287, opened 2020-03-18
Add a counter for chunks removed from the flush queue but not flushed


bboreham:
If a chunk is idle and added to the flush queue, then gets new samples shortly before it is dequeued, it will no longer be idle so simply ignored.

It seems like a nice property to be able to add up all reasons why chunks are added to the queue and all reasons they are removed and have them total the same number.

Specifically in the current code, the total `cortex_ingester_flush_reasons` does not equal the sum of `cortex_ingester_chunks_stored_total` and `cortex_ingester_dropped_chunks_total`.

(and saying that makes me realise `stored` and `dropped` should be labels on one metric)

pstibrany:
`cortex_ingester_flush_reasons` counts series, while `cortex_ingester_chunks_stored_total` and `cortex_ingester_dropped_chunks_total` count chunks, so they cannot be compared. We could add new metric or modify `cortex_ingester_flush_reasons` to `cortex_ingester_chunks_flush_reasons`.

We can fix `cortex_ingester_flush_reasons` by moving that increment from where it is now (just after series has been enqueued) to `flushUserSeries` method, after dequeing and checking flush reason again. That would fix issue you describe. WDYT? (I can put that into #2778 PR, if you agree).

bboreham:
I don't agree this should be fixed simply by not recording why series are added to the queue.

Good point that some metrics count chunks and some count series.

PR #2778, opened 2020-06-24
Fix chunks flushing bugs

### LINKED ISSUE #2743: Panic while flushing chunks
This panic happened while calling `/flush` endpoint on ingester in internal cluster. Ingester was running `grafana/cortex-cortex:r92-a0674c0c2` (https://github.com/grafana/cortex/commits/a0674c0c2f528f8aa5297768604e17ca7ebb047f)

```
level=info ts=2020-06-18T08:23:49.848675718Z caller=flush.go:42 msg="starting to flush all the chunks"
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x279a9b7]

goroutine 1288 [running]:
github.com/cortexproject/cortex/pkg/ingester.(*Ingester).flushChunks(0xc0006e8800, 0x3ee0260, 0xc002814d20, 0xc000e0c6b0, 0x3, 0x87ff6a90a1aa9836, 0xc0191ae160, 0xb, 0xb, 0xc1c44a5180, ...)
	/go/src/github.com/cortexproject/cortex/pkg/ingester/flush.go:344 +0x637
github.com/cortexproject/cortex/pkg/ingester.(*Ingester).flushUserSeries(0xc0006e8800, 0x20, 0xc000e0c6b0, 0x3, 0x87ff6a90a1aa9836, 0x1, 0x0, 0x0)
	/go/src/github.com/cortexproject/cortex/pkg/ingester/flush.go:291 +0x843
github.com/cortexproject/cortex/pkg/ingester.(*Ingester).flushLoop(0xc0006e8800, 0x20)
	/go/src/github.com/cortexproject/cortex/pkg/ingester/flush.go:205 +0x10a
created by github.com/cortexproject/cortex/pkg/ingester.(*Ingester).startFlushLoops
	/go/src/github.com/cortexproject/cortex/pkg/ingester/ingester.go:256 +0x68
```

pstibrany:
**What this PR does**: There were few bugs in chunks flushing code that this PR tries to fix:
- https://github.com/cortexproject/cortex/issues/2743
- Samples added during immediate flush were discarded
- If removal of old chunks happens during flush, incorrect chunks are marked as flushed (this can now happen since this PR keeps immediate-flushed chunks in memory)

In addition to that, this PR also:
-  introduces new metric `cortex_ingester_flush_series_in_progress` (needed to fix integration test)
- updates log messages related to flushing chunks (Fixes https://github.com/cortexproject/cortex/issues/2744)

Fixes #2743 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

(Notify @gouthamve @bboreham as chunk experts)


bboreham:
I wonder if it would be possible to fix #2287 at the same time?
Maybe if we move the point that the metric is computed to after we know what happened to the chunk, we can get away without adding a new metric?

pkg/ingester/flush.go
```diff
@@ -238,8 +247,11 @@ func (i *Ingester) flushUserSeries(flushQueueIndex int, userID string, fp model.
 		return nil
 	}
 
-	// shouldFlushSeries() has told us we have at least one chunk
-	chunks := series.chunkDescs
+	// shouldFlushSeries() has told us we have at least one chunk.
+	// Make a copy of chunks descriptors slice, to avoid possible issues related to removing (and niling) elements from chunkDesc.
+	// This can happen if first chunk is already flushed -- removeFlushedChunks may set such chunk to nil.
+	// Since elements in the slice are pointers, we can still safely update chunk descriptors after the copy.
+	chunks := append([]*desc(nil), series.chunkDescs...)
```
pracucci, line 254: A `make([]*desc, 0, len(series.chunkDescs)` + `copy()`  may me more efficient?

pkg/ingester/flush.go
```diff
@@ -275,6 +287,11 @@ func (i *Ingester) flushUserSeries(flushQueueIndex int, userID string, fp model.
 
 	userState.fpLocker.Unlock(fp)
 
+	// No need to flush these chunks again.
+	for len(chunks) > 0 && chunks[0].flushed {
```
pracucci, line 291: Question: is it always guaranteed that chunks are flushed in order? May end up in a situation where X is not flushed but X+1 is?

bboreham: Not in the current code; `Put()` is always implemented by iterated calls to `PutOne()`, and any error considers the whole set to have failed.
But there's nothing guaranteeing this.  Possibly we should remove `Put()` to avoid confusion.

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@
 * [CHANGE] Removed `-store.fullsize-chunks` option which was undocumented and unused (it broke ingester hand-overs). #2656
 * [CHANGE] Query with no metric name that has previously resulted in HTTP status code 500 now returns status code 422 instead. #2571
 * [CHANGE] Experimental Delete Series: Change target flag for purger from `data-purger` to `purger`. #2777
+* [CHANGE] Ingester: Chunks flushed via /flush stay in memory until retention period is reached. This affects `cortex_ingester_memory_chunks` metric. #2778
```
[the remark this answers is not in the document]
pstibrany: I can do that, but there are already changes merged on master, that are not in 1.2-rc0, so cleanup needs to happen when merging release branch to master. (Personally I think we should do the first merge as soon as 1.2-rc0 is tagged, to avoid this problem growing. It is a pain to clean up.)

pkg/ingester/flush.go
pstibrany: I don't think sightly complicating code to be more efficient here will get us much... there is typically only one or two chunks in the slice. And I assume (without checking) that compiler is smart enough to optimize for this case anyway.

pstibrany:
> I wonder if it would be possible to fix #2287 at the same time?
> Maybe if we move the point that the metric is computed to after we know what happened to the chunk, we can get away without adding a new metric?

I will take a look at it before merging.

pstibrany:
> > I wonder if it would be possible to fix #2287 at the same time?
> > Maybe if we move the point that the metric is computed to after we know what happened to the chunk, we can get away without adding a new metric?
> 
> I will take a look at it before merging.

I've sent separate PR #2802 for issue #2287.

pstibrany:
I will merge this tomorrow, if there are no objections.

PR #2779, opened 2020-06-24
Initialize RuntimeConfig module only if config is specified

annanay25:

**What this PR does**:
This PR is a follow up from #2092. Updated so that the `RuntimeConfig` module no longer initializes if the config path is empty.

Fixes #<issue number>

**Checklist**
- NA Tests updated
- NA Documentation added

### DIFF
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -122,6 +122,10 @@ func (t *Cortex) initRing() (serv services.Service, err error) {
 
 func (t *Cortex) initRuntimeConfig() (services.Service, error) {
 	if t.Cfg.RuntimeConfig.LoadPath == "" {
+		if t.Cfg.LimitsConfig.PerTenantOverrideConfig == "" {
+			// no need to initialize module if load path is empty
+			return nil, nil
+		}
 		t.Cfg.RuntimeConfig.LoadPath = t.Cfg.LimitsConfig.PerTenantOverrideConfig
 		t.Cfg.RuntimeConfig.ReloadPeriod = t.Cfg.LimitsConfig.PerTenantOverridePeriod
 	}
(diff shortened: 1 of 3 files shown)


pkg/cortex/modules.go
```diff
@@ -122,6 +122,10 @@ func (t *Cortex) initRing() (serv services.Service, err error) {
 
 func (t *Cortex) initRuntimeConfig() (services.Service, error) {
 	if t.Cfg.RuntimeConfig.LoadPath == "" {
+		if t.Cfg.LimitsConfig.PerTenantOverrideConfig == "" {
+			// no need to initialize module if load path is empty
+			return nil, nil
```
pstibrany, line 127: I'm pretty sure leaving `t.RuntimeConfig` set to nil will cause problems elsewhere.

pstibrany: Nitpick, but can we move this check *after* first check?

- First `if` -- check whether to use new or old-style options
- Second `if` -- if LoadPath is still empty, return quickly

The way it is done right now mixes these two concerns, that should be separate.

pkg/cortex/modules.go
```diff
@@ -122,6 +122,10 @@ func (t *Cortex) initRing() (serv services.Service, err error) {
...
+		}
 		t.Cfg.RuntimeConfig.LoadPath = t.Cfg.LimitsConfig.PerTenantOverrideConfig
 		t.Cfg.RuntimeConfig.ReloadPeriod = t.Cfg.LimitsConfig.PerTenantOverridePeriod
 	}
```
pstibrany, line 131:
```suggestion
	if t.Cfg.RuntimeConfig.LoadPath == "" {
		t.Cfg.RuntimeConfig.LoadPath = t.Cfg.LimitsConfig.PerTenantOverrideConfig
		t.Cfg.RuntimeConfig.ReloadPeriod = t.Cfg.LimitsConfig.PerTenantOverridePeriod
	}

	if t.Cfg.RuntimeConfig.LoadPath == "" {
		// no need to initialize module if load path is empty
		return nil, nil
	}
```

pkg/cortex/modules.go
annanay25: Sure, thanks @pstibrany! Making the changes.

PR #2783, opened 2020-06-24
new flag: ruler.outage-tolerance

owen-d:
## What

Exposes the underlying Prometheus manager option, `OutageTolerance`. This is the lookback period for which the metric store can be queried for past alert evaluations. It's used when each rule group is initially run. In Cortex, this has implications not only during restarts, but also when the ring membership changes. Without this configuration, previous rule state cannot be restored, meaning that after re-sharding/restarts, we must wait the entire `ForDuration` until an alert may be considered firing.

edit: defaulted to prometheus defaults for these flags and included a total of 3 configs at Jacob's suggestion:
```
# Max time to tolerate outage for restoring "for" state of alert.
# CLI flag: -ruler.for-outage-tolerance
[for_outage_tolerance: <duration> | default = 1h]

# Minimum duration between alert and restored "for" state. This is maintained
# only for alerts with configured "for" time greater than grace period.
# CLI flag: -ruler.for-grace-period
[for_grace_period: <duration> | default = 10m]

# Minimum amount of time to wait before resending an alert to Alertmanager.
# CLI flag: -ruler.resend-delay
[resend_delay: <duration> | default = 1m]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## master / unreleased
 
-* [CHANGE] Introduced `ruler.outage-tolerance`, the outage period after which previous alert state cannot be checked from metric storage and must be re-evaluated in full.
+* [FEATURE] Introduced `ruler.for-outage-tolerance`, Max time to tolerate outage for restoring "for" state of alert. #2783
+* [FEATURE] Introduced `ruler.for-grace-period`, Minimum duration between alert and restored "for" state. This is maintained only for alerts with configured "for" time greater than grace period. #2783
+* [FEATURE] Introduced `ruler.for-resend-delay`, Minimum amount of time to wait before resending an alert to Alertmanager. #2783
 * [CHANGE] Metric `cortex_kv_request_duration_seconds` now includes `name` label to denote which client is being used as well as the `backend` label to denote the KV backend implementation in use. #2648
 * [CHANGE] Experimental Ruler: Rule groups persisted to object storage using the experimental API have an updated object key encoding to better handle special characters. Rule groups previously-stored using object storage must be renamed to the new format. #2646
 * [CHANGE] Query Frontend now uses Round Robin to choose a tenant queue to service next. #2553
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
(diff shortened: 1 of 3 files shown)


pkg/ruler/ruler.go
```diff
@@ -132,6 +136,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.FlushCheckPeriod, "ruler.flush-period", 1*time.Minute, "Period with which to attempt to flush rule groups.")
 	f.StringVar(&cfg.RulePath, "ruler.rule-path", "/rules", "file path to store temporary rule files for the prometheus rule managers")
 	f.BoolVar(&cfg.EnableAPI, "experimental.ruler.enable-api", false, "Enable the ruler api")
+	f.DurationVar(&cfg.OutageTolerance, "ruler.outage-tolerance", 0, "outage period after which previous alert state cannot be cannot be checked from metric storage and must be re-evaluated in full.")
```
jtlisi, line 139: Can you update this flag to look more like:

https://github.com/prometheus/prometheus/blob/b788986717e1597452ca25e5219510bb787165c7/cmd/prometheus/main.go#L231-L232

Keep the `ruler.` prefix but in general it would be good to keep a similar naming scheme

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Introduced `ruler.outage-tolerance`, the outage period after which previous alert state cannot be checked from metric storage and must be re-evaluated in full.
```
jtlisi, line 5: I think this is a `FEATURE` change entry.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master / unreleased
 
+* [FEATURE] Introduced `ruler.for-outage-tolerance`, Max time to tolerate outage for restoring "for" state of alert. #2783
+* [FEATURE] Introduced `ruler.for-grace-period`, Minimum duration between alert and restored "for" state. This is maintained only for alerts with configured "for" time greater than grace period. #2783
+* [FEATURE] Introduced `ruler.for-resend-delay`, Minimum amount of time to wait before resending an alert to Alertmanager. #2783
```
pracucci, line 7: Uups! The flag is `ruler.resend-delay`. @owen-d Would you mind opening a PR to fix the CHANGELOG entry, please?

bboreham: Also put it in the correct place, like we ask in the checklist.

PR #2788, opened 2020-06-25
[memberlist] Add support to join memberlist via SRV records

periklis:
**What this PR does**:

This PR enables lookup for members to join via SRV records.

Follow-up on discussion thread [here](https://grafana.slack.com/archives/CEPJRLQNL/p1592313566141400)

**Checklist**
- [ ] Tests updated
- [x] Documentation added

cc @pstibrany

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 * [ENHANCEMENT] Experimental TSDB: Added `-experimental.tsdb.head-compaction-idle-timeout` option to force compaction of data in memory into a block. #2803
 * [ENHANCEMENT] Experimental TSDB: Added support for flushing blocks via `/flush`, `/shutdown` (previously these only worked for chunks storage) and by using `-experimental.tsdb.flush-blocks-on-shutdown` option. #2794
 * [ENHANCEMENT] Ingester: Added new metric `cortex_ingester_flush_series_in_progress` that reports number of ongoing flush-series operations. Useful when calling `/flush` handler: if `cortex_ingester_flush_queue_length + cortex_ingester_flush_series_in_progress` is 0, all flushes are finished. #2778
+* [ENHANCEMENT] Memberlist members can join cluster via SRV records. #2788
 * [BUGFIX] Fixed a bug in the index intersect code causing storage to return more chunks/series than required. #2796
 * [BUGFIX] Fixed the number of reported keys in the background cache queue. #2764
 * [BUGFIX] Fix race in processing of headers in sharded queries. #2762
--- a/pkg/ring/kv/memberlist/memberlist_client.go
+++ b/pkg/ring/kv/memberlist/memberlist_client.go
@@ -16,6 +16,8 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/hashicorp/memberlist"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/discovery/dns"
+	"github.com/thanos-io/thanos/pkg/extprom"
 
 	"github.com/cortexproject/cortex/pkg/ring/kv/codec"
 	"github.com/cortexproject/cortex/pkg/util"
@@ -164,7 +166,7 @@ func (cfg *KVConfig) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.BoolVar(&cfg.RandomizeNodeName, prefix+"memberlist.randomize-node-name", true, "Add random suffix to the node name.")
 	f.DurationVar(&cfg.StreamTimeout, prefix+"memberlist.stream-timeout", 0, "The timeout for establishing a connection with a remote node, and for read/write operations. Uses memberlist LAN defaults if 0.")
 	f.IntVar(&cfg.RetransmitMult, prefix+"memberlist.retransmit-factor", 0, "Multiplication factor used when sending out messages (factor * log(N+1)).")
-	f.Var(&cfg.JoinMembers, prefix+"memberlist.join", "Other cluster members to join. Can be specified multiple times.")
+	f.Var(&cfg.JoinMembers, prefix+"memberlist.join", "Other cluster members to join. Can be specified multiple times. It can be an IP, hostname or an entry specified in the DNS Service Discovery format (see https://cortexmetrics.io/docs/configuration/arguments/#dns-service-discovery for more details).")
 	f.DurationVar(&cfg.MinJoinBackoff, prefix+"memberlist.min-join-backoff", 1*time.Second, "Min backoff duration to join other cluster members.")
 	f.DurationVar(&cfg.MaxJoinBackoff, prefix+"memberlist.max-join-backoff", 1*time.Minute, "Max backoff duration to join other cluster members.")
 	f.IntVar(&cfg.MaxJoinRetries, prefix+"memberlist.max-join-retries", 10, "Max number of retries to join other cluster members.")
@@ -200,6 +202,9 @@ type KV struct {
 	services.Service
 
 	cfg KVConfig
+
+	// dns discovery provider
+	provider *dns.Provider
 
 	// Protects access to memberlist and broadcasts fields.
 	initWG     sync.WaitGroup
@@ -268,15 +273,23 @@ var (
 )
 
 // NewKV creates new gossip-based KV service. Note that service needs to be started, until then it doesn't initialize
-// gossiping part. Only after service is in Running state, it is really gossiping.
-// Starting the service will also trigger connecting to the existing memberlist cluster, if cfg.JoinMembers is set.
-// If that fails and AbortIfJoinFails is true, error is returned and service enters Failed state.
+// gossiping part. Only after service is in Running state, it is really gossiping. Starting the service will also
+// trigger connecting to the existing memberlist cluster. If that fails and AbortIfJoinFails is true, error is returned
+// and service enters Failed state.
 func NewKV(cfg KVConfig) *KV {
 	cfg.TCPTransport.MetricsRegisterer = cfg.MetricsRegisterer
 	cfg.TCPTransport.MetricsNamespace = cfg.MetricsNamespace
 
+	mr := extprom.WrapRegistererWithPrefix("cortex_",
+		extprom.WrapRegistererWith(
+			prometheus.Labels{"name": "memberlist"},
+			cfg.MetricsRegisterer,
+		),
+	)
+
 	mlkv := &KV{
 		cfg:            cfg,
+		provider:       dns.NewProvider(util.Logger, mr, dns.GolangResolverType),
 		store:          make(map[string]valueDesc),
 		codecs:         make(map[string]codec.Codec),
 		watchers:       make(map[string][]chan string),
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -128,6 +128,7 @@
   * `cortex_querier_blocks_consistency_checks_failed_total`
   * `cortex_querier_storegateway_refetches_per_query`
 * [ENHANCEMENT] Cortex is now built with Go 1.14. #2480 #2753
+* [ENHANCEMENT] Memberlist members can join cluster via SRV records. #
```
pracucci, line 131: 😉 
```suggestion
* [ENHANCEMENT] Memberlist members can join cluster via SRV records. #2788
```

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -475,6 +487,28 @@ func (m *KV) joinMembersOnStartup(ctx context.Context, members []string) error {
 	return lastErr
 }
 
+// Provides a simple SRV lookup to join a memberlist cluster w/o knowning members' addresses upfront.
+// Note: Overwrites flag `memberlist.join` if provided in addition to `memberlist.join-service`.
+func (m *KV) lookupMembers(srv string) error {
```
pracucci, line 492: I'm wondering if we could reuse https://cortexmetrics.io/docs/configuration/arguments/#dns-service-discovery instead of having another way to configure service discovery. The DNS service discovery we're currently using is based on Thanos one: you can see an example of its usage at `pkg/querier/blocks_store_balanced_set.go`.

Would it make sense to use it for gossip too?

periklis: @pracucci & @pstibrany Giving this a thought for the last couple of days, I believe it is a good idea to expand memberlist joins to use the same DNS service discovery in addition to our `joinMembers` flag. Although a simple `net.lookupSRV` is enough the use cases provided by Thanos' DNS provider make sense also for memberlists:
- Joining a cluster means resolving the addresses of all members upfront. However, leaving the resolution mechanism to a third-party dependency is good from a separation of concerns perspective and adds testability. OTOH my simple `net.lookupSRV` would need a function pointer handle to make it testable, that's not a big deal though.
- Re-Joining the cluster - e.g. periodically - should work the same as joining anew with the only difference of re-establishing connections.

AFAICS, the DNS service discovery mechanism is free of any caching of members and in addition to that the ultimate list of which members to connect to is maintained on memberlist side. In addition, reading through the hashicorp lib I don't see a blocker regarding how to resolve member addresses. WDYT?

@pracucci FWIW, using the DNS service disovery is not bound to DNS caching timeouts or so like that would demand to rejoin the cluster based on these timeouts right?

pracucci: > AFAICS, the DNS service discovery mechanism is free of any caching of members and in addition to that the ultimate list of which members to connect to is maintained on memberlist side. In addition, reading through the hashicorp lib I don't see a blocker regarding how to resolve member addresses. WDYT?

The way the Thanos DNS provider works is that `Addresses()` returns the result of the last `Resolve()`. For this specific use case you could just call `Resolve()` and then `Address()` in sequence.

> FWIW, using the DNS service disovery is not bound to DNS caching timeouts or so like that would demand to rejoin the cluster based on these timeouts right?

I'm not sure about the question. The DNS service discovery just uses the Go's DNS Lookup under the hood, so there's real difference compared to your current implementation, except that it would be more generic and we would reuse the same mechanism we have.

WDYT?

periklis: My concern question is rather a concern of the sort: Using the thanos DNS provider do I need to implement re-joins based on DNS TTLs?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -134,6 +135,7 @@ type KVConfig struct {
 	DeadNodeReclaimTime time.Duration `yaml:"dead_node_reclaim_time"`
 
 	// List of members to join
+	JoinService      string              `yaml:"join_service"`
```
pstibrany, line 138: Would it make sense to use this as a boolean flag to enable SRV records lookup for names in join_members?

pkg/ring/kv/memberlist/memberlist_client.go
pstibrany: I wouldn't bring TTLs into the question here.

periklis:
> I think the logic I'd like to see is that we use `dns.NewProvider`, pass configured `JoinMembers` to `(*Provider).Resolve` method, and use the result (`(*Provider).Addresses()`) to call `memberlist.Join` method. If `RejoinInterval` in memberlist client is configured, we would periodically repeat the process.
> 
> For this to work we perhaps don't even need extra options, as it should be backwards compatible.
> 
> What do you think, what are possible problems with such approach?

Indeed the boolean flag with the present `JoinMembers` flag will be sufficient and backward compatible.

pstibrany:
> Indeed the boolean flag with the present `JoinMembers` flag will be sufficient and backward compatible.

Maybe we can even do without it (I suggested bool flag before writing my other comment).

periklis:
> > Indeed the boolean flag with the present `JoinMembers` flag will be sufficient and backward compatible.
> 
> Maybe we can even do without it (I suggested bool flag before writing my other comment).

How would that work? Scan for `dns(srv|srvnoa)+` prefixes in `joinMembers`? Try DNS and fallback on err?

pstibrany:
> > Maybe we can even do without it (I suggested bool flag before writing my other comment).
> 
> How would that work? Scan for `dns(srv|srvnoa)+` prefixes in `joinMembers`? Try DNS and fallback on err?

Yes, or just look for ‘+’, which is not allowed in host names. If there is no +, we pass hostnames to memberlist directly. If there is, we do our own resolution.

periklis:
Good! I think we have a plan. I will amend this PR on Monday.

periklis:
@pracucci @pstibrany PTAL the last version is based on the DNS service discovery mechanism. Btw. I am trying to figure out how to write a proper integration test for this.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -475,6 +484,35 @@ func (m *KV) joinMembersOnStartup(ctx context.Context, members []string) error {
 	return lastErr
 }
 
+// Provides a dns-based member disovery to join a memberlist cluster w/o knowning members' addresses upfront.
+func (m *KV) discoverMembers() {
```
pracucci, line 488: I would prefer to not manipulate the config. I would suggest to pick in input `m.cfg.JoinMembers` and return a list of addresses and then passing these addresses to the `m.joinMembersOnStartup()` and `m.memberlist.Join()` instead of `m.cfg.JoinMembers`.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -377,10 +382,14 @@ func (m *KV) starting(_ context.Context) error {
 var errFailedToJoinCluster = errors.New("failed to join memberlist cluster on startup")
 
 func (m *KV) running(ctx context.Context) error {
+	// Lookup SRV records for given service to discover members.
+	// Overwrite any given `memberlist.join` members if any found via SRV lookup.
+	members := m.discoverMembers(m.cfg.JoinMembers)
+
```
pstibrany, line 388: Should this be inside `if len(m.cfg.JoinMembers) > 0` condition?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -475,6 +486,34 @@ func (m *KV) joinMembersOnStartup(ctx context.Context, members []string) error {
...
+	if len(m.cfg.JoinMembers) == 0 {
+		return nil
+	}
+
+	var ms []string
+
+	for _, member := range members {
+		if strings.Contains(member, "+") {
+
+			err := m.provider.Resolve(context.Background(), []string{member})
```
pstibrany, line 500: Should we resolve all "+" members at once, instead of one by one?

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -268,15 +272,16 @@ var (
...
+// gossiping part. Only after service is in Running state, it is really gossiping. Starting the service will also
+// trigger connecting to the existing memberlist cluster. If that fails and AbortIfJoinFails is true, error is returned
+// and service enters Failed state.
 func NewKV(cfg KVConfig) *KV {
 	cfg.TCPTransport.MetricsRegisterer = cfg.MetricsRegisterer
 	cfg.TCPTransport.MetricsNamespace = cfg.MetricsNamespace
 
 	mlkv := &KV{
 		cfg:            cfg,
+		provider:       dns.NewProvider(util.Logger, cfg.MetricsRegisterer, dns.GolangResolverType),
```
pracucci, line 284: The Prometheus registerer passed to the DNS provider should be something like this (to keep it consistent with other components and avoid "duplicated metrics registration"):

```
dnsProviderRegisterer := prometheus.WrapRegistererWithPrefix("cortex_", prometheus.WrapRegistererWith(prometheus.Labels{
		"name": "memberlist",
	}, cfg.MetricsRegisterer))
```

@pstibrany Make sense to you? Is it safe to use an hardcoded `memberlist` name?

periklis: IC, the `cfg.MetricsRegisterer` is only the defaultRegisterer up in the stack `pkg/cortex/modules.go` for memberlist. It makes sense to me to separate the provider metrics per component.

periklis:
> Thanks @periklis for patiently addressing all comments. Very much appreciated! I have one last comment (I just realized it) and then we should be good to go to me.

You are welcome! Although, I believe I should have catched most of them, if I put more focus time on this PR ;)

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -201,6 +202,9 @@ type KV struct {
 
 	cfg KVConfig
 
+	// dns discovery provider
+	provider *dns.Provider
```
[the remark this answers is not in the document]
periklis: TBH, i like to read members always with the given type without duplications between type and name, e.g. `provider of type dns.Provider`. I admit there is a small duplication here but it is prefixed with `dns.` so for me not an issue. Same for resolver, it duplicates method names: `resolver.resolve(members)`. Later reminds me on the old Java dictatorship `everything is a class`, whereas if I can build something as a simple function why calling it via a class with the same name.

Calling out `provider of type dns.Provider resolve(members)` sounds ok. WDYT?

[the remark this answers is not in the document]
pstibrany: It's not a big deal, my problem is that later in the code, `m.provider` seems too generic -- it's not clear what it actually provides without checking the type.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -401,7 +413,9 @@ func (m *KV) running(ctx context.Context) error {
 	for {
 		select {
 		case <-tickerChan:
-			reached, err := m.memberlist.Join(m.cfg.JoinMembers)
+			members := m.discoverMembers(m.cfg.JoinMembers)
+
+			reached, err := m.memberlist.Join(members)
```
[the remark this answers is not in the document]
periklis: Taking a closer look, `tickerChan` is a nil channel to close out the select case. Shouldn't the creation be bound to?
```
if m.cfg.RejoinInterval > 0 && len(m.cfg.JoinMembers) >0 {
...
}
```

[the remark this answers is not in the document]
pstibrany: Yes, that is what I was trying to suggest :)

periklis:
@pstibrany Furthermore I missed to pass the context from `running` to `discoverMembers`.

pkg/ring/kv/memberlist/memberlist_client.go
```diff
@@ -475,6 +493,33 @@ func (m *KV) joinMembersOnStartup(ctx context.Context, members []string) error {
...
+			resolve = append(resolve, member)
+		} else {
+			// No DNS SRV record to lookup, just append member
+			ms = append(ms, member)
+		}
+	}
+
+	err := m.provider.Resolve(ctx, resolve)
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "failed to resolve members from SRV records", "addrs", strings.Join(resolve, ","))
```
pstibrany, line 515: I'd drop `from SRV records` part from the message, resolver may support different ways of resolving.

PR #2794, opened 2020-06-25
Added support for flushing blocks

pstibrany:
**What this PR does**: This PR adds support for flushing blocks.

Flushing can be triggered by:
- using `-experimental.tsdb.flush-blocks-on-shutdown` option, honored when ingester shuts down
- using `/flush` endpoint (works for chunks too) -- this triggers flush and shipping (if enabled), but endpoint returns immediately. There are log messages written after flushing and shipping (if enabled) has finished
- using `/shutdown` endpoint (works for chunks too) -- this shuts down ingester with flush flag enabled. Endpoint waits until flushing and shipping has finished.

This follows the proposal from PR #2717.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -134,7 +134,9 @@ type TSDBState struct {
 	subservices *services.Manager
 
 	tsdbMetrics *tsdbMetrics
-	shipTrigger chan struct{}
+
+	forceCompactTrigger chan chan<- struct{}
+	shipTrigger         chan chan<- struct{}
 
 	// Head compactions metrics.
 	compactionsTriggered   prometheus.Counter
@@ -162,10 +164,11 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 		usersMetadata: map[string]*userMetricsMetadata{},
 		wal:           &noopWAL{},
 		TSDBState: TSDBState{
-			dbs:         make(map[string]*userTSDB),
-			bucket:      bucketClient,
-			tsdbMetrics: newTSDBMetrics(registerer),
-			shipTrigger: make(chan struct{}, 1),
+			dbs:                 make(map[string]*userTSDB),
+			bucket:              bucketClient,
+			tsdbMetrics:         newTSDBMetrics(registerer),
+			forceCompactTrigger: make(chan chan<- struct{}),
+			shipTrigger:         make(chan chan<- struct{}),
 
 			compactionsTriggered: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
 				Name: "cortex_ingester_tsdb_compactions_triggered_total",
@@ -220,7 +223,7 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
 	i.limiter = NewLimiter(limits, i.lifecycler, cfg.LifecyclerConfig.RingConfig.ReplicationFactor, cfg.ShardByAllLabels)
 	i.userStates = newUserStates(i.limiter, cfg, i.metrics)
 
-	i.Service = services.NewBasicService(i.startingV2, i.updateLoop, i.stoppingV2)
+	i.BasicService = services.NewBasicService(i.startingV2, i.updateLoop, i.stoppingV2)
 	return i, nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -1141,3 +1177,75 @@ sendLoop:
...
+
+	level.Info(util.Logger).Log("msg", "finished flushing and shipping TSDB blocks")
+}
+
+// Blocks version of Flush handler. It force-compacts blocks, and triggers shipping.
+func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
+	go func() {
+		ingCtx := i.BasicService.ServiceContext()
+		if ingCtx == nil || ingCtx.Err() != nil {
+			// Not started or already finished.
```
pracucci, line 1202: We should log an error here, otherwise we silently discard the flush request.

pkg/ingester/ingester_v2.go
```diff
@@ -1141,3 +1177,75 @@ sendLoop:
...
+		}
+
+		ch := make(chan struct{}, 1)
+
+		level.Info(util.Logger).Log("msg", "flushing TSDB blocks: triggering compaction")
+		select {
+		case i.TSDBState.forceCompactTrigger <- ch:
+			// Compacting now.
+		case <-ingCtx.Done():
+			level.Info(util.Logger).Log("msg", "failed to compact TSDB blocks, ingester not running anymore")
```
pracucci, line 1213: I would log it as warning.

pkg/ingester/ingester_v2.go
```diff
@@ -1141,3 +1177,75 @@ sendLoop:
...
+			return
+		}
+
+		// Wait until notified about compaction being finished.
+		select {
+		case <-ch:
+			level.Info(util.Logger).Log("msg", "finished compacting TSDB blocks")
+		case <-ingCtx.Done():
+			level.Info(util.Logger).Log("msg", "failed to compact TSDB blocks, ingester not running anymore")
```
pracucci, line 1222: I would log it as warning.

pkg/ingester/ingester_v2.go
```diff
@@ -1141,3 +1177,75 @@ sendLoop:
...
+		}
+
+		if i.cfg.TSDBConfig.ShipInterval > 0 {
+			level.Info(util.Logger).Log("msg", "flushing TSDB blocks: triggering shipping")
+
+			select {
+			case i.TSDBState.shipTrigger <- ch:
+				// shipping now
+			case <-ingCtx.Done():
+				level.Info(util.Logger).Log("msg", "failed to ship TSDB blocks, ingester not running anymore")
```
pracucci, line 1233: warning

pkg/ingester/ingester_v2.go
```diff
@@ -1141,3 +1177,75 @@ sendLoop:
...
+				return
+			}
+
+			// Wait until shipping finished.
+			select {
+			case <-ch:
+				level.Info(util.Logger).Log("msg", "shipping of TSDB blocks finished")
+			case <-ingCtx.Done():
+				level.Info(util.Logger).Log("msg", "failed to ship TSDB blocks, ingester not running anymore")
```
pracucci, line 1242: warning

pstibrany:
Thanks for review Marco. I did some changes in `compactBlocks` method after rebase, so would you please take a look again?

pkg/ingester/ingester_v2.go
```diff
@@ -1126,18 +1151,28 @@ func (i *Ingester) compactBlocks(ctx context.Context) {
...
+		reason := ""
+		switch {
+		case force:
+			reason = "forced"
 			err = userDB.CompactHead(tsdb.NewRangeHead(h, h.MinTime(), h.MaxTime()))
-		} else {
+
+		case i.cfg.TSDBConfig.HeadCompactionIdleTimeout > 0 && userDB.isIdle(time.Now(), i.cfg.TSDBConfig.HeadCompactionIdleTimeout):
+			reason = "idle"
+			level.Debug(util.Logger).Log("msg", "TSDB is idle, forcing compaction", "user", userID)
```
pracucci, line 1163: I would log as Info, since it should be unfrequent but noticed. I missed this in the previous review when this was introduced.

pstibrany: Sounds good. I was wondering whether we want to use `reason` as label for started/failed compaction counters? WDYT?

PR #2700, opened 2020-06-09
Fix Redis cache error when a query has no chunks to lookup

pracucci:
**What this PR does**:
Redis chunks cache lookup can fail with the error `ERR wrong number of arguments for 'mget' command` if a query has no chunks to lookup from storage. This happens because we don't correctly handle the case `SeriesStore.GetChunkRefs()` returns no chunks.

**Which issue(s) this PR fixes**:
Fixes #2133

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2796, opened 2020-06-26
Fixes an issue in the index chunks/series intersect code.

cyriltovena:
This was introduce in #2700, more specifically this line https://github.com/cortexproject/cortex/pull/2700/files#diff-10bca0f4f31a2ca1edc507d0289b143dR537

This causes any query with the first label matcher not matching anything to return all matches of all other labels.
This is a nasty one since, the code was relying on empty slice, and so it would skip nil values instead of returning no matches. I've added a regression test proving this is fixed everywhere. I think in Cortex it probably affect performance (since you have to download all chunks not required) but not read integrity.

I have found this with @slim-bean while deploying Loki, we fought a lot to pin down this exact issue while all queriers where OOMing.

/cc @gouthamve @pracucci @pstibrany We need to put on hold all current releases this is a dangerous bug.
/cc @rfratto @slim-bean Let's wait for this to be merged before applying the latest release of Loki.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci:
_Let's merge this fix quicky and then @bboreham can cherry-pick it into the release-1.2._

PR #2800, opened 2020-06-26
fix cache generation keys prefix

sandeepsukhani:
**What this PR does**:
Fixes a problem with removing cache generation number from keys prefixed with it.
Also adds required tests.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/cache/cache_gen_test.go
+++ b/pkg/chunk/cache/cache_gen_test.go
@@ -1,0 +1,39 @@
+package cache
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCacheGenNumCacheKeysPrefix(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/chunk/cache/cache_gen_test.go
```diff
@@ -0,0 +1,39 @@
...
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := InjectCacheGenNumber(context.Background(), tc.prefix)
+
+			prefixedKeys := addCacheGenNumToCacheKeys(ctx, keys)
+			for i, key := range prefixedKeys {
+				require.Equal(t, tc.prefix+keys[i], key)
+			}
+
```
gouthamve, line 32: Also worth checking if the length of the `keys`, `prefixedKeys` and `unprefixedKeys` is the same.

pstibrany:
Can you please take a look at changelog conflicts?

PR #2801, opened 2020-06-26
add support for delete chunks in index stores

sandeepsukhani:
**What this PR does**:
Adds support for delete chunks in stores pending support for it.

I will send a follow-up PR with integration tests for delete series which should verify this.

**Checklist**


pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -594,6 +590,27 @@ func (a dynamoDBStorageClient) PutChunks(ctx context.Context, chunks []chunk.Chu
 	return a.BatchWrite(ctx, dynamoDBWrites)
 }
 
+func (a dynamoDBStorageClient) DeleteChunk(ctx context.Context, chunkID string) error {
+	userID, err := user.ExtractOrgID(ctx)
+	if err != nil {
+		return err
+	}
```
pstibrany, line 597: This adds hidden dependency on having userID in the context, which was not there before. If DeleteChunk needs userID, it should be passed explicitly.

Furthermore, it's not really needed -- ParseExternalKey only uses it for verification.

Same comment applies to other `DeleteChunk` implementations in this PR.

sandeepsukhani: Changed it. Thanks for the review!

sandeepsukhani:
> `Client` interface looks little weird to me... `DeleteChunk` is the only method of the interface that takes userID as a parameter. If other methods use userID from the context, we should fix that too eventually.

@pstibrany In table-based stores for chunks we need `From` time from chunk to find the table to which chunk belongs. In case of `DeleteChunks` we are passing just the `ChunkID` which requires parsing it to find the chunk start time while for other methods we are passing the whole chunk which includes the time range of the chunk. The other option is passing chunkref which is built by parsing the `ChunkID` but has no data in it except what is derived from the `ChunkID` but I don't usually prefer that because it is a half baked chunk and you have to the read code to understand that it is not a full chunk if you are facing some problem. This is in the most explicit form I feel. Please let me know if you feel otherwise.

pkg/chunk/azure/blob_storage_client.go
```diff
@@ -202,6 +202,11 @@ func (b *BlobStorage) List(ctx context.Context, prefix string) ([]chunk.StorageO
 }
 
 func (b *BlobStorage) DeleteObject(ctx context.Context, chunkID string) error {
```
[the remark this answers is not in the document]
sandeepsukhani: I have changed it. Thanks!

PR #2802, opened 2020-06-26
Update seriesFlushReason when flushing, not enqueing.

### LINKED ISSUE #2287: Add a counter for chunks removed from the flush queue but not flushed

pstibrany:
**What this PR does**: This PR changes `cortex_ingester_flush_reasons`. It is renamed to `cortex_ingester_series_flush_reasons` since it counts series and not chunks, and it is now updated when flushing, not enqueing. This fixes problem described in issue #2287, when idle series could have been enqueued, but then updated before flushing.

Fixes #2287 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/metrics.go
+++ b/pkg/ingester/metrics.go
@@ -52,7 +52,7 @@ type ingesterMetrics struct {
 	chunkSize                     prometheus.Histogram
 	chunkAge                      prometheus.Histogram
 	memoryChunks                  prometheus.Gauge
-	flushReasons                  *prometheus.CounterVec
+	seriesFlushReasons            *prometheus.CounterVec
 	droppedChunks                 prometheus.Counter
 	oldestUnflushedChunkTimestamp prometheus.Gauge
 }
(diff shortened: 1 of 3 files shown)


pkg/ingester/metrics.go
```diff
@@ -187,9 +187,9 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
 			Name: "cortex_ingester_memory_chunks",
 			Help: "The total number of chunks in memory.",
 		}),
-		flushReasons: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
-			Name: "cortex_ingester_flush_reasons",
-			Help: "Total number of series scheduled for flushing, with reasons.",
+		seriesFlushReasons: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+			Name: "cortex_ingester_series_flush_reasons",
```
pracucci, line 191: Why adding the `_reasons`? To me, it should be `cortex_ingester_series_flushed_total`; the fact the `reason` is a dimension shouldn't be in the metric name.

bboreham:
This change makes the metric less useful. 
If you have an enormous queue you want to know why now, not in two hours when it is flushed.

PR #2803, opened 2020-06-26
Compact idle blocks.

pstibrany:
**What this PR does**: This PR adds compaction of blocks that are idle for longer than specified timeout.

~Draft: needs tests.~

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -52,6 +52,7 @@ type userTSDB struct {
 	seriesInMetric *metricCounter
 	limiter        *Limiter
 
+	// Used to detect idle TSDBs.
 	lastUpdateMutex sync.Mutex
 	lastUpdate      time.Time
 
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -110,14 +110,14 @@ func (u *userTSDB) PostDeletion(metrics ...labels.Labels) {
 
 func (u *userTSDB) isIdle(now time.Time, idle time.Duration) bool {
 	u.lastUpdateMutex.Lock()
-	defer u.lastUpdateMutex.Lock()
+	defer u.lastUpdateMutex.Unlock()
 
 	return u.lastUpdate.Add(idle).Before(now)
 }
 
 func (u *userTSDB) setLastUpdate(t time.Time) {
 	u.lastUpdateMutex.Lock()
-	defer u.lastUpdateMutex.Lock()
+	defer u.lastUpdateMutex.Unlock()
 	u.lastUpdate = t
 }
 
@@ -1119,8 +1119,7 @@ func (i *Ingester) compactBlocks(ctx context.Context) {
 
 		// Don't do anything, if there is nothing to compact.
 		h := userDB.Head()
-		minT, maxT := h.MinTime(), h.MaxTime()
-		if minT == maxT {
+		if h.NumSeries() == 0 {
 			return
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -105,6 +109,19 @@ func (u *userTSDB) PostDeletion(metrics ...labels.Labels) {
...
 }
 
+func (u *userTSDB) isIdle(now time.Time, idle time.Duration) bool {
+	u.lastUpdateMutex.Lock()
+	defer u.lastUpdateMutex.Unlock()
+
+	return u.lastUpdate.Add(idle).Before(now)
+}
+
+func (u *userTSDB) setLastUpdate(t time.Time) {
```
pracucci, line 119: This is called in the `v2Push()` which is quite a sensitive hot path. I'm a bit scared of the exclusive mutex. What do you think if we use an atomic storing the epoch?

pkg/ingester/ingester_v2.go
```diff
@@ -438,6 +455,8 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 	}
 	i.TSDBState.appenderCommitDuration.Observe(time.Since(startCommit).Seconds())
 
+	db.setLastUpdate(time.Now())
```
pracucci, line 458: We can use `startCommit` instead of `time.Now()`. When called in a hot path, `time.Now()` is a bit slow, so I would avoid if not strictly required (like in this case).

pkg/storage/tsdb/config.go
```diff
@@ -139,6 +140,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxTSDBOpeningConcurrencyOnStartup, "experimental.tsdb.max-tsdb-opening-concurrency-on-startup", 10, "limit the number of concurrently opening TSDB's on startup")
 	f.DurationVar(&cfg.HeadCompactionInterval, "experimental.tsdb.head-compaction-interval", 1*time.Minute, "How frequently does Cortex try to compact TSDB head. Block is only created if data covers smallest block range. Must be greater than 0 and max 5 minutes.")
 	f.IntVar(&cfg.HeadCompactionConcurrency, "experimental.tsdb.head-compaction-concurrency", 5, "Maximum number of tenants concurrently compacting TSDB head into a new block")
+	f.DurationVar(&cfg.HeadCompactionIdleTimeout, "experimental.tsdb.head-compaction-idle-timeout", 0, "If TSDB head is idle for this duration, it is compacted. 0 means disabled.")
```
pracucci, line 143: What if we enable by default with a reasonable timeout? (ie. 1h?)

pstibrany:
> The next ingester startup, we don't want to load the TSDB at all if all existing blocks on disk have passed the retention period. This part is missing (perfectly fine doing it in a separate PR, but I think should be done).

I agree this part is missing. I would like to address that in separate PR.

In this PR, I don't want to actually close idle TSDB, because that would cause issues with the way how we collect metrics, esp. in case we need to reopen it. But ignoring on startup is a good way to handle it, I think.

pkg/ingester/ingester_v2.go
pstibrany: All `time.Now` does is reading a specific address in memory, where kernel updates the current time. We have more `time.Now` calls here (hidden in `time.Since`). Should we optimize them all?

pstibrany: Ok, time.Now does little bit more, although not much.

pracucci: _Let's resolve this conversation and focus on real value. Sorry for that._

PR #2804, opened 2020-06-26
Release 1.2

bboreham:
I need to have a PR open so CI will run


k8s/alertmanager-dep.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: alertmanager
-        image: quay.io/cortexproject/cortex:v1.1.0
+        image: quay.io/cortexproject/cortex:v1.2.0
```
pstibrany, line 18: These YAML files will not work until version 1.2.0 is pushed.

VERSION
```diff
@@ -1 +1 @@
-1.1.0
\ No newline at end of file
+1.2.0-rc.1
```
pstibrany, line 1: Changelog says -rc.0 is the current one.

bboreham:
To repeat, the only reason this PR exists is to make CI run.  I will never be able to merge it on GitHub due to conflicts in CHANGELOG.

pstibrany:
> To repeat, the only reason this PR exists is to make CI run. I will never be able to merge it on GitHub due to conflicts in CHANGELOG.

Can we resolve conflicts and merge? We need to merge this branch to master eventually.

I feel like we should do a retrospective on release process and fix our document. There seems to be too much magic :-(

codesome:
In Prometheus, whenever there are conflicts, we just create a new branch out of `release-x.y` branch and do the fixing there, because we cannot add arbitrary commits in the release branch.

bboreham:
> > force-pushed the release-1.2 branch
> This is worrying too.

It was to fix a DCO error.  

I guess we should have a rule to do nothing at all by hand in the `release-x` branch, only in `prepare-release` branches which are PR'd into `release-x`

bboreham:
> Can we resolve conflicts and merge? 

If you mean via the "resolve conflicts" button, that merges master into release-1.2 which we definitely don't want.
I thought that #2791 would make the merge clean, but evidently not. 
The release instructions say to do it by hand on the command-line and turn off GitHub protection.

Open to other suggestions.

pstibrany:
> The release instructions say to do it by hand on the command-line and turn off GitHub protection.

I don't see any other way. As you point out, we definitely don't want to merge `master` into release branch.

bboreham:
Interesting: GitHub took my command-line merge as closing this PR.
And, I think it took the approval of this PR as meaning I could push without disabling any rules.

pstibrany:
Interesting. That may explain why I was able to do similar push once before without disabling the rule.

PR #2806, opened 2020-06-29
track oldest delete request age since their cancellation period is over

sandeepsukhani:
**What this PR does**:
Changes the tracking of the age of delete requests from "since creation time" to "since they are over their cancellation period".
This would help with defining alerts and SLO/SLAs since the cancellation period is configurable now.

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   * `cortex_bucket_stores_gate_queries_in_flight`
   * `cortex_bucket_stores_gate_duration_seconds`
 * [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_series_flushed_total`, and is now incremented during flush, not when series is enqueued for flushing. #2802
-* [CHANGE] Experimental Delete Series: Metric `purger_pending_delete_requests_count` would track age of delete requests since they are over their cancellation period instead of their creation time. #2806
+* [CHANGE] Experimental Delete Series: Metric `cortex_purger_oldest_pending_delete_request_age_seconds` would track age of delete requests since they are over their cancellation period instead of their creation time. #2806
 * [FEATURE] Introduced `ruler.for-outage-tolerance`, Max time to tolerate outage for restoring "for" state of alert. #2783
 * [FEATURE] Introduced `ruler.for-grace-period`, Minimum duration between alert and restored "for" state. This is maintained only for alerts with configured "for" time greater than grace period. #2783
 * [FEATURE] Introduced `ruler.resend-delay`, Minimum amount of time to wait before resending an alert to Alertmanager. #2783
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   * `cortex_bucket_stores_gate_queries_in_flight`
   * `cortex_bucket_stores_gate_duration_seconds`
 * [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_series_flushed_total`, and is now incremented during flush, not when series is enqueued for flushing. #2802
+* [CHANGE] Experimental Delete Series: Metric `purger_pending_delete_requests_count` would track age of delete requests since they are over their cancellation period instead of their creation time. #2806
 * [FEATURE] Introduced `ruler.for-outage-tolerance`, Max time to tolerate outage for restoring "for" state of alert. #2783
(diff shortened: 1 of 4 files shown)


pkg/chunk/purger/purger.go
```diff
@@ -473,7 +477,12 @@ func (p *Purger) pullDeleteRequestsToPlanDeletes() error {
 		p.executePlansChan <- req
 	}
 
-	p.metrics.oldestPendingDeleteRequestAgeSeconds.Set(float64(now.Sub(oldestPendingRequestCreatedAt) / time.Second))
+	// track age of oldest delete request since they are over their cancellation period
+	oldestPendingRequestAge := time.Duration(0)
```
pracucci, line 481: Have you considered changing the initial `oldestPendingRequestCreatedAt := now` setting it to `time.Time{}` and then here checking `if !oldestPendingRequestCreatedAt.IsZero()`? I'm mentioning it because I think it's a bit hard to understand why we check `! .Equal(now)` (which is theoretically legit) while you wanna actually check if it was never set.

pkg/chunk/purger/purger.go
```diff
@@ -248,6 +248,10 @@ func (p *Purger) workerJobCleanup(job workerJob) {
 			default:
 				// already sent
 			}
+		} else if len(p.usersWithPendingRequests) == 0 {
```
pracucci, line 251: [open question, just to understand] I'm wondering if this `else if` should actually be an `if {`. What if the previous `if` branch deletes the latest `p.usersWithPendingRequests`?

pkg/chunk/purger/purger.go
```diff
@@ -66,7 +66,7 @@ func newPurgerMetrics(r prometheus.Registerer) *purgerMetrics {
 	m.oldestPendingDeleteRequestAgeSeconds = promauto.With(r).NewGauge(prometheus.GaugeOpts{
 		Namespace: "cortex",
 		Name:      "purger_oldest_pending_delete_request_age_seconds",
-		Help:      "Age of oldest pending delete request in seconds",
+		Help:      "Age of oldest pending delete request in seconds, since they are over their cancellation period",
```
pracucci, line 69: I'm not sure I do understand what you mean by "since they are over their cancellation period". Could you clarify it?

pkg/chunk/purger/purger.go
sandeepsukhani: I have used `model.Time(0)` and checked for `0` value instead of `time.Time{}` because `CreatedAt` is of type `model.Time`.

pkg/chunk/purger/purger.go
sandeepsukhani: If the previous `if {` is true then that means there are more pending requests and the age of oldest pending request is non-zone. Even if I set it to 0 by changing this `else if` to an independent `if` condition then it would still get changed to a non-zero value when the requests get loaded. I don't see much of a difference except a dip to 0 if samples get scrapped before pending requests are found during the next load of requests.

pkg/chunk/purger/purger.go
sandeepsukhani: We do not process requests until they are over their cancellation period i.e they are not allowed to be cancelled anymore. I wanted to make it clear because people would think the age is since they are created which is not the case here.
I can remove it if it looks too confusing. Also, we have a metric to count the number of pending requests https://github.com/cortexproject/cortex/blob/35b756fa80c80210a678c59ff65a004d2bf0a018/pkg/chunk/purger/purger.go#L71 and here also pending requests are those which are over their cancellation period.
What do you think? I think we should either add that statement to both or none.

pracucci: Now I got it and I think having it in the description is fine. Agree that we should add the same to `pendingDeleteRequestsCount` too.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   * `cortex_bucket_stores_gate_queries_in_flight`
   * `cortex_bucket_stores_gate_duration_seconds`
 * [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_series_flushed_total`, and is now incremented during flush, not when series is enqueued for flushing. #2802
+* [CHANGE] Experimental Delete Series: Metric `purger_pending_delete_requests_count` would track age of delete requests since they are over their cancellation period instead of their creation time. #2806
```
pracucci, line 11:
```suggestion
* [CHANGE] Experimental Delete Series: Metric `cortex_purger_oldest_pending_delete_request_age_seconds` would track age of delete requests since they are over their cancellation period instead of their creation time. #2806
```

PR #2809, opened 2020-06-29
Feature/agnostic ruler

owen-d:
**What this PR does**:
Refactors & inhouses the prometheus `rules` package in order to allow it to evaluate arbitrary expressions and defer both pushing series & querying series to implementers. This primarily aids Loki which uses these interfaces to evaluate it's own alerts & maintain a synthetic series store in memory.

For instance, there is now a new interface responsible for restoring the `ALERTS_FOR_STATE` after i.e. restarts or ring re-sharding in cortex. The default implementation is still the same (pulled from prometheus via the `rules` pkg), but it also allows Loki to take advantage of the ruler as well.

```
// TenantAlertHistory is a tenant specific interface for restoring the for state of an alert after restarts/etc.
type TenantAlertHistory interface {
	RestoreForState(ts time.Time, alertRule *AlertingRule)
	Stop()
}
```

**Changes**:
- extends `ConcreteSeries` to support new manipulations & bound growth.
- pulls in prometheus rules pkg
- refactors said rules pkg via a few interfaces in order to support evaluating arbitrary rules

**Checklist**
- [x] Tests updated

/cc @jtlisi 
/cc @juliusv Would this be something Prometheus would be interested in?

edit: closed in favor of https://github.com/cortexproject/cortex/pull/2812


pkg/ruler/rules/recording.go
```diff
@@ -0,0 +1,202 @@
...
+	yaml "gopkg.in/yaml.v2"
+
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/util/strutil"
+)
+
+// A RecordingRule records its vector expression into new timeseries.
+type RecordingRule struct {
```
jtlisi, line 33: Do we need to fork this implementation? The Prometheus implementation is public so why not just use that instead of forking it?

PR #2811, opened 2020-06-29
Revert "Update seriesFlushReason when flushing, not enqueing. (#2802)"

bboreham:
This reverts #2802.

Operators need to know why the flush queue is growing at that time, not at a later time when the queue has cleared.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@
   * `cortex_bucket_stores_gate_queries_concurrent_max`
   * `cortex_bucket_stores_gate_queries_in_flight`
   * `cortex_bucket_stores_gate_duration_seconds`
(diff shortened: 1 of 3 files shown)


pstibrany:
> This reverts #2802.
> 
> Operators need to know why the flush queue is growing at that time, not at a later time when the queue has cleared.

> This change makes the metric less useful.
> If you have an enormous queue you want to know why now, not in two hours when it is flushed.

Fair enough, but isn't this metric useful too? I mean we can keep the current metric, and add metric that counts them when enqueing.

bboreham:
Sure, we can have both.

pracucci:
> Sure, we can have both.

Agree on having both.

pracucci:
Closing because superseded by #2818.

PR #2812, opened 2020-06-29
Agnostic Ruler

owen-d:
**What this PR does**:
Refactors & inhouses the prometheus `rules` package in order to allow it to evaluate arbitrary expressions and defer both pushing series & querying series to implementers. This primarily aids Loki which uses these interfaces to evaluate it's own alerts & maintain a synthetic series store in memory.

For instance, there is now a new interface responsible for restoring the `ALERTS_FOR_STATE` after i.e. restarts or ring re-sharding in cortex. The default implementation is still the same (pulled from prometheus via the `rules` pkg), but it also allows Loki to take advantage of the ruler as well.

```
// TenantAlertHistory is a tenant specific interface for restoring the for state of an alert after restarts/etc.
type TenantAlertHistory interface {
	RestoreForState(ts time.Time, alertRule *AlertingRule)
	Stop()
}
```

**Changes**:
- extends `ConcreteSeries` to support new manipulations & bound growth.
- pulls in prometheus rules pkg
- refactors said rules pkg via a few interfaces in order to support evaluating arbitrary rules

**Checklist**
- [x] Tests updated

**Disclaimer**:
This has been bundled into two commits. The first is responsible for bringing the prometheus rules package into cortex. All changes therein are denoted via `Cortex-Note` comments. I chose to subsume the entire rules package after first struggling with many unexported pieces of the package when trying to do it piecemeal. By including the entire package, it also allows us to easily refactor this again in the future should this be upstreamed in Prometheus.

The following commit shows the changes to the ruler to fulfill the required interfaces/etc.

### DIFF
--- a/pkg/ruler/compat.go
+++ b/pkg/ruler/compat.go
@@ -2,28 +2,44 @@ package ruler
 
 import (
 	"context"
-	"errors"
 	"time"
 
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql"
-	"github.com/prometheus/prometheus/rules"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/weaveworks/common/user"
 
 	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
 )
+
+// AppendableHistoryFunc allows creation of an Appendable and AlertHistory to be joined. The default implementation
+// does not take advantage of this option.
+type AppendableHistoryFunc = func(userID string, opts *rules.ManagerOptions) (rules.Appendable, rules.TenantAlertHistory)
+
+// This is the default implementation which returns an unlinked Appendable/AlertHistory.
+func DefaultAppendableHistoryFunc(p Pusher, q storage.Queryable) AppendableHistoryFunc {
+	return func(userID string, opts *rules.ManagerOptions) (rules.Appendable, rules.TenantAlertHistory) {
+		return &appender{
+			pusher: p,
+			userID: userID,
+		}, rules.NewMetricsHistory(q, opts)
+	}
+}
 
 // Pusher is an ingester server that accepts pushes.
 type Pusher interface {
 	Push(context.Context, *client.WriteRequest) (*client.WriteResponse, error)
 }
+
 type appender struct {
 	pusher  Pusher
 	labels  []labels.Labels
 	samples []client.Sample
 	userID  string
 }
+
+func (a *appender) Appender(_ rules.Rule) (storage.Appender, error) { return a, nil }
 
 func (a *appender) Add(l labels.Labels, t int64, v float64) (uint64, error) {
 	a.labels = append(a.labels, l)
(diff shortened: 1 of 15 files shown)


pkg/ruler/rules/recording.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/util/strutil"
+)
+
+// A RecordingRule records its vector expression into new timeseries.
+type RecordingRule struct {
+	name string
+	// Cortex-Note: this is changed to fmt.Stringer instead of a parser.Expr in order to facilitate arbitrary expressions.
+	vector fmt.Stringer
```
jtlisi, line 36: Correct me if I'm wrong. The main reason we are forking this file is to change vector to be a `fmt.Stringer` type? Why not just use a `parser.StringLiteral`?

https://github.com/prometheus/prometheus/blob/72425d4e3d14d209cc3f3f6e10e3240411303399/promql/parser/ast.go#L150-L154

This would fulfill the `parser.Expr` interface and be functionally equivalent to an `fmt.Stringer` if I'm not mistaken? It would also make it so we don't have to fork the entire Prometheus `rules` package.

pkg/ruler/rules/manager.go
```diff
@@ -0,0 +1,953 @@
...
+			}}, nil
+		default:
+			return nil, errors.New("rule result is not a vector or scalar")
+		}
+	}
+}
+
+// A Rule encapsulates a vector expression which is evaluated at a specified
+// interval and acted upon (currently either recorded or used for alerting).
+type Rule interface {
```
jtlisi, line 206: Instead of forking this definition, can't we just vendor and use the upstream intterface. All of the functions are Public?

pkg/ruler/rules/manager.go
```diff
@@ -0,0 +1,953 @@
...
+						log.With(m.logger, "alert", r.Alert),
+					))
+					continue
+				}
+				rules = append(rules, NewRecordingRule(
+					r.Record.Value,
+					expr,
+					labels.FromMap(r.Labels),
+				))
+			}
```
jtlisi, line 893: Instead of making a`ParseExpr` an anonymous function, why not make the entire body of this for loop a function that can be passed to the manager. That way Cortex can have an implementation that uses the default Prometheus rule implementations and Loki can have its own implementations. Since the body of this code is related to creating a `Rule` which is an interface, it should be relatively easy to make pluggable.

pkg/ruler/compat.go
```diff
@@ -2,29 +2,45 @@ package ruler
...
 	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
 )
 
+// AppendableHistoryFunc allows creation of an Appendable and AlertHistory to be joined. The default implementation
+// does not take advantage of this option.
+type AppendableHistoryFunc = func(userID string, opts *rules.ManagerOptions) (rules.Appendable, rules.TenantAlertHistory)
+
+// This is the default implementation which returns an unlinked Appendable/AlertHistory.
+func DefaultAppendableHistoryFunc(p Pusher, q storage.Queryable) AppendableHistoryFunc {
```
jtlisi, line 21: If this function is pluggable, can Loki just inject it's own implementation and Cortex keep the current implementation?

gouthamve:
This change made it upstream, would it be useful for the alert state restoring? https://github.com/prometheus/prometheus/pull/7472/

owen-d:
Closing this in favor of making these decisions in Loki. A small PR to cortex should follow instead.

PR #2817, opened 2020-06-30
fix a race in purger managing in process delete requests

sandeepsukhani:
**What this PR does**:
Fixed a race in purger while managing in process delete requests.
Refactored some code to manage them easily.

**Checklist**

### DIFF
--- a/pkg/chunk/purger/purger.go
+++ b/pkg/chunk/purger/purger.go
@@ -365,6 +365,7 @@ func (p *Purger) loadInprocessDeleteRequests() error {
 	}
 
 	for _, deleteRequest := range requestsWithBuildingPlanStatus {
+		deleteRequest := deleteRequest
 		req := makeDeleteRequestWithLogger(deleteRequest, util.Logger)
 		p.inProcessRequests.add(deleteRequest.UserID, &deleteRequest)
 
@@ -387,6 +388,7 @@ func (p *Purger) loadInprocessDeleteRequests() error {
 	}
 
 	for _, deleteRequest := range requestsWithDeletingStatus {
+		deleteRequest := deleteRequest
 		req := makeDeleteRequestWithLogger(deleteRequest, util.Logger)
 		level.Info(req.logger).Log("msg", "loaded in process delete requests with status deleting")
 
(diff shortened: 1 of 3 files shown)


pkg/chunk/purger/purger.go
```diff
@@ -454,10 +444,7 @@ func (p *Purger) pullDeleteRequestsToPlanDeletes() error {
 			return err
 		}
 
-		p.inProcessRequestIDsMtx.Lock()
-		p.inProcessRequests[deleteRequest.UserID] = deleteRequest
-		p.inProcessRequestIDsMtx.Unlock()
-
+		p.inProcessRequests.add(deleteRequest.UserID, &deleteRequest)
```
pstibrany, line 447: This takes address of variable declared in `for` loop. This address will be the same for all iterations.

For this to work correctly, copy of deleteRequest must be made:

```suggestion
		deleteRequestCopy := deleteRequest
		p.inProcessRequests.add(deleteRequest.UserID, &deleteRequestCopy)
```

pkg/chunk/purger/purger.go
```diff
@@ -369,10 +366,10 @@ func (p *Purger) loadInprocessDeleteRequests() error {
 
 	for _, deleteRequest := range requestsWithBuildingPlanStatus {
 		req := makeDeleteRequestWithLogger(deleteRequest, util.Logger)
+		p.inProcessRequests.add(deleteRequest.UserID, &deleteRequest)
```
pstibrany, line 369: Same problem with using address of for-loop variable.

pkg/chunk/purger/purger.go
```diff
@@ -393,7 +390,7 @@ func (p *Purger) loadInprocessDeleteRequests() error {
 		req := makeDeleteRequestWithLogger(deleteRequest, util.Logger)
 		level.Info(req.logger).Log("msg", "loaded in process delete requests with status deleting")
 
-		p.inProcessRequests[deleteRequest.UserID] = deleteRequest
+		p.inProcessRequests.add(deleteRequest.UserID, &deleteRequest)
```
pstibrany, line 393: Once again.

pkg/chunk/purger/purger.go
sandeepsukhani: Thanks for the catch! I totally missed it.

pkg/chunk/purger/purger.go
```diff
@@ -714,3 +691,56 @@ func makeDeleteRequestWithLogger(deleteRequest DeleteRequest, l log.Logger) dele
...
+type inProcessRequestsCollection struct {
+	requests map[string]*DeleteRequest
+	mtx      sync.RWMutex
+}
+
+func newInProcessRequestsCollection() *inProcessRequestsCollection {
+	return &inProcessRequestsCollection{requests: map[string]*DeleteRequest{}}
+}
+
+func (i *inProcessRequestsCollection) add(userID string, request *DeleteRequest) {
```
pracucci, line 706: I would call it `set()` because overrides any previous request (if any) and doesn't "add" the request to a list.

PR #2818, opened 2020-06-30
Metrics for enqueue/flush reason, and dequeue/flush outcome.

### LINKED ISSUE #2287: Add a counter for chunks removed from the flush queue but not flushed

pstibrany:
**What this PR does**: Re-added `cortex_ingester_flush_reasons` under the name `cortex_ingester_enqueued_series_total`.
Added `cortex_ingester_dequeued_series_total`, with outcome label. Outcome is a super set of flush reason.

Follow-up to #2802 and #2811.

Fixes #2287

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   * `cortex_bucket_stores_gate_queries_in_flight`
   * `cortex_bucket_stores_gate_duration_seconds`
 * [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_flushed_series_total`, and is now incremented during flush, not when series is enqueued for flushing. #2802, #2818
-* [CHANGE] Added `cortex_ingester_enqueued_series_total` with `reason` label and `cortex_ingester_dequeued_series_total` with `outcome` label (superset of reason). #2818
+* [CHANGE] Added `cortex_ingester_enqueued_series_total` with `reason` label (works the same as `cortex_ingester_flush_reasons` did before) and `cortex_ingester_dequeued_series_total` with `outcome` label (superset of reason). #2818
 * [CHANGE] Experimental Delete Series: Metric `cortex_purger_oldest_pending_delete_request_age_seconds` would track age of delete requests since they are over their cancellation period instead of their creation time. #2806
 * [CHANGE] Experimental TSDB: the store-gateway service is required in a Cortex cluster running with the experimental blocks storage. Removed the `-experimental.tsdb.store-gateway-enabled` CLI flag and `store_gateway_enabled` YAML config option. The store-gateway is now always enabled when the storage engine is `tsdb`. #2822
 * [CHANGE] Ingester: Chunks flushed via /flush stay in memory until retention period is reached. This affects `cortex_ingester_memory_chunks` metric. #2778
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   * `cortex_bucket_stores_gate_queries_concurrent_max`
   * `cortex_bucket_stores_gate_queries_in_flight`
   * `cortex_bucket_stores_gate_duration_seconds`
-* [CHANGE] Metric `cortex_ingester_flush_reasons` has been replaced with `cortex_ingester_flushing_enqueued_series_total` with `reason` label (works the same as `cortex_ingester_flush_reasons` did before) and `cortex_ingester_flushing_dequeued_series_total` with `outcome` label (superset of reason). #2802, #2818
+* [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_flushing_enqueued_series_total`, and new metric `cortex_ingester_flushing_dequeued_series_total` with `outcome` label (superset of reason) has been added. #2802, #2818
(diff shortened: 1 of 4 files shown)


pkg/ingester/metrics.go
```diff
@@ -187,8 +189,16 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
 			Name: "cortex_ingester_memory_chunks",
 			Help: "The total number of chunks in memory.",
 		}),
+		seriesEnqueuedForFlush: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+			Name: "cortex_ingester_enqueued_series_total",
```
pracucci, line 193: I've the feeling the metric name is too much generic. Have you considered something like `cortex_ingester_series_flushing_enqueued_total`?

pstibrany: Going with `cortex_ingester_flushing_enqueued_series_total` to keep `series` (unit of the metric) close to the end. Same for dequeued.

pkg/ingester/flush.go
```diff
@@ -241,19 +243,19 @@ func (i *Ingester) flushUserSeries(flushQueueIndex int, userID string, fp model.
...
-		return nil
+		return "NoSeries", nil
 	}
 
 	userState.fpLocker.Lock(fp)
 	reason := i.shouldFlushSeries(series, fp, immediate)
 	if reason == noFlush {
 		userState.fpLocker.Unlock(fp)
-		return nil
+		return reason.String(), nil
```
bboreham, line 258: This feels odd - calling a method on a constant value.

pkg/ingester/flush.go
```diff
@@ -241,19 +243,19 @@ func (i *Ingester) flushUserSeries(flushQueueIndex int, userID string, fp model.
 
 	userState, ok := i.userStates.get(userID)
 	if !ok {
-		return nil
+		return "NoUser", nil
```
bboreham, line 246: This means the possible values of label `outcome` are scattered around the code.
Why not return `flushReason` and extend it with `NoUser` and `NoSeries` ?

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@
   * `cortex_bucket_stores_gate_queries_concurrent_max`
   * `cortex_bucket_stores_gate_queries_in_flight`
   * `cortex_bucket_stores_gate_duration_seconds`
-* [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_series_flushed_total`, and is now incremented during flush, not when series is enqueued for flushing. #2802
+* [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_flushed_series_total`, and is now incremented during flush, not when series is enqueued for flushing. #2802, #2818
+* [CHANGE] Added `cortex_ingester_flushing_enqueued_series_total` with `reason` label (works the same as `cortex_ingester_flush_reasons` did before) and `cortex_ingester_flushing_dequeued_series_total` with `outcome` label (superset of reason). #2818
```
bboreham, line 11: Isn't `cortex_ingester_flushing_dequeued_series_total` also a superset of `cortex_ingester_flushed_series_total` now?
Do we need both?

pkg/ingester/flush.go
pstibrany: `noFlush.String()` didn't work for me, not sure why. But I will rework it to return `flushReason` instead, as you suggested elsewhere.

CHANGELOG.md
pstibrany: Not quite... `cortex_ingester_flushing_dequeued_series_total` will reflect possible flush errors, while `cortex_ingester_flushed_series_total` is incremented before flushing.

> Do we need both?
Likely not. I will remove `cortex_ingester_flushed_series_total`.

pracucci:
@bboreham Do you have any comment on the updated changes?

pkg/ingester/flush.go
```diff
@@ -146,6 +160,7 @@ func (i *Ingester) sweepSeries(userID string, fp model.Fingerprint, series *memo
 
 	flushQueueIndex := int(uint64(fp) % uint64(i.cfg.ConcurrentFlushes))
 	if i.flushQueues[flushQueueIndex].Enqueue(&flushOp{firstTime, userID, fp, immediate}) {
+		i.metrics.seriesEnqueuedForFlush.WithLabelValues(flush.String()).Inc()
 		util.Event().Log("msg", "add to flush queue", "userID", userID, "reason", flush, "firstTime", firstTime, "fp", fp, "series", series.metric, "nlabels", len(series.metric), "queue", flushQueueIndex)
 	}
```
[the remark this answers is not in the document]
pstibrany: It could signal that chunks are not flushed fast enough, but one can also see that from queue length.

PR #2819, opened 2020-07-01
Ingest samples older than 1h for block store

codesome:
**What this PR does**:

This PR is currently a super early draft for ingesting samples older than 1h. Lots of TODOs.

The fail in out-of-bound test shows that we are currently appending out-of-bound samples.

Fixes https://github.com/cortexproject/cortex/issues/2366

**Checklist**

- [x] Ingest samples older than 1h
- [x] Reload TSDBs on restart
- [x] Query samples from older TSDBs
- [x] Support query stream over old TSDBs
- [x] Automatically cleanup old TSDBs
- [x] Tests updated
- [ ] Documentation added

TODO in future PRs

- [ ] Metrics from new TSDBs
- [ ] Limits on the additional TSDB
- [ ] Automatically cleanup _stale_ TSDBs


bboreham:
Should probably clarify in the title this is for block store - chunks has ingested old samples forever.

pkg/storage/tsdb/config.go
```diff
@@ -244,3 +249,9 @@ func (cfg *BucketStoreConfig) Validate() error {
 func (cfg *Config) BlocksDir(userID string) string {
 	return filepath.Join(cfg.Dir, userID)
 }
+
+// BackfillBlocksDir returns the directory path where old TSDB blocks and wal should be
```
pracucci, line 253:
```suggestion
// BackfillBlocksDir returns the directory path where TSDB blocks and wal used for backfilling should be
```

pkg/storage/tsdb/config.go
```diff
@@ -146,6 +148,9 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.WALCompressionEnabled, "experimental.tsdb.wal-compression-enabled", false, "True to enable TSDB WAL compression.")
 	f.BoolVar(&cfg.StoreGatewayEnabled, "experimental.tsdb.store-gateway-enabled", false, "True if the Cortex cluster is running the store-gateway service and the querier should query the bucket store via the store-gateway.")
 	f.BoolVar(&cfg.FlushBlocksOnShutdown, "experimental.tsdb.flush-blocks-on-shutdown", false, "If true, and transfer of blocks on shutdown fails or is disabled, incomplete blocks are flushed to storage instead. If false, incomplete blocks will be reused after restart, and uploaded when finished.")
+
+	f.StringVar(&cfg.BackfillDir, "experimental.tsdb.backfill-dir", "backfill_tsdb", "Local directory to store backfill TSDBs in the ingesters.")
+	f.DurationVar(&cfg.BackfillLimit, "experimental.tsdb.backfill-limit", 6*time.Hour, "")
```
pracucci, line 153: Few comments:
1. I wouldn't call it "limit" but "max age" (so `backfill_max_age` and `.backfill-max-age`). Limit can be confused with our limits (like max series per user/metric).
2. Please add the flag description. I would say something like "Maximum accepted sample age by backfilling. 0 disables it."
3. I would support in the code a 0 value to disable it. Also, I would set a default to 0.

pkg/storage/tsdb/config.go
```diff
@@ -146,6 +148,9 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.WALCompressionEnabled, "experimental.tsdb.wal-compression-enabled", false, "True to enable TSDB WAL compression.")
 	f.BoolVar(&cfg.StoreGatewayEnabled, "experimental.tsdb.store-gateway-enabled", false, "True if the Cortex cluster is running the store-gateway service and the querier should query the bucket store via the store-gateway.")
 	f.BoolVar(&cfg.FlushBlocksOnShutdown, "experimental.tsdb.flush-blocks-on-shutdown", false, "If true, and transfer of blocks on shutdown fails or is disabled, incomplete blocks are flushed to storage instead. If false, incomplete blocks will be reused after restart, and uploaded when finished.")
+
+	f.StringVar(&cfg.BackfillDir, "experimental.tsdb.backfill-dir", "backfill_tsdb", "Local directory to store backfill TSDBs in the ingesters.")
```
pracucci, line 152: In the `Validate()` function please add a check to make sure this directory is set (non empty string) when the backfill limit/max age > 0. Add this case to the `Validate()` unit test too.

pkg/ingester/ingester_v2.go
```diff
@@ -1280,3 +1312,435 @@ func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
...
+
+	if err == nil {
+		err = app.Commit()
+	}
+
+	return err
+}
+
+// Assumes 1h bucket range for . TODO(codesome): protect stuff with locks.
+type backfillTSDBs struct {
```
pracucci, line 1421: This `ingester_v2.go` file is getting quite big. I would move `backfillTSDBs` to a dedicated `.go` file.

pkg/ingester/ingester_v2.go
```diff
@@ -1280,3 +1312,435 @@ func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
...
+	userStateMtx sync.RWMutex
```
pracucci, line 1422: I would name it `tsdbsMtx`

pkg/ingester/ingester_v2.go
```diff
@@ -1280,3 +1312,435 @@ func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
...
+	wg.Wait()
+	select {
+	case err := <-errC:
+		return nil, err
+	}
+
+	return result, nil
+}
+
+type tsdbBucket struct {
```
pracucci, line 1655: I would move this to the same new file containing `backfillTSDBs`

pkg/ingester/ingester_v2.go
```diff
@@ -1280,3 +1312,435 @@ func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
...
+	db                     *userTSDB
+	bucketStart, bucketEnd model.Time
```
pracucci, line 1657: Why not using `int64`? It's the data type used by TSDB.

pkg/ingester/ingester_v2.go
```diff
@@ -1280,3 +1312,435 @@ func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
...
+	lastAppend             time.Time
```
pracucci, line 1658: You already have this in `userTSDB`. We could remove this and use `userTSDB.lastUpdate`

pkg/ingester/ingester_v2.go
```diff
@@ -233,6 +238,13 @@ func (i *Ingester) startingV2(ctx context.Context) error {
 		return errors.Wrap(err, "opening existing TSDBs")
 	}
 
+	// Scan and open backfill TSDB's that already exist on disk.
+	// TODO(codesome): run this in parallel with opening main TSDBs
```
pracucci, line 242: Is this extra complexity worth? I'm not sure. Having backfilling TSDBs shouldn't be the normal case and I'm not sure adding extra complexity here is worth. I feel fine if we (concurrently) open all existing TSDBs first, and then we (concurrently) open all backfill TSDBs.

pkg/ingester/ingester_v2.go
```diff
@@ -405,13 +417,23 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 				}
 			}
 
+			cause := errors.Cause(err)
+			if cause == storage.ErrOutOfBounds && s.TimestampMs > db.Head().MaxTime()-i.cfg.TSDBConfig.BackfillLimit.Milliseconds() {
+				fmt.Println(s.TimestampMs, db.Head().MaxTime(), i.cfg.TSDBConfig.BackfillLimit.Milliseconds())
+				err := i.v2BackfillPush(userID, ts.Labels, s)
```
pracucci, line 423: It's not feasible to have 1 TSDB appender commit for every single sample. What I would do instead is:
1. While iterating `for _, s := range ts.Samples {`, I would keep a buffer with all the samples that can be backfilled
2. Below, we partition these backfill samples into buckets and we append all samples at once for each bucket

pkg/ingester/ingester_v2.go
```diff
@@ -1280,3 +1312,435 @@ func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
...
+			true, // Allow out of order uploads. It's fine in Cortex's context.
+		)
+	}
+
+	//i.TSDBState.tsdbMetrics.setRegistryForUser(userID, tsdbPromReg)
+	return userDB, nil
+}
+
+func (i *Ingester) v2BackfillPush(userID string, la []client.LabelAdapter, s client.Sample) error {
+	bucket, err := i.getOrCreateBackfillTSDB(userID, s.TimestampMs)
```
pracucci, line 1379: We shouldn't call `getOrCreateBackfillTSDB` for every single sample to append. As mentioned in another comment, I think would be better if we:
1. Get a list of all samples to backfill
2. Partition these samples into buckets
3. We ingest each bucket, 1 by 1, and for each bucket we "get or create" the backfill TSDB

pkg/ingester/ingester_v2.go
```diff
@@ -1280,3 +1312,435 @@ func (i *Ingester) v2FlushHandler(w http.ResponseWriter, _ *http.Request) {
 
 	w.WriteHeader(http.StatusNoContent)
 }
+
+// createBackfillTSDB creates a TSDB for a given userID and bucketName, and returns the created db.
+func (i *Ingester) createBackfillTSDB(userID, bucketName string, blockRange int64) (*userTSDB, error) {
```
pracucci, line 1317: There too much code duplicated with `createTSDB()`. We should refactor `createTSDB()` to do things once.

pkg/storage/tsdb/config.go
codesome: > I would support in the code a 0 value to disable it. Also, I would set a default to 0.

That was the plan going ahead. I will make these changes.

pkg/ingester/ingester_v2.go
codesome: I don't remember exactly why I used `model.Time`, maybe some methods in that that were useful. I will check again and change to `int64` if `model.Time` is not required.

pkg/ingester/ingester_v2.go
codesome: Yeah that sounds fine, we don't need additional complexity for this.

pkg/ingester/ingester_v2.go
codesome: That sounds like a good idea. I will make changes accordingly.

codesome:
The code looks more complex than I desire it to be. I am currently writing more units tests and once I am satisfied with that, I will spend some time in simplifying the code wherever possible before moving onto manual tests.

pkg/ingester/ingester_v2.go
```diff
@@ -1023,8 +1123,15 @@ func (i *Ingester) closeAllTSDB() {
 		}(userDB)
 	}
 
-	// Wait until all Close() completed
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		i.closeAllBackfillTSDBs()
+	}()
+
```
pracucci, line 1131: I think we should do it after `i.userStatesMtx.Unlock()`. Also to further simplify the code, I would run `wg.Wait()` to wait until all user TSDBs are closed, and then I would call `closeAllBackfillTSDBs()` outside of any go routine.

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,711 @@
...
+		}
+		if uploaded, err := db.shipper.Sync(context.Background()); err != nil {
+			level.Warn(util.Logger).Log("msg", "shipper failed to synchronize backfill TSDB blocks with the storage", "user", db.userID, "uploaded", uploaded, "bucket_dir", db.Dir(), "err", err)
+		} else {
+			level.Debug(util.Logger).Log("msg", "shipper successfully synchronized backfill TSDB blocks with storage", "user", db.userID, "uploaded", uploaded, "bucket_dir", db.Dir())
+		}
+	})
+}
+
+func (i *Ingester) runConcurrentBackfillWorkers(ctx context.Context, concurrency int, userFunc func(*userTSDB)) {
```
pracucci, line 398: I think `runConcurrentBackfillWorkers()` and `runConcurrentUserWorkers()` could be generalised into a single function like this:

```
func runConcurrentUserWorkers(ctx context.Context, userIDs []string, concurrency int, userFunc func(userID string))
```

Then:
- You need two functions to generate the list of user IDs. We already have `i.getTSDBUsers()` then you could do the same for the backfill TSDBs.
- The callback passed to `runConcurrentBackfillWorkers()` but looks easy passing the userID to the callback and then having the callback fetching the backfilling TSDBs. I understand the actual parallelisation is not the exact same, but we should strive for simplicity first (without compromising correctness) and then we can always optimise it. As already mentioned a bunch of times, I expect backfilling to be an uncommon use case (I don't want to keep multiple TSDBs open for every tenant any time)

_This is a refactoring that could be done in a preliminary PR._

pkg/ingester/ingester_v2.go
```diff
@@ -460,13 +503,27 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 				}
 			}
 
+			cause := errors.Cause(err)
+			if cause == storage.ErrOutOfBounds &&
+				i.cfg.BlocksStorageConfig.TSDB.BackfillLimit > 0 &&
+				s.TimestampMs > db.Head().MaxTime()-i.cfg.BlocksStorageConfig.TSDB.BackfillLimit.Milliseconds() {
```
pracucci, line 509: 1. There could be any case when `db.Head().MaxTime()` `math.MinInt64`?
2. But should be based on head max time or `time.Now()`? The "max age" could be also see an compared to "now". All other time-based limits we have are based on "now".

pkg/ingester/ingester_v2.go
```diff
@@ -495,8 +552,13 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 			}
 
 			// The error looks an issue on our side, so we should rollback
-			if rollbackErr := app.Rollback(); rollbackErr != nil {
-				level.Warn(util.Logger).Log("msg", "failed to rollback on error", "user", userID, "err", rollbackErr)
+			var merr tsdb_errors.MultiError
```
pracucci, line 555: Renaming `merr` to `rollbackErr` as was previously could help to clarify what this error is about.

codesome:
Superseded by https://github.com/cortexproject/cortex/pull/3025

PR #2822, opened 2020-07-01
Make store-gateway mandatory when running the blocks storage

pracucci:
**What this PR does**:
As of today, it's not mandatory to run the store-gateway when using the experimental blocks storage. However, over the time we've made many improvements to the blocks storage when using the store-gateway and running a blocks storage cluster without it is discouraged.

In this PR I'm proposing to do the next step and remove the bucket store support from the querier, which means the store-gateway becomes a required component when running the blocks storage.

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added


docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
 
-The in-memory samples are periodically flushed to disk - and the WAL truncated - when a new TSDB Block is cut, which by default occurs every 2 hours. Each new Block cut is then uploaded to the long-term storage and kept in the ingester for some more time, in order to give queriers enough time to discover the new Block from the storage and download its index header.
+The in-memory samples are periodically flushed to disk - and the WAL truncated - when a new TSDB Block is cut, which by default occurs every 2 hours. Each new Block cut is then uploaded to the long-term storage and kept in the ingester for some more time, in order to give queriers enough time to discover the new Block from the storage and download its index-header.
```
pstibrany, line 33:
```suggestion
The in-memory samples are periodically flushed to disk - and the WAL truncated - when a new TSDB Block is created, which by default occurs every 2 hours. Each newly created Block is then uploaded to the long-term storage and kept in the ingester for some more time, in order to give queriers and store-gateways enough time to discover the new Block on the storage and download its index-header.
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
 
 In order to effectively use the **WAL** and being able to recover the in-memory series upon ingester abruptly termination, the WAL needs to be stored to a persistent local disk which can survive in the event of an ingester failure (ie. AWS EBS volume or GCP persistent disk when running in the cloud). For example, if you're running the Cortex cluster in Kubernetes, you may use a StatefulSet with a persistent volume claim for the ingesters.
 
 ### The read path
 
-**Queriers** - at startup - iterate over the entire storage bucket to discover all tenants Blocks and - for each of them - download the index header. During this initial synchronization phase, a querier is not ready to handle incoming queries yet and its `/ready` readiness probe endpoint will fail.
+**Store-gateways** - at startup - iterate over the entire storage bucket to discover all tenants Blocks and - for each of them - download the `meta.json` and index-header. During this initial bucket synchronization phase, the store-gateway `/ready` readiness probe endpoint will fail.
```
pstibrany, line 39:
```suggestion
At startup **store-gateways** iterate over the entire storage bucket to discover all tenants Blocks and download the `meta.json` and index-header for each Block. During this initial bucket synchronization phase, the store-gateway `/ready` readiness probe endpoint will fail.
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
 
-Queriers also periodically re-iterate over the storage bucket to discover newly uploaded Blocks (by the ingesters) and find out Blocks deleted in the meanwhile, as effect of an optional retention policy.
+Similarly, **queriers** - at startup - iterate over the entire storage bucket to discover all tenants Blocks and - for each of them - download the `meta.json`. During this initial bucket scanning phase, a querier is not ready to handle incoming queries yet and its `/ready` readiness probe endpoint will fail.
```
pstibrany, line 41:
```suggestion
Similarly, **queriers** also iterate over the entire storage bucket at startup, to discover all tenants Blocks and download the `meta.json` for each Block. During this initial bucket scanning phase, a querier is not ready to handle incoming queries yet and its `/ready` readiness probe endpoint will fail.
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
 
-The blocks chunks and the entire index is never fully downloaded by the queriers. In the read path, a querier lookups the series label names and values using the in-memory index header and then download the required segments of the index and chunks for the matching series directly from the long-term storage using byte-range requests.
+Store-gateways and queriers also periodically re-iterate over the storage bucket to discover newly uploaded Blocks (by the ingesters) and find out Blocks marked for deletion or hard deleted in the meanwhile, as effect of compaction or an optional retention policy. The frequency at which this occurs is configured via `-experimental.tsdb.bucket-store.sync-interval`.
```
pstibrany, line 43:
```suggestion
Store-gateways and queriers also periodically re-iterate over the storage bucket to discover newly uploaded Blocks (by the ingesters) and find out Blocks marked for deletion or hard deleted in the meantime, as a result of compaction or an optional retention policy. The frequency at which this occurs is configured via `-experimental.tsdb.bucket-store.sync-interval`.
```

pstibrany:
> Based on title of PR, I didn't expect to see such massive changes to documentation.

Don't get me wrong, I really like the improvements :) (I would question using Blocks with capital B though)

pracucci:
> Don't get me wrong, I really like the improvements :)

Thanks @pstibrany for your review! Accepted all suggestions.

> (I would question using Blocks with capital B though)

Me too. Was a feedback originally received by Josh but looks a bit weird to me. Will rework it during the next documentation refactoring (planned for today).

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
 
 In order to effectively use the **WAL** and being able to recover the in-memory series upon ingester abruptly termination, the WAL needs to be stored to a persistent local disk which can survive in the event of an ingester failure (ie. AWS EBS volume or GCP persistent disk when running in the cloud). For example, if you're running the Cortex cluster in Kubernetes, you may use a StatefulSet with a persistent volume claim for the ingesters.
 
 ### The read path
 
-**Queriers** - at startup - iterate over the entire storage bucket to discover all tenants Blocks and - for each of them - download the index header. During this initial synchronization phase, a querier is not ready to handle incoming queries yet and its `/ready` readiness probe endpoint will fail.
+At startup **store-gateways** iterate over the entire storage bucket to discover all tenants Blocks and download the `meta.json` and index-header for each Block. During this initial bucket synchronization phase, the store-gateway `/ready` readiness probe endpoint will fail.
```
pstibrany, line 39:
```suggestion
At startup **store-gateways** iterate over the entire storage bucket to discover blocks for all tenants and download the `meta.json` and index-header for each Block. During this initial bucket synchronization phase, the store-gateway `/ready` readiness probe endpoint will fail.
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
+Similarly, **queriers** also iterate over the entire storage bucket at startup, to discover all tenants Blocks and download the `meta.json` for each Block. During this initial bucket scanning phase, a querier is not ready to handle incoming queries yet and its `/ready` readiness probe endpoint will fail.
 
-The blocks chunks and the entire index is never fully downloaded by the queriers. In the read path, a querier lookups the series label names and values using the in-memory index header and then download the required segments of the index and chunks for the matching series directly from the long-term storage using byte-range requests.
+Store-gateways and queriers also periodically re-iterate over the storage bucket to discover newly uploaded Blocks (by the ingesters) and find out Blocks marked for deletion or hard deleted in the meantime, as a result of compaction or an optional retention policy. The frequency at which this occurs is configured via `-experimental.tsdb.bucket-store.sync-interval`.
```
pstibrany, line 43:
```suggestion
Store-gateways and queriers periodically rescan the storage bucket to discover new Blocks (uploaded by the ingesters and compactor) and Blocks marked for deletion or fully deleted since last scan (as a result of compaction or an optional retention policy). The frequency at which this occurs is configured via `-experimental.tsdb.bucket-store.sync-interval`.
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
+
+Blocks can be optionally sharded and replicated across a pool of store-gateway instances. This feature can be enabled via `-experimental.store-gateway.sharding-enabled=true` and requires the  backend ring to be configured via `-experimental.store-gateway.sharding-ring.*` flags (or their respective YAML config options).
+
+When blocks sharding is **enabled**, store-gateway instances builds a ring and blocks are sharded accordingly. Queriers also need to have the same store-gateway sharding config in order to be able to correctly address blocks across the pool of store-gateways.
+
+When blocks sharding is **disabled**, queriers need the `-experimental.querier.store-gateway-addresses` CLI flag (or its respective YAML config option) being set to a comma separated list of store-gateway addresses in [DNS Service Discovery format]((../configuration/arguments.md#dns-service-discovery). Queriers will evenly balance the requests to query blocks across the resolved addresses.
+
+### Anatomy of a query execution
+
+When a querier receives a query range request, it contains few main information:
```
pstibrany, line 63:
```suggestion
When a querier receives a query range request, it contains following parameters:
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
+
+- `query`: the PromQL query expression itself (e.g. `rate(node_cpu_seconds_total[1m])`)
+- `start`: the start time
+- `end`: the end time
+- `step`: the query resolution (e.g. `30000` ms to have 1 resulting data point every 30s)
```
pstibrany, line 68: Step sent to Prometheus API endpoint is in seconds by default, but can be specified in "duration" format as well.

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
+
+The request sent to each store-gateway instance contains the list of Block IDs that are expected to be queried, and the response sent by the store-gateway back to the querier contains the list of Block IDs actually queried. The set of Blocks actually queried by a store-gateway may be a subset of the requested Blocks as effect of a Blocks re-sharding recently occurred (ie. last few seconds). The querier runs a consistency check on responses received from the store-gateways to ensure all expected Blocks have been queried; if not, the querier retries to fetch missing Blocks from different store-gateways (if the `-experimental.store-gateway.replication-factor` is greater than `1`) and if the consistency check fails after all retries, the query execution will fail as well (correctness is always guaranteed).
```
pstibrany, line 72:
```suggestion
The request sent to each store-gateway contains the list of Block IDs that are expected to be queried, and the response sent back by the store-gateway to the querier contains the list of Block IDs that were actually queried. This list may be a subset of the requested Blocks, for example due to recent Blocks resharding event (ie. last few seconds). The querier runs a consistency check on responses received from the store-gateways to ensure all expected Blocks have been queried; if not, the querier retries to fetch samples from missing Blocks from different store-gateways (if the `-experimental.store-gateway.replication-factor` is greater than `1`) and if the consistency check fails after all retries, the query execution fails as well (correctness is always guaranteed).
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
+
+Alongside the requests sent to the store-gateway, if the query time range covers a period within `-querier.query-ingesters-within`, the querier also sends a request to ingesters, in order to fetch samples which have not been persistent into a Block uploaded to the long-term storage yet.
```
pstibrany, line 74:
```suggestion
If the query time range covers a period within `-querier.query-ingesters-within` duration, the querier also sends the request to all ingesters, in order to fetch samples that have not been uploaded to the long-term storage yet.
```

docs/operations/blocks-storage.template
```diff
@@ -22,32 +22,59 @@ The rest of the document assumes you have read the [Cortex architecture](../arch
...
+
+Once all samples have been fetched from both store-gateways and ingesters, the querier proceed running the PromQL engine to execute the query and send back the result to the client.
```
pstibrany, line 76:
```suggestion
Once all samples have been fetched from both store-gateways and ingesters, the querier proceeds with running the PromQL engine to execute the query and send back the result to the client.
```

docs/operations/blocks-storage.template
```diff
@@ -57,7 +84,7 @@ The **compactor** is an optional - but highly recommended - service which compac
 
 The **vertical compaction** compacts all the Blocks of a tenant uploaded by any ingester for the same Block range period (defaults to 2 hours) into a single Block, de-duplicating samples that are originally written to N Blocks as effect of the replication.
 
-The **horizontal compaction** triggers after the vertical compaction and compacts several Blocks belonging to adjacent small range periods (2 hours) into a single larger Block. Despite the total block chunks size doesn't change after this compaction, it may have a significative impact on the reduction of the index size and its index header kept in memory by queriers.
+The **horizontal compaction** triggers after the vertical compaction and compacts several Blocks belonging to adjacent small range periods (2 hours) into a single larger Block. Despite the total block chunks size doesn't change after this compaction, it may have a significative impact on the reduction of the index size and its index-header kept in memory by queriers.
```
pstibrany, line 87:
```suggestion
The **horizontal compaction** triggers after the vertical compaction and compacts several Blocks with adjacent 2-hour range periods into a single larger Block. Even though the total size of block chunks doesn't change after this compaction, it may still significantly reduce the size of the index and index-header kept in memory by queriers and store-gateways.
```

docs/operations/blocks-storage.template
```diff
@@ -74,7 +101,7 @@ Whenever the pool of compactors increase or decrease (ie. following up a scale u
 
 ## Index cache
 
-The querier and store-gateway support a cache to speed up postings and series lookups from TSDB blocks indexes. Two backends are supported:
+The store-gateway support a cache to speed up postings and series lookups from TSDB blocks indexes. Two backends are supported:
```
pstibrany, line 104:
```suggestion
The store-gateway can use a cache to speed up lookups of postings and series from TSDB blocks indexes. Two backends are supported:
```

docs/operations/blocks-storage.template
```diff
@@ -105,15 +132,15 @@ For example, if you're running Memcached in Kubernetes, you may:
 
 ## Chunks cache
 
-Store-gateway and querier also support cache for storing chunks fetched from storage. Chunks contain actual samples, and can be reused if user query hits the same series for the same time range.
+Store-gateway also support cache for storing chunks fetched from storage. Chunks contain actual samples, and can be reused if user query hits the same series for the same time range.
```
pstibrany, line 135:
```suggestion
Store-gateway can also use a cache for storing chunks fetched from the storage. Chunks contain actual samples, and can be reused if user query hits the same series for the same time range.
```

PR #2826, opened 2020-07-02
Add -store.max-query-length support to blocks storage

pracucci:
**What this PR does**:
The `-store.max-query-length` is currently enforced in the query-frontend (before splitting) and in the chunks store. In this PR I propose to validate it in the querier so that it applies both to the chunks and blocks storage.

However, I've noticed the limit in the chunks store is also used by Loki, reason why I haven't removed it from there. /cc @cyriltovena 

WDYT?

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/util/validation/limits.go
+++ b/pkg/util/validation/limits.go
@@ -101,7 +101,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.MaxGlobalMetadataPerMetric, "ingester.max-global-metadata-per-metric", 0, "The maximum number of metadata per metric, across the cluster. 0 to disable.")
 
 	f.IntVar(&l.MaxChunksPerQuery, "store.query-chunk-limit", 2e6, "Maximum number of chunks that can be fetched in a single query.")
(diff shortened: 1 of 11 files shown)


pkg/util/validation/limits.go
```diff
@@ -101,7 +101,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.MaxGlobalMetadataPerMetric, "ingester.max-global-metadata-per-metric", 0, "The maximum number of metadata per metric, across the cluster. 0 to disable.")
 
 	f.IntVar(&l.MaxChunksPerQuery, "store.query-chunk-limit", 2e6, "Maximum number of chunks that can be fetched in a single query.")
-	f.DurationVar(&l.MaxQueryLength, "store.max-query-length", 0, "Limit to length of chunk store queries, 0 to disable.")
+	f.DurationVar(&l.MaxQueryLength, "store.max-query-length", 0, "Limit the query time range length (end - start time). This limit is enforced in the query-frontend (on the received query), in the querier (on the query eventually splitted by the query-frontend) and in the chunks storage. 0 to disable.")
```
pstibrany, line 104:
```suggestion
	f.DurationVar(&l.MaxQueryLength, "store.max-query-length", 0, "Limit the query time range length (end - start time). This limit is enforced in the query-frontend (on the received query), in the querier (on the query possibly split by the query-frontend) and in the chunks storage. 0 to disable.")
```

PR #2829, opened 2020-07-02
Add REWE digital case-study

gouthamve:
We did this with @weeco and I think this is great!!

Martin, please run the website using: https://cortexmetrics.io/docs/contributing/how-to-run-the-website-locally/ and let us know if it looks good? Once you give an LGTM, I'll merge this! Thanks for all your work on this!


weeco:
I ran it locally as well, LGTM

PR #2831, opened 2020-07-02
S3 Config Improvements

joe-elliott:
cc @jtlisi 

**Note**: There is still a lot of work to do on this PR (docs, changelog, yaml/cli parsing, tests, etc), but I wanted to a draft PR up so the maintainers would know it was in progress.

**What this PR does**:
- Extends the s3 client to support a more ergonomic config while remaining backwards compatible
- Adds the ability to configure SSE, UserMetaData, IdleConnTimeout and ResponseHeaderTimeout

Fixes #Nothing

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/chunk/aws/s3_storage_client.go
```diff
@@ -56,42 +76,122 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
+	if cfg.S3.URL != nil {
+		bucketNames = []string{strings.TrimPrefix(cfg.S3.URL.Path, "/")}
+	}
+
+	if cfg.BucketNames != "" {
+		bucketNames = strings.Split(cfg.BucketNames, ",") // comma separated list of bucket names
+	}
+
+	if len(bucketNames) == 0 {
+		return nil, nil, errors.New("Failed to initialize s3_storage_client:  At least one bucket name must be specified")
```
jtlisi, line 189: When returning errors in this function, I would just include the second half of the error and wrap/return the first half in the calling function.

pkg/chunk/aws/s3_storage_client_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/cortexproject/cortex/integration/e2e"
+	e2edb "github.com/cortexproject/cortex/integration/e2e/db"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
+)
+
+func TestS3Client(t *testing.T) {
+	networkName := "s3-test"
+	bucketName := "test"
+
+	s, err := e2e.NewScenario(networkName)
```
jtlisi, line 22: This test case should be in the `integrations/` directory. Otherwise the fixtures it relies on won't be configured correctly.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -38,8 +43,24 @@ func init() {
...
 	S3ForcePathStyle bool
+
+	BucketNames     string
+	Endpoint        string             `yaml:"endpoint"`
+	Region          string             `yaml:"region"`
+	AccessKeyID     string             `yaml:"access_key_id"`
+	SecretAccessKey string             `yaml:"secret_access_key"`
+	Insecure        bool               `yaml:"insecure"`
+	SSEEncryption   bool               `yaml:"sse_encryption"`
+	PutUserMetadata map[string]*string `yaml:"put_user_metadata"`
```
pracucci, line 55: How is this intended to be used?

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -38,8 +43,24 @@ func init() {
...
+	HTTPConfig      HTTPConfig         `yaml:"http_config"`
```
pracucci, line 56: Is it worth a nested config struct instead of embedding the 3 config options here?

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -53,43 +74,133 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
+	putUserMetadata map[string]*string
 }
 
 // NewS3ObjectClient makes a new S3-backed ObjectClient.
 func NewS3ObjectClient(cfg S3Config, delimiter string) (*S3ObjectClient, error) {
-	if cfg.S3.URL == nil {
-		return nil, fmt.Errorf("no URL specified for S3")
+	s3Config, bucketNames, err := buildS3Config(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to build s3 config: %v", err)
```
pracucci, line 101: Golang best practice is to never capitalise error messages. Could you change this and the error below, please?

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -53,43 +74,133 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
+		s3Config = s3Config.WithDisableSSL(true)
+	}
+
+	if cfg.Region != "" {
+		s3Config = s3Config.WithRegion(cfg.Region)
+	}
+
+	if cfg.AccessKeyID != "" && cfg.SecretAccessKey == "" ||
+		cfg.AccessKeyID == "" && cfg.SecretAccessKey != "" {
+		return nil, nil, errors.New("Must supply both an Access Key ID and Secret Access Key or neither")
```
pracucci, line 160: Lowercase first letter in the error message

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -53,43 +74,133 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
+	s3Config = s3Config.WithHTTPClient(&http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+				DualStack: true,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       cfg.HTTPConfig.IdleConnTimeout, //90 * time.Second,
```
pracucci, line 180: I would remove the comment

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -53,43 +74,133 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
+			return nil, nil, err
+		}
+	} else {
+		s3Config = &aws.Config{}
+		s3Config = s3Config.WithRegion("dummy")
+		s3Config = s3Config.WithCredentials(credentials.AnonymousCredentials)
+	}
 
 	s3Config = s3Config.WithMaxRetries(0) // We do our own retries, so we can monitor them
 	s3Config = s3Config.WithHTTPClient(&http.Client{Transport: defaultTransport})
```
pracucci, line 143: Do we need this? We're going to override it below.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -53,43 +74,133 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+				DualStack: true,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       cfg.HTTPConfig.IdleConnTimeout, //90 * time.Second,
+			MaxIdleConnsPerHost:   100,
+			TLSHandshakeTimeout:   3 * time.Second,
```
pracucci, line 182: Above you mentioned we use "golang defaults", but this is not the default.

pkg/chunk/aws/s3_storage_client.go
joe-elliott: I'm going to remove this option.  It can be added later if we want it.  It was still on my list to test how yaml would unmarshal to `map[string]*string`.  My current answer is:  I don't know :).

I'd rather get this PR in and not worry about this field, then add appropriate testing around this.

pkg/chunk/aws/s3_storage_client.go
joe-elliott: I don't have strong feelings here.  This was admittedly modeled after the thanos s3 client (https://github.com/thanos-io/thanos/tree/master/pkg/objstore/s3).  

If you'd like me to embed them in the S3Config struct I'd be glad to.

pkg/chunk/aws/s3_storage_client.go
joe-elliott: Adjusted the comment to clarify where that config came from and why it exists.

joe-elliott:
Dropping draft status.  There are still some open questions, but I think this is feature complete.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -53,45 +73,134 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
 	}
 
-	s3Config = s3Config.WithS3ForcePathStyle(cfg.S3ForcePathStyle) // support for Path Style S3 url if has the flag
-
-	s3Config = s3Config.WithMaxRetries(0) // We do our own retries, so we can monitor them
-	s3Config = s3Config.WithHTTPClient(&http.Client{Transport: defaultTransport})
 	sess, err := session.NewSession(s3Config)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create new s3 session: %v", err)
```
pracucci, line 105:
```suggestion
		return nil, errors.Wrap(err, "failed to create new s3 session")
```

integration/e2e/s3_storage_client_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/cortexproject/cortex/integration/e2e"
+	e2edb "github.com/cortexproject/cortex/integration/e2e/db"
+	s3 "github.com/cortexproject/cortex/pkg/chunk/aws"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
+)
+
+func TestS3Client(t *testing.T) {
+	networkName := "s3-test"
```
pracucci, line 21: As of today there's no need to define a dedicate network. You can just reuse the global `networkName`.

integration/e2e/s3_storage_client_test.go
```diff
@@ -0,0 +1,105 @@
...
+	bucketName := "test"
```
pracucci, line 22: Same here. You could just use the global `bucketName`.

PR #2837, opened 2020-07-05
Fix a bug about failure to add metrics of multi KV store

### LINKED ISSUE #2807: Failure to add metrics of multi KV store on startup with release-1.2
I have tried to run Cortex with release-1.2 branch.
But I found the error when starting Cortex with configurations below.

**Configs**

```yaml
...
"ingester":
  "lifecycler":
    "ring":
      "kvstore":
        "store": "multi"
        "consul":
          "host": "consul.host:8500"
        "multi":
          "primary": "consul"
          "secondary": "memberlist"
...
```

**Error logs**

```
panic: a previously registered descriptor with the same fully-qualified name as Desc{fqName: "cortex_kv_request_duration_seconds", help: "Time spent on kv store requests.", constLabels: {kv_name="ingester-ring",type="multi"}, variableLabels: [operation status_code]} has different label names or a different help string

goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*wrappingRegisterer).MustRegister(0xc000817b30, 0xc0003fa980, 0x1, 0x1)
	/my/path/cortex/vendor/github.com/prometheus/client_golang/prometheus/wrap.go:102 +0x194
github.com/prometheus/client_golang/prometheus/promauto.Factory.NewHistogramVec(0x5a04e80, 0xc000817b30, 0x509ffbf, 0x6, 0x0, 0x0, 0x50df2ab, 0x1b, 0x50f141d, 0x20, ...)
	/my/path/cortex/vendor/github.com/prometheus/client_golang/prometheus/promauto/auto.go:362 +0x158
github.com/cortexproject/cortex/pkg/ring/kv.newMetricsClient(0xc000453e30, 0x5, 0x5a487c0, 0xc000b6c5c0, 0x5a04e80, 0xc000817b30, 0x0, 0x0)
	/my/path/cortex/pkg/ring/kv/metrics.go:43 +0x251
github.com/cortexproject/cortex/pkg/ring/kv.createClient(0xc000453e30, 0x5, 0x50ab0f0, 0xb, 0xc0007f6520, 0x1a, 0x0, 0x0, 0x4a817c800, 0x0, ...)
	/my/path/cortex/pkg/ring/kv/client.go:159 +0x2c4
```

kwSeo:
Signed-off-by: Kyeongwon Seo <ruddnjs1230@gmail.com>

**What this PR does**:
This PR fixes a bug about failure to add metrics of multi KV store.
The multi KV store does not have to register metrics.
Because the metrics are registered when creating primary and secondary KV stores.

Fixes #2807

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [CHANGE] Experimental TSDB: the store-gateway service is required in a Cortex cluster running with the experimental blocks storage. Removed the `-experimental.tsdb.store-gateway-enabled` CLI flag and `store_gateway_enabled` YAML config option. The store-gateway is now always enabled when the storage engine is `tsdb`. #2822
 * [CHANGE] Ingester: Chunks flushed via /flush stay in memory until retention period is reached. This affects `cortex_ingester_memory_chunks` metric. #2778
 * [CHANGE] Querier: the error message returned when the query time range exceeds `-store.max-query-length` has changed from `invalid query, length > limit (X > Y)` to `the query time range exceeds the limit (query length: X, limit: Y)`. #2826
+* [CHANGE] KV: The `role` label which was a label of `multi` KV store client only has been added to metrics of every KV store client. If KV store client is not `multi`, then the value of `role` label is `primary`. #2837
 * [FEATURE] Introduced `ruler.for-outage-tolerance`, Max time to tolerate outage for restoring "for" state of alert. #2783
 * [FEATURE] Introduced `ruler.for-grace-period`, Minimum duration between alert and restored "for" state. This is maintained only for alerts with configured "for" time greater than grace period. #2783
 * [FEATURE] Introduced `ruler.resend-delay`, Minimum amount of time to wait before resending an alert to Alertmanager. #2783
@@ -47,7 +48,7 @@
 * [BUGFIX] Ingester: Flushing chunks via `/flush` endpoint could previously lead to panic, if chunks were already flushed before and then removed from memory during the flush caused by `/flush` handler. Immediate flush now doesn't cause chunks to be flushed again. Samples received during flush triggered via `/flush` handler are no longer discarded. #2778
 * [BUGFIX] Prometheus upgraded. #2849
   * Fixed unknown symbol error during head compaction
-* [BUGFIX] Fixed a bug about failure to add metrics of multi KV store. #2837
+* [BUGFIX] KV: Fixed a bug that triggered a panic due to metrics being registered with the same name but different labels when using a `multi` configured KV client. #2837
 
 ## 1.2.0 / 2020-07-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 * [BUGFIX] Ingester: Flushing chunks via `/flush` endpoint could previously lead to panic, if chunks were already flushed before and then removed from memory during the flush caused by `/flush` handler. Immediate flush now doesn't cause chunks to be flushed again. Samples received during flush triggered via `/flush` handler are no longer discarded. #2778
 * [BUGFIX] Prometheus upgraded. #2849
   * Fixed unknown symbol error during head compaction
+* [BUGFIX] Fixed a bug about failure to add metrics of multi KV store. #2837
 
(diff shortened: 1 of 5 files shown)


kwSeo:
I agree @jtlisi.
I think metrics with the same name would be better to have the same label name set.

jtlisi:
@kwSeo 👍 feel free to ping me when you want a second look

pracucci:
> I don't think this will fix the bug. This will work if you are running Cortex with a single ring. However, if you are running Cortex with a multi ring and a regular ring, this will trigger the same error.

@jtlisi If I understand what you mena, I'm not sure this is correct because we do wrap every KV registerer with a different `kv.RegistererWithKVName()`. Anyway, your proposed solution looks better so I'm all in!

pkg/ring/kv/client.go
```diff
@@ -139,6 +154,10 @@ func createClient(backend string, prefix string, cfg StoreConfig, codec codec.Co
 	case "multi":
 		client, err = buildMultiClient(cfg, codec, reg)
 
+	// this case is for testing. The mock KV client does not anything internally.
```
pracucci, line 157:
```suggestion
	// This case is for testing. The mock KV client does not do anything internally.
```

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 * [BUGFIX] Ingester: Flushing chunks via `/flush` endpoint could previously lead to panic, if chunks were already flushed before and then removed from memory during the flush caused by `/flush` handler. Immediate flush now doesn't cause chunks to be flushed again. Samples received during flush triggered via `/flush` handler are no longer discarded. #2778
 * [BUGFIX] Prometheus upgraded. #2849
   * Fixed unknown symbol error during head compaction
+* [BUGFIX] Fixed a bug about failure to add metrics of multi KV store. #2837
```
jtlisi, line 50:
```suggestion
* [BUGFIX] KV:  Fixed a bug that triggered a panic due to metrics being registered with the same name but different labels when using a `multi` configured KV client. #2837
```

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 * [BUGFIX] Ingester: Flushing chunks via `/flush` endpoint could previously lead to panic, if chunks were already flushed before and then removed from memory during the flush caused by `/flush` handler. Immediate flush now doesn't cause chunks to be flushed again. Samples received during flush triggered via `/flush` handler are no longer discarded. #2778
```
jtlisi, line 47: Can you add a `[CHANGE]` entry as well explaining the new label behaviour on metrics on the KV store.

pkg/ring/kv/client.go
kwSeo: Thanks! I fixed it.

CHANGELOG.md
kwSeo: Thanks!
I rewrote the change log with the your suggested log.

CHANGELOG.md
kwSeo: I added a `[CHANGE]` log that explain the `role` label to CHANGELOG.md

kwSeo:
@jtlisi @pracucci 
Thanks for your feedback.
I have updated my codes.
The mock KV client remains to test multi backend KV store client...

kwSeo:
@jtlisi
If I understand correctly, the current implementation of `multi` backend does not allow the same backend.
So two `inmemory` backends as primary and secondary could not be used.

pracucci:
> @jtlisi
> If I understand correctly, the current implementation of `multi` backend does not allow the same backend.
> So two `inmemory` backends as primary and secondary could not be used.

That's correct, reason why (even if I don't like "mock" very much) I haven't been able to suggest a different approach. I've also thought about skipping the check that multi backends must be different when running tests (eg. skip the check if they're both "inmemory") but the resulting test would be weaker because you wouldn't be able to ensure the "secondary" role label has been attached to the secondary backend, given both backends are "inmemory".

jtlisi:
@kwSeo Forgive me, I didn't realize it was not possible to use two `inmemory` backends. This is ready with the mock in that case. 👍

PR #2841, opened 2020-07-06
fix a flaky test in purger

### LINKED ISSUE #2840: Flake in TestPurger_Metrics
While running the CI on master I've seen this flaky test ([details here](https://app.circleci.com/pipelines/github/cortexproject/cortex/6838/workflows/f024cd69-9fb8-4ac0-99b5-93a8250d9d66/jobs/34385/steps)):

```
--- FAIL: TestPurger_Metrics (2.16s)
    purger_test.go:421: 1 != 0
FAIL
```

sandeepsukhani:
**What this PR does**:
In `TestPurger_Metrics` test I was starting the purger service [here](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/purger/purger_test.go#L418) after pulling the delete requests for processing [here](https://github.com/cortexproject/cortex/blob/master/pkg/chunk/purger/purger_test.go#L411) which usually should never happen because we always start the service first. This was done to capture an expected metric value.
This was causing Purger to work in an unexpected way causing it to fail the test sometimes. I have changed the test to stop existing purger and create a new one to start the service for delete request processing.

Fixes #2840 

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/purger/purger_test.go
+++ b/pkg/chunk/purger/purger_test.go
@@ -52,8 +52,6 @@ func setupTestDeleteStore(t *testing.T) *DeleteStore {
 }
 
 func setupStoresAndPurger(t *testing.T) (*DeleteStore, chunk.Store, chunk.ObjectClient, *Purger, *prometheus.Registry) {
-	registry := prometheus.NewRegistry()
-
 	deleteStore := setupTestDeleteStore(t)
 
 	chunkStore, err := testutils.SetupTestChunkStore()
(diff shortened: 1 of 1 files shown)


pkg/chunk/purger/purger_test.go
```diff
@@ -414,9 +417,17 @@ func TestPurger_Metrics(t *testing.T) {
 	require.InDelta(t, float64(2*86400), testutil.ToFloat64(purger.metrics.oldestPendingDeleteRequestAgeSeconds), 1)
 	require.Equal(t, float64(2), testutil.ToFloat64(purger.metrics.pendingDeleteRequestsCount))
 
-	// start loop to process requests
+	// stop the existing purger
+	require.NoError(t, services.StopAndAwaitTerminated(context.Background(), purger))
```
pracucci, line 421: At the beginning of the test there's a deferred call to `purger.StopAsync()`. Given we're now terminating it here, I believe that call should be removed.

PR #2851, opened 2020-07-07
Support multiple Alertmanagers explicitly in the Ruler

gotjosh:
**What this PR does**:

At the moment, the Ruler only support sending alerts to multiple
Alertmanagers via DNS based service discovery. However, sending to
multiple Alertmanager groups is something that Prometheus allows.

I believe the Ruler should support this too and this commit introduces
that.

To keep backward compatibility we're reusing the same flag but allowing
a list of 1+. With this, we can treat each URL as an Alertmanager group
where multiple Alertmanagers per group is only supported if DNS service
discovery is enabled.

~~*Please note that this would be effectively a breaking change on the ruler flags. In the docs, the ruler API is currently listed as experimental but not the ruler itself. I assume meant we would still need to support the flag for at least 2 more versions should we want to do what's on this PR*~~

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 * [ENHANCEMENT] Prometheus upgraded. #2798 #2849
   * Optimized labels regex matchers for patterns containing literals (eg. `foo.*`, `.*foo`, `.*foo.*`)
 * [ENHANCEMENT] Experimental Alertmanager: Alertmanager configuration persisted to object storage using an experimental API that accepts and returns YAML-based Alertmanager configuration. #2768
+* [ENHANCEMENT] Ruler: `-ruler.alertmanager-url` now supports multiple URLs. Each URL is treated as a separate Alertmanager group. Support for multiple Alertmanagers in a group can be achieved by using DNS service discovery. #2851
 * [BUGFIX] Fixed a bug in the index intersect code causing storage to return more chunks/series than required. #2796
 * [BUGFIX] Fixed the number of reported keys in the background cache queue. #2764
 * [BUGFIX] Fix race in processing of headers in sharded queries. #2762
--- a/pkg/ruler/notifier.go
+++ b/pkg/ruler/notifier.go
@@ -92,7 +92,7 @@ func buildNotifierConfig(rulerConfig *Config) (*config.Config, error) {
 		// hosts provided follow this specification: _service._proto.name
 		// e.g. _http._tcp.alertmanager.com
 		if rulerConfig.AlertmanagerDiscovery && !srvDNSregexp.MatchString(url.Host) {
-			return nil, fmt.Errorf("When alertmanager-discovery is on, host name must be of the form _portname._tcp.service.fqdn (is %q)", url.Host)
+			return nil, fmt.Errorf("when alertmanager-discovery is on, host name must be of the form _portname._tcp.service.fqdn (is %q)", url.Host)
 		}
 
 		validURLs = append(validURLs, url)
--- a/pkg/ruler/notifier.go
+++ b/pkg/ruler/notifier.go
@@ -2,7 +2,9 @@ package ruler
 
 import (
 	"context"
+	"fmt"
 	"net/url"
+	"regexp"
 	"sync"
 
 	gklog "github.com/go-kit/kit/log"
@@ -75,6 +77,7 @@ func (rn *rulerNotifier) stop() {
 func buildNotifierConfig(rulerConfig *Config) (*config.Config, error) {
 	validURLs := make([]*url.URL, 0, len(rulerConfig.AlertmanagerURL))
 
+	srvDNSregexp := regexp.MustCompile(`^_.+._.+`)
 	for _, h := range rulerConfig.AlertmanagerURL {
 		url, err := url.Parse(h)
 		if err != nil {
@@ -83,6 +86,13 @@ func buildNotifierConfig(rulerConfig *Config) (*config.Config, error) {
 
 		if url.String() == "" {
 			continue
+		}
+
+		// Given we only support SRV lookups as part of service discovery, we need to ensure
+		// hosts provided follow this specification: _service._proto.name
+		// e.g. _http._tcp.alertmanager.com
+		if rulerConfig.AlertmanagerDiscovery && !srvDNSregexp.MatchString(url.Host) {
+			return nil, fmt.Errorf("When alertmanager-discovery is on, host name must be of the form _portname._tcp.service.fqdn (is %q)", url.Host)
 		}
 
 		validURLs = append(validURLs, url)
--- a/pkg/ruler/notifier.go
+++ b/pkg/ruler/notifier.go
@@ -99,42 +99,7 @@ func buildNotifierConfig(rulerConfig *Config) (*config.Config, error) {
 
 	amConfigs := make([]*config.AlertmanagerConfig, 0, len(validURLs))
 	for _, url := range validURLs {
-
-		var sdConfig sd_config.ServiceDiscoveryConfig
-		if rulerConfig.AlertmanagerDiscovery {
-			sdConfig.DNSSDConfigs = []*dns.SDConfig{{
-				Names:           []string{url.Host},
-				RefreshInterval: model.Duration(rulerConfig.AlertmanagerRefreshInterval),
-				Type:            "SRV",
-				Port:            0, // Ignored, because of SRV.
-			}}
-		} else {
-			sdConfig.StaticConfigs = []*targetgroup.Group{{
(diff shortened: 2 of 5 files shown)


pkg/ruler/notifier.go
```diff
@@ -74,18 +72,24 @@ func (rn *rulerNotifier) stop() {
...
+	}
+
+	if len(validHosts) == 0 {
 		return &config.Config{}, nil
 	}
 
-	u := rulerConfig.AlertmanagerURL
 	var sdConfig sd_config.ServiceDiscoveryConfig
 	if rulerConfig.AlertmanagerDiscovery {
-		if !strings.Contains(u.Host, "_tcp.") {
```
gotjosh, line 84: I can't seem to find a history behind this change and the reason for its introduction but my investigation points out that this is not needed e.g. using a Kubernetes headless service both `_http._tcp.service.svc.local` and `service.svc.local` produce the same DNS query result.

pkg/ruler/ruler.go
```diff
@@ -74,14 +74,21 @@ type Config struct {
 	StoreConfig RuleStoreConfig `yaml:"storage"`
 	// Path to store rule files for prom manager.
 	RulePath string `yaml:"rule_path"`
-
+	// Sets the Authorization header on every request to the Alertmanager with the configured username.
```
gotjosh, line 77: Sadly, I had to introduce more flags given we no longer can assume a scheme, path, username and password based on URL like we did when it was just one. The good news is that this aligns with what the users of Prometheus would expect as this what the Alertmanager configuration supports.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Ruler: `-ruler.alertmanager-url` is now deprecated and has been replaced by `-ruler.alertmanager-hosts` along with `-ruler.alertmanager-path-prefix`, `-ruler.alertmanager-scheme`, `-ruler.alertmanager-username` and `-ruler.alertmanager-password`. `-ruler.alertmanager-url` will be removed in v1.4.0. #XXXX
```
gotjosh, line 5: I assume these are the deprecation rules based on what I read. I assume we only need to support not erroring when you used it as opposed to supporting it in the code itself.

gotjosh:
cc: @jtlisi @cstyan @gouthamve

CHANGELOG.md
bboreham: Please don't break my running system now we are past 1.0

gouthamve: Yep, we agreed on DMs that no breaking changes are to be made.

gotjosh:
> This seems to me like we're just adding the ability for multiple SD or static configs under one higher level alertmanager config?

That is correct, there's no notion of groups of groups of Alertmanagers under the current Cortex structure, and that change is much more involved than what I'm proposing here.

As an attempt to clear that confusion I have opened #2860 to discuss how to best proceed with that.

CHANGELOG.md
gotjosh: Based on the conversation in #2860, this has taken a new approach.

gotjosh:
Another look @gouthamve ?

pkg/ruler/ruler.go
```diff
@@ -128,12 +128,14 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 	f.DurationVar(&cfg.EvaluationDelay, "ruler.evaluation-delay-duration", 0, "Duration to delay the evaluation of rules to ensure they underlying metrics have been pushed to cortex.")
 	f.DurationVar(&cfg.PollInterval, "ruler.poll-interval", 1*time.Minute, "How frequently to poll for rule changes")
-	f.Var(&cfg.AlertmanagerURL, "ruler.alertmanager-url", "URL of the Alertmanager to send notifications to.")
-	f.BoolVar(&cfg.AlertmanagerDiscovery, "ruler.alertmanager-discovery", false, "Use DNS SRV records to discover alertmanager hosts.")
-	f.DurationVar(&cfg.AlertmanagerRefreshInterval, "ruler.alertmanager-refresh-interval", 1*time.Minute, "How long to wait between refreshing alertmanager hosts.")
-	f.BoolVar(&cfg.AlertmanangerEnableV2API, "ruler.alertmanager-use-v2", false, "If enabled requests to alertmanager will utilize the V2 API.")
+
+	f.Var(&cfg.AlertmanagerURL, "ruler.alertmanager-url", "URL(s) of the Alertmanager(s) to send notifications to. Each Alertmanager URL is treated as a separate group in the configuration. Multiple Alertmanagers in HA per group can be supported by using DNS resolution via -ruler.alertmanager-discovery.")
```
gouthamve, line 132: `ruler.alertmanager-url` do we specify the flag multiple times or do we do comma separated, but single flag? Not super clear from the help text.

pkg/ruler/notifier.go
```diff
@@ -74,67 +73,74 @@ func (rn *rulerNotifier) stop() {
...
+				BasicAuth: &config_util.BasicAuth{
+					Username: url.User.Username(),
+				},
+			}
+
+			if password, isSet := url.User.Password(); isSet {
+				amConfig.HTTPClientConfig.BasicAuth.Password = config_util.Secret(password)
+			}
 		}
+
```
gouthamve, line 136: Can we abstract this out into a function like `amConfigFromURL`? Its easier to read that way.

pkg/ruler/ruler.go
gotjosh: The documentation specifies a `StringSlice` as `list of strings` (not only for this one but others too) so I thought that might be sufficient. However, you're right - I have slightly changed the wording of the flag description to make it a bit more intuitive.

gotjosh:
> LGTM but as discussed offline, I think the _tcp bits are there for SRV records to know which port to look up. But my understanding of this is fuzzy, need to double check.

I don't think so, from what I can see it seems like `SRV`lookups (which are the only type we support) don't require a port(https://github.com/prometheus/prometheus/blob/master/discovery/dns/dns.go#L70) moreso, I don't see this check anywhere within Prometheus.

pkg/ruler/notifier.go
gotjosh: Cleared offline with @gouthamve, I have reintroduced the check given we need it. Added a test for it and a comment that explains a bit more in depth the reason.

pkg/ruler/notifier.go
```diff
@@ -74,68 +75,86 @@ func (rn *rulerNotifier) stop() {
...
-					Targets: []model.LabelSet{
-						{
-							model.AddressLabel: model.LabelValue(u.Host),
-						},
-					},
-				},
-			},
+
+		// Given we only support SRV lookups as part of service discovery, we need to ensure
+		// hosts provided follow this specification: _service._proto.name
```
gouthamve, line 92: `_portname._proto.name`

pkg/ruler/notifier.go
```diff
@@ -74,68 +75,86 @@ func (rn *rulerNotifier) stop() {
...
+		// e.g. _http._tcp.alertmanager.com
+		if rulerConfig.AlertmanagerDiscovery && !srvDNSregexp.MatchString(url.Host) {
+			return nil, fmt.Errorf("When alertmanager-discovery is on, host name must be of the form _portname._tcp.service.fqdn (is %q)", url.Host)
```
pracucci, line 95:
```suggestion
			return nil, fmt.Errorf("when alertmanager-discovery is on, host name must be of the form _portname._tcp.service.fqdn (is %q)", url.Host)
```

pkg/ruler/notifier_test.go
```diff
@@ -0,0 +1,190 @@
...
+				},
+			},
+		},
+		{
+			name: "with service discovery and an invalid URL",
+			cfg: &Config{
+				AlertmanagerURL:       []string{"http://_http.default.svc.cluster.local/alertmanager"},
+				AlertmanagerDiscovery: true,
+			},
+			err: fmt.Errorf("When alertmanager-discovery is on, host name must be of the form _portname._tcp.service.fqdn (is \"alertmanager.default.svc.cluster.local\")"),
```
pracucci, line 80:
```suggestion
			err: fmt.Errorf("when alertmanager-discovery is on, host name must be of the form _portname._tcp.service.fqdn (is \"alertmanager.default.svc.cluster.local\")"),
```

pkg/ruler/notifier.go
gotjosh: According to [wikipedia](https://en.wikipedia.org/wiki/SRV_record) this is suppose to be `_service`, I believe the port name convention `_portname` is a k8s thing.

pracucci: I think Josh is right.

PR #2852, opened 2020-07-08
Added -store.max-chunks-per-query limit support to blocks storage

### LINKED ISSUE #2844: Add MaxChunksPerQuery support to blocks storage
The Cortex chunks storage supports the `MaxChunksPerQuery` limit (`-store.query-chunk-limit` flag, defaults to 2M) implements this limit it in the querier and affects only chunks loaded from the long-term storage. It’s used to protect from queries OOMing.

We should support the same limit in the blocks storage as well.

### Proposal

- Add a chunks limiter in Thanos `BucketStore` (not strictly required, but better to limit asap)
  - https://github.com/thanos-io/thanos/issues/2845
- Enforce the limit in the querier's `blocksStoreQuerier` adding a chunks limiter tracked for each query. The number of received chunks from store-gateways can be computed from `storepb.Series.Chunks` while `stream.Recv()`

pracucci:
**What this PR does**:
In this PR:
- Added `-store.max-chunks-per-query` limit support to blocks storage
- Removed `max-sample-count` support because it's flawed in Thanos (it's per block and not per query as documented). I will work on this in a subsequent PR (I need to fix it in Thanos first) and the limit in store-gateway will be based on `-store.max-chunks-per-query` too.

Out of the scope of this PR:
- Add the limit to the store-gateway too (will be done in a separate PR, because I need to refactor and [fix Thanos first](https://github.com/thanos-io/thanos/pull/2858))
- Add a negative cache to avoid running the same query for the same tenant if we already know it will hit the max chunks limit ([issue](https://github.com/cortexproject/cortex/issues/2853))

Fixes #2844

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * [CHANGE] Metric `cortex_ingester_flush_reasons` has been renamed to `cortex_ingester_series_flushed_total`, and is now incremented during flush, not when series is enqueued for flushing. #2802
 * [CHANGE] Experimental Delete Series: Metric `cortex_purger_oldest_pending_delete_request_age_seconds` would track age of delete requests since they are over their cancellation period instead of their creation time. #2806
 * [CHANGE] Experimental TSDB: the store-gateway service is required in a Cortex cluster running with the experimental blocks storage. Removed the `-experimental.tsdb.store-gateway-enabled` CLI flag and `store_gateway_enabled` YAML config option. The store-gateway is now always enabled when the storage engine is `tsdb`. #2822
-* [CHANGE] Experimental TSDB: removed support for `-experimental.tsdb.bucket-store.max-sample-count` flag because the implementation was flawed. To limit the number of samples/chunks processed by a single query you can set `-store.query-chunk-limit`, which is now supported by the blocks storage too.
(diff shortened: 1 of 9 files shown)


pracucci:
> LGTM, good job.

Thanks!

> (I'm becoming more nervous about mixing direct and atomic access to the same variable, although in this specific case, there is clear happens-before relationship between the two, so I _think_ it is correct. [Talking about `numChunks` in `fetchSeriesFromStores` method])

It's safe because of the `g.Wait()` before reading `numChunks`, but let's be consistent: I've committed the usage of `atomic.LoadInt32()` when reading it in the `return` too.

pstibrany:
> (I'm becoming more nervous about mixing direct and atomic access to the same variable, although in this specific case, there is clear happens-before relationship between the two, so I _think_ it is correct. [Talking about `numChunks` in `fetchSeriesFromStores` method])

To expand on this comment, I think my reasoning is wrong. Program happens-before order, and atomic accesses are independent, so correct version of the code should use atomic load at the end of `fetchSeriesFromStores`. It would be cleaner to use uber-go/atomic package instead for this.

But no need to change it in this PR... we have similar code elsewhere.

pstibrany:
> 
> It's safe because of the `g.Wait()` before reading `numChunks`

This is what I thought as well. Unfortunately, there is no such guarantee.

PR #2854, opened 2020-07-08
Adding the Ruler to the Single Binary

### LINKED ISSUE #2833: Add sample configuration file with ruler config
Sample configuration files for running Cortex can be found under [docs/configuration](https://github.com/cortexproject/cortex/tree/54f10c2b1223bf16ca955ac517b55fa878d5de50/docs/configuration).

These show configuration examples for most components but not the Ruler.

joe-elliott:
**What this PR does**:
- Adds the ruler to the single binary
  - Adds labels `engine=querier` and `engine=ruler` to prevent metrics collisions on promql.Engine.
- Adds a new local storage engine for single binary local mode
  - Per discussions with @jtlisi it is read only.  Jacob, if you could review this piece in detail I would appreciate it.
- Adds appropriate unit/integration tests.
- Updates single-process config yamls 

Fixes #2833

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/ruler/rules/local/local.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// ListAllRuleGroups implements RuleStore
+func (l *Client) ListAllRuleGroups(ctx context.Context) (map[string]rules.RuleGroupList, error) {
+	lists := make(map[string]rules.RuleGroupList)
+
+	root := l.cfg.Directory
+	infos, err := ioutil.ReadDir(root)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to read dir "+root)
```
pracucci, line 47:
```suggestion
		return nil, errors.Wrapf(err, "unable to read dir %s", root)
```

pkg/ruler/rules/local/local.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	for _, info := range infos {
+		if !info.IsDir() {
+			continue
+		}
+
+		list, err := l.listAllRulesGroupsForUser(ctx, info.Name())
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to list rule group")
```
pracucci, line 57:
```suggestion
			return nil, errors.Wrapf(err, "failed to list rule groups for user %s", info.Name())
```

pkg/ruler/rules/local/local.go
```diff
@@ -0,0 +1,136 @@
...
+
+// DeleteRuleGroup implements RuleStore
+func (l *Client) DeleteRuleGroup(ctx context.Context, userID, namespace string, group string) error {
+	return errors.New("DeleteRuleGroup unsupported in rule local store")
+}
+
+func (l *Client) listAllRulesGroupsForUser(ctx context.Context, userID string) (rules.RuleGroupList, error) {
+	var allLists rules.RuleGroupList
+
+	root := path.Join(l.cfg.Directory, userID)
```
pracucci, line 93: You should use `filepath.Join()` instead of `path.Join()` when building filesystem paths.

pkg/ruler/rules/local/local.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to read dir "+root)
+	}
+
+	for _, info := range infos {
+		if info.IsDir() {
+			continue
+		}
+
+		list, err := l.ListRuleGroups(ctx, userID, info.Name())
```
pracucci, line 104: Could be more clear simply calling `listAllRulesGroupsForUserAndNamespace()`.

pkg/ruler/rules/local/local.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return allLists, nil
+}
+
+func (l *Client) listAllRulesGroupsForUserAndNamespace(ctx context.Context, userID string, namespace string) (rules.RuleGroupList, error) {
+	filename := path.Join(l.cfg.Directory, userID, namespace)
+
+	rulegroups, allErrors := rulefmt.ParseFile(filename)
+	if len(allErrors) > 0 {
+		return nil, errors.Wrap(allErrors[0], "error parsing "+filename)
```
pracucci, line 120:
```suggestion
		return nil, errors.Wrapf(allErrors[0], "error parsing %s", filename)
```

pkg/ruler/rules/local/local.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	allErrors = rulegroups.Validate()
+	if len(allErrors) > 0 {
+		return nil, errors.Wrap(allErrors[0], "error validating "+filename)
```
pracucci, line 125:
```suggestion
		return nil, errors.Wrapf(allErrors[0], "error validating %s", filename)
```

pkg/ruler/rules/local/local_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
+
+	rulefmt "github.com/cortexproject/cortex/pkg/ruler/legacy_rulefmt"
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
+)
+
+func TestLoadingRules(t *testing.T) {
```
pracucci, line 19: I would justbe more clear that it's testing only `ListAllRuleGroups()`.
```suggestion
func TestClient_ListAllRuleGroups(t *testing.T) {
```

pkg/cortex/cortex.go
```diff
@@ -219,9 +222,8 @@ type Cortex struct {
 	StoreGateway *storegateway.StoreGateway
 	MemberlistKV *memberlist.KVInitService
 
-	// Queryables that the querier should use to query the long
-	// term storage. It depends on the storage engine used.
-	StoreQueryables []querier.QueryableWithFilter
+	Queryable prom_storage.SampleAndChunkQueryable
+	Engine    *promql.Engine
```
pracucci, line 226: Naming it `QueryEngine` could help to avoid ambiguities.

pkg/cortex/modules.go
```diff
@@ -169,8 +169,6 @@ func (t *Cortex) initDistributor() (serv services.Service, err error) {
 }
 
 func (t *Cortex) initQuerier() (serv services.Service, err error) {
-	queryable, engine := querier.New(t.Cfg.Querier, t.Overrides, t.Distributor, t.StoreQueryables, t.TombstonesLoader, prometheus.DefaultRegisterer)
```
pracucci, line 172: Design wise, looks a bit weird to me that the querier is not created by `initQuerier()`. I'm probably missing something, but wouldn't possible to rollback this change and just have the Ruler module depending on the Querier so that it can reference `t.Queryable` and `t.Engine`?

pkg/cortex/cortex.go
```diff
@@ -219,9 +222,8 @@ type Cortex struct {
 	StoreGateway *storegateway.StoreGateway
 	MemberlistKV *memberlist.KVInitService
 
-	// Queryables that the querier should use to query the long
-	// term storage. It depends on the storage engine used.
-	StoreQueryables []querier.QueryableWithFilter
+	Queryable prom_storage.SampleAndChunkQueryable
```
pracucci, line 225: Why this change? Could you share more details about why this change is needed, please?

integration/e2ecortex/services.go
```diff
@@ -237,6 +237,9 @@ func NewSingleBinary(name string, flags map[string]string, image string, otherPo
 			"-ingester.concurrent-flushes":   "10",
 			"-ingester.max-transfer-retries": "10",
 			"-ingester.num-tokens":           "512",
+			// Ruler
+			"-ruler.storage.type":            "local",
+			"-ruler.storage.local.directory": "/tmp/cortex/rules",
```
pracucci, line 242: Where is this used? I think you override it when used. I'm wondering if we need to have this configuration at all in the generic `NewSingleBinary()` instead of injecting it when required (given in the integration test you override it anyway).

pkg/ruler/rules/local/local.go
joe-elliott: Thank you for your patient reviews.  These mistakes feel silly to me.  I have a lot to learn about Go.

integration/e2ecortex/services.go
joe-elliott: The single binary now requires _some_ amount of ruler configuration or it will fail to startup.  This is a relatively minimal configuration that will allow the single binary to start.  Without this tests such as `TestSingleBinaryWithMemberlist` will fail.

We could make the Ruler startup successfully with no configuration using an in memory implementation if desired.  I'd be concerned with starting up in a configuration like this however as the Ruler probably wouldn't work as expected.

joe-elliott:
Allright.  Feeling good about this one.  Dropping draft status.

@pracucci thank you for the early notes!

integration/e2ecortex/services.go
pstibrany: I think we should mention the fact that minimal ruler configuration is required in the changelog. Alternatively, we can disable ruler if no configuration is provided in single-binary mode (and log warning about it).

As it is now, people using single-binary will run into this problem as soon as they try to run Cortex, with pretty cryptic error:

```
cortex-1_1       | level=error ts=2020-07-14T11:48:54.9596355Z caller=log.go:140 msg="error running cortex" err="configs_api_url is not set or valid\nerror initialising module: ruler\ngithub.com/cortexproject/cortex/pkg/util/modules.(*Manager).InitModuleServices\n\t/Users/peter/Grafana/cortex/pkg/util/modules/modules.go:89\ngithub.com/cortexproject/cortex/pkg/cortex.(*Cortex).Run\n\t/Users/peter/Grafana/cortex/pkg/cortex/cortex.go:299\nmain.main\n\t/Users/peter/Grafana/cortex/cmd/cortex/main.go:153\nruntime.main\n\t/usr/local/Cellar/go/1.14.3/libexec/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/Cellar/go/1.14.3/libexec/src/runtime/asm_amd64.s:1373"
```

joe-elliott:
@pstibrany 
I added a check and an info level log to prevent the single binary from failing if the ruler was not configured.  The check is pretty gross, but I don't know if there's a cleaner option.  Integration tests should warn us if there's ever drift.

pstibrany:
> @pstibrany
> I added a check and an info level log to prevent the single binary from failing if the ruler was not configured. The check is pretty gross, but I don't know if there's a cleaner option. Integration tests should warn us if there's ever drift.

Thanks Joe. This works fine (I'd use a shared constant string between the two, but that's just a minor nit). I already approved.

integration/e2ecortex/services.go
joe-elliott: Per Peter's comments I have removed the need for ruler configuration in the single binary.  Resolving.

pkg/cortex/modules.go
```diff
@@ -449,9 +450,19 @@ func (t *Cortex) initTableManager() (services.Service, error) {
 }
 
 func (t *Cortex) initRuler() (serv services.Service, err error) {
+	// if the ruler is not configured and we're in single binary then let's just log an error and continue
+	// unfortunately there is no way to generate a "default" config and compare default against actual
+	// to determine if it's unconfigured.  the following check, however, correctly tests this.
+	// Single binary integration tests will break if this ever drifts
+	if t.Cfg.Target == All && t.Cfg.Ruler.StoreConfig.Type == "configdb" && t.Cfg.Ruler.StoreConfig.ConfigDB.ConfigsAPIURL.URL == nil {
```
jtlisi, line 457: No active work is being done on the configdb so I doubt this will break in the near future.

pkg/ruler/rules/local/local.go
```diff
@@ -0,0 +1,136 @@
...
+// RegisterFlags registers flags.
+func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.StringVar(&cfg.Directory, prefix+"local.directory", "", "Directory to scan for rules")
+}
+
+// Client expects to load already existing rules located at:
+//  cfg.Directory / userID / namespace
+type Client struct {
+	cfg Config
+}
```
annanay25, line 28: nit: This can be named RuleStore to be consistent with the obj store naming convention

PR #2857, opened 2020-07-08
Add a metric to track ruler update failures (errors in loading rules files).

cstyan:
**What this PR does**:
Adds a metric to track ruler update failures per user.

cc @gotjosh @jtlisi 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
   * `s3.http.insecure-skip-verify`
 * [ENHANCEMENT] Prometheus upgraded. #2798 #2849
   * Optimized labels regex matchers for patterns containing literals (eg. `foo.*`, `.*foo`, `.*foo.*`)
+* [ENHANCEMENT] Add metric `ruler_config_update_failures_total` to Ruler to track failures of loading rules files. #2857
 * [BUGFIX] Fixed a bug in the index intersect code causing storage to return more chunks/series than required. #2796
 * [BUGFIX] Fixed the number of reported keys in the background cache queue. #2764
 * [BUGFIX] Fix race in processing of headers in sharded queries. #2762
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@
   * `s3.http.insecure-skip-verify`
 * [ENHANCEMENT] Prometheus upgraded. #2798 #2849
   * Optimized labels regex matchers for patterns containing literals (eg. `foo.*`, `.*foo`, `.*foo.*`)
-* [ENHANCEMENT] Add metric `ruler_config_update_failures_total` to Ruler to track failures of loading rules files. #2857
+* [ENHANCEMENT] Add metric `cortex_ruler_config_update_failures_total` to Ruler to track failures of loading rules files. #2857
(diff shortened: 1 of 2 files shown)


pkg/ruler/ruler.go
```diff
@@ -50,6 +50,11 @@ var (
 		Name:      "ruler_config_updates_total",
 		Help:      "Total number of config updates triggered by a user",
 	}, []string{"user"})
+	configUpdateFailuresTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "ruler_config_update_failures_total",
+		Help:      "Total number of config updates triggered by a user",
```
gotjosh, line 56:
```suggestion
		Help:      "Total number of config update failures triggered by a user",
```

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
   * `s3.http.insecure-skip-verify`
 * [ENHANCEMENT] Prometheus upgraded. #2798 #2849
   * Optimized labels regex matchers for patterns containing literals (eg. `foo.*`, `.*foo`, `.*foo.*`)
+* [ENHANCEMENT] Add metric to Ruler to track failures of loading rules files. #2857
```
gotjosh, line 52: From what I see in other instance we tend to include the metric name in the changelog.

pracucci: Also please remove the trailing whitespace or run `make clean-white-noise` to fix it (otherwise the linter fails).

pkg/ruler/ruler.go
```diff
@@ -511,6 +516,7 @@ func (r *Ruler) syncManager(ctx context.Context, user string, groups store.RuleG
 		}
 		err = manager.Update(r.cfg.EvaluationInterval, files, nil)
 		if err != nil {
+			configUpdateFailuresTotal.WithLabelValues(user).Inc()
```
pracucci, line 519: Shouldn't we increment it also in case of `unable to create rule manager` few lines above?

cstyan: It's not clear to me if we should or shouldn't, the failure to create the ruler manager above seems less tied to the users config/rules files (which is what prompted this PR).

jtlisi: If the above command fails that would signal a error that is a result of a bad Cortex config, not a user rules config. We should instrument that error, or ensure there is a mechanism to communicate a failure. But it probably should be a separate metric.

jtlisi: A `reason` label should be an adequate solution

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
   * `s3.http.insecure-skip-verify`
 * [ENHANCEMENT] Prometheus upgraded. #2798 #2849
   * Optimized labels regex matchers for patterns containing literals (eg. `foo.*`, `.*foo`, `.*foo.*`)
+* [ENHANCEMENT] Add metric `ruler_config_update_failures_total` to Ruler to track failures of loading rules files. #2857
```
pracucci, line 53: Let's reference the right metric name :)

```suggestion
* [ENHANCEMENT] Add metric `cortex_ruler_config_update_failures_total` to Ruler to track failures of loading rules files. #2857
```

PR #2864, opened 2020-07-10
Add comment about prometheus sync with thanos

sylr:
**What this PR does**:

As Thanos depends on Cortex and Cortex depends of Thanos and both depends on Prometheus, upgrading prometheus on one side can cause build failures on the other side if breaking changes are shipped within the new version of prometheus.

I believe that syncing the upgrades of prometheus between Thanos & Cortex will make it more manageable.

CC https://github.com/thanos-io/thanos/pull/2871

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,8 @@ require (
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.10.0
+	// When upgrading prometheus please open a PR on github.com/thanos-io/thanos
+	// to upgrade it there as well with the same version.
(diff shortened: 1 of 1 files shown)


pracucci:
Thanks @sylr for your PR!

> As Thanos depends on Cortex and Cortex depends of Thanos and both depends on Prometheus, upgrading prometheus on one side can cause build failures on the other side if breaking changes are shipped within the new version of prometheus.

Does it? It should be unidirectional. We have this issue in Cortex, but Thanos shouldn't suffer it. Could you show me an example of such issue in Thanos, when Cortex upgrade Prometheus?

PR #2865, opened 2020-07-13
make fs object client windows compatible and add a method to get pathseparator from object clients

sandeepsukhani:
**What this PR does**:
Makes filesystem object client windows compatible by not using a hardcoded separator.
It also adds a method to `ObjectClient` to get what separator is being used by underlying storage.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/aws/s3_storage_client.go
+++ b/pkg/chunk/aws/s3_storage_client.go
@@ -325,3 +325,7 @@ func (a *S3ObjectClient) List(ctx context.Context, prefix string) ([]chunk.Stora
 
 	return storageObjects, commonPrefixes, nil
 }
+
+func (a *S3ObjectClient) PathSeparator() string {
+	return a.delimiter
+}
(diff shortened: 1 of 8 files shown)


pstibrany:
I am (weakly) opposed to this idea. I don't think Windows-specific filesystem issues should leak into APIs for all object clients. There are more issues and restrictions in Windows filesystem than this, I would not be happy to see them affecting the API. Furthermore, I don't see new `PathSeparator` method used anywhere.

I would prefer to see translation between separators to be done in filesystem object client (if needed). WDYT?

sandeepsukhani:
Sorry, I missed mentioning in the description that this is mostly required for boltdb shipper in Loki.
The only thing that is causing a problem for is the usage of `chunk.DirDelim` in FilesystemObjectClient which is a constant set to `\`. The usage of it is only [here](https://github.com/cortexproject/cortex/pull/2865/files#diff-d2aefc141a7c8f03082f08e8a4424fa8R129) which is the only change required to make it windows compatible. All the other path operations are using [filepath.Join](https://golang.org/pkg/path/filepath/#Join) which joins the elements using OS-specific separator. 

Now the shipper doesn't know whether the underlying implementation is a FilesystemObjectClient so can't do the path management on its own, which is why I have added a method to get the separator.

Yes, converting the separators is an option but it is not as clean as this is because Linux allows having `\` in filenames and we don't want to get into that rabbit hole of possibilities and use standard package instead.

cc looping in @slim-bean with whom I had this discussion, in case he wants to add something.

pstibrany:
Change makes sense and is easier than the alternative. (Although I cannot help myself but to link to https://en.wikipedia.org/wiki/Filename#In_Windows, esp. reserved file names section :))

PR #2870, opened 2020-07-13
Compact TSDB on open.

### LINKED ISSUE #2869: Ensure ingester's TSDB Compact() is called after WAL replay
When running the Cortex blocks storage, at ingester startup we do replay the WAL for any tenant's TSDB found in the local filesystem and, once done, the ingester switches to `READY` and start receiving samples from distributors.

There are scenarios where an ingester's TSDB may have a compactable HEAD right after the WAL replay but the HEAD is not compacted before the ingester switches to `READY`, because it just gets compacted after 1 minute as effect of the internal ticker used by `compactionLoop()`.

I propose to ensure that the ingester compacts the HEAD (if compactable) **after** the WAL replay and **before* switching to `READY`.

pstibrany:
**What this PR does**: This PR calls `Compact` method on newly openened TSDB. If there is data in head that needs to be cut into blocks, this will force it. If there is no such data, this is a noop.

Fixes #2869 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -891,7 +891,7 @@ func (i *Ingester) createTSDB(userID string) (*userTSDB, error) {
 		SeriesLifecycleCallback: userDB,
 	})
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to open TSDB: %s", udir)
 	}
 	db.DisableCompactions() // we will compact on our own schedule
 
@@ -900,7 +900,7 @@ func (i *Ingester) createTSDB(userID string) (*userTSDB, error) {
 	// local blocks compaction may still take place if configured.
 	err = db.Compact()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to compact TSDB: %s", udir)
 	}
 
 	userDB.DB = db
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -896,8 +896,8 @@ func (i *Ingester) createTSDB(userID string) (*userTSDB, error) {
 	db.DisableCompactions() // we will compact on our own schedule
 
 	// Run compaction before using this TSDB. If there is data in head that needs to be put into blocks,
-	// this will actually create the blocks. If there is no data (empty TSDB), this is a no-op.
-	// We don't do blocks compaction.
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -895,6 +895,14 @@ func (i *Ingester) createTSDB(userID string) (*userTSDB, error) {
 	}
 	db.DisableCompactions() // we will compact on our own schedule
 
+	// Run compaction before using this TSDB. If there is data in head that needs to be put into blocks,
+	// this will actually create the blocks. If there is no data (empty TSDB), this is a no-op.
+	// We don't do blocks compaction.
```
pracucci, line 900: It depends how you configure `-experimental.tsdb.block-ranges-period` (which is a list). If you configure multiple ranges there, then we do blocks compaction too. I wouldn't worry too much at this stage, but I would at least remove this comment because not necessarily true.

pkg/ingester/ingester_v2.go
```diff
@@ -895,6 +895,14 @@ func (i *Ingester) createTSDB(userID string) (*userTSDB, error) {
...
+	err = db.Compact()
+	if err != nil {
+		return nil, err
```
pracucci, line 903: I would wrap this error (and also the one from `tsdb.Open()`) to explain what we were doing.

pkg/ingester/ingester_v2_test.go
```diff
@@ -1602,3 +1603,76 @@ func pushSingleSample(t *testing.T, i *Ingester) {
...
+	defer services.StopAndAwaitTerminated(context.Background(), ingester) //nolint:errcheck
+
+	db := ingester.getTSDB(userID)
+	require.NotNil(t, db)
+
+	h := db.Head()
+
+	dur := time.Duration(h.MaxTime()-h.MinTime()) * time.Millisecond
+	require.True(t, dur < 4*time.Hour)
+	require.NotEmpty(t, db.Blocks())
```
pracucci, line 1677: I would assert on the number of blocks (should always be the same, given TSDB never rely on `time.Now()`). I would also set the retention explicitely in `ingesterCfg.TSDBConfig` above. This way we would have an even more significant test.

pstibrany:
Thanks for review, I've addressed your comments now.

pracucci:
Thanks @pstibrany! Let's merge it 😉

PR #2874, opened 2020-07-14
Expose hashsum of the config files

simonswine:
**What this PR does**:

This adds metrics for exposing the the sha256 hash of the config files used (before expansion). This allows to monitor the roll-out of a new config file across the cluster and can help to detect a mismatch in active configuration files.

The metric would look like this:

```
cortex_config_info{hash="sha256:86ec4c62f1a0ecc2eb111b752d6aa06db0d361c9f8d8fa6fc703ce6b85532458"} 1
```

As no expansion is going on this matches with the local file:

```
86ec4c62f1a0ecc2eb111b752d6aa06db0d361c9f8d8fa6fc703ce6b85532458  ./docs/configuration/single-process-config.yaml
```

No issue yet

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * [CHANGE] Added the `engine` label to the metrics exposed by the Prometheus query engine, to distinguish between `ruler` and `querier` metrics. #2854
 * [CHANGE] Added ruler to the single binary when started with `-target=all` (default). #2854
 * [CHANGE] Experimental TSDB: compact head when opening TSDB. This should only affect ingester startup after it was unable to compact head in previous run. #2870
+* [CHANGE] Metric `cortex_overrides_last_reload_successful` has been renamed to `cortex_runtime_config_last_reload_successful`. #2874
 * [FEATURE] Introduced `ruler.for-outage-tolerance`, Max time to tolerate outage for restoring "for" state of alert. #2783
 * [FEATURE] Introduced `ruler.for-grace-period`, Minimum duration between alert and restored "for" state. This is maintained only for alerts with configured "for" time greater than grace period. #2783
 * [FEATURE] Introduced `ruler.resend-delay`, Minimum amount of time to wait before resending an alert to Alertmanager. #2783
@@ -62,6 +63,7 @@
 * [ENHANCEMENT] Ruler: `-ruler.alertmanager-url` now supports multiple URLs. Each URL is treated as a separate Alertmanager group. Support for multiple Alertmanagers in a group can be achieved by using DNS service discovery. #2851
 * [ENHANCEMENT] Experimental TSDB: Cortex Flusher now works with blocks engine. Flusher needs to be provided with blocks-engine configuration, existing Flusher flags are not used (they are only relevant for chunks engine). Note that flush errors are only reported via log. #2877
 * [ENHANCEMENT] Flusher: Added `-flusher.exit-after-flush` option (defaults to true) to control whether Cortex should stop completely after Flusher has finished its work. #2877
+* [ENHANCEMENT] Added metrics `cortex_config_hash` and `cortex_runtime_config_hash` to expose hash of the currently active config file. #2874
 * [BUGFIX] Fixed a bug in the index intersect code causing storage to return more chunks/series than required. #2796
 * [BUGFIX] Fixed the number of reported keys in the background cache queue. #2764
 * [BUGFIX] Fix race in processing of headers in sharded queries. #2762
--- a/cmd/cortex/main.go
+++ b/cmd/cortex/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/sha256"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -29,11 +30,21 @@ var (
 	Revision string
 )
 
+// configHash exposes information about the loaded config
+var configHash *prometheus.GaugeVec = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Name: "cortex_config_hash",
+		Help: "Hash of the currently active config file.",
+	},
+	[]string{"sha256"},
+)
+
 func init() {
 	version.Version = Version
 	version.Branch = Branch
 	version.Revision = Revision
 	prometheus.MustRegister(version.NewCollector("cortex"))
+	prometheus.MustRegister(configHash)
 }
 
 const (
@@ -183,6 +194,12 @@ func LoadConfig(filename string, expandENV bool, cfg *cortex.Config) error {
 	if err != nil {
 		return errors.Wrap(err, "Error reading config file")
 	}
+
+	// create a sha256 hash of the config before expansion and expose it via
+	// the config_info metric
+	hash := sha256.Sum256(buf)
+	configHash.Reset()
+	configHash.WithLabelValues(fmt.Sprintf("%x", hash)).Set(1)
 
 	if expandENV {
 		buf = expandEnv(buf)
--- a/pkg/cortex/runtime_config.go
+++ b/pkg/cortex/runtime_config.go
@@ -1,7 +1,7 @@
 package cortex
 
 import (
-	"os"
+	"io"
 
 	"gopkg.in/yaml.v2"
 
@@ -19,15 +19,10 @@ type runtimeConfigValues struct {
 	Multi kv.MultiRuntimeConfig `yaml:"multi_kv_config"`
 }
 
-func loadRuntimeConfig(filename string) (interface{}, error) {
-	f, err := os.Open(filename)
-	if err != nil {
-		return nil, err
-	}
-
+func loadRuntimeConfig(r io.Reader) (interface{}, error) {
 	var overrides = &runtimeConfigValues{}
 
-	decoder := yaml.NewDecoder(f)
+	decoder := yaml.NewDecoder(r)
 	decoder.SetStrict(true)
 	if err := decoder.Decode(&overrides); err != nil {
 		return nil, err
--- a/pkg/util/runtimeconfig/manager.go
+++ b/pkg/util/runtimeconfig/manager.go
@@ -1,27 +1,36 @@
 package runtimeconfig
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
+	"errors"
 	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
 	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/services"
 )
 
 // Loader loads the configuration from file.
-type Loader func(filename string) (interface{}, error)
+type Loader func(r io.Reader) (interface{}, error)
 
 // ManagerConfig holds the config for an Manager instance.
 // It holds config related to loading per-tenant config.
 type ManagerConfig struct {
 	ReloadPeriod time.Duration `yaml:"period"`
-	LoadPath     string        `yaml:"file"`
-	Loader       Loader        `yaml:"-"`
+	// LoadPath contains the path to the runtime config file, requires an
+	// non-empty value
+	LoadPath string `yaml:"file"`
+	Loader   Loader `yaml:"-"`
 }
 
 // RegisterFlags registers flags.
@@ -44,20 +53,25 @@ type Manager struct {
 	config    interface{}
 
 	configLoadSuccess prometheus.Gauge
+	configHash        *prometheus.GaugeVec
 }
 
 // NewRuntimeConfigManager creates an instance of Manager and starts reload config loop based on config
 func NewRuntimeConfigManager(cfg ManagerConfig, registerer prometheus.Registerer) (*Manager, error) {
-	mgr := Manager{
-		cfg: cfg,
-		configLoadSuccess: prometheus.NewGauge(prometheus.GaugeOpts{
-			Name: "cortex_overrides_last_reload_successful",
-			Help: "Whether the last config reload attempt was successful.",
-		}),
+	if cfg.LoadPath == "" {
+		return nil, errors.New("LoadPath is empty")
 	}
 
-	if registerer != nil {
-		registerer.MustRegister(mgr.configLoadSuccess)
+	mgr := Manager{
+		cfg: cfg,
+		configLoadSuccess: promauto.With(registerer).NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_runtime_config_last_reload_successful",
+			Help: "Whether the last runtime-config reload attempt was successful.",
+		}),
+		configHash: promauto.With(registerer).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cortex_runtime_config_hash",
+			Help: "Hash of the currently active runtime config file.",
+		}, []string{"sha256"}),
 	}
 
 	mgr.Service = services.NewBasicService(mgr.start, mgr.loop, mgr.stop)
@@ -131,7 +145,14 @@ func (om *Manager) loop(ctx context.Context) error {
 // loadConfig loads configuration using the loader function, and if successful,
 // stores it as current configuration and notifies listeners.
 func (om *Manager) loadConfig() error {
-	cfg, err := om.cfg.Loader(om.cfg.LoadPath)
+	buf, err := ioutil.ReadFile(om.cfg.LoadPath)
+	if err != nil {
+		om.configLoadSuccess.Set(0)
+		return err
+	}
+	hash := sha256.Sum256(buf)
+
+	cfg, err := om.cfg.Loader(bytes.NewReader(buf))
 	if err != nil {
 		om.configLoadSuccess.Set(0)
 		return err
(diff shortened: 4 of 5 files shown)


simonswine:
@annanay25 if you get the chance can you take a look at this?

pstibrany:
Thanks for your first PR!

This looks like low-risk change, although I'm not sure what it's trying to solve.

For admin trying to see if his Cortex instances use the same config file, I think pre-expansion hash would make tiny bit more sense.

You make a good point about `-runtime-config.file` – I think it would be nice to extend the support for that one too.

simonswine:
> This looks like low-risk change, although I'm not sure what it's trying to solve.

I think I have buried that in the commit message: This allows to monitor the roll-out of a new config file across the
cluster and can help to detect a mismatch in configuration files.

> For admin trying to see if his Cortex instances use the same config file, I think pre-expansion hash would make tiny bit more sense.

That was my feeling, so I am assuming it's generally the node specific that's expanded using environment variables

> You make a good point about `-runtime-config.file` – I think it would be nice to extend the support for that one too.

I will have a look how this can be achieved, will come back to you on this

pkg/util/runtimeconfig/manager.go
```diff
@@ -44,20 +45,26 @@ type Manager struct {
...
 
 // NewRuntimeConfigManager creates an instance of Manager and starts reload config loop based on config
 func NewRuntimeConfigManager(cfg ManagerConfig, registerer prometheus.Registerer) (*Manager, error) {
 	mgr := Manager{
 		cfg: cfg,
 		configLoadSuccess: prometheus.NewGauge(prometheus.GaugeOpts{
-			Name: "cortex_overrides_last_reload_successful",
-			Help: "Whether the last config reload attempt was successful.",
+			Name: "cortex_runtime_config_last_reload_successful",
+			Help: "Whether the last runtime-config reload attempt was successful.",
```
simonswine, line 57: I have changed that because i found this mismatch between CLI argument and metric name quite confusing. 

Not too sure if this is subject to a deprecation process or if we just should use the `cortex_overrides` for both

pkg/util/runtimeconfig/manager_test.go
```diff
@@ -35,10 +38,10 @@ func (l *TestLimits) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return unmarshal((*plain)(l))
 }
 
-func testLoadOverrides(filename string) (interface{}, error) {
+func testLoadOverrides(filename string) (interface{}, []byte, error) {
```
simonswine, line 41: That is quite a lot of code duplication with `loadRuntimeConfig`, maybe with a func argument we could get create a exported method

simonswine:
I have updated the code to also support the runtime config.

A couple of notes:

* I still need to figure a nice way to observe the metric values for testing this. Are there any good examples how to do this?
* I went for having separate metrics for config and runtime-config. Is that ok, should we use a shared metric with a label `type=startup` and `type=runtime`? 
* Also look at my inline comments

Would be great if I can get some input from one of you @annanay25 / @pstibrany

pstibrany:
> * I still need to figure a nice way to observe the metric values for testing this. Are there any good examples how to do this?

Try to look for usage of `github.com/prometheus/client_golang/prometheus/testutil` package, there are many tests in Cortex codebase, either for individual metrics or entire registry.

> * I went for having separate metrics for config and runtime-config. Is that ok, should we use a shared metric with a label `type=startup` and `type=runtime`?

I think separate metrics is preferable in this case, since it's different components of the system.

> * Also look at my inline comments
Will try.

pkg/cortex/runtime_config.go
```diff
@@ -19,21 +21,30 @@ type runtimeConfigValues struct {
...
 	decoder := yaml.NewDecoder(f)
 	decoder.SetStrict(true)
 	if err := decoder.Decode(&overrides); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return overrides, nil
+	hash := sha256.New()
+	if _, err := f.Seek(0, 0); err != nil {
```
pstibrany, line 40: Let's read the file once, and use it for both parsing and hashing.

pkg/util/runtimeconfig/manager.go
```diff
@@ -14,7 +15,7 @@ import (
 )
 
 // Loader loads the configuration from file.
-type Loader func(filename string) (interface{}, error)
+type Loader func(filename string) (cfg interface{}, hash []byte, error error)
```
pstibrany, line 18: Hmm, changing this will break other users (Loki).

pkg/util/runtimeconfig/manager.go
pstibrany: I think it's OK to change (if you mention it in CHANGELOG.md and update existing alerts too in https://github.com/grafana/cortex-jsonnet repo) but I'm not sure if metric names are actually part of our V1 guarantees. It's not mentioned in the docs/configuration/v1-guarantees.md doc. @gouthamve what do you say?

pkg/util/runtimeconfig/manager.go
```diff
@@ -44,20 +45,26 @@ type Manager struct {
...
 	mgr := Manager{
 		cfg: cfg,
 		configLoadSuccess: prometheus.NewGauge(prometheus.GaugeOpts{
-			Name: "cortex_overrides_last_reload_successful",
-			Help: "Whether the last config reload attempt was successful.",
+			Name: "cortex_runtime_config_last_reload_successful",
+			Help: "Whether the last runtime-config reload attempt was successful.",
 		}),
+		configInfo: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cortex_runtime_config_info",
```
pstibrany, line 60: `_config_info` sounds too generic to me. I would suggest calling it `cortex_runtime_config_hash` with `sha256` label (instead of prefix). Same for metric in main package. (This is just a personal preference though)

pkg/util/runtimeconfig/manager.go
```diff
@@ -44,20 +45,26 @@ type Manager struct {
...
+			Help: "Information about the active runtime config.",
+		}, []string{"hash"}),
 	}
 
 	if registerer != nil {
 		registerer.MustRegister(mgr.configLoadSuccess)
+		registerer.MustRegister(mgr.configInfo)
```
pstibrany, line 67: Since you're touching this code, can we switch to using `promauto.With(registerer)...`? We're trying to do that everywhere in Cortex.

pkg/util/runtimeconfig/manager_test.go
pstibrany: Reason for that duplication is to 1) avoid inverse package dependency, 2) make it possible to use different structures to load config file into.

This code does *not* belong to `runtimeconfig.Manager`, but there can be a helper function to load YAML into specified structure and return its hash as well.

pkg/util/runtimeconfig/manager.go
simonswine: In theory it would probably be better to do the hashing outside of the Loader.

I guess that would require us to read the config file twice, which might actually read two different versions.

I think if we really change the interface here I would suggest we would use a `func(reader io.ReadCloser) (interface{}, error)` and then we can be quite flexible for future use cases

pstibrany: I think that would be fine too...

Current approach gives us a different flexibility -- namely that Loader can load config from anywhere, and can treat `filename` as URL, or even config data itself. But we don't really need that :) Let's use `func(reader io.ReadCloser) (interface{}, error)` and do hashing in Manager itself.

pkg/util/runtimeconfig/manager.go
simonswine: My thinking was that maybe in the future we also want to add more labels, like filepath to the metric, so I have kept it more generic.

I was not too sure if this use case is cover by (https://prometheus.io/docs/practices/naming/):

> foobar_build_info (for a pseudo-metric that provides metadata about the running binary)

pkg/util/runtimeconfig/manager.go
pstibrany: Checked with Goutham, we don't cover metrics in guarantees, so it's ok to rename it. (with changelog and alerts update)

annanay25: Yep, following this lets make the changes in the [cortex-mixin](https://github.com/grafana/cortex-jsonnet) as well.

pkg/util/runtimeconfig/manager.go
annanay25: > maybe in the future we also want to add more labels, like filepath to the metric

We can still add this info in the labels. I agree with Peter about naming it `cortex_runtime_config_hash`

cmd/cortex/main.go
```diff
@@ -184,6 +195,12 @@ func LoadConfig(filename string, expandENV bool, cfg *cortex.Config) error {
 		return errors.Wrap(err, "Error reading config file")
 	}
 
+	// create a sha256 hash of the config before expansion and expose it via
+	// the config_info metric
+	hash := sha256.Sum256(buf)
+	configInfo.Reset()
+	configInfo.WithLabelValues(fmt.Sprintf("sha256:%x", hash)).Set(1)
```
annanay25, line 202: Just took a look at how Alertmanager is doing this, and I think we should follow their approach. 

md5 sum of config loaded from disk, exposed as the metric _value itself_ should work better. Ref: https://github.com/prometheus/alertmanager/pull/751

Will be easier to build dashboards as well as alerts.

pstibrany: I agree it would be easier to build alert, I don't see us building alerts and dashboards for this.

That only works for short checksums (I'd argue it fails for MD5 which is 16 bytes and thus requires 2 float64 values, and even may lead to values like NaN, +Inf). I don't think that is a good pattern, and would stay with the current design.

pstibrany: Even then, building alert is not difficult... eg. `count(count(cortex_config_info) by (hash))` gives you number of different config files. If it's > 1, you can alert.

annanay25: Thanks for explaining @pstibrany; this makes perfect sense. :)

One last thing I'd like to add regarding dashboards, I envision a dashboard for monitoring a new config rollout to a multi-node cluster running the single binary - there it would be easy to track which nodes have picked up the new config if we plot the hash itself (it can be confusing to check labels to monitor the rollout). What would you recommend for that?

pstibrany: Would plotting `count(cortex_config_info) by (hash)` work? One hash would go down to 0 over time (and then disappear), another one up.

simonswine: Yes it think so, that should do what we need

pkg/util/runtimeconfig/manager.go
```diff
@@ -131,7 +139,21 @@ func (om *Manager) loop(ctx context.Context) error {
...
+		if err != nil {
+			om.configLoadSuccess.Set(0)
+			return err
+		}
+		reader = bytes.NewReader(buf)
+		hashSum := sha256.Sum256(buf)
+		hash = hashSum[:]
+	}
+
+	cfg, err := om.cfg.Loader(reader)
```
pstibrany, line 156: This will call loader even if there was no file, `reader` will be nil. I think that's not desirable.

pkg/util/runtimeconfig/manager.go
simonswine: Renamed it in the latest iteration

pkg/util/runtimeconfig/manager.go
simonswine: Maybe we should document a bit more how the loader method should behave, as right now there is not really a comment on. Maybe it needs to deal with a nil reader?

 I basically added that check, so I could make those two test cases work where there is no file provided (`TestOverridesManager_ListenerChannel` and `TestOverridesManager_StopClosesListenerChannels`)

Without a path this whole module wouldn't even be initialised (I think): https://github.com/cortexproject/cortex/blob/981bf9df1b3d8524a593e73405d6c057328ecc76/pkg/cortex/modules.go#L131

What would be a good approach? Rewrite those two test cases to use files instead?

pstibrany: > What would be a good approach? Rewrite those two test cases to use files instead?

Previously Manager wasn't involved in path check, now it is since it opens the file on its own. Tests need to be modified to actually use files now.

I think we should `NewRuntimeConfigManager` to return error if path is empty. Then there is no `nil` reader problem. Since we don't even try to initialize the Manager it in such case from the rest of Cortex, this seems to be easy way out. (Note that Loki may still be doing that... it should be fixed afterwards).

cmd/cortex/main.go
```diff
@@ -29,11 +30,21 @@ var (
 	Revision string
 )
 
+// configInfo exposes information about the loaded config
+var configInfo *prometheus.GaugeVec = prometheus.NewGaugeVec(
```
pracucci, line 34: [nit] We can call it `configHash` given it just exposes the hash.

cmd/cortex/main.go
```diff
@@ -29,11 +30,21 @@ var (
...
+	prometheus.GaugeOpts{
+		Name: "cortex_config_hash",
+		Help: "Hash of the currently active config.",
```
pracucci, line 37: I would add "file" because config can also be specified via CLI flags.

```suggestion
		Help: "Hash of the currently active config file.",
```

pkg/util/runtimeconfig/manager.go
```diff
@@ -44,20 +51,21 @@ type Manager struct {
 	config    interface{}
 
 	configLoadSuccess prometheus.Gauge
+	configInfo        *prometheus.GaugeVec
```
pracucci, line 54: I would call it `configHash` for clarity.

pkg/util/runtimeconfig/manager.go
pracucci: I agree. We shouldn't workaround the code to make tests pass, but improve tests. All tests should load config from file. You can generate a temporary file via code (with deferred deletion) and then load it in the test.

simonswine:
I think I have addressed most of the feedback. Github will not show my pushes for a while, as they have some problems, but once you are on 696b9bfe9, I should have addressed most concerns

pkg/util/runtimeconfig/manager.go
pracucci: @simonswine As Annanay correctly mentioned, please don't forget to open a PR to update the mixin once this PR will be merged, thanks! 🙏 🙏 🙏

pkg/util/runtimeconfig/manager.go
```diff
@@ -44,20 +51,21 @@ type Manager struct {
 	config    interface{}
 
 	configLoadSuccess prometheus.Gauge
+	configHash        *prometheus.GaugeVec
 }
 
 // NewRuntimeConfigManager creates an instance of Manager and starts reload config loop based on config
 func NewRuntimeConfigManager(cfg ManagerConfig, registerer prometheus.Registerer) (*Manager, error) {
 	mgr := Manager{
```
pstibrany, line 59: As cfg.LoadPath is required now, can we add a check and return error if it's empty?

pkg/util/runtimeconfig/manager.go
```diff
@@ -131,7 +139,15 @@ func (om *Manager) loop(ctx context.Context) error {
 // loadConfig loads configuration using the loader function, and if successful,
 // stores it as current configuration and notifies listeners.
 func (om *Manager) loadConfig() error {
-	cfg, err := om.cfg.Loader(om.cfg.LoadPath)
+
```
pstibrany, line 142: Extra empty line.

pkg/util/runtimeconfig/manager_test.go
```diff
@@ -35,24 +42,42 @@ func (l *TestLimits) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 }
 
+func newTestOverridesManagerConfig(t *testing.T, i int32) (*atomic.Int32, ManagerConfig, func()) {
+	var config = atomic.NewInt32(i)
+
+	// create empty file
+	tempFile, err := ioutil.TempFile("", "test-validation")
+	require.NoError(t, err)
+
+	cleanup := func() {
```
pstibrany, line 63: You can now register this function directly to `t.Cleanup`, no need to return it anymore. Makes code tiny bit nicer.

Cleanup function should not report errors. If this error check is part of the test (don't think that is the case here), it should not be in cleanup.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [CHANGE] Added the `engine` label to the metrics exposed by the Prometheus query engine, to distinguish between `ruler` and `querier` metrics. #2854
 * [CHANGE] Added ruler to the single binary when started with `-target=all` (default). #2854
 * [CHANGE] Experimental TSDB: compact head when opening TSDB. This should only affect ingester startup after it was unable to compact head in previous run. #2870
+* [CHANGE] Metric `cortex_overrides_last_reload_successful` has been renamed to `cortex_runtime_config_last_reload_successful` #2874
```
pstibrany, line 20:
```suggestion
* [CHANGE] Metric `cortex_overrides_last_reload_successful` has been renamed to `cortex_runtime_config_last_reload_successful`. #2874
```

(For consistency, same for other added entry)

pkg/util/runtimeconfig/manager.go
simonswine: :+1:

simonswine:
I think most if not all feedback should be addressed now :tada: 

The change in the jsonnet alerts has happened here: https://github.com/grafana/cortex-jsonnet/pull/139

simonswine:
Thanks for your responsive feedback. Have a great weekend :wave:

PR #2880, opened 2020-07-15
Querier: Unset Host Header if DownstreamUrl is configured

### LINKED ISSUE #2875: query-frontend: Different HTTP Headers on /api/v1/query and /api/v1/query_range calls
Hello, the query fronted sets different HTTP Host Headers depending on the Prometheus API call.
In front of Cortex is an Nginx which listen to `api-cortex.example.com`.
There is another Nginx in front of the Prometheus which should listen to `cortex-thanos.example.com`.
Depending on the Prometheus API endpoint Cortex sets different Host Headers.
This causes Nginx to chose the first server_name if finds as `api-cortex.example.com` is not configured.

Cortex is running inside Docker.
Testet with Cortex: v1.0.0, v1.1.0, v1.2.0

cortex.yml
```
auth_enabled: false
target: query-frontend
http_prefix:
server:
  http_listen_port: 9091
  http_listen_address: localhost
  grpc_listen_address: localhost
  log_level: debug
query_range:
  split_queries_by_interval: 24h
  align_queries_with_step: true
  cache_results: true
  results_cache:
    max_freshness: 1m
    cache:
      enable_fifocache: true
      fifocache:
        size: 5120
        validity: 3h
frontend:
  log_queries_longer_than: 5s
  compress_responses: true
  downstream_url: https://cortex-thanos.example.com
```
Nginx Logs:
```
26404#26404: *1 http request line: "GET /api/v1/query?query=1%2B1 HTTP/1.1"
26404#26404: *1 http header: "Host: api-cortex.example.com"
```
```
26405#26405: *2 http request line: "GET /api/v1/query_range?end=1594741440&query=[...]&start=1594684800&step=480 HTTP/1.1"
26405#26405: *2 http header: "Host: cortex-thanos.example.com"
```

tonobo:
**What this PR does**:

This change is necessary to correctly pass the Host Header to backend
service. The Host Header will be set by stdlib, so flushing them should
do the trick.

Fixes #2875

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/frontend/frontend_test.go
+++ b/pkg/querier/frontend/frontend_test.go
@@ -58,8 +58,9 @@ func TestFrontend(t *testing.T) {
 
 		assert.Equal(t, "Hello World", string(body))
 	}
-	testFrontend(t, handler, test, false)
-	testFrontend(t, handler, test, true)
+
+	testFrontend(t, defaultFrontendConfig(), handler, test, false)
+	testFrontend(t, defaultFrontendConfig(), handler, test, true)
 }
 
 func TestFrontendPropagateTrace(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tonobo:
Cool! Thanks!

pkg/querier/frontend/frontend_test.go
```diff
@@ -105,11 +106,67 @@ func TestFrontendPropagateTrace(t *testing.T) {
...
+	go downstreamServer.Serve(downstreamListen)           //nolint:errcheck
+
+	// Configure the query-frontend with the mocked downstream server.
+	config := defaultFrontendConfig()
+	config.DownstreamURL = fmt.Sprintf("http://%s", downstreamListen.Addr())
+
+	// Configure the test to send a request to the query-frontend and assert on the
+	// Host HTTP header received by the downstream server.
+	test := func(addr string) {
+		req, err := http.NewRequest("GET", fmt.Sprintf("http://%s/%s", addr, query), nil)
```
tonobo, line 142: I think your query wouldn't trigger the issue as query_range requests are not directly passed to the downstream if i get it correct.

tonobo: Hmm, strange. we only encountered the issue on instant queries ...

¯\_(ツ)_/¯

PR #2891, opened 2020-07-16
Allow for Custom Transport on Ruler s3 Client

joe-elliott:
**What this PR does**:
- Adds the ability to create an s3 storage client with custom http transport middleware
  - Adds simple test
- Splits out the RuleStore from the Ruler allowing for a custom rulestore backend to be created
- Ultimately this change is just a refactor to allow for more flexiblity for downstream projects.  There should be no behavior change.

cc @jtlisi 

Fixes #nothing

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/chunk/aws/s3_storage_client.go
```diff
@@ -94,7 +98,12 @@ type S3ObjectClient struct {
 
 // NewS3ObjectClient makes a new S3-backed ObjectClient.
 func NewS3ObjectClient(cfg S3Config, delimiter string) (*S3ObjectClient, error) {
-	s3Config, bucketNames, err := buildS3Config(cfg)
+	return NewS3ObjectClientWithMiddleware(cfg, delimiter, nil)
+}
+
+// NewS3ObjectClientWithMiddleware makes a new S3-backed ObjectClient with an option to add request middleware
+func NewS3ObjectClientWithMiddleware(cfg S3Config, delimiter string, injectMiddleware InjectRequestMiddleware) (*S3ObjectClient, error) {
```
jtlisi, line 105: **suggestion**: I think we should avoid adding `InjectRequestMiddleware` as a parameter to this function call. Instead we could make it a field in the config.

We can add it to the s3 config with a nil struct tag like:

```yaml
Inject InjectRequestMiddleware `yaml:"-"`
```

When `NewS3ObjectClient` is called, if the config value is set to nil it will use the default behavior. Otherwise, it will apply the middleware. 

**WDYT**?

joe-elliott: Works for me.  Requested changes pushed.

PR #2896, opened 2020-07-20
Recommend schema v9 and v10 for production

pracucci:
**What this PR does**:
The https://cortexmetrics.io/docs/architecture/ doc recommends schema v10, while https://cortexmetrics.io/docs/configuration/schema-configuration/ recommends v9. I'm proposing to be consistent and always recommend v10.

/cc @weeco 

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/configuration/schema-config-reference.md
+++ b/docs/configuration/schema-config-reference.md
@@ -7,7 +7,7 @@ slug: schema-configuration
 
 Cortex uses a NoSQL Store to store its index and optionally an Object store to store its chunks. Cortex has overtime evolved its schema to be more optimal and better fit the use cases and query patterns that arose.
 
-Currently there are 11 schemas that are used in production but we recommend running with `v10` schema when possible. You can move from one schema to another if a new schema fits your purpose better, but you still need to configure Cortex to make sure it can read the old data in the old schemas.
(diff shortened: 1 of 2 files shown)


pracucci:
> LGTM but I want @bboreham to weigh in. For some use-cases v9 is better, and I think Bryan initially had some apprehensions around making v10 the default.

Ok, will wait for @bboreham feedback. My only take is that we're already recommending v10 in https://cortexmetrics.io/docs/architecture/, so we should either fix it there or in the schema config. Right now it's inconsistent.

Wing924:
This is counter part: v10 have 16x more reads compared to v9.

bboreham:
I would recommend v9 with v10 if you know you will have enormous numbers of series under one row.

pracucci:
@gouthamve @jtlisi @bboreham I've updated the recommendation, mentioning to use v9 for normal use cases and v10 only if expect to have very high cardinality series. What do you think?

docs/configuration/schema-config-reference.md
```diff
@@ -23,7 +23,7 @@ from: <string>
 store: <string>
 # The object client to use. If none is specified, `store` is used for storing chunks as well. Valid options: s3, aws-dynamo, bigtable, bigtable-hashed, gcs, cassandra, filesystem.
 object_store: <string>
-# The schema version to use. Valid ones are v1, v2, v3,... v6, v9, v10, v11. Recommended for production: v9.
+# The schema version to use. Valid ones are v1, v2, v3,... v6, v9, v10, v11. Recommended for production: v10.
```
bboreham, line 26: should be "v9 or v10" ?

pracucci: v9, sorry. My mistake.

PR #2897, opened 2020-07-20
Upgrade grpc to v1.30.0

### LINKED ISSUE #2015: Fix deprecated gRPC naming.Watcher
Found as part of #1912 where we fixed most of the linting issues. The only outstanding item is as follows and has been ignored for now by adding a `//nolint:staticcheck`

```
pkg/querier/frontend/worker.go:60:10: SA1019: naming.Watcher is deprecated: please use package resolver.  (staticcheck)
	watcher naming.Watcher
```

https://github.com/cortexproject/cortex/blob/6774c628af69a54570e33f6a6d875c90f61f6d54/pkg/querier/frontend/worker.go#L60

This will have ripple effects that will go up into where we define the resolver 

https://github.com/cortexproject/cortex/blob/6774c628af69a54570e33f6a6d875c90f61f6d54/pkg/querier/frontend/worker.go#L76-L96

There is no direct Watcher class that exists in the resolver package referenced. So the migration is likely a little more in depth than just using resolver.Watcher or something similar. 

Reference to that Watcher class
https://godoc.org/google.golang.org/grpc/naming#Watcher

codesome:
**What this PR does**:

I was attempting Prometheus upgrade, but grpc is also upgraded with that. grpc v1.30.0 removes the deprecated `google.golang.org/grpc/naming` package which is used by query-frontend. Hence on @pracucci's suggestion, I am attempting the grpc upgrade separately before Prometheus upgrade. 

Fixes #2015

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/frontend/worker.go
+++ b/pkg/querier/frontend/worker.go
@@ -3,19 +3,19 @@ package frontend
 import (
 	"context"
 	"flag"
-	"fmt"
 	"math/rand"
 	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/pkg/errors"
 	"github.com/weaveworks/common/httpgrpc/server"
 	"github.com/weaveworks/common/middleware"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/naming"
+	"google.golang.org/grpc/resolver"
+	"google.golang.org/grpc/serviceconfig"
 
 	"github.com/cortexproject/cortex/pkg/querier"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/grpcclient"
 	"github.com/cortexproject/cortex/pkg/util/services"
 )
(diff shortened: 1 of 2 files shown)


codesome:
There is a big blocker. `go.etcd.io/etcd` [is using grpc v1.26.0](https://github.com/etcd-io/etcd/blob/master/go.mod#L46) and go modules update fails because of that.

pkg/querier/frontend/worker.go
```diff
@@ -88,54 +85,79 @@ func (w *worker) stopping(_ error) error {
...
+// Deprecated in favour of UpdateState.
+func (w *worker) NewServiceConfig(string) {}
 
-		w.resetConcurrency()
+// ParseServiceConfig implements resolver.ClientConn interface.
+func (w *worker) ParseServiceConfig(string) *serviceconfig.ParseResult {
+	return nil
+}
+
+// watchDNSLoop watches for changes in DNS and starts or stops workers.
```
pracucci, line 142: It's mentioning `watchDNSLoop`

codesome:
> I'm a bit puzzled by the changes in the worker. I can't understand how it works, because you instantiate a new gRPC DNS resolver but I can't understand who calls the resolver.ResolveNow() (when using the resolved with a gRPC client, it's the gRPC client calling, but that's not the case here).

From what I understood, it calls `UpdateState` whenever there is an update without needing any trigger from the code, and `ResolveNow` is just a hint to the resolver to resolve now which it can ignore.

> Also, I'm wondering if we need to use the gRPC DNS resolver at all. Looks like we need simple DNS-based service discovery, which is what we're currently achieving with the Thanos DNS resolver, so may be use it here as well, not using gRPC DNS resolver at all.

Yes makes sense. I will take a look at that part.

codesome:
Closing this as the grpc upgrade is at a halt because of etcd.

PR #2900, opened 2020-07-20
Clarified supported S3 endpoints for the blocks storage

### LINKED ISSUE #2885: Clarify S3 endpoints
Currently the documentation for `-experimental.tsdb.s3.endpoint` is "S3 endpoint without schema".  I failed to create a valid setting about four times.

If you use an address that doesn't resolve, this doesn't show up as an error until some time after startup when the shipper tries to use it.  It would be better to check the address and fail immediately.

The documentation for `-s3.url` is "S3 endpoint URL with escaped Key and Secret encoded. If only region is specified as a host, proper endpoint will be deduced. Use inmemory:///<bucket-name> to use a mock in-memory implementation."

We should clarify the docs at least.
We could make the "only region" case work for tsdb, or deprecate that feature, so they are consistent.

As an AWS user, I found this page useful: https://docs.aws.amazon.com/general/latest/gr/s3.html, but there are many S3-compatible services.

The "Key and Secret" bit should also be removed from the text since it's better to pass them in environment variables (or use role-based security).

pracucci:
**What this PR does**:
In this PR I'm clarifying the supported S3 endpoint for the blocks storage.

Fixes #2885

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/storage/backend/s3/config.go
+++ b/pkg/storage/backend/s3/config.go
@@ -25,6 +25,6 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.AccessKeyID, prefix+"s3.access-key-id", "", "S3 access key ID")
 	f.Var(&cfg.SecretAccessKey, prefix+"s3.secret-access-key", "S3 secret access key")
 	f.StringVar(&cfg.BucketName, prefix+"s3.bucket-name", "", "S3 bucket name")
-	f.StringVar(&cfg.Endpoint, prefix+"s3.endpoint", "", "S3 endpoint without schema")
+	f.StringVar(&cfg.Endpoint, prefix+"s3.endpoint", "", "The S3 bucket endpoint. It could be an AWS S3 endpoint listed at https://docs.aws.amazon.com/general/latest/gr/s3.html or the hostname and port of an S3-compatible service in the hostname:port format.")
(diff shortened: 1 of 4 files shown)


pkg/storage/backend/s3/config.go
```diff
@@ -25,6 +25,6 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.AccessKeyID, prefix+"s3.access-key-id", "", "S3 access key ID")
 	f.Var(&cfg.SecretAccessKey, prefix+"s3.secret-access-key", "S3 secret access key")
 	f.StringVar(&cfg.BucketName, prefix+"s3.bucket-name", "", "S3 bucket name")
-	f.StringVar(&cfg.Endpoint, prefix+"s3.endpoint", "", "S3 endpoint without schema")
+	f.StringVar(&cfg.Endpoint, prefix+"s3.endpoint", "", "The S3 bucket endpoint. It could be an AWS S3 endpoint listed at https://docs.aws.amazon.com/general/latest/gr/s3.html or the hostname and port of an S3-compatible service in the hostname:port format.")
```
bboreham, line 28: It's a little clumsy to repeat "hostname and port".  How about "or the address of an S3-compatible service in hostname:port format" ?

pracucci: Sure, done.

PR #2902, opened 2020-07-20
Upgrade Prometheus to release-2.20 branch

codesome:
Signed-off-by: Ganesh Vernekar <cs15btech11018@iith.ac.in>

**What this PR does**:

Upgrades Prometheus to release branch which contains some TSDB fixes

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


go.mod
```diff
@@ -28,49 +28,45 @@ require (
...
 	google.golang.org/grpc v1.29.1
 	gopkg.in/yaml.v2 v2.3.0
-	gopkg.in/yaml.v3 v3.0.0-20200603094226-e3079894b1e8
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	sigs.k8s.io/yaml v1.2.0
 )
 
-replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v36.2.0+incompatible
-
-replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.0+incompatible
```
codesome, line 72: This is also to be noted. There is a big bump here after the removal of these `replace`.

pracucci: You're right! Could you check the azure changelog for any significative change for us, please? I would focus on azure blob storage only.

codesome: Sure

codesome: Hm, the changelog is not useful at all https://github.com/Azure/azure-sdk-for-go/blob/master/CHANGELOG.md

codesome: No issues with `azure-sdk-for-go`, that is something Prometheus imports (for the service discovery) and follow [these guidelines](https://github.com/Azure/azure-sdk-for-go#avoiding-breaking-changes) to avoid breaking changes. 

The blob storage client uses `github.com/Azure/go-autorest` and the upgrade has only 1 breaking change which is not concerning (`By default, the DoRetryForStatusCodes functions will no longer infinitely retry a request when the response returns an HTTP status code of 429 (StatusTooManyRequests). To opt in to the old behavior set autorest.Count429AsRetry to false.`)

We should be good in terms of blob storage.

PR #2903, opened 2020-07-20
Remove global metrics registry and global logger in cache package

yeya24:

**What this PR does**:

This pr removes global prometheus registries and util.logger in the `cache` package. This is good for reusing this package in other project like Thanos.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/cache/background.go
+++ b/pkg/chunk/cache/background.go
@@ -47,17 +47,19 @@ func NewBackground(name string, cfg BackgroundConfig, cache Cache, reg prometheu
 		quit:     make(chan struct{}),
 		bgWrites: make(chan backgroundWrite, cfg.WriteBackBuffer),
 		name:     name,
-		droppedWriteBack: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
-			Namespace: "cortex",
-			Name:      "cache_dropped_background_writes_total",
-			Help:      "Total count of dropped write backs to cache.",
-		}, []string{"name"}).WithLabelValues(name),
+		droppedWriteBack: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Namespace:   "cortex",
+			Name:        "cache_dropped_background_writes_total",
+			Help:        "Total count of dropped write backs to cache.",
+			ConstLabels: prometheus.Labels{"name": name},
+		}),
 
-		queueLength: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
-			Namespace: "cortex",
-			Name:      "cache_background_queue_length",
-			Help:      "Length of the cache background write queue.",
-		}, []string{"name"}).WithLabelValues(name),
+		queueLength: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+			Namespace:   "cortex",
+			Name:        "cache_background_queue_length",
+			Help:        "Length of the cache background write queue.",
+			ConstLabels: prometheus.Labels{"name": name},
+		}),
 	}
 
 	c.wg.Add(cfg.WriteBackGoroutines)
--- a/pkg/chunk/cache/background.go
+++ b/pkg/chunk/cache/background.go
@@ -11,19 +11,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
-var (
-	droppedWriteBack = promauto.NewCounterVec(prometheus.CounterOpts{
-		Namespace: "cortex",
-		Name:      "cache_dropped_background_writes_total",
-		Help:      "Total count of dropped write backs to cache.",
-	}, []string{"name"})
-	queueLength = promauto.NewGaugeVec(prometheus.GaugeOpts{
-		Namespace: "cortex",
-		Name:      "cache_background_queue_length",
-		Help:      "Length of the cache background write queue.",
-	}, []string{"name"})
-)
-
 // BackgroundConfig is config for a Background Cache.
 type BackgroundConfig struct {
 	WriteBackGoroutines int `yaml:"writeback_goroutines"`
@@ -54,14 +41,23 @@ type backgroundWrite struct {
 }
 
 // NewBackground returns a new Cache that does stores on background goroutines.
-func NewBackground(name string, cfg BackgroundConfig, cache Cache) Cache {
+func NewBackground(name string, cfg BackgroundConfig, cache Cache, reg prometheus.Registerer) Cache {
 	c := &backgroundCache{
-		Cache:            cache,
-		quit:             make(chan struct{}),
-		bgWrites:         make(chan backgroundWrite, cfg.WriteBackBuffer),
-		name:             name,
-		droppedWriteBack: droppedWriteBack.WithLabelValues(name),
-		queueLength:      queueLength.WithLabelValues(name),
+		Cache:    cache,
+		quit:     make(chan struct{}),
+		bgWrites: make(chan backgroundWrite, cfg.WriteBackBuffer),
+		name:     name,
+		droppedWriteBack: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Namespace: "cortex",
+			Name:      "cache_dropped_background_writes_total",
+			Help:      "Total count of dropped write backs to cache.",
+		}, []string{"name"}).WithLabelValues(name),
+
+		queueLength: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: "cortex",
+			Name:      "cache_background_queue_length",
+			Help:      "Length of the cache background write queue.",
+		}, []string{"name"}).WithLabelValues(name),
 	}
 
 	c.wg.Add(cfg.WriteBackGoroutines)
--- a/pkg/chunk/cache/fifo_cache.go
+++ b/pkg/chunk/cache/fifo_cache.go
@@ -114,62 +114,69 @@ func NewFifoCache(name string, cfg FifoCacheConfig, reg prometheus.Registerer, l
 		entries:      make(map[string]*list.Element),
 		lru:          list.New(),
 
-		// TODO(bwplotka): There might be simple cache.Cache wrapper for those.
-		entriesAdded: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
-			Namespace: "querier",
-			Subsystem: "cache",
-			Name:      "added_total",
-			Help:      "The total number of Put calls on the cache",
-		}, []string{"cache"}).WithLabelValues(name),
+		entriesAdded: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Namespace:   "querier",
+			Subsystem:   "cache",
+			Name:        "added_total",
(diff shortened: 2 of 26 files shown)


pkg/chunk/cache/background.go
```diff
@@ -54,14 +41,23 @@ type backgroundWrite struct {
...
-		queueLength:      queueLength.WithLabelValues(name),
+		Cache:    cache,
+		quit:     make(chan struct{}),
+		bgWrites: make(chan backgroundWrite, cfg.WriteBackBuffer),
+		name:     name,
+		droppedWriteBack: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Namespace: "cortex",
+			Name:      "cache_dropped_background_writes_total",
+			Help:      "Total count of dropped write backs to cache.",
+		}, []string{"name"}).WithLabelValues(name),
```
pracucci, line 54: You should use `ConstLabels` in `CounterOpts` for any use case where the label value is constant.

_Same comment applies to the rest of the PR._

pkg/chunk/cache/fifo_cache.go
```diff
@@ -173,14 +115,61 @@ func NewFifoCache(name string, cfg FifoCacheConfig) *FifoCache {
...
-		totalGets:       cacheTotalGets.WithLabelValues(name),
-		totalMisses:     cacheTotalMisses.WithLabelValues(name),
-		staleGets:       cacheStaleGets.WithLabelValues(name),
-		memoryBytes:     cacheMemoryBytes.WithLabelValues(name),
+		entriesAdded: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Namespace: "querier",
+			Subsystem: "cache",
+			Name:      "added_total",
+			Help:      "The total number of Put calls on the cache",
+		}, []string{"cache"}).WithLabelValues(name),
```
pracucci, line 123: Use a constant label here as well, please.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -173,14 +115,61 @@ func NewFifoCache(name string, cfg FifoCacheConfig) *FifoCache {
 		lru:          list.New(),
 
 		// TODO(bwplotka): There might be simple cache.Cache wrapper for those.
```
pracucci, line 117: Could you also remove this? Looks like we're not going to do it.

pkg/chunk/cache/memcached.go
```diff
@@ -64,16 +55,25 @@ type Memcached struct {
...
-// TODO(bwplotka): Remove globals & util packages from cache package entirely (e.g util.Logger).
-func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string) *Memcached {
+func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string, reg prometheus.Registerer, logger log.Logger) *Memcached {
+	memcacheRequestDuration := promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "cortex",
+		Name:      "memcache_request_duration_seconds",
+		Help:      "Total time spent in seconds doing memcache requests.",
+		// Memecache requests are very quick: smallest bucket is 16us, biggest is 1s
+		Buckets: prometheus.ExponentialBuckets(0.000016, 4, 8),
+	}, []string{"method", "status_code", "name"})
```
pracucci, line 70: Name should be a constant label.

pkg/chunk/cache/memcached_client.go
```diff
@@ -102,10 +97,15 @@ func NewMemcachedClient(cfg MemcachedClientConfig, name string, r prometheus.Reg
...
+		logger:     logger,
+		provider:   dns.NewProvider(logger, dnsProviderRegisterer, dns.GolangResolverType),
 		quit:       make(chan struct{}),
 
-		numServers: memcacheServersDiscovered.WithLabelValues(name),
+		numServers: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: "cortex",
+			Name:      "memcache_client_servers",
+			Help:      "The number of memcache servers discovered.",
+		}, []string{"name"}).WithLabelValues(name),
```
pracucci, line 108: Constant label here as well.

pkg/chunk/storage/factory.go
```diff
@@ -117,19 +118,20 @@ func (cfg *Config) Validate() error {
 func NewStore(cfg Config, storeCfg chunk.StoreConfig, schemaCfg chunk.SchemaConfig, limits StoreLimits, reg prometheus.Registerer, cacheGenNumLoader chunk.CacheGenNumLoader) (chunk.Store, error) {
 	chunkMetrics := newChunkClientMetrics(reg)
 
-	indexReadCache, err := cache.New(cfg.IndexQueriesCacheConfig)
+	logger := log.NewNopLogger()
```
pracucci, line 121: Uhm, no, we should pick it in input, not using the nop logger :)

pkg/chunk/cache/instrumented.go
```diff
@@ -39,23 +21,33 @@ var (
...
 	}, []string{"name", "method"})
-)
 
-func init() {
-	requestDuration.Register()
-	prometheus.MustRegister(fetchedKeys)
-	prometheus.MustRegister(hits)
-	prometheus.MustRegister(valueSize)
-}
+	requestDuration := promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
```
pracucci, line 25: This is problematic because doesn't have `name` constant label, while it should have in order to avoid clashing metrics (which panic at registration). I'm OK adding the `name` here as well, but please mention it as `[CHANGE]` in the changelog.

pkg/chunk/cache/instrumented.go
```diff
@@ -39,23 +21,33 @@ var (
...
 
-		fetchedKeys:      fetchedKeys.WithLabelValues(name),
-		hits:             hits.WithLabelValues(name),
+		requestDuration: instr.NewHistogramCollector(requestDuration),
+
+		fetchedKeys: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Namespace: "cortex",
+			Name:      "cache_fetched_keys",
+			Help:      "Total count of keys requested from cache.",
+		}, []string{"name"}).WithLabelValues(name),
```
pracucci, line 43: `name` as constant label here as well.

pkg/chunk/storage/factory.go
yeya24: Thanks, yes what I am doing here is totally wrong :(.

But the logger is initialized in the service module, I need to do some work to propagate that logger here

yeya24:
Thanks for the review! @pracucci 
One problem I meet is that we are using `util.WarnExperimentalUse` in the `cache` package, for example https://github.com/cortexproject/cortex/blob/ac796657c651e399e7e37f3e1cadbeed5777d21b/pkg/chunk/cache/redis_cache.go.

Is it possible to get rid of this? This relates to the counter metric here https://github.com/cortexproject/cortex/blob/dd2376550347fbfafd620a0531882e560d975525/pkg/util/experimental.go#L9

pracucci:
> WarnExperimentalUse

I think we wanna keep it. If the problem is the logger, you could just pass the logger to it (in callers outside the `pkg/chunk/cache`, where logger is unavailable, you can just use `util.Logger`). If the problem is the metric, is it really a big deal in Thanos? Doesn't look dramatic exporting `cortex_experimental_features_in_use_total`.

yeya24:
> > WarnExperimentalUse
> 
> I think we wanna keep it. If the problem is the logger, you could just pass the logger to it (in callers outside the `pkg/chunk/cache`, where logger is unavailable, you can just use `util.Logger`). If the problem is the metric, is it really a big deal in Thanos? Doesn't look dramatic exporting `cortex_experimental_features_in_use_total`.

Thanks. I think it is okay. Any thoughts? @bwplotka 

All other comments are addressed except https://github.com/cortexproject/cortex/pull/2903#discussion_r458822633, will mark status as no WIP when I finish.

pkg/chunk/cache/fifo_cache.go
```diff
@@ -172,15 +114,69 @@ func NewFifoCache(name string, cfg FifoCacheConfig) *FifoCache {
...
-		entriesAdded:    cacheEntriesAdded.WithLabelValues(name),
-		entriesAddedNew: cacheEntriesAddedNew.WithLabelValues(name),
-		entriesEvicted:  cacheEntriesEvicted.WithLabelValues(name),
-		entriesCurrent:  cacheEntriesCurrent.WithLabelValues(name),
-		totalGets:       cacheTotalGets.WithLabelValues(name),
-		totalMisses:     cacheTotalMisses.WithLabelValues(name),
-		staleGets:       cacheStaleGets.WithLabelValues(name),
-		memoryBytes:     cacheMemoryBytes.WithLabelValues(name),
+		entriesAdded: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Namespace:   "querier",
```
pstibrany, line 118: These metric names look very weird to me, but that's not part of this PR.

pkg/chunk/cache/fifo_cache_test.go
```diff
@@ -37,7 +39,7 @@ func TestFifoCacheEviction(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		c := NewFifoCache(test.name, test.cfg)
+		c := NewFifoCache(test.name, test.cfg, prometheus.NewRegistry(), log.NewNopLogger())
```
pstibrany, line 42: Nit: I would prefer to pass nil as registerer, if we're not testing metrics.

pkg/chunk/cache/memcached.go
```diff
@@ -64,20 +55,28 @@ type Memcached struct {
...
+
+	logger log.Logger
 }
 
 // NewMemcached makes a new Memcache.
-// TODO(bwplotka): Fix metrics, get them out of globals, separate or allow prefixing.
-// TODO(bwplotka): Remove globals & util packages from cache package entirely (e.g util.Logger).
-func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string) *Memcached {
+func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string, reg prometheus.Registerer, logger log.Logger) *Memcached {
+	memcacheRequestDuration := promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
```
pstibrany, line 64: Nit: this could be inlined.

pkg/chunk/cache/instrumented.go
```diff
@@ -6,58 +6,54 @@ import (
...
+		ConstLabels: prometheus.Labels{"name": name},
+	}, []string{"method"})
 
-func init() {
-	requestDuration.Register()
-	prometheus.MustRegister(fetchedKeys)
-	prometheus.MustRegister(hits)
-	prometheus.MustRegister(valueSize)
-}
+	requestDuration := promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
```
pstibrany, line 26: Nit: This can be inlined.

PR #2904, opened 2020-07-21
Manipulate ingesters query min time when -querier.query-ingesters-within is set

pracucci:
**What this PR does**:
When running the Cortex blocks storage, the blocks retention in the ingesters may be way higher than the configured `-querier.query-ingesters-within` (eg. in the cortex-mixin we do have a default of 4 days blocks retention in the ingesters, to give operators enough time to recover from any issue causing blocks not being shipped from ingesters to storage, in case of an outage).

Running the blocks storage with a `-querier.query-ingesters-within` significantly higher than the retention means that (under normal conditions) all samples older than `-querier.query-store-after` and more recent than `-querier.query-ingesters-within` are both fetched from ingesters and storage, causing an undesired extra load on ingesters.

Similarly to what we do with `-querier.query-store-after` in the blocks storage, in this PR I'm proposing to manipulate the min time range of the query sent to ingesters to guarantee not being older than `-querier.query-ingesters-within`. This change applies **both to chunks and blocks storage**.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany:
My only concern is that test seems to rely on `time.Now()` in test and `time.Now()` in distributor queryable to not cross the 1s boundary, and may be flaky.

PR #2905, opened 2020-07-21
fix for when sharding is enable query frontend returns 500 for storage errors instead of actual error code

sandeepsukhani:
**What this PR does**:
When sharding middleware is enabled in query-frontend and storage returns an error with a status code, user gets 500 error code instead of the actual error code from the error. This is because we get a wrapped error from query engine which the code [here](https://github.com/cortexproject/cortex/blob/7b2283ff8f296fd231bcefae64386524842e5ed9/vendor/github.com/gogo/status/status.go#L144) fails to convert to an expected type to get the error code from.
I have seen the problem happening only with range queries being done for chunks storage and the query not having a metric name.
This PR fixes that issue by returning the cause of the issue using `errors.Cause`.

### DIFF
--- a/development/tsdb-blocks-storage-s3-single-binary/config/cortex.yaml
+++ b/development/tsdb-blocks-storage-s3-single-binary/config/cortex.yaml
@@ -57,6 +57,10 @@ tsdb:
 storage:
   engine: tsdb
 
+querier:
+  query_ingesters_within: 3h
+  query_store_after: 2h
+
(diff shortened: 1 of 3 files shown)


sandeepsukhani:
> LGTM! I just pushed a commit to add a code comment to explain why we do it. @sandeepsukhani could you check it, please?

LGTM! Thanks for taking care of it!

PR #2914, opened 2020-07-22
Move the fakeauth out of the Cortex module so it can be reused in other projects.

tomwilkie:


tomwilkie:
Thank you @pracucci :-)

PR #2916, opened 2020-07-22
docs for performing deletion in cortex chunks storage

sandeepsukhani:
**What this PR does**:
Adds docs on how to perform deletion in Cortex with its features.

**Checklist**
- [x] Documentation added

### DIFF
--- a/docs/guides/deleting-series.md
+++ b/docs/guides/deleting-series.md
@@ -1,0 +1,47 @@
+---
+title: "Deleting Series"
+linkTitle: "Deleting Series"
+weight: 3
+slug: deleting-series
+---
+
+:warning: This feature is currently experimental and is only supported for Chunks storage. :warning:
+
+Cortex supports deletion of series using [Prometheus compatible API](https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series).
+It however does not support [Prometheuses Clean Tombstones](https://prometheus.io/docs/prometheus/latest/querying/api/#clean-tombstones) API by trading it in for another feature for allowing cancellation of delete requests.
+
+### How it works
+
+A new service called `purger` is added which handles deletion APIs and does the processing of the requests.
+To store the requests, and some additional information while performing deletion it requires configuring an index and object store respectively for it.
+`purger` can be configured using the config defined [here](../configuration/config-file-reference.md#purger_config)
+
+#### Requesting Deletion
+
+By calling the `delete_series` API defined [here](https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series), you can request the deletion of series.
+The data would be deleted based on the configuration for `-purger.delete-request-cancel-period` flag which defaults to 24 hours.
+The delete requests would be picked up for processing only after they have aged for the duration configured using that flag.
+Cortex would do query time filtering of the data requested for deletion until the `purger` is done processing the delete requests.
+
+#### Cancellation of Delete Request
+
+Cortex allows cancellation of delete requests until they are not picked up for processing, which is controlled by a flag as explained in `Requesting Deletion` section.
+Since Cortex does query time filtering of data request for deletion until it is actually deleted, you can take an informed decision to cancel the delete request by calling the API defined below:
+
+```
+POST /api/v1/admin/tsdb/cancel_delete_request?request_id=<request_id>
+PUT /api/v1/admin/tsdb/cancel_delete_request?request_id=<request_id>
+```
+
+You can find the id of the request that you want to cancel by using the GET `delete_series` API defined below.
+
+#### Listing Delete Requests
+
+You can list the created delete requests using following API:
+
+```
+GET /api/v1/admin/tsdb/delete_series
+```
+
+**NOTE:** Cancelled delete requests are removed from the store, so they would not be there in the list of requests returned by the API.
+
--- a/docs/guides/deleting-series.md
+++ b/docs/guides/deleting-series.md
@@ -16,12 +16,23 @@ A new service called `purger` is added which handles deletion APIs and does the 
 To store the requests, and some additional information while performing deletion it requires configuring an index and object store respectively for it.
 `purger` can be configured using the config defined [here](../configuration/config-file-reference.md#purger_config)
 
+All the requests specified below needs to be sent to `purger`.
+
+**Note:** If you have enabled auth then same as other APIs it is required to have `X-Scope-OrgID` header set in the requests.
+
 #### Requesting Deletion
 
 By calling the `delete_series` API defined [here](https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series), you can request the deletion of series.
 The data would be deleted based on the configuration for `-purger.delete-request-cancel-period` flag which defaults to 24 hours.
 The delete requests would be picked up for processing only after they have aged for the duration configured using that flag.
 Cortex would do query time filtering of the data requested for deletion until the `purger` is done processing the delete requests.
+
+_Sample cURL command:_
+```
+curl -X POST \
+  '<purger_addr>/api/v1/admin/tsdb/delete_series?match%5B%5D=up&start=1591616227&end=1591619692' \
+  -H 'x-scope-orgid: <orgid>'
+```
 
 #### Cancellation of Delete Request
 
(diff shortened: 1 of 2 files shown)


docs/guides/deleting-series.md
```diff
@@ -0,0 +1,47 @@
...
+Cortex supports deletion of series using [Prometheus compatible API](https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series).
+It however does not support [Prometheuses Clean Tombstones](https://prometheus.io/docs/prometheus/latest/querying/api/#clean-tombstones) API by trading it in for another feature for allowing cancellation of delete requests.
+
+### How it works
+
+A new service called `purger` is added which handles deletion APIs and does the processing of the requests.
+To store the requests, and some additional information while performing deletion it requires configuring an index and object store respectively for it.
+`purger` can be configured using the config defined [here](../configuration/config-file-reference.md#purger_config)
+
+#### Requesting Deletion
```
gouthamve, line 19: Which service should we hit? Can we include some curl commands? Specifically, we should also include the `X-Scope-OrgId` header, etc.

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
+---
+title: "Deleting Series"
+linkTitle: "Deleting Series"
+weight: 3
+slug: deleting-series
+---
+
+:warning: This feature is currently experimental and is only supported for Chunks storage. :warning:
```
pracucci, line 8: There's no emoji replacement  :)

```suggestion
_This feature is currently experimental and is only supported for Chunks storage._
```

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+All the requests specified below needs to be sent to `purger`.
+
+**Note:** If you have enabled auth then same as other APIs it is required to have `X-Scope-OrgID` header set in the requests.
+
+#### Requesting Deletion
+
+By calling the `delete_series` API defined [here](https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series), you can request the deletion of series.
+The data would be deleted based on the configuration for `-purger.delete-request-cancel-period` flag which defaults to 24 hours.
+The delete requests would be picked up for processing only after they have aged for the duration configured using that flag.
+Cortex would do query time filtering of the data requested for deletion until the `purger` is done processing the delete requests.
```
pstibrany, line 28: This is true even before cancel period, right? Should it be mentioned?

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+title: "Deleting Series"
+linkTitle: "Deleting Series"
+weight: 3
+slug: deleting-series
+---
+
+:warning: This feature is currently experimental and is only supported for Chunks storage. :warning:
+
+Cortex supports deletion of series using [Prometheus compatible API](https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series).
+It however does not support [Prometheuses Clean Tombstones](https://prometheus.io/docs/prometheus/latest/querying/api/#clean-tombstones) API by trading it in for another feature for allowing cancellation of delete requests.
```
pracucci, line 11: Is "Prometheuses" intentional? Or should be "Prometheus'"

pracucci: Not sure "by trading it in for another feature for allowing cancellation of delete requests" is easy to understand. Maybe we could just say "because Cortex uses a different mechanism to manage deletions".

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+
+### How it works
+
+A new service called `purger` is added which handles deletion APIs and does the processing of the requests.
```
pracucci, line 15:
```suggestion
A new service called `purger` is added which exposes deletion APIs and does the processing of the requests.
```

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+To store the requests, and some additional information while performing deletion it requires configuring an index and object store respectively for it.
```
pracucci, line 16:
```suggestion
To store the requests, and some additional information while performing deletions, the purger requires configuring an index and object store respectively for it.
```

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+`purger` can be configured using the config defined [here](../configuration/config-file-reference.md#purger_config)
```
pracucci, line 17:
```suggestion
For more information about the `purger` configuration, please refer to the [config file reference](../configuration/config-file-reference.md#purger_config) documentation.
```

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+
+All the requests specified below needs to be sent to `purger`.
+
+**Note:** If you have enabled auth then same as other APIs it is required to have `X-Scope-OrgID` header set in the requests.
```
pracucci, line 21:
```suggestion
**Note:** If you have enabled multi-tenancy in your Cortex cluster then deletion APIs requests require to have the `X-Scope-OrgID` header set like for any other Cortex API.
```

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### Requesting Deletion
+
+By calling the `delete_series` API defined [here](https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series), you can request the deletion of series.
```
pracucci, line 25: A couple of suggestions:
1. Instead of `delete_series` mention the exact endpoint for clarity (eg. `/api/v1/admin/tsdb/delete_series`)
2. Try to avoid to link "here" and instead link some words related to the actual content in the link (eg. "here" > "Prometheus deletion API")

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+The data would be deleted based on the configuration for `-purger.delete-request-cancel-period` flag which defaults to 24 hours.
```
pracucci, line 26: I'm not sure this is clear enough. You could mention that deletion requests are immediately honored and deleted series are hidden from results (soft deletion), but the actual hard deletion of the series from the storage happens only after `-purger.delete-request-cancel-period` in order to give Cortex operators enough time to eventually cancel the request.

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+The delete requests would be picked up for processing only after they have aged for the duration configured using that flag.
+Cortex would do query time filtering of the data requested for deletion until the `purger` is done processing the delete requests.
+
+_Sample cURL command:_
+```
+curl -X POST \
+  '<purger_addr>/api/v1/admin/tsdb/delete_series?match%5B%5D=up&start=1591616227&end=1591619692' \
+  -H 'x-scope-orgid: <orgid>'
```
pracucci, line 34: I would change `<orgid>` to `<tenant-id>`. In the doc we always talk about multi-tenancy, so calling it "tenant" instead of "org" could help clarifying it.

_Same comment applies below._

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+#### Cancellation of Delete Request
+
+Cortex allows cancellation of delete requests until they are not picked up for processing, which is controlled by a flag as explained in `Requesting Deletion` section.
```
pracucci, line 39: Don't be afraid to repeat a flag for clarity.

```suggestion
Cortex allows cancellation of delete requests until they are not picked up for processing, which is controlled by the `-purger.delete-request-cancel-period` CLI flag or its respective YAML config option.
```

docs/guides/deleting-series.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+_Sample cURL command:_
+```
+curl -X GET \
+  <purger_addr>/api/v1/admin/tsdb/delete_series \
+  -H 'x-scope-orgid: <orgid>'
+```
+
+**NOTE:** Cancelled delete requests are removed from the store, so they would not be there in the list of requests returned by the API.
```
pracucci, line 71: What's about processed requests? You should mention the behaviour for them too.

docs/guides/deleting-series.md
sandeepsukhani: >Is "Prometheuses" intentional? Or should be "Prometheus'"
Yes, it is intentional. I can remove it if it looks weird.

PR #2926, opened 2020-07-24
retry processing of failed delete requests

sandeepsukhani:
**What this PR does**:
Adds retries for processing of failed delete requests.
It also fixes a bug in `chunk.Client` which was not encoding keys in `DeleteChunk` but it was in `PutChunk` and `GetChunk` causing it to fail during deletes.

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 * [ENHANCEMENT] Add TLS support to etcd client. #3102
 * [ENHANCEMENT] When a tenant accesses the Alertmanager UI or its API, if we have valid `-alertmanager.configs.fallback` we'll use that to start the manager and avoid failing the request. #3073
 * [ENHANCEMENT] Add `DELETE api/v1/rules/{namespace}` to the Ruler. It allows all the rule groups of a namespace to be deleted. #3120
+* [ENHANCEMENT] Experimental Delete Series: Retry processing of Delete requests during failures. #2926
 * [BUGFIX] Query-frontend: Fixed rounding for incoming query timestamps, to be 100% Prometheus compatible. #2990
 * [BUGFIX] Querier: Merge results from chunks and blocks ingesters when using streaming of results. #3013
 * [BUGFIX] Querier: query /series from ingesters regardless the `-querier.query-ingesters-within` setting. #3035
--- a/pkg/chunk/purger/purger.go
+++ b/pkg/chunk/purger/purger.go
@@ -221,7 +221,7 @@ func (p *Purger) retryFailedRequests() {
 	for _, userID := range userIDsWithFailedRequest {
 		deleteRequest := p.inProcessRequests.get(userID)
 		if deleteRequest == nil {
-			level.Error(util.Logger).Log("msg", "expected an in-process delete request", "user-id", userID)
+			level.Error(util.Logger).Log("msg", "expected an in-process delete request", "user", userID)
 			continue
 		}
 
@@ -229,7 +229,7 @@ func (p *Purger) retryFailedRequests() {
 		err := p.resumeStalledRequest(*deleteRequest)
 		if err != nil {
 			reqWithLogger := makeDeleteRequestWithLogger(*deleteRequest, util.Logger)
-			level.Error(reqWithLogger.logger).Log("msg", "failed to resume failed request")
+			level.Error(reqWithLogger.logger).Log("msg", "failed to resume failed request", "err", err)
 		}
 	}
 }
--- a/pkg/chunk/purger/purger.go
+++ b/pkg/chunk/purger/purger.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/gogo/protobuf/proto"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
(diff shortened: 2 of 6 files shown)


sandeepsukhani:
I have written tests which rely on changing file permissions to fail some of the operations in execution of delete requests but due to circleci running tests as a root user, it is still able to modify/create the files. I do not see any other easy way to write the tests without changing file permissions. Maybe run the tests without root user in circleci, if possible?

sandeepsukhani:
> I have written tests which rely on changing file permissions to fail some of the operations in execution of delete requests but due to circleci running tests as a root user, it is still able to modify/create the files. I do not see any other way to write the test for retries without changing file permissions. Maybe run the tests without root user in circleci, if possible?

I managed to fix the test by adding a mode to `chunk.MockStorage` :smile:

pkg/chunk/purger/purger.go
```diff
@@ -296,9 +323,15 @@ func (p *Purger) worker() {
 	}
 }
 
-func (p *Purger) executePlan(userID, requestID string, planNo int, logger log.Logger) error {
+func (p *Purger) executePlan(userID, requestID string, planNo int, logger log.Logger) (err error) {
 	logger = log.With(logger, "plan_no", planNo)
 
+	defer func() {
+		if err != nil {
+			p.inProcessRequests.setFailedRequestForUser(userID)
```
pstibrany, line 331: Instead of using `defer`, it would be clearer to explicitly call `setFailedRequestForUser` in branches where it is needed.

I don't think all errors need to be retried (eg. when parsing of chunkID fails).

Same comment applies to `buildDeletePlan`.

pstibrany:
I think this deserves ENHANCEMENT changelog entry, WDYT?

pkg/chunk/purger/purger.go
sandeepsukhani: I could not point out any acceptable failures which is why I considered all failures must be retried.
Having a non-parseable chunk id is something which needs someone's intervention because even queries would start failing.

We can always change this if we see cases where we are unnecessarily retrying a non-recoverable and ingorable case. Missing or ignoring a case like non-parseable chunkID could be something which is not expected to happen.

What do you think?

sandeepsukhani:
> I think this deserves ENHANCEMENT changelog entry, WDYT?

Sounds good. Will add it. Thanks for the review!

pkg/chunk/purger/purger.go
pstibrany: My thinking was more that I don't see a point in repeating failed request, if we expect the same failure (eg. that chunk parsing -- I would expect that to just fail over and over, or not?).

sandeepsukhani: Yes, it would fail over and over until someone manually intervenes and fixes the problem. The fix would be either fixing the chunk or removing it. Same would happen when someone queries data touching that chunk.
It could a problem which needs to be noticed by someone and not everyone would go and check the logs for errors like this.

pstibrany: I don't think wasting resources on jobs we know will fail again and again is a good approach. It should raise an alert instead.

sandeepsukhani: It would alert only if it fails. We have alerts setup for oldest pending request.
We also have a dashboard for showing failures in processing of request. Constant failures would be noticed but a single failure once in a while is not noticeable.

CHANGELOG.md
```diff
@@ -89,6 +89,7 @@
 * [ENHANCEMENT] Added metrics `cortex_config_hash` and `cortex_runtime_config_hash` to expose hash of the currently active config file. #2874
 * [ENHANCEMENT] Logger: added JSON logging support, configured via the `-log.format=json` CLI flag or its respective YAML config option. #2386
 * [ENHANCEMENT] Clarify limitations of the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints. #2953
+* [ENHANCEMENT] Experimental Delete Series: Retry processing of Delete requests during failures. #2926
```
pracucci, line 92: We're cutting Cortex `1.3.0` and this PR will be merged after that. Please rebase `master` and move the CHANGELOG entry under the `master / unreleased` section at the top of the CHANGELOG. Thanks!

pkg/chunk/purger/purger.go
```diff
@@ -207,6 +215,25 @@ func (p *Purger) stop(_ error) error {
 	return nil
 }
 
+func (p *Purger) retryFailedRequests() {
+	userIDsWithFailedRequest := p.inProcessRequests.listUsersWithFailedRequest()
+
+	for _, userID := range userIDsWithFailedRequest {
+		deleteRequest := p.inProcessRequests.get(userID)
+		if deleteRequest == nil {
+			level.Error(util.Logger).Log("msg", "expected an in-process delete request", "user-id", userID)
```
pracucci, line 224:
```suggestion
			level.Error(util.Logger).Log("msg", "expected an in-process delete request", "user", userID)
```

pkg/chunk/purger/purger.go
```diff
@@ -207,6 +215,25 @@ func (p *Purger) stop(_ error) error {
...
+			continue
+		}
+
+		p.inProcessRequests.unsetFailedRequestForUser(userID)
+		err := p.resumeStalledRequest(*deleteRequest)
+		if err != nil {
+			reqWithLogger := makeDeleteRequestWithLogger(*deleteRequest, util.Logger)
+			level.Error(reqWithLogger.logger).Log("msg", "failed to resume failed request")
```
pracucci, line 232:
```suggestion
			level.Error(reqWithLogger.logger).Log("msg", "failed to resume failed request", "err", err)
```

pracucci (commented):
Please remember to rebase this PR and move the CHANGELOG entry under "master / unreleased" section.

gouthamve (approved):
LGTM!

PR #2933, opened 2020-07-27
Clarify local website setup and upgrade deps

pracucci:
**What this PR does**:
While reviewing the PR https://github.com/cortexproject/cortex/pull/2895 I've noticed there's no guarantee the website's local dev env runs the same dependencies version of CI. Unfortunately `npm install -g` (global) doesn't support `package.json` so with my limited JS knowledge I haven't found a way to use `website/package.json*` in the `build-image`, but if there's any I will be very happy to reiterate on this.

/cc @cyriltovena 

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/build-image/Dockerfile
+++ b/build-image/Dockerfile
@@ -5,7 +5,11 @@ RUN apt-get update && apt-get install -y curl python-requests python-yaml file j
 	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
 RUN apt-get install -y nodejs npm && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-RUN npm install -g postcss-cli autoprefixer
+
+# Install website builder dependencies. Whenever you change these version, please also change website/package.json
+# and viceversa.
+RUN npm install -g postcss-cli@5.0.1 autoprefixer@9.8.5
+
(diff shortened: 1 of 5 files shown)


build-image/Dockerfile
```diff
@@ -5,7 +5,11 @@ RUN apt-get update && apt-get install -y curl python-requests python-yaml file j
 	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
 RUN apt-get install -y nodejs npm && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-RUN npm install -g postcss-cli autoprefixer
+
+# Install website builder dependencies. Whenever you change these version, please also change website/package.json
+# and viceversa.
+RUN npm install -g postcss-cli@5.0.1 autoprefixer@9.8.5
```
cyriltovena, line 11: You could may be have added the package.json inside the image. Not sure.

pracucci: I tried. The problem is that dependencies from `package.json` cannot be installed globally (`npm install -g` doesn't read `package.json` or the lock), so I'm not sure how to achieve it. Do you have any idea?

pracucci:
> 🙈LGTM, but I'm no expert on this stuff.

No worries. None is. Luckily we're better at scaling out Prometheus 😊

PR #2937, opened 2020-07-27
Cleanup blocks storage doc

pracucci:
**What this PR does**:
Over the past few months we did several changes to the experimental blocks storage. If hypothetically we would build it from scratch today, the config would probably look different and cleaner. In this PR I'm attempting to cleanup a bit the blocks storage config, in particular:

- The storage engine has been renamed from `tsdb` to `blocks` (we already call it "blocks storage" in the doc)
- The root level config option has been renamed from `tsdb` to `blocks_storage`
- The real TSDB config (as TSDB opened in the ingesters) has been moved under `blocks_storage` > `tsdb`

Changes are not huge but annoying like any config change. Since we're getting close to mark the blocks storage production grade, I think this is one of the last chances we have to cleanup a bit the config.

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/chunk/storage/factory.go
```diff
@@ -92,14 +92,14 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.Swift.RegisterFlags(f)
 	cfg.GrpcConfig.RegisterFlags(f)
 
-	f.StringVar(&cfg.Engine, "store.engine", "chunks", "The storage engine to use: chunks or tsdb. Be aware tsdb is experimental and shouldn't be used in production.")
+	f.StringVar(&cfg.Engine, "store.engine", "chunks", "The storage engine to use: chunks or blocks. Be aware blocks storage is experimental and shouldn't be used in production.")
```
pstibrany, line 95:
```suggestion
	f.StringVar(&cfg.Engine, "store.engine", "chunks", "The storage engine to use: chunks or blocks. Be aware that blocks storage is experimental and shouldn't be used in production.")
```

pkg/storage/tsdb/config.go
```diff
@@ -56,34 +56,17 @@ var (
 	errEmptyBlockranges             = errors.New("empty block ranges for TSDB")
 )
 
-// Config holds the config information for TSDB storage
+// Config holds the config information for the blocks storage.
 type Config struct {
```
pstibrany, line 60: Do we want to rename this to `BlocksStorageConfig`? (Right now, the full name is `tsdb.Config`, which looks bit weird considering we also have `tsdb.TSDBConfig`. Perhaps we should rename the package instead. Not necessarily in this PR)

pracucci: Agree. Let's rename it.

pracucci:
> We should also cleanup naming of internal types (V2, "tsdb" package), but that's for another PR.

Agree (on both).

pracucci:
@pstibrany I've addressed your comments, thanks.

PR #2941, opened 2020-07-28
Translate errors to PromQL errors by wrapping queryable.

pstibrany:
**What this PR does**: This is a refinement of PR #2571, which however only translated errors coming from the store, and ignored errors from ingesters. By registrering error-translation layer at API creation time, it will cover all possible error paths.

This PR also changes errors reported by various queryables to default to 500 instead of previous 422.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany:
Thanks Marco, I've fixed the problem -- it was caused by distributor_queryable.go wrapping *all* errors coming from ingester into `promql.ErrStorage` -- which is reported as 500.

I've removed `promql.ErrStorage` from most places where it was used, as new queryable error-translation feature introduced in this PR reports all unknown errors as `promql.ErrStorage` by default.

This change of default is a potentionally breaking change for *other places* which didn't use `promql.ErrStorage` wrapping before -- such errors were reported as 422. Now they will be wrapped into `ErrStorage` and thus 500. (In chunk_store_queryable.go, unknown errors already defaulted to ErrStorage wrapping though).

I think user-errors reported by queryables should use some explicit errors and we will add them to `pkg/api/queryable.go`.

pkg/querier/blocks_store_queryable.go
```diff
@@ -291,14 +290,7 @@ type blocksStoreQuerier struct {
 // Select implements storage.Querier interface.
 // The bool passed is ignored because the series is always sorted.
 func (q *blocksStoreQuerier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
-	set := q.selectSorted(sp, matchers...)
-
-	// We need to wrap the error in order to have Prometheus returning a 5xx error.
-	if err := set.Err(); err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
```
pstibrany, line 297: @pracucci: I'm not sure about this change... here if error was context.Canceled or DeadlineExceeded, it was reported to prom/api directly. Previously it meant 422, now it means 500. Not sure which one is better. For canceled -- if we can assume it was cancelled by user -- I think we should still return 422. What about DeadlineExceeded?

pracucci: `context.Canceled` should definitely by a 422. `context.DeadlineExceeded` is probably more fair to be considered as 500, but I'm open to discuss it.

pstibrany: Let's keep `context.DeadlineExceeded` as 500 for now (I don't see any "deadline exceeded" errors returned by our queriers or query-frontends in last week).

pstibrany:
@gotjosh @jtlisi Should ruler use this error-translation as well? (How exactly does ruler handle errors from querier?)

PR #2946, opened 2020-07-29
integration test for deletion of series

sandeepsukhani:
**What this PR does**:
Add integration tests for delete series, testing deletes for all index stores.
Fixes status code for add and cancel APIs to return `204` instead of `200`.

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
     - `-experimental.tsdb.flush-blocks-on-shutdown` changed to `-experimental.blocks-storage.tsdb.flush-blocks-on-shutdown`
 * [CHANGE] Flags `-bigtable.grpc-use-gzip-compression`, `-ingester.client.grpc-use-gzip-compression`, `-querier.frontend-client.grpc-use-gzip-compression` are now deprecated. #2940
 * [CHANGE] Limit errors reported by ingester during query-time now return HTTP status code 422. #2941
+* [CHANGE] Experimental Delete Series: `/api/v1/admin/tsdb/delete_series` and `/api/v1/admin/tsdb/cancel_delete_request` purger APIs to return status code `204` instead of `200` for success. #2946
 * [FEATURE] Introduced `ruler.for-outage-tolerance`, Max time to tolerate outage for restoring "for" state of alert. #2783
 * [FEATURE] Introduced `ruler.for-grace-period`, Minimum duration between alert and restored "for" state. This is maintained only for alerts with configured "for" time greater than grace period. #2783
 * [FEATURE] Introduced `ruler.resend-delay`, Minimum amount of time to wait before resending an alert to Alertmanager. #2783
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * `cortex_ingester_received_files`
   * `cortex_ingester_received_bytes_total`
   * `cortex_ingester_sent_bytes_total`
+* [CHANGE] Experimental Delete Series: `/api/v1/admin/tsdb/delete_series` and `/api/v1/admin/tsdb/cancel_delete_request` purger APIs to return status code `204` instead of `200` for success. #2946
 * [ENHANCEMENT] Add support for azure storage in China, German and US Government environments. #2988
 * [ENHANCEMENT] Query-tee: added a small tolerance to floating point sample values comparison. #2994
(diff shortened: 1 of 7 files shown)


integration/e2ecortex/client.go
```diff
@@ -358,3 +359,17 @@ func (c *Client) DeleteAlertmanagerConfig(ctx context.Context) error {
 
 	return nil
 }
+
+func (c *Client) PostRequest(url string, body io.Reader) (*http.Response, error) {
```
sandeepsukhani, line 363: I have not added a dedicated method for calling `delete_series` API here because it is not common usage. Adding a dedicated method would require changing the factory method for `Client` to accept the purger URL and in turn change all the usages to pass an empty string.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
     - `-experimental.tsdb.stripe-size` changed to `-experimental.blocks-storage.tsdb.stripe-size`
     - `-experimental.tsdb.wal-compression-enabled` changed to `-experimental.blocks-storage.tsdb.wal-compression-enabled`
     - `-experimental.tsdb.flush-blocks-on-shutdown` changed to `-experimental.blocks-storage.tsdb.flush-blocks-on-shutdown`
+* [CHANGE] Experimental Delete Series: `/api/v1/admin/tsdb/delete_series` and `/api/v1/admin/tsdb/cancel_delete_request` purger APIs to return status code `204` instead of `200` for success. #2946
```
pstibrany, line 43: I've recently found it more difficult to use endpoint that returned 204 from scripts (curl/wget both don't like 204 very much). I understand it's "nicer" from HTTP point of view, but 200 is more pragmatic.

sandeepsukhani: Actually this is to make API compatible with Prometheuses API which returns `204`. :shrug:

pstibrany: Makes sense.

integration/chunks_delete_series_test.go
```diff
@@ -0,0 +1,221 @@
...
+	// stop the purger and recreate it since we load requests for deletion every hour.
+	require.NoError(t, s.Stop(purger))
+
+	purger = e2ecortex.NewPurger("purger", mergeFlags(flags, map[string]string{
+		"-purger.delete-request-cancel-period": "-1m", // we retain delete requests for a day(default)+1m, changing it to a negative value so that delete requests get picked up immediately.
+	}), "")
+	purger.HTTPService.SetEnvVars(bigtableFlag)
+	require.NoError(t, s.StartAndWaitReady(purger))
+
+	time.Sleep(time.Second)
```
pracucci, line 176: I was testing this PR and this makes the test flaky (because may need more than 1 second). Let me get https://github.com/cortexproject/cortex/pull/2522 merged and then I will commit a change to this PR to fix it based on the PR #2522 I'm currently working on.

pracucci: Fixed it!

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
     - `-experimental.tsdb.flush-blocks-on-shutdown` changed to `-experimental.blocks-storage.tsdb.flush-blocks-on-shutdown`
 * [CHANGE] Flags `-bigtable.grpc-use-gzip-compression`, `-ingester.client.grpc-use-gzip-compression`, `-querier.frontend-client.grpc-use-gzip-compression` are now deprecated. #2940
 * [CHANGE] Limit errors reported by ingester during query-time now return HTTP status code 422. #2941
+* [CHANGE] Experimental Delete Series: `/api/v1/admin/tsdb/delete_series` and `/api/v1/admin/tsdb/cancel_delete_request` purger APIs to return status code `204` instead of `200` for success. #2946
```
pracucci, line 46: We're cutting Cortex `1.3.0` and this PR will be merged after that. Please rebase `master` and move the CHANGELOG entry under the `master / unreleased` section at the top of the CHANGELOG. Thanks!

sandeepsukhani:
@pracucci I ran this locally a couple of times and did not see any failures so not sure how to fix it :disappointed:

CHANGELOG.md
pracucci: I took care of this.

pracucci:
> @pracucci I ran this locally a couple of times and did not see any failures so not sure how to fix it 😞

Ok, no problem. I just rebased `master` and moved the CHANGELOG entry to the top (because we cut `1.3.0` in the meanwhile). If CI passes, we can merge it.

PR #2949, opened 2020-07-29
Install gsutil in the default Cortex image

pracucci:
**What this PR does**:
What's the sentiment to install the `gsutil` CLI in the default Cortex image, to provide cluster operators a ready to use CLI tool to manually upload local data to a GCS bucket? This could be useful while dealing with an outage.

The image size increases from 98MB to 179MB (because of Python).

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/cortex/Dockerfile
+++ b/cmd/cortex/Dockerfile
@@ -5,6 +5,18 @@ COPY       cortex /bin/cortex
 EXPOSE     80
 ENTRYPOINT [ "/bin/cortex" ]
 
+# Install the CLI tool "gsutil" in order to help cluster operators to run manual operations
+# when running the Cortex blocks storage with GCS.
+# IMPORTANT: if you change python version, remember to also change the plugin path at /etc/boto.cfg.
+RUN apk add --no-cache python3~=3.8 py3-pip && \
+    pip install google-compute-engine && \
+    wget https://storage.googleapis.com/pub/gsutil.tar.gz && \
+    tar -zxf gsutil.tar.gz -C /usr/local/bin && \
+    rm -f gsutil.tar.gz && \
+    mv /usr/local/bin/gsutil /usr/local/bin/gsutil-sources && \
+    ln -s /usr/local/bin/gsutil-sources/gsutil /usr/local/bin/gsutil && \
+    ln -s /usr/bin/python3 /usr/bin/python
+
 ARG revision
 LABEL org.opencontainers.image.title="cortex" \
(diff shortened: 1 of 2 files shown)


pstibrany:
I can see the benefit, but that's a too big increase :(

I think we should instead recommend people to use some other existing docker image for inspections like that. (Although having to start another pod for that is not as convenient).

pstibrany:
> > I think we should instead recommend people to use some other existing docker image for inspections like that. (Although having to start another pod for that is not as convenient).
> 
> While dealing with an issue in production, running another pod it's not just inconvenient, but sometimes not an option at all, because you need to copy blocks from a running ingester.

Yes, I've realized that while reading your https://github.com/grafana/cortex-jsonnet/pull/152. It would be great to have gsutil as small Go binary :)

gouthamve:
Should we just bundle the minio client? https://docs.min.io/docs/minio-client-quickstart-guide.html This way we have a single client for all cloud storages. As far as I understand `gsutil` is only GCP specific?

simonswine:
I think minio client can provide a quite generic interface to object stores and it comes at cheaper cost than adding cloud specific ones. 

It is also a golang project, so at some point we could benefit from a deeper integration with it.

pracucci:
> Should we just bundle the minio client? https://docs.min.io/docs/minio-client-quickstart-guide.html This way we have a single client for all cloud storages. As far as I understand `gsutil` is only GCP specific?

I tried it but it's not an option because of the authentication mechanism. I would like to have a seamless authentication when you run Cortex in GCP/GKE (automatically authenticate based on the node's service account without having to provide any credentials) and this doesn't work with the minio-client (while it does with the google SDK).

pracucci:
> Another option would be to add a target module to the `cortex` binary that handles this functionality and uses the built in config file to do so. This would mean we don't have to add anything to the image. Although this is not exactly a swiss army knife approach.

Looks like reinventing the wheel to me. While doing manual operations, having a swiss army knife to operate on the bucket is just very useful. We've seen it during our last outage (which I was dealing with).

simonswine:
If we are talking about a Kubernetes deployment, it would be possible to just run a second container (`sleep infinity`) with the appropriate image providing the object storage client and mount the necessary volumes into that sidecar.  That would avoid bloat of the cortex image and every deployment can just bundle their appropriate image. 

Longer term Kubernetes will eventually support [ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/) which might be the right tool .

pracucci:
_Closing for now_

PR #2954, opened 2020-07-30
Fixed basic lifecycler instance registration when only the address differs

### LINKED ISSUE #2950: Store gateway pods heartbeating stale IP addresses
Yesterday during an upgrade, we cycled out our 3 `store-gateway` pods in kubernetes, after which almost all queries against cortex started failing with a "tcp timeout: no route to host 10.x.x.x". Upon investigation, we discovered the `store-gateway` ring showed two out of three hosts reporting the incorrect IP address, even though they were still heartbeating on a regular basis. Below is a screenshot of the storage ring status, as well as a screenshot of the actual Kubernetes-assigned IP addresses of the `store-gateway` pods. I manually hit the "forget" button on the two stale IP addresses, and the correct IPs were immediately populated at the next heartbeat interval. We're running cortex v1.2.0, and can provide our config on request.

Screen Shot 2020-07-29 at 8 49 45 AM

Screen Shot 2020-07-29 at 8 49 50 AM

pracucci:
**What this PR does**:
I think I've tried to be too smart in the `BasicLifecycler` instance registration to avoid CAS operation when not required. Before this PR, if an instance already exists in the ring, it's not updated at registration phase if only the address differs. This PR fixes it.

I'm not 💯  sure, but could fix #2950.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,7 +101,7 @@
 * [BUGFIX] Query-frontend: Fix passing HTTP `Host` header if `-frontend.downstream-url` is configured. #2880
 * [BUGFIX] Ingester: Improve time-series distribution when `-experimental.distributor.user-subring-size` is enabled. #2887
 * [BUGFIX] Set content type to `application/x-protobuf` for remote_read responses. #2915
-* [BUGFIX] Fixed ruler and store-gateway instance registration in the ring (when sharding is enabled) when an instance replaces a previously registered instance which has been abruptly terminated and the only difference between the two instances is the address. #2954
+* [BUGFIX] Fixed ruler and store-gateway instance registration in the ring (when sharding is enabled) when a new instance replaces abruptly terminated one, and the only difference between the two instances is the address. #2954
 
 ## 1.2.0 / 2020-07-01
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -101,6 +101,7 @@
 * [BUGFIX] Query-frontend: Fix passing HTTP `Host` header if `-frontend.downstream-url` is configured. #2880
 * [BUGFIX] Ingester: Improve time-series distribution when `-experimental.distributor.user-subring-size` is enabled. #2887
 * [BUGFIX] Set content type to `application/x-protobuf` for remote_read responses. #2915
+* [BUGFIX] Fixed ruler and store-gateway instance registration in the ring (when sharding is enabled) when an instance replaces a previously registered instance which has been abruptly terminated and the only difference between the two instances is the address. #2954
```
pstibrany, line 104: I think this is little shorter and easier to read, but native speaker may rephrase it better.

```suggestion
* [BUGFIX] Fixed ruler and store-gateway instance registration in the ring (when sharding is enabled) when a new instance replaces abruptly terminated one, and the only difference between the two instances is the address. #2954
```

PR #2960, opened 2020-07-30
AM Metric: Add tenant label to valid/invalid configs

gotjosh:
Gives us a way to know whenever a tenant has an invalid configuration in
place.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/integration/alertmanager_test.go
+++ b/integration/alertmanager_test.go
@@ -30,7 +30,7 @@ func TestAlertmanager(t *testing.T) {
 		"",
 	)
 	require.NoError(t, s.StartAndWaitReady(alertmanager))
-	require.NoError(t, alertmanager.WaitSumMetrics(e2e.Equals(0), "cortex_alertmanager_invalid_config"))
+	require.NoError(t, alertmanager.WaitSumMetrics(e2e.Equals(0), "cortex_alertmanager_config_invalid"))
 
 	c, err := e2ecortex.NewClient("", "", alertmanager.HTTPEndpoint(), "", "user-1")
 	require.NoError(t, err)
@@ -81,7 +81,7 @@ func TestAlertmanagerStoreAPI(t *testing.T) {
 	require.NoError(t, err)
 
 	time.Sleep(2 * time.Second)
-	require.NoError(t, am.WaitSumMetrics(e2e.Equals(0), "cortex_alertmanager_invalid_config"))
+	require.NoError(t, am.WaitSumMetrics(e2e.Equals(0), "cortex_alertmanager_config_invalid"))
 
 	cfg, err := c.GetAlertmanagerConfig(context.Background())
 	require.NoError(t, err)
--- a/integration/alertmanager_test.go
+++ b/integration/alertmanager_test.go
@@ -5,6 +5,7 @@ package main
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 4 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -407,7 +406,7 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
 		if am.fallbackConfig == "" {
 			return fmt.Errorf("blank Alertmanager configuration for %v", cfg.User)
 		}
-		level.Info(am.logger).Log("msg", "blank Alertmanager configuration; using fallback", "user_id", cfg.User)
+		level.Info(am.logger).Log("msg", "blank Alertmanager configuration; using fallback", "user", cfg.User)
```
gotjosh, line 409: somewhat unrelated, noticed this is the only logline in this file with `user_id` instead of `user`.

jtlisi: 👍

gotjosh:
> I think it would make sense to unregister the metrics for that user? WDYT?

Originally, I thought that it might look strange that all your other metrics are registered except this one. But maybe it makes sense?

I have added the commit if we decide to not do it I can simply remove it.

gotjosh:
> Given the value the per-tenant metric can assume would be only 1 or 0

There's an extra scenario that this serves as a precedent. If you provide an invalid config whilst you have a valid config, Alertmanager will revert back to the last working configuration. 

https://github.com/cortexproject/cortex/blob/6db67a4efbbf62b1133fa037a95382a21f752bbf/pkg/alertmanager/multitenant.go#L416-L423

By having the extra-label adding a way to signal this becomes trivial and uniform from a UX perspective (anything other than `cortex_alertmanager_configs{status="valid"} == 1` is a failure).

With that said, I'm not fuzzed (as the above would just have to be a new metric then?) - Happy to go with what you think would align best with what we do in other parts of Cortex.

gotjosh:
Discussed offline, where Marco suggested that `prometheus_config_last_reload_successful` looks pretty similar to what we want to do and thus should be a separate metric.

gotjosh:
We've settled on `cortex_alertmanager_invalid_config`

integration/alertmanager_test.go
```diff
@@ -79,7 +80,8 @@ func TestAlertmanagerStoreAPI(t *testing.T) {
 	err = c.SetAlertmanagerConfig(context.Background(), cortexAlertmanagerUserConfigYaml, map[string]string{})
 	require.NoError(t, err)
 
-	require.NoError(t, am.WaitSumMetrics(e2e.Equals(1), "cortex_alertmanager_configs"))
+	time.Sleep(2 * time.Second)
```
gotjosh, line 83: I solved the CI failure using a quick hack, but I dislike `Sleep` is tests as it a general source of flaky tests. The problem here is the metric does not get initialise registered (given it is per tenant) until much later. 

A solution I can see would be to create a new method (or argument to `WaitSumMetrics`) that waits for a metric to appear instead of bailing out if we don't see the metric on the initial metrics page load. As I can see this being useful in other areas. 

Please let me know and I'll be happy to make the change.

pracucci: Let's work on it in a separate PR. Could you open an issue, please? I have this pending PR https://github.com/cortexproject/cortex/pull/2522 which could be used as a baseline to add a functional option to enable the logic you suggest.

gotjosh: Thanks marco, filed in https://github.com/cortexproject/cortex/issues/2975

pkg/alertmanager/multitenant.go
```diff
@@ -129,19 +126,17 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet) {
...
 
-	m.totalConfigs = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+	m.invalidConfig = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
-		Name:      "alertmanager_configs",
-		Help:      "How many configs the multitenant alertmanager knows about.",
-	}, []string{"status"})
-	m.totalConfigs.WithLabelValues(configStatusInvalid).Set(0)
-	m.totalConfigs.WithLabelValues(configStatusValid).Set(0)
+		Name:      "alertmanager_invalid_config",
```
gouthamve, line 137: As this is a bool, I think the name `alertmanager_config_invalid` would be better suited?

gotjosh: done!

PR #2962, opened 2020-07-30
Always return a result and result type in QueryRange

gotjosh:
**What this PR does**:

Prometheus always returns a result and result type when executing ranged
queries. This was inconsistent in Cortex where no responses would return
`null` for these two values.

https://github.com/prometheus/prometheus/blob/f482c7bdd734ba012f6cedd68ddc24b1416efb35/web/api/v1/api.go#L461-L465

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/querier/queryrange/query_range_test.go
+++ b/pkg/querier/queryrange/query_range_test.go
@@ -113,11 +113,15 @@ func TestMergeAPIResponses(t *testing.T) {
 		input    []Response
 		expected Response
 	}{
-		// No responses shouldn't panic.
+		// No responses shouldn't panic and return a non-empty result and result type.
 		{
 			input: []Response{},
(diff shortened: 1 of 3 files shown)


pkg/querier/queryrange/query_range_test.go
```diff
@@ -113,11 +113,15 @@ func TestMergeAPIResponses(t *testing.T) {
 		input    []Response
 		expected Response
 	}{
-		// No responses shouldn't panic.
+		// No responses shouldn't panic and return a non-empty result and result type.
```
pstibrany, line 116: This comment is bit misleading ... it returns empty (but not null) result now.

gotjosh: Is this better?

PR #2963, opened 2020-07-30
Ruler TenantManager interface

owen-d:
This PR creates a `TenantManager` interface and a default implementation (functionally equivalent to before) which allows different managers to be substituted (i.e. for Loki's ruler).

It also exposes the `SendAlerts` function for reuse in Loki and fixes some confusing naming issues with structs like `appender` which actually implemented `Appendable` and vice versa.

This also enables removing a few fields from the Ruler struct.

### DIFF
--- a/pkg/ruler/compat.go
+++ b/pkg/ruler/compat.go
@@ -79,8 +79,7 @@ func engineQueryFunc(engine *promql.Engine, q storage.Queryable, delay time.Dura
 	}
 }
 
-// TenantManagerFunc is a function adapter for the TenantManager interface
-type TenantManagerFunc func(
+type ManagerFactory = func(
 	ctx context.Context,
 	userID string,
 	notifier *notifier.Manager,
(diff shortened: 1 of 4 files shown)


pkg/ruler/compat.go
```diff
@@ -74,3 +78,51 @@ func engineQueryFunc(engine *promql.Engine, q storage.Queryable, delay time.Dura
 		return orig(ctx, qs, t.Add(-delay))
 	}
 }
+
+// TenantManagerFunc is a function adapter for the TenantManager interface
+type TenantManagerFunc func(
```
pstibrany, line 83: Let's just use the function directly, without having an interface for it?
I'd also suggest calling it `RulerManagerFactory` to better describe its intent.

owen-d: Fair point. I'll remove it :)

pkg/ruler/compat.go
```diff
@@ -74,3 +78,40 @@ func engineQueryFunc(engine *promql.Engine, q storage.Queryable, delay time.Dura
 		return orig(ctx, qs, t.Add(-delay))
 	}
 }
+
+type ManagerFactory = func(
```
pstibrany, line 82: This is a type alias, instead of new type (not sure it makes any difference here, but if it was a deliberate choice, I'd like to understand why?).
It should also go to ruler.go, where it is used (consumed).

PR #2970, opened 2020-07-31
Be consistent with span logging, set error=true and use otlog.Error

cstyan:
It's very confusing to look at a trace UI for two traces that seemed to have failed for the same reason, yet one has error ! icons and the other doesn't. This seems to be a side effect of not setting `error=true` on all error cases, and one case where we had `error="a string"`.

There's also a change required in upstream weaveworks/common to add another `error=true`

cc @gouthamve @gotjosh @joe-elliott 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -11,13 +11,13 @@ import (
 	"github.com/bradfitz/gomemcache/memcache"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	opentracing "github.com/opentracing/opentracing-go"
 	otlog "github.com/opentracing/opentracing-go/log"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	instr "github.com/weaveworks/common/instrument"
 
 	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 )
 
 type observableVecCollector struct {
@@ -146,19 +146,20 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 
 func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
 	var items map[string]*memcache.Item
-	err := instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(innerCtx context.Context) error {
-		sp := opentracing.SpanFromContext(innerCtx)
-		sp.LogFields(otlog.Int("keys requested", len(keys)))
+	method := "Memcache.GetMulti"
+	err := instr.CollectedRequest(ctx, method, c.requestDuration, memcacheStatusCode, func(innerCtx context.Context) error {
+		log, _ := spanlogger.New(innerCtx, method)
+		log.LogFields(otlog.Int("keys requested", len(keys)))
 
 		var err error
 		items, err = c.memcache.GetMulti(keys)
 
-		sp.LogFields(otlog.Int("keys found", len(items)))
+		log.LogFields(otlog.Int("keys found", len(items)))
 
 		// Memcached returns partial results even on error.
 		if err != nil {
-			sp.LogFields(otlog.Error(err))
-			level.Error(c.logger).Log("msg", "Failed to get keys from memcached", "err", err)
+			log.Error(err)
+			level.Error(log).Log("msg", "Failed to get keys from memcached", "err", err)
 		}
 		return err
 	})
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -158,7 +158,7 @@ func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, b
 
 		// Memcached returns partial results even on error.
 		if err != nil {
-			log.Error(err)
+			log.Error(err) //nolint:errcheck
 			level.Error(log).Log("msg", "Failed to get keys from memcached", "err", err)
 		}
 		return err
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -146,7 +146,7 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 
 func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
 	var items map[string]*memcache.Item
-	method := "Memcache.GetMulti"
+	const method = "Memcache.GetMulti"
 	err := instr.CollectedRequest(ctx, method, c.requestDuration, memcacheStatusCode, func(innerCtx context.Context) error {
 		log, _ := spanlogger.New(innerCtx, method)
 		log.LogFields(otlog.Int("keys requested", len(keys)))
--- a/pkg/chunk/cache/memcached.go
+++ b/pkg/chunk/cache/memcached.go
@@ -158,7 +158,7 @@ func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, b
 
 		// Memcached returns partial results even on error.
 		if err != nil {
-			log.Error(err) //nolint:errcheck
+			log.Error(err)
 			level.Error(log).Log("msg", "Failed to get keys from memcached", "err", err)
 		}
 		return err
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -358,7 +358,7 @@ func (s *storageClientV1) query(ctx context.Context, query chunk.IndexQuery, cal
 		return true
 	})
 	if err != nil {
-		log.Error(err)
+		log.Error(err) //nolint:errcheck
 		return errors.WithStack(err)
 	}
 	return nil
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -324,7 +324,7 @@ func (s *storageClientV1) QueryPages(ctx context.Context, queries []chunk.IndexQ
 func (s *storageClientV1) query(ctx context.Context, query chunk.IndexQuery, callback chunk_util.Callback) error {
 	const null = string('\xff')
 
-	log, _ := spanlogger.New(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
+	log, ctx := spanlogger.New(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
 	defer log.Finish()
 
 	table := s.client.Open(query.TableName)
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -13,13 +13,13 @@ import (
 	"cloud.google.com/go/bigtable"
 	"github.com/go-kit/kit/log"
 	ot "github.com/opentracing/opentracing-go"
-	otlog "github.com/opentracing/opentracing-go/log"
 	"github.com/pkg/errors"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
 	chunk_util "github.com/cortexproject/cortex/pkg/chunk/util"
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/grpcclient"
+	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 )
 
 const (
@@ -324,8 +324,8 @@ func (s *storageClientV1) QueryPages(ctx context.Context, queries []chunk.IndexQ
 func (s *storageClientV1) query(ctx context.Context, query chunk.IndexQuery, callback chunk_util.Callback) error {
 	const null = string('\xff')
 
-	sp, ctx := ot.StartSpanFromContext(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
-	defer sp.Finish()
+	log, _ := spanlogger.New(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
+	defer log.Finish()
 
 	table := s.client.Open(query.TableName)
 
@@ -358,7 +358,7 @@ func (s *storageClientV1) query(ctx context.Context, query chunk.IndexQuery, cal
 		return true
 	})
 	if err != nil {
-		sp.LogFields(otlog.String("error", err.Error()))
+		log.Error(err)
 		return errors.WithStack(err)
 	}
 	return nil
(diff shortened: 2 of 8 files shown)


pstibrany:
Isn’t this better solved by consistently using spanlogger that we use elsewhere?

cstyan:
Good point, looks like spanlogger `Error` via opentracings `ext.Error.Set(s.Span, true)` should solve most of this for us.

pkg/chunk/cache/memcached.go
```diff
@@ -146,19 +146,20 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 
 func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
 	var items map[string]*memcache.Item
-	err := instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(innerCtx context.Context) error {
-		sp := opentracing.SpanFromContext(innerCtx)
-		sp.LogFields(otlog.Int("keys requested", len(keys)))
+	method := "Memcache.GetMulti"
```
pracucci, line 149: Let's make it a `const`

pkg/chunk/cache/memcached.go
```diff
@@ -146,19 +146,20 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
...
 
-		sp.LogFields(otlog.Int("keys found", len(items)))
+		log.LogFields(otlog.Int("keys found", len(items)))
 
 		// Memcached returns partial results even on error.
 		if err != nil {
-			sp.LogFields(otlog.Error(err))
-			level.Error(c.logger).Log("msg", "Failed to get keys from memcached", "err", err)
+			log.Error(err)
+			level.Error(log).Log("msg", "Failed to get keys from memcached", "err", err)
```
pracucci, line 162: Don't we loose the logging to stdout/err switching from `c.logger` to `log` here?

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -320,8 +320,9 @@ func (s *storageClientV1) QueryPages(ctx context.Context, queries []chunk.IndexQ
 func (s *storageClientV1) query(ctx context.Context, query chunk.IndexQuery, callback chunk_util.Callback) error {
 	const null = string('\xff')
 
-	sp, ctx := ot.StartSpanFromContext(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
-	defer sp.Finish()
+	// sp, ctx := ot.StartSpanFromContext(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
```
pracucci, line 323: No commented code, please.

pkg/chunk/cache/memcached.go
```diff
@@ -146,19 +146,20 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 
 func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
 	var items map[string]*memcache.Item
-	err := instr.CollectedRequest(ctx, "Memcache.GetMulti", c.requestDuration, memcacheStatusCode, func(innerCtx context.Context) error {
-		sp := opentracing.SpanFromContext(innerCtx)
-		sp.LogFields(otlog.Int("keys requested", len(keys)))
+	method := "Memcache.GetMulti"
+	err := instr.CollectedRequest(ctx, method, c.requestDuration, memcacheStatusCode, func(innerCtx context.Context) error {
+		log, _ := spanlogger.New(innerCtx, method)
```
pracucci, line 151: `spanlogger.New()` starts a new span. Shouldn't we always replace `innerCtx` with the context returned by the `New()` function?

pkg/querier/queryrange/query_range.go
```diff
@@ -234,17 +235,16 @@ func (prometheusCodec) DecodeResponse(ctx context.Context, r *http.Response, _ R
 		body, _ := ioutil.ReadAll(r.Body)
 		return nil, httpgrpc.Errorf(r.StatusCode, string(body))
 	}
-
-	sp, _ := opentracing.StartSpanFromContext(ctx, "ParseQueryRangeResponse")
-	defer sp.Finish()
+	log, _ := spanlogger.New(ctx, "ParseQueryRangeResponse")
```
pracucci, line 238: New span here. Should replace the context, no?

pkg/querier/queryrange/results_cache.go
```diff
@@ -450,14 +450,14 @@ func (s resultsCache) get(ctx context.Context, key string) ([]Extent, bool) {
...
+	log, ctx := spanlogger.New(ctx, "unmarshal-extent")
+	defer log.Finish()
 
-	sp.LogFields(otlog.Int("bytes", len(bufs[0])))
+	log.LogFields(otlog.Int("bytes", len(bufs[0])))
 
 	if err := proto.Unmarshal(bufs[0], &resp); err != nil {
-		level.Error(s.logger).Log("msg", "error unmarshalling cached value", "err", err)
-		sp.LogFields(otlog.Error(err))
+		level.Error(log).Log("msg", "error unmarshalling cached value", "err", err)
```
pracucci, line 459: _Same comment about the logging here_

pkg/testexporter/correctness/simple.go
```diff
@@ -171,14 +170,14 @@ func verifySamples(log *spanlogger.SpanLogger, tc Case, pairs []model.SamplePair
 		expectedNumSamples := int(duration / cfg.ScrapeInterval)
 		if !epsilonCorrect(float64(len(pairs)), float64(expectedNumSamples), cfg.samplesEpsilon) {
 			level.Error(log).Log("msg", "wrong number of samples", "expected", expectedNumSamples, "actual", len(pairs))
-			log.LogFields(otlog.Error(fmt.Errorf("wrong number of samples")))
+			log.Error(fmt.Errorf("wrong value"))
```
pracucci, line 173: Isn't the previous message "wrong number of samples" more correct? It's also what we log.

pkg/testexporter/correctness/simple.go
```diff
@@ -171,14 +170,14 @@ func verifySamples(log *spanlogger.SpanLogger, tc Case, pairs []model.SamplePair
...
 			return false
 		}
 	} else {
 		expectedNumSamples := int(duration / cfg.ScrapeInterval)
 		if math.Abs(float64(expectedNumSamples-len(pairs))) > 2 {
 			level.Error(log).Log("msg", "wrong number of samples", "expected", expectedNumSamples, "actual", len(pairs))
-			log.LogFields(otlog.Error(fmt.Errorf("wrong number of samples")))
+			log.Error(fmt.Errorf("wrong value"))
```
pracucci, line 180: _Same comment here_

pkg/chunk/cache/memcached.go
cstyan: I don't think so. `spanlogger.New` or `FromContext` both create a `go-kit/log.Logger` from `util.Logger`, which we've initialized in `cmd/cortex/main.go`. If I'm missing something in terms of how this is all tied together let me know.

pkg/chunk/cache/memcached.go
cstyan: here I should have used `spanlogger.FromContext`

pkg/querier/queryrange/query_range.go
cstyan: I'm basing this off how we used the ot library `sp, _ := opentracing.StartSpanFromContext(ctx, "ParseQueryRangeResponse")`, where we ignored the returned context. I don't know the historical reason behind doing it that way, so if you think it should change here with the introduction of spanlogger that's fine by me.

pkg/chunk/cache/memcached.go
cstyan: hmmm, it looks like we never actually use innerCtx in the function other than to create the span? and without using `spanlogger.New` we don't get the `method` field in log lines for free

cstyan:
Thanks for the review @pracucci, I've replied to the comments that are still open, let me know what you think.

cstyan:
@pracucci can you let me know your preferences re: my replies to your comments?

pkg/chunk/cache/memcached.go
pracucci: You're right. Sorry for my previous misleading message. All good here.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -320,8 +320,8 @@ func (s *storageClientV1) QueryPages(ctx context.Context, queries []chunk.IndexQ
 func (s *storageClientV1) query(ctx context.Context, query chunk.IndexQuery, callback chunk_util.Callback) error {
 	const null = string('\xff')
 
-	sp, ctx := ot.StartSpanFromContext(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
-	defer sp.Finish()
+	log, _ := spanlogger.New(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
```
pracucci, line 323: Shouldn't we replace `ctx` with the one returned by `spanlogger.New()` here?

pkg/querier/queryrange/query_range.go
pracucci: I think the reason is that `ctx` is not used below within this function. I believe replacing it is more future proof to avoid mistakes (we should do it once we'll start using `ctx` below and it's easy to forget to do it).

pkg/querier/queryrange/results_cache.go
```diff
@@ -450,14 +450,14 @@ func (s resultsCache) get(ctx context.Context, key string) ([]Extent, bool) {
 	}
 
 	var resp CachedResponse
-	sp, _ := opentracing.StartSpanFromContext(ctx, "unmarshal-extent")
-	defer sp.Finish()
+	log, _ := spanlogger.New(ctx, "unmarshal-extent")
```
pracucci, line 453: Same comment about `ctx` just to make it future proof.

pstibrany:
While doing this change, would it make sense to move error-marking to `Log` method? If passed level is error, we would mark span with error flag, and if there is any `error` object passed, we would wrap first one into `otlog.Error`.

This would also make `(*SpanLogger).Error(error)` method  unnecessary.

(I thought it already works this way when making my earlier suggestion, but that doesn't seem to be the case)

pkg/chunk/cache/memcached.go
```diff
@@ -146,19 +146,20 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
...
 		items, err = c.memcache.GetMulti(keys)
 
-		sp.LogFields(otlog.Int("keys found", len(items)))
+		log.LogFields(otlog.Int("keys found", len(items)))
 
 		// Memcached returns partial results even on error.
 		if err != nil {
-			sp.LogFields(otlog.Error(err))
-			level.Error(c.logger).Log("msg", "Failed to get keys from memcached", "err", err)
+			log.Error(err) //nolint:errcheck
```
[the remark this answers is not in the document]
cstyan: I couldn't get that to work properly, for various variations of `(import path.Struct).Function`. If you have a suggestion for what to use as the entry in that file let me know.

pkg/testexporter/correctness/simple.go
```diff
@@ -171,14 +170,14 @@ func verifySamples(log *spanlogger.SpanLogger, tc Case, pairs []model.SamplePair
 		expectedNumSamples := int(duration / cfg.ScrapeInterval)
 		if !epsilonCorrect(float64(len(pairs)), float64(expectedNumSamples), cfg.samplesEpsilon) {
 			level.Error(log).Log("msg", "wrong number of samples", "expected", expectedNumSamples, "actual", len(pairs))
-			log.LogFields(otlog.Error(fmt.Errorf("wrong number of samples")))
+			log.Error(fmt.Errorf("wrong value")) //nolint:errcheck
```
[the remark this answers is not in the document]
cstyan: No you're right, `wrong number of samples` is more accurate. I was pretty sure I had updated all of these, maybe that change got lost in a rebase/force push. Sorry about that.

cstyan:
Thanks for the review again @pracucci I think everything is updated correctly now. There's just the lint exclusion left, I was not able to get an entry into the file that worked locally, which is why I went with the nolint comments.

@pstibrany are you referring to the spanloggers `Log` function? That's just a wrapper for `go-kit/log`'s `Log`, which has no notion of log level since that's done via the `go-kit/log/levels` wrappers. So there isn't actually any `Log(msg, level)` that could take `error` and log to error level and add the correct error tag to the span. Think we should just stick with compatibility with the logger we use.

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -320,8 +320,8 @@ func (s *storageClientV1) QueryPages(ctx context.Context, queries []chunk.IndexQ
 func (s *storageClientV1) query(ctx context.Context, query chunk.IndexQuery, callback chunk_util.Callback) error {
 	const null = string('\xff')
 
-	sp, ctx := ot.StartSpanFromContext(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
-	defer sp.Finish()
+	log, ctx := spanlogger.New(ctx, "QueryPages", ot.Tag{Key: "tableName", Value: query.TableName}, ot.Tag{Key: "hashValue", Value: query.HashValue})
```
cstyan, line 323: should the method here be `query` instead of `QueryPages`?

pstibrany:
> @pstibrany are you referring to the spanloggers `Log` function? That's just a wrapper for `go-kit/log`'s `Log`, which has no notion of log level since that's done via the `go-kit/log/levels` wrappers. So there isn't actually any `Log(msg, level)` that could take `error` and log to error level and add the correct error tag to the span. Think we should just stick with compatibility with the logger we use.

Yes, I'm referring to `Log` function. Log level is passed as just another key-value pair to that method (key being `level.Key()` and value being `level.ErrorValue()` for errors), and `Log` method is already processing all key-value pairs inside, so it can as well understand this special combination. I think it's worth exploring.

cstyan:
> Yes, I'm referring to `Log` function. Log level is passed as just another key-value pair to that method (key being `level.Key()` and value being `level.ErrorValue()` for errors), and `Log` method is already processing all key-value pairs inside, so it can as well understand this special combination. I think it's worth exploring.

Oh interesting, TIL. Good find, yeah we could then check `kvps` for a `level=error` combination and set the tag on the span. The question would be whether we want to assign all kvps as fields on the span as tags or log messages, or just look for a second field `err` or `error`. 

Are you okay if I just open a new issue to look into this, instead of doing it as part of this PR? Having the error tags in general is going to help with debugging the ruler already, which is what prompted me to make this change.

pkg/chunk/cache/memcached.go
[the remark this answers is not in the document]
cstyan: Please see the latest commit as well, if we want to have the `ctx` assignment in the bigtable index client for example, we need nolint comments again.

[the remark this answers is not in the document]
pstibrany: I think `(*github.com/cortexproject/cortex/pkg/util/spanlogger.SpanLogger).Error` should work (it does with standalone errcheck utility). When I update **`golangci-lint` in the build image to 1.30.0**, I no longer see this warning, even without changing .errcheck-exclude. 😕 

We could also remove error return value from this method, since no production code uses it.

pracucci:
@cstyan I've pushed a [commit](https://github.com/cortexproject/cortex/pull/2970/commits/f2d02ea2be272e97873b4b351cd675ca91896768) to try to globally exclude `SpanLogger.Error()` from errcheck (it works on my local machine, let's see what CI says...).

PR #2973, opened 2020-08-04
Move regex opt to after lookup

### LINKED ISSUE #2906: Regex optimisation on index does redundant lookups
When caching index entries, we basically lookup the entire row (to cache it) and then filter the entries that match after loading the full row: https://github.com/cortexproject/cortex/blob/071502d118a1d0687ef21f832b196341e20bdad1/pkg/chunk/util/util.go#L84-L116 and https://github.com/cortexproject/cortex/blob/071502d118a1d0687ef21f832b196341e20bdad1/pkg/chunk/storage/caching_index_client.go#L70-L71

Now we introduced an optimisation that does the following: https://github.com/cortexproject/cortex/blob/071502d118a1d0687ef21f832b196341e20bdad1/pkg/chunk/chunk_store.go#L463-L470

Basically if we have a lookup that looks like `label=~"a|b|c|d"` we are now splitting it into 4 different lookups of `label=a`, `label=b`, etc. This is causing us to load the entire row multiple times due to caching. We should make sure to push this optmisation down to `filteringBatchIter` and not load something multiple times.

gouthamve:
fixes #2906

When we use the caching client (which is what is used in most cases), we
load the entire row (tableName+HashKey) irrespective of what the
rangeKey parameters are. Which means with the optimisation, we are
loading the same row multiple times and then operating on the same data.
This PR moves the optimisation to after the data is loaded which should
be faster.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -561,6 +561,11 @@ func (c *baseStore) parseIndexEntries(_ context.Context, entries []IndexEntry, m
 			if _, ok := matchSet[string(labelValue)]; !ok {
 				continue
 			}
+
+			// If its in the set, then add it to set, we don't need to run
+			// matcher on it again.
+			result = append(result, chunkKey)
+			continue
 		}
 
 		if matcher != nil && !matcher.Matches(string(labelValue)) {
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -460,16 +460,6 @@ func (c *baseStore) lookupIdsByMetricNameMatcher(ctx context.Context, from, thro
 	} else if matcher.Type == labels.MatchEqual {
 		labelName = matcher.Name
 		queries, err = c.schema.GetReadQueriesForMetricLabelValue(from, through, userID, metricName, matcher.Name, matcher.Value)
-	} else if matcher.Type == labels.MatchRegexp && len(FindSetMatches(matcher.Value)) > 0 {
-		set := FindSetMatches(matcher.Value)
-		for _, v := range set {
-			var qs []IndexQuery
-			qs, err = c.schema.GetReadQueriesForMetricLabelValue(from, through, userID, metricName, matcher.Name, v)
-			if err != nil {
-				break
-			}
-			queries = append(queries, qs...)
-		}
 	} else {
 		labelName = matcher.Name
 		queries, err = c.schema.GetReadQueriesForMetricLabel(from, through, userID, metricName, matcher.Name)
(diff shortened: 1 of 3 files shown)


gouthamve:
```
➜  cortex git:(fix-2906) ✗ benchcmp old.txt new.txt 
benchmark                                     old ns/op     new ns/op     delta
BenchmarkParseIndexEntries500-8               342057        291739        -14.71%
BenchmarkParseIndexEntries2500-8              1718683       1516488       -11.76%
BenchmarkParseIndexEntries10000-8             6568014       6463280       -1.59%
BenchmarkParseIndexEntries50000-8             34212971      34284051      +0.21%
BenchmarkParseIndexEntriesRegexSet500-8       137045        89742         -34.52%
BenchmarkParseIndexEntriesRegexSet2500-8      647045        441737        -31.73%
BenchmarkParseIndexEntriesRegexSet10000-8     2772264       1881870       -32.12%
BenchmarkParseIndexEntriesRegexSet50000-8     14044719      10336336      -26.40%

benchmark                                     old allocs     new allocs     delta
BenchmarkParseIndexEntries500-8               1503           1504           +0.07%
BenchmarkParseIndexEntries2500-8              7503           7504           +0.01%
BenchmarkParseIndexEntries10000-8             30005          30007          +0.01%
BenchmarkParseIndexEntries50000-8             150007         150008         +0.00%
BenchmarkParseIndexEntriesRegexSet500-8       1503           1522           +1.26%
BenchmarkParseIndexEntriesRegexSet2500-8      7503           7522           +0.25%
BenchmarkParseIndexEntriesRegexSet10000-8     30005          30023          +0.06%
BenchmarkParseIndexEntriesRegexSet50000-8     150007         150024         +0.01%

benchmark                                     old bytes     new bytes     delta
BenchmarkParseIndexEntries500-8               96289         96320         +0.03%
BenchmarkParseIndexEntries2500-8              481975        482008        +0.01%
BenchmarkParseIndexEntries10000-8             1928507       1928559       +0.00%
BenchmarkParseIndexEntries50000-8             9606750       9606808       +0.00%
BenchmarkParseIndexEntriesRegexSet500-8       88222         88693         +0.53%
BenchmarkParseIndexEntriesRegexSet2500-8      441163        441627        +0.11%
BenchmarkParseIndexEntriesRegexSet10000-8     1764739       1765087       +0.02%
BenchmarkParseIndexEntriesRegexSet50000-8     8804033       8804335       +0.00%
```

Hrm, not such a big change, interesting but still worth having imo.

pkg/chunk/chunk_store.go
```diff
@@ -549,13 +539,29 @@ func (c *baseStore) parseIndexEntries(_ context.Context, entries []IndexEntry, m
...
 		}
 
+		// If the matcher is like a set (=~"a|b|c|d|...") and
+		// the label value is not in that set move on.
+		if len(matchSet) > 0 {
+			if _, ok := matchSet[string(labelValue)]; !ok {
+				continue
+			}
+		}
+
```
pstibrany, line 564: Several comments here:
1) Nit: if condition can be merged
2) No unit test covers `continue` branch. I think we should cover it by tests.
3) If `labelValue` is one of matched values, we still run regex on it on the next line. We can remove `matcher.Matches` in this case, and save some cpu cycles.

pstibrany:
This looks useful, and would be great to merge. @gouthamve would you find some time to look at the comments and get the PR into mergeable state?

pstibrany (approved):
LGTM, thanks!

PR #2974, opened 2020-08-04
Always query series and labels from TSDB head

pracucci:
**What this PR does**:
Yesterday we had an outage in our staging environment running the blocks storage, caused by all ingesters going OOM while concurrently running several heavy `/api/v1/series` queries.

The problem is that we run ingesters with a TSDB retention of 4 days and we ignore the query time range when handling "metadata" API requests (series and labels). In this PR I'm proposing to mitigate the issue forcing the time range to the TSDB head only, regardless of the TSDB retention (which could be several days). The new logic in this PR is conceptually similar to the Cortex chunks storage, where we do query only in-memory series.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -689,14 +689,20 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx context.Context, req *client.Me
...
+			return nil, ctx.Err()
+		}
+
 		seriesSet := q.Select(false, nil, matchers...)
 		if seriesSet.Err() != nil {
 			return nil, seriesSet.Err()
 		}
 
 		for seriesSet.Next() {
-			if seriesSet.Err() != nil {
```
pracucci, line 698: This check was useless, because `seriesSet.Next()` returns false on error.

PR #2976, opened 2020-08-04
Optimised /api/v1/series for blocks storage

pracucci:
**What this PR does**:
While working on #2794, I've compared our implementation to fetch series with Prometheus one and I've noticed Prometheus is easier. I've done the refactoring and written and benchmark and turned out it performs better too.

```
benchmark                                           old ns/op     new ns/op     delta
Benchmark_Ingester_v2MetricsForLabelMatchers-12     250057594     226011946     -9.62%

benchmark                                           old allocs     new allocs     delta
Benchmark_Ingester_v2MetricsForLabelMatchers-12     1104071        800196         -27.52%

benchmark                                           old bytes     new bytes     delta
Benchmark_Ingester_v2MetricsForLabelMatchers-12     62724828      51382654      -18.08%
```

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -674,7 +674,7 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx context.Context, req *client.Me
 			return nil, ctx.Err()
 		}
 
-		seriesSet := q.Select(false, nil, matchers...)
+		seriesSet := q.Select(true, nil, matchers...)
 		sets = append(sets, seriesSet)
 	}
 
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -665,11 +665,8 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx context.Context, req *client.Me
 	}
 	defer q.Close()
 
-	// Run a query for each matchers set and collect all the results
-	added := map[string]struct{}{}
-	result := &client.MetricsForLabelMatchersResponse{
-		Metric: make([]*client.Metric, 0),
-	}
+	// Run a query for each matchers set and collect all the results.
+	var sets []storage.SeriesSet
 
 	for _, matchers := range matchersSet {
 		// Interrupt if the context has been canceled.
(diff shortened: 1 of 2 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -695,37 +692,24 @@ func (i *Ingester) v2MetricsForLabelMatchers(ctx context.Context, req *client.Me
 		}
 
 		seriesSet := q.Select(false, nil, matchers...)
```
pstibrany, line 694: Sets with series are not sorted, so merging them later doesn't work properly. This can be shown by using modified request in the benchmark/test:

```
	req := &client.MetricsForLabelMatchersRequest{
		StartTimestampMs: now,
		EndTimestampMs:   now,
		MatchersSet: []*client.LabelMatchers{
			{Matchers: []*client.LabelMatcher{
				{Type: client.REGEX_MATCH, Name: model.MetricNameLabel, Value: "test.*"},
			}},
			{Matchers: []*client.LabelMatcher{
				{Type: client.REGEX_MATCH, Name: model.MetricNameLabel, Value: "test.*0"}, // ending with 0
			}},
		},
	}
```

This should still return all `numSeries` series, but returns more of them.

Can you rerun benchmarks with sorting, and see whether it is still an improvement or not?

pracucci: Wow, thanks for spotting it! I copied the logic from Prometheus:
https://github.com/prometheus/prometheus/blob/b521612042ec87103088b45b7b5dfee3bb8dc732/web/api/v1/api.go#L604-L608

Either Prometheus /series API endpoint doesn't guarantee any deduplication, or there's a bug in Prometheus too. I will open an issue there.

Back to us, I modified the logic to use `.Select(true, ...)` (sorted) for now and re-run the benchmark (updated PR description). There's still some benefit, less then before.

pracucci: Opened this issue in Prometheus: https://github.com/prometheus/prometheus/issues/7801. Will follow up this PR accordingly, if any change is required.

PR #2985, opened 2020-08-06
Add client ip address to duplicate message

MichelHollands:
**What this PR does**:

This depends on https://github.com/weaveworks/common/pull/193.

The `server.log_source_ips_enabled` setting in the config file has to be set to true to enable this.

Extract the IP address from the X-Forwarder-For value in the context. Add it to the duplicate samples error message. The error message now looks like this:

```
level=error ts=2020-08-11T12:40:28.205776Z caller=push.go:45 org_id=fake traceID=18c0d1b736132d93 sourceIPs="1.2.3.4, ::1" msg="push error" err="rpc error: code = Code(400) desc = user=fake: sample with repeated timestamp but different value; last value: 0.013667131, incoming value: 0.013651803 for series {__name__=\"scrape_duration_seconds\", instance=\"localhost:9090\", job=\"prometheus\"}"
```

Prometheus remote writes to a reverse proxy which forwards the remote write to Cortex. Prometheus, the reverse proxy and Cortex are running locally on a Mac hence the ::1 IP v6 address. The reverse proxy adds the X-Forwarded-For header of 1.2.3.4.

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/util/extract_forwarded.go
```diff
@@ -0,0 +1,91 @@
...
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+
+	"google.golang.org/grpc/metadata"
+)
+
+// ipAddressesKey is key for the GRPC metadata where the IP addresses are stored
+const ipAddressesKey = "ipaddresseskey"
```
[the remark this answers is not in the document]
MichelHollands: This had been changed to x-forwarded-for.

pkg/util/extract_forwarded.go
```diff
@@ -0,0 +1,91 @@
...
+		// Invalid IP address, just return it so it shows up in the logs
+		return address
+	}
+	return hostStr
+}
+
+// GetSource extracts the X-FORWARDED-FOR header from the given HTTP request
+// and returns a string with it and the remote address
+func GetSource(req *http.Request) string {
+	fwd := req.Header.Get("X-FORWARDED-FOR")
```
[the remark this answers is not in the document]
MichelHollands: This code has move to weaveworks/common. I'll add it there.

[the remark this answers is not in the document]
MichelHollands: I'm not too keen on restricting the X-Forwarded-For headers just yet. Customers haven't used this feature yet so there will be mistakes. By removing wrong addresses it will make it harder to see what was used.

pstibrany:
Note that PR https://github.com/cortexproject/cortex/pull/3064 will update weaveworks/common, which now also includes https://github.com/weaveworks/common/pull/193.

MichelHollands:
> Note that PR #3064 will update weaveworks/common, which now also includes [weaveworks/common#193](https://github.com/weaveworks/common/pull/193).

Thanks. I'll wait till your PR is in before making this ready for review.

pkg/util/push/push.go
```diff
@@ -6,19 +6,28 @@ import (
...
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
 // Handler is a http.Handler which accepts WriteRequests.
-func Handler(cfg distributor.Config, push func(context.Context, *client.WriteRequest) (*client.WriteResponse, error)) http.Handler {
+func Handler(cfg distributor.Config, serverCfg server.Config, push func(context.Context, *client.WriteRequest) (*client.WriteResponse, error)) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := util.WithContext(ctx, util.Logger)
+		sourceIPs, err := middleware.NewSourceIPs(serverCfg.LogSourceIPsHeader, serverCfg.LogSourceIPsRegex)
```
jtlisi, line 22: **thought**: Can we create this before we return the function, and reuse it on every call instead or creating a new one every time?

pkg/api/api.go
```diff
@@ -160,29 +160,29 @@ func (a *API) RegisterAPI(cfg interface{}) {
 }
 
 // RegisterDistributor registers the endpoints associated with the distributor.
-func (a *API) RegisterDistributor(d *distributor.Distributor, pushConfig distributor.Config) {
-	a.RegisterRoute("/api/v1/push", push.Handler(pushConfig, d.Push), true)
+func (a *API) RegisterDistributor(d *distributor.Distributor, pushConfig distributor.Config, serverConfig server.Config) {
```
jtlisi, line 163: **nitpick**: Forcing the serverconfig into each of these initialization functions seems like overkill for what we are trying to do.

**suggestion**: Instead of passing the serverConfig as a param to each of these functions. Would it be possible to create a single `middleware.SourceIPs` in the `initAPI` function and just have it as a set field in the API struct that is reused and passed directly to `push.Handler`?

pkg/util/push/push.go
MichelHollands: Yes, a pointer to a SourceIPExtractor is passed in as a parameter now.

pkg/api/api.go
MichelHollands: The SourceIpExtractor is now created (if needed) in the API struct and used in push.Handler. This change also exposed a bug. The SourceIpExtractor was created always even if the relevant config setting was false. This has been fixed now.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [CHANGE] Histogram `cortex_memcache_request_duration_seconds` `method` label value changes from `Memcached.Get` to `Memcached.GetBatched` for batched lookups, and is not reported for non-batched lookups (label value `Memcached.GetMulti` remains, and had exactly the same value as `Get` in nonbatched lookups).  The same change applies to tracing spans. #3046
 * [CHANGE] TLS server validation is now enabled by default, a new parameter `tls_insecure_skip_verify` can be set to true to skip validation optionally. #3030
 * [CHANGE] `cortex_ruler_config_update_failures_total` has been removed in favor of `cortex_ruler_config_last_reload_successful`. #3056
+* [FEATURE] Update in dependency `weaveworks/common`. Logging of IP of reverse proxy is now supported. #2985
```
pracucci, line 16: My suggestion when writing the CHANGELOG is to put yourself in the final user's shoes. The final user has no idea what `weaveworks/common` is but would love to know the new config options to control the logging of IP. So my suggestion is:
- Remove mentioning to the updated dependency
- Mention the new CLI flags introduced and what they do

docs/configuration/arguments.md
```diff
@@ -487,3 +487,20 @@ The DNS service discovery, inspired from Thanos DNS SD, supports different disco
   The domain name after the prefix is looked up as a SRV query, and then each SRV record is resolved as an A/AAAA record. For example: `dnssrv+memcached.namespace.svc.cluster.local`
 - **`dnssrvnoa+`**<br />
   The domain name after the prefix is looked up as a SRV query, with no A/AAAA lookup made after that. For example: `dnssrvnoa+memcached.namespace.svc.cluster.local`
+
+## Logging of IP of reverse proxy
+
+If a reverse proxy is used in front of Cortex it might be diffult to troubleshoot errors. The following 3 settings can be used to log the IP address passed along by the reverse proxy in headers like X-Forwarded-For.
+
+- `server.log_source_ips_enabled`
```
pracucci, line 495: When mentioning CLI flags, please always prefix with `-`. We haven't been consistent in the past, but we should improve it, to make it more clear they're CLI flags.

pkg/util/push/push_test.go
```diff
@@ -20,15 +21,17 @@ import (
 func TestHandler_remoteWrite(t *testing.T) {
 	req := createRequest(t, createPrometheusRemoteWriteProtobuf(t))
 	resp := httptest.NewRecorder()
-	handler := Handler(distributor.Config{MaxRecvMsgSize: 100000}, verifyWriteRequestHandler(t, client.API))
+	var sourceIPs *middleware.SourceIPExtractor
+	handler := Handler(distributor.Config{MaxRecvMsgSize: 100000}, sourceIPs, verifyWriteRequestHandler(t, client.API))
```
pracucci, line 25: Can be simplified, if your intention is to not pass the IP extractor.

```suggestion
	handler := Handler(distributor.Config{MaxRecvMsgSize: 100000}, nil, verifyWriteRequestHandler(t, client.API))
```

pkg/api/api_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/weaveworks/common/server"
+)
+
+type FakeLogger struct{}
+
+func (fl *FakeLogger) Log(keyvals ...interface{}) error {
+	return nil
+}
+
+func TestNewApiWithoutSourceIPExtractor(t *testing.T) {
```
pracucci, line 17: This test is a bit weak. It doesn't really assert on the functionality, but just that the `New()` doesn't return error.

_Same comment applies to `TestNewApiWithSourceIPExtractor`._

pkg/util/extract_forwarded.go
```diff
@@ -0,0 +1,53 @@
+package util
+
+import (
+	"context"
+
+	"google.golang.org/grpc/metadata"
+)
+
+// ipAddressesKey is key for the GRPC metadata where the IP addresses are stored
+const ipAddressesKey = "x-forwarded-for"
```
pracucci, line 10: You should define a custom `type` (for the context key) to avoid collisions if the same key is defined in other packages (see https://golang.org/pkg/context/ doc).

pkg/util/extract_forwarded.go
```diff
@@ -0,0 +1,53 @@
...
+
+// GetSourceFromOutgoingCtx extracts the source field from the GRPC context
```
pracucci, line 12: This is a generic `util` package. The function name `GetSourceFromOutgoingCtx()` looks a big vague to me. Maybe `GetSourceIPsFromOutgoingCtx` could me more clear? If so, other functions in this file should be renamed accordingly.

pkg/util/extract_forwarded_test.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Test extracting from incoming context
+			ctx := context.Background()
+			if tt.value != "" {
+				md := metadata.Pairs(tt.key, tt.value)
+				ctx = metadata.NewIncomingContext(ctx, md)
+			}
+			if got := GetSourceFromIncomingCtx(ctx); got != tt.want {
+				t.Errorf("GetSourceFromOutgoingCtx() = %v, want %v", got, tt.want)
```
pracucci, line 45: We use the `testify` framework in Cortex. You can simplify the assertions in this file using `assert.Equal(t, want, got)`.

pkg/util/extract_forwarded_test.go
```diff
@@ -0,0 +1,59 @@
...
+			name:  "Value in key",
+			key:   ipAddressesKey,
+			value: "172.16.1.1",
+			want:  "172.16.1.1",
+		},
+		{
+			name:  "Stored under wrong key",
+			key:   "wrongkey",
+			value: "172.16.1.1",
+			want:  "",
```
pracucci, line 33: If you define your custom data `type`, this test can be removed. You shouldn't really worry about the "wrong key".

pkg/util/push/push.go
```diff
@@ -6,19 +6,28 @@ import (
...
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
 // Handler is a http.Handler which accepts WriteRequests.
-func Handler(cfg distributor.Config, push func(context.Context, *client.WriteRequest) (*client.WriteResponse, error)) http.Handler {
+func Handler(cfg distributor.Config, sourceIPs *middleware.SourceIPExtractor, push func(context.Context, *client.WriteRequest) (*client.WriteResponse, error)) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := util.WithContext(ctx, util.Logger)
+		if sourceIPs != nil {
```
pracucci, line 21: Thinking loudly. Not a big fan of having it here. The context injection could be probably easily done in a generic middleware with which you wrap `push.Handler()`, but I don't have a good solution for the `logger`.

MichelHollands: Yes, the logger stumped me as well there. I'll have a think about that.

pkg/util/extract_forwarded_test.go
MichelHollands: As mentioned before GRPC metadata doesn't support custom types.

CHANGELOG.md
MichelHollands: Good idea. Updated.

docs/configuration/arguments.md
MichelHollands: A - prefix was added.

pkg/util/push/push_test.go
MichelHollands: Good catch. The variable declaration has been removed.

pkg/api/api_test.go
MichelHollands: Yeah, I was hoping to piggy back on some existing API test but there was nothing there. At least this shows the code doesn't panic. I haven't looked at the integration tests yet in case . Will do that next.

pkg/util/extract_forwarded.go
MichelHollands: This uses the GRPC metadata which uses strings as keys. So adding an extra type won't add much as it has to be converted back to a string when adding to the metadata.MD type.
I've made ipAddressKey a lot longer by adding the package name so it's less likely to clash.

pkg/util/extract_forwarded.go
MichelHollands: That naming is more definitely more consistent. It's changed.

pkg/util/extract_forwarded_test.go
MichelHollands: Changed.

pkg/util/extract_forwarded.go
pracucci: I think I was drunk when I left this comment. You're right, sorry.

pkg/util/push/push.go
pracucci: We can keep it as is for now.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [CHANGE] Histogram `cortex_memcache_request_duration_seconds` `method` label value changes from `Memcached.Get` to `Memcached.GetBatched` for batched lookups, and is not reported for non-batched lookups (label value `Memcached.GetMulti` remains, and had exactly the same value as `Get` in nonbatched lookups).  The same change applies to tracing spans. #3046
 * [CHANGE] TLS server validation is now enabled by default, a new parameter `tls_insecure_skip_verify` can be set to true to skip validation optionally. #3030
 * [CHANGE] `cortex_ruler_config_update_failures_total` has been removed in favor of `cortex_ruler_config_last_reload_successful`. #3056
+* [FEATURE] Logging of the source IP passed along by a reverse proxy is now supported by setting the `server.log-source-ips-enabled`. For non standard headers the settings `server.log-source-ips-header` and `server.log-source-ips-regex` can be used. #2985
```
pracucci, line 16:
```suggestion
* [FEATURE] Logging of the source IP passed along by a reverse proxy is now supported by setting the `-server.log-source-ips-enabled`. For non standard headers the settings `-server.log-source-ips-header` and `-server.log-source-ips-regex` can be used. #2985
```

pracucci (approved):
Thanks for patiently addressing and commenting on my feedback. Much appreciated! LGTM ❤️

PR #2989, opened 2020-08-06
Fix -ruler.alertmanager-url flag

gotjosh:
**What this PR does**:

- Fixes the flag `-ruler.alertmanager-url` by reverting it back to using a string (which is meant to be comma-separated) as opposed to a list of strings. As it was this broke compatibility with a file-based configuration where it required a list of strings instead of a single string.
- Unifies the usage of `StringsSlice` and `Strings` flag extensions.
- Adds an integration test for testing Alertmanager discovery.

Quoting the commits:

```
@gotjosh
fix: `-ruler.alertmanager-url` should be a string …
6b5a862
This was never space-separated, you needed to register the flag multiple times for it to append to the list of strings.

On top of that, the previous change implied that when using a config file you would need to provide a list instead of a string thus breaking existing configuration.
```
```
Unify the use of StringSlice flags …
1b81fa6
We had two very similar flag extensions: `Strings` and `StringSlice`.

With this, we unify its use by removing `Strings` in favour of `StringSlice` but keeping the string representation favoured in `Strings`.

The difference here is that by using `Sprintf` we would get the brackets representation as part of the string.
```

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/util/flagext/stringslice.go
```diff
@@ -1,15 +1,13 @@
...
-)
+import "fmt"
 
 // StringSlice is a slice of strings that implements flag.Value
 type StringSlice []string
 
 // String implements flag.Value
 func (v StringSlice) String() string {
-	return strings.Join(v, " ")
+	return fmt.Sprintf("%s", []string(v))
```
gotjosh, line 10: The unification of the flag extensions is basically that I kept the name `StringSlice` but "borrowed" the `String` casting function.

I like this because for empty values it renders `[]` instead of `""` which seems to be more accurate.

gotjosh:
I'm not sure what to include in the changelog - opinions are welcome.

gouthamve:
> I'm not sure what to include in the changelog

Nothing. I don't see any user-impact since the last release that is introduced by this PR.

PR #2990, opened 2020-08-06
Fixed query start/end timestamp rounding

### LINKED ISSUE #2932: All result timestamps off by 1ms (without alignment turned on)
If you run a query like this against the latest Cortex release, started with the single-process config at https://raw.githubusercontent.com/cortexproject/cortex/master/docs/configuration/single-process-config.yaml:

http://localhost:9009/api/prom/api/v1/query_range?query=time()&start=1595846748.806&end=1595846848.806&step=5

...then the first result sample should be at timestamp `1595846748.806`, but Cortex reports it at `1595846748.805`, and any subsequent samples are off by 1ms as well. Compare to the result by Prometheus at https://demo.promlabs.com/api/v1/query_range?query=time()&start=1595846748.806&end=1595846848.806&step=5.

I didn't turn on the step alignment middleware, so it shouldn't be related to that.

pracucci:
**What this PR does**:
The good @juliusv reported we're not rounding query start/end timestamps like Prometheus. This PR fixes it, specifically introduces the change done in Prometheus almost 3y ago https://github.com/prometheus/prometheus/pull/4941.

Fixes #2932

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/time_test.go
+++ b/pkg/util/time_test.go
@@ -38,3 +38,52 @@ func TestDurationWithJitter(t *testing.T) {
 func TestDurationWithJitter_ZeroInputDuration(t *testing.T) {
 	assert.Equal(t, time.Duration(0), DurationWithJitter(time.Duration(0), 0.5))
 }
+
+func TestParseTime(t *testing.T) {
+	ts, err := time.Parse(time.RFC3339Nano, "2015-06-03T13:21:58.555Z")
+	require.NoError(t, err)
+
+	var tests = []struct {
+		input  string
+		fail   bool
+		result time.Time
+	}{
+		{
+			input: "",
+			fail:  true,
+		}, {
+			input: "abc",
+			fail:  true,
+		}, {
+			input: "30s",
+			fail:  true,
+		}, {
+			input:  "123",
+			result: time.Unix(123, 0),
+		}, {
+			input:  "123.123",
+			result: time.Unix(123, 123000000),
+		}, {
+			input:  "2015-06-03T13:21:58.555Z",
+			result: ts,
+		}, {
+			input:  "2015-06-03T14:21:58.555+01:00",
+			result: ts,
+		}, {
+			// Test float rounding.
+			input:  "1543578564.705",
+			result: time.Unix(1543578564, 705*1e6),
+		},
+	}
+
+	for _, test := range tests {
+		ts, err := ParseTime(test.input)
+		if test.fail {
+			require.Error(t, err)
+			continue
+		}
+
+		require.NoError(t, err)
+		assert.Equal(t, TimeToMillis(test.result), ts)
+	}
+}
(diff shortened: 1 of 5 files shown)


juliusv:
:+1:

pkg/util/time_test.go
```diff
@@ -38,3 +38,52 @@ func TestDurationWithJitter(t *testing.T) {
...
+			fail:  true,
+		}, {
+			input:  "123",
+			result: time.Unix(123, 0),
+		}, {
+			input:  "123.123",
+			result: time.Unix(123, 123000000),
+		}, {
+			input:  "2015-06-03T13:21:58.555Z",
+			result: ts,
```
pstibrany, line 68:
```suggestion
			result: time.Unix(143333771, 855 * time.Millisecond.Nanoseconds()),
```

Same below. Otherwise it's just testing two calls to time.Parse against each other.

pkg/util/time_test.go
```diff
@@ -38,3 +38,52 @@ func TestDurationWithJitter(t *testing.T) {
...
+		}, {
+			input:  "2015-06-03T14:21:58.555+01:00",
+			result: ts,
+		}, {
+			// Test float rounding.
+			input:  "1543578564.705",
+			result: time.Unix(1543578564, 705*1e6),
+		},
+	}
```
pstibrany, line 77: What about `input: "2015-06-03T13:21:58.56789Z" ... should that be rounded on milliseconds precision?

pkg/util/time_test.go
pracucci: I copied the exact implementation and tests of Prometheus. Should we keep it 1-1 with Prometheus or change it? I'm open to discuss it.

pkg/util/time_test.go
pracucci: I copied the exact implementation and tests of Prometheus. Should we keep it 1-1 with Prometheus or change it? I'm open to discuss it.

pkg/util/time_test.go
pstibrany: I'd weakly prefer test to be clearer and using suggested version. I don't see small deviation from Prometheus to be an issue here.

pkg/util/time_test.go
pstibrany: No strong opinion on milliseconds rounding... We can keep it as-is to be consistent with Prometheus, and think about it when (if) someone hits that. Then we can clarify at both places (depending on what Prometheus decides)

pkg/util/time_test.go
pstibrany: Suggested version is incorrect. Correct ts is 1433337718 and 555 milliseconds of course. 🤦

pracucci: Done.

pkg/util/time_test.go
pracucci: I've added a test. I believe we should just use the Prometheus implementation as reference implementation (we want to be 100% compatible with Prometheus) and thus I've written the expected value with the current Prometheus rounding.

PR #2994, opened 2020-08-07
Query-tee: added a small tolerance to floating point sample values comparison

pracucci:
**What this PR does**:
While comparing query results between the Cortex chunks and blocks storage, we learned that floating point rounding could be different for the same exact query executed twice due to the non deterministic series ordering in PromQL engine (for more details please refer to https://github.com/prometheus/prometheus/pull/7757). 

This PR adds a small (configureable) tolerance to avoid such false positives.

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added


tools/querytee/response_comparator.go
```diff
@@ -186,16 +191,24 @@ func compareScalar(expectedRaw, actualRaw json.RawMessage) error {
...
+	if !compareSampleValue(expected.Value, actual.Value, tolerance) {
 		return fmt.Errorf("expected value %s for timestamp %v but got %s", expected.Value, expected.Timestamp, actual.Value)
 	}
 
 	return nil
 }
+
+func compareSampleValue(first, second model.SampleValue, tolerance float64) bool {
+	if tolerance <= 0 {
+		return first == second
```
brian-brazil, line 210: It's not safe to do equality comparisons on floats due to NaNs, use math.Float64bits

tools/querytee/response_comparator.go
```diff
@@ -186,16 +191,24 @@ func compareScalar(expectedRaw, actualRaw json.RawMessage) error {
...
+	} else {
+		return math.Abs(float64(first)-float64(second)) <= tolerance
```
brian-brazil, line 212: Tolerance is generally a ratio or number of decimal places, not a absolute difference.

tools/querytee/response_comparator.go
pracucci: Thanks for spotting it! The NaN comparison is also wrong when the tolerance is enabled (> 0).

tools/querytee/response_comparator.go
beorn7: Yeah. You might want to look at https://github.com/beorn7/floats/blob/master/equal.go (shameless plug).

tools/querytee/response_comparator.go
bwplotka: :+1: 

We sometimes do this: https://github.com/prometheus/prometheus/blob/e6d7cc5fa43716281bdf5d1cf40415f99a0e9623/storage/series.go#L270 to avoid issues (so use Float compare indeed)

Issue #1828, opened 2019-11-15
Concurrently preload TSDBs right after transfer


pracucci:
We should concurrently preload TSDBs in `TransferTSDB()`.

pstibrany:
Transfers for TSDB has been removed.

PR #2996, opened 2020-08-07
Removed ingesters blocks transfer support

### LINKED ISSUE #2966: Proposal: Remove support for TSDB blocks transfer between ingesters
When ingesters start and stop, they support transfer of data from leaving ingester to joining. This mechanism allows for fast rollouts of new ingesters, because otherwise ingesters would need to flush data from memory to storage, which can take a long time (up to 1h). Transfers were the primary solution to deal with slow rollouts, until WAL support for chunks was introduced in ingesters, which made transfers obsolete. Ingesters with WAL don't need to flush data to long-term storage, as they will reload same data into memory from WAL after restart.

When TSDB blocks support was introduced to ingesters, they also supported transfers.

In this issue, we suggest to **remove support for transfers of TSDB blocks between ingesters**. This doesn't affect transfer of chunks.

Reasons for this proposal:

- Transfers are rarely used with blocks storage: Blocks are always meant to be used with persistent disks to store the blocks, and since each TSDB also has WAL, ingester restart doesn't require flushing of blocks to storage. Restarts are already quite fast.
- Transfers complicate ingesters logic when using blocks. Removing transfers removes extra complexity, both in the code and operations.
- Dropping transfer support of blocks would simplify design of solution for issue https://github.com/cortexproject/cortex/issues/2366.
- There is an integration test for blocks transfers, but apart from that, it's not clear if anyone uses the feature.
- Blocks are still experimental feature, deprecating blocks transfer now doesn't break feature-guarantees. Once blocks are marked as production ready, it will be more difficult to remove this support.

pracucci:
**What this PR does**:
As discussed in #2966, this PR removes the blocks transfer support from ingesters.

/cc @codesome 

Fixes #2966 and #1828.

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/ingester/transfer.go
```diff
@@ -624,176 +387,3 @@ func (i *Ingester) findTargetIngester(ctx context.Context) (*ring.IngesterDesc,
 
 	return &ingesters[0], nil
 }
-
-// unshippedBlocks returns a ulid list of blocks that haven't been shipped
```
pstibrany, line 628: This function is useful for prometheus retention policy. We may want to revive it back soon.

pkg/storage/tsdb/config.go
```diff
@@ -172,7 +172,7 @@ func (cfg *TSDBConfig) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.HeadCompactionIdleTimeout, "experimental.blocks-storage.tsdb.head-compaction-idle-timeout", 1*time.Hour, "If TSDB head is idle for this duration, it is compacted. 0 means disabled.")
 	f.IntVar(&cfg.StripeSize, "experimental.blocks-storage.tsdb.stripe-size", 16384, "The number of shards of series to use in TSDB (must be a power of 2). Reducing this will decrease memory footprint, but can negatively impact performance.")
 	f.BoolVar(&cfg.WALCompressionEnabled, "experimental.blocks-storage.tsdb.wal-compression-enabled", false, "True to enable TSDB WAL compression.")
-	f.BoolVar(&cfg.FlushBlocksOnShutdown, "experimental.blocks-storage.tsdb.flush-blocks-on-shutdown", false, "If true, and transfer of blocks on shutdown fails or is disabled, incomplete blocks are flushed to storage instead. If false, incomplete blocks will be reused after restart, and uploaded when finished.")
+	f.BoolVar(&cfg.FlushBlocksOnShutdown, "experimental.blocks-storage.tsdb.flush-blocks-on-shutdown", false, "True to flush blocks to storage on shutdown. If false, incomplete blocks will be reused after restart, and uploaded when finished.")
```
pstibrany, line 175: I'd remove "uploaded when finished" part -- that depends on shipper configuration.

docs/architecture.md
```diff
@@ -153,17 +153,18 @@ Incoming series are not immediately written to the storage but kept in memory an
...
-
-5. `UNHEALTHY` is an ingester's state when it has failed to heartbeat to the ring's KV Store. While in this state, distributors skip the ingester while building the replication set for incoming series and the ingester does not receive write or read requests.
-
-For more information about the hand-over process, please check out the [Ingester hand-over](guides/ingester-handover.md) documentation.
+- **`PENDING`**<br />
+  The ingester has just started. While in this state the ingester doesn't receive neither write and read requests.
+- **`JOINING`**<br />
+  The ingester is starting up and joining the ring. While in this state the ingester doesn't receive neither write and read requests.
```
pstibrany, line 159: Old text has explained "joining the ring" part better -- ingester is adding the tokens, and possibly observing the ring for conflicts. It may also be receiving data from another ingester.

docs/architecture.md
```diff
@@ -153,17 +153,18 @@ Incoming series are not immediately written to the storage but kept in memory an
...
-
-4. `LEAVING` is an ingester's state when it is shutting down. It cannot receive write requests anymore, while it could still receive read requests for series it has in memory. While in this state, the ingester may look for a `PENDING` ingester to start a hand-over process with, used to transfer the state from `LEAVING` ingester to the `PENDING` one, during a rolling update (`PENDING` ingester moves to `JOINING` state during hand-over process). If there is no new ingester to accept hand-over, ingester in `LEAVING` state will flush data to storage instead.
-
-5. `UNHEALTHY` is an ingester's state when it has failed to heartbeat to the ring's KV Store. While in this state, distributors skip the ingester while building the replication set for incoming series and the ingester does not receive write or read requests.
-
-For more information about the hand-over process, please check out the [Ingester hand-over](guides/ingester-handover.md) documentation.
+- **`PENDING`**<br />
+  The ingester has just started. While in this state the ingester doesn't receive neither write and read requests.
```
pstibrany, line 157: It is also possibly waiting for data transfer request from another ingester.

docs/guides/ingesters-rolling-updates.md
```diff
@@ -0,0 +1,91 @@
...
+
+The Cortex chunks storage optionally supports a write-ahead log (WAL).
+The rolling update procedure for a Cortex cluster running the chunks storage depends whether the WAL is enabled or not.
+
+### Chunks storage with WAL enabled
+
+Similarly to the blocks storage, when Cortex is running the chunks storage with WAL enabled, it requires ingesters to run with a persistent disk where the WAL is stored (eg. a StatefulSet when deployed on Kubernetes).
+
+During a rolling update, the leaving ingester closes the WAL, synchronize the data to disk (`fsync`) and releases the disk resources.
+The new ingester, which is expected to reuse the same disk of the leaving one, will replay the WAL on startup in order to load back in memory the time series that have not been flushed to the storage yet.
```
pstibrany, line 37: It actually loads all data into memory, even flushed samples. WAL doesn't preserve "flushed" flag, so data will be possibly flushed twice.

docs/guides/ingesters-rolling-updates.md
```diff
@@ -0,0 +1,91 @@
...
+- `ACTIVE`
+- `LEAVING`
+
+On startup, an ingester goes into the **`PENDING`** state.
+In this state, the ingester is waiting for a hand-over from another ingester that is `LEAVING`.
+If no hand-over occurs within the configured timeout period ("auto-join timeout", configurable via `-ingester.join-after` option), the ingester will join the ring with a new set of random tokens (eg. during a scale up) and will switch its state to `ACTIVE`.
+
+A running ingester, in the **`ACTIVE`** state, is notified to shutdown via the Unix signal `SIGINT`.
+On receipt of this signal, the ingester switches to `LEAVING`: in this state it cannot receive write requests anymore, while it could still receive read requests for series it has in memory.
+
```
pstibrany, line 58:
```suggestion
When a running ingester in the **`ACTIVE`** state is notified to shutdown via `SIGINT` or `SIGTERM` Unix signal, the ingester switches to `LEAVING` state. In this state it cannot receive write requests anymore, but it can still receive read requests for series it has in memory.
```

docs/configuration/single-process-config-blocks.yaml
```diff
@@ -28,7 +28,7 @@ ingester_client:
     use_gzip_compression: true
 
 ingester:
-  # Disable blocks transfers on ingesters shutdown or rollout.
+  # Blocks storage doesn't support blocks transfer between ingesters on shutdown or rollout.
```
pstibrany, line 31: Remove this option completely? (in all YAML files using blocks)

pkg/ingester/transfer.go
pracucci: Yes, but could be slightly different (eg. moved it to a different struct). I would prefer to revive it when needed.

docs/configuration/single-process-config-blocks.yaml
pracucci: Better.

docs/architecture.md
pracucci: What do you think about the change done in https://github.com/cortexproject/cortex/pull/2996/commits/caa7dd453b918cd0674cbc52d999490e3ceebe0b ?

docs/architecture.md
pracucci: What do you think about the change done in https://github.com/cortexproject/cortex/pull/2996/commits/caa7dd453b918cd0674cbc52d999490e3ceebe0b ?

docs/guides/ingesters-rolling-updates.md
pracucci: Right. I'm not sure we should mention the "data will be possibly flushed twice" here, given this page is focused on rolling updates. Maybe it's something we should improve in the "Ingesters with WAL" page. WDYT?

In the meanwhile, I've rephrased it inhttps://github.com/cortexproject/cortex/pull/2996/commits/caa7dd453b918cd0674cbc52d999490e3ceebe0b.

pstibrany: Nit: In that change you have introduce "checkpoint" term, with no explanation. I think we can be bit more vague and just say that ingester reloads data from WAL.

docs/architecture.md
pstibrany: Changes look good. Thanks.

docs/guides/ingesters-rolling-updates.md
pracucci: Agree, done.

PR #2997, opened 2020-08-07
Cherry pick PR #2989 into release-1.3

pracucci:
**What this PR does**:
The PR #2989 fixes an unintentional breaking change introduced in the release `1.3.0-rc.0`. In this PR I'm cherry picking it.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/ruler/notifier.go
```diff
@@ -75,10 +76,11 @@ func (rn *rulerNotifier) stop() {
 // Builds a Prometheus config.Config from a ruler.Config with just the required
 // options to configure notifications to Alertmanager.
 func buildNotifierConfig(rulerConfig *Config) (*config.Config, error) {
-	validURLs := make([]*url.URL, 0, len(rulerConfig.AlertmanagerURL))
+	amURLs := strings.Split(rulerConfig.AlertmanagerURL, ",")
+	validURLs := make([]*url.URL, 0, len(amURLs))
 
 	srvDNSregexp := regexp.MustCompile(`^_.+._.+`)
```
pstibrany, line 82: This regular expression seems very weak and matches lot of unwanted inputs (eg. `_hello_world`).

pracucci: Could you open an issue for that and assign it to @gotjosh, please?

gotjosh: That was deliberate, `_hello_world` I'd assume could be a valid service name in:

`_service._proto.name. TTL class SRV priority weight port target.`

Taken from the [Wikipedia entry for SRV](https://en.wikipedia.org/wiki/SRV_record) records.

pstibrany: That regexp would treat `http://_hello_world/` as a valid URL for alert-manager discovery, even though it's missing `._tcp` part.

Slightly better regexp may be: `^_.+\._.+\.` or `^_.+\._tcp\.` (with case-insensitive match).

Probably not worth fixing, since  resolution for invalid names will fail anyway.

PR #2998, opened 2020-08-07
Add chunk flush outcome "Dropped", for short series

bboreham:
**What this PR does**:

If the series is below a configured length it gets dropped instead of flushed.  Add a separate `outcome` for this case, otherwise it is reported as "NoChunksToFlush".

**Checklist**
- NA Tests updated
- NA Documentation added

If we can get this into 1.3.0 then I don't think it needs a changelog entry; it is amending #2818 which didn't state any specific `outcome` values


pstibrany:
> If we can get this into 1.3.0 then I don't think it needs a changelog entry; it is amending #2818 which didn't state any specific outcome values

Perhaps we can update existing changelog entry to mention this PR number too?

pracucci:
> > If we can get this into 1.3.0 then I don't think it needs a changelog entry; it is amending #2818 which didn't state any specific outcome values
> 
> Perhaps we can update existing changelog entry to mention this PR number too?

I can do it when cherry-picking it for the 1.3.0 release.

PR #3000, opened 2020-08-08
Bump prometheus and thanos to master

yeya24:

**What this PR does**:

This pr updates Prometheus and Thanos version to the current master, required by [this pr](https://github.com/thanos-io/thanos/pull/2973)

Update Prometheus to https://github.com/prometheus/prometheus/commit/2899773b0159d0d4b06d65a77c0832060edacd4c.

Update Thanos to https://github.com/thanos-io/thanos/commit/9b578afb476374b68dd5ac5bfe50bcef70690760.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
Cortex changes look good to me.

yeya24:
> Cortex changes look good to me.

Thanks for the review! Do you think a changelog entry is needed?

pstibrany:
> > Cortex changes look good to me.
> 
> Thanks for the review! Do you think a changelog entry is needed?

I don't think so in this case. There are some new Thanos metrics (`thanos_memcached_client_info`, `thanos_memcached_operation_data_size_bytes`), that will get exposed by Cortex, but I don't think we need to document them.

pkg/ruler/compat.go
```diff
@@ -63,7 +63,7 @@ type PusherAppendable struct {
 }
 
 // Appender returns a storage.Appender
-func (t *PusherAppendable) Appender() storage.Appender {
+func (t *PusherAppendable) Appender(_ context.Context) storage.Appender {
```
annanay25, line 66: I think we should propagate context from here to `Pusher.Push()`. It will also help to remove `userID` parameter from `pusherAppender` struct. We can encode the userID into context, in ruler.go here https://github.com/cortexproject/cortex/blob/a01c1012643a8afbc259950db659e4a9e587259c/pkg/ruler/ruler.go#L543 
.. and extract userID from context in this file as part of `DefaultTenantManagerFactory`

pstibrany: > It will also help to remove userID parameter from pusherAppender struct. We can encode the userID into context, in ruler.go here

In other areas of the code, we prefer to pass userID explicitly, and not hide it into Context. But Cortex codebase is definitely not consistent about it :(

annanay25: Oh, thanks @pstibrany! Would you prefer we separate it out? Its definitely a by-product in this PR, its more important we pass context along :)

pkg/ruler/ruler.go
```diff
@@ -540,7 +540,12 @@ func (r *Ruler) newManager(ctx context.Context, userID string) (*promRules.Manag
 	r.userManagerMetrics.AddUserRegistry(userID, reg)
 
 	logger := log.With(r.logger, "user", userID)
-	return r.managerFactory(ctx, userID, notifier, logger, reg), nil
+	ctx = user.InjectUserID(ctx, userID)
+	manager := r.managerFactory(ctx, notifier, logger, reg)
+	if manager == nil {
+		return nil, errors.New("unable to create managerFactory")
```
annanay25, line 546: nit: s/managerFactory/manager :)

pkg/ruler/compat.go
pstibrany: > Would you prefer we separate it out?

Yes, I think it would be bit cleaner approach. Thanks! (And sorry for not reacting earlier)

annanay25: No worries @pstibrany :)

pracucci: Yes, let's be explicit here and pass `userID` as parameter.

pracucci:
Please be aware of this issue causing segfault. Not sure Prometheus we could be affected as well, but worth to keep an eye on it: https://github.com/prometheus/prometheus/issues/7776

yeya24:
Sorry for the delay, I just updated my pr and updates Prometheus version to current master. Please help review and let me know if I need to miss something.

pracucci:
@yeya24 FYI I rebased `master` and fixed the `CHANGELOG` entry ordering in [this commit](https://github.com/cortexproject/cortex/pull/3000/commits/b328d5b5a75afd710539dc8851bbd979a477eae7).

yeya24:
> @yeya24 FYI I rebased `master` and fixed the `CHANGELOG` entry ordering in [this commit](https://github.com/cortexproject/cortex/pull/3000/commits/b328d5b5a75afd710539dc8851bbd979a477eae7).

Awesome! Thanks for the help.

pracucci:
_I run a local stress test for about 30 minutes. No issue found._

pracucci:
> I would still prefer to revert change that pushes injecting of userID into context to upper layer, and instead keep dependency on userID as explicit as possible.

I take care of that so we can proceed merging it and unblocking several people (@annanay25 included).

PR #3015, opened 2020-08-11
Batch series in streaming ingester based on message sizes.

### LINKED ISSUE #2945: Stream-querying Ingester v2 hits message size limits
`v2QueryStream` method of Ingester can produce messages that are still too be to be sent over gRPC. Right now its logic is to send batches of 128 time series, but if timeseries contain many samples (eg. when asking for large time range), it may still hit gRPC message size limit.

With 6 samples/minute, over 12h time that is 4000+ samples for single timeseries. If we batch 128 of them, that is 500k + samples in single batch.

Batching in v2QueryStream ingester could be a little smarter to avoid hitting the limit.

pstibrany:
**What this PR does**: This PR modifies batching in blocks-based ingester, when it is streaming results back. Instead of using only number of series for batching, it also considers how much data is buffered in memory, and flushes batch if it get too big. This is to prevent problems with too large gRPC messages. "Too big" is hardcoded to 1 MiB, to comfortably fit into grpc default, and also to avoid using too much memory.

Note that if *single* series returns too much data, it can still reach gRPC limit for sending messages. Solution to that would be splitting single timeseries into multiple messages, but querier doesn't support that at the moment.

Fixes #2945 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -757,6 +757,8 @@ func createUserStats(db *userTSDB) *client.UserStatsResponse {
 	}
 }
 
+const queryStreamBatchMessageSize = 1 * 1024 * 1024
+
 // v2QueryStream streams metrics from a TSDB. This implements the client.IngesterServer interface
 func (i *Ingester) v2QueryStream(req *client.QueryRequest, stream client.Ingester_QueryStreamServer) error {
 	log, ctx := spanlogger.New(stream.Context(), "v2QueryStream")
@@ -792,7 +794,7 @@ func (i *Ingester) v2QueryStream(req *client.QueryRequest, stream client.Ingeste
 	}
 
 	timeseries := make([]client.TimeSeries, 0, queryStreamBatchSize)
-	batchSize := 0
+	batchBytesSize := 0 // In bytes
 	numSamples := 0
 	numSeries := 0
 	for ss.Next() {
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -809,21 +811,25 @@ func (i *Ingester) v2QueryStream(req *client.QueryRequest, stream client.Ingeste
 			ts.Samples = append(ts.Samples, client.Sample{Value: v, TimestampMs: t})
 		}
 		numSamples += len(ts.Samples)
-
-		timeseries = append(timeseries, ts)
 		numSeries++
-		batchSize++
-		if batchSize >= queryStreamBatchSize {
+		tsSize := ts.Size()
```
pracucci, line 815: How complicated would be adding a benchmark to the `v2QueryStream()`, to measure the performance impact of `ts.Size()`? I'm probably just paranoic, but if it's not much complicated we could measure it.

pstibrany:
Here is a result of benchmark before and after this change.

```
name                      old time/op    new time/op    deltanchstat before.txt after.txt 
Ingester_v2QueryStream-4    6.31ms ± 2%    6.46ms ± 1%  +2.38%  (p=0.001 n=10+8)

name                      old alloc/op   new alloc/op   delta
Ingester_v2QueryStream-4    3.45MB ± 0%    3.45MB ± 0%  +0.00%  (p=0.000 n=10+10)

name                      old allocs/op  new allocs/op  delta
Ingester_v2QueryStream-4     4.45k ± 0%     4.45k ± 0%  +0.02%  (p=0.000 n=10+10)
```

pracucci:
Thanks @pstibrany! LGTM

> Should we do it for the chunks storage too, to keep it specular? :)

Could you open an issue for that, please?

pstibrany:
> Could you open an issue for that, please?

I will take a look first if we can make it part of this PR (if it's small change).

PR #3017, opened 2020-08-11
Fixes the Alertmanager panicking when no external URL is provided

### LINKED ISSUE #2704: Alertmanager segfaults if I don't pass the -alertmanager.web.external-url flag
Not sure if it is an issue. But when I look at the documentation for alertmanager, it doesn't seem to me that `-alertmanager.web.external-url` flag is a required flag. Here is what happens when I don't pass that flag. It works fine with that flag though.

```

$ sudo docker run --rm  --name=alertmanager  --network=host  cortex:v1.0.1 alertmanager.configs.url=http://configs.cortex:3001  -server.http-listen-port 2001 -server.grpc-listen-port 2002 -log.level=debug  -target alertmanager
level=info ts=2020-06-09T15:57:57.010945071Z caller=server.go:147 http=[::]:2001 grpc=[::]:2002 msg="server listening on addresses"
level=debug ts=2020-06-09T15:57:57.012990789Z caller=cluster.go:149 component=cluster msg="resolved peers to following addresses" peers=
level=info ts=2020-06-09T15:57:57.024781121Z caller=cluster.go:161 component=cluster msg="setting advertise address explicitly" addr=10.0.2.15 port=9094
level=debug ts=2020-06-09T15:57:57.0262938Z caller=delegate.go:230 component=cluster received=NotifyJoin node=01EACWTVX1BCH35X2KV4CT1R4D addr=10.0.2.15:9094
level=debug ts=2020-06-09T15:57:57.026771062Z caller=cluster.go:233 component=cluster msg="joined cluster" peers=0
level=info ts=2020-06-09T15:57:57.027893802Z caller=main.go:109 msg="Starting Cortex" version="(version=, branch=, revision=)"
level=info ts=2020-06-09T15:57:57.029518485Z caller=cluster.go:623 component=cluster msg="Waiting for gossip to settle..." interval=200ms
level=debug ts=2020-06-09T15:57:57.029920472Z caller=module_service.go:48 msg="module waiting for initialization" module=alertmanager waiting_for=server
level=info ts=2020-06-09T15:57:57.030184558Z caller=module_service.go:58 msg=initialising module=alertmanager
level=debug ts=2020-06-09T15:57:57.046158906Z caller=multitenant.go:283 component=MultiTenantAlertmanager msg="initial configuration load" num_configs=2
level=debug ts=2020-06-09T15:57:57.046604378Z caller=multitenant.go:312 component=MultiTenantAlertmanager msg="adding configurations" num_configs=2
level=debug ts=2020-06-09T15:57:57.046985744Z caller=multitenant.go:400 component=MultiTenantAlertmanager msg="setting config" user=team-X
level=debug ts=2020-06-09T15:57:57.047624125Z caller=multitenant.go:429 component=MultiTenantAlertmanager msg="initializing new per-tenant alertmanager" user=team-X
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x14da0a0]

goroutine 337 [running]:
github.com/cortexproject/cortex/pkg/alertmanager.New(0xc00017c6e0, 0xc00017c690, 0xc00064f0e0, 0xc00068e300, 0xc)
	/go/src/github.com/cortexproject/cortex/pkg/alertmanager/alertmanager.go:171 +0xe90
github.com/cortexproject/cortex/pkg/alertmanager.(*MultitenantAlertmanager).newAlertmanager(0xc0007ea600, 0xc000654d6a, 0x6, 0xc0006ca770, 0x0, 0x0, 0xc00036c960)
	/go/src/github.com/cortexproject/cortex/pkg/alertmanager/multitenant.go:451 +0x183
github.com/cortexproject/cortex/pkg/alertmanager.(*MultitenantAlertmanager).setConfig(0xc0007ea600, 0xc000654d6a, 0x6, 0xc000122750, 0x8b, 0x0, 0x0, 0x0, 0x2f119a0, 0x3a49680)
	/go/src/github.com/cortexproject/cortex/pkg/alertmanager/multitenant.go:430 +0xac9
github.com/cortexproject/cortex/pkg/alertmanager.(*MultitenantAlertmanager).syncConfigs(0xc0007ea600, 0xc000835140)
	/go/src/github.com/cortexproject/cortex/pkg/alertmanager/multitenant.go:314 +0x2e0
github.com/cortexproject/cortex/pkg/alertmanager.(*MultitenantAlertmanager).starting(0xc0007ea600, 0x3a96880, 0xc00078a780, 0x1, 0xc000836530)
	/go/src/github.com/cortexproject/cortex/pkg/alertmanager/multitenant.go:247 +0x39
github.com/cortexproject/cortex/pkg/util/services.(*BasicService).main(0xc00004c630)
	/go/src/github.com/cortexproject/cortex/pkg/util/services/basic_service.go:134 +0x459
created by github.com/cortexproject/cortex/pkg/util/services.(*BasicService).StartAsync.func1
	/go/src/github.com/cortexproject/cortex/pkg/util/services/basic_service.go:96 +0xa8_
```

gotjosh:
**What this PR does**:

Two really small improvements, and a precedent to start the discussion around automatic fallback configuration.

Fixes #2704 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,8 @@
 * [BUGFIX] Query-frontend: Fixed rounding for incoming query timestamps, to be 100% Prometheus compatible. #2990
 * [BUGFIX] Querier: query /series from ingesters regardless the `-querier.query-ingesters-within` setting. #3035
 * [BUGFIX] Experimental blocks storage: Ingester is less likely to hit gRPC message size limit when streaming data to queriers. #3015
-* [BUGFIX] Fixes the Alertmanager panicking when no `alertmanager.web.external-url` is provided. #3017
 * [BUGFIX] Fix configuration for TLS server validation, TLS skip verify was hardcoded to true for all TLS configurations and prevented validation of server certificates. #3030
+* [BUGFIX] Fixes the Alertmanager panicking when no `-alertmanager.web.external-url` is provided. #3017
 
 ## 1.3.0 in progress
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * [BUGFIX] Query-frontend: Fixed rounding for incoming query timestamps, to be 100% Prometheus compatible. #2990
 * [BUGFIX] Querier: query /series from ingesters regardless the `-querier.query-ingesters-within` setting. #3035
 * [BUGFIX] Experimental blocks storage: Ingester is less likely to hit gRPC message size limit when streaming data to queriers. #3015
+* [BUGFIX] Fixes the Alertmanager panicking when no `alertmanager.web.external-url` is provided. #3017
 * [BUGFIX] Fix configuration for TLS server validation, TLS skip verify was hardcoded to true for all TLS configurations and prevented validation of server certificates. #3030
 
 ## 1.3.0 in progress
--- a/integration/e2ecortex/client.go
+++ b/integration/e2ecortex/client.go
@@ -212,6 +212,10 @@ func (c *Client) GetAlertmanagerConfig(ctx context.Context) (*alertConfig.Config
 		return nil, ErrNotFound
 	}
 
+	if resp.StatusCode/100 != 2 {
+		return nil, fmt.Errorf("getting config failed with status %d and error %v", resp.StatusCode, string(body))
+	}
+
 	var ss *ServerStatus
 	err = json.Unmarshal(body, &ss)
 	if err != nil {
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -484,7 +484,7 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	am.alertmanagersMtx.Unlock()
 
 	if !ok {
-		http.Error(w, "no Alertmanager for this user ID, Alertmanager not configured", http.StatusNotFound)
+		http.Error(w, "the Alertmanager has not been configured", http.StatusNotFound)
 		return
 	}
 
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -175,6 +175,10 @@ func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, logger log.L
 		return nil, fmt.Errorf("unable to create Alertmanager data directory %q: %s", cfg.DataDir, err)
 	}
 
+	if cfg.ExternalURL.URL == nil {
+		return nil, fmt.Errorf("unable to create Alertmanager, no 'web.external-url' configured")
+	}
+
 	var fallbackConfig []byte
 	if cfg.FallbackConfigFile != "" {
 		fallbackConfig, err = ioutil.ReadFile(cfg.FallbackConfigFile)
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -483,13 +483,8 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok {
-		http.Error(w, "the Alertmanager has not been configured", http.StatusNotFound)
-		return
-	}
-
-	if !userAM.IsActive() {
-		http.Error(w, "Alertmanager is not active, revise its configuration", http.StatusNotAcceptable)
+	if !ok || !userAM.IsActive() {
+		http.Error(w, "the Alertmanager is not configured", http.StatusNotFound)
 		return
 	}
 
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -483,10 +483,16 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok || !userAM.IsActive() {
-		http.Error(w, "no Alertmanager for this user ID", http.StatusNotFound)
+	if !ok {
+		http.Error(w, "no Alertmanager for this user ID, Alertmanager not configured", http.StatusNotFound)
 		return
 	}
+
+	if !userAM.IsActive() {
+		http.Error(w, "Alertmanager is not active, revise its configuration", http.StatusNotAcceptable)
+		return
+	}
+
 	userAM.mux.ServeHTTP(w, req)
 }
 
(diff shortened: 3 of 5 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -479,10 +483,16 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok || !userAM.IsActive() {
```
gotjosh, line 482: Ideally, what I'd like to do here is use the fallback configuration we support and start an Alertmanager instance. Then, we can let you that your Alertmanager instance is warming up. 

I believe this is the only component without a UI, as this is a pattern we've not seen before.

My question here, is that a step in the right direction? Is that something we'd like to include in Cortex? Authentication/Authorization is not a concern of Cortex, under that assumption, I'd expect that any user trying to access this URL should already have a manager starting - using the fallback configuration.

gouthamve: Hrm, why do we have the fallback configuration for if its not being used in this way?

gotjosh: It triggers when you have an empty configuration:

https://github.com/cortexproject/cortex/blob/e36e2ae2c08fd81cb7f8d0d0f8cba1357caa5d90/pkg/alertmanager/multitenant.go#L401-L404

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * [ENHANCEMENT] Add support for azure storage in China, German and US Government environments. #2988
 * [ENHANCEMENT] Query-tee: added a small tolerance to floating point sample values comparison. #2994
 * [BUGFIX] Query-frontend: Fixed rounding for incoming query timestamps, to be 100% Prometheus compatible. #2990
+* [BUGFIX] Fixes the Alertmanager panicking when no `alertmanager.web.external-url` is provided. #3017
```
pracucci, line 13:
```suggestion
* [BUGFIX] Fixes the Alertmanager panicking when no `-alertmanager.web.external-url` is provided. #3017
```

integration/e2ecortex/client.go
```diff
@@ -205,6 +205,10 @@ func (c *Client) GetAlertmanagerConfig(ctx context.Context) (*alertConfig.Config
 		return nil, ErrNotFound
 	}
 
+	if resp.StatusCode/100 != 2 {
+		return nil, fmt.Errorf("getting config failed with status %d and error %v", resp.StatusCode, string(body))
```
pracucci, line 209: Exposing the response body of the underlying storage is generally risky from a security perspective, because it may potentially leak sensitive data. I would discourage it.

pkg/alertmanager/multitenant.go
```diff
@@ -175,6 +175,10 @@ func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, logger log.L
 		return nil, fmt.Errorf("unable to create Alertmanager data directory %q: %s", cfg.DataDir, err)
 	}
 
+	if cfg.ExternalURL.URL == nil {
+		return nil, fmt.Errorf("unable to create Alertmanager, no 'web.external-url' configured")
```
pracucci, line 179: Either we display the exact CLI flag (`-alertmanager.web.external-url`) or we explain it in plain english. I would discourage the CLI flag because the user may use the YAML config file and have no idea about the flag.

```suggestion
		return nil, fmt.Errorf("unable to create Alertmanager because the external URL has not been configured")
```

pkg/alertmanager/multitenant.go
```diff
@@ -479,10 +483,16 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok || !userAM.IsActive() {
-		http.Error(w, "no Alertmanager for this user ID", http.StatusNotFound)
+	if !ok {
+		http.Error(w, "no Alertmanager for this user ID, Alertmanager not configured", http.StatusNotFound)
```
pracucci, line 487: Given the request comes from the user itself, we may rephrase it?

```suggestion
		http.Error(w, "the Alertmanager has not been configured", http.StatusNotFound)
```

pkg/alertmanager/multitenant.go
```diff
@@ -479,10 +483,16 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+		return
+	}
+
+	if !userAM.IsActive() {
+		http.Error(w, "Alertmanager is not active, revise its configuration", http.StatusNotAcceptable)
```
pracucci, line 492: I'm not sure about the message here. Internally we activate/deactivate the alertmanager because we can't completely remove it, but from the final user perspective it's the same exact thing as "Alertmanager has not been configured".

integration/e2ecortex/client.go
gotjosh: This is the test client though - is this still worth fixing?

pracucci: Definitely not. Sorry I was probably drunk 🤦

pkg/alertmanager/multitenant.go
gotjosh: I prefer the active voice instead of a passive one, but YMMV.

pkg/alertmanager/multitenant.go
gotjosh: You're completely right - paused is only used when the Alertmanager is gone. I'll have to rethink this one.

pkg/alertmanager/multitenant.go
pracucci: Sure, feel free to phrase it as you think is better, my comment was about:
1. not mention "user ID" ("user ID" is an internal thing, the final user just cares about themselves)
2. "no Alertmanager for this user" looks a bit unclear to me, while "Alertmanager not configured" looks more clear, so we should just say the Alertmanager hasn't been configured yet

pkg/alertmanager/multitenant_test.go
```diff
@@ -177,3 +179,69 @@ func TestLoadAllConfigs(t *testing.T) {
...
+	require.NoError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	// Create the Multitenant Alertmanager.
+	reg := prometheus.NewPedanticRegistry()
+	_, err = NewMultitenantAlertmanager(&MultitenantAlertmanagerConfig{
+		DataDir: tempDir,
+	}, log.NewNopLogger(), reg)
+
+	require.EqualError(t, err, "unable to create Alertmanager, no 'web.external-url' configured")
```
pracucci, line 194: Not sure why CI passed, but shouldn't given now the error message is different.

pkg/alertmanager/multitenant.go
pracucci: Ok, let's address this and then we can go.

pracucci: What do you think to rollback the logic, so we can unblock this PR and get it merged?

pracucci (approved):
LGTM, thanks!

PR #3019, opened 2020-08-12
Refactor Ruler, introduce MultiTenantManager interface

annanay25:
**What this PR does**:
This PR refactors ruler package by introducing a `MultiTenantManager` interface. This is done for multiple reasons. 

First, it splits out the Ruler struct into distinct components, and the Ruler is seen as a module that ties together the rule storage, ring for sharding and Manager for evaluation and notification on the rules.

```
+---------------------------------------------------------------+
|                                                               |
|                   Query       +-------------+                 |
|            +------------------>             |                 |
|            |                  |    Store    |                 |
|            | +----------------+             |                 |
|            | |     Rules      +-------------+                 |
|            | |                                                |
|            | |                                                |
|            | |                                                |
|       +----+-v----+   Filter  +------------+                  |
|       |           +----------->            |                  |
|       |   Ruler   |           |    Ring    |                  |
|       |           <-----------+            |                  |
|       +-------+---+   Rules   +------------+                  |
|               |                                               |
|               |                                               |
|               |                                               |
|               |    Load      +-----------------+              |
|               +-------------->                 |              |
|                              |     Manager     |              |
|                              |                 |              |
|                              +-----------------+              |
|                                                               |
+---------------------------------------------------------------+

```

Secondly, it allows us to add custom functionality into the Ruler depending on custom fields present in the RuleGroup description, by overriding the `DefaultMultiTenantManager`. For ex: This brings flexibility in using different appenders based on special config fields present in the RuleGroup definition. 

Fixes #<issue number>

**Checklist**
Will drop draft status once:
- [x] #3000 is merged (this PR is built on top of that one)
- [x] Tests updated
- [x] Documentation added

/cc @jtlisi


pkg/ruler/ruler_test.go
```diff
@@ -126,16 +140,14 @@ func TestNotifierSendsUserIDHeader(t *testing.T) {
 	cfg.AlertmanagerURL = ts.URL
 	cfg.AlertmanagerDiscovery = false
 
-	r, rcleanup := newTestRuler(t, cfg)
+	manager, rcleanup := newManager(t, cfg)
 	defer rcleanup()
-	defer services.StopAndAwaitTerminated(context.Background(), r) //nolint:errcheck
+	//defer services.StopAndAwaitTerminated(context.Background(), r) //nolint:errcheck
```
pracucci, line 145: Please remove commented code.

pkg/ruler/manager.go
```diff
@@ -0,0 +1,231 @@
...
+	promRules "github.com/prometheus/prometheus/rules"
+	"github.com/weaveworks/common/user"
+	"golang.org/x/net/context/ctxhttp"
+
+	store "github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+var (
+	managersTotal = promauto.NewGauge(prometheus.GaugeOpts{
```
pracucci, line 24: We try to avoid global metrics. Can you move it to `ManagerMetrics`?

pkg/ruler/ruler.go
```diff
@@ -266,32 +255,12 @@ func (r *Ruler) starting(ctx context.Context) error {
 // Stop stops the Ruler.
 // Each function of the ruler is terminated before leaving the ring
 func (r *Ruler) stopping(_ error) error {
-	r.notifiersMtx.Lock()
-	for _, n := range r.notifiers {
-		n.stop()
```
pracucci, line 271: This was done before stopping the subservices. I believe you should run `r.manger.Stop()` before stopping the subservices. WDYT?

pkg/ruler/manager.go
```diff
@@ -0,0 +1,231 @@
...
+		mapper:             newMapper(cfg.RulePath, logger),
+		userManagers:       map[string]*promRules.Manager{},
+		userManagerMetrics: userManagerMetrics,
+		registry:           reg,
+		logger:             logger,
+	}, nil
+}
+
+func (r *DefaultMultiTenantManager) SyncRuleGroups(ctx context.Context, ruleGroups map[string]store.RuleGroupList) {
+	// A lock is taken to ensure if syncManager is called concurrently, that each call
```
pracucci, line 83:
```suggestion
	// A lock is taken to ensure if this function is called concurrently, then each call
```

pkg/ruler/manager.go
```diff
@@ -0,0 +1,231 @@
...
+			go mngr.Stop()
+			delete(r.userManagers, userID)
+			level.Info(r.logger).Log("msg", "deleting rule manager", "user", userID)
+		}
+	}
+
+	managersTotal.Set(float64(len(r.userManagers)))
+}
+
+// syncManager maps the rule files to disk, detects any changes and will create/update the
```
pracucci, line 104:
```suggestion
// syncRulesToManager maps the rule files to disk, detects any changes and will create/update the
```

pkg/ruler/manager.go
annanay25: If I understand correctly, `ManagerMetrics` is actually something passed into the Prometheus manager and does not contain ruler level metrics. This particular `managersTotal` metric was already a global metric in `ruler.go` here - https://github.com/cortexproject/cortex/blob/e8a66862fa0500a65ceff575f96616460262c8db/pkg/ruler/ruler.go#L56-L60

Please do correct me if I'm wrong. :)

pkg/ruler/ruler.go
annanay25: I agree with you. I will move it.

pkg/ruler/manager.go
pracucci: Right. Then you can just define the metric in `DefaultMultiTenantManager`.

pkg/ruler/ruler.go
```diff
@@ -150,58 +140,53 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
-		return nil, err
-	}
-
-	userManagerMetrics := NewManagerMetrics()
-
-	if reg != nil {
-		reg.MustRegister(userManagerMetrics)
-	}
-
+func NewRuler(cfg Config, manager MultiTenantManager, reg prometheus.Registerer, logger log.Logger, ruleStore rules.RuleStore) (*Ruler, error) {
```
annanay25, line 183: @pracucci - I've made one more small change after you reviewed this PR (Sorry about this, I was going to create another follow up PR but unintentionally pushed changes as part of this itself 🤦 ). The change is that `NewRuler` now accepts a `MultiTenantManager` instead of a `ManagerFactory`.

pkg/ruler/ruler.go
```diff
@@ -150,58 +140,53 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 }
 
 // Ruler evaluates rules.
+//+---------------------------------------------------------------+
```
pstibrany, line 143: Please indent pre-formatted text, so that go doc and other tools display it correctly.

pkg/ruler/manager.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/prometheus/prometheus/notifier"
+	promRules "github.com/prometheus/prometheus/rules"
+	"github.com/weaveworks/common/user"
+	"golang.org/x/net/context/ctxhttp"
+
+	store "github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+type MultiTenantManager interface {
```
pstibrany, line 22: This should go next to Ruler, which uses the interface. Please add comments to this interface and its methods.

pkg/ruler/manager.go
```diff
@@ -0,0 +1,231 @@
...
+	if reg != nil {
+		reg.MustRegister(userManagerMetrics)
+	}
+
+	managersTotal := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "ruler_managers_total",
+		Help:      "Total number of managers registered and running in the ruler",
+	})
+	reg.MustRegister(managersTotal)
```
pstibrany, line 66: You can use `promauto.With(reg).NewGauge(...)` which handles the case when `reg` is `nil` (checked few lines earlier).

pkg/ruler/manager.go
```diff
@@ -66,6 +58,13 @@ func NewDefaultMultiTenantManager(cfg Config, managerFactory ManagerFactory, reg
 		reg.MustRegister(userManagerMetrics)
 	}
 
+	managersTotal := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "ruler_managers_total",
+		Help:      "Total number of managers registered and running in the ruler",
+	})
+	reg.MustRegister(managersTotal)
```
pracucci, line 66: Please do not use `reg.MustRegister()` but create the metric with `promauto.With(reg).NewGauge()`

pkg/ruler/ruler.go
pracucci: It's OK. It's part of this refactoring 👍

annanay25:
Thanks for the reviews, @pracucci & @pstibrany!

pkg/ruler/ruler.go
```diff
@@ -149,59 +139,64 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 }
 
+// MultiTenantManager is the interface of interaction with a Manager that is tenant aware.
+type MultiTenantManager interface {
+	// SyncRuleGroups is used to sync the Manager with rules from the RuleStore.
+	SyncRuleGroups(ctx context.Context, ruleGroups map[string]store.RuleGroupList)
+	// GetRules fetches rules for a particular tenant (userID).
+	GetRules(userID string) []*promRules.Group
+	// Stop stops all Manager components.
+	Stop()
```
pstibrany, line 149: Seeing `Stop` method in the generic interface looks like an anti-pattern to me. As Ruler is not creating the manager, I don't think it should be one stopping it either.

(But that can be addressed in another PR)

pkg/ruler/ruler.go
```diff
@@ -149,59 +139,64 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.ResendDelay, "ruler.resend-delay", time.Minute, `Minimum amount of time to wait before resending an alert to Alertmanager.`)
 }
 
+// MultiTenantManager is the interface of interaction with a Manager that is tenant aware.
```
pstibrany, line 142:
```suggestion
// MultiTenantManager interface describes interaction between Ruler, that periodically 
// loads rules from the store, and object that keeps rules in memory for each tenant.
```

It sounds to me like "multi-tenant aware rules manager" :)

pkg/ruler/ruler.go
annanay25: I agree, we need the `Stop()` function but I had a feeling the other function should be named `StartOrSyncRuleGroups()`. That sounds a little long, but do you have any other suggestions?

pkg/ruler/ruler.go
annanay25: > and object that keeps rules in memory for each tenant

The `DefaultMultiTenantManager` uses a mapper to write rules to a file and passes file path to the Prometheus Manager. Maybe we can modify this to

```
// MultiTenantManager interface describes interaction between Ruler, that periodically 
// loads rules from the store, and object that syncs rules to the Prometheus manager
// for every tenant.
```

PR #3025, opened 2020-08-12
Ingest samples older than 1h for block store

### LINKED ISSUE #2366: Blocks storage unable to ingest samples older than 1h after an outage

codesome:
**What this PR does**:

Supersedes #2819. I chose to open a new PR so that I can have the old one as reference for some parts.

This implements the updated design as described in https://github.com/cortexproject/cortex/issues/2366#issuecomment-667025973

Fixes #2366

- [x] Ingest samples older than 1h
- [x] Reload TSDBs on restart
- [x] Query samples from older TSDBs
- [x] Support query stream over old TSDBs
- [x] Automatically cleanup old TSDBs
- [x] Add metrics for counting backfill TSDBs open, errors in shipping/deleting, etc
- [x] Tests updated
- [ ] Documentation added

TODO in future PRs

- [ ] Metrics from new TSDBs
- [ ] Limits on the additional TSDB
- [ ] Automatically cleanup _stale_ TSDBs


pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,391 @@
...
+	userID      string
+	backfillAge time.Duration
+
+	// dbs[0] is the one which would be overlapping with the main TSDB.
+	// dbs[1] is the older TSDB among the two here.
+	// If the dbs[1] goes beyond the backfill age, it will be compacted
+	// and shipped and dbs[0] is moved to dbs[1]. During this process,
+	// there is only 1 TSDB ingesting old data, so there can be gaps till the compaction
+	// and shipping is going on.
+	// TODO(codesome): Avoid gaps by using a separate queue to compact and ship blocks.
```
codesome, line 34: Not for this PR

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,498 @@
...
+	return start, end
+}
+
+// getBucketName returns the string representation of the bucket.
+// YYYY_MM_DD_HH_YYYY_MM_DD_HH
+func getTSDBName(start, end int64) string {
+	startTime := model.Time(start).Time().UTC()
+	endTime := model.Time(end).Time().UTC()
+
+	return fmt.Sprintf(
```
pstibrany, line 437: `"<start>-<end>"` as unix timestamps may be easier to work with

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+	u.backfillTSDB.mtx.RUnlock()
+
+	if len(queriers) == 0 {
+		return nil, nil
+	}
+
+	result := make([]storage.SeriesSet, len(queriers))
+	for i, q := range queriers {
+		ss := q.Select(false, nil, matchers...)
```
pstibrany, line 269: By using unsorted Select, it will be difficult to merge correctly afterwards.

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,498 @@
...
+	defer b.compactMtx.Unlock()
+
+	var merr tsdb_errors.MultiError
+	secondCompacted := false
+	for i := range []int{1, 0} {
+		b.mtx.RLock()
+		db := b.dbs[i]
+		b.mtx.RUnlock()
+
+		if db == nil || (!force && db.end > mtime.Now().Add(-b.backfillAge-time.Hour).Unix()*1000) {
```
pstibrany, line 293: `db.end > mtime.Now().Add(-b.backfillAge-time.Hour).Unix()*1000` -- similar check repeats multiple times around the code... can we move it into a named function to explain its meaning?

codesome:
> I think naming-wise, dbs[0] and dbs[1] can perhaps be improved (eg. "older", "newer"?)

That was the first thing I tried. But it turned out to be a lot of duplicated code and maybe also makes it harder to approach the review. So I resorted to using an array and use the index to generalise the code. I will check again if explicit variable could be used after some simplifications now (and maybe more comments for context).

pkg/ingester/ingester_v2.go
```diff
@@ -442,13 +447,27 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 				}
 			}
 
+			cause := errors.Cause(err)
+			if cause == storage.ErrOutOfBounds &&
+				i.cfg.BlocksStorageConfig.TSDB.BackfillMaxAge > 0 &&
+				s.TimestampMs > mtime.Now().Add(-i.cfg.BlocksStorageConfig.TSDB.BackfillMaxAge-time.Hour).Unix()*1000 {
```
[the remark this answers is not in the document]
codesome: Sorry for not documenting it. 1hr comes form the main TSDB. I took the BackfillMaxAge to be additional time beyond the current 1hr old ingestion. I will add more comments, or would it be better for `BackfillMaxAge` to include the main TSDB? (in that case `BackfillMaxAge - 1h` should align with 24h)

pkg/ingester/ingester_v2.go
```diff
@@ -553,6 +582,16 @@ func (i *Ingester) v2Query(ctx context.Context, req *client.QueryRequest) (*clie
...
 	}
 
+	backfillSSs, err := db.backfillSelect(ctx, int64(from), int64(through), matchers)
+	if err != nil {
+		return nil, err
+	}
+	if len(backfillSSs) > 0 {
+		// TODO(codesome): If any TSDB in backfill buckets were overlapping
+		// with main TSDB, then use tsdb.NewMergedVerticalSeriesSet
+		ss = tsdb.NewMergedSeriesSet(append(backfillSSs, ss))
```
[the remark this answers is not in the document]
codesome: Uh, oh. Good catch!

pkg/ingester/ingester_v2.go
```diff
@@ -1092,6 +1160,62 @@ func (i *Ingester) openExistingTSDB(ctx context.Context) error {
...
+		start, end, userDB, err := openBackfillTSDB(backfillTSDBDirs[len(backfillTSDBDirs)-1].Name())
+		if err == nil {
+			db.backfillTSDB.dbs[0] = &backfillTSDBWrapper{
+				db:    userDB,
+				start: start,
+				end:   end,
+			}
+		}
+	}
+	if len(backfillTSDBDirs) > 1 {
```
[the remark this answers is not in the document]
codesome: Definitely! I missed that when I moved from explicit variables to an array.

pkg/ingester/ingester_v2_backfill.go
codesome: In case we have to debug, would it not be easier if the name told the exact hours without needing a manual translation? (thinking of the case where there might be multiple backfill TSDBs piled up due to various reasons).

pkg/ingester/ingester_v2_backfill.go
codesome: Definitely! Not sure why it was not already :)

pkg/ingester/ingester_v2_backfill.go
pstibrany: Yes, that's true, although there is a lot of precedent for unix timestamps at many places – eg. min/max time in meta.json of TSDB block, even block ID (ULID) has embedded timestamp in it. Your choice, this isn't a showstopper for me.

codesome:
> Avoid backward-dependancies from new types to ingester if possible. In general, if type A creates instances of type B, then B should not need to know about A: instead A needs to pass everything that B needs to it. I know ingester code doesn't always fit into this. :(

Need the ingester in `backfillAppender` only for `ingester.createNewTSDB` and `ingester.cfg.BlocksStorageConfig.TSDB.BackfillBlocksDir`. We could potentially keep those methods with `backfillTSDB` and use it from there instead of keeping the ingester around.

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,499 @@
...
+		startTime.Year(), startTime.Month(), startTime.Day(), startTime.Hour(),
+		endTime.Year(), endTime.Month(), endTime.Day(), endTime.Hour(),
+	)
+}
+
+func overlapsOpenInterval(mint1, maxt1, mint2, maxt2 int64) bool {
+	return mint1 < maxt2 && mint2 < maxt1
+}
+
+func getBackfillTSDBRanges(tsdbName string) (int64, int64, error) {
```
pracucci, line 449: You could simplify this a lot (and make it even stronger) using a single regular expression to extract all data.

codesome: Thanks for the regex idea. I was facing some issues with time.Parse. Regex should be simpler.

pkg/ingester/metrics.go
```diff
@@ -203,6 +207,17 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
 		}, []string{"user"})
 	}
 
+	if !createMetricsConflictingWithTSDB {
+		m.numUsersWithBackfillTSDBs = promauto.With(r).NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_ingester_tsdb_users_with_backfill_tsdb",
```
pracucci, line 212: I would suggest `cortex_ingester_backfill_tsdb_tenants`, because:
1. We're trying to move from "user" to "tenant" naming (will take ages)
2. We could prefix the backfill metrics with `cortex_ingester_backfill_tsdb`

pkg/ingester/metrics.go
```diff
@@ -203,6 +207,17 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
...
+			Help: "Total number of users with backfill TSDBs",
+		})
+		m.numBackfillTSDBsPerUser = promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cortex_ingester_tsdb_backfill_tsdb_per_user",
```
pracucci, line 216: I would suggest `cortex_ingester_backfill_tsdb_opened`. The dimension "user" shouldn't be part of the metric name, so I would not mention "per_user" in the metric name.

pkg/ingester/metrics.go
```diff
@@ -203,6 +207,17 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
...
+			Help: "Total number of backfill TSDBs per user",
```
pracucci, line 217:
```suggestion
			Help: "Total number of backfill TSDBs opened per tenant.",
```

pkg/ingester/metrics.go
```diff
@@ -203,6 +207,17 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
 		}, []string{"user"})
 	}
 
+	if !createMetricsConflictingWithTSDB {
+		m.numUsersWithBackfillTSDBs = promauto.With(r).NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_ingester_tsdb_users_with_backfill_tsdb",
+			Help: "Total number of users with backfill TSDBs",
```
pracucci, line 213:
```suggestion
			Help: "Total number of tenants with a backfill TSDB opened.",
```

pkg/ingester/ingester_v2.go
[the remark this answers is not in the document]
pracucci: Should 1h be hardcoded or should be based on the actual configured block range period? We allow to customise the block duration.

pkg/ingester/ingester_v2.go
```diff
@@ -547,12 +577,24 @@ func (i *Ingester) v2Query(ctx context.Context, req *client.QueryRequest) (*clie
...
 	// It's not required to return sorted series because series are sorted by the Cortex querier.
-	ss := q.Select(false, nil, matchers...)
+	ss := q.Select(len(backfillSSs) > 0, nil, matchers...)
 	if ss.Err() != nil {
 		return nil, ss.Err()
 	}
 
+	if len(backfillSSs) > 0 {
+		// TODO(codesome): If any TSDB in backfill buckets were overlapping
+		// with main TSDB, then use tsdb.NewMergedVerticalSeriesSet
```
pracucci, line 594: What's the impact of never using `NewMergedVerticalSeriesSet`?

pkg/ingester/ingester_v2.go
```diff
@@ -910,32 +964,52 @@ func (i *Ingester) getOrCreateTSDB(userID string, force bool) (*userTSDB, error)
...
 	blockRanges := i.cfg.BlocksStorageConfig.TSDB.BlockRanges.ToMilliseconds()
 
+	userDB, err := i.createNewTSDB(
+		userID,
+		i.cfg.BlocksStorageConfig.TSDB.BlocksDir(userID),
+		blockRanges[0],
+		blockRanges[len(blockRanges)-1],
+		tsdbPromReg,
+	)
+	if err == nil {
```
pracucci, line 977: I would reverse the `if`, using a guard:

```
if err != nil {
    return nil, err
}
```

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,499 @@
...
+	// dbs[0] is the one which would be overlapping with the main TSDB. "newer" TSDB.
+	// dbs[1] is the older TSDB among the two here. "older" TSDB.
+	// If the dbs[1] goes beyond the backfill age, it will be compacted
+	// and shipped and dbs[0] is moved to dbs[1]. During this process,
+	// there is only 1 TSDB ingesting old data, so there can be gaps till the compaction
+	// and shipping is going on.
+	// TODO(codesome): Avoid gaps by using a separate queue to compact and ship blocks.
+	dbs [2]*backfillTSDBWrapper
+
+	mtx sync.RWMutex
```
pracucci, line 40: Given we also have `compactMtx`, I would give this mutex a more explicit name.

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,499 @@
...
+	return err
+}
+
+func (a *backfillAppender) getAppender(s client.Sample) (storage.Appender, *userTSDB, error) {
+	if !a.backfillTSDB.isWithinBackfillAge(s.TimestampMs) {
+		return nil, nil, storage.ErrOutOfBounds
+	}
+
+	// Fast path to not take the lock.
+	if a.newerAppender != nil && s.TimestampMs >= a.newerAppender.start && s.TimestampMs < a.newerAppender.end {
```
pracucci, line 133: Is it safe to access `a.newerAppender` and `a.olderAppender`, given their references may be replaced any time (and when they're replaced, a lock is used)?

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,499 @@
...
+
+	// We could use RLock here to check the TSDBs, but because we create new TSDB if none exists,
+	// it's easier and cleaner to handle races by taking the write lock.
+	// TODO(codesome): If we see performance issues, explore ways to read lock here.
+	// TODO(codesome): IMPORTANT: fix the issue where the compaction moves the TSDB around between 2 appends to avoid using wrong appenders in older and newer spots.
+	a.backfillTSDB.mtx.Lock()
+	defer a.backfillTSDB.mtx.Unlock()
+
+	// Check if we already have TSDB created and use it.
+	if a.backfillTSDB.dbs[0] != nil && s.TimestampMs >= a.backfillTSDB.dbs[0].start && s.TimestampMs < a.backfillTSDB.dbs[0].end {
```
pracucci, line 147: You could move this long line logic into a simple function on `backfillTSDBWrapper.within(s.TimestampMs)`

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,499 @@
...
+// add requires the samples to be within the backfill age.
+// Note: This method should not be called concurrently.
+func (a *backfillAppender) add(la []client.LabelAdapter, s client.Sample) (err error) {
+	app, db, err := a.getAppender(s)
+	if err != nil {
+		return err
+	}
+	startAppend := time.Now()
+
+	cachedRef, cachedRefExists := db.refCache.Ref(startAppend, client.FromLabelAdaptersToLabels(la))
```
pracucci, line 108: To my understanding, `db.refCache` is shared across all TSDBs (main + 2 backfill) but is it safe to be shared?

pkg/ingester/ingester_v2.go
codesome: There could be overlapping chunks between the main TSDB and backfill TSDBs, and not vertically merging them might lead to wrong results. I think with recent refactoring in Prometheus, both the logic were merged into one. I will check that and replace stuff accordingly.

pkg/ingester/ingester_v2_backfill.go
codesome: The backfill TSDBs themselves are `userTSDB` and have their own `refCache` (each backfill TSDB getting one). They don't have any access to the main TSDB.

pkg/ingester/ingester_v2_backfill.go
codesome: Yes, it's safe. Note that it is 1 appender per `v2Push` call and not shared across multiple push. The idea here is that the goroutine which is getting the appender will use it sequentially (which is the case in `v2Push`, each sample is iterated sequentially). So, while the DBs can be started in parallel which is protected by the lock, it is safe to access `newerAppender` and `olderAppender` without lock as they are the part of the appender and not the backfillTSDB.

pkg/ingester/metrics.go
codesome: I will keep it `cortex_ingester_backfill_tsdb_open` as this is a gauge (current open TSDBs)

pkg/ingester/ingester_v2.go
codesome: > I think with recent refactoring in Prometheus, both the logic were merged into one.

That was only for chunks, specific to compaction. So we need to manually decide here whether to use vertical merging or not.

pracucci: If not using the vertical compaction could lead to incorrect results, then sure we should use it. I would always use it and we should be fine.

codesome: Right, in case of backfill TSDBs it sounds sane to use vertical always.

pkg/ingester/ingester_v2_backfill.go
```diff
@@ -0,0 +1,477 @@
...
+	// Fast path to not take the lock.
+	if a.newerAppender != nil && a.newerAppender.within(s.TimestampMs) {
+		return a.newerAppender.app, a.newerAppender.db, nil
+	} else if a.olderAppender != nil && a.olderAppender.within(s.TimestampMs) {
+		return a.olderAppender.app, a.olderAppender.db, nil
+	}
+
+	// We could use RLock here to check the TSDBs, but because we create new TSDB if none exists,
+	// it's easier and cleaner to handle races by taking the write lock.
+	// TODO(codesome): IMPORTANT: fix the issue where the compaction moves the TSDB around between 2 appends to avoid using wrong appenders in older and newer spots.
```
codesome, line 151: Regarding this `// TODO(codesome): IMPORTANT: fix the issue where the compaction moves the TSDB around between 2 appends to avoid using wrong appenders in older and newer spots.`, I think _this is fine_ with a check for the appender being `nil` to not overwrite it and lose data.

Reasons:
1. If the older TSDB moved (i.e. outside of backfill age), that means the sample is not eligible anyway for ingestion as it went out of backfill age. So it is fine to fail the append.
2. If the newer TSDB moved to the older spot, the new appender will be for the same TSDB as before and wont be an issue ingesting to it if the sample falls in that TSDB.
3. If the newer TSDB moved to the older spot and the new sample incoming belongs to the newer TSDB to be created, then we fail that append here which will be retried later.

So, no data loss because of the re-try of Prometheus remote-write. This is a tradeoff for a whole lot complexity it would bring to handle moving appenders with TSDBs. If it turns out to be a problem, we can iterate on it later.

codesome: I will update the comment (with `TODO`) explaining the case and that it is fine.

pracucci:
_Still super valid. Offline discussions are on-going._

codesome:
Nope nope, still valid

pracucci:
_Closing for now, since we're currently considering a different solution._

dilipkk-foyernet:
@pracucci can you please share any information about the new solution for backfilling the data which older than 1h into cortex?
I was wondering if (this)[https://github.com/grafana/cortex-tools/blob/main/pkg/backfill/backfill.go] will work or not. Please let me know your opinion on this,

Thanks,
Dilip

Issue #3014, opened 2020-08-11
Expose dynamodb backoff in config file.


tomwilkie:
https://github.com/cortexproject/cortex/blob/master/pkg/chunk/aws/dynamodb_storage_client.go#L61

pracucci:
Done in #3026

PR #3026, opened 2020-08-12
Expose dynamodb backoff config

jdbaldry:
Signed-off-by: Jack Baldry <jack.baldry@grafana.com>

**What this PR does**:
Exposes the `storage.aws.dynamodb.backoff_config` configuration file so that administrators may configure the DynamoDB backoff without the need for flags.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
Ah, sorry, had this openened in a tab for a while, haven't realized it's a draft now.

pracucci:
Fixes #3014

pracucci:
@jdbaldry Could you add a `CHANGELOG` entry please? Should be `[ENHANCEMENT]`.

jdbaldry:
Sorry for putting this in draft and then forgetting about it. I had a thought jump into my head that possibly only a subset of the backoffConfig was exposed via flags and intended to double check that I wasn't inverting the mismatch. I just double checked and they all match up so I'll remove the draft and add a `CHANGELOG` entry.

PR #3027, opened 2020-08-12
Update getting-started-chunks.md

brianpham:
**What this PR does**:
Add missing instructions to match setting up prometheus

No issue created

### DIFF
--- a/docs/getting-started/getting-started-chunks.md
+++ b/docs/getting-started/getting-started-chunks.md
@@ -20,7 +20,10 @@ Cortex can also make use of external memcacheds for caching and although these a
 
 For simplicity and to get started, we'll run it as a [single process](../configuration/single-process-config.md) with no dependencies:
 
+Clone and build prometheus
 ```sh
+$ git clone https://github.com/cortexproject/cortex.git
+$ cd cortex
 $ go build ./cmd/cortex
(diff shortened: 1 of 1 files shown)


brianpham:
Not sure if I need to update CHANGELOG for this change?

docs/getting-started/getting-started-chunks.md
```diff
@@ -21,6 +21,8 @@ Cortex can also make use of external memcacheds for caching and although these a
 For simplicity and to get started, we'll run it as a [single process](../configuration/single-process-config.md) with no dependencies:
 
 ```sh
+$ git clone git@github.com:cortexproject/cortex.git
```
pstibrany, line 24: I would suggest to use HTTP-based clone instead, as it doesn't depend on `ssh` and can use proxies.

```suggestion
$ git clone https://github.com/cortexproject/cortex.git
```

brianpham: That make sense. Fixing.

pstibrany:
Oh, I just realized it says “Prometheus“ instead of Cortex! My bad. Do you want to send PR to fix it? (If not, I will on Monday)

brianpham:
Ahh silly mistake. Just created a new PR.

PR #3043, opened 2020-08-14
Close spanlogger

joe-elliott:
**What this PR does**:
Correctly closes spanlogger in a handful of places where it was not closed.


pracucci:
> We could use a tool to check for this automatically.

@pstibrany Are you aware of any tool we could use to do it?

pstibrany:
> > We could use a tool to check for this automatically.
> 
> @pstibrany Are you aware of any tool we could use to do it?

Unfortunately not off the top of my head, would require some research.

PR #3046, opened 2020-08-15
Simpler instrumentation for non-batched memcached

bboreham:
Previously it would record a `Memcached.Get` operation with a nested `Memcached.GetMulti`; since they are in one-to-one correspondance this is pointless.

Skip the outer data collection for the non-batched case, and rename the batched one to `Memcached.GetBatched` which is more descriptive.

These names are used for metric labels and for tracing span names.

**Checklist**
- NA Tests updated
- NA Documentation added
- [x] `CHANGELOG.md` updated


pracucci (approved):
LGTM

PR #3050, opened 2020-08-17
Document all HTTP API endpoints

pracucci:
**What this PR does**:
I wanted to remote the "work in progress" from the API documentation page, so I did rework this part of the doc. In details:

1. Renamed "Cortex APIs" to "HTTP API", because:
  - We document only HTTP API
  - It's not necessarily "Cortex API", but also include "Prometheus API" exposed by Cortex or Alertmanager
2. I kept the existing documentation, but reworked and uniformed it
3. I obviously added mentioning to all missing endpoints (at least, all I found)
  - Whenever an API is Prometheus-compatible, I linked to Prometheus doc. I believe we shouldn't copy-paste Prometheus doc here, but just reference it, as further signal we're 100% compatible. When we're not compatible, I mentioned it

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added


docs/api/_index.md
```diff
@@ -0,0 +1,749 @@
...
+| [Configuration](#configuration) | _All services_ | `GET /config` |
+| [Services status](#services-status) | _All services_ | `GET /services` |
+| [Readiness probe](#readiness-probe) | _All services_ | `GET /ready` |
+| [Remote write](#remote-write) | Distributor | `POST /api/v1/push` |
+| [Tenants stats](#tenants-stats) | Distributor | `GET /distributor/all_user_stats` |
+| [HA tracker status](#ha-tracker-status) | Distributor | `GET /distributor/ha_tracker` |
+| [Flush chunks / blocks](#flush-chunks--blocks) | Ingester | `GET|POST /ingester/flush` |
+| [Shutdown](#shutdown) | Ingester | `GET|POST /ingester/shutdown` |
+| [Ingesters ring status](#ingesters-ring-status) | Ingester | `GET /ingester/ring` |
+| [Instant query](#instant-query) | Querier, Query-frontend | `GET|POST <prometheus-http-prefix>/api/v1/query` |
```
Wing924, line 30: `GET|POST <prometheus-http-prefix>/api/v1/query` is broken in table.

docs/api/_index.md
```diff
@@ -0,0 +1,749 @@
...
+| API | Service | Endpoint |
+| --- | ------- | -------- |
+| [Index page](#index-page) | _All services_ | `GET /` |
+| [Configuration](#configuration) | _All services_ | `GET /config` |
+| [Services status](#services-status) | _All services_ | `GET /services` |
+| [Readiness probe](#readiness-probe) | _All services_ | `GET /ready` |
+| [Remote write](#remote-write) | Distributor | `POST /api/v1/push` |
+| [Tenants stats](#tenants-stats) | Distributor | `GET /distributor/all_user_stats` |
+| [HA tracker status](#ha-tracker-status) | Distributor | `GET /distributor/ha_tracker` |
+| [Flush chunks / blocks](#flush-chunks--blocks) | Ingester | `GET|POST /ingester/flush` |
```
jtlisi, line 27: Screen Shot 2020-08-19 at 9 22 52 AM

I'm not sure what is causing it. However, I notice weird formatting in the rich text diff starting at this point.

weeco: I think this is because the pipe character is not escaped as Wing pointed out below

docs/api/_index.md
weeco, line 30:
```suggestion
| [Instant query](#instant-query) | Querier, Query-frontend | `GET\|POST <prometheus-http-prefix>/api/v1/query` |
```

docs/api/_index.md
pracucci: This looks a to a GitHub issue. If I build the website (I previewed it locally) it renders correctly. On the contrary, if I escape it with `\|` the markdown parser used by the website builder keeps it as `\|` which is not what we want.

docs/api/_index.md
pracucci: See https://github.com/cortexproject/cortex/pull/3050/files#r473089054

pracucci: See https://github.com/cortexproject/cortex/pull/3050/files#r473089054

docs/api/_index.md
pracucci: Example of website rendering:

Screenshot 2020-08-19 at 16 51 45

Wing924: How about change `|` to `, `?

pracucci: I would do it only if it turns out that the GitHub markdown rendering breaks once merged. I'm not sure if it's just the markdown diff preview we see in the PR.

docs/api/_index.md
```diff
@@ -0,0 +1,749 @@
...
+
+_For more information, please check out Prometheus [Remote storage integrations](https://prometheus.io/docs/prometheus/latest/storage/#remote-storage-integrations)._
+
+
+## Querier
+
+### Get tenant ingestion stats
+
+```
+GET /api/v1/user_stats
```
pstibrany, line 292: Interesting, I had no idea about this :)

docs/api/_index.md
pracucci: Yes, I had to use the `,` instead of `|`. Thanks for the suggestion!
https://github.com/cortexproject/cortex/pull/3081

PR #3053, opened 2020-08-18
Add validation of Alertmanager configs to API

gouthamve:
We current reply with a 200 even if bad config is applied. This
fixes that.


gotjosh (approved):
Awesome! LGTM.

pkg/alertmanager/api_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"testing"
+
+	"github.com/cortexproject/cortex/pkg/alertmanager/alerts"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAMConfigValidation(t *testing.T) {
```
gotjosh, line 11: One last nit (but maybe make sure we create an issue for it?) - I'm a big fan of tests talk to the handler directly over using [`httptest`](https://golang.org/pkg/net/http/httptest/#NewRequest). They can help quickly illustrate what the API expects and returns in terms of format.

I wouldn't mind if we just create the base structure with 1 fail / 1 pass but serves as a baseline to extend with other cases as we make changes here.

pkg/alertmanager/api.go
```diff
@@ -89,7 +94,13 @@ func (am *MultitenantAlertmanager) SetUserConfig(w http.ResponseWriter, r *http.
 		return
 	}
 
-	cfgDesc, _ := alerts.ToProto(cfg.AlertmanagerConfig, cfg.TemplateFiles, userID)
+	cfgDesc := alerts.ToProto(cfg.AlertmanagerConfig, cfg.TemplateFiles, userID)
+	if err := validateUserConfig(cfgDesc); err != nil {
+		level.Error(logger).Log("msg", errValidatingConfig, "err", err.Error())
```
pracucci, line 99: I don't think it's an error from the cluster operator view. I would say at most a warning. Realistically, there's probably nothing the operator could do, when Cortex is running in a multi-tenant env.

pkg/alertmanager/api.go
```diff
@@ -118,3 +129,48 @@ func (am *MultitenantAlertmanager) DeleteUserConfig(w http.ResponseWriter, r *ht
...
+
+	for _, tmpl := range cfg.Templates {
+		_, err := createTemplatesFile(tmpDir, cfg.User, tmpl.Filename, tmpl.Body)
+		if err != nil {
+			return err
+		}
+	}
+
+	templateFiles := make([]string, len(amCfg.Templates))
+	if len(amCfg.Templates) > 0 {
```
pracucci, line 159: You can remove this `if`. The next `for` will just not iterate over `amCfg.Templates` if it's empty.

pracucci (approved):
Good job @gouthamve! I left few nits I would be glad if could take a look, but consider it already approved from me.

pkg/alertmanager/api_test.go
```diff
@@ -62,13 +68,42 @@ receivers:
...
+		store:  noopAlertStore{},
+		logger: util.Logger,
+	}
 	for _, tc := range testCases {
-		cfgDesc := alerts.ToProto(tc.cfg.AlertmanagerConfig, tc.cfg.TemplateFiles, "fake")
-		err := validateUserConfig(cfgDesc)
+		payload, err := yaml.Marshal(&tc.cfg)
+		require.NoError(t, err)
+
+		req := httptest.NewRequest("GET", "http://foooo", bytes.NewReader(payload))
```
gotjosh, line 79:
```suggestion
		req := httptest.NewRequest(http.MethodPost, "http://alertmanager/api/v1/alerts", bytes.NewReader(payload))
```

gotjosh: Apologies as I understand this is pretty minimal but I find it helps illustrate the case much quicker.

pkg/alertmanager/api_test.go
```diff
@@ -0,0 +1,109 @@
...
+		ctx := context.Background()
+		ctx = user.InjectOrgID(ctx, "testing")
+		require.NoError(t, user.InjectOrgIDIntoHTTPRequest(ctx, req))
+		w := httptest.NewRecorder()
+		am.SetUserConfig(w, req)
+
+		resp := w.Result()
+		if tc.hasErr {
+			require.Equal(t, 400, resp.StatusCode)
+		} else {
```
gotjosh, line 89: I would check the message as well to ensure the format looks 👍  as this is what users would see.

gotjosh (approved):
My comments are nits and entirely optional. LGTM from here, thanks for taking care of my feedback!

PR #3054, opened 2020-08-18
Support checking cache behavior based on query request

### LINKED ISSUE #3049: feature: allow caller to say that query is not worth caching
We run weekly reports for customers, which make several sizeable queries over 1 week of data, which are extremely unlikely to be run again until next week, and then on different data.

These queries cause a lot of cache writes, and because Cortex caches everything, they may displace data in the cache that will be needed again.

So I would like some kind of "cache control header" to state this query is not worth caching.

yeya24:

**What this PR does**:

1. Add a `ShouldCacheFn func(r Request) bool` fucntion to cache middleware. This is useful for Thanos to extend the cache checking behavior based on Thanos requests and params.
2. Add a `noCache` field in query range protobuf. This value is set when cache control header value set to `no-store` in the http request.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/querier/queryrange/query_range.go
```diff
@@ -72,6 +72,8 @@ type Request interface {
 	GetStep() int64
 	// GetQuery returns the query of the request.
 	GetQuery() string
+	// GetNoCache returns the cache decision.
```
yeya24, line 75: Umm, maybe this comment is not very clear. Any suggestions are welcome :smile:

pkg/querier/queryrange/results_cache.go
```diff
@@ -161,6 +166,10 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
 		return nil, httpgrpc.Errorf(http.StatusBadRequest, err.Error())
 	}
 
+	if s.shouldCache != nil && !s.shouldCache(r) {
```
bwplotka, line 169: I would maybe create a separate quick middleware (:

You want to implement 

```go
// Middleware is a higher order Handler.
type Middleware interface {
	Wrap(Handler) Handler
}

```

yeya24: I checked the code and I found that this way makes the implementation much more complicated. Is this change worthwhile? And I think using the function in this pr is already very clear, is it really necessary to always follow the same pattern?

yeya24:
``` go
type cacheChecker struct {
	resultsCache queryrange.Handler

	// cacheNextHandler is the next handler of results cache
	cacheNextHandler queryrange.Handler
}

func newCacheChecker(cacheNextHandler queryrange.Handler) queryrange.Middleware {
	return queryrange.MiddlewareFunc(func(next queryrange.Handler) queryrange.Handler {
		return &cacheChecker{
			resultsCache:     next,
			cacheNextHandler: cacheNextHandler,
		}
	})
}

func (c *cacheChecker) Do(ctx context.Context, r queryrange.Request) (queryrange.Response, error) {
	if r.GetNoCache() {
		return c.cacheNextHandler.Do(ctx, r)
	}

	return c.resultsCache.Do(ctx, r)
}
```

Above is the example implementation for this middleware. But when initializing this middleware we need to explicitly pass the next handler of results cache. However, this is not doable I think because we can only get `Middleware` type instead of `Handler`. https://github.com/thanos-io/thanos/blob/master/pkg/queryfrontend/roundtrip.go#L93.

Sorry for the long comment. If I missed anything, please let me know!

bwplotka: There is some boilerplate to it, yes so for simple if, let's try your solution then, thanks for trying!

pracucci: I would have suggested the middleware too, but I agree to keep it simple if the middleware is complicated, so 👍

pkg/querier/queryrange/queryrange.proto
```diff
@@ -19,6 +19,7 @@ message PrometheusRequest {
   int64 step = 4;
   google.protobuf.Duration timeout = 5 [(gogoproto.stdduration) = true, (gogoproto.nullable) = false];
   string query = 6;
+  bool noCache = 7;
```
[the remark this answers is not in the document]
pracucci: I would make this a `CacheDirectives` struct to keep it more extensible (I left another comment about this).

yeya24:
@pracucci  @pstibrany Thanks for the detailed and careful reviews! I will update this pr soon.

pstibrany:
> @pracucci @pstibrany Thanks for the detailed and careful reviews! I will update this pr soon.

We gave you some conflicting suggestions :) I don't have a strong opinion on bool vs struct. I personally prefer to keep it simple and avoid doing design decisions before the need is there, but feel free to pick one (or not).

pkg/querier/queryrange/query_range.go
```diff
@@ -205,6 +207,14 @@ func (prometheusCodec) DecodeRequest(_ context.Context, r *http.Request) (Reques
 
 	result.Query = r.FormValue("query")
 	result.Path = r.URL.Path
+
+	for _, value := range r.Header.Values(cacheControlHeader) {
+		if value == noStoreValue {
```
pracucci, line 212: You should use the `string.Contains()` because of the reason explain in the previous comment. Iterating on `r.HeaderValues()` doesn't cover the case the header value is something like `no-cache,no-store`.

pracucci (approved):
LGTM (with a nit). Thanks!

pkg/querier/queryrange/query_range.go
yeya24: Okay, sorry I missed that. Umm, what if the request sets `Cache-Control=no-cache` and `Cache-Control=no-store` at the same time?  `r.Header.Get()` only returns the first value.

yeya24: I think I got the answer. Just want to double check. Seems it is not necessary and multi-value header is only valid in comma-separated format. Let me use `string.Contains()`. Thanks!

Screenshot from 2020-08-28 12-51-56

yeya24: Sorry, I missed that! Thanks for pointing this out!

pkg/querier/queryrange/queryrange.proto
```diff
@@ -59,3 +60,7 @@ message Extent  {
   string trace_id = 4 [(gogoproto.jsontag) = "-"];
   google.protobuf.Any response = 5 [(gogoproto.jsontag) = "response"];
 }
+
+message CacheDirectives {
+  bool disabled = 1;
```
pstibrany, line 65: Sorry for being annoying, but please let's not use "negative" case for the name, unless there is a good reason to (eg. need for default value in Go). I would suggest to be little more descriptive and call it `cachingEnabled`.

(For discussion, not a suggestion: Since this message is called `CachingDirectives`, should flag inside be named after "no-store" directive? Personally I wouldn't go this far, but I would also just call it `CachingOptions`.)

yeya24: Hi @pstibrany. It is better to use negative here since we need a default value `false`, https://github.com/cortexproject/cortex/pull/3054#discussion_r479318496.

`CachingOptions` seems a better name! Thanks, let me update the pr.

pstibrany: > Hi @pstibrany. It is better to use negative here since we need a default value `false`, [#3054 (comment)](https://github.com/cortexproject/cortex/pull/3054#discussion_r479318496).

Sorry, my mistake. I remember having exactly this discussion with Marco, but forgot it was for this PR :)

pstibrany (approved):

pstibrany:
Thank you for your work!

PR #3056, opened 2020-08-19
Add integration label to notification metrics

gouthamve:
Also, add a metric for when creation of rules manager would fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 * [ENHANCEMENT] Expose `storage.aws.dynamodb.backoff_config` configuration file field. #3026
 * [ENHANCEMENT] Added `cortex_request_message_bytes` and `cortex_response_message_bytes` histograms to track received and sent gRPC message and HTTP request/response sizes. Added `cortex_inflight_requests` gauge to track number of inflight gRPC and HTTP requests. #3064
 * [ENHANCEMENT] Add config validation to the experimental Alertmanager API. Invalid configs are no longer accepted. #3053
+* [ENHANCEMENT] Add "integration" as a label for `cortex_alertmanager_notifications_total` and `cortex_alertmanager_notifications_failed_total` metrics. #3056
+* [ENHANCEMENT] Add `cortex_ruler_config_last_reload_successful` and `cortex_ruler_config_last_reload_successful_seconds` to check status of users rule manager. #3056
 * [BUGFIX] Query-frontend: Fixed rounding for incoming query timestamps, to be 100% Prometheus compatible. #2990
 * [BUGFIX] Querier: query /series from ingesters regardless the `-querier.query-ingesters-within` setting. #3035
 * [BUGFIX] Experimental blocks storage: Ingester is less likely to hit gRPC message size limit when streaming data to queriers. #3015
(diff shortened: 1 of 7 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -135,7 +135,7 @@ func newMultitenantAlertmanagerMetrics(reg prometheus.Registerer) *multitenantAl
 	m.invalidConfig = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
 		Name:      "alertmanager_config_invalid",
-		Help:      "Whenever the Alertmanager config is invalid for a user.",
+		Help:      "Boolean set to 1 whenever the Alertmanager config is invalid for a user.",
```
gotjosh, line 138: 👍

pkg/ruler/manager.go
```diff
@@ -20,6 +20,24 @@ import (
...
+		Namespace: "cortex",
+		Name:      "ruler_config_updates_total",
+		Help:      "Total number of config updates triggered by a user",
+	}, []string{"user"})
+	configUpdateFailuresTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "cortex",
+		Name:      "ruler_config_update_failures_total",
+		Help:      "Total number of config update failures triggered by a user",
+	}, []string{"user", "reason"})
+	userManagerFailed = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
gotjosh, line 34: I had in mind that we could use the prometheus-style `last_reload_successful` type of thing.

https://github.com/prometheus/prometheus/blob/9438bf735a1e1e968f9403c2a1f606fde9a068a9/cmd/prometheus/main.go#L73-L80

I get the feeling that users might be a bit more familiar with this.

gotjosh (approved):
Great catch LGTM! I'm assuming this was introduced upstream but given we have custom metric struct we need to manually update to be on par as I don't see where exactly is the integration being set.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 * [ENHANCEMENT] Query-tee: added a small tolerance to floating point sample values comparison. #2994
 * [ENHANCEMENT] Query-tee: add support for doing a passthrough of requests to preferred backend for unregistered routes #3018
 * [ENHANCEMENT] Expose `storage.aws.dynamodb.backoff_config` configuration file field. #3026
+* [ENHANCEMENT] Add "integration" as a label for `cortex_alertmanager_notifications_total/failed_total` metrics. #3056
```
pracucci, line 17: I generally tend to suggest to be explicit:

```suggestion
* [ENHANCEMENT] Add "integration" as a label for `cortex_alertmanager_notifications_total` and `cortex_alertmanager_notifications_failed_total` metrics. #3056
```

pkg/ruler/manager.go
pracucci: [not a blocker to me] I would also be up for a metric with the timestamp. I believe you can alert similarly to the boolean, but it's also more expressive cause you can easily compute since how long it's broken.

pracucci, line 34: Instead of having global metrics, can we move them to DefaultMultiTenantManager? We already have managersTotal there.

pkg/ruler/manager.go
```diff
@@ -20,6 +20,24 @@ import (
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
+var (
+	configUpdatesTotal = promauto.NewCounterVec(prometheus.CounterOpts{
```
pracucci, line 24: I would move back `configUpdatesTotal` and `configUpdateFailuresTotal` to the ruler source file because they belong to the ruler, not the manager.

pkg/ruler/ruler.go
```diff
@@ -38,16 +38,6 @@ var (
 		Name:      "ruler_ring_check_errors_total",
 		Help:      "Number of errors that have occurred when checking the ring for ownership",
 	})
-	configUpdatesTotal = promauto.NewCounterVec(prometheus.CounterOpts{
-		Namespace: "cortex",
-		Name:      "ruler_config_updates_total",
-		Help:      "Total number of config updates triggered by a user",
-	}, []string{"user"})
-	configUpdateFailuresTotal = promauto.NewCounterVec(prometheus.CounterOpts{
```
pracucci, line 46: This metric has been removed (replaced by `cortex_ruler_config_last_reload_successful`) but CHANGELOG doesn't mention it. Could you mention it there, please?

pracucci (approved):
LGTM (with a nit about the CHANGELOG entry). Thanks!

PR #3060, opened 2020-08-20
Do not return error from Ring.Subring()

pracucci:
**What this PR does**:
While working on the shuffle sharding in the store-gateway, I've realised it's quite annoying that the `Ring.Subring()` can return an error. My take is that if the subring can't be constructed (for any reason) we shouldn't fail, but we should fallback to the full ring. In this PR I'm proposing it.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany (approved):
👍

PR #3065, opened 2020-08-20
Fixes the registration of the Alertmanager alert receiving API metrics

gotjosh:
**What this PR does**:

Fixes the registration of the Alertmanager API metrics for receiving alerts. Additionally, it removes a couple of TODOs I noticed along the way - they should be a no-op.

**Checklist**
- [x] Tests updated
- [x] Documentation added


pracucci (approved):
Oh yes, very good! I'm familiar with the problem (I submitted the PR https://github.com/prometheus/alertmanager/pull/2200) and your changes LGTM. Thanks!

gouthamve (approved):

gouthamve:
We need to make sure that two instances of AM are not created, but can't see that anywhere in the code, so we're good!

gotjosh:
> We need to make sure that two instances of AM are not created

Even if that would happen we only ever keep a singleton of the metrics.

PR #3067, opened 2020-08-21
Dont skip WAL logging for validation errors

### LINKED ISSUE #3059: Entire push skips WAL if there's a validation error in any of its samples
If I understand correctly, when using chunk-storage, if a single sample fails validation during ingestion (ie, due to out-of-order or duplicate timestamp), then none of the samples in the request will make it to the WAL either, even though some may have been added to the chunk: https://github.com/cortexproject/cortex/blob/master/pkg/ingester/ingester.go#L435

codesome:
Fixes #3059

I plan to add test for this on Monday.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@
 * [BUGFIX] Fix configuration for TLS server validation, TLS skip verify was hardcoded to true for all TLS configurations and prevented validation of server certificates. #3030
 * [BUGFIX] Fixes the Alertmanager panicking when no `-alertmanager.web.external-url` is provided. #3017
 * [BUGFIX] Fixes the registration of the Alertmanager API metrics `cortex_alertmanager_alerts_received_total` and `cortex_alertmanager_alerts_invalid_total`. #3065
-* [BUGFIX] Ingester: Don't skip WAL logging for validation errors. #3067
+* [BUGFIX] Ingester: If push request contained both valid and invalid samples, valid samples were ingested but not stored to WAL of the chunks storage. This has been fixed. #3067
 
 ## 1.3.0 / 2020-08-21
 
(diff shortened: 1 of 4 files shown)


codesome:
Tests added.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 * [BUGFIX] Fix configuration for TLS server validation, TLS skip verify was hardcoded to true for all TLS configurations and prevented validation of server certificates. #3030
 * [BUGFIX] Fixes the Alertmanager panicking when no `-alertmanager.web.external-url` is provided. #3017
 * [BUGFIX] Fixes the registration of the Alertmanager API metrics `cortex_alertmanager_alerts_received_total` and `cortex_alertmanager_alerts_invalid_total`. #3065
+* [BUGFIX] Ingester: Don't skip WAL logging for validation errors. #3067
```
pstibrany, line 32:
```suggestion
* [BUGFIX] Ingester: if push request contained both valid and invalid samples, valid samples were ingested but not stored to WAL of the chunks storage. This has been fixed. #3067
```

pstibrany (approved):
LGTM, thanks!

pkg/ingester/wal.go
```diff
@@ -186,6 +186,12 @@ func (w *walWrapper) Log(record *WALRecord) error {
 	case <-w.quit:
 		return nil
 	default:
+		if len(record.Series) == 0 && len(record.Samples) == 0 {
```
pracucci, line 189: Why the check is here? To me looks easier to understand if we move it up, right after `if record == nil`.

pracucci (approved):
LGTM (with a nit)

PR #3069, opened 2020-08-21
Store-gateway shuffle sharding

pracucci:
**What this PR does**:
In this PR I'm introducing shuffle sharding support to the store-gateway blocks sharding. The idea is that, when sharding is enabled, you can switch between the default sharding or shuffle sharding. When using shuffle sharding, each store-gateway is eligible to sync/load blocks for a subset of tenants.

Notes:
- I preferred to allow to configure `sharding_strategy` as string instead of a boolean, to keep the config more generic and open to future additional strategies
- The implementation in the store-gateway is slightly more tricky than initially planned because we have to make sure that, when a tenant "leaves" a store-gateway shard, blocks are unloaded and metrics correctly updated

Out of the scope of this PR:
- Fix subring implementation to do real "shuffling" (will be done in a dedicated PR)
- Trigger a re-sync for an user each time its shard size change (will be done in a dedicated PR because need to introduce events on overrides)

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/storegateway/sharding_util.go
```diff
@@ -0,0 +1 @@
+package storegateway
```
pstibrany, line 1: Empty file?

pstibrany (approved):
LGTM. Took me a while to understand why we're faking `Iter` response to be empty, if user is not included, but I believe it is to make `BucketStore` to *unload* blocks for such user. I wonder if we should just drop such `BucketStore` completely (after discovering that it has no blocks anymore)? Or would that complicate it even more (logic here is already quite complex).

pkg/storegateway/sharding_util.go
pracucci: Uh ops. Remoed.

pracucci:
> Took me a while to understand why we're faking `Iter` response to be empty, if user is not included, but I believe it is to make `BucketStore` to _unload_ blocks for such user.

Correct. There's no other way to unload blocks from BucketStore as of today (we could do it, but see below).

> I wonder if we should just drop such `BucketStore` completely (after discovering that it has no blocks anymore)? Or would that complicate it even more (logic here is already quite complex).

Let's say we add a way to `BucketStore` to "close it" unloading all blocks (eg. deleting them from disk). Even if we have it, it's tricky to deal with `BucketStore` and `MetadataFetcher` metrics. An option could be creating metrics structs that we pass to `BucketStore` and `MetadataFetcher` constructors, so that we retain them for each tenant even after unloading and re-use in case the same tenant will be re-added to the shard.

Technically it's doable, but I'm just wondering if it's worth the effort (will the final code be cleaner?).

pkg/ring/ring.go
```diff
@@ -475,3 +478,13 @@ func (r *Ring) GetInstanceState(instanceID string) (IngesterState, error) {
 
 	return instance.GetState(), nil
 }
+
+// Includes returns whether the ring includes the instance with the provided instanceID.
```
ranton256, line 482: Was this meant to say "HasInstance returns" instead of "Includes returns"?

pracucci: Uh ops, thanks for noticing it! I initially named the function `Includes` and then, after renaming, I forgot to update the comment 🤦

PR #3070, opened 2020-08-21
Add warning that schema configuration is not used in blocks storage mode

carlpett:
**What this PR does**:
When using block storage, schema-related configuration is not used. This PR adds a warning if schema has been configured anyway.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -190,6 +190,11 @@ func (c *Config) Validate(log log.Logger) error {
 	if err := c.TableManager.Validate(); err != nil {
 		return errors.Wrap(err, "invalid table_manager config")
 	}
+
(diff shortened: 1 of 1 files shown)


pkg/cortex/cortex.go
```diff
@@ -190,6 +190,11 @@ func (c *Config) Validate(log log.Logger) error {
 	if err := c.TableManager.Validate(); err != nil {
 		return errors.Wrap(err, "invalid table_manager config")
 	}
+
+	if c.Storage.Engine == storage.StorageEngineBlocks && len(c.Schema.Configs) > 0 {
```
pracucci, line 194: It's slightly more tricky because we also need to check for the secondary storage engine (used to migrate between storage engines). I'm going to push a fix for that.

pstibrany (approved):
👍

PR #3073, opened 2020-08-21
Start an Alertmanager when it is requested if we have a fallback config

gotjosh:
Opening this in the form a proposal to see what others think cc: @pracucci @gouthamve @bboreham 

Currently, fallback configurations work in the form of:

- Poll for all config in the object storage
- If the user has an empty config file and we have a fallback config, then use the fallback to start a new Alertmanager

With this PR, we start the am and upload an empty config if the alert manager path is hit (either by accessing the UI or receiving an alert). There are some downsides with this approach:

- If the user uploads a config and accesses the UI before we poll its config will be overwritten (although a check can be added as part of this process?)
- there's no way to know when an Alertmanager has gone "stale" and remove it from memory (although this might not be a concern of cortex?)

Clearly, there's thing missing but I just want to gauge opinions before I invest more effort into this.

### DIFF
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -463,12 +463,44 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok || !userAM.IsActive() {
+	if !ok && am.fallbackConfig != "" {
+		userAM, err = am.alertmanagerFromFallbackConfig(userID)
+		if err != nil {
+			http.Error(w, "Failed to initialize the Alertmanager", http.StatusInternalServerError)
+		}
+	}
+
+	if !userAM.IsActive() {
 		http.Error(w, "the Alertmanager is not configured", http.StatusNotFound)
 		return
 	}
 
 	userAM.mux.ServeHTTP(w, req)
+}
+
+func (am *MultitenantAlertmanager) alertmanagerFromFallbackConfig(userID string) (*Alertmanager, error) {
+	// There's a potential race condition here. If a tenant:
+	// a) uploads a config and b) receives an alert or access the UI
+	// before we're able the start the new manager we'll replace that config with an empty
+	// config.
+	cfg := &UserConfig{AlertmanagerConfig: ""}
+	cfgDesc := alerts.ToProto(cfg.AlertmanagerConfig, cfg.TemplateFiles, userID)
+	err := am.store.SetAlertConfig(context.Background(), cfgDesc)
+	if err != nil {
+		level.Error(am.logger).Log("msg", "unable to store fallback configuration", "user", userID, "err", err.Error())
+		return nil, err
+	}
+
+	// Calling setConfig with an empty configuration will use the fallback config.
+	err = am.setConfig(cfgDesc)
+	if err != nil {
+		return nil, err
+	}
+
+	// We need to check again in case they've had a new alertmanager started by the time we're ready
+	am.alertmanagersMtx.Lock()
+	defer am.alertmanagersMtx.Unlock()
+	return am.alertmanagers[userID], nil
 }
 
 // GetStatusHandler returns the status handler for this multi-tenant
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -463,31 +463,36 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok && am.fallbackConfig != "" {
-		userAM, err = am.alertmanagerFromFallbackConfig(userID)
-		if err != nil {
-			http.Error(w, "Failed to initialize the Alertmanager", http.StatusInternalServerError)
+	if ok {
+		if !userAM.IsActive() {
+			http.Error(w, "the Alertmanager is not configured", http.StatusNotFound)
+			return
 		}
-	}
 
-	if !userAM.IsActive() {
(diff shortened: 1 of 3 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -479,10 +479,46 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok || !userAM.IsActive() {
+	if !ok && am.fallbackConfig != "" {
```
pracucci, line 482: I would move this logic into a dedicate function, used to create the fallback config.

pkg/alertmanager/multitenant.go
```diff
@@ -479,10 +479,46 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+		}
+
+		newAM, err := am.newAlertmanager(userID, transformedConfig)
+		if err != nil {
+			level.Error(am.logger).Log("msg", "failed to start the Alertmanager", "user", userID, "err", err)
+			http.Error(w, "Failed to initialize the Alertmanager", http.StatusInternalServerError)
+		}
+
+		am.alertmanagersMtx.Lock()
+		am.alertmanagers[userID] = newAM
```
pracucci, line 503: There's a possible race condition here. Before overriding the alertmanager config for the user, you should check that it hasn't been already set in the meanwhile (between the previous lock and this lock).

pkg/alertmanager/multitenant.go
```diff
@@ -479,10 +479,46 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	userAM, ok := am.alertmanagers[userID]
 	am.alertmanagersMtx.Unlock()
 
-	if !ok || !userAM.IsActive() {
+	if !ok && am.fallbackConfig != "" {
+		// Give the user the fallback configuration and start its alertmanager
+		userAmConfig, err := amconfig.Load(am.fallbackConfig)
```
pracucci, line 484: Instead of all such duplicated logic, couldn't we just call `setConfig()` with an empty `RawConfig`?

pkg/alertmanager/multitenant.go
```diff
@@ -479,10 +479,46 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+			level.Error(am.logger).Log("msg", "failed to start the Alertmanager", "user", userID, "err", err)
+			http.Error(w, "Failed to initialize the Alertmanager", http.StatusInternalServerError)
+		}
+
+		am.alertmanagersMtx.Lock()
+		am.alertmanagers[userID] = newAM
+		userAM = newAM
+		am.alertmanagersMtx.Unlock()
+
+		go func() {
```
pracucci, line 507: Have you considered doing it synchronously? I would expect it to take few hundred ms and, in my opinion, it's perfectly fine (even if it takes few seconds, given this is a one time thing).

Moreover, if we do it synchronously, from a logic perspective I would do (in order):
1. `am.store.SetAlertConfig()`
2. Update `am.alertmanagers`

pracucci (commented):
Few things:
1. Can we test the changes?
2. Pay attention that the alertmanager integration test is failing (may be related)
3. Should deserve a CHANGELOG entry

pstibrany:
> * If the user uploads a config and accesses the UI before we poll its config will be overwritten (although a check can be added as part of this process?)

Would it be possible to initialize AM with the config without uploading default config to the object store? I think that would be safer than adding extra checks.

gotjosh:
> Would it be possible to initialize AM with the config without uploading default config to the object store? I think that would be safer than adding extra checks.

The problem with that is that the next poll of configs will de-activate the AM.

pkg/alertmanager/multitenant.go
```diff
@@ -463,12 +463,48 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
 	}
 
-	userAM.mux.ServeHTTP(w, req)
+	http.Error(w, "the Alertmanager is not configured", http.StatusNotFound)
+}
+
+func (am *MultitenantAlertmanager) alertmanagerFromFallbackConfig(userID string) (*Alertmanager, error) {
+	// Upload an empty config so that the Alertmanager is no de-activated in the next poll
+	cfg := &UserConfig{AlertmanagerConfig: ""}
+	cfgDesc := alerts.ToProto(cfg.AlertmanagerConfig, cfg.TemplateFiles, userID)
```
pstibrany, line 493: Tiny bit more explicit, `cfg` is otherwise unused:

```suggestion
	// Upload an empty config so that the Alertmanager is not de-activated in the next poll
	cfgDesc := alerts.ToProto("", nil, userID)
```

pkg/alertmanager/multitenant_test.go
```diff
@@ -241,3 +243,69 @@ func TestAlertmanager_ServeHTTP(t *testing.T) {
...
+route:
+  receiver: example-email
+receivers:
+  - name: example-email
+    email_configs:
+    - to: 'youraddress@example.org'
+`
+
+	// Create the Multitenant Alertmanager.
+	reg := prometheus.NewPedanticRegistry()
```
pstibrany, line 273: When not testing metrics, just use `nil` registry.

pstibrany (approved):
LGTM in the current form.

I'm not sure how much overwriting of changes is a concern worth addressing.

pkg/alertmanager/multitenant.go
gotjosh: good shout, done.

gotjosh:
> I'm not sure how much overwriting of changes is a concern worth addressing.

I think it's fine for now, it is likely that it'll happen at most _once_ while we try to spin up the first time and only when you have specified a fallback config.

gouthamve (approved):

PR #3080, opened 2020-08-24
Fail the config validation if an empty node is set at root level

### LINKED ISSUE #3079: Empty querier config block breaks querier
I encountered an issue where the querier component would consistently return 503:s on /api/prom/api/v1/query if given "unusual" configuration. There are no logs emitted that indicate what is going wrong, and adding tracing there isn't much to go on either:
image

The config is found here: https://gist.github.com/carlpett/544bc3f807140e51aca53aa8dd68fd5b. Notably, I had a querier block **without any content**. That is, like this:
```yaml
querier:
query_range:
  [...]
```

For some reason, this appears to break the querier setup such that it fails every request with `tsdb.ErrNotReady` here:
https://github.com/cortexproject/cortex/blob/2cb22f830dfd37b199978ad6191a18f444cdae61/vendor/github.com/prometheus/prometheus/web/api/v1/api.go#L263

I haven't determined exactly _why_ having this config breaks, yet.

(Following Slack discussions [here](https://cloud-native.slack.com/archives/CCYDASBLP/p1598006503011500))

pracucci:
**What this PR does**:
An user reported a tricky issue in #3079. I also tried to assist them on Slack and, just looking at the YAML config, I didn't understand the root cause. The problem is when the YAML config contains an [empty node](https://yaml.org/spec/1.2/spec.html#id2786563), by specification, the respective `struct` is reset to zero values and our default values are lost.

In this PR I'm attempting a workaround to spot when this happens for root level properties and, in this case, I make the config validation fail. I'm not proud of this PR, but I haven't found a nicer way to do it (unless using yaml v3, but would require a major refactoring and I believe it's far beyond the scope of the fix).

Fixes #3079

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/cortex/cortex.go
```diff
@@ -193,6 +199,33 @@ func (c *Config) Validate(log log.Logger) error {
 	return nil
 }
 
+// validateYAMLEmptyNodes ensure that no empty node has been specified in the YAML config file.
+// When an empty node is defined in YAML, the YAML parser sets the whole struct to its zero value
+// and so we loose all default values. It's very difficult to detect this case for the user, so we
```
carlpett, line 204: Nit (and pet peeve)
```suggestion
// and so we lose all default values. It's very difficult to detect this case for the user, so we
```

pkg/cortex/cortex.go
```diff
@@ -193,6 +199,33 @@ func (c *Config) Validate(log log.Logger) error {
...
+	// We expect all structs are the exact same. This check should never fail.
+	if cfgStruct.NumField() != defStruct.NumField() {
+		return errors.New("unable to validate configuration because of mismatching internal config data structure")
+	}
+
+	for i := 0; i < cfgStruct.NumField(); i++ {
+		// If the struct has been reset due to empty YAML value and the zero struct value
+		// doesn't match the default one, then we should warn the user about the issue.
+		if cfgStruct.Field(i).Kind() == reflect.Struct && cfgStruct.Field(i).IsZero() && !defStruct.Field(i).IsZero() {
+			return fmt.Errorf("the %s configuration in YAML has been specified as an empty YAML node", cfgStruct.Type().Field(i).Name)
```
pstibrany, line 222: I'm not sure I would understand this error message. Given that, should there be an option to use my (seemingly wrong) config anyway?

pkg/cortex/cortex.go
```diff
@@ -193,6 +199,33 @@ func (c *Config) Validate(log log.Logger) error {
...
+
+	defStruct := reflect.ValueOf(defaults)
+	cfgStruct := reflect.ValueOf(*c)
+
+	// We expect all structs are the exact same. This check should never fail.
+	if cfgStruct.NumField() != defStruct.NumField() {
+		return errors.New("unable to validate configuration because of mismatching internal config data structure")
+	}
+
+	for i := 0; i < cfgStruct.NumField(); i++ {
```
pstibrany, line 218: I expected to see recursive checks for nested structs. Can this problem happen with nested structs as well?

carlpett (commented):
:clap:
A thought, if you want to avoid the reflection: You could do some sentinel value that should always exist and detect if that is missing instead? Sort of the inverse of how a couple of unmarshalling libs has something similar to `XXX_unrecognized` (protobuf in this case), you could have a default-true bool `DefaultsApplied` or something.
Will lead to some clutter since it has to go in every top level node, of course, so :shrug:

pstibrany (approved):
> I'm not proud of this PR

I expected much worse after reading this. This PR is nice and clean, and looks good to me.

sandeepsukhani:
Is there a way to fix this by setting an expectation for certain values? Like for ints, some of them can't be set to 0 or should have a value above x or something. This way we protect it from having an invalid config i.e everything being set to go default by someone and we don't have to maintain code using `reflect` package.
It is just a thought.

pracucci:
> de using `reflect` package.
> It is just a thought, please feel free to dismiss it.

@sandeepsukhani Yes. This would require to enhance the existing `Validate()` functions and add them when missing. It's quite a long work (but probably worth it), while in this PR I was looking for a generic approach to detect the specific case of an empty YAML node, but I agree the proposed solution is not rock solid.

pkg/cortex/cortex.go
pracucci: This problem can happen with any struct deserialised from JSON. The problem is that the deeper we go, the more likely we hit the case "all zero values" is a valid config. After a which look at the configs at root level, I believe this is not a problem.

pracucci:
> Cou could do some sentinel value that should always exist and detect if that is missing instead? [...] Will lead to some clutter since it has to go in every top level node, of course, so 🤷

@carlpett Technically it's possible, but I see a couple of problems: (1) we'll forget to add to new config structs (2) as you mentioned, we'll end up with some "dirty fields" in the config just to cover this case, and I'm not sure it's what we want.

pracucci:
Agreed offline with @pstibrany to merge it even if not perfect. If anyone has a better idea, I would be happy to re-iterate on this in a following PR.

PR #3081, opened 2020-08-24
Fixed HTTP API documentation

pracucci:
**What this PR does**:
In this PR I'm doing a couple of fixes following up #3050:
1. Documented `/metrics` endpoint
2. The [comment](https://github.com/cortexproject/cortex/pull/3050#discussion_r473025579) did raise the point that the pipe in a table breaks the layout if not correctly escaped. I wanted to see if it was just the GitHub PR preview diff or the GitHub website rendering too. Turned out it's also when the markdown is rendered on GitHub, so in this PR I'm fixing it changing the pipe to comma (we can't use the `\|` because it doesn't render correctly in the Hugo website...)

N/A

**Checklist**
- [ ] Tests updated
- [x Documentation added


pstibrany (approved):
👍 Always nice to see improved docs.

PR #3088, opened 2020-08-26
Explain how gossiping works, how to tune and monitor it.

pstibrany:
**What this PR does**: This PR adds basic description of gossip algorithm, and some tuning options.

### DIFF
--- a/docs/configuration/arguments.md
+++ b/docs/configuration/arguments.md
@@ -186,7 +186,7 @@ prefix these flags with `distributor.ha-tracker.`
 
 #### memberlist
 
-Flags for configuring KV store based on memberlist library (works only for the [hash ring](../architecture.md#the-hash-ring), not for the HA Tracker).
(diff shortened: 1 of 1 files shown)


docs/configuration/arguments.md
```diff
@@ -188,6 +188,25 @@ prefix these flags with `distributor.ha-tracker.`
 
 Flags for configuring KV store based on memberlist library (works only for the [hash ring](../architecture.md#the-hash-ring), not for the HA Tracker).
```
pracucci, line 189: I think this should be moved below, before the listing of all flags (right before `- memberlist.nodename`).

pracucci (approved):
LGTM  (with a nit)

docs/configuration/arguments.md
pstibrany: Left the warning on top, moved "Flags ..." closer to flags.

pracucci: Very good, thanks!

PR #3090, opened 2020-08-26
Added shuffle sharding support to generate a subring

pracucci:
**What this PR does**:
The current `Subring()` implementation was conceived as an experimental shuffle sharding support in Cortex. It works fine in regards to build a subring of N nodes and guaranteeing stability and consistency but unfortunately suffers two issues:
1. It doesn't shuffle instances between tenants
2. It's not zone aware (the resulting subring may be very inbalanced in regards to zones)

In this PR I proposed an alternative `ShuffleShard()` implementation which is inspired by [AWS Route53 infima library](https://github.com/awslabs/route53-infima/blob/master/src/main/java/com/amazonaws/services/route53/infima/SimpleSignatureShuffleSharder.java) and guarantee stability and consistency, while offering nodes shuffling and zone awareness.

**Will be done in subsequent PRs**:
- Introduce a cache for each tenant subring
- Documentation

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/ring/ring.go
```diff
@@ -461,6 +462,101 @@ func (r *Ring) Subring(key uint32, n int) ReadRing {
...
+			// If one more instance has not been found, we can stop looking for
+			// more instances in this zone, because it means the zone has no more
+			// instances which haven't been already selected.
+			if !found {
+				break
+			}
+		}
+	}
+
+	r.mtx.RUnlock()
```
tomwilkie, line 544: Any way we can use `defer`? It'll make the code less brittle to future changes.

pkg/ring/ring.go
```diff
@@ -461,6 +462,101 @@ func (r *Ring) Subring(key uint32, n int) ReadRing {
...
+	seed := int64(binary.BigEndian.Uint64(checksum))
+
+	// Initialise the random generator used to select instances in the ring.
+	random := rand.New(rand.NewSource(seed))
+
+	r.mtx.RLock()
+
+	// We expect the shard size to be divisible by the number of zones, in order to
+	// have nodes balanced across zones. If it's not, we do round up.
+	numInstancesPerZone := int(math.Ceil(float64(size) / float64(len(r.ringZones))))
```
tomwilkie, line 503: Just want to check: in the case where we're not using zones, zone = ingesterID.  Will this algorithm end distributing all users to all ingesters in this case?

pkg/ring/ring.go
pracucci: Sure, done. I put it here to keep it locked for the shortest time, but it's really not an issue, especially considering the upcoming cache.

thorfour:
> It doesn't shuffle instances between tenants

I guess I don't understand this comment. Can you elaborate on what this problem is?

pkg/ring/ring.go
```diff
@@ -461,6 +462,100 @@ func (r *Ring) Subring(key uint32, n int) ReadRing {
...
+//   a different set of instances, with a reduced number of overlapping instances
+//   between two identifiers.
+func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
+	// Nothing to do if the shard size is not smaller then the actual ring.
+	if size <= 0 || r.IngesterCount() <= size {
+		return r
+	}
+
+	// Use the identifier to compute an hash we'll use to seed the random.
+	hasher := md5.New()
```
pstibrany, line 489: We use 64-bit hashing elsewhere (fnv64a), perhaps we should use it here as well, and not throw away half of the hash.

pkg/ring/ring.go
```diff
@@ -461,6 +462,100 @@ func (r *Ring) Subring(key uint32, n int) ReadRing {
 	return sub
 }
 
+// ShuffleShard returns a subring for the provided identifier (eg. a tenant ID)
+// and size (number of instances). The size is expected to be a multiple of the
+// number of zones and the returned subring will contain the same number of
+// instances per zone as far as there are enough registered instances in the ring.
+//
+// The algorithm used to build the subring is a shuffle sharder based on probabilistic
```
pstibrany, line 470: What do you mean by "probabilistic hashing"?

pstibrany (approved):
Great work!

pracucci:
> > It doesn't shuffle instances between tenants
> 
> I guess I don't understand this comment. Can you elaborate on what this problem is?

If we pick nodes following the "next" tokens in the ring (like `Subring()` is working right now) we pick consecutive nodes instead of shuffling them. Think about 2 tenants: one whose initial hash match the node X in the ring, and the second matches the node X+1 in the ring. They will share all nodes minus 1 in the subring. Shuffle sharding is about shuffling nodes to make this case probabilistically unlikely to happen, which is what this PR does.

thorfour:
> > > It doesn't shuffle instances between tenants
> > 
> > 
> > I guess I don't understand this comment. Can you elaborate on what this problem is?
> 
> If we pick nodes following the "next" tokens in the ring (like `Subring()` is working right now) we pick consecutive nodes instead of shuffling them. Think about 2 tenants: one whose initial hash match the node X in the ring, and the second matches the node X+1 in the ring. They will share all nodes minus 1 in the subring. Shuffle sharding is about shuffling nodes to make this case probabilistically unlikely to happen, which is what this PR does.

So is the value of this for really low tenant counts to make it more likely to evenly distribute load? Because the example of 2 tenants overlapping will always be true at large enough tenant counts.

pracucci:
> So is the value of this for really low tenant counts to make it more likely to evenly distribute load?

No. It works very well with a large number of tenants too (assuming a reasonably large number of nodes too, but order of magnitude smaller then tenants).

> Because the example of 2 tenants overlapping will always be true at large enough tenant counts.

It depends on the cluster size. If you have a cluster with 100 ingesters and you randomly pick 6 nodes for each tenant, you end up with 75M possible combinations, and the chances to have overlapping nodes (with shuffle sharding) are:
- Probability of 6 equal instances: 0.00000008%
- Probability of 5 equal instances: 0.00004731%
- Probability of 4 equal instances: 0.00550018%
- Probability of 3 equal instances: 0.22489615%
- Probability of 2 equal instances: 3.83729063%
- Probability of 1 equal instances: 27.62849251%
- Probability of 0 equal instances: 68.30377314%

I created [this spreadsheet](https://docs.google.com/spreadsheets/d/1FXbiWTXi6bdERtamH-IfmpgFq1fNL4GP_KX_yJvbRi4/edit#gid=0) to easily compute it.

For more information, you could take a look at [this](https://aws.amazon.com/blogs/architecture/shuffle-sharding-massive-and-magical-fault-isolation/) and the reference implementation linked in the PR description.

pkg/ring/ring.go
pracucci: > in the case where we're not using zones, zone = ingesterID

It was in the initial implementation but was soon fixed in #2404. When not using zones, the `zone` is an empty string, which works fine here (all ingesters belong to the same zone).

However, I'm wondering if we should  add a boolean flag to the ring to explicitly enable zone awareness, disabled by default instead of just relying on the fact zones are set (think about a rolling update adding zones to ingesters, you may want to rollout all ingesters first and then enable it).

pkg/ring/ring.go
pracucci: Uniqueness is not guaranteed, but it offsers a low collision probability. What we're using here is a [probabilistic data structure](https://towardsdatascience.com/hashes-power-probabilistic-data-structures-d1398d1335c6).

pkg/ring/ring.go
pracucci: I tried using fnv64a:

```
	// Use the identifier to compute an hash we'll use to seed the random.
	seed := client.HashAdd64a(client.HashNew64a(), identifier)
	// Initialise the random generator used to select instances in the ring.
	random := rand.New(rand.NewSource(int64(seed)))
```

But the `TestRing_ShuffleShard_Shuffling` fails. Apparently, md5 introduces a better randomization of seeds. Given there's no performance penalty I revert it.

docs/configuration/config-file-reference.md
```diff
@@ -2714,9 +2718,12 @@ The `limits_config` configures default and per-tenant limits imposed by Cortex s
...
 
-# Per-user subring to shard metrics to ingesters. 0 is disabled.
-# CLI flag: -experimental.distributor.user-subring-size
-[user_subring_size: <int> | default = 0]
+# The default tenant's shard size when the shuffle-sharding strategy is used.
+# Must be set both on ingesters and distributors. When this setting is specified
+# in the per-tenant overrides, a value of 0 disables shuffle sharding for the
+# tenant.
+# CLI flag: -distribution.ingestion-tenant-shard-size
+[ingestion_tenant_shard_size: <int> | default = 0]
```
ranton256, line 2726: I take it based on the similar "store_gateway_tenant_shard_size" flag from https://github.com/cortexproject/cortex/pull/3069/files you are planning to name all the similar flags "<thingy>_tenant_shard_size"?

ranton256: Also, are you planning to take/keep the -experimental  off this when you are ready to submit?

pstibrany:
There is additional work (caching of user rings) available in https://github.com/pstibrany/cortex/tree/fix-shuffle-sharding branch, not yet tested.

docs/configuration/config-file-reference.md
pracucci: > I take it based on the similar "store_gateway_tenant_shard_size" flag from https://github.com/cortexproject/cortex/pull/3069/files you are planning to name all the similar flags "_tenant_shard_size"?

Yes, I do. The reason is that overrides are specified within the same YAML node, so we need a way to clearly differentiate them. What do you think?

pracucci: > Also, are you planning to take/keep the -experimental off this when you are ready to submit?

I would plan to remove the experimental flag once ready to submit, unless you have any concerns. Generally speaking, I think originally adding the experimental prefix to CLI flags was a mistake and shouldn't be done anymore. What do you think?

pkg/ring/ring.go
pracucci: Agree @ranton256. Will do it in a separate PR (since it's unrelated from the work done in this PR).

pstibrany (approved):
Great job!

docs/configuration/config-file-reference.md
ranton256: I think you mean document as experimental but not name the flag experimental? That sounds fine to me.

pracucci: Yes Richard, you're correct (sorry for being unclear). This way, once we're all comfortable marking it stable, it will just be a doc change instead of a config change.

PR #3098, opened 2020-08-26
Make ruler evaluation delay configuration per user.

tomwilkie:

**What this PR does**:

Makes the ruler evaluation delay configuration per user.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


tomwilkie:
What did we say about compatibility and changing the config structs?

gotjosh:
> What did we say about compatibility and changing the config structs?



I think we [need to keep the old flag around for 2 more minor versions](https://cortexmetrics.io/docs/configuration/v1guarantees/#flags-config-and-minor-version-upgrades) but others might know best.

pkg/util/validation/limits.go
```diff
@@ -106,6 +111,8 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.CardinalityLimit, "store.cardinality-limit", 1e5, "Cardinality limit for index queries. This limit is ignored when running the Cortex blocks storage. 0 to disable.")
 	f.DurationVar(&l.MaxCacheFreshness, "frontend.max-cache-freshness", 1*time.Minute, "Most recent allowed cacheable result per-tenant, to prevent caching very recent results that might still be in flux.")
 
+	f.DurationVar(&l.EvaluationDelay, "ruler.evaluation-delay-duration", 0, "Duration to delay the evaluation of rules to ensure they underlying metrics have been pushed to cortex.")
```
gotjosh, line 114:
```suggestion
	f.DurationVar(&l.EvaluationDelay, "ruler.evaluation-delay-duration", 1*time.Minute, "Duration to delay the evaluation of rules to ensure the underlying metrics have been pushed to cortex.")
```

gotjosh: A typo and if we're making this per tenant I reckon `60s` might be "safer" approach?

pkg/util/validation/limits.go
```diff
@@ -62,6 +62,11 @@ type Limits struct {
 	CardinalityLimit    int           `yaml:"cardinality_limit"`
 	MaxCacheFreshness   time.Duration `yaml:"max_cache_freshness"`
 
+	// Ruler defaults and limits.
+	// Delay the evaluation of all rules by a set interval to give a buffer
```
gotjosh, line 66: I think this comment belongs in the ruler itself right before we lookup the tenant limit.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 * [ENHANCEMENT] Add "integration" as a label for `cortex_alertmanager_notifications_total` and `cortex_alertmanager_notifications_failed_total` metrics. #3056
 * [ENHANCEMENT] Add `cortex_ruler_config_last_reload_successful` and `cortex_ruler_config_last_reload_successful_seconds` to check status of users rule manager. #3056
 * [ENHANCEMENT] Memcached dial() calls now have an optional circuit-breaker to avoid hammering a broken cache #3051
+* [ENHANCEMENT] `-ruler.evaluation-delay-duration` is now overridable as a per-tenant limit, `evaluation_delay_duration` #3098
```
gotjosh, line 26: I think this would be considered a change.

gotjosh:
Nice @tomwilkie ! FWIW I'd be in favour of this change as I think the delay is certainly something tenants might not agree on.

pkg/ruler/compat.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/weaveworks/common/user"
 
 	"github.com/cortexproject/cortex/pkg/ingester/client"
+	"github.com/cortexproject/cortex/pkg/util/validation"
```
pstibrany, line 17: To avoid introducing unnecessary dependencies, and to simplify writing unit tests, we typically add new interface with the method used by ruler. (Examples: `BlocksStoreLimits`, `MetadataValidationConfig` and few others... all of which `validation.Overrides` implement).

Here we can add `RulesLimits` with `EvaluationDelay(userID)` method.

pkg/ruler/ruler.go
```diff
@@ -48,9 +48,6 @@ type Config struct {
 	ClientTLSConfig tls.ClientConfig `yaml:"ruler_client"`
 	// How frequently to evaluate rules by default.
 	EvaluationInterval time.Duration `yaml:"evaluation_interval"`
-	// Delay the evaluation of all rules by a set interval to give a buffer
-	// to metric that haven't been forwarded to cortex yet.
-	EvaluationDelay time.Duration `yaml:"evaluation_delay_duration"`
 	// How frequently to poll for updated rules.
```
pstibrany, line 51: We should keep this, but deprecate. We can remove it after two minor versions.

pkg/ruler/compat.go
```diff
@@ -102,10 +104,12 @@ func DefaultTenantManagerFactory(
 		logger log.Logger,
 		reg prometheus.Registerer,
 	) *rules.Manager {
+		evaluationDelay := overrides.EvaluationDelay(userID)
+
 		return rules.NewManager(&rules.ManagerOptions{
 			Appendable:      &PusherAppendable{pusher: p, userID: userID},
 			Queryable:       q,
-			QueryFunc:       engineQueryFunc(engine, q, cfg.EvaluationDelay),
+			QueryFunc:       engineQueryFunc(engine, q, evaluationDelay),
```
pstibrany, line 112: I think by passing EvaluationDelay function (or overrides) to `engineQueryFunc`, we can make runtime updates to user's delay work. WDYT?

CHANGELOG.md
pracucci: Yes, it's a YAML breaking change, because `evaluation_delay_duration` has been moved from `ruler_client` to `limits`. Technically, this is the kind of breaking changes we shouldn't introduce due to the 1.x guarantees we made.

pkg/ruler/compat.go
tomwilkie: Nice catch.

pkg/ruler/ruler.go
tomwilkie: Is there anything special I need to do to deprecated a config field?

CHANGELOG.md
tomwilkie: I've put the YAML field back, and as the flag hasn't changed and the flag behaviour is the same, I thought ENHANCEMENT would be more appropriate.  But not that fussed.

pkg/util/validation/limits.go
tomwilkie: Fixed typo, but we don't change default flag values for backwards compatibility reasons.  Don't want to surprise people on upgrades.

pkg/util/validation/limits.go
```diff
@@ -62,6 +62,9 @@ type Limits struct {
 	CardinalityLimit    int           `yaml:"cardinality_limit"`
 	MaxCacheFreshness   time.Duration `yaml:"max_cache_freshness"`
 
+	// Ruler defaults and limits.
+	EvaluationDelay time.Duration `yaml:"evaluation_delay_duration"`
```
pracucci, line 66: Being the YAML structure generic for all Cortex services, I'm wondering if `evaluation_delay_duration` is too much generic and `ruler_evaluation_delay_duration` may be more appropriate.

tomwilkie: I agree

CHANGELOG.md
pracucci, line 26:
```suggestion
* [ENHANCEMENT] `-ruler.evaluation-delay-duration` is now overridable as a per-tenant limit, `ruler_evaluation_delay_duration` #3098
```

pkg/ruler/ruler.go
```diff
@@ -109,7 +109,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.ExternalURL.URL, _ = url.Parse("") // Must be non-nil
 	f.Var(&cfg.ExternalURL, "ruler.external.url", "URL of alerts return path.")
 	f.DurationVar(&cfg.EvaluationInterval, "ruler.evaluation-interval", 1*time.Minute, "How frequently to evaluate rules")
-	f.DurationVar(&cfg.EvaluationDelay, "ruler.evaluation-delay-duration", 0, "Duration to delay the evaluation of rules to ensure they underlying metrics have been pushed to cortex.")
+	f.DurationVar(&cfg.EvaluationDelay, "ruler.evaluation-delay-duration-deprecated", 0, "Deprecated. Duration to delay the evaluation of rules to ensure they underlying metrics have been pushed to cortex.")
```
pracucci, line 112: It's tricky. With the changes in this PR we're keeping the `evaluation_delay_duration` in the ruler YAML config, but we're not reading it anymore, so we're basically introducing a breaking change anyway (but more subtle because if you configured it, we ignore it).

I'm wondering if we should make it simple: introduce a breaking change (mentioning it in the CHANGELOG) and remove the `evaluation_delay_duration` at all from the ruler config.

pstibrany: I agree that keeping the field around (for yaml), but silently ignoring is the worst combination. 

Alternative to breaking YAML config would be to use this value as a default value in overrides, if configured and  warn about it being deprecated, and removing after 2 releases.

pracucci: > Alternative to breaking YAML config would be to use this value as a default value in overrides, if configured and warn about it being deprecated, and removing after 2 releases.

Yes, but would require some refactoring in the limits overrides because when you read the setting for a user you don't know if that value is the default or not.

pstibrany: > Yes, but would require some refactoring in the limits overrides because when you read the setting for a user you don't know if that value is the default or not.

I think we can set the default into `t.Cfg.LimitsConfig.EvaluationDelay` from `t.Cfg.Ruler.EvaluationDelay` (if set, and `t.Cfg.LimitsConfig.EvaluationDelay` is 0) before calling `overrides.SetDefaultLimitsForYAMLUnmarshalling` in `initRuntimeConfig`.

CHANGELOG.md
pstibrany: If old YAML field is set, it still works (if new field is not set). We log a deprecation warning in such case. I've also added CHANGELOG entry about the rename.

pkg/ruler/ruler.go
pstibrany: I've implemented my suggestion.

docs/configuration/config-file-reference.md
```diff
@@ -846,9 +846,8 @@ ruler_client:
 # CLI flag: -ruler.evaluation-interval
 [evaluation_interval: <duration> | default = 1m]
 
-# Duration to delay the evaluation of rules to ensure they underlying metrics
-# have been pushed to cortex.
-# CLI flag: -ruler.evaluation-delay-duration
+# Deprecated. Please use -ruler.evaluation-delay-duration instead.
+# CLI flag: -ruler.evaluation-delay-duration-deprecated
```
gouthamve, line 850: Hrm, this deprecated notice is not super useful for the YAML config. Hrm. Not sure how to fix it though.

pkg/cortex/modules.go
```diff
@@ -126,6 +127,19 @@ func (t *Cortex) initRing() (serv services.Service, err error) {
...
+	// We need to modify LimitsConfig before calling SetDefaultLimitsForYAMLUnmarshalling later in this method
+	// but also if runtime-config is not used, for setting limits used by initOverrides.
+	// TODO: Remove this in Cortex 1.6.
+	if t.Cfg.Ruler.EvaluationDelay != 0 && t.Cfg.LimitsConfig.RulerEvaluationDelay == 0 {
+		t.Cfg.LimitsConfig.RulerEvaluationDelay = t.Cfg.Ruler.EvaluationDelay
+
+		// No need to report if this field isn't going to be used.
+		if t.Cfg.Target == All || t.Cfg.Target == Ruler {
+			flagext.DeprecatedFlagsUsed.Inc()
+			level.Warn(util.Logger).Log("msg", "Using DEPRECATED flag config YAML flag: ruler.evaluation_delay_duration, please use limits.ruler_evaluation_delay_duration instead.")
```
gouthamve, line 139: `Using DEPRECATED flag config YAML flag` --> `Using DEPRECATED YAML config field`?

gouthamve (approved):
LGTM with nits.

docs/configuration/config-file-reference.md
pstibrany: This flag only exists to have YAML field with "deprecated" documentation. Ideally we would hide CLI flag, but our tooling is not there yet. Given that this will be removed in Cortex 1.6, let's not waste too much time on it.

PR #3103, opened 2020-08-27
feat: return stdout and stderr from all concrete service exec commands

jtlisi:
**What this PR does**:

- Returns both the standard out and standard error when running the exec command on a concrete service.

In some cases the standard error output can be desired when running a service.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany (approved):

pracucci (approved):
Yep, good idea!

PR #3106, opened 2020-08-28
[WIP] Chnage the cache interface so that the cache client can be reused in Thanos

krasi-georgiev:
```
// Thanos
type Cache interface {
	Store(ctx context.Context, data map[string][]byte, ttl
time.Duration)
	Fetch(ctx context.Context, keys []string) map[string][]byte
}

// Cortex
type Cache interface {
	Store(ctx context.Context, key []string, buf [][]byte)
	Fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missing []string)
	Stop()
}

// Will work towards this nice middle ground:

type Cache interface {
	Store(ctx context.Context, data map[string][]byte)
	Fetch(ctx context.Context, keys []string) (results map[string][]byte, missing []string)
	Stop()
}

```

related to : https://github.com/thanos-io/thanos/pull/3032

Please let me know if think there could be some issues with this change 

Stil WIP as I want to test how well will this work on the Thanos side.

Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>

### DIFF
--- a/pkg/chunk/cache/background.go
+++ b/pkg/chunk/cache/background.go
@@ -36,8 +36,7 @@ type backgroundCache struct {
 }
 
 type backgroundWrite struct {
-	keys []string
-	bufs [][]byte
+	data map[string][]byte
 }
 
 // NewBackground returns a new Cache that does stores on background goroutines.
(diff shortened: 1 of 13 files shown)


pracucci:
I haven't checked all internal usages of the Cortex cache interface. I would say this change may make sense if doesn't bring performance penalties to Cortex (eg. because you have to convert slices to maps or viceversa). Alternatively we can just keep Cortex interface as is, and have a Cache adapter in Thanos.

krasi-georgiev:
I am looking at the code now so will have more detail soon.

pstibrany:
> Alternatively we can just keep Cortex interface as is, and have a Cache adapter in Thanos.

I think this may be the best option. Changing Cache interface in Cortex will cause lot of code changes with little benefit.

(If you are going to put the effort in, then I would consider using Thanos interface. I don't think Cortex uses `missing` values anywhere, and it's easy to compute at the call site, if needed).

krasi-georgiev:
Thanks for the suggestion for the missing values. Looking at the code doesn't seem like a lot of work so will spend a bit more time before jumping to conclusions.

krasi-georgiev:
Turns out changes are needed on more places then I anticipated at the beginning and looking at the PR on the Thanos side I should be able to do it without any interface changes.

PR #3112, opened 2020-08-31
change: separate distributor into api and service module

### LINKED ISSUE #3109: Distributor endpoints hosted for Ruler and Querier
The targets in Cortex suffer from a bit of an identity crisis since they end up hosting the APIs of all of the services they depend. The best example of this is the querier. Since it depends on the distributor, every querier hosts the `/api/v1/push` endpoint. I feel this should be corrected and specific targets should not host all of the endpoints of their dependencies.

jtlisi:
**What this PR does**:

- Decouples the distributor service from the API

Fixes #3109 

TLDR: Querier will no longer host distributor endpoints

**Checklist**

### DIFF
--- a/pkg/cortex/cortex_test.go
+++ b/pkg/cortex/cortex_test.go
@@ -87,5 +87,5 @@ func TestCortex(t *testing.T) {
 	require.NotNil(t, serviceMap[Server])
 	require.NotNil(t, serviceMap[Ingester])
 	require.NotNil(t, serviceMap[Ring])
-	require.NotNil(t, serviceMap[Distributor])
+	require.NotNil(t, serviceMap[DistributorService])
 }
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -50,7 +50,6 @@ const (
 	Server              string = "server"
 	Distributor         string = "distributor"
 	DistributorService  string = "distributor-service"
-	DistributorAPI      string = "distributor-api"
 	Ingester            string = "ingester"
 	Flusher             string = "flusher"
 	Querier             string = "querier"
@@ -168,7 +167,7 @@ func (t *Cortex) initDistributorService() (serv services.Service, err error) {
 	return t.Distributor, nil
 }
 
-func (t *Cortex) initDistributorAPI() (serv services.Service, err error) {
+func (t *Cortex) initDistributor() (serv services.Service, err error) {
 	t.API.RegisterDistributor(t.Distributor, t.Cfg.Distributor)
 
 	return nil, nil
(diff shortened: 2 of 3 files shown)


pkg/cortex/modules.go
```diff
@@ -630,7 +636,9 @@ func (t *Cortex) setupModuleManager() error {
 	mm.RegisterModule(MemberlistKV, t.initMemberlistKV, modules.UserInvisibleModule)
 	mm.RegisterModule(Ring, t.initRing, modules.UserInvisibleModule)
 	mm.RegisterModule(Overrides, t.initOverrides, modules.UserInvisibleModule)
-	mm.RegisterModule(Distributor, t.initDistributor)
+	mm.RegisterModule(Distributor, nil)
+	mm.RegisterModule(DistributorAPI, t.initDistributorAPI)
+	mm.RegisterModule(DistributorService, t.initDistributorService)
```
pstibrany, line 641: DistributorService should use `modules.UserInvisibleModule` option.

pkg/cortex/modules.go
```diff
@@ -650,24 +658,26 @@ func (t *Cortex) setupModuleManager() error {
...
-		Configs:        {API},
-		AlertManager:   {API},
-		Compactor:      {API, MemberlistKV},
-		StoreGateway:   {API, Overrides, MemberlistKV},
-		Purger:         {Store, DeleteRequestsStore, API},
-		All:            {QueryFrontend, Querier, Ingester, Distributor, TableManager, Purger, StoreGateway, Ruler},
+		API:                {Server},
+		Ring:               {API, RuntimeConfig, MemberlistKV},
+		Overrides:          {RuntimeConfig},
+		Distributor:        {DistributorService, DistributorAPI},
```
pstibrany, line 664: Do we need both `Distributor` and `DistributorAPI`?

pkg/cortex/cortex_test.go
```diff
@@ -87,5 +87,5 @@ func TestCortex(t *testing.T) {
 	require.NotNil(t, serviceMap[Server])
 	require.NotNil(t, serviceMap[Ingester])
 	require.NotNil(t, serviceMap[Ring])
-	require.NotNil(t, serviceMap[Distributor])
+	require.NotNil(t, serviceMap[DistributorService])
```
pstibrany, line 90: "All" target still includes `Distributor`, I think we can keep the previous version of the test.

pstibrany (approved):
Good idea.

pkg/cortex/modules.go
jtlisi: I think the `DistributorAPI` can be removed and the route registration can be moved to the `Distributor` module

pkg/cortex/cortex_test.go
jtlisi: The new `distributor` target doesn't return a service so this was failing.

pkg/cortex/modules.go
jtlisi: Makes sense

pstibrany (approved):
LGTM.

PR #3113, opened 2020-09-01
Shuffle-sharding of queriers in the query-frontend

pstibrany:
**What this PR does**: This PR implements shuffle-sharding of queriers in the query-frontend.

Since each query frontend knows about all queriers that are connecting to it, QF can select N queriers that will be handling requests from specific users.

Implementation details:
- Query-Frontend needs to know which queriers are connected, and group multiple connections. Protocol between QF and Q has been extended to support passing querierID. Old queriers will use empty string as ID. If new Querier connects to old Query-Frontend, it will not send its ID at all.
- If the same set of queriers is connected to all frontends, these frontends will all select the same subset of queriers for the given user.
- Each user has a different subset of queriers.
- When searching for next user's queue to handle request from, each querier will fairly iterate between its set of users, in round robin fashion, as before.
- Shuffle-sharding is enabled by setting either default (`-frontend.max-queriers-per-user`) or per-user limit ("max_queriers_per_user").

- [x] integration test
- [x] testing in dev cluster

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/util/validation/limits.go
```diff
@@ -61,6 +61,7 @@ type Limits struct {
 	MaxQueryParallelism int           `yaml:"max_query_parallelism"`
 	CardinalityLimit    int           `yaml:"cardinality_limit"`
 	MaxCacheFreshness   time.Duration `yaml:"max_cache_freshness"`
+	MaxQueriersPerUser  int           `yaml:"max_queries_per_user"`
```
joe-elliott, line 64: `max_queries_per_user` => `max_queriers_per_user`

pkg/querier/frontend/frontend_querier_queues.go
```diff
@@ -0,0 +1,216 @@
...
+	if maxQueriers == 0 || len(allSortedQueriers) <= maxQueriers {
+		// All queriers can be used.
+		return nil
+	}
+
+	result := make(map[string]struct{}, maxQueriers)
+	rnd := rand.New(rand.NewSource(userSeed))
+
+	for len(result) < maxQueriers {
+		ix := rnd.Intn(len(allSortedQueriers))
```
joe-elliott, line 204: this could roll the same querier multiple times?

pkg/querier/frontend/frontend_querier_queues.go
```diff
@@ -0,0 +1,216 @@
...
+		ix := sort.SearchStrings(q.sortedQueriers, querier)
+		q.sortedQueriers = append(q.sortedQueriers[:ix], q.sortedQueriers[ix+1:]...)
+
+		q.recomputeUserQueriers()
+	}
+}
+
+func (q *queues) recomputeUserQueriers() {
+	for _, uq := range q.userQueues {
+		uq.queriers = q.selectQueriersForUser(uq.seed, q.sortedQueriers, uq.maxQueriers)
```
joe-elliott, line 190: should there be locking here?  I feel like this will be accessed from multiple goroutines.

pkg/querier/frontend/frontend_querier_queues.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		u := q.users[uid]
+		if u == "" {
+			continue
+		}
+
+		q := q.userQueues[u]
+
+		if q.queriers != nil {
```
joe-elliott, line 143: i'm not sure how to quantify the effect, but this could unfairly skip users who don't happen to be assigned to the querier currently asking for a query.

pkg/querier/frontend/frontend.go
```diff
@@ -347,6 +366,29 @@ func (f *Frontend) Process(server Frontend_ProcessServer) error {
...
 }
 
+func getQuerierID(server Frontend_ProcessServer) (string, error) {
+	err := server.Send(&FrontendToClient{
+		Type: GET_ID,
+		// Old queriers don't support GET_ID, and will try to use the request.
+		// To avoid confusing them, include dummy request.
+		HttpRequest: &httpgrpc.HTTPRequest{
+			Method: "GET",
+			Url:    "/invalid_request_sent_by_frontend",
```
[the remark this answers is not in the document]
pstibrany: This request is passed to querier, which then passes it to the Server, that returns 404. This 404 response is then forwarded by querier to frontend, which ignores it here.

I don’t see why would querier fail to attach to the frontend.

pkg/querier/frontend/frontend.go
```diff
@@ -347,6 +366,29 @@ func (f *Frontend) Process(server Frontend_ProcessServer) error {
 	}
 }
 
+func getQuerierID(server Frontend_ProcessServer) (string, error) {
+	err := server.Send(&FrontendToClient{
```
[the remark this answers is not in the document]
pstibrany: Not via an argument — those can either be unary (single message) or stream, but not combined. Same for return values. Unfortunately we already use stream argument and return value.

This is ugly, but the best we could come up with with Marco.

pkg/querier/frontend/frontend_querier_queues.go
pstibrany: Yes it can. Alternative would be to make a copy of queriers, shuffle them and use first `maxQueriers` of them. Current version avoids this extra allocation at the cost of extra iterations. Not sure which one is better (benchmark time?)

pkg/querier/frontend/frontend_querier_queues.go
pstibrany: Adding / removing querier holds the lock in `*Frontend` already. All locking is done there, `*queues` doesn’t do any locking.

pkg/querier/frontend/frontend_querier_queues.go
pstibrany: I don’t quite follow here... skipping users who use different queriers is the goal of this PR.

pkg/querier/frontend/frontend_querier_queues.go
pstibrany: I've tried to use shuffling instead of trying random numbers. Updated benchmark shows little less time needed for `BenchmarkQueueRequest` (well, with +-7%, maybe not), but also little more memory allocations.

```
name              old time/op    new time/op    delta
GetNextRequest-4    59.7µs ± 4%    59.7µs ± 1%    ~     (p=0.645 n=10+9)
QueueRequest-4       626µs ± 7%     609µs ± 2%    ~     (p=0.105 n=10+10)

name              old alloc/op   new alloc/op   delta
GetNextRequest-4    1.60kB ± 0%    1.60kB ± 0%    ~     (all equal)
QueueRequest-4       322kB ± 0%     326kB ± 0%  +1.24%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
GetNextRequest-4       100 ± 0%       100 ± 0%    ~     (all equal)
QueueRequest-4       1.07k ± 0%     1.12k ± 0%  +4.68%  (p=0.000 n=10+10)
```

harry671003: I'm not sure if I understand this logic. Don't we want to shuffle-shard each user onto n distinct queriers? Does this function n distinct queriers?

pstibrany: "Distinctness" is guaranteed by 1) using a map, 2) condition in the for loop. This loop doesn't end until there are `maxQueriers` distinct queriers in the map. In addition to that, thanks to `len(allSortedQueriers) > maxQueriers` precondition checked at the beginning of this method, we know that loop will eventually finish.

docs/configuration/config-file-reference.md
```diff
@@ -2822,6 +2827,14 @@ The `limits_config` configures default and per-tenant limits imposed by Cortex s
 # CLI flag: -frontend.max-cache-freshness
 [max_cache_freshness: <duration> | default = 1m]
 
+# Maximum number of queriers that can handle requests for single user. If set to
+# 0, or value less than number of available queriers, all queriers will handle
+# given user's requests. Each frontend will select the same queriers (if they
+# are connected to all frontends). Note that this only works with queriers
+# connecting to the query-frontend, not when using downstream URL.
+# CLI flag: -frontend.max-queriers-per-user
+[max_queriers_per_user: <int> | default = 0]
```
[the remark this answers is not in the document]
pstibrany: Store-gateways can use different sharding strategies. In case of queriers, no sharding was done before (which corresponds to using `max_queriers_per_user: 0`), and only shuffle-sharding is available (if `max_queriers_per_user` is greater than 0). We can introduce sharding_strategy option too, but personally I don't see a need for it here.

docs/configuration/config-file-reference.md
```diff
@@ -2822,6 +2827,14 @@ The `limits_config` configures default and per-tenant limits imposed by Cortex s
 # CLI flag: -frontend.max-cache-freshness
 [max_cache_freshness: <duration> | default = 1m]
 
+# Maximum number of queriers that can handle requests for single user. If set to
+# 0, or value less than number of available queriers, all queriers will handle
```
ranton256, line 2831: Was this meant to say "value more than number of available queriers"?

pkg/querier/frontend/frontend_querier_queues.go
ranton256: Doing this via repeated random selections versus  shuffling the array seems like a probably unnecessary optimization at the expense of having a more deterministic outcome in terms of iterations required. I'm not sure it will matter practically, but seems like predictable would be better, meaning shuffle a copy of the array.

You don't have to shuffle the whole array, you can make a copy then from cnt over 0 to N first indices generate a random index between cnt+1 to N-1 inclisive and swap the value in the counter index with the random index so you can't ever select a duplicate, and you don't have to swap more values than the number you need to sellect. This is basically this: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm except I described it from 0 up and they describe it from N-1 down.

pstibrany: I've changed the implementation to use suggested algorithm. (Btw, it's the same what `rand.Shuffle()` does, but in our case we can stop early, and set elements in returned map without another iteration).

docs/configuration/config-file-reference.md
pstibrany: Well spotted, fixed.

pkg/querier/frontend/frontend_querier_queues_test.go
```diff
@@ -0,0 +1,249 @@
...
+	if uc != len(uq.userQueues) {
+		return fmt.Errorf("inconsistent number of users list and user queues")
+	}
+
+	return nil
+}
+
+func TestShuffleQueriers(t *testing.T) {
+	allQueriers := []string{"a", "b", "c", "d", "e"}
+
```
ranton256, line 238: Might be good to add a test that result of shuffleQueriersForUser has unique values and all were in the original input?

ranton256 (commented):
Changes look good to me. Thanks for the array shuffling algorithm change.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 * [ENHANCEMENT] When a tenant accesses the Alertmanager UI or its API, if we have valid `-alertmanager.configs.fallback` we'll use that to start the manager and avoid failing the request. #3073
 * [ENHANCEMENT] Add `DELETE api/v1/rules/{namespace}` to the Ruler. It allows all the rule groups of a namespace to be deleted. #3120
 * [ENHANCEMENT] Experimental Delete Series: Retry processing of Delete requests during failures. #2926
+* [ENHANCEMENT] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-user` globally, or using per-user limit `max_queriers_per_user`), each user's requests will be handled by different set of queriers. #3113
```
pracucci, line 40: I've just merged a PR to cut the CHANGELOG because of the upcoming `1.4.0` release. Could you rebase `master` and move your CHANGELOG entry to the top, under `## master / unreleased`, please?

docs/configuration/config-file-reference.md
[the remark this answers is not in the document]
pracucci: I agree it's not required here, but could help with config consistency.

pkg/querier/frontend/frontend_querier_queues.go
```diff
@@ -0,0 +1,227 @@
...
+		q.sortedQueriers = append(q.sortedQueriers, querier)
+		sort.Strings(q.sortedQueriers)
+
+		q.recomputeUserQueriers()
+	}
+}
+
+func (q *queues) removeQuerierConnection(querier string) {
+	conns := q.querierConnections[querier]
+	if conns <= 0 {
```
pracucci, line 171: Probably unactionable comment: if it's < 0 it's a serious bug because next connection from this querier will not be considered as "first connection".

pkg/querier/frontend/frontend_querier_queues.go
```diff
@@ -0,0 +1,227 @@
...
+		return
+	}
+
+	conns--
+	if conns > 0 {
+		q.querierConnections[querier] = conns
+	} else {
+		delete(q.querierConnections, querier)
+
+		ix := sort.SearchStrings(q.sortedQueriers, querier)
```
pracucci, line 181: Should we double check that `q.sortedQueriers[ix] == querier`? If there's no bug, it shouldn't happen. If there's a bug and we don't double check it (eg. logging), could be very difficult to spot it.

pkg/querier/frontend/frontend_querier_queues.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	return result
+}
+
+func getSeedForUser(user string) int64 {
+	d := md5.New()
+	_, _ = d.Write([]byte(user))
+	buf := d.Sum(nil)
+	return int64(binary.BigEndian.Uint64(buf[:8]) ^ binary.BigEndian.Uint64(buf[8:]))
```
pracucci, line 226: Any reason why this logic is different than https://github.com/cortexproject/cortex/pull/3090/files#diff-d8ef8c3e13558a58373d0cbfd02798e0R494-R500 ?

pkg/querier/frontend/frontend_querier_queues.go
```diff
@@ -0,0 +1,227 @@
...
+	scratchpad = append(scratchpad, allSortedQueriers...)
+
+	last := len(scratchpad) - 1
+	for i := 0; i < queriersToSelect; i++ {
+		r := rnd.Intn(last + 1)
+		result[scratchpad[r]] = struct{}{}
+		// move selected item to the end, it won't be selected anymore.
+		scratchpad[r], scratchpad[last] = scratchpad[last], scratchpad[r]
+		last--
+	}
```
pracucci, line 217: Is it preferable consistent hashing here? I know it's not strictly required, but I'm wondering if it's preferable. I believe this algorithm doesn't guarantee it because of "swap selected item to the end", which basically shuffle the slice if we add or remove a querier to the pool.

pracucci: Probably we don't care about consistent hashing, but I would like to have your opinion too.

pracucci (approved):
Fantastic job 👏  I also appreciated a lot the tests! I left minor comments, but overall LGTM!

I would mention this is experimental in the "v1 guarantees" doc (but not add the `experimental` CLI flag prefix which we've already seen is a pain) in order to be able to do any breaking change until we got enough confidence running it in production.

I think we also mentioned we want aggressive gRPC keepalive settings for querier->query-frontend, in order to quickly detect "dead" queriers. Could you check current settings (if any) and eventually fine tune it, please?

In a separate PR, I would also work on some doc.

pkg/querier/frontend/frontend.go
```diff
@@ -356,10 +395,12 @@ func (f *Frontend) queueRequest(ctx context.Context, req *request) error {
 	req.enqueueTime = time.Now()
 	req.queueSpan, _ = opentracing.StartSpanFromContext(ctx, "queued")
 
+	maxQueriers := f.limits.MaxQueriersPerUser(userID)
+
 	f.mtx.Lock()
 	defer f.mtx.Unlock()
 
-	queue := f.queues.getOrAddQueue(userID)
+	queue := f.queues.getOrAddQueue(userID, maxQueriers)
```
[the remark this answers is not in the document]
pstibrany: I'll add a check with comment that it can only happen if user is `""`.

pkg/querier/frontend/frontend_querier_queues.go
pstibrany: Correct. It should not happen, perhaps `panic` would be in order here?

pkg/querier/frontend/frontend_querier_queues.go
pstibrany: Main reason is to avoid dropping half of the result, which seems like a waste.

docs/configuration/config-file-reference.md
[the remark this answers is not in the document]
pstibrany: > Food for thought (not having a strong opinion): I agree it's not required here, but could help with config consistency.

If we want to be consistent with other config options, we should also support "default" (non-shuffle-sharding) strategy with non-zero shard size.

pkg/querier/frontend/frontend_querier_queues.go
pstibrany: Consistent as in when new querier is added, it doesn't totally change sharding for all users.

We currently don't do that and agreed that it's ok in this case. We may want to revisit that decision at some point.

pstibrany:
> I would mention this is experimental in the "v1 guarantees" doc (but not add the `experimental` CLI flag prefix which we've already seen is a pain) in order to be able to do any breaking change until we got enough confidence running it in production.
> 
> I think we also mentioned we want aggressive gRPC keepalive settings for querier->query-frontend, in order to quickly detect "dead" queriers. Could you check current settings (if any) and eventually fine tune it, please?

We need to configure these settings on query-frontend. Default values that Cortex uses are:
- `-server.grpc.keepalive.time=2h`
- `-server.grpc.keepalive.timeout=20s`

which means that after 2 hours of no activity on the connection, server will ping the client and wait for 20 seconds for response. On no reply, connection is closed. We can include this information in the docs for people to tune (eg. time=1m/timeout=20s, or something that makes sense for their setup).

> 
> In a separate PR, I would also work on some doc.

👍

pstibrany:
@pracucci I've addressed your feedback, please take a look again when time permits. Thanks!

pracucci:
> I've addressed your feedback, please take a look again when time permits. Thanks!

LGTM. Thanks to you! 🙏

docs/configuration/config-file-reference.md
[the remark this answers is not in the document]
pracucci: Let's keep it as is for now. We marked this feature as experimental, which will allow us to eventually fine-tune the config before marking it stable.

pkg/querier/frontend/frontend_querier_queues.go
pracucci: Ok to keep it as it. I believe consistent hashing is not a required property here and we can always revisit it.

PR #3115, opened 2020-09-01
Removed deprecated untyped record from chunks WAL

### LINKED ISSUE #2793: Remove deprecated record from chunks WAL after v1.2.0
New WAL records were introduced in v1.1.0, so with the v1 guarantees, we need to support the old records for up to 2 releases. As the old record was introduced in v1.0.0, it is supported till v1.2.0, hence can be removed for v1.3.0 and get rid of lot of grpc code + complexity in WAL replay.

I will pick this up in future unless someone does it before me :)

codesome:
**What this PR does**:
As the title says

Fixes #2793

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany (approved):
Looks good, thank you!

pracucci (approved):
LGTM, thanks!

PR #3119, opened 2020-09-01
makes rule group validation an interface

owen-d:
Prometheus was refactored to support [arbitrary rule loading](https://github.com/prometheus/prometheus/pull/7569), but we currently bind incoming rule groups to promql via the underlying `rulefmt` package implementation. This PR extracts said validation functionality into an interface method on the pre-existing `MultiTenantManager` interface which will allow plug-able implementations (attractive in my case for logql expressions in Loki).

### DIFF
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -273,29 +273,6 @@ var (
 	ErrBadRuleGroup = errors.New("unable to decoded rule group")
 )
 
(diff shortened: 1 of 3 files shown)


gotjosh (approved):
💯 LGTM.

annanay25 (approved):
lgtm!

gouthamve (approved):

PR #3120, opened 2020-09-01
Allow the Ruler to delete a whole namespace

gotjosh:
**What this PR does**:

Allow the Ruler API to delete a whole namespace. There's also a small bugfix that fixes responses in the ruler API.

Needs:

- [x] Integration test

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/integration/e2ecortex/client.go
+++ b/integration/e2ecortex/client.go
@@ -331,7 +331,7 @@ func (c *Client) DeleteNamespace(namespace string) error {
 	defer cancel()
 
 	// Execute HTTP request
-	_, err := c.httpClient.Do(req.WithContext(ctx))
+	_, err = c.httpClient.Do(req.WithContext(ctx))
 	if err != nil {
 		return err
 	}
--- a/integration/e2ecortex/client.go
+++ b/integration/e2ecortex/client.go
@@ -317,6 +317,7 @@ func (c *Client) DeleteRuleGroup(namespace string, groupName string) error {
 	return nil
 }
 
+// DeleteNamespace deletes all the rule groups (and the namespace itself).
 func (c *Client) DeleteNamespace(namespace string) error {
 	// Create HTTP request
 	req, err := http.NewRequest("DELETE", fmt.Sprintf("http://%s/api/prom/rules/%s", c.rulerAddress, url.PathEscape(namespace)), nil)
@@ -324,19 +325,17 @@ func (c *Client) DeleteNamespace(namespace string) error {
 		return err
 	}
 
-	req.Header.Set("Content-Type", "application/yaml")
 	req.Header.Set("X-Scope-OrgID", c.orgID)
 
 	ctx, cancel := context.WithTimeout(context.Background(), c.timeout)
 	defer cancel()
 
 	// Execute HTTP request
-	res, err := c.httpClient.Do(req.WithContext(ctx))
+	_, err := c.httpClient.Do(req.WithContext(ctx))
 	if err != nil {
 		return err
 	}
 
-	defer res.Body.Close()
 	return nil
 }
 
--- a/integration/e2ecortex/client.go
+++ b/integration/e2ecortex/client.go
@@ -317,6 +317,29 @@ func (c *Client) DeleteRuleGroup(namespace string, groupName string) error {
 	return nil
 }
 
+func (c *Client) DeleteNamespace(namespace string) error {
+	// Create HTTP request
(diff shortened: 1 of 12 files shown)


pkg/ruler/rules/compat.go
```diff
@@ -48,21 +48,26 @@ func FromProto(rg *RuleGroupDesc) rulefmt.RuleGroup {
...
 		newRule := rulefmt.RuleNode{
-			Record:      recordNode,
-			Alert:       alertNode,
 			Expr:        exprNode,
 			Labels:      client.FromLabelAdaptersToLabels(rl.Labels).Map(),
 			Annotations: client.FromLabelAdaptersToLabels(rl.Annotations).Map(),
 			For:         model.Duration(rl.GetFor()),
 		}
 
+		if rl.GetRecord() != "" {
```
gotjosh, line 61: Unrelated to the deletion of namespace but noticed this small bug that's been here for a while - only one of `alert` or `record` should be set.

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -149,6 +149,28 @@ func (o *RuleStore) DeleteRuleGroup(ctx context.Context, userID string, namespac
...
+	}
+
+	if len(ruleGroupObjects) == 0 {
+		return rules.ErrGroupNamespaceNotFound
+	}
+
+	for _, obj := range ruleGroupObjects {
+		level.Debug(util.Logger).Log("msg", "deleting rule group", "namespace", namespace, "key", obj.Key)
+		err = o.client.DeleteObject(ctx, obj.Key)
+		if err != nil {
```
gouthamve, line 166: Shouldn't we return the error here?

integration/e2ecortex/client.go
```diff
@@ -317,6 +317,29 @@ func (c *Client) DeleteRuleGroup(namespace string, groupName string) error {
...
 }
 
+func (c *Client) DeleteNamespace(namespace string) error {
+	// Create HTTP request
+	req, err := http.NewRequest("DELETE", fmt.Sprintf("http://%s/api/prom/rules/%s", c.rulerAddress, url.PathEscape(namespace)), nil)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/yaml")
```
pstibrany, line 327: There is no body, why does Content-Type need to be set?

integration/e2ecortex/client.go
```diff
@@ -317,6 +317,29 @@ func (c *Client) DeleteRuleGroup(namespace string, groupName string) error {
...
+	ctx, cancel := context.WithTimeout(context.Background(), c.timeout)
+	defer cancel()
+
+	// Execute HTTP request
+	res, err := c.httpClient.Do(req.WithContext(ctx))
+	if err != nil {
+		return err
+	}
+
+	defer res.Body.Close()
```
pstibrany, line 339: No need to defer.

pkg/ruler/rules/compat.go
pstibrany: Can we add a test showing the bug and fix?

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -149,6 +149,29 @@ func (o *RuleStore) DeleteRuleGroup(ctx context.Context, userID string, namespac
 	return err
 }
 
+// DeleteNamespace deletes all the rule groups in the specified namespace
+func (o *RuleStore) DeleteNamespace(ctx context.Context, userID, namespace string) error {
+	ruleGroupObjects, _, err := o.client.List(ctx, generateRuleObjectKey(userID, namespace, ""))
```
pstibrany, line 154: Implementation of `generateRuleObjectKey` relies on the fact that base64("") = "". For clarity, I would avoid using that assumption in `generateRuleObjectKey` method, because base64-decoding of strings like "=" or "==" also returns "", and it's not clear if encoding "" doesn't lead to one of those.

pstibrany (approved):
LGTM, thanks!

pkg/ruler/rules/compat.go
gotjosh: Line 204 on `pkg/ruler/api_test.go` is the assertion that was missing.

pstibrany: Ah, missed that. Thanks.

pkg/ruler/rules/objectclient/rule_store.go
gotjosh: Good shout, I found that a bit strange too but did not have a justification for a change. The case of `=` and `==` gives me that.

PR #3126, opened 2020-09-03
Index page should honor configured HTTP path prefix.

### LINKED ISSUE #3118: http_path_prefix not always respected
We are running cortex in a cluster form. so we are starting several containers with the target config set.

When starting a distributor with: `http_path_prefix: "/distributor"` in the config, the index page is loaded from the domain.com/distributor page in the browser.
when clicking on `usage statistics` it redirects to `/distributor/all_user_statistics`
but clicking any other link will redirect to their corresponding page on `domain.com` instead of `domain.com/distributor`

to me it seems like not all the urls respect the `http_path_prefix` from the config

using cortex:1.3.0

pstibrany:
**What this PR does**: This PR fixes index page, and adds configured HTTP path prefix to all internal links it renders.

Fixes #3118.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/api/handlers.go
+++ b/pkg/api/handlers.go
@@ -23,21 +23,21 @@ var indexPageContent = template.Must(template.New("main").Parse(`
 		<h1>Cortex</h1>
 		<p>Admin Endpoints:</p>
 		<ul>
-			<li><a href="{{ .JoinPath "/config" }}">Current Config</a></li>
-			<li><a href="{{ .JoinPath "/distributor/all_user_stats" }}">Usage Statistics</a></li>
-			<li><a href="{{ .JoinPath "/distributor/ha_tracker" }}">HA Tracking Status</a></li>
-			<li><a href="{{ .JoinPath "/multitenant_alertmanager/status" }}">Alertmanager Status</a></li>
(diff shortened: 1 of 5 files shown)


pkg/api/handlers.go
```diff
@@ -21,29 +23,39 @@ const indexPageContent = `
...
 
-func indexHandler(w http.ResponseWriter, _ *http.Request) {
-	if _, err := w.Write([]byte(indexPageContent)); err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+type templateInp struct {
+	Prefix string
+}
+
+func (i templateInp) Join(p string) string {
```
sandeepsukhani, line 49: Should we call this `PrefixedPath` or something?

sandeepsukhani (approved):
Just a minor suggestion otherwise it LGTM!

jtlisi (approved):
LGTM, good change

pkg/api/handlers.go
pstibrany: What about `AddPathPrefix`?

sandeepsukhani: Sounds good!

PR #3127, opened 2020-09-03
Improvements to blocksconvert

pstibrany:
**What this PR does**: This PR adds some improvements to blocksconvert suite:

- builders now export timestamp of current plan (each plan is for single day, this day's timestamp is exported)
- obsolete progress files now end up converted to error files by scheduler, not just deleted. This avoids crash looping of builders trying to build the same plan.
- scheduler now exports metrics about its queue: length, oldest/newest plan day timestamp
- scanner has new `-scanner.allowed-users` flag to only generate plans for selected users

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tools/blocksconvert/scanner/scanner.go
+++ b/tools/blocksconvert/scanner/scanner.go
@@ -48,8 +48,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.UploadFiles, "scanner.upload", true, "Upload plan files.")
 	f.BoolVar(&cfg.KeepFiles, "scanner.keep-files", false, "Keep plan files locally after uploading.")
 	f.IntVar(&cfg.TablesLimit, "scanner.tables-limit", 0, "Number of tables to convert. 0 = all.")
-	f.StringVar(&cfg.AllowedUsers, "scanner.allowed-users", "", "Allowed users that can be converted, comma-separated")
-	f.StringVar(&cfg.IgnoredUserPattern, "scanner.ignore-user", "", "Regex pattern with ignored users. Only used if all -scanner.allowed-users is not set.")
+	f.StringVar(&cfg.AllowedUsers, "scanner.allowed-users", "", "Allowed users that can be converted, comma-separated. If set, only these users have plan files generated.")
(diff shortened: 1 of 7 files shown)


tools/blocksconvert/scanner/scanner.go
```diff
@@ -92,8 +95,13 @@ func NewScanner(cfg Config, scfg blocksconvert.SharedConfig, l log.Logger, reg p
...
 	}
 
+	var users = blocksconvert.AllowAllUsers
+	if cfg.AllowedUsers != "" {
+		users = blocksconvert.ParseAllowedUsers(cfg.AllowedUsers)
+	}
+
 	var ignoredUserRegex *regexp.Regexp = nil
-	if cfg.IgnoredUserPattern != "" {
+	if users.AllUsersAllowed() && cfg.IgnoredUserPattern != "" {
```
codesome, line 104: `!users.AllUsersAllowed()`?

tools/blocksconvert/scheduler/scheduler.go
```diff
@@ -215,6 +223,29 @@ func (s *Scheduler) scanBucketForPlans(ctx context.Context) error {
 	return nil
 }
 
+func (s *Scheduler) deleteObsoleteProgressFiles(ctx context.Context, plan *plan, planBaseName string) {
+	for pg, t := range plan.ProgressFiles {
+		if time.Since(t) > s.cfg.MaxProgressFileAge {
```
codesome, line 228:
```go
if time.Since(t) <= s.cfg.MaxProgressFileAge {
	continue
}
```
And we can reduce the indentation for rest of the code.

tools/blocksconvert/scanner/scanner.go
pstibrany: This is correct. If all users are allowed (no config option for allowed users was passed), only then we want to ignore some users.

If "allowed users" option was used, only plans for those specific users are generated.

codesome: Right, I confused `Only used if all -scanner.allowed-users is not set.` to be allow all users :)

pstibrany: Perhaps I should improve description strings?

pracucci (approved):
Good job!

PR #3133, opened 2020-09-04
Index page should only show links that are valid for running Cortex

### LINKED ISSUE #3128: Having a single index-page for all cortex targets / components is misleading
Currently all cortex components expose a single `index.html` that looks like 
cotex-index-page

However, other than `current config`, which exposes the component's configuration, other endpoints are component specific.

```
Cortex
Admin Endpoints:
  - Current Config          // --> Specific / contextual to the current target / component
  - Usage Statistics
  - HA Tracking Status
  - Alertmanager Status    // --> Dependent on optional target
  - Ingester Ring Status   // --> Distributor service only
  - Ruler Ring Status      // --> Dependent on optional target's service, 404 if not deployed
  - Service Status
  - Compactor Ring Status (experimental blocks storage)
  - Store Gateway Ring (experimental blocks storage)


Dangerous:
 - Trigger a Flush           // --> Ingester service only
 - Trigger Ingester Shutdown // --> Ingester service only
```

It would be nice to have a per target rendering on the index page with urls that are pertinent to that target only. 

As a new user, I was confused for quite a while on why certain endpoints kept giving me 404s. I mistakenly took these 404s as signs of a faulty config on my part spending a good number of hours on it, even though the cortex stack was working as expected. Thus, I believe it could be treated as an UX issue.

pstibrany:
**What this PR does**: This PR makes index page dynamic. Cortex will now only show links that are valid for running instance.

Fixes #3128

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 * [BUGFIX] Ingester: If push request contained both valid and invalid samples, valid samples were ingested but not stored to WAL of the chunks storage. This has been fixed. #3067
 * [BUGFIX] Ruler: Config API would return both the `record` and `alert` in `YAML` response keys even when one of them must be empty. #3120
 * [BUGFIX] Index page now uses configured HTTP path prefix when creating links. #3126
+* [BUGFIX] Index page no longer shows links that are not valid for running Cortex instance. #3133
 
 ## 1.3.0 / 2020-08-21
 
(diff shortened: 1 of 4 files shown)


pracucci (approved):
The strategy you adopted is very smart. Good job! 👏

PR #3135, opened 2020-09-04
Expose number of entries in refcache as a metric.

pstibrany:
**What this PR does**: Blocks-ingester uses RefCache to remember "ref" for each pushed series. RefCache is purged periodically (every 5 minutes, hardcoded), and removes series that were not used in last 10 minutes (hardcoded too). This means that number entries in the refcache roughly corresponds to active series for a user.

This PR exposes number of entries in refcache as a per-user metric.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany:
Closing in favor of #3153.

PR #3147, opened 2020-09-08
test: provide unique network name for docker tests

### LINKED ISSUE #3146: Unable to run simultaneous integration tests on a single docker host
A docker network name must be unique and only a single one can exist at any given time. We use a global constant network name for our integration tests which leads to issues if multiple tests are ran on a single docker host.

jtlisi:
**What this PR does**:

Appends a randomly generated test ID string at launch of the integration tests to the network name used by each scenario. On linux, Docker bridge networks are implemented using IP tables and there should not be any issues with additional networks being created. The will only persist after tests that fail to shut down correctly.

Fixes #3146 

**Checklist**
- [x] Tests updated

### DIFF
--- a/integration/configs.go
+++ b/integration/configs.go
@@ -4,8 +4,8 @@ package integration
 
 import (
 	"fmt"
+	"math/rand"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"text/template"
 	"time"
@@ -18,17 +18,8 @@ type storeConfig struct {
 	From, IndexStore string
 }
 
-// Docker networks must be uniquely named. In order to run parallel integration
-// tests on the same docker host, a network name is generated using the current
-// timestamp in milliseconds.
-var networkName string
-
-func init() {
-	nowMillis := int(time.Now().UnixNano() / 1000)
-	networkName = "e2e-cortex-test-" + strconv.Itoa(nowMillis)
-}
-
 const (
+	networkNamePrefix      = "e2e-cortex-test-"
 	bucketName             = "cortex"
 	cortexConfigFile       = "config.yaml"
 	cortexSchemaConfigFile = "schema.yaml"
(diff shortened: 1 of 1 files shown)


integration/configs.go
```diff
@@ -27,7 +29,24 @@ const (
 	caCertFile             = "certs/root.crt"
 	serverCertFile         = "certs/server.crt"
 	serverKeyFile          = "certs/server.key"
-	storeConfigTemplate    = `
+)
+
+// Generate a 6 character (a-z,0-9) id for the test being run.
+// This ID is used to generate a network name for the test
+func genUniqueTestID() string {
+	rand.Seed(time.Now().UnixNano())
```
pstibrany, line 37: This is (posibly re-)seeding global RNG. In general I'd avoid doing that, and prefer to use local instance instead. But for integration tests, that should be fine.

pstibrany (approved):
LGTM. (Just curious, would it be possible to extend this to make integration tests run in parallel?)

jtlisi:
@pstibrany I was able to get the tests to run in parallel by generating the unique ID in the Scenario itself and adding `t.Parallel()` to each test case. However, running in parallel caused a bunch of otherwise healthy tests to fail due to WaitReady timeouts. I don't think it makes sense to parallelize integration tests in a single step. Rather distributing multiple integration test steps across multiple runners in parallel might make things a bit better.

I rolled back the changes I made on this PR to remove the extra parallelization code.

pracucci (commented):
Having a static network name was done in purpose, to cleanup the previous run if crashed for any reason. If you run integration tests, they crash, then you run again, if the network name is the same the 2nd run will cleanup (stop containers) left running from the previous run. With this change this is not true anymore.

What's the use case of running the same integration tests concurrently on the same node? Maybe we can find a different way to implement it, while keeping cleanup as well?

jtlisi:
@pracucci When running integration tests using a [Drone IO](https://docs.drone.io/) runner, the best way to spin up a docker environment is to use mount the docker socket of the underlying host. However, this issue comes into play when multiple jobs running the integration tests run on the same runner.

There may be a workaround by using a DIND service for integration tests instead of docker itself, like here: https://docker-runner.docs.drone.io/examples/service/docker_dind/. I will try that out and see if that works.

PR #3153, opened 2020-09-10
Configurable active user series

pstibrany:
**What this PR does**: This PR makes tracking of active series in the blocks storage configurable (there are two parameters: how often to purge old series from memory and update metrics, and how old must series be to be considered inactive).

This makes it possible to track active series more precisely even if they are still in the TSDB head, but haven't been appended any samples recently.

In the current state of the PR, this is only wired into the blocks engine, and not chunks engine.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added two options to control how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
+* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added options to control how whether active series tracking is enabled (`-ingester.active-series-enabled`, defaults to false), and how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
 
 ## 1.4.0-rc.0 in progress
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added options to control how whether active series tracking is enabled (`-ingester.active-series-enabled`, defaults to false), and how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
+* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added options to control whether active series tracking is enabled (`-ingester.active-series-enabled`, defaults to false), and how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
 
 ## 1.4.0-rc.0 in progress
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Experimental blocks storage: added new metric `cortex_ingester_tsdb_active_series` to track active series more accurately. Also added two options to control how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle`). #3153
+
 ## 1.4.0-rc.0 in progress
 
 * [CHANGE] Cassandra backend support is now GA (stable). #3180
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [ENHANCEMENT] Experimental blocks storage: added new metric `cortex_ingester_tsdb_active_series` to track active series more accurately. Also added two options to control how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle`). #3153
+* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added two options to control how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
 
 ## 1.4.0-rc.0 in progress
 
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -614,6 +614,10 @@ lifecycler:
 # CLI flag: -ingester.rate-update-period
 [rate_update_period: <duration> | default = 15s]
 
+# Enable tracking of active series.
+# CLI flag: -ingester.active-series-enabled
+[active_series_enabled: <boolean> | default = false]
+
 # How often to update active series metrics.
 # CLI flag: -ingester.active-series-update-period
 [active_series_update_period: <duration> | default = 1m]
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -614,13 +614,13 @@ lifecycler:
 # CLI flag: -ingester.rate-update-period
 [rate_update_period: <duration> | default = 15s]
 
-# How often to update active series metrics (blocks engine only).
+# How often to update active series metrics.
 # CLI flag: -ingester.active-series-update-period
-[active_series_update_period: <duration> | default = 5m]
+[active_series_update_period: <duration> | default = 1m]
 
-# After what time is series considered to be inactive (blocks engine only).
-# CLI flag: -ingester.active-series-idle
-[active_series_idle: <duration> | default = 15m]
+# After what time a series is considered to be inactive.
+# CLI flag: -ingester.active-series-idle-timeout
+[active_series_idle_timeout: <duration> | default = 15m]
 ```
 
 ### `querier_config`
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -614,17 +614,17 @@ lifecycler:
 # CLI flag: -ingester.rate-update-period
 [rate_update_period: <duration> | default = 15s]
 
-# Enable tracking of active series.
-# CLI flag: -ingester.active-series-enabled
-[active_series_enabled: <boolean> | default = false]
+# Enable tracking of active series and export them as metrics.
+# CLI flag: -ingester.active-series-metrics-enabled
+[active_series_metrics_enabled: <boolean> | default = false]
 
 # How often to update active series metrics.
-# CLI flag: -ingester.active-series-update-period
-[active_series_update_period: <duration> | default = 1m]
+# CLI flag: -ingester.active-series-metrics-update-period
+[active_series_metrics_update_period: <duration> | default = 1m]
 
 # After what time a series is considered to be inactive.
-# CLI flag: -ingester.active-series-idle-timeout
-[active_series_idle_timeout: <duration> | default = 15m]
+# CLI flag: -ingester.active-series-metrics-idle-timeout
+[active_series_metrics_idle_timeout: <duration> | default = 10m]
 ```
 
 ### `querier_config`
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -613,6 +613,14 @@ lifecycler:
 # Period with which to update the per-user ingestion rates.
 # CLI flag: -ingester.rate-update-period
 [rate_update_period: <duration> | default = 15s]
+
+# How often to update active series metrics (blocks engine only).
+# CLI flag: -ingester.active-series-update-period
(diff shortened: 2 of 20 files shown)


pkg/ingester/ingester.go
```diff
@@ -103,6 +105,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.MetadataRetainPeriod, "ingester.metadata-retain-period", 10*time.Minute, "Period at which metadata we have not seen will remain in memory before being deleted.")
 
 	f.DurationVar(&cfg.RateUpdatePeriod, "ingester.rate-update-period", 15*time.Second, "Period with which to update the per-user ingestion rates.")
+	f.DurationVar(&cfg.ActiveSeriesUpdatePeriod, "ingester.active-series-update-period", 5*time.Minute, "How often to update active series metrics (blocks engine only).")
```
pracucci, line 108: I would default to 1m, to have more accurate tracking (by default).

pkg/ingester/ingester.go
```diff
@@ -103,6 +105,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
+	f.DurationVar(&cfg.ActiveSeriesIdle, "ingester.active-series-idle", 15*time.Minute, "After what time is series considered to be inactive (blocks engine only).")
```
pracucci, line 109: Maybe we should call the CLI flag / YAML option `active-series-idle-time` or `active-series-idle-timeout`?

pracucci, line 109:
```suggestion
	f.DurationVar(&cfg.ActiveSeriesIdle, "ingester.active-series-idle", 15*time.Minute, "After what time a series is considered to be inactive (blocks engine only).")
```

pkg/storage/tsdb/active_series.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+// ActiveSeries is keeping track of recently active series for a single tenant.
+type ActiveSeries struct {
+	stripes [numActiveSeriesStripes]activeSeriesStripe
+}
+
+// activeSeriesStripe holds a subset of the series timestamps for a single tenant.
+type activeSeriesStripe struct {
+	refsMu        sync.Mutex
```
pracucci, line 26: It's called `refsMu` but doesn't project `ref` only. I would rename for clarity.

pkg/storage/tsdb/active_series.go
```diff
@@ -0,0 +1,168 @@
...
+		total += c.stripes[s].getActive()
+	}
+	return total
+}
+
+func (s *activeSeriesStripe) updateSeriesTimestamp(now time.Time, series labels.Labels, fp model.Fingerprint, labelsCopy func(labels.Labels) labels.Labels) {
+	s.refsMu.Lock()
+	defer s.refsMu.Unlock()
+
+	nanos := now.UnixNano()
```
pracucci, line 77: Renaming `nanos` to `nowNanos` could help read the code.

pkg/ingester/ingester_v2.go
```diff
@@ -364,6 +376,19 @@ func (i *Ingester) updateLoop(ctx context.Context) error {
...
 }
 
+func (i *Ingester) updateActiveSeries() {
+	for _, userID := range i.getTSDBUsers() {
+		userDB := i.getTSDB(userID)
+		if userDB == nil {
+			continue
+		}
+
+		now := time.Now()
```
pracucci, line 386: I would move this before the loop.

pracucci (commented):
Thanks for working on it! I have few concerns:
1. The new metric adds more cognitive load to users when trying to understand the differences between chunks and blocks storage. Don't have an answer, but if it could work for the chunks storage too would probably easier to understand.
2. I'm a bit scared of the lock contention in ActiveSeries (because it's an exclusive lock). I'm wondering if we switch the entry timestamp to an atomic, if a RWMutex could be used (and most of the time we would acquire the read because we expect the series to be already in the map).

pstibrany:
> Thanks for working on it! I have few concerns:
> 
> 1. The new metric adds more cognitive load to users when trying to understand the differences between chunks and blocks storage. Don't have an answer, but if it could work for the chunks storage too would probably easier to understand.

I agree. Let's do that. (Reason why I kept chunks out of this PR is basically to first see how well it works in practice.)

> 2. I'm a bit scared of the lock contention in ActiveSeries (because it's an exclusive lock). I'm wondering if we switch the entry timestamp to an atomic, if a RWMutex could be used (and most of the time we would acquire the read because we expect the series to be already in the map).

I will write a benchmark to see throughput of both approaches.

pstibrany:
Benchmark showing improvements when using RW lock and atomics to update timestamps.

```
name                                  old time/op    new time/op    delta
ActiveSeriesTest_single_series/50-4      154ns ± 2%      76ns ± 1%  -50.52%  (p=0.008 n=5+5)
ActiveSeriesTest_single_series/100-4     156ns ± 1%      76ns ± 1%  -51.25%  (p=0.008 n=5+5)
ActiveSeriesTest_single_series/500-4     158ns ± 1%      76ns ± 2%  -52.01%  (p=0.008 n=5+5)

name                                  old alloc/op   new alloc/op   delta
ActiveSeriesTest_single_series/50-4      0.00B          0.00B          ~     (all equal)
ActiveSeriesTest_single_series/100-4     0.00B          0.00B          ~     (all equal)
ActiveSeriesTest_single_series/500-4     0.00B          0.00B          ~     (all equal)

name                                  old allocs/op  new allocs/op  delta
ActiveSeriesTest_single_series/50-4       0.00           0.00          ~     (all equal)
ActiveSeriesTest_single_series/100-4      0.00           0.00          ~     (all equal)
ActiveSeriesTest_single_series/500-4      0.00           0.00          ~     (all equal)
```

pkg/ingester/active_series.go
```diff
@@ -0,0 +1,198 @@
...
+	for s := 0; s < numActiveSeriesStripes; s++ {
+		total += c.stripes[s].getActive()
+	}
+	return total
+}
+
+func (s *activeSeriesStripe) updateSeriesTimestamp(now time.Time, series labels.Labels, fp model.Fingerprint, labelsCopy func(labels.Labels) labels.Labels) {
+	nowNanos := now.UnixNano()
+
+	e := s.findEntryForSeries(fp, series)
```
pracucci, line 81: I think there's a race condition: we get a reference of an `activeSeriesEntry` and then we do update it. In the meanwhile, the purge could remove it, because the read lock is kept only during `findEntryForSeries()`. However, the next remote write push will add it back, right? If so, LGTM, but I would like your opinion too.

gouthamve:
> The new metric adds more cognitive load to users when trying to understand the differences between chunks and blocks storage. Don't have an answer, but if it could work for the chunks storage too would probably easier to understand.

I'd be okay adding this to chunks store too. How much overhead will this add?

pstibrany:
> > The new metric adds more cognitive load to users when trying to understand the differences between chunks and blocks storage. Don't have an answer, but if it could work for the chunks storage too would probably easier to understand.
> 
> I'd be okay adding this to chunks store too. How much overhead will this add?

By my counting, each `activeSeriesEntry` needs 40 bytes on its own + copy of the labels. For 1M series that's at least 40 MB, and if we estimate labels to take 200 bytes, that would be additional 200 MB.

(In blocks storage, we try to reuse labels copy if added to the ref cache at the same time. In chunks doing the same is not easy.)

pkg/ingester/active_series.go
pstibrany: Well spotted. Should be fixed now. (There was another problem... since atomic integer was not a pointer, but part of "entry", which was stored in a slice as a plain struct, without pointer, it could actually *move* in memory when slice grows. That's why atomic integers are now pointers).

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 * [ENHANCEMENT] When a tenant accesses the Alertmanager UI or its API, if we have valid `-alertmanager.configs.fallback` we'll use that to start the manager and avoid failing the request. #3073
 * [ENHANCEMENT] Add `DELETE api/v1/rules/{namespace}` to the Ruler. It allows all the rule groups of a namespace to be deleted. #3120
 * [ENHANCEMENT] Experimental Delete Series: Retry processing of Delete requests during failures. #2926
+* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added two options to control how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
```
pracucci, line 38: I've just merged a PR to cut the CHANGELOG because of the upcoming `1.4.0` release. Could you rebase `master` and move your CHANGELOG entry to the top, under `## master / unreleased`, please?

pracucci (approved):
Excellent job. I left a couple of nits. As discussed offline, I would add a `enabled` flag (default to false) to not let every Cortex user paying the cost of this tracking if not required (this is usually required by vendors billing systems).

pstibrany:
> Excellent job. I left a couple of nits. As discussed offline, I would add a `enabled` flag (default to false) to not let every Cortex user paying the cost of this tracking if not required (this is usually required by vendors billing systems).

Added `-ingester.active-series-enabled` flag (defaults to false), and switched fingerprinting to use xxHash 64-bit, which is faster than fnv64a (more visible for bigger labels, so benchmark now uses bigger label), but requires extra small allocation:

```
name                         old time/op    new time/op    delta
ActiveSeries_UpdateSeries-4    1.03µs ± 6%    0.74µs ±17%  -28.25%  (p=0.008 n=5+5)

name                         old alloc/op   new alloc/op   delta
ActiveSeries_UpdateSeries-4      206B ± 0%      253B ±11%  +23.01%  (p=0.008 n=5+5)

name                         old allocs/op  new allocs/op  delta
ActiveSeries_UpdateSeries-4      2.00 ± 0%      3.00 ± 0%  +50.00%  (p=0.008 n=5+5)
```

pstibrany:
I've also made sure that if `-ingester.active-series-enabled` is `false`, then new metric doesn't appear in the metrics at all. Previous version of the chunks-code actually exported 0 value for each user, which is useless. Unfortunately this made PR bigger :( (alternative would be to not register the metric into to metrics registry, but still create all in-memory structures).

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added options to control how whether active series tracking is enabled (`-ingester.active-series-enabled`, defaults to false), and how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
```
```suggestion
* [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added options to control whether active series tracking is enabled (`-ingester.active-series-enabled`, defaults to false), and how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
```

docs/configuration/config-file-reference.md
```diff
@@ -613,6 +613,18 @@ lifecycler:
 # Period with which to update the per-user ingestion rates.
 # CLI flag: -ingester.rate-update-period
 [rate_update_period: <duration> | default = 15s]
+
+# Enable tracking of active series.
+# CLI flag: -ingester.active-series-enabled
+[active_series_enabled: <boolean> | default = false]
```
pracucci, line 619: Should we prefix all such config options with `active_series_metrics` to make it more clear?

pkg/ingester/user_state.go
```diff
@@ -30,6 +31,8 @@ type userStates struct {
 	limiter *Limiter
 	cfg     Config
 	metrics *ingesterMetrics
+	// Is tracking of active series enabled?
+	activeSeriesEnabled bool
```
pracucci, line 35: Food for thought: tor simplicity, what if we remove this and initialise `*ActiveSeries` (will be never filled) and we just don't register the gauge metric to the registerer?

pkg/ingester/ingester_v2_test.go
```diff
@@ -262,6 +315,9 @@ func TestIngester_v2Push(t *testing.T) {
 			// Create a mocked ingester
 			cfg := defaultIngesterTestConfig()
 			cfg.LifecyclerConfig.JoinAfter = 0
+			if testData.disableActiveSeries {
+				cfg.ActiveSeriesEnabled = false
+			}
```
pracucci, line 320:
```suggestion
			cfg.ActiveSeriesEnabled = !testData.disableActiveSeries
```

pracucci (approved):
Good job, still LGTM! Left final nits.

pkg/ingester/user_state.go
pstibrany: I turned that option down at first as I expected lot of tests calling `newIngesterMetrics`. Turns out it was fewer places, so I've changed code to do that instead.

bboreham:
> In the current state of the PR, this is only wired into the blocks engine, and not chunks engine.

Is this still true?

pstibrany:
> > In the current state of the PR, this is only wired into the blocks engine, and not chunks engine.
> 
> Is this still true?

Not anymore. Also there is now a flag to enable this, which is disabled by default.

Issue #3063, opened 2020-08-20
Expose TLS server name parameter in TLS client options


simonswine:
I suggest adding add `tls-server-name` parameter to all TLS client options https://github.com/golang/go/blob/cd10f5f632983e63b156478abdfa0bd57e93898c/src/crypto/tls/common.go#L592

Without having the ability to configure a expected ServerName, it will be relatively hard to supply the correct certificates, as through our ring discovery we are mostly connecting to IP addresses and the certificate will then have the IP SAN set accordingly.

IP address can be quite hard to predict and this might get quite tricky in terms of cert management.

pracucci:
The PR #3156 is still open and should address this.

pracucci:
> The PR #3156 is still open and should address this.

_Still valid_

PR #3156, opened 2020-09-10
Support other TLS modes than mutual auth in Client

### LINKED ISSUE #3062: Support other TLS modes than mutual auth in Client
This is a follow-up on findings in PR https://github.com/cortexproject/cortex/pull/3030#discussion_r470503597.

Currently we are only enabling TLS, if either a client certificate or a ca certificate is enabled: https://github.com/cortexproject/cortex/blob/master/pkg/util/tls/tls.go#L38

We should also support TLS without those flags, for example for 

I propose:

- Adding a flag `tls-enable` to manually enable TLS. The flag will be implicitly set to true by specifying any of the other TLS flags)
- Add test coverage for those operational modes

simonswine:
**What this PR does**:

Increase TLS test coverage and support other TLS modes than mutual auth in Client by implementing explicit TLS enable flag.

This also adds a `tls-server-name` flag to require a specific server name instead of the hostname on the certificate.

Fixes #3062 #3063 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master / unreleased
 
 * [CHANGE] Ingester: don't update internal "last updated" timestamp of TSDB if tenant only sends invalid samples. This affects how "idle" time is computed. #3727
+* [CHANGE] Require explicit flag `-<prefix>.tls-enabled` to enable TLS in GRPC clients. Previously it was enough to specify a TLS flag to enable TLS validation. #3156
 * [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651
   - `-<prefix>.s3.sse.type`
   - `-<prefix>.s3.sse.kms-key-id`
@@ -13,6 +14,7 @@
   * Prevent compaction loop in TSDB on data gap.
 * [ENHANCEMENT] Return server side performance metrics for query-frontend (using Server-timing header). #3685
 * [ENHANCEMENT] Runtime Config: Add a `mode` query parameter for the runtime config endpoint. `/runtime_config?mode=diff` now shows the YAML runtime configuration with all values that differ from the defaults. #3700
+* [ENHANCEMENT] Add flag `-<prefix>.tls-server-name` to require a specific server name instead of the hostname on the certificate. #3156
 * [BUGFIX] HA Tracker: don't track as error in the `cortex_kv_request_duration_seconds` metric a CAS operation intentionally aborted. #3745
 
 ## 1.7.0 in progress
--- a/pkg/chunk/gcp/bigtable_index_client.go
+++ b/pkg/chunk/gcp/bigtable_index_client.go
@@ -51,6 +51,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.TableCacheEnabled, "bigtable.table-cache.enabled", true, "If enabled, once a tables info is fetched, it is cached.")
 	f.DurationVar(&cfg.TableCacheExpiration, "bigtable.table-cache.expiration", 30*time.Minute, "Duration to cache tables before checking again.")
 
+	// This overrides our default from TLS disabled to TLS enabled
+	cfg.GRPCClientConfig.TLSEnabled = true
 	cfg.GRPCClientConfig.RegisterFlagsWithPrefix("bigtable", f)
 }
 
@@ -73,8 +75,11 @@ type storageClientV1 struct {
 
 // NewStorageClientV1 returns a new v1 StorageClient.
 func NewStorageClientV1(ctx context.Context, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.IndexClient, error) {
-	opts := toOptions(cfg.GRPCClientConfig.DialOption(bigtableInstrumentation()))
-	client, err := bigtable.NewClient(ctx, cfg.Project, cfg.Instance, opts...)
+	dialOpts, err := cfg.GRPCClientConfig.DialOption(bigtableInstrumentation())
+	if err != nil {
+		return nil, err
+	}
+	client, err := bigtable.NewClient(ctx, cfg.Project, cfg.Instance, toOptions(dialOpts)...)
 	if err != nil {
 		return nil, err
 	}
@@ -97,8 +102,11 @@ func newStorageClientV1(cfg Config, schemaCfg chunk.SchemaConfig, client *bigtab
 
 // NewStorageClientColumnKey returns a new v2 StorageClient.
 func NewStorageClientColumnKey(ctx context.Context, cfg Config, schemaCfg chunk.SchemaConfig) (chunk.IndexClient, error) {
-	opts := toOptions(cfg.GRPCClientConfig.DialOption(bigtableInstrumentation()))
-	client, err := bigtable.NewClient(ctx, cfg.Project, cfg.Instance, opts...)
+	dialOpts, err := cfg.GRPCClientConfig.DialOption(bigtableInstrumentation())
+	if err != nil {
+		return nil, err
+	}
+	client, err := bigtable.NewClient(ctx, cfg.Project, cfg.Instance, toOptions(dialOpts)...)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/util/grpcclient/grpcclient.go
+++ b/pkg/util/grpcclient/grpcclient.go
@@ -26,6 +26,9 @@ type Config struct {
 
 	BackoffOnRatelimits bool               `yaml:"backoff_on_ratelimits"`
 	BackoffConfig       util.BackoffConfig `yaml:"backoff_config"`
+
+	TLSEnabled bool             `yaml:"tls_enabled"`
+	TLS        tls.ClientConfig `yaml:",inline"`
 }
 
 // RegisterFlags registers flags.
@@ -41,8 +44,11 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.Float64Var(&cfg.RateLimit, prefix+".grpc-client-rate-limit", 0., "Rate limit for gRPC client; 0 means disabled.")
 	f.IntVar(&cfg.RateLimitBurst, prefix+".grpc-client-rate-limit-burst", 0, "Rate limit burst for gRPC client.")
 	f.BoolVar(&cfg.BackoffOnRatelimits, prefix+".backoff-on-ratelimits", false, "Enable backoff and retry when we hit ratelimits.")
+	f.BoolVar(&cfg.TLSEnabled, prefix+".tls-enabled", cfg.TLSEnabled, "Enable TLS in the GRPC client. This flag needs to be enabled when any other TLS flag is set.")
 
 	cfg.BackoffConfig.RegisterFlags(prefix, f)
+
+	cfg.TLS.RegisterFlagsWithPrefix(prefix, f)
 }
 
 func (cfg *Config) Validate(log log.Logger) error {
@@ -67,7 +73,14 @@ func (cfg *Config) CallOptions() []grpc.CallOption {
 }
 
 // DialOption returns the config as a grpc.DialOptions.
-func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) []grpc.DialOption {
+func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) ([]grpc.DialOption, error) {
+	var opts []grpc.DialOption
+	tlsOpts, err := cfg.TLS.GetGRPCDialOptions(cfg.TLSEnabled)
+	if err != nil {
+		return nil, err
+	}
+	opts = append(opts, tlsOpts...)
+
 	if cfg.BackoffOnRatelimits {
 		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{NewBackoffRetry(cfg.BackoffConfig)}, unaryClientInterceptors...)
 	}
@@ -76,7 +89,8 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{NewRateLimiter(cfg)}, unaryClientInterceptors...)
 	}
 
-	return []grpc.DialOption{
+	return append(
+		opts,
 		grpc.WithDefaultCallOptions(cfg.CallOptions()...),
 		grpc.WithUnaryInterceptor(grpc_middleware.ChainUnaryClient(unaryClientInterceptors...)),
 		grpc.WithStreamInterceptor(grpc_middleware.ChainStreamClient(streamClientInterceptors...)),
@@ -85,31 +99,5 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 			Timeout:             time.Second * 10,
 			PermitWithoutStream: true,
 		}),
-	}
-}
-
-// ConfigWithTLS is the config for a grpc client with tls
-type ConfigWithTLS struct {
-	GRPC Config           `yaml:",inline"`
-	TLS  tls.ClientConfig `yaml:",inline"`
-}
-
-// RegisterFlagsWithPrefix registers flags with prefix.
-func (cfg *ConfigWithTLS) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
-	cfg.GRPC.RegisterFlagsWithPrefix(prefix, f)
-	cfg.TLS.RegisterFlagsWithPrefix(prefix, f)
-}
-
-func (cfg *ConfigWithTLS) Validate(log log.Logger) error {
-	return cfg.GRPC.Validate(log)
-}
-
-// DialOption returns the config as a grpc.DialOptions
-func (cfg *ConfigWithTLS) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) ([]grpc.DialOption, error) {
-	opts, err := cfg.TLS.GetGRPCDialOptions()
-	if err != nil {
-		return nil, err
-	}
-
-	return append(opts, cfg.GRPC.DialOption(unaryClientInterceptors, streamClientInterceptors)...), nil
+	), nil
 }
--- a/pkg/util/grpcclient/grpcclient.go
+++ b/pkg/util/grpcclient/grpcclient.go
@@ -44,7 +44,7 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.Float64Var(&cfg.RateLimit, prefix+".grpc-client-rate-limit", 0., "Rate limit for gRPC client; 0 means disabled.")
 	f.IntVar(&cfg.RateLimitBurst, prefix+".grpc-client-rate-limit-burst", 0, "Rate limit burst for gRPC client.")
 	f.BoolVar(&cfg.BackoffOnRatelimits, prefix+".backoff-on-ratelimits", false, "Enable backoff and retry when we hit ratelimits.")
-	f.BoolVar(&cfg.TLSEnabled, prefix+".tls-enabled", cfg.TLSEnabled, "Enable TLS in the GRPC client. This flag needs to be enabled when any other TLS flag is set.")
+	f.BoolVar(&cfg.TLSEnabled, prefix+".tls-enabled", cfg.TLSEnabled, "Enable TLS in the GRPC client. This flag needs to be enabled when any other TLS flag is set. If set to false, insecure connection to gRPC server will be used.")
 
 	cfg.BackoffConfig.RegisterFlags(prefix, f)
 
--- a/pkg/util/tls/tls.go
+++ b/pkg/util/tls/tls.go
@@ -16,6 +16,7 @@ type ClientConfig struct {
 	CertPath           string `yaml:"tls_cert_path"`
 	KeyPath            string `yaml:"tls_key_path"`
 	CAPath             string `yaml:"tls_ca_path"`
+	ServerName         string `yaml:"tls_server_name"`
 	InsecureSkipVerify bool   `yaml:"tls_insecure_skip_verify"`
 }
 
@@ -29,18 +30,15 @@ func (cfg *ClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet)
 	f.StringVar(&cfg.CertPath, prefix+".tls-cert-path", "", "Path to the client certificate file, which will be used for authenticating with the server. Also requires the key path to be configured.")
 	f.StringVar(&cfg.KeyPath, prefix+".tls-key-path", "", "Path to the key file for the client certificate. Also requires the client certificate to be configured.")
 	f.StringVar(&cfg.CAPath, prefix+".tls-ca-path", "", "Path to the CA certificates file to validate server certificate against. If not set, the host's root CA certificates are used.")
+	f.StringVar(&cfg.ServerName, prefix+".tls-server-name", "", "Override the expected name on the server certificate.")
 	f.BoolVar(&cfg.InsecureSkipVerify, prefix+".tls-insecure-skip-verify", false, "Skip validating server certificate.")
 }
 
 // GetTLSConfig initialises tls.Config from config options
 func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
-	// no tls config given at all
-	if cfg.CertPath == "" && cfg.KeyPath == "" && cfg.CAPath == "" {
-		return nil, nil
-	}
-
 	config := &tls.Config{
 		InsecureSkipVerify: cfg.InsecureSkipVerify,
+		ServerName:         cfg.ServerName,
 	}
 
 	// read ca certificates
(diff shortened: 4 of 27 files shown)


pkg/util/tls/tls_integration_test.go
```diff
@@ -0,0 +1,484 @@
...
+			conn, err := grpc.Dial(grpcHost, dialOptions...)
+			assert.NoError(t, err, tc.name)
+			require.NoError(t, err, tc.name)
+			require.NoError(t, err, tc.name)
+
+			client := grpc_health_v1.NewHealthClient(conn)
+
+			// TODO: Investigate why the client doesn't really receive the
+			// error about the bad certificate from the server side and just
+			// see connection closed instead
```
simonswine, line 154: Anyone having an idea how we can propagate the server error of "bad certificate" in the GRPC case?

pkg/util/tls/tls.go
```diff
@@ -13,9 +13,11 @@ import (
 
 // ClientConfig is the config for client TLS.
 type ClientConfig struct {
+	Enable             bool   `yaml:"tls_enable"`
```
pracucci, line 16: I would name it `tls_enabled` (more consistent with the rest of the config, even if in ancient times "enable" has been used too). Same comment the the CLI flag (`tls-enabled`).

As a consequence, I would rename the struct field to `Enabled`.

pkg/util/tls/tls.go
```diff
@@ -26,21 +28,28 @@ var (
...
+	f.StringVar(&cfg.ServerName, prefix+".tls-server-name", "", "Override the expected name on the server certificate.")
 	f.BoolVar(&cfg.InsecureSkipVerify, prefix+".tls-insecure-skip-verify", false, "Skip validating server certificate.")
 }
 
 // GetTLSConfig initialises tls.Config from config options
 func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
 	// no tls config given at all
-	if cfg.CertPath == "" && cfg.KeyPath == "" && cfg.CAPath == "" {
+	if cfg.CertPath != "" || cfg.KeyPath != "" || cfg.CAPath != "" || cfg.InsecureSkipVerify || cfg.ServerName != "" {
+		cfg.Enable = true
```
pracucci, line 43: We try to never manipulate configs at runtime (wherever possible). In this case I believe it's easy to avoid it, just using a local variable `enabled := cfg.Enable` which can be overridden here.

pkg/util/tls/tls_integration_test.go
```diff
@@ -0,0 +1,484 @@
+package tls_test
```
pracucci, line 1: I would keep it simple. Our `integration` tests are mainly focused on whenever you need to run the `cortex` binary, which is not the case of these tests. These tests are closer to unit tests then integration tests (even if you start an HTTP/gRPC server, but it's something we do in many unit tests) so I would keep them here.

```suggestion
package tls
```

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@
 * [ENHANCEMENT] When a tenant accesses the Alertmanager UI or its API, if we have valid `-alertmanager.configs.fallback` we'll use that to start the manager and avoid failing the request. #3073
 * [ENHANCEMENT] Add `DELETE api/v1/rules/{namespace}` to the Ruler. It allows all the rule groups of a namespace to be deleted. #3120
 * [ENHANCEMENT] Experimental Delete Series: Retry processing of Delete requests during failures. #2926
+* [ENHANCEMENT] Add flag `-<prefix>.tls-server-name` to require a specific server name instead of the hostname on the certificate. #3156
+* [ENHANCEMENT] Allow configuration of non-mutual TLS by using the flag `-<prefix>.tls-enable`. #3156
```
pracucci, line 49: Remember to update this too, once you rename the CLI flag.

pracucci (approved):
Amazing, thanks! Changes LGTM. I just left a couple of nits and we should be good to go 🚀

pkg/util/tls/tls_integration_test.go
simonswine: That makes all perfect sense. I think I can't easily change the package to `tls` as this would lead to an import path cycle through the used `grpcclient`.

pracucci: What's about a `test` subpackage?

pkg/util/tls/tls_integration_test.go
MichelHollands: https://github.com/grpc/grpc-go/issues/622#issuecomment-233107191 suggests that you need to use a WithBlock() call option to get the certificate error. This might not be possible with the existing clients.

pkg/util/tls/tls_integration_test.go
simonswine: Very good point, move it there

pkg/util/tls/tls_integration_test.go
simonswine: I will have to investigate a bit more but I have found that issue and tried `WithBlock()` and had the request being blocked for a while rather than having any change of error behaviour. I will take another look tomorrow

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@
 * [ENHANCEMENT] Experimental Delete Series: Retry processing of Delete requests during failures. #2926
 * [ENHANCEMENT] Improve performance of QueryStream() in ingesters. #3177
 * [ENHANCEMENT] Modules included in "All" target are now visible in output of `-modules` CLI flag. #3155
+* [ENHANCEMENT] Add flag `-<prefix>.tls-server-name` to require a specific server name instead of the hostname on the certificate. #3156
+* [ENHANCEMENT] Allow configuration of non-mutual TLS by using the flag `-<prefix>.tls-enabled`. #3156
```
pracucci, line 59: I've just merged a PR to cut the CHANGELOG because of the upcoming `1.4.0` release. Could you rebase `master` and move your CHANGELOG entry to the top, under `## master / unreleased`, please?

pkg/util/tls/tls.go
```diff
@@ -26,21 +28,29 @@ var (
 
 // RegisterFlagsWithPrefix registers flags with prefix.
 func (cfg *ClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.BoolVar(&cfg.Enabled, prefix+".tls-enabled", false, "Enable TLS in the client. This flag is implicitly enabled when any other TLS flag is set. This flag will only be respected by the GRPC client, as for the HTTP client the scheme used in the URL will take precedence.")
```
pstibrany, line 31: > This flag is implicitly enabled when any other TLS flag is set.

I assume this is for backward-compatibility, but it seems weird that using `-client.tls-enabled=false` will not be respected, if used with other tls flags.

We *could* make this a string variable and differentiate between empty string, "true" or "false" (or "enabled"/"disabled"). I don't like this idea, but don't have a better one.

pracucci: > I assume this is for backward-compatibility, but it seems weird that using -client.tls-enabled=false will not be respected, if used with other tls flags.

I haven't considered it, but you're right.

> We could make this a string variable and differentiate between empty string, "true" or "false" (or "enabled"/"disabled"). I don't like this idea, but don't have a better one.

But then in the config file reference documentation the default would be an empty string, which is weird, while having it documented as "false" would be better.

pstibrany: > > We could make this a string variable and differentiate between empty string, "true" or "false" (or "enabled"/"disabled"). I don't like this idea, but don't have a better one.
> 
> But then in the config file reference documentation the default would be an empty string, which is weird, while having it documented as "false" would be better.

Yes, it's bad solution. Personally I think we should require explicit use of `tls-enabled` flag, even though it's a breaking change.

pstibrany: > Yes, it's bad solution. Personally I think we should require explicit use of tls-enabled flag, even though it's a breaking change.

But one that's discovered and fixed quickly.

pracucci: I've raised a [public question in the CNCF Slack](https://cloud-native.slack.com/archives/CCYDASBLP/p1600339076337200).

pracucci: According to the current status of the Slack discussion, the general consensus is:
1. OK to introduce a breaking change (as Peter suggested)
2. Explicitly mention in `docs/configuration/v1-guarantees.md` that TLS support is experimental (there's a list of experimental features in the `Experimental features` section)

@simonswine WDYT?

simonswine: > 1. OK to introduce a breaking change (as Peter suggested)

While the string based approach is a bit hacky, my main concern is more that a `tls-enable=false` some scenarios will  be quite counter intuitive:

- I am specifying certs+keys, but no `tls-enable=true` => cortex exits with 1. This is not how other TLS enabled software works

- We have someone configuring `tls-enable=false`. But still the HTTP client is not really there to make the call. It is the request that will determine through the URL scheme if it is either `http://` or `https://`.

> 2. Explicitly mention in `docs/configuration/v1-guarantees.md` that TLS support is experimental (there's a list of experimental features in the `Experimental features` section)

This sounds reasonable, happy to add that

pstibrany: I think that:

- we should remove "tls_enabled" flag from `tls.ClientConfig` struct.

- for grpc clients, we should put "tls_enabled" bool flag into `grpcclient.Config`, and merge `grpcclient.ConfigWithTLS` with `grpcclient.Config`. When tls_enabled is false, TLS is not configured for grpc connection at all. We have some places in the code where we use `grpcclient.Config` (non-TLS version), which should be fixed, or TLS config put into `grpcclient.Config` directly.

- for HTTP clients (seems to be only ruler in Cortex), they will use TLS based on the scheme, with TLS settings coming from `tls.ClientConfig`.

- etcd client which has its own config should embed `tls.ClientConfig` instead, but keep `tls_enabled` flag.

pstibrany: > * I am specifying certs+keys, but no `tls-enable=true` => cortex exits with 1. This is not how other TLS enabled software works

TLS settings should only be consulted when needed (if enabled for grpc/etcd, or always for HTTP clients).

pstibrany: > > 1. Explicitly mention in `docs/configuration/v1-guarantees.md` that TLS support is experimental (there's a list of experimental features in the `Experimental features` section)
> 
> This sounds reasonable, happy to add that

Will send a separate PR to make it into Cortex 1.4.

simonswine: @pstibrany: I am currently looking into implementing your suggestions:

> * we should remove "tls_enabled" flag from `tls.ClientConfig` struct.
> * for grpc clients, we should put "tls_enabled" bool flag into `grpcclient.Config`, and merge `grpcclient.ConfigWithTLS` with `grpcclient.Config`. When tls_enabled is false, TLS is not configured for grpc connection at all. We have some places in the code where we use `grpcclient.Config` (non-TLS version), which should be fixed, or TLS config put into `grpcclient.Config` directly.

That's where I am currently at and this has come up so far:

- We have at least a GRPC client that has TLS currently enabled by default as part of the big table support:

https://github.com/cortexproject/cortex/blob/067412e4e79dd1abfe4fbc2ec74e3a13407c3b9d/pkg/chunk/gcp/bigtable_index_client.go#L38

- We seem to use sometimes `tls.ClientConfig` that is used later to generate GRPCConfigs from them e.g.:

https://github.com/cortexproject/cortex/blob/067412e4e79dd1abfe4fbc2ec74e3a13407c3b9d/pkg/ruler/ruler.go#L49

I just wanted to make sure you want me to go forward and for example:
- figure out a way so that we can have the GCP Bigtable with TLS enabled by default
- Add a `TLSEnabled bool` flag to every config part that contains a `tls.ClientConfig`

pstibrany: > I just wanted to make sure you want me to go forward and for example:

I still think this is a way to go. 

> 
> * figure out a way so that we can have the GCP Bigtable with TLS enabled by default

This can be achieved by overriding the config default in after this line: https://github.com/cortexproject/cortex/blob/067412e4e79dd1abfe4fbc2ec74e3a13407c3b9d/pkg/chunk/gcp/bigtable_index_client.go#L54

> * Add a `TLSEnabled bool` flag to every config part that contains a `tls.ClientConfig`

I'd say we should replace this with `grpcclient.ConfigWithTLS` instead. WDYT?

Thanks a lot for taking care of this issue!

pstibrany: > This can be achieved by overriding the config default after this line:
> https://github.com/cortexproject/cortex/blob/067412e4e79dd1abfe4fbc2ec74e3a13407c3b9d/pkg/chunk/gcp/bigtable_index_client.go#L54
> 

Sorry, I misremembered how we did it last time :) 

Take a look at this method: https://github.com/cortexproject/cortex/blob/85f93ff4abd7af81a5eeeae0e07379dd6ff2071d/pkg/chunk/table_provisioning.go#L33

and how it uses `cfg.ProvisionedWriteThroughput` as default value, which is then set by caller:

https://github.com/cortexproject/cortex/blob/85f93ff4abd7af81a5eeeae0e07379dd6ff2071d/pkg/chunk/purger/table_provisioning.go#L18

pstibrany: (I think the same technique should work here, thanks to default value of TLS enabled flag being `false)

simonswine: That sounds all good :+1: 

@pstibrany thanks for the "flags default override example", will keep you posted

pkg/util/tls/tls.go
```diff
@@ -74,12 +72,27 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
 	return config, nil
 }
 
+type ClientConfigWithEnabled struct {
+	ClientConfig `yaml:",inline"`
+	Enabled      bool `yaml:"tls_enabled"`
```
pstibrany, line 77: I expected to see `Enabled` flag in `grpcclient.Config`, and then using that instead of current `ClientConfigWithEnabled`. WDYT? (It may require passing `tlsEnabled` flag to `GetGRPCDialOptions`)

pkg/util/tls/tls.go
```diff
@@ -74,12 +72,27 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
...
+func (cfg *ClientConfigWithEnabled) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	cfg.ClientConfig.RegisterFlagsWithPrefix(prefix, f)
+	f.BoolVar(&cfg.Enabled, prefix+".tls-enabled", false, "Enable TLS in the GRPC client. This flag is needs to be enabled when any other TLS flag is set. This flag will only be respected by the GRPC client, as for the HTTP client the scheme used in the URL will take precedence.")
+}
+
 // GetGRPCDialOptions creates GRPC DialOptions for TLS
-func (cfg *ClientConfig) GetGRPCDialOptions() ([]grpc.DialOption, error) {
-	if tlsConfig, err := cfg.GetTLSConfig(); err != nil {
+func (cfg *ClientConfigWithEnabled) GetGRPCDialOptions() ([]grpc.DialOption, error) {
+	tlsConfig, err := cfg.GetTLSConfig()
```
pstibrany, line 88: This can return TLS-related error even if TLS is not enabled. I don't think we should let that happen.

simonswine: This should be resolved as of 342e17c592e6c63c3d01227732a7ea0ebdb27111

pkg/chunk/gcp/bigtable_index_client.go
```diff
@@ -52,6 +52,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.TableCacheExpiration, "bigtable.table-cache.expiration", 30*time.Minute, "Duration to cache tables before checking again.")
 
 	cfg.GRPCClientConfig.RegisterFlagsWithPrefix("bigtable", f)
+	// This overrides our default from TLS disabled to TLS enabled
+	cfg.GRPCClientConfig.TLS.Enabled = true
```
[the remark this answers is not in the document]
simonswine: Thanks for making that clearer, I think that should do it: https://github.com/cortexproject/cortex/pull/3156/commits/65f5bea09e462005c8e60a66436a3729274506a4

pkg/util/tls/tls.go
simonswine: I actually have implemented it that way the first time around.

It ended up being quite tedious in terms of duplicated work and passing the boolean around. 

Esp. in places where there is not a full `grpcclient.Config` used, like here:

https://github.com/cortexproject/cortex/blob/342e17c592e6c63c3d01227732a7ea0ebdb27111/pkg/querier/querier.go#L63
or there:
https://github.com/cortexproject/cortex/blob/342e17c592e6c63c3d01227732a7ea0ebdb27111/pkg/querier/blocks_store_replicated_set.go#L40

Wdyt if we maybe call it `tls.GRPCClientConfig` instead of `tls.ClientConfigWithEnabled` as this is more accurate in the way we use it

pkg/util/tls/tls.go
```diff
@@ -74,12 +72,27 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
...
 
+type ClientConfigWithEnabled struct {
+	ClientConfig `yaml:",inline"`
+	Enabled      bool `yaml:"tls_enabled"`
+}
+
+// RegisterFlagsWithPrefix registers flags with prefix.
+func (cfg *ClientConfigWithEnabled) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	cfg.ClientConfig.RegisterFlagsWithPrefix(prefix, f)
+	f.BoolVar(&cfg.Enabled, prefix+".tls-enabled", cfg.Enabled, "Enable TLS in the GRPC client. This flag is needs to be enabled when any other TLS flag is set. This flag will only be respected by the GRPC client, as for the HTTP client the scheme used in the URL will take precedence.")
```
pracucci, line 83:
```suggestion
	f.BoolVar(&cfg.Enabled, prefix+".tls-enabled", cfg.Enabled, "Enable TLS in the GRPC client. This flag needs to be enabled when any other TLS flag is set. This flag will only be respected by the GRPC client, as for the HTTP client the scheme used in the URL will take precedence.")
```

CHANGELOG.md
```diff
@@ -84,6 +84,8 @@
 * [ENHANCEMENT] Added `cortex_alertmanager_config_last_reload_successful_seconds` metric to show timestamp of last successful AM config reload. #3289
 * [ENHANCEMENT] Blocks storage: reduced number of bucket listing operations to list block content (applies to newly created blocks only). #3363
 * [ENHANCEMENT] Ruler: Include the tenant ID on the notifier logs. #3372
+* [ENHANCEMENT] Add flag `-<prefix>.tls-server-name` to require a specific server name instead of the hostname on the certificate. #3156
+* [ENHANCEMENT] Allow configuration of non-mutual TLS by using the flag `-<prefix>.tls-enabled`. #3156
```
pracucci, line 88: This looks a `[CHANGE]` to me, because TLS is no more implicitly enabled when other flags are set. You should mark it as a `[CHANGE]` and clarify it.

pkg/util/tls/tls.go
```diff
@@ -29,18 +30,15 @@ func (cfg *ClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet)
...
 	f.StringVar(&cfg.KeyPath, prefix+".tls-key-path", "", "Path to the key file for the client certificate. Also requires the client certificate to be configured.")
 	f.StringVar(&cfg.CAPath, prefix+".tls-ca-path", "", "Path to the CA certificates file to validate server certificate against. If not set, the host's root CA certificates are used.")
+	f.StringVar(&cfg.ServerName, prefix+".tls-server-name", "", "Override the expected name on the server certificate.")
 	f.BoolVar(&cfg.InsecureSkipVerify, prefix+".tls-insecure-skip-verify", false, "Skip validating server certificate.")
 }
 
 // GetTLSConfig initialises tls.Config from config options
 func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
-	// no tls config given at all
-	if cfg.CertPath == "" && cfg.KeyPath == "" && cfg.CAPath == "" {
```
pracucci, line 38: Isn't this a breaking change? What's the impact of this change for components that are directly using `ClientConfig`? Before we were retuning `nil` (TLS disabled) while now we do return `tls.Config{}` (TLS enabled).

pkg/util/tls/tls.go
pracucci, line 77: Could you move the `Enabled` field before `ClientConfig`, so that in the auto-generated config the `tls_enabled` will be documented before the other options?

pkg/util/grpcclient/grpcclient.go
```diff
@@ -27,6 +27,8 @@ type Config struct {
 
 	BackoffOnRatelimits bool               `yaml:"backoff_on_ratelimits"`
 	BackoffConfig       util.BackoffConfig `yaml:"backoff_config"`
+
+	TLS tls.ClientConfigWithEnabled `yaml:",inline"`
```
pracucci, line 31: This change is non an innocuous refactoring. The problem is that doing this change we add TLS to any gRPC client, even the ones for which we don't support it. For example BigTable client (see generated doc).

pkg/util/tls/tls.go
simonswine: My understanding is that for clients using `tls.Dial` which includes HTTP clients, the behaviour is the exact same if it is set to `nil` or `&tls.Config{}`:

https://github.com/golang/go/blob/master/src/crypto/tls/tls.go#L201

pkg/util/grpcclient/grpcclient.go
simonswine: This was done on on purpose as an outcome of this thread (https://github.com/cortexproject/cortex/pull/3156#discussion_r492013177), so there's an unified `grpcclient.Config`. We ensure that we enable TLS on the bigtable client by default, as TLS is on for bigtable by default in Cortex today.

I am not too sure if I caught every usage, but we should respect the TLS arguments (but I am not too sure if we want to support them according to your comment).

pkg/util/tls/tls.go
simonswine: I have done this in the latest master

CHANGELOG.md
simonswine: Updated

pkg/util/tls/tls.go
simonswine: Thanks for spotting this

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 * [CHANGE] Increased default `-<prefix>.redis.timeout` from `100ms` to `500ms`. #3301
 * [CHANGE] `cortex_alertmanager_config_invalid` has been removed in favor of `cortex_alertmanager_config_last_reload_successful`. #3289
 * [CHANGE] Query-frontend: POST requests whose body size exceeds 10MiB will be rejected. The max body size can be customised via `-frontend.max-body-size`. #3276
+* [CHANGE] Require explicit flag `-<prefix>.tls-enabled` to enable TLS in GRPC clients. Previously it was enough to specify a TLS flag to enable TLS validation. #3156
```
pracucci, line 52: We've just cut `1.6.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry to the top, under the "master / unreleased" section, please?

_Same comment applies to the CHANGELOG entry below._

pracucci: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

pstibrany:
I've rebased this patch on top of current master, and did the changes suggested in review comments:

- we should remove "tls_enabled" flag from tls.ClientConfig struct.

- for grpc clients, we should put "tls_enabled" bool flag into `grpcclient.Config`, and merge `grpcclient.ConfigWithTLS` with `grpcclient.Config`
- etcd client which has its own config should embed tls.ClientConfig instead, but keep tls_enabled flag.

I haven't updated ruler to use tls when communicating with AM, since https://github.com/cortexproject/cortex/pull/3752 is already doing that.

pkg/util/tls/tls.go
```diff
@@ -29,18 +30,15 @@ func (cfg *ClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet)
 	f.StringVar(&cfg.CertPath, prefix+".tls-cert-path", "", "Path to the client certificate file, which will be used for authenticating with the server. Also requires the key path to be configured.")
 	f.StringVar(&cfg.KeyPath, prefix+".tls-key-path", "", "Path to the key file for the client certificate. Also requires the client certificate to be configured.")
 	f.StringVar(&cfg.CAPath, prefix+".tls-ca-path", "", "Path to the CA certificates file to validate server certificate against. If not set, the host's root CA certificates are used.")
+	f.StringVar(&cfg.ServerName, prefix+".tls-server-name", "", "Override the expected name on the server certificate.")
 	f.BoolVar(&cfg.InsecureSkipVerify, prefix+".tls-insecure-skip-verify", false, "Skip validating server certificate.")
 }
 
 // GetTLSConfig initialises tls.Config from config options
 func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
```
pracucci, line 38: This function is also used by `pkg/configs/client/client.go`. Before it was returning `nil` if TLS was disabled, while now it always returns a config. Could you check the impact on the configs client? I've the feeling we're introducing a regression here.

pracucci (commented):
Thanks a lot @pstibrany for taking over this work 👏  A part form a couple of non blocking nits, I've the feeling we're introducing a regression in the configs client. Could you check it, please?

pkg/util/tls/tls.go
pstibrany: If operator doesn't set any TLS options, then I think that returned structure will work fine even when using TLS in configs.
Defaults we are setting are good for client-side use. (For server-side, `Certificates` shouldn't be nil).

pstibrany: I just tested http client with no `Transport` and `Transport: &http.Transport{TLSClientConfig: tlsCfg}` (where `tlsCfg` is result of calling `GetTLSConfig` on empty `cortex_tls.ClientConfig`) against public site using https, and I'm getting the same 200 response in both cases.

pstibrany: I think integration test in `tls_integration_test.go` covers this too.

pracucci (approved):
Thanks Peter for checking it. LGTM.

pstibrany (approved):

PR #3164, opened 2020-09-11
add store method for getting fetcher for a chunk

sandeepsukhani:
**What this PR does**:
Adds a new store method called `GetChunkFetcher` which helps to get a chunk fetcher for a chunk.
I have used end time of the chunk to find the store for it.

This helps with a use-case where Loki has a bunch of chunk ids which would be converted to chunkref using `ParseExternalKey` and passed to this method to get a fetcher for it.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/composite_store.go
+++ b/pkg/chunk/composite_store.go
@@ -176,7 +176,7 @@ func (c compositeStore) GetChunkRefs(ctx context.Context, userID string, from, t
 }
 
 func (c compositeStore) GetChunkFetcher(tm model.Time) *Fetcher {
-	// next, find the schema with the lowest start _after_ tm
+	// find the schema with the lowest start _after_ tm
 	j := sort.Search(len(c.stores), func(j int) bool {
 		return c.stores[j].start > tm
 	})
--- a/pkg/chunk/composite_store.go
+++ b/pkg/chunk/composite_store.go
@@ -184,7 +184,7 @@ func (c compositeStore) GetChunkFetcher(tm model.Time) *Fetcher {
 	// reduce it by 1 because we want a schema with start <= tm
 	j--
 
-	if j >= 0 && len(c.stores) > j {
+	if 0 <= j && j < len(c.stores) {
 		return c.stores[j].GetChunkFetcher(tm)
 	}
 
--- a/pkg/chunk/composite_store.go
+++ b/pkg/chunk/composite_store.go
@@ -32,6 +32,7 @@ type Store interface {
 	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([][]Chunk, []*Fetcher, error)
 	LabelValuesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string, labelName string) ([]string, error)
 	LabelNamesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string) ([]string, error)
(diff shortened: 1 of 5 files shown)


pkg/chunk/composite_store.go
```diff
@@ -174,6 +175,22 @@ func (c compositeStore) GetChunkRefs(ctx context.Context, userID string, from, t
...
+func (c compositeStore) GetChunkFetcher(chunk Chunk) *Fetcher {
+	// next, find the schema with the lowest start _after_ through
+	j := sort.Search(len(c.stores), func(j int) bool {
+		return c.stores[j].start > chunk.Through
+	})
+
+	// reduce it by 1 because we want a schema with start <= through
+	j--
+
+	if j >= 0 && len(c.stores) > j {
```
pstibrany, line 187:
```suggestion
	if 0 <= j && j < len(c.stores) {
```

(Same condition, but it's visually easier to see that `j` is bounded by 0 and `len(c.stores)`)

pkg/chunk/composite_store.go
```diff
@@ -174,6 +175,22 @@ func (c compositeStore) GetChunkRefs(ctx context.Context, userID string, from, t
 	return chunkIDs, fetchers, err
 }
 
+func (c compositeStore) GetChunkFetcher(chunk Chunk) *Fetcher {
```
pstibrany, line 178: Can we only pass single timestamp instead of `Chunk`?

(It is also not clear what should be returned if chunk crossed the boundary between two schemas)

pstibrany (approved):
LGTM

pkg/chunk/composite_store.go
sandeepsukhani: It sounds good, I wanted to do it like that but the code would change based on whether you are sending start time of chunk or end time of chunk. We will have to make it clearer by renaming the function or something that you are supposed to send end time of chunk otherwise the caller could do a mistake by sending chunk start time instead of the end time.

What do you think?

pstibrany: > It sounds good, I wanted to do it like that but the code would change based on whether you are sending start time of chunk or end time of chunk. We will have to make it clearer by renaming the function or something that you are supposed to send end time of chunk otherwise the caller could do a mistake by sending chunk start time instead of the end time. Maybe a comment and a relevant function parameter name like `chunkEndTime` should be enough?
> 
> What do you think?

I think the reason to use timestamp as parameter instead of full chunk would be to remove this question from the table *in the method*. There should still be some consistency between the calls, but now it would be up to client to be consistent about sending start or end timestamps.

I am wondering what difference does it really make whether we pass start or end time to the method though -- In both cases, the fetcher must be able to fetch the chunk, right?

sandeepsukhani: As discussed, I have added validation for not sharing the same date with multiple periodic configs.

pkg/chunk/composite_store.go
```diff
@@ -174,6 +175,22 @@ func (c compositeStore) GetChunkRefs(ctx context.Context, userID string, from, t
 	return chunkIDs, fetchers, err
 }
 
+func (c compositeStore) GetChunkFetcher(tm model.Time) *Fetcher {
+	// next, find the schema with the lowest start _after_ tm
```
pracucci, line 179:
```suggestion
	// find the schema with the lowest start _after_ tm
```

pracucci (approved):
LGTM!

PR #3167, opened 2020-09-11
Fix the querier returning 500 status code on query range limit errors

### LINKED ISSUE #3163: QueryRange: Catalogs range validation as an internal error
Time ranges for range queries are validated against a per-tenant limit.

https://github.com/cortexproject/cortex/blob/e0663b763a53352ae0d7cac0768a657fe5b80fbd/pkg/querier/querier.go#L285-L289

Within Prometheus API, the `query_range` endpoint does all of its validation upfront - meaning that incorrect input errors (400s) are validated before the query is ran.

https://github.com/prometheus/prometheus/blob/a282d2509971174301408c5a5f96946c478dfb0f/web/api/v1/api.go#L387-L452

As it is, Cortex will return a 500 for the validation of this limit which seems incorrect.  Any thoughts?

I'm not sure upstream will accept changes to encapsulate custom validation as part of the API, the other options is to copy the validation code of the code of this particular endpoint which seems unideal.

An example of the error:

```json
{
  "status": "error",
  "errorType": "internal",
  "error": "expanding series: the query time range exceeds the limit (query length: 4414h5m0s, limit: 768h0m0s)"
}
```

gotjosh:
Fixes #3163 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@
 * [BUGFIX] Index page no longer shows links that are not valid for running Cortex instance. #3133
 * [BUGFIX] Configs: prevent validation of templates to fail when using template functions. #3157
 * [BUGFIX] Configuring the S3 URL with an `@` but without username and password doesn't enable the AWS static credentials anymore. #3170
+* [BUGFIX] Limit errors on ranged queries (`api/v1/query_range`) no longer return a status code `500` but `422` instead. #3167
 
 ## 1.3.0 / 2020-08-21
 
--- a/integration/querier_test.go
+++ b/integration/querier_test.go
@@ -553,6 +553,7 @@ func TestQuerierWithChunksStorage(t *testing.T) {
 	querierFlags := mergeFlags(flags, map[string]string{
 		"-store.index-cache-read.memcached.addresses":  "dns+memcached0:11211",
 		"-store.index-cache-write.memcached.addresses": "dns+memcached1:11211",
+		"-store.max-query-length":                      "240h",
 	})
 
 	querier := e2ecortex.NewQuerier("querier", consul.NetworkHTTPEndpoint(), querierFlags, "")
@@ -592,9 +593,33 @@ func TestQuerierWithChunksStorage(t *testing.T) {
 
 	wg.Wait()
 
+	c, err := e2ecortex.NewClient("", querier.HTTPEndpoint(), "", "", "user-0")
+	require.NoError(t, err)
+
+	// Ensure limit errors on ranged queries don't return 500s.
+	start := now.Add(-264 * time.Hour)
+	end := now
+	step := 264 * time.Hour
+
+	r, body, err := c.QueryRangeRaw("series_1", start, end, step)
+	require.NoError(t, err)
+	require.Equal(t, 422, r.StatusCode)
+	expected := `
+	{
+		"error":"expanding series: the query time range exceeds the limit (query length: 264h5m0s, limit: 240h0m0s)",
+		"errorType":"execution",
+		"status":"error"
+	}
+	`
+	require.JSONEq(t, expected, string(body))
+
 	// Ensure no service-specific metrics prefix is used by the wrong service.
 	assertServiceMetricsPrefixes(t, Distributor, distributor)
 	assertServiceMetricsPrefixes(t, Ingester, ingester)
 	assertServiceMetricsPrefixes(t, Querier, querier)
 	assertServiceMetricsPrefixes(t, TableManager, tableManager)
+}
+
+func formatTime(t time.Time) string {
+	return strconv.FormatFloat(float64(t.Unix())+float64(t.Nanosecond())/1e9, 'f', -1, 64)
 }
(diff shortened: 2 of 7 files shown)


integration/querier_test.go
```diff
@@ -592,9 +593,33 @@ func TestQuerierWithChunksStorage(t *testing.T) {
...
+	require.JSONEq(t, expected, string(body))
+
 	// Ensure no service-specific metrics prefix is used by the wrong service.
 	assertServiceMetricsPrefixes(t, Distributor, distributor)
 	assertServiceMetricsPrefixes(t, Ingester, ingester)
 	assertServiceMetricsPrefixes(t, Querier, querier)
 	assertServiceMetricsPrefixes(t, TableManager, tableManager)
 }
+
+func formatTime(t time.Time) string {
```
pstibrany, line 623: Shall we expose function from e2ecortex package instead?

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
 * [BUGFIX] Ingester: If push request contained both valid and invalid samples, valid samples were ingested but not stored to WAL of the chunks storage. This has been fixed. #3067
 * [BUGFIX] Ruler: Config API would return both the `record` and `alert` in `YAML` response keys even when one of them must be empty. #3120
 * [BUGFIX] Index page now uses configured HTTP path prefix when creating links. #3126
+* [BUGFIX] Limit errors on ranged queries (`api/v1/query_range`) no longer return a status code `500` but `422` instead. #XXXX
```
pstibrany, line 49:
```suggestion
* [BUGFIX] Limit errors on ranged queries (`api/v1/query_range`) no longer return a status code `500` but `422` instead. #3167
```

pstibrany (approved):
LGTM, thank you!

pkg/querier/querier.go
```diff
@@ -68,6 +68,13 @@ type Config struct {
 	UseSecondStoreBeforeTime flagext.Time `yaml:"use_second_store_before_time"`
 }
 
+// UserError are errors caused by user input as opposed to errors executing the query.
+type UserError string
```
bboreham, line 72: I think it's great to add a specific type here, but most similar errors are currently using `httpgrpc.Errorf()`.
https://github.com/cortexproject/cortex/blob/5ede3635eb550e57ed22a8d0d48c8fe783036b54/pkg/querier/queryrange/query_range.go#L34-L36

Therefore I think you should either do the same here, or replace all similar ones with your new type.  
Perhaps it should go in package `validation`?

integration/querier_test.go
pracucci: I think so.

pkg/querier/querier.go
pracucci: > but most similar errors are currently using httpgrpc.Errorf()

I think we use `httpgrpc.Errorf()` when directly returning from a gRPC call (or it's parent is a direct gRPC call), but in this case it's a function called by the PromQL engine, which then wraps the error again, so it would look weird to me having a `httpgrpc.Errorf()` wrapped by a PromQL error.

I personally think that for this fix, having a dedicated error type is just fine and we could consider refactoring existing errors as a separate work. However, I would also argue that this error may be better placed in the `validation` package and maybe renamed to `LimitError` (`UserError` looks very generic to me).

gotjosh: > I think we use httpgrpc.Errorf() when directly returning from a gRPC call (or it's parent is a direct gRPC call), but in this case it's a function called by the PromQL engine, which then wraps the error again, so it would look weird to me having a httpgrpc.Errorf() wrapped by a PromQL error.

I was about to phrase that similarly as well - `NewQueryRange` is used by the engine directly.


> However, I would also argue that this error may be better placed in the validation package and maybe renamed to `LimitError` 

I think this is sensible.

integration/querier_test.go
gotjosh: This was never meant to be here, I've included it in the e2e package already - is never used in this file.

pracucci (approved):
LGTM, thanks!

pkg/querier/querier.go
bboreham: `httpgrpc.Errorf()` is used in non-gRPC situations, e.g. in `Distributor.Push()`.
To be clear, I hate httpgrpc, but I also hate inconsistency.

pracucci: I merged this PR because I didn't notice this comment, sorry! Let me open an issue.

pracucci: https://github.com/cortexproject/cortex/issues/3181

gotjosh: Apologies @bboreham, think I've missed that case - I'll take a deeper look and follow up on this.

bboreham: I don't understand what point you are making in "can't control the status code".
You added a case to the type switch not far above the line where it already decodes httpgrpc errors:
https://github.com/cortexproject/cortex/blob/f1ac40e05e9e9d8d4d16c61121b3a91848d4ec56/pkg/api/queryable.go#L42-L49

PR #3170, opened 2020-09-14
Fix @ in the S3 URL without username and password

### LINKED ISSUE #3034: IAM authentication support in ruler and alertmanager S3 client
I've been reported that the "chunks storage" S3 client (which is also used by the ruler and alertmanager) doesn't support IAM authentication, while the S3 client we use in the blocks storage does. We need investigate and eventually fix it.

_Longer shot would be re-using the blocks storage S3 client for the ruler and alertmanager as well, which is something discussed in https://github.com/cortexproject/cortex/issues/2262 and attempted in https://github.com/cortexproject/cortex/pull/2489._

pracucci:
**What this PR does**:
In this PR I'm upgrading `weaveworks/common` to bring in https://github.com/weaveworks/common/pull/198, which fixes an issue with the S3 URL contains `@` but no username and password.

Fixes #3034

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@
 * [BUGFIX] Index page now uses configured HTTP path prefix when creating links. #3126
 * [BUGFIX] Index page no longer shows links that are not valid for running Cortex instance. #3133
 * [BUGFIX] Configs: prevent validation of templates to fail when using template functions. #3157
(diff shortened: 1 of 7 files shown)


pstibrany (approved):
LGTM

PR #3173, opened 2020-09-14
Blocksconvert: store series and symbols on the disk to avoid using large heap

pstibrany:
**What this PR does**: This PR modifies `blocksconvert` to not keep details about all the series in memory, but put them into temporary files instead. Symbols and series are sorted in the file, and when iterating them during the index-building phase, they are merged and returned in sorted order.

This makes it possible to build large blocks with many series without using too much heap, at the cost of using more disk space.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tools/blocksconvert/builder/series_iterator.go
```diff
@@ -0,0 +1,165 @@
...
+	return si
+}
+
+func (sit *seriesIterator) buildHeap() {
+	for ix := 0; ix < len(sit.files); {
+		f := sit.files[ix]
+		next, err := f.hasNext()
+
+		if err != nil {
+			sit.errs.Add(err)
```
pracucci, line 32: Should we break the `for` iteration here?

tools/blocksconvert/builder/series_iterator.go
```diff
@@ -0,0 +1,165 @@
...
+func newSeriesIterator(files []*seriesFile) *seriesIterator {
+	si := &seriesIterator{
+		files: files,
+	}
+	si.buildHeap()
+	return si
+}
+
+func (sit *seriesIterator) buildHeap() {
+	for ix := 0; ix < len(sit.files); {
```
pracucci, line 27: I would add a comment that this `for` loop is used to remove files from which there's nothing more to read.

tools/blocksconvert/builder/series_iterator.go
```diff
@@ -0,0 +1,165 @@
...
+		if !next {
+			sit.errs.Add(f.close())
+			sit.files = append(sit.files[:ix], sit.files[ix+1:]...)
+			continue
+		}
+
+		ix++
+	}
+
+	for ix := len(sit.files) - 1; ix >= 0; ix-- {
```
pracucci, line 44: I would add a comment to explain what it does.

tools/blocksconvert/builder/series_iterator.go
```diff
@@ -0,0 +1,165 @@
...
+		files[ix], files[smallest] = files[smallest], files[ix]
+		heapifySeries(files, smallest)
+	}
+}
+
+type seriesFile struct {
+	f   *os.File
+	dec *gob.Decoder
+
+	next bool
```
pracucci, line 113: Maybe renaming it to `serIsNext` would be more clear?

tools/blocksconvert/builder/series_iterator.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (sf *seriesFile) peek() series {
+	if !sf.next {
+		panic("no next symbol")
+	}
+
+	return sf.ser
+}
+
+func (sf *seriesFile) advance() series {
```
pracucci, line 158: Maybe `pop()` is more clear?

tools/blocksconvert/builder/series_iterator.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (sit *seriesIterator) Close() error {
+	var errs errors.MultiError
+	for _, f := range sit.files {
+		errs.Add(f.close())
+	}
+	return errs.Err()
+}
+
+// heapify series at given index. root is the minimum (only labels are compared)
```
pracucci, line 89: I would improve this comment.

> root is the minimum

What do you mean? Do you mean that file at index 0 has the first series on sorting order?

tools/blocksconvert/builder/symbols_iterator.go
```diff
@@ -0,0 +1,176 @@
...
+
+		ix++
+	}
+
+	for ix := len(sit.files) - 1; ix >= 0; ix-- {
+		heapifySymbols(sit.files, ix)
+	}
+}
+
+// Advances iterator forward, and returns next element. If there is no next element, returns false.
```
pracucci, line 52:
```suggestion
// Next advances iterator forward, and returns next element. If there is no next element, returns false.
```

tools/blocksconvert/builder/symbols_iterator.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/prometheus/prometheus/tsdb/errors"
+)
+
+type symbolsIterator struct {
+	files []*symbolsFile
+	errs  errors.MultiError
+
+	// To avoid returning duplicates, we remember last returned symbol.
+	last     string
+	returned bool // Used to avoid skipping first "" symbol.
```
pracucci, line 18: Instead of `returned` have you considered `last *string` so that we use `nil` value to distinguish it?

tools/blocksconvert/builder/tsdb.go
```diff
@@ -166,31 +172,31 @@ func (d *tsdbBuilder) buildSingleSeries(metric labels.Labels, cs []chunk.Chunk)
...
-	d.addSeries(metric, chs, seriesSamples, minTime, maxTime)
-	return nil
-}
-
-func (d *tsdbBuilder) addSeries(m labels.Labels, cs []chunks.Meta, samples uint64, minTime, maxTime int64) {
-	sl := d.symbolsMap.toSymbolLabels(m)
-
-	d.seriesMu.Lock()
-	defer d.seriesMu.Unlock()
+	inmem, err := d.series.addSeries(metric, chs, seriesSamples, minTime, maxTime)
```
pracucci, line 175: Design wise, looks a bit weird to me that `addSeries()` returns the count of total series. At the cost of taking the read lock another time, I would probably add `d.series.count()` (or something similar) to update the `seriesInMemory` metric.

pracucci (approved):
Looks flawless to me, good job! I left minor comments, but consider it already approved 👍 

A couple of things:
1. The comments I've left to `seriesIterator` also apply to `symbolsIterator`.
2. I will reach out offline to better understanding the heapify functions (everything else looks easy to understand)

I initially was scared by the PR size, but I have to say the code is very easy to follow and understand (except the heapify which I find a bit tricky).

tools/blocksconvert/builder/series_iterator.go
pstibrany: Correct. We can also `return` directly, since `Next()` will report `false` on any error.

tools/blocksconvert/builder/series_iterator.go
pstibrany: I kept `next` but renamed field after to `nextSeries` and `nextSymbol`. It's a small struct with handful of tiny methods, I think this is enough.

tools/blocksconvert/builder/series_iterator.go
pstibrany: Abstracted generic `heapify` and added more comments. It's classic implementation of (min) heap sort.

pstibrany:
Thanks for review Marco. I've addressed your feedback, and added more comments about heapify. Please take a look again.

pracucci (approved):
Very good job! Looks perfect to me 👏

PR #3175, opened 2020-09-14
Fixed double tracing spans

### LINKED ISSUE #3168: Ingester emitting two identical trace spans for every Push call
Identical except for slight differences in timing.
Do we add the middleware twice?

image

pracucci:
**What this PR does**:
@bboreham reported in #3168 that we have some double tracing spans, due to how we inject the tracer for Thanos. According to my investigation, the issue is caused by the usage of `grpc_tracing.UnaryServerInterceptor()` and `grpc_tracing.StreamServerInterceptor()` (respectively called by Thanos `UnaryServerInterceptor()` and `StreamServerInterceptor()`) which create a new span.

This PR should fix it. I've manually tested it (both for unary and stream interceptors) and looks fixing the issue.

Fixes #3168

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -10,9 +10,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


bboreham (approved):

pstibrany (approved):
LGTM

Issue #3171, opened 2020-09-14
Slow ingester Push operations


bboreham:
This screengrab is from Jaeger, requesting all Push operations longer than 100ms on a single ingester in an hour:

![image](https://user-images.githubusercontent.com/8125524/93059392-88c70f00-f668-11ea-8785-c6649699f8e6.png)

The vertical clustering suggests to me that this is caused by blocking, in one case over a second.
I'm not yet sure of the cause, but my preferred theory is a lock held by a query.  Hence #3169 

Considering #3093 as a possible cause, I added a span for `userState.createSeriesWithFingerprint()` but it hasn't showed up in any of the slow traces I've looked at.

bboreham:
This symptom was considerably improved by #3177 when I tried it.
Suggests that Go is blocking until memory is made available by GC: although Go's stop-the-world pauses are very short, individual goroutines can get hit.

bboreham:
OK now I think I've got it.  Result of rolling out #3191:

![image](https://user-images.githubusercontent.com/8125524/93376851-20d41c80-f852-11ea-9509-f5f7729f318b.png)

PR #3177, opened 2020-09-14
Re-use memory for chunks in ingester QueryStream

bboreham:
This improves performance of queries and reduces garbage-collection.

It works by making an array of reusable objects, so that they all stick around until the batch-send function is called.
If there are many batches then the savings can be considerable.

Benchmark (using 2000 series so about 15 batches): 
```
before:
BenchmarkIngester_QueryStream-4              154           7448975 ns/op        16886320 B/op      32108 allocs/op
after:
BenchmarkIngester_QueryStream-4              187           6027117 ns/op         3771120 B/op      28647 allocs/op
```

Relates to #3171, but I don't think this will fix the issue.

**Checklist**
- [x] Tests updated
- NA Documentation added
- [x] `CHANGELOG.md` updated


pracucci (approved):
LGTM

pstibrany (approved):
LGTM.

PR #3178, opened 2020-09-14
Cancel abandoned operations in ReplicationSet.Do()

### LINKED ISSUE #3169: Abandoned QueryStream calls to ingesters should get canceled
Queries are satisfied when (N-1) of the calls return, but the Nth one can still be doing work and chewing up resources.

Illustrative trace:
image

I think the cancellation could be done in `ReplicationSet.Do()`

bboreham:
Generally it will start a set of operations in parallel, and return once enough of them have succeeded. Pass down a context to each one, and cancel that context when ReplicationSet.Do() returns.

Fixes #3169

**Checklist**
- NA Tests updated
- NA Documentation added
- [x] `CHANGELOG.md` updated


CHANGELOG.md
```diff
@@ -62,6 +62,7 @@
 * [BUGFIX] Index page no longer shows links that are not valid for running Cortex instance. #3133
 * [BUGFIX] Configs: prevent validation of templates to fail when using template functions. #3157
 * [BUGFIX] Configuring the S3 URL with an `@` but without username and password doesn't enable the AWS static credentials anymore. #3170
+* [BUGFIX] No-longer-needed ingester operations from queries are now canceled. #3178
```
pracucci, line 65:
```suggestion
* [BUGFIX] No-longer-needed ingester operations from queriers are now canceled. #3178
```

pracucci (approved):
Makes sense to me, and applies only to the read path (we can't do the same for the write path, but isn't touched).

bboreham:
> we can't do the same for the write path, but isn't touched

Note the top-level context does get canceled when the request returns at top level, so the write path takes precautions to avoid being impacted by that: https://github.com/cortexproject/cortex/blob/3ef425d741a1a093c3de8b3073251a7c105a24b4/pkg/distributor/distributor.go#L535-L536

The reason why I wanted this PR in the read path is that requests in the querier can run for a while after ingester requests, e.g. evaluating PromQL over the data.

CHANGELOG.md
bboreham: I realise it's a small thing, but I meant to write "queries".  The same code is called from the ruler, for instance.

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@
 * [BUGFIX] Configs: prevent validation of templates to fail when using template functions. #3157
 * [BUGFIX] Configuring the S3 URL with an `@` but without username and password doesn't enable the AWS static credentials anymore. #3170
 * [BUGFIX] Limit errors on ranged queries (`api/v1/query_range`) no longer return a status code `500` but `422` instead. #3167
+* [BUGFIX] No-longer-needed ingester operations from queries are now canceled. #3178
```
pracucci, line 79: I've just merged a PR to cut the CHANGELOG because of the upcoming `1.4.0` release. Could you rebase `master` and move your CHANGELOG entry to the top, under `## master / unreleased`, please?

pstibrany, line 79: Typo in `queriers`.

pstibrany (approved):
LGTM, thanks!

CHANGELOG.md
pracucci: I left the same comment, but Bryan correctly told it's intended to be "queries" because it affects both queriers and rulers.

pstibrany: Well, it confused two of us, so perhaps better wording like "operations caused by queries" would improve it?

PR #3185, opened 2020-09-15
Alertmanager Config API: Do not allow empty configurations

### LINKED ISSUE #3142: Alertmanager: Wrong input can wipe config
These are two examples I've found of POSTing wrong input to `/api/v1/alerts` that results in Alertmanager returning `2XX` and setting the config to an empty string:

- Wrong indentation of content
```
alertmanager_config: |
global: 
  smtp_from: 'andreas.opferkuch@grafana.com' 
route: 
  receiver: asdf
receivers: 
  - name: asdf
    email_configs: 
    - to: 'andreas.opferkuch@grafana.com' 
```

- Invalid field name (note the typo in `aalertmanager_config`)
```
aalertmanager_config: |
  global: 
    smtp_from: 'andreas.opferkuch@grafana.com' 
  route: 
    receiver: asdf
  receivers: 
    - name: asdf
      email_configs: 
      - to: 'andreas.opferkuch@grafana.com' 
```

gotjosh:
**What this PR does**:

Fixes #3142 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,7 @@
 * [BUGFIX] Configs: prevent validation of templates to fail when using template functions. #3157
 * [BUGFIX] Configuring the S3 URL with an `@` but without username and password doesn't enable the AWS static credentials anymore. #3170
 * [BUGFIX] Limit errors on ranged queries (`api/v1/query_range`) no longer return a status code `500` but `422` instead. #3167
+* [BUGFIX] Experimental Alertmanager: Do not allow empty Alertmanager configurations to be submitted through the configuration API. #XXXX
 
 ## 1.3.0 / 2020-08-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
 * [BUGFIX] Ruler: directories in the configured `rules-path` will be removed on startup and shutdown in order to ensure they don't persist between runs. #3195
 * [BUGFIX] Handle hash-collisions in the query path. #3192
-* [BUGFIX] Check for postgres rows errors. #3197
+* [BUGFIX] Experimental Alertmanager: Do not allow empty Alertmanager configurations or bad template filenames to be submitted through the configuration API. #3185
 
 ## 1.4.0-rc.0 in progress
 
@@ -94,7 +94,6 @@
 * [BUGFIX] Configs: prevent validation of templates to fail when using template functions. #3157
 * [BUGFIX] Configuring the S3 URL with an `@` but without username and password doesn't enable the AWS static credentials anymore. #3170
 * [BUGFIX] Limit errors on ranged queries (`api/v1/query_range`) no longer return a status code `500` but `422` instead. #3167
-* [BUGFIX] Experimental Alertmanager: Do not allow empty Alertmanager configurations to be submitted through the configuration API. #XXXX
 
 ## 1.3.0 / 2020-08-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
 * [BUGFIX] Ruler: directories in the configured `rules-path` will be removed on startup and shutdown in order to ensure they don't persist between runs. #3195
 * [BUGFIX] Handle hash-collisions in the query path. #3192
-* [BUGFIX] Experimental Alertmanager: Do not allow empty Alertmanager configurations or bad template filenames to be submitted through the configuration API. #3185
+* [BUGFIX] Check for postgres rows errors. #3197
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
 * [BUGFIX] Configs: prevent validation of templates to fail when using template functions. #3157
 * [BUGFIX] Configuring the S3 URL with an `@` but without username and password doesn't enable the AWS static credentials anymore. #3170
 * [BUGFIX] Limit errors on ranged queries (`api/v1/query_range`) no longer return a status code `500` but `422` instead. #3167
+* [BUGFIX] Experimental Alertmanager: Do not allow empty Alertmanager configurations to be submitted through the configuration API. #XXXX
```
pracucci, line 78: I've just merged a PR to cut the CHANGELOG because of the upcoming `1.4.0` release. Could you rebase `master` and move your CHANGELOG entry to the top, under `## master / unreleased`, please?

gotjosh: Done!

pkg/alertmanager/api.go
```diff
@@ -151,7 +156,8 @@ func validateUserConfig(cfg alerts.AlertConfigDesc) error {
 	for _, tmpl := range cfg.Templates {
 		_, err := createTemplateFile(tmpDir, cfg.User, tmpl.Filename, tmpl.Body)
 		if err != nil {
-			return err
+			level.Error(logger).Log("msg", "unable to create template file", "err", err, "user", cfg.User)
+			return fmt.Errorf("unable to create template file %s", tmpl.Filename)
```
gotjosh, line 160: I had to re-word this error has it was providing information about the FS - I don't think it could have been an attack vector but it felt like it had too much information in it.

pstibrany: I think it can, and if tmpl.Filename contains eg. `hello/../../../../../etc/passwd` it could possibly rewrite `/etc/password` even now (if permissions allow). I don't see any checks for path traversal.

pstibrany (approved):
This PR lgtm. Bug mentioned in the comment can be fixed in followup PR.

pkg/alertmanager/api.go
gotjosh: Done! WDYT?

pstibrany: Looks good, thanks. I'm wondering if we should put that check into `createTemplateFile` instead, since it's being called from multiple places?

gotjosh: This call is only internal and getting a useful error message would require some effort if we put inside the function given we're not taking the error message into account - not sure it's worth the effort at this point.

gotjosh: Done as part of https://github.com/cortexproject/cortex/pull/3185/commits/f731b9ab85e5ba3a49da4fdda4811eeb208cfba2

pstibrany (approved):
Thank you!

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
 * [BUGFIX] Ruler: directories in the configured `rules-path` will be removed on startup and shutdown in order to ensure they don't persist between runs. #3195
 * [BUGFIX] Handle hash-collisions in the query path. #3192
+* [BUGFIX] Experimental Alertmanager: Do not allow empty Alertmanager configurations or bad template filenames to be submitted through the configuration API. #3185
```
pracucci, line 15: I would make it clear it's the API, because the alertmanager itself is not experimental.

```suggestion
* [BUGFIX] Experimental Alertmanager API: Do not allow empty Alertmanager configurations or bad template filenames to be submitted through the configuration API. #3185
```

pracucci (approved):
LGTM thanks!

gouthamve:
Wait, there is a _valid case_ for when a user wants to upload an empty config, when they want to use the `fallback config`. See: https://github.com/cortexproject/cortex/blob/547aa290d16e6109620e063df794a70024781d64/pkg/alertmanager/multitenant.go#L385-L393

An empty configuration is not an _invalid_ configuration according to other parts of the code. Why are we rejecting it?

pkg/alertmanager/api.go
```diff
@@ -130,8 +131,12 @@ func (am *MultitenantAlertmanager) DeleteUserConfig(w http.ResponseWriter, r *ht
 	w.WriteHeader(http.StatusOK)
 }
 
-func validateUserConfig(cfg alerts.AlertConfigDesc) error {
-	// Validation copied from: https://github.com/prometheus/alertmanager/blob/8e861c646bf67599a1704fc843c6a94d519ce312/cli/check_config.go#L65-L96
+// Partially copied from: https://github.com/prometheus/alertmanager/blob/8e861c646bf67599a1704fc843c6a94d519ce312/cli/check_config.go#L65-L96
+func validateUserConfig(logger log.Logger, cfg alerts.AlertConfigDesc) error {
+	if cfg.RawConfig == "" {
```
gouthamve, line 136: Can we add a comment that we use the fallback config differently with this API, where we will provision an AM if you don't have any config. Hence we don't need to create an empty config here.

gotjosh: done!

gouthamve (approved):

PR #3189, opened 2020-09-16
Turn memcached circuit-breaker on by default

bboreham:
With these settings it will Will trip after 10 failures within 10 seconds

Also change circuit-breaker log fields to avoid clash: the names 'from' and 'to' are used elsewhere as dates, so avoid re-using them here as strings

**Checklist**
- NA Tests updated
- NA Documentation added
- [ ] `CHANGELOG.md` updated


pracucci (approved):
Thanks!

PR #3192, opened 2020-09-16
Dont rely on hashes for collecting chunks together

### LINKED ISSUE #717: Worry more about fingerprint clashes
The chunk store fetch code assumes that fingerprints uniquely identify a timeseries; this is fairly likely when they are looking at a single metric, but we still could get clashes.

Perhaps use the sha256 from the index?

gouthamve:
This fixes #717

We hit issue in production when our customer issued a query that issued
a simple query that touched 20K series, many of which had hash
collisions on `client.FastFingerprint`. This made our querier code
merge multiple series together which caused counters to go up
and down and caused rate() to return weird artifacts.

It is quite tricky to debug and I now have a test up which proves that
we will suffer if we don't handle collisions explicitly. After looking
at several solutions for fixing that, I've finally settled on something
super simple, just use a string as the map key.

Here is a benchmark where I insert and lookup 100K series:

```
goos: linux
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/distributor
BenchmarkSeriesMap-8         842          91633894 ns/op
PASS
ok      github.com/cortexproject/cortex/pkg/distributor 86.307s
```

Now I could have just used labels.String(), but the performance there is
quite bad:

```
goos: linux
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/distributor
BenchmarkSeriesMap-8         195         373104859 ns/op
PASS
ok      github.com/cortexproject/cortex/pkg/distributor 110.456s
```

Compare this to using client.Fingerprint for the hashing:
```
goos: linux
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/distributor
BenchmarkSeriesMap-8        1273          54778130 ns/op
```
This means that we've gotten 70% in this section, but as explained below
this is very small when compared to the overall query.

----------------

Now the reason I've stuck to the simplified case is because it takes
<100ms for doing this over 100K series, and in most cases the network
time to load all the chunks and iterate through them is several orders
of magnitude higher.

And tbh, I've looked at manually handling collsions, and we need to do
labels.Equal(l1, l2) to see if the series we're looking is the actual
series in the map or a collision, and the perf there actually worse.

I'm open to ideas here. Also note that the values of each map are
different, and any complex solution would require interface{} which is
arguably worse. sha256 has been tried and is worse in comparision.

**What this PR does**:

Fixes #717

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -178,7 +178,7 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 
 		// Parse any chunk series
 		for _, series := range response.Chunkseries {
-			key := LabelsToKeyString(client.FromLabelAdaptersToLabels(series.Labels))
+			key := client.LabelsToKeyString(client.FromLabelAdaptersToLabels(series.Labels))
 			existing := hashToChunkseries[key]
 			existing.Labels = series.Labels
 			existing.Chunks = append(existing.Chunks, series.Chunks...)
@@ -187,7 +187,7 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 
 		// Parse any time series
 		for _, series := range response.Timeseries {
-			key := LabelsToKeyString(client.FromLabelAdaptersToLabels(series.Labels))
+			key := client.LabelsToKeyString(client.FromLabelAdaptersToLabels(series.Labels))
 			existing := hashToTimeSeries[key]
 			existing.Labels = series.Labels
 			if existing.Samples == nil {
@@ -251,11 +251,4 @@ func sameSamples(a, b []ingester_client.Sample) bool {
 		}
 	}
 	return true
-}
-
-// LabelsToKeyString is used to form a string to be used as
-// the hashKey. Don't print, use l.String() for printing.
-func LabelsToKeyString(l labels.Labels) string {
-	b := make([]byte, 0, 1024)
-	return string(l.Bytes(b))
 }
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -170,32 +170,32 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 		return nil, err
 	}
 
-	hashToChunkseries := map[model.Fingerprint]ingester_client.TimeSeriesChunk{}
-	hashToTimeSeries := map[model.Fingerprint]ingester_client.TimeSeries{}
+	hashToChunkseries := map[string]ingester_client.TimeSeriesChunk{}
+	hashToTimeSeries := map[string]ingester_client.TimeSeries{}
 
 	for _, result := range results {
 		response := result.(*ingester_client.QueryStreamResponse)
 
 		// Parse any chunk series
 		for _, series := range response.Chunkseries {
-			hash := client.FastFingerprint(series.Labels)
-			existing := hashToChunkseries[hash]
+			key := LabelsToKeyString(client.FromLabelAdaptersToLabels(series.Labels))
+			existing := hashToChunkseries[key]
 			existing.Labels = series.Labels
 			existing.Chunks = append(existing.Chunks, series.Chunks...)
-			hashToChunkseries[hash] = existing
+			hashToChunkseries[key] = existing
 		}
 
 		// Parse any time series
 		for _, series := range response.Timeseries {
-			hash := client.FastFingerprint(series.Labels)
-			existing := hashToTimeSeries[hash]
+			key := LabelsToKeyString(client.FromLabelAdaptersToLabels(series.Labels))
+			existing := hashToTimeSeries[key]
 			existing.Labels = series.Labels
 			if existing.Samples == nil {
 				existing.Samples = series.Samples
 			} else {
 				existing.Samples = mergeSamples(existing.Samples, series.Samples)
 			}
-			hashToTimeSeries[hash] = existing
+			hashToTimeSeries[key] = existing
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


jtlisi (approved):
LGTM

I agree the extra time spent here is marginal with respect to the overall query path and the added complexity of a collision handling approach is probably not worth the time spent implementing it.

**suggestion**:  Since `label.Bytes()` resets the length of the provided []bytes slice I updated the code to do the following:

**query.go**

```go
// LabelsToKeyString is used to form a string to be used as
// the hashKey. Don't print, use l.String() for printing.
func LabelsToKeyString(l labels.Labels, b []byte) string {
	return string(l.Bytes(b))
}
```

**query_test.go**
```go
func benchmarkSeriesMap(numSeries int, b *testing.B) {
	series := makeSeries(numSeries)
	sm := map[string]int{}

        // instantiate a reusable slice
	buf := make([]byte, 0, 1024)
	b.ResetTimer()
	for n := 0; n < b.N; n++ {
		for i, s := range series {
			sm[LabelsToKeyString(s, buf)] = i
		}
```

By reusing the byte array you can shave off about 10%. Although this may not be worth it.

```console
$ /usr/local/bin/go test -benchmem -run=^$ github.com/cortexproject/cortex/pkg/distributor -bench ^BenchmarkSeriesMap
goos: darwin
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/distributor
BenchmarkSeriesMap-8   	      13	  82358056 ns/op	29421232 B/op	  200303 allocs/op
PASS
ok  	github.com/cortexproject/cortex/pkg/distributor	3.026s
```

pkg/distributor/query.go
```diff
@@ -252,3 +252,10 @@ func sameSamples(a, b []ingester_client.Sample) bool {
 	}
 	return true
 }
+
+// LabelsToKeyString is used to form a string to be used as
+// the hashKey. Don't print, use l.String() for printing.
+func LabelsToKeyString(l labels.Labels) string {
+	b := make([]byte, 0, 1024)
```
pstibrany, line 259: Using 1024 bytes slice seems like overkill. Here is a histogram of label lengths from single block with 949910 series:

```
66-155.3    █ 14629
155.3-244.6 ██████████████████████████████▏ 487865
244.6-333.9 ██████████████████ 292595
333.9-423.2 █▌ 23516
423.2-512.5 ▌ 6943
512.5-601.8 ███▊ 60915
601.8-691.1 ███▍ 54235
691.1-780.4 ▋ 8929
780.4-869.7 ▏ 277
869.7-959   ▏ 6

summary:
 p50=240, p90=562, p95=624, p99=690, min=66, max=959, avg=284.7459643545178
```

It shows that 90% of all series have labels shorter than 562 bytes.

(Alternatively: what about doing another pass through labels to get more accurate length to allocate?)

pkg/querier/chunk_store_queryable.go
```diff
@@ -53,10 +53,10 @@ func (q *chunkStoreQuerier) Select(_ bool, sp *storage.SelectHints, matchers ...
 
 // Series in the returned set are sorted alphabetically by labels.
 func partitionChunks(chunks []chunk.Chunk, mint, maxt int64, iteratorFunc chunkIteratorFunc) storage.SeriesSet {
-	chunksBySeries := map[model.Fingerprint][]chunk.Chunk{}
+	chunksBySeries := map[string][]chunk.Chunk{}
 	for _, c := range chunks {
-		fp := client.Fingerprint(c.Metric)
-		chunksBySeries[fp] = append(chunksBySeries[fp], c)
+		key := distributor.LabelsToKeyString(c.Metric)
```
pstibrany, line 58: Can we make this function to some utils or "client" package where other fingerprinting functions are?

pstibrany (approved):
I'm fine with this approach. Please include memory allocations in your benchmarks, and see comment about lowering default value.

pracucci (approved):
LGTM (modulo Peter's comments)

pkg/distributor/query.go
pracucci: Or using a pool?

gouthamve:
I tried the 3 options and am getting some very confusing answers:

With b := make([]byte, 0, 512)

```
goos: linux
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/ingester/client
BenchmarkSeriesMap-8         794          91165196 ns/op        28800429 B/op     200002 allocs/op
PASS
ok      github.com/cortexproject/cortex/pkg/ingester/client     82.069s
```

With b := make([]byte, 0, 1024)

```
goos: linux
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/ingester/client
BenchmarkSeriesMap-8         792          91004586 ns/op        28800442 B/op     200002 allocs/op
PASS
ok      github.com/cortexproject/cortex/pkg/ingester/client     81.759s
```

With 
```
numBytes := 0
for _, lv := range l {
	numBytes += 2 + len(lv.Name) + len(lv.Value)
}
b := make([]byte, 0, numBytes
```

```
goos: linux
goarch: amd64
pkg: github.com/cortexproject/cortex/pkg/ingester/client
BenchmarkSeriesMap-8         702          99405354 ns/op        57600479 B/op     400002 allocs/op
PASS
ok      github.com/cortexproject/cortex/pkg/ingester/client     80.696s
```

pstibrany:
> I tried the 3 options and am getting some very confusing answers:
> 

You're right. This `make` call doesn't really cause any allocation, thanks to Go optimizations:

```
./query.go:260:11: make([]byte, 0, 1024) does not escape
```

(running `go test -c -gcflags '-m'` in the package)

pkg/distributor/query.go
```diff
@@ -170,32 +170,32 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 		return nil, err
 	}
 
-	hashToChunkseries := map[model.Fingerprint]ingester_client.TimeSeriesChunk{}
-	hashToTimeSeries := map[model.Fingerprint]ingester_client.TimeSeries{}
+	hashToChunkseries := map[string]ingester_client.TimeSeriesChunk{}
+	hashToTimeSeries := map[string]ingester_client.TimeSeries{}
```
bboreham, line 174: these are now misnamed: the map key is not a hash.

PR #3195, opened 2020-09-17
fix: cleanup tempoary ruler files on startup and shutdown

### LINKED ISSUE #3008: restart cortex ruler not load rule config
I use configdb to make a http request to get rule group config, when first run cortex ruler, it will store user rule group and rules to ./rules/username/filename, but when i restart cortex ruler with no any update, it will get rule group from configdb, and then mapper to ./rules/username/filename, ruler check there is no any update, just continue to boot. As a result, it ignore load any rules...

In short, 
1. ruler first start and get rules from configdb which make a http request
2. ruler set user rules to userManagers and mapping rules to a file
3. restart ruler
4. ruler second get fules form configdb
5. ruler compare and mapping rules to file
6. no update, then ruler alse not load to userManager
7. as a result, no any rules load

jtlisi:
**What this PR does**:

On startup and shutdown the Ruler will clean up any directories in it's configured `rule-path`

Fixes #3008 

**Checklist**
- [x] Tests updated


pracucci (approved):
LGTM

PR #3197, opened 2020-09-17
Check for postgres rows errors

bvwells:
Signed-off-by: Ben Wells <b.v.wells@gmail.com>

**What this PR does**:

This PR checks for errors when reading postgress rows. The rows.Next() method may encounter an error when reading the next row. In this case a false value is returned and it is important to call rows.Err() to see whether an error was encountered.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
 * [BUGFIX] Ruler: directories in the configured `rules-path` will be removed on startup and shutdown in order to ensure they don't persist between runs. #3195
 * [BUGFIX] Handle hash-collisions in the query path. #3192
+* [BUGFIX] Check for postgres rows errors. #3197
 
 ## 1.4.0-rc.0 in progress
 
--- a/pkg/configs/db/postgres/postgres.go
+++ b/pkg/configs/db/postgres/postgres.go
@@ -134,6 +134,13 @@ func (d DB) findConfigs(filter squirrel.Sqlizer) (map[string]userconfig.View, er
 		cfg.DeletedAt = deletedAt.Time
 		cfgs[userID] = cfg
 	}
+
+	// Check for any errors encountered.
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
+
 	return cfgs, nil
 }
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
 * [BUGFIX] Ruler: directories in the configured `rules-path` will be removed on startup and shutdown in order to ensure they don't persist between runs. #3195
 * [BUGFIX] Handle hash-collisions in the query path. #3192
+* [BUGFIX] Check for postgres rows errors.
```
pstibrany, line 15:
```suggestion
* [BUGFIX] Check for postgres rows errors. #3197
```

pkg/configs/db/postgres/postgres.go
```diff
@@ -134,6 +134,13 @@ func (d DB) findConfigs(filter squirrel.Sqlizer) (map[string]userconfig.View, er
...
 		cfgs[userID] = cfg
 	}
+
+	// Check for any errors encountered.
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
+
 	return cfgs, nil
```
pstibrany, line 144:
```suggestion
	return cfgs, rows.Err()
```

Can we simplify it to this?

pstibrany (approved):
Thank you for your contribution!

CHANGELOG.md
bvwells: Good spot, updated.

pkg/configs/db/postgres/postgres.go
bvwells: I guess the only issue to change to this is that the method will end up returning a partial result and an error in the case that one of the rows returns an error when Next is called. I think that most examples of this I see would return a nil pointer and an error in this case.

Depends on how you want the API to behave. Let me know your thoughts, I can make the change if required.

Depends on whether this

pstibrany: > I guess the only issue to change to this is that the method will end up returning a partial result and an error in the case that one of the rows returns an error when Next is called. 

That's fine, code calling these two functions ignores partial result when error is returned. If you have a strong preference for returning `nil`, that's fine as well.

pracucci (approved):
LGTM, thanks!

pkg/configs/db/postgres/postgres.go
bvwells: Was going to have a look at this tonight, but looks like I'm too late! :-)

bvwells:
Thanks @pstibrany and @pracucci for the review and merge!

PR #3207, opened 2020-09-18
feat: add metric and fix querier int test flake

### LINKED ISSUE #3206: Flaky integration test -- TestQuerierNoSharding
Occasionally I hit failures on this test case. I think what is happening is the test finishes before the second queriers worker can connect to the query-frontend.

```
Error Trace:	querier_sharding_test.go:146 | 427s
Error:      	Max difference between 0 and 100 allowed is 20, but difference was -100 | 427s
Test:       	TestQuerierNoSharding
```

jtlisi:
**What this PR does**:

It's rare but I've noticed some flake on an integration test that is caused by a querier's worker not connecting to the frontend before the test finishes. This should help solve that issue.

Fixes #3206 

**Checklist**
- [x] Tests updated


pracucci (approved):
LGTM, thanks!

pstibrany (approved):
👍

PR #3208, opened 2020-09-18
Update ruler rules on API based update

gouthamve:

**What this PR does**:

Today, there is a 1min delay until the sees the update they made via the experimental API. In this PR, I am forcing a best effort config reload whenever a user updates their config.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/ruler/ruler.go
```diff
@@ -550,3 +563,71 @@ func (r *Ruler) Rules(ctx context.Context, in *RulesRequest) (*RulesResponse, er
...
+}
+
+// ReloadRules implements the rules service.
+func (r *Ruler) ReloadRules(ctx context.Context, _ *ReloadRulesRequest) (*emptypb.Empty, error) {
+	userID, err := user.ExtractOrgID(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("no user id found in context")
+	}
+
+	r.loadRules(ctx, userID)
```
jtlisi, line 631:
```suggestion
	go r.loadRules(ctx, userID)
```

Might as well run this in the background since no error is returned at this point.

pkg/ruler/ruler.go
```diff
@@ -550,3 +563,71 @@ func (r *Ruler) Rules(ctx context.Context, in *RulesRequest) (*RulesResponse, er
...
+
+			if err != nil {
+				level.Error(r.logger).Log("msg", "err reloading rules", "err", err, "ruler", addr)
+				return
+			}
+
+		}(r.logger, rlr.Addr, &wg)
+	}
+
+	wg.Wait()
```
jtlisi, line 621: **question**: Since this is best effort should we both waiting for every ruler to respond?

pracucci (commented):
When running multiple rulers, this best-effort approach works only for the ruler where the API was executed, correct?

pracucci:
> When running multiple rulers, this best-effort approach works only for the ruler where the API was executed, correct?

Ignore it. It was too quick. Your idea is good 👍

pkg/ruler/ruler.go
```diff
@@ -347,21 +349,32 @@ func (r *Ruler) run(ctx context.Context) error {
...
+	var err error
+
+	if userID == "" {
+		configs, err = r.store.ListAllRuleGroups(ctx)
+	} else {
+		var rgs rules.RuleGroupList
+		rgs, err = r.store.ListRuleGroups(ctx, userID, "")
+		configs[userID] = rgs
+	}
+
```
pstibrany, line 377: This is dangerous. `r.manager.SyncRuleGroups` always needs *All* users to be present in the map, with all of the rules. If it doesn't find the user in the supplied map, rules manager for given user is stopped and removed from memory.

pstibrany (changes requested):

gouthamve:
Superceded by #3268

PR #3209, opened 2020-09-18
Blocksconvert: download plan file locally

pstibrany:
**What this PR does**: This PR modifies "builder" tool from blocksconvert such that it first downloads the plan file to local disk, before it starts processing it. Reason for this change is that sometimes we see "connection reset" or "http2: server sent GOAWAY and closed the connection" errors from reading the plan. These can be fully avoided by downloading full plan before building the block.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


docs/configuration/v1-guarantees.md
```diff
@@ -48,3 +48,4 @@ Currently experimental features are:
 - Querier support for querying chunks and blocks store at the same time.
 - Tracking of active series and exporting them as metrics (`-ingester.active-series-metrics-enabled` and related flags)
 - Shuffle-sharding of queriers in the query-frontend (i.e. use of `-frontend.max-queriers-per-user` flag with non-zero value).
+- Blocksconvert tools
```
pracucci, line 51: Could you open a PR for this against the `release-1.4` branch too, please?

tools/blocksconvert/plan_file.go
```diff
@@ -19,15 +19,15 @@ import (
 
 type PlanEntry struct {
 	// Header
-	User     string `json:"user"`
-	DayIndex int    `json:"day_index"`
+	User     string `json:"user,omitempty"`
```
pracucci, line 22: Has the `omitempty` been added to avoid empty JSON properties when it gets serialised? Or is there a different reason?

tools/blocksconvert/scanner/scanner.go
```diff
@@ -42,7 +42,7 @@ type Config struct {
 }
 
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
-	f.StringVar(&cfg.TableName, "table", "", "Table to generate plan files from. If not used, tables are discovered via schema.")
+	f.StringVar(&cfg.TableNames, "scanner.tables", "", "Comma-separated tables to generate plan files from. If not used, all tables found via schema and scanning of Index store will be used.")
```
pracucci, line 45: Could you update the doc accordingly, please?

tools/blocksconvert/scanner/scanner.go
```diff
@@ -204,27 +204,30 @@ func (s *Scanner) running(ctx context.Context) error {
...
 	level.Info(s.logger).Log("msg", "total found tables", "count", len(allTables))
 
+	if s.cfg.TableNames != "" {
+		// Find tables from parameter.
+		tableNames := map[string]bool{}
+		for _, t := range strings.Split(s.cfg.TableNames, ",") {
+			tableNames[strings.TrimSpace(t)] = true
+		}
+
+		for ix := 0; ix < len(allTables); {
```
pracucci, line 214: `ix` is never increased

pracucci (commented):
Looks a nice improvement. Overall LGTM, but I think there's a little bug and the `blocksconvert` doc needs to be updated.

tools/blocksconvert/scanner/scanner.go
pstibrany: It is, on line 223. (There was bug before, perhaps you have seen buggy version of the code)

tools/blocksconvert/plan_file.go
pstibrany: That exactly is the reason. On manual inspection of some plan files I've realized that all fields are *always* included, which is wasteful and makes plan files bigger than necessary.

tools/blocksconvert/scanner/scanner.go
pracucci: Yes, I reviewed the previous version.

pracucci (approved):
LGTM!

PR #3210, opened 2020-09-18
Ruler Experimental API: Fix rule group validation

### LINKED ISSUE #2942: Ruler API should not accept invalid rules files
The Ruler API that allows users to write rules files to Cortex should return an error to users who try to upload invalid rules files, for example a rule group with no name.

cc @gotjosh

gotjosh:
**What this PR does**:

Fixed a few cases where the API would allow us to submit a bad rule group payload which would end up overwriting existing ones.

In our case, this is pretty egregious given we operate against individual rule groups you could end up in a situation where you have overwritten a working rule group by mistake. 

Fixes #2942 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -450,7 +450,7 @@ func (r *Ruler) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 			e = append(e, err.Error())
 		}
 
-		http.Error(w, strings.Join(e, ","), http.StatusBadRequest)
+		http.Error(w, strings.Join(e, ", "), http.StatusBadRequest)
 		return
 	}
 
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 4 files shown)


pkg/ruler/api.go
```diff
@@ -443,10 +444,13 @@ func (r *Ruler) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
...
 	errs := r.manager.ValidateRuleGroup(rg)
 	if len(errs) > 0 {
+		e := []string{}
 		for _, err := range errs {
 			level.Error(logger).Log("msg", "unable to validate rule group payload", "err", err.Error())
+			e = append(e, err.Error())
 		}
-		http.Error(w, errs[0].Error(), http.StatusBadRequest)
+
+		http.Error(w, strings.Join(e, ","), http.StatusBadRequest)
```
jtlisi, line 453:
```suggestion
		http.Error(w, strings.Join(e, ", "), http.StatusBadRequest)
```

May be worth adding a space so the errors aren't one continues string

jtlisi (approved):
LGTM modulo linter

pkg/ruler/api.go
gotjosh: good shout.

pracucci (approved):
LGTM, thanks!

PR #3216, opened 2020-09-21
Alertmanager: Remove outdated comment

gotjosh:
**What this PR does**:

Adds a new metric to keep track of when was the last uploaded configuration successfully applied to the Alertmanager. Also, improves documentation.

**Still needs a test but wanted to see whenever this was an acceptable solution - An alternative here is that we stop the Alertmanager completely if we fail to load the last configuration and symbolise the user the configuration did not work.** 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -348,9 +348,6 @@ func (am *MultitenantAlertmanager) syncConfigs(cfgs map[string]alerts.AlertConfi
 }
 
 func (am *MultitenantAlertmanager) transformConfig(userID string, amConfig *amconfig.Config) (*amconfig.Config, error) {
-	if amConfig == nil { // shouldn't happen, but check just in case
-		return nil, fmt.Errorf("no usable Cortex configuration for %v", userID)
-	}
 	if am.cfg.AutoWebhookRoot != "" {
 		for _, r := range amConfig.Receivers {
 			for _, w := range r.WebhookConfigs {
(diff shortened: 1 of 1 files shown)


pracucci (commented):
> [...] wanted to see whenever this was an acceptable solution - An alternative here is that we stop the Alertmanager completely if we fail to load the last configuration and symbolise the user the configuration did not work.

I have a question: how is the new metric different then what `cortex_alertmanager_config_invalid` already does?

pracucci:
@gotjosh Is there still interest in this PR?

gotjosh:
This was superseded by #3289 but the comments are still relevant - I'll modify the PR, as I think the comments are very relevant for upcoming work.

pkg/alertmanager/multitenant.go
```diff
@@ -405,15 +402,21 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
...
+			// working configuration.
 			return fmt.Errorf("invalid Cortex configuration for %v: %v", cfg.User, err)
 		}
 	}
 
+	// We can have an empty configuration here if:
+	// 1) the user had a previous alertmanager
+	// 2) then, submitted a non-working configuration (and we kept running the prev working config)
+	// 3) finally, the cortex AM instance is restarted and the running version is no longer present
+	if userAmConfig == nil {
```
jtlisi, line 416: **suggestion(non-blocking,if-minor):** Since this check has been moved from `transformConfig` consider removing the `transformConfig` function entirely an moving the for loop logic from the function here. `transformConfig` is only called here so it should be safe to move and I think it would still be quite readable as part of a single function call.

jtlisi (approved):
LGTM

pracucci (approved):
LGTM!

gotjosh:
@pracucci This should be good to go!

PR #3217, opened 2020-09-21
support snappy cache in query frontend

yeya24:

**What this PR does**:

Add an `enable_snappy` config to Results Cache.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


docs/configuration/config-file-reference.md
```diff
@@ -813,6 +813,10 @@ results_cache:
     # The CLI flags prefix for this block config is: frontend
     [fifocache: <fifo_cache_config>]
 
+  # Enable snappy compression on results cache.
+  # CLI flag: -frontend.enable-snappy-cache
+  [enable_snappy: <boolean> | default = false]
```
pracucci, line 818: I would allow to configure the compression algorithm name instead of a bool. We did this make for the gRPC compression in the past, which then lead to deprecating the boolean in favour of a string:

```
    # Deprecated: Use gzip compression when sending messages.  If true,
    # overrides grpc-compression flag.
    # CLI flag: -bigtable.grpc-use-gzip-compression
    [use_gzip_compression: <boolean> | default = false]

    # Use compression when sending messages. Supported values are: 'gzip',
    # 'snappy' and '' (disable compression)
    # CLI flag: -bigtable.grpc-compression
    [grpc_compression: <string> | default = ""]
```

I would suggest to go direct with a `compression string` which we'll accept `snappy` and empty string (disabled, default).

yeya24: It sounds much better! Thanks, I will update my pr and let you know.

pstibrany (approved):
LGTM, thank you!

pracucci (approved):
LGTM, thanks!

PR #3218, opened 2020-09-22
Speed up the listing of rule groups on the Ruler Configuration API

### LINKED ISSUE #3121: Ruler: GET /api/v1/rules is slow
We have observed consistent slowness (> 3s) when fetching the list of the rules/alerts from this endpoint. Today, while around the area I noticed that we do a `List` of objects in the bucket followed up by a `Get` of each in serial. In the case I mention, we have about 63 rule groups.

Looking around the current object storage we don't support batching but I think the Thanos one does (?). This might be yet another reason to perhaps move in that direction. If not, it feels like we need support for batching. 

Being fair, I'm not swinging to one solution or another just weighting down our options here.

gotjosh:
**What this PR does**:

Fixes #3121 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,6 @@
   - `POST <legacy-http-prefix>/push`
   - `POST /push`
   - `POST /ingester/push`
-* [CHANGE] Chunks GCS object storage client uses the `fields` selector to limit the payload size. #3218
 * [FEATURE] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-user` globally, or using per-user limit `max_queriers_per_user`), each user's requests will be handled by different set of queriers. #3113
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
 * [ENHANCEMENT] Expose additional HTTP configs for the S3 backend client. New flag are listed below: #3244
@@ -63,6 +62,7 @@
 * [ENHANCEMENT] Hash ring: added instance registered timestamp to the ring. #3248
 * [ENHANCEMENT] Reduce tail latency by smoothing out spikes in rate of chunk flush operations. #3191
 * [ENHANCEMENT] Experimental Ruler API: Fetch rule groups from object storage in parallel. #3218
+* [ENHANCEMENT] Chunks GCS object storage client uses the `fields` selector to limit the payload size when listing objects in the bucket. #3218
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
 * [BUGFIX] Ruler: directories in the configured `rules-path` will be removed on startup and shutdown in order to ensure they don't persist between runs. #3195
 * [BUGFIX] Handle hash-collisions in the query path. #3192
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
   - `POST <legacy-http-prefix>/push`
   - `POST /push`
   - `POST /ingester/push`
+* [CHANGE] Chunks GCS object storage client uses the `fields` selector to limit the payload size. #3218
 * [FEATURE] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-user` globally, or using per-user limit `max_queriers_per_user`), each user's requests will be handled by different set of queriers. #3113
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
 * [ENHANCEMENT] Expose additional HTTP configs for the S3 backend client. New flag are listed below: #3244
@@ -61,6 +62,7 @@
 * [ENHANCEMENT] Blocksconvert – Builder: retry block upload before giving up. #3245
 * [ENHANCEMENT] Hash ring: added instance registered timestamp to the ring. #3248
 * [ENHANCEMENT] Reduce tail latency by smoothing out spikes in rate of chunk flush operations. #3191
+* [ENHANCEMENT] Experimental Ruler API: Fetch rule groups from object storage in parallel. #3218
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
 * [BUGFIX] Ruler: directories in the configured `rules-path` will be removed on startup and shutdown in order to ensure they don't persist between runs. #3195
 * [BUGFIX] Handle hash-collisions in the query path. #3192
--- a/pkg/chunk/gcp/gcs_object_client.go
+++ b/pkg/chunk/gcp/gcs_object_client.go
@@ -109,8 +109,13 @@ func (s *GCSObjectClient) PutObject(ctx context.Context, objectKey string, objec
 func (s *GCSObjectClient) List(ctx context.Context, prefix, delimiter string) ([]chunk.StorageObject, []chunk.StorageCommonPrefix, error) {
 	var storageObjects []chunk.StorageObject
 	var commonPrefixes []chunk.StorageCommonPrefix
+	q := &storage.Query{Prefix: prefix, Delimiter: delimiter}
+	err := q.SetAttrSelection([]string{"Name", "Updated"})
+	if err != nil {
+		return nil, nil, err
+	}
 
-	iter := s.bucket.Objects(ctx, &storage.Query{Prefix: prefix, Delimiter: delimiter})
+	iter := s.bucket.Objects(ctx, q)
 	for {
 		if ctx.Err() != nil {
 			return nil, nil, ctx.Err()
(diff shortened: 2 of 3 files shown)


pstibrany:
I'm currently changing how rules are being loaded for shuffle sharding, and will need to only do it for some users, not all users. I'm thinking modifying `List` method to add `delimeter` parameter directly to it, to allow both recursive and non-recursive listings. Different users will then be loaded concurrently.

We may want to coordinate our efforts.

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -111,8 +111,10 @@ func (s *GCSObjectClient) PutObject(ctx context.Context, objectKey string, objec
 func (s *GCSObjectClient) List(ctx context.Context, prefix string) ([]chunk.StorageObject, []chunk.StorageCommonPrefix, error) {
 	var storageObjects []chunk.StorageObject
 	var commonPrefixes []chunk.StorageCommonPrefix
+	q := &storage.Query{Prefix: prefix, Delimiter: s.delimiter}
+	q.SetAttrSelection([]string{"name", "updated"})
```
pstibrany, line 115: This should be:
```suggestion
    q.SetAttrSelection([]string{"Name", "Updated"})
``` 

`SetAttrSelection` method will check and map these names to ones expected by API.

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -74,25 +75,56 @@ func (o *RuleStore) ListAllRuleGroups(ctx context.Context) (map[string]rules.Rul
...
+			for key := range ch {
+				user := decomposeRuleObjectKey(key)
+				if user == "" {
+					continue
+				}
+
+				rg, err := o.getRuleGroup(ctx, key)
+				if err != nil {
+					level.Error(util.Logger).Log("msg", "failed to get rule group", "key", key, "user", user)
+					errCh <- err
```
pstibrany, line 101: Given that `ListAllRuleGroups` returns first error, this gouroutine can stop after sending error to the channel.

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -74,25 +75,56 @@ func (o *RuleStore) ListAllRuleGroups(ctx context.Context) (map[string]rules.Rul
...
-		if user == "" {
-			continue
-		}
-
-		rg, err := o.getRuleGroup(ctx, obj.Key)
-		if err != nil {
-			return nil, err
-		}
+	ch := make(chan string)
+	errCh := make(chan error)
```
pstibrany, line 81: Can we use [errgroup](https://godoc.org/golang.org/x/sync/errgroup) here? That should remove need for `WaitGroup` and error channel, and will stop all goroutines when context is cancelled.

(Unbuffered channel here will cause leak of gouroutines if multiple workers try to send error)

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -74,25 +75,56 @@ func (o *RuleStore) ListAllRuleGroups(ctx context.Context) (map[string]rules.Rul
...
+					level.Error(util.Logger).Log("msg", "failed to get rule group", "key", key, "user", user)
+					errCh <- err
+				}
+
+				mtx.Lock()
+				defer mtx.Unlock()
+				if _, exists := userGroupMap[user]; !exists {
+					userGroupMap[user] = rules.RuleGroupList{}
+				}
+				userGroupMap[user] = append(userGroupMap[user], rg)
```
pstibrany, line 109:
```suggestion
				mtx.Lock()
				userGroupMap[user] = append(userGroupMap[user], rg)
				mtx.Unlock()
```

1. `defer` executes when function leaves, but this function doesn't until reading channel completely
2. we can simplify adding to the map

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -74,25 +75,56 @@ func (o *RuleStore) ListAllRuleGroups(ctx context.Context) (map[string]rules.Rul
...
+	}
 
-		if _, exists := userGroupMap[user]; !exists {
-			userGroupMap[user] = rules.RuleGroupList{}
+	for _, obj := range ruleGroupObjects {
+		select {
+		case ch <- obj.Key:
+			// ok
+		case <-ctx.Done():
+			return userGroupMap, ctx.Err()
```
pstibrany, line 119: By returning here without closing the channel, worker goroutines may get stuck on reading from channel.

I'd suggest 1) using buffered channel, with buffer equal to `len(ruleGroupObjects)`, 2) pushing all elements to the channel, and closing it, and only then 3) wait until all goroutines finish. Using `errgroup` will make this wait context-aware as well.

pstibrany (changes requested):
Thank you. Please take a look at comments.

pkg/chunk/gcp/gcs_object_client.go
gotjosh: Good catch.

gotjosh:
@pstibrany I have a feeling this should wait until #3235 is merged so that I can rebase and pick up the changes.

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -172,6 +163,58 @@ func (o *RuleStore) DeleteNamespace(ctx context.Context, userID, namespace strin
 	return nil
 }
 
+func (o *RuleStore) loadRuleGroupsConcurrently(ctx context.Context, rgObjects []chunk.StorageObject) (map[string]rules.RuleGroupList, error) {
+	ch := make(chan string, len(rgObjects))
+
+	for _, obj := range rgObjects {
+		level.Debug(util.Logger).Log("msg", "listing rule group", "key", obj.Key)
```
pstibrany, line 170: You may want to put this debug into fetching goroutine just before calling `o.getRuleGroup(gCtx, key)`, to avoid logging too early.

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -172,6 +163,58 @@ func (o *RuleStore) DeleteNamespace(ctx context.Context, userID, namespace strin
...
+			return nil
+		})
+	}
+
+	err := g.Wait()
+	if err != nil {
+		return map[string]rules.RuleGroupList{}, err
+	}
+
+	return result, nil
```
pstibrany, line 215:
```suggestion
	return result, err
```

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -103,18 +92,20 @@ func (o *RuleStore) ListRuleGroups(ctx context.Context, userID, namespace string
...
-			level.Error(util.Logger).Log("msg", "unable to retrieve rule group", "err", err, "key", obj.Key)
-			return nil, err
-		}
-		groups = append(groups, rg)
+	groups, err := o.loadRuleGroupsConcurrently(ctx, ruleGroupObjects)
+	if err != nil {
+		return nil, err
+	}
+
+	if _, exists := groups[userID]; !exists {
```
pstibrany, line 104: No need for extra `if` here... `nil` is a valid empty slice to return. Using just

```
    return groups[userID], nil`
```

or even
```
    return groups[userID], err
```

and removing previous `if err != nil` will work just fine.

pstibrany (approved):
Good job, thanks! I've left few nit comments. Integration test failure is unrelated.

pkg/ruler/rules/objectclient/rule_store.go
gotjosh: Doing it like this would return partial results on `result`, but I guess it does not really matter given uses of this check for  `err` and return immediately.

pkg/ruler/rules/objectclient/rule_store.go
gotjosh: I tested that in a playground with unsuccessful results https://play.golang.org/p/3-iF20WnP2I

Could be it be that Go's type system coerces nil to the empty return type so that `len` works?

pstibrany: Try this instead: https://play.golang.org/p/BGOaFKkQkzR

`nil` interface{} is different than `nil` slice or `nil` map.

gotjosh: Thanks! I have addressed the changes.

pstibrany (approved):
Thank you!

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
   - `POST <legacy-http-prefix>/push`
   - `POST /push`
   - `POST /ingester/push`
+* [CHANGE] Chunks GCS object storage client uses the `fields` selector to limit the payload size. #3218
```
pracucci, line 43: It's not a change impacting the user. Also, it's not even a change for embedders. It's more an `[ENHANCEMENT]`. We should also mention it affects only listing objects in the bucket.

```suggestion
* [ENHANCEMENT] Chunks GCS object storage client uses the `fields` selector to limit the payload size when listing objects in the bucket. #3218
```

pracucci (approved):
LGTM! Just left a nit about the CHANGELOG entry.

PR #3220, opened 2020-09-22
Share auth-injector tool

pracucci:
**What this PR does**:
I had this simple tool in my local repository since a while and I thought to propose to upstream it. It's just an HTTP reverse proxy offering a quick way to inject `X-Scope-OrgID` header.

For example, the following command listens on `8081` and proxies all requests to `8080` injecting `X-Scope-OrgID: load-balancer-1`:
```
go run ./tools/auth-injector -local-address=localhost:8081 -remote-address=http://localhost:8080 -tenant-id=load-balancer-1
```

Do you think it make sense to upstream it?

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,9 @@ website/content/en/docs
 e2e_integration_test*
 active-query-tracker
 dist/
+
+# Binaries built from ./cmd
+blocksconvert
+cortex
+query-tee
+test-exporter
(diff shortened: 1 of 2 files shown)


pstibrany (approved):
Sure, this can be useful in some situations, I don't see why not.

PR #3221, opened 2020-09-22
Arbitrary group loading in local rules backend

owen-d:
**Allow arbitrary group loading from local rule storage**:

This PR passes the upstream [GroupLoader](https://github.com/prometheus/prometheus/blob/master/rules/manager.go#L992-L996) interface to the local rules storage backend. This interface is already used in the Ruler API via the Prometheus rules manager to validate rules. This PR aims for parity with that and avoids hardcoding the implementation which allows downstream projects (Loki) to use the local backend store.

### DIFF
--- a/pkg/ruler/rules/local/local.go
+++ b/pkg/ruler/rules/local/local.go
@@ -8,7 +8,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
-	"github.com/prometheus/prometheus/pkg/rulefmt"
+	promRules "github.com/prometheus/prometheus/rules"
 
 	"github.com/cortexproject/cortex/pkg/ruler/rules"
 )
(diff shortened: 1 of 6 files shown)


pstibrany (approved):
LGTM

pkg/ruler/rules/local/local.go
```diff
@@ -143,7 +145,7 @@ func (l *Client) listAllRulesGroupsForUser(ctx context.Context, userID string) (
 func (l *Client) listAllRulesGroupsForUserAndNamespace(ctx context.Context, userID string, namespace string) (rules.RuleGroupList, error) {
 	filename := filepath.Join(l.cfg.Directory, userID, namespace)
 
-	rulegroups, allErrors := rulefmt.ParseFile(filename)
+	rulegroups, allErrors := l.loader.Load(filename)
```
pracucci, line 148: Note to myself: in Cortex we always use `FileLoader` whose `Load()` function calls `rulefmt.ParseFile()`, so this is a no-op.

pracucci (approved):
LGTM!

PR #3222, opened 2020-09-23
Blocksconvert: added support for specifying date range for tables scan

pstibrany:
**What this PR does**: This PR adds support for specifying date-range when scanning for tables. It also adds metrics to scanner to better track its progress.

Tables ignored due to range are logged:
```
level=info ts=2020-09-23T06:17:18.034230103Z caller=scanner.go:264 msg="table starts after period-end, ignoring" table=index_2646 table_start="2020-09-17 00:00:00 +0000 UTC" table_end="2020-09-24 00:00:00 +0000 UTC" period_end=2020-09-17T00:00:00Z

level=info ts=2020-09-23T06:17:18.034241493Z caller=scanner.go:259 msg="table ends before period-start, ignoring" table=index_2613 table_start="2020-01-30 00:00:00 +0000 UTC" table_end="2020-02-06 00:00:00 +0000 UTC" period_start=2020-02-06T00:00:00Z
```

### DIFF
--- a/tools/blocksconvert/scanner/scanner.go
+++ b/tools/blocksconvert/scanner/scanner.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/cortexproject/cortex/pkg/chunk"
 	"github.com/cortexproject/cortex/pkg/chunk/storage"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/services"
 	"github.com/cortexproject/cortex/tools/blocksconvert"
 )
@@ -30,6 +31,9 @@ import (
 type Config struct {
 	TableNames  string
 	TablesLimit int
+
+	PeriodStart flagext.DayValue
+	PeriodEnd   flagext.DayValue
 
 	OutputDirectory string
 	Concurrency     int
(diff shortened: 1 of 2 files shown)


tools/blocksconvert/scanner/scanner.go
```diff
@@ -231,15 +258,33 @@ func (s *Scanner) running(ctx context.Context) error {
 		return allTables[i].start.After(allTables[j].start)
 	})
 
+	for ix := 0; ix < len(allTables); {
+		t := allTables[ix]
+		if s.cfg.PeriodStart.IsSet() && !t.end.IsZero() && t.end.Unix() <= s.cfg.PeriodStart.Unix() {
```
pracucci, line 263: In the `PeriodStart` CLI flag description there's:
> Specified date is included in the range. 

If so, shouldn't the check here being `t.end.Unix() < s.cfg.PeriodStart.Unix()`?

pracucci: I actually realised the `t.end` is computed as:
`end:    start.Add(tablesConfig.Period),`

So the `end` is excluded. I would suggest to add a comment to document it. The `if` logic here is fine.

tools/blocksconvert/scanner/scanner.go
```diff
@@ -134,6 +144,14 @@ func NewScanner(cfg Config, scfg blocksconvert.SharedConfig, l log.Logger, reg p
 			Name: "cortex_blocksconvert_bigtable_parsed_index_entries_total",
 			Help: "Number of parsed index entries",
 		}),
+		currentTableRanges: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_blocksconvert_scanner_bigtable_ranges_in_current_table",
+			Help: "Number of ranges to scan from current table.",
+		}),
+		currentTableScannedRanges: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_blocksconvert_scanner_bigtable_scanned_ranges_from_current_table",
+			Help: "Number of scanned ranges from current table. (Goes from 0 on each new table)",
```
pracucci, line 153: > (Goes from 0 on each new table)

I find this a bit unclear. Maybe we could just say "Resets to 0 every time a table is getting scanned or its scan has completed."

pracucci (approved):
LGTM! Comments left are nits 🚀

tools/blocksconvert/scanner/scanner.go
pstibrany: > I would suggest to add a comment to document it.

`end` already has this comment:

```
	end    time.Time // Will not be set for non-periodic tables. Exclusive.
```

What do you suggest?

pracucci: All good. Sorry, but I didn't notice that comment.

pstibrany: Thanks

PR #3229, opened 2020-09-23
Fix consul in metric name potentially use with in memory kv store

### LINKED ISSUE #3226: The "consul" work in "cortex_consul_request_duration_seconds" metric is misleading
The name `consul` is wrong. It actually tracks any KV-store operation used by the ring, regardless you're using consul or another backend.

roidelapluie:
**What this PR does**:

Rename a metric to remove confusion

Fixes #3226

**Checklist**


pstibrany (approved):

pstibrany (changes requested):
Thank you.

Actually there already is `cortex_kv_request_duration_seconds`. In-memory client is using Consul interface, which is why internally we use Consul client to talk to it.

So this doesn't need changing in my opinion.

roidelapluie:
Yes I see what you mean, but that it still confusing. Closing this for now.

PR #3231, opened 2020-09-23
Upgrade vendored Alertmanager, Prometheus, and Thanos

jtlisi:
**What this PR does**:

This PR updates the vendored Alertmanager/Prometheus/Thanos to latest. Some changes needed to be rectified in the ruler notifier and the ingester WAL code to make this work.

A significant amount of code has been removed since Prometheus inverted and abstracted their service discovery implementations. We no longer need to vendor a significant amount of code. See https://github.com/prometheus/prometheus/pull/7701

/cc @codesome would you be able to sanity check the WAL changes?

Also Prometheus is now being built with Go 1.15.2 upstream. However, I have not updated the build image in this PR since it requires changing the version of Debian we use.


codesome (commented):
WAL changes look fine. -860,532 lines :raised_hands:

pstibrany:
Thank you! I've checked Prometheus and Thanos changes and haven't noticed anything worth being concerned about. Thanos store now uses chunks querier to serve `Series` call, let's see how that helps.

Thanos has introduced some new metrics that would be good to expose:

- `thanos_bucket_store_cached_series_fetch_duration_seconds`
- `thanos_bucket_store_cached_postings_fetch_duration_seconds`

(We can do that in separate PR)

pstibrany (approved):

pracucci (approved):
I reviewed Prometheus, Thanos, AWS changes and LGTM!

I just noticed a couple of new Thanos metrics which were not mapped in the store-gateway, so I pushed a commit do it. I have intentionally not mapped the new TSDB metric `prometheus_tsdb_data_replay_duration_seconds` because we already have `cortex_ingester_tsdb_wal_replay_duration_seconds`.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 * [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added options to control whether active series tracking is enabled (`-ingester.active-series-enabled`, defaults to false), and how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
 * [ENHANCEMENT] Blocksconvert – Builder: download plan file locally before processing it. #3209
 * [ENHANCEMENT] Store-gateway: added zone-aware replication support to blocks replication in the store-gateway. #3200
+* [ENHANCEMENT] Store-gateway: exported new metrics. #3231
+  - `cortex_bucket_store_cached_series_fetch_duration_seconds`
+  - `cortex_bucket_store_cached_postings_fetch_duration_seconds`
+  - `cortex_bucket_stores_gate_queries_max`
```
pstibrany, line 22: I can't see where this metric is defined.

pracucci: The `gate_queries_max` is a new metric added to the Thanos gate, which we use here:
https://github.com/cortexproject/cortex/blob/77b21c50c3ce2091abd381697790c31a0675b69e/pkg/storegateway/bucket_stores.go#L71-L72

PR #3235, opened 2020-09-24
Shuffle-sharding support in Ruler.

pstibrany:
**What this PR does**: This PR adds support for shuffle-sharding into ruler. Unfinished.

**TODO:**

- [x] Fix `FSObjectClient` to support new `List` method properly (Done separately in PR #3238. This PR is rebased on top of it)
- [x] Load user rules concurrently
- [x] Testing

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/ruler/ruler_test.go
```diff
@@ -216,3 +227,324 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rules.RuleGroupDesc
...
+		user3: {user3Group1},
+	}
+
+	type testCase struct {
+		sharding         bool
+		shardingStrategy string
+		shuffleShardSize int
+		setupRing        func(*ring.Desc)
+
+		expectedRulesForFuler1 map[string]rules.RuleGroupList
```
pracucci, line 261: `Fuler` made me smile :D Thanks!

```suggestion
		expectedRulesForRuler1 map[string]rules.RuleGroupList
```

pkg/ruler/ruler_test.go
```diff
@@ -216,3 +227,324 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rules.RuleGroupDesc
...
+				desc.AddIngester(ruler1, ruler1Addr, "", []uint32{user1Group1Token + 1, user2Group1Token + 1}, ring.ACTIVE)
+				desc.Ingesters[ruler2] = ring.IngesterDesc{
+					Addr:      ruler2Addr,
+					Timestamp: time.Now().Add(-time.Hour).Unix(),
+					State:     ring.ACTIVE,
+					Tokens:    []uint32{user1Group2Token + 1, user3Group1Token + 1},
+				}
+			},
+
+			// This ruler doesn't rulers from unhealthy ruler (RF=1).
```
pracucci, line 320: Did you mean... ?

```suggestion
			// This ruler doesn't get rules from unhealthy ruler (RF=1).
```

pkg/ruler/ruler_test.go
```diff
@@ -216,3 +227,324 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rules.RuleGroupDesc
...
+			sharding:         true,
+			shardingStrategy: ShardingStrategyDefault,
+
+			setupRing: func(desc *ring.Desc) {
+				desc.AddIngester(ruler1, ruler1Addr, "", []uint32{user1Group1Token + 1, user2Group1Token + 1}, ring.ACTIVE)
+				desc.AddIngester(ruler2, ruler2Addr, "", []uint32{user1Group2Token + 1, user3Group1Token + 1}, ring.JOINING)
+			},
+
+			// JOINING ruler will get the rules.
+			expectedRulesForFuler1: map[string]rules.RuleGroupList{
```
pracucci, line 382: Doesn't this expectation conflicts with `default sharding, JOINING ruler-1`?

pkg/ruler/ruler_test.go
```diff
@@ -216,3 +227,324 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rules.RuleGroupDesc
...
+		},
+
+		"shuffle sharding, multiple rulers, shard size 1": {
+			sharding:         true,
+			shardingStrategy: ShardingStrategyShuffle,
+			shuffleShardSize: 1,
+
+			setupRing: func(desc *ring.Desc) {
+				desc.AddIngester(ruler1, ruler1Addr, "", []uint32{userToken(user1, 0) + 1, userToken(user2, 0) + 1, userToken(user3, 0) + 1}, ring.ACTIVE)
+				desc.AddIngester(ruler2, ruler2Addr, "", []uint32{0}, ring.ACTIVE)
```
pracucci, line 406: Given this ruler has only the token 0, is this test case a fair one? The ruler 1 owns all required tokens for `allRules` so I suspect that would pass even if the shard size is 2. Am I wrong?

pracucci (commented):
Good job! I reviewed everything but the FS object store client changes (will do it in a 2nd pass). I left few comments here and there, but so far LGTM. Could you apply object store changes to https://github.com/cortexproject/cortex/pull/3238 please?

Also:
- Remember to `make doc`.
 - If https://github.com/cortexproject/cortex/pull/3236 will be merged before this PR, I would be glad if you could add ruler to the shuffle sharding too.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -289,8 +287,7 @@ func (a *S3ObjectClient) PutObject(ctx context.Context, objectKey string, object
 	})
 }
 
-// List objects and common-prefixes i.e synthetic directories from the store non-recursively
```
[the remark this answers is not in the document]
pstibrany: Done in https://github.com/cortexproject/cortex/pull/3238.

pkg/ring/ring.go
```diff
@@ -407,6 +407,16 @@ func (r *Ring) Collect(ch chan<- prometheus.Metric) {
 	)
 }
 
+func ShuffleShardSeed(identifier string) int64 {
```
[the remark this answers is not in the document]
pstibrany: I don't like the idea of query-frontend package having dependency on ring package for single method only. Let's move it to `util` instead.

pkg/ruler/rules/store.go
```diff
@@ -63,9 +69,26 @@ func NewConfigRuleStore(c client.Client) *ConfigRuleStore {
 	}
 }
 
-// ListAllRuleGroups implements RuleStore
-func (c *ConfigRuleStore) ListAllRuleGroups(ctx context.Context) (map[string]RuleGroupList, error) {
+func (c *ConfigRuleStore) ListAllUsers(ctx context.Context) ([]string, error) {
+	m, err := c.LoadAllRuleGroups(ctx)
+
+	// TODO: this should be optimized, if possible.
```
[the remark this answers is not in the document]
pstibrany: I'm not going to optimize it, but will leave TODO there if someone wants to use this combination.

pkg/ruler/ruler_test.go
pstibrany: > `Fuler` made me smile :D Thanks!

I hope it doesn't mean anything bad :)

pkg/ruler/ruler_test.go
pstibrany: Right. The comment is wrong. Those rules will in fact not be given to any ruler!

pstibrany: This is result of using replication factor 1 :( With higher RF, only one ruler would be joining, other would be ACTIVE or LEAVING and they would get the rule instead. With RF=1, we only select single ruler.

We could fix this by extending the replica set if JOINING ingester was selected.

pstibrany: Ruler now uses different replication strategy. Only ACTIVE Rulers can receive rules. JOINING/LEAVING/PENDING rulers have no rules.

pkg/ruler/ruler_test.go
pstibrany: With shard size 2, it was in fact second ruler that got all the rules. I've now modified the test (added group tokens to ruler2), and added same test with shardSize=2 to show that.

pstibrany:
@pracucci I've addressed your feedback, and I think this PR is ready for review now.

pstibrany:
Oh, I forgot about the "ring changed" use case. Let me address that.

pstibrany:
@annanay25 @jtlisi This PR also fixes problem that can occur when using sharding: when user is moved to a different ruler, and then moved back, `DefaultMultiTenantManager` would not re-create ruler manager anymore, because it would detect that rules are still the same in the `syncRulesToManager` method. Now the method also checks if manager exists.

When ruler manager is removed from memory, its metrics are also removed. When manager is recreated, new metrics are created. This works because `ManagerMetrics` exports all values per-user.

pkg/ruler/manager_metrics.go
```diff
@@ -99,8 +99,17 @@ func NewManagerMetrics() *ManagerMetrics {
...
 func (m *ManagerMetrics) AddUserRegistry(user string, reg *prometheus.Registry) {
 	m.regsMu.Lock()
+	defer m.regsMu.Unlock()
+
 	m.regs[user] = reg
-	m.regsMu.Unlock()
+}
+
+// DeleteUserRegistry removes user-specific Prometheus registry.
+func (m *ManagerMetrics) DeleteUserRegistry(user string) {
```
pracucci, line 108: I'm not sure this is correct. The `ManagerMetrics.Collect()` sends out aggregates, so when we remove an user, the aggregated counters will decrease and will be detected as resets even if they're not. Am I missing anything?

If I'm not missing anything, I believe that what we need to do is to never remove an user registry from `ManagerMetrics` but reuse that registry if already exists.

pkg/ruler/manager.go
```diff
@@ -129,11 +130,12 @@ func (r *DefaultMultiTenantManager) syncRulesToManager(ctx context.Context, user
 		return
 	}
 
-	if update {
-		level.Debug(r.logger).Log("msg", "updating rules", "user", "user")
+	manager, exists := r.userManagers[user]
+	if !exists || update {
```
pracucci, line 134: Is this fix testable?

pracucci (commented):
Good job! I haven't approved yet because I've the feeling the removal of the user registry from the `ManagerMetrics` is not correct (see dedicated comment and let's have a chat about it).

A part from this and few minor nits left:
- Please add a CHANGELOG entry, mentioning also the bug fix (https://github.com/cortexproject/cortex/pull/3235#issuecomment-699967405)
- Mention the new metric `cortex_ruler_sync_rules_total ` in the CHANGELOG

pkg/ruler/manager_metrics.go
pstibrany: `Collect` method sends out all metrics *per-user*, but doesn’t aggregate over users. This is important and makes `DeleteUserRegistry` work properly. When we remove user here, given user will disappear from `Collect` output, but it doesn’t affect any aggregate values.

Now I admit this is tricky, and we should put a comment into `Collect` method to make sure we *never* do aggregates over all users there.

Reusing registry for new ruler manager wouldn’t work, since ruler manager would try to register the same metrics and fail.

We could reuse entire ruler manager, by never removing existing one from memory for given user. (But removing rules instead).

Another option is to keep a list of registries for each user. That is instead of having `map[user]Registry`, we would keep `map[user][]Registry`, and each new ruler manager for user would get its own registry. When sending metrics, we would aggregate over all of them. I think this solution would be applicable for other places in Cortex (eg. closing and reopening TSDB), but I would prefer to implement that in a separate PR, since it’s not required here (yet).

pracucci: You're right. I'm fine keeping it as is until will be required. I would just add a clear and loud comment in `Collect()` function.

pracucci (approved):
Approving because my only concern was not an issue (my mistake). All good to me. I would still be glad if you could take a look at the non-blocking comments tho.

docs/guides/shuffle-sharding.md
```diff
@@ -92,3 +92,13 @@ When shuffle sharding is **enabled** via `-store-gateway.sharding-strategy=shuff
 _The shard size can be overridden on a per-tenant basis setting `store_gateway_tenant_shard_size` in the limits overrides configuration._
 
 _Please check out the [store-gateway documentation](../blocks-storage/store-gateway.md) for more information about how it works._
+
+### Ruler shuffle sharding
+
+Cortex ruler can run in three modes:
+
+1. **No sharding at all.** This is the most basic mode of the ruler. It is activated by using `-ruler.enable-sharding=false` (default) and works correctly only if single ruler is running. In this mode the Ruler loads all rules for all users.
```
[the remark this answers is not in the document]
pstibrany: Should we also fix `-frontend.max-queriers-per-user` to `-frontend.max-queriers-per-tenant` while we still can?

pkg/ruler/ruler_test.go
```diff
@@ -216,3 +227,329 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rules.RuleGroupDesc
...
+			setupRing: func(desc *ring.Desc) {
+				desc.AddIngester(ruler1, ruler1Addr, "", []uint32{userToken(user1, 0) + 1, userToken(user2, 0) + 1, userToken(user3, 0) + 1}, ring.ACTIVE)
+				desc.AddIngester(ruler2, ruler2Addr, "", []uint32{user1Group1Token + 1, user1Group2Token + 1, user2Group1Token + 1, user3Group1Token + 1}, ring.ACTIVE)
+			},
+
+			expectedRulesForRuler1: allRules,
+		},
+
+		// Same test as previous one, but with shard size=2. Second ruler gets all the rules.
+		"shuffle sharding, multiple rulers, shard size 2": {
```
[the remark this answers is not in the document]
pstibrany: I've changed the test to include expected rule groups for each ruler. That also allowed to remove some of the test cases.

pstibrany:
Thanks for review Marco. I've now addressed remaining comments. I've also added test for `Collect` method to verify that all returned metrics have `user` label. PTAL when you have time.

docs/guides/shuffle-sharding.md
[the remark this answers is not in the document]
pracucci: Probably yes. But I would suggest to do it in a separate PR.

pkg/ruler/manager.go
pracucci: Good job on the test!

pkg/ruler/ruler_test.go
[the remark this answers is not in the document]
pracucci: Waaaaaay better! Thanks!

pracucci (approved):
Looks very very good to me. No more comments 👏 👏 👏

pkg/ruler/ruler_test.go
```diff
@@ -216,3 +228,373 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rules.RuleGroupDesc
...
+				desc.AddIngester(ruler1, ruler1Addr, "", sortTokens([]uint32{userToken(user1, 0) + 1, userToken(user2, 1) + 1, user1Group1Token + 1, user1Group2Token + 1}), ring.ACTIVE)
+				desc.AddIngester(ruler2, ruler2Addr, "", sortTokens([]uint32{userToken(user1, 1) + 1, userToken(user3, 1) + 1, user2Group1Token + 1}), ring.ACTIVE)
+				desc.AddIngester(ruler3, ruler3Addr, "", sortTokens([]uint32{userToken(user2, 0) + 1, userToken(user3, 0) + 1, user3Group1Token + 1}), ring.ACTIVE)
+			},
+
+			expectedRules: expectedRulesMap{
+				ruler1: map[string]rules.RuleGroupList{
+					user1: {user1Group1, user1Group2},
+				},
+				ruler2: noRules, // It does user2's group token, but it doen's "own" user2 at all.
```
ranton256, line 483: typo: doen's -> doesn't

pkg/ruler/ruler_test.go
```diff
@@ -216,3 +228,373 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rules.RuleGroupDesc
...
+
+		// Same test as previous one, but with shard size=2. Second ruler gets all the rules.
+		"shuffle sharding, two rulers, shard size 2": {
+			sharding:         true,
+			shardingStrategy: ShardingStrategyShuffle,
+			shuffleShardSize: 2,
+
+			setupRing: func(desc *ring.Desc) {
+				desc.AddIngester(ruler1, ruler1Addr, "", sortTokens([]uint32{userToken(user1, 0) + 1, userToken(user2, 0) + 1, userToken(user3, 0) + 1}), ring.ACTIVE)
+				desc.AddIngester(ruler2, ruler2Addr, "", sortTokens([]uint32{user1Group1Token + 1, user1Group2Token + 1, user2Group1Token + 1, user3Group1Token + 1}), ring.ACTIVE)
```
ranton256, line 415: I'm confused why there are specific rule group tokens in the ring setup for this test.

ranton256 (commented):
I have some questions, may just result in clarifying my understanding rather than any code changes.

pkg/util/shard.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import (
+	"crypto/md5"
+	"encoding/binary"
+)
+
+// ShuffleShardSeed returns seed for random number generator, computed from provided identifier.
+func ShuffleShardSeed(identifier string) int64 {
```
harry671003, line 9: Why not refactor and reuse? https://github.com/cortexproject/cortex/blob/master/pkg/querier/frontend/frontend_querier_queues.go#L226

pkg/ruler/rules/store.go
```diff
@@ -21,8 +21,14 @@ var (
 
 // RuleStore is used to store and retrieve rules
 type RuleStore interface {
-	ListAllRuleGroups(ctx context.Context) (map[string]RuleGroupList, error)
```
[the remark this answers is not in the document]
pstibrany: I found it misleading for `List*` methods to also do loading, so I've renamed them and added comment too.

Once this PR is in, I plan to split Listing and Loading into separate methods, because right now Ruler loads all rule groups, even if it's going to throw them right away (due to sharding).

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -67,8 +68,30 @@ func (o *RuleStore) getRuleGroup(ctx context.Context, objectKey string) (*rules.
...
 
-// ListAllRuleGroups returns all the active rule groups
-func (o *RuleStore) ListAllRuleGroups(ctx context.Context) (map[string]rules.RuleGroupList, error) {
+func (o *RuleStore) ListAllUsers(ctx context.Context) ([]string, error) {
+	_, prefixes, err := o.client.List(ctx, rulePrefix, delim)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []string
```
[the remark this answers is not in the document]
pstibrany: Yes, ruler stores all rule groups into "rules/<user>/<namespace>/<group name>" objects (with base64-encoded namespace and group name). Right now this isn't configurable in any way.

pkg/ruler/ruler_replication_strategy.go
```diff
@@ -0,0 +1,38 @@
...
+	// Filter out unhealthy instances.
+	for i := 0; i < len(instances); {
+		if instances[i].IsHealthy(op, heartbeatTimeout) {
+			i++
+		} else {
+			instances = append(instances[:i], instances[i+1:]...)
+		}
+	}
+
+	// Each rule group is loaded to *exactly* one ruler, no matter the RF.
```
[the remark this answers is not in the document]
pstibrany: I will remove the comment from here. While it's true, it's not configured here but at https://github.com/cortexproject/cortex/blob/15d2d1431a9bb1d936f5f4edd272fa2e788bf6ff/pkg/ruler/ruler_ring.go#L94

This code doesn't deal with replication factor, so the comment doesn't belong here.

pkg/ruler/ruler_test.go
pstibrany: This test uses the exact same tokens setup as previous test `"shuffle sharding, multiple rulers, shard size 1"`, but with shard size=2. I've added comment about it.

pkg/ruler/rules/store.go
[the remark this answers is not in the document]
ranton256: Okay, makes sense.

pkg/ruler/ruler_replication_strategy.go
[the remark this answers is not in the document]
ranton256: Okay, I see, thanks for clarifying.

pkg/ruler/ruler_test.go
ranton256: 👍

ranton256 (approved):
Thanks for the clarifications.

PR #3238, opened 2020-09-24
ObjectClient List refactoring

pstibrany:
**What this PR does**: This PR refactors `ObjectClient` interface to explicitly pass `delimiter` to `List` method. Previously delimiter was property of `ObjectClient` passed to it when created, but that doesn't work if code using `ObjectClient` wants to do both recursive and non-recursive calls.

This PR also changes `FSObjectClient` such that it always expects paths using `/` separator, but maps it to `\` on Windows internally. Code using `FSObjectClient` will only see `/` returned.

This is a pre-requisite for PR #3235.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/gcp/gcs_object_client.go
+++ b/pkg/chunk/gcp/gcs_object_client.go
@@ -13,10 +13,9 @@ import (
 )
 
 type GCSObjectClient struct {
-	cfg       GCSConfig
-	client    *storage.Client
-	bucket    *storage.BucketHandle
-	delimiter string
+	cfg    GCSConfig
+	client *storage.Client
+	bucket *storage.BucketHandle
 }
 
 // GCSConfig is config for the GCS Chunk Client.
@@ -39,7 +38,7 @@ func (cfg *GCSConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 }
 
 // NewGCSObjectClient makes a new chunk.Client that writes chunks to GCS.
-func NewGCSObjectClient(ctx context.Context, cfg GCSConfig, delimiter string) (*GCSObjectClient, error) {
+func NewGCSObjectClient(ctx context.Context, cfg GCSConfig) (*GCSObjectClient, error) {
 	option, err := gcsInstrumentation(ctx, storage.ScopeReadWrite)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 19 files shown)


pkg/chunk/local/fs_object_client.go
```diff
@@ -95,53 +95,60 @@ func (f *FSObjectClient) PutObject(ctx context.Context, objectKey string, object
...
-	folderPath := filepath.Join(f.cfg.Directory, prefix)
+	folderPath := filepath.Join(f.cfg.Directory, filepath.FromSlash(prefix))
+	err := filepath.Walk(folderPath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
 
-	_, err := os.Stat(folderPath)
-	if err != nil {
-		if os.IsNotExist(err) {
```
sandeepsukhani, line 107: I think we should preserve this check to be consistent with other stores where it is not an error to call list with a prefix which does not exist.
What do you think?

sandeepsukhani (approved):
A small suggested change in FS object client otherwise it LGTM.

pkg/chunk/local/fs_object_client.go
pstibrany: Good point! Fixed now, PTAL.

pracucci (approved):
LGTM!

pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -108,11 +106,11 @@ func (s *GCSObjectClient) PutObject(ctx context.Context, objectKey string, objec
 }
 
 // List objects and common-prefixes i.e synthetic directories from the store non-recursively
```
pracucci, line 108: This comment was left, compared to other ones.

pkg/chunk/inmemory_storage_client.go
```diff
@@ -425,7 +425,7 @@ func (m *MockStorage) DeleteObject(ctx context.Context, objectKey string) error
 	return nil
 }
 
-func (m *MockStorage) List(ctx context.Context, prefix string) ([]StorageObject, []StorageCommonPrefix, error) {
+func (m *MockStorage) List(ctx context.Context, prefix, delimiter string) ([]StorageObject, []StorageCommonPrefix, error) {
```
pracucci, line 428: PathSeparator() below should be removed.

pkg/chunk/storage/factory.go
```diff
@@ -246,7 +246,7 @@ func NewChunkClient(name string, cfg Config, schemaCfg chunk.SchemaConfig, regis
 	case "inmemory":
 		return chunk.NewMockStorage(), nil
 	case "aws", "s3":
-		return newChunkClientFromStore(aws.NewS3ObjectClient(cfg.AWSStorageConfig.S3Config, chunk.DirDelim))
```
pracucci, line 249: I think you can remove chunk.DirDelim because unused.

pracucci (approved):
Few final nits!

pstibrany:
Addressed remaining comments. I've also copied `List` description from #3235 into this PR.

PR #3239, opened 2020-09-25
issues templates to gather all the required information in bug reports

sandeepsukhani:
**What this PR does**:
Adds issues templates to gather all the required info in bug reports which otherwise requires going back and forth with users to get it.

This is how it would look like: https://github.com/sandeepsukhani/test-issues-template/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,1 +1,0 @@
-blank_issues_enabled: false
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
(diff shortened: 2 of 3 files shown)


.github/ ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
sandeepsukhani, line 1: This is to hide the highlighted link in the screenshot. Users can still create an issue with a different format by clearing out all the text from the one of the existing templates.
image

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
sandeepsukhani, line 1: This is to hide the highlighted link in the screenshot. Users can still create an issue with a different format by clearing out all the text from the one of the existing templates.
image

pstibrany, line 1: Let's keep this enabled. It doesn't do any harm and is sometimes useful.

pstibrany (commented):
I'm not a big fan of these templates, and end up deleting the text more often than not. I'd only keep one for bug reports.

.github/ISSUE_TEMPLATE/config.yml
sandeepsukhani: Sounds good!

sandeepsukhani:
> I'm not a big fan of these templates, and end up deleting the text more often than not. I'd only keep one for bug reports.

It helps people who are new or not used to reporting bugs/requesting features to have a template for it which helps capture most of the required info. I was investigating a bug yesterday which didn't have much info so I had to go back and forth with him to get more information. For power users, it is always possible to just use a blank issue or clear out the template.

pstibrany (approved):

pracucci (approved):
Thanks!

PR #3244, opened 2020-09-27
feat: expose S3 HTTP configs for TSDB S3 client

jtlisi:
**What this PR does**:

- Expose additional HTTP configuration parameters for the Minio S3 client vendored from Thanos
**Checklist**
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master / unreleased
 
+* [CHANGE] Update the default HTTP configuration values for the S3 client to the upstream Thanos default values. #3244
+  - `-blocks-storage.s3.http.idle-conn-timeout` is set 90 seconds.
+  - `-blocks-storage.s3.http.response-header-timeout` is set to 2 minutes.
 * [CHANGE] Improved shuffle sharding support in the write path. This work introduced some config changes: #3090
   * Introduced `-distributor.sharding-strategy` CLI flag (and its respective `sharding_strategy` YAML config option) to explicitly specify which sharding strategy should be used in the write path
   * `-experimental.distributor.user-subring-size` flag renamed to `-distributor.ingestion-tenant-shard-size`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@
   - `POST /ingester/push`
 * [FEATURE] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-user` globally, or using per-user limit `max_queriers_per_user`), each user's requests will be handled by different set of queriers. #3113
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
+* [ENHANCEMENT] Expose additional HTTP configs for the S3 backend client. New flag are listed below: #3244
+  - `-blocks-storage.s3.http.idle-conn-timeout`
+  - `-blocks-storage.s3.http.response-header-timeout`
+  - `-blocks-storage.s3.http.insecure-skip-verify`
 * [ENHANCEMENT] Added `cortex_query_frontend_connected_clients` metric to show the number of workers currently connected to the frontend. #3207
 * [ENHANCEMENT] Shuffle sharding: improved shuffle sharding in the write path. Shuffle sharding now should be explicitly enabled via `-distributor.sharding-strategy` CLI flag (or its respective YAML config option) and guarantees stability, consistency, shuffling and balanced zone-awareness properties. #3090 #3214
 * [ENHANCEMENT] Ingester: added new metric `cortex_ingester_active_series` to track active series more accurately. Also added options to control whether active series tracking is enabled (`-ingester.active-series-enabled`, defaults to false), and how often this metric is updated (`-ingester.active-series-update-period`) and max idle time for series to be considered inactive (`-ingester.active-series-idle-timeout`). #3153
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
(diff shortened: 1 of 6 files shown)


pkg/storage/backend/s3/config.go
```diff
@@ -2,17 +2,38 @@ package s3
...
+	ResponseHeaderTimeout time.Duration `yaml:"response_header_timeout"`
+	InsecureSkipVerify    bool          `yaml:"insecure_skip_verify"`
+
+	// Allow upstream callers to inject a round tripper
+	Transport http.RoundTripper `yaml:"-"`
+}
+
+// RegisterFlagsWithPrefix registers the flags for TSDB s3 storage with the provided prefix
+func (cfg *HTTPConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.DurationVar(&cfg.IdleConnTimeout, prefix+"s3.http.idle-conn-timeout", 0, "If set, the time an idle connection will remain idle before closing.")
```
pstibrany, line 23: Should we use default values from vendor/github.com/thanos-io/thanos/pkg/objstore/s3/s3.go:49 (90s idle conns timeout, and 2min response header timeout) instead?

pstibrany (approved):
I think I would suggest using `DefaultConfig` from `github.com/thanos-io/thanos/pkg/objstore/s3` package for configuring default values. Otherwise LGTM.

pkg/storage/backend/s3/config.go
jtlisi: Yes, I think that makes sense.

pstibrany: I think that will be a change from what Cortex does now, so let's document that in the CHANGELOG as well.

pkg/storage/backend/s3/config.go
```diff
@@ -2,17 +2,38 @@ package s3
...
+
 // Config holds the config options for an S3 backend
 type Config struct {
 	Endpoint        string         `yaml:"endpoint"`
 	BucketName      string         `yaml:"bucket_name"`
 	SecretAccessKey flagext.Secret `yaml:"secret_access_key"`
 	AccessKeyID     string         `yaml:"access_key_id"`
 	Insecure        bool           `yaml:"insecure"`
+
+	HTTP HTTPConfig `yaml:"http_config"`
```
pracucci, line 36: I would remove the `_config` suffix. This also fixes the drift between CLI flags and YAML (because in the CLI flags it's `.http.`)

```suggestion
	HTTP HTTPConfig `yaml:"http"`
```

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master / unreleased
 
+* [CHANGE] Update the default HTTP configuration values for the S3 client to the upstream Thanos default values. #3244
```
pracucci, line 5: I would make it clear it affects the blocks storage.

```suggestion
* [CHANGE] Blocks storage: update the default HTTP configuration values for the S3 client to the upstream Thanos default values. #3244
```

pkg/storage/backend/s3/config.go
```diff
@@ -2,17 +2,38 @@ package s3
...
+
+	// Allow upstream callers to inject a round tripper
+	Transport http.RoundTripper `yaml:"-"`
+}
+
+// RegisterFlagsWithPrefix registers the flags for TSDB s3 storage with the provided prefix
+func (cfg *HTTPConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.DurationVar(&cfg.IdleConnTimeout, prefix+"s3.http.idle-conn-timeout", 90*time.Second, "The time an idle connection will remain idle before closing.")
+	f.DurationVar(&cfg.ResponseHeaderTimeout, prefix+"s3.http.response-header-timeout", 2*time.Minute, "The amount of time the client will wait for a servers response headers.")
+	f.BoolVar(&cfg.InsecureSkipVerify, prefix+"s3.http.insecure-skip-verify", false, "If enabled, the client will accept any certificate and hostname.")
```
pracucci, line 25: I would make it clear it's used when connecting via HTTPS, to try to avoid the confusion with `-blocks-storage. s3.insecure`.

```suggestion
	f.BoolVar(&cfg.InsecureSkipVerify, prefix+"s3.http.insecure-skip-verify", false, "If the client connects to S3 via HTTPS and this option is enabled, the client will accept any certificate and hostname.")
```

pracucci (approved):
LGTM with a couple of nits. Thanks!

pkg/storage/backend/s3/config.go
jtlisi: 👍

pkg/storage/backend/s3/config.go
jtlisi: 👍

CHANGELOG.md
jtlisi: 👍

pkg/storage/backend/s3/config.go
```diff
@@ -2,17 +2,38 @@ package s3
...
 )
 
+// HTTPConfig stores the http.Transport configuration for the s3 minio client.
+type HTTPConfig struct {
+	IdleConnTimeout       time.Duration `yaml:"idle_conn_timeout"`
+	ResponseHeaderTimeout time.Duration `yaml:"response_header_timeout"`
+	InsecureSkipVerify    bool          `yaml:"insecure_skip_verify"`
+
+	// Allow upstream callers to inject a round tripper
+	Transport http.RoundTripper `yaml:"-"`
```
bboreham, line 18: This isn't used, is it?

pracucci: This is used by downstream projects.

PR #3248, opened 2020-09-29
Added registered timestamp to instances in the ring

pracucci:
**What this PR does**:
I'm working on adding shuffle sharding support for ingesters on the read path. A preliminary step is to have the timestamp at which an instance registered to the ring, which is what I'm proposing in this PR.

The changeset is large because I had to add the registered timestamp as parameter of the `AddIngester()` function, which is widely used in tests. A part from this, logic changes are pretty minimal.

A part from unit tests, I've also run some manual tests both on ingesters ring (to test `Lifecycler`) and store-gateways ring (to test `BasicLifecycler`):
- Scale up / down an instance
- Abruptly terminate  (`SIGKILL`) an instance and then, after a while, restart it (the registered timestamp is preserved if the instance wasn't forgotten in the ring)
- Click "Forget" in the ring HTTP page

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/ring/basic_lifecycler.go
```diff
@@ -237,7 +247,7 @@ func (l *BasicLifecycler) registerInstance(ctx context.Context) error {
 		var exists bool
 		instanceDesc, exists = ringDesc.Ingesters[l.cfg.ID]
 		if exists {
-			level.Info(l.logger).Log("msg", "instance found in the ring", "instance", l.cfg.ID, "ring", l.ringName, "state", instanceDesc.GetState(), "tokens", len(instanceDesc.GetTokens()))
+			level.Info(l.logger).Log("msg", "instance found in the ring", "instance", l.cfg.ID, "ring", l.ringName, "state", instanceDesc.GetState(), "tokens", len(instanceDesc.GetTokens()), "registered_timestamp", instanceDesc.RegisteredTimestamp)
```
pstibrany, line 250: Can we also include formatted date in the log message?

pkg/ring/lifecycler.go
```diff
@@ -504,23 +517,32 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
 			}
 
 			// Either we are a new ingester, or consul must have restarted
 			level.Info(util.Logger).Log("msg", "instance not found in ring, adding with no tokens", "ring", i.RingName)
-			ringDesc.AddIngester(i.ID, i.Addr, i.Zone, []uint32{}, i.GetState())
+			ringDesc.AddIngester(i.ID, i.Addr, i.Zone, []uint32{}, i.GetState(), registeredAt)
 			return ringDesc, true, nil
 		}
 
+		// The instance already exists in the ring, so we can't change the registered timestamp (even if it's zero)
```
pstibrany, line 542: It may not be clear why we don't set "registered at" if it's zero. I would suggest extending  description of this field in ring.proto to describe its intended usage, and explaining why preserving zero is important.

pkg/ring/lifecycler.go
```diff
@@ -674,7 +696,7 @@ func (i *Lifecycler) updateConsul(ctx context.Context) error {
 		if !ok {
 			// consul must have restarted
 			level.Info(util.Logger).Log("msg", "found empty ring, inserting tokens", "ring", i.RingName)
-			ringDesc.AddIngester(i.ID, i.Addr, i.Zone, i.getTokens(), i.GetState())
+			ringDesc.AddIngester(i.ID, i.Addr, i.Zone, i.getTokens(), i.GetState(), i.getRegisteredAt())
```
pstibrany, line 699: In case ring entry already exists, we refresh all fields except registered timestamp. Let's include it as well.

pkg/ring/model.go
```diff
@@ -127,6 +133,16 @@ func (d *Desc) TokensFor(id string) (tokens, other Tokens) {
 	return myTokens, takenTokens
 }
 
+// GetRegisteredAt returns the timestamp when the instance has been registered to the ring
+// or a zero value if unknown.
+func (i *IngesterDesc) GetRegisteredAt() time.Time {
+	if i == nil {
```
pstibrany, line 139:
```suggestion
	if i == nil || i.RegisteredTimestamp == 0 {
```

(otherwise time.Unix(0, 0) will not be zero time.)

pkg/ring/model.go
```diff
@@ -434,12 +450,12 @@ func (d *Desc) getTokensByZone() map[string][]TokenDesc {
 type CompareResult int
 
 const (
-	Equal                   CompareResult = iota // Both rings contain same instances, tokens, states and timestamps.
-	EqualInstancesAndTokens                      // Both rings contain the same instances and tokens, but states and timestamps differ.
-	Different                                    // Rings have different set of instances, or their information don't match.
+	Equal                          CompareResult = iota // Both rings contain same exact instances.
+	EqualUnlessStatesAndTimestamps                      // Both rings contain the same instances with the same data except states and timestamps (may differ).
```
pstibrany, line 454: `EqualButStatesAndTimestamps` or `EqualExceptStatesAndTimestamps`

pstibrany (approved):
LGTM. Please take a look at comment at `(*IngesterDesc).GetRegisteredAt`.

pkg/ring/basic_lifecycler.go
pracucci: We can, but could be zero value. That's why I used the epoch. WDYT?

pkg/ring/model.go
pracucci: Good point. Fixed and added a test.

pkg/ring/model.go
pracucci: Ok. Picked `EqualButStatesAndTimestamps` just because it's slower.

pkg/ring/lifecycler.go
pracucci: Good point. Done.

pkg/ring/basic_lifecycler.go
pracucci: Anyway, the `.String()` of a zero value timestamp is `0001-01-01 00:00:00 +0000 UTC` so I guess it's quite clear to read that case.

pkg/ring/lifecycler.go
pracucci: I updated the command. I haven't specifically referenced the shuffle sharding because it's not a thing yet and this is a generic property, but I tried to make it clear to never change it once its set (even if it's 0).

WDYT?

pstibrany: What I meant is to explain purpose of the field. Something like: This field is used to find out subset of instances that could have possibly owned the specific token in the past.

pstibrany (approved):
LGTM, good job!

pkg/ring/lifecycler.go
pstibrany: > What I meant is to explain purpose of the field. Something like: This field is used to find out subset of instances that could have possibly owned the specific token in the past.

And from there, explain why updating to non-zero value is a bad idea.

pracucci: Ok, updated!

PR #3250, opened 2020-09-29
Add multi tenant query federation

### LINKED ISSUE #923: Query across multiple users / tenants / instances
As an administrator, I would like to query metrics that exist in multiple tenants.  For example  `prometheus_remote_storage_shards` - I'd like to see what the typical range is, look for outliers.

Previously discussed at #849 but that refers more to metrics generated inside Cortex.

I see two practical considerations: authorisation to access data from other users/instances, and performance.

Other notes from previous conversations:
 * the tentant ID could be exposed as a label…
 * suspect the locking strategy inside ingesters might need care
 * there is no list of users anywhere - you could use all the users with chunks in the ingesters, but that could miss people who haven’t sent anything in ~15mins
   * for many admin purposes, “active users” is fine.

Also related: #577

simonswine:
**What this PR does**:

- This intends to allow cross tenant querying. The implemenation follows this [proposal]

[proposal]: https://cortexmetrics.io/docs/proposals/cross-tenant-query-federation/ 

Fixes #923

**Checklist/TODO**

- [x] Merge #3486 first
- [x] Provide a design document with the details of the the form of the implementation (esp. what this would mean for configured tenant limits and metrics exposed about queries)
- [x] Tests updated
- [ ] Documentation added


docs/configuration/config-file-reference.md
```diff
@@ -858,6 +858,12 @@ store_gateway_client:
 # (ingesters shuffle sharding on read path is disabled).
 # CLI flag: -querier.shuffle-sharding-ingesters-lookback-period
 [shuffle_sharding_ingesters_lookback_period: <duration> | default = 0s]
+
+tenant_federation:
+  # If enabled, multi tenant query federation can be used by supplying multiple
+  # tenant IDs in the read path (experimental).
+  # CLI flag: -querier.tenant-federation.enabled
+  [enabled: <boolean> | default = false]
```
simonswine, line 866: If some cluster enables multi tenant querying, the flag should be set across all the instances and not only on the query frontend/scheduler, so that e.g. the distributor rejects multi tenant ingestion. I feel the prefix `-querier.` is misleading

integration/querier_tenant_federation_test.go
```diff
@@ -0,0 +1,181 @@
...
+	queueMetricName := "cortex_query_frontend_queue_length"
+	if cfg.querySchedulerEnabled {
+		queueComponent = queryScheduler
+		queueMetricName = "cortex_query_scheduler_queue_length"
+	}
+	require.NoError(t, queueComponent.WaitSumMetricsWithOptions(e2e.Equals(0), []string{queueMetricName}, e2e.WithLabelMatchers(
+		labels.MustNewMatcher(labels.MatchEqual, "user", strings.Join(tenantIDs, "|")))))
+
+	// TODO: check cache invalidation on tombstone cache gen increase
+	// TODO: check fairness in queryfrontend
```
simonswine, line 167: Not too sure if this is something I can test easily during integration testing

pkg/api/middlewares.go
```diff
@@ -14,13 +14,18 @@ import (
...
 				return
 			}
 
-			cacheGenNumber := cacheGenNumbersLoader.GetResultsCacheGenNumber(userID)
+			var cacheGenNumber string
+			// join the strings for multiple tenants
+			// TODO: Figure out how sensible that is
+			for _, tenantID := range tenantIDs {
+				cacheGenNumber += cacheGenNumbersLoader.GetResultsCacheGenNumber(tenantID)
+			}
```
simonswine, line 28: I am not too sure if I can `cacheGenNumber` like that

pkg/cortex/cortex.go
```diff
@@ -301,6 +302,12 @@ func New(cfg Config) (*Cortex, error) {
 		os.Exit(0)
 	}
 
+	// TODO: Find a better place for this
```
simonswine, line 305: I think it needs to be set as soon as possible, but I am not too sure if this is a good place. Thoughts?

jtlisi: **thought:** I don't think it's a big deal if this stays here. The alternative is making a small module and setting it as a dependency for the `Server` module. Having a module for such a small snippet may be overkill, but it doesn't necessarily have to be scoped only for this config, it could be a generic `init` module. The Auth config logic below could also be moved there for instance.

pkg/querier/merge_queryable.go
```diff
@@ -0,0 +1,280 @@
...
+		tenantID := m.tenantIDs[posTenant]
+		seriesSets[pos] = &addLabelsSeriesSet{
+			upstream: m.queriers[posTenant].Select(sortSeries, hints, filteredMatchers...),
+			labels: labels.Labels{
+				{
+					Name:  string(m.tenantLabelName),
+					Value: tenantID,
+				},
+			},
+		}
```
jtlisi, line 114: **suggestion(optional,if-minor):** have you considered running the `Select` for each tenant concurrently? It's probably not something that is needed for an initial implementation but it may be useful to help speed things up.

pkg/querier/merge_queryable.go
```diff
@@ -0,0 +1,280 @@
...
+
+	var result []string
+	for e := range resultMap {
+		result = append(result, e)
+	}
+	sort.Strings(result)
+	return result, warnings, nil
+}
+
+func filterValuesByMatchers(labelName string, labelValues []string, matchers ...*labels.Matcher) ([]int, []*labels.Matcher) {
```
jtlisi, line 246: It took me a while to figure out what this function does and why it is needed. Could you add comment explaining it a bit more. It also may be a bit easier if instead of having a separate function you just put this logic into the`mergeQuerier.Select` function and comment it there. I don't think it has much potential for reuse and having it inline would make things a bit easier to follow.

pkg/querier/merge_queryable.go
```diff
@@ -0,0 +1,280 @@
...
+
+// A collection of warnings for the whole set.
+// Warnings could be return even iteration has not failed with error.
+func (m *addLabelsSeriesSet) Warnings() storage.Warnings {
+	return m.upstream.Warnings()
+}
+
+type addLabelsSeries struct {
+	upstream storage.Series
+	labels   labels.Labels
```
jtlisi, line 150: **question:** why make this multiple labels, we only need one to store the tenant?

jtlisi (commented):
This looks great, I added a few questions and nits. I want to go through it one more time before I add an approval. However, I didn't any major issues in my first pass.

pkg/querier/merge_queryable.go
simonswine: I thought about having potentially multiple dimensions in the future (e.g. for federation across clusters)

pkg/querier/merge_queryable.go
```diff
@@ -0,0 +1,280 @@
...
+	resultMap := make(map[string]struct{})
+	for _, f := range funcs {
+		result, w, err := f()
+		if err != nil {
+			return nil, nil, err
+		}
+		for _, e := range result {
+			resultMap[e] = struct{}{}
+		}
+		warnings = append(warnings, w...)
```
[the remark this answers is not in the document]
simonswine: Done also made it a bit more efficient in terms of allocations

pkg/cortex/cortex.go
simonswine: Ok will keep it like that for now

pkg/querier/merge_queryable.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb/chunkenc"
+	"github.com/weaveworks/common/user"
+
+	"github.com/cortexproject/cortex/pkg/tenant"
+)
+
+const defaultTenantLabel = "__tenant_id__"
```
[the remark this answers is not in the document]
simonswine: This was something that @pracucci asked for future extensibility: https://github.com/cortexproject/cortex/pull/3364#discussion_r515871294. I think it makes sense to keep it

pkg/querier/merge_queryable.go
```diff
@@ -0,0 +1,280 @@
...
+	for pos := range m.tenantIDs {
+		querier := m.queriers[pos]
+		funcs[pos] = func() ([]string, storage.Warnings, error) {
+			return querier.LabelValues(name)
+		}
+	}
+	return mergeDistinctStringSlice(funcs...)
+}
+
+// LabelNames returns all the unique label names present in the block in sorted order.
```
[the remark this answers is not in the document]
simonswine: Added a lot more context to those calls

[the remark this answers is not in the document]
simonswine: Thanks for pointing that out, was just copy paste from the interface I guess. Should be much clearer now.

pkg/querier/merge_queryable.go
```diff
@@ -0,0 +1,280 @@
...
+	for _, additionalL := range additionalLabels {
+		setLabelRetainExisting(src, lb, additionalL)
+	}
+
+	return lb.Labels()
+}
+
+type stringSliceFunc func() ([]string, storage.Warnings, error)
+
+func mergeDistinctStringSlice(funcs ...stringSliceFunc) ([]string, storage.Warnings, error) {
```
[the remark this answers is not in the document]
simonswine: Done

pkg/querier/merge_queryable.go
simonswine: I have added two TODO regarding concurrency. I would leave it out of the initial implementation if that is ok, but I think it is something that can follow very soon

pkg/querier/merge_queryable.go
simonswine: I move the method close to Select and mentioned it in the comment specifically. If you think I should move it in line I can still do that...

docs/configuration/config-file-reference.md
simonswine: I have moved the whole implementation into it's own package with it's own config. This should be making it much clearer. 

The wrapping of the queryable now happens in modules.go so it should be more extensible...

jtlisi: 👍

integration/querier_tenant_federation_test.go
jtlisi: It should be possible. However, I wouldn't make it a priority for this PR.

pkg/querier/tenantfederation/tenant_federation.go
```diff
@@ -0,0 +1,14 @@
...
+	"flag"
+)
+
+type Config struct {
+	// Enabled switches on support for multi tenant query federation
+	Enabled bool `yaml:"enabled"`
+}
+
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.BoolVar(&cfg.Enabled, "tenant-federation.enabled", false, "If enabled, multi tenant query federation can be used by supplying multiple tenant IDs in the read path (experimental).")
```
jtlisi, line 13: **suggestion(optional):** I think this help text can be reworked to explain how to use multi-tenant queries using `|` character in the X-Scope-OrgID.

**nit(optional,if-minor):** It might make sense to reword the phrase `multi tenant query federation can be used`. Changing it to something that describes query data from multiple tenants without using phrasing similar to the config field. Ideally, the help text should describe what a feature means to someone who can't parse the meaning from the config field name.

jtlisi (commented):
LGTM

Most of the nits/suggestions are very minor except for the cachegen topic. Once that is sorted I'll  approve and see if we can get a second reviewer on this.

pkg/frontend/v1/frontend.go
```diff
@@ -250,17 +250,24 @@ func getQuerierID(server frontendv1pb.Frontend_ProcessServer) (string, error) {
...
 
 	req.enqueueTime = time.Now()
 	req.queueSpan, _ = opentracing.StartSpanFromContext(ctx, "queued")
 
-	maxQueriers := f.limits.MaxQueriersPerUser(userID)
+	// figure out the highest max querier per user
+	var maxQueriers int
+	for _, tenantID := range tenantIDs {
+		v := f.limits.MaxQueriersPerUser(tenantID)
+		if v > maxQueriers {
```
csmarchbanks, line 265: Should maxQueriers be the minimum value from all of the tenants being queried? Otherwise wouldn't it be possible to exceed the limit for a tenant with a low max queriers setting?

csmarchbanks (commented):
Nice work!

Do we want to have an entirely new package for `tenantfederation`? It seems like all that is there is a new type of queryable which could also just be added to the `querier` package. Similarly, should the configuration be a new top level, or nested under querier? One thing that would help me decide is if there is more cross cutting flags that will be needed? Right now the flag is only used in querier related code.

integration/querier_tenant_federation_test.go
```diff
@@ -0,0 +1,181 @@
...
+	memcached := e2ecache.NewMemcached()
+	consul := e2edb.NewConsul()
+	require.NoError(t, s.StartAndWaitReady(consul, memcached))
+
+	flags := mergeFlags(BlocksStorageFlags(), map[string]string{
+		"-querier.cache-results":             "true",
+		"-querier.split-queries-by-interval": "24h",
+		"-querier.query-ingesters-within":    "12h", // Required by the test on query /series out of ingesters time range
+		"-frontend.memcached.addresses":      "dns+" + memcached.NetworkEndpoint(e2ecache.MemcachedPort),
+		"-tenant-federation.enabled":         "true",
```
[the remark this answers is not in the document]
simonswine: @jtlisi Not entirely sure if you meant to do this locally once or do you want to have that as part of the integration test run by CI?

pkg/frontend/v1/frontend.go
simonswine: Very good point, I was also missing the handling of the `0` and the same code in the query scheduler.

I have done that now in this commit: https://github.com/cortexproject/cortex/pull/3250/commits/3bc1204434e9988248ad14c07ef5f21f6fb756a9

pkg/querier/tenantfederation/tenant_federation.go
simonswine: @jtlisi Can you have another look. I tried to improve it according to both of your suggestions

integration/querier_tenant_federation_test.go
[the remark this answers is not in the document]
simonswine: This is still a TODO, will look into this tomorrow

pkg/cortex/cortex.go
```diff
@@ -301,6 +305,12 @@ func New(cfg Config) (*Cortex, error) {
 		os.Exit(0)
 	}
 
+	// Swap out the default resolver to support multiple tenant IDs separated by a '|'
+	if cfg.TenantFederation.Enabled {
+		util.WarnExperimentalUse("multi-tenant-query-federation")
```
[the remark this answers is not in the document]
simonswine: Good point done this in the latest iteration

pkg/api/middlewares.go
simonswine: I have deferred that to a future PR as suggested and added a TODO comment

simonswine:
@csmarchbanks 

> Do we want to have an entirely new package for `tenantfederation`? [...]

In this [thread] we had some discussion about that. Basically we would like to avoid that users can use multiple tenants in the write path, this requires us to return errors once a multi tenant header is used. We do that by swapping out the `tenant.Resolver` used. So users are expected to enable that on all their components used and not only in the query path. 

I think in the future that struct might evolve further, e.g. by allowing to configure the label used to expose the `tenant_id` from the currently used value of `__tenant_id__`

I hope that make all sense.


[thread]:https://github.com/cortexproject/cortex/pull/3250#discussion_r531708974

pkg/querier/merge_queryable.go
jtlisi: I see, not sure if we will be able to reuse it. However, the flexibility of this implementation doesn't hurt either.

integration/querier_tenant_federation_test.go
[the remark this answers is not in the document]
jtlisi: No need to commit any changes here. Just smoke testing that the tests pass if you enable it for globally on your local should be enough.

pkg/querier/tenantfederation/tenant_federation.go
jtlisi: 👍 Perfect!

jtlisi (approved):
LGTM

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,308 @@
...
+	return &mergeQueryable{
+		upstream: upstream,
+	}
+}
+
+type mergeQueryable struct {
+	upstream storage.Queryable
+}
+
+// Querier returns a new mergeQuerier aggregating all tenants into the result
```
csmarchbanks, line 38: Godoc style comments should end with a full stop. Looks like the full stop needs to be added to a few godoc comments in here.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,308 @@
...
+	return storage.NewMergeSeriesSet(seriesSets, storage.ChainedSeriesMerge)
+}
+
+// filterValuesByMatchers applies matchers to inputed labelName and labelValue.
+// The index of matched values is returned and also all label matcher not
+// matching the labelName.
+// In case a label matcher is set on a label confliciting with tenantLabelName,
+// we need to rename this labelMatcher to it's original name.
+// This is used to as part of Select in the mergeQueryable, to ensure only
+// relevant queries are considered and the forwarded matchers do not container
```
csmarchbanks, line 206: I think `container` should be `contain`

csmarchbanks (commented):
Thank you for the context on the package, I missed that other thread.

I left a few questions/comments, all pretty minor, otherwise LGTM.

integration/querier_tenant_federation_test.go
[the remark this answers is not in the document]
simonswine: I have done that today and it didn't not come up with any unexpected failure (apart from the `Backward_compatibility` tests, which is expected as the feature flag is not recognised by them)

pkg/frontend/v1/frontend.go
```diff
@@ -256,17 +256,21 @@ func getQuerierID(server frontendv1pb.Frontend_ProcessServer) (string, error) {
...
 
-	maxQueriers := f.limits.MaxQueriersPerUser(userID)
+	// aggregate the max queriers limit in the case of a multi tenant query
+	maxQueriers, err := tenant.AggregatedLimitMinimumOfNonZeroValues(ctx, f.limits.MaxQueriersPerUser)
+	if err != nil {
+		return err
+	}
 
-	err = f.requestQueue.EnqueueRequest(userID, req, maxQueriers, nil)
+	err = f.requestQueue.EnqueueRequest(tenant.JoinTenantIDs(tenantIDs), req, maxQueriers, nil)
```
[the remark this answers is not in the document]
simonswine: Yes I agree, this needs solving before we can move that feature from an experimental one. 

In its current form this allows to by-pass the fairness guarantees provided by the queues.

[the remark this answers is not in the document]
csmarchbanks: :+1: Thanks, just wanted to make sure my understanding was correct. Resolving this thread.

pkg/frontend/v1/frontend.go
```diff
@@ -256,17 +256,21 @@ func getQuerierID(server frontendv1pb.Frontend_ProcessServer) (string, error) {
...
 	if err != nil {
 		return err
 	}
 
 	req.enqueueTime = time.Now()
 	req.queueSpan, _ = opentracing.StartSpanFromContext(ctx, "queued")
 
-	maxQueriers := f.limits.MaxQueriersPerUser(userID)
+	// aggregate the max queriers limit in the case of a multi tenant query
+	maxQueriers, err := tenant.AggregatedLimitMinimumOfNonZeroValues(ctx, f.limits.MaxQueriersPerUser)
```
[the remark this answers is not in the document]
simonswine: `validation` sounds like a better place to me to be honest. How about a `validate.MinimumOfNonZeroValues([]int) (int)` as this would not need the user lookup at all?

pkg/querier/tenantfederation/merge_queryable.go
simonswine: :+1: I also rephrase it a bit

pkg/querier/tenantfederation/merge_queryable.go
simonswine: Good spot, my fingers a too used to adding the `-er` nowadays :slightly_smiling_face:

pkg/frontend/v1/frontend.go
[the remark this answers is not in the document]
csmarchbanks: Sounds good to me.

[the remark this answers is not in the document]
simonswine: 6513f87 should have that now

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
   - limit for incoming gRPC messages has changed from 4194304 to 104857600 bytes
 * [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts. #3305
 * [FEATURE] Query-frontend: introduced query statistics logged in the query-frontend when enabled via `-frontend.query-stats-enabled=true`. When enabled, the metric `cortex_query_seconds_total` is tracked, counting the sum of the wall time spent across all queriers while running queries (on a per-tenant basis). The metrics `cortex_request_duration_seconds` and `cortex_query_seconds_total` are different: the first one tracks the request duration (eg. HTTP request from the client), while the latter tracks the sum of the wall time on all queriers involved executing the query. #3539
+* [FEATURE] Querier: Queries can be federated across multiple tenants. The tenants can IDs involved need to be specified separated by a `|` character in the `X-Scope-OrgID` header. This is an experimental feature, which can be enabled by setting `-tenant-federation.enabled` on all components. #3250
```
pracucci, line 15: We've just cut `1.6.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry to the top, under the "master / unreleased" section, please?

csmarchbanks (approved):
:+1: Awesome work, @jtlisi do the new changes look good to you as well? If so, let's get this in.

I am excited to try out this functionality soon!

Looks like a merge conflict in the changelog as well.

jtlisi (approved):
LGTM, including recent changes

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+		)
+	}
+	lb.Set(additional.Name, additional.Value)
+}
+
+func setLabelsRetainExisting(src labels.Labels, additionalLabels ...labels.Label) labels.Labels {
+	lb := labels.NewBuilder(src)
+
+	for _, additionalL := range additionalLabels {
+		setLabelRetainExisting(src, lb, additionalL)
```
pstibrany, line 305: nit: I would suggest to inline this call. `setLabelRetainExisting` is not called anywhere else, and this method is somewhat empty.

pstibrany (commented):
Great job! I've left some comments, mostly nits, here and there.

pkg/querier/tenantfederation/tenant_federation.go
```diff
@@ -0,0 +1,14 @@
...
+	"flag"
+)
+
+type Config struct {
+	// Enabled switches on support for multi tenant query federation
+	Enabled bool `yaml:"enabled"`
+}
+
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.BoolVar(&cfg.Enabled, "tenant-federation.enabled", false, "If enabled on all components, queries can be federated across multiple tenants. The tenant IDs involved need to be specified separated by a `|` character in the `X-Scope-OrgID` header (experimental).")
```
pstibrany, line 13: Let's also mention experimental feature in the `v1-guarantees.md` file.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+	var matchedValuesPos = make([]int, len(labelValues))
+	for pos := range labelValues {
+		matchedValuesPos[pos] = pos
+	}
+
+	for _, m := range matchers {
+		if m.Name != labelName {
+			// check if has the retained label name
+			if m.Name == retainExistingPrefix+labelName {
+				unrelatedMatchers = append(unrelatedMatchers, &labels.Matcher{
```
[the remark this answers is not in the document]
pstibrany: I would suggest adding a test for this. When I tried manually changing line https://github.com/cortexproject/cortex/pull/3250/files#diff-c0734bde53174a20024532ea6f70ef263b28bd98b02be5331512209b8e25c9b1R239 to use `labels.MatchRegexp`, it did panic.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [FEATURE] Querier: Queries can be federated across multiple tenants. The tenants can IDs involved need to be specified separated by a `|` character in the `X-Scope-OrgID` header. This is an experimental feature, which can be enabled by setting `-tenant-federation.enabled` on all components. #3250
```
```suggestion
* [FEATURE] Querier: Queries can be federated across multiple tenants. The tenants IDs involved need to be specified separated by a `|` character in the `X-Scope-OrgID` request header. This is an experimental feature, which can be enabled by setting `-tenant-federation.enabled=true` on all Cortex services. #3250
```

pracucci (commented):
- Mention this features in the list of experimental features at `docs/configuration/v1-guarantees.md`

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+	labelNames, warnings, err := m.mergeDistinctStringSlice(func(q storage.Querier) ([]string, storage.Warnings, error) {
+		return q.LabelNames()
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// check if the tenant label exists in the original result
+	var tenantLabelExists bool
+	for _, labelName := range labelNames {
```
pracucci, line 118: `labelNames` is sorted, so you can use `sort.SearchStrings()`.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	if tenantLabelExists {
+		// the tenant label exists in the original result, so it had been retained
+		labelNames = append(labelNames, retainExistingPrefix+string(m.tenantLabelName))
+	} else {
+		// the tenant label was not existing, so it need to get added
+		labelNames = append(labelNames, string(m.tenantLabelName))
+	}
+	sort.Strings(labelNames)
```
pracucci, line 132: Since you're just appending a single string, it would be more efficient to append it in the right place (look for it using `sort.SearchStrings()`) instead of appending to the end and then resorting it entirely.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+
+	return labelNames, warnings, nil
+
```
pracucci, line 135: [nit] newline to be removed

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+		}
+		for _, e := range result {
+			resultMap[e] = struct{}{}
+		}
+		for _, w := range resultWarnings {
+			warnings = append(warnings, fmt.Errorf("error querying tenant id %s: %w", tenantID, w))
+		}
+	}
+
+	var result []string
```
pracucci, line 160:
```suggestion
	result = make([]string, 0, len(resultMap))
```

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+	for e := range resultMap {
+		result = append(result, e)
+	}
+	sort.Strings(result)
+	return result, warnings, nil
+}
+
+// Close releases the resources of the Querier.
+func (m *mergeQuerier) Close() error {
+	var errs prometheus.MultiError
```
pracucci, line 170: [nit] Use `github.com/prometheus/prometheus/tsdb/errors` `NewMulti()` instead please. I would also add a `faillint` to ensure none uses this `prometheus.MultiError`.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+	for pos, tenantID := range m.tenantIDs {
+		if err := m.queriers[pos].Close(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to close querier for tenant id %s: %w", tenantID, err))
+		}
```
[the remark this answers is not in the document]
pracucci: I was going to give the same comment:
> Instead of creating a new error using `fmt.Errorf()`, please wrap it with `errors.Wrapf(err, "...")`

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+
+// filterValuesByMatchers applies matchers to inputed labelName and labelValue.
+// The index of matched values is returned and also all label matcher not
+// matching the labelName.
+// In case a label matcher is set on a label confliciting with tenantLabelName,
+// we need to rename this labelMatcher to it's original name.
+// This is used to as part of Select in the mergeQueryable, to ensure only
+// relevant queries are considered and the forwarded matchers do not contain
+// matchers on the tenantLabelName.
+func filterValuesByMatchers(labelName string, labelValues []string, matchers ...*labels.Matcher) ([]int, []*labels.Matcher) {
```
pracucci, line 209: I would name returned parameters in this case, just for clarity:

```suggestion
func filterValuesByMatchers(labelName string, labelValues []string, matchers ...*labels.Matcher) (matchedValuesPos []int, unrelatedMatchers []*labels.Matcher) {
```

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+		if err := m.queriers[pos].Close(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to close querier for tenant id %s: %w", tenantID, err))
+		}
+	}
+	return errs.MaybeUnwrap()
+}
+
+// Select returns a set of series that matches the given label matchers. If the
+// tenantLabelName is matched on it only considers those queriers matching. The
+// forwaded labelSelector is not containing those that operate on tenantLabelName.
```
pracucci, line 181:
```suggestion
// forwarded labelSelector is not containing those that operate on tenantLabelName.
```

pracucci (commented):
Very good job! I left few comments. Also:

- `pkg/frontend/transport/handler.go` is still usig `tenant.TenantID`. How should it behave there?
- What do you think about adding a `faillint` rule to fail on `tenant.TenantID` used in `pkg/frontend/...`, `pkg/scheduler/...` and `pkg/querier/tenantfederation/...` to make sure we handle it everywhere (in the future)?
- What's about `tenant.TenantID`  usage in `pkg/querier/queryrange`? Have you checked where we should handle it? For example, I think `limitsMiddleware` should be tenants federation aware at least. This is another package where having a `faillint` may help to prevent future regressions.
-

pkg/scheduler/scheduler.go
```diff
@@ -258,6 +260,7 @@ func (s *Scheduler) enqueueRequest(frontendContext context.Context, frontendAddr
 	}
 
 	userID := msg.GetUserID()
+	userCtx := user.InjectOrgID(context.TODO(), userID)
```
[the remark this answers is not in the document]
simonswine: Done this as part of b09d07603a8eeae1bca69dc41e48cbd1bc18d303

pkg/util/validation/limits.go
```diff
@@ -421,3 +421,16 @@ func (o *Overrides) getOverridesForUser(userID string) *Limits {
 	}
 	return o.defaultLimits
 }
+
+// MinimumOfNonZeroValues is returning the minimal value of the input values.
+// In many limits a value of 0 means unlimted so the method will return 0 if
+// all inputs have a limit of 0. Otherwise it is the minimum of > 0 values.
+func MinimumOfNonZeroValues(inputs []int) int {
```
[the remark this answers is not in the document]
simonswine: I switched to `MaxQueriersPerTenant` as suggested somewhere else in b09d07603a8eeae1bca69dc41e48cbd1bc18d303

pkg/querier/tenantfederation/merge_queryable.go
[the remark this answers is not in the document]
simonswine: Done in bd8c0cbe621939a0c79ef7834e6ade628266ef8e

pkg/querier/tenantfederation/merge_queryable.go
simonswine: Done exactly this in bd8c0cbe621939a0c79ef7834e6ade628266ef8e

pkg/querier/tenantfederation/merge_queryable.go
simonswine: Good point done in 4e1c8900bf9bfc7fb33a7f0cfbbb32e8ad567662

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+// matchers on the tenantLabelName.
+func filterValuesByMatchers(labelName string, labelValues []string, matchers ...*labels.Matcher) ([]int, []*labels.Matcher) {
+	// this contains the matchers which are not related to labelName
+	var unrelatedMatchers []*labels.Matcher
+
+	// this contains the pos of labelValues that are matched by the matchers
+	var matchedValuesPos = make([]int, len(labelValues))
+	for pos := range labelValues {
+		matchedValuesPos[pos] = pos
+	}
```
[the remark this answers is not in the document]
simonswine: Very good point makes much more sense: 4e1c8900bf9bfc7fb33a7f0cfbbb32e8ad567662

pkg/querier/tenantfederation/merge_queryable.go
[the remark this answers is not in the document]
simonswine: Thank you for pointing that out, this could have been a painful bug. 4dbdc884e83b42649e81d66a4b50fc104c31a724 adds a fix and a test

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+
+}
+
+type stringSliceFunc func(storage.Querier) ([]string, storage.Warnings, error)
+
+// mergeDistinctStringSlice is aggregating results from stringSliceFunc calls
+// on a querier. It removes duplicates and sorts the result.
+//
+// TODO: Consider running stringSliceFunc calls concurrently
+func (m *mergeQuerier) mergeDistinctStringSlice(f stringSliceFunc) ([]string, storage.Warnings, error) {
```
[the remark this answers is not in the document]
simonswine: Added some comment in 43854f7737a7d5ca234a0d968707b897e618149e.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,309 @@
...
+// mergeQuerier aggregates the results from underlying queriers and adds a
+// label tenantLabelName to identify the tenant ID that the metric resulted
+// from.
+// If the label tenantLabelName is already existing, its value is
+// overwritten by the tenant ID and the previous value is exposed through a new
+// label prefixed with "original_". This behaviour is not implemented recursively
+type mergeQuerier struct {
+	queriers        []storage.Querier
+	tenantIDs       []string
+	tenantLabelName model.LabelName
```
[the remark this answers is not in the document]
simonswine: Good point, see 43854f7737a7d5ca234a0d968707b897e618149e.

pkg/querier/tenantfederation/tenant_federation.go
simonswine: This is done in 891e692f79a1a65fe6a33961b6bb8f13121a5253

CHANGELOG.md
simonswine: Thanks for that suggestion. Has been added.

pkg/api/middlewares.go
```diff
@@ -14,13 +14,15 @@ import (
...
-			userID, err := tenant.TenantID(r.Context())
+			tenantIDs, err := tenant.TenantIDs(r.Context())
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusUnauthorized)
 				return
 			}
 
-			cacheGenNumber := cacheGenNumbersLoader.GetResultsCacheGenNumber(userID)
+			// len(tenantIDs) will always be > 0, as it otherwise errors
+			// TODO: Handle multiple tenants by creating reproducible aggregation of all individual cacheGenNumbers
```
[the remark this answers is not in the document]
simonswine: Can you take a look at 891e692f79a1a65fe6a33961b6bb8f13121a5253

pkg/querier/tenantfederation/merge_queryable.go
simonswine: Good point

pkg/frontend/v1/frontend.go
```diff
@@ -256,17 +257,27 @@ func getQuerierID(server frontendv1pb.Frontend_ProcessServer) (string, error) {
...
+	var maxQueriers int
+	if len(tenantIDs) == 1 {
+		maxQueriers = f.limits.MaxQueriersPerUser(tenantIDs[0])
+	} else {
+		maxQueriersPerTenant := make([]int, len(tenantIDs))
+		for pos := range maxQueriersPerTenant {
+			maxQueriersPerTenant[pos] = f.limits.MaxQueriersPerUser(tenantIDs[pos])
+		}
+		maxQueriers = validation.MinimumOfNonZeroValues(maxQueriersPerTenant)
+	}
```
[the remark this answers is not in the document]
simonswine: Done as part of b09d07603a8eeae1bca69dc41e48cbd1bc18d303

pkg/scheduler/scheduler.go
```diff
@@ -271,7 +274,21 @@ func (s *Scheduler) enqueueRequest(frontendContext context.Context, frontendAddr
...
+	var maxQueriers int
+	if len(tenantIDs) == 1 {
+		maxQueriers = s.limits.MaxQueriersPerUser(tenantIDs[0])
+	} else {
+		maxQueriersPerTenant := make([]int, len(tenantIDs))
+		for pos := range maxQueriersPerTenant {
+			maxQueriersPerTenant[pos] = s.limits.MaxQueriersPerUser(tenantIDs[pos])
+		}
+		maxQueriers = validation.MinimumOfNonZeroValues(maxQueriersPerTenant)
+	}
```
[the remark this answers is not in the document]
simonswine: Done as part of b09d07603a8eeae1bca69dc41e48cbd1bc18d303

pkg/cortex/modules.go
```diff
@@ -203,7 +204,15 @@ func (t *Cortex) initQueryable() (serv services.Service, err error) {
 	querierRegisterer := prometheus.WrapRegistererWith(prometheus.Labels{"engine": "querier"}, prometheus.DefaultRegisterer)
 
 	// Create a querier queryable and PromQL engine
-	t.QuerierQueryable, t.QuerierEngine = querier.New(t.Cfg.Querier, t.Overrides, t.Distributor, t.StoreQueryables, t.TombstonesLoader, querierRegisterer)
+	var queryable prom_storage.SampleAndChunkQueryable
+	queryable, t.QuerierEngine = querier.New(t.Cfg.Querier, t.Overrides, t.Distributor, t.StoreQueryables, t.TombstonesLoader, querierRegisterer)
+
+	// Enable merge querier if multi tenant query federation is enabled
+	if t.Cfg.TenantFederation.Enabled {
+		queryable = querier.NewSampleAndChunkQueryable(tenantfederation.NewQueryable(queryable))
```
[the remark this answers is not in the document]
simonswine: The proposal lists that under [future work] , right now I think we should not enable that just yet.

[future work]: https://github.com/cortexproject/cortex/blob/master/docs/proposals/cross-tenant-query-federation.md#cross-tenant-support-for-the-ruler

pkg/querier/tenantfederation/merge_queryable.go
simonswine: Inlined this per 6dde8c29a8827e0d0deb3bb32e809a846d37d932

simonswine:
Thank you both for your reviews, I have addressed most point, but I will need a bit more time tomorrow morning to take a look at the comments around usages of `TenantID` and future proofing changes through `faillint` as @pracucci suggested.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,311 @@
...
+	sort.Strings(result)
+	return result, warnings, nil
+}
+
+// Close releases the resources of the Querier.
+func (m *mergeQuerier) Close() error {
+	errs := tsdb_errors.NewMulti()
+	for pos, tenantID := range m.tenantIDs {
+		if err := m.queriers[pos].Close(); err != nil {
+			errs.Add(errors.Wrapf(m.queriers[pos].Close(), "failed to close querier for tenant id %s", tenantID))
```
pracucci, line 177: This is not correct. You're closing twice.

docs/guides/limitations.md
```diff
@@ -43,3 +43,7 @@ The Cortex chunks storage doesn't support queries without a metric name, like `c
 ## Query series and labels
 
 When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
+
+## Tenant federation
+
+When using tenant federation on a Cortex cluster with chunk storage, cache invalidation is not fully supported. Metrics that have been deleted can still appear in cached query results.
```
pracucci, line 49: I just would be clear about what "series deletion" means in this context, to avoid any misunderstanding:

```suggestion
When using tenant federation on a Cortex cluster with chunk storage, cache invalidation is not fully supported. Metrics that have been deleted via `/api/v1/admin/tsdb/delete_series` API endpoint can still appear in cached query results.
```

pkg/util/validation/limits.go
```diff
@@ -422,3 +422,18 @@ func (o *Overrides) getOverridesForUser(userID string) *Limits {
 	}
 	return o.defaultLimits
 }
+
+// MaxQueriersPerTenants is returning the smallest maxQueriers value of all
+// given tenant IDs. A value of 0 means unlimted queriers, so the method will
+// return 0 only if all inputs have a limit of 0. Otherwise it is the minimum
+// of > 0 values.
+func MaxQueriersPerTenants(tenantIDs []string, maxQueriersPerTenant func(string) int) int {
```
pracucci, line 430: This is one way of doing it, but I believe would be easier to use this was on `o *Overrides` so you don't need to pass `maxQueriersPerTenant` in input.

pracucci (approved):
Thanks for addressing my feedback! LGTM, but I would be glad to see last comments addressed, as well as the previous comments about `faillint`, before merging. However, if you can't address `faillint` comments in this PR, we could merge and address them in a follow up PR (but I'm quite keen on that and would love not loosing them).

pkg/querier/tenantfederation/merge_queryable.go
pstibrany: Right. `if err := ...` part should be removed. Both `Wrapf` and `Add` already handle case of `nil` error.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	return errs.Err()
+}
+
+// Select returns a set of series that matches the given label matchers. If the
+// tenantLabelName is matched on it only considers those queriers matching. The
+// forwarded labelSelector is not containing those that operate on
+// tenantLabelName.
+func (m *mergeQuerier) Select(sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
+	matchedTenants, filteredMatchers := filterValuesByMatchers(string(defaultTenantLabel), m.tenantIDs, matchers...)
```
pstibrany, line 188: Nit: no need to typecast `string(defaultTenantLabel)`

pkg/querier/tenantfederation/tenant_federation.go
pstibrany, line 13: > `If enabled on all components`

All components is somewhat wide... this is only required in query-frontend and querier, right? No other components can use this at the moment. I think we should be bit more specific, to allow people configure it correctly and also not give them false impression that other components use it somehow.

pkg/util/validation/limits.go
pstibrany: I would suggest different name: `MinimumOfMaxQueriersPerTenants` or `SmallestMaxQueriersPerTenant`. Now it looks like returning "max" value.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
...
 * [CHANGE] Blocks storage: compactor is now required when running a Cortex cluster with the blocks storage, because it also keeps the bucket index updated. #3583
 * [CHANGE] Blocks storage: block deletion marks are now stored in a per-tenant global markers/ location too, other than within the block location. The compactor, at startup, will copy deletion marks from the block location to the global location. This migration is required only once, so you can safely disable it via `-compactor.block-deletion-marks-migration-enabled=false` once new compactor has successfully started once in your cluster. #3583
+* [FEATURE] Querier: Queries can be federated across multiple tenants. The tenants IDs involved need to be specified separated by a `|` character in the `X-Scope-OrgID` request header. This is an experimental feature, which can be enabled by setting `-tenant-federation.enabled=true` on all Cortex services. #3250
```
pstibrany, line 8: Why all? Let's be specific to avoid confusing users.

pstibrany (approved):
Thanks for addressing my feedback. LGTM with few nit comments left.

pkg/querier/tenantfederation/merge_queryable.go
simonswine: Thanks very good catch and a lapse that can be dangerous: 42663d136

docs/guides/limitations.md
simonswine: Updated as of 9bfc3af1019ef4062ab1c4ef47f569d583a3a95b

pkg/querier/tenantfederation/merge_queryable.go
simonswine: Good point: e51291b2678580be88bfd2a24554bfcfd4ac13c2

CHANGELOG.md
simonswine: I think it needs to be enabled on all Cortex services, because otherwise e.g. the write path would interpret multiple tenants as a single new tenant. Once it is enabled the write path fails as it only supports a single tenant. I have aligned the CHANGELOG with the flags description.

pkg/querier/tenantfederation/tenant_federation.go
simonswine: I have aligned that to the CHANGELOG, see https://github.com/cortexproject/cortex/pull/3250#discussion_r547935405

pkg/util/validation/limits.go
simonswine: As discussed on slack, I have removed the method again from override and implement more generic helpers. Not that I am tottally happy with their names, but they should include all previous feedback about naming them:

9826281d5fd8726c5b8db145710cff2d8b9a0e78
0667242d7e0f347a350e8be367140d0f64524d11

simonswine:
>     * `pkg/frontend/transport/handler.go` is still usig `tenant.TenantID`. How should it behave there?
> 
>     * What do you think about adding a `faillint` rule to fail on `tenant.TenantID` used in `pkg/frontend/...`, `pkg/scheduler/...` and `pkg/querier/tenantfederation/...` to make sure we handle it everywhere (in the future)?
> 
>     * What's about `tenant.TenantID`  usage in `pkg/querier/queryrange`? Have you checked where we should handle it? For example, I think `limitsMiddleware` should be tenants federation aware at least. This is another package where having a `faillint` may help to prevent future regressions.

@pracucci I think I have addressed all of the remaining points (I think :slightly_smiling_face: ).

Esp those ones had the biggest changes, so certainly are in need for another review:

* The result cache gets skipped once there are more than one tenant b41a4df64a23b05c9d4702fc66e459171937369f
* limitsMiddleware is now using the "strictest" of the tenant's limits 0667242d7e0f347a350e8be367140d0f64524d11  
* Fail lint is checking the packages in the query path: e312687ac7538cd81a6c701a93711f31b78427f5

pracucci:
Thanks a lot @simonswine for your hard work and patiently address all comments! I agree on skipping the query results caching in this PR and address it in a follow up PR (given we have to design how to deal with cache invalidation done for deleted series).

bhiravabhatla:
When is this getting released. I see tenant_federation in cortex configuration documentation - but I dont think this enhancement was release as part of 1.6.

bhiravabhatla:
Is this released as part of 1.6. Could not find this in release notes. Its already present in configuration documentation though.

pstibrany:
> Is this released as part of 1.6. Could not find this in release notes. Its already present in configuration documentation though.

It will be part of Cortex 1.7.

Documentation on the website currently reflects `master` branch, with latest changes.

cabrinha:
In order to add Cortex as a datasource for grafana, when using auth and tenants... Whats the config for the datasource need to look like?

simonswine:
> In order to add Cortex as a datasource for grafana, when using auth and tenants... Whats the config for the datasource need to look like in order to see every tenant in one datasource?

@cabrinha unfortunately there is no way to specify this currently. You need to list the tenants explicitly (separated by a `|`). If you care about this enough it would be good to create a new issue with a feature request and detailing your use case a bit more.

PR #3252, opened 2020-09-30
Add ingesters shuffle sharding support on the read path

pracucci:
**What this PR does**:
This PR aims to implement the ingesters shuffle sharding on the read path, as described in [this design doc](https://cortexmetrics.io/docs/proposals/shuffle-sharding-on-the-read-path/#querier-ingesters-shuffle-sharding).

The idea is that, when distributors/ingesters shuffle sharding is enabled, we only query data from the subset of ingesters holding the data for a specific tenant.

This PR builds on top of #3299.

_Documentation has not been updated yet._

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/ingester_sharding_test.go
+++ b/integration/ingester_sharding_test.go
@@ -8,7 +8,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/cortexproject/cortex/integration/e2e"
@@ -43,8 +45,15 @@ func TestIngesterSharding(t *testing.T) {
 			defer s.Close()
 
 			flags := BlocksStorageFlags
+			flags["-distributor.shard-by-all-labels"] = "true"
 			flags["-distributor.sharding-strategy"] = testData.shardingStrategy
 			flags["-distributor.ingestion-tenant-shard-size"] = strconv.Itoa(testData.tenantShardSize)
+
+			if testData.shardingStrategy == "shuffle-sharding" {
+				// Enable shuffle sharding on read path but not lookback, otherwise all ingesters would be
+				// queried being just registered.
+				flags["-querier.shuffle-sharding-ingesters-lookback-period"] = "1ns"
+			}
 
 			// Start dependencies.
 			consul := e2edb.NewConsul()
@@ -56,6 +65,7 @@ func TestIngesterSharding(t *testing.T) {
 			ingester1 := e2ecortex.NewIngester("ingester-1", consul.NetworkHTTPEndpoint(), flags, "")
 			ingester2 := e2ecortex.NewIngester("ingester-2", consul.NetworkHTTPEndpoint(), flags, "")
 			ingester3 := e2ecortex.NewIngester("ingester-3", consul.NetworkHTTPEndpoint(), flags, "")
+			ingesters := e2ecortex.NewCompositeCortexService(ingester1, ingester2, ingester3)
 			querier := e2ecortex.NewQuerier("querier", consul.NetworkHTTPEndpoint(), flags, "")
 			require.NoError(t, s.StartAndWaitReady(distributor, ingester1, ingester2, ingester3, querier))
 
@@ -70,15 +80,19 @@ func TestIngesterSharding(t *testing.T) {
 
 			// Push series.
 			now := time.Now()
+			expectedVectors := map[string]model.Vector{}
 
-			client, err := e2ecortex.NewClient(distributor.HTTPEndpoint(), "", "", "", userID)
+			client, err := e2ecortex.NewClient(distributor.HTTPEndpoint(), querier.HTTPEndpoint(), "", "", userID)
 			require.NoError(t, err)
 
 			for i := 1; i <= numSeriesToPush; i++ {
-				series, _ := generateSeries(fmt.Sprintf("series_%d", i), now)
+				metricName := fmt.Sprintf("series_%d", i)
+				series, expectedVector := generateSeries(metricName, now)
 				res, err := client.Push(series)
 				require.NoError(t, err)
 				require.Equal(t, 200, res.StatusCode)
+
+				expectedVectors[metricName] = expectedVector
 			}
 
 			// Extract metrics from ingesters.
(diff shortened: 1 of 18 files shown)


pkg/ring/ring_test.go
```diff
@@ -609,6 +610,356 @@ func TestRing_ShuffleShard_Consistency(t *testing.T) {
...
+					rs, err := ring.ShuffleShardWithLookback(userID, event.shardSize, lookbackPeriod).GetAll(Read)
+					require.NoError(t, err)
+					assert.ElementsMatch(t, event.expected, rs.GetAddresses())
+				}
+			}
+		})
+	}
+}
+
+func TestRing_ShuffleShardWithLookback_CorrectnessWithFuzzy(t *testing.T) {
```
pracucci, line 839: Ignore this test for now because it's not correct. I should mock `time.Now()` in `ShuffleShardWithLookback()`. I'm working on a fix.

pracucci: After some investigation I've found out that, when the shard size is increased (eg. from 3 to 6) and there are multiple zones, the "consistency" property is not guaranteed (you can end up with all new instances in the subring) and this breaks this test.

_I've pushed a version of this test that breaks because of this. I will fix the issue in a separate PR, to keep the diff easier to review._

pracucci: With the fix done in #3299 this test passes.

pkg/querier/querier.go
```diff
@@ -64,10 +64,13 @@ type Config struct {
...
 	SecondStoreEngine        string       `yaml:"second_store_engine"`
 	UseSecondStoreBeforeTime flagext.Time `yaml:"use_second_store_before_time"`
+
+	ShuffleShardingIngestersLookbackPeriod time.Duration `yaml:"shuffle_sharding_ingesters_lookback_period"`
 }
 
 var (
-	errBadLookbackConfigs = errors.New("bad settings, query_store_after >= query_ingesters_within which can result in queries not being sent")
+	errBadLookbackConfigs                             = errors.New("bad settings, query_store_after >= query_ingesters_within which can result in queries not being sent")
+	errShuffleShardingLookbackLessThanQueryStoreAfter = errors.New("the shuffle-sharding lookback period should be greater or equal than the configured 'query store after'")
```
pstibrany, line 73: This is useful info and I'd recommend we put it into description of -querier.shuffle-sharding-ingesters-lookback-period flag.

pkg/querier/querier.go
```diff
@@ -88,18 +91,24 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.LookbackDelta, "querier.lookback-delta", 5*time.Minute, "Time since the last sample after which a time series is considered stale and ignored by expression evaluations.")
 	f.StringVar(&cfg.SecondStoreEngine, "querier.second-store-engine", "", "Second store engine to use for querying. Empty = disabled.")
 	f.Var(&cfg.UseSecondStoreBeforeTime, "querier.use-second-store-before-time", "If specified, second store is only used for queries before this timestamp. Default value 0 means secondary store is always queried.")
+	f.DurationVar(&cfg.ShuffleShardingIngestersLookbackPeriod, "querier.shuffle-sharding-ingesters-lookback-period", 0, "When distributor's sharding strategy is shuffle-sharding and this setting is > 0, queriers fetch in-memory series from the minimum set of required ingesters, selecting only ingesters which may have received series since 'now - lookback period'. If this setting is 0, queriers always query all ingesters (ingesters shuffle sharding on read path is disabled).")
```
pstibrany, line 94: Advice on what is the reasonable value to use or how to compute it would be useful.

pkg/ring/ring.go
```diff
@@ -471,18 +474,32 @@ func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
...
+}
 
-	// This deferred function will store newly computed ring into cache.
-	// This needs to happen *after* releasing Read lock (otherwise we get a deadlock).
-	defer func() {
-		// We cannot defer setCachedShuffledSubring directly, as result is still nil at this point,
-		// and that would be passed to the function, instead of updated value.
-		r.setCachedShuffledSubring(identifier, size, result)
-	}()
+func (r *Ring) shuffleShard(identifier string, size int, lookbackPeriod time.Duration, now time.Time) ReadRing {
```
pstibrany, line 499: Nit: We can return `*Ring` here to avoid cast in `ShuffleShard`.

pkg/ring/ring.go
```diff
@@ -471,18 +474,32 @@ func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
...
+	lookbackUntil := now.Add(-lookbackPeriod).Unix()
+
+	var result *Ring
```
pstibrany, line 502: `result` can be removed, and new ring returned directly.

pstibrany (approved):
LGTM, great job!

pkg/querier/querier.go
pracucci: Since adding all such details here may be "complicated" I was thinking to do it in the doc. WDYT?

pstibrany: Sounds good.

pkg/ring/ring_test.go
```diff
@@ -759,6 +761,369 @@ func TestRing_ShuffleShard_ConsistencyOnZonesChanged(t *testing.T) {
...
+				{what: add, instanceID: "instance-3", instanceDesc: generateRingInstanceWithInfo("instance-3", "zone-c", []uint32{userToken(userID, "zone-c", 2) + 2}, now.Add(-2*lookbackPeriod))},
+				{what: add, instanceID: "instance-4", instanceDesc: generateRingInstanceWithInfo("instance-4", "zone-a", []uint32{userToken(userID, "zone-a", 0) + 3}, now.Add(-2*lookbackPeriod))},
+				{what: add, instanceID: "instance-5", instanceDesc: generateRingInstanceWithInfo("instance-5", "zone-b", []uint32{userToken(userID, "zone-b", 1) + 3}, now.Add(-2*lookbackPeriod))},
+				{what: add, instanceID: "instance-6", instanceDesc: generateRingInstanceWithInfo("instance-6", "zone-c", []uint32{userToken(userID, "zone-c", 2) + 3}, now.Add(-2*lookbackPeriod))},
+				{what: test, shardSize: 3, expected: []string{"instance-1", "instance-2", "instance-3"}},
+				// Scale up.
+				{what: add, instanceID: "instance-7", instanceDesc: generateRingInstanceWithInfo("instance-7", "zone-a", []uint32{userToken(userID, "zone-a", 0) + 1}, now)},
```
ranton256, line 907: What about testing the behavior in between, when a full set of instances across each zone is only partly added, i.e. you have added in zone a, and b, but not c?

ranton256 (commented):
I have a few questions and comments, nothing I found was related to correctness or blocking.

integration/ingester_sharding_test.go
```diff
@@ -43,9 +45,16 @@ func TestIngesterSharding(t *testing.T) {
 			defer s.Close()
 
 			flags := BlocksStorageFlags
+			flags["-distributor.shard-by-all-labels"] = "true"
```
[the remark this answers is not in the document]
pracucci: No, it doesn't. I enabled it in this test just because it's how we (and to my knowledge most of users) run Cortex, so to have an integration test which better reflects real setups.

pkg/distributor/distributor.go
```diff
@@ -622,16 +625,8 @@ func (d *Distributor) send(ctx context.Context, ingester ring.IngesterDesc, time
...
 }
 
-// ForAllIngesters runs f, in parallel, for all ingesters
-func (d *Distributor) ForAllIngesters(ctx context.Context, reallyAll bool, f func(context.Context, client.IngesterClient) (interface{}, error)) ([]interface{}, error) {
-	replicationSet, err := d.ingestersRing.GetAll(ring.Read)
-	if err != nil {
-		return nil, err
-	}
-	if reallyAll {
-		replicationSet.MaxErrors = 0
```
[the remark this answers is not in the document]
pracucci: Good question. `reallyAll` was used only by `UserStats()` to make sure the request will fail if querying any ingester will fail (by default we allow 1 failure, due to quorum). The logic has **not** changed: I've just moved `replicationSet.MaxErrors = 0` to `UserStats()`.

pkg/ring/replication_set.go
```diff
@@ -85,6 +85,15 @@ func (r ReplicationSet) Includes(addr string) bool {
 	return false
 }
 
+// GetAddresses returns the addresses of all instances within the replication set.
+func (r ReplicationSet) GetAddresses() []string {
```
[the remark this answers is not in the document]
pracucci: Yep, done!

pkg/ring/ring_test.go
pracucci: Makes a lot of sense. Done it (and I did the same for the scale down test too).

tools/test
```diff
@@ -7,7 +7,7 @@ SLOW=
 TAGS=
 PARALLEL=
 RACE="-race -covermode=atomic"
-TIMEOUT=5m
+TIMEOUT=10m
```
pracucci, line 10: Tests run with `-race` in CI and, when race is enabled, they're very slow, especially new "stress" tests I've added for shuffle sharding. However, at this stage I believe we should prefer coverage over speed. We'll later work to speed up CI.

ranton256 (commented):
LGTM, thanks for the answers and the unit test.

PR #3254, opened 2020-09-30
Update comment to reflect current state.

gouthamve:
Prometheus now allows context, but the problem is the interface
`RulesRetriever` can never fail rule lookups. I can't see a good fix for
it right away, so decided to fix the comment for now.


jtlisi (approved):
LGTM

Progress at the very least

gotjosh (approved):
😭  The dream is no longer real.

pracucci (approved):

PR #3255, opened 2020-09-30
Make the Query Frontend horizontally scalable.

### LINKED ISSUE #1150: Query Frontend scalability
We're staring to have enough sustained query load that the limit of having two query frontends is hurting us.  The query frontends currently have to unmarshal the JSON from the querier (and the proto from the cache), combine them and then remarshal and send them.

The limit of two query frontends is so the queueing is fair.  The more replicas you add, the more queues you'll end up with and the more it will degrade to random load balancing, the problem the frontend was designs to solve.

One idea is to put the queue in a separate service.  This service would be responsible for queueing, scheduling, retires etc.  The frontend would then only be responsible for accepting the queries, consulting the cache, working out what to enqueue, accepting the response, writing back to the cache and the user.  This would allow the frontend to be stateless and scalable.  The queries would communicate directly with the frontend service for the bulk of the response; the new queueing service would be control-path only.

WDYT?

tomwilkie:
**What this PR does**:

Make the Query Frontend horizontally scalable.

- [x] Define the APIs.
- [x] Split out the logic into a separate struct.
- [ ] Figure out backwards compatibility.
- [ ] Make the scheduler a separate component.

Fixes #1150 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -421,7 +421,9 @@ func (a *API) registerQueryAPI(handler http.Handler) {
 // Cortex querier service. Currently this can not be registered simultaneously
 // with the Querier.
 func (a *API) RegisterQueryFrontend(f *frontend.Frontend) {
-	frontend.RegisterFrontendServer(a.server.GRPC, f)
+	frontend.RegisterFrontendV2Server(a.server.GRPC, f)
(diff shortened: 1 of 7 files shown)


pracucci:
Thanks @tomwilkie for kickstarting this work! I'm closing because superseeded by #3374.

PR #3256, opened 2020-09-30
Give ingester some timestamp tolerance

bboreham:
Inspired by https://github.com/prometheus/prometheus/pull/7976

If timestamps sent by remote-write clients are slightly off an exact period of seconds, quantize them to be exact.  This saves memory and storage space.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -568,6 +568,10 @@ lifecycler:
   # CLI flag: -ingester.availability-zone
   [availability_zone: <string> | default = ""]
 
+# quantize ingested timestamps to within this tolerance
+# CLI flag: -ingester.timestamp-tolerance
+[timestampTolerance: <int> | default = 0]
+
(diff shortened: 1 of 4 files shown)


roidelapluie:
Tackling this in cortex will likely be an issue whith deduplication when reading back data with prometheus remote read.

bboreham:
Good point.

I can see that there could be additional points in the series, causing more work in Prometheus, but do you think that is actually visible to an end-user via PromQL?

roidelapluie:
> Good point.
> 
> I can see that there could be additional points in the series, causing more work in Prometheus, but do you think that is actually visible to an end-user via PromQL?

It would break irate() and idelta()-like functions, they would be zero-ed.

bboreham:
> irate()

Good point.  I guess we could make it optional for people like me who never use `irate()`, or never use `remote_write`, but I'm not likely to progress this in the near term.

PR #3259, opened 2020-10-01
Disallow stale reads for Consul CAS

bboreham:
In order for the CAS to succeed, we need to have up-to-date data to compare against, so it doesn't make sense to allow stale reads.

More info: https://www.consul.io/api-docs/features/consistency

**Checklist**
- NA Tests updated
- NA Documentation added
- [x] `CHANGELOG.md` updated


pstibrany (approved):
Makes sense, good find!

pracucci (approved):
LGTM!

PR #3263, opened 2020-10-01
Use Promauto.With() where possible

roidelapluie:
Also ensures that other standalone objects are registered by adding a
registerer to their New() function.

**What this PR does**:

Partially #2204 (still miss the default registry metrics)

**Checklist**
- [ ] ~Tests updated~ there are existing tests covering this
- [ ] ~Documentation added~
- [ ] ~`CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`~

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
 	tsdberrors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/weaveworks/common/instrument"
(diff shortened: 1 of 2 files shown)


pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -47,8 +47,8 @@ type alertmanagerMetrics struct {
 	silencesPropagatedMessagesTotal *prometheus.Desc
 }
 
-func newAlertmanagerMetrics() *alertmanagerMetrics {
-	return &alertmanagerMetrics{
+func newAlertmanagerMetrics(r prometheus.Registerer) *alertmanagerMetrics {
```
pstibrany, line 50: `alertmanagerMetrics` implements `prometheus.Collector` interface. `Collectors` typically don't register into registry themselves, but are registered by client code using them. I think we should keep that, and not pass `registerer` into `newAlertmanagerMetrics`.

pkg/util/metrics_helper.go
```diff
@@ -465,11 +465,15 @@ func (d *HistogramData) Metric(desc *prometheus.Desc, labelValues ...string) pro
 }
 
 // Creates new histogram data collector.
-func NewHistogramDataCollector(desc *prometheus.Desc) *HistogramDataCollector {
-	return &HistogramDataCollector{
+func NewHistogramDataCollector(r prometheus.Registerer, desc *prometheus.Desc) *HistogramDataCollector {
```
pstibrany, line 468: As commented earleir, collectors should not require registrerer.

pstibrany (commented):
Thank you. The primary goal of #2204 was to get rid of using global registry. I don't think adding dependency on `prometheus.Registrerer` to types implementing `prometheus.Collector` is a way to go here.

pstibrany:
In general, I don't think object should register itself anywhere in its "new" method. (I'm sure you will find many counter-examples in Cortex codebase).

pkg/ruler/manager.go
```diff
@@ -54,10 +54,7 @@ func NewDefaultMultiTenantManager(cfg Config, managerFactory ManagerFactory, reg
 		return nil, err
 	}
 
-	userManagerMetrics := NewManagerMetrics()
-	if reg != nil {
-		reg.MustRegister(userManagerMetrics)
-	}
+	userManagerMetrics := NewManagerMetrics(reg)
```
pracucci, line 57: You can now initialise it directly below:

```
userManagerMetrics: NewManagerMetrics(reg),
```

roidelapluie: Do we want this New(reg) pattern or not? @pstibrany says no

pracucci: Sorry @roidelapluie for the long reply. Me and Peter had a discussion offline. We have a different opinion here. I commit to Peter's take: given `ManagerMetrics` is a `Collector`, and a `Collector` shouldn't take the registerer, let's not pass the registerer to `NewManagerMetrics()`.

_Same applies to alertmanager metrics._

pracucci (approved):
LGTM, thanks!

pstibrany (approved):

PR #3269, opened 2020-10-02
Split listing and loading rule groups in rules.RuleStore interface

pstibrany:
**What this PR does**: This PR splits listing and loading rules in `rules.RuleStore` interface and most importantly `objectclient.RuleStore` implementation. This makes it more efficient, since we don't need to load rules that ruler doesn't own. Other implementations (local, configdb) still combine listing and loading.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ruler/ruler.go
+++ b/pkg/ruler/ruler.go
@@ -50,6 +50,7 @@ const (
 	ShardingStrategyDefault = "default"
 	ShardingStrategyShuffle = "shuffle-sharding"
 
+	// Number of concurrent group list and group loads operations.
 	loadRulesConcurrency = 10
 
 	rulerSyncReasonInitial    = "initial"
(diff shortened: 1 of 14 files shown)


pkg/ruler/rules/store.go
```diff
@@ -70,7 +73,7 @@ func NewConfigRuleStore(c client.Client) *ConfigRuleStore {
 }
 
 func (c *ConfigRuleStore) ListAllUsers(ctx context.Context) ([]string, error) {
```
pracucci, line 75: Once this PR is merged (to keep diff easier), I would sugget to move `ConfigRuleStore` to a dedicated package `pkg/ruler/rules/configdb/` and keep in this file only the generic `type` definitions.

pkg/ruler/rules/store.go
```diff
@@ -19,16 +19,19 @@ var (
...
 type RuleStore interface {
 	ListAllUsers(ctx context.Context) ([]string, error)
+	ListAllRuleGroups(ctx context.Context) (map[string]RuleGroupList, error)
+	// ListRuleGroupsForUserAndNamespace returns all the active rule groups for a user from given namespace.
+	// If namespace is empty, groups from all namespaces are returned.
+	ListRuleGroupsForUserAndNamespace(ctx context.Context, userID string, namespace string) (RuleGroupList, error)
 
-	// Returns all rule groups, and loads rules for each group.
-	LoadAllRuleGroups(ctx context.Context) (map[string]RuleGroupList, error)
+	// LoadRuleGroups loads rules for each rule group in the map.
```
pracucci, line 32: I would clarify that the `groupsToLoad` **must** be the output of either `ListAllRuleGroups` or `ListRuleGroupsForUserAndNamespace`. This comment is required because the internal implemention expects it (some `LoadRuleGroups()` implementations are no-op because groups are already loaded when listing).

pracucci (approved):
Very good job! Changes LGTM and it's a nice improvement over the current situation. I just left few nits.

pkg/ruler/rules/objectclient/rule_store.go
```diff
@@ -27,24 +27,25 @@ import (
...
+func NewRuleStore(client chunk.ObjectClient, loadConcurrency int) *RuleStore {
 	return &RuleStore{
-		client: client,
+		client:          client,
+		loadConcurrency: loadConcurrency,
 	}
 }
 
-func (o *RuleStore) getRuleGroup(ctx context.Context, objectKey string) (*rules.RuleGroupDesc, error) {
+func (o *RuleStore) getRuleGroup(ctx context.Context, objectKey string, rg *rules.RuleGroupDesc) (*rules.RuleGroupDesc, error) {
```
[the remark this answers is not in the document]
pstibrany: Main reason for using `rg` is so that we can easily update entries in the map in `LoadRuleGroups` :) I will document it.

pkg/ruler/rules/store.go
pstibrany: Very good point, I'll add it.

pstibrany:
@jtlisi @gotjosh Do you want to review this before merging?

jtlisi:
@pstibrany I'd like to take a quick look

pkg/ruler/ruler.go
```diff
@@ -50,6 +50,7 @@ const (
 	ShardingStrategyDefault = "default"
 	ShardingStrategyShuffle = "shuffle-sharding"
 
+	// Number of concurrent group list and group loads operations.
 	loadRulesConcurrency = 10
```
jtlisi, line 54: Might make sense to move this variable to `pkg/ruler/storage.go` since that is where it is used.

gotjosh (approved):
LGTM! Great work Peter.

jtlisi (approved):
LGTM

pkg/ruler/ruler.go
pstibrany: It is also used in the ruler.go in sharded load.

pstibrany:
Thanks for reviews!

PR #3275, opened 2020-10-03
Add an ability to load an arbitrary list of Cortex modules

### LINKED ISSUE #3272: Add an ability to run mixed number of modules in one instance
We're running a flat small Cortex cluster in a single binary mode with a single tenant.

Currently we can either run 'all' target which is a predefined subset of modules or a single target.
Target 'all' does not include, for example, compactor which is needed and it can run in clustered mode using a hash ring.

Same goes, I guess, for alertmanager etc.

So we have to run it separately, although I think it's completely fine to run it inside the main instance I guess.
Maybe we can change **target** configuration parameter to a list and specify several modules to run? Or add another parameter e.g. **targets** to maintain backwards compatibility.

blind-oracle:
Signed-off-by: Igor Novgorodov <igor@novg.net>

**What this PR does**:
Adds an ability to load a comma-separated list of modules instead of a single module or 'all'.
This should be backwards-compatible with the current behaviour.

The moduleManager is extended a bit to keep track of already loaded modules.
I don't really like having **All** as a module, but I'll leave it for now as-is to keep changes compact.
And single-binary detection depends on this in a couple of places.
Ideally we should just alias the word **all** to a predefined list of modules and load them like they were specified on the command line or in config file.

Fixes #3272

**Checklist**
- [x] Tests updated
- [x] Documentation added


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master / unreleased
 
+* [FEATURE] Implement an ability to load multiple Cortex modules using a comma-separated list. #3272
+  - Should be backwards compatible with the current behavior.
+  - Target list like 'all,compactor' can be used to load the single-binary modules plus some other modules.
```
pracucci, line 7: I suggest a little rephrasing to simplify it:

```suggestion
* [ENHANCEMENT] Allow to specify multiple comma-separated Cortex services to `-target` CLI option (or its respective YAML config option). For example, `-target=all,compactor` can be used to start Cortex single-binary with compactor as well. #3272
```

We consider these improvements "enhancement" and not new core features. Please move the changelog entry to the right place (entries are grouped by type).

pkg/cortex/modules.go
```diff
@@ -173,7 +173,7 @@ func (t *Cortex) initDistributorService() (serv services.Service, err error) {
 	// Check whether the distributor can join the distributors ring, which is
 	// whenever it's not running as an internal dependency (ie. querier or
 	// ruler's dependency)
-	canJoinDistributorsRing := (t.Cfg.Target == All || t.Cfg.Target == Distributor)
+	canJoinDistributorsRing := t.Cfg.Modules[Distributor] || t.Cfg.Modules[All]
```
pracucci, line 176: You should do a similar change in `initAlertManager()` too.

pkg/cortex/cortex.go
```diff
@@ -71,10 +72,11 @@ import (
 
 // Config is the root config for Cortex.
 type Config struct {
-	Target      string `yaml:"target"`
-	AuthEnabled bool   `yaml:"auth_enabled"`
-	PrintConfig bool   `yaml:"-"`
-	HTTPPrefix  string `yaml:"http_prefix"`
+	Target      string          `yaml:"target"`
+	Modules     map[string]bool `yaml:"-"`
```
pracucci, line 76: I think keeping both `Target string`  and the new `Modules map[string]bool` may lead to bugs in the future. We may forget to check `Modules` instead of `Target`. A better option may be define a CLI flag custom type in `pkg/util/flagext` which is a `[]string` but implements `flag.Value` and `yaml.Marshaller/Unmarshaller` to marshal/unmarshal a comma separated string.

We wouldn't have a map, but checking if a module is within the `[]string` would be as easy as calling `util.StringsContain()`.

What do you think?

pkg/util/modules/modules.go
```diff
@@ -25,6 +25,12 @@ type module struct {
 // in the right order of dependencies.
 type Manager struct {
 	modules map[string]*module
+
+	// Modules that are already initialized
+	modulesLoaded map[string]bool
```
pracucci, line 30: I would rename "loaded" into "initialised", which is the naming we use here.

pkg/util/modules/modules.go
```diff
@@ -119,6 +132,11 @@ func (m *Manager) UserVisibleModuleNames() []string {
 	return result
 }
 
+// GetServicesMap returns services map
+func (m *Manager) GetServicesMap() map[string]services.Service {
```
pracucci, line 136: We may call it `GetInitialisedServices()` and update the comment accordingly.

pkg/util/modules/modules.go
```diff
@@ -67,41 +75,46 @@ func (m *Manager) AddDependency(name string, dependsOn ...string) error {
...
-
 	// initialize all of our dependencies first
-	deps := m.orderedDeps(target)
-	deps = append(deps, target) // lastly, initialize the requested module
+	deps := m.orderedDeps(name)
+	deps = append(deps, name) // lastly, initialize the requested module
 
 	for ix, n := range deps {
+		// Skip already loaded modules
+		if m.modulesLoaded[n] {
```
pracucci, line 89: Couldn't we just check `m.servicesMap`?

pracucci (commented):
Thanks for working on this! I left few comments. @pstibrany what do you think?

CHANGELOG.md
blind-oracle: Sure, forgot that they're grouped.

pkg/cortex/cortex.go
blind-oracle: There's already flagext.StringSlice, should I just use it?

pstibrany: `flagext.StringSlice` doesn't support comma-separated argument, only argument with multiple flags. I'm afraid that would be incompatible change in YAML config.

pkg/util/modules/modules.go
pstibrany: That wouldn't work... not all modules have a service.

pkg/util/modules/modules.go
```diff
@@ -67,41 +75,46 @@ func (m *Manager) AddDependency(name string, dependsOn ...string) error {
 // InitModuleServices initialises the target module by initialising all its dependencies
 // in the right order. Modules are wrapped in such a way that they start after their
 // dependencies have been started and stop before their dependencies are stopped.
-func (m *Manager) InitModuleServices(target string) (map[string]services.Service, error) {
-	if _, ok := m.modules[target]; !ok {
-		return nil, fmt.Errorf("unrecognised module name: %s", target)
+func (m *Manager) InitModuleServices(name string) error {
```
pstibrany, line 78: Could we make it work with this signature instead?

```suggestion
func (m *Manager) InitModuleServices(names ...string) (map[string]services.Service, error)
```

There are multiple projects using this system, and this is backwards compatible, while your suggestion isn't. Furthermore it keeps the state of initialization outside of this manager.

pkg/util/modules/modules.go
```diff
@@ -131,6 +149,12 @@ func (m *Manager) IsUserVisibleModule(mod string) bool {
 	return false
 }
 
+// IsModuleRegistered returns true if given module was registered
+func (m *Manager) IsModuleRegistered(mod string) bool {
```
pstibrany, line 153: Let's not introduce new public methods to manager, that are not needed externally.

pstibrany (commented):
Thank you for working on this! I suggest we keep the changes to `modules.Manager` backwards compatible and not introduce new public API to it (it's always easy to add new methods, and difficult to remove them once they get used), since there are several other projects using this other than Cortex.

pkg/util/modules/modules.go
blind-oracle: Sure, didn't know that anything else other than Cortex used it.

blind-oracle:
Thanks for the comments everyone.
I've tried to take all of them into account and improve accordingly:
* I've created a new type `flagext.StringSliceCSV` which implements serialisation to/from comma-separated string + tests.
I also had to add a workaround for this type to the doc-generator as it considers it a "list of strings".
* Reverted the API of module loader to be backwards-compatible without extra public methods
* Workaround Jaeger name generation as before it was suffixed by Target.
Now it will be suffixed only if there's only one module listed in Target, otherwise it will be just `cortex`. This should be backwards-compatible also.
* Fixed the CHANGELOG

pkg/util/modules/modules.go
```diff
@@ -25,6 +25,12 @@ type module struct {
 // in the right order of dependencies.
 type Manager struct {
 	modules map[string]*module
+
+	// Modules that are already initialized
+	modulesInitialized map[string]bool
```
pstibrany, line 30: Let’s not introduce more state to the Manager than necessary. In particular `modulesInitialized` and `servicesMap` don’t need to be fields of the manager.

Calling `InitModuleServices` twice should recreate everything from scratch.

pstibrany (commented):
Thank you for addressing my feedback! I have left more comments in the spirit of “minimize state”.

pkg/cortex/cortex.go
```diff
@@ -305,19 +317,27 @@ func (t *Cortex) setupThanosTracing() {
...
-		level.Warn(util.Logger).Log("msg", "selected target is an internal module, is this intended?", "target", t.Cfg.Target)
+func (t *Cortex) initModules() (err error) {
+	for _, module := range t.Cfg.Target {
+		if !t.ModuleManager.IsUserVisibleModule(module) {
+			level.Warn(util.Logger).Log("msg", "selected target is an internal module, is this intended?", "target", module)
+		}
 	}
 
-	serviceMap, err := t.ModuleManager.InitModuleServices(t.Cfg.Target)
+	t.ServiceMap, err = t.ModuleManager.InitModuleServices(t.Cfg.Target...)
```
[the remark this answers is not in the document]
blind-oracle: Makes sense, done

pkg/util/modules/modules.go
blind-oracle: Done

pkg/cortex/cortex.go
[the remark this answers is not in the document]
blind-oracle: I guess I didn't get the first time your idea, will try to get rid of it in `cortex` package.

pkg/util/modules/modules_test.go
```diff
@@ -39,13 +46,24 @@ func TestDependencies(t *testing.T) {
...
+	assert.Empty(t, svc)
+
+	svc, err = mm.InitModuleServices("serviceC")
 	assert.NoError(t, err)
+	assert.Equal(t, 3, len(svc))
 
-	svcs, err = mm.InitModuleServices("service_unknown")
-	assert.Nil(t, svcs)
-	assert.Error(t, err, fmt.Errorf("unrecognised module name: service_unknown"))
+	// Test loading of the module second time (should be noop)
```
pstibrany, line 63: It should not be noop -- it should in fact create new services. Can we extend the test to verify that services from first and second invocation of `mm.InitModuleServices("serviceC")` are different?

pstibrany (commented):
Thank you a lot for your patience. This is on the right track, few more minor nits.

pkg/util/modules/modules_test.go
blind-oracle: Yeah, the noop part is left from the previous stateful ModuleManager. Fixed

pkg/cortex/cortex.go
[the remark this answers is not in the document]
blind-oracle: Yep, for now made initModules() stateless and fixed tests.

blind-oracle:
Integration test failed, but looking at the logs it looks like some other transient issue.

pkg/cortex/cortex_test.go
```diff
@@ -68,24 +68,28 @@ func TestCortex(t *testing.T) {
 				},
 			},
 		},
-		Target: All,
 	}
 
+	cfg.Target = []string{All, Compactor}
```
pstibrany, line 73: We can put this back into the struct.

pstibrany (approved):
I have left some non-blocking nitpicky comments (mostly to minimize the diff), but LGTM. Thank you!

blind-oracle:
Done mentioned fixes

pstibrany (approved):
Thank you very much!

blind-oracle:
There's a conflict with newer PR https://github.com/cortexproject/cortex/pull/3287, I've merged current master and fixed.

blind-oracle:
Looks like tests are unstable.
I also occasionally get failed etcd-related tests when running `go test ./...` locally.

But now it's something else:
```
--- FAIL: TestShuffleShardWithCaching (5.03s)
    ring_test.go:790: lifecycler 0 failed: failed to pick tokens in the KV store, ring: test: failed to CAS test
    ring_test.go:834: 6 != 5
FAIL
```

pstibrany:
> Looks like tests are unstable.
> I also occasionally get failed etcd-related tests when running `go test ./...` locally.

Yes, this test is flaky. We need to fix it.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@
 * [CHANGE] Renamed `-redis.enable-tls` CLI flag to `-redis.tls-enabled`, and its respective YAML config option from `enable_tls` to `tls_enabled`. #3298
 * [FEATURE] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-tenant` globally, or using per-tenant limit `max_queriers_per_tenant`), each tenants's requests will be handled by different set of queriers. #3113 #3257
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
+* [ENHANCEMENT] Allow to specify multiple comma-separated Cortex services to `-target` CLI option (or its respective YAML config option). For example, `-target=all,compactor` can be used to start Cortex single-binary with compactor as well. #3272
```
pracucci, line 51:
```suggestion
* [ENHANCEMENT] Allow to specify multiple comma-separated Cortex services to `-target` CLI option (or its respective YAML config option). For example, `-target=all,compactor` can be used to start Cortex single-binary with compactor as well. #3275
```
```suggestion
* [ENHANCEMENT] Allow to specify multiple comma-separated Cortex services to `-target` CLI option (or its respective YAML config option). For example, `-target=all,compactor` can be used to start Cortex single-binary with compactor as well. #3272
```

pkg/cortex/cortex_test.go
```diff
@@ -68,15 +69,17 @@ func TestCortex(t *testing.T) {
...
 	}
 
 	c, err := New(cfg)
 	require.NoError(t, err)
 
-	serviceMap, err := c.ModuleManager.InitModuleServices(c.Cfg.Target)
+	serviceMap, err := c.ModuleManager.InitModuleServices(cfg.Target...)
 	require.NoError(t, err)
 	require.NotNil(t, serviceMap)
+	log.Println(serviceMap)
```
pracucci, line 82: I would remove this debug line.

```suggestion
```

pracucci (approved):
You did great! LGTM! Will merge as soon as the CI passes.

PR #3276, opened 2020-10-04
Fix for #3111 Read form values and request body before sending request to RoundTripper 

### LINKED ISSUE #3111: Query frontend slow log warning: invalid Read on closed Body
We are seeing a lot of warnings in the Thanos Query frontend slow log `http: invalid Read on closed Body`. Since it uses Cortex Query frontend as a library, I tested the Cortex one and it has the same problem.

The warning occurs when it logs a `POST` request. It doesn't affect anything but it is annoying.

```
level=warn ts=2020-08-31T15:01:31.642718692Z caller=frontend.go:198 org_id=fake traceID=f8ac3a930226002 msg="unable to parse form for request" err="http: invalid Read on closed Body"
level=info ts=2020-08-31T15:01:31.6427589Z caller=frontend.go:206 org_id=fake traceID=f8ac3a930226002 msg="slow query detected" method=POST host= path=/api/v1/query time_taken=1m3.598993508s
level=warn ts=2020-08-31T15:02:16.45874421Z caller=frontend.go:198 org_id=fake traceID=653a2ed25929e383 msg="unable to parse form for request" err="http: invalid Read on closed Body"
level=info ts=2020-08-31T15:02:16.458782837Z caller=frontend.go:206 org_id=fake traceID=653a2ed25929e383 msg="slow query detected" method=POST host= path=/api/v1/query time_taken=44.715073818s
level=warn ts=2020-08-31T15:02:28.766424824Z caller=frontend.go:198 org_id=fake traceID=2382a13b1e6ef89 msg="unable to parse form for request" err="http: invalid Read on closed Body"
level=info ts=2020-08-31T15:02:28.766461552Z caller=frontend.go:206 org_id=fake traceID=2382a13b1e6ef89 msg="slow query detected" method=POST host= path=/api/v1/query time_taken=12.206554463s
```

melkorm:
**What this PR does**:
Golang `http.RoundTripper` requires to close request body (https://golang.org/pkg/net/http/#RoundTripper), so accessing request body after running roundtrip can result in two scenarios:

a) silent - Body is NopReader and no form values is reported
b) warning - Body is ReadCloser and warning pops up about accessing closed body

This PR attempts to capture form values before running roundtrip which allows us to report form values and prevents us from breaking http (content length is X but body len is 0). 

Fixes #3111

**Checklist**
- [x] Tests updated 
- [ ] ~Documentation added~

PS. Let me know if I should put something in CHANGELOG regarding this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 * [CHANGE] Renamed `-<prefix>.redis.enable-tls` CLI flag to `-<prefix>.redis.tls-enabled`, and its respective YAML config option from `enable_tls` to `tls_enabled`. #3298
 * [CHANGE] Increased default `-<prefix>.redis.timeout` from `100ms` to `500ms`. #3301
 * [CHANGE] `cortex_alertmanager_config_invalid` has been removed in favor of `cortex_alertmanager_config_last_reload_successful`. #3289
+* [CHANGE] Query POST requests: body size is read up to 10MiB. #3276
 * [FEATURE] Shuffle sharding: added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-tenant` globally, or using per-tenant limit `max_queriers_per_tenant`), each tenants's requests will be handled by different set of queriers. #3113 #3257
 * [FEATURE] Shuffle sharding: added support for shuffle-sharding ingesters on the read path. When ingesters shuffle-sharding is enabled and `-querier.shuffle-sharding-ingesters-lookback-period` is set, queriers will fetch in-memory series from the minimum set of required ingesters, selecting only ingesters which may have received series since 'now - lookback period'. #3252
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
@@ -103,6 +104,7 @@
 * [BUGFIX] Fix common prefixes returned by List method of S3 client. #3358
 * [BUGFIX] Honor configured timeout in Azure and GCS object clients. #3285
 * [BUGFIX] Shuffle sharding: fixed max global series per user/metric limit when shuffle sharding and `-distributor.shard-by-all-labels=true` are both enabled in distributor. When using these global limits you should now set `-distributor.sharding-strategy` and `-distributor.zone-awareness-enabled` to ingesters too. #3369
+* [BUGFIX] Slow query logging: when using downstream server request parameters were not logged. #3276
 
 ## 1.4.0 / 2020-10-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@
 * [CHANGE] Renamed `-<prefix>.redis.enable-tls` CLI flag to `-<prefix>.redis.tls-enabled`, and its respective YAML config option from `enable_tls` to `tls_enabled`. #3298
 * [CHANGE] Increased default `-<prefix>.redis.timeout` from `100ms` to `500ms`. #3301
 * [CHANGE] `cortex_alertmanager_config_invalid` has been removed in favor of `cortex_alertmanager_config_last_reload_successful`. #3289
-* [CHANGE] Query POST requests: body size is read up to 10MiB. #3276
+* [CHANGE] Query-frontend: POST requests whose body size exceeds 10MiB will be rejected. The max body size can be customised via `-frontend.max-body-size`. #3276
 * [FEATURE] Shuffle sharding: added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-tenant` globally, or using per-tenant limit `max_queriers_per_tenant`), each tenants's requests will be handled by different set of queriers. #3113 #3257
 * [FEATURE] Shuffle sharding: added support for shuffle-sharding ingesters on the read path. When ingesters shuffle-sharding is enabled and `-querier.shuffle-sharding-ingesters-lookback-period` is set, queriers will fetch in-memory series from the minimum set of required ingesters, selecting only ingesters which may have received series since 'now - lookback period'. #3252
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
--- a/docs/configuration/prometheus-frontend.yml
+++ b/docs/configuration/prometheus-frontend.yml
@@ -56,5 +56,3 @@ frontend:
 
   # The Prometheus URL to which the query-frontend should connect to.
   downstream_url: http://prometheus.mydomain.com
-  # Max body size for downstream prometheus
-  downstream_max_body_size: 10485760
--- a/docs/configuration/prometheus-frontend.yml
+++ b/docs/configuration/prometheus-frontend.yml
@@ -56,3 +56,5 @@ frontend:
 
   # The Prometheus URL to which the query-frontend should connect to.
   downstream_url: http://prometheus.mydomain.com
+  # Max body size for downstream prometheus
+  downstream_max_body_size: 10485760
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -179,8 +179,6 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
-	startTime := time.Now()
-
 	// to parse form we need to be sure
 	// that roundtriper request gets not read reader
 	var buf bytes.Buffer
@@ -192,6 +190,7 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 	}
 	r.Body = ioutil.NopCloser(&buf)
 
+	startTime := time.Now()
 	resp, err := f.roundTripper.RoundTrip(r)
 	queryResponseTime := time.Since(startTime)
 
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -182,7 +182,7 @@ func (f *Frontend) Handler() http.Handler {
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
 
-	// buffer the body for later use in case of slow query
+	// Buffer the body for later use to track slow queries.
 	var buf bytes.Buffer
 	r.Body = http.MaxBytesReader(w, r.Body, int64(maxBodySize))
 	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -38,6 +38,7 @@ var (
 	errTooManyRequest   = httpgrpc.Errorf(http.StatusTooManyRequests, "too many outstanding requests")
 	errCanceled         = httpgrpc.Errorf(StatusClientClosedRequest, context.Canceled.Error())
 	errDeadlineExceeded = httpgrpc.Errorf(http.StatusGatewayTimeout, context.DeadlineExceeded.Error())
+	maxBodySize         = 10 * 1024 * 1024 // 10 MiB
 )
 
 // Config for a Frontend.
@@ -179,9 +180,11 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
-	// create buffer for later use in case of slow query
-	// to obtain request's form values
+	defer r.Body.Close()
+
+	// buffer the body for later use in case of slow query
 	var buf bytes.Buffer
+	r.Body = http.MaxBytesReader(w, r.Body, int64(maxBodySize))
 	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
 
 	startTime := time.Now()
@@ -190,46 +193,54 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 
 	if err != nil {
 		writeError(w, err)
-	} else {
-		hs := w.Header()
-		for h, vs := range resp.Header {
-			hs[h] = vs
-		}
-		w.WriteHeader(resp.StatusCode)
-		io.Copy(w, resp.Body)
+		return
+	}
+
+	hs := w.Header()
+	for h, vs := range resp.Header {
+		hs[h] = vs
+	}
+
+	w.WriteHeader(resp.StatusCode)
+	_, err = io.Copy(w, resp.Body)
+	if err != nil {
+		writeError(w, err)
+		return
 	}
 
 	f.reportSlowQuery(queryResponseTime, r, buf)
 }
 
-// reportSlowQuery reprots slow queries if LogQueriesLongerThan is set to <0, if it is set to 0 query logging
-// is disabled. This function shouldn't access request body as it will be already read by RoundTriper returning
-// no form values or reporting closed body warning
+// reportSlowQuery reprots slow queries if LogQueriesLongerThan is set to <0, where 0 disables logging
 func (f *Frontend) reportSlowQuery(queryResponseTime time.Duration, r *http.Request, bodyBuf bytes.Buffer) {
-	if f.cfg.LogQueriesLongerThan != 0 && queryResponseTime > f.cfg.LogQueriesLongerThan {
-		logMessage := []interface{}{
-			"msg", "slow query detected",
-			"method", r.Method,
-			"host", r.Host,
-			"path", r.URL.Path,
-			"time_taken", queryResponseTime.String(),
-		}
-
-		r.Body = ioutil.NopCloser(&bodyBuf)
-
-		// Ensure the form has been parsed so all the parameters are present
-		err := r.ParseForm()
-		if err != nil {
-			level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
-		}
-
-		// Attempt to iterate through the Form to log any filled in values
-		for k, v := range r.Form {
-			logMessage = append(logMessage, fmt.Sprintf("param_%s", k), strings.Join(v, ","))
-		}
-
-		level.Info(util.WithContext(r.Context(), f.log)).Log(logMessage...)
+	if f.cfg.LogQueriesLongerThan == 0 || queryResponseTime <= f.cfg.LogQueriesLongerThan {
+		return
 	}
+
+	logMessage := []interface{}{
+		"msg", "slow query detected",
+		"method", r.Method,
+		"host", r.Host,
+		"path", r.URL.Path,
+		"time_taken", queryResponseTime.String(),
+	}
+
+	// use previously buffered body
+	r.Body = ioutil.NopCloser(&bodyBuf)
+
+	// Ensure the form has been parsed so all the parameters are present
+	err := r.ParseForm()
+	if err != nil {
+		level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
+	}
+
+	// Attempt to iterate through the Form to log any filled in values
+	for k, v := range r.Form {
+		logMessage = append(logMessage, fmt.Sprintf("param_%s", k), strings.Join(v, ","))
+	}
+
+	level.Info(util.WithContext(r.Context(), f.log)).Log(logMessage...)
+
 }
 
 func writeError(w http.ResponseWriter, err error) {
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -47,7 +47,7 @@ type Config struct {
 	MaxOutstandingPerTenant int           `yaml:"max_outstanding_per_tenant"`
 	CompressResponses       bool          `yaml:"compress_responses"`
 	DownstreamURL           string        `yaml:"downstream_url"`
-	DownstreamMaxBodySize   int64         `yaml:"downstream_max_body_size"`
+	MaxBodySize             int64         `yaml:"max_body_size"`
 	LogQueriesLongerThan    time.Duration `yaml:"log_queries_longer_than"`
 }
 
@@ -56,7 +56,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxOutstandingPerTenant, "querier.max-outstanding-requests-per-tenant", 100, "Maximum number of outstanding requests per tenant per frontend; requests beyond this error with HTTP 429.")
 	f.BoolVar(&cfg.CompressResponses, "querier.compress-http-responses", false, "Compress HTTP responses.")
 	f.StringVar(&cfg.DownstreamURL, "frontend.downstream-url", "", "URL of downstream Prometheus.")
-	f.Int64Var(&cfg.DownstreamMaxBodySize, "frontend.downstream-max-body-size", defaultMaxBodySize, "Max body size for downstream prometheus.")
+	f.Int64Var(&cfg.MaxBodySize, "frontend.max-body-size", defaultMaxBodySize, "Max body size for downstream prometheus.")
 	f.DurationVar(&cfg.LogQueriesLongerThan, "frontend.log-queries-longer-than", 0, "Log queries that are slower than the specified duration. Set to 0 to disable. Set to < 0 to enable on all queries.")
 }
 
@@ -187,7 +187,7 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 
 	// Buffer the body for later use to track slow queries.
 	var buf bytes.Buffer
-	r.Body = http.MaxBytesReader(w, r.Body, f.cfg.DownstreamMaxBodySize)
+	r.Body = http.MaxBytesReader(w, r.Body, f.cfg.MaxBodySize)
 	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
 
 	startTime := time.Now()
--- a/pkg/querier/frontend/frontend.go
+++ b/pkg/querier/frontend/frontend.go
@@ -179,16 +179,10 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
-	// to parse form we need to be sure
-	// that roundtriper request gets not read reader
+	// create buffer for later use in case of slow query
+	// to obtain request's form values
 	var buf bytes.Buffer
 	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
-	// Ensure the form has been parsed so all the parameters are present
-	err := r.ParseForm()
-	if err != nil {
-		level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
-	}
-	r.Body = ioutil.NopCloser(&buf)
 
 	startTime := time.Now()
 	resp, err := f.roundTripper.RoundTrip(r)
@@ -205,13 +199,13 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 		io.Copy(w, resp.Body)
 	}
 
-	f.reportSlowQuery(queryResponseTime, r)
+	f.reportSlowQuery(queryResponseTime, r, buf)
 }
 
 // reportSlowQuery reprots slow queries if LogQueriesLongerThan is set to <0, if it is set to 0 query logging
 // is disabled. This function shouldn't access request body as it will be already read by RoundTriper returning
 // no form values or reporting closed body warning
-func (f *Frontend) reportSlowQuery(queryResponseTime time.Duration, r *http.Request) {
+func (f *Frontend) reportSlowQuery(queryResponseTime time.Duration, r *http.Request, bodyBuf bytes.Buffer) {
 	if f.cfg.LogQueriesLongerThan != 0 && queryResponseTime > f.cfg.LogQueriesLongerThan {
 		logMessage := []interface{}{
 			"msg", "slow query detected",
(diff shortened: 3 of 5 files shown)


pkg/querier/frontend/frontend.go
```diff
@@ -179,8 +179,19 @@ func (f *Frontend) Handler() http.Handler {
...
+	// that roundtriper request gets not read reader
+	var buf bytes.Buffer
+	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
+	// Ensure the form has been parsed so all the parameters are present
+	err := r.ParseForm()
+	if err != nil {
+		level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
+	}
+	r.Body = ioutil.NopCloser(&buf)
+
```
yeya24, line 194: What about moving `startTime := time.Now()` to this line?

melkorm: Good point 👍

pkg/querier/frontend/frontend_test.go
```diff
@@ -251,8 +255,75 @@ func TestFrontendCheckReady(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		client := http.Client{
+			Transport: &nethttp.Transport{},
+		}
+
+		resp, err := client.Do(req)
+		assert.NoError(t, err)
+		b, err := ioutil.ReadAll(resp.Body)
+		defer resp.Body.Close()
```
pracucci, line 309: I think there's no need to `defer` it. Since we've already `ioutil.ReadAll()` you should be able to directly close it.

pkg/querier/frontend/frontend_test.go
```diff
@@ -251,8 +255,75 @@ func TestFrontendCheckReady(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.Equal(t, 200, resp.StatusCode, string(b))
+
+		logs := buf.String()
```
pracucci, line 313: Should we also assert that doesn't contain `invalid Read on closed Body`?

pkg/querier/frontend/frontend_test.go
```diff
@@ -251,8 +255,75 @@ func TestFrontendCheckReady(t *testing.T) {
...
+
+	defer downstreamServer.Shutdown(context.Background()) //nolint:errcheck
+	go downstreamServer.Serve(downstreamListen)           //nolint:errcheck
+
+	// Configure the query-frontend with the mocked downstream server.
+	config := defaultFrontendConfig()
+	config.LogQueriesLongerThan = 1 * time.Microsecond
+	config.DownstreamURL = fmt.Sprintf("http://%s", downstreamListen.Addr())
+
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
pracucci, line 279: Do we need this? Given we're testing the case of the downstream URL, is this required at all? I found it confusing having both `w.Write([]byte(responseBody + responseBody + responseBody + responseBody + responseBody))` here and in the `downstreamServer`.

pkg/querier/frontend/frontend_test.go
```diff
@@ -251,8 +255,75 @@ func TestFrontendCheckReady(t *testing.T) {
...
-func testFrontend(t *testing.T, config Config, handler http.Handler, test func(addr string), matchMaxConcurrency bool) {
+func TestFrontend_LogsSlowQueriesFormValues_Issue3111(t *testing.T) {
+	// Create an HTTP server listening locally. This server mocks the downstream
+	// Prometheus API-compatible server.
+	downstreamListen, err := net.Listen("tcp", "localhost:0")
+	require.NoError(t, err)
+
+	downstreamServer := http.Server{
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, err := w.Write([]byte(responseBody + responseBody + responseBody + responseBody + responseBody))
```
pracucci, line 266: Could you comment on why are multiple `responseBody` chained together, please?

pracucci (commented):
Thanks for working on this fix! I confirm we wanna keep this log and your fixes makes sense to me. I left few comments / questions.

pkg/querier/frontend/frontend_test.go
melkorm: 👍  Changed by removing defer

pkg/querier/frontend/frontend_test.go
melkorm: Added assertion for this error message

pkg/querier/frontend/frontend_test.go
melkorm: Removed handler as it is no longer needed

pkg/querier/frontend/frontend_test.go
melkorm: Yes, at first I had to somehow trigger closing body and was playing with ResponseWriter which needs some data written to actually trigger request body close. 
Tested it now with short body and it also triggers this error with downstream server so it should be enough.

pkg/querier/frontend/frontend.go
```diff
@@ -179,6 +179,16 @@ func (f *Frontend) Handler() http.Handler {
...
+	// to parse form we need to be sure
+	// that roundtriper request gets not read reader
+	var buf bytes.Buffer
+	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
+	// Ensure the form has been parsed so all the parameters are present
+	err := r.ParseForm()
+	if err != nil {
+		level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
+	}
+	r.Body = ioutil.NopCloser(&buf)
```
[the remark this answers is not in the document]
melkorm: 🤔  That'a a good question. In case of parsing FormData I don't know what we should do as it should be (❓ ) rare, a specially when we have eliminated case of reading closed body. So, to not break current implementation I would leave it as is and perhaps wait if someone reports something odd with this behaviour (which current implementation uses but in different order). And then think about solution, but if you think we should do something different I am happy to implement it.

In case of
```
if err != nil {
		writeError(w, err)
	}
``` 
I think we should still report slow query even in case of request error, we could also improve it by adding log entry the same as for slow query to indicate actual roundtrip error with additional context.

pkg/querier/frontend/frontend.go
```diff
@@ -179,6 +179,16 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
+	// to parse form we need to be sure
+	// that roundtriper request gets not read reader
```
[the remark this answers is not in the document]
melkorm: Thank you ! Tried to put it in more accessible way but nothing could come to my mind and I agree that `request gets not read reader` sounds odd 🙈 
What do you think about: 
```
func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
	// we need to parse the request form from the body (if any) before delegating
	// the request to the roundtripper, ensuring the request body passed to the roundtripper
	// allows to re-read the full body
	var buf bytes.Buffer
```

pkg/querier/frontend/frontend.go
```diff
@@ -179,6 +179,16 @@ func (f *Frontend) Handler() http.Handler {
...
+	var buf bytes.Buffer
+	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
+	// Ensure the form has been parsed so all the parameters are present
+	err := r.ParseForm()
+	if err != nil {
+		level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
```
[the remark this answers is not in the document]
melkorm: Answered in https://github.com/cortexproject/cortex/pull/3276#discussion_r500426485

pkg/querier/frontend/frontend.go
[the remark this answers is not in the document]
pstibrany: What about:

> Buffer request body so that it can be read again by roundtripper.

pkg/querier/frontend/frontend.go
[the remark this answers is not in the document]
pstibrany: By continuing here, `buf` is in an unknown state... it can be completely empty (eg. for `POST` request with content type other than `application/x-www-form-urlencoded`, and wrong query string), or filled with broken / unfinished `x-www-form-urlencoded` data). In either case, it's different than it was before parsing the form.

pstibrany:
Instead of parsing the form at the beginning of `handle` method, I would suggest to leave parsing as-is, but buffer data from body:

1) read up to 1MB of data to buffer
2) set `r.Body` to `io.MultiReader(buffer, r.Body)`
3) call roundtripper
4) if request is slow, set `r.Body` to buffer again, call `r.ParseForm` and log as before.

By not parsing the form *before* calling roundtripper, we avoid some edge cases present in the current PR.

WDYT?

pkg/querier/frontend/frontend.go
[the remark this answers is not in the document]
pracucci: > So, to not break current implementation I would leave it as is and perhaps wait if someone reports something odd with this behaviour

I don't think we would break the current implementation if we return in case of error, because we would just skipping logging slow queries, which doesn't make much sense if the query failed. Requests logging is done in a middleware so would be done anyway.

melkorm:
> Instead of parsing the form at the beginning of `handle` method, I would suggest to leave parsing as-is, but buffer data from body:
> 
> 1. read up to 1MB of `r.Body` to buffer
> 2. set `r.Body` to `io.MultiReader(buffer, r.Body)`
> 3. call roundtripper
> 4. if request is slow, set `r.Body` to buffer again, call `r.ParseForm` and log as before.
> 
> By not parsing the form _before_ calling roundtripper, we avoid some edge cases present in the current PR.
> 
> WDYT?
> 
> (in point 4, we cannot reuse `bytes.Buffer` from step 1, since it was possibly already read)

Hmm I was thinking about this and couldn't see how it could solve our problem as slicing Reader into 2 separate Readers and joining them inside MultiReader would end up with 3 drained Readers as buf will be read same as r.Body which will result in empty MultiReader, this is example witch demonstrates this: https://play.golang.org/p/sJQtA2uknX4 
Perhaps I misunderstood something if you could provide example of what you meant it would be helpful 🙏 


but playing with your proposal I came to other idea that perhaps could make it less impactful to current implementation, instead of using MultiReader we could keep `TeeReader` assuming that RoundTrip will read the whole body and thus copy it to `&buf`, so if we hit slow query we could possibly reuse buffer assigning it to request and call ParseForm on it?

How it could look:
```
var buf bytes.Buffer
r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
...
// run Roudtrip which will read the body and fill our buffer trough TeeReader
resp, err := f.roundTripper.RoundTrip(r)
...
// in case of slow query we use buffer as our new body
// and call ParseForm to obtain parameters
r.Body = ioutil.NopCloser(&buf)
err := r.ParseForm()
```

I think it makes it best case scenario, as we don't read nor parse the body until slow query scenario. I've tested it and it works w/o issues.

Sorry for the late response but couldn't work on it for a while 😷

melkorm:
@pstibrany @pracucci I have pushed different approach to handling body discussed in https://github.com/cortexproject/cortex/pull/3276#issuecomment-706574867 so we use buffer and parse form only in case of slow query scenario.

Let me know what do you think :)

pstibrany:
> > (in point 4, we cannot reuse `bytes.Buffer` from step 1, since it was possibly already read)
> 
> Hmm I was thinking about this and couldn't see how it could solve our problem as slicing Reader into 2 separate Readers and joining them inside MultiReader would end up with 3 drained Readers as buf will be read same as r.Body which will result in empty MultiReader, this is example witch demonstrates this: https://play.golang.org/p/sJQtA2uknX4
> Perhaps I misunderstood something if you could provide example of what you meant it would be helpful 🙏

Yes, one needs to save the byte slice from `buf` and wrap it into a new `bytes.Buffer`.

> How it could look:
> 
> ```
> var buf bytes.Buffer
> r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))

Problem with this approach is that there is no limit in how much could be read into buf, so really nasty client can actually crash the query-frontend by sending gigabytes of data in the request (it can do that already today, but with this proposal, data would be stored in memory twice). My suggestion fixed that by limiting read to 1 MB only, but it didn't fix the problem that later in the roundtripper, we still read the full request into memory – we should limit that too.

melkorm:
> Problem with this approach is that there is no limit in how much could be read into buf, so really nasty client can actually crash the query-frontend by sending gigabytes of data in the request (it can do that already today, but with this proposal, data would be stored in memory twice). My suggestion fixed that by limiting read to 1 MB only, but it didn't fix the problem that later in the roundtripper, we still read the full request into memory – we should limit that too.

I think that request size should be handled at handlers level and used as advised here https://stackoverflow.com/a/28292505 - what I mean is that lower logic shouldn't enforce it as it would be needed to be copy pasted in few places where we access the body. Limiting it at the entry to the system would solve this for all cases and would make it easy to configure/change. I can create an issue for this to solve it in another PR.

> Yes, one needs to save the byte slice from `buf` and wrap it into a new `bytes.Buffer`.

If we assume that we put 1MB limit at the entry to the system (handlers) and assuming that Request Body is actually a stream (read data is gone after reading) then in my solution we don'y copy anything but with MultiReader we do because we have to have 1MB copy of Body:

 - `A`: bytes.Buffer for later use in slow query (it can't be read by anything thus can't be passed to RoundTripper), 
 - copy of `A` as first part of multi reader

Copying the byte slice would increase the logic and memory footprint comparing to the current solution imho.

In my second implementation we don't read body at all, we assume that RoundTripper will read the Body and it will be available in the buffer, that's my only concern tbh and if you look at the diff in introduces the least changes to current logic.

pstibrany:
> I think that request size should be handled at handlers level and used as advised here https://stackoverflow.com/a/28292505 - what I mean is that lower logic shouldn't enforce it as it would be needed to be copy pasted in few places where we access the body. Limiting it at the entry to the system would solve this for all cases and would make it easy to configure/change. I can create an issue for this to solve it in another PR.

Different handlers have different needs. If handler is just streaming the body somewhere, it doesn't need to be concerned with the length as much as handler that reads the full body into memory, which is what we do here.

> In my second implementation we don't read body at all, we assume that RoundTripper will read the Body and it will be available in the buffer, that's my only concern tbh and if you look at the diff in introduces the least changes to current logic.

My point was that there will be two copies of the body in memory using this approach. Once in `buf`, other copy created by `(*Frontend).RoundTrip` which calls `server.HTTPRequest` which reads the full body into buffer. But if we limit the body to some reasonable size, that's fine. (Not sure what reasonable size is here... but I'd start with 10MB and perhaps make it configurable).

pkg/querier/frontend/frontend.go
```diff
@@ -179,6 +179,10 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
+	// create buffer for later use in case of slow query
+	// to obtain request's form values
+	var buf bytes.Buffer
+	r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &buf))
```
pstibrany, line 185: Why `NopCloser` here?

pkg/querier/frontend/frontend.go
```diff
@@ -206,8 +215,10 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 			"time_taken", queryResponseTime.String(),
 		}
 
+		r.Body = ioutil.NopCloser(&bodyBuf)
```
pstibrany, line 218: Why `NopCloser` again?

pkg/querier/frontend/frontend.go
```diff
@@ -195,8 +199,13 @@ func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
 		io.Copy(w, resp.Body)
 	}
 
-	// If LogQueriesLongerThan is set to <0 we log every query, if it is set to 0 query logging
-	// is disabled
+	f.reportSlowQuery(queryResponseTime, r, buf)
```
pstibrany, line 202: Can we move the `if` condition outside, so that entire method isn't surrounded by it?

```suggestion
	if f.cfg.LogQueriesLongerThan != 0 && queryResponseTime > f.cfg.LogQueriesLongerThan {
		f.reportSlowQuery(queryResponseTime, r, buf)
	}
```

pkg/querier/frontend/frontend.go
pstibrany: Ah, sorry, here we need a `Closer`. My bad.

pkg/querier/frontend/frontend.go
pstibrany: Here we never close `r.Body`, so perhaps we should do that in `defer` statement at least.

melkorm: Good catch, added defer for `r.Body`

pkg/querier/frontend/frontend.go
melkorm: I've left the if statement in metod with return early approach, if you prefer having if outside of the method I can change it later.

pkg/querier/frontend/frontend.go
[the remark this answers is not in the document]
melkorm: Added return statements in case of errors, also added check for `_, err = io.Copy(w, resp.Body)`

pkg/querier/frontend/frontend.go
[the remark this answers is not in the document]
melkorm: This part of the code changed a bit, could you please see if it still makes sense ?

melkorm:
> My point was that there will be two copies of the body in memory using this approach. Once in `buf`, other copy created by `(*Frontend).RoundTrip` which calls `server.HTTPRequest` which reads the full body into buffer. But if we limit the body to some reasonable size, that's fine. (Not sure what reasonable size is here... but I'd start with 10MB and perhaps make it configurable).

Added limit for 10MiB

pkg/querier/frontend/frontend.go
```diff
@@ -38,6 +38,7 @@ var (
 	errTooManyRequest   = httpgrpc.Errorf(http.StatusTooManyRequests, "too many outstanding requests")
 	errCanceled         = httpgrpc.Errorf(StatusClientClosedRequest, context.Canceled.Error())
 	errDeadlineExceeded = httpgrpc.Errorf(http.StatusGatewayTimeout, context.DeadlineExceeded.Error())
+	maxBodySize         = 10 * 1024 * 1024 // 10 MiB
```
pstibrany, line 41: Should be const.

pkg/querier/frontend/frontend.go
```diff
@@ -179,46 +180,67 @@ func (f *Frontend) Handler() http.Handler {
...
-		err = r.ParseForm()
-		if err != nil {
-			level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
-		}
+	w.WriteHeader(resp.StatusCode)
+	_, err = io.Copy(w, resp.Body)
+	if err != nil {
+		writeError(w, err)
+		return
+	}
```
pstibrany, line 209: We shouldn't call `writeError` if copying body to response fails. `WriteError` cannot possibly succeed at this point, and may generate extra log messages since we try to send another status code. We can either log the message or silently ignore the error as before.

pstibrany (dismissed):
Thank you! LGTM, with minor nits.

pkg/querier/frontend/frontend.go
pracucci: Remember to mention this as a `[CHANGE]` in the changelog. We basically don't accept anymore any query POST request with a body larger than 10MB.

pkg/querier/frontend/frontend.go
```diff
@@ -179,46 +180,67 @@ func (f *Frontend) Handler() http.Handler {
 }
 
 func (f *Frontend) handle(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+
+	// buffer the body for later use in case of slow query
```
pracucci, line 185:
```suggestion
	// Buffer the body for later use to track slow queries.
```

pkg/querier/frontend/frontend.go
```diff
@@ -179,46 +180,67 @@ func (f *Frontend) Handler() http.Handler {
...
 
-		// Ensure the form has been parsed so all the parameters are present
-		err = r.ParseForm()
-		if err != nil {
-			level.Warn(util.WithContext(r.Context(), f.log)).Log("msg", "unable to parse form for request", "err", err)
-		}
+	w.WriteHeader(resp.StatusCode)
+	_, err = io.Copy(w, resp.Body)
+	if err != nil {
+		writeError(w, err)
```
pracucci: Will this `writeError()` work, considering we've already `w.WriteHeader()` and probably partially copied the `resp.Body` to `w`? I've the feeling this will fail and we should log instead of `writeError()`. Thoughts? /cc @pstibrany

pkg/querier/frontend/frontend.go
```diff
@@ -179,46 +180,67 @@ func (f *Frontend) Handler() http.Handler {
...
+	logMessage := []interface{}{
+		"msg", "slow query detected",
+		"method", r.Method,
+		"host", r.Host,
+		"path", r.URL.Path,
+		"time_taken", queryResponseTime.String(),
+	}
+
+	// use previously buffered body
+	r.Body = ioutil.NopCloser(&bodyBuf)
```
pracucci, line 226: Should we directly work on `r` or a `r.Clone()` of it? May be more future proof working on a clone. /cc @pstibrany

pkg/querier/frontend/frontend.go
```diff
@@ -32,6 +32,7 @@ import (
 const (
 	// StatusClientClosedRequest is the status code for when a client request cancellation of an http request
 	StatusClientClosedRequest = 499
+	maxBodySize               = 10 * 1024 * 1024 // 10 MiB
```
pracucci, line 35: Remember to also mention this as `[CHANGE]` in the `CHANGELOG` because after this PR gets merged we'll not allow anymore any query POST request with a body larger than 10MB.

pracucci (commented):
You did a great job. This deserves a `[BUGFIX]` entry in the `CHANGELOG` 😉

pkg/querier/frontend/frontend.go
melkorm: Yes, removed error handling and added comment saying why we don't handle it

pkg/querier/frontend/frontend.go
melkorm: See https://github.com/cortexproject/cortex/pull/3276#discussion_r508336476

pkg/querier/frontend/frontend.go
melkorm: Added, let me know if you want it phrased differently

pkg/querier/frontend/frontend.go
jtlisi, line 35: **thought(non-blocking):** This directly relates to an issue that Prometheus/Cortex users currently experience when they have a datasource configured to use a `GET` method. There is a `1 MiB` limit configured on the MaxHeaderSize in the Go server and queries with a large number of regex matchers will hit this limit causing queries to fail. The workaround is to switch the datasource to use a `POST` method. However, after this change, if a user hits this limit they will not have a readily available workaround. I personally think 10 MiB is reasonable as a max.

**suggestion:** I think a comment here and/or in the changelog denoting that this potentially could cause failures for large query strings with a very large number of regex matchers.

pkg/querier/frontend/frontend_test.go
```diff
@@ -251,8 +255,72 @@ func TestFrontendCheckReady(t *testing.T) {
 	}
 }
 
-func testFrontend(t *testing.T, config Config, handler http.Handler, test func(addr string), matchMaxConcurrency bool) {
+func TestFrontend_LogsSlowQueriesFormValues_Issue3111(t *testing.T) {
```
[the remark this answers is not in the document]
melkorm: Wouldn't call it strange as some projects use this convention, anyway removed the issue part

pkg/querier/frontend/frontend_test.go
```diff
@@ -251,8 +255,72 @@ func TestFrontendCheckReady(t *testing.T) {
...
+
+	var buf bytes.Buffer
+	l := log.NewLogfmtLogger(log.NewSyncWriter(&buf))
+
+	test := func(addr string) {
+		data := url.Values{}
+		data.Set("test", "form")
+		data.Set("issue", "3111")
+
+		req, _ := http.NewRequest(http.MethodPost, fmt.Sprintf("http://%s/", addr), strings.NewReader(data.Encode()))
```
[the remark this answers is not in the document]
melkorm: It won't work for `GET` method as https://pkg.go.dev/net/http#Request.ParseForm says:
```
// For POST, PUT, and PATCH requests, it also reads the request body, parses it
// as a form and puts the results into both r.PostForm and r.Form. Request body
// parameters take precedence over URL query string values in r.Form.
```

I've added assertion for parameters in request URL and could add similar test for `GET` method but I don't know how much of value it will have.

pkg/querier/frontend/frontend.go
melkorm: That made me think that this change is really dangerous as it will fail silently and potential users will be puzzled why some things doesn't work or give unexpected results. 

I think we could:

- add warning if Content-Length header says it's larger than 10MiB - without reading the body before the RoundTrip
- remove the limit for now
- leave it as is with 10MiB limit (no idea how many people run into request's body larger than 10 MiB 🤔 )

pstibrany: > That made me think that this change is really dangerous as it will fail silently and potential users will be puzzled why some things doesn't work or give unexpected results.

It doesn't fail silently, but with "http: request body too large" error, however we would send it back with status code 500, which isn't so good, since it's a client error and we should use 413 instead.

Unfortunately relevant error isn't public, so we could either 1) check for the error string from stdlib, which would be an ugly hack but it will work and with a good test we will catch change of the string easily, or 2) rework this completely, but nice thing about using `http.MaxBytesReader` is that it will also close the connection and not try to read the rest of the request. Personally I would be fine with approach 1) in `writeError` method (incl. the test for error reported by `http.MaxBytesReader`).

> I think we could:
> 
> * add warning if Content-Length header says it's larger than 10MiB - without reading the body before the RoundTrip

That's not enough, chunked requests have no content-length.

> * remove the limit for now

Query-frontend has no limit right now, and sending large enough request *does* crash it (easy to check using curl).

> * leave it as is with 10MiB limit (no idea how many people run into request's body larger than 10 MiB 🤔 )

I did a check on our clusters, and we don't receive query requests larger than 1 MiB. I suggested to use config option just in case, but I haven't insisted on using it since it never happens.

melkorm: > It doesn't fail silently, but with "http: request body too large" error, however we would send it back with status code 500, which isn't so good, since it's a client error and we should use 413 instead.

Yes, you are absolutely right, had LimitReader docs in head while writing this 😆 

I will add `"http: request body too large"` error handling in `writeError` and write a test for this.

melkorm: I have added new option to control body size and also handling for `http: request body too large`.

Currently the error message will look like this: `http: request body too large` but the original message is `readfrom tcp 127.0.0.1:54370->127.0.0.1:54365: http: request body too large`, my question is do we leave it as is or we create error inline with original message.

pstibrany (approved):
Thank you!

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@
 * [CHANGE] Renamed `-<prefix>.redis.enable-tls` CLI flag to `-<prefix>.redis.tls-enabled`, and its respective YAML config option from `enable_tls` to `tls_enabled`. #3298
 * [CHANGE] Increased default `-<prefix>.redis.timeout` from `100ms` to `500ms`. #3301
 * [CHANGE] `cortex_alertmanager_config_invalid` has been removed in favor of `cortex_alertmanager_config_last_reload_successful`. #3289
+* [CHANGE] Query POST requests: body size is read up to 10MiB. #3276
```
pracucci, line 51:
```suggestion
* [CHANGE] Query-frontend: POST requests whose body size exceeds 10MiB will be rejected. The max body size can be customised via `-frontend.downstream-max-body-size`. #3276
```

docs/configuration/prometheus-frontend.yml
```diff
@@ -56,3 +56,5 @@ frontend:
 
   # The Prometheus URL to which the query-frontend should connect to.
   downstream_url: http://prometheus.mydomain.com
+  # Max body size for downstream prometheus
+  downstream_max_body_size: 10485760
```
pracucci, line 60: I would remove this. Let's keep our examples as simples as possible.

pkg/querier/frontend/frontend.go
```diff
@@ -32,19 +32,22 @@ import (
...
+	errDeadlineExceeded      = httpgrpc.Errorf(http.StatusGatewayTimeout, context.DeadlineExceeded.Error())
+	errRequestEntityTooLarge = httpgrpc.Errorf(http.StatusRequestEntityTooLarge, "http: request body too large")
 )
 
 // Config for a Frontend.
 type Config struct {
 	MaxOutstandingPerTenant int           `yaml:"max_outstanding_per_tenant"`
 	CompressResponses       bool          `yaml:"compress_responses"`
 	DownstreamURL           string        `yaml:"downstream_url"`
+	DownstreamMaxBodySize   int64         `yaml:"downstream_max_body_size"`
```
pracucci, line 50: In the query-frontend config, the naming "downstream" is used for the connection between query-frontend and the backend Prometheus server (or Prometheus-compatible). The max body size configurable here is not the downstream one, but it's the max input body size. For this reason I would just call it `max_body_size` (same for CLI flag, fixing flag description too).

CHANGELOG.md
melkorm: Thanks, changed 👍

docs/configuration/prometheus-frontend.yml
melkorm: Done

pkg/querier/frontend/frontend.go
melkorm: Changed to `MaxBodySize`

pracucci (approved):
Thanks @melkorm for patiently address our feedback. Much appreciated 🙏 LGTM!

melkorm:
Thanks guys 🙏  Was my biggest contribution to date and I thought it's going to be a quick fix 😆 

Learned a lot about cortex and Go tho 🤟

PR #3278, opened 2020-10-04
Add injectable middlewares to bucket client

jtlisi:
**What this PR does**:

- This adds a config field for injecting middlewares when creating a bucket client to allow for the clients created within modules to be created with these middlewares.
- This approach is overtly simple and a factory approach similar to the Ruler may be preferable.

### DIFF
--- a/pkg/storage/tsdb/bucket_client.go
+++ b/pkg/storage/tsdb/bucket_client.go
@@ -32,6 +32,11 @@ func NewBucketClient(ctx context.Context, cfg BucketConfig, name string, logger 
 		return nil, err
 	}
 
+	// Wrap the client with any provided middleware
+	for _, wrap := range cfg.Middleware {
+		client = wrap(client)
+	}
+
 	return objstore.NewTracingBucket(bucketWithMetrics(client, name, reg)), nil
 }
 
--- a/pkg/storage/tsdb/bucket_client.go
+++ b/pkg/storage/tsdb/bucket_client.go
@@ -33,7 +33,7 @@ func NewBucketClient(ctx context.Context, cfg BucketConfig, name string, logger 
 	}
 
 	// Wrap the client with any provided middleware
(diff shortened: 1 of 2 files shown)


pkg/storage/tsdb/bucket_client.go
```diff
@@ -32,6 +32,11 @@ func NewBucketClient(ctx context.Context, cfg BucketConfig, name string, logger
 		return nil, err
 	}
 
+	// Wrap the client with any provided middleware
+	for _, wrap := range cfg.Middlewares {
+		client = wrap(client)
+	}
+
```
pstibrany, line 39: I would put this *after* wrapping into tracing bucket, so that spans only cover "real" bucket usage, not wrappers.

pkg/storage/tsdb/config.go
```diff
@@ -64,6 +65,10 @@ type BucketConfig struct {
 	GCS        gcs.Config        `yaml:"gcs"`
 	Azure      azure.Config      `yaml:"azure"`
 	Filesystem filesystem.Config `yaml:"filesystem"`
+
+	// Not used internally, meant to allow callers to wrap Buckets
+	// created using this config
+	Middlewares []func(objstore.Bucket) objstore.Bucket `yaml:"-"`
```
pstibrany, line 71: Might be good to have an option to return error too.

pstibrany (approved):
LGTM, with nits.

pracucci (approved):
It's simple but if it works for your use case it LGTM.

pkg/storage/tsdb/config.go
jtlisi: 👍

pkg/storage/tsdb/bucket_client.go
jtlisi: 👍

PR #3279, opened 2020-10-05
Make it possible to extend Makefile with custom settings.

pstibrany:
**What this PR does**: This PR modifies `Makefile` to include `Makefile.local` if it exists, so that it is possible to extend it with custom settings and rules without committing those in shared repository. Example `Makefile.local.example` is included for inspiration.

If `Makefile.local` is missing, nothing changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ blocksconvert
 cortex
 query-tee
 test-exporter
+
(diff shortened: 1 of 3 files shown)


jtlisi (approved):
LGTM

pracucci (approved):
LGTM, thanks!

PR #3283, opened 2020-10-06
Blocksconvert cleaner

pstibrany:
**What this PR does**: This PR adds `cleaner` tool to `blocksconvert`. It works similarly to `builder`, but instead of building blocks, it actually REMOVES CHUNKS and INDEX ENTRIES (!!!).

Similarly to builder, `cleaner` also asks scheduler for next plan to work on, downloads it, and then starts removing the data. It's best to use single cleaner at once, otherwise there may be chunks left in the storage and index client. Reason for this is that chunks can span multiple days, and chunk is fully deleted only when processing plan (day) when chunk started. At the same time, cleaner needs to download some chunks, because labels are not stored in the plans, only chunks, and to be able to clean up all index entries, cleaner needs to know the labels. When using multiple cleaners, it can happen that cleaner processing older plans will delete chunks required to properly clean up data in newer plans. When using single cleaner only, this is not a problem, since scheduler will send plans to cleaner in time-reveresed order.

Since builder and cleaner share parts of code to handle fetching and iterating the plan, this shared code has been refactored into `plan_processor_service.go`. This affects some metrics names.

It has been tested manually on dev data. Test consisted of cleaning data for specific user, and rescanning the tables. Scan produced all expected plans for other users, and no plans for user with cleaned data.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


tools/blocksconvert/planprocessor/service.go
```diff
@@ -0,0 +1,405 @@
...
+	SchedulerEndpoint string
+	NextPlanInterval  time.Duration
+	GrpcConfig        grpcclient.ConfigWithTLS
+
+	// Additional settings required for plan processor service.
+
+	PlansDirectory string                                                                            // Where to store plan files.
+	Bucket         objstore.Bucket                                                                   // Bucket client used for downloading plan files.
+	Cleanup        func(logger log.Logger) error                                                     // Cleanup function called on startup and after each build. Can be nil.
+	Factory        func(planLog log.Logger, userID string, dayStart, dayEnd time.Time) PlanProcessor // Factory for creating PlanProcessor. Called for each new plan.
```
pracucci, line 47: Design wise, I would have expected these to be arguments to `NewService()`. Thoughts?

pracucci (approved):
Good job! Large part of code changes are due to the refactoring, which looks good to me (just left a non-blocking comment). The cleaner makes sense to me, even if I didn't double checked all nuances of chunks storage.

Side note: would be great to see a doc update for this (separate PR is fine)!

PR #3285, opened 2020-10-06
fix timeout handling in GetObject for azure and gcs

sandeepsukhani:
**What this PR does**:
When a timeout is configured for gcs and azure object clients the `GetObject` function(which returns a reader) cancels the context in defer which means we return a reader with a cancelled context. This is causing the download operation to fail.

I think it was a mistake done by me when I refactored the code to split `GetChunk` method to have a `GetObject` method in object store clients which otherwise was downloading the chunks within the functions and not returning an object reader.

I have changed the code to not cancel the context and just set the timeout in context.

Fixes #<issue number>

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,7 +99,7 @@
 * [BUGFIX] Ingester: fail to start an ingester running the blocks storage, if unable to load any existing TSDB at startup. #3354
 * [BUGFIX] Blocks storage: Avoid deletion of blocks in the ingester which are not shipped to the storage yet. #3346
 * [BUGFIX] Fix common prefixes returned by List method of S3 client. #3358
-* [BUGFIX] fix timeout handling in `GetObject` method of azure and gcs object clients. #3285
+* [BUGFIX] Honor configured timeout in Azure and GCS object clients. #3285
 
 ## 1.4.0 / 2020-10-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@
 * [BUGFIX] Ingester: fail to start an ingester running the blocks storage, if unable to load any existing TSDB at startup. #3354
 * [BUGFIX] Blocks storage: Avoid deletion of blocks in the ingester which are not shipped to the storage yet. #3346
 * [BUGFIX] Fix common prefixes returned by List method of S3 client. #3358
+* [BUGFIX] fix timeout handling in `GetObject` method of azure and gcs object clients. #3285
 
 ## 1.4.0 / 2020-10-02
 
--- a/pkg/alertmanager/alerts/objectclient/store.go
+++ b/pkg/alertmanager/alerts/objectclient/store.go
@@ -60,7 +60,7 @@ func (a *AlertStore) getAlertConfig(ctx context.Context, key string) (alerts.Ale
 		return alerts.AlertConfigDesc{}, err
 	}
 
-	defer runutil.CloseWithLogOnErr(util.Logger, readCloser, "close object reader")
+	defer runutil.CloseWithLogOnErr(util.Logger, readCloser, "close alert config reader")
 
 	buf, err := ioutil.ReadAll(readCloser)
 	if err != nil {
--- a/pkg/alertmanager/alerts/objectclient/store.go
+++ b/pkg/alertmanager/alerts/objectclient/store.go
@@ -6,8 +6,11 @@ import (
 	"io/ioutil"
 	"path"
 
+	"github.com/thanos-io/thanos/pkg/runutil"
+
 	"github.com/cortexproject/cortex/pkg/alertmanager/alerts"
 	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/util"
 )
 
 // Object Alert Storage Schema
@@ -52,12 +55,14 @@ func (a *AlertStore) ListAlertConfigs(ctx context.Context) (map[string]alerts.Al
 }
 
 func (a *AlertStore) getAlertConfig(ctx context.Context, key string) (alerts.AlertConfigDesc, error) {
-	reader, err := a.client.GetObject(ctx, key)
+	readCloser, err := a.client.GetObject(ctx, key)
 	if err != nil {
 		return alerts.AlertConfigDesc{}, err
 	}
 
-	buf, err := ioutil.ReadAll(reader)
+	defer runutil.CloseWithLogOnErr(util.Logger, readCloser, "close object reader")
+
+	buf, err := ioutil.ReadAll(readCloser)
 	if err != nil {
 		return alerts.AlertConfigDesc{}, err
 	}
--- a/pkg/chunk/azure/blob_storage_client.go
+++ b/pkg/chunk/azure/blob_storage_client.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Azure/azure-storage-blob-go/azblob"
 
 	"github.com/cortexproject/cortex/pkg/chunk"
+	chunk_util "github.com/cortexproject/cortex/pkg/chunk/util"
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 )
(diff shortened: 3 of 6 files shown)


pkg/chunk/azure/blob_storage_client.go
```diff
@@ -111,10 +111,7 @@ func (b *BlobStorage) Stop() {}
 
 func (b *BlobStorage) GetObject(ctx context.Context, objectKey string) (io.ReadCloser, error) {
 	if b.cfg.RequestTimeout > 0 {
-		// The context will be cancelled with the timeout or when the parent context is cancelled, whichever occurs first.
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, b.cfg.RequestTimeout)
-		defer cancel()
+		ctx, _ = context.WithTimeout(ctx, b.cfg.RequestTimeout)
```
pracucci, line 114: Doesn't introduce a leak not calling the `cancel` function at all?

pkg/chunk/azure/blob_storage_client.go
```diff
@@ -110,13 +111,26 @@ func NewBlobStorage(cfg *BlobStorageConfig) (*BlobStorage, error) {
...
+	var cancel context.CancelFunc
 	if b.cfg.RequestTimeout > 0 {
-		// The context will be cancelled with the timeout or when the parent context is cancelled, whichever occurs first.
-		var cancel context.CancelFunc
 		ctx, cancel = context.WithTimeout(ctx, b.cfg.RequestTimeout)
-		defer cancel()
 	}
 
+	rc, err := b.getObject(ctx, objectKey)
+	if cancel != nil {
```
pstibrany, line 120: if `cancel` is nil, `err` is ignored

pkg/chunk/util/util.go
```diff
@@ -125,3 +127,21 @@ func EnsureDirectory(dir string) error {
...
+
+func NewReadCloserWithContextCancelFunc(readCloser io.ReadCloser, cancel context.CancelFunc) io.ReadCloser {
+	return ReadCloserWithContextCancelFunc{
+		ReadCloser: readCloser,
+		cancel:     cancel,
+	}
+}
+
+func (r ReadCloserWithContextCancelFunc) Close() error {
+	defer r.cancel()
```
pstibrany, line 145:
```suggestion
	r.cancel()
```

CHANGELOG.md
```diff
@@ -72,6 +72,7 @@
 * [BUGFIX] Ruler Experimental API: Don't allow rule groups without names or empty rule groups. #3210
 * [BUGFIX] Experimental Alertmanager API: Do not allow empty Alertmanager configurations or bad template filenames to be submitted through the configuration API. #3185
 * [BUGFIX] When using ruler sharding, moving all user rule groups from ruler to a different one and then back could end up with some user groups not being evaluated at all. #3235
+* [BUGFIX] fix timeout handling in `GetObject` method of azure and gcs object clients. #3285
```
pstibrany, line 75:
```suggestion
* [BUGFIX] Honor configured timeout in Azure and GCS object clients. #3285
```

pkg/chunk/util/util.go
pracucci: I thought it was more correct Sandeep's approach to use `defer`. Isn't safer to call the cancel function as the very last thing?

pkg/chunk/azure/blob_storage_client.go
pracucci: An easier way to implement it may be assigning a no-op static function to `cancel` if timeout is not not set, so that you don't have to worry whether `cancel` is defined or not (and performance wise there's no real penalty doing so).

pkg/chunk/azure/blob_storage_client.go
```diff
@@ -110,13 +111,26 @@ func NewBlobStorage(cfg *BlobStorageConfig) (*BlobStorage, error) {
...
 func (b *BlobStorage) GetObject(ctx context.Context, objectKey string) (io.ReadCloser, error) {
+	var cancel context.CancelFunc
 	if b.cfg.RequestTimeout > 0 {
-		// The context will be cancelled with the timeout or when the parent context is cancelled, whichever occurs first.
-		var cancel context.CancelFunc
 		ctx, cancel = context.WithTimeout(ctx, b.cfg.RequestTimeout)
-		defer cancel()
 	}
 
+	rc, err := b.getObject(ctx, objectKey)
```
pracucci, line 119: But rethinking about it: is even correct to apply the timeout to the download time (time it takes to download the whole response content)? We had a similar discussion in Thanos and we ended up it's very hard to set such timeouts (while we should circuit break on download speed below X bytes/sec).

pracucci (commented):
I left a couple of comments. Thanks!

pkg/chunk/util/util.go
pstibrany: If `return r.ReadCloser.Close()` reacts negatively on closed context, then yes... better to keep `defer. In practice, I don't expect this to make a difference, but just in case.

pkg/chunk/azure/blob_storage_client.go
pstibrany: I also find it strange to have timeout applied at this level, but it's better than having it applied at multiple call sites.

> while we should circuit break on download speed below X bytes/sec

You could apply the same logic at many places, but generally we don't, and try anyway. Returning "timeout" before time really ran out would be quite interesting.

sandeepsukhani: > while we should circuit break on download speed below X bytes/sec
It would be harder to do and would be complex for users to understand I feel. I am also not sure how many users would use this vs setting a timeout for queries which would anyways cancel the context.

pkg/chunk/azure/blob_storage_client.go
pracucci: @sandeepsukhani Have you got a chance to look at this comment?

sandeepsukhani: Sorry, I was held up with some other work. Will take care of this tomorrow.

CHANGELOG.md
sandeepsukhani: I have pushed this change. Thanks for the review!

pkg/chunk/azure/blob_storage_client.go
sandeepsukhani: I have pushed this change. Thanks for the review!

CHANGELOG.md
pstibrany: There are now two changelog entries.

pracucci: Could you remove this line, given you added a new one below, please?

pstibrany (approved):
LGTM

CHANGELOG.md
sandeepsukhani: Sorry, I guess I failed to rebase it properly. Will fix it. Thanks!

pkg/chunk/util/util.go
```diff
@@ -125,3 +127,21 @@ func EnsureDirectory(dir string) error {
...
 	return err
 }
+
+// ReadCloserWithContextCancelFunc helps with cancelling the context when closing a ReadCloser.
+type ReadCloserWithContextCancelFunc struct {
+	io.ReadCloser
+	cancel context.CancelFunc
+}
+
+func NewReadCloserWithContextCancelFunc(readCloser io.ReadCloser, cancel context.CancelFunc) io.ReadCloser {
```
pracucci, line 137: Should we also override the `Read()` function and cancel the context once `EOF` is reached? I'm not sure about the `ReadCloser` contract (if calling `Close()` is mandatory or not) and I can't find anything to clarify that in the golang doc. If a caller doesn't call `Close()` once the `EOF` is reached, we'll leak the memory. On the contrary, if we cancel once `EOF` (or maybe even error) is returned by `Close()` and we set `r.cancel = nil` (adding a check in `r.Close()` too), we'll protect from this case.

Thoughts?

pstibrany: I think we should specify in `GetObject` contract that returned `ReadCloser` must be closed in order to avoid leaking resources. That's usually implied, but comment doesn't hurt. I don't think we need `Read` react on `EOF` in such case.

pracucci: Ok documenting it, but we should also check callers in Cortex to make sure they do it, otherwise we introduce an issue.

pkg/chunk/util/util.go
```diff
@@ -125,3 +127,22 @@ func EnsureDirectory(dir string) error {
 	}
 	return err
 }
+
+// ReadCloserWithContextCancelFunc helps with cancelling the context when closing a ReadCloser.
+// Note: The consumer of ReadCloserWithContextCancelFunc should always call the Close method when it is done reading which otherwise could cause a resource leak.
```
pstibrany, line 132: This comment also belongs on the `GetObject` method of the `ObjectClient` interface, without referencing `ReadCloserWithContextCancelFunc`. Users of `ObjectClient` don't need to know internal implementation details.

pkg/alertmanager/alerts/objectclient/store.go
```diff
@@ -52,12 +55,14 @@ func (a *AlertStore) ListAlertConfigs(ctx context.Context) (map[string]alerts.Al
...
 
 func (a *AlertStore) getAlertConfig(ctx context.Context, key string) (alerts.AlertConfigDesc, error) {
-	reader, err := a.client.GetObject(ctx, key)
+	readCloser, err := a.client.GetObject(ctx, key)
 	if err != nil {
 		return alerts.AlertConfigDesc{}, err
 	}
 
-	buf, err := ioutil.ReadAll(reader)
+	defer runutil.CloseWithLogOnErr(util.Logger, readCloser, "close object reader")
```
pracucci, line 63:
```suggestion
	defer runutil.CloseWithLogOnErr(util.Logger, readCloser, "close alert config reader")
```

pracucci (approved):
LGTM, thanks!

pstibrany (approved):

PR #3287, opened 2020-10-06
Error if the store gateway is run on an invalid engine

roidelapluie:
This pull request returns an explicit error if the user asks to start
the store gateway but omits to set the storage type to `blocks`.

The previous behaviour was to silently ignore the -target=store-gateway
argument.

Note that this does not apply when target is `all`.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci (approved):
LGTM

pstibrany (approved):

PR #3288, opened 2020-10-06
feat: separate ruler into ruler and rulerAPI components

jtlisi:
**What this PR does**:

This PR decouples the ruler and rule store API into separate components and allow for their configuration and usage to be decoupled. In a later PR if the ruler is fully decoupled from the API, it will be possible to run the rules API as a standalone service which will make it easier to deal with the challenges of using an eventually consistent object storage backend.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/ruler/api_test.go
+++ b/pkg/ruler/api_test.go
@@ -27,10 +27,7 @@ func TestRuler_rules(t *testing.T) {
 	defer rcleanup()
 	defer services.StopAndAwaitTerminated(context.Background(), r) //nolint:errcheck
 
-	a := API{
-		ruler: r,
-		store: r.store,
-	}
+	a := NewAPI(r, r.store)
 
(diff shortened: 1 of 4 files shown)


pkg/ruler/api_test.go
```diff
@@ -27,9 +27,14 @@ func TestRuler_rules(t *testing.T) {
 	defer rcleanup()
 	defer services.StopAndAwaitTerminated(context.Background(), r) //nolint:errcheck
 
+	a := API{
```
pracucci, line 30: Could you use `NewAPI()` in tests too?

pracucci (approved):
LGTM with a nit! But is this a draft? :)

pkg/ruler/api_test.go
jtlisi: 👍 will do

PR #3292, opened 2020-10-07
Don't set selected attributes when using delimiter.

pstibrany:
**What this PR does**: This PR modifies GCS object client to avoid asking for selected attributes when using `List` with delimiter. Problem is that using both delimiter and selected attributes returns empty result, because `cloud.google.com/go/storage` package doesn't put `prefixes` into list of returned fields.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/gcp/gcs_object_client.go
+++ b/pkg/chunk/gcp/gcs_object_client.go
@@ -110,9 +110,16 @@ func (s *GCSObjectClient) List(ctx context.Context, prefix, delimiter string) ([
 	var storageObjects []chunk.StorageObject
 	var commonPrefixes []chunk.StorageCommonPrefix
 	q := &storage.Query{Prefix: prefix, Delimiter: delimiter}
-	err := q.SetAttrSelection([]string{"Name", "Updated"})
-	if err != nil {
-		return nil, nil, err
+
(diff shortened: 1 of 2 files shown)


pkg/chunk/gcp/gcs_object_client.go
```diff
@@ -130,7 +137,7 @@ func (s *GCSObjectClient) List(ctx context.Context, prefix, delimiter string) ([
 		}
 
 		// When doing query with Delimiter, Prefix is the only field set for entries which represent synthetic "directory entries".
-		if attr.Name == "" {
+		if attr.Prefix != "" {
```
pracucci, line 140: May you confirm Prefix is not also set objects?

pracucci (approved):
LGTM with a question

pkg/chunk/gcp/gcs_object_client.go
pstibrany: It is not an attribute returned in objects (“items”) in the API. Only Go exposes it via same struct.

See “Response” in https://cloud.google.com/storage/docs/json_api/v1/objects/list

pstibrany: And https://cloud.google.com/storage/docs/json_api/v1/objects for list of fields for objects.

Issue #3274, opened 2020-10-02
Proposal: Migrate Cortex CI/CD from CircleCI to GitHub Actions


AzfaarQureshi:
# Is your feature request related to a problem? Please describe.

The current CI/CD provider for Cortex, CircleCI, offers useful features but it also leaves room for a better developer experience. A better developer experience would include having easy access to test results, logs, and artifacts which is not currently possible with CircleCI. Currently, these are not available to developers until they provide CircleCI with OAuth access to their GitHub account. In addition, CircleCI workflows do not work out-of-box for forked repositories. One can personally setup an account with CircleCI and link it to their forked repository. However, their workflows would be subject to the free plan restrictions and can become a barrier for long term contribution when limited free-tier minutes run out. Alternatively, if one does not have a CircleCI account setup, CI tests will only run once a PR is opened upstream. This results in lower quality and unstable initial PRs which wastes development cycles. Another consideration that most developers working in large organizations have is security. Sharing sensitive data, such as OAuth tokens or private repositories to a third party creates potential security vulnerabilities which may be an issue for organizations with strict security requirements.

# Describe the solution you'd like

We propose that Cortex migrates from CircleCI to GitHub Actions as its CI/CD provider. Here are the reasons why:

### Ease Of Use

CI workflows written in GitHub Actions work out of the box on forked repositories. Currently with CircleCI, contributors are unable to run the CI pipeline until after an upstream pull request is made. Adopting GitHub Actions will ensure more stable and higher quality PRs before community effort is spent reviewing them. This helps increases developer velocity by reducing feedback cycles.

### Transparency

GitHub Actions help us align more with the core open source value of transparency. Test results, logs, and artifacts can be viewed freely from within the GitHub repository rather than having this information withheld behind a third party service.

### Rich Feature-set

GitHub Actions supports a wider variety of workflows than CircleCI. Github Actions has thirty unique [workflow triggers](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows) while CircleCI only has two: [push and schedule](https://circleci.com/docs/2.0/configuration-reference/#triggers). Workflows like pull request templates, closing stale issues and running custom jobs on new releases are all natively supported with GitHub Actions.

### Future Proofing

[...]

PR #1399, opened 2019-05-18
All should buld the docker images (and therefore the binaries and protos via dependancies), and not the exes directly.

tomwilkie:
Also:
- reduce the spam when building by hiding the docker run commands.
- check in the protos so Cortex can be used as a dependancy (fixes #1371).
- [x] TODO: verify generated proto code in CI

On master, running `make` builds the binaries twice and enters the build container too often.

```
>>>> Entering build container: exes
make: Entering directory '/go/src/github.com/cortexproject/cortex'
protoc -I /go/src:./vendor:./pkg/chunk/storage --gogoslick_out=plugins=grpc:./pkg/chunk/storage ./pkg/chunk/storage/caching_index_client.proto
protoc -I /go/src:./vendor:./pkg/ingester/client --gogoslick_out=plugins=grpc:./pkg/ingester/client ./pkg/ingester/client/cortex.proto
protoc -I /go/src:./vendor:./pkg/querier/frontend --gogoslick_out=plugins=grpc:./pkg/querier/frontend ./pkg/querier/frontend/frontend.proto
protoc -I /go/src:./vendor:./pkg/distributor --gogoslick_out=plugins=grpc:./pkg/distributor ./pkg/distributor/ha_tracker.proto
protoc -I /go/src:./vendor:./pkg/ring --gogoslick_out=plugins=grpc:./pkg/ring ./pkg/ring/ring.proto
CGO_ENABLED=0 go build -ldflags "-extldflags \"-static\" -s -w" -tags netgo -o cmd/cortex/cortex ./cmd/cortex
CGO_ENABLED=0 go build -ldflags "-extldflags \"-static\" -s -w" -tags netgo -o cmd/test-exporter/test-exporter ./cmd/test-exporter
make: Leaving directory '/go/src/github.com/cortexproject/cortex'

real	2m31.642s
user	0m0.031s
sys	0m0.016s

>>>> Entering build container: generated
make: Entering directory '/go/src/github.com/cortexproject/cortex'
make: Nothing to be done for 'generated'.
make: Leaving directory '/go/src/github.com/cortexproject/cortex'

real	0m5.574s
user	0m0.033s
sys	0m0.018s

>>>> Entering build container: cmd/cortex/cortex
make: Entering directory '/go/src/github.com/cortexproject/cortex'
CGO_ENABLED=0 go build -ldflags "-extldflags \"-static\" -s -w" -tags netgo -o cmd/cortex/cortex ./cmd/cortex
[...]

PR #1224, opened 2019-02-07
Save generated files in CI

bboreham:
It may be difficult to recreate those files exactly at a later date, so save what was used in the build.

PR #3302, opened 2020-10-08
 Migrate CircleCI workflows to GitHub Actions (1/3)

AzfaarQureshi:
### What this PR does

This is PR 1/3 of migrating Cortex's CI from CircleCI to GitHub Actions: 

**[Part 1/3](https://github.com/cortexproject/cortex/pull/3302)** 👈 
* create `test-build-deploy` workflow under `.github/workflows`
* add README
* adding the 3 base jobs
   * `lint`
   * `test`
   * `build`
* these jobs are run async and share no dependencies with each other
* workflow is run when a commit is pushed to `master` or when any PR is opened. However CD jobs will only run on the master branch and will be skipped otherwise
* all jobs in following PRs are dependant on one or more of these jobs finishing

**Part 2/3** 
* adding jobs dependant on `build`
    * `integration`
    * `integration-config-db`
* This PR completes the CI portion of migration. The following PR will finish the remaining CD portion of the migration

**Part 3/3** 
* This PR completes the migration from CircleCI to GitHub Actions
   * `deploy_website` is dependant on `build, test`
   * `deploy` is dependant on `build, test, lint, integration, integration-config-db`
* CD jobs will only run on the master branch and will be otherwise skipped

### Which issue(s) this PR fixes:
Fixes https://github.com/cortexproject/cortex/issues/3274

### Checklist
 - [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -1,0 +1,94 @@
+# GitHub Actions CI/CD
+
+The purpose of this workflow is to run all continuous integration (CI) and continuous deployment (CD) jobs when needed while respecting their internal dependencies. The continuous integration jobs serve to ensure new code passes linting, unit tests and integration tests before reaching the master branch. The continuous deployment jobs serve to deploy the latest version of the code to cortex and the website when merged with master.
+
+## Contributing
+
+If you wish to add a new CI or CD job, add it to the existing current test-build-deploy workflow and make sure it does not prevent any of the other jobs from passing. If you wish to change any of the build or testing images, update it in all sections are containers are often reused. If you wish to add an entirely new workflow, create a new yml file with separate triggers and filters. In all cases, clearly document any changes made to the workflows, images and dependencies below.
+
+## Test, Build and Deploy
+
+test-build-deploy.yml specifies a workflow that runs all Cortex continuous integration and continuous deployment jobs. The workflow is triggered on every pull request and commit to master, however the CD jobs only run when changes are merged onto master . The workflow combines both CI and CD jobs, because the CD jobs are dependent on artifacts produced the CI jobs.
+
+
+## Specific Jobs
+
+| Job                    | Description                                                                                                                   | Type |
+|------------------------|-------------------------------------------------------------------------------------------------------------------------------|------|
+| lint                   | Runs linting and ensures vendor directory, protos and generated documentation are consistent.                                 | CI   |
+| test                   | Runs units tests on Cassandra testing framework.                                                                              | CI   |
+| integration-configs-db | Integration tests for database configurations.                                                                                | CI   |
+| integration            | Runs integration tests after upgrading golang, pulling necessary docker images and downloading necessary module dependencies. | CI   |
+| build                  | Builds and saves an up-to-date Cortex image and website.                                                                      | CI   |
+| deploy_website         | Deploys the latest version of Cortex website to gh-pages branch. Triggered within workflow.                                   | CD   |
+| deploy                 | Deploys the latest Cortex image.                                                                                              | CD   |
+
+## Job Dependency Graph
+
+Internal dependencies between jobs illustrated below. Jobs run concurrently where possible but do not start until all jobs they depend on have completed successfully.
+
+
+![cortex_test-build-deploy](https://user-images.githubusercontent.com/20804975/95492784-9b7feb80-0969-11eb-9934-f44a4b1da498.png)
+
+### Key Details
+
+**Naming Convention**
+
+Each step in a job has a clear name that encapsulates the purpose of the command. The convention we are using is each word in the name should be capitalized except articles and prepositions. This creates consistent labeling when looking at the progress of the current workflow on GitHub.
+
+```yaml
+- name: Checkout Repo
+# commands
+- name: Get Dependencies
+# commands
+```
+
+**Checkout Version**
+
+Current build-image ships with an older version of Git which breaks github/actions@v2 so we are using actions/checkout@v1 for all jobs until the quay image is updated to ship with a more recent version of Git.
+
+```yaml
+- name: Checkout Repo
+  uses: actions/checkout@v1
+```
+
+**Symbolic Link to Expected Workspace**
+
+A significant number of commands in the Makefile are hardcoded with an assumed file structure of the CI container. To ensure paths specified in previous commands don’t break, a symlink was created from the hardcoded “expected” working directory `/go/src/github.com/cortexproject/cortex` to the actual working directory `$GITHUB_WORKSPACE`.
+
+```yaml
+- name: Sym link expected path to github workspace
+  run: |
+    mkdir -p /go/src/github.com/cortexproject/cortex
+    ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
+```
+
+**Sharing Artifacts Between Jobs**
+
+As of October 2020, GitHub Actions do not persist between different jobs in the same workflow. Each job is run on a fresh virtual environment (https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#runners). As such, we need to upload and download artifacts to share data between jobs.
+
+Artifact	Stored in	Used by	Purpose of storing artifact
+website public	build	deploy_website	share between jobs
+Docker Images	build	deploy, integration, integrations-config-db	share between jobs
+Frontend Protobuf	build	-	long term storage
+Caching Index Client Protobuf	build	-	long term storage
+Ring Protobuf 	build	-	long term storage
+Rules Protobuf	build	-	long term storage
+
+*Note:* Docker Images are zipped before uploading as a workaround. The images contain characters that are illegal in the upload-artifact action.
+```yaml
+- name: Compressing Images
+        run: tar -zcvf images.tar.gz /tmp/images
+      - name: Cache Images
+        uses: actions/upload-artifact@v2
+        with:
+          name: Docker Images
+          path: ./images.tar.gz
+```
+**Tags**
+
+As of Oct 2020, GitHub [does not support](https://github.community/t/using-regex-for-filtering/16427/2) regex for tag filtering. The regex /^v[0-9]+(\.[0-9]+){2}(-.+|[^-.]*)$/ was approximated using the available GitHub [filter patterns](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)
+```yaml
+tags:
+  - v[0-9]+.[0-9]+.[0-9]+**
+```
--- a/.github/README.md
+++ b/.github/README.md
@@ -67,13 +67,14 @@ A significant number of commands in the Makefile are hardcoded with an assumed f
 
 As of October 2020, GitHub Actions do not persist between different jobs in the same workflow. Each job is run on a fresh virtual environment (https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#runners). As such, we need to upload and download artifacts to share data between jobs.
 
-Artifact	Stored in	Used by	Purpose of storing artifact
-website public	build	deploy_website	share between jobs
-Docker Images	build	deploy, integration, integrations-config-db	share between jobs
-Frontend Protobuf	build	-	long term storage
-Caching Index Client Protobuf	build	-	long term storage
-Ring Protobuf 	build	-	long term storage
-Rules Protobuf	build	-	long term storage
+| Artifact                      | Stored In | Used By                                     | Purpose of Storing Artifact |
+|-------------------------------|-----------|---------------------------------------------|-----------------------------|
+| website public                | build     | deploy_website                              | share data between jobs     |
+| Docker Images                 | build     | deploy, integration, integrations-config-db | share data between jobs     |
+| Frontend Protobuf             | build     |                                             | long term storage           |
+| Caching Index Client Protobuf | build     |                                             | long term storage           |
+| Ring Protobuf                 | build     |                                             | long term storage           |
+| Rules Protobuf                | build     |                                             | long term storage           |
 
 *Note:* Docker Images are zipped before uploading as a workaround. The images contain characters that are illegal in the upload-artifact action.
 ```yaml
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -1,0 +1,31 @@
+name: ci
+on:
+  push:
+    branches: [master]
+    tags:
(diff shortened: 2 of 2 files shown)


.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,117 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/cortexproject/build-image:update-golang-1.14.9-eb0c8d4d2
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
+      - name: Sym Link Expected Path to Workspace
+        run: |
+          mkdir -p /go/src/github.com/cortexproject/cortex
+          ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
```
AzfaarQureshi, line 19: A significant number of commands in the Makefile were hardcoded with an assumed file structure of the CI container. make sure previous commands don’t break a symlink was created instead from the hardcoded “expected” working directory `go/src/github.com/cortexproject/cortex` to the actual working directory `$GITHUB_WORKSPACE`.

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,117 @@
+name: ci
+on:
+  push:
+    branches: [master]
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+** # Tag filters not as strict due to different regex system on Github Actions
```
AzfaarQureshi, line 6: As of Oct 2020, GitHub [does not support](https://github.community/t/using-regex-for-filtering/16427/2)regex for tag filtering. The old regex  `/^v[0-9]+(\.[0-9]+){2}(-.+|[^-.]*)$/` was approximated using the available GitHub [filter patterns](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,117 @@
...
+          mkdir /tmp/images
+          ln -s /tmp/images ./docker-images
+          make BUILD_IN_CONTAINER=false save-images
+      - name: Zip Images
+        run: tar -zcvf images.tar.gz /tmp/images
+      - name: Upload Images Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: Docker Images
+          path: ./images.tar.gz
```
AzfaarQureshi, line 117: As of Oct 2020 GitHub Actions does not persist between different jobs in the same workflow. Each job is run on a fresh [virtual environment](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#runners). As such, we need to upload and download artifacts to share data between jobs. 


Docker Images are zipped before uploading as a workaround. The images contain characters that are illegal in the upload-artifact action.

.github/README.md
```diff
@@ -0,0 +1,94 @@
...
+  run: |
+    mkdir -p /go/src/github.com/cortexproject/cortex
+    ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
+```
+
+**Sharing Artifacts Between Jobs**
+
+As of October 2020, GitHub Actions do not persist between different jobs in the same workflow. Each job is run on a fresh virtual environment (https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#runners). As such, we need to upload and download artifacts to share data between jobs.
+
+Artifact	Stored in	Used by	Purpose of storing artifact
```
pracucci, line 70: I think you wanted to format this as a table, but it's missing the table markdown syntax to render it correctly

.github/workflows/test-build-deploy.yml
pracucci: Looks good enough to me 👍

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,117 @@
...
+      image: quay.io/cortexproject/build-image:update-golang-1.14.9-eb0c8d4d2
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
+      - name: Sym Link Expected Path to Workspace
+        run: |
+          mkdir -p /go/src/github.com/cortexproject/cortex
+          ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
+      - name: Generate protobuf files
+        run: "protoc -I /go/src:./vendor/github.com/thanos-io/thanos/pkg:./vendor/github.com/gogo/protobuf:./vendor:./pkg/ruler --gogoslick_out=plugins=grpc,Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,:./pkg/ruler ./pkg/ruler/ruler.proto"
```
pracucci, line 21: Why is this required? The `make BUILD_IN_CONTAINER=false check-protos` should already build them. Have you had any issue?

My main concern here is that this is a copy of what we already have in the `Makefile` and we would have to keep it synched whenever we add protobuf definitions.

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,117 @@
...
+        image: cassandra:3.11
+        env:
+          JVM_OPTS: "-Xms1024M -Xmx1024M"
+        ports:
+        - 9042:9042
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
+      - name: Get Dependencies
+        run: go get -v -t -d ./...
```
pracucci, line 48: Could you explain why is this required, please?

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,117 @@
...
+  push:
+    branches: [master]
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+** # Tag filters not as strict due to different regex system on Github Actions
+  pull_request:
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/cortexproject/build-image:update-golang-1.14.9-eb0c8d4d2
```
pracucci, line 12: Do you think we can use something similar to `defaults: &defaults` in the CircleCI yaml, in order to avoid the repetition of `image: quay.io/cortexproject/build-image:update-golang-1.14.9-eb0c8d4d2` and being able to define the build image only once in this file?

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,117 @@
...
+          name: Ring Protobuf
+          path: pkg/ring/ring.pb.go
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Cortex Protobuf
+          path: pkg/ingester/client/cortex.pb.go
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Rules Protobuf
+          path: pkg/ruler/rules/rules.pb.go
```
pracucci, line 105: Have you figured out if we actually need to upload these artifacts? I know we do it in CircleCI but I have the feeling they're unused (at least I don't remember if we still use them anywhere).

pracucci (approved):
Very good job! Overall LGTM. I just left few questions here and there, but I can't see any blocker.

.github/workflows/test-build-deploy.yml
AzfaarQureshi: Oh, we included it because this PR [here](https://github.com/cortexproject/cortex/pull/1224) made it seem like there was an explicit decision to save these files. Has the situation changed?

.github/README.md
AzfaarQureshi: oops 🤦  you are right. Sorry this slipped by!

.github/workflows/test-build-deploy.yml
shovnik: There exists a "defaults" field for GitHub actions, but it automatically applies to all jobs in the workflow. The integration and integration-configs-db jobs that will be added with PR 2/3 of the migration do not use the default container so using a default would require explicitly removing the default container from those jobs by setting them to None.

We would like to reuse setup commands and default containers with something like commands in CircleCI, but this feature is not available yet on GitHub Actions.

Status: https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851

.github/workflows/test-build-deploy.yml
AzfaarQureshi: tl;dr: our mistake mb! It's removed now


We couldn't see that line in the [lint command](https://github.com/cortexproject/cortex/blob/master/Makefile#L141) in the makefile so we added it manually. The line also didn't show up in console output under the Lint section
Screen Shot 2020-10-09 at 3 12 40 PM

so we kept it in. But on a second look I realize I missed the `lint: protos` [line](https://github.com/cortexproject/cortex/blob/master/Makefile#L81)  and if we remove line 20, the `protoc -i` is run in `make lint` and the output shows in the console
Screen Shot 2020-10-09 at 3 16 11 PM

.github/workflows/test-build-deploy.yml
AzfaarQureshi: GHA doesn't support yaml anchors yet because of some security reasons but they are actively working on it: https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/14

.github/workflows/test-build-deploy.yml
AzfaarQureshi: you are right, this is vestigial code. apologies! next PR will be more thoroughly double checked

.github/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+**Symbolic Link to Expected Workspace**
+
+A significant number of commands in the Makefile are hardcoded with an assumed file structure of the CI container. To ensure paths specified in previous commands don’t break, a symlink was created from the hardcoded “expected” working directory `/go/src/github.com/cortexproject/cortex` to the actual working directory `$GITHUB_WORKSPACE`.
+
+```yaml
+- name: Sym link expected path to github workspace
+  run: |
+    mkdir -p /go/src/github.com/cortexproject/cortex
+    ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
```
pstibrany, line 63: This looks suspicious... why create many links instead of one:

```
mkdir -p /go/src/github.com/cortexproject/
ln -s "$GITHUB_WORKSPACE" /go/src/github.com/cortexproject/cortex
```

.github/README.md
```diff
@@ -0,0 +1,95 @@
...
+| Docker Images                 | build     | deploy, integration, integrations-config-db | share data between jobs     |
+| Frontend Protobuf             | build     |                                             | long term storage           |
+| Caching Index Client Protobuf | build     |                                             | long term storage           |
+| Ring Protobuf                 | build     |                                             | long term storage           |
+| Rules Protobuf                | build     |                                             | long term storage           |
+
+*Note:* Docker Images are zipped before uploading as a workaround. The images contain characters that are illegal in the upload-artifact action.
+```yaml
+- name: Compressing Images
+        run: tar -zcvf images.tar.gz /tmp/images
```
pstibrany, line 82: Not sure if compression gives us anything here. We can use `tar` without `-z` flag to make it run faster (but use more disk space).

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,113 @@
...
+  pull_request:
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/cortexproject/build-image:update-golang-1.14.9-eb0c8d4d2
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
+      - name: Sym Link Expected Path to Workspace
```
pstibrany, line 16: This linking is explained in README, apart from doing review, who reads that? I'd include the explanations from README directly into this file.

.github/workflows/test-build-deploy.yml
pstibrany: Would it be possible to use env variable instead?

.github/workflows/test-build-deploy.yml
```diff
@@ -0,0 +1,113 @@
...
+        run: |
+          mkdir -p /go/src/github.com/cortexproject/cortex
+          ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
+      - name: Lint
+        run: make BUILD_IN_CONTAINER=false lint
```
pstibrany, line 21: Should we set `BUILD_IN_CONTAINER=false` as env variable for entire workflow or at least individual jobs?

.github/workflows/test-build-deploy.yml
pstibrany: Since then we have added these files into Git, so I think #1224 is not relevant anymore.

pstibrany (approved):
Great job, thank you very much for doing this!

(Nit: In next PRs I would suggest moving the comments from README.md into YAML files, to keep it at one place.)

.github/workflows/test-build-deploy.yml
pstibrany: (In https://github.com/cortexproject/cortex/pull/1399)

.github/workflows/test-build-deploy.yml
pracucci: _This is not a block for getting this PR merged. We can discuss it separately._

.github/workflows/test-build-deploy.yml
pracucci: I also think we don't need to store these artifacts anymore.

.github/workflows/test-build-deploy.yml
pracucci: But `BUILD_IN_CONTAINER` is not an environment variable. At least, doesn't work if you set it as env variable. Am I missing anything?

pracucci (approved):
LGTM, thanks!

pracucci:
@AzfaarQureshi We merged the PR since it's in a good shape, but we would be glad if you could address the remaining comments in a subsequent PR. Thanks!

pstibrany:
> But BUILD_IN_CONTAINER is not an environment variable. At least, doesn't work if you set it as env variable. Am I missing anything?

"Every environment variable that make sees when it starts up is transformed into a make variable with the same name and value." (from [GNU Make docs](https://www.gnu.org/software/make/manual/html_node/Environment.html))

.github/README.md
shovnik: The compression was in fact unnecessary. We removed it in the subsequent migration PR: https://github.com/cortexproject/cortex/pull/3341

.github/workflows/test-build-deploy.yml
shovnik: Added a comment explaining this within the code itself in the subsequent PR to address this.
https://github.com/cortexproject/cortex/pull/3341

Issue #3304, opened 2020-10-09
Reduce ingester churn during rolling restart


csmarchbanks:
**Is your feature request related to a problem? Please describe.**
Whenever I update ingesters or the configuration for ingesters samples are spread to other ingesters by design. This causes very small and inefficient chunks to be created in an unneeded ingester when running with a reasonable replication factor and a WAL. In addition, if an ingester crashes for some reason (typically an OOM), spreading additional data to other ingesters can amplify the problem.

**Describe the solution you'd like**
A way to configure ingesters not to remove themselves from the ring when they are being restarted, and for distributors not to extend their replica set in the presence of a leaving or unhealthy ingester.

Also, happy to hear if this is a terrible idea :)

**Describe alternatives you've considered**
Keep the status quo and try to reduce how often ingesters are updated.

**Additional context**
This request is very similar to #2823, but for ingesters instead of the store-gateway.

pracucci:
This is something we recently discussed internally at Grafana Labs and we would like to do too. I haven't looked at the PR you opened yet, but I will do it soon!

Issue #467, opened 2017-06-19
Mysterious flush of underutilised chunks 1hr after ingester rollout


awh:
On the 15th of June:
![screenshot from 2017-06-19 17-23-20](https://user-images.githubusercontent.com/1504438/27295498-13ea51d0-5515-11e7-8d8c-2de8b93238b2.png)

On the 19th of June:
![screenshot from 2017-06-19 17-23-33](https://user-images.githubusercontent.com/1504438/27295513-1ff314b2-5515-11e7-8fcd-2e25cbcdf091.png)

Both of these happened approximately one hour after an ingester upgrade in which chunks were successfully transferred from terminating ingesters. Chunk max idle is 1h, possibly related?

CC @tomwilkie any thoughts?

tomwilkie:
Do you know if these are spurious?  Or did 800k/3 timeseries stop being
written to?

awh:
Oh - could those be series from metrics related to the ingesters which have now gone away?

awh:
Querying for `{job="cortex/ingester"}` returns only 1406 time series.

tomwilkie:
I assume this was part of a rolling upgrade of all of cortex?  What does `count({job=~"cortex/.*"})` return?

awh:
9862, and no, in both cases it was just the ingester that was rolled out (in the first instance we had upgraded all of cortex but delayed the ingester whilst we understood #458, and in the second instance we only needed to do the ingester to deploy #460)

awh:
It's almost like the state reconstructed from transferred chunks in the new ingesters is not identical to that in the old, and so they are making different decisions about what should be flushed...

awh:
This is also happening on rolling reboots.

leth:
As @awh said, we just had a single ingester replaced, due to a node reboot, and the queue peaked at 600K

leth:
I've been doing a bit of digging into the code around transferring chunks.

It looks like all chunks are transferred to a new ingester which was found in the PENDING state.
However whether this is the correct ingester to transfer to depends on whether

leth:
Oops

leth:
I had a brainwave last night; during ingester handover the old and new ingesters are no longer in `ACTIVE` state, so [the distributor will pick an extra ingester](https://github.com/weaveworks/cortex/blob/master/pkg/ring/ring.go#L184), which was not selected before the handover started.
[...]

PR #3305, opened 2020-10-09
Allow ingesters to stay in the ring during restart

### LINKED ISSUE #2823: Store-gateway blocks resharding during rollout
When running the blocks storage, store-gateways reshard blocks whenever the ring topology changes. This means that during a rollout of the store-gateways (ie. deploy a config change or version upgrade) blocks are resharded across instances. 

This is highly inefficient in a cluster with a large number of tenants or few very large tenants. Ideally, no blocks resharding should occur during a rollout (if the blocks replication factor is > 1).

## Rollouts

We could improve the system to avoid the blocks resharding when the following conditions are met:
- `-experimental.store-gateway.replication-factor` is `> 1` (so that while a store-gateway restarts all of its blocks are replicated to at least another instance)
- `-experimental.store-gateway.tokens-file-path` is configured (so that previous tokens are picked up on restart)
- The store-gateway instance ID is stable across restarts (ie. Kubernetes Statfulsets)

To avoid blocks resharding during store-gateways rollout we need the restarting store-gateway instance to **not** be unregistered from the ring during the restart.

When a store-gateway shutdowns, the instance could be left in the `LEAVING` state within the ring and we could change the `BlocksReplicationStrategy.ShouldExtendReplicaSet()` to **not** extend the replica set if an instance is in the `LEAVING` state.

This means that during the rollout, for the blocks hold by the restarting replica there will be N-1 replicas (contrary to the N desired replicas configured). Once the instance restarts, it will have the same instance ID and same tokens (assuming `tokens-file-path` is configured) and thus will replace its state from `LEAVING` to `JOINING` within the ring.

## Scale down

There's no way to distinguish between a rollout and a scale down: the process just receives a termination signal.

This means that during a scale down, the instance would be left in the `LEAVING` state within the ring. However, the store-gateway has an auto-forget feature which removes unhealthy instances after 10x heartbeat timeouts (default: 1m timeout = 10m before an unhealthy instance is forgotten).

A scale down of a number of instance < replication factor could leverage on the auto-forget. However, there's no easy way to have a smooth scale down unless we'll have a way to signal the process whether it's going to shutdown because of a scale down or rollout.

## Crashes

_In case a store-gateway crashes, there would be no difference compared to today._

csmarchbanks:
By keeping an ingester in the ring and not writing to additional
ingesters while it is being restarted, will stop small chunks from being
persisted to other ingesters. This will allow users running with a WAL
to avoid excess load and memory during a rolling restart.

I am curious what others think of this, if accepted, I think a similar
approach could be applied to the store-gateway (issue #2823). The
downside is that scaling a cortex cluster down while running with skip
unregister will involve manual forgetting of the ingester (or automation
to do this).

Fixes: #3304

**Checklist**
- [x] Tests updated (some tests use old way, some use new way).
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## master / unreleased
 
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
-* [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts.
+* [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts. #3305
 * [ENHANCEMENT] Added zone-awareness support on queries. When zone-awareness is enabled, queries will still succeed if all ingesters in a single zone will fail. #3414
 * [ENHANCEMENT] Blocks storage ingester: exported more TSDB-related metrics. #3412
   - `cortex_ingester_tsdb_wal_corruptions_total`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master / unreleased
 
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
+* [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts.
 * [ENHANCEMENT] Added zone-awareness support on queries. When zone-awareness is enabled, queries will still succeed if all ingesters in a single zone will fail. #3414
 * [ENHANCEMENT] Blocks storage ingester: exported more TSDB-related metrics. #3412
   - `cortex_ingester_tsdb_wal_corruptions_total`
--- a/pkg/ring/lifecycler.go
+++ b/pkg/ring/lifecycler.go
@@ -124,6 +124,7 @@ type Lifecycler struct {
 
 	// Whether to flush if transfer fails on shutdown.
 	flushOnShutdown *atomic.Bool
+	skipUnregister  *atomic.Bool
 
 	// We need to remember the ingester state, tokens and registered timestamp just in case the KV store
 	// goes away and comes back empty. The state changes during lifecycle of instance.
@@ -182,6 +183,7 @@ func NewLifecycler(cfg LifecyclerConfig, flushTransferer FlushTransferer, ringNa
 		RingName:        ringName,
 		RingKey:         ringKey,
 		flushOnShutdown: atomic.NewBool(flushOnShutdown),
+		skipUnregister:  atomic.NewBool(cfg.SkipUnregister),
 		Zone:            zone,
 
 		actorChan: make(chan func()),
@@ -490,7 +492,7 @@ heartbeatLoop:
 		}
 	}
 
-	if !i.cfg.SkipUnregister {
+	if !i.skipUnregister.Load() {
 		if err := i.unregister(context.Background()); err != nil {
 			return perrors.Wrapf(err, "failed to unregister from the KV store, ring: %s", i.RingName)
 		}
@@ -777,6 +779,16 @@ func (i *Lifecycler) FlushOnShutdown() bool {
 // Passing 'true' enables it, and 'false' disabled it.
 func (i *Lifecycler) SetFlushOnShutdown(flushOnShutdown bool) {
 	i.flushOnShutdown.Store(flushOnShutdown)
+}
+
+// SkipUnregister returns if unregistering is skipped on shutdown.
+func (i *Lifecycler) SkipUnregister() bool {
+	return i.skipUnregister.Load()
+}
+
+// SetSkipUnregister enables/disables unregistering on shutdown.
+func (i *Lifecycler) SetSkipUnregister(skipUnregister bool) {
+	i.skipUnregister.Store(skipUnregister)
 }
 
 func (i *Lifecycler) processShutdown(ctx context.Context) {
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
 * [CHANGE] `cortex_alertmanager_config_invalid` has been removed in favor of `cortex_alertmanager_config_last_reload_successful`. #3289
 * [FEATURE] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-tenant` globally, or using per-tenant limit `max_queriers_per_tenant`), each tenants's requests will be handled by different set of queriers. #3113 #3257
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
+* [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts.
```
csmarchbanks, line 53: Well, this is for myself, but may as well be consistent to make sure I don't forget :)

:wave: The v1.5 release just started and you will need to rebase to the latest master and move any CHANGELOG entries into the unreleased section. Thanks!

pkg/ring/lifecycler.go
```diff
@@ -56,12 +56,12 @@ type LifecyclerConfig struct {
 	FinalSleep       time.Duration `yaml:"final_sleep"`
 	TokensFilePath   string        `yaml:"tokens_file_path"`
 	Zone             string        `yaml:"availability_zone"`
+	SkipUnregister   bool          `yaml:"skip_unregister"`
```
pstibrany, line 59: I'd suggest calling it UnregisterFromRing, and make it true by default (for compatibility).

pkg/ring/lifecycler.go
```diff
@@ -778,6 +781,16 @@ func (i *Lifecycler) SetFlushOnShutdown(flushOnShutdown bool) {
 	i.flushOnShutdown.Store(flushOnShutdown)
 }
 
+// SkipUnregister returns if unregistering is skipped on shutdown.
+func (i *Lifecycler) SkipUnregister() bool {
+	return i.skipUnregister.Load()
+}
+
+// SetSkipUnregister enables/disables unregistering on shutdown.
```
pstibrany, line 789: This comment is slightly misleading, because by passing "true", we're actually disabling the unregister. (Why I suggested the rename)

pkg/ring/ring.go
```diff
@@ -126,6 +127,7 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.DurationVar(&cfg.HeartbeatTimeout, prefix+"ring.heartbeat-timeout", time.Minute, "The heartbeat timeout after which ingesters are skipped for reads/writes.")
 	f.IntVar(&cfg.ReplicationFactor, prefix+"distributor.replication-factor", 3, "The number of ingesters to write to and read from.")
 	f.BoolVar(&cfg.ZoneAwarenessEnabled, prefix+"distributor.zone-awareness-enabled", false, "True to enable the zone-awareness and replicate ingested samples across different availability zones.")
+	f.BoolVar(&cfg.ExtendWrites, prefix+"distributor.extend-writes", true, "Try writing to an additional ingester in the presence of an ingester not in the ACTIVE state.")
```
pstibrany, line 130: We can extend the description to explain why would someone want to disable this, and that it's perhaps best used with disabled unregister?

pstibrany (commented):
Thank you. I've left some comments/suggestions. Can we also modify api/_index.md to document that /shutdown will unregister from the ring, even when unregistering is disabled?

csmarchbanks:
Thanks for the review! I implemented all of the suggestions - the rename did mean changing some tests around that were depending on SkipUnregister being default false.

pstibrany (approved):
LGTM, thanks. Let's wait for @pracucci's review as well, before merging, please.

csmarchbanks:
Friendly reminder for @pracucci to review as requested :)

pkg/ring/lifecycler.go
```diff
@@ -778,6 +781,16 @@ func (i *Lifecycler) SetFlushOnShutdown(flushOnShutdown bool) {
 	i.flushOnShutdown.Store(flushOnShutdown)
 }
 
+// SkipUnregister returns if unregistering is skipped on shutdown.
```
jtlisi, line 784: **suggestion:**

```suggestion
// UnregisterFromRing returns if unregistering is skipped on shutdown.
```

pkg/ring/lifecycler.go
```diff
@@ -778,6 +781,16 @@ func (i *Lifecycler) SetFlushOnShutdown(flushOnShutdown bool) {
...
+func (i *Lifecycler) UnregisterFromRing() bool {
```
jtlisi, line 785: **suggestion(non-blocking,if-minor):** I would rename this function to parallel `SetUnregisterFromRing`. In my opinion `UnregisterFromRing()` connotes that it will actually perform the operation to unregister from the ring, not return the config value.

```suggestion
func (i *Lifecycler) GetUnregisterFromRing() bool {
```

jtlisi (approved):
LGTM, modulo some minor nits/suggestions

pkg/ring/lifecycler.go
csmarchbanks: Do'h, I really thought I renamed that too... Fixed.

pkg/ring/lifecycler.go
csmarchbanks: I changed it to `Should` which read nicer to me than `Get`.

pracucci (approved):
LGTM, thanks!

bboreham:
Ref #467 which I think this fixes, at least for the WAL restart-in-place case.

bboreham:
I'll mention another downside: you lose redundancy.

With replication-factor=3, if you turn off "extend writes" then do a rolling update, you are down to 2 copies of some data. If one of those is then lost then queries will be unstable due to https://github.com/cortexproject/cortex/issues/731.

PR #3309, opened 2020-10-09
Blocks compactionLoop: Respect context cancellation faster.

pstibrany:
**What this PR does**: [compactionLoop](https://github.com/cortexproject/cortex/blob/e26519995ae889d0f7a187a1e74dde99571fa39d/pkg/ingester/ingester_v2.go#L1194) method in blocks-ingester can ignore finished context if compaction takes too long. On next iteration, if multiple `select` cases are active (eg. both cancellation and ticker), Go runtime will [randomly](https://golang.org/ref/spec#Select_statements) select one. If it selects the ticker again, another compaction will run and block the shutdown.

This PR modifies the compaction loop to stop faster if context is cancelled.


pracucci (approved):
LGTM

PR #3310, opened 2020-10-09
[BUGFIX] Fix reporting of `cortex_prometheus_rule_group_duration_seconds`

gotjosh:
**What this PR does**:

Whilst building new dashboards for the ruler I noticed no data was being reported via this metric, turns out it was a bug.

Fixes a small bug in the ruler metrics, where we used the wrong name to
match against the upstream metric. This caused the metric to not report
any data at all.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ruler/manager_metrics.go
+++ b/pkg/ruler/manager_metrics.go
@@ -148,7 +148,7 @@ func (m *ManagerMetrics) Collect(out chan<- prometheus.Metric) {
 	// If same user is later re-added, all metrics will start from 0, which is fine.
 
 	data.SendSumOfSummariesPerUser(out, m.EvalDuration, "prometheus_rule_evaluation_duration_seconds")
-	data.SendSumOfSummariesPerUser(out, m.IterationDuration, "cortex_prometheus_rule_group_duration_seconds")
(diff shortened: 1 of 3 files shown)


pkg/ruler/manager_metrics.go
```diff
@@ -148,7 +148,7 @@ func (m *ManagerMetrics) Collect(out chan<- prometheus.Metric) {
 	// If same user is later re-added, all metrics will start from 0, which is fine.
 
 	data.SendSumOfSummariesPerUser(out, m.EvalDuration, "prometheus_rule_evaluation_duration_seconds")
-	data.SendSumOfSummariesPerUser(out, m.IterationDuration, "cortex_prometheus_rule_group_duration_seconds")
+	data.SendSumOfSummariesPerUser(out, m.IterationDuration, "prometheus_rule_group_duration_seconds")
```
gotjosh, line 151: Names need to match against the upstream metric, otherwise, we won't be able to match them against the new renamed metric that's part of `m.IterationDuration`

pstibrany (approved):

pracucci (approved):
Good job spotting and fixing it. Thanks!

PR #3312, opened 2020-10-09
Auth injector defaults

pstibrany:
**What this PR does**: This PR improves auth injector tool a bit:
- It changes default :80 to :8080
- It checks for empty remote address explicitly, since otherwise proxy would start forwarding to empty URL.
- It prints address that it's listening on, and forwarding to


pracucci (approved):
Fantastic!

PR #3313, opened 2020-10-09
Fix flaky test.

pstibrany:
**What this PR does**: This PR attempts to fix flaky test in `ring_test.go` by adding more CAS retries and using randomized delay between CAS retries.

~This PR will fix flaky test in ring_test.go. For now it just captures and logs output when test fails to get full understanding on what's going on.~ (I couldn't reproduce failure on CI machine)


pracucci (approved):
Discussed offline. Not a big fan, but if there's no better solution than let's fix the flaky test this way.

PR #3314, opened 2020-10-09
Fixed gRPC connections leaking in rulers when rulers sharding is enabled and APIs called

pracucci:
**What this PR does**:
We just found out that the ruler is leaking gRPC connections when rulers sharding is enabled and APIs called. This is an hot fix. The proper fix (that will be done later) is to use the gRPC clients pool like we do in any other place.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


gotjosh (approved):

pracucci:
I've created an issue for the proper fix we should do: https://github.com/cortexproject/cortex/issues/3315

PR #3320, opened 2020-10-11
feat: simplify query related modules

jtlisi:
**What this PR does**:

- Simplify the querier handler code by consolidating the querier specific latency metrics and the custom handler for the internal querier into a single function.
- Separates the `querier` module into two modules: `querier` and `queryable`.
- Decouples the instantiation of the `queryrange` tripperware from the `query-frontend`

The goal of this PR is to simplify the setup of the querier and make it clearer how requests are routed to the querier when running with or without the query-frontend enabled simultaneously and to decouple the initialization of the default queryable and the initialization of the request processing/handling code.

I did my best in this PR to comment/clarify how the routing for the querier is instantiated when running standalone or as a part of single-binary.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -6,11 +6,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
-
 	"github.com/felixge/fgprof"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/weaveworks/common/middleware"
 	"github.com/weaveworks/common/server"
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1,28 +1,17 @@
 package api
 
 import (
-	"context"
-	"errors"
 	"flag"
 	"net/http"
-	"regexp"
 	"strings"
 	"time"
 
-	"github.com/opentracing-contrib/go-stdlib/nethttp"
-	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
-	dto "github.com/prometheus/client_model/go"
 
 	"github.com/felixge/fgprof"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/gorilla/mux"
-	"github.com/prometheus/common/route"
-	"github.com/prometheus/prometheus/config"
-	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/storage"
-	v1 "github.com/prometheus/prometheus/web/api/v1"
 	"github.com/weaveworks/common/middleware"
 	"github.com/weaveworks/common/server"
 
@@ -104,12 +93,6 @@ func New(cfg Config, serverCfg server.Config, s *server.Server, logger log.Logge
 // RegisterRoute registers a single route enforcing HTTP methods. A single
 // route is expected to be specific about which HTTP methods are supported.
 func (a *API) RegisterRoute(path string, handler http.Handler, auth bool, method string, methods ...string) {
-	a.registerRouteWithRouter(a.server.HTTP, path, handler, auth, method, methods...)
-}
-
-// RegisterRoute registers a single route to a router, enforcing HTTP methods. A single
-// route is expected to be specific about which HTTP methods are supported.
-func (a *API) registerRouteWithRouter(router *mux.Router, path string, handler http.Handler, auth bool, method string, methods ...string) {
 	methods = append([]string{method}, methods...)
 
 	level.Debug(a.logger).Log("msg", "api: registering route", "methods", strings.Join(methods, ","), "path", path, "auth", auth)
@@ -117,10 +100,10 @@ func (a *API) registerRouteWithRouter(router *mux.Router, path string, handler h
 		handler = a.authMiddleware.Wrap(handler)
 	}
 	if len(methods) == 0 {
-		router.Path(path).Handler(handler)
+		a.server.HTTP.Path(path).Handler(handler)
 		return
 	}
-	router.Path(path).Methods(methods...).Handler(handler)
+	a.server.HTTP.Path(path).Methods(methods...).Handler(handler)
 }
 
 func (a *API) RegisterRoutesWithPrefix(prefix string, handler http.Handler, auth bool, methods ...string) {
@@ -133,20 +116,6 @@ func (a *API) RegisterRoutesWithPrefix(prefix string, handler http.Handler, auth
 		return
 	}
 	a.server.HTTP.PathPrefix(prefix).Methods(methods...).Handler(handler)
-}
-
-// Latest Prometheus requires r.RemoteAddr to be set to addr:port, otherwise it reject the request.
-// Requests to Querier sometimes doesn't have that (if they are fetched from Query-Frontend).
-// Prometheus uses this when logging queries to QueryLogger, but Cortex doesn't call engine.SetQueryLogger to set one.
-//
-// Can be removed when (if) https://github.com/prometheus/prometheus/pull/6840 is merged.
-func fakeRemoteAddr(handler http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.RemoteAddr == "" {
-			r.RemoteAddr = "127.0.0.1:8888"
-		}
-		handler.ServeHTTP(w, r)
-	})
 }
 
 // RegisterAlertmanager registers endpoints associated with the alertmanager. It will only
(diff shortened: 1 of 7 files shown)


pkg/cortex/modules.go
```diff
@@ -397,7 +413,9 @@ func (t *Cortex) initDeleteRequestsStore() (serv services.Service, err error) {
 	return
 }
 
-func (t *Cortex) initQueryFrontend() (serv services.Service, err error) {
+// initTripperware instantiates the tripperware used by the query frontend
+// to optimize Prometheus query requests.
+func (t *Cortex) initTripperware() (serv services.Service, err error) {
```
pracucci, line 418: Maybe we could rename it to `QueryFrontTripperware()` for clarify?

pkg/api/handlers.go
```diff
@@ -109,3 +128,106 @@ func configHandler(cfg interface{}) http.HandlerFunc {
...
+		Duration:         querierRequestDuration,
+		RequestBodySize:  receivedMessageSize,
+		ResponseBodySize: sentMessageSize,
+		InflightRequests: inflightRequests,
+	}
+	cacheGenHeaderMiddleware := getHTTPCacheGenNumberHeaderSetterMiddleware(tombstonesLoader)
+	middlewares := middleware.Merge(fakeRemoteAddr(), inst, cacheGenHeaderMiddleware)
+	router.Use(middlewares.Wrap)
+
+	promRouter := route.New().WithPrefix(cfg.ServerPrefix + cfg.PrometheusHTTPPrefix + "/api/v1")
```
pracucci, line 207: Am I missing anything, or `promRouter` and `legacyPromRouter` don't have `middlewares` attached?

pkg/cortex/cortex.go
```diff
@@ -260,10 +261,12 @@ type Cortex struct {
 	DeletesStore     *purger.DeleteStore
 	Frontend         *frontend.Frontend
 	TableManager     *chunk.TableManager
-	Cache            cache.Cache
 	RuntimeConfig    *runtimeconfig.Manager
 	Purger           *purger.Purger
 	TombstonesLoader *purger.TombstonesLoader
+	QuerierQueryable prom_storage.SampleAndChunkQueryable
+	QuerierEngine    *promql.Engine
+	Tripperware      frontend.Tripperware
```
pracucci, line 269: Should we rename it `QueryFrontendTripperware` for clarity, even if longer?

pkg/cortex/modules.go
```diff
@@ -189,55 +189,71 @@ func (t *Cortex) initDistributor() (serv services.Service, err error) {
...
+	} else {
+		// Single binary mode requires a query frontend endpoint for the worker. If no frontend endpoint
+		// is configured, Cortex will default to using localhost on it's own GRPC listening port.
+		if t.Cfg.Worker.Address == "" {
+			address := fmt.Sprintf("127.0.0.1:%d", t.Cfg.Server.GRPCListenPort)
+			level.Warn(util.Logger).Log("msg", "Worker address is empty in single binary mode.  Attempting automatic worker configuration.  If queries are unresponsive consider configuring the worker explicitly.", "address", address)
+			t.Cfg.Worker.Address = address
+		}
+
+		// If queries processed using the external HTTP Server, we need wrap the internalHandler with
```
pracucci, line 247:
```suggestion
		// If queries are processed using the external HTTP Server, we need wrap the internalHandler with
```

pracucci: Also, what's `internalHandler`?

pkg/api/api.go
```diff
@@ -293,112 +263,22 @@ func (a *API) RegisterCompactor(c *compactor.Compactor) {
...
-	api.Register(promRouter)
-	cacheGenHeaderMiddleware := getHTTPCacheGenNumberHeaderSetterMiddleware(tombstonesLoader)
-	promHandler := fakeRemoteAddr(inst.Wrap(cacheGenHeaderMiddleware.Wrap(promRouter)))
-
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/read", querier.RemoteReadHandler(queryable), true, "POST")
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/query", promHandler, true, "GET", "POST")
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/query_range", promHandler, true, "GET", "POST")
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/labels", promHandler, true, "GET", "POST")
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/label/{name}/values", promHandler, true, "GET")
-	a.registerRouteWithRouter(router, a.cfg.PrometheusHTTPPrefix+"/api/v1/series", promHandler, true, "GET", "POST", "DELETE")
```
pracucci, line 365: Who register these now?

pracucci (commented):
@jtlisi I tried my best to follow this refactoring but I found it a bit hard. The new design looks good to me, but I find difficult to understand if we're introducing any regression. I left few questions.

pkg/cortex/modules.go
jtlisi: In this case the internalHandler is the router that's returned by `NewQuerierHandler`. I say internalHandler in the sense that it is not the main router that is registered with the weaveworks common HTTP server to handle external requests. Any request forwarded to the internalHandler must either go through the external HTTP server router first or be passed to it from the frontend worker.

Maybe a clearer term would be `internalQuerierRouter`?

pkg/api/handlers.go
jtlisi: The middleware that was previously added to each of these routes has been moved to the router we instantiate earlier and register these routers with below. As a result any requests that reach these two routers will automatically have the associated middleware already applied.

We could wrap these two routers individually with the middleware as we did previously. However, that seems a bit redundant.

pkg/api/api.go
jtlisi: These routes are registered with the two routers when we call [api.Register](https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L259) on them. We end up having to register two routers because the upstream prometheus [route library](https://github.com/prometheus/common/blob/master/route/route.go#L88) and registration function requires us to register separately for each path prefix.

Previously we explicitly registered every route in the Prometheus API to query handler. However, now we only explicitly register the `metadata` and `read` routes because they aren't handled by the upstream API struct. For the rest we register the prefix and since the `promHandler` and `legacyPromHandler` are themselves routers, the requests should be forwarded correctly. 

Also I should note that the only way to hit these routes is by a request  being routed to the `internalQuerierHandler` which can only happen if it hit's the routes registered in `RegisterQueryAPI`, so no new routes should have been exposed and no existing routes should have been removed. This PR just simplified the router that already was only acting as a middleman to the upstream Prometheus routers we already registered with the API struct.

pkg/cortex/modules.go
jtlisi: Makes sense

pkg/cortex/modules.go
```diff
@@ -56,8 +54,10 @@ const (
 	IngesterService     string = "ingester-service"
 	Flusher             string = "flusher"
 	Querier             string = "querier"
+	Queryable           string = "queryable"
 	StoreQueryable      string = "store-queryable"
 	QueryFrontend       string = "query-frontend"
+	QueryTripperware    string = "query-tripperware"
```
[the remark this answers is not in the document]
jtlisi: Makes sense

jtlisi:
@pracucci I added a diagram that I hope makes things a bit clearer. Let me know if you find it helpful or not. Otherwise I can remove it since it adds a good amount of LOC.

pkg/api/middlewares.go
```diff
@@ -10,6 +10,22 @@ import (
 	"github.com/cortexproject/cortex/pkg/querier/queryrange"
 )
 
+// Latest Prometheus requires r.RemoteAddr to be set to addr:port, otherwise it reject the request.
+// Requests to Querier sometimes doesn't have that (if they are fetched from Query-Frontend).
+// Prometheus uses this when logging queries to QueryLogger, but Cortex doesn't call engine.SetQueryLogger to set one.
+//
+// Can be removed when (if) https://github.com/prometheus/prometheus/pull/6840 is merged.
+func fakeRemoteAddr() middleware.Interface {
```
pstibrany, line 18: We can remove this now.

pkg/cortex/modules.go
```diff
@@ -398,7 +452,9 @@ func (t *Cortex) initDeleteRequestsStore() (serv services.Service, err error) {
 	return
 }
 
-func (t *Cortex) initQueryFrontend() (serv services.Service, err error) {
+// initQueryFrontendTripperware instantiates the tripperware used by the query frontend
+// to optimize Prometheus query requests.
+func (t *Cortex) initQueryFrontendTripperware() (serv services.Service, err error) {
```
pstibrany, line 457: Nice!

pkg/api/handlers.go
```diff
@@ -109,3 +129,108 @@ func configHandler(cfg interface{}) http.HandlerFunc {
...
+	router.PathPrefix(cfg.PrometheusHTTPPrefix + "/api/v1/").Handler(promRouter)
+
+	//TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
+	// https://github.com/prometheus/prometheus/pull/7125/files
+	router.Path(cfg.LegacyHTTPPrefix + "/api/v1/metadata").Handler(querier.MetadataHandler(distributor))
+	router.Path(cfg.LegacyHTTPPrefix + "/api/v1/read").Handler(querier.RemoteReadHandler(queryable))
+	// A prefix is fine because external routes will be registered explicitly
+	router.PathPrefix(cfg.LegacyHTTPPrefix + "/api/v1/").Handler(legacyPromRouter)
+
+	// Since we have a new router and the request will not go trough the default server
```
pstibrany, line 230: Does this comment still apply?

pkg/cortex/modules.go
```diff
@@ -190,55 +190,109 @@ func (t *Cortex) initDistributor() (serv services.Service, err error) {
...
+
+	// Register the default endpoints that are always enabled for the querier module
+	t.API.RegisterQueryable(t.QuerierQueryable, t.Distributor)
+
+	return nil, nil
+}
+
+// initQuerier registers an internal HTTP router with a Prometheus API backed by the
+// Cortex Queryable. Then it does one of the following:
+// 1. Query-Frontend Enabled: If Cortex has an All or QueryFrontend target, the internal
```
pstibrany, line 209: To make sure comment is formatted properly, please put empty lines between paragraphs (or list items), and don't use extra space at the beginning of the lines (except for diagram below) -- such lines are formatted as code according to Go doc conventions.

https://blog.golang.org/godoc

pkg/cortex/modules.go
```diff
@@ -190,55 +190,109 @@ func (t *Cortex) initDistributor() (serv services.Service, err error) {
...
+		if t.Cfg.Worker.Address == "" {
+			address := fmt.Sprintf("127.0.0.1:%d", t.Cfg.Server.GRPCListenPort)
+			level.Warn(util.Logger).Log("msg", "Worker address is empty in single binary mode.  Attempting automatic worker configuration.  If queries are unresponsive consider configuring the worker explicitly.", "address", address)
+			t.Cfg.Worker.Address = address
+		}
+
+		// If queries are processed using the external HTTP Server, we need wrap the internal querier with
+		// HTTP router with middleware to parse the tenant ID from the HTTP header and inject it into the
+		// request context.
+		internalQuerierRouter = middleware.AuthenticateUser.Wrap(internalQuerierRouter)
```
pstibrany, line 288: We didn't do this before, did we? Why is it needed now?

pstibrany (approved):
Makes sense to me why we want this, although it looks bit scary. 🤞

pkg/cortex/modules.go
jtlisi: Actually, we did do this before because we previously used the `registerRouteWithRouter` function built into the API struct which applies this middleware. Now we just register the routes directly with with the router using the `Path|PathPrefix` function.

pkg/api/handlers.go
jtlisi: I think it would make sense to reword it or remove it but keep the tracing instrumentation since it can be useful to have a span for this particular router

pkg/cortex/modules.go
jtlisi: I will update this to match the conventions

pstibrany: Thanks. It helps tools to format it properly. (eg. GoLand :))

pkg/api/middlewares.go
jtlisi: 👍

pracucci (approved):
LGTM, thanks!

Issue #3318, opened 2020-10-10
Obtain tenant ID from Prometheus timeseries label


blind-oracle:
We're using several HA pairs of Prometheus and want to be able to ingest to different tenants from a single HA pair based on labels in a trusted environment. We need this to separate teams/application groups.

Currently out of box this is not supported and I wrote a small proxy that sits between Prometheus & Cortex and does the job (https://github.com/blind-oracle/cortex-tenant). It looks for a predefined label and uses it as a tenant ID.

Do you think it's worth to implement something like this in Cortex directly somewhere in the Distributor?

roidelapluie:
There is a auth gateway in the roadmap

https://cortexmetrics.io/docs/roadmap/

blind-oracle:
@roidelapluie 

Yeah, but from proposed functionality it doesn't look like that it would allow multiple tenants per single Prometheus...
But anyway, if such entity is planned then probably this feature should go there.

pracucci:
There are thoughts about adding relabeling support during ingestion (in distributors). This use case could be a good candidate for a custom relabelling rule (where you can change the tenant ID). @gouthamve thoughts?

blind-oracle:
Any chance on this?

pracucci:
> Any chance on this?

No progress on this yet. I think doing it at a proxy level is still the only (and probably) best way to do it. In a previous message I suggested whether we could use relabelling support but, after more thoughts, I'm not sure it's easy to fit into the distributor.

roidelapluie:
> > Any chance on this?
> 
> No progress on this yet. I think doing it at a proxy level is still the only (and probably) best way to do it. In a previous message I suggested whether we could use relabelling support but, after more thoughts, I'm not sure it's easy to fit into the distributor.

I tried that but many questions rise: which limits to apply, should the metrics be fed twice in the distributor ? (and how to prevent "loops" then), etc..

pracucci:
> I tried that but many questions rise: which limits to apply, should the metrics be fed twice in the distributor ? (and how to prevent "loops" then), etc..

I see the pain and I think it doesn't really fit. The whole Cortex has been designed to receive the tenant in the input request, so the distributor (like other services) have been designed this way. I think doing it at a proxy level is still the right (and best) way to do.

blind-oracle:
Ok, thanks guys, then I guess let's leave it as is then for now and use the proxy 👍

blind-oracle:
[...]

PR #3329, opened 2020-10-12
Support metric_relabel_configs in distributor

### LINKED ISSUE #1507: Consider allowing metric relabelling (or just "drop") during ingest
I think a nice feature would be to provide a way to use metric relabelling in the distributors.

The main use case I want is to be able to turn off certain sources of metrics quickly without having to change the remote write config of the prometheus instances running in each of my clusters.

roidelapluie:

**What this PR does**:
Support metric_relabel_configs in distributor

Fixes #1507 

**Checklist**
- [X] Tests updated
- [X] Documentation added


csmarchbanks (commented):
:partying_face:, I am excited for this!

Do you think it is worthwhile to allow this to be specified per user rather than just globally? I imagine a common use case will be to stop very high cardinality or high churn data from a specific tenant without disrupting data for all tenants.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 * [CHANGE] Increased default `-<prefix>.redis.timeout` from `100ms` to `500ms`. #3301
 * [FEATURE] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-tenant` globally, or using per-tenant limit `max_queriers_per_tenant`), each tenants's requests will be handled by different set of queriers. #3113 #3257
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
+* [FEATURE] Support `metric_relabel_confis` in distributor. #3329
```
drequena, line 52: configs instead of confis (just a simples typo) ;)

pkg/distributor/distributor.go
```diff
@@ -170,6 +171,9 @@ type Config struct {
 	// when true the distributor does not validate the label name, Cortex doesn't directly use
 	// this (and should never use it) but this feature is used by other projects built on top of it
 	SkipLabelNameValidation bool `yaml:"-"`
+
+	// Metrics relabeling
+	MetricRelabelConfigs []*relabel.Config `yaml:"metric_relabel_configs,omitempty" doc:"nocli|description=List of metric relabel configurations. Note that in most situations, it is more effective to use metrics relabeling directly in the Prometheus server, e.g. remote_write.write_relabel_configs."`
```
pstibrany, line 176: We should note that this is global, for ALL tenants.

pstibrany (commented):
Can we move the configuration into [Overrides](https://github.com/cortexproject/cortex/blob/c0d39957f7e509600773d028913ad765698538e6/pkg/util/validation/limits.go)? Then it can be per-user, and reloaded when changed on disk, without restarting Cortex. WDYT?

roidelapluie:
I was already thinking about https://github.com/cortexproject/cortex/issues/3318#issuecomment-706976139 when I set this up. WDYT? Should we have both?

pstibrany:
> I was already thinking about [#3318 (comment)](https://github.com/cortexproject/cortex/issues/3318#issuecomment-706976139) when I set this up. WDYT? Should we have both?

If by "both" you refer to global and per-tenant config, then Overrides / Limits already support that, and yes, I think it makes sense to have both. For example in cloud offering, we wouldn't use global one, but could use per-tenant config in specific cases. On the other hand, some people running Cortex internally may only need global one.

bboreham:
Thanks!  Any idea what the CPU/memory overhead is?

roidelapluie:
That depends on the relabeling rules..

roidelapluie:
I have moved it to limits

pkg/distributor/distributor.go
```diff
@@ -455,6 +456,11 @@ func (d *Distributor) Push(ctx context.Context, req *client.WriteRequest) (*clie
 			removeLabel(labelName, &ts.Labels)
 		}
 
+		if mrc := d.limits.MetricRelabelConfigs(userID); len(mrc) > 0 {
```
jtlisi, line 459: **question**: Is there any instrumentation or debug logging that could be added that would be helpful for operators/users when working with this feature?

jtlisi (approved):
LGTM!

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 * [CHANGE] Increased default `-<prefix>.redis.timeout` from `100ms` to `500ms`. #3301
 * [FEATURE] Added support for shuffle-sharding queriers in the query-frontend. When configured (`-frontend.max-queriers-per-tenant` globally, or using per-tenant limit `max_queriers_per_tenant`), each tenants's requests will be handled by different set of queriers. #3113 #3257
 * [FEATURE] Query-frontend: added `compression` config to support results cache with compression. #3217
+* [FEATURE] Support `metric_relabel_configs` in distributor. #3329
```
jtlisi, line 52:
```suggestion
* [FEATURE] Added the support for applying a Prometheus metrics relabel config on series received by the distributor. The `metric_relabel_configs` field was added to the limits config. #3329
```

pkg/distributor/distributor.go
roidelapluie: distributor_received_samples_total and distributor_samples_in_total covers that.

CHANGELOG.md
roidelapluie: Thanks, I updated my comment

pstibrany:
Thank you, this looks very good! Would it be possible to add small test showing parsing of relabeling YAML rules into per-user limits?

roidelapluie:
> Thank you, this looks very good! Would it be possible to add small test showing parsing of relabeling YAML rules into per-user limits?

Do we have tests like this already?

pstibrany:
> > Thank you, this looks very good! Would it be possible to add small test showing parsing of relabeling YAML rules into per-user limits?
> 
> Do we have tests like this already?

Until now structures we've put into overrides YAML file were pretty basic. I can only see `TestLimitsLoadingFromYaml`. Something similar in size but with relabel config to make sure the wiring is all correct would be good to have. WDYT?

roidelapluie:
Done :)

pracucci (commented):
Could we mark it experimental in `docs/configuration/v1-guarantees.md`, please?

roidelapluie:
done

docs/configuration/config-file-reference.template
```diff
@@ -24,6 +24,7 @@ To specify which configuration file to load, pass the `-config.file` flag at the
 * `<string>`: a regular string
 * `<url>`: an URL
 * `<prefix>`: a CLI flag prefix based on the context (look at the parent configuration block to see which CLI flags prefix should be used)
+* `<relabel_config>`: a [prometheus relabeling configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config).
```
pracucci, line 27:
```suggestion
* `<relabel_config>`: a [Prometheus relabeling configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config).
```

pracucci (approved):
Thanks for working on this and sorry for the super late review (I would like days of 48h 😢 ). Looks flawless to me. I just left a nit, definitely not a blocker.

pstibrany (approved):
Thanks a lot Julien!

docs/configuration/config-file-reference.template
pstibrany: Sorry, I've committed this thinking that I will fix DCO right after, but obviously I cannot push to your branch. 🤦

pstibrany: (Well, except via GitHub UI apparently)

roidelapluie: Yes you can

pstibrany: You're correct. I got confused there. Pushed DCO fix, when CI passes, I'll merge.

bboreham:
I just used this feature in production to remove a high-cardinality metric; worked super well!

Even better if we had a metric to track how many samples got dropped this way - right now they disappear from `cortex_distributor_received_samples_total` and don't appear in `cortex_discarded_samples_total`.

PR #3341, opened 2020-10-13
 Migrate CircleCI workflows to GitHub Actions (2/3)

shovnik:
### What this PR does

This is PR 2/3 of migrating Cortex's CI from CircleCI to GitHub Actions: 

**[Part 1/3](https://github.com/cortexproject/cortex/pull/3302)** 
* create `test-build-deploy` workflow under `.github/workflows`
* add readme
* adding the 3 base jobs
   * `lint`
   * `test`
   * `build`
* these jobs are run async and share no dependencies with each other
* workflow is run when a commit is pushed to `master` or when any PR is opened. However CD jobs will only run on the master branch and will be skipped otherwise
* all jobs in following PRs are dependant on one or more of these jobs finishing

**[Part 2/3](https://github.com/cortexproject/cortex/pull/3341)**  👈 
* adding jobs dependant on `build`
    * `integration`
    * `integration-config-db`
* This PR completes the CI portion of migration. The following PR will finish the remaining CD portion of the migration
* Addresses comments on PR 1/3: removing compression for tar and removing useless artifacts from README

**Part 3/3** 
* This PR completes the migration from CircleCI to GitHub Actions
   * `deploy_website` is dependant on `build, test`
   * `deploy` is dependant on `build, test, lint, integration, integration-config-db`
* CD jobs will only run on the master branch and will be otherwise skipped

cc - @alolita
### Which issue(s) this PR fixes:
fixes a part of https://github.com/cortexproject/cortex/issues/3274

### Checklist
 - [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
+        # Commands in the Makefile are hardcoded with an assumed file structure of the CI container
+        # Symlink ensures paths specified in previous commands don’t break
       - name: Sym Link Expected Path to Workspace
         run: |
           mkdir -p /go/src/github.com/cortexproject/cortex
@@ -80,10 +82,97 @@ jobs:
           mkdir /tmp/images
           ln -s /tmp/images ./docker-images
           make BUILD_IN_CONTAINER=false save-images
-      - name: Zip Images
-        run: tar -zcvf images.tar.gz /tmp/images
-      - name: Upload Images Artifact
+      - name: Create Docker Images Archive
+        run: tar -cvf images.tar.gz /tmp/images
+      - name: Upload Docker Images Artifact
         uses: actions/upload-artifact@v2
         with:
           name: Docker Images
           path: ./images.tar.gz
+
+  integration:
+    needs: build
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Install Docker Client
+        run: |
+          set -x
+          VER="17.03.0-ce"
+          curl -L -o /tmp/docker-$VER.tgz https://download.docker.com/linux/static/stable/x86_64/docker-$VER.tgz
+          tar -xz -C /tmp -f /tmp/docker-$VER.tgz
+          sudo mv /tmp/docker/* /usr/bin
+      - name: Upgrade golang
+        run: |
+          cd /tmp
+          wget https://dl.google.com/go/go1.14.9.linux-amd64.tar.gz
+          tar -zxvf go1.14.9.linux-amd64.tar.gz
+          sudo rm -fr /usr/local/go
+          sudo mv /tmp/go /usr/local/go
+          cd -
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Sym Link Expected Path to Workspace
+        run: |
+          sudo mkdir -p /go/src/github.com/cortexproject/cortex
+          sudo ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
+      - name: Download Docker Images Artifacts
+        uses: actions/download-artifact@v2
+        with:
+          name: Docker Images
+      - name: Create Docker Images Archive
+        run: tar -xvf images.tar.gz -C /
+      - name: Load Docker Images
+        run: |
+          ln -s /tmp/images ./docker-images
+          make BUILD_IN_CONTAINER=false load-images
+      - name: Preload Images
+        # We download docker images used by integration tests so that all images are available
+        # locally and the download time doesn't account in the test execution time, which is subject
+        # to a timeout
+        run: |
+          docker pull minio/minio:RELEASE.2019-12-30T05-45-39Z
+          docker pull amazon/dynamodb-local:1.11.477
+          docker pull consul:0.9
+          docker pull gcr.io/etcd-development/etcd:v3.4.7
+          docker pull quay.io/cortexproject/cortex:v0.6.0
+          docker pull quay.io/cortexproject/cortex:v0.7.0
+          docker pull quay.io/cortexproject/cortex:v1.0.0
+          docker pull quay.io/cortexproject/cortex:v1.1.0
+          docker pull shopify/bigtable-emulator:0.1.0
+          docker pull rinscy/cassandra:3.11.0
+          docker pull memcached:1.6.1
+          docker pull bouncestorage/swift-aio:55ba4331
+      - name: Integration Tests
+        run: |
+          export CORTEX_IMAGE_PREFIX="${IMAGE_PREFIX:-quay.io/cortexproject/}"
+          export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:${TAG:-$(./tools/image-tag)}"
+          export CORTEX_CHECKOUT_DIR="/go/src/github.com/cortexproject/cortex"
+          echo "Running integration tests with image: $CORTEX_IMAGE"
+          go test -tags=requires_docker -timeout 1200s -v -count=1 ./integration/...
+        env:
+          IMAGE_PREFIX: ${{ secrets.IMAGE_PREFIX }}
+          TAG: ${{ github.event.push.tag_name }}
+  integration-configs-db:
+    needs: build
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Install Docker Client
+        run: |
+          set -x
+          VER="17.03.0-ce"
+          curl -L -o /tmp/docker-$VER.tgz https://download.docker.com/linux/static/stable/x86_64/docker-$VER.tgz
+          tar -xz -C /tmp -f /tmp/docker-$VER.tgz
+          sudo mv /tmp/docker/* /usr/bin
+      - name: Download Docker Images Artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: Docker Images
+      - name: Extract Docker Images Archive
+        run: tar -xvf images.tar.gz -C /
+      - name: Run Integration Configs Tests
+        run: |
+          touch build-image/.uptodate
+          MIGRATIONS_DIR=$(pwd)/cmd/cortex/migrations
+            make BUILD_IMAGE=quay.io/cortexproject/build-image:upgrade-build-image-debian-491e60715-WIP TTY='' configs-integration-test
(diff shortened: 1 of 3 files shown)


.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
+          tar -zxvf go1.14.9.linux-amd64.tar.gz
+          sudo rm -fr /usr/local/go
+          sudo mv /tmp/go /usr/local/go
+          cd -
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Sym Link Expected Path to Workspace
+        run: |
+          sudo mkdir -p /go/src/github.com/cortexproject/cortex
+          sudo ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
```
shovnik, line 117: Linking the directory instead of the contents of the folder did not work when we tested it.

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
+          docker pull bouncestorage/swift-aio:55ba4331
+      - name: Integration Tests
+        run: |
+          export CORTEX_IMAGE_PREFIX="${IMAGE_PREFIX:-quay.io/cortexproject/}"
+          export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:${TAG:-$(./tools/image-tag)}"
+          export CORTEX_CHECKOUT_DIR="/go/src/github.com/cortexproject/cortex"
+          echo "Running integration tests with image: $CORTEX_IMAGE"
+          go test -tags=requires_docker -timeout 1200s -v -count=1 ./integration/...
+        env:
+          IMAGE_PREFIX: ${{ secrets.IMAGE_PREFIX }}
```
AzfaarQureshi, line 153: This secret will need to be added in the forked repo if one wants to change the image_prefix

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
           mkdir /tmp/images
           ln -s /tmp/images ./docker-images
           make BUILD_IN_CONTAINER=false save-images
-      - name: Zip Images
-        run: tar -zcvf images.tar.gz /tmp/images
-      - name: Upload Images Artifact
+      - name: Create Docker Images Archive
+        run: tar -cvf images.tar.gz /tmp/images
```
pracucci, line 86: Given we're not compressing anymore, could you remove the `.gz` suffix?

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
         uses: actions/upload-artifact@v2
         with:
           name: Docker Images
           path: ./images.tar.gz
+
+  integration:
+    needs: build
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Install Docker Client
```
pracucci, line 97: This is also done in `build`. Is there any way to build a shared step? Just asking, not a blocker.

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
+        uses: actions/checkout@v2
+      - name: Sym Link Expected Path to Workspace
+        run: |
+          sudo mkdir -p /go/src/github.com/cortexproject/cortex
+          sudo ln -s $GITHUB_WORKSPACE/* /go/src/github.com/cortexproject/cortex
+      - name: Download Docker Images Artifacts
+        uses: actions/download-artifact@v2
+        with:
+          name: Docker Images
+      - name: Create Docker Images Archive
```
pracucci, line 122: I guess you mean "Extract" instead of "Create"

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
+        # We download docker images used by integration tests so that all images are available
+        # locally and the download time doesn't account in the test execution time, which is subject
+        # to a timeout
+        run: |
+          docker pull minio/minio:RELEASE.2019-12-30T05-45-39Z
+          docker pull amazon/dynamodb-local:1.11.477
+          docker pull consul:0.9
+          docker pull gcr.io/etcd-development/etcd:v3.4.7
+          docker pull quay.io/cortexproject/cortex:v0.6.0
+          docker pull quay.io/cortexproject/cortex:v0.7.0
```
pracucci, line 138: Could you remove this, please? Not required anymore.

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
+          docker pull quay.io/cortexproject/cortex:v1.0.0
+          docker pull quay.io/cortexproject/cortex:v1.1.0
```
pracucci, line 140: Could you also add cortex version 1.2.0, 1.3.0, 1.4.0 please? They're used in `integration/backward_compatibility_test.go` (I forgot to add them before)

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
+          docker pull rinscy/cassandra:3.11.0
+          docker pull memcached:1.6.1
+          docker pull bouncestorage/swift-aio:55ba4331
+      - name: Integration Tests
+        run: |
+          export CORTEX_IMAGE_PREFIX="${IMAGE_PREFIX:-quay.io/cortexproject/}"
+          export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:${TAG:-$(./tools/image-tag)}"
+          export CORTEX_CHECKOUT_DIR="/go/src/github.com/cortexproject/cortex"
+          echo "Running integration tests with image: $CORTEX_IMAGE"
+          go test -tags=requires_docker -timeout 1200s -v -count=1 ./integration/...
```
pracucci, line 151: I recently increased this timeout to 1800s. Could you update it, please?

.github/workflows/test-build-deploy.yml
```diff
@@ -80,10 +82,97 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          name: Docker Images
+      - name: Extract Docker Images Archive
+        run: tar -xvf images.tar.gz -C /
+      - name: Run Integration Configs Tests
+        run: |
+          touch build-image/.uptodate
+          MIGRATIONS_DIR=$(pwd)/cmd/cortex/migrations
+            make BUILD_IMAGE=quay.io/cortexproject/build-image:upgrade-build-image-debian-491e60715-WIP TTY='' configs-integration-test
```
pracucci, line 178: Could you add a comment to explain why we need `TTY=''`, please?

pracucci (commented):
Good job on this 2nd part too! 👏  I left few comments and then we should be good to go 🚀

.github/workflows/test-build-deploy.yml
AzfaarQureshi: We could replace this with a YAML anchor (which GitHub Actions has plans to release [shortly](https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/14)) or we could write a [custom action](https://github.com/actions/runner/issues/646) and use that instead. With the custom action we would replace the `Install Docker Client` step with something like this in both Build and Integration.

```yaml
steps:
  - name: Install Docker Client
     uses: cortexproject/install-docker-client
  - run: ...
``` 

This means you'll have to maintain another actions repo and that might bring unwanted toil when upgrading versions.

shovnik: A simple way to have this logic in one place would be to create an install-docker.sh script and call it whenever an image requires a clean docker client installation. This would also make it easier to update the docker version in a single place. Is that close to what you had in mind?

AzfaarQureshi: ^ This is a better idea imo. This way you won't need to update the docker client version manually every time the `install docker client` step is called. The script can simply read `$VER` which can be set as a global variable for the entire workflow at the top of the file.

.github/workflows/test-build-deploy.yml
shovnik: Removed

.github/workflows/test-build-deploy.yml
shovnik: Slight brain fart 😅 , thanks

.github/workflows/test-build-deploy.yml
shovnik: Removed

.github/workflows/test-build-deploy.yml
shovnik: Added

.github/workflows/test-build-deploy.yml
shovnik: Updated

.github/workflows/test-build-deploy.yml
shovnik: I committed the solution using an install-docker script. Let me know if it is okay. @pracucci

.github/workflows/test-build-deploy.yml
AzfaarQureshi: The default runners in GitHub Actions don't support TTY :( https://github.com/actions/runner/issues/241 and a GitHub member responded saying its fairly low prio in their backlog.

This is what happens when we remove `TTY=''`:

Screen Shot 2020-10-15 at 1 14 55 PM

However, there is a workaround we can do if you'd like it back: https://man7.org/linux/man-pages/man1/script.1.html
(script -e -c `$COMMAND_HERE`)
The snippet would be changed to: 

```yaml

   steps:
    - uses: actions/checkout@v1
      run: |
        touch build-image/.uptodate
        MIGRATIONS_DIR=$(pwd)/cmd/cortex/migrations
        script -e -c "make BUILD_IMAGE=quay.io/cortexproject/build-image:upgrade-build-image-debian-491e60715-WIP configs-integration-test"
```

This works and enables the command to run with TTY but this seemed kind of hacky to me so I left it as `TTY=''`

Would you like us to change the line to use `script -e -c` or just add a comment saying GHA doesnt support TTY in their default runners yet

shovnik:
Hey @pracucci , we addressed all the changes requested in the comments on the PR. Is this PR in a state ready to be merged?

pstibrany (approved):
Thanks for your work!

.github/workflows/test-build-deploy.yml
pstibrany: I think comment explaining why TTY='' is fine.

pracucci: > just add a comment saying GHA doesnt support TTY in their default runners yet

Let's just add a comment, but not a block so you can do it in a following PR.

pracucci (approved):
LGTM, thanks!

Issue #3282, opened 2020-10-05
Potential Prometheus memory leak in ingesters


roystchiang:
**Describe the bug**
We encountered an issue with Ingester/Prometheus unable to recover from WAL due to `invalid block sequence: block time ranges overlap` issue. Repeatedly calling this Ingester with ingestion request results in OOM.

When an ingester service starts up, it tries to recover TSDB for all UserIds. If a TSDB failed to be recovered, [ingester will skip it and continue on](https://github.com/cortexproject/cortex/blob/012fc0b0fa2fce785388917959b1a8168d29d809/pkg/ingester/ingester_v2.go#L1103). When an ingestion request for the corrupted TSDB comes, ingester will try to create a TSDB for this UserID. Since there is already a TSDB for this user, this action will result in [Prometheus performing recovery once again](https://github.com/cortexproject/cortex/blob/012fc0b0fa2fce785388917959b1a8168d29d809/pkg/ingester/ingester_v2.go#L404). 

I noticed that for the same request, ingesters log different errors

```
level=warn ts=2020-10-01T19:36:19.800983863Z caller=grpc_logging.go:38 method=/cortex.Ingester/Push duration=28.647562844s err="user=some-user-id: failed to open TSDB: /data/tsdb/some-user-id: invalid block sequence: block time ranges overlap: [mint: 1601488821086, maxt: 1601490567894, range: 29m6s, blocks: 2]: <ulid: 01EKG4DJNVRS9KSPQZHVHKD07G, mint: 1601481600000, maxt: 1601490567894, range: 2h29m27s>, <ulid: 01EKGQHG7FP40NRB4SA0GFAV4E, mint: 1601488821086, maxt: 1601496000000, range: 1h59m38s>" msg="gRPC\n"
```

```
level=warn ts=2020-10-01T19:15:41.737350414Z caller=grpc_logging.go:38 method=/cortex.Ingester/Push duration=13m46.278392928s err="user=some-user-id: failed to open TSDB: /data/tsdb/some-user-id: mmap files, file: /data/tsdb/some-user-id/chunks_head/000036: mmap: cannot allocate memory" msg="gRPC\n
```

I see the same chunk-head is memory-mapped about 200 times. I also saw that an ingester is able to handle about 200 requests for this corrupted TSDB before going OOM. I suspect Promethues is memory-mapping files during WAL replay. When Prometheus fails, this mmap is not cleaned up.

```
/ # sysctl vm.max_map_count
[...]

PR #3344, opened 2020-10-14
Avoid large blocks on forced compaction of blocks ingester

### LINKED ISSUE #3308: Blocks: Compaction (idle/force) should respect configured time range for blocks.
**Describe the bug**
When Cortex ingester is [idle- or force-compacting the block](https://github.com/cortexproject/cortex/blob/e26519995ae889d0f7a187a1e74dde99571fa39d/pkg/ingester/ingester_v2.go#L1219), it uses full time-range of the head as input parameter to compaction. This doesn't respect configured block range, and is especially bad when head contains data spanning several days or even weeks – it would produce huge block. We should fix that.

**Expected behavior**
Compaction (idle/force) should respect configured time range for blocks.

**Storage Engine**
- [x] Blocks
- [ ] Chunks

codesome:
**What this PR does**:

Creates smaller blocks out of Head during idle and forced compaction.

Fixes #3308

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/ingester/ingester_v2.go
```diff
@@ -1216,8 +1217,30 @@ func (i *Ingester) compactionLoop(ctx context.Context) error {
...
+			return err
+		}
+	}
+	return userDB.CompactHead(tsdb.NewRangeHead(h, h.MinTime(), h.MaxTime()))
+}
+
 // Compacts all compactable blocks. Force flag will force compaction even if head is not compactable yet.
 func (i *Ingester) compactBlocks(ctx context.Context, force bool) {
+	i.TSDBState.compactBlocksMtx.Lock()
+	defer i.TSDBState.compactBlocksMtx.Unlock()
```
codesome, line 1242: As the access to the mint and maxt of Head and then compaction is not atomic, we need this protection

CHANGELOG.md
```diff
@@ -92,6 +92,7 @@
 * [BUGFIX] Use a valid grpc header when logging IP addresses. #3307
 * [BUGFIX] Fixed the metric `cortex_prometheus_rule_group_duration_seconds` in the Ruler, it wouldn't report any values. #3310
 * [BUGFIX] Fixed gRPC connections leaking in rulers when rulers sharding is enabled and APIs called. #3314
+* [BUGFIX] Block Storage: Avoid creating large blocks on forced compaction and when TSDB is idle. #3344
```
pracucci, line 95:
```suggestion
* [BUGFIX] Blocks storage: Avoid creating blocks larger than configured block range period on forced compaction and when TSDB is idle. #3344
```

pkg/ingester/ingester_v2.go
```diff
@@ -1216,8 +1217,30 @@ func (i *Ingester) compactionLoop(ctx context.Context) error {
 	return nil
 }
 
+// compactHead compact the Head block at 2h intervals,
+// avoiding a single huge block.
+func compactHead(userDB *userTSDB) error {
```
pracucci, line 1222: What do you think to move it to `userTSDB.compactHead()`?

pkg/ingester/ingester_v2.go
```diff
@@ -1216,8 +1217,30 @@ func (i *Ingester) compactionLoop(ctx context.Context) error {
...
+	// The Compact() will produces multiple 2h blocks if Head has a lot of data.
```
pracucci, line 1223: Do not hardcode "2h" in the comment because it depends on the configured block range period

pkg/ingester/ingester_v2.go
pracucci: I would make it per TSDB, what do you think?

If yes, then we need to wrap the call to `userDB.Compact()` to take the lock there too. Moving in this direction, to make the code more clear, I would then embed `*tsdb.DB` with a name in `type userTSDB struct` so that we explicitly wrap the `tsdb.DB` functions we need to call.

pracucci (commented):
Thanks @codesome ! Good work, LGTM. I left few minor comments, mostly design. I would also be glad if you could add a unit test: we already have `TestIngester_flushing` and, if I'm not missing anything, could be easy to add another test case where you push samples spanning over different days and make sure it generates 2 output blocks.

pkg/ingester/ingester_v2.go
```diff
@@ -1216,8 +1217,30 @@ func (i *Ingester) compactionLoop(ctx context.Context) error {
...
+	if err := userDB.Compact(); err != nil {
+		return err
+	}
+	twentyFourHrs := 24 * time.Hour.Milliseconds()
+	h := userDB.Head()
+	if (h.MinTime()/twentyFourHrs)*twentyFourHrs != (h.MaxTime()/twentyFourHrs)*twentyFourHrs {
```
pracucci, line 1229: Shouldn't we check the configured `TSDBConfig.BlockRanges` and compact the head if mint and maxt don't belong to the same range period?

pkg/ingester/ingester_v2.go
codesome: Makes sense to make it per TSDB

pkg/ingester/ingester_v2.go
codesome: That would be a better one

pkg/ingester/ingester_v2.go
codesome: Taking a look now, and I think we don't need to make `*tsdb.DB` with a name right now. We don't need a wrapper around `Compact()`, but rather `CompactHead`, but the wrapper wont match the original `CompactHead` function signature. So I am thinking it to keep it simple and have `userDB.compactHead()`. WDYT? Embeding `*tsdb.DB` with a name sounds fine too if that would help later.

codesome: Oh I see now, it will be easier to use the lock if we also wrap `Compact()`. I will embed the tsdb.DB as a name in the struct.

pracucci: Embedding `tsdb.DB` with a name may help to avoid misuses, because functions must be explicitly wrapped.

codesome:
PR is ready for another review. I have added a unit test and refactored the code as discussed.

pkg/ingester/ingester_v2.go
```diff
@@ -48,12 +48,13 @@ type Shipper interface {
...
-	activeSeries   *ActiveSeries
-	seriesInMetric *metricCounter
-	limiter        *Limiter
+	db               *tsdb.DB
+	userID           string
+	refCache         *cortex_tsdb.RefCache
+	activeSeries     *ActiveSeries
+	seriesInMetric   *metricCounter
+	limiter          *Limiter
+	compactBlocksMtx sync.Mutex
```
pstibrany, line 57: Typically mutex is on top of the block of variables that it protects. This mutex is used differently though. I would suggest adding a comment about it.

pkg/ingester/ingester_v2.go
```diff
@@ -66,14 +67,63 @@ type userTSDB struct {
...
+	defer u.compactBlocksMtx.Unlock()
+
+	return u.db.Compact()
+}
+
+// compactHead compacts the Head block at specified block durations avoiding a single huge block.
+func (u *userTSDB) compactHead(blockDuration int64) error {
+	u.compactBlocksMtx.Lock()
+	defer u.compactBlocksMtx.Unlock()
+	// The Compact() will produces multiple smaller blocks if Head has a lot of data.
```
pstibrany, line 103: typo: ... produce ...

pkg/ingester/ingester_v2.go
```diff
@@ -66,14 +67,63 @@ type userTSDB struct {
...
+	h := u.Head()
+	if (h.MinTime()/blockDuration)*blockDuration != (h.MaxTime()/blockDuration)*blockDuration {
+		// Remaining data in Head spans across 2 block ranges, so we break it into 2 blocks.
+		// Block maxt is exclusive, so we do a -1 here for maxt.
+		maxt := (h.MaxTime()/blockDuration)*blockDuration - 1
+		if err := u.db.CompactHead(tsdb.NewRangeHead(h, h.MinTime(), maxt)); err != nil {
+			return err
+		}
+	}
+	return u.db.CompactHead(tsdb.NewRangeHead(h, h.MinTime(), h.MaxTime()))
```
pstibrany, line 116: We have two `u.db.CompactHead` calls here, that only differ in max time. Let's use one, using maxt variable.

pstibrany (approved):
LGTM with some nits.

pkg/ingester/ingester_v2.go
pstibrany: Actually... I don't understand *why* we even need this, given that compaction can *only* be done from our own loop, and we never compact the same block concurrently.

pkg/ingester/ingester_v2.go
codesome: Not exactly, we are actually doing 2 compactions here

pstibrany: You're right. I've introduced a bug into your PR 🤦

pstibrany:
Based on offline discussion with @codesome, I've removed compaction mutex, since it's not needed.

Additionally, I've replaced `u.db.Compact()` call with a for-loop, since during head compaction (in `if` block), we could receive more data spanning 2 blocks again, which would again create bigger block in last head compact call. Using for-loop fixes that.

codesome:
> Using for-loop fixes that.

Actually it does not, it's similar to calling a `Compact()`. Once you exit the loop here, you can still get more samples after that.

Additionally, a brute force loop might end up creating unaligned block since it is not aligning them like TSDB does (this brings one more thing that we will need to duplicate other checks that TSDB does and keep them in sync, maintenance burden?).

I propose keeping the `Compact()` call and instead rely on [this](https://github.com/cortexproject/cortex/issues/3282#issuecomment-714285039) for no samples during force flushing.

codesome:
> Actually it does not, it's similar to calling a Compact(). Once you exit the loop here, you can still get more samples after that.

Sorry, I am wrong, I see you are re-using the maxTime and not taking it from the Head. But other checks during a `Compact()` still holds.

pkg/ingester/ingester_v2.go
```diff
@@ -66,14 +66,61 @@ type userTSDB struct {
...
+}
+
+// compactHead compacts the Head block at specified block durations avoiding a single huge block.
+func (u *userTSDB) compactHead(blockDuration int64) error {
+	h := u.Head()
+
+	minTime, maxTime := h.MinTime(), h.MaxTime()
+
+	for (minTime/blockDuration)*blockDuration != (maxTime/blockDuration)*blockDuration {
+		// Data in Head spans across multiple block ranges, so we break it into 2 blocks.
```
codesome, line 102:
```suggestion
		// Data in Head spans across multiple block ranges, so we break it here.
```

codesome (commented):
LGTM! This is more robust against what Peter mentioned above. Ignore my previous comment.

pkg/ingester/ingester_v2_test.go
```diff
@@ -1815,6 +1815,48 @@ func TestIngester_flushing(t *testing.T) {
...
 			},
 		},
+
+		"noBlockSpanningOverADayOnFlush": {
+			setupIngester: func(cfg *Config) {
+				cfg.BlocksStorageConfig.TSDB.FlushBlocksOnShutdown = false
+			},
+
+			action: func(t *testing.T, i *Ingester, m *shipperMock) {
+				// Pushing 5 samples, spanning over 3 days.
```
pracucci, line 1825: > spanning over 3 days

Aren't they over 2 days? However I believe the test would be even stronger if samples would span 3 days and you would expect 3 blocks in output.

pkg/ingester/ingester_v2_test.go
```diff
@@ -1815,6 +1815,48 @@ func TestIngester_flushing(t *testing.T) {
 				m.AssertNumberOfCalls(t, "Sync", 1)
 			},
 		},
+
+		"noBlockSpanningOverADayOnFlush": {
```
pracucci, line 1819: > noBlockSpanningOverADayOnFlush

What does "noBlock" mean here? Also could you write it in plain english, plese?

pkg/ingester/ingester_v2_test.go
```diff
@@ -1815,6 +1815,48 @@ func TestIngester_flushing(t *testing.T) {
...
+				pushSingleSampleAtTime(t, i, 25*time.Hour.Milliseconds())
+				pushSingleSampleAtTime(t, i, 26*time.Hour.Milliseconds())
+
+				// Nothing shipped yet.
+				m.AssertNumberOfCalls(t, "Sync", 0)
+
+				i.FlushHandler(httptest.NewRecorder(), httptest.NewRequest("POST", "/flush", nil))
+
+				// Flush handler only triggers compactions, but doesn't wait for them to finish. Let's wait for a moment, and then verify.
+				time.Sleep(3 * time.Second)
```
pracucci, line 1838: Instead of `time.Sleep()` you could use `util.Poll()` checking the number of times `Sync()` has been called. This would make test faster and avoid flaky test if it takes more than 3s for any reason.

pracucci (approved):
Good job! LGTM, thanks! I just left a couple of minor nits.

pkg/ingester/ingester_v2_test.go
codesome: > However I believe the test would be even stronger if samples would span 3 days and you would expect 3 blocks in output.

codesome: > However I believe the test would be even stronger if samples would span 3 days and you would expect 3 blocks in output.

It will depend on the actual samples, because by default it will try to create 2h blocks and not over a day. So a test of creating smaller blocks + breaking the block when it spans 2 day might be fine. I will update the test and maybe have 3 days itself anyway.

pkg/ingester/ingester_v2_test.go
codesome: I will update it :P

PR #3346, opened 2020-10-14
Avoid deletion of blocks which are not shipped

### LINKED ISSUE #2868: Blocks should never be deleted by ingester's TSDB retention until shipped to the storage
When running Cortex blocks storage, the TSDB retention in the ingester is configured to `-experimental.tsdb.retention-period`. TSDB will then delete any block whose MaxTime is older then that, regardless the block has been successfully shipped to the storage (TSDB has no meaning of the shipper).

There are at least two scenarios where this could happen:

1. Ingester is failing to upload blocks to the storage (eg. auth issue, networking issue, ...)
2. Ingester's TSDB Head compaction is failing (eg. due to an in-memory corruption), ingester is restarted after several hours (more than `retention-period`), WAL gets replayed, blocks compacted but they're immediately deleted by retention, before they're successfully uploaded to the storage

codesome:
This is for blocks storage ingesters

Fixes #2868

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/ingester/ingester_v2.go
```diff
@@ -109,6 +110,38 @@ func (u *userTSDB) PostDeletion(metrics ...labels.Labels) {
...
+
+	if u.shipper == nil {
+		return deletable
+	}
+
+	shipperMeta, err := shipper.ReadMetaFile(u.Dir())
+	if err != nil {
+		// If there is any issue with the shipper, we should be conservative and not delete anything.
+		level.Error(util.Logger).Log("msg", "failed to read shipper meta during deletion of blocks", "user", u.userID, "err", err)
+		return nil
```
codesome, line 128: Should we have a metric for this to alert on? Because if this issue persists, queries could skip some data and the disk can run out of space.

roidelapluie:
Can we invert this and make deleteable `shipperMeta.Uploaded` ?

pkg/ingester/ingester_v2.go
```diff
@@ -109,6 +110,37 @@ func (u *userTSDB) PostDeletion(metrics ...labels.Labels) {
...
+	if err != nil {
+		// If there is any issue with the shipper, we should be conservative and not delete anything.
+		level.Error(util.Logger).Log("msg", "failed to read shipper meta during deletion of blocks", "user", u.userID, "err", err)
+		return map[ulid.ULID]struct{}{}
+	}
+
+Outer:
+	for id := range deletable {
+		for _, shippedID := range shipperMeta.Uploaded {
+			if shippedID == id {
```
pracucci, line 133: I would use `if bytes.Equal(shippedID, id)`

pracucci (commented):
Thanks @codesome for working on this! A couple of things:
1. How difficult would be adding a unit test on this?
2. I was checking in one of our production clusters if `thanos.shipper.json` actually contains all shipped blocks and I've found an ingester with some blocks missing in the shipper meta file (but successfully shipped to the bucket). I'm investigating the root cause, but this could be a blocker for this PR

pracucci:
> 2. I was checking in one of our production clusters if `thanos.shipper.json` actually contains all shipped blocks and I've found an ingester with some blocks missing in the shipper meta file (but successfully shipped to the bucket). I'm investigating the root cause, but this could be a blocker for this PR

This could fix it. https://github.com/thanos-io/thanos/pull/3321

codesome:
> Can we invert this and make deleteable shipperMeta.Uploaded ? Or would that be 'too aggressive'?

Not super sure about how querying is setup, but queriers depend on ingesters for some data for upto some hours and there might be gaps in queries if we aggressively delete blocks like that.

codesome:
> How difficult would be adding a unit test on this?

I will check, might need some mocking of shipper

pracucci:
> > Can we invert this and make deleteable shipperMeta.Uploaded ? Or would that be 'too aggressive'?
> 
> Not super sure about how querying is setup, but queriers depend on ingesters for some data for upto some hours and there might be gaps in queries if we aggressively delete blocks like that.

We should never delete a block until the (configured) retention period is reached. What we're doing in this PR is adding an extra protection to not delete a block until shipped (eg. authentication or networking issues).

pracucci:
> > How difficult would be adding a unit test on this?
> 
> I will check, might need some mocking of shipper

A couple of things:
1. In unit tests we already have a `shipperMock` 
2. If you run the ingester with the local filesystem as backend, you should be able to actually run the real shipper

codesome:
@pracucci I have added a unit test now. I think we have to first get https://github.com/thanos-io/thanos/pull/3321 vendored and then go with this, right?

pracucci:
> @pracucci I have added a unit test now. I think we have to first get [thanos-io/thanos#3321](https://github.com/thanos-io/thanos/pull/3321) vendored and then go with this, right?

@codesome Correct. I opened a PR https://github.com/cortexproject/cortex/pull/3363 for the Thanos upgrade.

pracucci (approved):
LGTM, thanks! But let's wait for https://github.com/cortexproject/cortex/pull/3363 before merging.

pkg/ingester/ingester_v2.go
```diff
@@ -110,6 +112,37 @@ func (u *userTSDB) PostDeletion(metrics ...labels.Labels) {
 	}
 }
 
+// blocksToDelete filters the input blocks and returns the blocks which are safe to be deleted from the ingester.
+func (u *userTSDB) blocksToDelete(blocks []*tsdb.Block) map[ulid.ULID]struct{} {
+	if u.DB == nil {
+		return map[ulid.ULID]struct{}{}
```
pstibrany, line 118: Nit:

```suggestion
		return nil
```

pkg/ingester/ingester_v2.go
```diff
@@ -110,6 +112,37 @@ func (u *userTSDB) PostDeletion(metrics ...labels.Labels) {
...
+		for _, shippedID := range shipperMeta.Uploaded {
+			if bytes.Equal(shippedID[:], id[:]) {
+				continue Outer
+			}
+		}
+		// Not shipped yet.
+		delete(deletable, id)
+	}
+
+	return deletable
```
pstibrany, line 143: Can be simplified to:

```suggestion
	result := map[ulid.ULID]struct{}{}
	for _, shippedID := range shipperMeta.Uploaded {
		if _, ok := deletable[shippedID]; ok {
			result[shippedID] = struct{}{}
		}
	}
	return result
```

pstibrany (approved):
LGTM, left non-blocking nits.

pkg/ingester/ingester_v2.go
pstibrany: (Same below)

codesome: I thought `nil` maps would cause panics, but apparently it does not

pstibrany: `nil` maps are fine for reading, iterating and deleting, but not for writing new elements into them.

pkg/ingester/ingester_v2.go
pstibrany: In our alerts we do alert on disk running out of space. I'd say that + error message is enough.

pracucci (approved):
Thanos upgrade PR has been merged, so we can merge this PR too.

PR #3345, opened 2020-10-14
Update prometheus vendor to master

gouthamve:
Wanted to pull in https://github.com/prometheus/prometheus/pull/8050

Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

PR #3347, opened 2020-10-14
Query the entire ingester range for series.

gouthamve:
With https://github.com/prometheus/prometheus/pull/8050 we shouldn't be
paging in chunks for the series API lookups. This would mean much lower
memory utilisation and is similar to doing a range query for that
period with the same query.

With that optimisation, we should allow series API to query all of the
ingester instead of just head block.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


gouthamve:
Depending on https://github.com/cortexproject/cortex/pull/3345

pracucci (commented):
I'm a bit dubious about always querying `0, math.MaxInt64`. We have clusters with several days retention on ingesters and we would end up querying all ingesters blocks every time, even when the start/end time range is way smaller (think about "Last 1h" dashboards).

What if we try to do something smarter, like:
- If start/end time range is within the retention period, honor start/end time range
- Otherwise, query only the head (we would return series/labels different than what the user expect anyway)

integration/querier_test.go
```diff
@@ -414,6 +414,291 @@ func TestQuerierWithBlocksStorageRunningInSingleBinaryMode(t *testing.T) {
 	}
 }
 
+func TestMetadataQueriesWithBlocksStorage(t *testing.T) {
```
pracucci, line 417: This test has been copied from the above ones. I think you could just add your assertions to the two tests above. You could create a function to assert on metadata queries, and call it from the two tests above.

pracucci (approved):
Good job, thanks. LGTM! Could you add a CHANGELOG entry, please?

_Note: @pstibrany also plan to review it by the end of the day. Do you mind waiting for his review too, please?_

pstibrany:
> _Note: @pstibrany also plan to review it by the end of the day. Do you mind waiting for his review too, please?_

By my initial quick look it looks good, but haven't fully checked the integration test yet. I will do that later today.

integration/querier_test.go
```diff
@@ -410,6 +414,170 @@ func TestQuerierWithBlocksStorageRunningInSingleBinaryMode(t *testing.T) {
...
 				require.NoError(t, cluster.WaitSumMetrics(e2e.Equals(float64((11+2)*seriesReplicationFactor)), "thanos_memcached_operations_total")) // as before + 2 gets
 			}
+
+			// Query metadata.
+			testMetadataQueriesWithBlocksStorage(t, c, series1, series2, series3, blockRangePeriod)
+		})
+	}
+}
+
+func testMetadataQueriesWithBlocksStorage(t *testing.T, c *e2ecortex.Client, series1, series2, series3 []prompb.TimeSeries, blockRangePeriod time.Duration) {
```
pstibrany, line 424: Nit: I've found it bit confusing to express `from` and `to` in terms of series, and would suggest passing three points in time instead, relevant to the test:

- minT of oldest block in ingester
- minT of head
- maxT of head

pstibrany (approved):
LGTM, Thanks!

integration/querier_test.go
pracucci: Reason I passed the timeseries is because we also need labels (eg. `series3[0].Labels`)

Issue #3337, opened 2020-10-13
Unit tests in GitHub actions workflow are not running with -race


pracucci:
**Describe the bug**
Unit tests running in CircleCI are running with `-race`. This is enabled in `tools/test` when `$CIRCLECI` is valued. In PR #3302 we did the work to run unit tests on GitHub actions, but we didn't notice this and GitHub actions test are not running with `-race`.

**Expected behavior**
Unit tests should run with the same exact command between CircleCI and GitHub actions.

pstibrany:
> GitHub actions test are running with `-race`

You mean they are NOT running with `-race`, right? And we want to enable it.

pracucci:
> > GitHub actions test are running with `-race`
> 
> You mean they are NOT running with `-race`, right? And we want to enable it.

Yes, it was a typo. Fixed. Thanks!

AzfaarQureshi:
We can either manually set $CIRCLECI before we call tools/test or we can change the makefile/scripts to replace $CIRCLECI with another default GitHub Actions [env variable](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables)

pracucci:
My take:
1. I would keep GitHub actions workflow clean (no hacks setting CIRCLECI env vars)
2. We should make sure the functionality is preserved in CircleCI too (we'll run both for a while, until GitHub actions migration is completed)

That being said you could add another condition to `tools/test`, to make sure the `-race` and code coverage is enabled when running CI in GitHub actions too.

AzfaarQureshi:
Filed a PR here: https://github.com/cortexproject/cortex/pull/3351

PR #3351, opened 2020-10-14
BugFix: setting -race and code coverage flags for GitHub Actions workflow

AzfaarQureshi:
**What this PR does**:

This PR adds a condition in `tools/test.sh` to add `-race` and code coverage flags when running tests in GitHub Actions. This is accomplished by checking the existence of `$CI` which is a [default env variable](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables) in GitHub Actions and is always set to `true`

Fixes https://github.com/cortexproject/cortex/issues/3337

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

cc: @alolita

### DIFF
--- a/tools/test
+++ b/tools/test
@@ -53,7 +53,7 @@ done
 
 GO_TEST_ARGS=(-tags "${TAGS[@]}" -cpu 4 -timeout $TIMEOUT $VERBOSE)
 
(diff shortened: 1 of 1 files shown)


pracucci (approved):
LGTM, thanks!

gouthamve (approved):

PR #3354, opened 2020-10-15
Fail fast an ingester if unable to load existing TSDBs

pracucci:
**What this PR does**:
Tonight we had an issue in one ingester which had TSDB head chunks corrupted (root cause will be discussed separately). When a similar issue happen, the ingester skips the corrupted TSDB at startup, it joins the ring with `ACTIVE` state and, as soon as receive any write request from the tenant with the corrupted TSDB it will try to reopen the TSDB for *every single write request*. This leads to an undesirable situation, which will soon get the ingester to get killed (due to OOM).

In this PR I'm proposing to fail fast the ingester if unable to load an existing TSDB. It's a loud and clear signal to the Cortex cluster operator and, in my opinion, it's better to fail fast an ingester before it start receiving write requests instead of having it failing few minutes after running.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -19,6 +19,7 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb"
+	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/shipper"
@@ -253,6 +254,9 @@ func NewV2ForFlusher(cfg Config, registerer prometheus.Registerer) (*Ingester, e
 
 func (i *Ingester) startingV2ForFlusher(ctx context.Context) error {
 	if err := i.openExistingTSDB(ctx); err != nil {
+		// Try to rollback and close opened TSDBs before halting the ingester.
+		i.closeAllTSDB()
+
 		return errors.Wrap(err, "opening existing TSDBs")
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -1057,8 +1064,13 @@ func (i *Ingester) openExistingTSDB(ctx context.Context) error {
...
 
-	err := filepath.Walk(i.cfg.BlocksStorageConfig.TSDB.Dir, func(path string, info os.FileInfo, err error) error {
+	// Keep track of all errors that could occur.
+	errs := tsdb_errors.MultiError{}
+	errsMx := sync.Mutex{}
+
+	walkErr := filepath.Walk(i.cfg.BlocksStorageConfig.TSDB.Dir, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
+			level.Warn(util.Logger).Log("msg", "skipped filesystem entry when looking for existing TSDB to open", "path", path, "err", err)
 			return filepath.SkipDir
```
pstibrany, line 1074: If there is any error traversing the tree, shouldn't we return such error?  (Esp. if we're halting ingester if we fail to open TSDB)

pkg/ingester/ingester_v2.go
```diff
@@ -1114,14 +1131,23 @@ func (i *Ingester) openExistingTSDB(ctx context.Context) error {
 		return filepath.SkipDir // Don't descend into directories
 	})
 
+	if walkErr != nil {
```
pstibrany, line 1134: This cannot happen currently, since walkFn will filter out any error. (see other comment)

pstibrany (approved):
LGTM, but please take a look at comment.

pkg/ingester/ingester_v2.go
pracucci: Yes, better to do error out. Done.

pkg/ingester/ingester_v2.go
pracucci: Given I addressed the other comment, why can't happen? The `Walk()` interrupts and returns error as soon as we return error. Errors returned by `Walk()` itself are not filtered again via `Walk()`.

pstibrany: Now it can.

pstibrany (approved):

pkg/ingester/ingester_v2.go
pracucci: Even before it could, in case `os.Open(path)` or `f.Readdirnames(1)` failed.

pstibrany: You are right.

PR #3271, opened 2020-10-02
Update Thanos to latest master.

pstibrany:
**What this PR does**: This PR updates Thanos dependency to [latest master](https://github.com/thanos-io/thanos/commits/bb1662a1d13c). Main reason for update is to get https://github.com/thanos-io/thanos/pull/3261 in.

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3363, opened 2020-10-19
Upgrade Thanos to latest master

pracucci:
**What this PR does**:
This builds upon #3271 and updates to the latest Thanos master to also get https://github.com/thanos-io/thanos/pull/3321 in.

About the Thanos upgrade: I checked the whole `vendor/` diff and, assuming no bug is in `vendor/github.com/thanos-io/thanos/pkg/store/labelpb/label.go`, the rest of the changes should be fine.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/querier/block.go
```diff
@@ -55,15 +56,15 @@ func (bqss *blockQuerierSeriesSet) Next() bool {
 		return false
 	}
 
-	currLabels := bqss.series[bqss.next].Labels
+	currLabels := labelpb.ZLabelsToPromLabels(bqss.series[bqss.next].Labels)
```
pracucci, line 59: I've a doubt. In this Thanos upgrade, `bqss.series[bqss.next].Labels` is a `labelpb.ZLabel`, which means the whole protobuf message `[]byte` will be retained until the query execution completes. Don't we risk to end up utilising more memory than before? In queriers I'm more worried about memory utilisation than memory allocations rate.

Thoughts @pstibrany @bwplotka ?

pstibrany: If I understand this correctly, problem you're suggesting should not be an issue in querier. Previously querier would receive the message and then extract individual strings (making a copy), and (possibly) discard the message. Now we skip the extraction part at the cost of keeping the message in memory longer. As long as we don't retain these labels somewhere, I think this should decrease memory usage. But it's risky, and we better keep an eye on it.

pstibrany (approved):

pkg/querier/block.go
pstibrany: (I think `*storepb.Series` already keep reference to entire original message via `c.Raw.Data`)

pracucci: I think we should better investigate / understand how this behave. If a query execution is slow, we may potentially increase the memory utilisation with this PR's change.

_I'm going to merge this PR to unblock other work based on this, but let's keep an eye on it._

bwplotka: As I mentioned offline. Thanks to ZLabels, it's really up to you. If you don't want to hold memory, just use `labelpb.DeepCopy` function to do copy manually (: But you are right if you hold labels for longer time, it's very likely you want to copy those.

bwplotka: We are still experimenting, so you are right. Might be risky

PR #3364, opened 2020-10-19
Add cross tenant query federation proposal

### LINKED ISSUE #923: Query across multiple users / tenants / instances

simonswine:
Signed-off-by: Christian Simon <simon@swine.de>

**What this PR does**:

Add cross tenant query federation proposal

Proposal to work on a fix for #923

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/proposals/cross-tenant-query-federation.md
+++ b/docs/proposals/cross-tenant-query-federation.md
@@ -1,0 +1,108 @@
+---
+title: "Cross-Tenant Query Federation"
+linkTitle: "Cross-Tenant Query Federation"
+weight: 1
+slug: "cross-tenant-query-federation"
+---
+
+- Author: [Christian Simon](https://github.com/simonswine)
+- Date: October 2020
+- Status: Accepted
+
+## Overview
+
+This document aims to describe how to implement the ability to allow queries to cover data from more than a single Cortex tenant.
+
+## Reasoning
+
+Adopting a tenancy model within an organization with each tenant representing a department comes with the disadvantage that it will prevent queries from spanning multiple departments. This proposal tries to overcome those limitations.
+
+## Alternatives considered
+
+In theory PromQL API clients could be aggregating/correlating query results from multiple tenants. For example Grafana could be used with multiple data sources and a cross tenant query could be achieved through using [Transformations][grafana_transformation].
+
+[grafana_transformation]: https://grafana.com/docs/grafana/latest/panels/transformations/
+
+## Challenges
+
+### Aggregate data without overlaps
+
+#### Challenge
+
+The series in different tenants might have exactly the same labels and hence potentially collide which each other.
+
+#### Proposal
+
+In order to be able to always identify the tenant correctly, queries using multiple tenants should inject a tenant label named `__tenant_id__` and its value containing the tenant ID into the results. A potentially existing label with the same name should be stored in a label with the prefix `original_`.
+
+Label selectors containing the tenant label should behave like any other labels. This can be achieved by selecting the tenants used in a multi tenant query.
+
+### Exposure of feature to the user
+
+#### Challenge
+
+The tenant ID is currently read from the `X-Scope-OrgID` HTTP header. The tenant ID has those [documented limitations][cortex_tenant_id] of values being allowed.
+
+[cortex_tenant_id]: https://cortexmetrics.io/docs/guides/limitations/#tenant-id-naming
+
+#### Proposal
+
+For the query path a user should be able to specify a `X-Scope-OrgID` header with multiple tenant IDs.  Multiple tenant IDs should then be propagating throughout out the system until it reaches the querier. The `Queryable` instance returned by the querier component, is wrapped by a `mergeQueryable`, which will aggregate the results from a `Queryable` per tenant and hence treated by the downstream components as a single tenant query.
+
+To allow such queries to be processed we suggest that an experimental configuration flag `-querier.tenant-federation.enabled` will be added, which is switched off by default. Once enabled the value of the `X-Scope-OrgID` header should be interpreted as `|` separated list of tenant ids. Components which are not expecting multiple tenant ids (e.g. the ingress path) must signal an error if multiple are used.
+
+
+### Implementing Limits, Fairness and Observability for Cross-Tenant queries
+
+#### Challenge
+
+In Cortex the tenant id is used as the primary identifier for those components:
+
+- The limits that apply to a certain query.
+
+- The query-frontend maintains a per tenant query queue to implement fairness.
+
+- Relevant metrics about the query are exposed under a `user=` label.
+
+Having a query spanning multiple tenants, the existing methods are no longer correct.
+
+#### Proposal
+
+The identifier for aforementioned features for queries involving more than a single tenant should be derived from: An ordered, distinct list of tenant IDs, which is joined by a `|`.
+
+While this feature is considered experimental, this provides some insights and ability to limit multi-tenant queries with these short comings:
+
+- Cardinality costs to the possible amount of tenant ID combinations.
+
+- Query limits applied to single tenants part of a multi tenant query are ignored.
+
+
+## Conclusion
+
+| Challenge                                                                | Status                       |
+|--------------------------------------------------------------------------|------------------------------|
+| Aggregate data without overlap                                           | Implementation in PR [#3250] |
+| Exposure of feature to the user                                          | Implementation in PR [#3250] |
+| Implementing Limits, Fairness and Observability for Cross-Tenant queries | Implementation in PR [#3250] |
+
+[#3250]: https://github.com/cortexproject/cortex/pull/3250
+
+### Future work
+
+Those features are currently out of scope for this proposal, but we can foresee some interest implementing them after this proposal.
+
+#### Cross Tenant support for the ruler
+
+Ability to use multi tenant queries in the ruler.
+
+#### Allow the identifier for limits, fairness and observability to be switched out
+
+It would be great if the source identifier could be made more pluggable. This could allow to for example base all of those features on another dimension (e.g. users rather than tenants)
+
+#### Allow customisation of the label used to expose tenant ids
+
+As per this proposal the label name `__tenant_id__` is fixed, but users might want to be able to modify that through a configuration option.
+
+#### Retain overlapping tenant id label values recursively
+
+As per this proposal the tenant label injection retains an existing label value, but this is not implemented recursively. So if the result already contains `__tenant_id__` and `original__tenant_id__` labels, the value of the latter would be lost.
--- a/docs/proposals/cross-tenant-query-federation.md
+++ b/docs/proposals/cross-tenant-query-federation.md
@@ -84,7 +84,7 @@ Having a query spanning multiple tenants, the existing methods are no longer cor
 
 #### Proposal
 
-The identifier for aforementioned features for queries involving more than a single tenant should be derived from: An ordered, distinct list of tenant IDs, which is joined by a `|`.
+The identifier for aforementioned features for queries involving more than a single tenant should be derived from: An ordered, distinct list of tenant IDs, which is joined by a `|`. This will produce a reproducible identifier for the same set of tenants no matter which order they have been specified.
 
 While this feature is considered experimental, this provides some insights and ability to limit multi-tenant queries with these short comings:
 
--- a/docs/proposals/cross-tenant-query-federation.md
+++ b/docs/proposals/cross-tenant-query-federation.md
@@ -19,9 +19,25 @@ Adopting a tenancy model within an organization with each tenant representing a 
 
 ## Alternatives considered
 
+### Aggregation in PromQL API clients
+
 In theory PromQL API clients could be aggregating/correlating query results from multiple tenants. For example Grafana could be used with multiple data sources and a cross tenant query could be achieved through using [Transformations][grafana_transformation].
 
+As this approach comes with the following disadvantages, it was not considered further:
+
+- Every PromQL API client needs to support the aggregation from various sources.
+
+- Queries that are written in PromQL can't be used without extra work across tenants.
+
 [grafana_transformation]: https://grafana.com/docs/grafana/latest/panels/transformations/
+
+### Multi tenant aggregation in the query frontends
+
+Another approach to multi tenant query federation could be achieved by aggregation of partial query results within the query frontend. For this a query needs to be split into sub queries per tenant and afterwards the partial results need reduced into the final result.
+
+The [astmapper] package goes down a similar approach, but it cannot parallelize all query types. Ideally multi-tenant query federation should support the full PromQL language and the algorithms necessary would differ per query functions and operators used. This approach was deemed as a fairly complex way to achieve that tenant query federation.
+
+[astmapper]: https://github.com/cortexproject/cortex/blob/f0c81bb59bf202db820403812e8dabcb64347bfd/pkg/querier/astmapper/parallel.go#L27
 
 ## Challenges
 
@@ -62,7 +78,7 @@ In Cortex the tenant id is used as the primary identifier for those components:
 
 - The query-frontend maintains a per tenant query queue to implement fairness.
 
-- Relevant metrics about the query are exposed under a `user=` label.
+- Relevant metrics about the query are exposed under a `user` label.
 
 Having a query spanning multiple tenants, the existing methods are no longer correct.
 


docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,82 @@
...
+
+#### Proposal
+
+In order to identify the series uniquely, queries using multiple tenants should inject a label with the name `__tenant__` and its value containing the tenant ID into the results. An potentially existing label with the same name should be overwritten.
+
+### Exposure of feature to the user
+
+#### Challenge
+
+The tenant ID is currently read from the `X-Scope-OrgID` HTTP header. There is no limitation on character classes defined, that can be used as part of the tenant ID.
```
jtlisi, line 42: **thought:** There is no limit on character classes explicitly documented in this repository. However, it is unlikely Cortex would work correctly with certain characters such as `/`. This is because `/` is commonly used as a delimiter in our storage backends and having a tenant ID with this character would likely issues in most of our processes. 

The exact restrictions on the tenant id in Cortex would vary based on the storage backends being used.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,82 @@
...
+
+#### Proposal
+
+For the query path a user should be able to specify a `X-Scope-OrgID` per tenant ID and the information should be propagating as a string slice throughout out the system.
+
+To allow such queries to be processed we suggest that an experimental configuration flag `-experimental.querier.tenant-federation` will be added, which is switched off by default.
+
+### Implementing Limits and Observability for Cross-Tenant queries
+
+#### Challenge
```
jtlisi, line 52: **Thought:** An additional challenge would be how we will round robin queries from the query-frontend to the queriers when multiple tenant ids exist in a single query.

Currently, we keep a queue for each tenant in the query frontend and round-robin queries from each queue to keep scheduling fair between tenants.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,82 @@
...
+|           |          |        |
+|           |          |        |
+|           |          |        |
+
+
+### TODO: Future work
+
+#### Cross Tenant support for the ruler
+
+#### Cross Cluster Query Federation
```
jtlisi, line 80: **suggestion:** I think this feature is out of scope for this document
```suggestion
```

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Aggregate data without overlaps
+
+#### Challenge
+
+The series in different tenants might have exactly the same labels and hence potentially collide which each other.
+
+#### Proposal
+
+In order to identify the series uniquely, queries using multiple tenants should inject a label with the name `__tenant__` and its value containing the tenant ID into the results. An potentially existing label with the same name should be overwritten.
```
jtlisi, line 36: 👍 I like the approach of using a reserved label.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Exposure of feature to the user
+
+#### Challenge
+
+The tenant ID is currently read from the `X-Scope-OrgID` HTTP header. There is no limitation on character classes defined, that can be used as part of the tenant ID.
+
+#### Proposal
+
+For the query path a user should be able to specify a `X-Scope-OrgID` per tenant ID and the information should be propagating as a string slice throughout out the system.
```
jtlisi, line 46: **question:** Will the entire set of tenant IDs be propagated throughout the entire query, from frontend to block storage? At any point will the context of a query be reduced to a single tenant? 

The queryable used by the querier comes to mind as an ideal place split these requests on a per-tenant basis.

docs/proposals/cross-tenant-query-federation.md
simonswine: I have added this in the latest commit

docs/proposals/cross-tenant-query-federation.md
simonswine: Thanks for pointing this out, I have clarified this point in the latest commit

docs/proposals/cross-tenant-query-federation.md
csmarchbanks: Could the injected label be configurable? Depending on the organization, I imagine tenants could be used similarly to external_labels where all data for certain environments or products are split into different tenants. Then the label could be much more meaningful to a user than `__tenant__`.

The behavior where potentially existing labels being overwritten is interesting, the exact opposite approach is taken by Prometheus default remote behavior - if the label already exists in the series it is not overwritten by `external_labels`. There is also the behavior of scraping which can add `exported_<original-label-name>`. Personally I would prefer one of those approaches than overwriting the label but am not set strongly.

simonswine: @csmarchbanks yes a very good point. The injected label should be configurable. 

I think the tenant_id label value needs to take precedence over what is stored on the series, as the tenant isolation is IMHO much more important, than preserving the user's label data. (Thanos also seems to do this overwrite without keeping the original value: https://github.com/thanos-io/thanos/blob/4b89eea19e6c477e5d272aab9db6b46afee8d2bb/pkg/store/labelpb/label.go#L242)

I do agree that it might be helpful to preserve it in a similar way that the prefix `exporter_` works and it should be something relatively  straightforward to add to the proposal. What do you think should we do with this:

- Is `exporter_` the right prefix, or should we do something like `overwritten_by_tenant_` or maybe `original_`?

- Do we need to enforce that recursively let's say our prefix is `original_` what are we gonna do in the case when the input contains `__tenant__` and `original___tenant__`?

csmarchbanks: I think `original_` is nice, and would not overlap with `exported_` from scrapes. I guess `original_` could become `original_original_...` or something in the case of multiple matches (though that seems very unlikely).

Considering that this is the display layer we are talking about, not actually accessing the data, I am not sure I am convinced that the tenant info is more important than being able to easily transport queries from say a local Prometheus to Cortex. I could also understand an argument where one would wish that how tenants are structured in Cortex is transparent from a user's perspective.

csmarchbanks: Also, would it be possible to query on `__tenant__`? It could be unexpected to get a label back in a result set that would not be queryable.

simonswine: > Also, would it be possible to query on `__tenant__`? It could be unexpected to get a label back in a result set that would not be queryable.

Yes it will be possible to query, the labels matchers are applied to figure out which tenants to use

> Considering that this is the display layer we are talking about, not actually accessing the data, I am not sure I am convinced that the tenant info is more important than being able to easily transport queries from say a local Prometheus to Cortex. I could also understand an argument where one would wish that how tenants are structured in Cortex is transparent from a user's perspective.

I still think it is very important to display the accurate tenant_id in `__tenant__` above anything else. 

Just a quick example: you have a manager A and two teams B, C each with their own tenant. The boss evaluates the performance of the teams through a multi-tenant query. If the raw data takes precedence team C could manipulate some data to make team B look bad just by using the `__tenant__` label.

csmarchbanks: Your use case is different from mine :) I sure hope I never have to deal with that situation.

Cool, I think `original_` is a good approach then and make sure there is a tenant label for everything. If someone wants they can aggregate it away anyway.

docs/proposals/cross-tenant-query-federation.md
pracucci: We know for sure parts of Cortex don't work with `/` and whitespaces in the tenant ID.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Aggregate data without overlaps
+
+#### Challenge
+
+The series in different tenants might have exactly the same labels and hence potentially collide which each other.
+
+#### Proposal
+
+In order to be able to always identify the tenant correctly, queries using multiple tenants should inject a tenant label and its value containing the tenant ID into the results. An potentially existing label with the same name should be stored in a label with the prefix `original_`. This also should be done recursively until no more conflicts arise.
```
pracucci, line 36: > An potentially existing label with the same name should be stored in a label with the prefix `original_`.

Is this complexity required? In Prometheus, label names starting with `__` are considered internal / reserved. We could just use the label name `__tenant__` and override any existing one by design (but we expect it doesn't exist).

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Exposure of feature to the user
+
+#### Challenge
+
+The tenant ID is currently read from the `X-Scope-OrgID` HTTP header. There is no limitation on character classes defined, that can be used as part of the tenant ID.
+
+#### Proposal
+
+For the query path a user should be able to specify a `X-Scope-OrgID` per tenant ID and the information should be propagating as a string slice throughout out the system until it reaches the querier. The `Queryable` instances in the querier component, are wrapped by a `mergeQueryable`, which will aggregate the results from a `Queryable` per tenant and hence treated by the downstream components as a single tenant query.
```
pracucci, line 50: > For the query path a user should be able to specify a `X-Scope-OrgID` per tenant ID

If you mean specifying the HTTP header multiple times, be aware of the HTTP specifications:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

In particular:
> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

docs/proposals/cross-tenant-query-federation.md
csmarchbanks: You can see the use case for the configurable label/a discussion here: https://github.com/cortexproject/cortex/pull/3364#discussion_r508762843. In the case of making the label configurable the complexity is required.

The use case for a configurable tenant label is something like team, organization, product, etc... could be an obvious tenant separation and much clearer to the end user as to where that metric is coming from. For a MVP only allowing `__tenant__` would be acceptable to me though, I would just like it to be configurable in the future.

docs/proposals/cross-tenant-query-federation.md
simonswine: On the back of this Jacob created https://github.com/cortexproject/cortex/issues/3365, will reference it once we have it documented

docs/proposals/cross-tenant-query-federation.md
simonswine: Very good point. I wasn't aware that the requirement is that strict.

After doing some research, I have found this implementation of splitting a header value without reducing the character set:

https://github.com/golang/gddo/blob/master/httputil/header/header.go#L86

I think that mechanism should work for us too

pracucci: Thanks for sharing. I took a quick look. So looks like it has to be comma separated, but values can (not must) be quoted to be able to use commas within the single value. Correct?

simonswine: Yes that's how I read the BNF as part of the RFC (https://tools.ietf.org/html/rfc2616#section-2.2).

It's further more possible to include `"` by using `\"`.

jtlisi: Instead of using a third party library to handle mutli-part headers, could we just assume each tenant ID header value will be URL encoded using `url.Path(Escape|Unescape)`? This will not affect anyone not currently encoding URL encoding their tenant ID but should accommodate this use case. The only bug this might introduce is if a user currently has a tenant ID that includes a `%` character, it may trigger an error unless they ensure the `%` is URL escaped. However, I can't imagine a `%` character in a tenant ID is something that is very common.

See more here: https://play.golang.org/p/YjS5fWT8K-y

pstibrany: (Joining bike-shedding party) I'd suggest using `base64/` prefix and using base64-encoded comma-separated list of tenants after it (eg. `base64/aGVsbG8sd29ybGQ=` for "hello,world"). If header value doesn't have this prefix, we wouldn't interpret it. This should work since we already don't support `/` in tenants.

I wouldn't recommend using `url.PathEscape` since it has different purpose (namely escaping strings in URL paths as name suggests).

(That said, other changes related to propagating this through Cortex will be more "fun").

simonswine: > (Joining bike-shedding party) I'd suggest using `base64/` prefix and using base64-encoded comma-separated list of tenants after it (eg. `base64/aGVsbG8sd29ybGQ=` for "hello,world"). If header value doesn't have this prefix, we wouldn't interpret it. This should work since we already don't support `/` in tenants.

I think that could be a good way forward. We still need to think about escaping of a comma for example. I would suggest we just use `json.Marshal([]string{"tenant-,",  "tenant-b"})` and then encode it using base64 as suggested.

> (That said, other changes related to propagating this through Cortex will be more "fun").

I think it probably entails having our own implementation of `weaveworks/common/user` and then make it other libraries in `weaveworks/common` extensible in the sense to bring your own implementation of it using an interface.

simonswine: I think that basically both options could work fairly well. I think we should make a decision about this soon, so I can update the proposal and prepare the work that is required to make this happen

A :rocket: ) Use a single header, if the header starts with prefix `base64/`, the remainder of the value is a base64 encoded ([URL variant]) JSON list.

[URL variant]: https://github.com/golang/go/blob/d4c1ad882973e407ff85b977f4ce5b9435451190/src/encoding/base64/base64.go#L103

B :tada:) Support multiple header values and as RFC2616 4.2 demands a comma separated list as discussed above.

I think I am leaning towards option (A :rocket: ) as it will allow us to extend it in the future and en-/decoding will be super straightforward to implement. It comes with the advantage of limited readability of the header itself

/cc opinions/votes welcome @jtlisi @pracucci @tomwilkie @gouthamve

docs/proposals/cross-tenant-query-federation.md
simonswine: Thanks for adding the limitations to the documentation @pracucci

docs/proposals/cross-tenant-query-federation.md
pracucci: I've the feeling we're making it too complicated. Some thoughts:
1. Comma `,` is not an officially supported character for tenant ID (see https://cortexmetrics.io/docs/guides/limitations/#tenant-id-naming).
2. We should try to make it easy to use (think about engineers running a `curl -H "X-ScopeOrgID: ..."` command. Doing the base64 encoding of a json is very annoying on the command line.
3. We could conditionally enable the parsing of `,` in the `X-Scope-OrgID` only when query federation is enabled. This would make sure we don't introduce any breaking change in Cortex tenant ID if we just split the header value on comma.

pracucci: Even better, we could use `|` instead of `,`. The pipe in the tenant ID should be even less common and it's considered a [character to avoid](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html) by S3.

docs/proposals/cross-tenant-query-federation.md
jtlisi: I'm in favor of restricting the label to be `__tenant__`. I think the design around configuring the injected label and recursively prefixing should be move to a **Future Work** section.

docs/proposals/cross-tenant-query-federation.md
jtlisi: >Even better, we could use | instead of ,. The pipe in the tenant ID should be even less common and it's considered a character to avoid by S3 (why S3? because if a character is unsupported by S3 or discouraged by S3, we could assume none is using it in Cortex).

I'm 100% behind this approach.

docs/proposals/cross-tenant-query-federation.md
csmarchbanks: :+1: on moving to future work.

docs/proposals/cross-tenant-query-federation.md
simonswine: Thanks everyone for you input. I agree @pracucci 's proposal is the right sized solution for this. I have updated the proposal and mark this as resolved.

docs/proposals/cross-tenant-query-federation.md
simonswine: I have amended the proposal accordingly. Thanks for your comments.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Aggregate data without overlaps
+
+#### Challenge
+
+The series in different tenants might have exactly the same labels and hence potentially collide which each other.
+
+#### Proposal
+
+In order to be able to always identify the tenant correctly, queries using multiple tenants should inject a tenant label named `__tenant__` and its value containing the tenant ID into the results. A potentially existing label with the same name should be stored in a label with the prefix `original_`.
```
pracucci, line 36: > queries using multiple tenants should inject a tenant label named `__tenant__`

In the blocks storage we already have this:

```
TenantIDExternalLabel = "__org_id__"
```

It's an unfortunate label name we picked before deciding to standardise on "tenant ID", but I would try to aim for consistency. We should be able to move the blocks storage to the new label name, but I would suggest to use `__tenant_id__` instead of `__tenant__`. The naming `__tenant_id__` is more flexible because leaves the `__tenant` prefix open for future additional internal labels (eg. `__tenant_<something else>_`).

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+The tenant ID is currently read from the `X-Scope-OrgID` HTTP header. The tenant ID has those [documented limitations][cortex_tenant_id] of values being allowed.
+
+[cortex_tenant_id]: https://cortexmetrics.io/docs/guides/limitations/#tenant-id-naming
+
+#### Proposal
+
+For the query path a user should be able to specify a `X-Scope-OrgID` header with multiple tenant IDs.  Multiple tenant IDs should then be propagating throughout out the system until it reaches the querier. The `Queryable` instance returned by the querier component, is wrapped by a `mergeQueryable`, which will aggregate the results from a `Queryable` per tenant and hence treated by the downstream components as a single tenant query.
+
+To allow such queries to be processed we suggest that an experimental configuration flag `-experimental.querier.tenant-federation.enabled` will be added, which is switched off by default. Once enabled the value of the `X-Scope-OrgID` header should be interpreted as `|` separated list of tenant ids. Components which are not expecting multiple tenant ids (e.g. the ingress path) must signal an error if multiple are used.
```
pracucci, line 52: Please don't use the ` --experimental` prefix. We're not going to do it anymore, because it's a pain to remove it once the feature is stable. Experimental features are just listed to `docs/configuration/v1-guarantees.md`, so the CLI flag name can just be `-querier.tenant-federation.enabled`.

pracucci (approved):
LGTM, with few nits

docs/proposals/cross-tenant-query-federation.md
simonswine: Makes sense. Added this change.

docs/proposals/cross-tenant-query-federation.md
simonswine: Good point and :crossed_fingers: we get it all consistent one day

pracucci: Once you introduce `__tenant_id__` I will take care of migrating to it in the blocks storage too.

pracucci:
@pstibrany @jtlisi Could you take a final look? I would be glad to merge it soon 🙏

csmarchbanks (approved):
:+1:

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+The tenant ID is currently read from the `X-Scope-OrgID` HTTP header. The tenant ID has those [documented limitations][cortex_tenant_id] of values being allowed.
+
+[cortex_tenant_id]: https://cortexmetrics.io/docs/guides/limitations/#tenant-id-naming
+
+#### Proposal
+
+For the query path a user should be able to specify a `X-Scope-OrgID` header with multiple tenant IDs.  Multiple tenant IDs should then be propagating throughout out the system until it reaches the querier. The `Queryable` instance returned by the querier component, is wrapped by a `mergeQueryable`, which will aggregate the results from a `Queryable` per tenant and hence treated by the downstream components as a single tenant query.
+
+To allow such queries to be processed we suggest that an experimental configuration flag `-querier.tenant-federation.enabled` will be added, which is switched off by default. Once enabled the value of the `X-Scope-OrgID` header should be interpreted as `|` separated list of tenant ids. Components which are not expecting multiple tenant ids (e.g. the ingress path) must signal an error if multiple are used.
```
pstibrany, line 52: I would expect that components which are not expecting multiple tenant IDs aren’t affected by this change.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Overview
+
+This document aims to describe how to implement the ability to allow queries to cover data from more than a single Cortex tenant.
+
+## Reasoning
+
+Adopting a tenancy model within an organization with each tenant representing a department comes with the disadvantage that it will prevent queries from spanning multiple departments. This proposal tries to overcome those limitations.
+
+## Alternatives considered
```
pstibrany, line 20: Another alternative is handling this entirely in query-frontend or another component in front of it. Have you considered that?

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
```
pstibrany, line 21: What are pros/cons of such solution?

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+#### Challenge
+
+The series in different tenants might have exactly the same labels and hence potentially collide which each other.
+
+#### Proposal
+
+In order to be able to always identify the tenant correctly, queries using multiple tenants should inject a tenant label named `__tenant_id__` and its value containing the tenant ID into the results. A potentially existing label with the same name should be stored in a label with the prefix `original_`.
+
+Label selectors containing the tenant label should behave like any other labels. This can be achieved by selecting the tenants used in a multi tenant query.
```
pstibrany, line 38: What about `__tenant_id__!~”a.*”`, how is that going to work?

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Challenge
+
+In Cortex the tenant id is used as the primary identifier for those components:
+
+- The limits that apply to a certain query.
+
+- The query-frontend maintains a per tenant query queue to implement fairness.
+
+- Relevant metrics about the query are exposed under a `user=` label.
```
pstibrany, line 65: (super nitty nit: `=` is not part of label)

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+Having a query spanning multiple tenants, the existing methods are no longer correct.
```
pstibrany, line 67: That depends on selected implementation.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Proposal
+
+The identifier for aforementioned features for queries involving more than a single tenant should be derived from: An ordered, distinct list of tenant IDs, which is joined by a `|`.
```
pstibrany, line 71: Why is ordering important?

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,108 @@
...
+
+[#3250]: https://github.com/cortexproject/cortex/pull/3250
+
+### Future work
+
+Those features are currently out of scope for this proposal, but we can foresee some interest implementing them after this proposal.
+
+#### Cross Tenant support for the ruler
+
+Ability to use multi tenant queries in the ruler.
```
pstibrany, line 96: Rule groups are user specific. How do you imagine for cross-tenant rules to work?

pstibrany (commented):
This is an interesting idea, but I have my doubts about the complexity and invasiveness introduced by proposed changes. I would prefer to explore the idea of doing cross-tenant querying in some query-frontend-like component.

simonswine (commented):
Thanks @pstibrany for your comments, I have just added some comments and added some details to the alternatives considered. Let me know what you think. Ideally I would like to get a consensus on the approach fairly soon.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Challenge
+
+The tenant ID is currently read from the `X-Scope-OrgID` HTTP header. The tenant ID has those [documented limitations][cortex_tenant_id] of values being allowed.
+
+[cortex_tenant_id]: https://cortexmetrics.io/docs/guides/limitations/#tenant-id-naming
+
+#### Proposal
+
+For the query path a user should be able to specify a `X-Scope-OrgID` header with multiple tenant IDs.  Multiple tenant IDs should then be propagating throughout out the system until it reaches the querier. The `Queryable` instance returned by the querier component, is wrapped by a `mergeQueryable`, which will aggregate the results from a `Queryable` per tenant and hence treated by the downstream components as a single tenant query.
```
pstibrany, line 66: I am wondering if this is really the way to go, as it prohibits some scenarios.

I would suggest a different approach... define tenant grouping in a Cortex configuration. What I mean by that is that Cortex would have artificial tenant, which would actually map into group of tenants that are really queried.

This could be an overrideable per-tenant setting. By default, tenant would only map to itself, but it would be possible to say that tenant A is really querying data from “B,C,D” or even regex like “A-.*”.

This solves problem with having single user in the X-Scope-OrgID, and allows for more interesting scenarios. Eg. having “global” tenant that simply maps to “.*” (not expanding this recursively).

It also makes authentication easier — it’s not clear to me how authentication would work with multiple values in X-Scope-OrgID header, but it’s clear how it works for single value.

Is it too late to consider this alternative?

docs/proposals/cross-tenant-query-federation.md
pstibrany: Yes, that works, but feels very limited. I would like to be able to expand list of tenant dynamically, based on tenants in the system. See my other comment about possible alternative.

docs/proposals/cross-tenant-query-federation.md
pstibrany: By using “virtual” tenants that map to “real” tenants, we can have new limits / metrics / fairness defined for “virtual” tenant instead, and many existing methods would actually work as expected.

docs/proposals/cross-tenant-query-federation.md
pstibrany: Fair enough, my bad.

pstibrany: > I can see that eventually a concept for a User identity across Cortex, with a User having access to multiple tenants

Funnily enough, this is basically what I suggested in another comment.

docs/proposals/cross-tenant-query-federation.md
pstibrany: Still not quite sure why “reproducible string identifier” is considered important in this context. It’s already human readable (no base64 or anything else), and for querying it doesn’t seem to matter.

docs/proposals/cross-tenant-query-federation.md
pstibrany: Thanks. I now better understand why this wouldn’t work as easily as I thought.

docs/proposals/cross-tenant-query-federation.md
simonswine: > I would suggest a different approach... define tenant grouping in a Cortex configuration. What I mean by that is that Cortex would have artificial tenant, which would actually map into group of tenants that are really queried.

At some point I have considered using exactly that approach defining "virtual tenants". I was just a bit concerned by those factors:
- What happens if a virtual tenant uses paths were we don't have multi tenant support. (e.g. the read part) 
- Should we allow recursive virtual tenants?
- Changing of tenant mappings would require cache invalidations

> This could be an overrideable per-tenant setting. By default, tenant would only map to itself, but it would be possible to say that tenant A is really querying data from “B,C,D” or even regex like “A-.*”.

Are we having the ability to list all existing tenants easily or does that only work for tenants recently queried/ingested to?

> It also makes authentication easier — it’s not clear to me how authentication would work with multiple values in X-Scope-OrgID header, but it’s clear how it works for single value.

I agree that it makes the auth proxy simpler, but I think pretty much the same logic will be part of cortex.

> Is it too late to consider this alternative?

No, I think it's better we have consensus here. 

@pstibrany not too sure how busy you are today, but maybe we can schedule a short 1-1 meeting before the next weekly and talk through it?

pstibrany: > What happens if a virtual tenant uses paths were we don't have multi tenant support. (e.g. the read part)

On read path, we would have support. On write path, we don't expect any data for such user. We could add a check to disallow pushing data with mapping enabled.

> Should we allow recursive virtual tenants?

I think that would make it too complicated.

> Changing of tenant mappings would require cache invalidations

Very good point! We can solve it by storing the actual used mapping (=list of tenants) into cache and so that we can verify that cached data is still correct.

> Are we having the ability to list all existing tenants easily or does that only work for tenants recently queried/ingested to?

When using blocks, we do. When using chunks, we don't (= no regex based mapping, but explicit listing is still possible).

> I agree that it makes the auth proxy simpler, but I think pretty much the same logic will be part of cortex.

Cortex doesn't do anything other than checking that X-Scope-OrgId is not empty, and existing proxies / gateways can treat virtual tenants as normal users.

pstibrany: After internal discussion, the main difference between this proposal and my suggestion is about where the mapping is stored (externally in case of this proposal, or within Cortex in my suggestion).

I think implementing this proposal will be a good step forward, and if we ever want to have mapping stored within Cortex, it should be easy to add in the future.

pstibrany (approved):
Thank you for your patience and discussion. I understand now that my worries about big changes touching all components are not founded anymore, as adopted solution is only going to use single value in x-scope-orgid and context. Also this proposal should be easy to extend in the future, if we ever want to have mapping to target tenants within the Cortex.

docs/proposals/cross-tenant-query-federation.md
```diff
@@ -0,0 +1,124 @@
+---
+title: "Cross-Tenant Query Federation"
+linkTitle: "Cross-Tenant Query Federation"
+weight: 1
+slug: "cross-tenant-query-federation"
+---
+
+- Author: [Christian Simon](https://github.com/simonswine)
+- Date: October 2020
+- Status: Proposed
```
jtlisi, line 10:
```suggestion
- Status: Accepted
```

😉

jtlisi (approved):
LGTM, great work

PR #3366, opened 2020-10-19
Per tenant ruler rule group limits

gotjosh:
**What this PR does**:

Limits the number of rules per rule group, total rule groups and rules a particular tenant can create. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -472,7 +472,7 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if err := a.ruler.AssertMaxRulesPerRuleGroupPerUser(userID, len(rg.Rules)); err != nil {
+	if err := a.ruler.AssertMaxRulesPerRuleGroup(userID, len(rg.Rules)); err != nil {
 		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
@@ -485,7 +485,7 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if err := a.ruler.AssertMaxRuleGroupsPerUser(userID, len(rgs)); err != nil {
+	if err := a.ruler.AssertMaxRuleGroups(userID, len(rgs)); err != nil {
 		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -480,7 +480,7 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 
 	rgs, err := a.store.ListRuleGroupsForUserAndNamespace(req.Context(), userID, "")
 	if err != nil {
-		level.Error(logger).Log("msg", "unable to fetch current rule groups for validation", "err", err.Error())
+		level.Error(logger).Log("msg", "unable to fetch current rule groups for validation", "err", err.Error(), "user", userID)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -472,6 +472,25 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	if err := a.ruler.AssertMaxRulesPerRuleGroupPerUser(userID, len(rg.Rules)); err != nil {
+		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	rgl, err := a.store.ListRuleGroupsForUserAndNamespace(req.Context(), userID, "")
+	if err != nil {
+		level.Error(logger).Log("msg", "unable to fetch current rule groups for validation", "err", err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if err := a.ruler.AssertMaxRuleGroupsPerUser(userID, len(rgl)); err != nil {
+		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
 	rgProto := store.ToProto(userID, namespace, rg)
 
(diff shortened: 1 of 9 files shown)


pstibrany (commented):
Pretty straightforward, LGTM, with some nitpicks: we use "tenant" terminology in limits recently instead of  "user". Also cli flags and YAML options are not consistent (should be the same word by word except `_` / `-`).

pkg/ruler/api.go
```diff
@@ -472,6 +472,25 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
...
 
+	if err := a.ruler.AssertMaxRulesPerRuleGroupPerUser(userID, len(rg.Rules)); err != nil {
+		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	rgs, err := a.store.ListRuleGroupsForUserAndNamespace(req.Context(), userID, "")
+	if err != nil {
+		level.Error(logger).Log("msg", "unable to fetch current rule groups for validation", "err", err.Error())
```
pracucci, line 483: I would add the user here too

```suggestion
		level.Error(logger).Log("msg", "unable to fetch current rule groups for validation", "err", err.Error(), "user", userID)
```

pkg/ruler/ruler.go
```diff
@@ -56,6 +56,10 @@ const (
 	rulerSyncReasonInitial    = "initial"
 	rulerSyncReasonPeriodic   = "periodic"
 	rulerSyncReasonRingChange = "ring-change"
+
+	// Limit errors
+	errMaxRuleGroupsPerUserLimitExceeded        = "per-user rule groups limit (limit %d actual: %d) exceeded"
```
pracucci, line 61:
```suggestion
	errMaxRuleGroupsPerUserLimitExceeded        = "per-user rule groups limit (limit: %d actual: %d) exceeded"
```

pkg/ruler/ruler.go
```diff
@@ -736,3 +740,26 @@ func (r *Ruler) Rules(ctx context.Context, in *RulesRequest) (*RulesResponse, er
 
 	return &RulesResponse{Groups: groupDescs}, nil
 }
+
+// AssertMaxRuleGroupsPerUser limit has not been reached compared to the current
+// number of total rule groups in input and returns an error if so.
+func (r *Ruler) AssertMaxRuleGroupsPerUser(userID string, rg int) error {
+	limit := r.limits.RulerMaxRuleGroupsPerUser(userID)
```
pracucci, line 747: What do you think to skip checking the limit if it's `<= 0`? We're used to use `0` to disable a limit. I would also mention it in the limit CLI flag description (eg. `0 disables it`).

_Same comment applies to `AssertMaxRulesPerRuleGroupPerUser`_

pkg/ruler/ruler.go
```diff
@@ -736,3 +740,26 @@ func (r *Ruler) Rules(ctx context.Context, in *RulesRequest) (*RulesResponse, er
...
+
+	if rg < limit {
```
pracucci, line 749: Hitting the limit is still OK, so:

```suggestion
	if rg <= limit {
```

_Same comments applies to `AssertMaxRulesPerRuleGroupPerUser()` too_

pracucci (commented):
Good job! Left few minor comments and then LGTM.

pkg/ruler/ruler.go
gotjosh: This is just as we're about to create a new one, e.g. In the case where limit is 3 and we have 2, the above would be true and would create the 3rd.

pkg/util/validation/limits.go
```diff
@@ -71,8 +71,10 @@ type Limits struct {
 	MaxQueriersPerTenant int           `yaml:"max_queriers_per_tenant"`
 
 	// Ruler defaults and limits.
-	RulerEvaluationDelay time.Duration `yaml:"ruler_evaluation_delay_duration"`
-	RulerTenantShardSize int           `yaml:"ruler_tenant_shard_size"`
+	RulerEvaluationDelay             time.Duration `yaml:"ruler_evaluation_delay_duration"`
+	RulerTenantShardSize             int           `yaml:"ruler_tenant_shard_size"`
+	RulerMaxRulesPerRuleGroupPerUser int           `yaml:"ruler_max_rules_per_rule_group_per_user"`
```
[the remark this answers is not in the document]
gotjosh: Good shout.

pkg/ruler/ruler.go
gotjosh: Good shout! Done.

pkg/util/validation/limits.go
```diff
@@ -71,8 +71,10 @@ type Limits struct {
 	MaxQueriersPerTenant int           `yaml:"max_queriers_per_tenant"`
 
 	// Ruler defaults and limits.
-	RulerEvaluationDelay time.Duration `yaml:"ruler_evaluation_delay_duration"`
-	RulerTenantShardSize int           `yaml:"ruler_tenant_shard_size"`
+	RulerEvaluationDelay        time.Duration `yaml:"ruler_evaluation_delay_duration"`
+	RulerTenantShardSize        int           `yaml:"ruler_tenant_shard_size"`
+	RulerMaxRulesPerRuleGroup   int           `yaml:"ruler_max_rules_per_rule_group"`
+	RulerMaxRuleGroupsPerTenant int           `yaml:"ruler_max_rule_groups"`
```
pracucci, line 77:
```suggestion
	RulerMaxRuleGroupsPerTenant int           `yaml:"ruler_max_rule_groups_per_tenant"`
```

pracucci (approved):
Thanks for patiently address my comments!

pkg/util/validation/limits.go
```diff
@@ -124,6 +126,8 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 
 	f.DurationVar(&l.RulerEvaluationDelay, "ruler.evaluation-delay-duration", 0, "Duration to delay the evaluation of rules to ensure the underlying metrics have been pushed to Cortex.")
 	f.IntVar(&l.RulerTenantShardSize, "ruler.tenant-shard-size", 0, "The default tenant's shard size when the shuffle-sharding strategy is used by ruler. When this setting is specified in the per-tenant overrides, a value of 0 disables shuffle sharding for the tenant.")
+	f.IntVar(&l.RulerMaxRulesPerRuleGroup, "ruler.max-rules-per-rule-group", 15, "Maximum number of rules per rule group per-tenant. 0 to disable.")
```
jtlisi, line 129: **suggestion:** I think we should consider setting 0 as the default value for both new options. Otherwise this is a breaking change for some users. If we leave the default as is at least consider mentioning the new default value in the changelog.

jtlisi (approved):
One minor suggestion.

Otherwise, LGTM

pkg/util/validation/limits.go
gotjosh: Great shout jacob - I've disabled them.

PR #3369, opened 2020-10-20
Fixed max global series per user/metric limit when shuffle sharding is enabled

### LINKED ISSUE #3367: Global limits are not applied correctly when using shuffle sharding
**Description**

Global limits, at least for ingesters and maybe for other components like querier, are not being calculated correctly when 

For example here:
pkg/ingester/limiter.go : https://github.com/cortexproject/cortex/blob/0211d278730677ba5516c1252566be69357ed4d9/pkg/ingester/limiter.go#L202

The formula seems to assume a tenant’s series are evenly distributed across all ingesters, but if subring/shuffle sharding is enabled, the tenant’s series end up in only a subset of ingesters.

I think to be correct, this needs to use the default and per-tenant setting, i.e. IngestionTenantShardSize

https://github.com/cortexproject/cortex/blob/f3589368be28605fe0089d94b510ec0509b51350/pkg/util/validation/limits.go#L48
pkg/util/validation/limits.go:48

**To Reproduce**

1. run cortex with shuffle sharding on the ingestion path enabled for ingesters
2. set max_global_series_per_metric and/or max_global_series_per_metric values
3. attempt to ingest up to max_global_series_per_metric or max_global_series_per_metric values.

**Expected behavior**
You can ingest up to global limit / actual number of ingesters in user subring size.
Actual behavior is global number of ingesters total is used as denominator, so limit is lower than expected.

**Environment:**
 - Infrastructure: [EKS]
 - Deployment tool: [helm]

**Storage Engine**
- [x] Blocks
- [ ] Chunks

**Additional Context**

pracucci:
**What this PR does**:
The max global series per user/metric limit doesn't work correctly when shuffle sharding and shard-by-all-labels are both enabled because it considers series to be distributed across all ingesters, while actually series are only distributed across a subset of ingesters (the shard size).

Fixes #3367

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/configuration/arguments.md
+++ b/docs/configuration/arguments.md
@@ -489,7 +489,7 @@ Valid per-tenant limits are (with their corresponding flags for default values):
 
    Like `max_series_per_user` and `max_series_per_metric`, but the limit is enforced across the cluster. Each ingester is configured with a local limit based on the replication factor, the `-distributor.shard-by-all-labels` setting and the current number of healthy ingesters, and is kept updated whenever the number of ingesters change.
 
-   Requires `-distributor.replication-factor`, `-distributor.shard-by-all-labels` and `-distributor.sharding-strategy` set for the ingesters too.
+   Requires `-distributor.replication-factor`, `-distributor.shard-by-all-labels`, `-distributor.sharding-strategy` and `-distributor.zone-awareness-enabled` set for the ingesters too.
 
 - `max_series_per_query` / `-ingester.max-series-per-query`
 - `max_samples_per_query` / `-ingester.max-samples-per-query`
@@ -506,7 +506,7 @@ Valid per-tenant limits are (with their corresponding flags for default values):
 
    Like `max_metadata_per_user` and `max_metadata_per_metric`, but the limit is enforced across the cluster. Each ingester is configured with a local limit based on the replication factor, the `-distributor.shard-by-all-labels` setting and the current number of healthy ingesters, and is kept updated whenever the number of ingesters change.
 
-   Requires `-distributor.replication-factor`, `-distributor.shard-by-all-labels` and `-distributor.sharding-strategy` set for the ingesters too.
+   Requires `-distributor.replication-factor`, `-distributor.shard-by-all-labels`, `-distributor.sharding-strategy` and `-distributor.zone-awareness-enabled` set for the ingesters too.
 
 ## Storage
 
--- a/docs/configuration/arguments.md
+++ b/docs/configuration/arguments.md
@@ -489,12 +489,24 @@ Valid per-tenant limits are (with their corresponding flags for default values):
 
    Like `max_series_per_user` and `max_series_per_metric`, but the limit is enforced across the cluster. Each ingester is configured with a local limit based on the replication factor, the `-distributor.shard-by-all-labels` setting and the current number of healthy ingesters, and is kept updated whenever the number of ingesters change.
 
-   Requires `-distributor.replication-factor` and `-distributor.shard-by-all-labels` set for the ingesters too.
+   Requires `-distributor.replication-factor`, `-distributor.shard-by-all-labels` and `-distributor.sharding-strategy` set for the ingesters too.
 
 - `max_series_per_query` / `-ingester.max-series-per-query`
 - `max_samples_per_query` / `-ingester.max-samples-per-query`
 
   Limits on the number of timeseries and samples returns by a single ingester during a query.
+
+- `max_metadata_per_user` / `-ingester.max-metadata-per-user`
+- `max_metadata_per_metric` / `-ingester.max-metadata-per-metric`
+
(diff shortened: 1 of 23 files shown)


pracucci:
_I found a general issue with integration tests (which I believe is the reason why they're failing) I'm going to address in a separate PR and, once merged, I will rebase `master` in this PR._

docs/configuration/arguments.md
```diff
@@ -489,13 +489,25 @@ Valid per-tenant limits are (with their corresponding flags for default values):
...
 
-   Requires `-distributor.replication-factor` and `-distributor.shard-by-all-labels` set for the ingesters too.
+   Requires `-distributor.replication-factor`, `-distributor.shard-by-all-labels`, `-distributor.sharding-strategy` and `-distributor.zone-awareness-enabled` set for the ingesters too.
 
 - `max_series_per_query` / `-ingester.max-series-per-query`
 - `max_samples_per_query` / `-ingester.max-samples-per-query`
 
   Limits on the number of timeseries and samples returns by a single ingester during a query.
 
+- `max_metadata_per_user` / `-ingester.max-metadata-per-user`
```
pracucci, line 499: Metadata limits were not documented, do I've added it.

pkg/util/shard.go
```diff
@@ -3,6 +3,13 @@ package util
 import (
 	"crypto/md5"
 	"encoding/binary"
+	"math"
+)
+
+const (
+	// Sharding strategies.
+	ShardingStrategyDefault = "default"
+	ShardingStrategyShuffle = "shuffle-sharding"
```
pracucci, line 12: I moved constants here to be able to reference them in ingester package without having to import distributor. I took the opportunity to use these constants in the store-gateway too.

pkg/ring/lifecycler_test.go
```diff
@@ -17,18 +18,6 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/test"
 )
 
-type flushTransferer struct {
-	lifecycler *Lifecycler
-}
-
-func (f *flushTransferer) Flush() {}
-func (f *flushTransferer) TransferOut(ctx context.Context) error {
```
pracucci, line 25: Unless I'm missing anything, the logic of this `flushTransferer` was not used, so I've removed it in favour of the tiny `nopFlushTransferer`

pkg/ingester/limiter.go
```diff
@@ -198,11 +211,34 @@ func (l *Limiter) convertGlobalToLocalLimit(globalLimit int) int {
 
 	// May happen because the number of ingesters is asynchronously updated.
 	// If happens, we just temporarily ignore the global limit.
-	if numIngesters > 0 {
-		return int((float64(globalLimit) / float64(numIngesters)) * float64(l.replicationFactor))
+	if numIngesters == 0 {
+		return 0
```
ranton256, line 215: So now local limit is 0 if num healthy ingesters is zero, right? I think that's probably okay, but the comment above is confusing now, since that's not what "ignore the global limit" makes me think.

ranton256 (commented):
LGTM, thanks for the fast turnaround.

pkg/ingester/limiter.go
pstibrany: 0 means unlimited.

pkg/util/shard.go
pstibrany: Same constants are used by ruler now.

pkg/ingester/limiter_test.go
```diff
@@ -17,65 +18,124 @@ func TestSeriesLimit_maxSeriesPerMetric(t *testing.T) {
 		maxLocalSeriesPerMetric  int
 		maxGlobalSeriesPerMetric int
 		ringReplicationFactor    int
+		ringZoneAwarenessEnabled bool
```
pstibrany, line 21: Would it be possible to refactor limiter code to avoid the same repeated tests? (In a follow-up PR, if at all).

pstibrany (approved):
LGTM, good job!

pkg/ingester/limiter.go
ranton256: I see, thanks.

ranton256 (approved):
Thanks for the fix, and also for updating the docs for max_metadata_per_user and max_metadata_per_metric.

pkg/util/shard.go
pracucci: Good point, done!

pkg/ingester/limiter.go
pracucci: Let me also outline the logic hasn't changed. If the ring client doesn't see any ingester (0), the global rate limit was temporarily disabled even before.

pkg/ingester/limiter_test.go
pracucci: I think could be possible. Added to my TODO list for a separate PR.

PR #3371, opened 2020-10-20
Blocksconvert: Fix chunks ordering in the block.

pstibrany:
**What this PR does**: On recent Prometheus Storage Working Group meeting, @brian-brazil has pointed out that chunks should be stored in the same order as series. While this isn't a requirement for current Prometheus, it is what Prometheus does and future optimizations may require it.

This PR fixes blocksconvert Builder to reorder chunks correctly. Unfortunately it also doubles the disk space required during the block build process (chunks are first stored randomly as they are downloaded, and reordered in the second phase).

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,7 +107,7 @@
 * [BUGFIX] Honor configured timeout in Azure and GCS object clients. #3285
 * [BUGFIX] Shuffle sharding: fixed max global series per user/metric limit when shuffle sharding and `-distributor.shard-by-all-labels=true` are both enabled in distributor. When using these global limits you should now set `-distributor.sharding-strategy` and `-distributor.zone-awareness-enabled` to ingesters too. #3369
 * [BUGFIX] Slow query logging: when using downstream server request parameters were not logged. #3276
-* [BUGFIX] Blocksconvert: fix chunks ordering in the block. #3371
+* [BUGFIX] Blocksconvert: fix chunks ordering in the block. Chunks in different order than series work just fine in TSDB blocks at the moment, but it's not consistent with what Prometheus does and future Prometheus and Cortex optimizations may rely on this ordering. #3371
 
 ## 1.4.0 / 2020-10-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,7 @@
 * [BUGFIX] Honor configured timeout in Azure and GCS object clients. #3285
 * [BUGFIX] Shuffle sharding: fixed max global series per user/metric limit when shuffle sharding and `-distributor.shard-by-all-labels=true` are both enabled in distributor. When using these global limits you should now set `-distributor.sharding-strategy` and `-distributor.zone-awareness-enabled` to ingesters too. #3369
 * [BUGFIX] Slow query logging: when using downstream server request parameters were not logged. #3276
+* [BUGFIX] Blocksconvert: fix chunks ordering in the block. #3371
 
 ## 1.4.0 / 2020-10-02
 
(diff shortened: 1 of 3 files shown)


pstibrany:
Prometheus PR: https://github.com/prometheus/prometheus/pull/8085

CHANGELOG.md
```diff
@@ -98,6 +98,7 @@
 * [BUGFIX] Fixed Gossip memberlist members joining when addresses are configured using DNS-based service discovery. #3360
 * [BUGFIX] Ingester: fail to start an ingester running the blocks storage, if unable to load any existing TSDB at startup. #3354
 * [BUGFIX] Blocks storage: Avoid deletion of blocks in the ingester which are not shipped to the storage yet. #3346
+* [BUGFIX] Blocksconvert: fix chunks ordering in the block. #3371
```
pracucci, line 101: I would clarify a bit better. In particular, I think the user may interested to know what's the impact of this bug.

tools/blocksconvert/builder/tsdb.go
```diff
@@ -211,10 +214,27 @@ func (d *tsdbBuilder) finishBlock(source string, labels map[string]string) (ulid
...
+
+	const (
+		indexWriterName          = "index writer"
+		unsortedChunksReaderName = "unsorted chunks reader"
+		chunksWriterName         = "chunks writer"
+	)
+
 	indexWriter, err := index.NewWriter(context.Background(), filepath.Join(d.tmpBlockDir, "index"))
 	if err != nil {
 		return ulid.ULID{}, errors.Wrap(err, "new index writer")
```
pracucci, line 235:
```suggestion
		return ulid.ULID{}, errors.Wrap(err, indexWriterName)
```

tools/blocksconvert/builder/tsdb.go
```diff
@@ -223,16 +243,39 @@ func (d *tsdbBuilder) finishBlock(source string, labels map[string]string) (ulid
 
 	level.Info(d.log).Log("msg", "added symbols to index", "count", symbols)
 
-	stats, err := addSeriesToIndex(indexWriter, d.series)
+	unsortedChunksReader, err := chunks.NewDirReader(filepath.Join(d.tmpBlockDir, unsortedChunksDir), nil)
+	if err != nil {
+		return ulid.ULID{}, errors.Wrap(err, "chunks reader")
```
pracucci, line 248:
```suggestion
		return ulid.ULID{}, errors.Wrap(err, unsortedChunksReaderName)
```

tools/blocksconvert/builder/tsdb.go
```diff
@@ -223,16 +243,39 @@ func (d *tsdbBuilder) finishBlock(source string, labels map[string]string) (ulid
...
+	}
+	toClose[unsortedChunksReaderName] = unsortedChunksReader
+
+	chunksWriter, err := chunks.NewWriter(filepath.Join(d.tmpBlockDir, "chunks"))
+	if err != nil {
+		return ulid.ULID{}, errors.Wrap(err, "chunks reader")
```
pracucci, line 254: Here "chunks reader" is not correct.

```suggestion
		return ulid.ULID{}, errors.Wrap(err, chunksWriterName)
```

tools/blocksconvert/builder/tsdb.go
```diff
@@ -223,16 +243,39 @@ func (d *tsdbBuilder) finishBlock(source string, labels map[string]string) (ulid
...
+		delete(toClose, k)
+
+		err := c.Close()
+		if err != nil {
+			return ulid.ULID{}, errors.Wrapf(err, "closing %s", k)
+		}
+	}
+
+	// Delete unsorted chunks, they are no longer needed.
+	if err := os.RemoveAll(filepath.Join(d.tmpBlockDir, unsortedChunksDir)); err != nil {
```
pracucci, line 277: Should be done in the `defer` cleanup function too? The one defined above.

pracucci (approved):
LGTM! Just left few nits

tools/blocksconvert/builder/tsdb.go
pstibrany: Defer only closes open resources on error, but doesn't modify the outcome.

Deleting unsorted chunks is part of normal flow, which can end up with error, although it will just make metrics wrong (when counting block dir). Still, I'd keep it as-is.

PR #3379, opened 2020-10-22
Fixed panic in gRPC opentracing instrumentation

pracucci:
**What this PR does**:
We got an incident due to ruler frequently panicking. We pinpointed the issue to BigTable gRPC tracing instrumentation (commenting it out stopped to panic) and, after further investigation, we found a timing issue in the tracing instrumentation library, for which I've proposed a fix here https://github.com/opentracing-contrib/go-grpc/pull/12.

However, given the severity of the issue for us and that the project doesn't look much active, I would be glad if we could temporarily override it to pull in the fix.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


jtlisi (approved):
LGTM

gotjosh (approved):
LGTM. 

There's another change being introduced besides the fix on `OpenTracingServerInterceptor` but that seems potential harmless?

PR #3385, opened 2020-10-24
Added option to enable/disable specific users for compaction.

pstibrany:
**What this PR does**: This PR adds ability to enable/disable specific users for compaction and cleanup.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/compactor/compactor.go
```diff
@@ -71,6 +75,9 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 		"If not 0, blocks will be marked for deletion and compactor component will delete blocks marked for deletion from the bucket. "+
 		"If delete-delay is 0, blocks will be deleted straight away. Note that deleting blocks immediately can cause query failures, "+
 		"if store gateway still has the block loaded, or compactor is ignoring the deletion because it's compacting the block at the same time.")
+
+	f.Var(&cfg.EnabledUsers, "compactor.enabled-users", "Comma separated list of users that can be compacted. If specified, only these users will be compacted by compactor, otherwise all users can be compacted. Subject to sharding.")
+	f.Var(&cfg.DisabledUsers, "compactor.disabled-users", "Comma separated list of users that cannot be compacted by this compactor. If specified, and compactor would normally pick given user for compaction (via -compactor.enabled-users or sharding), it will be ignored instead.")
```
pracucci, line 80: A couple of comments:
1. Should we cal them tenants in the CLI flag?
2. I may understand why you're keeping it hidden, but not having documented anywhere it's like doesn't exist. For this reason, maybe we should expose it in the doc and add the related YAML config option too?

pracucci (approved):
LGTM! Just left a non-blocking comment.

pkg/compactor/compactor.go
pstibrany: Agree, renamed and exposed.

PR #3388, opened 2020-10-26
Expose the config hash as a metric.

gouthamve:
If there is a config drift, then we won't dedupe metrics. This can be
used to monitor things. See: https://github.com/prometheus/alertmanager/issues/596

This is set by the config coordinator upstream which we dont use, hence
we're setting it manually.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -2,6 +2,8 @@ package alertmanager
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/binary"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -70,6 +72,11 @@ type Alertmanager struct {
 	// The Dispatcher is the only component we need to recreate when we call ApplyConfig.
 	// Given its metrics don't have any variable labels we need to re-use the same metrics.
 	dispatcherMetrics *dispatch.DispatcherMetrics
+	// This needs to be set to the hash of the config. All the hashes need to be same
+	// for deduping of alerts to work, hence we need this metric. See https://github.com/prometheus/alertmanager/issues/596
+	// Further, in upstream AM, this metric is handled using the config coordinator which we don't use
+	// hence we need to generate the metric ourselves.
+	configHashMetric prometheus.Gauge
 
 	activeMtx sync.Mutex
 	active    bool
@@ -97,9 +104,16 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 		stop:      make(chan struct{}),
 		active:    false,
 		activeMtx: sync.Mutex{},
+		configHashMetric: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "cortex_alertmanager_config_hash",
+			Help: "Hash of the currently loaded alertmanager configuration.",
+		}),
 	}
 
 	am.registry = reg
+	if am.registry != nil {
+		am.registry.MustRegister(am.configHashMetric)
+	}
 
 	am.wg.Add(1)
 	nflogID := fmt.Sprintf("nflog:%s", cfg.UserID)
@@ -182,7 +196,7 @@ func clusterWait(p *cluster.Peer, timeout time.Duration) func() time.Duration {
 }
 
 // ApplyConfig applies a new configuration to an Alertmanager.
-func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config) error {
+func (am *Alertmanager) ApplyConfig(userID string, conf *config.Config, rawCfg string) error {
 	templateFiles := make([]string, len(conf.Templates))
 	if len(conf.Templates) > 0 {
 		for i, t := range conf.Templates {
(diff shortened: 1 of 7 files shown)


roidelapluie:
I think we should make that change cortex-wise. Cortex is exposing hashes as labels.

pkg/alertmanager/multitenant.go
```diff
@@ -429,7 +431,7 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
 	} else if am.cfgs[cfg.User].RawConfig != cfg.RawConfig || hasTemplateChanges {
```
gotjosh, line 431: It might be unrelated but related - I think we can `userAmConfig.String()` instead of `rawConfig`.

In the case where you have:
```
global:
  smtp_from: 'youraddress@example.org'
```
and submit:

```
global:
  smtp_from: 'youraddress@example.org'
```

They're effectively the same but we treat them as different.

gotjosh (approved):
Nice find @gouthamve! I just have one question.

pkg/alertmanager/multitenant.go
roidelapluie: Doesn't string() hide secrets?

pkg/alertmanager/alertmanager.go
```diff
@@ -97,9 +104,16 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 		stop:      make(chan struct{}),
 		active:    false,
 		activeMtx: sync.Mutex{},
+		configHashMetric: prometheus.NewGauge(prometheus.GaugeOpts{
```
pracucci, line 107: You can use `promauto.With(reg)`. It gracefully handle the case `reg == nil`.

pkg/alertmanager/multitenant.go
pracucci: Yes, Julien is right. We shouldn't use `String()` because of this.

pkg/alertmanager/alertmanager.go
pracucci, line 107: Who adds the `user` label to this metric, when it gets exported?

pracucci, line 107: We also have `cortex_runtime_config_hash` where we expose `sha256` label. Would it work if we do the same here, to keep consistency with that metric?

pracucci (commented):
Could you add a `CHANGELOG` entry, please?

pkg/alertmanager/alertmanager.go
roidelapluie: in general I prefer alertmanager's approach, then you could have the label with `count_values("hash", cortex_runtime_config_hash) without()`. Then you can run `changes()` on it.

pkg/alertmanager/alertmanager.go
```diff
@@ -367,3 +384,12 @@ func buildReceiverIntegrations(nc *config.Receiver, tmpl *template.Template, log
...
 }
+
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data)
+	// We only want 48 bits as a float64 only has a 53 bit mantissa.
+	smallSum := sum[0:6]
+	var bytes = make([]byte, 8)
+	copy(bytes, smallSum)
+	return float64(binary.LittleEndian.Uint64(bytes))
+}
```
pstibrany, line 395: This is super ugly. As Marco pointed out above, I'd suggest to expose hash as a label (with algorithm in the label name).

pkg/alertmanager/multitenant.go
gouthamve: I think we should alert if two different configs like that end up in AM, I'd rather keep the current structure. I don't see why the configs in two different AMs would end up different like that.

pkg/alertmanager/alertmanager.go
gouthamve: See: https://github.com/cortexproject/cortex/blob/master/pkg/alertmanager/multitenant.go#L442-L460

gouthamve:
@pracucci and @pstibrany, main reason I'm doing it this way is so that we track the upstream `alertmanager_config_hash`. Having said that I don't see why we should. Changing it to sha256 as a label.

pkg/alertmanager/alertmanager.go
```diff
@@ -97,6 +104,10 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 		stop:      make(chan struct{}),
 		active:    false,
 		activeMtx: sync.Mutex{},
+		configHashMetric: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
```
pracucci, line 107: Does this metric get exposed? The input `reg` is not the exposed one, but an internal one added to  `alertmanagerMetrics`, where the metric is not exposed. Am I missing anything? Could we just add an assertion to an existing integration test? You could just assert on the count of this metric, for simplicity.

pracucci (commented):
Thanks @gouthamve for addressing my feedback. Changes LGTM, but I've the feeling the metric is not exposed (see related comment).

pkg/alertmanager/alertmanager.go
gouthamve: You're right. I've added the tests now.

pracucci (approved):
LGTM. I don't have a strong opinion about the hash value. I would have preferred to keep consistency with Cortex, but since you prefer this way let's merge it.

PR #3391, opened 2020-10-26
Move the compactor to use its own ring operation

pracucci:
**What this PR does**:
This is an internal change with no user impact. Similarly to the ruler, I've added a ring operation dedicated for the `Compactor` to make it more clear and avoid misunderstanding in the future.

/cc @MichelHollands 

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/model.go
+++ b/pkg/ring/model.go
@@ -165,6 +165,9 @@ func (i *IngesterDesc) IsHealthy(op Operation, heartbeatTimeout time.Duration) b
 
 	case Ruler:
 		healthy = i.State == ACTIVE
+
+	case Compactor:
+		healthy = i.State == ACTIVE
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/ring/model.go
```diff
@@ -165,6 +165,9 @@ func (i *IngesterDesc) IsHealthy(op Operation, heartbeatTimeout time.Duration) b
 
 	case Ruler:
 		healthy = i.State == ACTIVE
+
+	case Compactor:
+		healthy = i.State == ACTIVE
```
pracucci, line 170: Slightly different than the previous `Read`, but this one should be more accurate.

pstibrany (approved):
LGTM (no changes in tests? hmmm... 🤔)

pracucci:
> LGTM (no changes in tests? hmmm... 🤔)

Yes, no changes in tests, because unit tests run compactor and don't need to directly use the ring operation.

PR #3392, opened 2020-10-26
Refactored ingester Limiter tests

pracucci:
**What this PR does**:
In the PR #3369 I've received the comment:
> Would it be possible to refactor limiter code to avoid the same repeated tests? (In a follow-up PR, if at all).

This is the follow-up PR. The diff is not much readable (sorry), so I would suggest to just check out the actual code. Tests are the exact same, just refactored to avoid repeating all duplicated test cases.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany (approved):
Nice! Thank you!

PR #3393, opened 2020-10-26
Refactored TSDB opening concurrency at startup

pracucci:
**What this PR does**:
This is an internal refactoring with no user impact. In the PR #3354 I received the feedback:

> feel like error-handling code could be clearer by using errgroup. Also splitting the concerns would help -- one goroutine doing the scan, other N goroutines doing opening. Now both are mixed.

In this PR I'm addressing this feedback splitting the logic between the "scan" and the "opening".

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -16,7 +16,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
-	"github.com/prometheus/prometheus/pkg/gate"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb"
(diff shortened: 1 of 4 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -1096,13 +1095,44 @@ func (i *Ingester) closeAllTSDB() {
...
+				startTime := time.Now()
+
+				db, err := i.createTSDB(userID)
+				if err != nil {
+					errsMx.Lock()
+					errs.Add(errors.Wrapf(err, "unable to open TSDB for user %s", userID))
+					errsMx.Unlock()
+
+					level.Error(util.Logger).Log("msg", "unable to open TSDB", "err", err, "user", userID)
+					return
```
pstibrany, line 1121: By returning here, we lose a consumer of `queue` channel, which can lead to blocking of dir walker, when it tries to send another item to the channel.

pstibrany (commented):
This is still bit more complicated than I would prefer. Suggestion to use `errgroup` would give us 1) handling of error (no need for `errs` and `errsMx`), 2) cancellation of context (right now if opening of one TSDB fails, other will still be tried), 3) simplified waiting (no need for `wg`).

When using `errgroup`, I'd suggest to spawn dir-walker as another goroutine, which can return its own error. Dir walker should react on group context cancellation, and by closing channel via defer, it will let tsdb-opening goroutines to stop as well.

Currently this code can return multiple errors – using `errgroup` will only return first one [I don't think this matters too much], but we can still log them all – as it is now.

pkg/ingester/ingester_v2.go
pracucci: Uh good point!

pracucci:
Thanks @pstibrany for your thoughtful code review. Could you take another look, please?

pkg/ingester/ingester_v2.go
```diff
@@ -1096,98 +1095,105 @@ func (i *Ingester) closeAllTSDB() {
...
 func (i *Ingester) openExistingTSDB(ctx context.Context) error {
 	level.Info(util.Logger).Log("msg", "opening existing TSDBs")
-	wg := &sync.WaitGroup{}
-	openGate := gate.New(i.cfg.BlocksStorageConfig.TSDB.MaxTSDBOpeningConcurrencyOnStartup)
 
-	// Keep track of all errors that could occur.
-	errs := tsdb_errors.MultiError{}
-	errsMx := sync.Mutex{}
+	queue := make(chan string)
+	group, groupCtx := errgroup.WithContext(context.Background())
```
pstibrany, line 1100: We can use `ctx` instead of `context.Background`.

pkg/ingester/ingester_v2.go
```diff
@@ -1096,98 +1095,105 @@ func (i *Ingester) closeAllTSDB() {
...
-		if err != nil {
-			// If the root directory doesn't exist, we're OK (not needed to be created upfront).
-			if os.IsNotExist(err) && path == i.cfg.BlocksStorageConfig.TSDB.Dir {
-				return filepath.SkipDir
-			}
+	// Create a pool of workers which will open existing TSDBs.
+	for n := 0; n < i.cfg.BlocksStorageConfig.TSDB.MaxTSDBOpeningConcurrencyOnStartup; n++ {
+		group.Go(func() error {
+			for userID := range queue {
+				// Interrupt in case a failure occurred in another goroutine.
```
pstibrany, line 1106: Nit: I think this is unnecessary since dir-walker reacts on `groupCtx.Err()` and closes `queue`.

pstibrany (approved):
LGTM, Thanks!

PR #3399, opened 2020-10-27
Wait until container is actually running, not just created.

pstibrany:
**What this PR does**: this PR modifies integration tests to wait for container to be in Running state, and not just created. I've observed up to 30 seconds between container was created and actually running. Problem is that when integration tests try to get ports for container, and it's only in Created state, getting ports fails. In Running state, it succeeds.


pracucci (approved):
LGTM

PR #3403, opened 2020-10-27
Documented limitations and glossary

### LINKED ISSUE #3023: Add a glossary to the documentation
While supporting Cortex users, I've noticed many times there's no clear understanding of some terms we're used to frequently mention in our doc (or while giving support). It's perfectly fine for a newcomer and I believe we should do a better job to explain terms.

Some terminology we should explain:
- tenant / user / org (see https://github.com/cortexproject/cortex/issues/2987)
- flush / flushing
- series
- sample
- chunk (both related to chunks and blocks storage)

_Please comment with other terms you think we should document._

### LINKED ISSUE #3365: Document the restrictions/recommendations for naming `Tenants`

Currently Cortex does not have any guideline for naming a tenant ID. However, there are implications/restrictions when naming a tenant. For instance, a tenant should not contain a `/` character since that character is used as a delimiter by our various backend systems and using `/` in the tenant ID would likely lead to user issues. Another issue would be the length of the tenant ID itself. S3 for instance has a maximum key length of 1024 characters. A sufficiently long tenant ID would make it impossible to store objects persisted by Cortex.

A set of guidelines to follow when naming a tenant. Also updating the tenant parsing middleware to enforce certain restrictions on incoming tenants to ensure they are valid. Ideally, these restrictions should be designed to ensure any tenant ID chosen by a user would work in all possible storage backends.

pracucci:
**What this PR does**:
In this PR I'm proposing to introduce a couple of new documentation pages:
- Limitations
- Glossary

Fixes #3023
Fixes #3365

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/guides/glossary.md
+++ b/docs/guides/glossary.md
@@ -93,7 +93,7 @@ node_cpu_seconds_total{instance="10.0.0.2",mode="user"}
 
 ### Tenant
 
-A tenant (also called "user" or "org") is the owner of a set of series written to and queried from Cortex. Cortex multi-tenancy support allows you to isolate series belonging to different tenants. For example, if you have two tenants `team-A` and `team-B`, `team-A` series will be isolated from `team-B`, and each team will be able to query only their own series. 
+A tenant (also called "user" or "org") is the owner of a set of series written to and queried from Cortex. Cortex multi-tenancy support allows you to isolate series belonging to different tenants. For example, if you have two tenants `team-A` and `team-B`, `team-A` series will be isolated from `team-B`, and each team will be able to query only their own series.
 
 For more information, please refer to:
 
--- a/docs/guides/glossary.md
+++ b/docs/guides/glossary.md
@@ -97,7 +97,7 @@ A tenant (also called "user" or "org") is the owner of a set of series written t
 
 For more information, please refer to:
 
-- [HTTP API Authentication](../api/_index.md#authentication)
+- [HTTP API authentication](../api/_index.md#authentication)
 - [Tenant ID limitations](./limitations.md#tenant-id-naming)
 
 ### Time series
--- a/docs/guides/glossary.md
+++ b/docs/guides/glossary.md
@@ -45,7 +45,7 @@ For more information, please refer to the guide "[Ingesters rolling updates](./i
 
 ### Hash ring
 
-The hash ring is a distributed data structure used by Cortex for sharding, replication and service discovery. The hash ring data structure gets shared across Cortex replicas via Gossip or a key-value store.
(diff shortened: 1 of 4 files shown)


docs/guides/limitations.md
```diff
@@ -0,0 +1,17 @@
...
+linkTitle: "Limitations"
+weight: 998
+slug: limitations
+---
+
+## Tenant ID naming
+
+The tenant ID (also called "user ID" or "org ID") shouldn't include the following characters:
+
+- `/`
```
jtlisi, line 12: **thought:** Beyond these two characters I think we should consider the built-in restrictions for characters in S3 listed [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys). I haven't taken the time to verify which of the special characters listed are actually a problem and restrictions unique to other backends. 

Also it may be worth noting the 1024 character limit puts a max size on the tenant ID for blocks:

`<tenant-id>` + `/` + `<block ULID -- 26 chars>` + `/chunks/` + `<chunk ID -- 6 chars>`

1024 - 1 - 26 - 8 - 6 = 1024 - 41 = 983 maximum length of tenant ID for blocks storage

The same max length as above applies to the rules API. However, rulegroup names/namespaces are also unbounded right now so the max length is more about the interplay between the three. I will add it to a separate issue but it may make sense to have limits configured on each of these things that keep the max length in the sane category so users can't accidentally create objects to big to store in S3.

docs/guides/glossary.md
```diff
@@ -0,0 +1,115 @@
...
+node_cpu_seconds_total{instance="10.0.0.2",mode="user"}
+```
+
+### Tenant
+
+A tenant (also called "user" or "org") is the owner of a set of series written to and queried from Cortex. Cortex multi-tenancy support allows you to isolate series belonging to different tenants. For example, if you have two tenants `team-A` and `team-B`, `team-A` series will be isolated from `team-B`, and each team will be able to query only their own series.
+
+For more information, please refer to:
+
+- [HTTP API Authentication](../api/_index.md#authentication)
```
jtlisi, line 100:
```suggestion
- [HTTP API authentication](../api/_index.md#authentication)
```

jtlisi (approved):
LGTM

docs/guides/limitations.md
pracucci: Good point. Unfortunately the tenant ID max length is way less because of memcached (we have the tenant ID as part of the cache key). Memcached max key length is 250 bytes. I will put a conservative 150 characters in the doc.

pracucci:
Thanks @jtlisi for your review! I've applied your suggestions. I will merge once the CI passes in order to move on (I have few other doc improvements in my pipeline), but feel free to leave any comment post-merge and I will promptly address them.

PR #3412, opened 2020-10-28
Expose all TSDB metrics in the ingester

pracucci:
**What this PR does**:
In past few days we experienced TSDB WAL corruption twice. Given the root cause investigation (and possible fix) is a separate issue, I've also realised we're not mapping all TSDB metrics so in this PR I'm fixing this gap.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Blocks storage ingester: exported more TSDB-related metrics. #3412
+  - `cortex_ingester_tsdb_wal_corruptions_total`
+  - `cortex_ingester_tsdb_head_truncations_failed_total`
+  - `cortex_ingester_tsdb_head_truncations_total`
+  - `cortex_ingester_tsdb_head_gc_duration_seconds`
+
 ## 1.5.0 in progress
 
 * [CHANGE] Blocks storage: update the default HTTP configuration values for the S3 client to the upstream Thanos default values. #3244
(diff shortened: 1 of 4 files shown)


pkg/ingester/ingester_v2_test.go
```diff
@@ -1858,12 +1858,12 @@ func TestIngester_flushing(t *testing.T) {
 				i.FlushHandler(httptest.NewRecorder(), httptest.NewRequest("POST", "/flush", nil))
 
 				// Flush handler only triggers compactions, but doesn't wait for them to finish. Let's wait for a moment, and then verify.
-				test.Poll(t, 1*time.Second, true, func() interface{} {
+				test.Poll(t, 5*time.Second, uint64(0), func() interface{} {
```
pracucci, line 1861: This test was flaky again on my local machine, so I had:
1. Increase polling timeout from 1s to 5s
2. Compared on the actual number, so we can see what's the value of `NumSeries()` when it fails

codesome (approved):

CHANGELOG.md
```diff
@@ -89,6 +89,11 @@
 * [ENHANCEMENT] Blocks storage: reduced number of bucket listing operations to list block content (applies to newly created blocks only). #3363
 * [ENHANCEMENT] Ruler: Include the tenant ID on the notifier logs. #3372
 * [ENHANCEMENT] Blocks storage Compactor: Added `-compactor.enabled-tenants` and `-compactor.disabled-tenants` to explicitly enable or disable compaction of specific tenants. #3385
+* [ENHANCEMENT] Blocks storage ingester: exported more TSDB-related metrics. #3412
```
csmarchbanks, line 92: :wave: The v1.5 release just started and you will need to rebase to the latest master and move any CHANGELOG entries into the unreleased section. Thanks!

gouthamve (approved):

PR #3414, opened 2020-10-28
Add topology aware read

MichelHollands:
**What this PR does**:
Add zone awareness to the read path. A `MaxUnavailableZones` field is added to the `ReplicationSet` struct. The value of that field is derived from the number of zones.

In `ReplicationSet.Do()` the zones of the of the failing requests is taken into account for zone aware requests. If too many zones have failures then the request fails.

Unit tests for the non-zone aware invocations of `ReplicationSet.Do()` are added as well.
 
Fixes #<issue number>

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## master / unreleased
 
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
-* [FEATURE] Add support for zone aware reads. #3414
+* [ENHANCEMENT] Added zone-awareness support on queries. When zone-awareness is enabled, queries will still succeed if all ingesters in a single zone will fail. #3414
 * [ENHANCEMENT] Blocks storage ingester: exported more TSDB-related metrics. #3412
   - `cortex_ingester_tsdb_wal_corruptions_total`
   - `cortex_ingester_tsdb_head_truncations_failed_total`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master / unreleased
 
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
+* [FEATURE] Add support for zone aware reads. #3414
 * [ENHANCEMENT] Blocks storage ingester: exported more TSDB-related metrics. #3412
   - `cortex_ingester_tsdb_wal_corruptions_total`
   - `cortex_ingester_tsdb_head_truncations_failed_total`
--- a/docs/guides/zone-replication.md
+++ b/docs/guides/zone-replication.md
@@ -11,6 +11,8 @@ It is completely possible that all the replicas for the given data are held with
 
 For this reason, Cortex optionally supports zone-aware replication. When zone-aware replication is **enabled**, replicas for the given data are guaranteed to span across different availability zones. This requires Cortex cluster to run at least in a number of zones equal to the configured replication factor.
 
+Reads from a zone-aware replication enabled Cortex Cluster can withstand zone failures as long as there are more than replication factor / 2 zones available without any failing instances. 
+
 The Cortex services supporting **zone-aware replication** are:
 
 - **[Distributors and Ingesters](#distributors-and-ingesters-time-series-replication)**
--- a/docs/guides/zone-replication.md
+++ b/docs/guides/zone-replication.md
@@ -11,7 +11,7 @@ It is completely possible that all the replicas for the given data are held with
 
 For this reason, Cortex optionally supports zone-aware replication. When zone-aware replication is **enabled**, replicas for the given data are guaranteed to span across different availability zones. This requires Cortex cluster to run at least in a number of zones equal to the configured replication factor.
 
-Reads from a zone-aware replication enabled Cortex Cluster can withstand zone failures as long as there are more than replication factor / 2 zones available without any failing instances. 
+Reads from a zone-aware replication enabled Cortex Cluster can withstand zone failures as long as there are more than replication factor / 2 zones available without any failing instances.
 
 The Cortex services supporting **zone-aware replication** are:
 
--- a/docs/guides/zone-replication.md
+++ b/docs/guides/zone-replication.md
@@ -26,6 +26,8 @@ The Cortex time-series replication is used to hold multiple (typically 3) replic
 2. Rollout ingesters to apply the configured zone
 3. Enable time-series zone-aware replication via the `-distributor.zone-awareness-enabled` CLI flag (or its respective YAML config option). Please be aware this configuration option should be set to distributors, queriers and rulers.
 
+The `-distributor.shard-by-all-labels` setting has an impact on zone replication. When set to `true` there will be more series stored that will be spread out over more instances. When an instance goes down fewer series will be impacted.
+
 ## Store-gateways: blocks replication
 
 The Cortex [store-gateway](../blocks-storage/store-gateway.md) (used only when Cortex is running with the [blocks storage](../blocks-storage/_index.md)) supports blocks sharding, used to horizontally scale blocks in a large cluster without hitting any vertical scalability limit.
--- a/docs/guides/zone-replication.md
+++ b/docs/guides/zone-replication.md
@@ -11,7 +11,7 @@ It is completely possible that all the replicas for the given data are held with
 
 For this reason, Cortex optionally supports zone-aware replication. When zone-aware replication is **enabled**, replicas for the given data are guaranteed to span across different availability zones. This requires Cortex cluster to run at least in a number of zones equal to the configured replication factor.
 
-Reads from a zone-aware replication enabled Cortex Cluster can withstand zone failures as long as there are more than replication factor / 2 zones available without any failing instances.
+Reads from a zone-aware replication enabled Cortex Cluster can withstand zone failures as long as there are no more than `floor(replication factor / 2)` zones with failing instances.
 
 The Cortex services supporting **zone-aware replication** are:
 
@@ -28,7 +28,9 @@ The Cortex time-series replication is used to hold multiple (typically 3) replic
 2. Rollout ingesters to apply the configured zone
 3. Enable time-series zone-aware replication via the `-distributor.zone-awareness-enabled` CLI flag (or its respective YAML config option). Please be aware this configuration option should be set to distributors, queriers and rulers.
 
-The `-distributor.shard-by-all-labels` setting has an impact on zone replication. When set to `true` there will be more series stored that will be spread out over more instances. When an instance goes down fewer series will be impacted.
+The `-distributor.shard-by-all-labels` setting has an impact on read availability. When enabled, a metric is sharded across all ingesters and querier needs to fetch series from all ingesters while, when disabled, a metric is sharded only across `<replication factor>` ingesters.
+
+In the event of a large outage impacting ingesters in more than 1 zone, when `-distributor.shard-by-all-labels=true` all queries will fail, while when disabled some queries may still succeed if the ingesters holding the required metric are not impacted by the outage.
 
 ## Store-gateways: blocks replication
 
--- a/integration/e2e/service.go
+++ b/integration/e2e/service.go
@@ -172,6 +172,11 @@ func (s *ConcreteService) Kill() error {
 		logger.Log(string(out))
 		return err
 	}
+
+	// Wait until the container actually stopped. However, this could fail if
+	// the container already exited, so we just ignore the error.
+	_, _ = RunCommandAndGetOutput("docker", "wait", s.containerName())
+
 	s.usedNetworkName = ""
 
 	return nil
--- a/integration/e2e/service.go
+++ b/integration/e2e/service.go
@@ -168,7 +168,7 @@ func (s *ConcreteService) Kill() error {
 
 	logger.Log("Killing", s.name)
 
-	if out, err := RunCommandAndGetOutput("docker", "stop", "--time=0", s.containerName()); err != nil {
+	if out, err := RunCommandAndGetOutput("docker", "kill", s.containerName()); err != nil {
 		logger.Log(string(out))
 		return err
 	}
--- a/pkg/ring/replication_set.go
+++ b/pkg/ring/replication_set.go
@@ -17,16 +17,16 @@ type ReplicationSet struct {
 	MaxUnavailableZones int
 }
 
-type ingesterError struct {
-	err error
-	ing *IngesterDesc
-}
-
 // Do function f in parallel for all replicas in the set, erroring is we exceed
 // MaxErrors and returning early otherwise.
 func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(context.Context, *IngesterDesc) (interface{}, error)) ([]interface{}, error) {
+	type instanceError struct {
+		err      error
+		instance *IngesterDesc
+	}
+
 	var (
-		errs        = make(chan ingesterError, len(r.Ingesters))
+		errs        = make(chan instanceError, len(r.Ingesters))
 		resultsChan = make(chan interface{}, len(r.Ingesters))
 		minSuccess  = len(r.Ingesters) - r.MaxErrors
 		forceStart  = make(chan struct{}, r.MaxErrors)
@@ -49,9 +49,9 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 			}
 			result, err := f(ctx, ing)
 			if err != nil {
-				errs <- ingesterError{
-					err: err,
-					ing: ing,
+				errs <- instanceError{
+					err:      err,
+					instance: ing,
 				}
 			} else {
 				resultsChan <- result
@@ -68,20 +68,14 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 	for numSuccess < minSuccess {
 		select {
 		case err := <-errs:
-			numErrs++
-
 			if r.MaxUnavailableZones > 0 {
-				// Zone aware path
-				zoneFailureCount[err.ing.Zone]++
+				zoneFailureCount[err.instance.Zone]++
 
 				if len(zoneFailureCount) > r.MaxUnavailableZones {
 					return nil, errorTooManyZoneFailures
 				}
 			} else {
-				// Non zone aware path
-				// Errors per zone : only RF-1 zone can be failing
-				// if failingZones > r.MaxUnavailableZones
-
+				numErrs++
 				if numErrs > r.MaxErrors {
 					return nil, err.err
 				}
--- a/pkg/ring/replication_set.go
+++ b/pkg/ring/replication_set.go
@@ -7,7 +7,7 @@ import (
 	"time"
 )
 
-var errorTooManyZoneFailures = errors.New("Too many zones failed")
+var errorTooManyZoneFailures = errors.New("too many zones failed")
 
 // ReplicationSet describes the ingesters to talk to for a given key, and how
 // many errors to tolerate.
--- a/pkg/ring/replication_set.go
+++ b/pkg/ring/replication_set.go
@@ -63,13 +63,13 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 		numErrs          int
 		numSuccess       int
 		results          = make([]interface{}, 0, len(r.Ingesters))
-		zoneFailureCount = make(map[string]int)
+		zoneFailureCount = make(map[string]struct{})
 	)
 	for numSuccess < minSuccess {
 		select {
 		case err := <-errs:
 			if r.MaxUnavailableZones > 0 {
-				zoneFailureCount[err.instance.Zone]++
+				zoneFailureCount[err.instance.Zone] = struct{}{}
 
 				if len(zoneFailureCount) > r.MaxUnavailableZones {
 					return nil, errorTooManyZoneFailures
--- a/pkg/ring/replication_set.go
+++ b/pkg/ring/replication_set.go
@@ -2,22 +2,32 @@ package ring
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"sort"
 	"time"
 )
 
+var errorTooManyZoneFailures = errors.New("Too many zones failed")
+
 // ReplicationSet describes the ingesters to talk to for a given key, and how
 // many errors to tolerate.
 type ReplicationSet struct {
-	Ingesters []IngesterDesc
-	MaxErrors int
+	Ingesters           []IngesterDesc
+	MaxErrors           int
+	MaxUnavailableZones int
+}
+
+type ingesterError struct {
+	err error
+	ing *IngesterDesc
 }
 
 // Do function f in parallel for all replicas in the set, erroring is we exceed
 // MaxErrors and returning early otherwise.
 func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(context.Context, *IngesterDesc) (interface{}, error)) ([]interface{}, error) {
 	var (
-		errs        = make(chan error, len(r.Ingesters))
+		errs        = make(chan ingesterError, len(r.Ingesters))
 		resultsChan = make(chan interface{}, len(r.Ingesters))
 		minSuccess  = len(r.Ingesters) - r.MaxErrors
 		forceStart  = make(chan struct{}, r.MaxErrors)
@@ -38,9 +48,14 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 				case <-after.C:
 				}
 			}
+			fmt.Printf("About to run for addr %v and zone %v\n", ing.Addr, ing.Zone)
 			result, err := f(ctx, ing)
+			fmt.Printf("Addr %v and zone %v result %v\n", ing.Addr, ing.Zone, err)
 			if err != nil {
-				errs <- err
+				errs <- ingesterError{
+					err: err,
+					ing: ing,
+				}
 			} else {
 				resultsChan <- result
 			}
@@ -48,17 +63,37 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 	}
 
 	var (
-		numErrs    int
-		numSuccess int
-		results    = make([]interface{}, 0, len(r.Ingesters))
+		numErrs          int
+		numSuccess       int
+		results          = make([]interface{}, 0, len(r.Ingesters))
+		zoneFailureCount = make(map[string]int)
 	)
+	fmt.Printf("minSuccess %v\n", minSuccess)
 	for numSuccess < minSuccess {
+		fmt.Printf("numSuccess %v\n", numSuccess)
 		select {
 		case err := <-errs:
 			numErrs++
-			if numErrs > r.MaxErrors {
-				return nil, err
+
+			if r.MaxUnavailableZones > 0 {
+				// Non zone aware path
+				fmt.Printf("Incrementing count for Zone %v\n", err.ing.Zone)
+				zoneFailureCount[err.ing.Zone]++
+				fmt.Printf("Count map %v\n", zoneFailureCount)
+
+				if len(zoneFailureCount) > r.MaxUnavailableZones {
+					return nil, errorTooManyZoneFailures
+				}
+			} else {
+				// Non zone aware path
+				// Errors per zone : only RF-1 zone can be failing
+				// if failingZones > r.MaxUnavailableZones
+
+				if numErrs > r.MaxErrors {
+					return nil, err.err
+				}
 			}
+
 			// force one of the delayed requests to start
 			forceStart <- struct{}{}
 
--- a/pkg/ring/replication_set.go
+++ b/pkg/ring/replication_set.go
@@ -2,12 +2,9 @@ package ring
 
 import (
 	"context"
-	"errors"
 	"sort"
 	"time"
 )
-
-var errorTooManyZoneFailures = errors.New("too many zones failed")
 
 // ReplicationSet describes the ingesters to talk to for a given key, and how
 // many errors to tolerate.
@@ -20,24 +17,37 @@ type ReplicationSet struct {
 // Do function f in parallel for all replicas in the set, erroring is we exceed
 // MaxErrors and returning early otherwise.
 func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(context.Context, *IngesterDesc) (interface{}, error)) ([]interface{}, error) {
+	type instanceResult struct {
+		res      interface{}
+		instance *IngesterDesc
+	}
+
 	type instanceError struct {
 		err      error
 		instance *IngesterDesc
 	}
 
+	// Initialise the result tracker, which is use to keep track of successes and failures.
+	var tracker replicationSetResultTracker
+	if r.MaxUnavailableZones > 0 {
+		tracker = newZoneAwareResultTracker(r.Ingesters, r.MaxUnavailableZones)
+	} else {
+		tracker = newDefaultResultTracker(r.Ingesters, r.MaxErrors)
+	}
+
 	var (
 		errs        = make(chan instanceError, len(r.Ingesters))
-		resultsChan = make(chan interface{}, len(r.Ingesters))
-		minSuccess  = len(r.Ingesters) - r.MaxErrors
+		resultsChan = make(chan instanceResult, len(r.Ingesters))
 		forceStart  = make(chan struct{}, r.MaxErrors)
 	)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	// Spawn a goroutine for each instance.
 	for i := range r.Ingesters {
 		go func(i int, ing *IngesterDesc) {
-			// wait to send extra requests
-			if i >= minSuccess && delay > 0 {
+			// Wait to send extra requests. Works only when zone-awareness is disabled.
+			if delay > 0 && r.MaxUnavailableZones == 0 && i >= len(r.Ingesters)-r.MaxErrors {
 				after := time.NewTimer(delay)
 				defer after.Stop()
 				select {
@@ -54,39 +64,33 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 					instance: ing,
 				}
 			} else {
-				resultsChan <- result
+				resultsChan <- instanceResult{
+					res:      result,
+					instance: ing,
+				}
 			}
 		}(i, &r.Ingesters[i])
 	}
 
-	var (
-		numErrs          int
-		numSuccess       int
-		results          = make([]interface{}, 0, len(r.Ingesters))
-		zoneFailureCount = make(map[string]struct{})
-	)
-	for numSuccess < minSuccess {
+	results := make([]interface{}, 0, len(r.Ingesters))
+
+	for !tracker.succeeded() {
 		select {
 		case err := <-errs:
-			if r.MaxUnavailableZones > 0 {
-				zoneFailureCount[err.instance.Zone] = struct{}{}
+			tracker.done(err.instance, err.err)
 
-				if len(zoneFailureCount) > r.MaxUnavailableZones {
-					return nil, errorTooManyZoneFailures
-				}
-			} else {
-				numErrs++
-				if numErrs > r.MaxErrors {
-					return nil, err.err
-				}
+			if tracker.failed() {
+				return nil, err.err
 			}
 
 			// force one of the delayed requests to start
-			forceStart <- struct{}{}
+			if delay > 0 && r.MaxUnavailableZones == 0 {
+				forceStart <- struct{}{}
+			}
 
 		case result := <-resultsChan:
-			numSuccess++
-			results = append(results, result)
+			tracker.done(result.instance, nil)
+			results = append(results, result.res)
 
 		case <-ctx.Done():
 			return nil, ctx.Err()
--- a/pkg/ring/replication_set.go
+++ b/pkg/ring/replication_set.go
@@ -71,7 +71,7 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 			numErrs++
 
 			if r.MaxUnavailableZones > 0 {
-				// Non zone aware path
+				// Zone aware path
 				zoneFailureCount[err.ing.Zone]++
 
 				if len(zoneFailureCount) > r.MaxUnavailableZones {
--- a/pkg/ring/replication_set.go
+++ b/pkg/ring/replication_set.go
@@ -25,10 +25,6 @@ type ReplicationSet struct {
 func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(context.Context, *IngesterDesc) (interface{}, error)) ([]interface{}, error) {
 	type instanceResult struct {
 		res      interface{}
-		instance *IngesterDesc
-	}
-
-	type instanceError struct {
 		err      error
 		instance *IngesterDesc
 	}
@@ -42,9 +38,8 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 	}
 
 	var (
-		errs        = make(chan instanceError, len(r.Ingesters))
-		resultsChan = make(chan instanceResult, len(r.Ingesters))
-		forceStart  = make(chan struct{}, r.MaxErrors)
+		ch         = make(chan instanceResult, len(r.Ingesters))
+		forceStart = make(chan struct{}, r.MaxErrors)
 	)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
@@ -64,16 +59,10 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 				}
 			}
 			result, err := f(ctx, ing)
-			if err != nil {
-				errs <- instanceError{
-					err:      err,
-					instance: ing,
-				}
-			} else {
-				resultsChan <- instanceResult{
-					res:      result,
-					instance: ing,
-				}
+			ch <- instanceResult{
+				res:      result,
+				err:      err,
+				instance: ing,
 			}
 		}(i, &r.Ingesters[i])
 	}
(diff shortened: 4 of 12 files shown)


CHANGELOG.md
```diff
@@ -55,6 +55,7 @@
 * [FEATURE] Add OpenStack Swift support to blocks storage. #3303
 * [FEATURE] Added support for applying Prometheus relabel configs on series received by the distributor. A `metric_relabel_configs` field has been added to the per-tenant limits configuration. #3329
 * [FEATURE] Support for Cassandra client SSL certificates. #3384
+* [FEATURE] Add support for zone aware reads. #3414
```
csmarchbanks, line 58: :wave: The v1.5 release just started and you will need to rebase to the latest master and move any CHANGELOG entries into the unreleased section. Thanks!

MichelHollands: Thanks for letting me know. A rebase has been done.

pkg/ring/ring.go
```diff
@@ -326,16 +326,37 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+		for _, ingester := range r.ringDesc.Ingesters {
+			zones[ingester.Zone] = true
+			ingesters = append(ingesters, ingester)
+		}
+		maxUnavailableZones = len(zones) / 2
+		// Assume ingesters are evenly spread over zones
+		// The maximum number of failures is all the ingesters in the maximum number of failing zones
+		maxErrors := (len(ingesters) / len(zones)) * maxUnavailableZones
+
+		return ReplicationSet{
```
pracucci, line 344: Design-wise I'm not a big fan of this early return. I would be glad if you could investigate if we can cleanly improve the `GetAll()` implementation without falling into two different implementations (in this PR it's basically a big if branch). If it's not clean, then we could get rediscuss it.

pkg/ring/ring.go
```diff
@@ -326,16 +326,37 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+		zones := make(map[string]bool)
+
+		for _, ingester := range r.ringDesc.Ingesters {
+			zones[ingester.Zone] = true
+			ingesters = append(ingesters, ingester)
+		}
+		maxUnavailableZones = len(zones) / 2
+		// Assume ingesters are evenly spread over zones
+		// The maximum number of failures is all the ingesters in the maximum number of failing zones
+		maxErrors := (len(ingesters) / len(zones)) * maxUnavailableZones
```
pracucci, line 342: No, this is not correct. With `MaxErrors` you're not telling "max number of ingesters failing in 1 specific zone", but you're telling "max number of ingesters failing across the whole cluster". When zone awareness is enabled, max errors is... it depends:

1. If no instance has been filtered out because unhealthy, then it's equal to `maxUnavailable` as computed below (when zone-awareness is disabled)
2. If instances in 1 zone have been all filtered out (which should be when the number of failing instances in a single zone are > `maxUnavailable`) then `MaxErrors: 0` (because we can't sustain any other failure)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [FEATURE] Add support for zone aware reads. #3414
```
pracucci, line 5: We usually consider feature a branch new feature. In this case we're improving an existing one.

```suggestion
* [ENHANCEMENT] Added zone-awareness support on queries. When zone-awareness is enabled, queries will still succeed if all ingesters in a single zone will fail. #3414
```

pracucci (commented):
Good job @MichelHollands! I think we're on the good track. I left you many comments, but most are small, while `GetAll()` will need a bit more work.

That being said, I would be glad if you could add 1 integration test. I would test this scenario:
- Start a cortex cluster with 6 ingesters and zone-awareness enabled
- Push 100 series
- Query back 100 series > all good
- SIGKILL 1 ingester in 1 zone
- Query back 100 series > all good
- SIGKILL 1 more ingester in the same zone
- Query back 100 series > all good
- SIGKILL 1 more ingester in a different zone
- Query back 100 series > fail

pkg/ring/replication_set.go
```diff
@@ -2,22 +2,31 @@ package ring
...
 // many errors to tolerate.
 type ReplicationSet struct {
-	Ingesters []IngesterDesc
-	MaxErrors int
+	Ingesters           []IngesterDesc
+	MaxErrors           int
+	MaxUnavailableZones int
+}
+
+type ingesterError struct {
```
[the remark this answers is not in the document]
MichelHollands: Changed to instance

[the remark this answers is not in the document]
MichelHollands: Moved

pkg/ring/replication_set.go
```diff
@@ -2,22 +2,31 @@ package ring
...
+	err error
+	ing *IngesterDesc
```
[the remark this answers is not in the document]
MichelHollands: Changed manually

pkg/ring/replication_set.go
```diff
@@ -40,25 +49,44 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
...
+				// Zone aware path
+				zoneFailureCount[err.ing.Zone]++
+
+				if len(zoneFailureCount) > r.MaxUnavailableZones {
+					return nil, errorTooManyZoneFailures
+				}
+			} else {
+				// Non zone aware path
+				// Errors per zone : only RF-1 zone can be failing
+				// if failingZones > r.MaxUnavailableZones
```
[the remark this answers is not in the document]
MichelHollands: The comment has been removed

pkg/ring/replication_set.go
```diff
@@ -40,25 +49,44 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
...
 	)
 	for numSuccess < minSuccess {
 		select {
 		case err := <-errs:
 			numErrs++
-			if numErrs > r.MaxErrors {
-				return nil, err
+
+			if r.MaxUnavailableZones > 0 {
+				// Zone aware path
```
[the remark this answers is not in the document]
MichelHollands: The comment has been removed

pkg/ring/replication_set.go
```diff
@@ -40,25 +49,44 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
...
-		results    = make([]interface{}, 0, len(r.Ingesters))
+		numErrs          int
+		numSuccess       int
+		results          = make([]interface{}, 0, len(r.Ingesters))
+		zoneFailureCount = make(map[string]int)
 	)
 	for numSuccess < minSuccess {
 		select {
 		case err := <-errs:
 			numErrs++
```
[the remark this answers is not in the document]
MichelHollands: It has been moved

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +37,231 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+			}
+		}
+		return 1, nil
+	}
+}
+
+func TestReplicationSet_Do(t *testing.T) {
+	tests := []struct {
+		name                string
+		ingesters           []IngesterDesc
```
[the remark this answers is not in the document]
MichelHollands: Changed manually

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +37,231 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+		maxErrors           int
+		maxUnavailableZones int
+		f                   func(context.Context, *IngesterDesc) (interface{}, error)
+		delay               time.Duration
+		closeContextDelay   time.Duration
+		want                []interface{}
+		wantErr             bool
```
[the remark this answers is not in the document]
MichelHollands: Changed so wantErr is `removed` and the check is done on an actual error.

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +37,231 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+			name:      "1 error, no errors expected",
+			ingesters: []IngesterDesc{{}},
+			f: func(c context.Context, id *IngesterDesc) (interface{}, error) {
+				return nil, errors.New("Dummy error")
+			},
+			want:    nil,
+			wantErr: true,
+		},
+		{
+			name:      "3 ingesters, last call fails",
```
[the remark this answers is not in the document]
MichelHollands: Changed manually

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +37,231 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+
+func TestReplicationSet_Do(t *testing.T) {
+	tests := []struct {
+		name                string
+		ingesters           []IngesterDesc
+		maxErrors           int
+		maxUnavailableZones int
+		f                   func(context.Context, *IngesterDesc) (interface{}, error)
+		delay               time.Duration
+		closeContextDelay   time.Duration
```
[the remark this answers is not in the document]
MichelHollands: Renamed manually

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +37,231 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+			f:         failingFunctionAfter(2, 10*time.Millisecond),
+			delay:     100 * time.Millisecond,
+			want:      nil,
+			wantErr:   true,
+		},
+		{
+			name:              "3 ingesters, context fails",
+			ingesters:         []IngesterDesc{{}, {}, {}},
+			maxErrors:         1,
+			f:                 failingFunctionAfter(0, 200*time.Millisecond),
```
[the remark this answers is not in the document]
MichelHollands: `f` was changed so it doesn't fail.

pkg/ring/ring.go
```diff
@@ -326,16 +326,37 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
 		return ReplicationSet{}, ErrEmptyRing
 	}
 
+	var maxUnavailableZones int
+	var numRequired int
```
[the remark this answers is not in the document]
MichelHollands: The definition was removed

pkg/ring/ring.go
MichelHollands: The zoneware code was moved to be in the main code so it reuses more of the variables.

pkg/ring/ring.go
```diff
@@ -326,16 +326,37 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+	ingesters := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+	if r.cfg.ZoneAwarenessEnabled {
+		zones := make(map[string]bool)
```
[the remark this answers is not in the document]
MichelHollands: This code has been rewritten so this no longer applies.

CHANGELOG.md
MichelHollands: Applied directly.

pkg/ring/ring.go
MichelHollands: The cases should now be handled as well.

pkg/ring/ring.go
```diff
@@ -326,16 +326,37 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+	numRequired = len(r.ringDesc.Ingesters)
 	if numRequired < r.cfg.ReplicationFactor {
 		numRequired = r.cfg.ReplicationFactor
 	}
 	maxUnavailable := r.cfg.ReplicationFactor / 2
 	numRequired -= maxUnavailable
 
-	ingesters := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
 	for _, ingester := range r.ringDesc.Ingesters {
 		if r.IsHealthy(&ingester, op) {
```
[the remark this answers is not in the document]
MichelHollands: Code has been added to handle the case where instances in a zone are failing.

pkg/ring/ring.go
```diff
@@ -326,29 +339,51 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
-			ingesters = append(ingesters, ingester)
+			instances = append(instances, ingester)
+
+		} else {
+			zoneFailures[ingester.Zone]++
+		}
+	}
+	maxErrors := len(instances) - numRequired
+
+	if r.cfg.ZoneAwarenessEnabled {
```
pracucci, line 360: This would simplify the logic inside the if.

```suggestion
	if r.cfg.ZoneAwarenessEnabled && if len(zoneFailures) > 0 {
```

pracucci (commented):
Thanks @MichelHollands for addressing my feedback! I left few more comments. Please make sure that any fix you do is also covered by a unit test to avoid any regression (either in this PR or future changes).

pkg/ring/ring.go
```diff
@@ -326,29 +339,51 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
 		return ReplicationSet{}, ErrEmptyRing
 	}
 
+	maxUnavailableZones := len(r.ringZones) / 2
```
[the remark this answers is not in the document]
MichelHollands: It was changed to `max(0, len(zones) - maxUnavailable -1)`. Without the minus 1 for example for RF=3 and `len(zones)` = 3 then it would result in 2. Similarly for RF = 5 and `len(zones)` = 5 that would result in 3.

pkg/ring/ring.go
```diff
@@ -326,29 +339,51 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+	numRequired := calculateRequiredInstances(len(r.ringDesc.Ingesters), r.cfg.ReplicationFactor)
 
-	ingesters := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+	instances := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+	zoneFailures := make(map[string]int)
 	for _, ingester := range r.ringDesc.Ingesters {
 		if r.IsHealthy(&ingester, op) {
-			ingesters = append(ingesters, ingester)
+			instances = append(instances, ingester)
+
```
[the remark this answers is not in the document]
MichelHollands: Oops, that wasn't supposed to be there.

pkg/ring/ring.go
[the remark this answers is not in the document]
MichelHollands: Added

pkg/ring/ring.go
```diff
@@ -326,29 +339,51 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+			zoneFailures[ingester.Zone]++
+		}
+	}
+	maxErrors := len(instances) - numRequired
+
+	if r.cfg.ZoneAwarenessEnabled {
+		filteredInstances := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+		if len(zoneFailures) > 0 {
+			for _, ingester := range instances {
+				_, present := zoneFailures[ingester.Zone]
```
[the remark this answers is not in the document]
MichelHollands: Fixed

pkg/ring/ring.go
```diff
@@ -326,29 +339,51 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+				if !present {
+					filteredInstances = append(filteredInstances, ingester)
+				}
+			}
+		}
+
+		if len(filteredInstances) != 0 {
+			instances = filteredInstances
+			maxUnavailableZones = 0
+			numRequired = calculateRequiredInstances(len(instances), r.cfg.ReplicationFactor)
```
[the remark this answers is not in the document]
MichelHollands: Yes, that's correct. `calculateRequiredInstances` has been removed now.

pkg/ring/ring.go
```diff
@@ -326,29 +339,51 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
 	for _, ingester := range r.ringDesc.Ingesters {
 		if r.IsHealthy(&ingester, op) {
-			ingesters = append(ingesters, ingester)
+			instances = append(instances, ingester)
+
+		} else {
+			zoneFailures[ingester.Zone]++
+		}
+	}
+	maxErrors := len(instances) - numRequired
```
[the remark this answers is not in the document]
MichelHollands: Moved.

pkg/ring/ring.go
MichelHollands: Good catch. The if statement has been updated and the redundant if statement inside has been removed.

pkg/ring/ring.go
```diff
@@ -335,20 +338,49 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
 	maxUnavailable := r.cfg.ReplicationFactor / 2
 	numRequired -= maxUnavailable
 
-	ingesters := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+	maxUnavailableZones := len(r.ringZones) - maxUnavailable - 1
```
pracucci, line 341: I think I've deceived you in my previous comment, showing you a *wrong* pseudo code. I think it should be `maxUnavailableZones := maxUnavailable`. Am I missing anything?

pkg/ring/ring.go
```diff
@@ -335,20 +338,49 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+	if maxUnavailableZones < 0 {
```
pracucci, line 342: Given the comment about you will not need this anymore, but if you will need again in the future please be aware you can just use `util.Max(0, ...)`

pkg/ring/ring.go
```diff
@@ -335,20 +338,49 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+			zoneFailures[ingester.Zone]++
+		}
+	}
+
+	if r.cfg.ZoneAwarenessEnabled && len(zoneFailures) > 0 {
+		if len(zoneFailures) > maxUnavailableZones {
+			return ReplicationSet{}, ErrTooManyFailedIngesters
+		}
+
+		filteredInstances := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
```
pracucci, line 361: I would add a comment explaining why we do this. Something like this: "Given we can tolerate a zone failure, we can entirely skip querying instances in that zone."

pkg/ring/ring.go
```diff
@@ -335,20 +338,49 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
+		for _, ingester := range instances {
+			if _, ok := zoneFailures[ingester.Zone]; !ok {
+				filteredInstances = append(filteredInstances, ingester)
+			}
+		}
+
+		if len(filteredInstances) != 0 {
```
pracucci, line 368: Are there cases where we don't enter this `if`, given the current logic?

pkg/ring/ring.go
```diff
@@ -335,20 +338,49 @@ func (r *Ring) GetAll(op Operation) (ReplicationSet, error) {
...
 	for _, ingester := range r.ringDesc.Ingesters {
 		if r.IsHealthy(&ingester, op) {
-			ingesters = append(ingesters, ingester)
+			instances = append(instances, ingester)
+		} else {
+			zoneFailures[ingester.Zone]++
+		}
+	}
+
+	if r.cfg.ZoneAwarenessEnabled && len(zoneFailures) > 0 {
```
pracucci, line 356: Should we also add `&& maxUnavailableZones > 0`? I think we shouldn't enter this logic if we can't tolerate any zone failure.

pracucci (commented):
I just realised that there are places where triggering the zone-awareness logic in the `GetAll()` is not correct, because it will filter out even healthy instances (eg. `NewRingServiceDiscovery()`, `distributor.GetAll()`, etc..). Let's talk offline about this.

pkg/ring/ring.go
MichelHollands: No, that's what it comes down to :)

pkg/ring/ring.go
MichelHollands: The comment was added.

pkg/ring/ring.go
MichelHollands: Indeed this will always apply.

pkg/ring/ring.go
MichelHollands: `maxUnavailableZones = 0` when the replication factor = 1. If any zone failures occur that will be caught by `if len(zoneFailures) > maxUnavailableZones` on the next line. So I don't think it's absolutely necessary.

On the other hand it won't hurt so it's been added. I've added extra test cases that have a replication factor of 1 and 1 or more zones. These tests up the coverage as well.

pkg/ring/replication_set.go
```diff
@@ -40,25 +49,38 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
...
 	}
 
 	var (
-		numErrs    int
-		numSuccess int
-		results    = make([]interface{}, 0, len(r.Ingesters))
+		numErrs          int
+		numSuccess       int
+		results          = make([]interface{}, 0, len(r.Ingesters))
+		zoneFailureCount = make(map[string]int)
```
pstibrany, line 66: Number of failures within the zone is not used. To avoid confusion about why using `int` as value type, we can use `map[string]struct{}` instead.

pkg/ring/ring.go
```diff
@@ -371,21 +371,47 @@ func (r *Ring) GetReplicationSetForOperation(op Operation) (ReplicationSet, erro
 	}
 	maxUnavailable := r.cfg.ReplicationFactor / 2
 	numRequired -= maxUnavailable
+	numRequiredZones := util.Min(len(r.ringZones), r.cfg.ReplicationFactor)
+	maxUnavailableZones := util.Max(0, numRequiredZones-maxUnavailable-1)
 
-	ingesters := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+	instances := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+	zoneFailures := make(map[string]int)
```
pstibrany, line 378: Let's use `map[string]struct{}` if we don't use the value.

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +37,233 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+
+var (
+	errFailure     = errors.New("failed")
+	errZoneFailure = errors.New("zone failed")
+)
+
+// Return a function that fails starting from failAfter times
+func failingFunctionAfter(failAfter int, delay time.Duration) func(context.Context, *IngesterDesc) (interface{}, error) {
+	var mutex = &sync.RWMutex{}
+	count := 0
```
pstibrany, line 49: using `atomic.Int32` can simplify the test a bit.


```suggestion
func failingFunctionAfter(failAfter int32, delay time.Duration) func(context.Context, *IngesterDesc) (interface{}, error) {
	count := atomic.NewInt32(0)
	return func(context.Context, *IngesterDesc) (interface{}, error) {
		time.Sleep(delay)
		if count.Inc() > failAfter {
			return nil, errFailure
		}
		return 1, nil
	}
}
```

pstibrany (commented):
First review pass. I need to get better understanding of the tests.

pkg/ring/replication_set.go
MichelHollands: Good idea. Originally I thought the count for each zone was needed but that was not the case.

pkg/ring/ring.go
MichelHollands: Same here.

pkg/ring/replication_set_tracker_test.go
```diff
@@ -0,0 +1,266 @@
...
+				assert.False(t, tracker.succeeded())
+				assert.False(t, tracker.failed())
+
+				// Zone-b
+				tracker.done(&instance3, nil)
+				assert.False(t, tracker.succeeded())
+				assert.False(t, tracker.failed())
+
+				// Zone-a
+				tracker.done(&instance1, nil)
```
MichelHollands, line 254: Shouldn't this be instance2?

pracucci: Ops, yes! Fixed. Test was passing anyway because trackers don't keep track of "which" instance has done, only the count.

pkg/ring/replication_set.go
```diff
@@ -10,25 +10,50 @@ import (
...
+
+	// Maximum number of different zones in which instances can fail. Max unavailable zones and
+	// max errors are mutually exclusive.
+	MaxUnavailableZones int
 }
 
 // Do function f in parallel for all replicas in the set, erroring is we exceed
 // MaxErrors and returning early otherwise.
 func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(context.Context, *IngesterDesc) (interface{}, error)) ([]interface{}, error) {
+	type instanceResult struct {
```
pstibrany, line 26: This method would benefit from merging `instanceResult` and `instanceError` and the two channels (`errs` and `resultsChan`). It gets shorter and easier to follow. (See https://gist.github.com/pstibrany/153b237d0dcd861828a706c089ebc8cf)

pkg/ring/replication_set_tracker.go
```diff
@@ -0,0 +1,104 @@
...
+func (t *zoneAwareResultTracker) done(instance *IngesterDesc, err error) {
+	t.waitingByZone[instance.Zone]--
+
+	if err != nil {
+		t.failuresByZone[instance.Zone]++
+	}
+}
+
+func (t *zoneAwareResultTracker) succeeded() bool {
+	actualSucceededZones := 0
```
pstibrany, line 79: `successfulZones`

pkg/ring/replication_set_tracker.go
```diff
@@ -0,0 +1,104 @@
...
+func newZoneAwareResultTracker(instances []IngesterDesc, maxUnavailableZones int) *zoneAwareResultTracker {
+	t := &zoneAwareResultTracker{
+		waitingByZone:       make(map[string]int),
+		failuresByZone:      make(map[string]int),
+		maxUnavailableZones: maxUnavailableZones,
+	}
+
+	for _, instance := range instances {
+		t.waitingByZone[instance.Zone]++
+		t.failuresByZone[instance.Zone] = 0
```
pstibrany, line 63: By not initializing this to 0 (which is default value anyway, and we don't test for element in the map anywhere in this code), we could use `failedZones` = len(t.failuresByZone)` in `failed()`.

pkg/ring/replication_set_tracker.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (t *defaultResultTracker) succeeded() bool {
+	return t.numSucceeded >= t.minSucceeded
+}
+
+func (t *defaultResultTracker) failed() bool {
+	return t.numErrors > t.maxErrors
+}
+
+type zoneAwareResultTracker struct {
```
pstibrany, line 47: suggestion: I would add comment that this tracker expects that *all* instances within a zone must succeed in order for zone to succeed. (Fortunately it's quite clear from the code)

pkg/ring/ring_test.go
```diff
@@ -418,6 +418,315 @@ func TestRing_GetReplicationSetForOperation(t *testing.T) {
...
+		"RF=3, only 1 zone, two instances per zone, one instance unhealthy": {
+			ringInstances: map[string]IngesterDesc{
+				"instance-1": {Addr: "127.0.0.1", Zone: "zone-a", Tokens: GenerateTokens(128, nil)},
+				"instance-2": {Addr: "127.0.0.2", Zone: "zone-a", Tokens: GenerateTokens(128, nil)},
+			},
+			expectedAddresses:  []string{},
+			unhealthyInstances: []string{"instance-2"},
+			replicationFactor:  3,
+			expectedError:      ErrTooManyFailedIngesters,
+		},
```
pstibrany, line 600: Can we add `RF=3, one zone, 4 instances, one unhealthy` scenario? (I would expect that we can actually tolerate single unhealthy instance in this case – right? – but it actually returns too many failed ingesters error)

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +38,226 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+			}, {
+				Zone: "zone3",
+			}},
+			f:                   failingFunctionOnZones("zone1", "zone2"),
+			maxUnavailableZones: 1,
+			expectedError:       errZoneFailure,
+		},
+		{
+			name: "max unavailable zones = 1, should succeed on instances failing in 1 out of 3 zones (6 instances)",
+			instances: []IngesterDesc{{
```
pstibrany, line 171: Can we reformat these long vertical blocks to be more compact and readable? (In entire test)

pstibrany (commented):
Some initial comments. I still need to review `TestReplicationSet_Do` and integration test.

pkg/ring/replication_set_tracker.go
MichelHollands: Added

pkg/ring/replication_set_tracker.go
MichelHollands: Changed

pkg/ring/replication_set_tracker.go
MichelHollands: Renamed

pkg/ring/replication_set.go
MichelHollands: It is indeed shorter and easier to follow. Applied.

pkg/ring/ring.go
```diff
@@ -363,29 +363,73 @@ func (r *Ring) GetReplicationSetForOperation(op Operation) (ReplicationSet, erro
...
+		// instance, we have to decrease the max unavailable zones accordingly.
+		maxUnavailableZones -= len(zoneFailures)
+	} else {
+		// Calculate the number of required ingesters;
+		// ensure we always require at least RF-1 when RF=3.
+		numRequired := len(r.ringDesc.Ingesters)
+		if numRequired < r.cfg.ReplicationFactor {
+			numRequired = r.cfg.ReplicationFactor
+		}
+		maxUnavailable := r.cfg.ReplicationFactor / 2
```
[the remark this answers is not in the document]
MichelHollands: Done

pkg/ring/ring.go
```diff
@@ -363,29 +363,73 @@ func (r *Ring) GetReplicationSetForOperation(op Operation) (ReplicationSet, erro
...
-	numRequired := len(r.ringDesc.Ingesters)
-	if numRequired < r.cfg.ReplicationFactor {
-		numRequired = r.cfg.ReplicationFactor
-	}
-	maxUnavailable := r.cfg.ReplicationFactor / 2
-	numRequired -= maxUnavailable
-
-	ingesters := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
+	// Build the initial replication set, excluding unhealthy instances.
+	instances := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
```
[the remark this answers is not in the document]
MichelHollands: Renamed

pkg/ring/ring_test.go
MichelHollands: We don't support a use case where there are less zones than the replication factor. This causes problems on the write path. It is documented in https://cortexmetrics.io/docs/guides/zone-aware-replication/

pstibrany: Right, thanks.

pkg/ring/ring.go
```diff
@@ -363,29 +363,73 @@ func (r *Ring) GetReplicationSetForOperation(op Operation) (ReplicationSet, erro
...
+		if len(zoneFailures) > maxUnavailableZones {
+			return ReplicationSet{}, ErrTooManyFailedIngesters
+		}
+
+		if len(zoneFailures) > 0 {
+			// We remove all instances (even healthy ones) from zones with at least
+			// 1 failing ingester. Due to how replication works when zone-awareness is
+			// enabled (data is replicated to RF different zones), there's no benefit in
+			// querying healthy instances from "failing zones".
+			filteredInstances := make([]IngesterDesc, 0, len(r.ringDesc.Ingesters))
```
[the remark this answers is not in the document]
MichelHollands: The tokens in a zone are randomly distributed. So if one of the ingesters in a zone fails the only way to make sure you have all the data is to loop over all the tokens and see if there is data for all of them. We're trying to avoid having to do that computation.
If all the ingesters in a zone are healthy then we don't have to worry about that.

pkg/ring/ring.go
```diff
@@ -363,29 +363,73 @@ func (r *Ring) GetReplicationSetForOperation(op Operation) (ReplicationSet, erro
...
+	// to 0 and then we update them whether zone-awareness is enabled or not.
+	maxErrors := 0
+	maxUnavailableZones := 0
+
+	if r.cfg.ZoneAwarenessEnabled {
+		// Given data is replicated to RF different zones, we can tolerate a number of
+		// RF/2 failing zones. However, we need to protect from the case the ring currently
+		// contains instances in a number of zones < RF.
+		numReplicatedZones := util.Min(len(r.ringZones), r.cfg.ReplicationFactor)
+		minSuccessZones := (numReplicatedZones / 2) + 1
```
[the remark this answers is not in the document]
MichelHollands: Added a special case for rf=2 and nr of zones =2/

pkg/ring/replication_set_test.go
MichelHollands: Any suggestion how to achieve that? I could add a function to create the list of instances which could perhaps also return the expectedAddresses. Not sure that would be much more readable though.

pstibrany: I didn't mean anything fancy, just removing extra newlines would go a long way here. Eg.:

```
			instances:           []IngesterDesc{{Zone: "zone1"}, {Zone: "zone1"}, {Zone: "zone2"}, {Zone: "zone2"}, {Zone: "zone3"}, {Zone: "zone3"}},
```

instead of 

```
			instances: []IngesterDesc{{
				Zone: "zone1",
			}, {
				Zone: "zone1",
			}, {
				Zone: "zone2",
			}, {
				Zone: "zone2",
			}, {
				Zone: "zone3",
			}, {
				Zone: "zone3",
			}}
```

pkg/ring/replication_set_test.go
```diff
@@ -33,3 +38,226 @@ func TestReplicationSet_GetAddresses(t *testing.T) {
...
+			}
+			ctx := context.Background()
+			if tt.cancelContextDelay > 0 {
+				var cancel context.CancelFunc
+				ctx, cancel = context.WithCancel(ctx)
+				go func() {
+					time.AfterFunc(tt.cancelContextDelay, func() {
+						cancel()
+					})
+				}()
```
pstibrany, line 252: no need for `go func()...` here... `time.AfterFunc` does that already

pkg/ring/ring.go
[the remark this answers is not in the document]
pstibrany: I'm actually not sure anymore if my previous statement is correct. Reason is that during write, we would accept it if we get *single* OK. But during read, we don't know which single zone has received the sample, so we need to ask both.

I'd check with @pracucci to be 100% sure. But now I think my previous comment here was incorrect.

pkg/ring/ring.go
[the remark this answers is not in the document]
pstibrany: I commented on this to improve the comment in the code.

Comment in the code says "there's no benefit querying healthy instances from failing zones", but your explanation suggests that it would actually be harmful (too slow). I don't quite see why we would need to loop over all tokens – we know which ingester(s) "own" the token (main + replicas), and we only need to ask those.

I think the real reason why we can skip all ingesters in this zone, is that we either treat entire zone as failed if there is single error, or successul if there are no errors.

(Emphasis on "I think". I may be wrong about it)

pkg/ring/replication_set_test.go
MichelHollands: Removed.

pkg/ring/replication_set_test.go
MichelHollands: My mistake, I was looking in the wrong place. The newlines have been removed.

pkg/ring/ring.go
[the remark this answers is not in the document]
MichelHollands: Changed

pkg/ring/ring.go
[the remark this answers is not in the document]
MichelHollands: The comment was updated.

docs/guides/zone-replication.md
```diff
@@ -11,6 +11,8 @@ It is completely possible that all the replicas for the given data are held with
 
 For this reason, Cortex optionally supports zone-aware replication. When zone-aware replication is **enabled**, replicas for the given data are guaranteed to span across different availability zones. This requires Cortex cluster to run at least in a number of zones equal to the configured replication factor.
 
+Reads from a zone-aware replication enabled Cortex Cluster can withstand zone failures as long as there are more than replication factor / 2 zones available without any failing instances.
```
pracucci, line 14: > as long as there are more than replication factor / 2 zones available without any failing instances

This is not true. Think about 5 zones and RF=3. You're asserting reads will succeed as far as there are 2 healthy zones, but actually you need 4 healthy zones.

```suggestion
Reads from a zone-aware replication enabled Cortex Cluster can withstand zone failures as long as there are no more than `floor(replication factor / 2)` zones with failing instances.
```

docs/guides/zone-replication.md
```diff
@@ -26,6 +28,8 @@ The Cortex time-series replication is used to hold multiple (typically 3) replic
 2. Rollout ingesters to apply the configured zone
 3. Enable time-series zone-aware replication via the `-distributor.zone-awareness-enabled` CLI flag (or its respective YAML config option). Please be aware this configuration option should be set to distributors, queriers and rulers.
 
+The `-distributor.shard-by-all-labels` setting has an impact on zone replication. When set to `true` there will be more series stored that will be spread out over more instances. When an instance goes down fewer series will be impacted.
```
pracucci, line 31:
```suggestion
The `-distributor.shard-by-all-labels` setting has an impact on read availability. When enabled, a metric is sharded across all ingesters and querier needs to fetch series from all ingesters while, when disabled, a metric is sharded only across `<replication factor>` ingesters.

In the event of a large outage impacting ingesters in more than 1 zone, when `-distributor.shard-by-all-labels=true` all queries will fail, while when disabled some queries may still succeed if the ingesters holding the required metric are not impacted by the outage.
```

pkg/ring/ring.go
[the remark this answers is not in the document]
pracucci: What's the current status of this conversation?

`DefaultReplicationStrategy.Filter()` calculate `minSuccess := (replicationFactor / 2) + 1`. The `minSuccess` is the write quorum. If RF=2, then we need to write to both ingesters in order to succeed. Because of this, it's safe to have `maxUnavailableZones=1` when `RF=2`.

I suggest changing the logic as follow (the idea is that :
```
maxUnavailableZones = minSuccessZones - 1
```

Tests should be updated accordingly.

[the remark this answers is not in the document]
pstibrany: > What's the current status of this conversation?

I convinced myself that with RF=2, it is enough to write to single ingester. But that's not a majority 🤦

pracucci (approved):
Thanks @MichelHollands for patiently addressing my feedback. Well done, LGTM! 👏

pkg/ring/ring.go
```diff
@@ -363,29 +363,74 @@ func (r *Ring) GetReplicationSetForOperation(op Operation) (ReplicationSet, erro
...
+		if len(zoneFailures) > maxUnavailableZones {
+			return ReplicationSet{}, ErrTooManyFailedIngesters
+		}
+
+		if len(zoneFailures) > 0 {
+			// We remove all instances (even healthy ones) from zones with at least
+			// 1 failing ingester. Due to how replication works when zone-awareness is
+			// enabled (data is replicated to RF different zones), there's no benefit in
+			// querying healthy instances from "failing zones". A zone is considered
+			// failed if there is single error.
```
ranton256, line 399: I don't follow the logic behind this.
If I have 1 failed replica in zone A, and 1 failed in zone B, and none failed in zone C, won't this cause an outage for some timeseries when there would have been none in some cases(i.e. the instance in A and the instance in B didn't host a replica of any of the same series). With this logic of removing everything from a zone if there is a failure, won't this make it more likely to have an outage in some cases?

pkg/ring/ring.go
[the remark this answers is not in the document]
ranton256: > I don't quite see why we would need to loop over all tokens – we know which ingester(s) "own" the token (main + replicas), and we only need to ask those.

I also don't understand why we would need to loop over all of them.

integration/e2e/service.go
```diff
@@ -168,10 +168,15 @@ func (s *ConcreteService) Kill() error {
 
 	logger.Log("Killing", s.name)
 
-	if out, err := RunCommandAndGetOutput("docker", "stop", "--time=0", s.containerName()); err != nil {
+	if out, err := RunCommandAndGetOutput("docker", "kill", s.containerName()); err != nil {
```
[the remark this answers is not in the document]
pracucci: The `Kill()` function, used by end-to-end tests, is expected to send a `SIGKILL`. We were using `docker stop --time=0`, which is expected to send a SIGTERM and then a `SIGKILL` after time `0`, but we wondered if there could be some timing issues and a graceful shutdown of the stopped process could actually happen (or at least start). To make it more obvious we're going to send a `SIGKILL`, we decided to just use `docker kill`.

pkg/ring/ring.go
pracucci: It depends.

Assuming RF=3, if `-distributor.shard-by-all-labels` is disabled, then this function is not used and we only query the 3 ingesters holding all the series for the queried metric (and we just need 2 success out of 3 queried ingesters).

However, if `-distributor.shard-by-all-labels` is enabled, we don't know which ingesters hold all the series for a given metric name, so we need to query all of them. When zone-awareness is disabled, we need to query all ingester minus 1, while when zone-awareness is enabled we need to query all ingesters in all zones minus 1 (thus tolerating all ingesters in 1 zone failing).

About shuffle sharding, this function is called after the subring has been selected.

Am I missing anything?

ranton256: Okay, I follow now, and am convinced the logic is correct, but maybe less enthusiastic about some of the implications, but since the value of  _all ingesters_ when -distributor.shard-by-all-labels is enabled is the subring seems more reasonable.

ranton256 (approved):

PR #3417, opened 2020-10-28
Cut release 1.5.0-rc.0

csmarchbanks:
The changelog was all in the correct order, and I did not see anything missing when from it.


jtlisi (approved):
LGTM

bboreham:
[I think you should have a PR open for the actual release but I don't see it so commenting here]

One thought: should we segregate the list by command?  In this case Cortex and Blocksconvert.
The latter is extremely niche so the list of enhancements is noise for most people.

csmarchbanks:
I will open another PR for the full release once the candidate has a few days at least to run in our clusters.

 >One thought: should we segregate the list by command?

I I think that is a reasonable idea, and I can make the change for the final release or next release candidate. We could also do a segmentation of Cortex vs tools if we didn't want multiple sections for blocksconvert, query-tee, etc...

PR #3419, opened 2020-10-29
Deploy when running github action on a tag

csmarchbanks:
When releasing 1.5.0-rc.0 the docker image is not being created, nor is the website updated. It appears that github actions is skipping the deploy job for tags, this tries to fix that.

I have no idea what I am doing with github actions, so please tell me if this is totally off base.

### DIFF
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -173,7 +173,7 @@ jobs:
 
   deploy_website:
     needs: [build, test]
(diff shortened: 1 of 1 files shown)


pracucci (approved):
I'm not a GitHub actions expert too, but looking at the GH actions doc your changes make sense to me 👍  Let's try!

csmarchbanks:
Well, it now deployed the image, but the tag is `HEAD-df64c73` rather than `v1.5.0-rc.0`. Investigating...

Issue #3348, opened 2020-10-14
/flush call on Cortex Blocks can result in overlapping blocks


codesome:
**Describe the bug**

`/flush` call on blocks based ingesters can cause creation of overlapping blocks.

This is because we call `CompactHead` for the entire Head block range which includes the valid time for samples to be ingested where TSDB doesn't safeguard much, hence some samples can slip in overlapping with the block being created parallely in the `/flush` call.

More info is in this issue https://github.com/prometheus/prometheus/issues/8055

**Storage Engine**
- [x] Blocks
- [ ] Chunks

pstibrany:
Fixed by #3422.

PR #3422, opened 2020-10-29
Fail ingestion when a flush is going on

### LINKED ISSUE #3282: Potential Prometheus memory leak in ingesters

codesome:
**What this PR does**:

Fails a push request with a 5xx when there is an ongoing flush. This is according to the idea proposed in this comment https://github.com/cortexproject/cortex/issues/3282#issuecomment-714285039

Fixes #3282 #3348

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci (commented):
Thanks @codesome about this! We had an offline discussion about a race condition. Ganesh mentioned he will work on this tomorrow. Thanks!

pkg/ingester/ingester_v2.go
```diff
@@ -93,6 +97,16 @@ func (u *userTSDB) Compact() error {
 
 // compactHead compacts the Head block at specified block durations avoiding a single huge block.
 func (u *userTSDB) compactHead(blockDuration int64) error {
+	u.forcedCompactionInProgressMtx.Lock()
+	u.forcedCompactionInProgress = true
+	u.forcedCompactionInProgressMtx.Unlock()
+	defer func() {
+		u.forcedCompactionInProgress = false
```
pracucci, line 104: You need to lock around this

pracucci (approved):
LGTM, but with a little fix, thanks!

pkg/ingester/ingester_v2.go
codesome: Instead of lock, what about using an atomic variable? Both would be fine actually.

codesome: Added the lock

pkg/ingester/ingester_v2_test.go
```diff
@@ -2422,3 +2422,39 @@ func TestIngesterNotDeleteUnshippedBlocks(t *testing.T) {
 		require.NotEqual(t, b.Meta().ULID, newBlocks2[1].Meta().ULID)
 	}
 }
+
+func TestIngesterPushErrorDuringForcedCompaction(t *testing.T) {
+	i, cleanup, err := newIngesterMockWithTSDBStorage(defaultIngesterTestConfig(), nil)
+	require.NoError(t, err)
+	t.Cleanup(cleanup)
```
ranton256, line 2429: I think ideally it would be good to also add a test that verifies compactHead waits on no pushes to be in flight and locks the muutex.

pracucci:
@pstibrany Me and @codesome are discussion a potentially better way to handle this. Please don't merge it you approve it.

pkg/ingester/ingester_v2.go
```diff
@@ -54,6 +54,10 @@ type userTSDB struct {
 	seriesInMetric *metricCounter
 	limiter        *Limiter
 
+	forcedCompactionInProgress    bool
+	forcedCompactionInProgressMtx sync.RWMutex
+	pushesInFlight                sync.WaitGroup
```
pstibrany, line 59: Mutex is typically on top of the block of fields that it "protects".

pstibrany (approved):
LGTM, thanks Ganesh!

pracucci:
> @pstibrany Me and @codesome are discussion a potentially better way to handle this. Please don't merge it you approve it.

@pstibrany Ignore this. After further discussion, Ganesh has no bandwidth to experiment alternative solutions (eg. disable blocks compaction at all in TSDB and enable overlapping blocks in TSB, making sure query performances don't degrade significantly on overlapping blocks), so I think we can get this PR in and then we may reconsider this approach if/when Ganesh will have bandwidth.

_So we can merge if you approve, but let's wait for Ganesh adding the test suggested by Richard first._

pkg/ingester/ingester_v2_test.go
codesome: I tried figuring out possible ways to test this, but since it uses a `sync.WaitGroup` for the in-flight requests and all the operations are relatively quick, I am not sure how to deterministically test for no in-flight requests. Do you have any ideas?

pstibrany: One possibility is to call `acquireAppendLock` directly from the test, then call compaction, verify that it updated bool hasn't started compaction, and then call `releaseAppendLock` to simulate finished push.

Verification of compaction could be done via TSDB metrics, or Head minT.

codesome: Nice one, I have added a test with that logic.

pkg/ingester/ingester_v2_test.go
```diff
@@ -2422,3 +2422,96 @@ func TestIngesterNotDeleteUnshippedBlocks(t *testing.T) {
...
+
+	require.NoError(t, services.StartAndAwaitRunning(context.Background(), i))
+	t.Cleanup(func() {
+		_ = services.StopAndAwaitTerminated(context.Background(), i)
+	})
+
+	// Wait until it's ACTIVE
+	test.Poll(t, 10*time.Millisecond, ring.ACTIVE, func() interface{} {
```
pracucci, line 2474: The 10millis is the timeout, not the poll frequency. Because of this, value should be higher. I would suggest at least 1s.

_This comment applies to the other test too._

pkg/ingester/ingester_v2_test.go
```diff
@@ -2422,3 +2422,96 @@ func TestIngesterNotDeleteUnshippedBlocks(t *testing.T) {
...
+	require.NoError(t, testutil.GatherAndCompare(registry, strings.NewReader(`
+		# HELP cortex_ingester_tsdb_compactions_total Total number of TSDB compactions that were executed.
+		# TYPE cortex_ingester_tsdb_compactions_total counter
+		cortex_ingester_tsdb_compactions_total 0
+	`), "cortex_ingester_tsdb_compactions_total"))
+
+	// No requests in flight after this.
+	db.releaseAppendLock()
+
+	// Let's wait for a moment for flush to finish, and then verify.
```
pracucci, line 2503:
```suggestion
	// Let's wait until all head series have been flushed.
```

pracucci (approved):
Very good job. LGTM, modulo a couple of nits. Thanks!

pkg/ingester/ingester_v2_test.go
```diff
@@ -2422,3 +2422,96 @@ func TestIngesterNotDeleteUnshippedBlocks(t *testing.T) {
...
+	})
+
+	// Push a sample, it should succeed.
+	pushSingleSample(t, i)
+
+	// We mock a flushing by setting the boolean.
+	db := i.getTSDB(userID)
+	require.NotNil(t, db)
+	require.False(t, db.forcedCompactionInProgress)
+	db.forcedCompactionInProgress = true
```
pstibrany, line 2448: To avoid -race reports, I think we should use mutex around the check and setting.

pkg/ingester/ingester_v2_test.go
```diff
@@ -2422,3 +2422,96 @@ func TestIngesterNotDeleteUnshippedBlocks(t *testing.T) {
...
+
+	// Ingestion should fail with a 503.
+	req, _, _ := mockWriteRequest(labels.Labels{{Name: labels.MetricName, Value: "test"}}, 0, util.TimeToMillis(time.Now()))
+	ctx := user.InjectOrgID(context.Background(), userID)
+	_, err = i.v2Push(ctx, req)
+	require.Equal(t, httpgrpc.Errorf(http.StatusServiceUnavailable, wrapWithUser(errors.New("forced compaction in progress"), userID).Error()), err)
+
+	// Ingestion is successful after a flush.
+	require.True(t, db.forcedCompactionInProgress)
+	db.forcedCompactionInProgress = false
```
pstibrany, line 2458: To avoid -race reports, I think we should use mutex around the check and setting the value.

pstibrany (approved):
Thanks. I think there may be race errors in the test, but otherwise LGTM.

PR #3425, opened 2020-10-30
Intern Prometheus MultiError to solve blocking circular dependency between Cortex and Thanos

pracucci:
**What this PR does**:
Cortex and Thanos vendoring are a circular dependency. Cortex depends on Thanos (for blocks storage) and Thanos depends on Cortex (for query-frontend). This is currently blocking Prometheus upgrade in both projects.

The reason is that Prometheus `MultiError` has been refactored. We can't upgrade Prometheus in Cortex because it breaks Thanos, but can't upgrade Prometheus in Thanos too because breaks Cortex. In this PR I'm interning the `MultiError` into Cortex to unblock this annoying situation and being able to upgrade Prometheus in Thanos first, and then upgrade Prometheus + Thanos in Cortex.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,7 @@ lint:
 	# Ensure no blacklisted package is imported.
 	faillint -paths "github.com/bmizerany/assert=github.com/stretchr/testify/assert,\
 		golang.org/x/net/context=context,\
+		github.com/prometheus/prometheus/tsdb/errors=util,\
 		sync/atomic=go.uber.org/atomic" ./pkg/... ./cmd/... ./tools/... ./integration/...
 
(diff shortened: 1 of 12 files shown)


gouthamve (approved):
If I understand correctly we just copied the multierror package into Cortex, and we'll revert back to the prometheus errors package once we vendor Thanos + Prometheus?

pracucci:
> If I understand correctly we just copied the multierror package into Cortex, and we'll revert back to the prometheus errors package once we vendor Thanos + Prometheus?

Correct

PR #3426, opened 2020-10-30
Upgrade Prometheus and Thanos

pracucci:
**What this PR does**:
In this PR I'm upgrading Prometheus and Thanos, in order to backport the TSDB WAL corruption (partial) fix done in https://github.com/prometheus/prometheus/pull/8125.

In this PR I'm also removing the temporarily interned `MultiError` (https://github.com/cortexproject/cortex/pull/3425) which was done as a trick to avoid a circular breaking change between Cortex and Thanos, blocking both projects from upgrading Prometheus.

I've checked Thanos changes and there's nothing that we should change on our side (eg. no new metrics to map).

/cc @codesome 

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/table_manager.go
+++ b/pkg/chunk/table_manager.go
@@ -14,6 +14,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
+	tsdberrors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/weaveworks/common/instrument"
 	"github.com/weaveworks/common/mtime"
 
@@ -469,7 +470,7 @@ func (m *TableManager) partitionTables(ctx context.Context, descriptions []Table
 
 func (m *TableManager) createTables(ctx context.Context, descriptions []TableDesc) error {
 	numFailures := 0
-	merr := util.NewMultiError()
+	merr := tsdberrors.NewMulti()
 
 	for _, desc := range descriptions {
 		level.Info(util.Logger).Log("msg", "creating table", "table", desc.Name)
@@ -486,7 +487,7 @@ func (m *TableManager) createTables(ctx context.Context, descriptions []TableDes
 
 func (m *TableManager) deleteTables(ctx context.Context, descriptions []TableDesc) error {
 	numFailures := 0
-	merr := util.NewMultiError()
+	merr := tsdberrors.NewMulti()
 
 	for _, desc := range descriptions {
 		level.Info(util.Logger).Log("msg", "table has exceeded the retention period", "table", desc.Name)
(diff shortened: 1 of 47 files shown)


pkg/chunk/table_manager.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
+	tsdberrors "github.com/prometheus/prometheus/tsdb/errors"
```
codesome, line 17: To be consistent 
```suggestion
	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
```

pkg/ingester/wal.go
```diff
@@ -794,7 +795,8 @@ func processWALWithRepair(startSegment int, userStates *userStates, params walRe
 	if err != nil {
 		level.Error(util.Logger).Log("msg", "error in repairing WAL", "err", err)
 	}
-	multiErr := util.NewMultiError(err)
+	multiErr := tsdb_errors.NewMulti()
+	multiErr.Add(err)
 	multiErr.Add(w.Close())
 
 	return multiErr.Err()
```
codesome, line 802: All this can be just `tsdb_errors.NewMulti(err, w.Close()).Err()`, but maybe that can be done in follow up PRs and not required here.

codesome (approved):

pkg/chunk/table_manager.go
pracucci: Good point, done!

pkg/ingester/wal.go
pracucci: Yes, done. Generally speaking I'm not a big fan of such contracted versions (to me it's less readable) but this is your WAL, so your pick ;)

vendor/github.com/thanos-io/thanos/pkg/compact/compact.go
```diff
@@ -515,7 +516,7 @@ func (e HaltError) Error() string {
 // IsHaltError returns true if the base error is a HaltError.
 // If a multierror is passed, any halt error will return true.
 func IsHaltError(err error) bool {
-	if multiErr, ok := errors.Cause(err).(terrors.MultiError); ok {
+	if multiErr, ok := errors.Cause(err).(errutil.MultiError); ok {
```
pstibrany, line 519: `terrors` :joy:

pstibrany (approved):
LGTM

vendor/github.com/thanos-io/thanos/pkg/compact/compact.go
pracucci: Ahah... yes, was a misleading name :D

PR #3429, opened 2020-10-30
Add integration test on query-scheduler

pracucci:
**What this PR does**:
In this PR I've added integration tests for the query-scheduler. The diff will be easier to review if you enable "hide whitespace changes".

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/frontend2/querier_scheduler_worker.go
+++ b/pkg/querier/frontend2/querier_scheduler_worker.go
@@ -83,8 +83,8 @@ func NewQuerierSchedulerWorkers(cfg QuerierWorkersConfig, handler RequestHandler
 	}
 
 	frontendClientsGauge := promauto.With(reg).NewGauge(prometheus.GaugeOpts{
-		Name: "cortex_query_scheduler_worker_frontend_clients",
-		Help: "The current number of frontend clients.",
+		Name: "cortex_querier_query_frontend_clients",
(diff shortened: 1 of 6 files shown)


pkg/querier/frontend2/querier_scheduler_worker.go
```diff
@@ -83,8 +83,8 @@ func NewQuerierSchedulerWorkers(cfg QuerierWorkersConfig, handler RequestHandler
 	}
 
 	frontendClientsGauge := promauto.With(reg).NewGauge(prometheus.GaugeOpts{
-		Name: "cortex_query_scheduler_worker_frontend_clients",
-		Help: "The current number of frontend clients.",
+		Name: "cortex_querier_query_frontend_clients",
```
pracucci, line 86: I would prefer it `cortex_query_scheduler` metrics prefix are reserved for the scheduler. Thoughts?

pracucci:
Flaky test should be fixed by this PR: https://github.com/cortexproject/cortex/pull/3434

pkg/querier/frontend2/querier_scheduler_worker.go
pstibrany: :+1: Good idea, thanks for spotting.

pstibrany (approved):
Great job, thank you very much!

PR #3431, opened 2020-10-30
Ensure grpc keepalive on all clients

pracucci:
**What this PR does**:
We're not enabling gRPC keepalive for inter-service gRPC connections 😱  I guess it's time to do it. A sane hardcoded default should be fine for all cases. The gRPC storage plugin also had a timeout of 10m which looks a mistake.

_New gRPC clients introduced for the query-scheduler (https://github.com/cortexproject/cortex/pull/3374) will benefit this too._

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   - `cortex_ingester_tsdb_head_truncations_failed_total`
   - `cortex_ingester_tsdb_head_truncations_total`
   - `cortex_ingester_tsdb_head_gc_duration_seconds`
+* [ENHANCEMENT] Enforced keepalive on all gRPC clients used for inter-service communication. #3431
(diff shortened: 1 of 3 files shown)


jtlisi (approved):
LGTM

pstibrany (approved):

pstibrany:
These settings are too low, and will cause early disconnect of clients on idle connections.

Cortex gRPC server uses 5 minutes as minimum time between pings. If client sends pings more often, it gets "ping strike". When client reaches 3 "ping strikes", it is disconnected. With 20s keepalive ping, this happens after 1 minute.

https://github.com/weaveworks/common/pull/205 will allow to configure lower settings on the gRPC server.

pstibrany:
Ping strikes are implemented in https://github.com/cortexproject/cortex/blob/8cfa79c9265492fbd78a3d89d29e526d35e3db9b/vendor/google.golang.org/grpc/internal/transport/http2_server.go#L672, with hard-coded max value of 2.

PR #3433, opened 2020-10-30
Moved query-scheduler to pkg/scheduler

pracucci:
**What this PR does**:
This is a refactoring PR to move the query-scheduler to `pkg/scheduler` and avoid cross references between `scheduler` and `frontend2` package. I followed this design principle:

- Scheduler moved to `pkg/scheduler`
- Frontend2 is still in `pkg/querier/frontend2` (out of scope, its refactoring will be done in a separate PR)
- Scheduler proto moved to `pkg/scheduler/schedulerpb` (similarly to what we did in store-gateway)
- Faillint guarantees frontend2 can't reference `pkg/scheduler`
- Failling guarantees `pkg/scheduler` can't reference `pkg/querier/...`

The whole refactoring has been done using Goland refactoring tool. No logic has been touched.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


Makefile
```diff
@@ -150,8 +150,12 @@ lint:
 		github.com/prometheus/prometheus/tsdb/errors=util,\
 		sync/atomic=go.uber.org/atomic" ./pkg/... ./cmd/... ./tools/... ./integration/...
 
+	# Ensure clean pkg structure.
+	faillint -paths "github.com/cortexproject/cortex/pkg/scheduler"   ./pkg/querier/...
+	faillint -paths "github.com/cortexproject/cortex/pkg/querier/..." ./pkg/scheduler/...
+
 	# Validate Kubernetes spec files. Requires:
-	# https://kubeval.instrumenta.dev
+	# https://kubeval.instrumenta.dev
```
pracucci, line 158: There was a non-breaking space

pkg/querier/frontend2/frontend.proto
```diff
@@ -10,36 +10,6 @@ import "github.com/weaveworks/common/httpgrpc/httpgrpc.proto";
...
-    httpgrpc.HTTPRequest httpRequest = 2;
-
-    // Where should querier send HTTP Response to (using FrontendForQuerier interface).
-    string frontendAddress = 3;
-
-    // User who initiated the request. Needed to send reply back to frontend.
-    string userID = 4;
-}
-
 // Frontend interface exposed to Queriers. Used by queriers to report back the result of the query.
```
pstibrany, line 13: I think it's little unfortunate that frontend.proto is now split into two files – I considered it a plus that single proto file defined the entire workflow.

pstibrany (approved):
Nice, thank you!

pkg/querier/frontend2/frontend.proto
pracucci: I personally believe every Cortex service should expose its own proto. This way, you can have a clear separation of concerns: "which service exposes what". I don't think the architecture is more clear to understand just because all RPC are defined in a single file. On the contrary, if each service exposes its own `.proto` you can have a better understanding of which service exposes a specific RPC. Overall architecture and workflow should be documented elsewhere IMO, and definitely not in a protobuf-style (better to explain it in plain english, with diagrams like you did in the query-scheduler PR description).

PR #3434, opened 2020-10-30
Fixed TestSchedulerEnqueueWithFrontendDisconnect flakyness

pracucci:
**What this PR does**:
The test `TestSchedulerEnqueueWithFrontendDisconnect` is flaky ([CI failed here](https://github.com/cortexproject/cortex/runs/1332670597?check_suite_focus=true)). The reason is that there's no guarantee that the query-frontend disconnected from the scheduler and its request canceled before `verifyQuerierDoesntReceiveRequest` is called. The change done in this test should help to reduce this likelihood.

I also took the opportunity to:

1. Rename the metric `cortex_query_scheduler_connected_workers` to `cortex_query_scheduler_connected_querier_clients`, because:
  - Now it's more clear which workers we're talking about
  - In the frontend v1 the metric is called `_clients`, so I've renamed for consistency
2. Add the metric `cortex_query_scheduler_connected_frontend_clients`

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/frontend2/scheduler.go
+++ b/pkg/querier/frontend2/scheduler.go
@@ -42,9 +42,10 @@ type Scheduler struct {
 	pendingRequests map[requestKey]*schedulerRequest // Request is kept in this map even after being dispatched to querier. It can still be canceled at that time.
 
 	// Metrics.
-	connectedWorkers prometheus.GaugeFunc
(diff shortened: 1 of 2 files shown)


pstibrany (approved):
LGTM, thanks!

PR #3436, opened 2020-10-30
fix: register querier routes explicitly with handler for instrumentation

### LINKED ISSUE #3432: cortex_querier_.* metrics no longer differentiates api/prom/v1 routes
After deploying 1.5.0-rc.0 I noticed that some of our dashboards broke, upon digging in I noticed that the `cortex_querier_.*` metrics truncate routes to just `api_prom_api_v1` rather than e.g. `api_prom_api_v1_query`. Metadata is still split into `api_prom_api_v1_metadata`, but not query, query_range, series, etc...

I think this might be related to https://github.com/cortexproject/cortex/pull/3320, but didn't have time yet to fully dig in.

@jtlisi would you be able to see if it is related to #3320?

jtlisi:
**What this PR does**:

- Registers the Querier internal routes explicitly to ensure the instrumentation utilizes the correct route names

Fixes #3432 

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,7 +92,6 @@
 * [ENHANCEMENT] Ruler: Include the tenant ID on the notifier logs. #3372
 * [ENHANCEMENT] Blocks storage Compactor: Added `-compactor.enabled-tenants` and `-compactor.disabled-tenants` to explicitly enable or disable compaction of specific tenants. #3385
 * [ENHANCEMENT] Blocks storage ingester: Creating checkpoint only once even when there are multiple Head compactions in a single `Compact()` call. #3373
-* [BUGFIX] Querier API: Fix issues #3432 where `cortex_querier_request_` metrics where truncated routes. #3436
 * [BUGFIX] Blocks storage ingester: Read repair memory-mapped chunks file which can end up being empty on abrupt shutdowns combined with faulty disks. #3373
 * [BUGFIX] Blocks storage ingester: Close TSDB resources on failed startup preventing ingester OOMing. #3373
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@
 * [ENHANCEMENT] Ruler: Include the tenant ID on the notifier logs. #3372
 * [ENHANCEMENT] Blocks storage Compactor: Added `-compactor.enabled-tenants` and `-compactor.disabled-tenants` to explicitly enable or disable compaction of specific tenants. #3385
 * [ENHANCEMENT] Blocks storage ingester: Creating checkpoint only once even when there are multiple Head compactions in a single `Compact()` call. #3373
+* [BUGFIX] Querier API: Fix issues #3432 where `cortex_querier_request_` metrics where truncated routes. #3436
 * [BUGFIX] Blocks storage ingester: Read repair memory-mapped chunks file which can end up being empty on abrupt shutdowns combined with faulty disks. #3373
 * [BUGFIX] Blocks storage ingester: Close TSDB resources on failed startup preventing ingester OOMing. #3373
 * [BUGFIX] No-longer-needed ingester operations for queries triggered by queriers and rulers are now canceled. #3178
--- a/pkg/api/handlers.go
+++ b/pkg/api/handlers.go
@@ -213,19 +213,29 @@ func NewQuerierHandler(
 	legacyPromRouter := route.New().WithPrefix(cfg.ServerPrefix + cfg.LegacyHTTPPrefix + "/api/v1")
 	api.Register(legacyPromRouter)
 
-	//TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
+	// TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
 	// https://github.com/prometheus/prometheus/pull/7125/files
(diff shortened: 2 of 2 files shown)


pkg/api/handlers.go
```diff
@@ -213,19 +213,29 @@ func NewQuerierHandler(
...
+	router.Path(cfg.PrometheusHTTPPrefix+"/api/v1/series").Methods("GET", "POST", "DELETE").Handler(promRouter)
+	router.Path(cfg.PrometheusHTTPPrefix + "/api/v1/metadata").Methods("GET").Handler(promRouter)
+
+	// TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
 	// https://github.com/prometheus/prometheus/pull/7125/files
 	router.Path(cfg.LegacyHTTPPrefix + "/api/v1/metadata").Handler(querier.MetadataHandler(distributor))
 	router.Path(cfg.LegacyHTTPPrefix + "/api/v1/read").Handler(querier.RemoteReadHandler(queryable))
-	// A prefix is fine because external routes will be registered explicitly
-	router.PathPrefix(cfg.LegacyHTTPPrefix + "/api/v1/").Handler(legacyPromRouter)
+	router.Path(cfg.PrometheusHTTPPrefix + "/api/v1/read").Methods("POST").Handler(legacyPromRouter)
```
csmarchbanks, line 232: Should these also be `LegacyHTTPPrefix`? Looks like you already register them for `PrometheusHTTPPrefix` above.

jtlisi: Yes, good catch

pkg/api/handlers.go
```diff
@@ -213,19 +213,29 @@ func NewQuerierHandler(
...
-	//TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
+	// TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
 	// https://github.com/prometheus/prometheus/pull/7125/files
 	router.Path(cfg.PrometheusHTTPPrefix + "/api/v1/metadata").Handler(querier.MetadataHandler(distributor))
 	router.Path(cfg.PrometheusHTTPPrefix + "/api/v1/read").Handler(querier.RemoteReadHandler(queryable))
-	// A prefix is fine because external routes will be registered explicitly
-	router.PathPrefix(cfg.PrometheusHTTPPrefix + "/api/v1/").Handler(promRouter)
-
-	//TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
+	router.Path(cfg.PrometheusHTTPPrefix + "/api/v1/read").Methods("POST").Handler(promRouter)
```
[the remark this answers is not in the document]
jtlisi: They shouldn't interact. RegisterQueryAPI registers the route with the router embedded in the weaveworks/common server. This router is registered as a handler to the weaveworks/common router. The only way this routing could break is if a middleware in the weaveworks common server rewrites the URL or method of a request, and then that request is passed to this routers `ServeHTTP` function. I don't think anything is in place that could do this.

CHANGELOG.md
```diff
@@ -92,6 +92,7 @@
 * [ENHANCEMENT] Ruler: Include the tenant ID on the notifier logs. #3372
 * [ENHANCEMENT] Blocks storage Compactor: Added `-compactor.enabled-tenants` and `-compactor.disabled-tenants` to explicitly enable or disable compaction of specific tenants. #3385
 * [ENHANCEMENT] Blocks storage ingester: Creating checkpoint only once even when there are multiple Head compactions in a single `Compact()` call. #3373
+* [BUGFIX] Querier API: Fix issues #3432 where `cortex_querier_request_` metrics where truncated routes. #3436
```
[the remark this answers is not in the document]
jtlisi: Removed

pkg/api/handlers.go
jtlisi: I think this should be sorted.

pkg/api/handlers.go
```diff
@@ -207,25 +207,39 @@ func NewQuerierHandler(
...
-	router.Path(cfg.PrometheusHTTPPrefix + "/api/v1/metadata").Handler(querier.MetadataHandler(distributor))
-	router.Path(cfg.PrometheusHTTPPrefix + "/api/v1/read").Handler(querier.RemoteReadHandler(queryable))
-	// A prefix is fine because external routes will be registered explicitly
-	router.PathPrefix(cfg.PrometheusHTTPPrefix + "/api/v1/").Handler(promRouter)
-
-	//TODO(gotjosh): This custom handler is temporary until we're able to vendor the changes in:
+	router.Path(prefix + "/api/v1/metadata").Handler(querier.MetadataHandler(distributor))
+	router.Path(prefix + "/api/v1/read").Handler(querier.RemoteReadHandler(queryable))
+	router.Path(prefix + "/api/v1/read").Methods("POST").Handler(promRouter)
+	router.Path(prefix+"/api/v1/query").Methods("GET", "POST").Handler(promRouter)
```
csmarchbanks, line 225: Nit: could we have spaces around each `+` to let the eye quickly follow the pattern?

csmarchbanks (commented):
One nit, but LGTM. I will let Marco approve as well before merging though.

Generally my plan is to release another release candidate early next week pending any more bug reports.

pkg/api/handlers.go
jtlisi: That bothered me as well. I tried to fix it, but the`gofmt` lint complains for some reason.

csmarchbanks: :cry: one of the first times I am not ok with gofmt. Thanks for trying!

jtlisi: Same haha

csmarchbanks (approved):

PR #3439, opened 2020-10-30
chore: ensure internalQuerier utilizes identical authMiddleware to API

jtlisi:
**What this PR does**:

This PR ensures the authentication middleware used by the internalQuerier is the same that is used by the API struct. I was worried this might be causing an issue with queries for single binary users with `no org id` errors, however, I manually verified it is not. Currently, this isn't causing an issue because the query-frontend will create the `x-scope-orgid` header with `fake` as a value so the `AuthenticateUser` middleware is valid. However, this will ensure any changes made to the auth middleware used by the API in the future will also be added to this internalHandler..

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -56,12 +56,13 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 }
 
 type API struct {
-	cfg            Config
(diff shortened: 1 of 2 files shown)


jtlisi:
Moved this PR into #3436

pracucci (approved):
LGTM, thanks!

PR #3440, opened 2020-11-02
Fix race in TestFrontend_LogsSlowQueriesFormValues.

pstibrany:
**What this PR does**: This PR fixes race in `TestFrontend_LogsSlowQueriesFormValues`. Race is between `buf.String()` and writes to the buf from other goroutines. ([CI Failure](https://github.com/cortexproject/cortex/runs/1340754022?check_suite_focus=true))

**Checklist**
- [x] Tests updated


pracucci (approved):
LGTM, thanks!

PR #3443, opened 2020-11-02
Remove dummy credentials from AWS config

somdoron:
**What this PR does**:

the current dummy credentials deny using EKS IAM service accounts, removing this line fallback to the AWS SDK default which is the following:

* Environment Variables
* Shared Credentials file
* Shared Configuration file (if SharedConfig is enabled)
* EC2 Instance Metadata (credentials only)

reference: https://docs.aws.amazon.com/sdk-for-go/api/aws/session/#NewSession

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/aws/s3_storage_client.go
+++ b/pkg/chunk/aws/s3_storage_client.go
@@ -137,7 +137,6 @@ func buildS3Config(cfg S3Config) (*aws.Config, []string, error) {
 	} else {
 		s3Config = &aws.Config{}
 		s3Config = s3Config.WithRegion("dummy")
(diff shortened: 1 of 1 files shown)


pracucci:
Thanks @somdoron. Pinging @bboreham and @gouthamve about this.

pracucci (approved):
Talked to @joe-elliott who did this initially and they agree with this change, so LGTM. Thanks!

pstibrany (commented):
🙈 Approving given the Joe's approval.

pstibrany (approved):

PR #3444, opened 2020-11-02
Move query-frontend to pkg/frontend

pracucci:
**What this PR does**:
This is a pkg refactoring PR. Cortex logic, behaviour or tests haven't changed. This is a first PR to move the query-frontend to its own package, separating the logic of the query-frontend and the "frontend worker" which is actually running in the querier. This refactoring (once completed) should help to separate the logic and also reduce the dependencies imported by Thanos.

Out of the scope of this PR:

- Refactor packages used by query sharding (astmapper, queryrange) otherwise the diff would be too large and hard to review. Will attempt it in a following PR, if this PR gets accepted.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany:
I wonder if we should just use single `querier/frontend`, with both worker implementations. There is a lot of copied code between the two, and I plan to refactor it to clean it up. (We can also merge the two packages at that time).

pstibrany (approved):
LGTM, good job!

pracucci:
> I wonder if we should just use single `querier/frontend`, with both worker implementations. There is a lot of copied code between the two, and I plan to refactor it to clean it up. (We can also merge the two packages at that time).

I think so, but I would do it in a separate PR. In this PR I just tried to refactor query-fronted pkg, not the querier's one.

pracucci:
> > I wonder if we should just use single `querier/frontend`, with both worker implementations. There is a lot of copied code between the two, and I plan to refactor it to clean it up. (We can also merge the two packages at that time).
> 
> I think so, but I would do it in a separate PR. In this PR I just tried to refactor query-fronted pkg, not the querier's one.

For the records, I wouldn't even call the package "frontend". Maybe `pkg/querier/worker` could be more appropriate.

bboreham:
Our staging service broke when the querier from this PR auto-deployed.
Queues filled up and everything 429'd.
I have rolled back to 1.5.0-rc.0.

pracucci:
> Our staging service broke when the querier from this PR auto-deployed.
> Queues filled up and everything 499'd.
> I have rolled back to 1.5.0-rc.0.

Sorry @bboreham for that. Will look into it shortly and I will follow up as soon as I understand what's wrong.

pracucci:
> > Our staging service broke when the querier from this PR auto-deployed.
> > Queues filled up and everything 499'd.
> > I have rolled back to 1.5.0-rc.0.
> 
> Sorry @bboreham for that. Will look into it shortly and I will follow up as soon as I understand what's wrong.

This PR https://github.com/cortexproject/cortex/pull/3454 should fix it.

PR #3446, opened 2020-11-02
Cut release 1.5.0-rc.1

csmarchbanks:
Cut the next release candidate. I also split the changelog into groups by binary rather than listing all of the changes together.

cc: @bboreham let me know what you think of the changelog split.


pracucci (approved):
LGTM, thanks!

pstibrany (approved):

bboreham (commented):
Yes the split of blocksconvert is an improvement.

Shouldn't the release notes say what is different between .0 and .1? (#3320, which is a fix to an unanounced change, if I am reading things correctly)

csmarchbanks:
> Shouldn't the release notes say what is different between .0 and .1?

We had decided it didn't need to be in the official release notes since the bug was never released, I hesitate to add it to the CHANGELOG, then remove it when merging everything back into master, I can write a note at the top of the Github release for what is different, does that sound ok?

I have added a second commit to this PR now, a cherry pick of https://github.com/cortexproject/cortex/pull/3439 to avoid a regression. I would appreciate another look!

pracucci (approved):
LGTM, thanks! Just run `make clean-white-noise` to make the linter happy.

bboreham (approved):

PR #3447, opened 2020-11-02
fix: register querier routes explicitly with handler for instrumentation

jtlisi:
**What this PR does**:

Port #3436 to master


pracucci (approved):
Thanks!

PR #3452, opened 2020-11-03
Added -querier.max-query-lookback and fixed -querier.max-query-into-future

pracucci:
**What this PR does**:
Currently, the only option to limit the time range of queries is `-store.max-look-back-period`, but it suffers the following issues:

1. Applies to chunks storage only
2. Applies to instance/range queries only (not a strict requirement today, but we'll soon work on querying series/labels from storage in the blocks storage)
3. Can't be overridden on a per-tenant basis

In this PR I'm adding `-querier.max-query-lookback` which is expected is superseed `-store.max-look-back-period` solving the issues above. While working on tests, I've also noticed that `-querier.max-query-into-future` is currently broken (doesn't really enforce it), so I've fixed it in this PR.

Out of the scope of this PR:
- Enforcing the limit in the query-frontend too (will be done in a separate PR)

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -74,7 +74,7 @@ func (cfg *StoreConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.WriteDedupeCacheConfig.RegisterFlagsWithPrefix("store.index-cache-write.", "Cache config for index entry writing. ", f)
 
 	f.Var(&cfg.CacheLookupsOlderThan, "store.cache-lookups-older-than", "Cache index entries older than this period. 0 to disable.")
-	f.Var(&cfg.MaxLookBackPeriod, "store.max-look-back-period", "Limit how long back data can be queried")
+	f.Var(&cfg.MaxLookBackPeriod, "store.max-look-back-period", "Deprecated: use -querier.max-query-lookback instead. Limit how long back data can be queried. This setting applies to chunks storage only.") // To be removed in Cortex 1.8.
 }
 
 // Validate validates the store config.
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -18,6 +19,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/chunk/cache"
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/extract"
+	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/spanlogger"
 	"github.com/cortexproject/cortex/pkg/util/validation"
 )
@@ -78,7 +80,12 @@ func (cfg *StoreConfig) RegisterFlags(f *flag.FlagSet) {
 }
 
 // Validate validates the store config.
-func (cfg *StoreConfig) Validate() error {
+func (cfg *StoreConfig) Validate(logger log.Logger) error {
+	if cfg.MaxLookBackPeriod > 0 {
+		flagext.DeprecatedFlagsUsed.Inc()
+		level.Warn(logger).Log("msg", "running with DEPRECATED flag -store.max-look-back-period, use -querier.max-query-lookback instead.")
+	}
+
 	if err := cfg.ChunkCacheConfig.Validate(); err != nil {
 		return err
 	}
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -218,13 +218,13 @@ func NewQueryable(distributor QueryableWithFilter, stores []QueryableWithFilter,
 		}
 
 		q := querier{
-			cfg:              cfg,
-			ctx:              ctx,
-			mint:             mint,
-			maxt:             maxt,
-			chunkIterFn:      chunkIterFn,
-			tombstonesLoader: tombstonesLoader,
-			limits:           limits,
+			ctx:                ctx,
+			mint:               mint,
+			maxt:               maxt,
+			chunkIterFn:        chunkIterFn,
+			tombstonesLoader:   tombstonesLoader,
+			limits:             limits,
+			maxQueryIntoFuture: cfg.MaxQueryIntoFuture,
 		}
 
 		dqr, err := distributor.Querier(ctx, mint, maxt)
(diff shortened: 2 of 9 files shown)


pkg/querier/querier.go
```diff
@@ -204,18 +205,20 @@ func NewQueryable(distributor QueryableWithFilter, stores []QueryableWithFilter,
 	return storage.QueryableFunc(func(ctx context.Context, mint, maxt int64) (storage.Querier, error) {
 		now := time.Now()
 
-		if cfg.MaxQueryIntoFuture > 0 {
```
pracucci, line 207: Moved to `validateQueryTimeRange()`, which is also called in the `Select()`. If this limit is not enforced in the `Select()` too, we're not really enforcing it. Unit tests were weak: I spotted this issue improving unit tests.

pkg/chunk/chunk_store.go
```diff
@@ -74,7 +74,7 @@ func (cfg *StoreConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.WriteDedupeCacheConfig.RegisterFlagsWithPrefix("store.index-cache-write.", "Cache config for index entry writing. ", f)
 
 	f.Var(&cfg.CacheLookupsOlderThan, "store.cache-lookups-older-than", "Cache index entries older than this period. 0 to disable.")
-	f.Var(&cfg.MaxLookBackPeriod, "store.max-look-back-period", "Limit how long back data can be queried")
+	f.Var(&cfg.MaxLookBackPeriod, "store.max-look-back-period", "Deprecated: use -querier.max-query-lookback instead. Limit how long back data can be queried. This setting applies to chunks storage only.") // To be removed in Cortex 1.8.
 }
 
 // Validate validates the store config.
```
gouthamve, line 80: Can we warn users if this is set in validate?

```
flagext.DeprecatedFlagsUsed.Inc()
level.Warn(logger).Log("msg", "running with DEPRECATED ..... use ... instead")
```

pkg/querier/querier.go
gouthamve: Is this because `Select` uses hints?

gouthamve (approved):
LGTM with nits :)

pkg/querier/querier.go
pracucci: Exactly.

pkg/querier/querier.go
```diff
@@ -475,3 +499,41 @@ func UseBeforeTimestampQueryable(queryable storage.Queryable, ts time.Time) Quer
...
+	endTime := model.Time(endMs)
+
+	// Clamp time range based on max query into future.
+	if maxQueryIntoFuture > 0 && endTime.After(now.Add(maxQueryIntoFuture)) {
+		origEndTime := endTime
+		endTime = now.Add(maxQueryIntoFuture)
+
+		// Make sure to log it in traces to ease debugging.
+		level.Debug(spanlogger.FromContext(ctx)).Log(
+			"msg", "the start time of the query has been manipulated because of the 'max query into future' setting",
```
pstibrany, line 515:
```suggestion
			"msg", "the end time of the query has been manipulated because of the 'max query into future' setting",
```

pkg/querier/querier.go
```diff
@@ -475,3 +499,41 @@ func UseBeforeTimestampQueryable(queryable storage.Queryable, ts time.Time) Quer
...
+			"original", origEndTime, "updated", endTime)
+
+		if endTime.Before(startTime) {
+			return 0, 0, errEmptyTimeRange
+		}
```
pstibrany, line 520: This check is repeated in both `if` conditions. Should we do it only once, after all manipulations?

pkg/querier/querier.go
```diff
@@ -253,6 +256,8 @@ func NewQueryable(distributor QueryableWithFilter, stores []QueryableWithFilter,
 }
 
 type querier struct {
+	cfg Config
```
pstibrany, line 259: I would prefer to avoid depending on entire Config struct, and rather be explicit about what querier actually uses (`MaxQueryIntoFuture` only for now).

pkg/querier/querier_test.go
```diff
@@ -473,6 +474,214 @@ func TestQuerier_ValidateQueryTimeRange(t *testing.T) {
 	}
 }
 
+func TestQuerier_ValidateQueryTimeRange_MaxQueryLookback(t *testing.T) {
+	const (
+		engineLookbackDelta = 5 * time.Minute
+		month               = 30 * 24 * time.Hour
```
pstibrany, line 480: I would call this differently (`thirtyDays`?) as month lengths vary.

pkg/util/validation/limits.go
```diff
@@ -122,6 +123,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 
 	f.IntVar(&l.MaxChunksPerQuery, "store.query-chunk-limit", 2e6, "Maximum number of chunks that can be fetched in a single query. This limit is enforced when fetching chunks from the long-term storage. When running the Cortex chunks storage, this limit is enforced in the querier, while when running the Cortex blocks storage this limit is both enforced in the querier and store-gateway. 0 to disable.")
 	f.DurationVar(&l.MaxQueryLength, "store.max-query-length", 0, "Limit the query time range (end - start time). This limit is enforced in the query-frontend (on the received query), in the querier (on the query possibly split by the query-frontend) and in the chunks storage. 0 to disable.")
+	f.DurationVar(&l.MaxQueryLookback, "querier.max-query-lookback", 0, "Limit how long back data (series and metadata) can be queried, up until <lookback> time ago. 0 to disable.")
```
pstibrany, line 126:
```suggestion
	f.DurationVar(&l.MaxQueryLookback, "querier.max-query-lookback", 0, "Limit how long back data (series and metadata) can be queried, up until <lookback> duration ago. 0 to disable.")
```

pstibrany (approved):
LGTM, with some nits.

pkg/querier/querier.go
pracucci: I think it's more future proof doing it in both places. Doesn't look an overcomplication.

pracucci:
Thanks @pstibrany for your thoughtful review. I've addressed your comments.

pkg/querier/querier.go
pstibrany: > I think it's more future proof doing it in both places

I don't quite understand what you mean by "future proof" here. (It's not overcomplication, but looks like unnecessary duplication. Plus it would cover case when input start time is already after end time).

pracucci: As of today, we can remove the `if` and move it to the end because the two validations are done one on start time and the other one on end time. If in the future we add more validation (either on start or end), we may end up in bugs which we don't realise because of an invalid time range (start > end) introduced by a previous validation.

pstibrany: I see, although I think this check is pretty universal, and hopefully unit tests will warn us about newly introduced bugs. Thanks for explaining your thinking.

ranton256 (approved):
LGTM

PR #3455, opened 2020-11-03
Querier workers refactor, step 1

pstibrany:
This is first step in querier workers refactor. There are two workers (one using frontend, one using scheduler) with lot of copied code. End-goal is to refactor common parts.

First step is only moving `querier_scheduler_worker.go` to `pkg/querier/frontend` package, and then renaming `pkg/querier/frontend` to `pkg/querier/worker`, all done via Goland refactoring. No change in logic was done.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/frontend/frontend_test.go
+++ b/pkg/frontend/frontend_test.go
@@ -29,7 +29,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/frontend/transport"
 	"github.com/cortexproject/cortex/pkg/frontend/v1/frontendv1pb"
 	"github.com/cortexproject/cortex/pkg/querier"
-	"github.com/cortexproject/cortex/pkg/querier/worker"
+	querier_worker "github.com/cortexproject/cortex/pkg/querier/worker"
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/services"
 )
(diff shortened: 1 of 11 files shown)


pkg/frontend/frontend_test.go
```diff
@@ -282,14 +282,14 @@ func testFrontend(t *testing.T, config CombinedFrontendConfig, handler http.Hand
 	go httpServer.Serve(httpListen) //nolint:errcheck
 	go grpcServer.Serve(grpcListen) //nolint:errcheck
 
-	var worker services.Service
-	worker, err = frontend.NewWorker(workerConfig, querierConfig, httpgrpc_server.NewServer(handler), logger)
+	var wrkr services.Service
```
pracucci, line 285: If you name the import `querier_worker` like you did elsewhere, you can rollback the variable name to `worker`

pkg/querier/worker/worker.go
```diff
@@ -23,6 +23,7 @@ import (
 )
 
 // WorkerConfig is config for a worker.
+// nolint:golint
```
pracucci, line 26: Instead of ignoring the lint, could we just rename it to `Config`?

pracucci (approved):
LGTM, thanks! Just a couple of nits, but feel free to ignore them.

pkg/querier/worker/worker.go
pstibrany: This is done in next PR, which is why it's just ignored here.

PR #3456, opened 2020-11-03
Return 502 if the alertmanager is not available

gotjosh:
Often, we find it confusing that in the unlikely event of:

a) creating a configuration b) then deleting it

or

a) no fallback config b) making a request to the alertmanager

Our users (and ruler) receive a 404. This is generally, acceptable as the system might not want to reveal that a configuration for a particular user exists. However, we deliberately within Cortex do not make any assumptions wrt authentication/authorization.

I feel like we need to be more indicative and let clients know what exactly is  is happening. There are many status codes that could potentially fit the bill so any suggestions are welcome.

According to [HTTP standard](https://tools.ietf.org/html/rfc7231#section-6.5.8) 502 represents:

    The 502 (Bad Gateway) status code indicates that the server, while
    acting as a gateway or proxy, received an invalid response from an
    inbound server it accessed while attempting to fulfill the request.

Given this is not cacheable by default and represents a state within the server (a user could create a config to restore the service), it seems to be like appropriate.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Return 502 if the Alertmanager of a particular tenant is not available. #3456
 * [ENHANCEMENT] Blocks storage ingester: exported more TSDB-related metrics. #3412
   - `cortex_ingester_tsdb_wal_corruptions_total`
   - `cortex_ingester_tsdb_head_truncations_failed_total`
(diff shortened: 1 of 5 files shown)


gotjosh:
This PR has a second intention - should the ruler be aware of this status and make it count it as a different type of error? A Cortex operator would typically alert on "alerts dropped" but in this scenario, the "drop of the alert" is not due to a system failure but a user "configuration".

gotjosh:
Discussed with @gouthamve online, we're exploring another avenue here.

PR #3461, opened 2020-11-04
Query store for series lookups

gouthamve:

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/api/_index.md
+++ b/docs/api/_index.md
@@ -279,7 +279,7 @@ GET,POST <prometheus-http-prefix>/api/v1/series
 GET,POST <legacy-http-prefix>/api/v1/series
 ```
 
-Find series by label matchers. Differently than Prometheus and due to scalability and performances reasons, Cortex currently ignores the `start` and `end` request parameters and always fetches the series from in-memory data stored in the ingesters.
+Find series by label matchers. Differently than Prometheus and due to scalability and performances reasons, Cortex currently ignores the `start` and `end` request parameters and always fetches the series from in-memory data stored in the ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
 
 _For more information, please check out the Prometheus [series endpoint](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers) documentation._
 
--- a/docs/configuration/v1-guarantees.md
+++ b/docs/configuration/v1-guarantees.md
@@ -24,7 +24,7 @@ The Cortex maintainers commit to ensuring future version of Cortex can read data
 Cortex strives to be 100% API compatible with Prometheus (under `/api/prom/*`); any deviation from this is considered a bug, except:
 
 - Requiring the `__name__` label on queries when querying the [chunks storage](../chunks-storage/_index.md) (queries to ingesters or clusters running the blocks storage are not affected).
-- For queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters.
+- For queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
 - Additional API endpoints for creating, removing and modifying alerts and recording rules.
 - Additional API around pushing metrics (under `/api/push`).
 - Additional API endpoints for management of Cortex itself, such as the ring.  These APIs are not part of the any compatibility guarantees.
@@ -56,3 +56,4 @@ Currently experimental features are:
 - OpenStack Swift storage support.
 - Metric relabeling in the distributor.
 - Scalable query-frontend (when using query-scheduler)
+- Querying store for series, labels APIs (`-querier.query-store-for-labels-enabled`)
--- a/docs/guides/limitations.md
+++ b/docs/guides/limitations.md
@@ -42,4 +42,4 @@ The Cortex chunks storage doesn't support queries without a metric name, like `c
 
 ## Query series and labels
 
-When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters.
+When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
--- a/pkg/querier/blocks_store_queryable.go
+++ b/pkg/querier/blocks_store_queryable.go
@@ -382,7 +382,7 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
 
 		// Fetch series from stores. If an error occur we do not retry because retries
 		// are only meant to cover missing blocks.
-		seriesSets, queriedBlocks, warnings, numChunks, err := q.fetchSeriesFromStores(spanCtx, clients, minT, maxT, matchers, convertedMatchers, maxChunksLimit, leftChunksLimit)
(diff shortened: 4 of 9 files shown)


gouthamve:
@pracucci and @pstibrany This is now ready for review. I am still querying the head even if the data is purely in object store because the abstractions are different and I can't think of a clean way to pass this flag down to `v2MetricsForLabelMatchers`.

Also, I know the flag description says labels APIs as well, but I'll add them in future PRs.

pkg/querier/querier.go
```diff
@@ -84,6 +85,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.IngesterStreaming, "querier.ingester-streaming", true, "Use streaming RPCs to query ingester.")
 	f.IntVar(&cfg.MaxSamples, "querier.max-samples", 50e6, "Maximum number of samples a single query can load into memory.")
 	f.DurationVar(&cfg.QueryIngestersWithin, "querier.query-ingesters-within", 0, "Maximum lookback beyond which queries are not sent to ingester. 0 means all queries are sent to ingester.")
+	f.BoolVar(&cfg.QueryStoreForLabels, "querier.query-store-for-labels", false, "Query long-term store for series, label values and label names APIs. Works only with blocks engine.")
```
pstibrany, line 88: I'm wondering whether we need this flag at all, or we should explicitly call it out as temporary? Can we make it true for blocks, and false for chunks?

pstibrany (approved):
LGTM. I didn't know store gateways already support this by using SkipChunks, nice!

pkg/querier/querier.go
pracucci: I would add the `-enabled` suffix (and same for the YAML config option).

pkg/querier/querier.go
```diff
@@ -287,7 +291,7 @@ func (q querier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Mat
 	// querying the long-term storage.
 	// Also, in the recent versions of Prometheus, we pass in the hint but with Func set to "series".
 	// See: https://github.com/prometheus/prometheus/pull/8050
-	if sp == nil || sp.Func == "series" {
+	if sp == nil || sp.Func == "series" && !q.queryStoreForLabels {
```
pracucci, line 294: I would make it more clear to read:
```suggestion
	if (sp == nil || sp.Func == "series") && !q.queryStoreForLabels {
```

pkg/querier/blocks_store_queryable.go
```diff
@@ -546,7 +547,7 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(
 	return seriesSets, queriedBlocks, warnings, int(numChunks.Load()), nil
 }
 
-func createSeriesRequest(minT, maxT int64, matchers []storepb.LabelMatcher, blockIDs []ulid.ULID) (*storepb.SeriesRequest, error) {
+func createSeriesRequest(minT, maxT int64, matchers []storepb.LabelMatcher, sp *storage.SelectHints, blockIDs []ulid.ULID) (*storepb.SeriesRequest, error) {
```
pracucci, line 550: I would pass only a `skipChunks boolean` instead of the whole `sp`. The reason is that we should never read mint/maxt from `sp *storage.SelectHints` and always use the input `minT/maxT`. If we don't pass `sp *storage.SelectHints`, then we would avoid any mistake.

pracucci (approved):
Good job @gouthamve ! I left few minor comments. A part from this, few more things:
1. Could you mark it as experimental in `docs/configuration/v1-guarantees.md`?
2. In the doc pages `docs/configuration/v1-guarantees.md`, `docs/api/_index.md` and `docs/guides/limitations.md` we mention the `/api/v1/series` API can't query the long-term storage. Could you improve the documentation, clearly explaining it's not supported for the chunks storage but the blocks storage supports it when the flag is enabled?

pkg/querier/querier.go
```diff
@@ -39,6 +39,7 @@ type Config struct {
 	IngesterStreaming    bool          `yaml:"ingester_streaming"`
 	MaxSamples           int           `yaml:"max_samples"`
 	QueryIngestersWithin time.Duration `yaml:"query_ingesters_within"`
+	QueryStoreForLabels  bool          `yaml:"query_store_for_labels"`
```
pracucci, line 42: YAML config options must match CLI flags.

```suggestion
	QueryStoreForLabels  bool          `yaml:"query_store_for_labels_enabled"`
```

docs/api/_index.md
```diff
@@ -279,7 +279,7 @@ GET,POST <prometheus-http-prefix>/api/v1/series
 GET,POST <legacy-http-prefix>/api/v1/series
 ```
 
-Find series by label matchers. Differently than Prometheus and due to scalability and performances reasons, Cortex currently ignores the `start` and `end` request parameters and always fetches the series from in-memory data stored in the ingesters.
+Find series by label matchers. Differently than Prometheus and due to scalability and performances reasons, Cortex currently ignores the `start` and `end` request parameters and always fetches the series from in-memory data stored in the ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `querier.query-store-for-labels-enabled` is set.
```
pracucci, line 282: In the doc, we prefix CLI flags with `-` to help the users better recognize them.

```suggestion
Find series by label matchers. Differently than Prometheus and due to scalability and performances reasons, Cortex currently ignores the `start` and `end` request parameters and always fetches the series from in-memory data stored in the ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
```

docs/configuration/v1-guarantees.md
```diff
@@ -24,7 +24,7 @@ The Cortex maintainers commit to ensuring future version of Cortex can read data
 Cortex strives to be 100% API compatible with Prometheus (under `/api/prom/*`); any deviation from this is considered a bug, except:
 
 - Requiring the `__name__` label on queries when querying the [chunks storage](../chunks-storage/_index.md) (queries to ingesters or clusters running the blocks storage are not affected).
-- For queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters.
+- For queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `querier.query-store-for-labels-enabled` is set.
```
pracucci, line 27:
```suggestion
- For queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
```

docs/configuration/v1-guarantees.md
```diff
@@ -56,3 +56,4 @@ Currently experimental features are:
 - OpenStack Swift storage support.
 - Metric relabeling in the distributor.
 - Scalable query-frontend (when using query-scheduler)
+- Querying store for series, labels APIs (`querier.query-store-for-labels-enabled`)
```
pracucci, line 59:
```suggestion
- Querying store for series, labels APIs (`-querier.query-store-for-labels-enabled`)
```

docs/guides/limitations.md
```diff
@@ -42,4 +42,4 @@ The Cortex chunks storage doesn't support queries without a metric name, like `c
 
 ## Query series and labels
 
-When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters.
+When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `querier.query-store-for-labels-enabled` is set.
```
pracucci, line 45:
```suggestion
When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
```

pracucci (approved):
Thanks @gouthamve for addressing my feedback! Final comments and then we're good to go 🚀

pracucci (approved):
Thanks a lot @gouthamve ! LGTM

PR #3405, opened 2020-10-27
Add documentation for '/debug/*' endpoints

Jigar3:
Signed-off-by: Jigar <chavada.jigar.3@gmail.com>

**What this PR does**:
- Adds documentation of `/debug/*` endpoints

**Which issue(s) this PR fixes**:
Fixes #3174 

**Checklist**
- [ ] Tests updated
- [x] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3464, opened 2020-11-05
Disable DNS resolution for testing.

pstibrany:
**What this PR does**: This PR fixes flaky `TestResetConcurrency` test ([failure](https://github.com/cortexproject/cortex/pull/3405/checks?check_run_id=1357010241)). It has used "localhost" as address to resolve, which could add additional unexpected targets to the worker.

### DIFF
--- a/pkg/querier/worker/worker.go
+++ b/pkg/querier/worker/worker.go
@@ -123,16 +123,23 @@ func newQuerierWorkerWithProcessor(cfg Config, log log.Logger, processor process
 		processor: processor,
 	}
 
-	w, err := util.NewDNSWatcher(address, cfg.DNSLookupPeriod, f)
-	if err != nil {
-		return nil, err
+	// Empty address is only used in tests, where individual targets are added manually.
+	if address != "" {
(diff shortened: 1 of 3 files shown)


pkg/querier/worker/worker.go
```diff
@@ -123,23 +123,33 @@ func newQuerierWorkerWithProcessor(cfg Config, log log.Logger, processor process
...
+		if err != nil {
+			return nil, err
+		}
+
+		servs = append(servs, w)
 	}
 
-	servs = append(servs, w)
+	if len(servs) > 0 {
+		subservices, err := services.NewManager(servs...)
```
pracucci, line 137: Should we make the manager work with an empty list of services? Would simplify the logic in the worker.

pracucci (approved):
LGTM

pkg/querier/worker/worker.go
pstibrany: I'd prefer not to complicate manager's logic more than it already is.

PR #3468, opened 2020-11-06
Frontend refactor

pstibrany:
**What this PR does**: This is another frontend refactoring. This time it extracts common code between Frontend v1 and scheduler, specifically the queue. Queue has been moved into its own package (`pkg/scheduler/queue`, but should perhaps be moved to `pkg/util/queue`), and now has a type `*RequestQueue`. This is used by both frontendv1 and scheduler. Some benchmarks from original frontend package were moved here as well.

The only logical thing that has changed is discarding of the expired requests -- previously this was done in `GetNextRequestForQuerier` method (in both scheduler and frontend), but since queue now doesn't deal with request in any way (request is now basically `interface{}`), it has been moved into scheduler/frontend instead.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/frontend/v1/queue_test.go
```diff
@@ -39,194 +49,118 @@ func TestDequeuesExpiredRequests(t *testing.T) {
...
-	require.Nil(t, err)
-	require.NotNil(t, req)
-
-	// ensure either one or two queues are fully drained, depending on which was requested first
-	_, ok := f.queues.userQueues[userID]
-	if ok {
-		// if the second user's queue was chosen for the last request,
-		// the first queue should still contain 4 (expired) requests.
-		require.Equal(t, 4, len(f.queues.getOrAddQueue(userID, 0)))
+	// Replay entire conversation in one second.
```
pracucci, line 73: I find this comment unclear. Is there a better way to phrase it?

pkg/frontend/v1/queue_test.go
```diff
@@ -39,194 +49,118 @@ func TestDequeuesExpiredRequests(t *testing.T) {
...
 	}
 
-	ctx := context.Background()
-	idx := -1
-	for i := 0; i < 100; i++ {
-		req, nidx, err := f.getNextRequestForQuerier(ctx, idx, "")
-		require.NoError(t, err)
-		require.NotNil(t, req)
-		idx = nidx
+	// Replay entire conversation.
```
pracucci, line 110: Same here. I would clarify this comment.

pracucci (approved):
Wow, good job! It took me a while to compare with the current code in `master` but I haven't noticed any issue! I just left few non blocking comments. Thanks!

pkg/frontend/v1/queue_test.go
pstibrany: I've tried:

```
	// Calling Process will only return when client disconnects or context is finished.
	// We use context timeout to stop Process call.
```

pstibrany:
Thanks for your review @pracucci!

pkg/frontend/v1/queue_test.go
pracucci: Way more clear, thanks!

pracucci:
Thanks @pstibrany ! Let's ship this ;)

PR #3470, opened 2020-11-09
blocksconvert: read chunks concurrently when finishing block

pstibrany:
**What this PR does**: when blocksconvert builder is finishing block, it needs to reorder chunks in the final block. It does that by reading chunks for each appended series (series are iterated in correct order), and writing those chunks again in correct order. This PR modifies reading process to use multiple goroutines. This was motivated by observing slowness in this phase of the process.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tools/blocksconvert/builder/tsdb.go
+++ b/tools/blocksconvert/builder/tsdb.go
@@ -325,7 +325,7 @@ func addSeriesToIndex(indexWriter *index.Writer, sl *seriesList, unsortedChunksR
 		}()
 	}
 
-	ix := 0
+	seriesRef := 0
 	for s, ok := it.Next(); ok; s, ok = it.Next() {
 		l := s.Metric
 		cs := s.Chunks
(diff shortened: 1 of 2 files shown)


tools/blocksconvert/builder/tsdb.go
```diff
@@ -297,20 +301,61 @@ func addSeriesToIndex(indexWriter *index.Writer, sl *seriesList, unsortedChunksR
...
 	}
 
-	ix := 0
+	type chunkToRead struct {
+		ref   uint64
+		chunk *chunkenc.Chunk
+		err   *error
+	}
+
+	ch := make(chan chunkToRead)
```
pracucci, line 310: Would help performances if we add a small buffer to the channel?

pracucci (approved):
LGTM, thanks!

tools/blocksconvert/builder/tsdb.go
pstibrany: I don't see how it would help the performance. Can you elaborate?

pracucci: No worries, it's OK as is.

PR #3471, opened 2020-11-09
Document query scheduler

pstibrany:
**What this PR does**: This PR adds section about query scheduler to architecture overview.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/operations/scalable-query-frontend.md
+++ b/docs/operations/scalable-query-frontend.md
@@ -11,7 +11,7 @@ Note that these instructions apply in both the HA single binary scenario or micr
 
 ## Scaling the Query Frontend
 
-For every query frontend the querier adds a [configurable number of goroutines](https://github.com/cortexproject/cortex/blob/1797adfed2979f6096c3305b0dc9162c1ec0c046/pkg/querier/worker/worker.go#L212)
+For every query frontend the querier adds a [configurable number of concurrent workers](https://github.com/cortexproject/cortex/blob/1797adfed2979f6096c3305b0dc9162c1ec0c046/pkg/querier/worker/worker.go#L212)
 which are each capable of executing a query.
 Therefore, scaling the query frontend impacts the amount of work each individual querier is attempting to do at any given time.
 
@@ -37,11 +37,11 @@ This makes scaling query frontend easier, as it allows running multiple query fr
 In order to use query scheduler, both query frontend and queriers must be configured with query scheduler address
 (using `-frontend.scheduler-address` and `-querier.scheduler-address` options respectively).
 
-Note that querier will only fetch queries from query frontend OR query scheduler, but not both.
-If configured with both `-querier.frontend-address` and `-querier.scheduler-address`, querier will only use scheduler address.
+Note that querier will only fetch queries from query frontend or query scheduler, but not both.
+`-querier.frontend-address` and `-querier.scheduler-address` options are mutually exclusive, and at most one can be set.
 
 When using query scheduler, it is recommended to run two query scheduler instances.
-Running only one query scheduler poses a risk of increased query latency during scheduler restarts.
+Running only one query scheduler poses a risk of increased query latency when single scheduler crashes or restarts.
 It shouldn't be necessary to run more than two schedulers.
 
 When using single-binary mode, Cortex defaults to run **without** query scheduler.
@@ -49,8 +49,8 @@ When using single-binary mode, Cortex defaults to run **without** query schedule
 ### DNS Configuration / Readiness
 
 When a new frontend is first created on scale up it will not immediately have queriers attached to it.
-The existing endpoint `/ready` was updated to only return http 200 when the query frontend was ready to serve queries.
-Make sure to configure this endpoint as a healthcheck in your load balancer.
-Otherwise a query frontend scale up event might result in failed queries or high latency for a bit while queriers attach.
+The existing endpoint `/ready` returns HTTP 200 status code only when the query frontend is ready to serve queries.
+Make sure to configure this endpoint as a healthcheck in your load balancer,
+otherwise a query frontend scale up event might result in failed queries or high latency for a bit while queriers attach.
 
 When using query frontend with query scheduler, `/ready` will report 200 status code only after frontend discovers some schedulers via DNS resolution.
--- a/docs/operations/scalable-query-frontend.md
+++ b/docs/operations/scalable-query-frontend.md
@@ -13,7 +13,7 @@ Note that these instructions apply in both the HA single binary scenario or micr
 
 For every query frontend the querier adds a [configurable number of concurrent workers](https://github.com/cortexproject/cortex/blob/1797adfed2979f6096c3305b0dc9162c1ec0c046/pkg/querier/worker/worker.go#L212)
 which are each capable of executing a query.
-Therefore, scaling the query frontend impacts the amount of work each individual querier is attempting to do at any given time.
+Each worker is connected to a single query frontend instance, therefore scaling up the query frontend impacts the amount of work each individual querier is attempting to do at any given time.
 
 Scaling up may cause a querier to attempt more work than they are configured for due to restrictions such as memory and cpu limits.
 Additionally, the PromQL engine itself is limited in the number of queries it can do as configured by the `-querier.max-concurrent` parameter.
@@ -42,7 +42,7 @@ Note that querier will only fetch queries from query frontend or query scheduler
 
 When using query scheduler, it is recommended to run two query scheduler instances.
 Running only one query scheduler poses a risk of increased query latency when single scheduler crashes or restarts.
-It shouldn't be necessary to run more than two schedulers.
+Running two query-schedulers should be enough even for large Cortex clusters with an high QPS.
 
 When using single-binary mode, Cortex defaults to run **without** query scheduler.
 
--- a/docs/operations/scalable-query-frontend.md
+++ b/docs/operations/scalable-query-frontend.md
@@ -5,21 +5,48 @@ weight: 5
 slug: scaling-query-frontend
 ---
 
-Historically scaling the Cortex query frontend has [posed some challenges](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/).
-This document aims to detail how to use some of the added configuration parameters to correctly scale the frontend.
+Historically scaling the Cortex query frontend has [posed some challenges](../proposals/scalable-query-frontend.md).
+This document aims to detail how to use some added configuration parameters to correctly scale the frontend.
 Note that these instructions apply in both the HA single binary scenario or microservices mode.
 
-## Using Query Scheduler
+## Scaling the Query Frontend
 
-Query scheduler is new service that moves the in-memory queue from query frontend to a separate component.
-This makes scaling query frontend easier, as it allows running multiple query frontends without increasing the number of queues,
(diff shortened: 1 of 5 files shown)


docs/operations/scalable-query-frontend.md
```diff
@@ -5,15 +5,35 @@ weight: 5
...
 ---
 
-Historically scaling the Cortex query frontend has [posed some challenges](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/).  This document aims to detail how to use some of the added configuration parameters to correctly scale the frontend.  Note that these instructions apply in both the HA single binary scenario or microservices mode.
+Historically scaling the Cortex query frontend has [posed some challenges](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/).
+This document aims to detail how to use some of the added configuration parameters to correctly scale the frontend.
+Note that these instructions apply in both the HA single binary scenario or microservices mode.
+
+## Using Query Scheduler
+
+Query scheduler is new service that moves the in-memory queue from query frontend to a separate component.
```
pracucci, line 14: Remember doc is here to stay. Try to avoid time-based references, like "new" or "now".

```suggestion
Query scheduler is a service that moves the in-memory queue from query frontend to a separate component.
```

docs/operations/scalable-query-frontend.md
```diff
@@ -5,15 +5,35 @@ weight: 5
...
+This makes scaling query frontend easier, as it allows running multiple query frontends without increasing the number of queues,
+which makes life difficult for queriers (as above-linked document explains).
```
pracucci, line 16: > as above-linked document explains

I would just re-link it. BTW, links in the doc should be relative and not absolute.

docs/operations/scalable-query-frontend.md
```diff
@@ -5,15 +5,35 @@ weight: 5
...
+
+When using query frontend with query scheduler, `/ready` will report 200 status code only after frontend discovers some schedulers via DNS resolution.
 
 ## Querier Max Concurrency
 
-Make sure to configure the querier frontend worker to match max concurrency.  This will allow the operator to freely scale the frontend up and down without impacting the amount of work an individual querier is attempting to perform.  More details [here](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/#dynamic-querier-concurrency).
+When using query frontend (possibly in combination with query scheduler), queriers need to connect to frontends or schedulers.
+To make sure that querier doesn't receive more queries that it can handle at the same time, make sure to configure the querier to match its PromQL concurrency with number of connections.
+This allows the operator to freely scale the frontend or scheduler up and down without impacting the amount of work an individual querier is attempting to perform.
+More details [here](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/#dynamic-querier-concurrency).
```
pracucci, line 36: I'm a bit unsure about linking proposals in the doc. The reason is that we don't keep proposals updated (and we shouldn't) but proposals get outdated over the time. Linking to proposals may be misleasing.

docs/operations/scalable-query-frontend.md
```diff
@@ -5,15 +5,35 @@ weight: 5
...
+Make sure to configure this endpoint as a healthcheck in your load balancer.
+Otherwise a query frontend scale up event might result in failed queries or high latency for a bit while queriers attach.
+
+When using query frontend with query scheduler, `/ready` will report 200 status code only after frontend discovers some schedulers via DNS resolution.
 
 ## Querier Max Concurrency
 
-Make sure to configure the querier frontend worker to match max concurrency.  This will allow the operator to freely scale the frontend up and down without impacting the amount of work an individual querier is attempting to perform.  More details [here](https://cortexmetrics.io/docs/proposals/scalable-query-frontend/#dynamic-querier-concurrency).
+When using query frontend (possibly in combination with query scheduler), queriers need to connect to frontends or schedulers.
+To make sure that querier doesn't receive more queries that it can handle at the same time, make sure to configure the querier to match its PromQL concurrency with number of connections.
```
pracucci, line 34: > make sure to configure the querier to match its PromQL concurrency with number of connections

Could you mention the involved CLI flags? May not be clear to the reader.

pracucci (approved):
Thanks @pstibrany for working on the doc! LGTM. I just left minor comments.

docs/operations/scalable-query-frontend.md
pstibrany: There is an example after this paragraph.

pstibrany:
Thanks for your review @pracucci. I've now reworked "Scaling the Query Frontend":
- added section about difficulties of scaling query frontend (copied and modified from proposal)
- inlined config options
- expanded Query scheduler section
- removed links to proposal, and only left one on the top (for "historical" context)
- made links relative

PTAL when you find some time. Thank you!

docs/operations/scalable-query-frontend.md
```diff
@@ -5,25 +5,52 @@ weight: 5
...
+Historically scaling the Cortex query frontend has [posed some challenges](../proposals/scalable-query-frontend.md).
+This document aims to detail how to use some added configuration parameters to correctly scale the frontend.
+Note that these instructions apply in both the HA single binary scenario or microservices mode.
 
-## DNS Configuration / Readiness
+## Scaling the Query Frontend
 
-When a new frontend is first created on scale up it will not immediately have queriers attached to it.  The existing endpoint `/ready` was updated to only return http 200 when the query frontend was ready to serve queries.  Make sure to configure this endpoint as a healthcheck in your load balancer.  Otherwise a query frontend scale up event might result in failed queries or high latency for a bit while queriers attach.
+For every query frontend the querier adds a [configurable number of goroutines](https://github.com/cortexproject/cortex/blob/1797adfed2979f6096c3305b0dc9162c1ec0c046/pkg/querier/worker/worker.go#L212)
```
pracucci, line 14: You know what I think about mentioning goroutines in the doc :) I don't think our doc should either talk about goroutines or link to code.

docs/operations/scalable-query-frontend.md
```diff
@@ -5,25 +5,52 @@ weight: 5
...
+### Query Scheduler
+
+Query scheduler is a service that moves the in-memory queue from query frontend to a separate component.
+This makes scaling query frontend easier, as it allows running multiple query frontends without increasing the number of queues.
+
+In order to use query scheduler, both query frontend and queriers must be configured with query scheduler address
+(using `-frontend.scheduler-address` and `-querier.scheduler-address` options respectively).
+
+Note that querier will only fetch queries from query frontend OR query scheduler, but not both.
+If configured with both `-querier.frontend-address` and `-querier.scheduler-address`, querier will only use scheduler address.
```
pracucci, line 41: We should add a config validation to make it fail if both addresses are specified. What do you think?

docs/operations/scalable-query-frontend.md
```diff
@@ -5,25 +5,52 @@ weight: 5
...
+
+When using query scheduler, it is recommended to run two query scheduler instances.
+Running only one query scheduler poses a risk of increased query latency during scheduler restarts.
```
pracucci, line 44: Also for HA purposes.

docs/operations/scalable-query-frontend.md
```diff
@@ -5,25 +5,52 @@ weight: 5
...
+It shouldn't be necessary to run more than two schedulers.
+
+When using single-binary mode, Cortex defaults to run **without** query scheduler.
+
+### DNS Configuration / Readiness
+
+When a new frontend is first created on scale up it will not immediately have queriers attached to it.
+The existing endpoint `/ready` was updated to only return http 200 when the query frontend was ready to serve queries.
```
pracucci, line 52: I would not mention "was updated" but I would say:

```suggestion
The query-frontend `/ready` endpoint returns HTTP 200 status code only when the query-frontend is ready to serve queries.
```

pracucci (approved):
Good job! Way better than before. I left few more comments.

docs/operations/scalable-query-frontend.md
pstibrany: > ... when single scheduler crashes or restarts.

OK?

docs/operations/scalable-query-frontend.md
```diff
@@ -5,25 +5,52 @@ weight: 5
...
+This document aims to detail how to use some added configuration parameters to correctly scale the frontend.
+Note that these instructions apply in both the HA single binary scenario or microservices mode.
 
-## DNS Configuration / Readiness
+## Scaling the Query Frontend
 
-When a new frontend is first created on scale up it will not immediately have queriers attached to it.  The existing endpoint `/ready` was updated to only return http 200 when the query frontend was ready to serve queries.  Make sure to configure this endpoint as a healthcheck in your load balancer.  Otherwise a query frontend scale up event might result in failed queries or high latency for a bit while queriers attach.
+For every query frontend the querier adds a [configurable number of concurrent workers](https://github.com/cortexproject/cortex/blob/1797adfed2979f6096c3305b0dc9162c1ec0c046/pkg/querier/worker/worker.go#L212)
+which are each capable of executing a query.
+Therefore, scaling the query frontend impacts the amount of work each individual querier is attempting to do at any given time.
```
pracucci, line 16:
```suggestion
Each worker is connected to a single query-frontend instance, therefore scaling up the query-frontend impacts the amount of work each individual querier is attempting to do at any given time.
```

docs/operations/scalable-query-frontend.md
```diff
@@ -5,25 +5,52 @@ weight: 5
...
+This makes scaling query frontend easier, as it allows running multiple query frontends without increasing the number of queues.
+
+In order to use query scheduler, both query frontend and queriers must be configured with query scheduler address
+(using `-frontend.scheduler-address` and `-querier.scheduler-address` options respectively).
+
+Note that querier will only fetch queries from query frontend or query scheduler, but not both.
+`-querier.frontend-address` and `-querier.scheduler-address` options are mutually exclusive, and at most one can be set.
+
+When using query scheduler, it is recommended to run two query scheduler instances.
+Running only one query scheduler poses a risk of increased query latency when single scheduler crashes or restarts.
```
pracucci, line 44: > poses a risk of increased query latency

Increased query latency or failing queries? If the query-frontend can't enqueue a query in the query-scheduler, doesn't the query fail?

docs/operations/scalable-query-frontend.md
```diff
@@ -5,25 +5,52 @@ weight: 5
...
+It shouldn't be necessary to run more than two schedulers.
```
pracucci, line 45:
```suggestion
Running two query-schedulers should be enough even for large Cortex clusters with an high QPS.
```

pracucci (approved):
Good job! I left very final non-blocking comments, but feel free to merge it without any other review from my side. Thanks!

docs/operations/scalable-query-frontend.md
pstibrany: If frontend has no workers attempting to forward messages to scheduler (eg. because no scheduler is running), or frontend workers are unable to connect (send "INIT" message and get "OK" back), then frontend will get stuck while trying to pass query to the channel -- but no worker is reading from the channel. Once workers connect to schedulers, they will process the query. From user's perspective, query is just slower.

If workers are connected and read from channel, then forwarding of the query to scheduler may fail. In this case frontend will retry to send query to the channel. This adds latency. If this forwarding fails too many times, query fails.

PR #3472, opened 2020-11-09
add support for slicing of chunk in encodings

sandeepsukhani:
**What this PR does**:
We slice a chunk when deletion interval covers a chunk partially. This PR adds support for slicing off a chunk in encodings which helps add support for deletion in Loki because it implements its own encoding for storing logs in chunks. 

There was existing `Slice` method in chunk encoding interface which is specifically built for query optimization and is a noop for some of the encodings so instead of changing that function, I have added a new method called `Rebound` which slices a chunk for a given time range.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunk/chunk.go
+++ b/pkg/chunk/chunk.go
@@ -20,15 +20,12 @@ import (
 	"github.com/cortexproject/cortex/pkg/prom1/storage/metric"
 )
 
-// Errors that decode can return
 const (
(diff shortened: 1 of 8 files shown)


pstibrany (approved):
LGTM, thanks!

pracucci (approved):
LGTM, thanks!

PR #3476, opened 2020-11-10
Expose WAL segment size setting

pracucci:
**What this PR does**:
There are some use cases (eg. Cortex clusters with very small tenants) where reducing the WAL segment size may reduce the disk requirements. In this PR I'm proposing to allow to customise it.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany (approved):

codesome (approved):
Should we have some minimum size for the flag? I don't expect anyone to set it very low, but for example if I set it to 1, it won't work.

PR #3477, opened 2020-11-10
Added option to ignore regexp-matching users by scheduler.

pstibrany:
**What this PR does**: Added option for blocksconvert scheduler to ignore users based on regexp.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tools/blocksconvert/scheduler/scheduler.go
```diff
@@ -51,19 +54,29 @@ func NewScheduler(cfg Config, scfg blocksconvert.SharedConfig, l log.Logger, reg
 		users = blocksconvert.ParseAllowedUsers(cfg.AllowedUsers)
 	}
 
-	s := newSchedulerWithBucket(l, b, scfg.BucketPrefix, users, cfg, reg)
+	var ignoredUserRegex *regexp.Regexp = nil
+	if users.AllUsersAllowed() && cfg.IgnoredUserPattern != "" {
```
pracucci, line 58: Why not applying ignored users even if `users.AllUsersAllowed()` is false?

pracucci (approved):

tools/blocksconvert/scheduler/scheduler.go
pstibrany: You're right. I copied the logic from scanner, where these two options are exclusive. It works slightly differently here, although I would only recommend using one option.

pstibrany: I've modified scanner to have the same logic wrt. to options as scheduler. (Internally the logic was already the same, just config options were treated differently).

pracucci: Ok. Makes sense to me.

pracucci (approved):
Still LGTM 😉  I think we can merge!

PR #3479, opened 2020-11-10
Merge release-1.5 back to master

csmarchbanks:
Final step of release 1.5.0, merge all of the release-branch changes back into master. In order to keep commit history, I will be merging this rather than squashing.


pracucci (approved):
LGTM, thanks!

pstibrany:
> In order to keep commit history, I will be merging this rather than squashing.

It also makes release tags reachable from master, which is important for Go tooling.

PR #3480, opened 2020-11-10
Support custom log.Logger instances in spanlogger

rfratto:
This PR adds functionality to the `spanlogger` package to support passing in custom log.Logger instances. This will be useful for me within [grafana/agent](https://github.com/grafana/agent) where loggers with different labels are passed around to various components. 

Signed-off-by: Robert Fratto <robertfratto@gmail.com>

### DIFF
--- a/pkg/util/spanlogger/spanlogger.go
+++ b/pkg/util/spanlogger/spanlogger.go
@@ -12,17 +12,30 @@ import (
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
+type loggerCtxMarker struct{}
+
+var (
+	loggerCtxKey = &loggerCtxMarker{}
+)
+
 // SpanLogger unifies tracing and logging, to reduce repetition.
 type SpanLogger struct {
 	log.Logger
 	opentracing.Span
 }
 
-// New makes a new SpanLogger.
+// New makes a new SpanLogger, where logs will be sent to the global logger.
 func New(ctx context.Context, method string, kvps ...interface{}) (*SpanLogger, context.Context) {
+	return NewWithLogger(ctx, util.Logger, method, kvps...)
+}
+
+// NewWithLogger makes a new SpanLogger with a custom log.Logger to send logs
+// to. The provided context will have the logger attached to it and can be
+// retrieved with FromContext or FromContextWithFallback.
+func NewWithLogger(ctx context.Context, l log.Logger, method string, kvps ...interface{}) (*SpanLogger, context.Context) {
 	span, ctx := opentracing.StartSpanFromContext(ctx, method)
 	logger := &SpanLogger{
-		Logger: log.With(util.WithContext(ctx, util.Logger), "method", method),
+		Logger: log.With(util.WithContext(ctx, l), "method", method),
 		Span:   span,
 	}
 	if len(kvps) > 0 {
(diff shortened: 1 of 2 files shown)


pracucci (approved):
LGTM, thanks!

jtlisi (approved):
LGTM

pkg/util/spanlogger/spanlogger.go
```diff
@@ -31,18 +44,29 @@ func New(ctx context.Context, method string, kvps ...interface{}) (*SpanLogger,
...
 func FromContext(ctx context.Context) *SpanLogger {
+	return FromContextWithFallback(ctx, util.Logger)
+}
+
+// FromContextWithContext returns a span logger using the current parent span.
+// IF there is no parent span, the SpanLogger will only log to the logger
+// within the context. If the context doesn't have a logger, the fallback
+// logger is used.
+func FromContextWithFallback(ctx context.Context, fallback log.Logger) *SpanLogger {
+	logger, ok := ctx.Value(loggerCtxKey).(log.Logger)
```
pstibrany, line 60: I don't see any code that would be *storing* value under this key. What am I missing?

pkg/util/spanlogger/spanlogger.go
```diff
@@ -31,18 +44,29 @@ func New(ctx context.Context, method string, kvps ...interface{}) (*SpanLogger,
...
-// If there is no parent span, the Spanlogger will only log to global logger.
+// FromContext returns a span logger using the current parent span. If there
+// is no parent span, the SpanLogger will only log to the logger
+// in the context. If the context doesn't have a logger, the global logger
+// is used.
 func FromContext(ctx context.Context) *SpanLogger {
+	return FromContextWithFallback(ctx, util.Logger)
+}
+
+// FromContextWithContext returns a span logger using the current parent span.
```
pstibrany, line 55:
```suggestion
// FromContextWithFallback returns a span logger using the current parent span.
```

pkg/util/spanlogger/spanlogger.go
rfratto: Thanks, I copied this in a rush from [my temporary fork of the package](https://github.com/rfratto/agent/blob/b2af97528144f62fbab936ef2fedeae212e2e6a7/pkg/util/spanlogger/spanlogger.go#L40) and forgot that one part 😞 

I'll add a test!

pkg/util/spanlogger/spanlogger_test.go
```diff
@@ -19,3 +20,32 @@ func TestSpanLogger_Log(t *testing.T) {
...
+	span = FromContextWithFallback(ctx, log.NewNopLogger())
+	span.Log("msg", "restored spanlogger")
+
+	span = FromContextWithFallback(context.Background(), logger)
+	span.Log("msg", "fallback spanlogger")
+
+	expect := [][]interface{}{
+		{"method", "test", "msg", "original spanlogger"},
+		{"msg", "restored spanlogger"},
+		{"msg", "fallback spanlogger"},
```
rfratto, line 42: It's not clear to me if it's intentional that the method key is only attached to the original span logger and not the restored ones, but that's the way it behaved prior to my change.

pstibrany: That is suspicious indeed, and perhaps a bug, but let's save that for a separate PR.

pstibrany (approved):
LGTM, thanks!

pkg/util/spanlogger/spanlogger_test.go
rfratto: On second thought, I think it makes sense even if it's not very intuitive: `FromContext`/`FromContextWithFallback` will be used from different methods than the one the span creates. I think it's probably correct that they're not annotated with the same method name since the logs are coming from different functions.

PR #3483, opened 2020-11-11
Improve compactor blocks cleaner when running with a large number of tenants

pracucci:
**What this PR does**:
I'm testing the compactor in a cluster with a large number of tenants (6K) and I've found out the blocks cleaner doesn't scale. The problem is that cleaning 1 tenant at a time is too slow, so I'm proposing to add concurrency to it.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -121,12 +121,8 @@ func (c *BlocksCleaner) cleanUsers(ctx context.Context) error {
 		return errors.Wrap(err, "failed to discover users from bucket")
 	}
 
-	return concurrency.ForeachUser(ctx, users, c.cfg.CleanupConcurrency, func(userID string) error {
-		if cleanErr := c.cleanUser(ctx, userID); cleanErr != nil {
-			return errors.Wrapf(cleanErr, "failed to delete user blocks (user: %s)", userID)
-		}
-
-		return nil
+	return concurrency.ForEachUser(ctx, users, c.cfg.CleanupConcurrency, func(ctx context.Context, userID string) error {
+		return errors.Wrapf(c.cleanUser(ctx, userID), "failed to delete user blocks (user: %s)", userID)
 	})
 }
 
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -11,7 +11,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/thanos-io/thanos/pkg/block"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/compact"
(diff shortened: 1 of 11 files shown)


pkg/compactor/compactor.go
```diff
@@ -270,7 +273,8 @@ func (c *Compactor) starting(ctx context.Context) error {
 		DataDir:             c.compactorCfg.DataDir,
 		MetaSyncConcurrency: c.compactorCfg.MetaSyncConcurrency,
 		DeletionDelay:       c.compactorCfg.DeletionDelay,
-		CleanupInterval:     util.DurationWithJitter(c.compactorCfg.CompactionInterval, 0.05),
+		CleanupInterval:     util.DurationWithJitter(c.compactorCfg.CompactionInterval, 0.1),
```
pracucci, line 276: Increased a bit the jitter.

pkg/util/concurrency/runner.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"sync"
+
+	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
+)
+
+// ForeachUser runs the provided userFunc for each userIDs up to concurrency concurrent workers.
+// In case userFunc returns error, it will continue to process remaining users but returns an
+// error with all errors userFunc has returned.
+func ForeachUser(ctx context.Context, userIDs []string, concurrency int, userFunc func(userID string) error) error {
```
pstibrany, line 13: Nit: `ForEachUser`

pkg/util/concurrency/runner.go
```diff
@@ -0,0 +1,64 @@
...
+		go func() {
+			defer wg.Done()
+
+			for userID := range ch {
+				// Ensure the context has not been canceled (ie. shutdown has been triggered).
+				if ctx.Err() != nil {
+					break
+				}
+
+				if err := userFunc(userID); err != nil {
```
pstibrany, line 32: Should we pass context to the `userFunc` instead of passing it via local variables (as callers of `ForEachUser` do now)?

pkg/compactor/blocks_cleaner.go
```diff
@@ -120,20 +121,13 @@ func (c *BlocksCleaner) cleanUsers(ctx context.Context) error {
 		return errors.Wrap(err, "failed to discover users from bucket")
 	}
 
-	errs := tsdb_errors.NewMulti()
-	for _, userID := range users {
-		// Ensure the context has not been canceled (ie. shutdown has been triggered).
-		if ctx.Err() != nil {
-			return ctx.Err()
+	return concurrency.ForeachUser(ctx, users, c.cfg.CleanupConcurrency, func(userID string) error {
+		if cleanErr := c.cleanUser(ctx, userID); cleanErr != nil {
```
pstibrany, line 125: Inner function body can be replaced with

```
        return errors.Wrapf(c.cleanUser(ctx, userID), "failed to delete user blocks (user: %s)", userID)
```

pstibrany (approved):
LGTM, nice! I've left some non-blocking nits.

pkg/util/concurrency/runner.go
pracucci: Yes, I think it's a better approach.

pracucci:
Thanks @pstibrany, I've addressed all your comments!

Issue #2827, opened 2020-07-02
Wait stable store-gateway ring state before running the initial blocks sync


pracucci:
In the event of a cluster cold start or scale up of 2+ store-gateway instances at the same time we may end up in a situation where each new store-gateway instance starts at a slightly different time and thus each one syncs blocks based on a different state of the ring.

## Example of the issue

Consider this cold start scenario (SG = store-gateway):
- T1: SG1 starts
- T2: SG2 starts
- T3: SG1 begins synching blocks based on the current state of the ring (ring: SG1)
- T3: SG3 starts
- T4: SG2 begins synching blocks based on the current state of the ring (ring: SG1, SG2)
- T5: SG3 begins synching blocks based on the current state of the ring (ring: SG1, SG2, SG3)

In case of a replication factor of 1 (but the same applies to clusters with a replication factor > 1), SG1 will initially sync all blocks, SG2 will initially sync 50% of blocks, SG3 will initially sync 30% of blocks.

## Proposal

An idea would be introducing a buffer period before starting synching blocks. During this buffer period we do monitor the ring and wait until the ring state is stable (doesn't change) for at least X seconds, before running the initial sync.

The same approach could be applied to the sync done by the store-gateway each time the ring topology changes, in order to cover the case of existing store-gateway instances re-synching after a scale up (so that they will re-synch blocks once all scaled up instances have started).

Under normal conditions, it's not necessary to have a very large buffer period. For example, 1 minute could be a reasonable default to cover a cluster deployed to Kubernetes.

pracucci:
_Still valid._

amckinley:
➕ 1, this has bitten us when doing a cold start after migrating Cortex from one k8s cluster to another. We normally run 9 `store-gateways` so each one normally starts up pretty quickly, but when cold starting, all `store-gateways` attempt to fetch the index for all blocks, causing them all to eventually fail when they fill their disks.

jtlisi:
We should verify that the `token-observe-period` feature is operative for all KV stores. If so, that is a valid workaround for this issue.

PR #3484, opened 2020-11-11
Wait for a stable ring at compactor startup

pracucci:
**What this PR does**:
This is like https://github.com/cortexproject/cortex/issues/2827 but for the compactor. The idea is to add the ability to wait for a stable ring at compactor startup. If this proposal is accepted, I will then work on the same for the store-gateway.

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ring/util.go
+++ b/pkg/ring/util.go
@@ -82,6 +82,38 @@ func WaitInstanceState(ctx context.Context, r *Ring, instanceID string, state In
 	return backoff.Err()
 }
 
+// WaitRingStability monitors the ring topology for the provided operation and waits until it
+// keeps stable for at least minStability.
+func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability, maxWaiting time.Duration) error {
+	const pollingFrequency = time.Second
+
+	// Configure the max waiting time as a context deadline.
+	ctx, cancel := context.WithTimeout(ctx, maxWaiting)
+	defer cancel()
+
+	// Get the initial ring state.
+	ringLastState, _ := r.GetAllHealthy(op) // nolint:errcheck
+	ringLastStateTs := time.Now()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(pollingFrequency):
+			// We ignore the error because in case of error it will return an empty
+			// replication set which we use to compare with the previous state.
+			currRingState, _ := r.GetAllHealthy(op) // nolint:errcheck
+
+			if HasReplicationSetChanged(ringLastState, currRingState) {
+				ringLastState = currRingState
+				ringLastStateTs = time.Now()
+			} else if ringLastStateTs.Add(minStability).Before(time.Now()) {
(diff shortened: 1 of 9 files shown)


pkg/ring/util.go
```diff
@@ -82,6 +82,38 @@ func WaitInstanceState(ctx context.Context, r *Ring, instanceID string, state In
 	return backoff.Err()
 }
 
+// WaitRingStability monitors the ring topology for the provided operation and waits until it
+// keeps stable for at least minStability.
+func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability, maxWaiting time.Duration) error {
+	const pollingFrequency = time.Second
+
+	// Configure the max waiting time as a context deadline.
```
pstibrany, line 90: I'd suggest removing `maxWaiting` and rely on context only. Client calling this can already setup context deadline, if needed. The code right now returns `context.DeadlineExceeded` when `maxWaiting` is reached, which client may confuse with its own context deadline.

pkg/ring/util.go
```diff
@@ -82,6 +82,38 @@ func WaitInstanceState(ctx context.Context, r *Ring, instanceID string, state In
...
+
+	// Get the initial ring state.
+	ringLastState, _ := r.GetAllHealthy(op) // nolint:errcheck
+	ringLastStateTs := time.Now()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(pollingFrequency):
```
pstibrany, line 102: Use ticker?

pkg/ring/util.go
```diff
@@ -82,6 +82,38 @@ func WaitInstanceState(ctx context.Context, r *Ring, instanceID string, state In
...
+			// We ignore the error because in case of error it will return an empty
+			// replication set which we use to compare with the previous state.
+			currRingState, _ := r.GetAllHealthy(op) // nolint:errcheck
+
+			if HasReplicationSetChanged(ringLastState, currRingState) {
+				ringLastState = currRingState
+				ringLastStateTs = time.Now()
+			} else if ringLastStateTs.Add(minStability).Before(time.Now()) {
```
pstibrany, line 110:
```suggestion
			} else if time.Since(ringLastStateTs) > minStability {
```

pstibrany (approved):
LGTM, thanks. My comments are not blocking.

pkg/ring/util.go
pracucci: It's more lines of code. Any benefit of using a ticker?

pkg/ring/util.go
pracucci: I thought the same, but using it from the caller perspective is more annoying, because it's not something you can do inline. You need to get a new context, defer the cancel(), call the WaitRingStability().

pkg/ring/util.go
pracucci: Discussed offline. Will move to ticker.

PR #3485, opened 2020-11-11
Fix makefile for SELinux enabled machines

simonswine:
**What this PR does**:

Changes in the makefile how containers are used to mount volumes into contianer, break on my SELinux enabled Fedora workstation. This changes ensure, that the volumes are mounted using the 'z' label, which makes sure to configure them correctly to work with SELinux

**Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,9 +113,9 @@ GO_FLAGS := -ldflags "-X main.Branch=$(GIT_BRANCH) -X main.Revision=$(GIT_REVISI
 
 ifeq ($(BUILD_IN_CONTAINER),true)
 
-GOVOLUMES=	-v $(shell pwd)/.cache:/go/cache:delegated \
-			-v $(shell pwd)/.pkg:/go/pkg:delegated \
-			-v $(shell pwd):/go/src/github.com/cortexproject/cortex:delegated
+GOVOLUMES=	-v $(shell pwd)/.cache:/go/cache:delegated,z \
+			-v $(shell pwd)/.pkg:/go/pkg:delegated,z \
+			-v $(shell pwd):/go/src/github.com/cortexproject/cortex:delegated,z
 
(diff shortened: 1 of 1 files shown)


pracucci (approved):
LGTM, thanks! Manually tested on OSX and works fine (to make sure doesn't introduce any issue).

pstibrany (approved):

bboreham:
Hi, could I ask that you put a little more detail on "break"?  The Docker docs say "If you use selinux you can add the z or Z options to modify the selinux label", but there isn't any suggestion that it is mandatory (or indeed any explanation of what it does or where to look for more info).

I'm asking because I maintain other projects that have similar builds, and would like to know if this change should be made there.

simonswine:
@bboreham yes, no problem the PR was not really going into too much detail.

Once you have a SELinux enforcing system all directories that are mounted in a container will need to be labelled certain labels, so that the can be read/written to at all.

While the root file system and volumes that are managed by docker automatically come with the correct labels, mounting a local path, e.g. in your home directory, won't have those labels to be protected by containers that have elevated privileges. The docker command will succeed, but subsequent access to those files will within the container will be failing with a permission denied. Adding the `z` flag to the docker volume mounts will change the label to those paths to use `container_file_t` (only as long docker daemon is started with `--selinux-enabled`).

More on this can be found here: https://github.com/containers/container-selinux/blob/master/container_selinux.8
(I am only really familiar with the RedHat policies, so that is at least valid for RHEL, Centos, Fedora)

I usually try to avoid mounting paths in `docker run` build steps and rather use something like this:
- `docker create` to create but not start the container
- `docker cp` to copy source files in
- `doker start` to exectue the build step
- `docker cp` to the built artifact out

This also avoids issues with UIDs/GIDS, but can get tricky too if it is meant to operate on the whole repo.

PR #3491, opened 2020-11-12
Close TSDB and delete local data when TSDB is idle for a long time.

pstibrany:
**What this PR does**: This PR adds ability for ingester to close idle TSDB and delete local data when TSDB is idle for a long time (no data is appended to it). Closing/deletion only happens if head is empty (possibly thanks to previous force-compaction due to being idle), and all blocks were shipped.

~This PR also takes care of hiding user metrics, while keeping the registry to avoid breaking the counters over all users.~ Support for metrics removal has been extracted to separate PR, https://github.com/cortexproject/cortex/pull/3511.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/blocks-storage/store-gateway.md
+++ b/docs/blocks-storage/store-gateway.md
@@ -703,6 +703,15 @@ blocks_storage:
     # CLI flag: -blocks-storage.tsdb.flush-blocks-on-shutdown
     [flush_blocks_on_shutdown: <boolean> | default = false]
 
+    # If TSDB has not received any data for this duration, and all blocks from
+    # TSDB have been shipped, TSDB is closed and deleted from local disk. If set
+    # to positive value, this value should be equal or higher than
+    # -querier.query-ingesters-within flag to make sure that TSDB is not closed
+    # prematurely, which could cause partial query results. 0 or negative value
+    # disables closing of idle TSDB.
+    # CLI flag: -blocks-storage.tsdb.close-idle-tsdb-timeout
+    [close_idle_tsdb_timeout: <duration> | default = 0s]
+
     # limit the number of concurrently opening TSDB's on startup
     # CLI flag: -blocks-storage.tsdb.max-tsdb-opening-concurrency-on-startup
     [max_tsdb_opening_concurrency_on_startup: <int> | default = 10]
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -311,10 +311,7 @@ type TSDBState struct {
 	appenderAddDuration    prometheus.Histogram
 	appenderCommitDuration prometheus.Histogram
 	refCachePurgeDuration  prometheus.Histogram
-
-	// Idle TSDB metrics.
-	idleTsdbChecks      prometheus.Counter
-	idleTsdbCheckResult *prometheus.CounterVec
+	idleTsdbCheckResult    *prometheus.CounterVec
 }
 
 func newTSDBState(bucketClient objstore.Bucket, registerer prometheus.Registerer) TSDBState {
@@ -370,10 +367,6 @@ func newTSDBState(bucketClient objstore.Bucket, registerer prometheus.Registerer
 			Buckets: prometheus.DefBuckets,
 		}),
 
-		idleTsdbChecks: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
-			Name: "cortex_ingester_idle_tsdb_checks_total",
-			Help: "The total number of checks for idle TSDB.",
-		}),
 		idleTsdbCheckResult: idleTsdbCheckResult,
 	}
 }
@@ -1556,8 +1549,6 @@ func (i *Ingester) closeAndDeleteIdleUserTSDBs(ctx context.Context) error {
 		if ctx.Err() != nil {
 			return nil
 		}
-
-		i.TSDBState.idleTsdbChecks.Inc()
 
 		result := i.closeAndDeleteUserTSDBIfIdle(userID)
 
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -48,6 +48,31 @@ type Shipper interface {
 	Sync(ctx context.Context) (uploaded int, err error)
 }
 
+type tsdbState int
+
+const (
+	active          tsdbState = iota // Pushes are allowed only in this state.
+	forceCompacting                  // TSDB is being force-compacted.
+	closing                          // Used while closing idle TSDB.
+	closed                           // Used to avoid setting closing back to active in closeAndDeleteIdleUsers method.
+)
+
+// Describes result of TSDB idle check (for closing). String is used as metric label.
+type tsdbIdleCheckResult string
+
+const (
+	tsdbIdle              tsdbIdleCheckResult = "idle" // Not reported via metrics. Metrics use tsdbIdleClosed on success.
+	tsdbShippingDisabled  tsdbIdleCheckResult = "shipping_disabled"
+	tsdbNotIdle           tsdbIdleCheckResult = "not_idle"
+	tsdbNotCompacted      tsdbIdleCheckResult = "not_compacted"
+	tsdbNotShipped        tsdbIdleCheckResult = "not_shipped"
+	tsdbCheckFailed       tsdbIdleCheckResult = "check_failed"
+	tsdbCloseFailed       tsdbIdleCheckResult = "close_failed"
+	tsdbNotActive         tsdbIdleCheckResult = "not_active"
+	tsdbDataRemovalFailed tsdbIdleCheckResult = "data_removal_failed"
+	tsdbIdleClosed        tsdbIdleCheckResult = "idle_closed" // Success.
+)
+
 type userTSDB struct {
 	db             *tsdb.DB
 	userID         string
@@ -56,13 +81,12 @@ type userTSDB struct {
 	seriesInMetric *metricCounter
 	limiter        *Limiter
 
-	forcedCompactionInProgressMtx sync.RWMutex
-	forcedCompactionInProgress    bool
-
-	pushesInFlight sync.WaitGroup
+	stateMtx       sync.RWMutex
+	state          tsdbState
+	pushesInFlight sync.WaitGroup // Increased with Read lock held, only if state == active.
 
 	// Used to detect idle TSDBs.
-	lastUpdate *atomic.Int64
+	lastUpdate atomic.Int64
 
 	// Thanos shipper used to ship blocks to the storage.
 	shipper Shipper
@@ -102,16 +126,25 @@ func (u *userTSDB) StartTime() (int64, error) {
 	return u.db.StartTime()
 }
 
+func (u *userTSDB) casState(from, to tsdbState) bool {
+	u.stateMtx.Lock()
+	defer u.stateMtx.Unlock()
+
+	if u.state != from {
+		return false
+	}
+	u.state = to
+	return true
+}
+
 // compactHead compacts the Head block at specified block durations avoiding a single huge block.
 func (u *userTSDB) compactHead(blockDuration int64) error {
-	u.forcedCompactionInProgressMtx.Lock()
-	u.forcedCompactionInProgress = true
-	u.forcedCompactionInProgressMtx.Unlock()
-	defer func() {
-		u.forcedCompactionInProgressMtx.Lock()
-		u.forcedCompactionInProgress = false
-		u.forcedCompactionInProgressMtx.Unlock()
-	}()
+	if !u.casState(active, forceCompacting) {
+		return errors.New("TSDB head cannot be compacted because it is not in active state (possibly being closed)")
+	}
+
+	defer u.casState(forceCompacting, active)
+
 	// Ingestion of samples in parallel with forced compaction can lead to overlapping blocks.
 	// So we wait for existing in-flight requests to finish. Future push requests would fail until compaction is over.
 	u.pushesInFlight.Wait()
@@ -190,7 +223,7 @@ func (u *userTSDB) blocksToDelete(blocks []*tsdb.Block) map[ulid.ULID]struct{} {
 		return deletable
 	}
 
-	shipperMeta, err := shipper.ReadMetaFile(u.db.Dir())
+	shippedBlocks, err := u.getShippedBlocks()
 	if err != nil {
 		// If there is any issue with the shipper, we should be conservative and not delete anything.
 		level.Error(util.Logger).Log("msg", "failed to read shipper meta during deletion of blocks", "user", u.userID, "err", err)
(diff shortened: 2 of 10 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -102,16 +110,25 @@ func (u *userTSDB) StartTime() (int64, error) {
...
-	u.forcedCompactionInProgressMtx.Lock()
-	u.forcedCompactionInProgress = true
-	u.forcedCompactionInProgressMtx.Unlock()
-	defer func() {
-		u.forcedCompactionInProgressMtx.Lock()
-		u.forcedCompactionInProgress = false
-		u.forcedCompactionInProgressMtx.Unlock()
-	}()
+	if !u.setState(active, forceCompacting) {
+		return errors.New("TSDB is not active")
```
pracucci, line 127: I would improve the log message saying "TSDB head can't be compacted because ..."

pkg/ingester/ingester_v2.go
```diff
@@ -216,6 +242,37 @@ func (u *userTSDB) setLastUpdate(t time.Time) {
 	u.lastUpdate.Store(t.Unix())
 }
 
+func (u *userTSDB) canIdleClose(idleTimeout time.Duration) (bool, error) {
```
pracucci, line 245: I would just return `error`, so that when we call it we can log the specific reason why we can't.

pkg/ingester/metrics.go
```diff
@@ -443,14 +449,20 @@ func (sm *tsdbMetrics) Collect(out chan<- prometheus.Metric) {
...
-	data.SendSumOfCountersPerUser(out, sm.tsdbChunksCreatedTotal, "prometheus_tsdb_head_chunks_created_total")
-	data.SendSumOfCountersPerUser(out, sm.tsdbChunksRemovedTotal, "prometheus_tsdb_head_chunks_removed_total")
 	data.SendSumOfCounters(out, sm.tsdbMmapChunkCorruptionTotal, "prometheus_tsdb_mmap_chunk_corruptions_total")
 	data.SendSumOfCounters(out, sm.checkpointDeleteFail, "prometheus_tsdb_checkpoint_deletions_failed_total")
 	data.SendSumOfCounters(out, sm.checkpointDeleteTotal, "prometheus_tsdb_checkpoint_deletions_total")
 	data.SendSumOfCounters(out, sm.checkpointCreationFail, "prometheus_tsdb_checkpoint_creations_failed_total")
 	data.SendSumOfCounters(out, sm.checkpointCreationTotal, "prometheus_tsdb_checkpoint_creations_total")
 
+	// Before sending per-user metrics, we hide removed users.
+	data.RemoveUsersMatchingPredicate(func(user string) bool {
```
pracucci, line 459: Discussed offline. We should improve this. A way could be leverage on #3506 and add removal logic in the metrics helper.

pracucci (commented):
Very good job! I did a first pass review and I will do a more accurate review once initial feedback has been discussed and/or addressed. I would also mark this feature as experimental in the "v1 guarantees" doc page: WDYT?

pkg/storage/tsdb/config.go
```diff
@@ -206,6 +212,9 @@ func (cfg *TSDBConfig) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.WALCompressionEnabled, "blocks-storage.tsdb.wal-compression-enabled", false, "True to enable TSDB WAL compression.")
 	f.IntVar(&cfg.WALSegmentSizeBytes, "blocks-storage.tsdb.wal-segment-size-bytes", wal.DefaultSegmentSize, "TSDB WAL segments files max size (bytes).")
 	f.BoolVar(&cfg.FlushBlocksOnShutdown, "blocks-storage.tsdb.flush-blocks-on-shutdown", false, "True to flush blocks to storage on shutdown. If false, incomplete blocks will be reused after restart.")
+	f.DurationVar(&cfg.CloseIdleTSDBInterval, "blocks-storage.tsdb.close-idle-tsdb-interval", 0, "How often to check for idle TSDB that can be closed. 0 means disabled.")
+	f.IntVar(&cfg.CloseIdleTSDBConcurrency, "blocks-storage.tsdb.close-idle-tsdb-concurrency", 10, "Limit the number of concurrently closing TSDB when idle.")
+	f.DurationVar(&cfg.CloseIdleTSDBTimeout, "blocks-storage.tsdb.close-idle-tsdb-timeout", 12*time.Hour, "If TSDB has not received any data for this duration, and all blocks from TSDB have been shipped, TSDB is closed and deleted from local disk. Only used if -blocks-storage.tsdb.close-idle-tsdb-interval is set.")
```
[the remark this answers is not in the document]
pstibrany: Flushing is not enough... We don't want to close TSDBs too early to avoid wrong query results. This setting should be set to the same value (or higher) than `-querier.query-ingesters-within`.

pkg/storage/tsdb/config.go
```diff
@@ -206,6 +212,9 @@ func (cfg *TSDBConfig) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.WALCompressionEnabled, "blocks-storage.tsdb.wal-compression-enabled", false, "True to enable TSDB WAL compression.")
 	f.IntVar(&cfg.WALSegmentSizeBytes, "blocks-storage.tsdb.wal-segment-size-bytes", wal.DefaultSegmentSize, "TSDB WAL segments files max size (bytes).")
 	f.BoolVar(&cfg.FlushBlocksOnShutdown, "blocks-storage.tsdb.flush-blocks-on-shutdown", false, "True to flush blocks to storage on shutdown. If false, incomplete blocks will be reused after restart.")
+	f.DurationVar(&cfg.CloseIdleTSDBInterval, "blocks-storage.tsdb.close-idle-tsdb-interval", 0, "How often to check for idle TSDB that can be closed. 0 means disabled.")
+	f.IntVar(&cfg.CloseIdleTSDBConcurrency, "blocks-storage.tsdb.close-idle-tsdb-concurrency", 10, "Limit the number of concurrently closing TSDB when idle.")
```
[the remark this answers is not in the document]
pstibrany: You're right, sequential should be fine.

pkg/ingester/ingester_v2.go
pstibrany: I considered that, but "not idle" didn't seem like a good error message :)

pkg/storage/tsdb/config.go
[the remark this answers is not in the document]
pracucci: You're right. I would improve the comment about the relationship with `-querier.query-ingesters-within`.

pkg/ingester/ingester_v2.go
```diff
@@ -216,6 +242,37 @@ func (u *userTSDB) setLastUpdate(t time.Time) {
...
+	if !u.isIdle(time.Now(), idleTimeout) {
```
pracucci, line 246: I just realised this check has a problem: the `lastUpdate` is not correct after an ingester restart. An option could be checking the timestamp against `max(lastUpdate, TSDB max time)`.

pstibrany: When TSDB is opened, we set `lastUpdate` to current time to avoid premature idle flush and close. This will only be a problem if ingesters never stay up for longer than configured idle timeout.

pstibrany: Instead of current time, we can set it to last sample time from TSDB when opening, as you suggest.

pracucci: > When TSDB is opened, we set lastUpdate to current time to avoid premature idle flush and close. 

Then we're protected 👍 

> Instead of current time, we can set it to last sample time from TSDB when opening, as you suggest.

Could make sense.

pkg/ingester/metrics.go
pstibrany: I've improved metric removal logic, but will move it out to a separate PR for easier review.

pkg/ingester/ingester_v2.go
```diff
@@ -102,16 +110,25 @@ func (u *userTSDB) StartTime() (int64, error) {
 	return u.db.StartTime()
 }
 
+func (u *userTSDB) setState(from, to tsdbState) bool {
```
jtlisi, line 113: **suggestion:** Consider changing this function call to be `casState` since this looks to be a compare and swap operation

```suggestion
func (u *userTSDB) casState(from, to tsdbState) bool {
```

docs/blocks-storage/store-gateway.md
```diff
@@ -698,6 +698,15 @@ blocks_storage:
     # CLI flag: -blocks-storage.tsdb.flush-blocks-on-shutdown
     [flush_blocks_on_shutdown: <boolean> | default = false]
 
+    # If TSDB has not received any data for this duration, and all blocks from
+    # TSDB have been shipped, TSDB is closed and deleted from local disk. If set
+    # to positive value, this value should be equal or higher than
+    # -querier.query-ingesters-within flag to make sure that TSDB is not closed
```
ranton256, line 704: Does this comment apply on the store-gateway as well, on being greater than query-ingesters-within?

ranton256 (approved):

docs/blocks-storage/store-gateway.md
pstibrany: No, it doesn’t, but store-gateways share the same config struct, which is why it shows up in this documentation as well. We need to fix that.

pstibrany: Store-gateways don’t have TSDB open at all.

pkg/ingester/ingester_v2.go
```diff
@@ -1438,6 +1510,78 @@ func (i *Ingester) compactBlocks(ctx context.Context, force bool) {
...
+			return nil
+		}
+
+		i.closeAndDeleteUserTSDBIfIdle(userID)
+	}
+
+	return nil
+}
+
+func (i *Ingester) closeAndDeleteUserTSDBIfIdle(userID string) {
```
pracucci, line 1525: I think would be great to have a couple of metrics (counters) on succeeded and failed closing idle TSDBs. WDYT?

pracucci (approved):
LGTM, thanks!

jtlisi (approved):
LGTM

pkg/storage/tsdb/config.go
```diff
@@ -209,6 +213,7 @@ func (cfg *TSDBConfig) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.WALCompressionEnabled, "blocks-storage.tsdb.wal-compression-enabled", false, "True to enable TSDB WAL compression.")
 	f.IntVar(&cfg.WALSegmentSizeBytes, "blocks-storage.tsdb.wal-segment-size-bytes", wal.DefaultSegmentSize, "TSDB WAL segments files max size (bytes).")
 	f.BoolVar(&cfg.FlushBlocksOnShutdown, "blocks-storage.tsdb.flush-blocks-on-shutdown", false, "True to flush blocks to storage on shutdown. If false, incomplete blocks will be reused after restart.")
+	f.DurationVar(&cfg.CloseIdleTSDBTimeout, "blocks-storage.tsdb.close-idle-tsdb-timeout", 12*time.Hour, "If TSDB has not received any data for this duration, and all blocks from TSDB have been shipped, TSDB is closed and deleted from local disk. If set to positive value, this value should be equal or higher than -querier.query-ingesters-within flag to make sure that TSDB is not closed prematurely, which could cause invalid query results. 0 or negative value disables closing of idle TSDB.")
```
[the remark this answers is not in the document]
pstibrany: > Is that flag, or `-querier.query-store-after` or both? I think the minimum value is the max between `-querier.query-ingesters-within` and `-querier.query-store-after`.

`-querier.query-store-after` is about querying long-term storage, which in my opinion is unrelated.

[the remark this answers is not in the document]
pracucci: Yes, you're right.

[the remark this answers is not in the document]
pstibrany: > Also 12h default looks a bit high. The default value for `-blocks-storage.tsdb.head-compaction-idle-timeout`, so I think 2h may be safe for this default. In our OSS jsonnet we'll override it according to our setup config.

On the other hand, `-querier.query-ingesters-within` defaults to 0, so perhaps we should disable closing of idle TSDBs by default?

pkg/ingester/ingester_v2.go
```diff
@@ -1438,6 +1510,78 @@ func (i *Ingester) compactBlocks(ctx context.Context, force bool) {
 	})
 }
 
+func (i *Ingester) closeAndDeleteIdleUsers(ctx context.Context) error {
```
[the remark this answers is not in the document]
pstibrany: It's both :) Going with `closeAndDeleteIdleUserTSDBs`.

pkg/storage/tsdb/config.go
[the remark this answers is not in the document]
pracucci: We may want to default it to 0 until we're comfortable it works fine (BTW, we may mark it experimental) but I don't think we should default to 0 because `-querier.query-ingesters-within` is `0`. Blocks are kept in the ingester up to the retention period (not forever), but practically they will be available to query back from store-gateway way before `12h` so I think `12h` is a good default.

pkg/ingester/ingester_v2.go
pstibrany: I've added metrics for number of idle-checks, and various check results. PTAL

pkg/ingester/ingester_v2.go
```diff
@@ -277,6 +369,12 @@ func newTSDBState(bucketClient objstore.Bucket, registerer prometheus.Registerer
 			Help:    "The total time it takes to purge the TSDB series reference cache for a single tenant.",
 			Buckets: prometheus.DefBuckets,
 		}),
+
+		idleTsdbChecks: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
```
pracucci, line 373: I'm not sure this metric is worth it. It looks redundant to me, b
ecause it's equal to `sum(cortex_ingester_idle_tsdb_check_results_total)`.

pracucci: If you end up removing it, then we could rename `cortex_ingester_idle_tsdb_check_results_total` into `cortex_ingester_idle_tsdb_checks_total`.

pstibrany: That's true.

pkg/storage/tsdb/config.go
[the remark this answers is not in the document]
pstibrany: > practically they will be available to query back from store-gateway way before 12h so I think 12h is a good default.

Fair enough. I will keep it at 0 while it's experimental, and let's use non-zero value once we test that it works as expected.

pracucci:
Thanks @pstibrany for addressing my comments. LGTM 👍

PR #3495, opened 2020-11-13
Update auth documentation

### LINKED ISSUE #3493: Tenant-ID confusion
I'm trying to figure out/understand the multi-tenantcy. It's quite unclear for me and the documentation ([Authentication and Authorisation](https://cortexmetrics.io/docs/guides/auth/)) is quite short/high level. After searching through older issues in github, i found links to both the [glossary](https://cortexmetrics.io/docs/guides/glossary/#tenant) and the [limitations](https://cortexmetrics.io/docs/guides/limitations/#tenant-id-naming) which gave me more insights. 

Also the whole concept of multi tentantcy is a bit unclear. lets say i have 2 zones where my prometheis are remote-writing to cortex. They both have their own tenant-id. So it's easy to track how much data each zone/tenant is writing to cortex. But on the query side you should use the same tenant-id if you want to query that data. There is no way to query both zones as far as i'm aware. If this is true, we should use the same tenant-id and find data by the `cluster` label in the metrics. But then i lose the overview of which tenant is writing how much data its sending. But if we should use the same tenant-id, then you can only limit who can/can't access the data, right? Maybe a solution can be to allow for multiple values in the `X-Scope-OrgID` header? not sure if this is feasible. 

- link from [Authentication and Authorisation](https://cortexmetrics.io/docs/guides/auth/) to either/both: [glossary](https://cortexmetrics.io/docs/guides/glossary/#tenant) and/or [limitations](https://cortexmetrics.io/docs/guides/limitations/#tenant-id-naming)
- explain more in depth how it's working
- add tips/tricks
- support for multiple values in `X-Scope-OrgID`?

TomL-dev:
**What this PR does**:

update documentation regarding authentication and authorization

Fixes #3493

**Checklist**
- [ ] Tests updated
- [x] Documentation added


jtlisi (approved):
LGTM, thanks for the PR.

pstibrany (approved):
Thanks.

PR #3499, opened 2020-11-17
Improve logging in the chunk store.

cyriltovena:
This prints statistics and stringify matchers only once.
I realized we are pretty verbose and it doesn't actually add value to logs one stats by one, at the same it's also the hot path and this can consume CPU(around 10% while looking at queriers in Loki).

I've added a benchmark, I'm planning to use this one more in the future to improve index querying memory allocations..

bench diff:

```
benchmark                     old ns/op     new ns/op     delta
Benchmark_GetRefsChunk-16     461135        423417        -8.18%

benchmark                     old allocs     new allocs     delta
Benchmark_GetRefsChunk-16     1723           1539           -10.68%

benchmark                     old bytes     new bytes     delta
Benchmark_GetRefsChunk-16     104342        95305         -8.66%
```

### DIFF
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -411,10 +411,7 @@ func (c *store) getMetricNameChunks(ctx context.Context, userID string, from, th
 
 func (c *store) lookupChunksByMetricName(ctx context.Context, userID string, from, through model.Time, matchers []*labels.Matcher, metricName string) ([]Chunk, error) {
 	log, ctx := spanlogger.New(ctx, "ChunkStore.lookupChunksByMetricName")
-	defer func() {
-
-		log.Finish()
-	}()
+	defer log.Finish()
 
 	// Just get chunks for metric if there are no matchers
 	if len(matchers) == 0 {
--- a/pkg/chunk/chunk_store.go
+++ b/pkg/chunk/chunk_store.go
@@ -265,9 +265,18 @@ func (c *baseStore) LabelValuesForMetricName(ctx context.Context, userID string,
 
 // LabelNamesForMetricName retrieves all label names for a metric name.
 func (c *store) LabelNamesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string) ([]string, error) {
+	var indexTotalChunks, totalFilteredChunks int
 	log, ctx := spanlogger.New(ctx, "ChunkStore.LabelNamesForMetricName")
-	defer log.Span.Finish()
-	level.Debug(log).Log("from", from, "through", through, "metricName", metricName)
+	defer func() {
+		level.Debug(log).Log(
+			"from", from,
+			"through", through,
+			"metricName", metricName,
+			"Chunks in index", indexTotalChunks,
+			"Chunks post filtering", totalFilteredChunks,
+		)
+		log.Span.Finish()
+	}()
 
 	shortcut, err := c.validateQueryTimeRange(ctx, userID, &from, &through)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pstibrany:
> I realized we are pretty verbose and it doesn't actually add value to logs one stats by one

Added value is seeing in the trace when exactly was operation in given stage, which gives you additional timing information in the log and span. This is now lost (it could easily be put back).

I'm personally fine with that, but would like to double-check with @bboreham and other chunks users.

pstibrany:
Another difference is that now we will log that summary message even in the case of error, and it will have incorrect (0) values logged.

cyriltovena:
> Another difference is that now we will log that summary message even in the case of error, and it will have incorrect (0) values logged.



> Added value is seeing in the trace and log when exactly was operation in given stage, which gives you additional timing information. This is now lost (it could easily be put back).

I'm happy to change and keep the trace log, the output log was mostly my concern.

bboreham:
Yes there's a little bit of timing information lost.
I don't have a strong memory of using that information, but generally everything may be useful.

If this is expensive, it should be worth investigating why.  E.g. is it in some boxing operation where all the arguments are made into `interface{}` ?  Maybe some special-purpose methods on spanlogger would help that, skip the work if debug logging is off and the current span is not sampled.

(I don't think it is reasonable to complain about the amount of work done to log when debug is turned on - this is by definition not something you would normally do in production)

cyriltovena:
> I don't have a strong memory of using that information, but generally everything may be useful.

I never used the timing informations, I barely used once to debug the count informations.

I used this log debug at high scale and it takes cpu time as this in the hot path, but also allocations, the problem with gokit is that even if you don't have debug activated it still takes ressources.

I could try to keep the timing but only logging in the span, removing the log entirely, happy to spent some time in this direction if we are ok with that.

Current performance improvement should also be taken into consideration 8~10% improvement, for losing timing is not a bad deal :), specially if you never used it or don't use debug.

bboreham:
> the parameters are evaluated

I don't see anything more complicated than a `len()`, and your change does the same number of those.

Your 8% is 40 microseconds.  Should we look at how come these routines are called so often that that difference is valuable?
If something here is getting called 1,000 times a second we probably don't want to log anything from it, but move the information to a higher level.

cyriltovena:
> > the parameters are evaluated
> 
> I don't see anything more complicated than a `len()`, and your change does the same number of those.
> 

the biggest pain was the formatMatcher, then the actual log, which I improved already by computing this only once. You're right len() wasn't the problem, although still think gokit is.

> Your 8% is 40 microseconds. Should we look at how come these routines are called so often that that difference is valuable?

Our new Loki calls the index in the ingesters, which causes a multiplication of index queries, although local with boltdb, this was showing in cpu and memory profile.

> If something here is getting called 1,000 times a second we probably don't want to log anything from it, but move the information to a higher level.

That was my first assumption, I thought we didn't really needed this anymore, that's why I packed everything into a single log.
I could remove debug log too, but  I like to run debug in production and  level.Debug() is also taking ressources even if it doesn't output anything.

There's definitely other part to improve, but this one was a low hanging fruit.

Should I just roll back and remove only the multiple occurence of formatMatcher ?

bboreham:
I don't know what formatMatcher is.  

Definitely sounds like you want to have a different path where this info isn't logged in detail.
Maybe create a new "spam" log level?

cyriltovena:
This is loki ingester index queries:

image


As you can see quite some CPU time in the log of the span and gokit, but also chunk.formatMatcher. Allocated object look pretty much the same.

I'll appreciate consideration if you don't use it, turning off debug is not giving much benefits.

bboreham:
Looks to me like (a) most of the time goes in the actual logging, not in evaluating parameters.

My sense is you don't want this info logged at all, and I would support a flag that gives you that option.

cyriltovena:
I think Cortex has enough feature flags, I don't think it's bad choosing what is important to log for the user, I've tried to get best of both world here, log less and keep relevant info.

bboreham:
I don't mean a user-visible flag, I mean in the code.

bboreham:
I don't see this as a "best of both worlds" thing, I see that you are degrading the information available in one case, and logging millions of lines you are never going to look at in the other case.  Is that incorrect?

PR #3507, opened 2020-11-18
Alertmanager: Unregister tenant-specific metrics when the configuration is removed 

gotjosh:
**What this PR does**:

Unregisters tenant-specific metrics for the Alertmanager when the configuration is removed.

**Checklist**
- [x] Tests updated
- N/A Documentation added


gouthamve:
Why are we still doing active and non-active? Can we just remove the tenant from memory?

gotjosh:
@gouthamve Do we need that now? That'll probably be a more involved refactor and would like to avoid it, if possible. With the new proposal, I'm about to submit this is all going away anyways.

pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -151,6 +151,13 @@ func (m *alertmanagerMetrics) addUserRegistry(user string, reg *prometheus.Regis
 	m.regsMu.Unlock()
 }
 
+func (m *alertmanagerMetrics) deleteUserRegistry(user string) {
+	m.regsMu.Lock()
+	defer m.regsMu.Unlock()
+
+	delete(m.regs, user)
+}
+
```
pstibrany, line 160: This only works properly if no aggregate metrics are reported. See `TestMetricsArePerUser` test in ruler for how this can be verified, or wait until https://github.com/cortexproject/cortex/pull/3511 lands for "soft" deletion of metrics, if you need aggregates (doesn't seem like it though).

pkg/alertmanager/multitenant.go
```diff
@@ -430,11 +431,18 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
...
 		err := existing.ApplyConfig(cfg.User, userAmConfig, rawCfg)
 		if err != nil {
 			return fmt.Errorf("unable to apply Alertmanager config for user %v: %v", cfg.User, err)
 		}
+		activeAfter := existing.IsActive()
+
+		// If we went from paused to active, we need to re-add the user-regitry.
+		if !activeBefore && activeAfter {
+			am.alertmanagerMetrics.addUserRegistry(cfg.User, existing.registry)
+		}
```
jtlisi, line 445: **suggestion(if-minor):**
```suggestion

		// If we went from paused to active, we need to re-add the user-regitry.
		if !activeBefore && existings.IsActive() {
			am.alertmanagerMetrics.addUserRegistry(cfg.User, existing.registry)
		}
```

pkg/alertmanager/alertmanager_metrics.go
jtlisi: The alertmanager aggregates a significant number of its metrics so this PR is blocked by #3511

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## master / unreleased
 
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
+* [CHANGE] Alertmanager: Tenant-specific metrics will be unregistered when the alertmanager configuration is removed. #XXXX
```
jtlisi, line 6: **suggestion(non-blocking):** 
```suggestion
* [CHANGE] Alertmanager: Per-tenant metrics will be unregistered when the alertmanager configuration for a tenant is removed. #3507
```

`s/Tenant-specific/Per-tenant/g`

pkg/alertmanager/alertmanager_metrics.go
pracucci: Given #3511 has been merged, is there any reason why this PR is not using the new `util.UserRegistries` and 9 RemoveUserRegistry()`?

jtlisi: IMO, It would make sense to use soft deletes now that #3511 is merged.

gotjosh: I actually went for soft as well.

jtlisi (approved):
LGTM

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [CHANGE] Blocks storage: increased `-blocks-storage.bucket-store.chunks-cache.attributes-ttl` default from `24h` to `168h` (1 week). #3528
 * [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts. #3305
 * [ENHANCEMENT] Added zone-awareness support on queries. When zone-awareness is enabled, queries will still succeed if all ingesters in a single zone will fail. #3414
+* [CHANGE] Alertmanager: Tenant-specific metrics will be unregistered when the alertmanager configuration is removed. #3507
```
pracucci, line 9: We've just cut `1.6.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry to the top, under the "master / unreleased" section, please?

pkg/alertmanager/multitenant.go
```diff
@@ -435,11 +436,17 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
...
+		activeBefore := existing.IsActive()
 		// If the config changed, apply the new one.
 		err := existing.ApplyConfig(cfg.User, userAmConfig, rawCfg)
 		if err != nil {
 			return fmt.Errorf("unable to apply Alertmanager config for user %v: %v", cfg.User, err)
 		}
+
+		// If we went from paused to active, we need to re-add the user-regitry.
+		if !activeBefore && existing.IsActive() {
+			am.alertmanagerMetrics.addUserRegistry(cfg.User, existing.registry)
```
pracucci, line 448: If I'm not missing anything, this should not be correct. The problem is that `existing.registry` is the previous Prometheus registry and not a new one. This means that the registered metrics are there since before `deleteUserRegistry()` has been called and this would lead to counters/histograms/summaries to be double tracked as soon as you re-add it via `addUserRegistry()`. Why? Because when we call `deleteUserRegistry()` the last value of all counters/histograms/summaries is "recorded" in the `UserRegistries`. When you add back the same registry via `addUserRegistry()`, the metrics in the registry will be summed up to the last value of the metrics when `deleteUserRegistry()` was called, but if the registry is the same you will end up with double countings.

Assuming we fix this issue, the only benefit I see of this PR is let **gauge metrics** values (only gauges) disappear from aggregated metrics when `deleteUserRegistry()`. If this is the desiderata of this PR, then to fix it we should use a new registry each time we call `addUserRegistry()`.

_Am I missing anything?_

gotjosh: Thanks, Marco for pointing this out - I think you're right on the soft deletion part. We need new registry.

My intention was to remove _all_ metrics, but under the current circumstances (where we can't distinguish between a user moving away from a shard or being removed completely from the system) if this is the best we can do (without incurring in more complexity) - I'll take it.

pkg/alertmanager/multitenant.go
```diff
@@ -435,11 +436,17 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
...
+		activeBefore := existing.IsActive()
 		// If the config changed, apply the new one.
 		err := existing.ApplyConfig(cfg.User, userAmConfig, rawCfg)
 		if err != nil {
 			return fmt.Errorf("unable to apply Alertmanager config for user %v: %v", cfg.User, err)
 		}
+
+		// If we went from paused to active, we need to re-add an user-regitry.
+		if !activeBefore && existing.IsActive() {
+			am.alertmanagerMetrics.addUserRegistry(cfg.User, prometheus.NewRegistry())
```
pracucci, line 448: I can't understand this change honestly. You're just setting a new prometheus registry to `alertmanagerMetrics` but that registry will never get alertmanager metrics registered on.

I still believe this PR has not been conceived correctly. What I think you should do is completely release the alertmanager resources once a tenant is removed from this replica, and create a new alertmanager (with a new registry which is then registered to `alertmanagerMetrics`) when the tenant is re-added back. Basically this should completely remove the `IsActive()` logic.

gotjosh: > I can't understand this change honestly. You're just setting a new prometheus registry to alertmanagerMetrics but that registry will never get alertmanager metrics registered on.

Sorry Marco, I completely left this PR in an abandoned state but my ultimate intention was to close it. I quickly realised exactly what you pointed out that the new registry is never initialised with the corresponding metrics.

Given I'm doing work around the area with the scalability, and completely removing the `IsActive` is a bit more involved than I initially expect, I'll close this and defer to later.

PR #3508, opened 2020-11-18
Query frontend with scheduler fixes

pstibrany:
**What this PR does**: Small fixes to query-frontend code:
- use context from the stream, not just worker context (in practice, it doesn't make a difference – it doesn't detect closed connection to the scheduler earlier)
- check for returned status when forwarding cancellation request
- fix docker-compose config for querier-with-scheduler
- added GoLand launch configs for querier-scheduler and query-frontend-with-scheduler

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci (approved):
We pair investigated it, so LGTM ;)

PR #3511, opened 2020-11-19
Support removal of registry from UserRegistries

pstibrany:
**What this PR does**: This PR enhances `UserRegistries` to support "soft" removal of user-specific registry, and registration of new registry for user. Soft-removal preserves latest metric values (except for gauges) in order to avoid counter resets. This is required to keep metrics consistent when moving user around different components due to resharding.

It could be argued that once we have this, we don't need "hard" delete, as used now by Ruler.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/metrics_helper.go
+++ b/pkg/util/metrics_helper.go
@@ -142,15 +142,7 @@ func (d MetricFamiliesPerUser) SendSumOfCountersWithLabels(out chan<- prometheus
 }
 
 func (d MetricFamiliesPerUser) SendSumOfCountersPerUser(out chan<- prometheus.Metric, desc *prometheus.Desc, counter string) {
-	for _, userEntry := range d {
-		if userEntry.user == "" {
-			continue
-		}
-
-		v := userEntry.metrics.SumCounters(counter)
-
-		out <- prometheus.MustNewConstMetric(desc, prometheus.CounterValue, v, userEntry.user)
-	}
+	d.SendSumOfCountersPerUserWithLabels(out, desc, counter)
 }
 
 // SendSumOfCountersPerUserWithLabels provides metrics with the provided label names on a per-user basis. This function assumes that `user` is the
@@ -182,6 +174,10 @@ func (d MetricFamiliesPerUser) SendSumOfGauges(out chan<- prometheus.Metric, des
 
 func (d MetricFamiliesPerUser) SendSumOfGaugesWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, gauge string, labelNames ...string) {
 	d.sumOfSingleValuesWithLabels(gauge, gaugeValue, labelNames).WriteToMetricChannel(out, desc, prometheus.GaugeValue)
+}
+
+func (d MetricFamiliesPerUser) SendSumOfGaugesPerUser(out chan<- prometheus.Metric, desc *prometheus.Desc, gauge string) {
+	d.SendSumOfGaugesPerUserWithLabels(out, desc, gauge)
 }
 
 // SendSumOfGaugesPerUserWithLabels provides metrics with the provided label names on a per-user basis. This function assumes that `user` is the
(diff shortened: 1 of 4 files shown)


pkg/util/metrics_helper.go
```diff
@@ -507,27 +526,71 @@ func NewUserRegistries() *UserRegistries {
...
 	r.regs = append(r.regs, UserRegistry{
 		user: user,
 		reg:  reg,
 	})
 }
 
+func (r *UserRegistries) softRemoveUserRegistry(ur *UserRegistry) {
+	last, err := ur.reg.Gather()
+	if err != nil {
+		level.Warn(Logger).Log("msg", "failed to gather metrics from registry", "user", ur.user, "err", err)
```
pracucci, line 553: Should we hard-delete instead of proceeding with an empty `last`?

pkg/util/metrics_helper.go
```diff
@@ -507,27 +526,71 @@ func NewUserRegistries() *UserRegistries {
...
+			ix++
+		} else {
+			// Remove gauges and unknowns.
+			last = append(last[:ix], last[ix+1:]...)
+		}
+	}
+
+	ur.lastGather, err = NewMetricFamilyMap(last)
+	if err != nil {
+		level.Warn(Logger).Log("msg", "failed to gather metrics from registry", "user", ur.user, "err", err)
```
pracucci, line 569: Should we hard-delete instead of proceeding with an empty `lastGather`?

pkg/util/metrics_helper_test.go
```diff
@@ -508,6 +510,463 @@ func TestFloat64PrecisionStability(t *testing.T) {
...
+			summary_user_sum{user="2"} 10
+			summary_user_count{user="2"} 5
+			summary_user_sum{user="4"} 20
+			summary_user_count{user="4"} 5
+			summary_user_sum{user="5"} 25
+			summary_user_count{user="5"} 5
+	`)))
+}
+
+func TestReplaceRegistry(t *testing.T) {
```
pracucci, line 776:
```suggestion
func TestUserRegistries_AddUserRegistry_ReplaceRegistry(t *testing.T) {
```

pkg/util/metrics_helper_test.go
```diff
@@ -508,6 +510,463 @@ func TestFloat64PrecisionStability(t *testing.T) {
...
+			summary_user_count{user="1"} 5
+			summary_user_sum{user="2"} 10
+			summary_user_count{user="2"} 5
+			summary_user_sum{user="4"} 20
+			summary_user_count{user="4"} 5
+			summary_user_sum{user="5"} 25
+			summary_user_count{user="5"} 5
+	`)))
+}
+func TestRemoveUserRegistryHard(t *testing.T) {
```
pracucci, line 684:
```suggestion
func TestUserRegistries_RemoveUserRegistry_HardRemoval(t *testing.T) {
```

pkg/util/metrics_helper_test.go
```diff
@@ -508,6 +510,463 @@ func TestFloat64PrecisionStability(t *testing.T) {
...
+		summary_user_sum{user="3"} 15
+		summary_user_count{user="3"} 5
+		summary_user_sum{user="4"} 20
+		summary_user_count{user="4"} 5
+		summary_user_sum{user="5"} 25
+		summary_user_count{user="5"} 5
+	`)))
+}
+
+func TestRemoveUserRegistrySoft(t *testing.T) {
```
pracucci, line 593:
```suggestion
func TestUserRegistries_RemoveUserRegistry_SoftRemoval(t *testing.T) {
```

pkg/util/metrics_helper_test.go
```diff
@@ -508,6 +510,463 @@ func TestFloat64PrecisionStability(t *testing.T) {
 	}
 }
 
+func TestMetrics(t *testing.T) {
```
pracucci, line 513: What is this test about? Is it really worth to keep it? If yes, I would suggest to rename it to make it more clear to understand.

pracucci (approved):
Very good job! The design you picked ended up to an implementation simpler than I thought. I left few non-blocking comments.

pkg/util/metrics_helper_test.go
pstibrany: This is a baseline for other tests, for comparison. I'd keep it, I'll try to rename it with other tests.

pracucci:
@pstibrany Still LGTM, but please be aware `TestUserRegistries_AddUserRegistry_ReplaceRegistry` is now failing.

pstibrany:
> @pstibrany Still LGTM, but please be aware `TestUserRegistries_AddUserRegistry_ReplaceRegistry` is now failing.

Thanks! I've actually introduced a bug, and test caught it. Phew :)

PR #3512, opened 2020-11-19
Refactor consistency check so it can be reused

gouthamve:
This is in anticipation of thanos-io/thanos#3469 where we would need to
utilise the consistency check to query for labels as well.


pracucci (approved):
Very good refactoring! LGTM 👏

PR #3515, opened 2020-11-19
Improve allocations while doing index queries

cyriltovena:
This contains 3 optimizations.

1 - Reuse the iterator when iterating through batches
2 - Avoid the string copy when parsing the data from the index.
3 - Use an intermediary `IndexQueryEntry` that does not contains the tableName and hashValue strings for lookupEntriesByQueries that doesn't require it. 

Hard to get a non noisy benchmark but that's how it looks like before and after:

```
benchmark                                   old ns/op     new ns/op     delta
Benchmark_GetRefsChunkWithManyChunks-16     73513148      72606738      -1.23%

benchmark                                   old allocs     new allocs     delta
Benchmark_GetRefsChunkWithManyChunks-16     741487         641464         -13.49%

benchmark                                   old bytes     new bytes     delta
Benchmark_GetRefsChunkWithManyChunks-16     78208523      64817304      -17.12%
```

Performance is unchanged but allocation is reduced.

Why I'm doing this because I found out that we do allocate a lot in size and object.

image


pkg/chunk/storage_client.go
```diff
@@ -48,7 +48,7 @@ type WriteBatch interface {
 
 // ReadBatch represents the results of a QueryPages.
 type ReadBatch interface {
-	Iterator() ReadBatchIterator
+	Iterator(it ReadBatchIterator) ReadBatchIterator
```
pstibrany, line 51: Please document why this parameter is used, and if `nil` is allowed.

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -587,11 +587,17 @@ type dynamoDBReadResponse struct {
...
-		i:                    -1,
-		dynamoDBReadResponse: b,
+func (b *dynamoDBReadResponse) Iterator(iter chunk.ReadBatchIterator) chunk.ReadBatchIterator {
+	if iter == nil {
+		return &dynamoDBReadResponseIterator{
+			i:                    -1,
+			dynamoDBReadResponse: b,
+		}
 	}
+	concrete := iter.(*dynamoDBReadResponseIterator)
```
pstibrany, line 597: This will panic when used with different implementation. Following lines will panic if `concrete` is `nil`, but `iter` was not `nil`.

pkg/chunk/aws/dynamodb_storage_client.go
```diff
@@ -587,11 +587,17 @@ type dynamoDBReadResponse struct {
...
+	concrete.i = -1
+	concrete.dynamoDBReadResponse = b
+	return concrete
```
pstibrany, line 600: Returning `iter` would avoid extra interface conversion. (Unit test and benchmark would be nice).

pkg/chunk/purger/delete_requests_store.go
```diff
@@ -212,8 +212,9 @@ func (ds *DeleteStore) GetPendingDeleteRequestsForUser(ctx context.Context, user
 
 func (ds *DeleteStore) queryDeleteRequests(ctx context.Context, deleteQuery []chunk.IndexQuery) ([]DeleteRequest, error) {
 	deleteRequests := []DeleteRequest{}
+	var itr chunk.ReadBatchIterator
 	err := ds.indexClient.QueryPages(ctx, deleteQuery, func(query chunk.IndexQuery, batch chunk.ReadBatch) (shouldContinue bool) {
-		itr := batch.Iterator()
+		itr = batch.Iterator(itr)
```
pstibrany, line 217: I think this is introducing a bug, if there are multiple `delete queries`, since this iterator will now be shared between multiple concurrently running goroutines.

If there is only one, there is no need to reuse iterator.

Access to `deleteRequests` is not protected either. :confused:

pkg/chunk/storage/caching_index_client.go
```diff
@@ -134,13 +134,13 @@ func (s *cachingIndexClient) QueryPages(ctx context.Context, queries []chunk.Ind
...
 	}
-
+	var iter chunk.ReadBatchIterator
 	err = s.IndexClient.QueryPages(ctx, cacheableMissed, func(cacheableQuery chunk.IndexQuery, r chunk.ReadBatch) bool {
 		resultsMtx.Lock()
 		defer resultsMtx.Unlock()
 		key := queryKey(cacheableQuery)
 		existing := results[key]
-		for iter := r.Iterator(); iter.Next(); {
+		for iter = r.Iterator(iter); iter.Next(); {
```
pstibrany, line 143: Bug: Reuses iterator between concurrent goroutines.

pstibrany (commented):
> This contains 3 optimizations.

I would have preferred to see three separate PRs with their own bechmarks. One referenced in this PR isn't too convincing.

> 1 - Reuse the iterator when iterating through batches

This has introduced bug everywhere where `QueryPages` is used with multiple queries, and implementations run them concurrently. (Which is all implementations, except testing ones.) In the end, this optimization may work only at very few places.

> 2 - Avoid the string copy when parsing the data from the index.

I would prefer to keep using standard library for conversion in `chunk.go`. Stdlib is typically optimized faster and hand-in-hand with runtime optimizations. By switching to 3rd party library for such basic operations, we're risking not getting those anymore.

As to the change, personally I'm not a fan of replacing strings with `[]byte`, as it brings complexity to the code (security and concurrency aspects are completely different) That said, I can see how this specific optimization may be useful.

> 3 - Use an intermediary `IndexQueryEntry` that does not contains the tableName and hashValue strings for lookupEntriesByQueries that doesn't require it.

This saves 32 bytes per instance. For 1M entries, that's 32 MB. On first sight it looks like very tiny optimization to have a major impact. Can you provide some context where this is useful?

pkg/chunk/purger/delete_requests_store.go
cyriltovena: I sent a PR to rework this function to accept only one request.

pkg/chunk/storage/caching_index_client.go
cyriltovena: I don't think that's true, but I agree that it's not super clear and may be dangerous to maintain. There's a lock.

cyriltovena:
> > This contains 3 optimizations.
> 
> I would have preferred to see three separate PRs with their own bechmarks. One referenced in this PR isn't too convincing.
> 
> > 1 - Reuse the iterator when iterating through batches
> 
> This has introduced bug everywhere where `QueryPages` is used with multiple queries, and implementations run them concurrently. (Which is all implementations, except testing ones.) In the end, this optimization may work only at very few places.
> 
> > 2 - Avoid the string copy when parsing the data from the index.
> 
> I would prefer to keep using standard library for conversion in `chunk.go`. Stdlib is typically optimized faster and hand-in-hand with runtime optimizations. By switching to 3rd party library for such basic operations, we're risking not getting those anymore.
> 
> As to the change, personally I'm not a fan of replacing strings with `[]byte`, as it brings complexity to the code (security and concurrency aspects are completely different) That said, I can see how this specific optimization may be useful.
> 
> > 3 - Use an intermediary `IndexQueryEntry` that does not contains the tableName and hashValue strings for lookupEntriesByQueries that doesn't require it.
> 
> This saves 32 bytes per instance. For 1M entries, that's 32 MB. On first sight it looks like very tiny optimization to have a major impact. Can you provide some context where this is useful?

The problem we have in Loki is that compare to cortex, we have many chunks in a small time frame, and so index queries are very expensive because they returns a lot of hits. Now I think I'm going to park this PR because it gave me another idea. I think the problem is more how GetRefChunk works. Most of the time, I need only the first few refs. I'll see if I can refactor this way instead.

I agree that some of those optimization are not yielding enough improvement.

pkg/chunk/storage/caching_index_client.go
pstibrany: You are right, I missed the lock here.

cyriltovena:
More context 

image


I've pin down the problem. With actually honor the timerange on the labels and series API. Cortex never did on the chunk store. Still thinking a lot about this.

PR #3516, opened 2020-11-19
Improve delete index query signature.

cyriltovena:
At first I thought `queryDeleteRequests` was creating a race because it can be called with multiple index query and querypage from the index client can run in parallel.

However then I realized that it is always used with a single query and so it will always run sequentially.

So I changed the signature to make sure we never use this with multiple queries in the future.


jtlisi (approved):
LGTM

pstibrany (approved):

PR #3520, opened 2020-11-20
Query Label Names and Values from blocks store

### LINKED ISSUE #2681: Labels and series lookups only query ingesters
https://github.com/cortexproject/cortex/blob/master/pkg/ingester/ingester.go#L833-L834

> // TODO Right now we ignore start and end.

This breaks away from the Prometheus API standard for /api/v1/series (https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers) which causes problems with things like Grafana generating variable values from labels (https://grafana.com/docs/grafana/latest/features/datasources/prometheus/#templating) for something not occurring in the current timeframe

gouthamve:

**What this PR does**:

Fixes #2681

**Checklist**
- [x] Tests updated
- [x] Documentation added


integration/querier_test.go
```diff
@@ -590,7 +590,7 @@ func testMetadataQueriesWithBlocksStorage(
 				for _, val := range lvt.resp {
 					exp = append(exp, model.LabelValue(val))
 				}
-				require.ElementsMatch(t, exp, labelsRes)
+				require.Equal(t, exp, labelsRes)
```
pracucci, line 593: We were using `ElementsMatch()` to avoid flakyness if results are not sorted as the expected value. Is it different now?

pkg/storegateway/bucket_stores.go
```diff
@@ -247,6 +247,42 @@ func (u *BucketStores) Series(req *storepb.SeriesRequest, srv storepb.Store_Seri
...
+	if store == nil {
+		return nil, nil
+	}
+
+	return store.LabelNames(ctx, req)
+}
+
+// LabelValues implements the Storegateway proto service.
+func (u *BucketStores) LabelValues(ctx context.Context, req *storepb.LabelValuesRequest) (*storepb.LabelValuesResponse, error) {
+	spanLog, spanCtx := spanlogger.New(ctx, "BucketStores.LabelNames")
```
pracucci, line 270:
```suggestion
	spanLog, spanCtx := spanlogger.New(ctx, "BucketStores.LabelValues")
```

pkg/storegateway/bucket_stores.go
```diff
@@ -247,6 +247,42 @@ func (u *BucketStores) Series(req *storepb.SeriesRequest, srv storepb.Store_Seri
...
+	defer spanLog.Span.Finish()
+
+	userID := getUserIDFromGRPCContext(spanCtx)
+	if userID == "" {
+		return nil, fmt.Errorf("no userID")
+	}
+
+	store := u.getStore(userID)
+	if store == nil {
+		return nil, nil
```
pracucci, line 262: The caller code (in the querier expects either an error or a `storepb.LabelNamesResponse`, otherwise it panics (I think). Why don't we return an empty response here?

pkg/storegateway/bucket_stores.go
```diff
@@ -247,6 +247,42 @@ func (u *BucketStores) Series(req *storepb.SeriesRequest, srv storepb.Store_Seri
...
```
pracucci, line 280: The caller code (in the querier expects either an error or a storepb.LabelValuesResponse, otherwise it panics (I think). Why don't we return an empty response here?

pkg/querier/blocks_store_queryable.go
```diff
@@ -292,14 +294,74 @@ func (q *blocksStoreQuerier) Select(_ bool, sp *storage.SelectHints, matchers ..
 	return q.selectSorted(sp, matchers...)
 }
 
-func (q *blocksStoreQuerier) LabelValues(name string) ([]string, storage.Warnings, error) {
-	// Cortex doesn't use this. It will ask ingesters for metadata.
-	return nil, nil, errors.New("not implemented")
+func (q *blocksStoreQuerier) LabelNames() ([]string, storage.Warnings, error) {
```
pracucci, line 297: I would like to see unit tests for this. You could get inspired by `Select()` unit tests.

pkg/querier/blocks_store_queryable.go
```diff
@@ -292,14 +294,74 @@ func (q *blocksStoreQuerier) Select(_ bool, sp *storage.SelectHints, matchers ..
...
+		return nil, nil, err
+	}
+
+	return strutil.MergeSlices(resNameSets...), resWarnings, nil
 }
 
-func (q *blocksStoreQuerier) LabelNames() ([]string, storage.Warnings, error) {
-	// Cortex doesn't use this. It will ask ingesters for metadata.
-	return nil, nil, errors.New("not implemented")
+func (q *blocksStoreQuerier) LabelValues(name string) ([]string, storage.Warnings, error) {
```
pracucci, line 332: I would like to see unit tests for this. You could get inspired by `Select()` unit tests.

pkg/querier/querier.go
```diff
@@ -373,11 +377,96 @@ func (q querier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Mat
...
+	if !q.queryStoreForLabels {
+		return q.metadataQuerier.LabelValues(name)
+	}
+
+	if len(q.queriers) == 1 {
+		return q.queriers[0].LabelValues(name)
+	}
+
+	// Using an errgroup here instead of channels, etc because this
+	// is a better model imo and we should move to this everywhere.
```
pracucci, line 389: This looks more part of the PR description than a code comment. I would remove it.

pkg/querier/querier.go
```diff
@@ -373,11 +377,96 @@ func (q querier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Mat
...
+		// Need to reassign as the original variable will change and can't be relied on in a goroutine.
+		querier := querier
+		g.Go(func() error {
+			myValues, myWarnings, err := querier.LabelValues(name)
+			if err != nil {
+				return err
+			}
+
+			// We need values to be sorted we can merge them.
+			sort.Strings(myValues)
```
pracucci, line 408: May you confirm store-gateway and "distributor" don't return it already sorted, right? If store-gateway does, then we could add it to the API spec (response is guaranteed to be sorted) and we could just sort it in the "distributor".

_Same comment applies to `LabelNames()`_

pkg/querier/querier.go
```diff
@@ -373,11 +377,96 @@ func (q querier) Select(_ bool, sp *storage.SelectHints, matchers ...*labels.Mat
...
+	if !q.queryStoreForLabels {
+		return q.metadataQuerier.LabelNames()
+	}
+
+	if len(q.queriers) == 1 {
+		return q.queriers[0].LabelNames()
+	}
+
+	// Using an errgroup here instead of channels, etc because this
+	// is a better model imo and we should move to this everywhere.
```
pracucci, line 437: No need for this comment here either.

pkg/querier/blocks_store_queryable.go
```diff
@@ -292,14 +294,74 @@ func (q *blocksStoreQuerier) Select(_ bool, sp *storage.SelectHints, matchers ..
...
+
+		return queriedBlocks, nil
+	}
+
+	err := q.queryWithConsistencyCheck(spanCtx, spanLog, minT, maxT, queryFunc)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return strutil.MergeSlices(resNameSets...), resWarnings, nil
```
pracucci, line 329: Each set in `resNameSets` must be already sorted to work correctly with `strutil.MergeSlices()`. Are they?

_Same comment applies to `LabelValues()`_

pkg/querier/blocks_store_queryable.go
```diff
@@ -584,6 +646,160 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(
...
+	name string,
+	clients map[BlocksStoreClient][]ulid.ULID,
+	minT int64,
+	maxT int64,
+) ([][]string, storage.Warnings, []ulid.ULID, error) {
+	var (
+		reqCtx        = grpc_metadata.AppendToOutgoingContext(ctx, cortex_tsdb.TenantIDExternalLabel, q.userID)
+		g, gCtx       = errgroup.WithContext(reqCtx)
+		mtx           = sync.Mutex{}
+		nameSets      = [][]string{}
```
pracucci, line 735: I think this should be called `valueSets`

pkg/querier/blocks_store_queryable.go
```diff
@@ -584,6 +646,160 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(
...
+	// Concurrently fetch series from all clients.
+	for c, blockIDs := range clients {
+		// Change variables scope since it will be used in a goroutine.
+		c := c
+		blockIDs := blockIDs
+
+		g.Go(func() error {
+			req, err := createLabelValuesRequest(minT, maxT, name, blockIDs)
+			if err != nil {
+				return errors.Wrapf(err, "failed to create label names request")
```
pracucci, line 750:
```suggestion
				return errors.Wrapf(err, "failed to create label values request")
```

pkg/querier/blocks_store_queryable.go
```diff
@@ -584,6 +646,160 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(
...
+			valuesResp, err := c.LabelValues(gCtx, req)
+			if err != nil {
+				return errors.Wrapf(err, "failed to fetch series from %s", c)
+			}
+
+			myQueriedBlocks := []ulid.ULID(nil)
+			if valuesResp.Hints != nil {
+				hints := hintspb.LabelValuesResponseHints{}
+				if err := types.UnmarshalAny(valuesResp.Hints, &hints); err != nil {
+					return errors.Wrapf(err, "failed to unmarshal label names hints from %s", c)
```
pracucci, line 762:
```suggestion
					return errors.Wrapf(err, "failed to unmarshal label values hints from %s", c)
```

pkg/querier/blocks_store_queryable.go
```diff
@@ -584,6 +646,160 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(
...
+
+				ids, err := convertBlockHintsToULIDs(hints.QueriedBlocks)
+				if err != nil {
+					return errors.Wrapf(err, "failed to parse queried block IDs from received hints")
+				}
+
+				myQueriedBlocks = ids
+			}
+
+			level.Debug(spanLog).Log("msg", "received label names from store-gateway",
```
pracucci, line 773:
```suggestion
			level.Debug(spanLog).Log("msg", "received label values from store-gateway",
```

pkg/querier/blocks_store_queryable.go
```diff
@@ -584,6 +646,160 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(
...
+				"instance", c,
+				"num labels", len(valuesResp.Values),
```
pracucci, line 775:
```suggestion
				"num values", len(valuesResp.Values),
```

pkg/querier/blocks_store_queryable.go
```diff
@@ -611,6 +827,69 @@ func createSeriesRequest(minT, maxT int64, matchers []storepb.LabelMatcher, skip
 	}, nil
 }
 
+func createLabelNamesRequest(minT, maxT int64, blockIDs []ulid.ULID) (*storepb.LabelNamesRequest, error) {
+	req := &storepb.LabelNamesRequest{
+		Start: minT,
+		End:   maxT,
+	}
+
+	if len(blockIDs) == 0 {
```
pracucci, line 836: Is this a real case? When does it happen?

pkg/querier/blocks_store_queryable.go
```diff
@@ -611,6 +827,69 @@ func createSeriesRequest(minT, maxT int64, matchers []storepb.LabelMatcher, skip
...
+				Type:  storepb.LabelMatcher_RE,
+				Name:  block.BlockIDLabel,
+				Value: strings.Join(convertULIDsToString(blockIDs), "|"),
+			},
+		},
+	}
+
+	anyHints, err := types.MarshalAny(hints)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to marshal request hints")
```
pracucci, line 853:
```suggestion
		return nil, errors.Wrapf(err, "failed to marshal label names request hints")
```

pkg/querier/blocks_store_queryable.go
```diff
@@ -611,6 +827,69 @@ func createSeriesRequest(minT, maxT int64, matchers []storepb.LabelMatcher, skip
...
```
pracucci, line 885:
```suggestion
		return nil, errors.Wrapf(err, "failed to marshal label values request hints")
```

pracucci (commented):
Good job! This work is pretty solid 👏  I left some comments. Most are minor thing and small improvements, but there's a comment about the ordering requirement of `strutil.MergeSlices()` which I would like you to investigate and also I would be glad to see some unit tests on blocks store queryable. Thanks! 🚀

pkg/querier/blocks_store_queryable.go
gouthamve: When Thanos querier queries our storegateways. I know it's a rare-case (maybe even non-existent) and can remove it if you'd like.

gouthamve:
@pracucci Thanks for the detailed review! I fixed all the suggestions and added the tests. This is good for another round of review. I'll dig into and add comments about the sortedness of names and values shortly.

pkg/querier/blocks_store_queryable_test.go
```diff
@@ -544,6 +545,474 @@ func TestBlocksStoreQuerier_Select(t *testing.T) {
...
+				cortex_querier_storegateway_refetches_per_query_count 1
+			`,
+		},
+	}
+
+	for testName, testData := range tests {
+		t.Run(testName, func(t *testing.T) {
+			// Splitting it because we need a new registry for names and values.
+			// And also the initial expectedErr checking needs to be done for both.
+			for _, testFunc := range []string{"LabelNames", "LabelValues"} {
```
gouthamve, line 958: I feel like there should be a saner approach than this but my brain can't come up with any. Any ideas @pracucci / @pstibrany?

pracucci: I think you did great. I would have probably done the same. We have other tests where we adopt this approach. LGTM.

pracucci (commented):
Thanks @gouthamve for addressing my feedback! LGTM (modulo a couple of nits). Looking forward to your comment about sorting and then we can go 🚀

pstibrany (commented):
Sorry for late review. Overall it looks good to me (with Marco's comments fixed). I'd like to see less duplicated code, if possible, but that can also be addressed later.

integration/querier_test.go
gouthamve: Yes its different now. We return sorted label values now.

pkg/querier/querier.go
gouthamve: I have this change made in the distributor, but but the querier interface we rely on is the Prometheus interface which doesn't guarantee sorted-ness.

We can always say in Cortex we always return sorted, but not sure if it will bite us sometime in the future.

gouthamve:
@pracucci I've addressed all your comments, but @pstibrany I kinda kept the suggestion of deduplicating code for another PR. I'll soon open a follow up PR for this once I have some bandwidth.

pracucci (approved):
LGTM, thanks!

PR #3521, opened 2020-11-20
Add chunk.EncodeTo to allow to pass your own buffer.

cyriltovena:
Useful if you want to re-use memory after flushing data.

### DIFF
--- a/pkg/chunk/chunk.go
+++ b/pkg/chunk/chunk.go
@@ -183,8 +183,7 @@ var writerPool = sync.Pool{
 
 // Encode writes the chunk into a buffer, and calculates the checksum.
 func (c *Chunk) Encode() error {
-	var buf bytes.Buffer
-	return c.EncodeTo(&buf)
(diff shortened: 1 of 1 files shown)


pkg/chunk/chunk.go
```diff
@@ -184,7 +184,14 @@ var writerPool = sync.Pool{
 // Encode writes the chunk into a buffer, and calculates the checksum.
 func (c *Chunk) Encode() error {
 	var buf bytes.Buffer
+	return c.EncodeTo(&buf)
```
pracucci, line 187: Can we just do `return c.EncodeTo(nil)`?

pracucci (approved):
LGTM (with a nit)

pkg/chunk/chunk.go
cyriltovena: yup definitively.

jtlisi (approved):
LGTM

PR #3527, opened 2020-11-23
Scan tenants upfront in the querier.BlocksScanner in order to improve its cacheability

pracucci:
**What this PR does**:
A "list objects" API call is called once the iteration is done, but it's cached with the timestamp of when the iteration has started. If the iteration is slow, the caching of the "list objects" API response is not much effective. Similarly to what we already did in other components, I would like to propose to get the full list of tenants upfront in the querier (so that the response will be cached) and then sync each tenant blocks.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/querier/blocks_scanner.go
+++ b/pkg/querier/blocks_scanner.go
@@ -47,6 +47,7 @@ type BlocksScanner struct {
 	logger          log.Logger
 	bucketClient    objstore.Bucket
 	fetchersMetrics *storegateway.MetadataFetcherMetrics
+	usersScanner    *cortex_tsdb.UsersScanner
 
 	// We reuse the metadata fetcher instance for a given tenant both because of performance
 	// reasons (the fetcher keeps a in-memory cache) and being able to collect and group metrics.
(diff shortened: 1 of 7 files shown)


pkg/querier/blocks_scanner.go
```diff
@@ -210,21 +218,16 @@ func (d *BlocksScanner) scanBucket(ctx context.Context) (returnErr error) {
...
-		userID := strings.TrimSuffix(s, "/")
+	// Push a job for each user whose blocks need to be discovered.
+	for _, userID := range userIDs {
 		select {
 		case jobsChan <- userID:
-			return nil
+		// Nothing to do.
 		case <-ctx.Done():
-			return ctx.Err()
+			resMx.Lock()
```
pstibrany, line 227: This will add many context errors to `resErrs` (one per remaining user ID). Previously only one was added, which looks like enough.

pstibrany (approved):
LGTM

pkg/querier/blocks_scanner.go
pracucci: So glad you spotted it, thanks! Should be fixed now.

pracucci:
Thanks @pstibrany for your valuable review! I addressed your comment. Could you look at it again, please?

pkg/querier/blocks_scanner_test.go
```diff
@@ -162,7 +162,7 @@ func TestBlocksScanner_StopWhileRunningTheInitialScanOnManyTenants(t *testing.T)
 	time.Sleep(time.Second)
 
 	stopTime := time.Now()
-	require.NoError(t, services.StopAndAwaitTerminated(context.Background(), s))
+	_ = services.StopAndAwaitTerminated(context.Background(), s)
```
pracucci, line 165: I've noticed this and the following test are flaky because in this case `StopAndAwaitTerminated()` may either return `nil` or `context.Canceled`.

pstibrany (approved):

PR #3528, opened 2020-11-23
Added support to cache meta.json attributes

pracucci:
**What this PR does**:
I just realised we issue `Attributes()` calls on `meta.json` but we don't cache it. This PR adds support for caching `meta.json` attributes API call too, other than increasing the default attributes TTL from 1d to 7d.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany (approved):
LGTM.

We don't need to use this for NoCompactMarkFilename, since that's only used by compactor, which doesn't use metadata cache at all.

PR #3529, opened 2020-11-23
add api to list all alertmanager configs and rule groups

### LINKED ISSUE #3530: support list all users alertmanager configs and rule groups

Aa a administrator, I want to see all alertmangaer configs and rule groups

Add a api or web to list all configs and rules

AllenzhLi:
**What this PR does**:

add api to list all user alertmangaer configs

Fixes #3530 

**Checklist**
- [x] Tests updated
- [x] Documentation added


pracucci:
Thanks @AllenzhLi for your PR. @gotjosh could you do the first pass review, please?

gotjosh:
Hi @AllenzhLi, apologies for the delay on this and thank you for your contribution - I have it on agenda to review this soon. Before I do, could you please add at least an unit test to this?

pracucci (commented):
Thanks for your contribution! A couple of quick things I would add:

- A CHANGELOG entry
- Documentation of new endpoints to `docs/api/_index.md`
- Unit tests

AllenzhLi:
> Thanks for your contribution! A couple of quick things I would add:
> 
> * A CHANGELOG entry
> * Documentation of new endpoints to `docs/api/_index.md`
> * Unit tests

The above is done.

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@
 * [ENHANCEMENT] Memberlist: client can now keep a size-bounded buffer with sent and received messages and display them in the admin UI (/memberlist) for troubleshooting. #3581 #3602
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Query-Frontend: `cortex_query_seconds_total` now return seconds not nanoseconds. #3589
+* [ENHANCEMENT] Add api to list all tenant alertmanager configs and ruler rules. #3259
```
pracucci, line 23: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

AllenzhLi: Updated.

CHANGELOG.md
```diff
@@ -80,6 +80,9 @@
 * [BUGFIX] Compactor: fixed "could not guess file size" log when uploading blocks deletion marks to the global location. #3807
 * [BUGFIX] Prevent panic at start if the http_prefix setting doesn't have a valid value. #3796
 * [BUGFIX] Memberlist: fixed panic caused by race condition in `armon/go-metrics` used by memberlist client. #3724
+* [ENHANCEMENT] Add api to list all tenant alertmanager configs and ruler rules. #3259
```
pstibrany, line 83: Cortex release 1.8.0 is now in progress. Could you please rebase master and move the CHANGELOG entry under the master / unreleased section?

AllenzhLi: Updated.

docs/api/_index.md
```diff
@@ -663,13 +673,23 @@ GET /status
 
 Displays a web page with the current status of the Alertmanager, including the Alertmanager cluster members.
 
+<<<<<<< HEAD
```
pstibrany, line 676: Please remove merge conflict marks.

docs/api/_index.md
```diff
@@ -663,13 +673,23 @@ GET /status
...
+List all Alertmanager configurations that are currently loaded. This endpoint returns a YAML dictionary with all the Alertmanager configurations and `200` status code on success.
+=======
 ### Alertmanager ring status
 
 ```
 GET /multitenant_alertmanager/ring
 ```
 
 Displays a web page with the Alertmanager hash ring status, including the state, healthy and last heartbeat time of each Alertmanager instance.
+>>>>>>> eb02c0dc3bd9c4010baa9c77b84647111b1a013b
```
pstibrany, line 692: Please remove merge conflict marks.

docs/api/_index.md
AllenzhLi: done.

docs/api/_index.md
AllenzhLi: done.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 * [ENHANCEMENT] Ruler: added the following metrics when ruler sharding is enabled: #3916
   * `cortex_ruler_clients`
   * `cortex_ruler_client_request_duration_seconds`
+* [ENHANCEMENT] Add api to list all tenant alertmanager configs and ruler rules. #3259
+   - `GET /multitenant_alertmanager/configs`
+   - `GET /ruler/rules`
```
pstibrany, line 11:
```suggestion
* [ENHANCEMENT] Alertmanager: Add API endpoint to list all tenant alertmanager configs: `GET /multitenant_alertmanager/configs`. #3259
* [ENHANCEMENT] Ruler: Add API endpoint to list all tenant ruler rule groups: `GET /ruler/rules`. #3259
```

CHANGELOG.md
```diff
@@ -107,7 +110,7 @@
 * [BUGFIX] Prevent panic at start if the http_prefix setting doesn't have a valid value. #3796
 * [BUGFIX] Memberlist: fixed panic caused by race condition in `armon/go-metrics` used by memberlist client. #3724
 * [BUGFIX] Querier: returning 422 (instead of 500) when query hits `max_chunks_per_query` limit with block storage. #3895
-* [BUGFIX] Alertmanager: Ensure that experimental `/api/v1/alerts` endpoints work when `-http.prefix` is empty. #3905
+* [BUGFIX] Alertmanager: Ensure that experimental `/api/v1/alerts` endpoints work when `-http.prefix` is empty. #3905*
```
pstibrany, line 113: This looks like unintended change.

docs/api/_index.md
```diff
@@ -431,6 +433,14 @@ GET /ruler_ring
 
 Displays a web page with the ruler hash ring status, including the state, healthy and last heartbeat time of each ruler.
 
+### Ruler rules
+
+```
+GET /ruler/rules
```
pstibrany, line 439: We're listing rule groups, should we call the endpoint that way? Ie. `/ruler/rule_groups`.

More importantnly, let's mention that this endpoint is not part of ruler-API which is enabled by the flag, but is always available. It should not be exposed to end users.

docs/api/_index.md
```diff
@@ -431,6 +433,14 @@ GET /ruler_ring
...
+```
+
+List all tenant rules that are currently loaded. This endpoint returns a YAML dictionary with all the rule groups for each tenant and `200` status code on success.
```
pstibrany, line 442: > List all tenant rules that are currently loaded. 

I think this is misleading. Each ruler will have different set of rule groups loaded in memory, but this endpoint will always list them all.

docs/api/_index.md
```diff
@@ -663,6 +673,13 @@ GET /status
 
 Displays a web page with the current status of the Alertmanager, including the Alertmanager cluster members.
 
+### Alertmanager configs
+
+```
+GET /multitenant_alertmanager/configs
+```
+
+List all Alertmanager configurations that are currently loaded. This endpoint returns a YAML dictionary with all the Alertmanager configurations and `200` status code on success.
```
pstibrany, line 682: Same comment about "loaded"... all AM configs are returned. Once AM sharding is finished, AM will only have subset of configs loaded in memory, but endpoint will still return all configs.

I'd also add mention that this endpoint is available regardless of whether alertmanager-API is enabled or not.

pkg/alertmanager/api.go
```diff
@@ -184,3 +185,42 @@ func validateUserConfig(logger log.Logger, cfg alertspb.AlertConfigDesc) error {
 
 	return nil
 }
+
+func (am *MultitenantAlertmanager) ListUserConfig(w http.ResponseWriter, r *http.Request) {
+	logger := util_log.WithContext(r.Context(), am.logger)
+	userIDs, err := am.store.ListAllUsers(r.Context())
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to list users of alertmanager")
```
pstibrany, line 193:
```suggestion
		level.Error(logger).Log("msg", "failed to list users of alertmanager", "err", err)
```

pkg/alertmanager/api.go
```diff
@@ -184,3 +185,42 @@ func validateUserConfig(logger log.Logger, cfg alertspb.AlertConfigDesc) error {
...
+		http.Error(w, fmt.Sprintf("%s: %s", errListAllUser, err.Error()), http.StatusInternalServerError)
+		return
+	}
+	userConfigMap := make(map[string]*UserConfig, len(userIDs))
+	for _, userID := range userIDs {
+		cfg, err := am.store.GetAlertConfig(r.Context(), userID)
```
pstibrany, line 199: New `GetAlertConfigs` could be used to efficiently load configs for all users at once (or perhaps smaller batches). 

I'm bit afraid what happens when someone calls this on large alertmanager installation with thousands of users -- this will be slow and can take a lot of memory, possibly crashing AM. It would be best to stream output instead of building it in memory first. WDYT?

pkg/alertmanager/api.go
```diff
@@ -184,3 +185,42 @@ func validateUserConfig(logger log.Logger, cfg alertspb.AlertConfigDesc) error {
...
+		if err != nil {
+			if err == alertspb.ErrNotFound {
+				http.Error(w, err.Error(), http.StatusNotFound)
```
pstibrany, line 202: If config for single user is not found (perhaps it was just deleted), I don't think we should stop early... instead we should just ignore it, and return the rest. Note that `GetAlertConfigs` handles this.

pkg/alertmanager/api.go
```diff
@@ -184,3 +185,42 @@ func validateUserConfig(logger log.Logger, cfg alertspb.AlertConfigDesc) error {
...
+			} else {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
+			return
+		}
+		userConfigMap[userID] = &UserConfig{
+			TemplateFiles:      alertspb.ParseTemplates(cfg),
+			AlertmanagerConfig: cfg.RawConfig,
```
pstibrany, line 210: Here I'd suggest writing partial output to response, and not keeping it in memory.

pkg/alertmanager/api.go
```diff
@@ -184,3 +185,42 @@ func validateUserConfig(logger log.Logger, cfg alertspb.AlertConfigDesc) error {
...
+		level.Error(logger).Log("msg", errMarshallingYAML, "err", err)
+		http.Error(w, fmt.Sprintf("%s: %s", errMarshallingYAML, err.Error()), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/yaml")
+	if _, err := w.Write(d); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
pstibrany, line 225: We can use `util.WriteYAMLResponse()`

pkg/ruler/api.go
```diff
@@ -551,3 +551,28 @@ func (a *API) DeleteRuleGroup(w http.ResponseWriter, req *http.Request) {
 
 	respondAccepted(w, logger)
 }
+
+func (a *API) ListAllRules(w http.ResponseWriter, req *http.Request) {
```
pstibrany, line 555: Similar to listing alert groups, I think we should
1) get all users
2) get rule groups for batch of users
3) send this partial output, and then go to step 2 and repeat

Otherwise loading all rule groups into memory first may take a while, and its memory consumption is not bounded in any way. WDYT?

pstibrany (commented):
Thank you for your work on this. I agree that these two endpoints are useful addition, but I have some comments on implementation.

pkg/alertmanager/api.go
AllenzhLi: Update to use GetAlertConfigs.

pkg/alertmanager/api.go
AllenzhLi: Yes, util.WriteYAMLResponse() seems great.

pkg/alertmanager/api.go
AllenzhLi: partial output？
I don't quite understand this, can you elaborate more?

pkg/ruler/api.go
AllenzhLi: Yes, I agree with you. 

But the ruler store have a ListAllRuleGroups interface which returns all rule groups for all users.

pkg/alertmanager/api.go
AllenzhLi: Do you meaning output user configurations in batches, such as outputting 10 user configurations at a time?

pkg/ruler/api.go
AllenzhLi: Ruler sstore have no interface like GetAlertConfigs which allow get configs by input batch users.

Is there any plan to support such a interface? I can work for that, but it may lasts a long time.

pkg/alertmanager/api.go
pstibrany: What I meant is marshalling and writing individual values to the response, which are then sent to client immediately, instead of building the entire `userConfigMap` in memory first. It would need to be done in such a way that what client see is single valid YAML document. I'm not sure how feasible that is.

In other words, instead of storing `userConfigMap[userID] = &UserConfig{...}`, we could marshal `&UserConfig{...}` into YAML and write it to response, possibly with some newlines. It's just an idea, and could be explored.

What worries me a bit is that we have some clusters with 1K+ tenants in Alertmanager (and it could easily be 10K+ in a year). Building response in memory for all those tenants could cause OOM.

pkg/ruler/api.go
pstibrany: > But the ruler store have a ListAllRuleGroups interface which returns all rule groups for all users.

I don't think we should use it -- it suffers the same issue as described above -- loading all configs into memory could cause OOM. Ruler also doesn't do that when sharding is enabled.

I would not advocate for using `GetAlertConfigs` for *ALL* users at once. That is why I suggested to use "batches" of users instead.

pkg/alertmanager/api.go
AllenzhLi: Get it, I will try.

pkg/ruler/api.go
AllenzhLi: Get it.

pkg/alertmanager/api.go
AllenzhLi: I have update to use partial output. It seems ok but only except that every output will add a '---'.

pkg/ruler/api.go
AllenzhLi: Also updated.

Now the api use iterator mode.

There is a problem:

```
// for k, v := range tt.value {
// 	iter.Put(map[string]*testStruct{k: v})
// }
```

TestStreamWriteYAMLResponse use iter not work, but TestAMConfigListUserConfig and TestRuler_ListAllRules work well, suspected that httptest flush is not a real flush.

AllenzhLi: This problem have resolved and the reson is '---'.

Yaml unmarshal will stop when find '---'.

pkg/alertmanager/api.go
AllenzhLi: update from yaml.NewEncoder to yaml.Marshal and everything seems good.

I think I have achieved: marshal single &UserConfig{...} into YAML and write it to response.

pkg/ruler/api.go
AllenzhLi: Similar to listing alert groups. The `partial output` seems good now.

AllenzhLi:
Friendly ping (:

pstibrany (changes requested):
Thank you for working on it, and apologies for super-slow review cycle.

I've tried to test this code (built Cortex, and started Alertmanager), but unfortunately it doesn't work. Reason is that `StreamWriteResponse` expects that response writer implements `http.Flusher`, but Cortex uses response writer internally, which doesn't do that. In that case `StreamWriteResponse` returns error, and never drains the channel (`iter`), which causes goroutines trying to call `Put` (send to channel) block forever.

Please do the following changes:
- remove `RespIterator`
- change `StreamWriteResponse` to not *require* `http.Flusher`. If response implements `http.Flusher`, then this method can call `Flush` method, but it should not *require* it.
- change functions passed to `concurrency.ForEachUser` in both `ListAllConfigs` and `ListAllRules` as follows: Instead of calling `iter.Put(data)`, use the following idiom (assuming `iter` is now channel):

```
select {
case iter <- data:
case <-done:
}
```

This allows both `ListAllConfigs` and `ListAllRules` to stop early, if sending response has already finished.

- Instead of using `done <- struct{}{}` to signal that response has been sent, please use `close(done)` instead. The difference is that closed channel can be read repeatedly without blocking (it returns default value, and optional bool to indicate if read succeeded or channel was closed), while if we send empty struct to channel, this can be only read once -- perhaps by the select as shown above, but then additional read `<-done` at the end of `ListAllConfigs` / `ListAllRules` would never end.

pkg/util/http.go
```diff
@@ -98,6 +98,54 @@ func RenderHTTPResponse(w http.ResponseWriter, v interface{}, t *template.Templa
 	}
 }
 
+// RespIterator is a iterator for stream http response
+type RespIterator interface {
```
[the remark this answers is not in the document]
AllenzhLi: I have remove RespIterator and update to use select idiom.

AllenzhLi:
> * change `StreamWriteResponse` to not _require_ `http.Flusher`. If response implements `http.Flusher`, then this method can call `Flush` method, but it should not _require_ it.

Only response implements `http.Flusher`, then we can get rule groups for batch of users and send(w.Write and http.Flusher) this partial output, and then repeat.

If no `http.Flusher` method, test found that the http.ResponseWriter only accept the first `w.Write(partial_content)`, so we must cache all content and do `w.Write(all_content)`.

Do you have any other suggestions?

pstibrany:
> If no `http.Flusher` method, test found that the http.ResponseWriter only accept the first `w.Write(partial_content)`, so we must cache all content and do `w.Write(all_content)`.

This sounds very suspicious. It’s perfectly valid to call `response.Write` multiple times, and all those writes will be sent to HTTP client. I would expect that simply removing `Flush` calls should not affect the test, if it’s written correctly (which I haven’t checked).

AllenzhLi:
> > If no `http.Flusher` method, test found that the http.ResponseWriter only accept the first `w.Write(partial_content)`, so we must cache all content and do `w.Write(all_content)`.
> 
> This sounds very suspicious. It’s perfectly valid to call `response.Write` multiple times, and all those writes will be sent to HTTP client. I would expect that simply removing `Flush` calls should not affect the test, if it’s written correctly (which I haven’t checked).

You are right. I have make a misstake.

The new commit have implement stream write yaml to `http.ResponseWriter`.

I also update `Content-type` to `application/yaml` which will become a file to download in chrome. This action is keep same with the `GetUserConfig` API.

I have make a three tenants alertmanger test, and use chrome access the `/multitenant_alertmanager/configs`, then I get a file download and the content of the file as follow:

```
u3:
  template_files: {}
  alertmanager_config: |
    global:
      # The smarthost and SMTP sender used for mail notifications.
      smtp_from: 'alertmanager@example.org'
      smtp_auth_username: 'alertmanager'
      smtp_auth_password: 'password'

    # The directory from which notification templates are read.
    templates:
    - '/etc/alertmanager/template/*.tmpl'

    # The root route on which each incoming alert enters.
    route:
      # The labels by which incoming alerts are grouped together. For example,
      # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
      # be batched into a single group.
      #
      # To aggregate by all possible labels use '...' as the sole label name.
      # This effectively disables aggregation entirely, passing through all
      # alerts as-is. This is unlikely to be what you want, unless you have
      # a very low alert volume or your upstream notification system performs
      # its own grouping. Example: group_by: [...]
      group_by: ['alertname', 'cluster', 'service']

      # When a new group of alerts is created by an incoming alert, wait at
      # least 'group_wait' to send the initial notification.
      # This way ensures that you get multiple alerts for the same group that start
      # firing shortly after another are batched together on the first
      # notification.
      group_wait: 30s

      # When the first notification was sent, wait 'group_interval' to send a batch
      # of new alerts that started firing for that group.
      group_interval: 5m

      # If an alert has successfully been sent, wait 'repeat_interval' to
      # resend them.
      repeat_interval: 3h

      # A default receiver
      receiver: team-X-mails

      # All the above attributes are inherited by all child routes and can
      # overwritten on each.

      # The child route trees.
      routes:
      # This routes performs a regular expression match on alert labels to
      # catch alerts that are related to a list of services.
      - match_re:
          service: ^(foo1|foo2|baz)$
        receiver: team-X-mails
        # The service has a sub-route for critical alerts, any alerts
        # that do not match, i.e. severity != critical, fall-back to the
        # parent node and are sent to 'team-X-mails'
        routes:
        - match:
            severity: critical
          receiver: team-X-pager
      - match:
          service: files
        receiver: team-Y-mails

        routes:
        - match:
            severity: critical
          receiver: team-Y-pager

      # This route handles all alerts coming from a database service. If there's
      # no team to handle it, it defaults to the DB team.
      - match:
          service: database
        receiver: team-DB-pager
        # Also group alerts by affected database.
        group_by: [alertname, cluster, database]
        routes:
        - match:
            owner: team-X
          receiver: team-X-pager
          continue: true
        - match:
            owner: team-Y
          receiver: team-Y-pager


    # Inhibition rules allow to mute a set of alerts given that another alert is
    # firing.
    # We use this to mute any warning-level notifications if the same alert is
    # already critical.
    inhibit_rules:
    - source_match:
        severity: 'critical'
      target_match:
        severity: 'warning'
      # Apply inhibition if the alertname is the same.
      # CAUTION:
      #   If all label names listed in `equal` are missing
      #   from both the source and target alerts,
      #   the inhibition rule will apply!
      equal: ['alertname', 'cluster', 'service']


    receivers:
    - name: 'team-X-mails'
      email_configs:
      - to: 'team-X+alerts@example.org'

    - name: 'team-X-pager'
      email_configs:
      - to: 'team-X+alerts-critical@example.org'
      pagerduty_configs:
      - service_key: <team-X-key>

    - name: 'team-Y-mails'
      email_configs:
      - to: 'team-Y+alerts@example.org'

    - name: 'team-Y-pager'
      pagerduty_configs:
      - service_key: <team-Y-key>

    - name: 'team-DB-pager'
      pagerduty_configs:
      - service_key: <team-DB-key>
u1:
  template_files: {}
  alertmanager_config: |
    global:
      # The smarthost and SMTP sender used for mail notifications.
      smtp_from: 'alertmanager@example.org'
      smtp_auth_username: 'alertmanager'
      smtp_auth_password: 'password'

    # The directory from which notification templates are read.
    templates:
    - '/etc/alertmanager/template/*.tmpl'

    # The root route on which each incoming alert enters.
    route:
      # The labels by which incoming alerts are grouped together. For example,
      # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
      # be batched into a single group.
      #
      # To aggregate by all possible labels use '...' as the sole label name.
      # This effectively disables aggregation entirely, passing through all
      # alerts as-is. This is unlikely to be what you want, unless you have
      # a very low alert volume or your upstream notification system performs
      # its own grouping. Example: group_by: [...]
      group_by: ['alertname', 'cluster', 'service']

      # When a new group of alerts is created by an incoming alert, wait at
      # least 'group_wait' to send the initial notification.
      # This way ensures that you get multiple alerts for the same group that start
      # firing shortly after another are batched together on the first
      # notification.
      group_wait: 30s

      # When the first notification was sent, wait 'group_interval' to send a batch
      # of new alerts that started firing for that group.
      group_interval: 5m

      # If an alert has successfully been sent, wait 'repeat_interval' to
      # resend them.
      repeat_interval: 3h

      # A default receiver
      receiver: team-X-mails

      # All the above attributes are inherited by all child routes and can
      # overwritten on each.

      # The child route trees.
      routes:
      # This routes performs a regular expression match on alert labels to
      # catch alerts that are related to a list of services.
      - match_re:
          service: ^(foo1|foo2|baz)$
        receiver: team-X-mails
        # The service has a sub-route for critical alerts, any alerts
        # that do not match, i.e. severity != critical, fall-back to the
        # parent node and are sent to 'team-X-mails'
        routes:
        - match:
            severity: critical
          receiver: team-X-pager
      - match:
          service: files
        receiver: team-Y-mails

        routes:
        - match:
            severity: critical
          receiver: team-Y-pager

      # This route handles all alerts coming from a database service. If there's
      # no team to handle it, it defaults to the DB team.
      - match:
          service: database
        receiver: team-DB-pager
        # Also group alerts by affected database.
        group_by: [alertname, cluster, database]
        routes:
        - match:
            owner: team-X
          receiver: team-X-pager
          continue: true
        - match:
            owner: team-Y
          receiver: team-Y-pager


    # Inhibition rules allow to mute a set of alerts given that another alert is
    # firing.
    # We use this to mute any warning-level notifications if the same alert is
    # already critical.
    inhibit_rules:
    - source_match:
        severity: 'critical'
      target_match:
        severity: 'warning'
      # Apply inhibition if the alertname is the same.
      # CAUTION:
      #   If all label names listed in `equal` are missing
      #   from both the source and target alerts,
      #   the inhibition rule will apply!
      equal: ['alertname', 'cluster', 'service']


    receivers:
    - name: 'team-X-mails'
      email_configs:
      - to: 'team-X+alerts@example.org'

    - name: 'team-X-pager'
      email_configs:
      - to: 'team-X+alerts-critical@example.org'
      pagerduty_configs:
      - service_key: <team-X-key>

    - name: 'team-Y-mails'
      email_configs:
      - to: 'team-Y+alerts@example.org'

    - name: 'team-Y-pager'
      pagerduty_configs:
      - service_key: <team-Y-key>

    - name: 'team-DB-pager'
      pagerduty_configs:
      - service_key: <team-DB-key>
u2:
  template_files: {}
  alertmanager_config: |
    global:
      # The smarthost and SMTP sender used for mail notifications.
      smtp_from: 'alertmanager@example.org'
      smtp_auth_username: 'alertmanager'
      smtp_auth_password: 'password'

    # The directory from which notification templates are read.
    templates:
    - '/etc/alertmanager/template/*.tmpl'

    # The root route on which each incoming alert enters.
    route:
      # The labels by which incoming alerts are grouped together. For example,
      # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
      # be batched into a single group.
      #
      # To aggregate by all possible labels use '...' as the sole label name.
      # This effectively disables aggregation entirely, passing through all
      # alerts as-is. This is unlikely to be what you want, unless you have
      # a very low alert volume or your upstream notification system performs
      # its own grouping. Example: group_by: [...]
      group_by: ['alertname', 'cluster', 'service']

      # When a new group of alerts is created by an incoming alert, wait at
      # least 'group_wait' to send the initial notification.
      # This way ensures that you get multiple alerts for the same group that start
      # firing shortly after another are batched together on the first
      # notification.
      group_wait: 30s

      # When the first notification was sent, wait 'group_interval' to send a batch
      # of new alerts that started firing for that group.
      group_interval: 5m

      # If an alert has successfully been sent, wait 'repeat_interval' to
      # resend them.
      repeat_interval: 3h

      # A default receiver
      receiver: team-X-mails

      # All the above attributes are inherited by all child routes and can
      # overwritten on each.

      # The child route trees.
      routes:
      # This routes performs a regular expression match on alert labels to
      # catch alerts that are related to a list of services.
      - match_re:
          service: ^(foo1|foo2|baz)$
        receiver: team-X-mails
        # The service has a sub-route for critical alerts, any alerts
        # that do not match, i.e. severity != critical, fall-back to the
        # parent node and are sent to 'team-X-mails'
        routes:
        - match:
            severity: critical
          receiver: team-X-pager
      - match:
          service: files
        receiver: team-Y-mails

        routes:
        - match:
            severity: critical
          receiver: team-Y-pager

      # This route handles all alerts coming from a database service. If there's
      # no team to handle it, it defaults to the DB team.
      - match:
          service: database
        receiver: team-DB-pager
        # Also group alerts by affected database.
        group_by: [alertname, cluster, database]
        routes:
        - match:
            owner: team-X
          receiver: team-X-pager
          continue: true
        - match:
            owner: team-Y
          receiver: team-Y-pager


    # Inhibition rules allow to mute a set of alerts given that another alert is
    # firing.
    # We use this to mute any warning-level notifications if the same alert is
    # already critical.
    inhibit_rules:
    - source_match:
        severity: 'critical'
      target_match:
        severity: 'warning'
      # Apply inhibition if the alertname is the same.
      # CAUTION:
      #   If all label names listed in `equal` are missing
      #   from both the source and target alerts,
      #   the inhibition rule will apply!
      equal: ['alertname', 'cluster', 'service']


    receivers:
    - name: 'team-X-mails'
      email_configs:
      - to: 'team-X+alerts@example.org'

    - name: 'team-X-pager'
      email_configs:
      - to: 'team-X+alerts-critical@example.org'
      pagerduty_configs:
      - service_key: <team-X-key>

    - name: 'team-Y-mails'
      email_configs:
      - to: 'team-Y+alerts@example.org'

    - name: 'team-Y-pager'
      pagerduty_configs:
      - service_key: <team-Y-key>

    - name: 'team-DB-pager'
      pagerduty_configs:
      - service_key: <team-DB-key>

```

pkg/util/http.go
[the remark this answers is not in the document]
AllenzhLi: The select idion realy great. Thank you very match.

AllenzhLi:
Just now, I have test one hundred tenants alertmanager instance, work well.

When I increase the number of instance to two hundred, get error `ctx canceled`.

AllenzhLi:
When I update the `http_server_write_timeout` from default 30s to 5m all is work well and I can increase alertmanager tenants to 1000, that spend 4m20s to return all configs.

pkg/alertmanager/api_test.go
```diff
@@ -199,3 +207,91 @@ func TestMultitenantAlertmanager_DeleteUserConfig(t *testing.T) {
...
+			RawConfig: cfg.AlertmanagerConfig,
+		})
+		require.NoError(t, err)
+	}
+
+	externalURL := flagext.URLValue{}
+	err := externalURL.Set("http://localhost:8080/alertmanager")
+	require.NoError(t, err)
+
+	tempDir, err := ioutil.TempDir(os.TempDir(), "alertmanager")
```
pstibrany, line 266: Nit: We can use new `t.TempDir()` which creates unique dir per top-level test, and is automatically removed at the end of test.

pkg/alertmanager/api_test.go
```diff
@@ -199,3 +207,91 @@ func TestMultitenantAlertmanager_DeleteUserConfig(t *testing.T) {
...
+	req := httptest.NewRequest("GET", "https://localhost:8080/multitenant_alertmanager/configs", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+
+	resp := w.Result()
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	require.Equal(t, "application/yaml", resp.Header.Get("Content-Type"))
+	body, _ := ioutil.ReadAll(resp.Body)
+	old, _ := yaml.Marshal(testCases)
+	require.YAMLEq(t, string(old), string(body))
```
pstibrany, line 293: Using YAMLEq is smart!

pkg/alertmanager/api_test.go
```diff
@@ -199,3 +207,91 @@ func TestMultitenantAlertmanager_DeleteUserConfig(t *testing.T) {
...
+
+	// It succeeds and the Alertmanager is started
+	require.Len(t, am.alertmanagers, 2)
```
pstibrany, line 296: Nit: Shouldn't we test this after `loadAndSyncConfigs`?

pstibrany (commented):
Thank you for addressing my feedback. I have few more comments, but this PR is almost ready now.

pkg/util/http.go
```diff
@@ -98,6 +98,19 @@ func RenderHTTPResponse(w http.ResponseWriter, v interface{}, t *template.Templa
 	}
 }
 
+// StreamWriteYAMLResponse stream writes data as http response
+func StreamWriteYAMLResponse(w http.ResponseWriter, iter chan interface{}) {
+	w.Header().Set("Content-Type", "application/yaml")
+	for v := range iter {
+		data, err := yaml.Marshal(v)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
AllenzhLi: You are right. Get it.

pkg/util/http.go
```diff
@@ -98,6 +98,19 @@ func RenderHTTPResponse(w http.ResponseWriter, v interface{}, t *template.Templa
...
+			return
+		}
+		_, _ = w.Write(data)
```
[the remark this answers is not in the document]
AllenzhLi: I agree with you. I will add a log here.

pkg/ruler/ruler.go
```diff
@@ -812,3 +817,44 @@ func (r *Ruler) DeleteTenantConfiguration(w http.ResponseWriter, req *http.Reque
...
+
+		select {
+		case iter <- data:
+		case <-done: // stop early, if sending response has already finished
+		}
+
+		return nil
+	})
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to list all alertmanager configs", "err", err)
```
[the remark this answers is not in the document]
AllenzhLi: updated.

pkg/alertmanager/api_test.go
AllenzhLi: I agree with you. This test should after loadAndSyncConfigs.

pkg/alertmanager/api_test.go
AllenzhLi: `t.TempDir()` is good. I learned.

The `tempDir` seems to be useless here, I will delete it.

pstibrany:
Latest changes added two extra files:
- .vscode/launch.json
- docs/configuration/single-process-config-alertmanager.yaml

pkg/util/http.go
```diff
@@ -98,6 +100,23 @@ func RenderHTTPResponse(w http.ResponseWriter, v interface{}, t *template.Templa
 	}
 }
 
+// StreamWriteYAMLResponse stream writes data as http response
+func StreamWriteYAMLResponse(w http.ResponseWriter, iter chan interface{}, logger log.Logger) {
+	w.Header().Set("Content-Type", "application/yaml")
+	for v := range iter {
+		data, err := yaml.Marshal(v)
+		if err != nil {
+			level.Error(logger).Log("msg", "yaml marshal filed", "err", err)
```
pstibrany, line 109: typo: filed => failed

pkg/util/http.go
```diff
@@ -98,6 +100,23 @@ func RenderHTTPResponse(w http.ResponseWriter, v interface{}, t *template.Templa
...
+			continue
+		}
+		_, err = w.Write(data)
+		if err != nil {
+			level.Error(logger).Log("msg", "write http response filed", "err", err)
```
pstibrany, line 114: same typo

pstibrany (approved):
I think this is in a good state now (after fixing extra files and typos).

Thank you very much for working on this!

pkg/util/http.go
AllenzhLi: updated

pkg/util/http.go
AllenzhLi: updated

AllenzhLi:
Aleady fixing extra files and typos.

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
 * [ENHANCEMENT] Ruler: added the following metrics when ruler sharding is enabled: #3916
   * `cortex_ruler_clients`
   * `cortex_ruler_client_request_duration_seconds`
+* [ENHANCEMENT] Alertmanager: Add API endpoint to list all tenant alertmanager configs: `GET /multitenant_alertmanager/configs`. #3259
```
pracucci, line 30:
```suggestion
* [ENHANCEMENT] Alertmanager: Add API endpoint to list all tenant alertmanager configs: `GET /multitenant_alertmanager/configs`. #3529
```

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
...
+* [ENHANCEMENT] Ruler: Add API endpoint to list all tenant ruler rule groups: `GET /ruler/rule_groups`. #3259
```
pracucci, line 31:
```suggestion
* [ENHANCEMENT] Ruler: Add API endpoint to list all tenant ruler rule groups: `GET /ruler/rule_groups`. #3529
```

pkg/ruler/ruler.go
```diff
@@ -840,3 +845,44 @@ func (r *Ruler) DeleteTenantConfiguration(w http.ResponseWriter, req *http.Reque
...
+	iter := make(chan interface{})
+
+	go func() {
+		util.StreamWriteYAMLResponse(w, iter, logger)
+		close(done)
+	}()
+
+	err = concurrency.ForEachUser(req.Context(), userIDs, fetchRulesConcurrency, func(ctx context.Context, userID string) error {
+		rg, err := r.store.ListRuleGroupsForUserAndNamespace(ctx, userID, "")
+		if errors.Is(err, chunk.ErrStorageObjectNotFound) {
```
pracucci, line 869: Checking for `chunk.ErrStorageObjectNotFound` is not correct because this error is specific to a single backend implementation. However, the "object not found" error cannot be returned by `ListRuleGroupsForUserAndNamespace()` so should be safe to remove this if condition.

pkg/ruler/ruler_test.go
```diff
@@ -843,6 +845,38 @@ type ruleGroupKey struct {
...
+	defer cleanup()
+
+	r, rcleanup := newTestRuler(t, cfg)
+	defer rcleanup()
+	defer services.StopAndAwaitTerminated(context.Background(), r) //nolint:errcheck
+
+	router := mux.NewRouter()
+	router.Path("/ruler/rule_groups").Methods(http.MethodGet).HandlerFunc(r.ListAllRules)
+
+	// Verify namespace1 rules are there.
```
pracucci, line 859: This comment looks incorrect.

pkg/ruler/ruler_test.go
```diff
@@ -843,6 +845,38 @@ type ruleGroupKey struct {
...
+	router.ServeHTTP(w, req)
+
+	resp := w.Result()
+	body, _ := ioutil.ReadAll(resp.Body)
+
+	// Check status code and header
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	require.Equal(t, "application/yaml", resp.Header.Get("Content-Type"))
+
+	// Testing the running rules for user1 in the mock store
```
pracucci, line 871: This comment looks incorrect.

pkg/ruler/ruler_test.go
```diff
@@ -843,6 +845,38 @@ type ruleGroupKey struct {
...
+	gs := make(map[string]map[string][]rulefmt.RuleGroup) // user:namespace:[]rulefmt.RuleGroup
+	for userID := range mockRules {
+		gs[userID] = mockRules[userID].Formatted()
+	}
+	expectedResponse, _ := yaml.Marshal(gs)
```
pracucci, line 876: As a good practice I would suggest to always check errors (in tests too).

```suggestion
	expectedResponse, err := yaml.Marshal(gs)
	require.NoError(t, err)
```

pkg/util/http_test.go
```diff
@@ -84,6 +88,52 @@ func TestWriteTextResponse(t *testing.T) {
...
+		value:               make(map[string]*testStruct),
+	}
+	for i := 0; i < rand.Intn(100); i++ {
+		ts := testStruct{
+			Name:  "testName" + strconv.Itoa(i),
+			Value: i,
+		}
+		tt.value[ts.Name] = &ts
+	}
+	d, _ := yaml.Marshal(tt.value)
```
pracucci, line 117:
```suggestion
	d, err := yaml.Marshal(tt.value)
	require.NoError(t, err)
```

pkg/alertmanager/api.go
```diff
@@ -184,3 +193,48 @@ func validateUserConfig(logger log.Logger, cfg alertspb.AlertConfigDesc) error {
...
+	iter := make(chan interface{})
+
+	go func() {
+		util.StreamWriteYAMLResponse(w, iter, logger)
+		close(done)
+	}()
+
+	err = concurrency.ForEachUser(r.Context(), userIDs, fetchConcurrency, func(ctx context.Context, userID string) error {
+		cfg, err := am.store.GetAlertConfig(ctx, userID)
+		if errors.Is(err, chunk.ErrStorageObjectNotFound) {
```
pracucci, line 216: This is not correct. You should check for `alertspb.ErrNotFound`.

```suggestion
		if errors.Is(err, alertspb.ErrNotFound) {
```

pkg/alertmanager/api_test.go
```diff
@@ -199,3 +206,84 @@ func TestMultitenantAlertmanager_DeleteUserConfig(t *testing.T) {
...
+	router := mux.NewRouter()
+	router.Path("/multitenant_alertmanager/configs").Methods(http.MethodGet).HandlerFunc(am.ListAllConfigs)
+	req := httptest.NewRequest("GET", "https://localhost:8080/multitenant_alertmanager/configs", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+
+	resp := w.Result()
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	require.Equal(t, "application/yaml", resp.Header.Get("Content-Type"))
+	body, _ := ioutil.ReadAll(resp.Body)
```
pracucci, line 286:
```suggestion
	body, err := ioutil.ReadAll(resp.Body)
	require.NoError(t, err)
```

pkg/alertmanager/api_test.go
```diff
@@ -199,3 +206,84 @@ func TestMultitenantAlertmanager_DeleteUserConfig(t *testing.T) {
...
+	old, _ := yaml.Marshal(testCases)
```
pracucci, line 287:
```suggestion
	old, err := yaml.Marshal(testCases)
	require.NoError(t, err)
```

pracucci (commented):
Thanks for working on this! LGTM modulo few last comments. Once addressed we can merge. Thanks for your patience! 🙏  🙇

AllenzhLi:
> Thanks for working on this! LGTM modulo few last comments. Once addressed we can merge. Thanks for your patience! 🙏 🙇

Thank you very much for your detailed review.

pracucci (approved):
Thanks to patiently address our feedback! ❤️ 🙏

PR #3534, opened 2020-11-24
Validate block range periods config in the compactor

pracucci:
**What this PR does**:
Running a compactor with block range periods which are non divisible is not an issue per se, but doesn't bring the result the user may expect. To avoid mistakes, I would suggest to take a bold position in this case and do not allow block range periods which are not divisible by the previous one. I got tricked by this too and I would prefer to avoid our users get tricked too.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * [ENHANCEMENT] /metrics now supports OpenMetrics output. HTTP and gRPC servers metrics can now include exemplars. #3524
 * [ENHANCEMENT] Expose gRPC keepalive policy options by gRPC server. #3524
 * [ENHANCEMENT] Blocks storage: enabled caching of `meta.json` attributes, configurable via `-blocks-storage.bucket-store.metadata-cache.metafile-attributes-ttl`. #3528
+* [ENHANCEMENT] Compactor: added a config validation check to fail fast if the compactor has been configured invalid block range periods (each period is expected to be a multiple of the previous one). #3534
 * [BUGFIX] Blocks storage ingester: fixed some cases leading to a TSDB WAL corruption after a partial write to disk. #3423
 * [BUGFIX] Blocks storage: Fix the race between ingestion and `/flush` call resulting in overlapping blocks. #3422
 * [BUGFIX] Querier: fixed `-querier.max-query-into-future` which wasn't correctly enforced on range queries. #3452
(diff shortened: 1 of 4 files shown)


pstibrany:
>  but doesn't bring the result the user may expect

I'd like to understand what you mean. Can you elaborate please?

pstibrany (approved):

pracucci:
> > but doesn't bring the result the user may expect
> 
> I'd like to understand what you mean. Can you elaborate please?

Yes, definitely. I've been too cryptic. Let's pick my real case: I configure the block ranges as `2h, 12h, 24h, 7 days, 30 days`. The problem is that the 7 days and 30 days time boundaries don't align, so a bunch of 7 days blocks (crossing the 30d time boundary) will be never compacter either in one "30 days block" or another one. They will just be left as 7 days blocks. The solution was to set it to `2h, 12h, 24h, 7 days, 28 days`. This check would have spotted, but took me a couple of days before realising it.

pracucci:
> > > but doesn't bring the result the user may expect
> > 
> > 
> > I'd like to understand what you mean. Can you elaborate please?
> 
> Yes, definitely. I've been too cryptic. Let's pick my real case: I configure the block ranges as `2h, 12h, 24h, 7 days, 30 days`. The problem is that the 7 days and 30 days time boundaries don't align, so a bunch of 7 days blocks (crossing the 30d time boundary) will be never compacter either in one "30 days block" or another one. They will just be left as 7 days blocks. The solution was to set it to `2h, 12h, 24h, 7 days, 28 days`. This check would have spotted, but took me a couple of days before realising it.

WDYT?

pstibrany:
> Yes, definitely. I've been too cryptic. Let's pick my real case: I configure the block ranges as `2h, 12h, 24h, 7 days, 30 days`. The problem is that the 7 days and 30 days time boundaries don't align, so a bunch of 7 days blocks (crossing the 30d time boundary) will be never compacter either in one "30 days block" or another one. They will just be left as 7 days blocks. The solution was to set it to `2h, 12h, 24h, 7 days, 28 days`. This check would have spotted, but took me a couple of days before realising it.

Thanks for explanation. That makes sense. I would perhaps expect to see 35 days block instead (also not what we really want, but closer to 30 than 7 :)).

PR #3535, opened 2020-11-24
Added metrics to track on-going compaction

pracucci:
**What this PR does**:
Currently we don't have a way to track the progress of on-going compaction when there are many tenants. In this PR I'm proposing to add metrics to be able to track such progress (metrics are reset to 0). I haven't added any unit test because they're pretty hard to write, given these metrics track progress and get reset 0 once the compaction run is done.

I manually tested it in a staging env.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -138,12 +138,16 @@ type Compactor struct {
 	ringSubservicesWatcher *services.FailureWatcher
 
 	// Metrics.
-	compactionRunsStarted     prometheus.Counter
-	compactionRunsCompleted   prometheus.Counter
-	compactionRunsFailed      prometheus.Counter
-	compactionRunsLastSuccess prometheus.Gauge
-	blocksMarkedForDeletion   prometheus.Counter
-	garbageCollectedBlocks    prometheus.Counter
+	compactionRunsStarted          prometheus.Counter
+	compactionRunsCompleted        prometheus.Counter
+	compactionRunsFailed           prometheus.Counter
+	compactionRunsLastSuccess      prometheus.Gauge
+	compactionRunDiscoveredTenants prometheus.Gauge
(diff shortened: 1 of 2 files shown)


pkg/compactor/compactor.go
```diff
@@ -403,21 +433,25 @@ func (c *Compactor) compactUsers(ctx context.Context) error {
 
 		// Ensure the user ID belongs to our shard.
 		if owned, err := c.ownUser(userID); err != nil {
+			c.compactionRunSkippedTenants.Inc()
```
pstibrany, line 436: Or `compactionRunFailedTenants`?

pstibrany (approved):
LGTM, thanks!

pkg/compactor/compactor.go
pracucci: I think skipped is more correct in this case because we're actually skipping that tenant. The `compactionRunFailedTenants` is when a tenant is not skipped and its compaction has failed.

pstibrany: We are skipping tenant, but not because tenant **doesn't belong** to the compactor, but because of **failure to check** if tenant belongs to the compactor. It's not a failed compaction, so I agree that `compactionRunFailedTenants` is not the best metric. Ring has a special metric for this, perhaps compactor should have one as well?

pracucci: Not sure how much it's worth adding another metric, considering the actual reasons why it can happen. I believe another metric could just add more confusion to the user, especially in this context where we want to watch the progress of a compaction run.

pracucci: Merging to move forward, but we can continue the discussion and I will work on a follow-up PR if we decide to act differently.

PR #3536, opened 2020-11-24
API: HTTP GZIP API compression

gotjosh:
**What this PR does**:

Introduces optional compression for both the Ruler and Alertmanager
config and prometheus-compatible API. It defaults to level 6 compression
which seems like acceptable level - we use the same on in the query
front-end.

In practice, we have noticed that tenants might have a high-number of
rules and alerts which tends to size up to 3MB+ of YAML/JSON being output by
the API endpoints.

While testing this in our environments, we saw up to 91% response
compression.

**Checklist**
- [x] Tests updated
- [x] Documentation added


pstibrany (commented):
LGTM, left some minor comments.

docs/configuration/config-file-reference.md
```diff
@@ -69,6 +69,12 @@ Where default_value is the value to use if the environment variable is undefined
 [http_prefix: <string> | default = "/api/prom"]
 
 api:
+  # Use compression for the Ruler and Alertmanager configuration API responses.
+  # Some endpoints of these components serve large YAML or JSON blobs which can
+  # benefit from compression.
+  # CLI flag: -api.enable-response-compression
+  [enable_response_compression: <boolean> | default = false]
```
pracucci, line 76: Would you mind to call it `response_compression_enabled` (and rename CLI flag accordingly) to keep consistency with the new bool naming? In the past we had no consistency, but now we're trying to name it `*_enabled`.

integration/e2ecortex/client.go
```diff
@@ -363,6 +329,40 @@ type userConfig struct {
 	AlertmanagerConfig string            `yaml:"alertmanager_config"`
 }
 
+// GetAlertmanagerConfig gets the status of an alertmanager instance
+func (c *Client) GetAlertmanagerConfig(ctx context.Context) (*alertConfig.Config, error) {
```
pracucci, line 333: Any reason why was moved? Not that I'm against code re-organisation, but it makes diff harder to read. I tried to compare it with `master` and I've the feeling nothing has changed, but maybe I'm missing something.

pkg/api/api.go
```diff
@@ -112,6 +118,14 @@ func (a *API) RegisterRoute(path string, handler http.Handler, auth bool, method
 	a.server.HTTP.Path(path).Methods(methods...).Handler(handler)
 }
 
+func (a *API) RegisterRouteWithCompression(path string, handler http.Handler, auth bool, method string, methods ...string) {
```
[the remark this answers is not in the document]
pracucci: 👍  on this.

pracucci (commented):
Thanks @gotjosh. LGTM. I also share @pstibrany's feedback about checking if we can just reuse `gziphandler.GzipHandler`.

integration/e2ecortex/client.go
gotjosh: Sorry! I'm happy to move it back if needbe - nothing changed, this was grouped with the functions that related to the ruler so I moved it further down to be with the functions that relate to the Alertmanager.

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/golang-migrate/migrate/v4 v4.7.0
 	github.com/golang/protobuf v1.4.3
 	github.com/golang/snappy v0.0.2
+	github.com/gorilla/handlers v1.5.1
```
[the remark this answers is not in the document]
gotjosh: Good shout! Completely missed this.

docs/configuration/config-file-reference.md
gotjosh: Good shout, I tried to mimic other flags we had in the ruler. I'll always err on the side of consistency.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
 * [CHANGE] Blocks storage: increased `-blocks-storage.bucket-store.chunks-cache.attributes-ttl` default from `24h` to `168h` (1 week). #3528
 * [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts. #3305
+* [ENHANCEMENT] Alertmanager and Ruler API: Add HTTP compression to the API responses. Compression can be enabled via `-api.enable-response-compression`. #3536
```
jtlisi, line 8: Is scoping this change to the alertmanager and ruler accurate? Every route is registered with this middleware, so they should all be able to take advantage of it, right?

gotjosh: I changed the scope but did not updated the changelog, let me amend.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
 * [CHANGE] Blocks storage: increased `-blocks-storage.bucket-store.chunks-cache.attributes-ttl` default from `24h` to `168h` (1 week). #3528
 * [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts. #3305
+* [ENHANCEMENT] API: Add HTTP compression to the API responses. Compression can be enabled via `-api.response-compression-enabled`. #3536
```
jtlisi, line 8: **suggestion(non-blocking,if-minor):** May be worth being explicit about the type of compression used.

```suggestion
* [ENHANCEMENT] API: Add GZIP HTTP compression to the API responses. Compression can be enabled via `-api.response-compression-enabled`. #3536
```

pkg/api/api.go
```diff
@@ -46,6 +49,7 @@ type Config struct {
 
 // RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.BoolVar(&cfg.ResponseCompression, "api.response-compression-enabled", false, "Use compression for API responses. Some endpoints serve large YAML or JSON blobs which can benefit from compression.")
```
jtlisi, line 52: **suggestion:** same as above

```suggestion
	f.BoolVar(&cfg.ResponseCompression, "api.response-compression-enabled", false, "Use gzip compression for API responses. Some endpoints serve large YAML or JSON blobs which can benefit from compression.")
```

jtlisi (approved):
LGTM, assuming CI passes

gotjosh:
The integration tests will fail, I can't figure out why the response is not returning the correct header despite the handler being wrapped (you can tell by the `Accept-Encoding; Vary`) in the response. It [should](https://github.com/nytimes/gziphandler/blob/master/gzip_test.go#L70) return the header as expected. 🤔

pstibrany (approved):
LGTM, thanks!

pracucci (commented):
I have a doubt: when response compression is enabled both in the `query-frontend` (see `initQueryFrontend()`) and in the API (new flag in this PR), the gzip handler will be injected twice. Doesn't this break the query-frontend?

pstibrany:
> I have a doubt: when response compression is enabled both in the `query-frontend` (see `initQueryFrontend()`) and in the API (new flag in this PR), the gzip handler will be injected twice. Doesn't this break the query-frontend?

I checked the code, and I think having two of gzip handlers will work fine (when outer one sets content encoding to "gzip", inner one will not do any gzipping at that point, as it requires content-encoding to be ""). 

I think we should deprecate frontend-specific setting in favor of this PR, and remove in in couple of releases.

pracucci:
> I think we should deprecate frontend-specific setting in favor of this PR, and remove it in couple of releases.

Agree. But if it's not an issue in this PR, then we can do it in a follow up PR. Merging it!

PR #3537, opened 2020-11-24
Exported process metrics to monitor the number of memory map areas allocated

pracucci:
**What this PR does**:
The Cortex blocks storage (in particular TSDB) does an extensive usage of mmap-ed files. The linux kernel has a limit on the max number of memory map areas a process can allocate (defaults to 65K) and I would like to alert on it. Unfortunately neither cAdvisor or default process metrics export it, so I'm proposing to export them directly from Cortex.

I looked at `procfs` and it supports reading both the maps and the limit, but it does way more than we need (eg. it parse every single map entry) so I decided to just open the file and count/parse by myself, so that we don't waste CPU and memory parsing something we don't need.

Tested in a Kubernetes cluster running on Linux:
Screenshot 2020-11-24 at 15 48 14

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/process/collector.go
+++ b/pkg/process/collector.go
@@ -1,0 +1,121 @@
+package process
+
+import (
+	"bufio"
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/procfs"
+)
+
+type Options struct {
+	Pid            int
+	ProcMountPoint string
+}
+
+type processCollector struct {
+	opts Options
+
+	// Metrics.
+	currMaps *prometheus.Desc
+	maxMaps  *prometheus.Desc
+}
+
+// NewProcessCollector makes a new custom process collector used to collect process metrics the
+// default instrumentation doesn't support.
+func NewProcessCollector(opts Options) prometheus.Collector {
+	// Apply default options.
+	if opts.Pid == 0 {
+		opts.Pid = os.Getpid()
+	}
+	if opts.ProcMountPoint == "" {
+		opts.ProcMountPoint = procfs.DefaultMountPoint
+	}
+
+	c := &processCollector{
+		opts: opts,
+		currMaps: prometheus.NewDesc(
+			"process_memory_map_areas",
+			"Number of memory map areas allocated by the process.",
+			nil, nil,
+		),
+		maxMaps: prometheus.NewDesc(
+			"process_memory_map_areas_limit",
+			"Maximum number of memory map ares the process can allocate.",
+			nil, nil,
+		),
+	}
+
+	return c
+}
+
+// Describe returns all descriptions of the collector.
+func (c *processCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.currMaps
+	ch <- c.maxMaps
+}
+
+// Collect returns the current state of all metrics of the collector.
+func (c *processCollector) Collect(ch chan<- prometheus.Metric) {
+	if value, err := c.getMapsCount(); err == nil {
+		ch <- prometheus.MustNewConstMetric(c.currMaps, prometheus.GaugeValue, value)
+	}
+
+	if value, err := c.getMapsCountLimit(); err == nil {
+		ch <- prometheus.MustNewConstMetric(c.maxMaps, prometheus.GaugeValue, value)
+	}
+}
+
+// getMapsCount returns the number of memory map ares the process has allocated.
+func (c *processCollector) getMapsCount() (float64, error) {
+	file, err := os.Open(processMapsPath(c.opts.ProcMountPoint, c.opts.Pid))
+	if err != nil {
+		return 0, err
+	}
+	defer file.Close()
+
+	count := 0
+	for scan := bufio.NewScanner(file); scan.Scan(); {
+		count++
+	}
+
+	return float64(count), nil
+}
+
+// getMapsCountLimit returns the maximum of memory map ares the process can allocate.
+func (c *processCollector) getMapsCountLimit() (float64, error) {
+	file, err := os.Open(vmMapsLimitPath(c.opts.ProcMountPoint))
+	if err != nil {
+		return 0, err
+	}
+	defer file.Close()
+
+	content, err := ioutil.ReadAll(file)
+	if err != nil {
+		return 0, err
+	}
+
+	content = bytes.TrimSpace(content)
+
+	// A base value of zero makes ParseInt infer the correct base using the
+	// string's prefix, if any.
+	const base = 0
+	value, err := strconv.ParseInt(string(content), base, 64)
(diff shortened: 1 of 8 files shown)


pkg/process/collector.go
```diff
@@ -0,0 +1,121 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/procfs"
+)
+
+type Options struct {
```
pstibrany, line 15: I suggest removing Options, to simplify the code, and hardcode "/proc/self/maps" as file to read. Mounting procfs to other place breaks so many other things that we can safely assume this.

pkg/process/collector.go
```diff
@@ -0,0 +1,121 @@
...
+	opts Options
+
+	// Metrics.
+	currMaps *prometheus.Desc
+	maxMaps  *prometheus.Desc
+}
+
+// NewProcessCollector makes a new custom process collector used to collect process metrics the
+// default instrumentation doesn't support.
+func NewProcessCollector(opts Options) prometheus.Collector {
```
pstibrany, line 30: I'd suggest moving this to linux specific file, so that we can avoid doing the work (opening non-existant files) on other platforms.

For general implementation, NewProcessCollector can return nil, or empty collector.

pstibrany (approved):

pkg/process/collector.go
pracucci: This was actually done for unit tests. We don't let Cortex users to configure it, but having options looks an easy way to mock `/proc` for unit tests. Same for the pid.

pkg/process/collector.go
pracucci: Good suggestion. Will do.

pkg/process/collector.go
```diff
@@ -0,0 +1,121 @@
+package process
```
jtlisi, line 1: **question:** This package isn't scoped to a particular cortex-module. In the past packages of this nature have been stored in`pkg/util/*`. Any reason you chose not to? IMO reducing the usage of the util package as a catch all would lead us towards more discrete package design overall.

jtlisi (approved):
LGTM, one repo-meta question

pkg/process/collector.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"bufio"
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/procfs"
```
pstibrany, line 12: This dependency is only used for “/proc” const string. As “Go proverbs” go, “A little copying is better than a little dependency”.

pkg/process/collector.go
pstibrany: Another (simpler?) option would be to pass those two files for testing. This is just a non-blocking nit.

pstibrany (approved):

pkg/process/collector.go
pracucci: Right. Moved to `pkg/util/process`.

pkg/process/collector.go
pracucci: Good point. Done.

pkg/process/collector.go
pracucci: Re-thinking it, if I do it, unit tests will not run anymore on OSX. Is this a real issue? Is it a big deal if we try to open a couple of non-existing files on unsupported systems, given Cortex is prod is likely to run on Linux only?

pstibrany: It's not a big deal, but it is a wasted work. As an alternative, we can keep the code generic, but check for those files once, and only setup compactor when they exist.

pracucci: Ok. I've added a check to see if the proc FS is supported.

pkg/util/process/collector.go
```diff
@@ -0,0 +1,133 @@
...
+// default instrumentation doesn't support.
+func NewProcessCollector() prometheus.Collector {
+	return newProcessCollector(os.Getpid(), DefaultProcMountPoint)
+}
+
+func newProcessCollector(pid int, procMountPoint string) prometheus.Collector {
+	c := &processCollector{
+		pid:            pid,
+		procMountPoint: procMountPoint,
+		supported:      isSupported(procMountPoint),
```
pstibrany, line 39: Super-Nit: If not supported, this method should return empty collector or nil. No need for having `supported` field and check it on each collection.

pkg/util/process/collector.go
```diff
@@ -0,0 +1,133 @@
...
+	ch <- c.maxMaps
+}
+
+// Collect returns the current state of all metrics of the collector.
+func (c *processCollector) Collect(ch chan<- prometheus.Metric) {
+	if !c.supported {
+		return
+	}
+
+	if value, err := c.getMapsCount(); err == nil {
```
pstibrany, line 71: If err is not nil, we should return -1. By skipping this metric on error, we break metric collection. Same is true below.

pkg/util/process/collector.go
```diff
@@ -0,0 +1,133 @@
...
+		return 0, err
+	}
+	defer file.Close()
+
+	count := 0
+	for scan := bufio.NewScanner(file); scan.Scan(); {
+		count++
+	}
+
+	return float64(count), nil
```
pstibrany, line 93: We should return `scan.Err()` as error.

pkg/util/process/collector.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	content = bytes.TrimSpace(content)
+
+	// A base value of zero makes ParseInt infer the correct base using the
+	// string's prefix, if any.
+	const base = 0
+	value, err := strconv.ParseInt(string(content), base, 64)
```
pstibrany, line 114: Base=0 seems wrong. If value starts with zero, it will be treated as 8-based. I don't think that's correct.

pstibrany (approved):
LGTM, but I have few more non-blocking comments.

pkg/util/process/collector.go
pracucci: I picked it from`procfs` implementation. I trust them. I would keep it.

pkg/util/process/collector.go
pracucci: I asked Beorn cause I had a doubt whether this is true or not:
> It is OK to return a descriptor for which you never (or not always) return a metric.

Basically it's not a problem here. I think we've seen the issue you describe long time ago when working with the metrics aggregator helper in Cortex.

pkg/util/process/collector.go
pstibrany: I don't think that is a correct decision, but let's commit to it and see how it works.

PR #3539, opened 2020-11-25
Add basic query stats collection & logging.

tomwilkie:
**What this PR does**:

Current collects & logs wall clock time and number of series in the queries.

TODO:
- [ ] collection number of samples
- [ ] propagate & aggregate in the query frontend
- [ ] add some metrics.
- [ ] for labels and series endpoints

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/api/handlers.go
```diff
@@ -217,13 +218,15 @@ func NewQuerierHandler(
 	legacyPromRouter := route.New().WithPrefix(legacyPrefix + "/api/v1")
 	api.Register(legacyPromRouter)
 
+	statsMiddleware := stats.NewMiddleware(logger)
```
pracucci, line 221: Should we conditionally enable it with a flag?

pkg/querier/blocks_store_queryable.go
```diff
@@ -357,6 +359,8 @@ func (q *blocksStoreQuerier) selectSorted(sp *storage.SelectHints, matchers ...*
 		storage.EmptySeriesSet()
 	}
 
+	stats.Series += len(resSeriesSets)
```
pracucci, line 362: `len(resSeriesSets)` is not the number of series, but series set. A single set contains all series returned by 1 store-gateway. We may return the number of series back from store-gateway response (because counting it here would mean iterating the entire set and it's a waste of resources), but this wouldn't be an accurate count anyway because (1) doesn't take in account series from ingesters (2) series from store-gateways may be duplicated.

An option may be wrapping the `storage.SeriesSet` returned by `querier.Select()` and counting the number of `Series()` returned while iterating `SeriesSet`, updating the stats in the context once the the end of set (or error) is reached.

pkg/querier/stats/report_middleware.go
```diff
@@ -0,0 +1,62 @@
...
+		stats, ctx := ContextWithEmptyStats(r.Context())
+		r = r.WithContext(ctx)
+
+		startTime := time.Now()
+		next.ServeHTTP(w, r)
+
+		// Track statistics.
+		m.querySeconds.WithLabelValues(userID).Add(float64(stats.LoadWallTime()))
+
+		level.Info(m.logger).Log(
```
pstibrany, line 53: Instead of using `m.logger` directly, can we use `util.WithContext(r.Context(), m.logger)` to make sure that also trace ID is logged, if available?

pkg/querier/stats/report_middleware.go
```diff
@@ -0,0 +1,62 @@
...
+			"msg", "query stats",
+			"user", userID,
+			"method", r.Method,
+			"path", r.URL.Path,
```
pstibrany, line 57: I think we should log the query itself as well. Perhaps it would make sense to move this stats logging code to the `transport.Handler`, where frontend also logs slow queries. WDYT?

pkg/frontend/config.go
```diff
@@ -31,6 +32,7 @@ func (cfg *CombinedFrontendConfig) RegisterFlags(f *flag.FlagSet) {
 
 	f.BoolVar(&cfg.CompressResponses, "querier.compress-http-responses", false, "Compress HTTP responses.")
 	f.StringVar(&cfg.DownstreamURL, "frontend.downstream-url", "", "URL of downstream Prometheus.")
+	f.BoolVar(&cfg.QueryStatsEnabled, "frontend.query-stats-enabled", false, "True to enable query statistics tracking. When enabled, a message with some statistics is logged for every query. This configuration option must be set both on query-frontend and querier.")
```
pstibrany, line 35: If we move stats-setup in the context into handler, this flag should go to HandlerConfig.

pkg/frontend/v2/frontend.go
```diff
@@ -228,7 +229,12 @@ enqueueAgain:
 		return nil, ctx.Err()
 
 	case resp := <-freq.response:
-		return resp, nil
+		if stats.ShouldTrackHTTPGRPCResponse(resp.HttpResponse) {
+			stats := stats.FromContext(ctx)
+			stats.Merge(resp.Stats)
```
pstibrany, line 234: I'd add comment that it's ok if `stats` is nil here, otherwise it looks like it may panic.

pkg/querier/stats/stats.go
```diff
@@ -0,0 +1,62 @@
...
+	if s == nil || other == nil {
+		return
+	}
+
+	s.AddWallTime(other.LoadWallTime())
+}
+
+func ShouldTrackHTTPGRPCResponse(r *httpgrpc.HTTPResponse) bool {
+	// Do no track statistics for requests failed because of a server error.
+	return r.Code < 500 || r.Code > 599
```
pstibrany, line 61: There are no HTTP status codes defined above 599. Perhaps more explicit `200 <= code && code <= 399` would be clearer?

pstibrany (approved):
LGTM

I'd suggest moving context-setup code and stats logging into frontend `handler` package, to be able to log the query itself, and not just HTTP method and request URI. I think these stats are pointless without having logging the query itself.

pkg/querier/stats/report_middleware.go
pracucci: Sure, makes sense.

pkg/querier/stats/stats.go
pracucci: 4xx should be included, but I picked up your comment about being more explicit.

pracucci:
Thanks @pstibrany for your review! I should have addressed your comments.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   - limit for outgoing gRPC messages has changed from 2147483647 to 16777216 bytes
   - limit for incoming gRPC messages has changed from 4194304 to 104857600 bytes
 * [FEATURE] Distributor/Ingester: Provide ability to not overflow writes in the presence of a leaving or unhealthy ingester. This allows for more efficient ingester rolling restarts. #3305
+* [FEATURE] Query-frontend: introduced query statistics logged in the query-frontend when enabled via `-frontend.query-stats-enabled=true`. When enabled the following metrics are also tracked: #3539
+  - `cortex_query_seconds_total`
```
bboreham, line 13: Can you explain somewhere (e.g. in the docs) how this metric differers from `cortex_request_duration_seconds{route="api_prom_api_v1_query_range"}`

gouthamve:
Note: https://github.com/prometheus/prometheus/pull/6890 is the right way to track samples and series and is currently blocked because some benchmarks are significantly worse.

When we try to implement the _series and samples_ tracking ourselves, we should keep an eye on the benchmarks.

pracucci:
> Note: [prometheus/prometheus#6890](https://github.com/prometheus/prometheus/pull/6890) is the right way to track samples and series and is currently blocked because some benchmarks are significantly worse.
> 
> When we try to implement the _series and samples_ tracking ourselves, we should keep an eye on the benchmarks.

Agree 💯  on the benchmarks. However, I believe there's no reasonable way to implement it ourselves. I think should be tracked by PromQL engine to get the count done correctly.

CHANGELOG.md
pracucci: Yes, sure. Done.

pstibrany (approved):
LGTM, thanks!

pkg/frontend/transport/handler.go
```diff
@@ -34,11 +39,13 @@ var (
 type HandlerConfig struct {
 	LogQueriesLongerThan time.Duration `yaml:"log_queries_longer_than"`
 	MaxBodySize          int64         `yaml:"max_body_size"`
+	QueryStatsEnabled    bool          `yaml:"query_stats_enabled"`
 }
 
 func (cfg *HandlerConfig) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.LogQueriesLongerThan, "frontend.log-queries-longer-than", 0, "Log queries that are slower than the specified duration. Set to 0 to disable. Set to < 0 to enable on all queries.")
 	f.Int64Var(&cfg.MaxBodySize, "frontend.max-body-size", 10*1024*1024, "Max body size for downstream prometheus.")
+	f.BoolVar(&cfg.QueryStatsEnabled, "frontend.query-stats-enabled", false, "True to enable query statistics tracking. When enabled, a message with some statistics is logged for every query. This configuration option must be set both on query-frontend and querier.")
```
bboreham, line 48: Why did this get the "frontend" name when it goes on both?

pstibrany: It's now only set on frontend: https://github.com/cortexproject/cortex/pull/3595

pkg/querier/stats/report_middleware.go
pstibrany: It can be useful if you want to bill your customers based on number of queries.

bboreham: That requirement seems better met by a metric than logs.

bboreham: Also could already be met by setting `-frontend.log-queries-longer-than` negative.

pracucci: > Also could already be met by setting -frontend.log-queries-longer-than negative.

That would track the HTTP response time but no the wall time aggregated by all queriers (which was the need we had here).

pracucci: > Also could already be met by setting -frontend.log-queries-longer-than negative.

Yes, we also track `cortex_query_seconds_total`. Whether the Cortex operator would use metric or log really depends on their use case.

bboreham: OK I will make a PR to have a separate option.

PR #3540, opened 2020-11-25
Add option to use v2 signatures for S3 blocks store

simonswine:
**What this PR does**:

- This allows to configure the block store client to use V2 signatures for S3 authentication. 
- This updates thanos to include https://github.com/thanos-io/thanos/pull/3496

**Checklist**
- [ ] Tests updated
- [X] Documentation added

### DIFF
--- a/pkg/storage/backend/s3/config.go
+++ b/pkg/storage/backend/s3/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	SecretAccessKey flagext.Secret `yaml:"secret_access_key"`
 	AccessKeyID     string         `yaml:"access_key_id"`
 	Insecure        bool           `yaml:"insecure"`
+	SignatureV2     bool           `yaml:"signature_v2"`
 
 	HTTP HTTPConfig `yaml:"http"`
 }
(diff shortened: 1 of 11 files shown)


pkg/storage/backend/s3/config.go
```diff
@@ -32,6 +32,7 @@ type Config struct {
 	SecretAccessKey flagext.Secret `yaml:"secret_access_key"`
 	AccessKeyID     string         `yaml:"access_key_id"`
 	Insecure        bool           `yaml:"insecure"`
+	SignatureV2     bool           `yaml:"signature_v2"`
```
pstibrany, line 35: I think we’re missing setting of the corresponding field here: https://github.com/cortexproject/cortex/blob/574df71ae758a2d055008f599afd36b632c2f690/pkg/storage/backend/s3/bucket_client.go#L20

pracucci: Yes, this field is currently unused. Needs to be passed to the S3 client.

Another comment I have. Such booleans are bad for extensibility in the future, in case a new version scheme will be introduced. You may achieve the same result allowing to configure the version number `signature_version: 2|4`, introducing a check in `Config.Validate()` (function to be added and called from parents) to make sure it's either 2 or 4. If in the future v6 will be introduced, this flag will support it too.

simonswine: Thanks for spotting the missing passing-through :grimacing: 

On using an enum parameter: I have done that as part of the last commit. I think it might be misrepresenting what is actually happening, I am not too sure as in the background the minio client seems to use https://github.com/minio/minio-go/blob/26addf203d4a7d73e33f0731b67a7c02a910632a/pkg/credentials/signature-type.go#L28 `SignatureDefault`, which is equal to `v4`. It might be more accurate to rename the parameters from `v4` to `default`, but I felt that is more confusing.

Btw: I have purposely used strings with `v` prefix to avoid YAML parsing issues around integers vs. strings.

pstibrany (approved):
LGTM, thanks!

pracucci (approved):
LGTM, thanks!

PR #3544, opened 2020-11-26
[Querier] Deprecate `-querier.compress-http-responses-deprecated`

gotjosh:
**What this PR does**:
Deprecate `-querier.compress-http-responses-deprecated` In favour of `-api.response-compression-enabled` which applies to every other API endpoint we register.

A follow up from #3536 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,9 @@
 
 ## master / unreleased
 
-<<<<<<< HEAD
 ## 1.6.0-rc.0 in progress
 
-=======
 * [CHANGE] Query Frontend: deprecate `-querier.compress-http-responses` in favour of `-api.response-compression-enabled`. #3544
->>>>>>> 572de8827 ([Querier] Deprecate `-querier.compress-http-responses`)
 * [CHANGE] Querier: deprecated `-store.max-look-back-period`. You should use `-querier.max-query-lookback` instead. #3452
 * [CHANGE] Blocks storage: increased `-blocks-storage.bucket-store.chunks-cache.attributes-ttl` default from `24h` to `168h` (1 week). #3528
 * [CHANGE] Blocks storage: the config option `-blocks-storage.bucket-store.index-cache.postings-compression-enabled` has been deprecated and postings compression is always enabled. #3538
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 ## master / unreleased
 
+<<<<<<< HEAD
 ## 1.6.0-rc.0 in progress
 
+=======
+* [CHANGE] Query Frontend: deprecate `-querier.compress-http-responses` in favour of `-api.response-compression-enabled`. #3544
+>>>>>>> 572de8827 ([Querier] Deprecate `-querier.compress-http-responses`)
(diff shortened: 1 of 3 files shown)


pkg/cortex/modules.go
```diff
@@ -517,13 +516,7 @@ func (t *Cortex) initQueryFrontend() (serv services.Service, err error) {
 
 	// Wrap roundtripper into Tripperware.
 	roundTripper = t.QueryFrontendTripperware(roundTripper)
-
-	handler := transport.NewHandler(t.Cfg.Frontend.Handler, roundTripper, util.Logger)
```
pracucci, line 521: If we deprecate removing the code as well we introduce a breaking change (which we can't, I think). We should keep the logic as is, deprecated it in the CLI flag description + CHANGELOG and in 2 minor versions we'll remove it. @gouthamve may you confirm it, please?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Query Frontend: depcrated `-querier.compress-http-responses-deprecated`. You should use `-api.response-compression-enabled` instead. #3544
```
jtlisi, line 5:
```suggestion
* [CHANGE] Query Frontend: deprecate `-querier.compress-http-responses-deprecated` in favor of `-api.response-compression-enabled`. #3544
```

pkg/cortex/modules.go
jtlisi: @gotjosh on a related not, does having this flag and the new flag set simultaneously cause an issues? If so you could add a check below to only inject the gzip handler if the query-frontend flag is enabled but not the generic compression flag.

pracucci: > does having this flag and the new flag set simultaneously cause an issues?

We checked when we reviewed the previuos PR (where the new flag has been introduced) and it shouldn't cause any issue.

CHANGELOG.md
bboreham: The flag appears to be named `-querier.compress-http-responses` ?

pkg/cortex/modules.go
gotjosh: Is this better?

pkg/frontend/config.go
```diff
@@ -20,16 +20,20 @@ type CombinedFrontendConfig struct {
 	FrontendV1 v1.Config               `yaml:",inline"`
 	FrontendV2 v2.Config               `yaml:",inline"`
 
-	CompressResponses bool   `yaml:"compress_responses"`
-	DownstreamURL     string `yaml:"downstream_url"`
+	// Deprecated. Replaced with pkg/api/Config.ResponseCompression field.
+	// TODO: To be removed in Cortex 1.7.
```
pracucci, line 24:
```suggestion
	// TODO: To be removed in Cortex 1.8.
```

pracucci (approved):
LGTM (with a nit)

pkg/frontend/config.go
gotjosh: done!

jtlisi (approved):
LGTM

PR #3547, opened 2020-11-28
blocksconvert: Normalize labels from chunks.

pstibrany:
**What this PR does**: During conversion of some old data, we have found chunks that have duplicate labels written in them. As long as label values are the same, we can still convert such data.

This PR implements chunks label normalization – makes sure that labels are sorted, and removes duplicate labels. If duplicate labels have different values, it will error out.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci (approved):
Good job, LGTM!

PR #3549, opened 2020-12-01
Tenant deletion when using blocks storage

pstibrany:
**What this PR does**: This PR introduces new API for tenant deletion when using blocks storage. There are two API endpoints: one to trigger tenant deletion (it really only writes mark file with tenant prefix), another one to check for status (= check for blocks).

In this PR ingester checks for tenant deletion mark, and 1) stops shipping blocks for such tenant, and 2) closes TSDB faster (if closing is enabled).

In subsequent PRs, more components will react on deletion mark:
- compactor will clean up blocks in the bucket
- ruler will clean up rule groups, and stop rule evaluation [ruler will use separate marker in its own bucket]
- alertmanager will clean up its per-tenant config and state [alertmanager will use separate marker in its own bucket]

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -215,10 +215,10 @@ func (a *API) RegisterIngester(i Ingester, pushConfig distributor.Config) {
 	a.RegisterRoute("/push", push.Handler(pushConfig, a.sourceIPs, i.Push), true, "POST") // For testing and debugging.
 }
 
-// RegisterPurger registers the endpoints associated with the Purger/DeleteStore. They do not exactly
+// RegisterChunksPurger registers the endpoints associated with the Purger/DeleteStore. They do not exactly
 // match the Prometheus API but mirror it closely enough to justify their routing under the Prometheus
 // component/
-func (a *API) RegisterPurger(store *purger.DeleteStore, deleteRequestCancelPeriod time.Duration) {
+func (a *API) RegisterChunksPurger(store *purger.DeleteStore, deleteRequestCancelPeriod time.Duration) {
 	deleteRequestHandler := purger.NewDeleteRequestHandler(store, deleteRequestCancelPeriod, prometheus.DefaultRegisterer)
 
 	a.RegisterRoute(a.cfg.PrometheusHTTPPrefix+"/api/v1/admin/tsdb/delete_series", http.HandlerFunc(deleteRequestHandler.AddDeleteRequestHandler), true, "PUT", "POST")
@@ -229,6 +229,11 @@ func (a *API) RegisterPurger(store *purger.DeleteStore, deleteRequestCancelPerio
 	a.RegisterRoute(a.cfg.LegacyHTTPPrefix+"/api/v1/admin/tsdb/delete_series", http.HandlerFunc(deleteRequestHandler.AddDeleteRequestHandler), true, "PUT", "POST")
 	a.RegisterRoute(a.cfg.LegacyHTTPPrefix+"/api/v1/admin/tsdb/delete_series", http.HandlerFunc(deleteRequestHandler.GetAllDeleteRequestsHandler), true, "GET")
 	a.RegisterRoute(a.cfg.LegacyHTTPPrefix+"/api/v1/admin/tsdb/cancel_delete_request", http.HandlerFunc(deleteRequestHandler.CancelDeleteRequestHandler), true, "PUT", "POST")
+}
+
+func (a *API) RegisterBlocksPurger(api *purger.BlocksPurgerAPI) {
+	a.RegisterRoute("/purger/delete_tenant", http.HandlerFunc(api.DeleteTenant), true, "POST")
+	a.RegisterRoute("/purger/delete_tenant_status", http.HandlerFunc(api.DeleteTenantStatus), true, "GET")
 }
 
 // RegisterRuler registers routes associated with the Ruler service.
--- a/pkg/chunk/purger/blocks_purger_api.go
+++ b/pkg/chunk/purger/blocks_purger_api.go
@@ -1,0 +1,121 @@
+package purger
+
+import (
+	"context"
+	"net/http"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/oklog/ulid"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/objstore"
+
+	"github.com/cortexproject/cortex/pkg/storage/bucket"
+	cortex_tsdb "github.com/cortexproject/cortex/pkg/storage/tsdb"
+	"github.com/cortexproject/cortex/pkg/tenant"
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+type BlocksPurgerAPI struct {
+	bucketClient objstore.Bucket
+	logger       log.Logger
+}
+
+func NewBlocksPurgerAPI(storageCfg cortex_tsdb.BlocksStorageConfig, logger log.Logger, reg prometheus.Registerer) (*BlocksPurgerAPI, error) {
+	bucketClient, err := createBucketClient(storageCfg, logger, reg)
+	if err != nil {
+		return nil, err
+	}
+
+	return newBlocksPurgerAPI(bucketClient, logger), nil
+}
+
+func newBlocksPurgerAPI(bkt objstore.Bucket, logger log.Logger) *BlocksPurgerAPI {
+	return &BlocksPurgerAPI{bucketClient: bkt, logger: logger}
+}
+
+func (api *BlocksPurgerAPI) DeleteTenant(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	userID, err := tenant.TenantID(ctx)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	err = cortex_tsdb.WriteTenantDeletionMark(r.Context(), api.bucketClient, userID)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	level.Info(api.logger).Log("msg", "tenant deletion marker created", "user", userID)
+
+	w.WriteHeader(http.StatusOK)
+}
+
+type DeleteTenantStatusResponse struct {
+	TenantID                  string `json:"tenant_id"`
+	BlocksDeleted             bool   `json:"blocks_deleted"`
+	RuleGroupsDeleted         bool   `json:"rule_groups_deleted"`
+	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted"`
+}
+
+func (api *BlocksPurgerAPI) DeleteTenantStatus(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	userID, err := tenant.TenantID(ctx)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	result := DeleteTenantStatusResponse{}
+	result.TenantID = userID
+	result.BlocksDeleted, err = api.checkBlocksForUser(ctx, userID)
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	util.WriteJSONResponse(w, result)
+}
+
+func (api *BlocksPurgerAPI) checkBlocksForUser(ctx context.Context, userID string) (bool, error) {
+	var errBlockFound = errors.New("block found")
+
+	userBucket := bucket.NewUserBucketClient(userID, api.bucketClient)
+	err := userBucket.Iter(ctx, "", func(s string) error {
+		s = strings.TrimSuffix(s, "/")
+
+		_, err := ulid.Parse(s)
+		if err != nil {
+			// not block, keep looking
+			return nil
+		}
+
+		// Used as shortcut to stop iteration.
+		return errBlockFound
+	})
+
+	if errors.Is(err, errBlockFound) {
+		return false, nil
+	}
+
+	if err != nil {
+		return false, err
+	}
+
+	// No blocks found, all good.
+	return true, nil
+}
+
+func createBucketClient(cfg cortex_tsdb.BlocksStorageConfig, logger log.Logger, reg prometheus.Registerer) (objstore.Bucket, error) {
+	bucketClient, err := bucket.NewClient(context.Background(), cfg.Bucket, "purger", logger, reg)
+	if err != nil {
+		return nil, errors.Wrap(err, "create bucket client")
+	}
+
+	return bucketClient, nil
+}
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -58,21 +58,26 @@ const (
 	closed                           // Used to avoid setting closing back to active in closeAndDeleteIdleUsers method.
 )
 
-// Describes result of TSDB idle check (for closing). String is used as metric label.
-type tsdbIdleCheckResult string
+// Describes result of TSDB-close check. String is used as metric label.
+type tsdbCloseCheckResult string
 
 const (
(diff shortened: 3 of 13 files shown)


pkg/storage/tsdb/tenant_deletion_mark.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"encoding/json"
+	"path"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+const TenantDeletionMarkFilename = "tenant-deleted.json"
```
pracucci, line 14: Two things:
1. Can we store it under the `marker/` prefix, within the tenant location? This way will work nicely with the bucket index too (we'll store the block deletion marks there).
2. What do you think to name it `tenant-deletion-mark.json` to keep naming consistent with the block deletion mark?

pkg/storage/tsdb/tenant_deletion_mark.go
```diff
@@ -0,0 +1,39 @@
...
+
+type TenantDeletionMark struct {
+	// Unix timestamp when deletion marker was created.
+	Created int64 `json:"created"`
```
pracucci, line 18: I would keep it consistent with the block deletion mark:

```
	DeletionTime int64 `json:"deletion_time"`
```

pkg/storage/tsdb/tenant_deletion_mark.go
```diff
@@ -0,0 +1,39 @@
...
+func WriteTenantDeletionMark(ctx context.Context, bkt objstore.Bucket, userID string) error {
+	m := &TenantDeletionMark{Created: time.Now().Unix()}
+
+	data, err := json.Marshal(m)
+	if err != nil {
+		return errors.Wrap(err, "serialize deletion mark")
+	}
+
+	markerFile := path.Join(userID, TenantDeletionMarkFilename)
+	return errors.Wrap(bkt.Upload(ctx, markerFile, bytes.NewReader(data)), "upload deletion mark")
```
pracucci, line 38:
```suggestion
	return errors.Wrap(bkt.Upload(ctx, markerFile, bytes.NewReader(data)), "upload tenant deletion mark")
```

pkg/storage/tsdb/tenant_deletion_mark.go
```diff
@@ -0,0 +1,39 @@
...
+	return bkt.Exists(ctx, markerFile)
+}
+
+// Uploads deletion mark to the tenant "directory".
+func WriteTenantDeletionMark(ctx context.Context, bkt objstore.Bucket, userID string) error {
+	m := &TenantDeletionMark{Created: time.Now().Unix()}
+
+	data, err := json.Marshal(m)
+	if err != nil {
+		return errors.Wrap(err, "serialize deletion mark")
```
pracucci, line 34:
```suggestion
		return errors.Wrap(err, "serialize tenant deletion mark")
```

pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -0,0 +1,116 @@
...
+
+func newBlocksPurgerAPI(bkt objstore.Bucket) *BlocksPurgerAPI {
+	return &BlocksPurgerAPI{bucketClient: bkt}
+}
+
+func (api *BlocksPurgerAPI) DeleteTenant(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	userID, err := tenant.TenantID(ctx)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
pracucci, line 40: We return `http.StatusUnauthorized` is other APIs. We should be consistent here too.

_Same comment applies to `DeleteTenantStatus()`._

pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+
+	err = cortex_tsdb.WriteTenantDeletionMark(r.Context(), api.bucketClient, userID)
```
pracucci, line 44: Should we check if already exists, in order to not overwrite it? May be useful to not overwrite, to make sure that the "deletion time" in the marker is the oldest one.

pracucci (commented):
Very good job! Design and implementation LGTM. I just left few minor comments, some related to my work on bucketindex.

Have you considered:
- Add a CHANGELOG entry
- Update the HTTP API doc (mentioning it's experimental)
- Add the tenant deletion to the list of experimental features in the doc

pkg/ingester/ingester_v2.go
```diff
@@ -1454,6 +1471,26 @@ func (i *Ingester) shipBlocks(ctx context.Context) {
...
+		if userDB.deletionMarkFound.Load() {
+			return nil
+		}
+
+		if time.Since(time.Unix(userDB.lastDeletionMarkCheck.Load(), 0)) > cortex_tsdb.DeletionMarkCheckInterval {
+			userDB.lastDeletionMarkCheck.Store(time.Now().Unix())
+
+			deletionMarkExists, err := cortex_tsdb.TenantDeletionMarkExists(ctx, i.TSDBState.bucket, userID)
+			if err != nil {
+				// If we cannot check for deletion mark, it's likely that shipping would fail anyway.
```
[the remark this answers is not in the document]
pstibrany: I've changed that since it makes tests that have no bucket defined to fail.

pkg/ingester/ingester_v2.go
```diff
@@ -1454,6 +1471,26 @@ func (i *Ingester) shipBlocks(ctx context.Context) {
...
+				level.Warn(util.Logger).Log("msg", "failed to check for tenant deletion mark before shipping", "user", userID, "err", err)
+				return nil
+			} else if deletionMarkExists {
+				userDB.deletionMarkFound.Store(true)
```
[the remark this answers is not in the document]
pstibrany: I don't think we should. Supporting rollback is entire new can of worms (compactor will actually delete data when it finds the mark, it cannot do undelete), and I don't think we should give false hope here by setting it back to false. Furthermore, once `deletionMarkFound` is set to true, we don't even get to this point.

pkg/storage/tsdb/tenant_deletion_mark.go
pstibrany: Sure.

pkg/chunk/purger/blocks_purger_api.go
pstibrany: I don't think this is correct. This endpoint uses authentication middleware, which already checks if there is correct x-scope-orgID header, and stores tenant into context. Auth middleware already returns `http.StatusUnauthorized`. If we get error here, it can be for different reasons, e.g. using multi-tenants.

pstibrany: (Btw, I purger's other handler uses `http.StatusBadRequest`, instead of `http.StatusUnauthorized`. I copied it from there. I think it's more correct.)

pkg/chunk/purger/blocks_purger_api.go
pstibrany: It doesn't make difference from API point of view, and overwriting with more recent timestamp will not do any harm. Additionally, it would be best-effort-attempt only. I suggest that we instead log a message on successful creation, so that it is possible to find out if there were multiple markers, if needed for troubleshooting.

pkg/chunk/purger/blocks_purger_api.go
pracucci: I see your point. I'm OK keeping it as is in this PR. However, I think it's worth opening a GitHub issue about it, cause we should keep a keep a consistent approach within Cortex.

pkg/chunk/purger/blocks_purger_api.go
pracucci: It's true doesn't cause any harm and it's true would be a best-effort attempt, but I still believe keeping the original deletion timestamp in the marker would be useful for debugging (eg. you're assuming every Cortex users collects and stores Cortex logs for a reasonable amount of time, which I'm not sure is true).

Reversing your answer, I may say that even checking for existence and returning a "no content" status code without overwriting it doesn't cause any harm and would avoid to overwrite it under normal circumnstances.

In any case, not a blocker to me.

pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	level.Info(api.logger).Log("msg", "tenant deletion marker created", "user", userID)
+
+	http.Error(w, "tenant deletion marker created", http.StatusOK)
```
pracucci, line 54: Looks a bit weird using `http.Error()` to report a success. Maybe we could just return a 200 with empty response body (all in all the message doesn't bring any value) or, if we wanna keep the message, create an `util.WriteTextResponse()` similar to the already existing `util.WriteJSONResponse()`?

pracucci (approved):
Thanks for addressing my comments! LGTM. I left few more comments, mostly nits, but please take a look before merging.

pkg/storage/tsdb/tenant_deletion_mark.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"encoding/json"
+	"path"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+const TenantDeletionMarkFilename = "marker/tenant-deletion-mark.json"
```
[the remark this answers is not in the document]
pstibrany: >  I don't remember if it's safe with regards of filesystem-based bucket.

If it's not, I'd argue that filesystem-bucket is broken and should be fixed. Since we want to use `markers` prefix, we need to use some separator, and we already use `/` elsewhere. (E.g. `UserBucketClient` uses it in its `fullName` method)

Also renamed "filename" to path.

pkg/api/api.go
```diff
@@ -231,6 +231,11 @@ func (a *API) RegisterPurger(store *purger.DeleteStore, deleteRequestCancelPerio
 	a.RegisterRoute(a.cfg.LegacyHTTPPrefix+"/api/v1/admin/tsdb/cancel_delete_request", http.HandlerFunc(deleteRequestHandler.CancelDeleteRequestHandler), true, "PUT", "POST")
 }
 
+func (a *API) RegisterBlocksPurger(api *purger.BlocksPurgerAPI) {
+	a.RegisterRoute("/purger/delete_tenant", http.HandlerFunc(api.DeleteTenant), true, "POST")
+	a.RegisterRoute("/purger/delete_tenant_status", http.HandlerFunc(api.DeleteTenantStatus), true, "GET")
```
jtlisi, line 236: **Thought:** It's not an official convention, however, requests that are scoped to a particular tenant universally have a `/api/v1` prefix. The only exception I know of is the Alertmanager UI which is a bit of an edge case. It might be worth following/codifying that convention in this PR. However, I can see an argument to be made that this is a bit different since it eliminates the tenant from existence.

```suggestion
	a.RegisterRoute("/api/v1/delete_tenant", http.HandlerFunc(api.DeleteTenant), true, "POST")
	a.RegisterRoute("/api/v1/delete_tenant_status", http.HandlerFunc(api.DeleteTenantStatus), true, "GET")
```

pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -0,0 +1,120 @@
...
+	level.Info(api.logger).Log("msg", "tenant deletion marker created", "user", userID)
+
+	w.WriteHeader(http.StatusOK)
+}
+
+type DeleteTenantStatusResponse struct {
+	TenantID                  string `json:"tenant_id"`
+	BlocksDeleted             bool   `json:"blocks_deleted"`
+	RuleGroupsDeleted         bool   `json:"rule_groups_deleted"`
+	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted"`
```
jtlisi, line 61:
```
	RuleGroupsDeleted         bool   `json:"rule_groups_deleted"`
	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted"`
```

I don't think these two fields are used anywhere?

jtlisi (approved):
LGTM, modulo a few non-blocking nits

pkg/api/api.go
pstibrany: I am also conflicted about this. On one hand, this is part of API, and we will keep it stable once development has finished.

On the other hand, it's part of "admin" API (eg. like /shutdown), and don't want to send wrong signal by putting it under `/api/v1` prefix. Someone may accidentally expose entire `/api/v1/*` to the users, which in case of `delete_tenant` would be pretty bad.

I'd suggest to keep it as is for now, we can always decide on supporting different path in the future.

pkg/chunk/purger/blocks_purger_api.go
pstibrany: You're right. For now they are just an example. More PRs are coming with support for deleting tenant's ruler and alertmanager data (config, state), and then they will be used as well (perhaps in a changed form).

pkg/ingester/ingester_v2.go
```diff
@@ -259,8 +271,12 @@ func (u *userTSDB) setLastUpdate(t time.Time) {
 	u.lastUpdate.Store(t.Unix())
 }
 
-// Reports tsdbIdle if TSDB can be closed, or some other tsdbIdleCheckResult.
-func (u *userTSDB) canIdleClose(idleTimeout time.Duration) (tsdbIdleCheckResult, error) {
+// Reports tsdbIdle if TSDB can be closed, or some other tsdbCloseCheckResult.
```
pracucci, line 274: > Reports tsdbIdle if TSDB can be closed

Can also be closed on `tsdbTenantMarkedForDeletion`. Maybe we should just generalise the comment and say that "Checks whether the TSDN can be closed.".

pkg/ingester/ingester_v2.go
```diff
@@ -1463,6 +1480,27 @@ func (i *Ingester) shipBlocks(ctx context.Context) {
...
+
+		if time.Since(time.Unix(userDB.lastDeletionMarkCheck.Load(), 0)) > cortex_tsdb.DeletionMarkCheckInterval {
+			// Even if check fails with error, we don't want to repeat it too often.
+			userDB.lastDeletionMarkCheck.Store(time.Now().Unix())
+
+			deletionMarkExists, err := cortex_tsdb.TenantDeletionMarkExists(ctx, i.TSDBState.bucket, userID)
+			if err != nil {
+				// If we cannot check for deletion mark, we continue anyway, even though in production shipper will likely fail too.
+				// This however simplifies unit tests, where tenant deletion check is enabled by default, but tests don't setup bucket.
+				level.Warn(util.Logger).Log("msg", "failed to check for tenant deletion mark before shipping", "user", userID, "err", err)
```
pracucci, line 1495:
```suggestion
				level.Warn(util.Logger).Log("msg", "failed to check for tenant deletion mark before shipping blocks", "user", userID, "err", err)
```

pkg/ingester/ingester_v2.go
```diff
@@ -1620,11 +1661,16 @@ func (i *Ingester) closeAndDeleteUserTSDBIfIdle(userID string) tsdbIdleCheckResu
...
 	// And delete local data.
 	if err := os.RemoveAll(dir); err != nil {
-		level.Error(util.Logger).Log("msg", "failed to delete idle TSDB", "user", userID, "err", err)
+		level.Error(util.Logger).Log("msg", "failed to delete local TSDB", "user", userID, "err", err)
 		return tsdbDataRemovalFailed
 	}
 
-	level.Info(util.Logger).Log("msg", "deleted data for idle TSDB", "user", userID, "dir", dir)
+	if tenantDeleted {
+		level.Info(util.Logger).Log("msg", "deleted local TSDB, user marked for deletion", "user", userID, "dir")
```
pracucci, line 1669: The log messages ends with `"dir"` but the dif value is missing.

```suggestion
		level.Info(util.Logger).Log("msg", "deleted local TSDB, user marked for deletion", "user", userID, "dir", dir)
```

pracucci (approved):
Thanks for addressing my feedback. LGTM! I left few final nits, but no need for having me re-review it again. Thanks! 👏

PR #3552, opened 2020-12-01
Close old TSDB faster.

pstibrany:
**What this PR does**: When opening existing TSDB, set last update time of TSDB to oldest timestamp from the head. This will make ingester to flush and close it sooner (unless more data is pushed to TSDB).

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci (approved):
LGTM

PR #3553, opened 2020-12-01
Define bucket index struct

pracucci:
**What this PR does**:
This is the first PR to introduce the bucket index, as described in this [design doc](https://docs.google.com/document/d/1XCUc7Z6rWqETNZpnVvG05Axzbd--Rm_smDeknTJre2M/edit). In this PR I'm just defining the structs and using them in the querier. The logic hasn't changed (unless bugs).

I've also upgraded Thanos (minor changes) as preliminary step for a follow-up PR.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/storage/tsdb/bucketindex/index.go
```diff
@@ -0,0 +1,195 @@
...
+// Index contains all known blocks and markers of a tenant.
+type Index struct {
+	// Version of the index format.
+	Version int `json:"version"`
+
+	// List of complete blocks (partial blocks are excluded from the index).
+	Blocks []*Block `json:"blocks"`
+
+	// List of block deletion marks.
+	BlockDeletionMarks []*BlockDeletionMark `json:"block_deletion_marks"`
```
pracucci, line 35: I explicitly called it "Block deletion mark" to avoid misunderstandings with other deletion marks we may have in the system (eg. "Tenant deletion mark").

pkg/querier/blocks_store_queryable.go
```diff
@@ -68,7 +68,7 @@ type BlocksFinder interface {
 
 	// GetBlocks returns known blocks for userID containing samples within the range minT
 	// and maxT (milliseconds, both included). Returned blocks are sorted by MaxTime descending.
-	GetBlocks(userID string, minT, maxT int64) ([]*BlockMeta, map[ulid.ULID]*metadata.DeletionMark, error)
+	GetBlocks(ctx context.Context, userID string, minT, maxT int64) (bucketindex.Blocks, map[ulid.ULID]*bucketindex.BlockDeletionMark, error)
```
pracucci, line 71: I've introduced `context.Context`, even if not used yet, as preliminary step for a follow-up PR.

pkg/querier/blocks_store_queryable.go
```diff
@@ -170,7 +170,6 @@ func NewBlocksStoreQueryableFromConfig(querierCfg Config, gatewayCfg storegatewa
 		TenantsConcurrency:       storageCfg.BucketStore.TenantSyncConcurrency,
 		MetasConcurrency:         storageCfg.BucketStore.MetaSyncConcurrency,
 		CacheDir:                 storageCfg.BucketStore.SyncDir,
-		ConsistencyDelay:         storageCfg.BucketStore.ConsistencyDelay,
```
pracucci, line 173: Was unused. Removed it.

pkg/storage/tsdb/bucketindex/index.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// MinTime and MaxTime specify the time range all samples in the block are in (millis precision).
+	MinTime int64 `json:"min_time"`
+	MaxTime int64 `json:"max_time"`
+
+	// SegmentsFormat and SegmentsNum stores the format and number of chunks segments
+	// in the block, if they match a known pattern. We don't store the full segments
+	// files list in order to keep the index small. SegmentsFormat is empty if segments
+	// are unknown or don't match a known format.
+	SegmentsFormat string `json:"segments_format"`
```
pstibrany, line 55: I'd suggest to omit empty segments_format and segments_num fields.

pkg/storage/tsdb/bucketindex/index.go
```diff
@@ -0,0 +1,195 @@
...
+	cortex_tsdb "github.com/cortexproject/cortex/pkg/storage/tsdb"
+	"github.com/cortexproject/cortex/pkg/util"
+)
+
+const (
+	IndexFilename = "bucket-index.json"
+	IndexVersion1 = 1
+
+	SegmentsFormatUnknown       = ""
+	SegmentsFormat1Based6Digits = "1b6d"
```
pstibrany, line 23: Can we add a comment on what this format is? This looks quite cryptic without explanation.

pkg/storage/tsdb/bucketindex/index.go
```diff
@@ -0,0 +1,195 @@
...
+			},
+			SegmentFiles: m.thanosMetaSegmentFiles(),
+		},
+	}
+}
+
+func (m *Block) thanosMetaSegmentFiles() (files []string) {
+	if m.SegmentsFormat == SegmentsFormat1Based6Digits {
+		for i := 1; i <= m.SegmentsNum; i++ {
+			files = append(files, fmt.Sprintf("%06d", i))
```
pstibrany, line 91: Nice trick with named return value initialized to `nil`.

pstibrany (approved):
Nice work!

pkg/storage/tsdb/bucketindex/index.go
pracucci: Good point!

pkg/storage/tsdb/bucketindex/index.go
pracucci: Sure

PR #3558, opened 2020-12-02
Tenant deletion: compactor changes

pstibrany:
**What this PR does**: This PR builds on #3549. In this PR, compactor component checks for tenant deletion marker, and 1) it will skip such users when doing compaction, and 2) its blocks cleaner component will actually remove all blocks for such users.

~Draft until #3549 is merged.~ Rebased on top of master.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -130,14 +130,14 @@ func (c *BlocksCleaner) cleanUsers(ctx context.Context) error {
 	allUsers := append(users, deleted...)
 	return concurrency.ForEachUser(ctx, allUsers, c.cfg.CleanupConcurrency, func(ctx context.Context, userID string) error {
 		if isDeleted[userID] {
-			return errors.Wrapf(c.cleanUserCompletely(ctx, userID), "failed to delete blocks for user marked for deletion: %s", userID)
+			return errors.Wrapf(c.deleteUser(ctx, userID), "failed to delete blocks for user marked for deletion: %s", userID)
 		}
 		return errors.Wrapf(c.cleanUser(ctx, userID), "failed to delete blocks for user: %s", userID)
 	})
 }
 
 // Remove all blocks for user marked for deletion.
-func (c *BlocksCleaner) cleanUserCompletely(ctx context.Context, userID string) error {
+func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
 	userLogger := util.WithUserID(userID, c.logger)
 	userBucket := bucket.NewUserBucketClient(userID, c.bucketClient)
 
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -100,31 +100,84 @@ func (c *BlocksCleaner) ticker(ctx context.Context) error {
 }
 
 func (c *BlocksCleaner) runCleanup(ctx context.Context) {
-	level.Info(c.logger).Log("msg", "started hard deletion of blocks marked for deletion")
+	level.Info(c.logger).Log("msg", "started hard deletion of blocks marked for deletion, and blocks for tenants marked for deletion")
 	c.runsStarted.Inc()
 
 	if err := c.cleanUsers(ctx); err == nil {
-		level.Info(c.logger).Log("msg", "successfully completed hard deletion of blocks marked for deletion")
+		level.Info(c.logger).Log("msg", "successfully completed hard deletion of blocks marked for deletion, and blocks for tenants marked for deletion")
 		c.runsCompleted.Inc()
 		c.runsLastSuccess.SetToCurrentTime()
 	} else if errors.Is(err, context.Canceled) {
-		level.Info(c.logger).Log("msg", "canceled hard deletion of blocks marked for deletion", "err", err)
+		level.Info(c.logger).Log("msg", "canceled hard deletion of blocks marked for deletion, and blocks for tenants marked for deletion", "err", err)
 		return
 	} else {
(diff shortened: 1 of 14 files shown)


pkg/compactor/blocks_cleaner.go
```diff
@@ -100,31 +100,84 @@ func (c *BlocksCleaner) ticker(ctx context.Context) error {
...
+	return concurrency.ForEachUser(ctx, allUsers, c.cfg.CleanupConcurrency, func(ctx context.Context, userID string) error {
+		if isDeleted[userID] {
+			return errors.Wrapf(c.cleanUserCompletely(ctx, userID), "failed to delete blocks for user marked for deletion: %s", userID)
+		}
+		return errors.Wrapf(c.cleanUser(ctx, userID), "failed to delete blocks for user: %s", userID)
+	})
+}
+
+// Remove all blocks for user marked for deletion.
+func (c *BlocksCleaner) cleanUserCompletely(ctx context.Context, userID string) error {
```
pracucci, line 140: `deleteUser()` may be more clear

pkg/compactor/blocks_cleaner.go
```diff
@@ -100,31 +100,84 @@ func (c *BlocksCleaner) ticker(ctx context.Context) error {
...
+	userLogger := util.WithUserID(userID, c.logger)
+	userBucket := bucket.NewUserBucketClient(userID, c.bucketClient)
+
+	level.Info(userLogger).Log("msg", "deleting blocks for user marked for deletion")
+
+	var deleted, failed int
+	err := userBucket.Iter(ctx, "", func(name string) error {
```
pracucci, line 147: Here we delete only blocks, but the users scanner will continue to report this user as far as there's an object in the user prefix (eg. markers/, debug/, bucket-index.json). How should that work?

_Not something we necessarily have to address in this PR, but we need a plan for that._

pracucci (approved):
Good job! LGTM. I left few comments, but no blocker 👏

pkg/storage/tsdb/users_scanner.go
```diff
@@ -29,25 +29,47 @@ func NewUsersScanner(bucketClient objstore.Bucket, isOwned func(userID string) (
...
 }
 
-// ScanUsers returns a fresh list of users found in the storage. If sharding is enabled,
-// the returned list contains only the users owned by this instance.
-func (s *UsersScanner) ScanUsers(ctx context.Context) ([]string, error) {
+// ScanUsers returns a fresh list of users found in the storage, that are not marked for deletion,
+// and list of users marked for deletion.
+//
+// If sharding is enabled, returned lists contains only the users owned by this instance.
+func (s *UsersScanner) ScanUsers(ctx context.Context) ([]string, []string, error) {
```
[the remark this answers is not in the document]
pstibrany: I considered that, and in this case I think named return values would be indeed useful.

pkg/compactor/blocks_cleaner.go
pstibrany: My plan is this: when `deleteUser` is finished, it will update deletion mark with timestamp when it has finished.

When `deleteUser` is called next time, there are two cases:
- if there are more blocks to delete, `deleteUser` will delete them, and update tenant deletion mark again with "finished" time.
- if no more blocks are to be deleted, `deleteUser` will check "finished" time in the marker, and after some time (say a week), it will do remaining cleanup work: deleting objects with the "user/" prefix completely (with marker file being the last one).

pkg/storage/tsdb/users_scanner.go
```diff
@@ -29,25 +29,47 @@ func NewUsersScanner(bucketClient objstore.Bucket, isOwned func(userID string) (
...
 		} else if !owned {
-			return nil
+			users = append(users[:ix], users[ix+1:]...)
+			continue
 		}
 
-		users = append(users, userID)
-		return nil
-	})
+		deletionMarkExists, err := TenantDeletionMarkExists(ctx, s.bucketClient, userID)
```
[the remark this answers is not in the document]
pstibrany: Added, by re-using metafile config options for Exists.

pkg/compactor/blocks_cleaner.go
pracucci: > it will update deletion mark with timestamp when it has finished

Uhm. This makes the deletion mark no more immutable. Let's talk about it offline. Not to be addressed in this PR.

pkg/storage/tsdb/caching_bucket.go
```diff
@@ -169,6 +171,10 @@ func isTenantsDir(name string) bool {
 	return name == ""
 }
 
+func isTenantDeletionMark(name string) bool {
```
pracucci, line 174: Any reason why not adding it to `isMetaFile()` like we did for the block deletion mark?

pracucci (approved):
Still LGTM. Left a nit, but feel free to merge it without having me re-reviewing it.

pkg/compactor/blocks_cleaner.go
pstibrany: That's correct. I don't think it's a problem -- other components just check for existence. Compactor will be the only component actually reading the tenant deletion mark.

If you think this is still an issue, we could also write "finished" timestamp into a separate file.

pkg/storage/tsdb/caching_bucket.go
pstibrany: No special reason, I can merge it.

pkg/compactor/blocks_cleaner.go
pracucci: > Compactor will be the only component actually reading the tenant deletion mark.

True. You're right, probably not an issue. I was thinking about adding it to the bucket index too, but we actually just need to store the "existence" in the index.

PR #3560, opened 2020-12-02
Add signature v2 support for S3 chunks client

simonswine:
**What this PR does**:

* This allows to configure the chunks S3 client to use V2 signatures for S3 authentication (is also used for the alertmanager and ruler store).

* This follows the interface of #3540 for the blocks client

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
   * `cortex_compactor_tenants_processing_succeeded`
   * `cortex_compactor_tenants_processing_failed`
 * [ENHANCEMENT] Added new experimental API endpoints: `POST /purger/delete_tenant` and `GET /purger/delete_tenant_status` for deleting all tenant data. Only works with blocks storage. #3549
+* [ENHANCEMENT] Chunks storage: add option to use V2 signatures for S3 authentication. #3560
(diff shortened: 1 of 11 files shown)


pkg/chunk/grpc/grpc.pb.go
```diff
@@ -7,17 +7,18 @@ import (
 	bytes "bytes"
 	context "context"
 	fmt "fmt"
+	io "io"
+	math "math"
+	math_bits "math/bits"
+	reflect "reflect"
+	strings "strings"
+
```
jtlisi, line 15: **nit:** Not sure you meant to commit this. This looks like a autoformatter hitting a generated file.

jtlisi (approved):
LGTM

There is one nit that needs to be fixed before this can be merged.

pkg/chunk/grpc/grpc.pb.go
simonswine: Yes that was a too aggressive `go-imports` run

pstibrany (approved):
LGTM

pracucci (commented):
Post-merge comment: good job! 👏

PR #3554, opened 2020-12-01
Proposal: introduce bucket index for the blocks storage

pracucci:
**What this PR does**:
As previously shared in the CNCF Slack channel, I'm proposing to add a per-tenant bucket index to improve scalability of Cortex blocks storage when running a cluster with thousands of tenants. This PR is contains the design doc originally shared via [google doc](https://docs.google.com/document/d/1XCUc7Z6rWqETNZpnVvG05Axzbd--Rm_smDeknTJre2M/edit).

**Which issue(s) this PR fixes**:
N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3561, opened 2020-12-02
Write block deletion marks in the global location too

pracucci:
**What this PR does**:
This is another preliminary step required to implement the bucket index, as designed in https://github.com/cortexproject/cortex/pull/3554. In this PR:

- Introduce bucket index reader and writer (but not plugged in yet)
- Introduce a bucket client wrapper used to store block deletion marks in the per-tenant global location too
- Plug the bucket client wrapper in the compactor (which is the component deleting blocks)

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/storage/tsdb/bucketindex/markers.go
+++ b/pkg/storage/tsdb/bucketindex/markers.go
@@ -1,0 +1,43 @@
+package bucketindex
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/oklog/ulid"
+	"github.com/thanos-io/thanos/pkg/block"
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+)
+
+const (
+	MarkersPathname = "markers"
+)
+
+// BlockDeletionMarkFilepath returns the path, relative to the tenant's bucket location,
+// of a block deletion mark in the bucket markers location.
+func BlockDeletionMarkFilepath(blockID ulid.ULID) string {
+	return fmt.Sprintf("%s/%s-%s", MarkersPathname, blockID.String(), metadata.DeletionMarkFilename)
+}
+
+// BlockDeletionMarkFilename returns the filename to use when storing a block deletion mark
+// in the bucket markers location.
+func BlockDeletionMarkFilename(blockID ulid.ULID) string {
+	return fmt.Sprintf("%s-%s", blockID.String(), metadata.DeletionMarkFilename)
+}
+
+// IsBlockDeletionMarkFilename returns whether the input filename matches the expected pattern
+// of block deletion markers stored in the markers location.
+func IsBlockDeletionMarkFilename(name string) (ulid.ULID, bool) {
+	parts := strings.SplitN(name, "-", 2)
+	if len(parts) != 2 {
+		return ulid.ULID{}, false
+	}
+
+	// Ensure the 2nd part matches the block deletion mark filename.
+	if parts[1] != metadata.DeletionMarkFilename {
+		return ulid.ULID{}, false
+	}
+
+	// Ensure the 1st part is a valid block ID.
+	return block.IsBlockDir(parts[0])
+}
--- a/pkg/storage/tsdb/bucketindex/markers.go
+++ b/pkg/storage/tsdb/bucketindex/markers.go
@@ -19,12 +19,6 @@ func BlockDeletionMarkFilepath(blockID ulid.ULID) string {
 	return fmt.Sprintf("%s/%s-%s", MarkersPathname, blockID.String(), metadata.DeletionMarkFilename)
 }
 
-// BlockDeletionMarkFilename returns the filename to use when storing a block deletion mark
-// in the bucket markers location.
-func BlockDeletionMarkFilename(blockID ulid.ULID) string {
-	return fmt.Sprintf("%s-%s", blockID.String(), metadata.DeletionMarkFilename)
-}
-
 // IsBlockDeletionMarkFilename returns whether the input filename matches the expected pattern
 // of block deletion markers stored in the markers location.
 func IsBlockDeletionMarkFilename(name string) (ulid.ULID, bool) {
--- a/pkg/storage/tsdb/bucketindex/markers.go
+++ b/pkg/storage/tsdb/bucketindex/markers.go
@@ -2,10 +2,10 @@ package bucketindex
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/oklog/ulid"
-	"github.com/thanos-io/thanos/pkg/block"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 )
 
@@ -39,5 +39,6 @@ func IsBlockDeletionMarkFilename(name string) (ulid.ULID, bool) {
 	}
 
 	// Ensure the 1st part is a valid block ID.
-	return block.IsBlockDir(parts[0])
+	id, err := ulid.Parse(filepath.Base(parts[0]))
+	return id, err == nil
 }
--- a/pkg/storage/tsdb/bucketindex/markers_bucket_client.go
+++ b/pkg/storage/tsdb/bucketindex/markers_bucket_client.go
@@ -15,14 +15,14 @@ import (
 )
 
 type globalMarkersBucket struct {
-	objstore.InstrumentedBucket
+	objstore.Bucket
 }
 
 // BucketWithGlobalMarkers wraps the input bucket into a bucket which also keeps track of markers
 // in the global markers location.
-func BucketWithGlobalMarkers(b objstore.InstrumentedBucket) objstore.InstrumentedBucket {
+func BucketWithGlobalMarkers(b objstore.Bucket) objstore.Bucket {
 	return &globalMarkersBucket{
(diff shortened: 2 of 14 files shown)


pkg/storage/tsdb/bucketindex/markers_bucket_client.go
```diff
@@ -0,0 +1,85 @@
...
+	"path"
+	"strings"
+
+	"github.com/oklog/ulid"
+	"github.com/thanos-io/thanos/pkg/block"
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+type globalMarkersBucket struct {
```
pstibrany, line 17: It's not clear what "global" refers to. Comment would help. (Perhaps "userMarkersBucket" name too?)

pkg/storage/tsdb/bucketindex/markers_bucket_client.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// Parse the block ID in the path. If there's not block ID, then it's not the per-block
+	// deletion mark.
+	parts := strings.Split(name, "/")
+	if len(parts) < 2 {
+		return ulid.ULID{}, false
+	}
+
+	return block.IsBlockDir(parts[len(parts)-2])
```
pstibrany, line 84:
```suggestion
	return block.IsBlockDir(path.Dir(name))
```

and avoid split.

pkg/storage/tsdb/bucketindex/reader.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, errors.Wrap(err, "read bucket index")
+	}
+	defer runutil.CloseWithLogOnErr(logger, reader, "close bucket index reader")
+
+	// Read all the content.
+	gzipReader, err := gzip.NewReader(reader)
+	if err != nil {
+		return nil, ErrIndexCorrupted
+	}
+	defer runutil.CloseWithLogOnErr(logger, gzipReader, "close bucket index gzip reader")
```
pstibrany, line 41: Comment: No need to close gzip reader, since we're ignoring its error anyway. It only checks if final checksum was correct. We would already detect error (eg. json parsing) earlier.

pkg/storage/tsdb/bucketindex/reader.go
```diff
@@ -0,0 +1,55 @@
...
+
+	content, err := ioutil.ReadAll(gzipReader)
```
pstibrany, line 43: Nit: We can skip full reading to memory, and use `json.NewDecoder()` instead for parsing the `gzipReader` directly. (Given that index is expected to be small, even when uncompressed, this isn't a big deal).

pkg/storage/tsdb/bucketindex/writer.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	return out, nil
+}
+
+func (w *Writer) generateBlockIndexEntry(ctx context.Context, id ulid.ULID) (*Block, error) {
+	metaFile := path.Join(id.String(), block.MetaFilename)
+
+	// Get the block's meta.json file.
+	r, err := w.bkt.Get(ctx, metaFile)
```
pstibrany, line 159: comment: I'm bit surprised to see that we don't use some Thanos function for reading meta.json file. But I cannot find such public function in Thanos. :confused:

pkg/storage/tsdb/bucketindex/writer.go
```diff
@@ -0,0 +1,273 @@
...
+		out = append(out, m)
+	}
+
+	return out, nil
+}
+
+func (w *Writer) generateBlockDeletionMarkIndexEntry(ctx context.Context, id ulid.ULID) (*BlockDeletionMark, error) {
+	markFile := path.Join(id.String(), metadata.DeletionMarkFilename)
+
+	// Get the block's deletion mark file.
```
pstibrany, line 247: We can replace marker reading code in this method with:

```
	m := metadata.DeletionMark{}
	err := metadata.ReadMarker(ctx, w.logger, w.bkt, id.String(), &m)
```

(`UserBucketClient` already implements required `InstrumentedBucket`)

pstibrany (approved):
Nice job! I have mostly minor nits and comments, nothing blocking.

Unrelated to this PR: have you considered naming `bucketindex.Block` to `BlockMeta`.

pracucci:
> have you considered naming bucketindex.Block to BlockMeta

Yes. I would prefer to not call it "BlockMeta" to avoid misunderstanding. We don't index the BlockMeta, but some information about the Block and I would prefer not couple it with the `meta.json` (eg. we may also index block information which is not in the `meta.json`).

pkg/storage/tsdb/bucketindex/markers_bucket_client.go
pracucci: It's not the same thing. The object prefix could be `/user-1/<block-id>/deletion-mark.json`. We need to pick the 2nd last segment, not the whole `path.Dir()`.

pkg/storage/tsdb/bucketindex/reader.go
pracucci: I think it's more future proof keeping the call to `.Close()`.

pkg/storage/tsdb/bucketindex/reader.go
pracucci: Sure, makes sense!

pkg/storage/tsdb/bucketindex/markers_bucket_client.go
pstibrany: > We need to pick the 2nd last segment, not the whole path.Dir().

`block.IsBlockDir` already covers that.

pkg/storage/tsdb/bucketindex/writer.go
pracucci: There's `metadata.Read()` and `metadata.ReadMarker()` from Thanos but:
- with `metadata.Read()` we would have no way to distinguish a corrupted meta from a network error
- `metadata.ReadMarker()` requires an `InstrumentedBucket` and it's quite annoying to propagate till here (I have tried). I can try again, but I would prefer to keep it in a separate PR.

pkg/storage/tsdb/bucketindex/writer.go
pracucci: Very good point. Done, thanks!

pkg/storage/tsdb/bucketindex/writer.go
pracucci: We could check if the `error` returned by `metadata.Read()` is a `json.SyntaxError` but unfortunately that doesn't work because the returned error is a `MultiErr`. We would need this https://github.com/prometheus/prometheus/pull/8133 in.

pracucci:
Thanks a lot @pstibrany for your valuable feedback! I addressed most comments and replied to few other ones. Could you take another look, please?

pkg/storage/tsdb/bucketindex/markers.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	MarkersPathname = "markers"
+)
+
+// BlockDeletionMarkFilepath returns the path, relative to the tenant's bucket location,
+// of a block deletion mark in the bucket markers location.
+func BlockDeletionMarkFilepath(blockID ulid.ULID) string {
+	return fmt.Sprintf("%s/%s-%s", MarkersPathname, blockID.String(), metadata.DeletionMarkFilename)
```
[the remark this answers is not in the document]
pracucci: I removed `BlockDeletionMarkFilename()`. Was not used right now.

pstibrany (approved):
LGTM, nice work.

pkg/storage/tsdb/bucketindex/markers_bucket_client.go
pracucci: I missed that, sorry. Fixed!

AllenzhLi:
In bucket_index.json aleady have block ids which is marked delete,
Why save a delete file for every block separately?

pracucci:
> In bucket_index.json aleady have block ids which is marked delete,
> Why save a delete file for every block separately?

@AllenzhLi Without a per-tenant global `markers/` location we wouldn't have an easy way to find out which blocks have been marked for deletion. Without that, we would have to check the deletion mark for every single block with 1 GET object operation per block (which is what we do right now). With the global location, we can just issue 1 LIST objects API call to find all blocks marked for deletion.

AllenzhLi:
Thank you for your reply.

I have read the design doc, in the doc
image
In the bucket_index.json have the list of block deletion marks, why not use it?

pstibrany:
> In the bucket_index.json have the list of block deletion marks, why not use it?

Bucket index JSON is created based on found blocks and deletion marks. Querier and store-gateway only use index (and ignore blocks marked for deletion). During compaction, compactor will only create deletion marks, but not update the index. Index update is a separate loop, that looks for deletion marks.

AllenzhLi:
Understood, thanks.

PR #3562, opened 2020-12-03
Add opstrace to adopters list

gouthamve:
@jgehrcke I'll soon share the screenshot of how the website will look!

PR #3563, opened 2020-12-03
Revert "Add opstrace to adopters list"

gouthamve:
Reverts cortexproject/cortex#3562

We didn't have a public LGTM from any Opstrace employees. We had one on Twitter here https://twitter.com/gehrcke/status/1334535786671304704 but we should ideally have an LGTM on Github too specially as we're adding their Logo to the website.


pstibrany:
I think we have one now in #3562.

PR #3574, opened 2020-12-08
[Proposal] Scalable Alertmanager

gotjosh:
**What this PR does**:

Introduces a series of changes for scaling the Alertmanager. Given the proposal is quite lengthy, I have tried to keep it to the point on most parts.

Please do let me know if there's an area that requires expanding.

**Checklist**
- N/A Tests updated
- [x] Documentation added
- N/A `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`


docs/proposals/scalable-alertmanager.md
```diff
@@ -0,0 +1,78 @@
...
+To implement the sharding strategy we need a component in charge of handling incoming alerts and API requests, distributing it to the corresponding shard. 
+
+The idea here is to have an “Alertmanager Distributor” as a first stop in the reception of alerts. Once alerts are received, the component is in charge of validating the alerts against the limits. Validated alerts are then sent to multiple managers in parallel.
+
+The individual pieces of this component (sharding, limits) cannot be optional - the optional part of it is where we decide to run it.
+
+We can either run it as a separate service or embed it. **I propose we simply embed it**. At its core it’ll be simpler to operate. With future work making it possible to run as a separate service so that operators can scale when/if needed.
+
+
+## Conclusion
```
gotjosh, line 68: One thing I have not talked about is sequencing - is it worth mentioning that now? My plan is to leave the current alertmanager struct untouched and create a new one can be specified via a flag. The PRs can be slices of the sections described. Keeping in mind that I can break those down as I deemed fit to prioritise simpler and less risky reviews.

pracucci (approved):
We discussed offline and LGTM, thanks! I just pushed a commit to fix the image location.

docs/proposals/scalable-alertmanager.md
```diff
@@ -0,0 +1,78 @@
...
+
+Under the assumption we implement the options proposed above, our architecture looks like this:
+
+![Scalable Alertmanager Architecture](/images/proposals/scalable-am.png)
+
+**POST /api/v1/alerts (from the ruler) can go to any Alertmanager replica.** The AM distributor uses the ring to write alerts to a quorum of AM managers (reusing the existing code). We continue to use the same in-memory data structure from the upstream Alertmanager to save alerts and notify other pieces
```
jtlisi, line 74: **question:** In this architecture the ruler will still use HTTP to forward alerts to the Alertmanager. In order to avoid multiple deliveries will the Ruler only be configured with a single address? If so does this assume the user will need to place a loadbalancer between the `ruler` and the `am-distributor` to achieve an equal distribution of traffic and ensure availability?

gotjosh: I think the short answer is yes. 

Just like today, you need to use a headless service (or include all the alertmanager URLs manually) to avoid load balancing traffic between Alertmanagers. You'll need to ensure you have a way to load balance traffic between the ruler <> Alertmanager.

jtlisi: 👍

jtlisi (approved):
LGTM

docs/proposals/scalable-alertmanager.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Design
+
+This is a big design, and as such we have divided the problem up into 4 areas: Routing & Sharding, Persistence & State, Replication & Consistency, and Architecture. The options and solutions I propose are divided along these lines to aid clarity.
+
+### Routing & Sharding
+
+To achieve horizontal scalability, we need to distribute the workload among replicas of the service. We need to choose an appropriate field to use to distribute the workload. The field must be present on all the API requests to the Alertmanager service.
+
+**We propose the sharding on Tenant ID**. The simplicity of this implementation, would allow us to get up and running relatively quickly whilst helping us validate assumptions. We intend to use the existing ring code to manage this. Other options such as tenant ID + receiver or Tenant ID + route are relatively complex as distributor components (in this case the Ruler) would need to be aware of Alertmanager configuration.
```
ranton256, line 37: I think this con is fine for the time being If the current design can support ~2000 average tenants, it could with a little handholding support a tenant 2000 times normal on by fiddling with sharding overrides, so I don't think this limitation is reason enough to not choose the simple option 1, shard on tenant id.

docs/proposals/scalable-alertmanager.md
ranton256: I'd be interested in your ideas for sequencing, but seems like it could be separate from this PR.

ranton256 (approved):
The proposal LGTM. Thanks for doing this.

docs/proposals/scalable-alertmanager.md
gotjosh: I'm prepping the first PR, I'll tag you in it so that you can get an idea of what my plan is.

pracucci:
_Going to merge it. As usual, we'll address any post-merge comment._

PR #3576, opened 2020-12-09
Removed CircleCI support

pracucci:
**What this PR does**:
On October 30th I've disabled the CircleCI webhook because our CI was already migrated to GitHub actions. We already released Cortex 1.5.0 relying on GitHub actions only so, in this PR, I would propose to decommission CircleCI support at all.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


README.md
```diff
@@ -1,10 +1,9 @@
 <p align="center"><img src="images/logo.png" alt="Cortex Logo"></p>
 
-[![Circle CI](https://circleci.com/gh/cortexproject/cortex/tree/master.svg?style=shield)](https://circleci.com/gh/cortexproject/cortex/tree/master)
+[![CI](https://github.com/cortexproject/cortex/workflows/ci/badge.svg)](https://github.com/cortexproject/cortex/actions)
 [![GoDoc](https://godoc.org/github.com/cortexproject/cortex?status.svg)](https://godoc.org/github.com/cortexproject/cortex)
 <a href="https://goreportcard.com/report/github.com/cortexproject/cortex"><img src="https://goreportcard.com/badge/github.com/cortexproject/cortex" alt="Go Report Card" /></a>
 <a href="https://cloud-native.slack.com/messages/cortex/"><img src="https://img.shields.io/badge/join%20slack-%23cortex-brightgreen.svg" alt="Slack" /></a>
-[![TODOs](https://badgen.net/https/api.tickgit.com/badgen/github.com/cortexproject/cortex)](https://www.tickgit.com/browse?repo=github.com/cortexproject/cortex)
```
pracucci, line 7: Removed because was only referenced here, but not in the website homepage. I personally believe doesn't bring any value but, if there are objections, I can bring it back.

docs/contributing/how-to-update-the-build-image.md
```diff
@@ -10,6 +10,5 @@ The build image currently can only be updated by a Cortex maintainer. If you're
 1. Update `build-image/Docker`
 2. Build the image running `make build-image/.uptodate`
 3. Publish the image to the repository running `docker push quay.io/cortexproject/build-image:TAG` (this can only be done by a maintainer)
-4. Replace the image tag in `.circleci/config.yml` (_there may be multiple references_)
-5. Replace the image tag in `.github/workflows/*` (_there may be multiple references_)
-6. Open a PR and make sure the CI with new build-image passes
+4. Replace the image tag in `.github/workflows/*` (_there may be multiple references_)
```
bboreham, line 13: Are there?  Can we fix that?

bboreham (approved):
LGTM.

I wonder if we can remove the whole `tools/test` script, since it now doesn't do much beyond what `go test ./...` would do.

docs/contributing/how-to-update-the-build-image.md
pracucci: There are and, to my knowledge, they're not easy to fix because can't have default values in GitHub actions yaml definition. I'm not aware of an easy way to fix it.

roidelapluie (approved):

pstibrany (approved):

jtlisi (approved):
LGTM

pracucci:
> I wonder if we can remove the whole tools/test script, since it now doesn't do much beyond what go test ./... would do.

Yes, it should be enough to just run:
`go test -tags netgo -timeout 20m -race -covermode=atomic -count=1 ./cmd/... ./pkg/... ./tools/...`

However, if I run `go test ./pkg/...` ring KV tests for etcd are flaky (and sometimes panic too) while if I run `go test` on each individual package (like `tools/test`) does it works. I spent some time investigating and I haven't understood the root cause yet. I'm going to merge this PR and open a GitHub issue about it, sharing some details.

PR #3578, opened 2020-12-09
Add docker-compose setup using gossip.

pstibrany:
**What this PR does**: This PR copies existing `tsdb-blocks-storage-s3` docker-compose setup, but instead of Consul, it uses gossiped ring for ingesters. Eventually I'd like to use gossip for other components too (compactor, ruler), but for now ingesters' ring is already giving some problematic behaviour that I'd like to explore before switching other components.

**Checklist**
- [NA] Tests updated
- [NA] Documentation added


pracucci (approved):
Very useful, thanks!

PR #3592, opened 2020-12-11
Fix Cortex logo in the website

pracucci:
**What this PR does**:
I finally understood the root cause of the truncated Cortex logo in the published website:
Screenshot 2020-12-11 at 12 19 26

It's caused by the minification. Wasn't able to reproduce it locally because minification was disabled. I've now enabled `--minify` in `make web-serve` too and removing the minification fixes it:
Screenshot 2020-12-11 at 12 20 32

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany (approved):

cyriltovena:
👏

PR #3593, opened 2020-12-11
Add junit reporting to integration tests

simonswine:
**What this PR does**:

This might help getting a better overview over integration tests failing in the Github actions. I sometimes even use the xml format locally to get a quicker idea which tests have failed.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -147,9 +147,18 @@ jobs:
           export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:$IMAGE_TAG"
           export CORTEX_CHECKOUT_DIR="/go/src/github.com/cortexproject/cortex"
           echo "Running integration tests with image: $CORTEX_IMAGE"
-          go test -tags=requires_docker -timeout 1800s -v -count=1 ./integration/...
+          go get -u github.com/jstemmer/go-junit-report@v0.9.1
+          go test -tags=requires_docker -timeout 1800s -v -count=1 ./integration/... 2>&1 | tee >(go-junit-report > test_integration.xml)
         env:
           IMAGE_PREFIX: ${{ secrets.IMAGE_PREFIX }}
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v1.6
+        if: always()
+        with:
+          check_name: Integration Test Results
+          report_individual_runs: true
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          files: 'test_*.xml'
 
   integration-configs-db:
     needs: build
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -147,18 +147,11 @@ jobs:
           export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:$IMAGE_TAG"
           export CORTEX_CHECKOUT_DIR="/go/src/github.com/cortexproject/cortex"
           echo "Running integration tests with image: $CORTEX_IMAGE"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build-deploy.yml
```diff
@@ -147,9 +147,15 @@ jobs:
           export CORTEX_IMAGE="${CORTEX_IMAGE_PREFIX}cortex:$IMAGE_TAG"
           export CORTEX_CHECKOUT_DIR="/go/src/github.com/cortexproject/cortex"
           echo "Running integration tests with image: $CORTEX_IMAGE"
-          go test -tags=requires_docker -timeout 1800s -v -count=1 ./integration/...
+          go get -u github.com/jstemmer/go-junit-report@v0.9.1
+          go test -tags=requires_docker -timeout 1800s -v -count=1 ./integration/... 2>&1 | tee >(go-junit-report > test_integration.xml)
         env:
           IMAGE_PREFIX: ${{ secrets.IMAGE_PREFIX }}
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v1
```
pstibrany, line 155: Bit scary to use random build actions from untrusted sources, and giving them your github token. Are we comfortable with that?

pracucci: I see the concerns about security, but what are alternatives? I took a quick look at official actions and I can't see anything related to test results reporting.

pracucci (commented):
Could you investigate and fix the failure, please? See:
https://github.com/cortexproject/cortex/runs/1537317182?check_suite_focus=true

.github/workflows/test-build-deploy.yml
simonswine: The only alternative I could think of would be to use a golang library instead

pracucci: As a data point, the `GITHUB_TOKEN` [permissions](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) are quite wide, but:

> The token expires when the job is finished.

Also, it's enough to open a PR with a modified workflow to run your custom code authenticating with `GITHUB_TOKEN` (eg. the junit reporter run in this PR's CI workflow). Isn't a security issue as well? I would create my malign action and open PRs around. I guess I'm missing something.

pracucci: > The only alternative I could think of would be to use a golang library instead

I don't think there would be any benefit, unless we review (from a security standpoint) that library. I personally see the potential issue, but to me it's **not a blocker** (see also my comment ☝️ ). I'm personally OK approving it, but I would like to see it working first 😉

pstibrany: > Isn't a security issue as well? I would create my malign action and open PRs around. I guess I'm missing something.

Interesting. If that runs within a context of target project (and not the fork), I think you are onto something. Definitely worth trying.

The fact that token expires quickly helps, but also note that action writers would prefer to have read-only token as well: https://github.community/t/read-only-token-for-ci/17295

In OSS project I’m worried less about this than I would be for commercial one as possible harm is very limited.

pstibrany:
Where can one see if it works as intended? Where to find the report?

simonswine:
While it is working now, I am not yet happy with the PR and close it for now, as otherwise it becomes a bit of a time sink.

Basically currently it groups the output of the integration tests into collapsible groups like that:

scrn-2020-12-14-14-14-07

scrn-2020-12-14-14-14-19

Which makes the output quite readable. But for some reason the streaming of the stdout is only showing up at the very end, when all tests are finished (running it locally it streams the output as the integration tests are executed). I am also worried it might hide important error information in the error cases.

PR #3595, opened 2020-12-11
Propagate whether query stats is enabled from query-frontend down the request path

pracucci:
**What this PR does**:
In the PR https://github.com/cortexproject/cortex/pull/3539 I've introduced query stats metering support, but it required to enable it both in the query-frontend and querier. I didn't like it very much because I think it's error prone and hard to understand for our users so, in this PR, I've introduced the propagation of the `StatsEnabled` flag down the request path.

_Other than existing integration tests, I've manually tested it both with and without query-scheduler._

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Querier: it's no more required to set `-frontend.query-stats-enabled=true` in the querier to enable query statistics logging in the query-frontend. The flag is now required to be configured only in the query-frontend and the information will be propagated to the queriers in the internal request. #3595
```
pstibrany, line 5:
```suggestion
* [CHANGE] Querier: it's not required to set `-frontend.query-stats-enabled=true` in the querier anymore to enable query statistics logging in the query-frontend. The flag is now required to be configured only in the query-frontend and it will be propagated to the queriers. #3595
```

pstibrany (approved):
LGTM, great job!

PR #3598, opened 2020-12-13
Fix max query lookback type

Wing924:
**What this PR does**:
Allow `-querier.max-query-lookback` use `y|w|d` suffix. (ex. `60d`, `1y`)

In #3452, `-store.max-look-back-period` was deprecated, and we should use `-querier.max-query-lookback`.
The old one support `y|w|d` suffix but new one don't support it.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany (approved):
Thank you!

pracucci (approved):
Amazing, thanks!

PR #3601, opened 2020-12-14
Optimised Ring.ShuffleShard() and disabled subring cache in store-gateway, ruler and compactor

pracucci:
**What this PR does**:
We've been reported an high memory utilisation in the store-gateway when shuffle-sharding is enabled and the cluster has a large number of tenants and replicas. This memory utilisation increase is caused by the in-memory subring cache used by the shuffle-sharding, which is used to reduce the CPU utilisation due to the expensive computation of the subring. 

However, store-gateway, ruler and compactor use `Ring.ShuffleShard()` only to check whether a tenant/block/rulegroup belongs to their own shard and this operation is run, for each tenant, only at a regular interval (it doesn't depend on QPS). 

Disabling subring cache is a possible solution (proposed in this PR), but then CPU utilisation is expected to increase. For this reason, in this PR I've also worked to optimise `Ring.ShuffleShard()` paying attention to not increase CPU time for `Ring.Get()` (problem is that it's easy to optimise `ShuffleShard()` but hard to do without impacting `Get()`).

## BenchmarkRing_ShuffleShard_512Tokens (512 tokens per instance)

```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkRing_ShuffleShard_512Tokens-12     1396063       389304        -72.11%

benchmark                                   old allocs     new allocs     delta
BenchmarkRing_ShuffleShard_512Tokens-12     60             59             -1.67%

benchmark                                   old bytes     new bytes     delta
BenchmarkRing_ShuffleShard_512Tokens-12     681775        58226         -91.46%
```

## BenchmarkRing_Get

```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkRing_Get-12                        853           764           -10.43%

benchmark                                   old allocs     new allocs     delta
BenchmarkRing_Get-12                        0              0              +0.00%

benchmark                                   old bytes     new bytes     delta
BenchmarkRing_Get-12                        0             0             +0.00%
```

## BenchmarkRing_ShuffleShard (128 tokens per instance)

```
BenchmarkRing_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_3-12              100674        32002         -68.21%
BenchmarkRing_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_10-12             393978        152904        -61.19%
BenchmarkRing_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_30-12             1316209       639339        -51.43%
BenchmarkRing_ShuffleShard/num_instances_=_50,_num_zones_=_3,_shard_size_=_3-12              117988        48928         -58.53%
BenchmarkRing_ShuffleShard/num_instances_=_50,_num_zones_=_3,_shard_size_=_10-12             525024        183298        -65.09%
BenchmarkRing_ShuffleShard/num_instances_=_50,_num_zones_=_3,_shard_size_=_30-12             1221208       558804        -54.24%
BenchmarkRing_ShuffleShard/num_instances_=_100,_num_zones_=_1,_shard_size_=_3-12             112511        37656         -66.53%
BenchmarkRing_ShuffleShard/num_instances_=_100,_num_zones_=_1,_shard_size_=_10-12            408735        153631        -62.41%
BenchmarkRing_ShuffleShard/num_instances_=_100,_num_zones_=_1,_shard_size_=_30-12            1328229       655980        -50.61%
BenchmarkRing_ShuffleShard/num_instances_=_100,_num_zones_=_3,_shard_size_=_3-12             121937        49367         -59.51%
BenchmarkRing_ShuffleShard/num_instances_=_100,_num_zones_=_3,_shard_size_=_10-12            466278        176984        -62.04%
BenchmarkRing_ShuffleShard/num_instances_=_100,_num_zones_=_3,_shard_size_=_30-12            1265773       562446        -55.57%
BenchmarkRing_ShuffleShard/num_instances_=_1000,_num_zones_=_1,_shard_size_=_3-12            95741         31072         -67.55%
BenchmarkRing_ShuffleShard/num_instances_=_1000,_num_zones_=_1,_shard_size_=_10-12           376408        143209        -61.95%
BenchmarkRing_ShuffleShard/num_instances_=_1000,_num_zones_=_1,_shard_size_=_30-12           1328692       616708        -53.59%
BenchmarkRing_ShuffleShard/num_instances_=_1000,_num_zones_=_3,_shard_size_=_3-12            114967        47096         -59.04%
BenchmarkRing_ShuffleShard/num_instances_=_1000,_num_zones_=_3,_shard_size_=_10-12           441547        176075        -60.12%
BenchmarkRing_ShuffleShard/num_instances_=_1000,_num_zones_=_3,_shard_size_=_30-12           1288778       551126        -57.24%
```

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ring/ring.go
+++ b/pkg/ring/ring.go
@@ -278,7 +278,7 @@ func (r *Ring) loop(ctx context.Context) error {
 		}
 
 		now := time.Now()
-		ringTokens := ringDesc.getTokens()
+		ringTokens := ringDesc.GetTokens()
 		ringTokensByZone := ringDesc.getTokensByZone()
 		ringInstanceByToken := ringDesc.getTokensInfo()
 		ringZones := getZones(ringTokensByZone)
@@ -721,7 +721,7 @@ func (r *Ring) shuffleShard(identifier string, size int, lookbackPeriod time.Dur
 		cfg:              r.cfg,
 		strategy:         r.strategy,
 		ringDesc:         shardDesc,
-		ringTokens:       shardDesc.getTokens(),
+		ringTokens:       shardDesc.GetTokens(),
 		ringTokensByZone: shardTokensByZone,
 		ringZones:        getZones(shardTokensByZone),
 
--- a/pkg/ring/ring.go
+++ b/pkg/ring/ring.go
@@ -113,6 +113,10 @@ type Config struct {
 	ReplicationFactor    int           `yaml:"replication_factor"`
 	ZoneAwarenessEnabled bool          `yaml:"zone_awareness_enabled"`
 	ExtendWrites         bool          `yaml:"extend_writes"`
+
+	// Whether the shuffle-sharding subring cache is disabled. This option is set
+	// internally and never exposed to the user.
+	SubringCacheDisabled bool `yaml:"-"`
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet with a specified prefix
@@ -707,6 +711,10 @@ func (r *Ring) HasInstance(instanceID string) bool {
 }
 
 func (r *Ring) getCachedShuffledSubring(identifier string, size int) *Ring {
+	if r.cfg.SubringCacheDisabled {
+		return nil
+	}
+
 	r.mtx.RLock()
 	defer r.mtx.RUnlock()
 
@@ -731,7 +739,7 @@ func (r *Ring) getCachedShuffledSubring(identifier string, size int) *Ring {
 }
 
 func (r *Ring) setCachedShuffledSubring(identifier string, size int, subring *Ring) {
-	if subring == nil {
+	if subring == nil || r.cfg.SubringCacheDisabled {
 		return
 	}
 
--- a/pkg/ring/ring.go
+++ b/pkg/ring/ring.go
@@ -34,6 +34,10 @@ const (
 
 	// CompactorRingKey is the key under which we store the compactors ring in the KVStore.
 	CompactorRingKey = "compactor"
+
+	// GetBufferSize is the suggested size of buffers passed to Ring.Get(). It's based on
+	// a typical replication factor 3, plus extra room for a JOINING + LEAVING instance.
+	GetBufferSize = 5
 )
 
 // ReadRing represents the read interface to the ring.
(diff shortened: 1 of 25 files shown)


pkg/ring/ring.go
```diff
@@ -113,6 +113,10 @@ type Config struct {
 	ReplicationFactor    int           `yaml:"replication_factor"`
 	ZoneAwarenessEnabled bool          `yaml:"zone_awareness_enabled"`
 	ExtendWrites         bool          `yaml:"extend_writes"`
+
+	// Whether the shuffle-sharding subring cache is disabled. This option is set
+	// internally and never exposed to the user.
+	SubringCacheDisabled bool `yaml:"-"`
```
pracucci, line 119: The config option is "disabled bool" instead of "enabled bool" so that the default will be "enabled" (because the zero value of a bool is false).

pstibrany (approved):
Makes sense to me

pkg/ring/util.go
```diff
@@ -34,6 +34,11 @@ func GenerateTokens(numTokens int, takenTokens []uint32) []uint32 {
 		i++
 	}
 
+	// Ensure returned tokens are sorted.
+	sort.Slice(tokens, func(i, j int) bool {
```
pracucci, line 38: In the ring lifecycler it's already guaranteed all tokens are sorted. I've introduced sorting here as well to have same guarantee in tests too.

pkg/ring/ring.go
```diff
@@ -34,16 +34,20 @@ const (
...
 type ReadRing interface {
 	prometheus.Collector
 
 	// Get returns n (or more) ingesters which form the replicas for the given key.
-	// buf is a slice to be overwritten for the return value
-	// to avoid memory allocation; can be nil.
-	Get(key uint32, op Operation, buf []IngesterDesc) (ReplicationSet, error)
+	// bufDescs, bufHosts and bufZones are slices to be overwritten for the return value
+	// to avoid memory allocation; can be nil, or created with ring.MakeBuffersForGet().
+	Get(key uint32, op Operation, bufDescs []IngesterDesc, bufHosts, bufZones []string) (ReplicationSet, error)
```
pstibrany, line 50: Nit: it makes sense for caller to pass buffer for ingesters, because `Get` returns it back. Hosts/zones buffers are only optimizations, and using a ring-specific pool (shared with subrings) may be a good alternative, without a need to change `Get` call.

pstibrany (approved):
LGTM so far.

pkg/ring/ring.go
pracucci: I tried it, but unfortunately it makes is way slower on the hot path:

```
BenchmarkRing_Get-12                        853           1126          +32.00%
```

pstibrany: Fair enough.

PR #3603, opened 2020-12-14
Memberlist: fix "forget"

pstibrany:
**What this PR does**: This PR fixes bug in "forget" feature when using memberlist.

When using memberlist (gossip), "forgetting" an entry doesn't delete it from the KV store, but sets entry to the "LEFT" state instead. Reason is that memberlist needs something to gossip about, and it can gossip about "LEFT" entry but not deleted entry. Unfortunately during this step, we didn't set Timestamp to current time, but reused last Timestamp of the entry. If there was already another message in the gossip network with higher timestamp, such message would eventually propagate everywhere, "resurrecting" the forgotten entry (because of higher timestamp, even though it's still in the past).

Solution is to set entry timestamp to "now" when setting entry to LEFT state.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [ENHANCEMENT] Memberlist: add status page (/memberlist) with available details about memberlist-based KV store and memberlist cluster. It's also possible to view KV values in Go struct or JSON format, or download for inspection. #3575
 * [ENHANCEMENT] Memberlist: client can now keep a size-bounded buffer with sent and received messages and display them in the admin UI (/memberlist) for troubleshooting. #3581
 * [BUGFIX] Query-Frontend: `cortex_query_seconds_total` now return seconds not nanoseconds. #3589
+* [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603 
 
 ## 1.6.0-rc.0 in progress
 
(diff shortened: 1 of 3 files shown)


docs/proposals/tenant-deletion.md
```diff
@@ -0,0 +1,152 @@
+---
+title: "Deletion of Tenant Data from Blocks Storage"
```
ranton256, line 2: I'm guessing you didn't mean to have this file in the review since it looks unrelated?

ranton256 (commented):
LGTM except the tenant deletion proposal is included, and assume that was meant to be separate.

docs/proposals/tenant-deletion.md
pstibrany: Oh, right. Thanks. That is part of different PR.

pracucci (approved):
Good finding, fix LGTM!

nicolai86:
this change improves the issue of "ghost" entries in the ring after scaling down but doesn't seem to fix the issue entirely.

for verification of this patch we quickly scaled up store-gateways from 12 to 80 instances and back down to 12, simulating a burst in auto scaling. About 3 1/2 hours after downscaling without any further scaling activity the store gateway ring reports unhealthy members in ring every 20 minutes apart.
Screen Shot 2021-03-23 at 1 28 50 PM

instance ids are random members from the previous scale up. Any ideas how to triage this issue and resolve it for good?

pracucci:
> this change improves the issue of "ghost" entries in the ring after scaling down but doesn't seem to fix the issue entirely.

@nicolai86 I would suggest you to open an issue instead of commenting on a merged PR (then the issue can reference this PR). Moving forward the discussion in an issue looks the more appropriate place.

Issue #3570, opened 2020-12-04
Crash and restart of store-gateway puts extra pressure on other store-gateways


pstibrany:
During recent incident, we have observed following behaviour:

1. Some store gateways have crashed, and restarted
2. On restart, store gateways join the ring and start the initial sync of blocks. During this sync, they are in the JOINING state.
3. Other (ACTIVE) store-gateways notice that ring has changed, because of change of state from ACTIVE to JOINING (for restarted gateways). This triggers synchronization of the blocks on these ACTIVE store-gateways.
4. ACTIVE store-gateways will also load blocks that belong to JOINING instances. 
5. After JOINING store-gateways are finished with initial-sync, they change state to ACTIVE again, which triggers another sync in other store-gateways (this will actually unload blocks from other instances).

Step 4. is problematic, because it puts extra pressure on store-gateways that hasn't crashed -- they will have disproportionally bigger number of blocks loaded. This can eventually make them crash as well, if they run out of resources. If they crash, they will start initial-sync in JOINING state again, making other ACTIVE store-gateways start loading more blocks.

pracucci:
> Step 4. is problematic, because it puts extra pressure on store-gateways that hasn't crashed -- they will have disproportionally bigger number of blocks loaded.

Agree and we need to fix it. The reason why I did it was so to cover the scale you scale up to a number of blocks >= replication factor. Let's say you start 10 new store-gateway at the same time. They will all be JOINING while synching blocks: if other ACTIVE store-gateways start to offload blocks before the JOINING store-gateways are still loading blocks, there will be a period of time during which some blocks are not loaded by any store-gateway and queries will fail.

Thoughts?

pstibrany:
> Agree and we need to fix it.
> 
> The reason why I did it was so to cover the scale you scale up to a number of blocks >= replication factor. Let's say you start 10 new store-gateway at the same time. They will all be JOINING while synching blocks: if other ACTIVE store-gateways start to offload blocks before the JOINING store-gateways are still loading blocks, there will be a period of time during which some blocks are not loaded by any store-gateway and queries will fail.

You are right, we don't want to existing store-gateways to start *offloading* blocks before JOINING store-gateways are ready to serve requests. On the other hand, if existing ACTIVE gateways *don't have the blocks yet*, they should also not *start loading* them either because JOINING gateways are already doing that.

pracucci:
_Still valid_

pracucci:
[...]

PR #3604, opened 2020-12-14
Store unregister shutdown

### LINKED ISSUE #2823: Store-gateway blocks resharding during rollout

andrejbranch:
**What this PR does**: Allows prevention of unregistering store gateways from the ring on shutdown. Allows prevention of replicating store gateway shard blocks when leaving the ring.

Fixes #2823

**Checklist**
- [X] Tests updated
- [X] Documentation added

Will update change log after feedback

### DIFF
--- a/docs/configuration/config-file-reference.md
+++ b/docs/configuration/config-file-reference.md
@@ -4165,6 +4165,18 @@ sharding_ring:
   # CLI flag: -store-gateway.sharding-ring.instance-availability-zone
   [instance_availability_zone: <string> | default = ""]
 
+  # Unregister from the ring upon clean shutdown. It can be useful to disable
+  # for rolling restarts with consistent naming
+  # CLI flag: -store-gateway.sharding-ring.unregister-on-shutdown
+  [unregister_on_shutdown: <boolean> | default = true]
+
+  # Try writing to an additional store gateway in the presence of a gateway not
+  # in the ACTIVE state. It is useful to disable this along with
+  # -store-gateway.sharding-ring.unregister-on-shutdown=false in order to not
+  # spread blocks to extra gateways during rolling restarts with consistent naming.
+  # CLI flag: -store-gateway.sharding-ring.extend-writes
+  [extend_writes: <boolean> | default = true]
+
 # The sharding strategy to use. Supported values are: default, shuffle-sharding.
 # CLI flag: -store-gateway.sharding-strategy
 [sharding_strategy: <string> | default = "default"]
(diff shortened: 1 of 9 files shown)


pstibrany:
> Allows prevention of unregistering store gateways from the ring on shutdown. Allows prevention of replicating store gateway shard blocks when leaving the ring.

Keeping store-gateways in the ring isn't enough. When store-gateway restarts, and starts syncing blocks, it sets itself to JOINING state in the ring. Other store-gateways will notice this change and they will start loading the blocks too. See issue  #3570 for discussion.

pstibrany:
> Keeping store-gateways in the ring isn't enough. When store-gateway restarts, and starts syncing blocks, it sets itself to JOINING state in the ring. Other store-gateways will notice this change and they will start loading the blocks too. See issue #3570 for discussion.

Scratch that... it may be enough if only single store-gateway is restarting. I need to rethink this.

bboreham:
@andrejbranch do you think you will come back to work on this?

andrejbranch:
@bboreham happy too but need some direction

PR #3611, opened 2020-12-16
Replace existing docker-compose.yml with generated one.

pstibrany:
**What this PR does**: This PR replaces `development/tsdb-blocks-storage-s3/docker-compose.yml ` with jsonnet file, and runs jsonnet from `compose-up.sh` script.

This simplifies working with the file, and allows for more possibilities (not in this PR, but we could eg. generate Prometheus and Grafana Agent config files from the same jsonnet, or merge different docker-compose setups into one, and just use different config flags).

Opening this PR to gather feedback from other developers using Cortex docker-compose setup.

**Checklist**
- [NA] Tests updated
- [NA] Documentation added


pracucci (commented):
I don't know. I see what you're trying to achieve, but I fear this made represent a barrier for Cortex contributors which need an easy way to play with a local Cortex setup without knowing jsonnet. I'm using to jsonnet and I personally find the whole `docker-compose.jsonnet` a bit hard to read.

pstibrany:
> I don't know. I see what you're trying to achieve, but I fear this made represent a barrier for Cortex contributors which need an easy way to play with a local Cortex setup without knowing jsonnet. I'm using to jsonnet and I personally find the whole `docker-compose.jsonnet` a bit hard to read.

I think I went too far with the jsonnet, and I could make it bit more readable.

Assuming that, what do you think about having both, jsonnet and yaml version, and adding a lint check to make sure they stay in sync? That way people don't need to have jsonnet for just running the setup, and only need it when they want to commit modifications.

pracucci:
> Assuming that, what do you think about having both, jsonnet and yaml version, and adding a lint check to make sure they stay in sync? That way people don't need to have jsonnet for just running the setup, and only need it when they want to commit modifications.

This may be a good compromise. I would be up for that.

pstibrany:
I still have hope.

pracucci (approved):

PR #3613, opened 2020-12-17
Perform final tenant cleanup after last block is deleted.

pstibrany:
**What this PR does**: This PR implements final tenant cleanup after some delay since last tenant's block has been deleted. Cleanup will remove all remaining marker files (including tenant deletion), and meta.json files stored under `<tenant>/debug/meta` prefix.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/chunk/purger/blocks_purger_api.go
+++ b/pkg/chunk/purger/blocks_purger_api.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -44,7 +45,7 @@ func (api *BlocksPurgerAPI) DeleteTenant(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	err = cortex_tsdb.WriteTenantDeletionMark(r.Context(), api.bucketClient, userID)
+	err = cortex_tsdb.WriteTenantDeletionMark(r.Context(), api.bucketClient, userID, cortex_tsdb.NewTenantDeletionMark(time.Now()))
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
@@ -58,8 +59,8 @@ func (api *BlocksPurgerAPI) DeleteTenant(w http.ResponseWriter, r *http.Request)
 type DeleteTenantStatusResponse struct {
 	TenantID                  string `json:"tenant_id"`
 	BlocksDeleted             bool   `json:"blocks_deleted"`
-	RuleGroupsDeleted         bool   `json:"rule_groups_deleted"`
-	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted"`
+	RuleGroupsDeleted         bool   `json:"rule_groups_deleted,omitempty"`
+	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted,omitempty"`
 }
 
 func (api *BlocksPurgerAPI) DeleteTenantStatus(w http.ResponseWriter, r *http.Request) {
--- a/pkg/chunk/purger/blocks_purger_api.go
+++ b/pkg/chunk/purger/blocks_purger_api.go
@@ -59,8 +59,8 @@ func (api *BlocksPurgerAPI) DeleteTenant(w http.ResponseWriter, r *http.Request)
 type DeleteTenantStatusResponse struct {
 	TenantID                  string `json:"tenant_id"`
 	BlocksDeleted             bool   `json:"blocks_deleted"`
-	RuleGroupsDeleted         bool   `json:"rule_groups_deleted,omitempty"`
-	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted,omitempty"`
+	RuleGroupsDeleted         bool   `json:"rule_groups_deleted,omitempty"`          // Not yet supported.
+	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted,omitempty"` // Not yet supported.
 }
 
 func (api *BlocksPurgerAPI) DeleteTenantStatus(w http.ResponseWriter, r *http.Request) {
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -227,6 +227,8 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
 
 	if deleted > 0 {
 		level.Info(userLogger).Log("msg", "deleted blocks for user marked for deletion", "deletedBlocks", deleted)
+	} else {
+		level.Info(userLogger).Log("msg", "no blocks found")
 	}
 
 	mark, err := cortex_tsdb.ReadTenantDeletionMark(ctx, c.bucketClient, userID)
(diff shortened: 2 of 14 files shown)


pkg/compactor/blocks_cleaner.go
```diff
@@ -223,10 +225,72 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
...
+		level.Info(userLogger).Log("msg", "updating finished time in tenant deletion mark")
+		mark.FinishedTime = time.Now().Unix()
+		return errors.Wrap(cortex_tsdb.WriteTenantDeletionMark(ctx, c.bucketClient, userID, mark), "failed to update tenant deletion mark")
+	}
+
+	if time.Since(time.Unix(mark.FinishedTime, 0)) < c.cfg.TenantCleanupDelay {
+		return nil
+	}
+
+	level.Info(userLogger).Log("msg", "cleaning up remaining tenant data")
```
pracucci, line 255: Is this log message useful? If so, I would improve the log message. This log comes from the compactor and this generic message may not be clear to the user.

pkg/compactor/blocks_cleaner.go
```diff
@@ -223,10 +225,72 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
...
+	if deleted, err := deletePrefix(ctx, userBucket, bucketindex.MarkersPathname, userLogger); err != nil {
+		return errors.Wrap(err, "failed to delete marker files for user marked for deletion")
+	} else if deleted > 0 {
+		level.Info(userLogger).Log("msg", "deleted marker files", "count", deleted)
+	}
+
 	return nil
 }
 
+func deletePrefix(ctx context.Context, bkt objstore.Bucket, prefix string, logger log.Logger) (int, error) {
```
pracucci, line 274: Given this is a generic helper, what do you think to move it to `pkg/storage/bucket`?

pkg/compactor/blocks_cleaner.go
```diff
@@ -223,10 +225,72 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
...
+
+	if time.Since(time.Unix(mark.FinishedTime, 0)) < c.cfg.TenantCleanupDelay {
+		return nil
+	}
+
+	level.Info(userLogger).Log("msg", "cleaning up remaining tenant data")
+
+	// Let's do final cleanup of tenant.
+	if deleted, err := deletePrefix(ctx, userBucket, block.DebugMetas, userLogger); err != nil {
+		return errors.Wrap(err, "failed to delete debug/metas for user marked for deletion")
```
pracucci, line 259:
```suggestion
		return errors.Wrapf(err, "failed to delete %s for user marked for deletion", block.DebugMetas)
```

pkg/compactor/blocks_cleaner.go
```diff
@@ -223,10 +225,72 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
...
+	} else if deleted > 0 {
+		level.Info(userLogger).Log("msg", "deleted debug/metas files", "count", deleted)
```
pracucci, line 261:
```suggestion
		level.Info(userLogger).Log("msg", "deleted debug/metas files for tenant marked for deletion", "count", deleted)
```

pkg/compactor/blocks_cleaner.go
```diff
@@ -223,10 +225,72 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
 	c.tenantBlocks.DeleteLabelValues(userID)
 	c.tenantMarkedBlocks.DeleteLabelValues(userID)
 
-	level.Info(userLogger).Log("msg", "finished deleting blocks for user marked for deletion", "deletedBlocks", deleted)
+	if deleted > 0 {
+		level.Info(userLogger).Log("msg", "deleted blocks for user marked for deletion", "deletedBlocks", deleted)
```
pracucci, line 229: Log messages here mix "tenant" and "user". I would change them to get stick to "tenant" for naming consistency.

pkg/compactor/blocks_cleaner_test.go
```diff
@@ -100,6 +108,7 @@ func testBlocksCleanerWithOptions(t *testing.T, options testBlocksCleanerOptions
 		CleanupInterval:                    time.Minute,
 		CleanupConcurrency:                 options.concurrency,
 		BlockDeletionMarksMigrationEnabled: options.markersMigrationEnabled,
+		TenantCleanupDelay:                 0, // immediately
```
pracucci, line 111: Would be great to set this to an higher value (adjusting `user-4` `FinishedTime` accordingly), and add another user which has the `FinishedTime` < delay, to make sure it's not cleaned up.

pracucci (approved):
Very good job, LGTM! I just left few nits.

pkg/compactor/blocks_cleaner.go
```diff
@@ -223,10 +225,72 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
...
+	mark, err := cortex_tsdb.ReadTenantDeletionMark(ctx, c.bucketClient, userID)
+	if err != nil {
+		return errors.Wrap(err, "failed to read tenant deletion mark")
+	}
+	if mark == nil {
+		return errors.Wrap(err, "cannot find tenant deletion mark anymore")
+	}
+
+	// If we have just deleted some blocks, update "finished" time. Also update "finished" time if it wasn't set yet, but there are no blocks.
+	// Note: this UPDATES the tenant deletion mark. Components that use caching bucket will NOT SEE this update,
```
ranton256, line 243: Any chance of multiple callers updating it? I assume if they do they will write similar times and it will be ok?

pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -58,8 +59,8 @@ func (api *BlocksPurgerAPI) DeleteTenant(w http.ResponseWriter, r *http.Request)
 type DeleteTenantStatusResponse struct {
 	TenantID                  string `json:"tenant_id"`
 	BlocksDeleted             bool   `json:"blocks_deleted"`
-	RuleGroupsDeleted         bool   `json:"rule_groups_deleted"`
-	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted"`
+	RuleGroupsDeleted         bool   `json:"rule_groups_deleted,omitempty"`
+	AlertManagerConfigDeleted bool   `json:"alert_manager_config_deleted,omitempty"`
```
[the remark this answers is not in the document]
pstibrany: Reason for hiding them was exactly because they are not supported yet. Once support is added, I would remove `omitempty`.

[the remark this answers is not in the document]
pracucci: Ok, as you prefer. I think it was clear to state they're not supported having them always exported with the `false` value, but your way works too.

pkg/compactor/blocks_cleaner.go
pstibrany: I will improve it. I think it's useful to know when final deletion was done.

pkg/compactor/blocks_cleaner.go
pstibrany: Within one compactor, it is only called once.

Multiple compactors may happen to call this method concurrently (eg. one compactor is doing compaction, and another compactor starts and "takes over" the user). In that case, they may overwrite each-others timestamp if they both happen to delete some blocks or read tenant deletion mark with no timestamp. In that case last write will win, which will delay eventual final cleanup of data for tenant.

pkg/compactor/blocks_cleaner.go
```diff
@@ -223,10 +225,72 @@ func (c *BlocksCleaner) deleteUser(ctx context.Context, userID string) error {
 	c.tenantBlocks.DeleteLabelValues(userID)
 	c.tenantMarkedBlocks.DeleteLabelValues(userID)
 
-	level.Info(userLogger).Log("msg", "finished deleting blocks for user marked for deletion", "deletedBlocks", deleted)
+	if deleted > 0 {
+		level.Info(userLogger).Log("msg", "deleted blocks for user marked for deletion", "deletedBlocks", deleted)
+	} else {
+		level.Info(userLogger).Log("msg", "no blocks found")
```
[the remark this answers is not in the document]
pstibrany: I will remove this message, since it would repeat on each compaction run until tenant is fully deleted.

pstibrany:
> Very good job, LGTM! I just left few nits.

Thanks for your review. I've addressed your feedback, and PR is ready for another look.

pracucci (approved):
LGTM, thanks!

PR #3614, opened 2020-12-17
Add bucket index support to querier

pracucci:
**What this PR does**:
In this PR I'm adding the bucket index support to querier. Using the bucket index in the querier is optional and disabled by default (and will be until marked experimental).

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/storage/tsdb/caching_bucket.go
```diff
@@ -92,6 +94,8 @@ func (cfg *MetadataCacheConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix
 	f.DurationVar(&cfg.MetafileContentTTL, prefix+"metafile-content-ttl", 24*time.Hour, "How long to cache content of the metafile.")
 	f.IntVar(&cfg.MetafileMaxSize, prefix+"metafile-max-size-bytes", 1*1024*1024, "Maximum size of metafile content to cache in bytes.")
 	f.DurationVar(&cfg.MetafileAttributesTTL, prefix+"metafile-attributes-ttl", 168*time.Hour, "How long to cache attributes of the block metafile.")
+	f.DurationVar(&cfg.BucketIndexContentTTL, prefix+"bucket-index-content-ttl", 5*time.Minute, "How long to cache content of the metafile.")
```
pstibrany, line 97: s/metafile/bucket index/

pkg/storage/tsdb/bucketindex/loader.go
```diff
@@ -0,0 +1,253 @@
...
+func (l *Loader) updateCachedIndexes(ctx context.Context) error {
+	var (
+		toUpdate []string
+		toDelete []string
+	)
+
+	// Build a list of users for which we should update or delete the index.
+	l.indexesMx.RLock()
+	for userID, entry := range l.indexes {
+		if time.Since(entry.getRequestedAt()) >= l.cfg.IdleTimeout {
```
pstibrany, line 151: Nit: Candidate for `switch { ... }` to reduces some else-ifs.

pkg/storage/tsdb/bucketindex/loader.go
```diff
@@ -0,0 +1,253 @@
...
+		} else if entry.err == nil && time.Since(entry.getUpdatedAt()) >= l.cfg.UpdateOnStaleInterval {
+			toUpdate = append(toUpdate, userID)
+		}
+	}
+	l.indexesMx.RUnlock()
+
+	// Delete unused indexes, if confirmed they're still unused.
+	l.indexesMx.Lock()
+	for _, userID := range toDelete {
+		if idx := l.indexes[userID]; time.Since(idx.getRequestedAt()) >= l.cfg.IdleTimeout {
```
pstibrany, line 164: nit: reuse `time.Now()` between iterations?

pkg/storage/tsdb/bucketindex/loader.go
```diff
@@ -0,0 +1,253 @@
...
+			delete(l.indexes, userID)
+			level.Info(l.logger).Log("msg", "unloaded bucket index", "user", userID, "reason", "idle")
+		}
+	}
+	l.indexesMx.Unlock()
+
+	// Update actively used indexes.
+	for _, userID := range toUpdate {
+		readCtx, cancel := context.WithTimeout(ctx, readIndexTimeout)
```
pstibrany, line 173: To avoid linters panicking due to defer in the loop, we could move entire for body into a method. It would also make `defer` more efficient.

pkg/storage/tsdb/bucketindex/loader.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+type cachedIndex struct {
+	// We cache either the index or the error occurred while fetching it. They're
+	// mutually exclusive.
+	index *Index
+	err   error
+
+	// Unix timestamp (seconds) of when the index has been updated from the storage the last time.
+	updatedAt *atomic.Int64
```
pstibrany, line 230: Nit: No need for pointers on atomic values, if outer structure (`cachedIndex`) is used via pointers.

pkg/storage/tsdb/bucketindex/loader.go
```diff
@@ -0,0 +1,253 @@
...
+func (l *Loader) GetIndex(ctx context.Context, userID string) (*Index, error) {
+	l.indexesMx.RLock()
+	if entry := l.indexes[userID]; entry != nil {
+		idx := entry.index
+		err := entry.err
+		l.indexesMx.RUnlock()
+
+		// We don't check if the index is stale because it's the responsibility
+		// of the background job to keep it updated.
+		entry.requestedAt.Store(time.Now().Unix())
```
pstibrany, line 100: I'd suggest making update of requestedAt timestamp a method on `*cachedIndex`, to keep the logic at single place.

pkg/querier/blocks_finder_bucket_index_test.go
```diff
@@ -0,0 +1,219 @@
...
+		idx.Blocks = append(idx.Blocks, &bucketindex.Block{ID: id, MinTime: minT, MaxTime: maxT})
+	}
+	for i := 1; i <= numDeletionMarks; i++ {
+		id := ulid.MustNew(uint64(i), nil)
+		idx.BlockDeletionMarks = append(idx.BlockDeletionMarks, &bucketindex.BlockDeletionMark{ID: id, DeletionTime: time.Now().Unix()})
+	}
+	require.NoError(b, bucketindex.WriteIndex(ctx, bkt, userID, idx))
+	finder := prepareBucketIndexBlocksFinder(b, bkt)
+
+	for n := 0; n < b.N; n++ {
```
pstibrany, line 145: `b.ResetTimer()` before running benchmark itself?

pkg/storage/tsdb/bucketindex/loader_test.go
```diff
@@ -0,0 +1,426 @@
...
+	assert.NoError(t, testutil.GatherAndCompare(reg, bytes.NewBufferString(`
+		# HELP cortex_bucket_index_load_failures_total Total number of bucket index loading failures.
+		# TYPE cortex_bucket_index_load_failures_total counter
+		cortex_bucket_index_load_failures_total 0
+		# HELP cortex_bucket_index_loaded Number of bucket indexes currently loaded in-memory.
+		# TYPE cortex_bucket_index_loaded gauge
+		cortex_bucket_index_loaded 1
+	`),
+		"cortex_bucket_index_load_failures_total",
+		"cortex_bucket_index_loaded",
```
pstibrany, line 184: Should we verify `cortex_bucket_index_loads_total` as well? (I guess it could be bit flaky though)

pstibrany (approved):
Wow, great job! No major problems found, logic is solid. :clap:

pkg/querier/blocks_finder_bucket_index.go
```diff
@@ -0,0 +1,137 @@
...
+	// Filter blocks containing samples within the range.
+	for _, block := range idx.Blocks {
+		if !block.Within(minT, maxT) {
+			continue
+		}
+
+		matchingBlocks[block.ID] = block
+	}
+
+	for _, mark := range idx.BlockDeletionMarks {
```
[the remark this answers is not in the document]
pracucci: Yes, that was an idea. Then I wrote a benchmark (1K blocks, 100 marks) and it takes 40Kns for each `GetBlocks()` call, reason why I believe it's not worth it.

pkg/querier/blocks_finder_bucket_index.go
```diff
@@ -0,0 +1,137 @@
...
+		loader: bucketindex.NewLoader(cfg.IndexLoader, bkt, logger, reg),
+	}
+
+	var err error
+	f.subservices, err = services.NewManager(f.loader)
+	if err != nil {
+		return nil, err
+	}
+
+	f.Service = services.NewBasicService(f.starting, f.running, f.stopping)
```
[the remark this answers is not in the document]
pracucci: Very good point!

pkg/storage/tsdb/bucketindex/loader_test.go
pracucci: I didn't in this (and few other) test because it would be flaky. There's no easy way to check it in a stable way, because we have no control over the background checks.

pracucci:
Thanks @pstibrany for your deep review! After reading your comments I realised it was definitely a draft what I've submitted 😉  I've addressed all comments, except a couple I've commented about.

docs/blocks-storage/bucket-index.md
```diff
@@ -0,0 +1,42 @@
+---
+title: "Bucket Index"
+linkTitle: "Bucket Index"
+weight: 5
+slug: bucket-index
+---
+
+The bucket index is a **per-tenant file containing the list of blocks and block deletion marks** in the storage. The bucket index itself is stored in the backend object storage, is periodically updated by the compactor and used by queriers to discover blocks in the storage.
+
+The bucket index usage is **optional** and can be enabled via `-blocks-storage.bucket-store.bucket-index.enabled=true` (or its respective YAML config option).
```
pstibrany, line 10: Explain why it is useful?

docs/blocks-storage/bucket-index.md
```diff
@@ -0,0 +1,42 @@
...
+  This option configures how frequently a cached bucket index should be refreshed.
+- `-blocks-storage.bucket-store.bucket-index.update-on-error-interval`<br />
+  If downloading a bucket index fails, the failure is cached for a short time in order to avoid hammering the backend storage. This option configures how frequently a bucket index, which previously failed to load, should be tried to load again.
+
+If a bucket index is unused for a long time (configurable via `-blocks-storage.bucket-store.bucket-index.idle-timeout`), e.g. because that querier instance is not receiving any query from the tenant, the querier will offload it, stopping to keep it updated at regular intervals. This is particularly for tenants which are resharded to different queriers when [shuffle sharding](../guides/shuffle-sharding.md) is enabled.
+
+Finally, the querier, at query time, checks how old is a bucket index (based on its `updated_at`) and fail a query if its age is older than `-blocks-storage.bucket-store.bucket-index.max-stale-period`. This circuit breaker is used to ensure queriers will not return any partial query results due to a stale view over the long-term storage.
```
pstibrany, line 42: You have also added caching of index into caching-bucket. Is that worth mentioning as well?

docs/blocks-storage/compactor.md
```diff
@@ -10,7 +10,7 @@ slug: compactor
 The **compactor** is an service which is responsible to:
 
 - Compact multiple blocks of a given tenant into a single optimized larger block. This helps to reduce storage costs (deduplication, index size reduction), and increase query speed (querying fewer blocks is faster).
-- Keep the per-tenant bucket index updated. The bucket index is used by [queriers](./querier.md) and [store-gateways](./store-gateway.md) to discover new blocks in the storage.
+- Keep the per-tenant bucket index updated. The [bucket index](./bucket-index.md) is used by [queriers](./querier.md) and [store-gateways](./store-gateway.md) to discover new blocks in the storage.
```
pstibrany, line 13: In the bucket-index.md only queriers are mentioned. Let's make it consistent (either way is fine, but store-gateways support is not yet there).

docs/blocks-storage/querier.md
```diff
@@ -13,12 +13,28 @@ The querier is **stateless**.
...
 
 While running, queriers periodically iterate over the storage bucket to discover new tenants and recently uploaded blocks. Queriers do **not** download any content from blocks except a small `meta.json` file containing the block's metadata (including the minimum and maximum timestamp of samples within the block).
 
 Queriers use the metadata to compute the list of blocks that need to be queried at query time and fetch matching series from the [store-gateway](./store-gateway.md) instances holding the required blocks.
 
+### Bucket index enabled
+
+When [bucket index](./bucket-index.md) is enabled, queriers lazily download the bucket index upon the 1st query received for a given tenant, cache it in memory and periodically keep it update. The bucket index contains the list of blocks and block deletion marks of a tenant, which is later used during the query execution to find the set of blocks that need to be queried for the given query.
```
pstibrany, line 31: s/1st/first/

docs/blocks-storage/compactor.md
pracucci: My plan is to update doc for store-gateway once support will be there. I will do it in the store-gateway PR.

pstibrany: My point is that this document already mentions store-gateways, but bucket-index.md doesn't, which may be confusing. Either mention it at both places, or not at all?

pkg/storage/tsdb/bucketindex/loader.go
```diff
@@ -0,0 +1,269 @@
...
+// 2. Update indexes which have been updated last time since >= update timeout
+func (l *Loader) checkCachedIndexes(ctx context.Context) error {
+	var (
+		toUpdate []string
+		toDelete []string
+		now      = time.Now()
+	)
+
+	// Build a list of users for which we should update or delete the index.
+	l.indexesMx.RLock()
```
ranton256, line 150: I would suggest making this a helper method to calculate toDelete and toUpdate since it's the scope that needs the lock and doing something separate to make things slightly easier to read.

docs/blocks-storage/_index.md
```diff
@@ -29,7 +29,7 @@ When running the Cortex blocks storage, the Cortex architecture doesn't signific
 
 The **[store-gateway](./store-gateway.md)** is responsible to query blocks and is used by the [querier](./querier.md) at query time. The store-gateway is required when running the blocks storage.
 
-The **[compactor](./compactor.md)** is responsible to merge and deduplicate smaller blocks into larger ones, in order to reduce the number of blocks stored in the long-term storage for a given tenant and query them more efficiently. It also keeps the bucket index updated and, for this reason, it's a required component.
+The **[compactor](./compactor.md)** is responsible to merge and deduplicate smaller blocks into larger ones, in order to reduce the number of blocks stored in the long-term storage for a given tenant and query them more efficiently. It also keeps the [bucket index](./bucket-index.md) updated and, for this reason, it's a required component.
```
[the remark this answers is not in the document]
pracucci: > Is this going to cause issues for users running with compactor now(not sure if there are such people)?

Did you mean "with compactor" or "without compactor"? I'm no sure I understand the question.

> Also, bucket-index.md says the bucket index is optional which seems like it contradicts this statement.

The compactor always writes the bucket index. The flag to enable bucket index is whether it should be used or not in queriers (and in the store-gateway in the upcoming PR too). The point is that, before enabling the bucket index in the queriers and store-gateway, you have to rollout the compactor first, so that the bucket index for all tenants is created before you enable it in querier and store-gateway. To simplify it (at least I thought it would have simplified), I've kept it always enabled in the compactor.

docs/blocks-storage/bucket-index.md
```diff
@@ -0,0 +1,42 @@
...
```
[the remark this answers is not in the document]
pracucci: I don't think falling back is a viable solution. The "bucket scan" logic requires a preventive bucket scanning, which we don't do when we enable the bucket index. Lazily bucket scanning would be too slow to do at query time. Moreover, fallback logic unfrequently exercised would bring further risks that fallback logic doesn't work as expected.

In my opinion, when you run Cortex with bucket index, the bucket index is an essential part of the system and it's required to be kept updated. The compactor already exports a metric with the timestamp of the last time the bucket index of each tenant has been updated, so that we can alert on it before the `max-stale-period` is reached.

docs/blocks-storage/querier.md
```diff
@@ -587,6 +604,14 @@ blocks_storage:
       # CLI flag: -blocks-storage.bucket-store.metadata-cache.metafile-attributes-ttl
       [metafile_attributes_ttl: <duration> | default = 168h]
 
+      # How long to cache content of the bucket index.
+      # CLI flag: -blocks-storage.bucket-store.metadata-cache.bucket-index-content-ttl
+      [bucket_index_content_ttl: <duration> | default = 5m]
+
+      # Maximum size of bucket index content to cache in bytes.
+      # CLI flag: -blocks-storage.bucket-store.metadata-cache.bucket-index-max-size-bytes
+      [bucket_index_max_size_bytes: <int> | default = 1048576]
```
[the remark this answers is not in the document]
pracucci: Yes. I will update the doc.

To answer you, it doesn't get cached (skipped to be cached). Bucket index should never hit such big size tho. We see less 50KB compressed for 1K blocks.

pkg/storage/tsdb/bucketindex/loader.go
pracucci: Good point, done. What do you think=

docs/blocks-storage/querier.md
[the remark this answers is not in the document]
pracucci: Done. Does the new description clarifies it?

docs/blocks-storage/bucket-index.md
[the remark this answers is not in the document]
pracucci: What do you think?

pkg/storage/tsdb/bucketindex/loader.go
ranton256: Thanks, LGTM

ranton256 (approved):

docs/blocks-storage/bucket-index.md
pracucci: Yes, sure, done. WDYT?

docs/blocks-storage/compactor.md
pracucci: Oh right. I removed the mention to store-gateway. I will update it in the store-gateway PR.

docs/blocks-storage/bucket-index.md
```diff
@@ -0,0 +1,42 @@
...
+- **`block_deletion_marks`**<br />
+  List of block deletion marks.
+- **`updated_at`**<br />
+  Unix timestamp (seconds precision) of when the index has been updated (written in the storage) the last time.
+
+## How it gets updated
+
+The [compactor](./compactor.md) periodically scans the bucket and uploads an updated bucket index to the storage. The frequency at which the bucket index is updated can be configured via `-compactor.cleanup-interval`.
+
+The bucket index is built and updated by the compactor even if `-blocks-storage.bucket-store.bucket-index.enabled` has **not** been enabled. This is intentional and the overhead introduced by keeping the bucket index is non significative.
```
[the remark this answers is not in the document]
pracucci: I've updated the doc. Is it more clear now?

docs/blocks-storage/bucket-index.md
pracucci: Done. WDYT?

pracucci:
@pstibrany @ranton256 I'm going to merge this PR, but feel free to add further comments (if any). I will promptly address post-merge comments too.

Issue #3615, opened 2020-12-17
Ruler: rule group limit enforcement has off-by-one error


mattmendick:
**Describe the bug**
The ruler seems to have an issue where the rules per group limit is forcing one fewer than the limit, yielding an error message like this:
```
per-user rules per rule group limit (limit: 15 actual: 15)
```

**To Reproduce**
Submit a rule group with the same number of rules as the limit, ie: 15 rules in a group with the limit set to 15.

**Expected behavior**
The submitted rule group should be allowed to contain up to and containing the limit specified. The limit set to 15 should allow a group with 15 rules to be accepted.

pracucci:
_Still valid_

mtanda:
@pracucci 
We hit this issue in Grafana Cloud. Do you have plan to fix this?

pracucci:
> Do you have plan to fix this?

There's an open PR https://github.com/cortexproject/cortex/pull/3616 which got stuck. I talked to Matt -- the PR author -- and they're fine if anyone takes it over.

pracucci:
Fixed in #3616.

mtanda:
Thanks for quick fix!

PR #3616, opened 2020-12-17
Small fix for off-by-one rules limit enforcement

mattmendick:
**What this PR does**:

Fixes an off by one error where rules in a rule group were not able to be equal to the limit.

https://github.com/cortexproject/cortex/issues/3615

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * [BUGFIX] Query-frontend: Fix issue where cached entry size keeps increasing when making tiny query repeatedly. #3968
 * [BUGFIX] Compactor: `-compactor.blocks-retention-period` now supports weeks (`w`) and years (`y`). #4027
 * [BUGFIX] Querier: returning 422 (instead of 500) when query hits `max_chunks_per_query` limit with block storage, when the limit is hit in the store-gateway. #3937
+* [BUGFIX] Ruler: Rule group limit enforcement should now allow the same number of rules in a group as the limit. #3615
 
 ## Blocksconvert
 
--- a/pkg/ruler/api_test.go
+++ b/pkg/ruler/api_test.go
@@ -300,7 +300,7 @@ func TestRuler_DeleteNamespace(t *testing.T) {
 	require.Equal(t, "{\"status\":\"error\",\"data\":null,\"errorType\":\"server_error\",\"error\":\"unable to delete rg\"}", w.Body.String())
 }
 
-func TestRuler_Limits(t *testing.T) {
+func TestRuler_LimitsPerGroup(t *testing.T) {
 	cfg, cleanup := defaultRulerConfig(newMockRuleStore(make(map[string]rulespb.RuleGroupList)))
 	defer cleanup()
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Query-Frontend: `cortex_query_seconds_total` now return seconds not nanoseconds. #3589
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
+* [BUGFIX] Ruler: Rule group limit enforcement should now allow the same number of rules in a group as the limit.#3615
```
pracucci, line 19:
```suggestion
* [BUGFIX] Ruler: Rule group limit enforcement should now allow the same number of rules in a group as the limit. #3615
```

pracucci (approved):
Thanks!

gouthamve (approved):

pracucci:
@mattmendick Could you fix the test `TestRuler_Limits/when_exceeding_the_rule_group_limit` accordingly?

bboreham:
Will you update the tests?
Looks like someone kindly wrote a test to check the erroneous behaviour.

mattmendick:
Absolutely! Will work on this soon. Thanks for both of you taking a look!

jtlisi (approved):
LGTM

jtlisi:
After a closer look I think I figured out the issue causing tests to fail. When `AssertMaxRuleGroups` is called, it's only checking against the length of the number of existing rule groups without including the one you are adding. You need to add 1 to the length before passing to the function unless the rule being created already exists. To check if the rule already exists you can loop through the existing groups and see if any groups exist with the same group name and namespace.

pkg/ruler/api_test.go
```diff
@@ -337,24 +337,74 @@ rules:
...
-			output: "per-user rules per rule group limit (limit: 1 actual: 1) exceeded\n",
+			output: "per-user rule groups limit (limit: 1 actual: 2) exceeded\n",
 		},
 	}
 
 	for _, tt := range tc {
+		// define once so the requests build on each other so the number of rules can be tested
+		router := mux.NewRouter()
+		router.Path("/api/v1/rules/{namespace}").Methods("POST").HandlerFunc(a.CreateRuleGroup)
+
```
bboreham, line 406: I don't understand this change.  It's not run "once" - it's done again for each element in `tc`.
And I can't see what practical difference moving it outside of `Run()` makes.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 * [ENHANCEMENT] Querier: Implement result caching for tenant query federation. #3640
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
+* [BUGFIX] Ruler: Rule group limit enforcement should now allow the same number of rules in a group as the limit. #3616
```
pracucci, line 30: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

bboreham:
@mattmendick do you think you will come back to this?

pkg/ruler/api_test.go
mattmendick: Right you are, definitely not going to work that way :)

mattmendick:
I'd like to take another shot at this, many apologies for the length of time this has been taking, and I really appreciate all the guidance and help you all have been giving this! 🙏

CHANGELOG.md
pstibrany: Cortex release 1.8.0 is now in progress. Could you rebase master and move the CHANGELOG entry under the master / unreleased section, please?

pstibrany:
Based on https://github.com/cortexproject/cortex/issues/3615#issuecomment-815944848, rebased on top of master, and squashed to avoid DCO errors.

gotjosh (approved):
LGTM

Issue #3624, opened 2020-12-21
Data race in ruler (test flake)


bboreham:
Seen in one [unit test run](https://github.com/cortexproject/cortex/runs/1588947707?check_suite_focus=true) but passed on re-run.

I think the error indicates a failure to adhere to this instruction on `sync.WaitGroup.Add()`:
```
// Note that calls with a positive delta that occur when the counter is zero
// must happen before a Wait. 
```

logs:
```
level=info msg="ruler up and running"
level=info msg="ruler up and running"
level=info msg="ruler up and running"
level=info msg="ruler up and running"
level=info msg="ruler up and running"
==================
WARNING: DATA RACE
Write at 0x00c000aa6fa0 by goroutine 61:
  internal/race.Write()
      /usr/local/go/src/internal/race/race.go:41 +0x114
  sync.(*WaitGroup).Wait()
      /usr/local/go/src/sync/waitgroup.go:128 +0x115
  github.com/cortexproject/cortex/pkg/ruler.(*rulerNotifier).stop()
      /__w/cortex/cortex/pkg/ruler/notifier.go:71 +0x99
  github.com/cortexproject/cortex/pkg/ruler.(*DefaultMultiTenantManager).Stop()
      /__w/cortex/cortex/pkg/ruler/manager.go:233 +0x153
  github.com/cortexproject/cortex/pkg/ruler.(*Ruler).stopping()
      /__w/cortex/cortex/pkg/ruler/ruler.go:313 +0x7a
  github.com/cortexproject/cortex/pkg/ruler.(*Ruler).stopping-fm()
      /__w/cortex/cortex/pkg/ruler/ruler.go:312 +0x55
  github.com/cortexproject/cortex/pkg/util/services.(*BasicService).main()
      /__w/cortex/cortex/pkg/util/services/basic_service.go:184 +0x488

Previous read at 0x00c000aa6fa0 by goroutine 16:
  internal/race.Read()
      /usr/local/go/src/internal/race/race.go:37 +0x1e8
  sync.(*WaitGroup).Add()
      /usr/local/go/src/sync/waitgroup.go:71 +0x1fb
  github.com/cortexproject/cortex/pkg/ruler.(*rulerNotifier).run()
      /__w/cortex/cortex/pkg/ruler/notifier.go:43 +0x63

Goroutine 61 (running) created at:
  github.com/cortexproject/cortex/pkg/util/services.(*BasicService).StartAsync.func1()
[...]

PR #3626, opened 2020-12-22
Synchronous call to rulerNotifier.run()

pracucci:
**What this PR does**:
I was checking the issue reported at https://github.com/cortexproject/cortex/issues/3624. I've not been able to reproduce it (I tried adding few `time.Sleep()`), but something that popped my eyes is that `rulerNotifier.run()` should be called synchronously and not via `go n.run()`.

Maybe #3624 but not sure.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ruler/manager.go
+++ b/pkg/ruler/manager.go
@@ -206,7 +206,7 @@ func (r *DefaultMultiTenantManager) getOrCreateNotifier(userID string) (*notifie
 		},
 	}, log.With(r.logger, "user", userID))
 
-	go n.run()
+	n.run()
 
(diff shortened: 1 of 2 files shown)


bboreham (approved):
Good catch!

jtlisi (approved):
Nice

jtlisi:
@pracucci Is there a possible bug this fixes that would necessitate a changelog entry?

pracucci:
> @pracucci Is there a possible bug this fixes that would necessitate a changelog entry?

I don't think so. I believe the only problem is if you call `stop()` right after starting it, which should realistically happen only in tests.

PR #3627, opened 2020-12-22
Compactor: retry compaction of a single tenant on failure instead of re-running compaction for all tenants

pracucci:
**What this PR does**:
Currently, the compactor re-run compaction for all tenants in case at least 1 tenant failed. Despite re-running the compaction for a previously succeeded tenant doesn't redo the work, but considering each single compaction run needs to re-scan the bucket, this ends up to be pretty inefficient for Cortex clusters running with a large number of tenants.

In this PR I'm proposing to retry compaction on failure only for the failed tenants.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -461,7 +461,6 @@ func (c *Compactor) compactUsers(ctx context.Context) {
 	}
 
 	succeeded = true
-	return
 }
 
 func (c *Compactor) compactUserWithRetries(ctx context.Context, userID string) error {
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -367,7 +367,7 @@ func (c *Compactor) stopping(_ error) error {
 
 func (c *Compactor) running(ctx context.Context) error {
 	// Run an initial compaction before starting the interval.
-	c.compactUsers(ctx)
+	_ = c.compactUsers(ctx)
 
 	ticker := time.NewTicker(util.DurationWithJitter(c.compactorCfg.CompactionInterval, 0.05))
 	defer ticker.Stop()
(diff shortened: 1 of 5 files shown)


ranton256:
Thanks for this, should reduce the amount of list calls on s3 or similar.

ranton256 (approved):
LGTM

pkg/compactor/compactor.go
```diff
@@ -381,41 +381,26 @@ func (c *Compactor) running(ctx context.Context) error {
...
 }
 
-func (c *Compactor) compactUsersWithRetries(ctx context.Context) {
-	retries := util.NewBackoff(ctx, util.BackoffConfig{
-		MinBackoff: c.compactorCfg.retryMinBackoff,
-		MaxBackoff: c.compactorCfg.retryMaxBackoff,
-		MaxRetries: c.compactorCfg.CompactionRetries,
-	})
-
+func (c *Compactor) compactUsers(ctx context.Context) (returnErr error) {
```
pstibrany, line 384: Returned error is not used anywhere. We should remove it (we already log all errors anyway), and perhaps also simplify `returnErr` to boolean flag for updating metrics.

pracucci: Does it look better now?

pstibrany: Looks better. As a final change, I'd suggest removing unused return value completely, and convert `succeeded` into local variable.

pstibrany (approved):

pkg/compactor/compactor.go
pracucci: Oh yeah, sure, done.

PR #3629, opened 2020-12-22
Added block index attributes caching support to metadata cache

pracucci:
**What this PR does**:
The Thanos index-header reader issue an `Attributes()` request, other than get range requests:

https://github.com/thanos-io/thanos/blob/e9260e30f265f60e17d3f61ff361fb7357b3d34a/pkg/block/indexheader/binary_reader.go#L132-L134

The `Attributes()` API call is issued for each new block loaded in the store-gateway. Considering we replica blocks across store-gateway (defaults to a replication factor = 3), we may save some object store API calls caching the attributes in the metadata cache (block index is immutable, so it's safe to cache it).

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


ranton256 (approved):

pstibrany (approved):
LGTM, nice catch.

PR #3632, opened 2020-12-23
Do not close idle TSDBs while blocks shipping is in progress

### LINKED ISSUE #3620: ingester_v2 reporting quite a bit of "failed to delete idle TSDB" error
**Describe the bug**
We are seeing error message like 

```
level=error ts=2020-12-15T21:33:44.395830698Z caller=ingester_v2.go:1623 msg="failed to delete idle TSDB" user=some-user-id err="unlinkat /data/tsdb/some-user-id: directory not empty"
```
We are using [this commit](https://github.com/cortexproject/cortex/tree/af9e20c54ee97f409008f3e86541c5dfa5038e22)

**To Reproduce**
I am not sure

**Expected behavior**

**Environment:**
 - Infrastructure: EKS 
 - Deployment tool: Helm

**Storage Engine**
- [ X ] Blocks
- [ ] Chunks

**Additional Context**
The error happens during a call to `os.RemoveAll()` call, research shows that `RemoveAll()` might return "directory not empty" error if file are created under the directory while the removal process is on going. So I am wondering if there is any background process in ingester that maybe interfering with the local TSDB deletion process?

pracucci:
**What this PR does**:
In this PR I'm attempting one possible solution (there are multiples) to fix the race condition described in #3620.

Fixes #3620

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -1506,7 +1506,7 @@ func (i *Ingester) shipBlocks(ctx context.Context) {
 		// Run the shipper's Sync() to upload unshipped blocks. Make sure the TSDB state is active, in order to
 		// avoid any race condition with closing idle TSDBs.
 		if !userDB.casState(active, activeShipping) {
-			level.Warn(util.Logger).Log("msg", "shipper shipped because the TSDB is not active", "user", userID)
+			level.Info(util.Logger).Log("msg", "shipper skipped because the TSDB is not active", "user", userID)
 			return nil
 		}
 		defer userDB.casState(activeShipping, active)
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -52,7 +52,8 @@ type Shipper interface {
 type tsdbState int
 
 const (
-	active          tsdbState = iota // Pushes are allowed only in this state.
+	active          tsdbState = iota // Pushes are allowed.
+	activeShipping                   // Pushes are allowed. Blocks shipping is in progress.
 	forceCompacting                  // TSDB is being force-compacted.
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -1501,7 +1503,14 @@ func (i *Ingester) shipBlocks(ctx context.Context) {
 			}
 		}
 
-		// Run the shipper's Sync() to upload unshipped blocks.
+		// Run the shipper's Sync() to upload unshipped blocks. Make sure the TSDB state is active, in order to
+		// avoid any race condition with closing idle TSDBs.
+		if !userDB.casState(active, activeShipping) {
+			level.Warn(util.Logger).Log("msg", "shipper shipped because the TSDB is not active", "user", userID)
```
pstibrany, line 1509: typo: "shipper skipped ..."

Should this be a warning? Looks more like a debug message to me.

pstibrany (approved):
LGTM, good job!

pkg/ingester/ingester_v2.go
pracucci: > typo: "shipper skipped ..."

Ops, fixed!

> Should this be a warning? Looks more like a debug message to me.

Let's find a common ground. I lowered it to Info, but I believe it's worth to notice and most users don't run Cortex with debug log level.

pstibrany: My reasoning for "debug" level is that there are only two reasons how this can happen:
- during force-compaction, In this case, after force-compaction, flushing will call shipper anyway
- when tsdb is closed (or being closed), in this case we already know that shipper has nothing to ship

I think that in both cases, fact that shipper has skipped its cycle is not interesting.

Issue #3622, opened 2020-12-21
Feature request: be able to specify the retention time for blocks


AllenzhLi:
**Is your feature request related to a problem? Please describe.**
Now the compactor only compact the blocks, there is no retention.

**Describe the solution you'd like**
1. add a default retention config
2. add a api to update retention time range

AllenzhLi:
Is there any one work on this ?

pracucci:
We had some preliminary discussions to add a per-tenant rentention support in the compactor, but we haven't started the work yet. If you plan to work on it, I would strongly recommend you to write a design document to share, so that we can discuss the design along with the community before jumping on coding it.

AllenzhLi:
I want to work on it.

Later I will write a preliminary design document with my understanding.

AllenzhLi:
I have written a initial design document #3638

blind-oracle:
By the way, is there any retention policy at all for blocks storage?
Not per-tenant, but at least global?

pracucci:
> By the way, is there any retention policy at all for blocks storage?
> Not per-tenant, but at least global?

The current approch for a global retention is to define a bucket policy to automatically delete blocks after X time.

blind-oracle:
@pracucci Thanks.

But that would be a hard-delete, right?
The queriers would just lose those blocks right away without marking them as `to-be-deleted` as compactor does?
Wouldn't queries fail?

pracucci:
> But that would be a hard-delete, right?
> The queriers would just lose those blocks right away without marking them as `to-be-deleted` as compactor does?
> Wouldn't queries fail?

Yes, that would be an hard delete. However, Cortex has a config option to set how back data should be queried:

```
# Limit how long back data (series and metadata) can be queried, up until
# <lookback> duration ago. This limit is enforced in the query-frontend, querier
# and ruler. If the requested time range is outside the allowed range, the
# request will not fail but will be manipulated to only query data within the
# allowed time range. 0 to disable.
# CLI flag: -querier.max-query-lookback
[max_query_lookback: <duration> | default = 0s]
```

Let's say you wanna set a retention of 400 days. You can configure:
- `-querier.max-query-lookback=9600h`
- Bucket policy retention to 401 days, to have 1 day buffer and no failing query

blind-oracle:
Thanks, got it! I was wondering how to do it properly :)

Yes, having per-tenant policies would be a very nice feature

pracucci:
[...]

PR #3638, opened 2020-12-30
Proposal for tenant retention

AllenzhLi:
This is a proposal / design doc for tenant retention data.

### DIFF
--- a/docs/proposals/tenant_retention.md
+++ b/docs/proposals/tenant_retention.md
@@ -20,39 +20,26 @@ Metric data is growing over time per-tenant, at the same time, the value of data
 
 ## Background
 
-### tenant
+### tenants
 When using blocks storage, Cortex stores tenant’s data in object store for long-term storage of blocks, tenant id as part of the object store path. We discover all tenants via scan the root dir of bucket.
 
 ### runtime config
-Using the "overrides" mechanism (part of runtime config) already allows for per-tenant settings. See https://cortexmetrics.io/docs/configuration/arguments/#runtime-configuration-file for more details. Using it for tenant retention would fit nicely. Admin could set per-tenant retention here, and also have a single global value for tenants that don't have custom value set.
+Using the "overrides" mechanism (part of runtime config) already allows for per-tenant settings. See [runtime-configuration-file](https://cortexmetrics.io/docs/configuration/arguments/#runtime-configuration-file) for more details. Using it for tenant retention would fit nicely. Admin could set per-tenant retention here, and also have a single global value for tenants that don't have custom value set.
 
 ## Proposal
 
-### retention component
+### retention period field
 
-We propose to introduce an new runtime config component `retention`.
+We propose to introduce just one new field `RetentionPeriodPerTenant` in the Limits struct(defined at pkg/util/validation/limits.go).
 
-We can store per-tenant settings and global setting in the retention component. As a runtime config, the settings is reload period (which defaults to 10 seconds), so we can update the settings onfly.
+`RetentionPeriodPerTenant` setting how long historical metric data retention period per-tenant. `0` is disable.
 
-Per-tenant retention configuration will be applyed for in follow order:
+As a runtime config, the settings is reload period (which defaults to 10 seconds), so we can update the settings onfly.
 
-* tenant specified value
-* global value
-
-For each tenant, if a tenant-specific value exists, it will be used directly, otherwise, if a global value exists, then the global value will be used; If neither exists, do nothing.
+For each tenant, if a tenant-specific *runtime_config* value exists, it will be used directly, otherwise, if a default *limits_config* value exists, then the default value will be used; If neither exists, do nothing.
 
 ### Implementation
 
-The runtime config will add a retention component which include a dictionary to store per-tenant retention value and global retention value.
+A BlocksCleaner within the Compactor run periodically(which defaults to 1 hour) and the retention logic will insert into it. The logic should compare retention value to block `maxTime` and blocks that match `maxTime < now - retention` will be marked for delete.
 
-Compactor will check per-tanant retention value and compare to meta info of blocks. The blocks that match `maxTime < now - retention` will be marked for delete.
-
-## Compactor
-
-Upon discovering the tenant, the compactor will check following markers:
-
-- tenant delete marker
-- block delete marker
-
-If tenant delete marker exists, compactor use TenantCleaner delete tenant all data regardless of whether the block delete marker exists.
-If only block delete marker exists, compactor use BlocksCleaner delete the block after `deletion_delay` duration.
+Blocks deletion is not immediate, but follows a two steps process. See [soft-and-hard-blocks-deletion](https://cortexmetrics.io/docs/blocks-storage/compactor/#soft-and-hard-blocks-deletion)
--- a/docs/proposals/tenant_retention.md
+++ b/docs/proposals/tenant_retention.md
@@ -1,0 +1,64 @@
+---
+title: "Retention of Tenant Data from Blocks Storage"
+linkTitle: "Retention of Tenant Data from Blocks Storage"
+weight: 1
+slug: tenant-retention
+---
+
+- Author: [Allenzhli](https://github.com/Allenzhli)
+- Date: December 2020
+- Status: Proposed
+
+## Retention of tenant data
+
+## Problem
+
+In Cortex, we are missing retention of blocks storage. The metric data is growing over time per-tenant, at the same time, the value of data decreases. We want to support retention time of metric data like prometheus does.
+
+## Background
+
+When using blocks storage, Cortex stores tenant’s data in object store for long-term storage of blocks, tenant id as part of the object store path. We discover all tenants via scan the root dir of bucket.
+
+## Proposal
+
+### API Endpoints
+
+#### POST /compactor/retention_tenant
+
+We propose to introduce an `/compactor/retention_tenant` API endpoint to set data retention for the tenant.
+
+While this endpoint works as an “admin” endpoint and should not be exposed directly to tenants, it still needs to know under which tenant it should operate.
+For consistency with other endpoints this endpoint would therefore require an `X-Scope-OrgID` header and use the tenant from it.
+
+It is safe to call “retention_tenant” multiple times.
+
+#### GET /compactor/retention_tenant
+
+To query the state of data retention, an endpoint will be available: `/compactor/retention_tenant`.
+This will return the retention of current tenant.
+For consistency with other endpoints this endpoint would therefore require an `X-Scope-OrgID` header and use the tenant from it.
+
+### Implementation (asynchronous)
+
+Compactor will implement both API endpoints.
+Upon receiving the call to `/compactor/retention_tenant` endpoint, Compactor will initiate the retention by writing “retetion marker” objects for specific tenant to blocks bucket.
+
+Retention marker for the tenant will be an object stored under tenant prefix, eg. “/<tenant>/retention”.
+This object will contain the retention time when it was created, so that we can calculate and check if a block should be deleted.
+For every clean user period, we can check if any blocks should be retention, and the other should be delete.
+Delete a block also use deletetion marker.
+
+We could reuse a subset of the proposed Thanos tombstones format, or use custom format.
+
+### Blocks deletion marker
+
+Blocks deletion marker will be used by compactor, querier and store-gateway.
+
+#### Compactor
+
+Upon discovering the tenant, the compactor will check following markers:
+
+- tenant delete marker
+- tenant retention marker
+
+if only tenant retention exists, the compactor will check retention of all blocks that belong to the tenant after compact. The blocks which match `maxTime < now - retention` will be marked for delete.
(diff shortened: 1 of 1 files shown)


pstibrany:
We typically use "runtime config" (overrides) for configuring per-tenant limits and options. Is there a good reason not to do that for tenant retention? If so, I'd suggest to include those reasons in the design document. Otherwise, I'd suggest that we use "overrides" for configuring tenant retention as well.

AllenzhLi:
> We typically use "runtime config" (overrides) for configuring per-tenant limits and options. Is there a good reason not to do that for tenant retention? If so, I'd suggest to include those reasons in the design document. Otherwise, I'd suggest that we use "overrides" for configuring tenant retention as well. It would also simplify design – there will be no need to expose API for configuring or checking tenant retention.

Thank you for your reply.

The design is inspired by your tenant-deletion design.
 
The other reason to not use runtime config is that I does't find where can store the setting for every tenant.

pstibrany:
> The other reason to not use runtime config is that I does't find where can store the setting for every tenant. Maybe I missing something.

Using the "overrides" mechanism (part of runtime config) already allows for per-tenant settings. See https://cortexmetrics.io/docs/configuration/arguments/#runtime-configuration-file for more details. Using it for tenant retention would fit nicely. Admin could set per-tenant retention here, and also have a single global value for tenants that don't have custom value set.

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Author: [Allenzhli](https://github.com/Allenzhli)
+- Date: December 2020
+- Status: Proposed
+
+## Retention of tenant data
+
+## Problem
+
+In Cortex, we are missing retention of blocks storage. The metric data is growing over time per-tenant, at the same time, the value of data decreases. We want to support retention time of metric data like prometheus does.
```
pracucci, line 16: This is not entire correct. Data retention is typically achieved be achieved via a bucket policy. However, we should explain this has two main issues:

1. Not every backend storage support bucket policies
2. Bucket policies don't easily allow a per-tenant custom retention

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,64 @@
...
+
+When using blocks storage, Cortex stores tenant’s data in object store for long-term storage of blocks, tenant id as part of the object store path. We discover all tenants via scan the root dir of bucket.
+
+## Proposal
+
+### API Endpoints
+
+#### POST /compactor/retention_tenant
+
+We propose to introduce an `/compactor/retention_tenant` API endpoint to set data retention for the tenant.
```
pracucci, line 28: I wouldn't introduce any new API. We should use the already existing mechanism used to set per-tenant overrides (via the runtime config "overrides").

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,64 @@
...
+To query the state of data retention, an endpoint will be available: `/compactor/retention_tenant`.
+This will return the retention of current tenant.
+For consistency with other endpoints this endpoint would therefore require an `X-Scope-OrgID` header and use the tenant from it.
+
+### Implementation (asynchronous)
+
+Compactor will implement both API endpoints.
+Upon receiving the call to `/compactor/retention_tenant` endpoint, Compactor will initiate the retention by writing “retetion marker” objects for specific tenant to blocks bucket.
+
+Retention marker for the tenant will be an object stored under tenant prefix, eg. “/<tenant>/retention”.
```
pracucci, line 46: I wouldn't use any retention marker. The per-tenant retention should be part of the config.

pracucci (commented):
Thanks for working on this proposal! I left a couple of high level comments. I would love going deeper into the review once these are addressed. Thanks!

AllenzhLi:
Sorry, I haven't paid attention to it for a while and got stuck，now I'm back.  

I have test the runtime_config, it is looks nice and simplify except that I can't update the setting via a http api.

AllenzhLi:
#3622

pstibrany:
> I have test the runtime_config, it is looks nice and simplify except that I can't update the setting via a http api.

There is no API to change runtime config. You can modify YAML file directly, and Cortex will reload it automatically.

AllenzhLi:
> There is no API to change runtime config. You can modify YAML file directly, and Cortex will reload it automatically.

Yeah，it will reload periodly. 

But update the retention duration onfly is also needed.

https://grafana.com/docs/loki/latest/operations/storage/retention/
Loki: Currently, the retention policy can only be set globally. A per-tenant retention policy with an API to delete ingested logs is still under development.

pstibrany:
> Yeah，it will reload periodly.
> 
> But update the retention duration onfly is also needed.
> 
> https://grafana.com/docs/loki/latest/operations/storage/retention/
> Loki: Currently, the retention policy can only be set globally. A per-tenant retention policy with an API to delete ingested logs is still under development.

You are right, but that is a separate problem from retention. I'd suggest focusing on retention in this design doc, with options we already have (runtime config), and leave solution for API for a different design document. (Which then also needs to discuss where to store such modified options to survive restarts).

AllenzhLi:
> You are right, but that is a separate problem from retention. I'd suggest focusing on retention in this design doc, with options we already have (runtime config), and leave solution for API for a different design document. (Which then also needs to discuss where to store such modified options to survive restarts).

I agree with you. I will update the design document as soon as possible.

docs/proposals/tenant_retention.md
AllenzhLi: I agree with you. I will update this.

docs/proposals/tenant_retention.md
AllenzhLi: I will delete the api design.

docs/proposals/tenant_retention.md
AllenzhLi: I agree with you, this will be delete.

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,58 @@
...
+When using blocks storage, Cortex stores tenant’s data in object store for long-term storage of blocks, tenant id as part of the object store path. We discover all tenants via scan the root dir of bucket.
+
+### runtime config
+Using the "overrides" mechanism (part of runtime config) already allows for per-tenant settings. See https://cortexmetrics.io/docs/configuration/arguments/#runtime-configuration-file for more details. Using it for tenant retention would fit nicely. Admin could set per-tenant retention here, and also have a single global value for tenants that don't have custom value set.
+
+## Proposal
+
+### retention component
+
+We propose to introduce an new runtime config component `retention`.
```
pracucci, line 33: I don't think this is necessary. We can just have 1 new field in the `Limits` struct (defined at `pkg/util/validation/limits.go`) which allows to specify the per-tenant retention (0 to disable it).

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,58 @@
...
+Per-tenant retention configuration will be applyed for in follow order:
+
+* tenant specified value
+* global value
+
+For each tenant, if a tenant-specific value exists, it will be used directly, otherwise, if a global value exists, then the global value will be used; If neither exists, do nothing.
+
+### Implementation
+
+The runtime config will add a retention component which include a dictionary to store per-tenant retention value and global retention value.
```
pracucci, line 46: See above. No need for a new dictionary/map. We can just reuse `Limits`, which is the way we currently implement any per-tenant config.

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,58 @@
...
+
+Compactor will check per-tanant retention value and compare to meta info of blocks. The blocks that match `maxTime < now - retention` will be marked for delete.
```
pracucci, line 48: FYI, we have a blocks cleaner component within the compactor. That could be the right place to add the retention logic.

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Compactor
```
pracucci, line 50: I'm not sure what's the purpose of this section. Is it to explain how things are working today in the compactor?

pracucci (commented):
Thanks for the follow up! The updated proposal LGTM, modulo few last comments.

docs/proposals/tenant_retention.md
AllenzhLi: You are right.  I didn't understand the runtime config correctly.

docs/proposals/tenant_retention.md
AllenzhLi: Same as above. I will reuse the `Limits`.

docs/proposals/tenant_retention.md
AllenzhLi: I agree with you. I tend to insert the logic in function `cleanUser`.

docs/proposals/tenant_retention.md
AllenzhLi: I want to explain how to implement blocks delete.

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,45 @@
...
+When using blocks storage, Cortex stores tenant’s data in object store for long-term storage of blocks, tenant id as part of the object store path. We discover all tenants via scan the root dir of bucket.
+
+### runtime config
+Using the "overrides" mechanism (part of runtime config) already allows for per-tenant settings. See [runtime-configuration-file](https://cortexmetrics.io/docs/configuration/arguments/#runtime-configuration-file) for more details. Using it for tenant retention would fit nicely. Admin could set per-tenant retention here, and also have a single global value for tenants that don't have custom value set.
+
+## Proposal
+
+### retention period field
+
+We propose to introduce just one new field `RetentionPeriodPerTenant` in the Limits struct(defined at pkg/util/validation/limits.go).
```
pracucci, line 33: All fields in the `Limits` struct are per-tenant so no need to make it explicit.

```suggestion
We propose to introduce just one new field `RetentionPeriod` in the Limits struct(defined at pkg/util/validation/limits.go).
```

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,45 @@
...
+
+`RetentionPeriodPerTenant` setting how long historical metric data retention period per-tenant. `0` is disable.
+
+As a runtime config, the settings is reload period (which defaults to 10 seconds), so we can update the settings onfly.
```
pracucci, line 37:
```suggestion
Runtime config is reloaded periodically (defaults to 10 seconds), so we can update the retention settings on-the-fly.
```

pracucci (approved):
The proposal LGTM, thanks! I left a couple of non-blocking nits.

@gouthamve @pstibrany @ranton256 (which was interested as well) could you take a look, please?

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,45 @@
...
+
+`RetentionPeriodPerTenant` setting how long historical metric data retention period per-tenant. `0` is disable.
+
+Runtime config is reloaded periodically (defaults to 10 seconds), so we can update the retention settings on-the-fly.
+
+For each tenant, if a tenant-specific *runtime_config* value exists, it will be used directly, otherwise, if a default *limits_config* value exists, then the default value will be used; If neither exists, do nothing.
+
+### Implementation
+
+A BlocksCleaner within the Compactor run periodically(which defaults to 1 hour) and the retention logic will insert into it. The logic should compare retention value to block `maxTime` and blocks that match `maxTime < now - retention` will be marked for delete.
```
pstibrany, line 43:
```suggestion
A BlocksCleaner within the Compactor run periodically (which defaults to 15 minutes) and the retention logic will insert into it. The logic should compare retention value to block `maxTime` and blocks that match `maxTime < now - retention` will be marked for delete.
```

docs/proposals/tenant_retention.md
```diff
@@ -0,0 +1,45 @@
...
+### runtime config
+Using the "overrides" mechanism (part of runtime config) already allows for per-tenant settings. See [runtime-configuration-file](https://cortexmetrics.io/docs/configuration/arguments/#runtime-configuration-file) for more details. Using it for tenant retention would fit nicely. Admin could set per-tenant retention here, and also have a single global value for tenants that don't have custom value set.
+
+## Proposal
+
+### retention period field
+
+We propose to introduce just one new field `RetentionPeriod` in the Limits struct(defined at pkg/util/validation/limits.go).
+
+`RetentionPeriodPerTenant` setting how long historical metric data retention period per-tenant. `0` is disable.
```
pstibrany, line 35: I assume `RetentionPeriod` and `RetentionPeriodPerTenant` are the same. We should use just one name here.

pstibrany (approved):
LGTM, thank you!

docs/proposals/tenant_retention.md
AllenzhLi: Yes, they are same. Both should be `RetentionPeriod`.

pracucci:
This design doc looks solid to me. Simple yet effective. 👍

ranton256 (approved):
Thanks for writing this up. I concur with other reviews, this is straightforward to address the problem reusing the existing override mechanism and Compactor component, LGTM.

PR #3639, opened 2020-12-30
Add HTTP endpoint to show overrides (#1324)

### LINKED ISSUE #1324: Idea: Add HTTP endpoint to show overrides.
Like /ring, /user_stats etc.

ethervoid:
**What this PR does**:

This PR introduces a new endpoint that would retrieve the override configuration, per tenant, that is loaded as part of the runtime configuration

Fixes #1324

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
   - Ruler: `-ruler.storage.swift.auth-version`, `-ruler.storage.swift.max-retries`, `-ruler.storage.swift.connect-timeout`, `-ruler.storage.swift.request-timeout`.
 * [ENHANCEMENT] Disabled in-memory shuffle-sharding subring cache in the store-gateway, ruler and compactor. This should reduce the memory utilisation in these services when shuffle-sharding is enabled, without introducing a significantly increase CPU utilisation. #3601
 * [ENHANCEMENT] Shuffle sharding: optimised subring generation used by shuffle sharding. #3601
-* [ENHANCEMENT] New /runtime_config endpoint that returns the defined runtime configuration in YAML format. #3639
+* [ENHANCEMENT] New /runtime_config endpoint that returns the defined runtime configuration in YAML format. The returned configuration includes overrides. #3639
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
 * [BUGFIX] Ingester: do not close idle TSDBs while blocks shipping is in progress. #3630
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
   - Ruler: `-ruler.storage.swift.auth-version`, `-ruler.storage.swift.max-retries`, `-ruler.storage.swift.connect-timeout`, `-ruler.storage.swift.request-timeout`.
 * [ENHANCEMENT] Disabled in-memory shuffle-sharding subring cache in the store-gateway, ruler and compactor. This should reduce the memory utilisation in these services when shuffle-sharding is enabled, without introducing a significantly increase CPU utilisation. #3601
 * [ENHANCEMENT] Shuffle sharding: optimised subring generation used by shuffle sharding. #3601
-* [ENHANCEMENT] New /overrides endpoint that returns the limit overrides configuration in YAML format
+* [ENHANCEMENT] New /runtime_config/overrides endpoint that returns the defined runtime configuration in YAML format. #3639
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
 * [BUGFIX] Ingester: do not close idle TSDBs while blocks shipping is in progress. #3630
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
   - Ruler: `-ruler.storage.swift.auth-version`, `-ruler.storage.swift.max-retries`, `-ruler.storage.swift.connect-timeout`, `-ruler.storage.swift.request-timeout`.
 * [ENHANCEMENT] Disabled in-memory shuffle-sharding subring cache in the store-gateway, ruler and compactor. This should reduce the memory utilisation in these services when shuffle-sharding is enabled, without introducing a significantly increase CPU utilisation. #3601
 * [ENHANCEMENT] Shuffle sharding: optimised subring generation used by shuffle sharding. #3601
-* [ENHANCEMENT] New /runtime_config/overrides endpoint that returns the defined runtime configuration in YAML format. #3639
+* [ENHANCEMENT] New /runtime_config endpoint that returns the defined runtime configuration in YAML format. #3639
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
 * [BUGFIX] Ingester: do not close idle TSDBs while blocks shipping is in progress. #3630
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
   - Ruler: `-ruler.storage.swift.auth-version`, `-ruler.storage.swift.max-retries`, `-ruler.storage.swift.connect-timeout`, `-ruler.storage.swift.request-timeout`.
 * [ENHANCEMENT] Disabled in-memory shuffle-sharding subring cache in the store-gateway, ruler and compactor. This should reduce the memory utilisation in these services when shuffle-sharding is enabled, without introducing a significantly increase CPU utilisation. #3601
 * [ENHANCEMENT] Shuffle sharding: optimised subring generation used by shuffle sharding. #3601
+* [ENHANCEMENT] New /overrides endpoint that returns the limit overrides configuration in YAML format
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
 * [BUGFIX] Ingester: do not close idle TSDBs while blocks shipping is in progress. #3630
--- a/docs/api/_index.md
+++ b/docs/api/_index.md
@@ -19,6 +19,7 @@ For the sake of clarity, in this document we have grouped API endpoints by servi
 | --- | ------- | -------- |
 | [Index page](#index-page) | _All services_ | `GET /` |
 | [Configuration](#configuration) | _All services_ | `GET /config` |
+| [Overrides](#overrides) | _All services_ | `GET /overrides` |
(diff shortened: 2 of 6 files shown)


pkg/cortex/modules.go
```diff
@@ -170,6 +170,8 @@ func (t *Cortex) initRuntimeConfig() (services.Service, error) {
 
 func (t *Cortex) initOverrides() (serv services.Service, err error) {
 	t.Overrides, err = validation.NewOverrides(t.Cfg.LimitsConfig, tenantLimitsFromRuntimeConfig(t.RuntimeConfig))
+
+	t.API.RegisterOverrides(t.RuntimeConfig)
```
pstibrany, line 174: This should be in `initRuntimeConfig` instead, after `t.RuntimeConfig` is set.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * [ENHANCEMENT] Memberlist: client can now keep a size-bounded buffer with sent and received messages and display them in the admin UI (/memberlist) for troubleshooting. #3581 #3602
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
+* [ENHANCEMENT] New /overrides endpoint that returns the limit overrides configuration in YAML format
```
pstibrany, line 24: Enhancements go before BUGFIX. Also please add PR number to the end.

pstibrany (commented):
Thanks for your PR! It looks good, I've left some comments to check before we can merge.

pkg/api/api.go
```diff
@@ -174,6 +176,13 @@ func (a *API) RegisterAPI(httpPathPrefix string, cfg interface{}) {
 	a.RegisterRoute("/debug/fgprof", fgprof.Handler(), false, "GET")
 }
 
+// RegisterOverrides registers the endpoints associates with the configuration overrides
+func (a *API) RegisterOverrides(runtimeCfgManager *runtimeconfig.Manager) {
+	a.indexPage.AddLink(SectionAdminEndpoints, "/overrides", "Current Overrides Config")
```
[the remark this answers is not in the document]
pstibrany: (Note that when issue #1324 was created, there was no "runtime config", only overrides. Which is why it doesn't mention it.)

pkg/api/handlers.go
```diff
@@ -131,6 +132,22 @@ func configHandler(cfg interface{}) http.HandlerFunc {
 	}
 }
 
+func overridesHandler(runtimeCfgManager *runtimeconfig.Manager) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
pstibrany: Please note that this has now been done in https://github.com/cortexproject/cortex/pull/3645 too. To avoid duplicate work, it may be good idea to rebase on top of recent master, once https://github.com/cortexproject/cortex/pull/3645 is merged.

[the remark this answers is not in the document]
ethervoid: Great, thank you for pointing me to the new change. I'll wait until that PR is merged to make the rebase

pkg/api/api.go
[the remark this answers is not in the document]
ethervoid: Done. Now overrides is in the path `/runtime_config/overrides`

pkg/cortex/modules.go
ethervoid: Change done

CHANGELOG.md
ethervoid: Done

ethervoid:
@pstibrany I've made the requested changes but I'm waiting for the mentioned PR to be merged to do the last one referring to the refactoring of the YAML response code

pkg/api/api.go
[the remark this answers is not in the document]
pstibrany: Sorry for misunderstanding... what I meant is to expose entire “runtime configuration” object under `/runtime_config` path. Overrides are part of “runtime configuration”, so they will be automatically included by doing that.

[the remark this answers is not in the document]
pstibrany: By **“runtime configuration” object** I mean value returned by calling `t.RuntimeConfig.GetConfig()` method.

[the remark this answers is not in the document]
ethervoid: Oh, I see. Sorry for the noise, I've made the requested changes. Now we're exposing the runtime configuration through the `/runtime_config` endpoint.

docs/api/_index.md
```diff
@@ -109,6 +110,14 @@ GET /config
 
 Displays the configuration currently applied to Cortex (in YAML format), including default values and settings via CLI flags. Sensitive data is masked. Please be aware that the exported configuration **doesn't include the per-tenant overrides**.
 
+### Runtime Configuration
+
+```
+GET /runtime_config
+```
+
+Displays the runtime configuration currently applied to Cortex (in YAML format), including default values.
```
pstibrany, line 119: Please add a note saying that if Cortex is not configured with any `-runtime-config.file`, then this will not be available.

pkg/api/api.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cortexproject/cortex/pkg/util/runtimeconfig"
```
pstibrany, line 10: Nit: this should go between cortex imports, and not be standalone.

pkg/api/api.go
```diff
@@ -174,6 +176,13 @@ func (a *API) RegisterAPI(httpPathPrefix string, cfg interface{}) {
 	a.RegisterRoute("/debug/fgprof", fgprof.Handler(), false, "GET")
 }
 
+// RegisterRuntimeConfig registers the endpoints associates with the runtime configuration
+func (a *API) RegisterRuntimeConfig(runtimeCfgManager *runtimeconfig.Manager) {
+	a.indexPage.AddLink(SectionAdminEndpoints, "/runtime_config", "Current Runtime Config")
```
pstibrany, line 181: I'd mention it includes overrides too: "Current Runtime Config (incl. Overrides)"

pkg/cortex/modules.go
```diff
@@ -165,11 +165,13 @@ func (t *Cortex) initRuntimeConfig() (services.Service, error) {
 
 	serv, err := runtimeconfig.NewRuntimeConfigManager(t.Cfg.RuntimeConfig, prometheus.DefaultRegisterer)
 	t.RuntimeConfig = serv
+	t.API.RegisterRuntimeConfig(t.RuntimeConfig)
 	return serv, err
 }
 
 func (t *Cortex) initOverrides() (serv services.Service, err error) {
 	t.Overrides, err = validation.NewOverrides(t.Cfg.LimitsConfig, tenantLimitsFromRuntimeConfig(t.RuntimeConfig))
+
```
pstibrany, line 174: nit: extra empty line added

pstibrany (commented):
Thank you, this looks good. One more detail that may be worth handling: if `runtimeCfgManager.GetConfig()` in the handler returns nil (which can happen if you define `-runtime-config.file` but file doesn't exist, then the output of `/runtime_config` is just "null". Perhaps some message like "runtime config file doesn't exist" would be more appropriate?

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 * [ENHANCEMENT] Compactor: tenants marked for deletion will now be fully cleaned up after some delay since deletion of last block. Cleanup includes removal of remaining marker files (including tenant deletion mark file) and files under `debug/metas`. #3613
 * [ENHANCEMENT] Compactor: retry compaction of a single tenant on failure instead of re-running compaction for all tenants. #3627
 * [ENHANCEMENT] Querier: Implement result caching for tenant query federation. #3640
+* [ENHANCEMENT] New /runtime_config endpoint that returns the defined runtime configuration in YAML format. #3639
```
pstibrany, line 28: I would also mention that it includes overrides here.

pstibrany:
Can we mention new `/runtime_config` endpoint in `docs/configuration/arguments.md` file ("Runtime Configuration file" section) too?

pkg/api/handlers.go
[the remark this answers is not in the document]
pracucci: I just merged #3645, so you can rebase `master` and just use that helper.

pracucci (approved):
Pretty cool! LGTM, modulo Peter's feedback.

In a **separate PR**, would be fantastic if you could also add the `mode=diff` support like in https://github.com/cortexproject/cortex/pull/3645 because it's a bit hard to read the output due to the large quantity of default valued fields. If you don't, please open an issue to leave it for someone else 🙏

pstibrany:
> In a **separate PR**, would be fantastic if you could also add the `mode=diff` support like in #3645 because it's a bit hard to read the output due to the large quantity of default valued fields. If you don't, please open an issue to leave it for someone else 🙏

This doesn’t apply to runtime config. `/runtime_config` only shows your changes (yaml file) nothing else.

ethervoid:
I"ve been checking about the diff option and looks like we're setting the default values [here](https://github.com/cortexproject/cortex/blob/2dd7692140fc1777acef736993d89211ee826bd2/pkg/cortex/modules.go#L171) to have it in the YAML unmarshal operation so I think we can add a diff option but it would be only for the `overrides` part

pstibrany:
> This doesn’t apply to runtime config. `/runtime_config` only shows your changes (yaml file) nothing else.

I am wrong about this. I tested this PR earlier today, but without actually setting any overrides 🤦 and just other parts of runtime config.

> I"ve been checking about the diff option and looks like we're setting the default values [here](https://github.com/cortexproject/cortex/blob/2dd7692140fc1777acef736993d89211ee826bd2/pkg/cortex/modules.go#L171) to have it in the YAML unmarshal operation so I think we can add a diff option but it would be only for the `overrides` part

I think diff makes sense for entire runtime config, but is mostly interesting for overrides. However it is also somewhat tricky for overrides... when diffing, each per-tenant override should be diffed against default limits individually, and not the whole structure.

As Marco suggested, let's leave that for another PR to get this merged.

pkg/api/handlers.go
```diff
@@ -221,6 +222,17 @@ func configHandler(actualCfg interface{}, defaultCfg interface{}) http.HandlerFu
 	}
 }
 
+func runtimeConfigHandler(runtimeCfgManager *runtimeconfig.Manager) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		runtimeConfig := runtimeCfgManager.GetConfig()
+		if runtimeConfig == nil {
+			http.Error(w, "runtime config file doesn't exist", http.StatusInternalServerError)
```
pstibrany, line 229: This isn’t error, it just means that overrides file is empty or doesn’t exist. Both are perfectly fine. Writing text message with code 200 is preferred.

docs/api/_index.md
```diff
@@ -123,6 +124,14 @@ GET /config?mode=defaults
 
 Displays the configuration using only the default values.
 
+### Runtime Configuration
+
+```
+GET /runtime_config
+```
+
+Displays the runtime configuration currently applied to Cortex (in YAML format), including default values. Please be aware that the configuration will be available only if Cortex is configured with a valid `-runtime-config.file`.
```
pstibrany, line 133: `-runtime-config.file` option must be set for this endpoint to be available, but actual file doesn’t need to exist or be valid. But that’s just nitpicking.

pstibrany (approved):
Thanks for patiently addressing our feedback. Let’s fix the nil case, and merge.

ethervoid:
@pstibrany I think that's all. I want to thank you for your amazing patiente guiding me and providing really valuable feedback. I've learn a lot about the the project and the way the thins are done so I can be more accurate for future PRs :-)

pstibrany:
Thank you for addressing my feedback!

PR #3640, opened 2020-12-31
Implement result cache for tenant query federation

simonswine:
**What this PR does**:

Tenant query federation PR #3250 did skip result caching for tenant query federation. This PR will implement it.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/guides/limitations.md
+++ b/docs/guides/limitations.md
@@ -43,7 +43,3 @@ The Cortex chunks storage doesn't support queries without a metric name, like `c
 ## Query series and labels
 
 When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
-
-## Tenant federation
-
-When tenant federation is enabled on a Cortex cluster, result caching is disabled for queries spanning more than a single tenant. Result caching is planned to be implemented before general availability of this feature.
--- a/pkg/chunk/purger/tombstones.go
+++ b/pkg/chunk/purger/tombstones.go
@@ -3,6 +3,7 @@ package purger
 import (
 	"context"
 	"sort"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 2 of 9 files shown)


pkg/util/validation/limits.go
```diff
@@ -474,3 +474,19 @@ func SmallestPositiveNonZeroDurationPerTenant(tenantIDs []string, f func(string)
 	}
 	return *result
 }
+
+// BiggestDurationPerTenant is returning the maximum duration
+// per tenant. Without tenants given it will return a 0 duration.
+func BiggestDurationPerTenant(tenantIDs []string, f func(string) time.Duration) time.Duration {
```
jtlisi, line 480: **suggestion(optional):** I prefer MaxDurationPerTenant

```suggestion
// MaxDurationPerTenant is returning the maximum duration
// per tenant. Without tenants given it will return a 0 duration.
func MaxDurationPerTenant(tenantIDs []string, f func(string) time.Duration) time.Duration {
```

jtlisi (changes requested):
LGTM besides some minor nits.

Also could you add a `[ENHANCEMENT]` entry to the CHANGELOG describing the change in behaviour.

docs/guides/limitations.md
```diff
@@ -43,7 +43,3 @@ The Cortex chunks storage doesn't support queries without a metric name, like `c
 ## Query series and labels
 
 When running queries to the `/api/v1/series`, `/api/v1/labels` and `/api/v1/label/{name}/values` endpoints, query's time range is ignored and the data is always fetched from ingesters. There is experimental support to query the long-term store with the *blocks* storage engine when `-querier.query-store-for-labels-enabled` is set.
-
-## Tenant federation
-
-When tenant federation is enabled on a Cortex cluster, result caching is disabled for queries spanning more than a single tenant. Result caching is planned to be implemented before general availability of this feature.
```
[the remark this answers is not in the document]
pstibrany: > **question:** Why are you removing this?

I think this PR removes the limitation.

pstibrany:
@sandeepsukhani Would you mind taking a look at this PR please?

pkg/util/validation/limits.go
```diff
@@ -474,3 +474,19 @@ func SmallestPositiveNonZeroDurationPerTenant(tenantIDs []string, f func(string)
...
+	var result *time.Duration
```
pstibrany, line 481: Duration is a simple number, and we already return 0 if we don't find any. I'd suggest simplifying this code to avoid pointers, and just start with `result := time.Duration(0)`

pstibrany (commented):
Changes look good to me, but I would prefer if @sandeepsukhani took a look as well, as he is an expert on this stuff.

pkg/chunk/purger/tombstones.go
```diff
@@ -246,14 +247,64 @@ func (tl *TombstonesLoader) loadPendingTombstones(userID string) error {
...
+			}
+
+			// set first tenant string whatever happens next
+			result.results = numbers.results
+			result.store = numbers.store
+		}
+
+		// set results number string if it's higher than the ones before
+		if numbers.results != "" {
+			results, err := strconv.Atoi(numbers.results)
```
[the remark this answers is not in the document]
sandeepsukhani: Yes, the gen numbers are actually timestamps, we simply set them to current timestamp whenever we need to change it.

sandeepsukhani (approved):
The changes look good to me.

pkg/util/validation/limits.go
simonswine: :+1: Agreed, done that

pkg/util/validation/limits.go
simonswine: :+1:  done that

docs/guides/limitations.md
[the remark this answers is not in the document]
simonswine: Yes we won't have that limitation anymore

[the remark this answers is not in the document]
jtlisi: I clearly didn't read the actual text.

jtlisi (approved):
LGTM

pstibrany (approved):

pkg/chunk/purger/tombstones.go
```diff
@@ -246,14 +247,64 @@ func (tl *TombstonesLoader) loadPendingTombstones(userID string) error {
...
+			if err != nil {
+				level.Error(util.Logger).Log("msg", "error parsing resultsCacheGenNumber", "tenant", tenantID, "err", err)
```
pracucci, line 288: Across the code base, the tenant ID is logged with the field name `user` instead of `tenant`. Do you mind fixing this for consistency, please?

_Same comment applies below._

simonswine: Good point see: https://github.com/cortexproject/cortex/pull/3680

PR #3645, opened 2021-01-05
Only output config parameters that differ from the defaults for `/config` endpoint

simonswine:
**What this PR does**:

This changes the behaviour of the config endpoint to only show the config values that have really changed towards the default config.

I personally get lost quite often in the very verbose output of the user's configuration that includes the defaults. This tries to solve that by only show values that differ from the defaults.

Fixes #<issue number>

**Checklist**
- [ ] Gather feedback if this is something useful
- [ ] Tests updated
- [ ] Documentation added


simonswine:
This is the example output for using `docs/configuration/single-process-config-blocks.yaml`

```yaml
auth_enabled: false
blocks_storage:
  backend: filesystem
  bucket_store:
    sync_dir: /tmp/cortex/tsdb-sync
  filesystem:
    dir: ./data/tsdb
  tsdb:
    dir: /tmp/cortex/tsdb
compactor:
  data_dir: /tmp/cortex/compactor
  sharding_ring:
    kvstore:
      store: inmemory
distributor:
  shard_by_all_labels: true
frontend_worker:
  match_max_concurrent: true
ingester:
  lifecycler:
    address: 127.0.0.1
    final_sleep: 0s
    min_ready_duration: 0s
    num_tokens: 512
    ring:
      kvstore:
        store: inmemory
      replication_factor: 1
ingester_client:
  grpc_client_config:
    grpc_compression: gzip
    max_send_msg_size: 104857600
ruler:
  enable_api: true
  storage:
    local:
      directory: /tmp/cortex/rules
    type: local
server:
  grpc_server_max_concurrent_streams: 1000
  grpc_server_max_recv_msg_size: 104857600
  grpc_server_max_send_msg_size: 104857600
  http_listen_port: 9009
storage:
  engine: blocks
```

pracucci:
@simonswine I see a value in this change, but what do you think about enabling this logic via a query string parameter passed to the `/config` endpoint? This way we would preserve the current behaviour while adding support to the diff only.

pkg/cortex/modules.go
```diff
@@ -80,6 +83,95 @@ const (
...
+			}
+		default:
+			return nil, fmt.Errorf("I don't know about type %T!\n", v)
+		}
+	}
+
+	return output, nil
+}
+
+func (t *Cortex) diffConfig() (map[interface{}]interface{}, error) {
```
pstibrany, line 157: I think it would be useful to have unit test for this, to make sure that `marshallCycle` or `diffConfig` don't return error in the future, when we modify our Config struct.

pkg/api/api.go
```diff
@@ -166,10 +166,12 @@ func (a *API) RegisterAlertmanager(am *alertmanager.MultitenantAlertmanager, tar
...
 
 // RegisterAPI registers the standard endpoints associated with a running Cortex.
-func (a *API) RegisterAPI(httpPathPrefix string, cfg interface{}) {
+func (a *API) RegisterAPI(httpPathPrefix string, actualCfg interface{}, defaultCfg interface{}) {
 	a.indexPage.AddLink(SectionAdminEndpoints, "/config", "Current Config")
 
-	a.RegisterRoute("/config", configHandler(cfg), false, "GET")
+	a.RegisterRoute("/config", configHandler(actualCfg), false, "GET")
+	a.RegisterRoute("/config/defaults", configHandler(defaultCfg), false, "GET")
+	a.RegisterRoute("/config/diff", configDiffHandler(actualCfg, defaultCfg), false, "GET")
```
pstibrany, line 174: Having three routes for config seems like overkill. I think single `/config` is just fine (we're also adding `/runtime_config` in another PR), with a `diff` param switching to diff view. Also I'd suggest to include it in the index page.

pkg/api/handlers.go
```diff
@@ -115,19 +117,120 @@ func indexHandler(httpPathPrefix string, content *IndexPageContent) http.Handler
...
+	}
+}
+
+func configHandler(cfg interface{}) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		marshalYAMLResponse(w, r, cfg)
+	}
+}
+
+func marshalYAMLResponse(w http.ResponseWriter, r *http.Request, object interface{}) {
```
pstibrany, line 221: Let's move this to `util` package, where we already have `WriteTextResponse` and `WriteJSONResponse`.

pkg/api/api.go
simonswine: Feedback I got from github.com/leventebalogh (who is working on Config UI) was that it would be helpful to have just the default values as well.

Does it make sense to use `mode=defaults`/`mode=diff` instead?

pstibrany: > Does it make sense to use `mode=defaults`/`mode=diff` instead?

👍 Sure, thanks!

pkg/api/handlers.go
simonswine: Done that,

**Note**: I have followed `WriteJSONResponse` and not included the error log line `level.Error(util.Logger).Log("msg", "error writing response", "err", err)`

pkg/cortex/modules.go
simonswine: Added that, totally agree, otherwise it might be quite brittle. I have splitted testing into two packages.

pkg/api/handlers.go
```diff
@@ -115,20 +116,109 @@ func indexHandler(httpPathPrefix string, content *IndexPageContent) http.Handler
 	}
 }
 
-func configHandler(cfg interface{}) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		out, err := yaml.Marshal(cfg)
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
+func yamlMarshallUnmarshall(in interface{}) (map[interface{}]interface{}, error) {
```
pstibrany, line 119: super-nit: for consistency with Go methods: `yamlMarshalUnmarshal`

pstibrany (approved):
LGTM, thank you! This is very useful!

pkg/cortex/modules.go
```diff
@@ -80,6 +81,13 @@ const (
 	All                      string = "all"
 )
 
+func newDefaultConfig() *Config {
+	defaultConfig := &Config{}
+	defaultFS := flag.NewFlagSet("", flag.PanicOnError)
+	defaultConfig.RegisterFlags(defaultFS)
```
pracucci, line 87: [non blocking, just FYI] You can achieve the same with `flagext.DefaultValues()`.

pracucci (approved):
I love it, well done! 👏

PR #3647, opened 2021-01-05
Fix bucket index sync interval config inconsistency

pracucci:
**What this PR does**:
We currently have 2 options to configure how frequently the bucket should be synced:

- `-blocks-storage.bucket-store.bucket-index.update-on-stale-interval`: used only by the querier when bucket index is enabled
- `-blocks-storage.bucket-store.sync-interval`: used by store-gateway in any case, and also by querier when bucket index is disabled

I believe this would lead to misunderstandings, so in this PR I'm proposing to remove the `update-on-stale-interval` config at all and always use `sync-interval`. The bucket index is marked as experimental, so this is not a breaking change.

_No CHANGELOG entry because bucket index was introduced after cutting `1.6.0`._

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -243,7 +243,7 @@ func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.BucketIndex.RegisterFlagsWithPrefix(f, "blocks-storage.bucket-store.bucket-index.")
 
 	f.StringVar(&cfg.SyncDir, "blocks-storage.bucket-store.sync-dir", "tsdb-sync", "Directory to store synchronized TSDB index headers.")
-	f.DurationVar(&cfg.SyncInterval, "blocks-storage.bucket-store.sync-interval", 5*time.Minute, "How frequently scan the bucket - or refresh the bucket index (if enabled) - to look for changes (new blocks shipped by ingesters and blocks deleted by retention or compaction).")
+	f.DurationVar(&cfg.SyncInterval, "blocks-storage.bucket-store.sync-interval", 5*time.Minute, "How frequently to scan the bucket, or to refresh the bucket index (if enabled), in order to look for changes (new blocks shipped by ingesters and blocks deleted by retention or compaction).")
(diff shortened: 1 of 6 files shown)


pkg/storage/tsdb/config.go
```diff
@@ -243,7 +243,7 @@ func (cfg *BucketStoreConfig) RegisterFlags(f *flag.FlagSet) {
 	cfg.BucketIndex.RegisterFlagsWithPrefix(f, "blocks-storage.bucket-store.bucket-index.")
 
 	f.StringVar(&cfg.SyncDir, "blocks-storage.bucket-store.sync-dir", "tsdb-sync", "Directory to store synchronized TSDB index headers.")
-	f.DurationVar(&cfg.SyncInterval, "blocks-storage.bucket-store.sync-interval", 5*time.Minute, "How frequently scan the bucket - or fetch the bucket index (if enabled) - to look for changes (new blocks shipped by ingesters and blocks removed by retention or compaction). 0 disables it.")
+	f.DurationVar(&cfg.SyncInterval, "blocks-storage.bucket-store.sync-interval", 5*time.Minute, "How frequently scan the bucket - or refresh the bucket index (if enabled) - to look for changes (new blocks shipped by ingesters and blocks deleted by retention or compaction).")
```
jtlisi, line 246: **nit:** Adjust the phrasing slightly and use and em dash instead of an en dash.
```suggestion
	f.DurationVar(&cfg.SyncInterval, "blocks-storage.bucket-store.sync-interval", 5*time.Minute, "How frequently to scan the bucket -- or to refresh the bucket index (if enabled) -- in order to look for changes (new blocks shipped by ingesters and blocks deleted by retention or compaction).")
```

jtlisi (approved):
LGTM besides a minor nit. Looks like there isn't an existing changelog entry for this config value so it should be good to go.

pkg/storage/tsdb/config.go
pstibrany: While I agree with you, we currently don't use `--`  anywhere else :(

pstibrany (approved):
LGTM, thanks.

ranton256 (approved):
LGTM, thanks for cleaning it up. I agree this is clearer and should be less error prone.

pkg/storage/tsdb/config.go
pracucci: I picked up the suggestion, changing `--` into `,`. Hope it works for you.

Issue #3644, opened 2021-01-05
frontend_worker.frontend_address ignored in 1.6.0


jakubgs:
# Description

I upgraded to `1.6.0` and noticed the value of `frontend_worker.frontend_address` was ignored.

# Details

If I check the config in `1.6.0` I can see it set to the IP I want:
```
 > curl -s http://10.1.33.0:9092/config | grep -A2 frontend_worker
frontend_worker:
  frontend_address: 10.1.31.152:9095
  scheduler_address: ""
```
But when I look at the connections I can see it connecting to localhost:
```
 > sudo netstat -pnt | grep 10.1.31.152:9095

 > sudo netstat -pnt | grep 127.0.0.1:9095  
tcp        0      0 127.0.0.1:48792         127.0.0.1:9095          ESTABLISHED 1510541/cortex
tcp6       0      0 127.0.0.1:9095          127.0.0.1:48792         ESTABLISHED 1510541/cortex
```
But when I downgrade to `1.5.0` it works again as expected:
```
 > sudo netstat -pnt | grep 10.1.31.152:9095
tcp        0      0 10.1.31.155:44850       10.1.31.152:9095        ESTABLISHED 519751/cortex

 > sudo netstat -pnt | grep 127.0.0.1:9095
```

jakubgs:
I checked the logs and it's clearly trying to connect to `127.0.0.1:9095`:
```
 > sudo journalctl -fu cortex | grep 'Starting querier'
caller=worker.go:112 msg="Starting querier worker connected to query-frontend" frontend=127.0.0.1:9095
```
Despite config clearly stating otherwise:
```
 > curl -s http://localhost:9092/config | grep frontend_address    
  frontend_address: 10.1.31.152:9095
```
I don't get how this is possible considering the value seems to be directly taken from the YAML:
https://github.com/cortexproject/cortex/blob/2f7d37d7368cce61ee028a27c2a4a21f9d8fe552/pkg/cortex/cortex.go#L97
https://github.com/cortexproject/cortex/blob/2f7d37d7368cce61ee028a27c2a4a21f9d8fe552/pkg/querier/worker/worker.go#L22-L23
https://github.com/cortexproject/cortex/blob/2f7d37d7368cce61ee028a27c2a4a21f9d8fe552/pkg/querier/worker/worker.go#L111-L112

jakubgs:
I tried using the CLI flag instead but it makes no difference:
```
[...]

PR #3650, opened 2021-01-06
fix default incorrectly overriding FrontendAddress

jakubgs:
The issue introduced in https://github.com/cortexproject/cortex/commit/eba13a81e73d9835a9adaf29ce6d8cd0d385c44f that causes `Worker.FrontendAddress` to be overridden by the default even when set by user.

Resolves: https://github.com/cortexproject/cortex/issues/3644

Ping: @pstibrany

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
 * [BUGFIX] Ingester: do not close idle TSDBs while blocks shipping is in progress. #3630
 * [BUGFIX] Ingester: correctly update `cortex_ingester_memory_users` and `cortex_ingester_active_series` when a tenant's idle TSDB is closed, when running Cortex with the blocks storage. #3646
+* [BUGFIX] Querier: fix default value incorrectly overriding `-querier.frontend-address` in single-binary mode. #3650
 
 ## 1.6.0
 
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -295,9 +295,9 @@ func (t *Cortex) initQuerier() (serv services.Service, err error) {
 		// and internal using the default instrumentation when running as a standalone service.
 		internalQuerierRouter = t.Server.HTTPServer.Handler
 	} else {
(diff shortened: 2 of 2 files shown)


pstibrany (approved):
LGTM, thanks!

pstibrany:
This fixes a bug. Can you please also add a CHANGELOG entry?

jakubgs:
>Can you please also add a CHANGELOG entry?

Ah, good point. There you go.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
 * [BUGFIX] Ingester: do not close idle TSDBs while blocks shipping is in progress. #3630
 * [BUGFIX] Ingester: correctly update `cortex_ingester_memory_users` and `cortex_ingester_active_series` when a tenant's idle TSDB is closed, when running Cortex with the blocks storage. #3646
+* [BUGFIX] Worker: fix default incorrectly overriding FrontendAddress. #3650
```
pstibrany, line 31:
```suggestion
* [BUGFIX] Querier: fix default value incorrectly overriding `-querier.frontend-address` in single-binary mode.  #3650
```

jakubgs: I guess I was confused by the file naming(`pkg/querier/worker/worker.go`) and config section name `frontend_worker`.

pstibrany: No worries, it *is* confusing. Querier is the top-level Cortex component, which internally runs "frontend worker" that connects to frontend component. Thanks for your changes!

jakubgs:
Is there any timeline for `1.6.1` release that would include this? Or should I just roll back to `1.5.0` for the time being?

pstibrany:
> Is there any timeline for `1.6.1` release that would include this? Or should I just roll back to `1.5.0` for the time being?

We rarely do patch-point releases. Perhaps you can use master instead, once this is merged? (PR needs another approval first) We don't have binaries, but there are docker images built from master.

jakubgs:
>We rarely do patch-point releases.

And ignoring configuration for a core component is not enough to make a patch release?

Ah well, then I guess I'll just build master. I assume `make dist` is what I want.

khaines (approved):
LGTM!

PR #3651, opened 2021-01-06
Adds support to S3 server side encryption using AWS KMS

### LINKED ISSUE #3437: Support multiple AWS S3 SSE (Server side encryption)
We have S3 buckets that support just `aws:kms` server side encryption.
I can't use Cortex with those buckets because the S3 storage client hardcodes `AES256` https://github.com/cortexproject/cortex/blob/master/pkg/chunk/aws/s3_storage_client.go#L116

When using `-ruler.storage.s3.sse-encryption` option (Ruler is just one example) I would like to specify what type of encryption to use, instead of just true or false.

**Describe alternatives you've considered**
If changing the CLI parameter's value type is not an option, then I would add another option, which is
`-ruler.storage.s3.sse-encryption-type` of type String, where values can be `AES256` or `aws:kms`

**Additional infomation**
This might affect Loki as well

lucasvmiguel:
**What this PR does**: Adds support to S3 server side encryption using KMS.

Fixes #3437

**Checklist**
- [x] Tests updated
- [X] Documentation added
- [x] Manually tests to verify if S3 SSE is working
  - [x] Data stored without SSE
  - [x] Data stored with deprecated SSE (`s3.sse-encryption=true`)
  - [x] Data stored with S3 SSE (`sse-config.type=SSE-S3`)
  - [x] Data store with S3 KMS (`sse-config.type=SSE-KMS`)

_TIP: Just to help future folks that want to use this feature (because I lost sometime to understand how I'd fetch the data back from the S3 bucket). You basically have to create the following IAM policy and attach to an IAM role._
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "kms:Decrypt",
                "s3:GetObject"
            ],
            "Resource": [
                "<AWS KMS KEY ARN>",
                "<AWS S3 BUCKET ARN>/*"
            ]
        }
    ]
}
```


pkg/chunk/aws/s3_storage_client.go
```diff
@@ -100,7 +108,10 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.AccessKeyID, prefix+"s3.access-key-id", "", "AWS Access Key ID")
 	f.StringVar(&cfg.SecretAccessKey, prefix+"s3.secret-access-key", "", "AWS Secret Access Key")
 	f.BoolVar(&cfg.Insecure, prefix+"s3.insecure", false, "Disable https on s3 connection.")
-	f.BoolVar(&cfg.SSEEncryption, prefix+"s3.sse-encryption", false, "Enable AES256 AWS Server Side Encryption")
+	f.BoolVar(&cfg.SSEEncryption, prefix+"s3.sse-encryption", false, "Enable AWS Server Side Encryption [Deprecated: Use sse_config instead. if s3.sse-encryption is enabled, it assumes sse_config SSE-S3 type]")
+	f.StringVar(&cfg.SSEConfig.Type, prefix+"s3.sse_config.type", "", "Enable AWS Server Side Encryption. Only SSE-S3 and SSE-KMS are supported")
+	f.StringVar(&cfg.SSEConfig.KMSKeyID, prefix+"s3.sse_config.kms_key_id", "", "KMS Key ID used to encrypt objects in S3")
+	f.Var(&cfg.SSEConfig.KMSEncryptionContext, prefix+"s3.sse_config.kms_encryption_context", "KMS Encryption Context used for object encryption")
```
pracucci, line 114: CLI flags are `-` separated, while YAML config options are `_` separated.

```suggestion
	f.StringVar(&cfg.SSEConfig.KMSKeyID, prefix+"s3.sse_config.kms-key-id", "", "KMS Key ID used to encrypt objects in S3")
	f.Var(&cfg.SSEConfig.KMSEncryptionContext, prefix+"s3.sse_config.kms-encryption-context", "KMS Encryption Context used for object encryption")
```

pracucci (commented):
Thanks for working on this! I haven't review it properly. Just leaving a comment about a little thing I've noticed and then will do a proper review.

lucasvmiguel:
**Manual tests performed**

|  bucket name   | params  | result  |
|---|---|---|
| cortex-test-sse | none | stored without SSE |
| cortex-test-sse-2  | s3.sse-encryption=true | stored encrypted data with SSE-S3 |
| cortex-test-sse-3 | s3.sse-config.type=SSE-S3 | stored encrypted data with SSE-S3 |
| cortex-test-sse-4 |  s3.sse-config.type=SSE-KMS and s3.sse-config.kms-key-id=<KEY> | stored encrypted data with SSE-KMS
  |

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -100,7 +108,10 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.AccessKeyID, prefix+"s3.access-key-id", "", "AWS Access Key ID")
 	f.StringVar(&cfg.SecretAccessKey, prefix+"s3.secret-access-key", "", "AWS Secret Access Key")
 	f.BoolVar(&cfg.Insecure, prefix+"s3.insecure", false, "Disable https on s3 connection.")
-	f.BoolVar(&cfg.SSEEncryption, prefix+"s3.sse-encryption", false, "Enable AES256 AWS Server Side Encryption")
```
pracucci, line 103: Please mention this deprecation in the CHANGELOG.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -100,7 +108,10 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
...
+	f.BoolVar(&cfg.SSEEncryption, prefix+"s3.sse-encryption", false, "Enable AWS Server Side Encryption [Deprecated: Use sse-config instead. if s3.sse-encryption is enabled, it assumes sse-config SSE-S3 type]")
+	f.StringVar(&cfg.SSEConfig.Type, prefix+"s3.sse-config.type", "", "Enable AWS Server Side Encryption. Only SSE-S3 and SSE-KMS are supported")
```
pracucci, line 112: We're used to define `RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet)` on the receiver config. I would suggest you to do the same on `SSEConfig`, passing `prefix+"s3.sse."` as prefix to `SSEConfig.RegisterFlagsWithPrefix()`.

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -100,7 +108,10 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.AccessKeyID, prefix+"s3.access-key-id", "", "AWS Access Key ID")
 	f.StringVar(&cfg.SecretAccessKey, prefix+"s3.secret-access-key", "", "AWS Secret Access Key")
 	f.BoolVar(&cfg.Insecure, prefix+"s3.insecure", false, "Disable https on s3 connection.")
-	f.BoolVar(&cfg.SSEEncryption, prefix+"s3.sse-encryption", false, "Enable AES256 AWS Server Side Encryption")
+	f.BoolVar(&cfg.SSEEncryption, prefix+"s3.sse-encryption", false, "Enable AWS Server Side Encryption [Deprecated: Use sse-config instead. if s3.sse-encryption is enabled, it assumes sse-config SSE-S3 type]")
```
pracucci, line 111: Please add also a `// TODO Remove in Cortex 1.9.0` (because we guarantee deprecated config to be removed at least after 2 minor releases).

pkg/chunk/aws/sse_config.go
```diff
@@ -0,0 +1,68 @@
...
+	SSEKMS     = "SSE-KMS"
+	sseKMSType = "aws:kms"
+	// SSES3 config type constant to configure S3 server side encryption with AES-256
+	// https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html
+	SSES3     = "SSE-S3"
+	sseS3Type = "AES256"
+)
+
+// SSEEncryptionConfig configures server side encryption (SSE)
+type SSEEncryptionConfig struct {
```
pracucci, line 22: Why do we need this when we already have `SSEConfig`?

pkg/chunk/aws/sse_config.go
```diff
@@ -0,0 +1,68 @@
...
+	ServerSideEncryption string
+	KMSKeyID             *string
+	KMSEncryptionContext *string
```
pracucci, line 25: I'm not a big fan of pointers in config structs. You can achieve the same assuming empty string means not configured (which should be fine in this context).

pkg/chunk/aws/s3_storage_client.go
```diff
@@ -72,6 +72,7 @@ type S3Config struct {
 	SSEEncryption    bool       `yaml:"sse_encryption"`
 	HTTPConfig       HTTPConfig `yaml:"http_config"`
 	SignatureVersion string     `yaml:"signature_version"`
+	SSEConfig        SSEConfig  `yaml:"sse_config"`
```
pracucci, line 75: In recent times we stopped appending `_config` to sub-config structs cause it's a bit redundant. I would remove it from here too.

```suggestion
	SSEConfig        SSEConfig  `yaml:"sse"`
```

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## master / unreleased
 
+
+* [FEATURE] Adds support to S3 server side encryption using KMS. The following config fields have been added. #3651
```
pracucci, line 6: We're used to group CHANGELOG entries together by type. Please move this entry below, under the group of `[FEATURE]`.

pracucci (commented):
Good job @lucasvmiguel and sorry for the extremely delay in the review! Overall LGTM, just left few comments to polish it a bit before merging. Thanks!

pkg/chunk/aws/s3_storage_client.go
lucasvmiguel: If we decide to not use `_config`, we won't be compatible with [Thanos config](https://thanos.io/tip/thanos/storage.md/#s3).

I don't mind to remove `_config`, I just thought that we wanted the config interface exactly the same as Thanos. (to not "break" any future migration)

pkg/chunk/aws/sse_config.go
lucasvmiguel: AWS lib uses string pointers several times in there. Removing string pointers from `sse_config.go (SSEEncryptionConfig)` means that I will have to treat that on the file `s3_storage_client.go`, which has more logic that should have.

pkg/chunk/aws/s3_storage_client.go
pracucci: We're not 1:1 compatible with Thanos even in the blocks storage. It's perfectly fine. We just wanna let configure the same things, so that when we'll add KMS support to blocks storage too we can keep the same config struct.

pkg/chunk/aws/sse_config.go
lucasvmiguel: `SSEConfig` is the struct responsible to receive the config from the CLI or a config file. After `SSEConfig` is filled with the values provided by the user, a `SSEEncryptionConfig` struct will be created to be worked inside the application. In the end, `SSEEncryptionConfig` will have different values than `SSEConfig`.

CHANGELOG.md
```diff
@@ -24,13 +24,17 @@
   * `-cluster.peer-timeout` in favor of `-alertmanager.cluster.peer-timeout`
 * [CHANGE] Blocks storage: the default value of `-blocks-storage.bucket-store.sync-interval` has been changed from `5m` to `15m`. #3724
 * [FEATURE] Querier: Queries can be federated across multiple tenants. The tenants IDs involved need to be specified separated by a `|` character in the `X-Scope-OrgID` request header. This is an experimental feature, which can be enabled by setting `-tenant-federation.enabled=true` on all Cortex services. #3250
-* [ENHANCEMENT] Allow specifying JAEGER_ENDPOINT instead of sampling server or local agent port. #3682
+* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `<prefix>.s3.sse-encryption`, you should use the following config fields that have been added. #3651
```
pracucci, line 27: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

CHANGELOG.md
```diff
@@ -24,13 +24,17 @@
...
+  - `<prefix>.s3.sse.type`
+  - `<prefix>.s3.sse.kms-key-id`
+  - `<prefix>.s3.sse.kms-encryption-context`
 * [FEATURE] Alertmanager: introduced the experimental option `-alertmanager.sharding-enabled` to shard tenants across multiple Alertmanager instances. This feature is still under heavy development and its usage is discouraged. The following new metrics are exported by the Alertmanager: #3664
   * `cortex_alertmanager_ring_check_errors_total`
   * `cortex_alertmanager_sync_configs_total`
   * `cortex_alertmanager_sync_configs_failed_total`
   * `cortex_alertmanager_tenants_discovered`
   * `cortex_alertmanager_tenants_owned`
+* [ENHANCEMENT] Allow specifying JAEGER_ENDPOINT instead of sampling server or local agent port. #3682
```
pracucci, line 37: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master / unreleased
 
+* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `<prefix>.s3.sse-encryption`, you should use the following config fields that have been added. #3651
```
```suggestion
* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  - `<prefix>.s3.sse.type`
```
pracucci, line 6:
```suggestion
  - `-<prefix>.s3.sse.type`
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  - `<prefix>.s3.sse.kms-key-id`
```
pracucci, line 7:
```suggestion
  - `-<prefix>.s3.sse.kms-key-id`
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  - `<prefix>.s3.sse.kms-encryption-context`
```
pracucci, line 8:
```suggestion
  - `-<prefix>.s3.sse.kms-encryption-context`
```

pkg/chunk/aws/sse_config.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// SSEConfig configures S3 server side encryption
+// struct that is going to receive user input (through config file or CLI)
+type SSEConfig struct {
+	Type                 string     `yaml:"type"`
+	KMSKeyID             string     `yaml:"kms_key_id"`
+	KMSEncryptionContext chunk.Tags `yaml:"kms_encryption_context"`
```
pracucci, line 37: I don't think we should use `chunk.Tags` here. I would simplify it just taking in input the context json as a `string`. Moreover, I would clarify in the CLI flag description that it expects a JSON as a string.

pkg/chunk/aws/sse_config_test.go
```diff
@@ -0,0 +1,101 @@
...
+				kmsEncryptionContext: kmsEncryptionContext,
+			},
+			expected: &SSEParsedConfig{
+				ServerSideEncryption: sseKMSType,
+				KMSKeyID:             &kmsKeyID,
+				KMSEncryptionContext: &parsedKMSEncryptionContext,
+			},
+		},
+		{
+			name: "Test invalid SEE type",
```
pracucci, line 76:
```suggestion
			name: "Test invalid SSE type",
```

pkg/chunk/aws/sse_config_test.go
```diff
@@ -0,0 +1,101 @@
...
+				kmsEncryptionContext: kmsEncryptionContext,
+			},
+			expected: &SSEParsedConfig{
+				ServerSideEncryption: sseKMSType,
+				KMSKeyID:             &kmsKeyID,
+				KMSEncryptionContext: &parsedKMSEncryptionContext,
+			},
+		},
+		{
+			name: "Test SSE encryption with SSEKMS type with context",
```
pracucci, line 63: This test is duplicated from above.

pracucci (commented):
Thanks for the follow up! We're very close! I left last comments.

Please mention that S3 server-side encryption is experimental in `docs/configuration/v1-guarantees.md` (there's a list of experimental features). This will allow us to introduce config breaking changes in the near future if we find out any issue (or improvement).

pkg/chunk/aws/sse_config.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, errors.New("SSE type is empty or invalid")
+	}
+}
+
+func parseKMSEncryptionContext(kmsEncryptionContext string) (*string, error) {
+	if kmsEncryptionContext == "" {
+		return nil, nil
+	}
+
+	jsonKMSEncryptionContext, err := json.Marshal(json.RawMessage(kmsEncryptionContext))
```
pracucci, line 77: If `kmsEncryptionContext` is already a JSON as a string, why do we need to re-marshal it? Can't we just use it as is and base64 encode it?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master / unreleased
 
+* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651
```
pracucci, line 5: The order of entries should be [CHANGE], [FEATURE], [ENHANCEMENT], [BUGFIX]. Please move this entry after `[CHANGE]` below.

pracucci (commented):
Thanks for the follow up! I have a last question about context parsing.

pkg/chunk/aws/sse_config.go
lucasvmiguel: Using `json.Marshal`, I can validate if the string is a valid `JSON`. Giving the user a better error message in case they pass an invalid `JSON`.

pracucci: Ok, fair enough. Please add a code comment to explain it.

pracucci (approved):

pkg/chunk/aws/sse_config.go
```diff
@@ -0,0 +1,85 @@
...
+
+// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
+func (cfg *SSEConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.StringVar(&cfg.Type, prefix+"type", "", "Enable AWS Server Side Encryption. Only SSE-S3 and SSE-KMS are supported")
+	f.StringVar(&cfg.KMSKeyID, prefix+"kms-key-id", "", "KMS Key ID used to encrypt objects in S3")
+	f.StringVar(&cfg.KMSEncryptionContext, prefix+"kms-encryption-context", "", "KMS Encryption Context used for object encryption. It expects a JSON as a string.")
+}
+
+// NewSSEParsedConfig creates a struct to configure server side encryption (SSE)
+func NewSSEParsedConfig(sseType string, kmsKeyID string, kmsEncryptionContext string) (*SSEParsedConfig, error) {
```
pstibrany, line 46: Nit: Perhaps this should have been a method on `SSEConfig` or take it as a parameter?

pkg/chunk/aws/sse_config.go
pstibrany: Can we also add a unit test with invalid JSON?

pkg/chunk/aws/sse_config_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewSSEParsedConfig(t *testing.T) {
+	kmsKeyID := "test"
+	kmsEncryptionContext := `{"a": "bc", "b": "cd"}`
+	parsedKMSEncryptionContext := "eyJhIjoiYmMiLCJiIjoiY2QifQ=="
```
pstibrany, line 13:
```suggestion
    parsedKMSEncryptionContext = base64.StdEncoding.EncodeToString([]byte(`{"a":"bc","b":"cd"}`)) // compact form of kmsEncryptionContext
````

pstibrany (approved):
LGTM, thank you! Please take a look at final comments before merging.

PR #3652, opened 2021-01-06
Put trace span from QueryStream under Select

bboreham:
Illustration of the problem lots of spans apparently at the same level:

image

This PR puts `Distributor.QueryStream` under `distributorQuerier.Select`.

`distributorQuerier.Select` should really go under `querier.Select`, but that's harder to fix because we are using a Prometheus interface that doesn't have a Context parameter.

**Checklist**
- NA Tests updated
- NA Documentation added
- NA `CHANGELOG.md` updated


pstibrany (approved):
👍

Issue #3648, opened 2021-01-05
Frontend query partition can generate 50x queries


bboreham:
**Describe the bug**
From tracing, I see many times today: the input to `handleHit` is a simple query:

```
query  "sum(core_non_notified_transaction_total > 0)" 
start  "2021-01-05 15:42:30 +0000 UTC"
end    "2021-01-05 16:42:30 +0000 UTC" 
step   15000
```

but `handleHit` issues about 50 calls to queriers:
```
start  "2021-01-05 15:37:30 +0000 UTC" end  "2021-01-05 15:43:00 +0000 UTC"
start  "2021-01-05 15:38:15 +0000 UTC" end  "2021-01-05 15:44:00 +0000 UTC" 
start  "2021-01-05 15:50:15 +0000 UTC  end  "2021-01-05 15:56:00 +0000 UTC" 
start  "2021-01-05 15:44:15 +0000 UTC" end  "2021-01-05 15:50:00 +0000 UTC" 
start  "2021-01-05 15:45:15 +0000 UTC" end  "2021-01-05 15:51:00 +0000 UTC" 
...
```

This is expensive.

**To Reproduce**
No idea.
Looking at [`partition()`](https://github.com/cortexproject/cortex/blob/6614def85fd42fc526cf934809d046ce26a7e4e3/pkg/querier/queryrange/results_cache.go#L419), I suspect it is getting several partial results from the cache, in some arbitrary order.

Instead, the code should (a) ensure that `extents` is sorted before scanning it and (b) coalesce sub-queries where the gap is short.

**Expected behavior**
A single sub-query, since the total extent is an hour we would expect almost all the same chunks to be used in any split.

**Environment:**
Cortex 1.6.0

**Storage Engine**
- [ ] Blocks
- [x] Chunks

bboreham:
I dumped out the `extents`; they are sorted, but somehow have an extraordinary pattern of gaps:

```
queryrange.Extent{Start:1609918515, End:1609918515, 
queryrange.Extent{Start:1609918560, End:1609918575, 
queryrange.Extent{Start:1609918620, End:1609918635, 
queryrange.Extent{Start:1609918680, End:1609918695, 
queryrange.Extent{Start:1609918740, End:1609918755, 
queryrange.Extent{Start:1609918800, End:1609918815, 
queryrange.Extent{Start:1609918860, End:1609918875, 
queryrange.Extent{Start:1609918920, End:1609918935, 
[...]

PR #3653, opened 2021-01-06
Query-frontend: discard cache extents under 5 minutes

bboreham:
**What this PR does**:
Quick but effective workaround for the case where cache contains many small extents, leading to tiny subqueries which are expensive.

The minimum allowed is passed into `partition()` so that the unit test doesn't need extensive changes.

Also added a couple of comments.

It's not really a fix, but it cures the symptom of #3648

**Checklist**
- [x] Tests updated
- NA Documentation added

### DIFF
--- a/pkg/querier/queryrange/results_cache.go
+++ b/pkg/querier/queryrange/results_cache.go
@@ -412,6 +412,7 @@ func toExtent(ctx context.Context, req Request, res Response) (Extent, error) {
 }
 
 // partition calculates the required requests to satisfy req given the cached data.
+// extents must be in order by start time.
 func partition(req Request, extents []Extent, extractor Extractor, minCacheExtent int64) ([]Request, []Response, error) {
 	var requests []Request
 	var cachedResponses []Response
@@ -441,6 +442,7 @@ func partition(req Request, extents []Extent, extractor Extractor, minCacheExten
 		start = extent.End
 	}
 
+	// Lastly, make a request for any data missing at the end.
 	if start < req.GetEnd() {
 		r := req.WithStartEnd(start, req.GetEnd())
 		requests = append(requests, r)
--- a/pkg/querier/queryrange/results_cache.go
+++ b/pkg/querier/queryrange/results_cache.go
@@ -296,7 +296,8 @@ func (s resultsCache) handleHit(ctx context.Context, r Request, extents []Extent
 	log, ctx := spanlogger.New(ctx, "handleHit")
 	defer log.Finish()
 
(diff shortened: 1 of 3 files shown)


pkg/querier/queryrange/results_cache.go
```diff
@@ -301,7 +301,8 @@ func (s resultsCache) handleHit(ctx context.Context, r Request, extents []Extent
 	log, ctx := spanlogger.New(ctx, "handleHit")
 	defer log.Finish()
 
-	requests, responses, err := partition(r, extents, s.extractor)
+	const minCacheExtent = 5 * 60 * 1000 // 5 minutes
```
tomwilkie, line 304: Does it make any sense to make this configurable? (I leaning toward "no")

pkg/querier/queryrange/results_cache.go
```diff
@@ -416,7 +417,8 @@ func toExtent(ctx context.Context, req Request, res Response) (Extent, error) {
 }
 
 // partition calculates the required requests to satisfy req given the cached data.
-func partition(req Request, extents []Extent, extractor Extractor) ([]Request, []Response, error) {
+// extents must be in order by start time.
+func partition(req Request, extents []Extent, extractor Extractor, minCacheExtent int64) ([]Request, []Response, error) {
```
tomwilkie, line 421: Can you make this a function on `(s resultsCache)` so we don't have to pass `extractor`?

And if `minCacheExtent`, can we stick it in this function / in the package so we don't have so many args?

edit: I see its for testing.  Can we make it a field?   I just think so many args is a bit of smell.

tomwilkie (approved):
Couple of min nits, otherwise LGTM.

Is the test fail a flake?  Can we get it to be green before merge?

bboreham:
The test fail is a repeatable crash in `TestResultsCacheMaxFreshness()`.  

PS: why are CI tests taking 20 minutes?

pkg/querier/queryrange/results_cache.go
pracucci: I would use `(5 *time.Minute).Milliseconds()` to make it more readable.

pracucci: > Does it make any sense to make this configurable?

I would start hardcoded and eventually make it configurable if a good use case arise for that.

pracucci (approved):

bboreham:
I have applied your suggestions.

pracucci (approved):
LGTM! 👏

PR #3654, opened 2021-01-06
Make distributor rate-limit return 5xx

### LINKED ISSUE #2037: Cortex should tell remote write clients to slow down on rate-limiting

bboreham:
This means the client will retry so the data isn't dropped.

I added a new metric for requests rejected in this way, because `cortex_discarded_samples_total` doesn't describe this case - we don't expect it to be discarded forever.

Still to do:
 * put a limit on how much this can happen, so clients don't build up data indefinitely. I left the 'hard' case in the code to allow for this enhancement.
 *  update tests to expect the different behaviour

Fixes #2037

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -592,11 +592,18 @@ func (d *Distributor) Push(ctx context.Context, req *ingester_client.WriteReques
 		// Ensure the request slice is reused if the request is rate limited.
 		ingester_client.ReuseSlice(req.Timeseries)
 
-		// Return a 4xx here to have the client discard the data and not retry. If a client
-		// is sending too much data consistently we will unlikely ever catch up otherwise.
-		validation.DiscardedSamples.WithLabelValues(validation.RateLimited, userID).Add(float64(validatedSamples))
-		validation.DiscardedMetadata.WithLabelValues(validation.RateLimited, userID).Add(float64(len(validatedMetadata)))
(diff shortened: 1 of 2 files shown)


jtlisi (commented):
I'm not sure we should return a 5xx for a rate limited sample. The downside of having Prometheus retry rate limited samples is that it will keep retrying writes and start to fall behind. If actions not taken, Prometheus could fall far enough behind that no writes will succeed since the data is outside the valid time window. This can lead to a more significant loss of data than rate limiting on its own. We have ran into this scenario a number of times and it can be very difficult to recover.

bboreham:
Data outside the valid time window will get a 4xx error hence be dropped; however I touched on your point under "still to do".

bboreham:
At the Cortex community meeting it was agreed that it would be better to do this inside Prometheus, have it react to 429s by retransmitting up to a point, or by dropping the data if it is way behind.

bboreham:
As of Prometheus 2.25, retry on 429 can be configured at the sending side https://github.com/prometheus/prometheus/pull/8477

Combined with #3833 we don't need this PR any more.

PR #3664, opened 2021-01-08
Alertmanager: Allow sharding of alertmanager tenants

gotjosh:
**What this PR does**:

The first part of the proposed as part of https://github.com/cortexproject/cortex/pull/3574, introduces sharding via the ring for the Alertmanager component.

I have a few things to test and a couple of doubts, I'm going to keep a checklist here for major visibility:

- [x] Do we need to mark the flag as experimental? No.
- [x] Do I need a changelog entry now? Yes.
- [ ] How does this behave with the current clustering logic, perhaps disabling clustering while sharding is the way to go until we're done.
- [x] Add an integration test for both the sharding and sharding + clustering.
- [x] Do I need any additional documentation now or can that be added later on?

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/alertmanager/multitenant.go
```diff
@@ -22,23 +23,28 @@ import (
...
 	"github.com/cortexproject/cortex/pkg/alertmanager/alerts"
+	"github.com/cortexproject/cortex/pkg/ring"
+	"github.com/cortexproject/cortex/pkg/ring/kv"
 	"github.com/cortexproject/cortex/pkg/tenant"
 	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/cortexproject/cortex/pkg/util/services"
 )
 
-var backoffConfig = util.BackoffConfig{
```
gotjosh, line 31: I think this might be effectively a breaking change. Before, we would wait for all configurations to load indefinitely when starting up, whereas now we wouldn't and simply abort if we fail to do the initial sync. 

I find the new behaviour a bit better as it is clearer whenever the service would start or not but I'm open to reverting this back if we think this might be too complex to communicate back to our users.

pkg/alertmanager/multitenant.go
```diff
@@ -181,6 +202,12 @@ type MultitenantAlertmanager struct {
 	multitenantMetrics  *multitenantAlertmanagerMetrics
 
 	peer *cluster.Peer
+
+	registry          prometheus.Registerer
+	ringCheckErrors   prometheus.Counter
+	tenantsOwned      prometheus.Gauge
+	tenantsDiscovered prometheus.Gauge
+	SyncTotal         *prometheus.CounterVec
```
pstibrany, line 210: Why is this public?

pkg/alertmanager/multitenant.go
```diff
@@ -250,29 +315,150 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
...
+		defer ringTicker.Stop()
+		ringTickerChan = ringTicker.C
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-am.subservicesWatcher.Chan():
+			return errors.Wrap(err, "alertmnanager subservices failed")
```
pstibrany, line 432: nit: typo in the "alertmanager"

pkg/alertmanager/multitenant.go
```diff
@@ -283,45 +469,71 @@ func (am *MultitenantAlertmanager) stopping(_ error) error {
...
+
+	am.tenantsDiscovered.Set(float64(len(configs)))
+	am.tenantsOwned.Set(float64(len(ownedConfigs)))
+	return ownedConfigs, nil
 }
 
-// poll the alert store. Not re-entrant.
-func (am *MultitenantAlertmanager) poll() (map[string]alerts.AlertConfigDesc, error) {
-	cfgs, err := am.store.ListAlertConfigs(context.Background())
+func isConfigOwner(userID string, c alerts.AlertConfigDesc, r ring.ReadRing, instanceAddr string, log log.Logger) (bool, error) {
```
pstibrany, line 525: There are multiple unused parameters to this function.

pkg/ring/model_test.go
```diff
@@ -63,6 +63,54 @@ func TestIngesterDesc_IsHealthy_ForIngesterOperations(t *testing.T) {
 	}
 }
 
+func TestIngesterDesc_IsHealthy_ForStoreAlertmanagerOperations(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		instance *IngesterDesc
+		timeout  time.Duration
```
pstibrany, line 71: nit: timeout is set to Minute in all tests. Doesn't need to be a variable.

pstibrany (approved):
LGTM. This follows the same design as ruler, except we only shard by userID, so shuffle-sharding is not necessary.

codesome (commented):
Can we also add `/alertmanager/ring` admin endpoint for this like other rings, please?

gotjosh:
@pstibrany comments addressed. Also renamed a few flags/config to match the most recent similar ones.

@codesome done!

pstibrany:
I don't see a way to retract my approval since it's still a draft, but I plan to recheck this PR in light of recent issues with default configuration (somewhat related to #3679). That is, what happens when alertmanager no longer owns the tenant, but it receives HTTP calls for it – it should definitely *not* upload blank config file to the store. I want to double check that.

pstibrany:
Also, please rebase to use changed ring operation code -- ring operation should now be defined directly in alertmanager package.

pkg/alertmanager/alertmanager_ring.go
```diff
@@ -0,0 +1,108 @@
...
+	cfg.KVStore.RegisterFlagsWithPrefix(rfprefix, "alertmanagers/", f)
+	f.DurationVar(&cfg.HeartbeatPeriod, rfprefix+"heartbeat-period", 15*time.Second, "Period at which to heartbeat to the ring.")
+	f.DurationVar(&cfg.HeartbeatTimeout, rfprefix+"heartbeat-timeout", time.Minute, "The heartbeat timeout after which alertmanagers are considered unhealthy within the ring.")
+	f.IntVar(&cfg.ReplicationFactor, rfprefix+"replication-factor", 3, "The replication factor to use when sharding the alertmanager.")
+
+	// Instance flags
+	cfg.InstanceInterfaceNames = []string{"eth0", "en0"}
+	f.Var((*flagext.StringSlice)(&cfg.InstanceInterfaceNames), rfprefix+"instance-interface-names", "Name of network interface to read address from.")
+	f.StringVar(&cfg.InstanceAddr, rfprefix+"instance-addr", "", "IP address to advertise in the ring.")
+	f.IntVar(&cfg.InstancePort, rfprefix+"instance-port", 0, "Port to advertise in the ring (defaults to server.grpc-listen-port).")
```
codesome, line 74: We are not using http for alertmanager, so should it default to `server.http-listen-port`?

codesome (commented):
I just noticed when trying this PR that we are not setting the port in the ring config before getting the lifecycler config. Hence setting port `0` in the ring for AM's entry.

pkg/alertmanager/multitenant.go
pracucci: I don't think this is a breaking change. Changes like this happen frequently and, to me, are just enhnancements.

pkg/alertmanager/multitenant.go
```diff
@@ -22,23 +23,28 @@ import (
...
-}
-
 const (
 	// If a config sets the webhook URL to this, it will be rewritten to
 	// a URL derived from Config.AutoWebhookRoot
 	autoWebhookURL = "http://internal.monitor"
 
+	// Reasons for (re)syncing alertmanager configurations from object storage.
+	reasonPeriodic   = "initial"
+	reasonInitial    = "periodic"
```
pracucci, line 41: Values are inverted:

```suggestion
	reasonPeriodic   = "periodic"
	reasonInitial    = "initial"
```

pkg/alertmanager/multitenant.go
```diff
@@ -236,10 +263,48 @@ func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, logger log.L
 		return nil, err
 	}
 
-	return createMultitenantAlertmanager(cfg, fallbackConfig, peer, store, logger, registerer), nil
+	ringStore, err := kv.NewClient(
```
pracucci, line 266: Should be created only if sharding is enabled.

pkg/alertmanager/multitenant.go
```diff
@@ -236,10 +263,48 @@ func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, logger log.L
...
+		kv.RegistererWithKVName(registerer, "alertmanager"),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "create KV store client")
+	}
+
+	return createMultitenantAlertmanager(cfg, fallbackConfig, peer, store, ringStore, logger, registerer)
+}
+
+func enableSharding(am *MultitenantAlertmanager, ringStore kv.Client) error {
```
pracucci, line 278: [nit] I would move this function inline. It's called only in 1 place.

pkg/alertmanager/replication_strategy.go
```diff
@@ -0,0 +1,37 @@
...
+var (
+	RingOp = ring.NewOp([]ring.IngesterState{ring.ACTIVE}, func(s ring.IngesterState) bool {
+		// Only ACTIVE Alertmanager get requests. If instance is not ACTIVE, we need to find another Alertmanager.
+		return s != ring.ACTIVE
+	})
+)
+
+type alertmanagerReplicationStrategy struct{}
+
+func (r alertmanagerReplicationStrategy) Filter(instances []ring.IngesterDesc, op ring.Operation, _ int, heartbeatTimeout time.Duration, _ bool) (healthy []ring.IngesterDesc, maxFailures int, err error) {
```
pracucci, line 20: The same exact replication strategy is now used by store-gateway, ruler and alertmanager. I would create a `IgnoreUnhealthyInstancesReplicationStrategy` in the `pkg/ring` and replace it everywhere to avoid such duplicated code.

pkg/cortex/modules.go
```diff
@@ -685,6 +685,8 @@ func (t *Cortex) initConfig() (serv services.Service, err error) {
 }
 
 func (t *Cortex) initAlertManager() (serv services.Service, err error) {
+	t.Cfg.Alertmanager.Ring.ListenPort = t.Cfg.Server.HTTPListenPort
```
pracucci, line 688: This is the port advertised in the ring. Given the ring is just used internally, shouldn't be the GRPC port like all other services? Why do we need to advertise the HTTP port?

pkg/alertmanager/multitenant.go
```diff
@@ -250,29 +315,150 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
 		peer:                peer,
 		store:               store,
 		logger:              log.With(logger, "component", "MultiTenantAlertmanager"),
+		registry:            registerer,
+		ringCheckErrors: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
+			Name: "cortex_alertmanager_ring_check_errors_total",
```
pracucci, line 320: You need to document all introduced metrics in the `CHANGELOG`.

pracucci (commented):
> Do I need a changelog entry now?

I would say yes.

>  Do we need to mark the flag as experimental?

Not the flag (we don't do anymore because it's pain once the feature moves to stable), but you should mention the feature is experimental in `docs/configuration/v1-guarantees.md`.

pracucci (commented):
I did another pass and left more comments (mostly minor nits). I haven't reviewed tests yet: will do it in the next pass, but I would prefer to see these comments addressed first.

pkg/alertmanager/alertmanager_ring.go
```diff
@@ -0,0 +1,108 @@
...
+	// in order to simplify the config.
+	RingNumTokens = 128
+)
+
+// RingConfig masks the ring lifecycler config which contains
+// many options not really required by the alertmanager ring. This config
+// is used to strip down the config to the minimum, and avoid confusion
+// to the user.
+type RingConfig struct {
+	KVStore           kv.Config     `yaml:"kvstore" doc:"description=The key-value store used to share the hash ring across multiple instances. This option needs be set when running in microservices mode."`
```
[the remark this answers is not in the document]
gotjosh: Good catch, it was a bad a copy / paste.

pkg/alertmanager/multitenant.go
```diff
@@ -250,29 +315,150 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
...
+				ringLastState = currRingState
+				if err := am.iteration(ctx, reasonRingChange); err != nil {
+					level.Warn(am.logger).Log("msg", "error updating configs", "err", err)
+				}
+			}
+		}
+	}
+}
+
+func (am *MultitenantAlertmanager) iteration(ctx context.Context, syncReason string) error {
```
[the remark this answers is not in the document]
gotjosh: You're right, iteration made more sense when I was using a timed service - under this new structure the latter is clearer.

pkg/alertmanager/multitenant.go
```diff
@@ -250,29 +315,150 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
...
+		level.Info(am.logger).Log("msg", "alertmanager is ACTIVE in the ring")
+	}
+
 	return nil
 }
 
-func (am *MultitenantAlertmanager) iteration(ctx context.Context) error {
-	err := am.updateConfigs()
+func (am *MultitenantAlertmanager) run(ctx context.Context) error {
+	level.Info(am.logger).Log("msg", "alertmanager up and running")
```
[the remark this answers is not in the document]
gotjosh: I think it's useful to know when we moved from `starting` to `running`. If you don't mind I would love to keep it around, this effectively symbolises that the instance is ready to serve requests.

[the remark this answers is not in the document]
pstibrany: Cortex also logs message when it “starts” — all services transitioned to “running” state. When alertmanager runs in microserbices mode, this extra log is redundant.

pkg/alertmanager/multitenant.go
```diff
@@ -283,45 +469,71 @@ func (am *MultitenantAlertmanager) stopping(_ error) error {
...
+			am.ringCheckErrors.Inc()
+			level.Error(am.logger).Log("msg", "failed to load user configuration", "user", userID, "err", err)
+			continue
+		}
+
+		if owned {
+			level.Debug(am.logger).Log("msg", "configuration owned", "user", userID)
+			ownedConfigs[userID] = cfg
+		} else {
+			level.Debug(am.logger).Log("msg", "configuration not owned, ignoring", "user", userID)
```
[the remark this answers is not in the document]
gotjosh: The ruler has this exact same logs and in the past, they've proven to be useful in understanding who owns what. Configuration mishaps might happen at runtime that would not end up running a particular tenant's instance. 

I'd advocate to keep them in.

pkg/alertmanager/multitenant.go
```diff
@@ -283,45 +469,71 @@ func (am *MultitenantAlertmanager) stopping(_ error) error {
...
+			level.Warn(am.logger).Log("msg", "failed to leave the cluster", "err", err)
+		}
+
+	}
+
+	if am.subservices != nil {
+		// subservices manages ring and lifecycler, if sharding was enabled.
+		_ = services.StopManagerAndAwaitStopped(context.Background(), am.subservices)
 	}
 	level.Debug(am.logger).Log("msg", "stopping")
```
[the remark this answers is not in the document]
gotjosh: Removed.

pkg/cortex/modules.go
gotjosh: I think we do. I'm not entirely sure what using a different port entitles, but this ring is also used by the distributors. Could you point me in the right direction to understand how port exposure makes a difference here?

pkg/alertmanager/replication_strategy.go
gotjosh: On it, I'll leave this refactor for last. Once I get a good run of all test since it should be a no-op.

pkg/alertmanager/multitenant.go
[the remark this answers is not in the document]
gotjosh: I'll take it.

pkg/alertmanager/multitenant.go
```diff
@@ -250,29 +315,150 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
...
+		return err
+	}
+
+	if am.cfg.ShardingEnabled {
+		// With the initial iteration now completed, we should have loaded all assigned alertmanager configurations to this instance. We can switch it to ACTIVE and start serving requests.
+		if err := am.ringLifecycler.ChangeState(ctx, ring.ACTIVE); err != nil {
+			return errors.Wrapf(err, "switch instance to %s in the ring", ring.ACTIVE)
+		}
+
+		// Wait until the ring client detected this instance in the ACTIVE state.
```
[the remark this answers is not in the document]
gotjosh: Given the above operation to change state executes async, there could be a chance where the ring client does not take the instance as active yet. With this, we can guarantee it.

pkg/alertmanager/multitenant.go
gotjosh: I documented the ones I have introduced manually. However, the one introduces by the ring I have not. Should I?

pkg/cortex/modules.go
gotjosh: cc: @codesome

codesome: Currently AM is serving only on http requests. So for a start AM distributors would be contacting AM via http (to keep the changes simpler). So should it not be http port for now? And later when/if we switch to using grpc, use the grpc port in ring?

pkg/alertmanager/replication_strategy.go
gotjosh: Done.

pkg/alertmanager/multitenant.go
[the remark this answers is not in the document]
pracucci: Not sure I still understand why we need it. Let's say we remove this piece of code. The `run()` function will run and the `loadAndSyncConfigs()` will be re-triggered as soon as the ring client detects the switch to `ACTIVE` (because `HasReplicationSetChanged()` will detect it has a change).

pkg/alertmanager/multitenant.go
```diff
@@ -250,29 +290,175 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
...
 	if err != nil {
-		level.Warn(am.logger).Log("msg", "error updating configs", "err", err)
+		am.syncFailures.WithLabelValues(syncReason).Inc()
+		return err
 	}
-	// Returning error here would stop "MultitenantAlertmanager" service completely,
-	// so we return nil to keep service running.
+
+	am.syncConfigs(cfgs)
+	am.syncTotal.WithLabelValues(syncReason).Inc()
```
pracucci, line 461: This is metric is tracking the total number of attempts, not successes. You ned to move it before calling `loadAlertmanagerConfigs()` otherwise it doesn't match with `syncFailures`.

pkg/alertmanager/multitenant.go
```diff
@@ -289,41 +475,61 @@ func (am *MultitenantAlertmanager) stopping(_ error) error {
...
+		am.tenantsOwned.Set(float64(len(configs)))
+		return configs, nil
+	}
+
+	ownedConfigs := map[string]alerts.AlertConfigDesc{}
+	for userID, cfg := range configs {
+		owned, err := am.isConfigOwned(userID)
+		if err != nil {
+			am.ringCheckErrors.Inc()
+			level.Error(am.logger).Log("msg", "failed to load user configuration", "user", userID, "err", err)
```
pracucci, line 505:
```suggestion
			level.Error(am.logger).Log("msg", "failed to load alertmanager configuration for user", "user", userID, "err", err)
```

pkg/alertmanager/multitenant.go
```diff
@@ -289,41 +475,61 @@ func (am *MultitenantAlertmanager) stopping(_ error) error {
...
+			continue
+		}
+
+		if owned {
+			level.Debug(am.logger).Log("msg", "configuration owned", "user", userID)
```
pracucci, line 510:
```suggestion
			level.Debug(am.logger).Log("msg", "alertmanager configuration owned", "user", userID)
```

pkg/alertmanager/multitenant.go
```diff
@@ -289,41 +475,61 @@ func (am *MultitenantAlertmanager) stopping(_ error) error {
...
+			ownedConfigs[userID] = cfg
+		} else {
+			level.Debug(am.logger).Log("msg", "configuration not owned, ignoring", "user", userID)
```
pracucci, line 513:
```suggestion
			level.Debug(am.logger).Log("msg", "alertmanager configuration not owned, ignoring", "user", userID)
```

pkg/alertmanager/multitenant_test.go
```diff
@@ -368,3 +395,456 @@ receivers:
...
+
+	require.NoError(t, am.StartAsync(ctx))
+	err = am.AwaitRunning(ctx)
+	require.Error(t, err)
+	require.Equal(t, services.Failed, am.State())
+	require.False(t, am.ringLifecycler.IsRegistered())
+	require.NotNil(t, am.ring)
+}
+
+func generateSortedTokens(numTokens int) ring.Tokens {
```
pracucci, line 841: You can directly use `GenerateTokens()` because now it returns sorted tokens (we changed it recently).

pkg/alertmanager/multitenant_test.go
```diff
@@ -41,26 +46,64 @@ receivers:
...
+	tempDir, err := ioutil.TempDir(os.TempDir(), "alertmanager")
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		err := os.RemoveAll(tempDir)
+		require.NoError(t, err)
+	})
+
+	cfg := &MultitenantAlertmanagerConfig{
+		PollInterval: time.Minute,
```
pracucci, line 95: I think you need to set this after calling `flagext.DefaultValues()`. Doesn't `DefaultValues()` override it?

pkg/alertmanager/multitenant_test.go
```diff
@@ -368,3 +395,456 @@ receivers:
...
+		{
+			name:              "sharding enabled, 3 instances, RF = 2",
+			withSharding:      true,
+			instances:         3,
+			replicationFactor: 2,
+			configs:           10,
+			expectedTenants:   20, // configs * replication factor
+		},
+		{
+			name:              "sharding enabled, 5 instance, RF = 3",
```
pracucci, line 531:
```suggestion
			name:              "sharding enabled, 5 instances, RF = 3",
```

pkg/alertmanager/multitenant_test.go
```diff
@@ -368,3 +395,456 @@ receivers:
...
+				// The alertmanager is ready to be tested once all instances are ACTIVE and the ring settles.
+				for _, am := range instances {
+					for _, id := range instanceIDs {
+						require.NoError(t, ring.WaitInstanceState(ctx, am.ring, id, ring.ACTIVE))
+					}
+				}
+			}
+
+			// Now that the ring has settled, sync configs with the instances.
+			var NumConfigs, NumInstances int
```
pracucci, line 605: No need to export these variables. Please name with with a lowercase first letter.

pkg/alertmanager/multitenant_test.go
```diff
@@ -368,3 +395,456 @@ receivers:
...
+			for _, am := range instances {
+				err := am.loadAndSyncConfigs(ctx, reasonRingChange)
+				require.NoError(t, err)
+				NumConfigs += len(am.cfgs)
+				NumInstances += len(am.alertmanagers)
+
```
pracucci, line 611: Extra newline.

pkg/alertmanager/multitenant_test.go
```diff
@@ -368,3 +395,456 @@ receivers:
...
+			}))
+
+			// Assert if we expected a sync or not.
+			if tt.expected {
+				test.Poll(t, time.Second, float64(2), func() interface{} {
+					metrics := regs.BuildMetricFamiliesPerUser()
+					return metrics.GetSumOfCounters("cortex_alertmanager_sync_configs_total")
+				})
+			} else {
+				time.Sleep(250 * time.Millisecond)
```
pracucci, line 771: Why this sleep? We try to avoid sleeps in tests because are a source of flakyness. Can be converted into a `test.Poll()`?

pracucci (approved):
Good job! LGTM, modulo few last nits and the memberlist setup fix in `modules.go`. I've also merged the PR https://github.com/cortexproject/cortex/pull/3677. Could you rebase, please?

docs/configuration/v1-guarantees.md
```diff
@@ -66,3 +66,4 @@ Currently experimental features are:
   - The bucket index support in the querier and store-gateway (enabled via `-blocks-storage.bucket-store.bucket-index.enabled=true`) is experimental
   - The block deletion marks migration support in the compactor (`-compactor.block-deletion-marks-migration-enabled`) is temporarily and will be removed in future versions
 - Querier: tenant federation
+- [January 2021] Alertmanager: Sharding of tenants across multiple instances
```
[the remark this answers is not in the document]
gotjosh: I'm fine with this but I felt the date was useful and something we could perhaps start? Maybe a better off discussion as separate PR.

pkg/alertmanager/multitenant_test.go
gotjosh: I tried originally, but couldn't. The problem here is that we have no way of saying "wait for at least X polls before checking". 250ms is 2.5x the time needed for sync to trigger so hopefully this doesn't become a flake for a while..

The assertion below is comparing that the value did not change because only the initial sync happened. Hence a value of 1.

pkg/alertmanager/multitenant.go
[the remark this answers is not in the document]
gotjosh: > will be re-triggered as soon as the ring client detects the switch to ACTIVE (because HasReplicationSetChanged() will detect it has a change

It's exactly that. We're trading off that extra sync which should not be needed for waiting a little bit more for the instance to be active. 

I think that additionally If we experience slow reads for the ring state this can become a problem where we know this instance is active and is part of the replication set but is not being detected. By waiting, we're guaranteeing we were able to read back its state before moving it on. 

I originally saw it as part of the storegateway and I thought it seemed like a sensible thing to include as well - I like the idea of guaranteeing that we won't need that initial sync because we waiting to read it was active.

pstibrany (approved):
👍

pkg/ring/replication_strategy_test.go
```diff
@@ -10,93 +10,155 @@ import (
 
 func TestRingReplicationStrategy(t *testing.T) {
 	for i, tc := range []struct {
-		RF, LiveIngesters, DeadIngesters int
-		ExpectedMaxFailure               int
-		ExpectedError                    string
+		replifcationFactor, liveIngesters, deadIngesters int
```
pracucci, line 13: `replifcationFactor` > `replicationFactor`

pracucci (approved):
LGTM with a  nit

PR #3667, opened 2021-01-08
Add gossip interval flag for alertmanger

simonswine:
Signed-off-by: Christian Simon <simon@swine.de>

**What this PR does**:

This allows to configure a different gossip interval for alertmanager. This allow to adapt the default of `200ms` to different values, which can help to address delayed gossip notifications in high volume and latency deployments.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


pkg/alertmanager/multitenant.go
```diff
@@ -117,6 +118,7 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet) {
 
 	f.StringVar(&cfg.ClusterBindAddr, "cluster.listen-address", defaultClusterAddr, "Listen address for cluster.")
 	f.StringVar(&cfg.ClusterAdvertiseAddr, "cluster.advertise-address", "", "Explicit address to advertise in cluster.")
+	f.DurationVar(&cfg.ClusterGossipInterval, "cluster.gossip-interval", cluster.DefaultGossipInterval, "The interval between sending messages that need to be gossiped that haven't been able to piggyback on probing messages. If this is set to zero, non-piggyback gossip is disabled. By lowering this value (more frequent) gossip messages are propagated across cluster more quickly at the expense of increased bandwidth.")
```
simonswine, line 121: Not too sure why we are not using an `-alertmanager.` prefix for all of those config flags, as currently it might mislead users to think it is affecting other parts of the system to (e.g. memberlist)

pstibrany: What is non-piggyback gossip?

If we allow configuring this value (which I think is already pretty low for gossip, and wouldn't recommend going to lower values), we should also enable configuration of pull/push interval for faster state convergence. (In addition to that, configuring GossipNodes would be useful too, but that doesn't seem configurable in AM).

pstibrany (approved):
LGTM, thanks!

pkg/alertmanager/multitenant.go
pracucci: > Not too sure why we are not using an -alertmanager. prefix for all of those config flags, as currently it might mislead users to think it is affecting other parts of the system to (e.g. memberlist)

I think it's a bug and we should fix it (but need to address it in a separate PR). I've opened an issue in the meanwhile: https://github.com/cortexproject/cortex/issues/3670

gotjosh: I have filed https://github.com/cortexproject/cortex/pull/3677 to fix #3670. Would be great to align these two PRs to make sure the nomenclature and prefix make sense on both. Happy to merge this as is and I can quickly fix it the other.

simonswine: > What is non-piggyback gossip?

That was a copy paste from the memberlist code comment. I think it is confusing more than helping so I reverted to the alertmanager description.

> pull/push interval 

Added that

> Gossip Nodes

Yes that seems to be fixed at quorum or 3 whatever is bigger: https://github.com/prometheus/alertmanager/blob/ce108378d4c8d580804452dcc3a31222067793b2/cluster/cluster.go#L184-L187 

> Aliging to #3677 

I have added some comments [here] (but meanwhile I used HA in the same way)

[here]:https://github.com/cortexproject/cortex/pull/3677/files#r555644111

pkg/alertmanager/multitenant.go
```diff
@@ -117,6 +119,8 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet) {
 
 	f.StringVar(&cfg.ClusterBindAddr, "cluster.listen-address", defaultClusterAddr, "Listen address for cluster.")
 	f.StringVar(&cfg.ClusterAdvertiseAddr, "cluster.advertise-address", "", "Explicit address to advertise in cluster.")
+	f.DurationVar(&cfg.HAGossipInterval, "alertmanager.ha.gossip-interval", cluster.DefaultGossipInterval, "The interval between sending gossip messages. By lowering this value (more frequent) gossip messages are propagated across cluster more quickly at the expense of increased bandwidth usage.")
+	f.DurationVar(&cfg.HAPushPullInterval, "alertmanager.ha.push-pull-interval", cluster.DefaultPushPullInterval, "The interval between gossip state syncs. Setting this interval lower (more frequent) will increase convergence speeds across larger clusters at the expense of increased bandwidth usage.")
```
pracucci, line 123: I wouldn't introduce a new prefix `ha` but use `cluster`, which we're already using. My suggestion is to name these new CLI flags (and their YAML field counterpart):

- `alertmanager.cluster-gossip-interval`
- `alertmanager.cluster-push-pull-interval`

pracucci (commented):
LGTM, but I've a comment about config naming.

pkg/alertmanager/multitenant.go
pracucci: Oh sorry. Naming is also related to https://github.com/cortexproject/cortex/pull/3677. Let's discuss offline about it.

pracucci:
@simonswine We merged #3677. Could you rebase `master` and move your new config options under the new `ClusterConfig` struct, please?

simonswine:
@pracucci this is ready for another review now

pracucci (approved):
Yeah, you did great. Thanks!

PR #3671, opened 2021-01-08
Alertmanager Distributor

codesome:
**What this PR does**:

Adds sharding capabilities to alertmanager for the `/alerts` API and introduces a new Distributor component in it based on the design [here](https://cortexmetrics.io/docs/proposals/scalable-alertmanager/).

**Checklist**
- [x] Tests updated
- [ ] Documentation added


gotjosh:
Depends on https://github.com/cortexproject/cortex/pull/3664/files

pracucci:
Let's get #3664 merged and then we'll review this PR, otherwise it's getting quite complicated to follow. Merging of #3664 should happen relatively soon (looks in a good shape to me).

gotjosh:
#3664 is now merged, feel free to rebase.

codesome:
I have rebased, but I need to fix some TODOs, add some metrics and more tests. I will tell when it is ready for review.

codesome:
It is ready for a review. I still have to fix the metrics part of unit test and have to test running the distributor and alertmanager in a single binary to see if routing is happening properly (with no conflicting route registrations).

codesome:
PR is ready for review. I am still testing some bits locally to make sure everything is working fine. I will be adding more tests to alertmanager.

The current idea is to handle sharding only for `/alerts` API and incrementally add support for others. cc @gotjosh

pkg/alertmanager/multitenant.go
```diff
@@ -735,6 +762,59 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+		am.distributor.DistributeRequest(w, req)
+		return
+	}
+
+	// If sharding is not enabled or Distributor does not support this path,
+	// it is a simple pass-through.
+	am.serveHTTP(w, req)
+}
+
+// HandleRequest serves the Alertmanager's web UI and API sen via gRPC.
```
ranton256, line 775: nit: typo sen[t]

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,104 @@
...
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health/grpc_health_v1"
+
+	am_client "github.com/cortexproject/cortex/pkg/alertmanager/alertmanagerpb"
```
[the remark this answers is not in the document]
codesome: I changed it just before you commented :)

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,277 @@
...
+	services.Service
+
+	cfg              Config
+	requestsInFlight sync.WaitGroup
+
+	alertmanagerRing ring.ReadRing
+	alertmanagerPool *client.Pool
+	replication      int
+
+	// Manager for subservices (AlertmanagerSet)
```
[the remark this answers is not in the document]
codesome: This felt like the simplest way to achieve this without needing a lot of changes in tests or code. I just followed how it was done in current distributors for samples, happy to make it an interface.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,277 @@
...
+)
+
+// Config contains the configuration required to
+// create a Distributor
+type Config struct {
+	RemoteTimeout      time.Duration    `yaml:"remote_timeout"`
+	AlertmanagerClient tls.ClientConfig `yaml:"alertmanager_client"`
+
+	// For testing and for extending the ingester by adding calls to the client
+	AlertmanagerClientFactory client.PoolFactory `yaml:"-"`
```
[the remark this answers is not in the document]
codesome: So the problem with `GetClientsFor` is that it gets all clients and we do all in parallel - but when we are sharding we need to return as soon as the quorum is satisfied. That is already implemented by `ring.DoBatch` and the only requirement is getting client for the given address. I can move it to be an interface passed as function argument but the method would be `GetClientFor(addr string) (AlertmanagerClient, error)`.

pkg/alertmanager/distributor/alertmanager_replication_set.go
```diff
@@ -0,0 +1 @@
+package distributor
```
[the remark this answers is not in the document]
codesome: Don't need as of now

pkg/alertmanager/distributor/distributor.go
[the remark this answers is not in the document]
codesome: Now that I think about it, from my other comment, the quorum logic is already implemented in `ring.DoBatch`, so the only thing we need is a factory to get clients for address.  And that is like an interface (a simple function signature). The moving from manually spawning goroutines and returning after all requests end -> to -> using `ring.DoBatch` which takes care of quorum logic was the main motivation to simplify it down and remove `AlertmanagerSet`.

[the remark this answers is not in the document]
codesome: Also, my bad, will change *client.Pool to an interface :)

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (d *Distributor) doWrite(userID string, w http.ResponseWriter, r *http.Request) {
+	source := util.GetSourceIPsFromOutgoingCtx(r.Context())
+
+	var body []byte
+	var err error
+	if r.Body != nil {
+		body, err = ioutil.ReadAll(r.Body)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
```
gotjosh, line 173: We need to log the error here if we're not letting the user know what happened.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,283 @@
...
+		w.WriteHeader(http.StatusOK)
+		return
+	}
+
+	d.respondFromError(err, w)
+}
+
+func (d *Distributor) doRead(userID string, w http.ResponseWriter, r *http.Request) {
+	key := shardByUser(userID)
+	replicationSet, err := d.alertmanagerRing.Get(key, ring.Read, nil, nil, nil)
```
gotjosh, line 218:
```suggestion
	replicationSet, err := d.alertmanagerRing.Get(key, alertmanager.RingOp, nil, nil, nil)
```

Ditto from above, given the current semantics I don't think we want to read from anything that's not active.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
gotjosh, line 220: Let's make sure we log the error here and present the user with a more friendly error. I don't think the users need to know about our ring failures.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+}
+
+func (d *Distributor) doWrite(userID string, w http.ResponseWriter, r *http.Request) {
+	source := util.GetSourceIPsFromOutgoingCtx(r.Context())
+
+	var body []byte
+	var err error
+	if r.Body != nil {
+		body, err = ioutil.ReadAll(r.Body)
```
gotjosh, line 171: We already read the body as we do `server.HTTPRequest(r)`, so any errors we might expect from it can be handled on that error condition. Is there a reason why we need to do it here?

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,283 @@
...
+		localCtx = user.InjectOrgID(localCtx, userID)
+		if sp := opentracing.SpanFromContext(r.Context()); sp != nil {
+			localCtx = opentracing.ContextWithSpan(localCtx, sp)
+		}
+		// Get clientIP(s) from Context and add it to localCtx
+		localCtx = util.AddSourceIPsToOutgoingContext(localCtx, source)
+
+		// http.Request.Clone() does not do a deep copy, hence copying the body.
+		// https://github.com/golang/go/issues/36095.
+		rClone := r.Clone(r.Context())
```
gotjosh, line 192: Do we need to close the request? (this is a genuine question) I'm not sure we do if we're using `server.HTTPRequest` as the wrapper for the request.

pkg/alertmanager/distributor/distributor.go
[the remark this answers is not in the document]
gotjosh: Makes sense, I think we can stick with `DoBatch` for simplifying things.

gotjosh (commented):
Good job @codesome, we're almost there. 🎉

pkg/alertmanager/distributor/distributor.go
codesome: When we have multiple requests spawning form here, only one of them will be able to read it full and will cause panics in others. So we copy the body here to be able to use in multiple requests.

pkg/alertmanager/distributor/distributor.go
codesome: Looks like we don't need to clone the entire request but only need to clone the body (which this clone does not do, it is done separately). I thought of not cloning and creating grpc request manually from 1 request, but the code inside `server.HTTPRequest` has some unexposed methods. I guess it should be fine to copy that small method here and skip this cloning?

codesome: Yes I will copy that small method and skip cloning.

codesome:
(PR is ready for review)

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,306 @@
...
+		return
+	}
+
+	http.Error(w, string(resp.HttpResponse.Body), int(resp.HttpResponse.Code))
+}
+
+func (d *Distributor) respondFromError(err error, w http.ResponseWriter) {
+	httpResp, ok := httpgrpc.HTTPResponseFromError(errors.Cause(err))
+	if !ok {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
gotjosh, line 258: I don't think the user needs to know about this error, let's log and use a generic message something like: `Failed to process the request to the alertmanager`

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,306 @@
...
+
+		if resp.GetStatus() != alertmanagerpb.OK {
+			return errors.New("alertmanager grpc request not ok")
+		}
+
+		return nil
+	}, func() {})
+
+	if err == nil {
+		w.WriteHeader(http.StatusOK)
```
gotjosh, line 209: I think we need to respond with any of the bodies from the downstream alertmanager either pick the first one or replicate the response from upstream.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,306 @@
...
+	http.Error(w, string(httpResp.Body), int(httpResp.Code))
+}
+
+func (d *Distributor) doRequest(ctx context.Context, am ring.InstanceDesc, r *http.Request, requestBody []byte) (*alertmanagerpb.Response, error) {
+	amClient, err := d.alertmanagerClientsPool.GetClientFor(am.Addr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get alertmanager from pool %s", am.Addr)
+	}
+
+	resp, err := amClient.HandleRequest(ctx, &alertmanagerpb.Request{
```
gotjosh, line 275: If you use `server.HTTPRequest` from [`weaveworks/common`](https://github.com/weaveworks/common/blob/master/httpgrpc/server/server.go#L157-L168) you don't have to a) read the body initially and instead just copy the req, b) have `httpTogrpchttpHeaders` and c) manually construct the `httpgrpc.HTTPRequest` struct

pkg/alertmanager/multitenant.go
```diff
@@ -737,6 +766,59 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+		return
+	}
+
+	// If sharding is not enabled or Distributor does not support this path,
+	// it is a simple pass-through.
+	am.serveHTTP(w, req)
+}
+
+// HandleRequest serves the Alertmanager's web UI and API sent via gRPC.
+func (am *MultitenantAlertmanager) HandleRequest(ctx context.Context, in *alertmanagerpb.Request) (*alertmanagerpb.Response, error) {
```
gotjosh, line 780: nit: I have the feeling we can use [this handler](https://github.com/weaveworks/common/blob/master/httpgrpc/server/server.go#L50-L71) to reduce some of the boilerplate but I guess we could do that in a follow up PR.

gotjosh (commented):
Thanks for addressing my feedback! My comments should be the last set of minor details.

pkg/alertmanager/distributor/distributor.go
codesome: Yeah, I can do that for reads for now. Writes will require some additional overheads of creating another body reader and eventually consuming it again (more allocations), so for writes I think copying the body and not using `server.HTTPRequest` is better.

pkg/alertmanager/multitenant.go
codesome: A future PR sounds good. At the moment I am sending all non 200 as errors so that the quorum can catch the failures of any kind properly. The linked handler would return errors only for 5xx and not 4xx hence will require some special handling in the distributor.

pkg/alertmanager/multitenant.go
```diff
@@ -737,6 +766,59 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 		return
 	}
 
+	if am.cfg.ShardingEnabled && am.distributor.IsPathSupported(req.URL.Path) {
+		am.distributor.DistributeRequest(w, req)
+		return
+	}
+
+	// If sharding is not enabled or Distributor does not support this path,
+	// it is a simple pass-through.
```
pracucci, line 775:
```suggestion
	// it is served by this instance.
```

pkg/alertmanager/multitenant.go
```diff
@@ -737,6 +766,59 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 		return
 	}
 
+	if am.cfg.ShardingEnabled && am.distributor.IsPathSupported(req.URL.Path) {
```
pracucci, line 769: Concern: once this PR gets merged, the distributor will be always used if sharding is enabled. Shouldn't we introduce a temporarily flag to enable the distributor (disabled by default) so that we'll have control over it while testing / rolling out?

pkg/alertmanager/multitenant.go
```diff
@@ -737,6 +766,59 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+	var body []byte
+	httpResp := w.Result()
+	if httpResp.Body != nil {
+		body, err = ioutil.ReadAll(httpResp.Body)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// All non 200 status are sent as errors.
```
pracucci, line 800: Why? Shouldn't we just forward the HTTP response "as is", whatever is the response code? Also you're checking for `== 200` but potentially there may be other successful codes (eg. 2xx, 3xx).

pkg/alertmanager/multitenant.go
```diff
@@ -226,12 +233,15 @@ func newMultitenantAlertmanagerMetrics(reg prometheus.Registerer) *multitenantAl
 // organizations.
 type MultitenantAlertmanager struct {
 	services.Service
+	alertmanagerpb.AlertmanagerClient
+	grpc_health_v1.HealthClient
```
pracucci, line 237: Why the `MultitenantAlertmanager` needs to implement `grpc_health_v1.HealthClient`?

pkg/alertmanager/multitenant.go
```diff
@@ -768,6 +850,14 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
 	http.Error(w, "the Alertmanager is not configured", http.StatusNotFound)
 }
 
+func (am *MultitenantAlertmanager) Check(_ context.Context, _ *grpc_health_v1.HealthCheckRequest, _ ...grpc.CallOption) (*grpc_health_v1.HealthCheckResponse, error) {
```
pracucci, line 853: Why you need this? Given we have 1 global gRPC server per process, aren't we already covered by the health server registered in `pkg/cortex/cortex.go`?

pkg/alertmanager/multitenant.go
```diff
@@ -795,6 +885,11 @@ func (am *MultitenantAlertmanager) GetStatusHandler() StatusHandler {
 	}
 }
 
+// Close implements ring_client.PoolClient.
+func (am *MultitenantAlertmanager) Close() error {
```
pracucci, line 889: Why do we need it on `MultitenantAlertmanager`?

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (f *alertmanagerClientsPool) GetClientFor(addr string) (AlertmanagerClient, error) {
+	c, err := f.pool.GetClientFor(addr)
+	if err != nil {
+		return nil, err
+	}
+	return c.(AlertmanagerClient), nil
+}
+
+func dialAlertmanagerClient(cfg grpcclient.Config, addr string, rd *prometheus.HistogramVec) (*alertmanagerClient, error) {
```
pracucci, line 100: Please rename `rd` to `requestDuration` to make it more clear.

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,133 @@
...
+	cfg.TLS.RegisterFlagsWithPrefix(prefix, f)
+}
+
+type alertmanagerClientsPool struct {
+	pool *client.Pool
+}
+
+func newAlertmanagerClientsPool(discovery client.PoolServiceDiscovery, amClientCfg AlertmanagerClientConfig, logger log.Logger, reg prometheus.Registerer) AlertmanagerClientsPool {
+	// We prefer sane defaults instead of exposing further config options.
+	// TODO: Figure out if these defaults make sense.
```
pracucci, line 53: Remove this TODO. Some feedback on the following defaults:

```
MaxRecvMsgSize:      100 << 20,
```
It's 100MB. Looks quite a lot. Maybe we could reduce to 16MB.

```
MaxSendMsgSize:      16 << 20,
```

It's 16MB. Looks a lot for the request. Maybe we can reduce to 4MB.

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,133 @@
...
+		RateLimitBurst:      0,
+		BackoffOnRatelimits: false,
+		TLSEnabled:          amClientCfg.TLSEnabled,
+		TLS:                 amClientCfg.TLS,
+	}
+
+	rd := promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:   "cortex",
+		Name:        "alertmanager_client_request_duration_seconds",
+		Help:        "Time spent executing requests to the alertmanager.",
```
pracucci, line 68:
```suggestion
		Help:        "Time spent executing requests from an alertmanager to another alertmanager.",
```

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,133 @@
...
+		Buckets:     prometheus.ExponentialBuckets(0.008, 4, 7),
+		ConstLabels: prometheus.Labels{"client": "alertmanager-distributor"},
```
pracucci, line 70: The client should be the user-visible service, which is the "alertmanager".

```suggestion
		ConstLabels: prometheus.Labels{"client": "alertmanager"},
```

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,133 @@
...
+		RateLimit:           0,
+		RateLimitBurst:      0,
+		BackoffOnRatelimits: false,
+		TLSEnabled:          amClientCfg.TLSEnabled,
+		TLS:                 amClientCfg.TLS,
+	}
+
+	rd := promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:   "cortex",
+		Name:        "alertmanager_client_request_duration_seconds",
```
pracucci, line 67: Should be `alertmanager_distributor_client_request_duration_seconds`?

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,133 @@
...
+	poolCfg := client.PoolConfig{
+		CheckInterval:      time.Minute,
+		HealthCheckEnabled: true,
+		HealthCheckTimeout: 10 * time.Second,
+	}
+
+	clientsCount := promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "alertmanager_distributor_alertmanager_clients",
+		Help:      "The current number of alertmanager clients in the pool.",
```
pracucci, line 86:
```suggestion
		Help:      "The current number of alertmanager distributor clients in the pool.",
```

pkg/alertmanager/distributor/alertmanager_client.go
```diff
@@ -0,0 +1,133 @@
...
+
+	poolCfg := client.PoolConfig{
+		CheckInterval:      time.Minute,
+		HealthCheckEnabled: true,
+		HealthCheckTimeout: 10 * time.Second,
+	}
+
+	clientsCount := promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "alertmanager_distributor_alertmanager_clients",
```
pracucci, line 85:
```suggestion
		Name:      "alertmanager_distributor_clients",
```

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,317 @@
...
+	if alertmanagerClientsPool == nil {
+		alertmanagerClientsPool = newAlertmanagerClientsPool(client.NewRingServiceDiscovery(alertmanagersRing), cfg.AlertmanagerClient, logger, reg)
+	}
+
+	d = &Distributor{
+		cfg:                     cfg,
+		logger:                  logger,
+		alertmanagerRing:        alertmanagersRing,
+		alertmanagerClientsPool: alertmanagerClientsPool,
+		//replication:             alertmanagersRing.ReplicationFactor(),
```
pracucci, line 87: ?

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,317 @@
...
+
+	receivedRequests  *prometheus.CounterVec
+	amSends           *prometheus.CounterVec
+	amSendFailures    *prometheus.CounterVec
+	replicationFactor prometheus.Gauge
+}
+
+// New constructs a new Distributor
+func New(cfg Config, alertmanagersRing *ring.Ring, alertmanagerClientsPool AlertmanagerClientsPool, logger log.Logger, reg prometheus.Registerer) (d *Distributor, err error) {
+	if logger == nil {
```
pracucci, line 74: I would remove this and always pass a logger. That's what we're used to.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,317 @@
...
+
+	d = &Distributor{
+		cfg:                     cfg,
+		logger:                  logger,
+		alertmanagerRing:        alertmanagersRing,
+		alertmanagerClientsPool: alertmanagerClientsPool,
+		//replication:             alertmanagersRing.ReplicationFactor(),
+	}
+
+	d.initMetrics(reg)
```
pracucci, line 90: Inline it.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,317 @@
...
+	d.replicationFactor.Set(float64(alertmanagersRing.ReplicationFactor()))
```
pracucci, line 91: Do we need this metric at all? We currently have it only for the distributor (the series distributor) and we don't have it for the store-gateway, and I don't think we felt missing it.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,317 @@
...
+	d.requestsInFlight.Wait()
+	return nil
+}
+
+func (d *Distributor) stopping(_ error) error {
+	return services.StopManagerAndAwaitStopped(context.Background(), d.subservices)
+}
+
+// IsPathSupported returns true if the given route is currently supported by the Distributor.
+// TODO: This will go away in future after we gradually add support for the entire API.
```
pracucci, line 138: I would remove this TODO. Doesn't add much value. All code is subject to change by its nature ;)

pracucci (commented):
Good job!

I did a first pass review. I will continue the review tomorrow. In particular, I need to double check the distributor logic.

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+	}
+
+	firstSuccessfulResponse := make(chan *alertmanagerpb.Response, 1)
+	grpcHeaders := httpTogrpchttpHeaders(r.Header)
+	err = ring.DoBatch(r.Context(), RingOp, d.alertmanagerRing, []uint32{shardByUser(userID)}, func(am ring.InstanceDesc, _ []int) error {
+		d.amSends.WithLabelValues(am.Addr).Inc()
+
+		// Use a background context to make sure all alertmanagers get the request even if we return early.
+		localCtx, cancel := context.WithTimeout(context.Background(), d.cfg.RemoteTimeout)
```
pracucci, line 190: Shouldn't the `RemoteTimeout` be applied to reads too?

pkg/alertmanager/distributor/distributor.go
```diff
@@ -0,0 +1,317 @@
...
+// RingOp is the operation used for reading/writing to the alertmanagers.
+var RingOp = ring.NewOp([]ring.IngesterState{ring.ACTIVE}, func(s ring.IngesterState) bool {
+	// Only ACTIVE Alertmanager get requests. If instance is not ACTIVE, we need to find another Alertmanager.
+	return s != ring.ACTIVE
+})
+
+// Config contains the configuration required to
+// create a Distributor
+type Config struct {
+	RemoteTimeout      time.Duration            `yaml:"remote_timeout"`
```
pracucci, line 40: I would move this inside `AlertmanagerClientConfig`. Pratically it's a timeout applied on the client.

pkg/alertmanager/multitenant.go
codesome: Discussed offline, not required

pkg/alertmanager/multitenant.go
codesome: After some new discussions, I have used this ready handler now and modified distributor code accordingly.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	d = &Distributor{
+		cfg:                     cfg,
+		logger:                  logger,
+		alertmanagerRing:        alertmanagersRing,
+		alertmanagerClientsPool: alertmanagerClientsPool,
+	}
+
+	d.receivedRequests = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
```
pracucci, line 59: The naming of these metrics is a bit ambiguous. The prefix is `cortex_alertmanager_distributor_` so it looks to be from the alertmanager distributor's perspective. If so, what you call "received request" is actually "sent request" and viceversa. The distributor is the client, the alertmanager is the server.

Other than this, few more comments:
1. Do we need "receivedRequests" to be per user?
2. What's the "ingester" label in "amSends" and "amSendFailures"? Looks wrong to me
3. Since we have to rename these metrics, I would suggest you to take a look at current cortex request/response metric names to keep these specular (to find them, you can just open Prometheus and type "cortexrequest")

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	d.amSends = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Name: "cortex_alertmanager_distributor_alertmanager_send_total",
+		Help: "The total number of requests sent to the alertmanager.",
+	}, []string{"ingester"})
+	d.amSendFailures = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Name: "cortex_alertmanager_distributor_alertmanager_send_failures_total",
+		Help: "The total number of requests sent to the alertmanager that failed.",
+	}, []string{"ingester"})
+
+	d.Service = services.NewBasicService(d.starting, d.running, d.stopping)
```
pracucci, line 72: Since `starting()` and `stopping()` are empty functions, you can remove them and pass `nil` here. It's supported by the service implementation.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	}
+
+	logger := log.With(d.logger, "user", userID)
+
+	d.receivedRequests.WithLabelValues(userID).Inc()
+
+	if r.Method == http.MethodGet {
```
pracucci, line 115: Also `Method.Head` that the client may potentially issue (eg. the browser).

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+		level.Error(logger).Log("msg", "failed to get grpc request from http request", "err", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(r.Context(), d.cfg.AlertmanagerClient.RemoteTimeout)
+	defer cancel()
+	// Until we have a mechanism to combine the results from multiple alertmanagers,
+	// we forward the request to only only of the alertmanagers.
+	amDesc := replicationSet.Ingesters[0]
```
pracucci, line 209: I would pick a random one instead of `0`, in order to balance traffic (and write a comment to explain why we do it). Thoughts?

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+		return
+	}
+
+	http.Error(w, string(resp.HttpResponse.Body), int(resp.HttpResponse.Code))
+}
+
+func (d *Distributor) respondFromError(err error, w http.ResponseWriter, logger log.Logger) {
+	httpResp, ok := httpgrpc.HTTPResponseFromError(errors.Cause(err))
+	if !ok {
+		level.Error(logger).Log("msg", "Failed to process the request to the alertmanager", "err", err)
```
pracucci, line 224:
```suggestion
		level.Error(logger).Log("msg", "failed to process the request to the alertmanager", "err", err)
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		d.amSendFailures.WithLabelValues(amDesc.Addr).Inc()
+		d.respondFromError(err, w, logger)
+		return
+	}
+
+	http.Error(w, string(resp.HttpResponse.Body), int(resp.HttpResponse.Code))
+}
+
+func (d *Distributor) respondFromError(err error, w http.ResponseWriter, logger log.Logger) {
```
pracucci, line 221: Can be an utility function decoupled from `Distributor`. In other words, no need to take `Distributor` as receiver.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	httpResp, ok := httpgrpc.HTTPResponseFromError(errors.Cause(err))
+	if !ok {
+		level.Error(logger).Log("msg", "Failed to process the request to the alertmanager", "err", err)
+		http.Error(w, "Failed to process the request to the alertmanager", http.StatusInternalServerError)
+		return
+	}
+	d.respondFromGRPCHTTPResponse(w, httpResp)
+}
+
+func (d *Distributor) respondFromGRPCHTTPResponse(w http.ResponseWriter, httpResp *httpgrpc.HTTPResponse) {
```
pracucci, line 231: Can be an utility function decoupled from `Distributor`. In other words, no need to take `Distributor` as receiver.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+			w.Header().Add(h.Key, v)
+		}
+	}
+	http.Error(w, string(httpResp.Body), int(httpResp.Code))
+}
+
+func (d *Distributor) doRequest(ctx context.Context, am ring.InstanceDesc, req *httpgrpc.HTTPRequest) (*alertmanagerpb.Response, error) {
+	amClient, err := d.alertmanagerClientsPool.GetClientFor(am.Addr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get alertmanager from pool %s", am.Addr)
```
pracucci, line 243:
```suggestion
		return nil, errors.Wrapf(err, "failed to get alertmanager client from pool (alertmanager address: %s)", am.Addr)
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func shardByUser(userID string) uint32 {
+	ringHasher := fnv.New32a()
+	// Hasher never returns err.
+	_, _ = ringHasher.Write([]byte(userID))
+	return ringHasher.Sum32()
+}
+
+func httpTogrpchttpHeaders(hs http.Header) []*httpgrpc.Header {
```
pracucci, line 263:
```suggestion
func httpToHttpgrpcHeaders(hs http.Header) []*httpgrpc.Header {
```

pkg/alertmanager/distributor.go
pracucci, line 231: It's called "httpgrpc" not "grpchttp".

```suggestion
func (d *Distributor) respondFromHTTPGRPCResponse(w http.ResponseWriter, httpResp *httpgrpc.HTTPResponse) {
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	d.respondFromGRPCHTTPResponse(w, httpResp)
+}
+
+func (d *Distributor) respondFromGRPCHTTPResponse(w http.ResponseWriter, httpResp *httpgrpc.HTTPResponse) {
+	for _, h := range httpResp.Headers {
+		for _, v := range h.Values {
+			w.Header().Add(h.Key, v)
+		}
+	}
+	http.Error(w, string(httpResp.Body), int(httpResp.Code))
```
pracucci, line 237: A couple of things:
1. Using `http.Error()` is a bit misleading (we're not necessarily returning an error)
2. `http.Error()` overrides `Content-Type` and `X-Content-Type-Options`. I don't think we should

Because of this, I would not use `http.Error()` at all. You can `w.WriteHeader(code)` and then write the `httpResp.Body` to `w`.

_Can you improve tests to cover this regression too? We should't change any response HTTP header._

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	amDesc := replicationSet.Ingesters[0]
+	d.amSends.WithLabelValues(amDesc.Addr).Inc()
+	resp, err := d.doRequest(ctx, amDesc, req)
+	if err != nil {
+		d.amSendFailures.WithLabelValues(amDesc.Addr).Inc()
+		d.respondFromError(err, w, logger)
+		return
+	}
+
+	http.Error(w, string(resp.HttpResponse.Body), int(resp.HttpResponse.Code))
```
pracucci, line 218: See comment I left to `respondFromGRPCHTTPResponse()`

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	firstSuccessfulResponse := make(chan *alertmanagerpb.Response, 1)
+	grpcHeaders := httpTogrpchttpHeaders(r.Header)
+	err = ring.DoBatch(r.Context(), RingOp, d.alertmanagerRing, []uint32{shardByUser(userID)}, func(am ring.InstanceDesc, _ []int) error {
+		d.amSends.WithLabelValues(am.Addr).Inc()
+
+		// Use a background context to make sure all alertmanagers get the request even if we return early.
+		localCtx, cancel := context.WithTimeout(context.Background(), d.cfg.AlertmanagerClient.RemoteTimeout)
+		defer cancel()
+		localCtx = user.InjectOrgID(localCtx, userID)
+		if sp := opentracing.SpanFromContext(r.Context()); sp != nil {
```
pracucci, line 145: Why getting the span from context instead of creating a new span?

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+
+	if r.Method == http.MethodGet {
+		d.doRead(userID, w, r, logger)
+	} else {
+		d.doWrite(userID, w, r, logger)
+	}
+}
+
+func (d *Distributor) doWrite(userID string, w http.ResponseWriter, r *http.Request, logger log.Logger) {
+	source := util.GetSourceIPsFromOutgoingCtx(r.Context())
```
pracucci, line 123: I think this logic has been copied from `pkg/distributor` but how does that work in the alertmanager context? Who injects the source IPs into `r.Context()`? What do you want this logic convering? I'm a bit lost.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+			Url:     r.RequestURI,
+			Body:    body,
+			Headers: grpcHeaders,
+		})
+		if err != nil {
+			d.amSendFailures.WithLabelValues(am.Addr).Inc()
+			return err
+		}
+
+		if resp.HttpResponse.Code/100 != 2 {
```
pracucci, line 162: What's about redirects (3xx)?

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	if r.Body != nil {
+		body, err = ioutil.ReadAll(r.Body)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to read the request body during write", "err", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+	}
+
+	firstSuccessfulResponse := make(chan *alertmanagerpb.Response, 1)
```
pracucci, line 136: I'm missing something. The channel has buffer 1, so the writer starting from the 2nd will block until a reader reads. But the reader reads once `ring.DoBatch()` has returned, so how does it work with a replication factor > 1?

Generally speaking, I think the usage of a channel here is tricky and probably overkilled. A simple variable protected with a mutex would be enough.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+		d.respondFromError(err, w, logger)
+	}
+
+	select {
+	case resp := <-firstSuccessfulResponse:
+		d.respondFromGRPCHTTPResponse(w, resp.HttpResponse)
+	default:
+		// This should not happen, but we have this default case to prevent
+		// and bugs deadlocking at this point.
+		level.Warn(logger).Log("msg", "first successful response not available on nil error")
```
pracucci, line 184: The log is not much clear. I would clarify making it understandable from the Cortex operator.

pracucci (commented):
Thanks @codesome for addressing my previous feedback. Here is a second pass, mostly focused on the distributor.

pkg/alertmanager/distributor.go
codesome: Random ones sounds good

pkg/alertmanager/distributor.go
codesome: Should we handle redirects here? If we are not handling, then it is like a failure since we dont persist the alerts

pkg/alertmanager/multitenant.go
```diff
@@ -410,6 +434,19 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
 	return am, nil
 }
 
+type handlerForGRPCServer struct {
```
pstibrany, line 437: This mechanism and entire "AlertmanagerServer -> am.grpcServer -> handlerForGRPCServer -> internal serveHTTPHelper" flow should be documented.

pkg/alertmanager/multitenant.go
```diff
@@ -737,6 +774,24 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+}
+
+// HandleRequest serves the Alertmanager's web UI and API sent via gRPC.
+func (am *MultitenantAlertmanager) HandleRequest(ctx context.Context, in *alertmanagerpb.Request) (*alertmanagerpb.Response, error) {
+	resp, err := am.grpcServer.Handle(ctx, in.HttpRequest)
+	return &alertmanagerpb.Response{HttpResponse: resp}, err
+}
+
+// serveHTTPHelper serves the Alertmanager's web UI and API.
+func (am *MultitenantAlertmanager) serveHTTPHelper(w http.ResponseWriter, req *http.Request) {
```
pstibrany, line 794: It's not really a helper, but rather internal implementation of HTTP handler, without "sharding" check. Can we find a better name?

pkg/alertmanager/multitenant.go
```diff
@@ -737,6 +774,24 @@ func (am *MultitenantAlertmanager) ServeHTTP(w http.ResponseWriter, req *http.Re
...
+		am.distributor.DistributeRequest(w, req)
+		return
+	}
+
+	// If sharding is not enabled or Distributor does not support this path,
+	// it is served by this instance.
+	am.serveHTTPHelper(w, req)
+}
+
+// HandleRequest serves the Alertmanager's web UI and API sent via gRPC.
```
pstibrany, line 787: I'd suggest extending this comment a bit:

```suggestion
// HandleRequest implements gRPC Alertmanager service, which receives request from AlertManager-Distributor
// and forwards them to internal http handler (without doing the distribution logic again).
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+}
+
+func (d *Distributor) doWrite(userID string, w http.ResponseWriter, r *http.Request, logger log.Logger) {
+	source := util.GetSourceIPsFromOutgoingCtx(r.Context())
+
+	var body []byte
+	var err error
+	if r.Body != nil {
+		body, err = ioutil.ReadAll(r.Body)
```
pstibrany, line 128: There is no limit here. Hostile client can easily sends request with large body and crash Alertmanager. You can use `http.MaxBytesReader` to wrap `r.Body` and get error when too many bytes are read.

pkg/alertmanager/distributor.go
pstibrany: Sending is done via `select` with a `default` branch, so if there is no reader, writer will not do the send.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	select {
+	case resp := <-firstSuccessfulResponse:
+		d.respondFromGRPCHTTPResponse(w, resp.HttpResponse)
+	default:
+		// This should not happen, but we have this default case to prevent
+		// and bugs deadlocking at this point.
+		level.Warn(logger).Log("msg", "first successful response not available on nil error")
+		w.WriteHeader(http.StatusOK)
```
pstibrany, line 185: OK is misleading here. We don't really know what happened, I think we should not tell client OK in this case.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,272 @@
...
+
+	ctx, cancel := context.WithTimeout(r.Context(), d.cfg.AlertmanagerClient.RemoteTimeout)
+	defer cancel()
+	// Until we have a mechanism to combine the results from multiple alertmanagers,
+	// we forward the request to only only of the alertmanagers.
+	amDesc := replicationSet.Ingesters[0]
+	d.amSends.WithLabelValues(amDesc.Addr).Inc()
+	resp, err := d.doRequest(ctx, amDesc, req)
+	if err != nil {
+		d.amSendFailures.WithLabelValues(amDesc.Addr).Inc()
```
pstibrany, line 213: If one fails, should we try another one?

pstibrany (commented):
Good job! First pass review, will look at distributor logic in more detail.

pkg/alertmanager/distributor.go
codesome: Yes it was copied from pkg/distributor. I don't think it is required here.

development/tsdb-blocks-storage-s3/config/cortex.yaml
```diff
@@ -86,6 +86,15 @@ ruler:
 
 alertmanager:
   enable_api: true
+  sharding_enabled: true
```
pstibrany, line 89: Should we run multiple alertmanagers?

pkg/alertmanager/distributor.go
codesome: These request metrics would be taken care by the grpc server, so none of these metrics are required.

pkg/alertmanager/distributor.go
codesome: Ignore this metric, the "failure" is misleading here, it's removed now. The err is actually the error returned by the AM. The spillover is taken care by the `DoBatch` if an AM is not healthy, it is similar how we are doing for distributor->ingester. I don't think we have to re-try for any error after we have selected AM from the ring.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,237 @@
...
+			return
+		}
+	}
+
+	var firstSuccessfulResponse *httpgrpc.HTTPResponse
+	var respMtx sync.Mutex
+	grpcHeaders := httpTogrpchttpHeaders(r.Header)
+	err = ring.DoBatch(r.Context(), RingOp, d.alertmanagerRing, []uint32{shardByUser(userID)}, func(am ring.InstanceDesc, _ []int) error {
+		// Use a background context to make sure all alertmanagers get the request even if we return early.
+		localCtx, cancel := context.WithTimeout(context.Background(), d.cfg.AlertmanagerClient.RemoteTimeout)
```
pracucci, line 113: Why do you need it? I understand the comment, but I can't understand who will cancel the context if "we return early".

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,237 @@
...
+				http.Error(w, "Request body too large", http.StatusBadRequest)
+			}
+			level.Error(logger).Log("msg", "failed to read the request body during write", "err", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+	}
+
+	var firstSuccessfulResponse *httpgrpc.HTTPResponse
+	var respMtx sync.Mutex
```
pracucci, line 109: [just for clarity] Name mutex like the variable you're protecting with the `Mtx` suffix, so `firstSuccessfulResponseMtx`

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,237 @@
...
+	grpcHeaders := httpTogrpchttpHeaders(r.Header)
+	err = ring.DoBatch(r.Context(), RingOp, d.alertmanagerRing, []uint32{shardByUser(userID)}, func(am ring.InstanceDesc, _ []int) error {
+		// Use a background context to make sure all alertmanagers get the request even if we return early.
+		localCtx, cancel := context.WithTimeout(context.Background(), d.cfg.AlertmanagerClient.RemoteTimeout)
+		defer cancel()
+		localCtx = user.InjectOrgID(localCtx, userID)
+		sp, localCtx := opentracing.StartSpanFromContext(localCtx, "DistributeRequest.doWrite")
```
pracucci, line 116:
```suggestion
		sp, localCtx := opentracing.StartSpanFromContext(localCtx, "Distributor.doWrite")
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to get grpc request from http request", "err", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(r.Context(), d.cfg.AlertmanagerClient.RemoteTimeout)
+	defer cancel()
+
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "DistributeRequest.doRead")
```
pracucci, line 178:
```suggestion
	sp, ctx := opentracing.StartSpanFromContext(ctx, "Distributor.doRead")
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,237 @@
...
+	for _, h := range httpResp.Headers {
+		for _, v := range h.Values {
+			w.Header().Add(h.Key, v)
+		}
+	}
+	w.WriteHeader(int(httpResp.Code))
+	w.Write(httpResp.Body) //nolint
+}
+
+func (d *Distributor) doRequest(ctx context.Context, am ring.InstanceDesc, req *httpgrpc.HTTPRequest) (*httpgrpc.HTTPResponse, error) {
```
pracucci, line 212: [nit] Since you need to always apply the `RemoteTimeout` and since `doRequest()` takes `Distributor` as receiver (and you have access to the distributor config) you could apply the `RemoteTimeout` here so that we do it only once and for all.

pkg/alertmanager/distributor_test.go
```diff
@@ -0,0 +1,297 @@
...
+					contentTypeOptions = []string{"not-ok-option-1", "not-ok-option-2"}
+				}
+				require.Equal(t, contentType, resp.Header.Values("Content-Type"))
+				require.Equal(t, contentTypeOptions, resp.Header.Values("X-Content-Type-Options"))
+			}
+
+			// Since the response is sent as soon as the quorum is reached, when we
+			// reach this point the 3rd AM may not have received the request yet.
+			// To avoid flaky test we retry until we hit the desired state within a reasonable timeout.
+			test.Poll(t, time.Second, true, func() interface{} {
```
pracucci, line 113: This and the next suggestion below improves the error reporting on failure (you will see the actual value instead of just `false`):
```suggestion
			test.Poll(t, time.Second, c.expectedTotalCalls, func() interface{} {
```

pkg/alertmanager/distributor_test.go
```diff
@@ -0,0 +1,297 @@
...
+				totalReqCount := 0
+				for _, a := range ams {
+					reqCount := a.requestsCount(route)
+					// AM should not get duplicate requests.
+					require.True(t, reqCount <= 1, "duplicate requests %d", reqCount)
+					totalReqCount += reqCount
+				}
+
+				return c.expectedTotalCalls == totalReqCount
```
pracucci, line 122:
```suggestion
				return totalReqCount
```

pkg/alertmanager/distributor_test.go
```diff
@@ -0,0 +1,297 @@
...
+	grpc_health_v1.HealthClient
+	t *testing.T
+	// receivedRequests is map of route -> statusCode -> number of requests.
+	receivedRequests map[string]map[int]int
+	mtx              sync.Mutex
+	myAddr           string
+	happy            bool
+}
+
+func newMockAlertmanager(t *testing.T, idx int, happy bool) *mockAlertmanager {
```
pracucci, line 199: Looks a bad smell passing `testing.T` to a mock. I would remove it. See related comments below.

pkg/alertmanager/distributor_test.go
```diff
@@ -0,0 +1,297 @@
...
+		happy:            happy,
+	}
+}
+
+func (am *mockAlertmanager) HandleRequest(_ context.Context, in *httpgrpc.HTTPRequest, _ ...grpc.CallOption) (*httpgrpc.HTTPResponse, error) {
+	am.mtx.Lock()
+	defer am.mtx.Unlock()
+
+	u, err := url.Parse(in.Url)
+	require.NoError(am.t, err)
```
pracucci, line 213: You can return error if `url.Parse()` fails.

pkg/alertmanager/distributor_test.go
```diff
@@ -0,0 +1,297 @@
...
+	am.mtx.Lock()
+	defer am.mtx.Unlock()
+
+	routesReceived := len(am.receivedRequests)
+	if routesReceived == 0 {
+		// This AM did not receive any requests.
+		return 0
+	}
+
+	require.Equal(am.t, 1, routesReceived)
```
pracucci, line 262: A generic mock should assert on the number of requests received. I would remove this.

pkg/alertmanager/distributor_test.go
```diff
@@ -0,0 +1,297 @@
...
+	routeMap, ok := am.receivedRequests[route]
+	require.True(am.t, ok)
```
pracucci, line 264: You can remove this assertion and change it to:

```
if !ok {
return 0
}
```

pkg/alertmanager/distributor_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+func (am *mockAlertmanager) requestsCount(route string) int {
+	am.mtx.Lock()
+	defer am.mtx.Unlock()
+
+	routesReceived := len(am.receivedRequests)
+	if routesReceived == 0 {
+		// This AM did not receive any requests.
+		return 0
+	}
```
pracucci, line 260: You can remove this check entirely. See below, but basically `routeMap, ok := am.receivedRequests[route]` will return `ok = false` if there are no routes

pracucci (commented):
Thanks a lot for addressing all comments. Change LGTM and most of the left comments are last nit, **but** there are a couple of serious ones reason why I haven't marked as approved yet. Thanks!

pkg/distributor/distributor.go
```diff
@@ -180,7 +180,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.HATrackerConfig.RegisterFlags(f)
 	cfg.DistributorRing.RegisterFlags(f)
 
-	f.IntVar(&cfg.MaxRecvMsgSize, "distributor.max-recv-msg-size", 100<<20, "remote_write API max receive message size (bytes).")
+	f.IntVar(&cfg.MaxRecvMsgSize, "distributor.max-recv-body-size", 100<<20, "remote_write API max receive message size (bytes).")
```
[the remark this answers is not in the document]
codesome: Not sure how this changed. Will revert back.

integration/alertmanager_test.go
```diff
@@ -269,7 +269,7 @@ func TestAlertmanagerSharding(t *testing.T) {
 		require.NoError(t, am.WaitSumMetrics(e2e.Equals(float64(30)), "cortex_alertmanager_tenants_discovered"))
 		// We know that the ring has settled when every instance has some tenants and the total number of tokens have been assigned.
 		require.NoError(t, am.WaitSumMetrics(e2e.Greater(float64(0)), "cortex_alertmanager_tenants_owned"))
-		require.NoError(t, am.WaitSumMetrics(e2e.Equals(float64(384)), "cortex_ring_tokens_total"))
+		require.NoError(t, am.WaitSumMetrics(e2e.Equals(float64(384)), `cortex_ring_tokens_total`))
```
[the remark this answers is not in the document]
codesome: Some back and forth test cleanup left behind. Will revert this.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+
+func (d *Distributor) doWrite(userID string, w http.ResponseWriter, r *http.Request, logger log.Logger) {
+	var body []byte
+	var err error
+	if r.Body != nil {
+		body, err = ioutil.ReadAll(r.Body)
+		if err != nil {
+			if err.Error() == "http: request body too large\n" {
```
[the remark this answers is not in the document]
codesome: > I believe the ending \n is a mistake

`http.MaxBytesReader` has `\n` in it's error :/

pkg/alertmanager/distributor.go
codesome: This logic is copied from the pkg/distributor code. So you mean not required here?

CHANGELOG.md
```diff
@@ -10,6 +10,18 @@
   - `-<prefix>.s3.sse.kms-encryption-context`
 * [FEATURE] Querier: Enable `@ <timestamp>` modifier in PromQL using the new `-querier.at-modifier-enabled` flag. #3744
 * [FEATURE] Overrides Exporter: Add `overrides-exporter` module for exposing per-tenant resource limit overrides as metrics. It is not included in `all` target, and must be explicitly enabled. #3785
+* [FEATURE] Experimental thanosconvert: introduce an experimental tool `thanosconvert` to migrate Thanos block metadata to Cortex metadata. #3770
+* [FEATURE] Alertmanager: It now shards the `/api/v1/alerts` API using the ring when sharding is enabled. #3671
+  * Added `NewMaxBytesHandler` in the utils package for limiting the size of http request body.
```
pracucci, line 15: We should document the CHANGELOG for the final user, so:

```suggestion
  * Added `-alertmanager.max-recv-msg-size` (defaults to 16M) to limit the size of HTTP request body handled by the alertmanager.
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	}
+
+	var firstSuccessfulResponse *httpgrpc.HTTPResponse
+	var firstSuccessfulResponseMtx sync.Mutex
+	grpcHeaders := httpTogrpchttpHeaders(r.Header)
+	err = ring.DoBatch(r.Context(), RingOp, d.alertmanagerRing, []uint32{shardByUser(userID)}, func(am ring.InstanceDesc, _ []int) error {
+		// Use a background context to make sure all alertmanagers get the request even if we return early.
+		localCtx := user.InjectOrgID(context.Background(), userID)
+		sp, localCtx := opentracing.StartSpanFromContext(localCtx, "Distribute.doWrite")
```
pracucci, line 115:
```suggestion
		sp, localCtx := opentracing.StartSpanFromContext(localCtx, "Distributor.doWrite")
```

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	req, err := server.HTTPRequest(r)
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to get grpc request from http request", "err", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	sp, ctx := opentracing.StartSpanFromContext(r.Context(), "Distribute.doRead")
```
pracucci, line 174:
```suggestion
	sp, ctx := opentracing.StartSpanFromContext(r.Context(), "Distributor.doRead")
```

pkg/alertmanager/distributor.go
pracucci: This comment is still valid :)

pkg/alertmanager/multitenant.go
```diff
@@ -132,6 +138,7 @@ const (
 func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.DataDir, "alertmanager.storage.path", "data/", "Base path for data storage.")
 	f.DurationVar(&cfg.Retention, "alertmanager.storage.retention", 5*24*time.Hour, "How long to keep data for.")
+	f.Int64Var(&cfg.MaxRecvMsgSize, "alertmanager.max-recv-msg-size", 16<<20, "Maximum size (bytes) of an accepted HTTP request body size.")
```
pracucci, line 141: "size" is repeated

```suggestion
	f.Int64Var(&cfg.MaxRecvMsgSize, "alertmanager.max-recv-msg-size", 16<<20, "Maximum size (bytes) of an accepted HTTP request body.")
```

pracucci (approved):
Good job, LGTM. Thanks! (just left few minor nits)

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,235 @@
...
+	defer sp.Finish()
+	// Until we have a mechanism to combine the results from multiple alertmanagers,
```
ranton256, line 176: What would combining mean in terms of behavior?

ranton256 (approved):
LGTM

pkg/alertmanager/distributor.go
codesome: Combining would deduplicate the results from those alertmanagers, basically a union

gotjosh (approved):

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,235 @@
...
+
+	firstSuccessfulResponseMtx.Lock() // Another request might be ongoing after quorum.
+	resp := firstSuccessfulResponse
+	firstSuccessfulResponseMtx.Unlock()
+
+	if resp != nil {
+		respondFromHTTPGRPCResponse(w, resp)
+	} else {
+		// This should not happen.
+		level.Error(logger).Log("msg", "distributor did not receive response from alertmanager though no errors")
```
pstibrany, line 153:
```suggestion
		level.Error(logger).Log("msg", "distributor did not receive any response from alertmanagers, but there were no errors")
````

File: pkg/util/http.go [hunk 79]
```diff
@@ -235,3 +245,24 @@ func SerializeProtoResponse(w http.ResponseWriter, resp proto.Message, compressi
 	}
 	return nil
 }
+
+type MaxBytesHandler struct {
```
pstibrany, line 249: Using handler for switching `r.Body` to `http.MaxBytesReader` is too disconnected from where we actually read the body. Let's use more direct code.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,235 @@
...
+	requestsInFlight sync.WaitGroup
+
+	alertmanagerRing        ring.ReadRing
+	alertmanagerClientsPool ClientsPool
+
+	logger log.Logger
+}
+
+// NewDistributor constructs a new Distributor
+func NewDistributor(cfg *MultitenantAlertmanagerConfig, alertmanagersRing *ring.Ring, alertmanagerClientsPool ClientsPool, logger log.Logger, reg prometheus.Registerer) (d *Distributor, err error) {
```
pstibrany, line 42: Let's extract options used by Distributor to its own Config structure. No need to pass entire unrelated config structure to Distributor.

pstibrany (commented):
Great job, this looks good! I've left some last comments.

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,235 @@
...
+		level.Error(logger).Log("msg", "failed to get grpc request from http request", "err", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	sp, ctx := opentracing.StartSpanFromContext(r.Context(), "Distributor.doRead")
+	defer sp.Finish()
+	// Until we have a mechanism to combine the results from multiple alertmanagers,
+	// we forward the request to only only of the alertmanagers.
+	amDesc := replicationSet.Ingesters[rand.Intn(len(replicationSet.Ingesters))]
```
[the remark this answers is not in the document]
codesome: The strategy used for AM ring makes sure there is at least 1 AM, so do we need the check here?

pkg/alertmanager/distributor.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+}
+
+func (d *Distributor) doWrite(userID string, w http.ResponseWriter, r *http.Request, logger log.Logger) {
+	var body []byte
+	var err error
+	if r.Body != nil {
+		body, err = ioutil.ReadAll(r.Body)
+		if err != nil {
+			if errors.Is(util.ErrRequestBodyTooLarge, err) {
```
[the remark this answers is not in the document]
codesome: If we are setting a limit on http body, I would expect it to be for every request that comes to alertmanager and not special cased for some requests. And this is not the only place where we are reading the body, `server.HTTPRequest` also reads the body (which makes me realise we need this check there too). Also when alertmanager is processing the request without going through the distributor.

pkg/util/http.go
codesome: I feel the other way round. Having it at handler level allows us to not think about handling the size limit everywhere (if we miss somewhere, potential attack surface?). For example, we use the upstream Prometheus alertmanager which takes the http request and we don't have control over it.

pkg/alertmanager/distributor.go
codesome: Turns out we only need `cfg.AlertmanagerClient`, so instead of creating a new distributor config, I will pass this one instead

pkg/alertmanager/distributor.go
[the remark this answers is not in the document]
pstibrany: Setting generic limit on http handler is wrong approach for several reasons:
- in general, large http request body is not a problem
- handlers that don’t read body **at all** don’t need the limit, nothing will break. Other handlers may need different limits (eg. if handler streams body to a file, 1MB limit may be too low).
- **it only becomes a problem when full body is read to memory** like here – we risk OOMs
- so **this** code needs to be protected – and that protection exists, but is hidden. But **this** code cannot check whether it is already using limited body or not, and blindly relies on the fact that it is. If we copy this handler code in the future to a different place, suddenly the protection may not be there.

For `server.HTTPRequest` I’d suggest giving it explicit max-body-size parameter – that breaks existing API, but in this case it may be good as it forces us to fix it everywhere.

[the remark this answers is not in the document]
codesome: Made it local

pracucci (approved):
Still LGTM. Time to merge it!

CHANGELOG.md
```diff
@@ -10,6 +10,18 @@
   - `-<prefix>.s3.sse.kms-encryption-context`
 * [FEATURE] Querier: Enable `@ <timestamp>` modifier in PromQL using the new `-querier.at-modifier-enabled` flag. #3744
 * [FEATURE] Overrides Exporter: Add `overrides-exporter` module for exposing per-tenant resource limit overrides as metrics. It is not included in `all` target, and must be explicitly enabled. #3785
+* [FEATURE] Experimental thanosconvert: introduce an experimental tool `thanosconvert` to migrate Thanos block metadata to Cortex metadata. #3770
+* [FEATURE] Alertmanager: It now shards the `/api/v1/alerts` API using the ring when sharding is enabled. #3671
```
gouthamve, line 14: What flags did we add to enable the sharding, I don't see those here.

CHANGELOG.md
```diff
@@ -10,6 +10,18 @@
...
+  * Added `-alertmanager.max-recv-msg-size` (defaults to 16M) to limit the size of HTTP request body handled by the alertmanager.
+  * New flags added for communication between alertmanagers:
```
gouthamve, line 16: Can we move this to a new [FEATURE] if it doesn't have to do with sharding?

gouthamve (approved):

CHANGELOG.md
codesome: `[FEATURE] Alertmanager: introduced the experimental option -alertmanager.sharding-enabled to shard tenants across multiple Alertmanager instances. This feature is still under heavy development and its usage is discouraged. The following new metrics are exported by the Alertmanager: #3664 ` - in 1.7.0

CHANGELOG.md
codesome: https://github.com/cortexproject/cortex/pull/3671#discussion_r578356324
It is related to sharding and we have a flag for that

PR #3677, opened 2021-01-11
Alertmanager: Deprecate (but keep compatibility) of cluster flags

### LINKED ISSUE #3670: Alertmanager cluster CLI flags have the wrong prefix
**Describe the bug**
The multitenant alertmanager CLI flags used to configure the alertmanager cluster are missing the `-alertmanager.` prefix (they start with `-cluster.` which is quite generic and may lead to misunderstandings):
https://github.com/cortexproject/cortex/blob/e4867d81778014c9fbb053e3f05743fb0b39853f/pkg/alertmanager/multitenant.go#L118-L121

According to the v1 guarantees, only the alertmanager APIs are marked experimental, so we can't change these CLI flags straightaway but we need to introduce the new flags, deprecate the old ones (while keeping backward compatibility) and remove the old ones after 2 releases.

gotjosh:
**What this PR does**:

Deprecates flags with the cluster prefix within the Alertmanager. I've
opted for a "high-availability" nomenclature which feel a bit more
aligned with what the flags are actually doing.

Fixes #3670 

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/alertmanager/multitenant.go
```diff
@@ -94,6 +94,11 @@ type MultitenantAlertmanagerConfig struct {
 	Peers                flagext.StringSlice `yaml:"peers"`
 	PeerTimeout          time.Duration       `yaml:"peer_timeout"`
 
+	HABindAddr      string                 `yaml:"ha_bind_address"`
+	HAAdvertiseAddr string                 `yaml:"ha_advertise_address"`
+	HAPeers         flagext.StringSliceCSV `yaml:"ha_initial_peers"`
+	HAPeerTimeout   time.Duration          `yaml:"ha_peer_timeout"`
```
simonswine, line 100: I am not too sure if `ha` is an improvement. I would prefer staying with `cluster` as alertmanager upstream is using:

```
      --cluster.listen-address="0.0.0.0:9094"
                                 Listen address for cluster. Set to empty string to disable HA mode.
      --cluster.advertise-address=CLUSTER.ADVERTISE-ADDRESS
                                 Explicit address to advertise in cluster.
      --cluster.peer=CLUSTER.PEER ...
                                 Initial peers (may be repeated).
      --cluster.peer-timeout=15s
                                 Time to wait between peers to send notifications.
      --cluster.gossip-interval=200ms
                                 Interval between sending gossip messages. By lowering this value (more frequent) gossip messages are propagated across the cluster more quickly at the expense of increased bandwidth.
      --cluster.pushpull-interval=1m0s
                                 Interval for gossip state syncs. Setting this interval lower (more frequent) will increase convergence speeds across larger clusters at the expense of increased bandwidth usage.
      --cluster.tcp-timeout=10s  Timeout for establishing a stream connection with a remote node for a full state sync, and for stream read and write operations.
      --cluster.probe-timeout=500ms
                                 Timeout to wait for an ack from a probed node before assuming it is unhealthy. This should be set to 99-percentile of RTT (round-trip time) on your network.
      --cluster.probe-interval=1s
                                 Interval between random node probes. Setting this lower (more frequent) will cause the cluster to detect failed nodes more quickly at the expense of increased bandwidth usage.
      --cluster.settle-timeout=1m0s
                                 Maximum time to wait for cluster connections to settle before evaluating notifications.
      --cluster.reconnect-interval=10s
                                 Interval between attempting to reconnect to lost peers.
      --cluster.reconnect-timeout=6h0m0s
```
Addtiionally I would also rather have a separate `ClusterConfig struct` so that the flag name separation with `.` is consistent with the YAML structure.

Also it might make sense to align the naming of parameters with our memberlist flags from here: https://github.com/cortexproject/cortex/blob/77c73aecaad5f81cba3fc4dd2bf5d4b501cb776c/pkg/ring/kv/memberlist/memberlist_client.go#L168-L186

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 * [CHANGE] Blocks storage: compactor is now required when running a Cortex cluster with the blocks storage, because it also keeps the bucket index updated. #3583
 * [CHANGE] Blocks storage: block deletion marks are now stored in a per-tenant global markers/ location too, other than within the block location. The compactor, at startup, will copy deletion marks from the block location to the global location. This migration is required only once, so you can safely disable it via `-compactor.block-deletion-marks-migration-enabled=false` once new compactor has successfully started once in your cluster. #3583
 * [CHANGE] OpenStack Swift: the default value for the `-ruler.storage.swift.container-name` and `-swift.container-name` config options has changed from `cortex` to empty string. If you were relying on the default value, you should set it back to `cortex`. #3660
+* [CHANGE] Alertmanager: Deprecate `cluster.` flags in favor of their `alertmanager.ha.` equivalent. The flags deprecated are: #3677
```
pracucci, line 9:
```suggestion
* [CHANGE] Alertmanager: Deprecate `cluster.` flags in favor of their `alertmanager.ha.` equivalent. The deprecated flags (and their respective YAML config options) are: #3677
```

pkg/alertmanager/multitenant.go
```diff
@@ -115,10 +123,17 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet) {
...
-	f.Var(&cfg.Peers, "cluster.peer", "Initial peers (may be repeated).")
-	f.DurationVar(&cfg.PeerTimeout, "cluster.peer-timeout", time.Second*15, "Time to wait between peers to send notifications.")
+	// Flags prefixed with `cluster` are deprecated in favor of their `alertmanager` prefix equivalent.
+	// TODO: New flags introduced in Cortex 1.7, remove old ones in Cortex 1.9
+	f.StringVar(&cfg.ClusterBindAddr, "cluster.listen-address", defaultClusterAddr, "Deprecated. Use -alertmanager.ha.listen-address instead.")
+	f.StringVar(&cfg.ClusterAdvertiseAddr, "cluster.advertise-address", "", "Deprecated. Use -alertmanager.ha.advertise-address instead.")
+	f.Var(&cfg.Peers, "cluster.peer", "Deprecated. Use -alertmanager.ha.peers instead.")
+	f.DurationVar(&cfg.PeerTimeout, "cluster.peer-timeout", time.Second*15, "Deprecated. Use -alertmanager.ha.peer-timeout instead.")
+
+	f.StringVar(&cfg.HABindAddr, "alertmanager.ha.listen-address", defaultClusterAddr, "Listen address and port for the cluster. Not specifying this flag disables high-availability mode.")
```
pracucci, line 133: If `ha` has a dot after it, then the YAML config option should be a sub-struct. Otherwise, keeping the current YAML config as is (my preference) then CLI flags should be name `alertmanager.ha_*`.

pkg/alertmanager/multitenant.go
```diff
@@ -94,6 +94,11 @@ type MultitenantAlertmanagerConfig struct {
 	Peers                flagext.StringSlice `yaml:"peers"`
 	PeerTimeout          time.Duration       `yaml:"peer_timeout"`
 
+	HABindAddr      string                 `yaml:"ha_bind_address"`
+	HAAdvertiseAddr string                 `yaml:"ha_advertise_address"`
+	HAPeers         flagext.StringSliceCSV `yaml:"ha_initial_peers"`
```
pracucci, line 99: YAML field name doesn't match its CLI flag.

pkg/alertmanager/multitenant.go
```diff
@@ -94,6 +94,11 @@ type MultitenantAlertmanagerConfig struct {
 	Peers                flagext.StringSlice `yaml:"peers"`
 	PeerTimeout          time.Duration       `yaml:"peer_timeout"`
 
+	HABindAddr      string                 `yaml:"ha_bind_address"`
```
pracucci, line 97: YAML field name doesn't match its CLI flag.

gotjosh:
I spent some time trying to debug the integration test w/o much luck. It seems like the DNS resolver for docker is not working as expected on CI as the past pass locally. My understanding is that Docker uses the host's `/etc/resolv.conf` but I'm not able to inspect that file by logging on the machine directly - any help?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 * [CHANGE] Blocks storage: compactor is now required when running a Cortex cluster with the blocks storage, because it also keeps the bucket index updated. #3583
 * [CHANGE] Blocks storage: block deletion marks are now stored in a per-tenant global markers/ location too, other than within the block location. The compactor, at startup, will copy deletion marks from the block location to the global location. This migration is required only once, so you can safely disable it via `-compactor.block-deletion-marks-migration-enabled=false` once new compactor has successfully started once in your cluster. #3583
 * [CHANGE] OpenStack Swift: the default value for the `-ruler.storage.swift.container-name` and `-swift.container-name` config options has changed from `cortex` to empty string. If you were relying on the default value, you should set it back to `cortex`. #3660
+* [CHANGE] Alertmanager: Deprecate `cluster.` flags in favor of their `alertmanager.cluster.` equivalent. The deprecated flags (and their respective YAML config options) are: #3677
```
pracucci, line 9:
```suggestion
* [CHANGE] Alertmanager: Deprecated `-cluster.` CLI flags in favor of their `-alertmanager.cluster.` equivalent. The deprecated flags (and their respective YAML config options) are: #3677
```

pkg/alertmanager/multitenant.go
```diff
@@ -115,14 +126,54 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet) {
...
+func (cfg *ClusterConfig) RegisterFlags(f *flag.FlagSet) {
+	prefix := "alertmanager.cluster."
+	f.StringVar(&cfg.ListenAddr, prefix+"listen-address", defaultClusterAddr, "Listen address and port for the cluster. Not specifying this flag disables high-availability mode.")
+	f.StringVar(&cfg.AdvertiseAddr, prefix+"advertise-address", "", "Explicit address or hostname to advertise in cluster.")
+	f.Var(&cfg.Peers, prefix+"peers", "Comma-separated list of initial peers.")
+	f.DurationVar(&cfg.PeerTimeout, prefix+"peer-timeout", defaultPeerTimeout, "Time to wait between peers to send notifications.")
+}
+
+// SupportPreviousFlagset ensures we support the previous set of cluster flags.
+func (cfg *ClusterConfig) SupportPreviousFlagset(amCfg *MultitenantAlertmanagerConfig, logger log.Logger) error {
```
pracucci, line 151: No need to return `error`. Can't happen.

pkg/alertmanager/multitenant.go
```diff
@@ -226,7 +281,7 @@ func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, logger log.L
 		}
 		err = peer.Join(cluster.DefaultReconnectInterval, cluster.DefaultReconnectTimeout)
 		if err != nil {
-			level.Warn(logger).Log("msg", "unable to join gossip mesh", "err", err)
+			level.Warn(logger).Log("msg", "unable to join gossip mesh while initializing high availability mode", "err", err)
```
pracucci, line 284:
```suggestion
			level.Warn(logger).Log("msg", "unable to join gossip mesh while initializing cluster for high availability mode", "err", err)
```

pkg/alertmanager/multitenant.go
```diff
@@ -97,12 +97,23 @@ type MultitenantAlertmanagerConfig struct {
...
 
-	Store AlertStoreConfig `yaml:"storage"`
+	Store   AlertStoreConfig `yaml:"storage"`
+	Cluster ClusterConfig    `yaml:"cluster"`
 
 	EnableAPI bool `yaml:"enable_api"`
 }
 
-const defaultClusterAddr = "0.0.0.0:9094"
+type ClusterConfig struct {
```
pracucci, line 106: To avoid mistakes in the future (until the deprecated flags won't be fully removed), I would suggest you to prefix all deprecated struct field names (not the YAML) in `MultitenantAlertmanagerConfig ` with `Deprecated`:

```
	DeprecatedClusterBindAddr      string              `yaml:"cluster_bind_address"`
	DeprecatedClusterAdvertiseAddr string              `yaml:"cluster_advertise_address"`
	DeprecatedPeers                flagext.StringSlice `yaml:"peers"`
	DeprecatedPeerTimeout          time.Duration       `yaml:"peer_timeout"`
```

_I left the comment here cause I can't comment there._

pracucci (approved):
You did great! LGTM with a couple of nits.

_I'm now going to look into the integration test issue._

pracucci:
> I'm now going to look into the integration test issue.

To keep compatibility across platforms, a container needs to be address by the host `<network name>-<container name>`, which is what the e2e framework does when you use functions like `NetworkEndpoint()`. However, such functions can't be used in this specific integration test because you have chicken-egg problem: to call `NetworkEndpoint()` you need the service first, but you need to create the first you need you need the peers list first.

I've refactored the code and fixed it (see [commit](https://github.com/cortexproject/cortex/pull/3677/commits/792f9c6f14aa875170e4d6ca7924235e9efba7a8)).

gotjosh:
> which is what the e2e framework does when you use functions like NetworkEndpoint()

Thanks @pracucci, I initially suspected it had something to do with that but when I inspected the return value of `NetworkEndpoint()` locally in the debugger it did not have the network name as a prefix so I quickly discarded it. 🤦

pracucci (approved):
LGTM, thanks!

pstibrany (approved):
LGTM. Is it worth mentioning that old flags still take precedence, if set to non-default values?

PR #3678, opened 2021-01-12
Disable access to /alertmanager/metrics, which exposes default registerer

pstibrany:
**What this PR does**: This PR disables access to `/alertmanager/metrics` path, which exposes default registerer = all Cortex metrics on given instance.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci (approved):
Makes sense to me! 👍 

Looking at the alertmanager `ui.Register()` I'm seeing other paths we may want to block, like `/debug/*` and `/-/reload`. WDYT?

pstibrany:
> Looking at the alertmanager `ui.Register()` I'm seeing other paths we may want to block, like `/debug/*` and `/-/reload`. WDYT?

Makes sense. Added.

pracucci (approved):
LGTM

PR #3679, opened 2021-01-12
Disable access to Alertmanager (user, or sending alerts), if it is not running.

pstibrany:
**What this PR does**: This PR disables access to user's Alertmanager (by user, or services trying to send alert) until `MultitenantAlertmanager` has started and synced configuration for each user.

Call to tenant's AM before `MultitenantAlertmanager` is running may result in loss of configuration for this tenant.

**Checklist**
- [x] Tests updated
- [na] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,8 @@
 * [BUGFIX] Compactor: delete `deletion-mark.json` at last when deleting a block in order to not leave partial blocks without deletion mark in the bucket if the compactor is interrupted while deleting a block. #3660
 * [BUGFIX] Blocks storage: do not cleanup a partially uploaded block when `meta.json` upload fails. Despite failure to upload `meta.json`, this file may in some cases still appear in the bucket later. By skipping early cleanup, we avoid having corrupted blocks in the storage. #3660
 * [BUGFIX] Alertmanager: disable access to `/alertmanager/metrics` (which exposes all Cortex metrics), `/alertmanager/-/reload` and `/alertmanager/debug/*`, which were available to any authenticated user with enabled AlertManager. #3678
-* [BUGFIX] Alertmanager: don't serve HTTP requests until Alertmanager has fully started. Serving HTTP requests earlier may result in loss of configuration for the user. #3679
 * [BUGFIX] Query-Frontend: avoid creating many small sub-queries by discarding cache extents under 5 minutes #3653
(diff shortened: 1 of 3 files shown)


gouthamve (dismissed):

jtlisi (commented):
Is this enough? In addition, shouldn't we require a check to the backend before storing the fallback? Othwerwise, we could end up in a similar situation in the future. For instance, once we start sharding Alertmanager tenants if a request somehow gets routed to an Alertmanager without the tenant loaded it will override the config.

pstibrany:
> Is this enough?

I think yes, for now.

> In addition, shouldn't we require a check to the backend before storing the fallback? Othwerwise, we could end up in a similar situation in the future. For instance, once we start sharding Alertmanager tenants if a request somehow gets routed to an Alertmanager without the tenant loaded it will override the config.

That's a valid point, and something keep an eye on when reviewing AM sharding. I think we should also reconsider idea of uploading "empty" config to the same path as user-uploaded config. Eg. we could give empty config ".default" suffix.

gotjosh (approved):
Good catch @pstibrany !

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 * [BUGFIX] Compactor: delete `deletion-mark.json` at last when deleting a block in order to not leave partial blocks without deletion mark in the bucket if the compactor is interrupted while deleting a block. #3660
 * [BUGFIX] Blocks storage: do not cleanup a partially uploaded block when `meta.json` upload fails. Despite failure to upload `meta.json`, this file may in some cases still appear in the bucket later. By skipping early cleanup, we avoid having corrupted blocks in the storage. #3660
 * [BUGFIX] Alertmanager: disable access to `/alertmanager/metrics` (which exposes all Cortex metrics), `/alertmanager/-/reload` and `/alertmanager/debug/*`, which were available to any authenticated user with enabled AlertManager. #3678
+* [BUGFIX] Alertmanager: don't serve HTTP requests until AlertManager has fully started. Serving HTTP requests earlier may result in loss of configuration for the user. #3679
```
gotjosh, line 47:
```suggestion
* [BUGFIX] Alertmanager: don't serve HTTP requests until Alertmanager has fully started. Serving HTTP requests earlier may result in loss of configuration for the user. #3679
```

jtlisi (approved):
LGTM modulo changelog suggestion

pracucci (approved):
LGTM!

PR #3682, opened 2021-01-13
Improve decompression within ParseProtoReader.

cyriltovena:
This uses the last weaveworks/common, which include a fix for the content-type size to be set correctly allowing to allocate buffers correctly instead of let them grow naturally.

I've also implemented an alternative decompression from a `bytes.Buffer` which avoids allocating and reading a new buffer when using httpgrpc since the underlying reader is already a `bytes.Buffer`.

I took the liberty to also improve the validation when using RawSnappy, I'm checking the length using the snappy header. This will avoid wasting CPU when the decompressed size is too big since we will know in advance.

I've added also missing  tests that covers the old and new code.

I'm planning to use this in Loki, but it should also benefits Cortex, this method is used within the push and remote_read code.


bboreham:
Should note https://github.com/weaveworks/common/pull/204, Allow specifying JAEGER_ENDPOINT, in the changelog.

Do you think we could drop framed Snappy?  It was only in a small range of Prometheus clients, a long time ago.

cyriltovena:
> Should note [weaveworks/common#204](https://github.com/weaveworks/common/pull/204), Allow specifying JAEGER_ENDPOINT, in the changelog.
> 
> Do you think we could drop framed Snappy? It was only in a small range of Prometheus clients, a long time ago.

We don't use it in Loki and it does make the code more complex, so I'm in for this, if you want.

cyriltovena:
I also realized during writing tests, that FrameSnappy weight more.

bboreham:
This was where framed Snappy was removed, before Prometheus 1.7: https://github.com/prometheus/prometheus/pull/2696

cyriltovena:
> This was where framed Snappy was removed, before Prometheus 1.7: [prometheus/prometheus#2696](https://github.com/prometheus/prometheus/pull/2696)

I wonder if this is fine to remove ? I guess we won't support 1.7 then ? Is this fine ? I'm not totally sure.

@gouthamve @pracucci any chance you might have a reason to keep this ? Do we expect all users above 1.7 Prometheus ?

pracucci:
> @gouthamve @pracucci any chance you might have a reason to keep this ? Do we expect all users above 1.7 Prometheus ?

Yesterday we had a brief discussion in this [Slack thread](https://cloud-native.slack.com/archives/CCYDASBLP/p1610549969043000) and LGTM from me and Goutham. As Bryan said:

> I think it's exactly 1.6 that we would remove support for, since Cortex did not support <1.6 already
> 1.6 was April 2017 and 1.7 June 2017

As far as the CHANGELOG mention we're dropping support for Prometheus 1.7, LGTM.

cyriltovena:
I would have prefer another PR. but ho well too late.

bboreham:
Not clear why it's too late. I don't mind it being another PR.

pkg/util/http.go
```diff
@@ -81,38 +82,63 @@ type CompressionType int
...
-// ParseProtoReader parses a compressed proto from an io.Reader.
-func ParseProtoReader(ctx context.Context, reader io.Reader, expectedSize, maxSize int, req proto.Message, compression CompressionType) error {
-	var body []byte
-	var err error
+func decompressRequest(ctx context.Context, reader io.Reader, expectedSize, maxSize int, compression CompressionType) (body []byte, err error) {
 	sp := opentracing.SpanFromContext(ctx)
 	if sp != nil {
 		sp.LogFields(otlog.String("event", "util.ParseProtoRequest[start reading]"))
+		defer func() {
+			sp.LogFields(otlog.String("event", "util.ParseProtoRequest[unmarshal]"),
```
pracucci, line 115: I would move this to `ParseProtoReader()` for clarity.

pkg/util/http.go
```diff
@@ -81,38 +82,63 @@ type CompressionType int
...
+}
+
+func decompressFromReader(reader io.Reader, expectedSize, maxSize int, compression CompressionType, sp opentracing.Span) ([]byte, error) {
+	var (
+		buf  bytes.Buffer
+		body []byte
+		err  error
+	)
 	if expectedSize > 0 {
 		if expectedSize > maxSize {
```
pracucci, line 140: Why is this check here, instead of the beginning of `ParseProtoReader()`?

pkg/util/http.go
```diff
@@ -122,45 +148,72 @@ func ParseProtoReader(ctx context.Context, reader io.Reader, expectedSize, maxSi
...
 		body = buf.Bytes()
 		if sp != nil {
 			sp.LogFields(otlog.String("event", "util.ParseProtoRequest[decompress]"),
 				otlog.Int("size", len(body)))
 		}
-		if err == nil && len(body) <= maxSize {
+		if err != nil {
+			return nil, err
+		}
+		if len(body) > maxSize {
```
pracucci, line 161: Why don't we call `decompressFromBuffer()` instead of the code below? Looks doing the same thing.

pracucci (commented):
Good job! Changes make a lot of sense to me. I left few comments, but overall LGTM 👏

pkg/util/http.go
cyriltovena: moved that one into `ParseProtoReader`.

pkg/util/http.go
cyriltovena: Good feedback, refactored a bit and made the change.

cyriltovena: I didn't had tests before and I was reluctant to change too much, but now I do have tests and they still all pass.

pkg/util/http.go
cyriltovena: I actually moved it here for clarity but still applied your feedback, it's more or less the same as you need to push down the span into all branch.

My first implementation only start the usage of the span in the `decompressRequest`.

pracucci (approved):
Good job, LGTM!

pkg/util/http.go
```diff
@@ -163,6 +115,88 @@ func ParseProtoReader(ctx context.Context, reader io.Reader, expectedSize, maxSi
...
+		buf.Grow(expectedSize + bytes.MinRead) // extra space guarantees no reallocation
+	}
+	switch compression {
+	case NoCompression:
+		// Read from LimitReader with limit max+1. So if the underlying
+		// reader is over limit, the result will be bigger than max.
+		_, err = buf.ReadFrom(io.LimitReader(reader, int64(maxSize)+1))
+		body = buf.Bytes()
+	case RawSnappy:
+		_, err = buf.ReadFrom(reader)
```
bboreham, line 152: Is this a possible DoS attack? (I see it is what the old code did)
Seems we could use the same `ReadFrom` in either case, then check `compression` after.

bboreham: Could fix it up later; I don't mind merging this as-is.

bboreham (approved):
Thanks!

pstibrany (approved):
LGTM.

Nit: Both `decompressFromReader` and `decompressFromBuffer` could benefit from using early returns.

pkg/util/http.go
cyriltovena: Yeah I see the potential issue, not sure if there's a better option then using a limitReader in both case ?

Decoding the length still leave us open for hijacked/fake requests.

I made the change let me know what you think ?

pkg/util/http.go
```diff
@@ -163,6 +115,88 @@ func ParseProtoReader(ctx context.Context, reader io.Reader, expectedSize, maxSi
 	return nil
 }
 
+func decompressRequest(ctx context.Context, reader io.Reader, expectedSize, maxSize int, compression CompressionType, sp opentracing.Span) (body []byte, err error) {
```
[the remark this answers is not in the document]
cyriltovena: Yep this was leftover from previous span extraction code.

pkg/util/http.go
bboreham: I think using LimitReader in all paths is correct, if the point is to stop someone blowing up the process.

Is 'result bigger than max' actually detected in the NoCompression case now?
Suggest just having one `ReadFrom` call then check the len, before we get into a routine named `decompress`

pkg/util/http.go
```diff
@@ -163,6 +115,89 @@ func ParseProtoReader(ctx context.Context, reader io.Reader, expectedSize, maxSi
...
+		err  error
+	)
+	if expectedSize > 0 {
+		buf.Grow(expectedSize + bytes.MinRead) // extra space guarantees no reallocation
+	}
+	reader = io.LimitReader(reader, int64(maxSize)+1)
+	switch compression {
+	case NoCompression:
+		// Read from LimitReader with limit max+1. So if the underlying
+		// reader is over limit, the result will be bigger than max.
```
pracucci, line 149: Move this comment above, where `io.LimitReader` is used, please.

pkg/util/http.go
cyriltovena: yes because we read a bit more the defer  in `decompressRequest` will return the error.

PR #3684, opened 2021-01-13
Add test for anchored YAML loading.

gouthamve:
Also we were git-ignoring that folder so fix config to ignore only the
binary in root folder.


pstibrany (approved):

pracucci (commented):
Fantastic, thanks!

PR #3687, opened 2021-01-14
fix: Stale series with ruler evaluation delay

simonswine:
**What this PR does**:

When ruler evaluation delay is enabled, the upstream rule evaluation logic adds staleness markers for missing series at the actual timestamp when the evaluation is run rather than the timestamp when the series was missing: (see:
https://github.com/prometheus/prometheus/blob/6c56a1faaaad07317ff585bda75b99bdba0517ad/rules/manager.go#L647-L660).

This can cause larger gaps than there actually are, as once the data becomes available again it can't be written as it would be out-of-order.

This fix shifts those staleness marker back in time by the evaluation delay set for the individual tenant.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
 * [BUGFIX] Blocks storage: do not cleanup a partially uploaded block when `meta.json` upload fails. Despite failure to upload `meta.json`, this file may in some cases still appear in the bucket later. By skipping early cleanup, we avoid having corrupted blocks in the storage. #3660
 * [BUGFIX] Alertmanager: disable access to `/alertmanager/metrics` (which exposes all Cortex metrics), `/alertmanager/-/reload` and `/alertmanager/debug/*`, which were available to any authenticated user with enabled AlertManager. #3678
 * [BUGFIX] Query-Frontend: avoid creating many small sub-queries by discarding cache extents under 5 minutes #3653
+* [BUGFIX] Ruler: Ensure the stale markers generated for evaluated rules respect the configured `-ruler.evaluation-delay-duration`. This will avoid issues with samples with NaN be persisted with timestamps set ahead of the next rule evaluation. #3687
 
 ## 1.6.0
 
--- a/integration/ruler_test.go
+++ b/integration/ruler_test.go
@@ -5,6 +5,7 @@ package integration
 import (
 	"context"
 	"fmt"
+	"math"
 	"net/http"
 	"path/filepath"
 	"strings"
@@ -14,6 +15,8 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"github.com/prometheus/prometheus/pkg/value"
+	"github.com/prometheus/prometheus/prompb"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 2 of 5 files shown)


gotjosh:
cc: @gouthamve

integration/ruler_test.go
```diff
@@ -179,45 +182,106 @@ func TestRulerEvaluationDelay(t *testing.T) {
...
 	user := "fake"
 
+	evaluationDelay := time.Minute * 5
+
 	configOverrides := map[string]string{
 		"-ruler.storage.local.directory":   filepath.Join(e2e.ContainerSharedDir, "ruler_configs"),
 		"-ruler.poll-interval":             "2s",
 		"-ruler.rule-path":                 filepath.Join(e2e.ContainerSharedDir, "rule_tmp/"),
-		"-ruler.evaluation-delay-duration": "5m", // 5 minutes is clarifying when seeing when a rule is evaluated
+		"-ruler.evaluation-delay-duration": evaluationDelay.String(), // 5 minutes is clarifying when seeing when a rule is evaluated
```
jdbaldry, line 191: I'm not sure I understand the comment here

jdbaldry (commented):
One typo

integration/ruler_test.go
jdbaldry: Oh I realize you didn't add the comment, sorry.

simonswine: I don't understand it, so it didn't touch it

simonswine: I have removed it now

integration/ruler_test.go
```diff
@@ -179,45 +182,106 @@ func TestRulerEvaluationDelay(t *testing.T) {
...
+	matrix := result.(model.Matrix)
+	require.GreaterOrEqual(t, 1, matrix.Len(), "expect at least a series returned")
+
+	// Iterate through the values recorded and ensure they exist as expected.
+	inputPos := 0
 	for _, m := range matrix {
 		for _, v := range m.Values {
-			diff := now.Unix() - int64(v.Value)
-			require.GreaterOrEqual(t, diff, maxDiff)
+			// Skip values for stale posistions
```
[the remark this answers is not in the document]
simonswine: Thanks  :clap:

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@
 * [BUGFIX] Blocks storage: do not cleanup a partially uploaded block when `meta.json` upload fails. Despite failure to upload `meta.json`, this file may in some cases still appear in the bucket later. By skipping early cleanup, we avoid having corrupted blocks in the storage. #3660
 * [BUGFIX] Alertmanager: disable access to `/alertmanager/metrics` (which exposes all Cortex metrics), `/alertmanager/-/reload` and `/alertmanager/debug/*`, which were available to any authenticated user with enabled AlertManager. #3678
 * [BUGFIX] Query-Frontend: avoid creating many small sub-queries by discarding cache extents under 5 minutes #3653
+* [BUGFIX] Ruler: Minimize gaps on rule evaluations with stale input and enabled ruler evaluation delay. #3687
```
jtlisi, line 48: **suggestion:** I think it would be good to make it clear what was fixed in addition to the symptom is solves.
```suggestion
* [BUGFIX] Ruler: Ensure the stale markers generated for evaluated rules respect the set evaluation delay. This will avoid issues with samples with NaN be persisted with timestamps set ahead of the next rule evaluation. #3687
```

jtlisi (approved):
LGTM, modulo on minor changelog nit

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@
 * [BUGFIX] Blocks storage: do not cleanup a partially uploaded block when `meta.json` upload fails. Despite failure to upload `meta.json`, this file may in some cases still appear in the bucket later. By skipping early cleanup, we avoid having corrupted blocks in the storage. #3660
 * [BUGFIX] Alertmanager: disable access to `/alertmanager/metrics` (which exposes all Cortex metrics), `/alertmanager/-/reload` and `/alertmanager/debug/*`, which were available to any authenticated user with enabled AlertManager. #3678
 * [BUGFIX] Query-Frontend: avoid creating many small sub-queries by discarding cache extents under 5 minutes #3653
+* [BUGFIX] Ruler: Ensure the stale markers generated for evaluated rules respect the set evaluation delay. This will avoid issues with samples with NaN be persisted with timestamps set ahead of the next rule evaluation. #3687
```
pracucci, line 48:
```suggestion
* [BUGFIX] Ruler: Ensure the stale markers generated for evaluated rules respect the configured `-ruler.evaluation-delay-duration`. This will avoid issues with samples with NaN be persisted with timestamps set ahead of the next rule evaluation. #3687
```

integration/ruler_test.go
```diff
@@ -179,45 +182,106 @@ func TestRulerEvaluationDelay(t *testing.T) {
...
+	// Insert metrics
+	res, err := c.Push([]prompb.TimeSeries{series})
 	require.NoError(t, err)
-	require.Equal(t, model.ValMatrix, result.Type())
+	require.Equal(t, 200, res.StatusCode)
 
-	// Iterate through the values recorded and ensure they exist in the past.
-	matrix := result.(model.Matrix)
+	// Wait until the rule is evaluated for the first time
+	require.NoError(t, cortex.WaitSumMetrics(e2e.Greater(1.0), "cortex_prometheus_rule_evaluations_total"))
```
pracucci, line 236: Since it's greater than 1 it will wait for 2 evaluations. Is that expected or you wanted `e2e.Greater(0)`?

integration/ruler_test.go
```diff
@@ -179,45 +182,106 @@ func TestRulerEvaluationDelay(t *testing.T) {
...
+	// Generate series that includes stale nans
+	var samplesToSend int = 10
+	series := prompb.TimeSeries{
+		Labels: []prompb.Label{
+			{Name: "__name__", Value: "a_sometimes_stale_nan_series"},
+			{Name: "instance", Value: "sometimes-stale"},
+		},
+	}
+	series.Samples = make([]prompb.Sample, samplesToSend)
+	posStale := map[int]struct{}{
```
pracucci, line 215: `posStale := 2` may be clearer.

integration/ruler_test.go
```diff
@@ -179,45 +182,106 @@ func TestRulerEvaluationDelay(t *testing.T) {
...
+		2: {},
+	}
+
+	// Create samples, that are delayed by the evaluation delay with increasing values.
+	for pos := range series.Samples {
+		series.Samples[pos].Timestamp = e2e.TimeToMilliseconds(now.Add(-evaluationDelay)) + 1000*int64(pos)
```
pracucci, line 221: Is this the same but more clear, or am I missing something?

```suggestion
		series.Samples[pos].Timestamp = e2e.TimeToMilliseconds(now.Add(-evaluationDelay).Add(pos * time.Second))
```

integration/ruler_test.go
```diff
@@ -179,45 +182,106 @@ func TestRulerEvaluationDelay(t *testing.T) {
...
+	require.Equal(t, 1, vector.Len(), "expect one sample returned")
 
 	// 290 seconds gives 10 seconds of slack between the rule evaluation and the query
-	// to account for CI latency, but ensures the latest evalation was in the past.
-	var maxDiff int64 = 290
+	// to account for CI latency, but ensures the latest evaluation was in the past.
+	var maxDiff int64 = 290_000
+	require.GreaterOrEqual(t, e2e.TimeToMilliseconds(time.Now())-int64(vector[0].Value)*1000, maxDiff)
+
+	// Wait until the rule is evaluated once per sample sent
```
pracucci, line 251: I'm not sure I do understand this. Why you need to wait `samplesToSend` evaluations?

integration/ruler_test.go
pracucci: Also, I believe this test can be flaky because of this check. What you want is to wait 1 more evaluation after the `Push()`, so ideally you should get the the current `cortex_prometheus_rule_evaluations_total` and `WaitSumMetric()` by that value + 1.

pracucci (commented):
Thanks for fixing this annoying issue! I took a look if there's a cleaner way to implement the ruler evaluation delay at all (to avoid the "hack/fix" done in the appender in this PR) but I haven't found any, so LGTM.

Before merging, I would like you to take a look at the comment I've left. Thanks!

integration/ruler_test.go
simonswine: Good point, I have implemented as suggest, `Push -> SumMetric -> WaitSumMetric(Greater(result))`

integration/ruler_test.go
simonswine: That's a remnant from a time when I had two staleness periods :+1:

integration/ruler_test.go
simonswine: I have added a comment to clarify:
```
	// Wait until all the pushed samples have been evaluated by the rule. This
	// ensures that rule results are successfully written even after a
	// staleness period.```

pracucci (approved):
LGTM, thanks!

PR #3700, opened 2021-01-15
Runtime config endpoint now supports diff parameter

### LINKED ISSUE #3672: Add the diff option for the runtime_config endpoint

The new `runtime_config` endpoint could be a little verbose and hard to read because is returning all the defaults values

We need to add a new diff option, like the config endpoint, to just show the changed values and discard the defaults values

**Describe alternatives you've considered**

**Additional context**

This comes from #1324

ethervoid:
**What this PR does**:

This PR introduces a new parameter for the `/runtime_config` endpoint that will allow the user to filter the outcome and just get the non-default values of the runtime configuration.

I've tried to add tests for the `runtimeConfigurationHandler` but given that the handler depends on `runtimeConfigManager`  the tests become very complex unless we're able to mock the `runtimeConfigManager`

Fixes #3672

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   * Avoid unnecessary `runtime.GC()` during compactions.
   * Prevent compaction loop in TSDB on data gap.
 * [ENHANCEMENT] Return server side performance metrics for query-frontend (using Server-timing header). #3685
+* [ENHANCEMENT] Runtime Config: Add a `mode` query parameter for the runtime config endpoint. `/runtime_config?mode=diff` now shows the YAML runtime configuration with all values that differ from the defaults. #3700
 * [BUGFIX] HA Tracker: don't track as error in the `cortex_kv_request_duration_seconds` metric a CAS operation intentionally aborted. #3745
 
 ## 1.7.0 in progress
--- a/pkg/cortex/runtime_config.go
+++ b/pkg/cortex/runtime_config.go
@@ -112,27 +112,47 @@ func runtimeConfigHandler(runtimeCfgManager *runtimeconfig.Manager, defaultLimit
 		}
 		switch r.URL.Query().Get("mode") {
 		case "diff":
-			defaultLimitsObj, err := util.YAMLMarshalUnmarshal(defaultLimits)
-			if err != nil {
-				http.Error(w, err.Error(), http.StatusInternalServerError)
-				return
-			}
 			runtimeCfgObj, err := util.YAMLMarshalUnmarshal(runtimeConfig)
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
-			limitsCfgObj, err := util.YAMLMarshalUnmarshal(runtimeCfgObj["overrides"])
-			if err != nil {
-				http.Error(w, err.Error(), http.StatusInternalServerError)
-				return
+			if runtimeCfgObj["overrides"] != nil {
+				defaultLimitsObj, err := util.YAMLMarshalUnmarshal(defaultLimits)
+				if err != nil {
+					http.Error(w, err.Error(), http.StatusInternalServerError)
+					return
+				}
+				limitsCfgObj, err := util.YAMLMarshalUnmarshal(runtimeCfgObj["overrides"])
+				if err != nil {
+					http.Error(w, err.Error(), http.StatusInternalServerError)
+					return
+				}
+				limitsDiff, err := diffLimitsConfig(defaultLimitsObj, limitsCfgObj)
+				if err != nil {
+					http.Error(w, err.Error(), http.StatusInternalServerError)
+					return
+				}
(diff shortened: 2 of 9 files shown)


ethervoid:
@pstibrany Don't know where I should ping to ask for review so given that you gently did the last review... :D

pstibrany (commented):
Thank you for your PR! Very useful, especially for overrides!

As it is now, "diff" doesn't show diff for entire "runtime config", only for overrides.

I'd also suggest defining the handler in the `pkg/cortex/runtime_config.go` file, and passing only fully constructed `http.Handler` to the API registration. This may require moving some methods (eg. `yamlMarshalUnmarshal`) to `util` package. There should be no need for `api` package to depend on `validation`.

CHANGELOG.md
```diff
@@ -36,6 +36,8 @@
 * [ENHANCEMENT] Disabled in-memory shuffle-sharding subring cache in the store-gateway, ruler and compactor. This should reduce the memory utilisation in these services when shuffle-sharding is enabled, without introducing a significantly increase CPU utilisation. #3601
 * [ENHANCEMENT] Shuffle sharding: optimised subring generation used by shuffle sharding. #3601
 * [ENHANCEMENT] New /runtime_config endpoint that returns the defined runtime configuration in YAML format. The returned configuration includes overrides. #3639
+* [ENHANCEMENT] API: Add a `mode` query parameter for the runtime config endpoint: #3645
```
pracucci, line 39: We're used to link the PR number, not the issue number. Please change `#3645` > `#3700`. Thanks!

pracucci: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

ethervoid: On it!

ethervoid:
@pstibrany I'll tackle the changes today, thank you for the review.

ethervoid:
@pstibrany Changes made.

I've made it possible to do a diff in the kv.Multi object but right now I can't see clear default values for that struct, like the ones you can see in the overrides.

pkg/cortex/runtime_config.go
```diff
@@ -83,3 +85,78 @@ func multiClientRuntimeConfigChannel(manager *runtimeconfig.Manager) func() <-ch
...
+				}
+				runtimeCfgObj["multi_kv_config"] = multKVDiff
+			}
+			output = runtimeCfgObj
+		default:
+			output = runtimeConfig
+		}
+		util.WriteYAMLResponse(w, output)
+	}
+}
```
pstibrany, line 162:
```suggestion
func runtimeConfigHandler(runtimeCfgManager *runtimeconfig.Manager, defaultLimits validation.Limits) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		cfg, ok := runtimeCfgManager.GetConfig().(*runtimeConfigValues)
		if !ok || cfg == nil {
			util.WriteTextResponse(w, "runtime config file doesn't exist")
			return
		}

		var output interface{}
		switch r.URL.Query().Get("mode") {
		case "diff":
			// Default runtime config is just empty struct, but to make diff work,
			// we set defaultLimits for every tenant that exists in runtime config.
			defaultCfg := runtimeConfigValues{}
			defaultCfg.TenantLimits = map[string]*validation.Limits{}
			for k, v := range cfg.TenantLimits {
				if v != nil {
					defaultCfg.TenantLimits[k] = &defaultLimits
				}
			}

			cfgYaml, err := util.YAMLMarshalUnmarshal(cfg)
			if err != nil {
				http.Error(w, err.Error(), http.StatusInternalServerError)
				return
			}

			defaultCfgYaml, err := util.YAMLMarshalUnmarshal(defaultCfg)
			if err != nil {
				http.Error(w, err.Error(), http.StatusInternalServerError)
				return
			}

			output, err = util.DiffConfig(defaultCfgYaml, cfgYaml)
			if err != nil {
				http.Error(w, err.Error(), http.StatusInternalServerError)
				return
			}

		default:
			output = cfg
		}
		util.WriteYAMLResponse(w, output)
	}
}
```

pstibrany (commented):
Thank you for your continuous work on this PR!

I took a liberty to modify and simplify `runtimeConfigHandler`, to avoid comparing individual fields of `runtimeConfigValues` but work with default value which is simply empty struct. (See suggestion)

The rest of the PR looks good.

pkg/cortex/runtime_config.go
ethervoid: Thank you for taking the time to do it. 

I tried to do something similar in the past PR but I couldn't because of package visibility. TBH didn't think about this change in this PR so thank you!

pstibrany (approved):
Thank you for PR and addressing my feedback, this is going to be very useful!

pracucci (approved):
LGTM thanks!

PR #3703, opened 2021-01-17
Include parameter name when returning HTTP 400 for query_range API

### LINKED ISSUE #3699: Better error message when parameter is missing when calling query_range API
Currently if I forget to give the `step` and only give `start` and `end` parameters when calling query_range, I get the following error message: 

```
cannot parse "" to a valid duration
```
It took me a while to find out that `step` is required. It would had be nice the error message mentions about the `step`  parameter.

alvinlin123:
**What this PR does**:
Add parameter name to HTTP 400 message returned to the user for self-debugging.

Fixes #3699

**Checklist**
- [x] Tests updated
- [] Documentation added


CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 * [ENHANCEMENT] Disabled in-memory shuffle-sharding subring cache in the store-gateway, ruler and compactor. This should reduce the memory utilisation in these services when shuffle-sharding is enabled, without introducing a significantly increase CPU utilisation. #3601
 * [ENHANCEMENT] Shuffle sharding: optimised subring generation used by shuffle sharding. #3601
 * [ENHANCEMENT] New /runtime_config endpoint that returns the defined runtime configuration in YAML format. The returned configuration includes overrides. #3639
+* [ENHANCEMENT] query_range API to include parameter name in HTTP 400 message. #3699
```
pracucci, line 47: A couple of suggestions:
1. We're used to link the PR number instead of the issue number
2. If the change affects only 1 specific Cortex service then we prefix it

```suggestion
* [ENHANCEMENT] Query-frontend: included the parameter name failed to validate in HTTP 400 message. #3703
```

pracucci (approved):
LGTM, thanks!

pstibrany (approved):
LGTM (with Marco's comments).

(Personally I find `util.ParseTime` returning httpgrpc.Error directly ugly, leading to ugliness in `decorateWithParamName`, but there is no need to spend more time on it.)

CHANGELOG.md
alvinlin123: Thanks for the suggestion and fixing it for me.

PR #3705, opened 2021-01-18
Expose a metric tracking the oldest timestamp of unshipper block in ingester

pracucci:
**What this PR does**:
Currently, there's no way to alert if any compacted block is not shipped from ingester to storage within a reasonable time. In this PR I'm proposing to introduce a new metric `cortex_ingester_oldest_unshipped_block_timestamp_seconds` tracking the unix timestamp of the oldest TSDB block not shipped to the storage yet.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@
 * [ENHANCEMENT] Distributor: change the error message returned when a received series has too many label values. The new message format has the series at the end and this plays better with Prometheus logs truncation. #3718
   - From: `sample for '<series>' has <value> label names; limit <value>`
   - To: `series has too many labels (actual: <value>, limit: <value>) series: '<series>'`
+* [ENHANCEMENT] Ingester: exposed metric `cortex_ingester_oldest_unshipped_block_timestamp_seconds`, tracking the unix timestamp of the oldest TSDB block not shipped to the storage yet. #3705
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
 * [BUGFIX] Memberlist: Entry in the ring should now not appear again after using "Forget" feature (unless it's still heartbeating). #3603
 * [BUGFIX] Ingester: do not close idle TSDBs while blocks shipping is in progress. #3630
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -1575,16 +1575,19 @@ func (i *Ingester) shipBlocks(ctx context.Context) {
 		}
 		defer userDB.casState(activeShipping, active)
 
-		if uploaded, err := userDB.shipper.Sync(ctx); err != nil {
+		uploaded, err := userDB.shipper.Sync(ctx)
+		if err != nil {
 			level.Warn(util.Logger).Log("msg", "shipper failed to synchronize TSDB blocks with the storage", "user", userID, "uploaded", uploaded, "err", err)
 		} else {
 			level.Debug(util.Logger).Log("msg", "shipper successfully synchronized TSDB blocks with storage", "user", userID, "uploaded", uploaded)
 		}
 
 		// The shipper meta file could be updated even if the Sync() returned an error,
-		// so it's safer to always update it.
-		if err := userDB.updateCachedShippedBlocks(); err != nil {
-			level.Error(util.Logger).Log("msg", "failed to update cached shipped blocks after shipper synchronisation", "user", userID, "err", err)
+		// so it's safer to update it each time at least a block has been uploaded.
+		if uploaded > 0 {
+			if err := userDB.updateCachedShippedBlocks(); err != nil {
+				level.Error(util.Logger).Log("msg", "failed to update cached shipped blocks after shipper synchronisation", "user", userID, "err", err)
+			}
 		}
 
 		return nil
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -108,6 +108,10 @@ type userTSDB struct {
 	// for statistics
 	ingestedAPISamples  *ewmaRate
 	ingestedRuleSamples *ewmaRate
+
+	// Cached shipped blocks.
+	shippedBlocksMtx sync.Mutex
+	shippedBlocks    map[ulid.ULID]struct{}
 }
 
 // Explicitly wrapping the tsdb.DB functions that we use.
@@ -237,15 +241,10 @@ func (u *userTSDB) blocksToDelete(blocks []*tsdb.Block) map[ulid.ULID]struct{} {
 		return deletable
 	}
 
-	shippedBlocks, err := u.getShippedBlocks()
-	if err != nil {
-		// If there is any issue with the shipper, we should be conservative and not delete anything.
-		level.Error(util.Logger).Log("msg", "failed to read shipper meta during deletion of blocks", "user", u.userID, "err", err)
-		return nil
-	}
+	shippedBlocks := u.getCachedShippedBlocks()
 
 	result := map[ulid.ULID]struct{}{}
-	for _, shippedID := range shippedBlocks {
+	for shippedID := range shippedBlocks {
 		if _, ok := deletable[shippedID]; ok {
 			result[shippedID] = struct{}{}
 		}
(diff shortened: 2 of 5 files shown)


pkg/ingester/ingester_v2.go
```diff
@@ -262,6 +262,38 @@ func (u *userTSDB) getShippedBlocks() ([]ulid.ULID, error) {
 	return shipperMeta.Uploaded, nil
 }
 
+// getOldestUnshippedBlockTime returns the unix timestamp with milliseconds precision of the oldest
+// TSDB block not shipped to the storage yet, or 0 if all blocks have been shipped.
+func (u *userTSDB) getOldestUnshippedBlockTime() (uint64, error) {
+	// Read the list of shipper blocks. If the shipper metadata file doesn't exist
+	// (eg. because the shipper hasn't run yet) we're considering it as if no block
+	// has been shipped.
+	shipped, err := u.getShippedBlocks()
```
pstibrany, line 271: `getShippedBlocks` is used twice, each time we build a map from it. Perhaps `getShippedBlocks` should build and return map directly?

pkg/ingester/ingester_v2.go
```diff
@@ -412,10 +434,16 @@ func NewV2(cfg Config, clientConfig client.Config, limits *validation.Overrides,
...
 	if registerer != nil {
 		registerer.Unregister(i.metrics.memSeries)
+
 		promauto.With(registerer).NewGaugeFunc(prometheus.GaugeOpts{
 			Name: "cortex_ingester_memory_series",
 			Help: "The current number of series in memory.",
-		}, i.numSeriesInTSDB)
+		}, i.getMemorySeriesMetric)
+
+		promauto.With(registerer).NewGaugeFunc(prometheus.GaugeOpts{
```
pstibrany, line 443: I'm wondering if it makes sense to include oldest sample as well, but I guess it doesn't give us anything extra.

pkg/ingester/ingester_v2.go
```diff
@@ -1438,6 +1466,22 @@ func (i *Ingester) numSeriesInTSDB() float64 {
...
 
+// getOldestUnshippedBlockMetric returns the unix timestamp of the oldest unshipped block or
+// 0 if all blocks have been shipped.
+func (i *Ingester) getOldestUnshippedBlockMetric() float64 {
+	i.userStatesMtx.RLock()
+	defer i.userStatesMtx.RUnlock()
+
+	oldest := uint64(0)
+	for _, db := range i.TSDBState.dbs {
+		if ts, err := db.getOldestUnshippedBlockTime(); err == nil && (oldest == 0 || ts < oldest) {
```
pstibrany, line 1477: Instead of computing this value on each metric collection, should we cache it and update after each call to the shipper? As it is now, each metric collection needs to read shipper json file for each tenant, which could be slow if there are many tenants.

pstibrany (approved):
LGTM, but I'd like to see your take on caching the value instead.

pkg/ingester/ingester_v2.go
pracucci: I can benchmark it and then we can decide if extra complexity is worth it. But agree checking it.

pkg/ingester/ingester_v2.go
pracucci: I'm not sure which value would give exposing the oldest sample (I guess you mean the oldest block `mint`). I believe that from the user's perspective is easier to reason about the time the block has been compacted so that this metric effectively measure the delay between it and when it was successfully shipped to the storage.

pstibrany: > I guess you mean the oldest block `mint`

Yes. It was just an idea said out loud while thinking if block timestamp is the right thing to expose, but I don't see much value in it.

pkg/ingester/ingester_v2.go
pracucci: `blocksToDelete()` doesn't build a map of out `getShipperBlocks()`. It just iterates over it.

pkg/ingester/ingester_v2.go
pracucci: I've written a simple benchmark ([commit](https://github.com/cortexproject/cortex/pull/3705/commits/9447e9a184064c4f1d91c846b80083034f28d7f5)) and on my MacBook it takes 31μs/op. If you have 1K tenants, it would be 31ms. I've the feeling caching for it, at this stage, would just be premature optimisation.

```
BenchmarkUserTSDB_getOldestUnshippedBlockTime-12    	   37930	     30898 ns/op	    4486 B/op	      26 allocs/op
```

pstibrany: Is this a valid benchmark? On a busy server `getOldestUnshippedBlockTime` will not benefit from heavy caching of all IO operations as much as it does in this standalone test.

I don't think it would be a premature optimization, just a better approach. For 1K tenants, current approach needs to open/read/close 1000 files on each metric collection (=3000 syscalls). Instead we could simply read one atomic variable per tenant, and update it only when it can actually change. I don't see it as "premature optimization".

ranton256: +1, I share concern this could be pretty inefficient for large number of tenants.

ranton256 (commented):
Thanks for doing this, functionally, in terms of the metric exported it looks good to me, but I second pstibrany's concern about the cost of calculating the metric at scrape time so would prefer the alternative implementation suggested.

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@
 * [ENHANCEMENT] Distributor: change the error message returned when a received series has too many label values. The new message format has the series at the end and this plays better with Prometheus logs truncation. #3718
   - From: `sample for '<series>' has <value> label names; limit <value>`
   - To: `series has too many labels (actual: <value>, limit: <value>) series: '<series>'`
+* [ENHANCEMENT] Ingester: exposed metric `cortex_ingester_oldest_unshipped_block_timestamp_seconds`, tracking the unix timestamp of the oldest TSDB block not shipped to the storage yet. #3705
```
pracucci, line 71: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

pkg/ingester/ingester_v2.go
pracucci: Ok, got it. I pushed a change to keep the shipped blocks in memory (they're tiny) and reload them each time any block is uploaded. I've also took the opportunity to improve the tests to not use a mocked shipper but the real one backed by a filesystem bucket.

What do you think?

pkg/ingester/ingester_v2.go
```diff
@@ -1438,8 +1479,27 @@ func (i *Ingester) numSeriesInTSDB() float64 {
...
+	}
+
+	return float64(oldest / 1000)
+}
+
 func (i *Ingester) shipBlocksLoop(ctx context.Context) error {
-	shipTicker := time.NewTicker(i.cfg.BlocksStorageConfig.TSDB.ShipInterval)
+	// We add a slight jitter to make sure that if the head compaction interval and ship interval are set to the same
+	// value they don't clash (if they both continuously run at the same exact time, the head compaction may not run
+	// because can't successfully cas the state).
```
ranton256, line 1501: What do you mean by "cas the state"? Not sure if I'm confused about a term  I don't know or a typo ?

ranton256 (approved):
Thanks for the update.

pkg/ingester/ingester_v2.go
pracucci: By cas I mean "Compare And Swap" (we have a function called `casState()`). Maybe "CAS" (upper case) is more clear? Should I explode it to "compare and swap"?

pkg/ingester/ingester_v2.go
```diff
@@ -1512,12 +1572,21 @@ func (i *Ingester) shipBlocks(ctx context.Context) {
...
 
-		if uploaded, err := userDB.shipper.Sync(ctx); err != nil {
+		uploaded, err := userDB.shipper.Sync(ctx)
+		if err != nil {
 			level.Warn(util.Logger).Log("msg", "shipper failed to synchronize TSDB blocks with the storage", "user", userID, "uploaded", uploaded, "err", err)
 		} else {
 			level.Debug(util.Logger).Log("msg", "shipper successfully synchronized TSDB blocks with storage", "user", userID, "uploaded", uploaded)
 		}
 
+		// The shipper meta file could be updated even if the Sync() returned an error,
```
pstibrany, line 1582: Shipper meta file may have changed even if no blocks were uploaded -- when shipper has removed obsolete entries from the list. I guess we don't need to care about that.

pstibrany (approved):
:+1: Thanks!

Also nice cleanup in the tests!

pkg/ingester/ingester_v2.go
pracucci: Yes, right. Shouldn't be a problem for our use case because technically the blocks removed from the meta because of the retention have been shipped anyway. I've updated the code comment.

PR #3706, opened 2021-01-18
Do not log "failed to load config" if runtime config file is empty

pracucci:
**What this PR does**:
We recently introduced an empty `runtime.yaml` in the local dev env setup `development/tsdb-blocks-storage-s3/config/`, but I've just realised the config fails to load because empty (no YAML node). Should an empty config be legit? If so, then this PR fixes it.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/runtime_config.go
+++ b/pkg/cortex/runtime_config.go
@@ -31,8 +31,7 @@ func loadRuntimeConfig(r io.Reader) (interface{}, error) {
 	decoder.SetStrict(true)
 
 	// Decode the first document. An empty document (EOF) is OK.
-	err := decoder.Decode(&overrides)
-	if err != nil && !errors.Is(err, io.EOF) {
+	if err := decoder.Decode(&overrides); err != nil && !errors.Is(err, io.EOF) {
 		return nil, err
 	}
 
--- a/pkg/cortex/runtime_config.go
+++ b/pkg/cortex/runtime_config.go
@@ -35,7 +35,7 @@ func loadRuntimeConfig(r io.Reader) (interface{}, error) {
 		return nil, err
 	}
 
-	// Ensure the provided YAML config is not composed by multiple documents,
+	// Ensure the provided YAML config is not composed of multiple documents,
 	if err := decoder.Decode(&runtimeConfigValues{}); !errors.Is(err, io.EOF) {
 		return nil, errMultipleDocuments
 	}
--- a/pkg/cortex/runtime_config.go
+++ b/pkg/cortex/runtime_config.go
@@ -1,14 +1,18 @@
 package cortex
 
 import (
+	"errors"
 	"io"
-	"io/ioutil"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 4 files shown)


pkg/cortex/runtime_config.go
```diff
@@ -20,11 +21,13 @@ type runtimeConfigValues struct {
...
+	content, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
 
-	decoder := yaml.NewDecoder(r)
-	decoder.SetStrict(true)
-	if err := decoder.Decode(&overrides); err != nil {
+	var overrides = &runtimeConfigValues{}
+	if err := yaml.UnmarshalStrict(content, overrides); err != nil {
```
pstibrany, line 30: I think we should revert back to original code, and extend the checks as follows:
- If first returned error is EOF, we return `nil`. This covers empty document case.
- We call `decoder.Decode` second time, and **if it doesn't return EOF**, then `loadRuntimeConfig` will return error.

Second check covers the case when there are multiple YAML documents on the input, eg.

```
---
# This is an empty YAML.
---
multi_kv_config:
  mirror_enabled: true
```

In this case, first two calls to `Decode` return `nil` (no error), so our current version of `loadRuntimeConfig` (or version suggested in this PR, which only decodes first document) would return no error, which is confusing.

WDYT?

pstibrany (commented):
I think we should revert back to using `

pkg/cortex/runtime_config.go
pracucci: Very good point about multiple documents (that should be forbidden). I've changed it. What do you think?

pkg/cortex/runtime_config.go
```diff
@@ -24,10 +29,18 @@ func loadRuntimeConfig(r io.Reader) (interface{}, error) {
 
 	decoder := yaml.NewDecoder(r)
 	decoder.SetStrict(true)
-	if err := decoder.Decode(&overrides); err != nil {
+
+	// Decode the first document. An empty document (EOF) is OK.
+	err := decoder.Decode(&overrides)
+	if err != nil && !errors.Is(err, io.EOF) {
```
pstibrany, line 35:
```suggestion
	if err := decoder.Decode(&overrides); err != nil && !errors.Is(err, io.EOF) {
```

(for "consistency" with next `if`)

pkg/cortex/runtime_config.go
```diff
@@ -24,10 +29,18 @@ func loadRuntimeConfig(r io.Reader) (interface{}, error) {
...
 		return nil, err
 	}
 
+	// Ensure the provided YAML config is not composed by multiple documents,
```
pstibrany, line 39:
```suggestion
	// Ensure the provided YAML config is not composed of multiple documents,
```

pstibrany (approved):
LGTM, nice. Remaining comments are just non-blocking nits.

Issue #3031, opened 2020-08-13
Bad entry in overrides should be more noticeable


bboreham:
Symptom:

```
level=error ts=2020-08-13T10:03:45.497348374Z caller=manager.go:85 msg="failed to load config" err="yaml: unmarshal errors:\n  line 58: key \"redacted\" already set in map"
```

I need sirens going off and lights flashing, because now it has dropped all of my overrides and data is being unfairly rejected.

Related: #2957, #1433

bboreham:
In the community meeting @gouthamve pointed me at `cortex_runtime_config_last_reload_successful`

bboreham:
Also it was suggested that a failure on first load should be a fatal error, which would halt a rolling update.

pracucci:
_Still valid_

pracucci:
_Still valid_

pracucci:
Done in https://github.com/cortexproject/cortex/pull/3707.

PR #3707, opened 2021-01-18
Fail to startup Cortex if provided runtime config is invalid

pracucci:
**What this PR does**:
Currently, Cortex can successfully startup if the provided runtime config (typically limits overrides) is invalid. We recently had an incident because of this: we rolled out ingesters with an invalid runtime config, ingesters start up correctly, but overrides were missing causing an impact on customers. Other alerts fired pretty soon, so the impact of the incident was limited.

As a retrospective, I believe we should fast fail Cortex at startup if the provided runtime config is invalid. For example, in the case of a rolling update, ingesters would stop to rollout until fixed.

I haven't changed the behaviour of an invalid config loaded while Cortex is running because, in that case, we keep the previous config and we definitely don't want to shutdown all ingesters because of this.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
 * [ENHANCEMENT] Disabled in-memory shuffle-sharding subring cache in the store-gateway, ruler and compactor. This should reduce the memory utilisation in these services when shuffle-sharding is enabled, without introducing a significantly increase CPU utilisation. #3601
 * [ENHANCEMENT] Shuffle sharding: optimised subring generation used by shuffle sharding. #3601
 * [ENHANCEMENT] New /runtime_config endpoint that returns the defined runtime configuration in YAML format. The returned configuration includes overrides. #3639
+* [ENHANCEMENT] Fail to startup Cortex if provided runtime config is invalid. #3707
 * [BUGFIX] Allow `-querier.max-query-lookback` use `y|w|d` suffix like deprecated `-store.max-look-back-period`. #3598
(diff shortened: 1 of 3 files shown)


gouthamve (approved):

pstibrany (approved):
I think changing it to fail makes sense. This behaviour was originally introduced by @bboreham in https://github.com/cortexproject/cortex/commit/55697b593c66436eca01f22dac6ffdf17e1208ba.

jtlisi (approved):
LGTM

bboreham:
> This behaviour was originally introduced by @bboreham

Disagree.

bboreham:
The removal of an error check on startup happens here: https://github.com/cortexproject/cortex/commit/870fe73107bd2b7fe35976e3c310ae1a21ad76be#diff-f7e299904986266ff7106a24f1f542c4cab3fc980a401e416ab714627ba6585aL56-L58

bboreham:
Fixes #3031

pstibrany:
> The removal of an error check on startup happens here: [870fe73#diff-f7e299904986266ff7106a24f1f542c4cab3fc980a401e416ab714627ba6585aL56-L58](https://github.com/cortexproject/cortex/commit/870fe73107bd2b7fe35976e3c310ae1a21ad76be#diff-f7e299904986266ff7106a24f1f542c4cab3fc980a401e416ab714627ba6585aL56-L58)

👍 You're right. In that case I don't think there is any disagreement that we should merge this.

PR #3713, opened 2021-01-19
reduce file permissions as per gosec recommendation

nicolai86:
**What this PR does**:
`gosec` reports that directories and files created by cortex have too open permissions associated with it.
This CR follows the `gosec` recommendation, limiting all file and directory permissions flagged to just the user the cortex process runs on, following the Principle of Least Privilege.

In most cases this shouldn't create an issue; I'd expect production deployments to execute the cortex binary as a dedicated user already. Similarly for development purposes it's easy to assume the correct user to allow access to files for ie debugging purposes.

Unless this change raises a red flag somewhere I think we should consider adding the `gosec` testing `301`, `302` and `306` specifically to the GH workflow to ensure we don't regress.

**Checklist**
~- [ ] Tests updated~
~- [ ] Documentation added~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Reduce permissions on all files and directories created by cortex to just the user it executes as. #3713
 * [CHANGE] Ruler: removed the flag `-ruler.evaluation-delay-duration-deprecated` which was deprecated in 1.4.0. Please use the `ruler_evaluation_delay_duration` per-tenant limit instead. #3693
 * [CHANGE] Removed the flags `-<prefix>.grpc-use-gzip-compression` which were deprecated in 1.3.0: #3693
   * `-query-scheduler.grpc-client-config.grpc-use-gzip-compression`: use `-query-scheduler.grpc-client-config.grpc-compression` instead
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -262,7 +262,7 @@ type MultitenantAlertmanager struct {
 
 // NewMultitenantAlertmanager creates a new MultitenantAlertmanager.
 func NewMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, logger log.Logger, registerer prometheus.Registerer) (*MultitenantAlertmanager, error) {
-	err := os.MkdirAll(cfg.DataDir, 0777)
+	err := os.MkdirAll(cfg.DataDir, 0700)
 	if err != nil {
 		return nil, fmt.Errorf("unable to create Alertmanager data directory %q: %s", cfg.DataDir, err)
 	}
@@ -817,7 +817,7 @@ func createTemplateFile(dataDir, userID, fn, content string) (bool, error) {
 	}
 
 	dir := filepath.Join(dataDir, "templates", userID, filepath.Dir(fn))
-	err := os.MkdirAll(dir, 0755)
+	err := os.MkdirAll(dir, 0700)
 	if err != nil {
 		return false, fmt.Errorf("unable to create Alertmanager templates directory %q: %s", dir, err)
 	}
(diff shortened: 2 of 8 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -817,7 +817,7 @@ func createTemplateFile(dataDir, userID, fn, content string) (bool, error) {
 	}
 
 	dir := filepath.Join(dataDir, "templates", userID, filepath.Dir(fn))
-	err := os.MkdirAll(dir, 0755)
+	err := os.MkdirAll(dir, 0750)
```
ranton256, line 820: Is this creation mask intentionally different from the others? it seems to be user and group whereas others are just user?

nicolai86: Not intentionally, I'll reduce this to 0700 as well.

roidelapluie:
That sounds like a breaking change to me, I am not sure we should remove group read.

pstibrany:
Another way to solve this by admin is to set per-process umask, and we can use wide permissions in Cortex. I'm fine with PR or keeping it as is, since admin already has control over it anyway.

roidelapluie:
For what it is worth Prometheus and TSDB relies on umask and set the permissions to 0666. I guess that that part of cortex will too, then. And this is on purpose.

pstibrany:
> For what it is worth Prometheus and TSDB relies on umask and set the permissions to 0666. I guess that that part of cortex will too, then. And in Prometheus, this is on purpose.

Thanks Julien. I personally think Cortex should follow the Prometheus here for consistency.

nicolai86:
@pstibrany let me update the PR to keep TSDB related code on `0666`.
@roidelapluie I don't have context why prometheus would chose `0666` as default. can you elaborate or point me to a reasoning for this?

nicolai86:
also, @pstibrany would you be open to make the file permissions configurable? This way we can use more restrictive permissions without affecting other users of cortex.

pstibrany:
> also, @pstibrany would you be open to make the file permissions configurable? This way we can use more restrictive permissions without affecting other users of cortex.

I won’t block the PR that simply adjusts permissions around the Cortex code, but it seems to me that this can be easily solved by setting correct umask before starting Cortex process. If that’s difficult, perhaps Cortex itself could set umask (`golang.org/x/sys/unix.Umask`) globally at the start. That seems to me like a better solution, since we cannot easily change all vendored code that creates new filesystem objects. WDYT?

roidelapluie:
https://github.com/prometheus/prometheus/pull/4462 but there are others as well.

As said this is probably a breaking change.

roidelapluie:
I'd also note that setting umask is easy, and therefore file permissions are 'de facto' already configurable.

nicolai86:
the umask can be changed by the process while it's running. so while the permissions are configurable, it doesn't serve as a security mechanism. The goal of this PR was to reduce permissions to the required minimum to prevent too permissive permissions from being abused.

nicolai86:
@pstibrany I'll circle back on cortex setting a umask in a bit. Just need to get a heartbeat from our security engineer on it.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Reduce permissions on all files and directories created by cortex to just the user it executes as. #3713
```
pracucci, line 5: We've just released `1.7.0-rc.0`. Could you rebase `master` and move the CHANGELOG entry under the `master / unreleased` section, please?

nicolai86:
i'll close the PR in favor of using umask, as suggested by @pstibrany . thanks for the feedback folks :)

pracucci:
@nicolai86 Thanks for being so open discussing it. Would you be willing to submit a PR to add a new page to our documentation (I would say in the "guides" folder) to explain how to reduce file permissions with umask? I think may be a solid addition to our documentation for the benefit of the whole community.

nicolai86:
@pracucci I can contribute the documentation for cortex running inside k8s for sure.

pracucci:
> @pracucci I can contribute the documentation for cortex running inside k8s for sure.

Thanks!

PR #3717, opened 2021-01-20
Improve bucket index loader to handle edge cases

pracucci:
**What this PR does**:
While rolling out the bucket index in our of our clusters I've noticed that we have some tenants with rules but not remote writing the cluster (or started very soon to remote write so their blocks haven't been uploaded to bucket yet). Since there are no blocks in the bucket, their bucket index doesn't exist and this leads the `bucketindex.Loader` to continuously try to load their bucket index.

In this PR I'm proposing 3 changes that should relief the pressure when such edge case happens:
- Keep cached the "bucket index not found" error during background checks
- Don't increase `cortex_bucket_index_load_failures_total` on `ErrIndexNotFound` error (we already didn't do in the background check)
- In case of `ErrIndexNotFound`, update it every `UpdateOnStaleInterval` instead of `UpdateOnErrorInterval`

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
   * `cortex_alertmanager_sync_configs_failed_total`
   * `cortex_alertmanager_tenants_discovered`
   * `cortex_alertmanager_tenants_owned`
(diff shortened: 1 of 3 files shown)


pstibrany (approved):
LGTM, good job.

PR #3719, opened 2021-01-20
Move ExtendWrites field to distributor.

pstibrany:
**What this PR does**: After recent change to ring code (#3675), `ExtendWrites` field introduced in #3305 was no longer used inside ring code, but only in distributor. This PR removes the field from ring config structure and moves it to distributor config structure.

This breaks YAML config, but doesn't affect command line flag. Note that "extend writes" is still [experimental](https://cortexmetrics.io/docs/configuration/v1guarantees/).

**Checklist**
- [x] Tests updated
- [x] Documentation updated


pracucci (approved):
LGTM

PR #3720, opened 2021-01-20
Better layout for companies logos in the website

pracucci:
**What this PR does**:
The companies logos layout in the website is not much pretty. I think we should give each logo the same visibility. In this PR I'm proposing to use a fixed size box to include the company logo. I've cropped few logos, trimming transparent pixels, to let them scale nicely within the box.

**Before**:

Screenshot 2021-01-20 at 13 15 48

**After**:

Screenshot 2021-01-20 at 13 15 56

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


website/assets/scss/_styles_project.scss
```diff
@@ -50,3 +50,34 @@
...
+    text-align: center;
+}
+
+.adopter-logo {
+    display: inline-block;
+    margin: 10px;
+    padding: 10px;
+    width: 180px;
+    height: 100px;
+    border: 1px solid #dedede;
```
pstibrany, line 69: On your screenshot the border looks like leftover from debugging. Do you plan to keep it?

pstibrany (approved):

gouthamve (approved):
<3

website/assets/scss/_styles_project.scss
pracucci: Yes, I wanted to introduce visual boxes. WDYT?

pstibrany: As it is now, they are too subtle, and look like forgotten leftover. If you want to keep them, I would suggest using higher contrast. (Personally I would leave them out, I don't think boxes around these logos add anything).

pracucci: The box visually helps to give each logo the same weight. I've added a light drop shadow to better emphasis it.

Screenshot 2021-01-21 at 12 53 46

pstibrany: That's subtle, but looks better to me. Now it doesn't look like a bug anymore :)

PR #3722, opened 2021-01-21
Alertmanager: Remove alertmanager instead of pausing

gotjosh:
**What this PR does**:

Completely removes a tenant's Alertmanagers when synchronising configurations.

Before, we would "pause" the Alertmanager which would never free up the resources, and leave the Alertmanager running in-memory. 

I've also bundled a small change to a logger in the ruler that was double-logging the userID. 

NA/NA

**Checklist**
- [x] Tests updated
- N/A Documentation added

### DIFF
--- a/pkg/alertmanager/alertmanager_metrics.go
+++ b/pkg/alertmanager/alertmanager_metrics.go
@@ -155,7 +155,9 @@ func (m *alertmanagerMetrics) addUserRegistry(user string, reg *prometheus.Regis
 }
 
 func (m *alertmanagerMetrics) removeUserRegistry(user string) {
-	m.regs.RemoveUserRegistry(user, true)
+	// We neeed to go for a soft deletion here, as hard deletion requires
+	// that _all_ metrics except gauges are per-user.
+	m.regs.RemoveUserRegistry(user, false)
 }
 
 func (m *alertmanagerMetrics) Describe(out chan<- *prometheus.Desc) {
--- a/pkg/alertmanager/alertmanager_metrics.go
+++ b/pkg/alertmanager/alertmanager_metrics.go
@@ -154,6 +154,10 @@ func (m *alertmanagerMetrics) addUserRegistry(user string, reg *prometheus.Regis
 	m.regs.AddUserRegistry(user, reg)
 }
 
+func (m *alertmanagerMetrics) removeUserRegistry(user string) {
+	m.regs.RemoveUserRegistry(user, true)
+}
+
 func (m *alertmanagerMetrics) Describe(out chan<- *prometheus.Desc) {
 	out <- m.alertsReceived
 	out <- m.alertsInvalid
--- a/pkg/alertmanager/alertmanager_metrics_test.go
+++ b/pkg/alertmanager/alertmanager_metrics_test.go
@@ -257,6 +257,226 @@ func TestAlertmanagerMetricsStore(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestAlertmanagerMetricsRemoval(t *testing.T) {
+	mainReg := prometheus.NewPedanticRegistry()
+
+	alertmanagerMetrics := newAlertmanagerMetrics()
+	mainReg.MustRegister(alertmanagerMetrics)
+	alertmanagerMetrics.addUserRegistry("user1", populateAlertmanager(1))
+	alertmanagerMetrics.addUserRegistry("user2", populateAlertmanager(10))
+	alertmanagerMetrics.addUserRegistry("user3", populateAlertmanager(100))
+
+	alertmanagerMetrics.removeUserRegistry("user3")
+
+	//TODO: Verify that these metrics match what I expect.
+	err := testutil.GatherAndCompare(mainReg, bytes.NewBufferString(`
+    		# HELP cortex_alertmanager_alerts How many alerts by state.
+    		# TYPE cortex_alertmanager_alerts gauge
+    		cortex_alertmanager_alerts{state="active",user="user1"} 1
+    		cortex_alertmanager_alerts{state="active",user="user2"} 10
+    		cortex_alertmanager_alerts{state="suppressed",user="user1"} 2
+    		cortex_alertmanager_alerts{state="suppressed",user="user2"} 20
+
+    		# HELP cortex_alertmanager_alerts_invalid_total The total number of received alerts that were invalid.
+    		# TYPE cortex_alertmanager_alerts_invalid_total counter
+    		cortex_alertmanager_alerts_invalid_total{user="user1"} 2
+    		cortex_alertmanager_alerts_invalid_total{user="user2"} 20
+
+    		# HELP cortex_alertmanager_alerts_received_total The total number of received alerts.
+    		# TYPE cortex_alertmanager_alerts_received_total counter
+    		cortex_alertmanager_alerts_received_total{user="user1"} 10
+    		cortex_alertmanager_alerts_received_total{user="user2"} 100
+
+    		# HELP cortex_alertmanager_config_hash Hash of the currently loaded alertmanager configuration.
+    		# TYPE cortex_alertmanager_config_hash gauge
+    		cortex_alertmanager_config_hash{user="user1"} 0
+    		cortex_alertmanager_config_hash{user="user2"} 0
+
+    		# HELP cortex_alertmanager_nflog_gc_duration_seconds Duration of the last notification log garbage collection cycle.
+    		# TYPE cortex_alertmanager_nflog_gc_duration_seconds summary
+    		cortex_alertmanager_nflog_gc_duration_seconds_sum 11
+    		cortex_alertmanager_nflog_gc_duration_seconds_count 2
+
+    		# HELP cortex_alertmanager_nflog_gossip_messages_propagated_total Number of received gossip messages that have been further gossiped.
+    		# TYPE cortex_alertmanager_nflog_gossip_messages_propagated_total counter
+    		cortex_alertmanager_nflog_gossip_messages_propagated_total 11
+
+    		# HELP cortex_alertmanager_nflog_queries_total Number of notification log queries were received.
+    		# TYPE cortex_alertmanager_nflog_queries_total counter
+    		cortex_alertmanager_nflog_queries_total 11
+
+    		# HELP cortex_alertmanager_nflog_query_duration_seconds Duration of notification log query evaluation.
+    		# TYPE cortex_alertmanager_nflog_query_duration_seconds histogram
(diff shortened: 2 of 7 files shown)


pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -154,6 +154,10 @@ func (m *alertmanagerMetrics) addUserRegistry(user string, reg *prometheus.Regis
 	m.regs.AddUserRegistry(user, reg)
 }
 
+func (m *alertmanagerMetrics) removeUserRegistry(user string) {
+	m.regs.RemoveUserRegistry(user, true)
```
pstibrany, line 158: "hard" remove is only possible under certain conditions:
1) all metrics are grouped "per user"
2) metrics that are not "per user" are only gauges

Unfortunately in this case there are multiple counters (eg. `silencesQueriesTotal`), and hard-remove will break them.

pkg/alertmanager/alertmanager_metrics_test.go
```diff
@@ -257,6 +257,226 @@ func TestAlertmanagerMetricsStore(t *testing.T) {
...
+func TestAlertmanagerMetricsRemoval(t *testing.T) {
+	mainReg := prometheus.NewPedanticRegistry()
+
+	alertmanagerMetrics := newAlertmanagerMetrics()
+	mainReg.MustRegister(alertmanagerMetrics)
+	alertmanagerMetrics.addUserRegistry("user1", populateAlertmanager(1))
+	alertmanagerMetrics.addUserRegistry("user2", populateAlertmanager(10))
+	alertmanagerMetrics.addUserRegistry("user3", populateAlertmanager(100))
+
+	alertmanagerMetrics.removeUserRegistry("user3")
```
pstibrany, line 269: I'd suggest adding a test before removal and after removal. Make sure that counters don't go down -- that would be interpreted as a reset by Prometheus.

pkg/ruler/manager.go
```diff
@@ -171,8 +171,7 @@ func (r *DefaultMultiTenantManager) newManager(ctx context.Context, userID strin
 	reg := prometheus.NewRegistry()
 	r.userManagerMetrics.AddUserRegistry(userID, reg)
 
-	logger := log.With(r.logger, "user", userID)
```
pstibrany, line 174: Why is per-user logger removed?

pkg/alertmanager/alertmanager_metrics.go
gotjosh: Thanks for breaking that down - I went through a rabbit hole trying to understand when soft vs hard was possible by analysing `TestMetricsArePerUser` in the ruler but couldn't really make sense of it. I then noticed that most of the alertmanager metrics (even some I think should be) are not per user and that this was causing a problem with hard deletion.

In the world where we're going with a RF=3 as a default, I'm thinking that most of these metrics most be per user to make them useful.

pkg/ruler/manager.go
gotjosh: It's already set as part of the managerFactory.

https://github.com/cortexproject/cortex/blob/5682f6678cd96ce5f3514b4a3484b7d44b22e45a/pkg/ruler/compat.go#L136

pkg/alertmanager/alertmanager_metrics.go
pstibrany: If you go with hard delete, then I'd suggest adding a test that verifies the two conditions I wrote above. See `TestMetricsArePerUser` test in ruler that verifies (1).

gotjosh:
One more Look @pstibrany? 🙏

pkg/alertmanager/alertmanager_metrics_test.go
```diff
@@ -257,6 +258,474 @@ func TestAlertmanagerMetricsStore(t *testing.T) {
...
+	mainReg := prometheus.NewPedanticRegistry()
+
+	alertmanagerMetrics := newAlertmanagerMetrics()
+	mainReg.MustRegister(alertmanagerMetrics)
+	alertmanagerMetrics.addUserRegistry("user1", populateAlertmanager(1))
+	alertmanagerMetrics.addUserRegistry("user2", populateAlertmanager(10))
+	alertmanagerMetrics.addUserRegistry("user3", populateAlertmanager(100))
+
+	// Assertion before removal.
+	err := testutil.GatherAndCompare(mainReg, bytes.NewBufferString(`
```
pstibrany, line 271: Nit: It's difficult to see what's being tested here. Couple of metrics (some gauge, some per-user, some counter) would be enough, instead of all :)

pstibrany (approved):
LGTM, thanks!

pkg/alertmanager/alertmanager_metrics_test.go
gotjosh: Let me add a more in-depth comment, I tend to prefer these kinds of holistic tests as they're susceptible to small changes. With that, it makes it clear to see when something does change which tends to be helpful with regressions.

pkg/alertmanager/multitenant.go
```diff
@@ -604,17 +604,16 @@ func (am *MultitenantAlertmanager) syncConfigs(cfgs map[string]alerts.AlertConfi
...
-			level.Info(am.logger).Log("msg", "deactivating per-tenant alertmanager", "user", user)
-			userAM.Pause()
-			delete(am.cfgs, user)
-			am.multitenantMetrics.lastReloadSuccessful.DeleteLabelValues(user)
-			am.multitenantMetrics.lastReloadSuccessfulTimestamp.DeleteLabelValues(user)
-			level.Info(am.logger).Log("msg", "deactivated per-tenant alertmanager", "user", user)
+	for userID, userAM := range am.alertmanagers {
+		if _, exists := cfgs[userID]; !exists {
+			level.Info(am.logger).Log("msg", "deactivating per-tenant alertmanager", "user", userID)
+			go userAM.Stop()
```
pracucci, line 610: I would try to avoid calling `Stop()` is a goroutine. Calling it synchronously would be safer in my opinion.

The `userAM.Stop()` does basically two things:
1. Cancel contexts
2. `am.wg.Wait()`

The real blocking thing here could be the wait, so I would suggest the following:
1. Run `userAM.Stop()` synchronously (no gorouting here)
2. Remove the `am.wg.Wait()` from the `Stop()`
3. Create `StopAndWait()` which calls `Stop()` and then `am.wg.Wait()`, replacing existing calls to `Stop()` with `StopAndWait()` (except for this one)

WDYT @gotjosh @pstibrany ?

pkg/alertmanager/multitenant.go
```diff
@@ -244,11 +244,11 @@ type MultitenantAlertmanager struct {
...
 	fallbackConfig string
 
-	// All the organization configurations that we have. Only used for instrumentation.
-	cfgs map[string]alerts.AlertConfigDesc
-
 	alertmanagersMtx sync.Mutex
 	alertmanagers    map[string]*Alertmanager
+	// Stores the current set of configurations we're running in each tenant's Alertmanager.
+	// Used for comparing configurations as we synchronize them.
+	cfgs map[string]alerts.AlertConfigDesc
```
pracucci, line 251: I would expect `cfgs` to be protected by `alertmanagerMtx` (because there may be concurrent accesses to it) but its usage in `setConfig()` is not protected. We should fix it.

pracucci (commented):
Good job @gotjosh! I left a couple of comments.

pkg/alertmanager/multitenant.go
gotjosh: Good shout. 👌

pkg/alertmanager/multitenant.go
gotjosh: I think this makes sense, nice catch. 

I have one question though (mostly for the benefit of my understanding), in practice the split of context cancelling and `wg.Wait()` means that we instruct the channels to close but don't wait for them to drained? If so, what happens if they never drain? I guess it would no better from before where we were just fire and forgetting about the stopping.

pkg/alertmanager/alertmanager_metrics_test.go
gotjosh: Done!

pkg/alertmanager/multitenant.go
gotjosh: I went for a lock at the top of the function as there are several accesses to its existence and the config setting throughout the function, as opposed to lock/unlock multiple times throughout the method. This feels simpler than trying to optimise in multiple accesses for now but let me know what you think!

pracucci: I don't think we should lock the function. There are some blocking calls down there, while we should keep the lock the least as possible (considering it's an exclusive global lock). There are only two usages of `am.cfgs` in `setConfig()` so doesn't look prohibitive took a lock at the right place.

In particular, to my understanding of the function logic, `am.cfgs[cfg.User]` (used in the `if`) can be read along with the following:

```
am.alertmanagersMtx.Lock()
existing, hasExisting := am.alertmanagers[cfg.User]
am.alertmanagersMtx.Unlock()
```

What do you think?

gotjosh: Generally, I'd yes, but if we look at the problem holistically - I think it makes sense to just lock the whole function.

First, the only way we can call `setConfig` concurrently is if we have reached the poll interval or there was a change of topology in the ring. On the latter case, there will certainly be configs that are going to be removed/added from a particular alertmanager.

Second, as you're already aware `setConfig` operates on the assumption that the alertmanager exists or not for that particular tenant. So the moment we do:
```
am.alertmanagersMtx.Lock()
existing, hasExisting := am.alertmanagers[cfg.User]
am.alertmanagersMtx.Unlock()
```

We're running the function under the assumption that the tenant exists for this particular AM but that's not necessarily true. If we release the lock another run of `syncConfigs` could have removed it by the time we do the work to create/update the alertmanager. 

I understand the chances of this race are fairly slim, but I genuinely think this is the safest route. Is there anything I'm missing?

pracucci: > First, the only way we can call setConfig concurrently is if we have reached the poll interval or there was a change of topology in the ring.

`setConfig()` can also be called by `alertmanagerFromFallbackConfig()`, which can be called while serving an HTTP request from the user.

> We're running the function under the assumption that the tenant exists for this particular AM but that's not necessarily true. If we release the lock another run of syncConfigs could have removed it by the time we do the work to create/update the alertmanager.

This is true. `loadAndSyncConfigs()` cannot be called concurrently, but `setConfig()` can be both called concurrently by `loadAndSyncConfigs()` (its calls chain) and an HTTP request setting up the alertmanager with the default config.

LGTM 👍

pracucci (approved):
Thanks!

pstibrany (approved):
👍

PR #3724, opened 2021-01-21
Changed default value of -blocks-storage.bucket-store.sync-interval from 5m to 15m

pracucci:
**What this PR does**:
While rolling out the bucket index in our cluster we've realised the default value of `-blocks-storage.bucket-store.sync-interval` (5m) doesn't play nicely with the consistency check. The consistency check gives a grace period for blocks recently uploaded (and potentially not yet discovered by store-gateway) which is equal to 3x the sync-interval (so 15m by default). Considering a default caching TTL for the bucket index of 5m, there may be edge cases where the querier is including a recently uploaded block in the consistency check but the store-gateway hasn't loaded it yet.

To fix the default config, in this PR I'm proposing to increase the default value of `sync-interval` from 5m to 15m (which also look a better default value anyway).

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany (approved):
👍

PR #3727, opened 2021-01-21
Don't update lastUpdateTime if TSDB isn't updated.

pstibrany:
**What this PR does**: This PR modified ingester to not update "lastUpdate" time if push request didn't contain any sample that was appended to TSDB.

Without doing this change, continuous push of bad samples will keep TSDB in memory with unflushed data. If "good" sample arrives much later, all old data will be finally put to block.

With this change, ingester will eventually flush and close such TSDB, and remove local data. After closing, next push of valid or invalid sample will reopen TSDB again (and then later close, if no valid samples arrive), but at least old samples will not be kept in the memory for a long time.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/ingester_v2.go
+++ b/pkg/ingester/ingester_v2.go
@@ -767,7 +767,10 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 	}
 	i.TSDBState.appenderCommitDuration.Observe(time.Since(startCommit).Seconds())
 
-	db.setLastUpdate(time.Now())
+	// If only invalid samples are pushed, don't change "last update", as TSDB was not modified.
+	if succeededSamplesCount > 0 {
+		db.setLastUpdate(time.Now())
+	}
 
 	// Increment metrics only if the samples have been successfully committed.
 	// If the code didn't reach this point, it means that we returned an error
--- a/pkg/ingester/ingester_v2_test.go
+++ b/pkg/ingester/ingester_v2_test.go
@@ -1894,6 +1894,51 @@ func (m *shipperMock) Sync(ctx context.Context) (uploaded int, err error) {
 	return args.Int(0), args.Error(1)
 }
 
+func TestIngester_invalidSamplesDontChangeLastUpdateTime(t *testing.T) {
+	cfg := defaultIngesterTestConfig()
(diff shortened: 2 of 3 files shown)


ranton256 (approved):
Thanks for the fix. The changes LGTM.

pkg/ingester/ingester_v2_test.go
```diff
@@ -1894,6 +1935,51 @@ func (m *shipperMock) Sync(ctx context.Context) (uploaded int, err error) {
...
+		req, _, _ := mockWriteRequest(labels.Labels{{Name: labels.MetricName, Value: "test"}}, 0, sampleTimestamp)
+		_, err = i.v2Push(ctx, req)
+		require.NoError(t, err)
+	}
+
+	db := i.getTSDB(userID)
+	lastUpdate := db.lastUpdate.Load()
+
+	// Wait until 1 second passes.
+	test.Poll(t, 1*time.Second, time.Now().Unix()+1, func() interface{} {
```
pracucci, line 1968: Why not `time.Sleep(time.Second)`?

pracucci (approved):

pracucci (commented):
Could you also add a CHANGELOG entry, please?

pkg/ingester/ingester_v2_test.go
pstibrany: `test.Poll` allows us to wait shorter.

pkg/ingester/ingester_v2.go
```diff
@@ -767,7 +767,10 @@ func (i *Ingester) v2Push(ctx context.Context, req *client.WriteRequest) (*clien
 	}
 	i.TSDBState.appenderCommitDuration.Observe(time.Since(startCommit).Seconds())
 
-	db.setLastUpdate(time.Now())
+	// If only invalid samples are pushed, don't change "last update", as TSDB was not modified.
+	if succeededSamplesCount > 0 {
+		db.setLastUpdate(time.Now())
```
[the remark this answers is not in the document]
pstibrany: I don't think we need to avoid `time.Now()` here. Furthermore there are cases where commit takes quite a while. We only have bucket up to 10 seconds, but metrics show that we hit it regularly on some ingesters (!), although in practice that should still be good enough.

PR #3734, opened 2021-01-22
Move some utility functions out of `util` and into their own packages

kminehart:
**What this PR does**:

This PR reduces the number of imports in the `pkg/chunk/cache` for use outside of cortex.

When importing `github.com/cortexproject/cortex/pkg/chunk/cache` I was getting a lot of extra dependencies because `cache` imports `util` which imports packages like `grpc-go`.

This involved moving functions out of `util` and into their own packages.

To maintain compatibility and reduce the number of changes, I've left the functions in `util`, and had them call their equivalent function in their new package. If we don't want the cruft of having these defined twice, I'd be happy to comb through and refactor it all.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


jtlisi (commented):
> To maintain compatibility and reduce the number of changes, I've left the functions in util, and had them call their equivalent function in their new package. If we don't want the cruft of having these defined twice, I'd be happy to comb through and refactor it all.

I was about to comment about this, but then I re-read you description. I do think it would make sense to remove the original files and to update everywhere that calls these functions. It would also make the diff a bit cleaner since a lot of the files should be a 1-1 copy over besides the package name references.

pkg/util/ctxutil/ctx.go
```diff
@@ -0,0 +1 @@
+package ctxutil
```
pstibrany, line 1: Why do we need this empty package?

pkg/util/logutil/warn.go
```diff
@@ -0,0 +1,21 @@
+package logutil
```
pstibrany, line 1: More like `experimental.go`

pkg/util/logutil/log.go
```diff
@@ -0,0 +1,7 @@
+package logutil
```
pstibrany, line 1: I'd suggest not repeating "util" suffix in these package names, and only use non-conflicting names on imports.

pkg/util/experimental.go
```diff
@@ -1,21 +1,8 @@
...
-	prometheus.CounterOpts{
-		Namespace: "cortex",
-		Name:      "experimental_features_in_use_total",
-		Help:      "The number of experimental features in use.",
-	},
-)
+import "github.com/cortexproject/cortex/pkg/util/logutil"
 
 // WarnExperimentalUse logs a warning and increments the experimental features metric.
 func WarnExperimentalUse(feature string) {
```
pstibrany, line 6: Alternative:

```
var WarnExperimentalUse = logutil.WarnExperimentalUse
```

I don't think we even need this. Let's just update the references.

pkg/util/log.go
```diff
@@ -9,16 +9,15 @@ import (
...
 
-	"github.com/cortexproject/cortex/pkg/tenant"
+	"github.com/cortexproject/cortex/pkg/util/logutil"
 )
 
 var (
 	// Logger is a shared go-kit logger.
 	// TODO: Change all components to take a non-global logger via their constructors.
-	Logger = log.NewNopLogger()
+	Logger = logutil.Logger
```
pstibrany, line 20: If we want to keep this var to avoid lot of code changes, let's at least mark it as deprecated.

pkg/util/math.go
```diff
@@ -1,33 +1,25 @@
 package util
 
+import (
```
pstibrany, line 3: There are only handful references for each of these functions, I'd suggest replacing them with new functions and deleting this file.

pstibrany (commented):
> I was about to comment about this, but then I re-read you description. I do think it would make sense to remove the original files and to update everywhere that calls these functions. It would also make the diff a bit cleaner since a lot of the files should be a 1-1 copy over besides the package name references.

I agree with Jacob that we should just remove old functions and variables. Note however that we should then also update  projects that use this (eg. Loki, Thanos), to avoid the pain for them.

pkg/util/experimental.go
kminehart: hahahaha good point. I  completely forgot about this.

pkg/util/logutil/log.go
kminehart: What name would you suggest in this case?

pkg/util/ctxutil/ctx.go
kminehart: I put `WithContext` in here, but then I realized it needed to be grouped with logging functions. I must have forgotten to remove it.I'll remove this empty package.

pkg/util/logutil/log.go
pstibrany: `pkg/util/log`, and import using `import util_log "github.com/cortexproject/cortex/pkg/util/log"`.

kminehart:
* Renamed logutil/mathutil to log/math
* Removed old definitions in `util` and replaced references to these functions throughout the codebase
* When a conflict arose (pretty common with these 2 package names), I went with `util_log` or `util_math` to show that they were subpackages of the `util` package.

pstibrany (approved):

pkg/alertmanager/api.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/cortexproject/cortex/pkg/alertmanager/alerts"
 	"github.com/cortexproject/cortex/pkg/tenant"
-	"github.com/cortexproject/cortex/pkg/util"
+	util_log "github.com/cortexproject/cortex/pkg/util/log"
```
pracucci, line 12: I think the `pkg/util/log` should be renamed to `pkg/util/logutil` to avoid misunderstandings with `github.com/go-kit/kit/log`.

pkg/ingester/ingester_v2_test.go
```diff
@@ -41,6 +41,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/ring"
 	cortex_tsdb "github.com/cortexproject/cortex/pkg/storage/tsdb"
 	"github.com/cortexproject/cortex/pkg/util"
+	util_math "github.com/cortexproject/cortex/pkg/util/math"
```
pracucci, line 44: `pkg/util/math` should be renamed to `pkg/util/mathutil` for the same reason.

pkg/util/log/log.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"os"
+
+	"github.com/go-kit/kit/log"
+	kitlog "github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+)
+
+var (
+	Logger = kitlog.NewNopLogger()
```
pracucci, line 13: This should be marked deprecated too.

pracucci (commented):
Thanks for working on this! I agree with the changes, but let's rename the util packages for clarity (and to avoid pkg naming clashes):

- `pkg/util/log` > `pkg/util/logutil`
- `pkg/util/math` > `pkg/util/mathutil`

pkg/alertmanager/api.go
kminehart: that's what they were originally; @pstibrany requested that I rename them :see_no_evil:

pstibrany: > I think the `pkg/util/log` should be renamed to `pkg/util/logutil` to avoid misunderstandings with `github.com/go-kit/kit/log`.

:) In my previous comment I actually suggested we do the opposite, to avoid repeating “util” in the package name. I don’t mind the conflicting names, and actually prefer “util_log” and “util_math” in imports (I wouldn’t mind to use that everywhere for consistency, but there is no need to).

pracucci: Didn't know you already discussed it. I missed that, sorry. Let's discuss it offline, but definitely not a good reason to block this PR.

pracucci (approved):

PR #3744, opened 2021-01-26
Enable @ modifier with result cache handling

codesome:
**What this PR does**:

Enables the newly added @ modifier in PromQL with cache handling for it.

If the time in @ is > query end time, then we don't cache the result. If @ time is <= end, then we cache, since even if the start and end were actually in the future we would end up caching the result anyway.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/querier/queryrange/results_cache.go
```diff
@@ -257,6 +266,44 @@ func (s resultsCache) shouldCacheResponse(ctx context.Context, r Response) bool
...
+	}
+
+	end := r.GetEnd()
+	atModAfterEnd := false
+	parser.Inspect(expr, func(n parser.Node, _ []parser.Node) error {
+		switch e := n.(type) {
+		case *parser.VectorSelector:
+			if e.Timestamp != nil && *e.Timestamp > end {
+				atModAfterEnd = true
+				return errors.New("at modifier after end")
```
pstibrany, line 287: Nit: Since we're going to run this method for every single query, perhaps we can just create this sentinel error value once, and not on every invocation.

pstibrany (approved):
👍

pstibrany:
Should we keep `@` support protected by flag? Or is it enabled in Prometheus by default?

codesome:
> Should we keep @ support protected by flag? Or is it enabled in Prometheus by default?

It is disabled by default in Prometheus because it has no in-built caching and so that users of PromQL can enable when they have fixed the caching to use this modifier. Since we handle the caching inside Cortex itself, I don't think we need to have a flag for that and just enable it by default.

codesome:
After some offline discussions, I have added a flag for @ modifier.

gouthamve (approved):

pkg/querier/queryrange/results_cache.go
```diff
@@ -257,6 +263,54 @@ func (s resultsCache) shouldCacheResponse(ctx context.Context, r Response) bool
...
+	//      below maxCacheTime. In such cases if any tenant is intentionally
+	//      playing with old data, we could cache empty result if we look
+	//      beyond query end.
+	query := r.GetQuery()
+	if !strings.Contains(query, "@") {
+		return true
+	}
+	expr, err := parser.ParseExpr(query)
+	if err != nil {
+		// We are being pessimistic in such cases.
```
pracucci, line 283: A warning log may be useful here. WDYT?

pracucci (commented):
Good job! I left a nit post-merge comment.

pkg/querier/queryrange/results_cache.go
codesome: I will open another PR for the warning.

PR #3668, opened 2021-01-08
HA tracker limits

pstibrany:
**What this PR does**: This PR verifies that replica label isn't larger than configured max label value limit, and also introduces new limit for maximum number of clusters tracked by HA tracker.

**Checklist**
- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3745, opened 2021-01-26
Don't track as error an HA tracker CAS operation intentionally aborted

pracucci:
**What this PR does**:
The PR https://github.com/cortexproject/cortex/pull/3668 introduced a little regression causing intentionally aborted CAS operations (by HA tracker) to be tracked as `500` in the `cortex_kv_request_duration_seconds` metric instead of `200`. The reason of this subtle bug is in the introduction of `replicasNotMatchError` (which can be returned by the `CAS()` operation function), while the previous version was returning a `httpgrpc.Errorf(http.StatusAccepted)` which is correctly converted into a `200` error code by `pkg/ring/kv/metrics.go`.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/metrics_helper.go
+++ b/pkg/util/metrics_helper.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/prometheus/pkg/labels"
 )
 
 // Data for single value (counter/gauge) with labels.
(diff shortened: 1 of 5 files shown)


pkg/util/metrics_helper.go
```diff
@@ -661,3 +662,39 @@ func (r *UserRegistries) BuildMetricFamiliesPerUser() MetricFamiliesPerUser {
...
+	builder := labels.NewBuilder(nil)
+	for _, pair := range pairs {
+		builder.Set(pair.GetName(), pair.GetValue())
+	}
+	return builder.Labels()
+}
+
+// GetSumOfHistogramSampleCount returns the sum of samples count of histograms matching the provided metric name
+// and optional label matchers. Returns 0 if no metric matches.
+func GetSumOfHistogramSampleCount(families []*dto.MetricFamily, metricName string, matchers labels.Selector) uint64 {
```
pracucci, line 677: In the unit tests I've added in the distributor I couldn't use the typical `testutil.GatherAndCompare()` because the histogram tracks the request duration (which is unpredictable). Since we don't have a way to just test the histogram sample count, I've introduced this function to easy it (and future tests too).

pstibrany (approved):
Thanks for spotting and fixing this!

gouthamve:
This should be cherry-picked to `1.7.0` branch or not? @khaines @pracucci

khaines:
@gouthamve I'm of the opinion it should be brought into the 1.7.0 branch.

khaines:
#3746 - adds this change to the 1.7 Branch

PR #3749, opened 2021-01-27
Fixing ambiguous double negative for ingesters

warmfusion:
Fixing ambiguous terminology from double negatives;

"doesn't receive neither this and that". As neither is Not Either, this could suggest that it accepts both concurrently as 'either' binds the two operations together, but it really wants to be suggesting that these operations are mutually exclusive.

Not sure this is the clearest explanation I've ever created on a ticket, but i hope it makes sense.

Reading this section made my eye twitch which may just be my own problem so if you think the current version is valid that's fine by me. I'm not an expert in English grammar :-)

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -152,9 +152,9 @@ Incoming series are not immediately written to the storage but kept in memory an
 Ingesters contain a **lifecycler** which manages the lifecycle of an ingester and stores the **ingester state** in the [hash ring](#the-hash-ring). Each ingester could be in one of the following states:
 
 - **`PENDING`**<br />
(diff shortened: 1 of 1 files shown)


pracucci (approved):
LGTM, thanks!

pstibrany (approved):
Thanks!

pstibrany:
@warmfusion Can you please check and fix DCO (explained at https://github.com/cortexproject/cortex/pull/3749/checks?check_run_id=1777331604)?

pracucci:
> @warmfusion Can you please check and fix [DCO failure](https://github.com/cortexproject/cortex/pull/3749/checks?check_run_id=1777331604)? Thanks! [Here is some explanation about it](https://github.com/apps/dco).

@warmfusion Gentle ping. Could you sign your commit, so we can merge, please?

PR #3750, opened 2021-01-27
Tenant deletion: ruler support.

pstibrany:
**What this PR does**: This PR extends tenant deletion API to support deletion of rule groups. API handler deletes rule groups directly, without using "deletion mark" as proposed in the design document. 

In addition to that, it also changes ruler such that when ruler is no longer processing given rule group (due to resharding, or rule group deleted from bucket), it removes locally cached rule group.

PR also extends test suite for to verify rule store behaviour used by tenant deletion API handler (eg. deleting all rule groups for a user).

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@
   * `cortex_ha_tracker_replicas_cleanup_marked_for_deletion_total`
   * `cortex_ha_tracker_replicas_cleanup_deleted_total`
   * `cortex_ha_tracker_replicas_cleanup_delete_failed_total`
-* [ENHANCEMENT] Tenant deletion endpoints now support deletion of ruler groups. #3750
+* [ENHANCEMENT] Tenant deletion endpoints now support deletion of ruler groups. This must be explicitly enabled by using `-tenant-deletion.rule-groups-enabled` and only works when using rule store that supports deletion. #3750
 * [BUGFIX] Cortex: Fixed issue where fatal errors and various log messages where not logged. #3778
 * [BUGFIX] HA Tracker: don't track as error in the `cortex_kv_request_duration_seconds` metric a CAS operation intentionally aborted. #3745
 * [BUGFIX] Querier / ruler: do not log "error removing stale clients" if the ring is empty. #3761
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@
   * `cortex_ha_tracker_replicas_cleanup_marked_for_deletion_total`
   * `cortex_ha_tracker_replicas_cleanup_deleted_total`
   * `cortex_ha_tracker_replicas_cleanup_delete_failed_total`
-* [ENHANCEMENT] Tenant deletion endpoints now support deletion of ruler groups. This must be explicitly enabled by using `-tenant-deletion.rule-groups-enabled` and only works when using rule store that supports deletion. #3750
+* [ENHANCEMENT] Tenant deletion endpoints now support deletion of ruler groups. This only works when using rule store that supports deletion. #3750
 * [BUGFIX] Cortex: Fixed issue where fatal errors and various log messages where not logged. #3778
 * [BUGFIX] HA Tracker: don't track as error in the `cortex_kv_request_duration_seconds` metric a CAS operation intentionally aborted. #3745
 * [BUGFIX] Querier / ruler: do not log "error removing stale clients" if the ring is empty. #3761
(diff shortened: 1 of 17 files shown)


pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -13,28 +14,38 @@ import (
...
 
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
 	"github.com/cortexproject/cortex/pkg/storage/bucket"
 	cortex_tsdb "github.com/cortexproject/cortex/pkg/storage/tsdb"
 	"github.com/cortexproject/cortex/pkg/tenant"
 	"github.com/cortexproject/cortex/pkg/util"
 )
 
+type TenantDeletionConfig struct {
+	EnableDeletionOfRuleGroups bool `yaml:"rule_groups_enabled"`
```
pracucci, line 25: A couple of feedback on the config:
1. What's the use case to enable it conditionally? Do we really need this config?
1. I would make this config a child of the purger config and not a root config entry

pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -83,7 +113,21 @@ func (api *BlocksPurgerAPI) DeleteTenantStatus(w http.ResponseWriter, r *http.Re
...
-func (api *BlocksPurgerAPI) checkBlocksForUser(ctx context.Context, userID string) (bool, error) {
+func (api *BlocksPurgerAPI) isRulesForUserDeleted(ctx context.Context, userID string) (bool, error) {
+	if api.ruleStore == nil {
+		// If purger doesn't have access to rule store, then we cannot say that rules have been deleted.
+		return false, nil
+	}
+
+	list, err := api.ruleStore.ListRuleGroupsForUserAndNamespace(ctx, userID, "")
+	if err != nil {
+		return false, nil
```
pracucci, line 124: Shouldn't we return error in this case? Otherwise this function never returns an error.

pracucci (approved):
Great job 👏 👏 👏  I just have a comment about the config. We can talk offline if you prefer.

pkg/chunk/purger/blocks_purger_api.go
pstibrany: > 1. What's the use case to enable it conditionally? Do we really need this config?

Reason for introducing this option is that not all rule-stores allow deletion of rule groups, and instead return error for any attempt. To make tenant deletion API usable when using such rule stores, it is possible to disable deleting of rule groups.

> 2. I would make this config a child of the purger config and not a root config entry
Done.

pracucci: > Reason for introducing this option is that not all rule-stores allow deletion of rule groups, and instead return error for any attempt. To make tenant deletion API usable when using such rule stores, it is possible to disable deleting of rule groups.

Can we auto-detect it?

pkg/chunk/purger/blocks_purger_api.go
pstibrany: Right. Thanks for spotting this.

pkg/chunk/purger/blocks_purger_api.go
pstibrany: Yes, we can. I've changed the code to do that now, and removed the option. PTAL.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
   * `cortex_ingester_tsdb_symbol_table_size_bytes`
   * `cortex_ingester_tsdb_storage_blocks_bytes`
   * `cortex_ingester_tsdb_time_retentions_total`
+* [ENHANCEMENT] Tenant deletion endpoints now support deletion of ruler groups. This must be explicitly enabled by using `-tenant-deletion.rule-groups-enabled` and only works when using rule store that supports deletion. #3750
```
pracucci, line 46: The CHANGELOG entry needs to be updated. The CLI flag has been removed.

pkg/ruler/rules/store.go
```diff
@@ -36,8 +39,16 @@ type RuleStore interface {
...
 	SetRuleGroup(ctx context.Context, userID, namespace string, group *RuleGroupDesc) error
+
+	// DeleteRuleGroup deletes single rule group.
 	DeleteRuleGroup(ctx context.Context, userID, namespace string, group string) error
+
+	// DeleteNamespace lists rule groups for given user and namespace, and deletes all rule groups.
+	// If namespace is empty, deletes all rule groups for user.
 	DeleteNamespace(ctx context.Context, userID, namespace string) error
+
+	// SupportsModifications true if this RuleStore supports operations that modify rule groups in the store.
```
pracucci, line 50:
```suggestion
	// SupportsModifications returns true if this RuleStore supports operations that modify rule groups in the store.
```

pracucci (approved):
LGTM, thanks! (modulo a couple of nits)

PR #3755, opened 2021-01-28
Enable wrapping of distributor push.

jeschkies:
Summary:
Introduces `PushFunc` and `PushWrapper` types which are used to defined a
configurable wrapper around the push function of the distributor. It is
similar to an HTTP middleware but receives the already deserialized
write request.

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   * Prevent compaction loop in TSDB on data gap.
 * [ENHANCEMENT] Return server side performance metrics for query-frontend (using Server-timing header). #3685
 * [ENHANCEMENT] Runtime Config: Add a `mode` query parameter for the runtime config endpoint. `/runtime_config?mode=diff` now shows the YAML runtime configuration with all values that differ from the defaults. #3700
+* [ENHANCEMENT] Distributor: Enable downstream projects to wrap distributor push function and access the deserialized write requests berfore/after they are pushed. #3755
 * [BUGFIX] HA Tracker: don't track as error in the `cortex_kv_request_duration_seconds` metric a CAS operation intentionally aborted. #3745
 
 ## 1.7.0 in progress
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -35,6 +35,9 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/push"
 )
 
+// DistributorPushWrapper wraps around a push. It is similar to middleware.Interface.
+type DistributorPushWrapper func(next push.Func) push.Func
+
 type Config struct {
 	ResponseCompression bool `yaml:"response_compression_enabled"`
 
@@ -45,6 +48,10 @@ type Config struct {
 	ServerPrefix       string               `yaml:"-"`
 	LegacyHTTPPrefix   string               `yaml:"-"`
 	HTTPAuthMiddleware middleware.Interface `yaml:"-"`
+
+	// This allows downstream projects to wrap the distributor push function
+	// and access the deserialized write requests before/after they are pushed.
+	DistributorPushWrapper DistributorPushWrapper `yaml:"-"`
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet.
@@ -57,6 +64,15 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.AlertmanagerHTTPPrefix, prefix+"http.alertmanager-http-prefix", "/alertmanager", "HTTP URL path under which the Alertmanager ui and api will be served.")
 	f.StringVar(&cfg.PrometheusHTTPPrefix, prefix+"http.prometheus-http-prefix", "/prometheus", "HTTP URL path under which the Prometheus api will be served.")
+}
+
+// Push either wraps the distributor push function as configured or returns the distributor push directly.
+func (cfg *Config) wrapDistributorPush(d *distributor.Distributor) push.Func {
+	if cfg.DistributorPushWrapper != nil {
+		return cfg.DistributorPushWrapper(d.Push)
+	}
+
+	return d.Push
 }
 
 type API struct {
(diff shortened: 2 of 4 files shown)


pstibrany:
Would it make sense to reuse recently introduced `PushOnlyIngesterServer` instead?

pracucci (commented):
LGTM to me. Just left minor comments.

> Would it make sense to reuse recently introduced PushOnlyIngesterServer instead?

@pstibrany I'm not sure. I talked to Karsten offline and the use case here is to run the distributor as is but being able, in a downstream project, to access the `WriteRequest` before (or after) it has been handled by the `Distributor.Push()`. I think a wrapper here is what we want. WDYT?

pstibrany:
> @pstibrany I'm not sure. I talked to Karsten offline and the use case here is to run the distributor as is but being able, in a downstream project, to access the `WriteRequest` before (or after) it has been handled by the `Distributor.Push()`. I think a wrapper here is what we want. WDYT?

What I meant was to modify `util/push.Handler` like this:

```
func Handler(cfg distributor.Config int, sourceIPs *middleware.SourceIPExtractor, push client.PushOnlyIngesterServer) http.Handler
```

Which would work and wouldn't introduce new dependencies, but it doesn't really solve the problem this PR is solving.

While working on this, we I'd suggest we reduce the dependency from `util/push` to `distributor` package, by
1) moving `DistributorPushFunc` to `util/push`,
2) applying the push wrapper before calling `util/push.Handler`, and
3) passing `cfg.MaxRecvMsgSize` as separate parameter to Handler.

WDYT?

jeschkies:
@pracucci the linter complains
```
pkg/distributor/distributor.go:399:6: type name will be used as distributor.DistributorPushFunc by other packages, and that stutters; consider calling this PushFunc (golint)
type DistributorPushFunc func(context.Context, *ingester_client.WriteRequest) (*ingester_client.WriteResponse, error)
     ^
```
Should I ignore it?

jeschkies:
> moving DistributorPushFunc to util/push (interfaces are typically declared at place where they are "consumed", ie. called)

@pstibrany, I've tried that but it introduces a circular dependency since the config in `distrobutor.go` must know the type `DistributorPushWrapper`.

> applying the push wrapper before calling util/push.Handler, and

I was thinking about this. However, the wrapper is defined in the config. Since the config is passed to the handler why not evaluate it there?

> passing cfg.MaxRecvMsgSize as separate parameter to Handler.
Ah, I start to understand. You don't want the dependency on config and thus avoid the circular dependency. I'm open to that. @pracucci, what's your take?

pracucci:
> Ah, I start to understand. You don't want the dependency on config and thus avoid the circular dependency. I'm open to that. @pracucci, what's your take?

@jeschkies I'm fine with Peter's suggestion 👍

pracucci:
> type DistributorPushFunc func(context.Context, *ingester_client.WriteRequest) (*ingester_client.WriteResponse, error)
> Should I ignore it?

Once you will move it to `pkg/util` you can just name it `PushFunc` because it will be a generic utility there with no "knowledge" of the distributor.

jeschkies:
> Once you will move it to pkg/util you can just name it PushFunc because it will be a generic utility there with no "knowledge" of the distributor.

I called it `Func` because the linter would complain about stuttering. Unfortunately my tests don't make any sense anymore. How can I test the logic in the `RegisterDistributor` method?

pkg/api/api.go
```diff
@@ -189,7 +189,15 @@ func (a *API) RegisterRuntimeConfig(runtimeConfigHandler http.HandlerFunc) {
...
 	client.RegisterPushOnlyIngesterServer(a.server.GRPC, d)
 
-	a.RegisterRoute("/api/v1/push", push.Handler(pushConfig, a.sourceIPs, d.Push), true, "POST")
+
+	var pushFunc push.Func
+	if pushConfig.DistributorPushWrapper != nil {
+		pushFunc = pushConfig.DistributorPushWrapper(d.Push)
+	} else {
+		pushFunc = d.Push
+	}
```
jeschkies, line 198: How can I test this change? Should I make it part of the `distributor.Config`?

```golang
func (c *Config) Push(d *Distributor) push.Func {
    if pushConfig.DistributorPushWrapper != nil {
        return c.DistributorPushWrapper(d.Push)
    } else {
        return d.Push
    }
}
```

pstibrany: I wouldn't worry about it too much, it's pretty straightforward logic.

pstibrany (approved):
Thanks for addressing our feedback!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * [ENHANCEMENT] Return server side performance metrics for query-frontend (using Server-timing header). #3685
 * [ENHANCEMENT] Runtime Config: Add a `mode` query parameter for the runtime config endpoint. `/runtime_config?mode=diff` now shows the YAML runtime configuration with all values that differ from the defaults. #3700
 * [BUGFIX] HA Tracker: don't track as error in the `cortex_kv_request_duration_seconds` metric a CAS operation intentionally aborted. #3745
+* [ENHANCEMENT] Distributor: Enable downstream projects to wrap distributor push function and access the deserialized write requests berfore/after they are pushed. #3755
```
pracucci, line 17: Move this up, along with the group of `[ENHANCEMENT]`.

pkg/api/api.go
```diff
@@ -59,6 +66,15 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.PrometheusHTTPPrefix, prefix+"http.prometheus-http-prefix", "/prometheus", "HTTP URL path under which the Prometheus api will be served.")
 }
 
+// Push either wraps the distributor push function as configured or returns the distributor push directly.
+func (cfg *Config) Push(d *distributor.Distributor) push.Func {
```
pracucci, line 70: A couple of things:
1. Since it doesn't push, but it wraps the push function, I would call it `wrapDistributorPush()`
2. No need to export it, so name should start with a lowercase letter (eg. `wrapDistributorPush()`)

pracucci (approved):
LGTM (modulo a couple of nits). Thanks!

PR #3498, opened 2020-11-17
Add index-header lazy loading support to store-gateway

pracucci:
**What this PR does**:
When running a blocks storage cluster with many tenants, you may end up with a large amount of blocks (and thus index-headers) managed by each store-gateway. This could end up to hit the limit of the max per-process mmap areas (because each index-header gets mmap-ed). To overcome this, I built into Thanos the ability to lazy mmap index-headers (https://github.com/thanos-io/thanos/pull/3431) and in this PR I'm upgrading Thanos to bring this enhancement (optional and experimental).

I checked other Thanos changes and should be fine.

**Which issue(s) this PR fixes**:
N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3758, opened 2021-01-29
Expose index-header lazy load option and recommends bucket index

pracucci:
**What this PR does**:
The index-header lazy loading was built in #3498 and released in Cortex 1.6.0, but we never exposed its config because we were dubious about potential side effects. We're now running in production and looks fine, so I would suggest to expose its config option.

Moreover, our confidence on bucket index has increased, so I would start recommending it in our production tips.

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added


pstibrany (approved):

jtlisi (approved):
LGTM

PR #3760, opened 2021-01-29
Chore: renamed some pkg/ingester ingester occurrences to instance

pracucci:
**What this PR does**:
The ring is currently used by a large number of services, not just ingesters, but the ring codes (and some errors) still mention "ingester". For the new logic we've added, we've named it "instance" (instead of "ingester") in recent times, but all old code still calls it "ingester".

In this chore PR I'm renaming some occurrences of "ingester" to "instance" in `pkg/ring/`. Protobuf has not been touched and same for config because LifecyclerConfig and RingConfig is actually used only by ingesters (other services use the new basic lifecycler and a custom ring config).

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


khaines (approved):
This was good to clean up! LGTM

pstibrany (approved):

PR #3761, opened 2021-01-29
Fix RingServiceDiscovery logic on empty ring

pracucci:
**What this PR does**:
While checking https://github.com/cortexproject/cortex/issues/3743 I've noticed that the `ErrEmptyRing` is treated as an error, but it's actually a valid state with regards to the ring service discovery, because it just means the ring is empty. This PR fixes it.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


khaines (approved):

jtlisi (approved):
LGTM,

Would this warrant a bugfix entry in the changelog?

pracucci:
> Would this warrant a bugfix entry in the changelog?

Yes, why not. The logic before the fix shouldn't cause any real harm (the pool skips to health check clients until the ring is empty) except a misleading log error.

PR #3763, opened 2021-01-29
Make runtime configuration struct public

56quarters:
Make the struct used for runtime configuration public to allow subscribers
to `runtimeconfig.Manager` update channels to cast messages and use the
configuration values they receive. This is useful for allowing extension
modules to use runtime configuration.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cortex/runtime_config.go
+++ b/pkg/cortex/runtime_config.go
@@ -17,17 +17,17 @@ var (
 	errMultipleDocuments = errors.New("the provided runtime configuration contains multiple documents")
 )
 
-// runtimeConfigValues are values that can be reloaded from configuration file while Cortex is running.
+// RuntimeConfigValues are values that can be reloaded from configuration file while Cortex is running.
 // Reloading is done by runtime_config.Manager, which also keeps the currently loaded config.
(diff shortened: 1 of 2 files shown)


jtlisi (approved):
LGTM

pstibrany:
Can you please elaborate on how you plan to use the fact that `runtimeConfigValues` is exported? It seems to me that it would only give you access to fields already declared in this struct, is that really useful?

Wouldn’t it be more interesting to have some generic extension field inside `runtimeConfigValues` and create a possibility for observing that?

pstibrany:
Asking because this struct was made private on purpose, as it’s very Cortex-specific. Eg. Loki uses its own. I assume that other derived projects could use the same approach and define custom structure too.

pstibrany:
Discussed offline, with some suggested alternatives.

I'd just add that making `runtimeConfigValues` public would likely lead to import cycles when trying to use it, since this struct lives in `pkg/cortex`, which is basically root package for Cortex package-depedency tree. (That again was done on purpose to avoid trying to reuse it.)

56quarters:
Closing in favor of a different approach

PR #3764, opened 2021-01-29
Optimised Images

agarwalutkarsh554:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pracucci:
I'm a bit dubious about the benefit of this PR. To review the correctness we would have to check every single image and the benefit (in terms of size reduction) doesn't look like worth the time. I'm going to pass reviewing this, unless someone else wants to jump it.

pstibrany:
👍 for closing. Our website is not so image heavy that this would bring much benefit.

Issue #3579, opened 2020-12-09
Simplify unit tests runner


pracucci:
Currently, unit tests are run via `tools/test` script. This script looks overcomplicated for what we need right now and the following command should be enough:

```
go test -tags netgo -timeout 20m -race -covermode=atomic -count=1 ./cmd/... ./pkg/... ./tools/...
```

However, if I run `go test ./pkg/...` some ring KV tests for etcd fail while if I run go test on each individual package (like tools/test) does it works. I spent some time investigating and I haven't understood the root cause yet.

_Below you can find the issues I've seen._

## Panic

```
panic: invalid Go type int for field go_etcd_io.etcd.etcdserver.etcdserverpb.loggablePutRequest.value_size

goroutine 419 [running]:
google.golang.org/protobuf/internal/impl.newSingularConverter(0x21606e0, 0x1d9d2e0, 0x21605c0, 0xc0005420e0, 0x214dc01, 0xc000a89700)
	/Users/marco/workspace/src/github.com/cortexproject/cortex/vendor/google.golang.org/protobuf/internal/impl/convert.go:143 +0xf17
google.golang.org/protobuf/internal/impl.NewConverter(0x21606e0, 0x1d9d2e0, 0x21605c0, 0xc0005420e0, 0x203000, 0xc000b38240)
	/Users/marco/workspace/src/github.com/cortexproject/cortex/vendor/google.golang.org/protobuf/internal/impl/convert.go:60 +0xdc
google.golang.org/protobuf/internal/impl.fieldInfoForScalar(0x21605c0, 0xc0005420e0, 0x1d8b86f, 0x9, 0x0, 0x0, 0x21606e0, 0x1d9d2e0, 0x1d8b87a, 0x2e, ...)
	/Users/marco/workspace/src/github.com/cortexproject/cortex/vendor/google.golang.org/protobuf/internal/impl/message_reflect_field.go:234 +0x115
google.golang.org/protobuf/internal/impl.(*MessageInfo).makeKnownFieldsFunc(0xc00022fa40, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xc000b38150, 0xc000b38180, 0xc000b381b0, 0xc000b381e0)
	/Users/marco/workspace/src/github.com/cortexproject/cortex/vendor/google.golang.org/protobuf/internal/impl/message_reflect.go:67 +0x97d
[...]

PR #3767, opened 2021-01-30
Fixing tests

pstibrany:
**What this PR does**: This PR updates snappy package to latest master version (it has fix for arm64 assembly code, which panics in v0.0.2 when running on upcoming go1.16) and fixes tests to run on darwin/arm64 platform.

This also partially fixes https://github.com/cortexproject/cortex/issues/3579 (`TestWatchPrefix/etcd` test).

(Running with go1.16rc1)

**Checklist**
- [x] Tests updated
- [na] Documentation added


pkg/ring/kv/kv_test.go
```diff
@@ -197,6 +201,7 @@ func TestWatchPrefix(t *testing.T) {
 				}
 				require.NoError(t, err)
 			}
+			fmt.Println("Generator finished in", time.Since(start))
```
pracucci, line 204: If you need to keep these `fmt.Println()`, what do you think about `t.Log()` instead? Otherwise I would drop them.

pracucci (approved):
LGTM, thanks! So, should I assume you've bought the new ARM-based MacBook? Jokes apart, is the snappy upgrade worth a CHANGELOG entry about what it fixes?

pkg/ring/kv/kv_test.go
pstibrany: > is the snappy upgrade worth a CHANGELOG entry about what it fixes?

It only fixes arm64 code on (still unreleased) go1.16. Given that, I don't think it's worth the changelog entry.

pstibrany:
> is the snappy upgrade worth a CHANGELOG entry about what it fixes?

It only fixes arm64 code on (still unreleased) go1.16. Given that, I don't think it's worth the changelog entry.

PR #3770, opened 2021-02-01
thanosconvert: convert block meta from Thanos to Cortex

### LINKED ISSUE #2920: Build a tool to migrate TSDB blocks from Thanos/Prometheus to Cortex blocks storage
I'm noticing an increased interest to move TSDB blocks from Thanos and/or Prometheus to the Cortex blocks storage. In the PR https://github.com/cortexproject/cortex/pull/2919 I've documented the expected procedure, but would be great to have a tool to automate the rewrite the `meta.json` in the bucket.

My idea would be to leave the blocks copy/upload procedure to the user (because it highly depends on the user's setup), while offering a tool to automatically rewrite the `meta.json` in the bucket. The procedure is documented in the [PR](https://github.com/cortexproject/cortex/pull/2919) but its TL;DR is:

1. Iterate the tenants in the bucket
2. For each tenant, iterate the blocks in the bucket
3. For each block, download the `meta.json`, decode, [manipulate](https://github.com/cortexproject/cortex/pull/2919), re-encode and re-upload

milesbxf:
**What this PR does**:

thanosconvert is a small tool to iterate over a bucket and convert block-level metadata (in meta.json) so that it is readable by Cortex. This tool assumes that users have already copied Thanos blocks over to a destination bucket, with blocks organised by user at the top level.

This is a relatively simple serial implementation - it's pretty quick to run so making it concurrent is probably unnecessary, at least for a v1. I've only tested this with S3, but since it uses `github.com/thanos-io/thanos/pkg/objstore` it should work with all the other supported block stores. It reuses the same `pkg/storage/bucket.Config` struct for bucket configuration so should be easy to configure for users who already have the store-gateway/querier running.

It's idempotent so is safe to run multiple times against the same bucket, and leaves Cortex blocks untouched. It will log any errors it encounters but keeps going.

I'm happy that this is mostly complete, but there's a few things left to do - I'd love some feedback before I finish it off!

Fixes #2920

**Checklist**
- [x] Add CLI usage message
- [x] Clean up, comment public API methods
- [x] Documentation added

### DIFF
--- a/cmd/thanosconvert/main.go
+++ b/cmd/thanosconvert/main.go
@@ -17,16 +17,17 @@ import (
 	"github.com/cortexproject/cortex/tools/thanosconvert"
 )
 
-var (
-	configFilename string
-	dryRun         bool
-)
-
 func main() {
+	var (
+		configFilename string
+		dryRun         bool
+		cfg            bucket.Config
+	)
 
 	logfmt, loglvl := logging.Format{}, logging.Level{}
 	logfmt.RegisterFlags(flag.CommandLine)
 	loglvl.RegisterFlags(flag.CommandLine)
+	cfg.RegisterFlags(flag.CommandLine)
 	flag.StringVar(&configFilename, "config", "", "Path to bucket config YAML")
 	flag.BoolVar(&dryRun, "dry-run", false, "Don't make changes; only report what needs to be done")
 	flag.Usage = func() {
@@ -43,26 +44,26 @@ func main() {
 		os.Exit(1)
 	}
 
-	if configFilename == "" {
-		fmt.Fprintf(flag.CommandLine.Output(), "Error: -config flag is required\n\n")
+	if configFilename != "" {
+		buf, err := ioutil.ReadFile(configFilename)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to load config file", "err", err, "filename", configFilename)
+			os.Exit(1)
+		}
+		err = yaml.UnmarshalStrict(buf, &cfg)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to parse config", "err", err)
+			os.Exit(1)
+		}
+	}
+
+	if err := cfg.Validate(); err != nil {
+		fmt.Fprintf(flag.CommandLine.Output(), "Error: Bucket config is invalid. Error: %v\n\n", err)
 		flag.Usage()
 		os.Exit(1)
 	}
 
 	ctx := context.Background()
-
-	cfg := bucket.Config{}
-
-	buf, err := ioutil.ReadFile(configFilename)
-	if err != nil {
-		level.Error(logger).Log("msg", "failed to load config file", "err", err, "filename", configFilename)
-		os.Exit(1)
-	}
-	err = yaml.UnmarshalStrict(buf, &cfg)
-	if err != nil {
-		level.Error(logger).Log("msg", "failed to parse config", "err", err)
-		os.Exit(1)
-	}
 
 	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, logger)
 	if err != nil {
--- a/cmd/thanosconvert/main.go
+++ b/cmd/thanosconvert/main.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 
 	"github.com/go-kit/kit/log/level"
-	"github.com/weaveworks/common/server"
+	"github.com/weaveworks/common/logging"
 	"gopkg.in/yaml.v2"
 
 	"github.com/cortexproject/cortex/pkg/storage/bucket"
@@ -23,16 +23,29 @@ var (
 )
 
 func main() {
-	serverConfig := server.Config{}
-	serverConfig.RegisterFlags(flag.CommandLine)
+
+	logfmt, loglvl := logging.Format{}, logging.Level{}
+	logfmt.RegisterFlags(flag.CommandLine)
+	loglvl.RegisterFlags(flag.CommandLine)
 	flag.StringVar(&configFilename, "config", "", "Path to bucket config YAML")
 	flag.BoolVar(&dryRun, "dry-run", false, "Don't make changes; only report what needs to be done")
+	flag.Usage = func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "%s is a tool to convert block metadata from Thanos to Cortex.\nPlease see %s for instructions on how to run it.\n\n", os.Args[0], "https://cortexmetrics.io/docs/blocks-storage/migrate-storage-from-thanos-and-prometheus/")
+		fmt.Fprintf(flag.CommandLine.Output(), "Flags:\n")
+		flag.PrintDefaults()
+	}
 	flag.Parse()
 
-	util.InitLogger(&serverConfig)
+	logger, err := util.NewPrometheusLogger(loglvl, logfmt)
+	if err != nil {
+		fmt.Printf("failed to create logger: %v\n", err)
+		flag.Usage()
+		os.Exit(1)
+	}
 
 	if configFilename == "" {
-		level.Error(util.Logger).Log("msg", "-config is required")
+		fmt.Fprintf(flag.CommandLine.Output(), "Error: -config flag is required\n\n")
+		flag.Usage()
 		os.Exit(1)
 	}
 
@@ -42,25 +55,25 @@ func main() {
 
 	buf, err := ioutil.ReadFile(configFilename)
 	if err != nil {
-		level.Error(util.Logger).Log("msg", "failed to load config file", "err", err, "filename", configFilename)
+		level.Error(logger).Log("msg", "failed to load config file", "err", err, "filename", configFilename)
 		os.Exit(1)
 	}
 	err = yaml.UnmarshalStrict(buf, &cfg)
 	if err != nil {
-		level.Error(util.Logger).Log("msg", "failed to parse config", "err", err)
+		level.Error(logger).Log("msg", "failed to parse config", "err", err)
 		os.Exit(1)
 	}
 
-	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, util.Logger)
+	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, logger)
 	if err != nil {
-		level.Error(util.Logger).Log("msg", "failed to initialize", "err", err)
+		level.Error(logger).Log("msg", "failed to initialize", "err", err)
 		os.Exit(1)
 	}
 
 	iterCtx := context.Background()
 	results, err := converter.Run(iterCtx)
 	if err != nil {
-		level.Error(util.Logger).Log("msg", "error while iterating blocks", "err", err)
+		level.Error(logger).Log("msg", "error while iterating blocks", "err", err)
 		os.Exit(1)
 	}
 
--- a/cmd/thanosconvert/main.go
+++ b/cmd/thanosconvert/main.go
@@ -13,7 +13,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/cortexproject/cortex/pkg/storage/bucket"
-	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/pkg/util/log"
 	"github.com/cortexproject/cortex/tools/thanosconvert"
 )
 
@@ -36,7 +36,7 @@ func main() {
 	}
 	flag.Parse()
 
-	logger, err := util.NewPrometheusLogger(loglvl, logfmt)
+	logger, err := log.NewPrometheusLogger(loglvl, logfmt)
 	if err != nil {
 		fmt.Printf("failed to create logger: %v\n", err)
 		flag.Usage()
--- a/cmd/thanosconvert/main.go
+++ b/cmd/thanosconvert/main.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/go-kit/kit/log/level"
 	"github.com/weaveworks/common/logging"
 	"gopkg.in/yaml.v2"
 
@@ -39,44 +38,33 @@ func main() {
 
 	logger, err := log.NewPrometheusLogger(loglvl, logfmt)
 	if err != nil {
-		fmt.Printf("failed to create logger: %v\n", err)
-		flag.Usage()
-		os.Exit(1)
+		fatal("failed to create logger: %v", err)
 	}
 
 	if configFilename != "" {
 		buf, err := ioutil.ReadFile(configFilename)
 		if err != nil {
-			level.Error(logger).Log("msg", "failed to load config file", "err", err, "filename", configFilename)
-			os.Exit(1)
+			fatal("failed to load config file from %s: %v", configFilename, err)
 		}
 		err = yaml.UnmarshalStrict(buf, &cfg)
 		if err != nil {
-			level.Error(logger).Log("msg", "failed to parse config", "err", err)
-			os.Exit(1)
+			fatal("failed to parse config file: %v", err)
 		}
 	}
 
 	if err := cfg.Validate(); err != nil {
-		fmt.Fprintf(flag.CommandLine.Output(), "Error: Bucket config is invalid. Error: %v\n\n", err)
-		flag.Usage()
-		os.Exit(1)
+		fatal("bucket config is invalid: %v", err)
 	}
 
 	ctx := context.Background()
 
 	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, logger)
 	if err != nil {
-		level.Error(logger).Log("msg", "failed to initialize", "err", err)
-		os.Exit(1)
+		fatal("couldn't initilize converter: %v", err)
 	}
 
 	iterCtx := context.Background()
 	results, err := converter.Run(iterCtx)
-	if err != nil {
-		level.Error(logger).Log("msg", "error while iterating blocks", "err", err)
-		os.Exit(1)
-	}
 
 	fmt.Println("Results:")
 	for user, res := range results {
@@ -86,4 +74,13 @@ func main() {
 		fmt.Printf("  Failed %d:\n  %s", len(res.FailedBlocks), strings.Join(res.FailedBlocks, ","))
 	}
 
+	if err != nil {
+		fatal("converter failed: %v", err)
+	}
+
+}
+
+func fatal(msg string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, msg+"\n", args...)
+	os.Exit(1)
 }
--- a/cmd/thanosconvert/main.go
+++ b/cmd/thanosconvert/main.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/go-kit/kit/log/level"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/weaveworks/common/server"
 	"gopkg.in/yaml.v2"
 
@@ -37,8 +36,6 @@ func main() {
 		os.Exit(1)
 	}
 
-	registry := prometheus.DefaultRegisterer
-
 	ctx := context.Background()
 
 	cfg := bucket.Config{}
@@ -54,7 +51,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, util.Logger, registry)
+	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, util.Logger)
 	if err != nil {
 		level.Error(util.Logger).Log("msg", "failed to initialize", "err", err)
 		os.Exit(1)
@@ -63,7 +60,7 @@ func main() {
 	iterCtx := context.Background()
 	results, err := converter.Run(iterCtx)
 	if err != nil {
-		level.Error(util.Logger).Log("msg", "iterate blocks", "err", err)
+		level.Error(util.Logger).Log("msg", "error while iterating blocks", "err", err)
 		os.Exit(1)
 	}
 
--- a/cmd/thanosconvert/main.go
+++ b/cmd/thanosconvert/main.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/weaveworks/common/server"
+	"gopkg.in/yaml.v2"
+
+	"github.com/cortexproject/cortex/pkg/storage/bucket"
+	"github.com/cortexproject/cortex/pkg/util"
+	"github.com/cortexproject/cortex/tools/thanosconvert"
+)
+
+var (
+	configFilename string
+	dryRun         bool
+)
+
+func main() {
+	serverConfig := server.Config{}
+	serverConfig.RegisterFlags(flag.CommandLine)
+	flag.StringVar(&configFilename, "config", "", "Path to bucket config YAML")
+	flag.BoolVar(&dryRun, "dry-run", false, "Don't make changes; only report what needs to be done")
+	flag.Parse()
+
+	util.InitLogger(&serverConfig)
+
+	if configFilename == "" {
+		level.Error(util.Logger).Log("msg", "-config is required")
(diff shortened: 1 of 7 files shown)


cmd/thanosconvert/main.go
```diff
@@ -0,0 +1,78 @@
...
+	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, util.Logger, registry)
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "failed to initialize", "err", err)
+		os.Exit(1)
+	}
+
+	iterCtx := context.Background()
+	results, err := converter.Run(iterCtx)
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "iterate blocks", "err", err)
```
pracucci, line 66:
```suggestion
		level.Error(util.Logger).Log("msg", "error while iterating blocks", "err", err)
```

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (r *PerUserResults) AddUnchanged(blockID string) {
+	r.UnchangedBlocks = append(r.UnchangedBlocks, blockID)
+}
+
+type Results map[string]PerUserResults
+
+// NewThanosBlockConverter creates a ThanosBlockConverter
+func NewThanosBlockConverter(ctx context.Context, cfg bucket.Config, dryRun bool, logger log.Logger, reg prometheus.Registerer) (*ThanosBlockConverter, error) {
```
pracucci, line 53: I would remove `reg prometheus.Registerer` given this tool doesn't export metrics anyway, and just pass `nil` as registered to `bucket.NewClient()` (when `nil` metrics registration will be skipped).

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,195 @@
...
+// Run iterates over all blocks from all users in a bucket
+func (c ThanosBlockConverter) Run(ctx context.Context) (Results, error) {
+	results := make(Results)
+
+	// discover users in the bucket
+	var users []string
+	err := c.bkt.Iter(ctx, "", func(o string) error {
+		users = append(users, strings.TrimSuffix(o, "/"))
+		return nil
+	})
```
pracucci, line 75: You may want to reuse the scanner we already have:

```
scanner := cortex_tsdb.NewUsersScanner(c.bkt, cortex_tsdb.AllUsers)
users, _, err := scanner.ScanUsers(ctx)
```

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,195 @@
...
+			return results, errors.Wrap(err, fmt.Sprintf("error converting user %s", u))
+		}
+	}
+	return results, nil
+}
+
+func (c ThanosBlockConverter) convertUser(ctx context.Context, user string) (PerUserResults, error) {
+	results := PerUserResults{}
+
+	err := c.bkt.Iter(ctx, user, func(o string) error {
```
pracucci, line 94: I would wrap `c.bkt` with `bucket.NewUserBucketClient()`. This will mask you the first object prefix segment (which is the `<tenant ID>/`).

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,195 @@
...
+		// get block ID from path
+
+		parts := strings.Split(o, "/")
```
pracucci, line 97: Once you use the `UserBuckerClient`, this should just be a matter of:

```
o = strings.TrimSuffix(o, "/")
blockID, err := ulid.Parse(o)
```

If the parse fail, just skip it (no error) because it means it's not a block. We don't have just blocks in bucket.

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,195 @@
...
+		r, err := c.bkt.Get(ctx, metaKey)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "Get meta.json", "block", blockID, "meta_key", metaKey, "err", err.Error())
+			results.AddFailed(blockID)
+			return nil
+		}
+
+		meta := &metadata.Meta{}
+		decoder := json.NewDecoder(r)
+		err = decoder.Decode(meta)
```
pracucci, line 117: Please use `metadata.Read()` from Thanos package `github.com/thanos-io/thanos/pkg/block/metadata/`

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,195 @@
...
+func convertMetadata(meta metadata.Meta, expectedUser string) (metadata.Meta, []string) {
+	var changesRequired []string
+
+	if meta.Thanos.Labels == nil {
+		// TODO: no "thanos" entry could mean this block is corrupt - should we bail out instead?
+		meta.Thanos.Labels = map[string]string{}
+		// don't need to add to changesRequired, since the code below will notice lack of org id
+	}
+
+	org, ok := meta.Thanos.Labels["__org_id__"]
```
pracucci, line 177: Instead of the hardcoded string `"__org_id__"`, please use `cortex_tsdb.TenantIDExternalLabel`.

tools/thanosconvert/thanosconvert_test.go
```diff
@@ -0,0 +1,355 @@
...
+			}
+
+			ctx := context.Background()
+			results, err := converter.Run(ctx)
+			test.assertions(t, bkt, results, err)
+		})
+	}
+}
+
+func CortexMeta(user string) metadata.Meta {
```
pracucci, line 167: No need to export this function.

tools/thanosconvert/thanosconvert_test.go
```diff
@@ -0,0 +1,355 @@
...
+	return metadata.Meta{
+		Thanos: metadata.Thanos{
+			Labels: map[string]string{
+				"__org_id__": user,
+			},
+		},
+	}
+}
+
+func ThanosMeta() metadata.Meta {
```
pracucci, line 177: No need to export this function.

tools/thanosconvert/thanosconvert_test.go
```diff
@@ -0,0 +1,355 @@
...
+	return metadata.Meta{
+		Thanos: metadata.Thanos{
+			Labels: map[string]string{
+				"cluster": "foo",
+			},
+		},
+	}
+}
+
+func GetLogger() log.Logger {
```
pracucci, line 187: No need to export this function.

tools/thanosconvert/thanosconvert_test.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	logger, err := util.NewPrometheusLogger(l, f)
+	if err != nil {
+		panic(err)
+	}
+	return logger
+}
+
+func MockBucket(data fakeBucket) *bucket.ClientMock {
```
pracucci, line 205: No need to export this function.

tools/thanosconvert/thanosconvert_test.go
```diff
@@ -0,0 +1,355 @@
...
+				assert.Len(t, results["user1"].FailedBlocks, 0)
+				assert.Len(t, results["user1"].ConvertedBlocks, 0)
+				assert.Len(t, results["user1"].UnchangedBlocks, 0)
+			},
+		},
+		{
+			name: "bucket fully converted is a noop",
+			bucketData: fakeBucket{
+				"user1": map[string]metadata.Meta{
+					"block1": CortexMeta("user1"),
```
pracucci, line 59: Once you start using `ulid.Parse()` these tests will fail because `"block1"` is not a valid block ID. You could just define at the beginning of the test function something like the following, and then use such variables across the test cases:

```
block1 := ulid.MustNew(1, nil)
block2 := ulid.MustNew(2, nil)
```

pracucci (commented):
Fantastic job! We wanted this since a long time, thanks! I left few comments here and there, but mostly nits that should be very quick to address.

Other than the comments:
- Could you add it to the list of experimental features listed in `docs/configuration/v1-guarantees.md`, please?
- Could you mention the tool in `docs/blocks-storage/migrate-storage-from-thanos-and-prometheus.md`? There's a sentence like "no tool is provided to migrate TSDB blocks from Thanos" while now there is 😉

milesbxf:
> Fantastic job! We wanted this since a long time, thanks! I left few comments here and there, but mostly nits that should be very quick to address.
> 
> Other than the comments:
> 
> * Could you add it to the list of experimental features listed in `docs/configuration/v1-guarantees.md`, please?
> * Could you mention the tool in `docs/blocks-storage/migrate-storage-from-thanos-and-prometheus.md`? There's a sentence like "no tool is provided to migrate TSDB blocks from Thanos" while now there is 😉

@pracucci thanks for the detailed review! I think I've addressed all your comments - I've added a new commit to make it easier to review the changes, but can squash it down once you're happy. I'll make the docs changes today.

milesbxf:
👋  @pracucci I've updated the docs and added a CHANGELOG entry. Is there anything else I need to do?

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,188 @@
...
+			return nil
+		}
+		blockID := blockULID.String()
+
+		level.Debug(c.logger).Log("msg", "Processing block", "block", blockID)
+
+		// retrieve meta.json
+
+		metaKey := fmt.Sprintf("%s/%s/meta.json", user, blockID)
+		r, err := c.bkt.Get(ctx, metaKey)
```
pracucci, line 106: Use the `userBucketClient` please. It means the path to `meta.json` will be:
`metaKey := fmt.Sprintf("%s/meta.json", blockID)`

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,188 @@
...
+		return errors.Wrap(err, "upload to bucket")
+	}
+	return nil
+}
+
+func convertMetadata(meta metadata.Meta, expectedUser string) (metadata.Meta, []string) {
+	var changesRequired []string
+
+	if meta.Thanos.Labels == nil {
+		// TODO: no "thanos" entry could mean this block is corrupt - should we bail out instead?
```
pracucci, line 165: I would not bail out. If we don't require the Thanos entry, the tool should work with Prometheus blocks too, which I believe is a plus for the community.

pracucci (approved):
Good job, thanks! I left a couple of nits. Can you confirm you manually tested it after changes? I haven't got the time to manually test it.

@pstibrany could you do the 2nd review, please?

docs/blocks-storage/migrate-storage-from-thanos-and-prometheus.md
```diff
@@ -18,18 +18,57 @@ The Cortex blocks storage has few requirements that should be considered when mi
...
-### Manipulate `meta.json` file
+### Migrate block metadata (`meta.json`) to Cortex
+
+For each block copied/uploaded to the Cortex bucket, there are a few changes required to the `meta.json`.
+
+#### Automatically migrate metadata using the `thanosconvert` tool
+
+`thanosconvert` can iterate over a Cortex bucket and make sure that each `meta.json` has the correct `thanos` > `labels` layout.
+
+⚠️  `thanosconvert` will modify files in the bucket you specify. It's recommended that you have backups or enable object versioning before running this tool.
```
pstibrany, line 33:
```suggestion
⚠ Warning ⚠ `thanosconvert` will modify files in the bucket you specify. It's recommended that you have backups or enable object versioning before running this tool.
```

docs/blocks-storage/migrate-storage-from-thanos-and-prometheus.md
```diff
@@ -18,18 +18,57 @@ The Cortex blocks storage has few requirements that should be considered when mi
...
 
-For each block copied/uploaded to the Cortex bucket, the `meta.json` should be manipulated. The easiest approach would be iterating the tenants and blocks in the bucket and for each block:
+To run `thanosconvert`, you need to provide a YAML file with bucket configuration in the same format as the [blocks storage bucket configuration](../configuration/config-file-reference.md#blocks_storage_config). For example, for s3:
```
pstibrany, line 35:
```suggestion
To run `thanosconvert`, you need to provide it with the bucket configuration in the same format as the [blocks storage bucket configuration](../configuration/config-file-reference.md#blocks_storage_config).
```

(command line options work too here)

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,188 @@
...
+	dryRun bool
+}
+
+type PerUserResults struct {
+	FailedBlocks, ConvertedBlocks, UnchangedBlocks []string
+}
+
+func NewPerUserResults() PerUserResults {
+	return PerUserResults{
+		FailedBlocks:    []string{},
```
pstibrany, line 33: Nit: this creates 0-sized slice, which is unnecessary. As it is now, `PerUserResults{}` is perfectly fine default value, and doesn't even need "New" function -- which isn't used anywhere.

(Furthermore, since methods are defined on `*PerUserResults`, this method should return pointer. But I'd suggest to remove it completely.)

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// Run iterates over all blocks from all users in a bucket
+func (c ThanosBlockConverter) Run(ctx context.Context) (Results, error) {
+	results := make(Results)
+
+	scanner := cortex_tsdb.NewUsersScanner(c.bkt, cortex_tsdb.AllUsers, c.logger)
+	users, _, err := scanner.ScanUsers(ctx)
+	if err != nil {
+		return results, errors.Wrap(err, "error while scanning users")
```
pstibrany, line 73: Nit: unify "users" vs "tenants" terminology in the log and error messages.

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,188 @@
...
+	return results, nil
+}
+
+func (c ThanosBlockConverter) convertUser(ctx context.Context, user string) (PerUserResults, error) {
+	results := PerUserResults{}
+
+	userBucketClient := bucket.NewUserBucketClient(user, c.bkt)
+	err := userBucketClient.Iter(ctx, "", func(o string) error {
+		// get block ID from path
+		o = strings.TrimSuffix(o, "/")
```
pstibrany, line 93: We can shorten the code here by using `id, ok := block.IsBlockDir(o)`.

tools/thanosconvert/thanosconvert.go
pstibrany: After using `userBucketClient`, we can also use `block.DownloadMeta()` which will download and parse the meta.json file (it doesn't check for Cortex orgid label, so that's fine), to simplify the code here.

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,188 @@
...
+			return nil
+		}
+
+		// convert and upload if appropriate
+
+		newMeta, changesRequired := convertMetadata(*meta, user)
+
+		if len(changesRequired) > 0 {
+			if c.dryRun {
+				level.Debug(c.logger).Log("msg", "Block requires changes, not uploading due to dry run", "block", blockID, "changes_required", strings.Join(changesRequired, ","))
```
pstibrany, line 126: Do we need to repeat "not uploading due to dry run" message for each block? Perhaps just short "(dry-run)" would do.

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,188 @@
...
+		results[u] = r
+		if err != nil {
+			return results, errors.Wrap(err, fmt.Sprintf("error converting user %s", u))
+		}
+	}
+	return results, nil
+}
+
+func (c ThanosBlockConverter) convertUser(ctx context.Context, user string) (PerUserResults, error) {
+	results := PerUserResults{}
```
pstibrany, line 88: I'd suggest to simplify this code by not keeping `PerUserResults` at all. The tool could simply use logging to log converted blocks, progress and failures, and not retain anything in the memory after processing of the user is done. It's only printed anyway -- we could print it immediately.

cmd/thanosconvert/main.go
```diff
@@ -0,0 +1,88 @@
...
+	configFilename string
+	dryRun         bool
+)
+
+func main() {
+
+	logfmt, loglvl := logging.Format{}, logging.Level{}
+	logfmt.RegisterFlags(flag.CommandLine)
+	loglvl.RegisterFlags(flag.CommandLine)
+	flag.StringVar(&configFilename, "config", "", "Path to bucket config YAML")
```
pstibrany, line 30: Nit: I'd suggest supporting command line options too (or perhaps as the only way of configuring). Bucket config needs just a couple of options.

cmd/thanosconvert/main.go
```diff
@@ -0,0 +1,88 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/cortexproject/cortex/pkg/storage/bucket"
+	"github.com/cortexproject/cortex/pkg/util/log"
+	"github.com/cortexproject/cortex/tools/thanosconvert"
+)
+
+var (
+	configFilename string
+	dryRun         bool
```
pstibrany, line 22: Nit: no need for global variables, move to main() method.

pstibrany (approved):
Thank you, this is useful tool. I left bunch of comments suggestions if you feel like working on this code bit more, but they are not blocking... the tool already looks useful as-is.

tools/thanosconvert/thanosconvert.go
milesbxf: Turns out I wasn't even using this constructor anyway 😄

tools/thanosconvert/thanosconvert.go
milesbxf: I've gone with "users" - which one is more correct? Is there a distinction?

tools/thanosconvert/thanosconvert.go
milesbxf: Yes, good point - in my head I was thinking of putting out more sophisticated reporting (perhaps saving to a file), but this is probably unnecessary. On the other hand, this makes testing significantly easier, as we can run assertions on the contents of the results

milesbxf:
> Thank you, this is useful tool. I left bunch of comments suggestions if you feel like working on this code bit more, but they are not blocking... the tool already looks useful as-is.

Thanks for the review - I figured I'd implement a few of your suggestions whilst the PR is still open 🙏

milesbxf:
> Good job, thanks! I left a couple of nits. Can you confirm you manually tested it after changes? I haven't got the time to manually test it.

Thanks! Just to confirm, I've manually tested this after the latest batch of changes (again, only with S3)

milesbxf:
Looks like an unrelated test failed:
```
--- FAIL: TestAlertmanager_ServeHTTP (0.24s)
    multitenant_test.go:90: 
        	Error Trace:	multitenant_test.go:90
        	            				testing.go:846
        	            				testing.go:876
        	            				testing.go:1048
        	Error:      	Received unexpected error:
        	            	unlinkat /tmp/alertmanager319454365: directory not empty
        	Test:       	TestAlertmanager_ServeHTTP
FAIL
coverage: 65.6% of statements in github.com/cortexproject/cortex/pkg/alertmanager
FAIL	github.com/cortexproject/cortex/pkg/alertmanager	32.268s
FAIL
```

I can report an issue for this if you think it's a flake - in the meantime how do I rerun CI?

pstibrany:
> in the meantime how do I rerun CI

There is "Rerun all jobs" button in top-right corner of the workflow run. I've restarted it now, just in case you don't have permissions and don't see that button.

tools/thanosconvert/thanosconvert.go
pstibrany: There is no distinction in Cortex. Our codebase uses both terms interchangeably, and it is sometimes confusing for people.

tools/thanosconvert/thanosconvert.go
pstibrany: Good point about testing.

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,167 @@
...
+			return nil
+		}
+
+		// convert and upload if appropriate
+
+		newMeta, changesRequired := convertMetadata(meta, user)
+
+		if len(changesRequired) > 0 {
+			if c.dryRun {
+				level.Info(c.logger).Log("msg", "Block requires changes (dry-run)", "block", blockID.String(), "changes_required", strings.Join(changesRequired, ","))
```
pstibrany, line 105: I'd suggest to add "user" into all log messages here. (Or use `ulogger := log.With(c.logger, "user", user)` use wrapped ulogger only in entire method.)

tools/thanosconvert/thanosconvert.go
```diff
@@ -0,0 +1,167 @@
...
+	if err := userBucketClient.Upload(ctx, fmt.Sprintf("%s/meta.json", blockID), &body); err != nil {
+		return errors.Wrap(err, "upload to bucket")
+	}
+	return nil
+}
+
+func convertMetadata(meta metadata.Meta, expectedUser string) (metadata.Meta, []string) {
+	var changesRequired []string
+
+	if meta.Thanos.Labels == nil {
```
pstibrany, line 144: No need for this check or creating of empty `meta.Thanos.Labels`, since it will be replaced anyway. Reading from `nil` map, or deleting from `nil` map will work as expected. (I'd suggest adding unit test for that)

cmd/thanosconvert/main.go
```diff
@@ -0,0 +1,89 @@
...
+		fmt.Fprintf(flag.CommandLine.Output(), "%s is a tool to convert block metadata from Thanos to Cortex.\nPlease see %s for instructions on how to run it.\n\n", os.Args[0], "https://cortexmetrics.io/docs/blocks-storage/migrate-storage-from-thanos-and-prometheus/")
+		fmt.Fprintf(flag.CommandLine.Output(), "Flags:\n")
+		flag.PrintDefaults()
+	}
+	flag.Parse()
+
+	logger, err := log.NewPrometheusLogger(loglvl, logfmt)
+	if err != nil {
+		fmt.Printf("failed to create logger: %v\n", err)
+		flag.Usage()
```
pstibrany, line 43: Printing usage hides the error message. I'd suggest removing flag.Usage() calls after print errors.

cmd/thanosconvert/main.go
```diff
@@ -0,0 +1,89 @@
...
+		os.Exit(1)
+	}
+
+	if configFilename != "" {
+		buf, err := ioutil.ReadFile(configFilename)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to load config file", "err", err, "filename", configFilename)
```
pstibrany, line 50: This methods uses three different ways to print errors: logger, `fmt.Printf` (stdout) and `fmt.Fprintf(flag.Commandline.Output())` which uses stderr. I'd suggest to print errors followed by os.Exit(1) to stderr.

cmd/thanosconvert/main.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	iterCtx := context.Background()
+	results, err := converter.Run(iterCtx)
+	if err != nil {
+		level.Error(logger).Log("msg", "error while iterating blocks", "err", err)
+		os.Exit(1)
+	}
+
+	fmt.Println("Results:")
```
pstibrany, line 81: We use logger for logging the progress in `Run` method. Perhaps we should use logger for logging the result too?

cmd/thanosconvert/main.go
```diff
@@ -0,0 +1,89 @@
...
+
+	converter, err := thanosconvert.NewThanosBlockConverter(ctx, cfg, dryRun, logger)
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to initialize", "err", err)
+		os.Exit(1)
+	}
+
+	iterCtx := context.Background()
+	results, err := converter.Run(iterCtx)
+	if err != nil {
```
pstibrany, line 76: There may be some results already when this error happens. Should this code simply log results first, and error (if any) as last?

pstibrany (approved):
Thanks for addressing my feedback. I've left few more (non-blocking) comments to take a look at, if you want.

milesbxf:
> Thanks for addressing my feedback. I've left few more (non-blocking) comments to take a look at, if you want.

Thanks @pstibrany - I've addressed some of these 🙏 I probably won't be able to spend much more time on this, so let me know if there's anything blocking the merge you'd like me to address

pracucci:
Thanks a lot @milesbxf for your hard work!

> I probably won't be able to spend much more time on this, so let me know if there's anything blocking the merge you'd like me to address

Sure. It's a great tool, let's ship it! 🚀

Issue #3154, opened 2020-09-10
Ring state will be inconsistent between memory and consul after a CAS error


bboreham:
The change in memory state is made before updating Consul, and no attempt is made to revert the former if the latter fails:

https://github.com/cortexproject/cortex/blob/a87c25fd994a2eec5ab5af0a920cc529b17c0030/pkg/ring/lifecycler.go#L710-L711

I noticed this because I got this log message:
```
level=warn ts=2020-09-09T19:59:32.324235593Z caller=grpc_logging.go:55 duration=15.010918473s method=/cortex.Ingester/TransferChunks err="Transfer: ChangeState: failed to CAS collectors/ring" msg="gRPC\n"
```

That's coming from here: https://github.com/cortexproject/cortex/blob/f27cef893d92e38395de0504f922231fc15bb7d8/pkg/ingester/transfer.go#L204

The `defer` in that function should then log "TransferChunks failed" and go back to PENDING state, but I don't see that log, which is explained by this line checking the in-memory state: https://github.com/cortexproject/cortex/blob/f27cef893d92e38395de0504f922231fc15bb7d8/pkg/ingester/transfer.go#L185

(Also odd: metrics show it did go to ACTIVE state)

bboreham:
@pstibrany explained the last point: the next heartbeat will save the in-memory state to Consul.

So, maybe all we need is a better check in `Ingester.transfer()` ?

pstibrany:
What do you think should happen? How do you suggest to modify check in `Ingester.transfer()`?

Changing state doesn't seem appropriate. As of now  the only possible transition from ACTIVE state is to LEAVING state. I don't think that's correct answer.

Other possibilities seem even worse (going back to PENDING/JOINING), because transfer has already finished.

PR #2565, opened 2020-05-06
Allow starting ingester in JOINING state to auto join

csmarchbanks:
If an ingester crashes during token transfer, or fails to rollback a
token transfer then it will start up with a ring entry in the JOINING
state. This will mean the ingester will no longer auto join because it
is not in the PENDING state, and cannot receive a transfer request
either. It will just stay in the JOINING state until manually deleted
and forgotten from the ring.

Instead, handle this case upon init, and set the state back to PENDING.
This will allow the newly started ingester to properly receive a
transfer request or auto join.

Signed-off-by: Chris Marchbanks <csmarchbanks@gmail.com>

**What this PR does**:

**Which issue(s) this PR fixes**:
Fixes #2508

**Checklist**
- [x] Tests updated
- [x] Documentation added - Not needed
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3774, opened 2021-02-02
Fix ring lifecycler when 'unregister on shutdown' is disabled

pracucci:
**What this PR does**:
The PR https://github.com/cortexproject/cortex/pull/3305 introduced an option to let an ingester stay in the ring on shutdown. In some of our clusters, we're already testing it (in conjunction with the option to extend writes on failure) to avoid series resharding on rollout.

When the option is used, I've noticed the following log logged by `ingester-8` at startup:
```
level=warn ts=2021-01-30T17:32:47.008963494Z caller=lifecycler.go:235 msg="found an existing instance(s) with a problem in the ring, this instance cannot become ready until this problem is resolved. The /ring http endpoint on the distributor (or single binary) provides visibility into the ring." ring=ingester err="instance ingester-8 in state LEAVING"
```

Basically, the `ingester-8` was logging that it's not ready because an ingester is in LEAVING state. Well, the ingester was itself and, despite technically correct, I found the log a bit misleading.

This made me start this PR with the aim of just improving `Lifecycler.CheckReady()` and I've ended up discovering other potential side effects of the current ring lifecycler implementation when 'unregister on shutdown' is disabled. I tried to address them in this PR:

1.  `Lifecycler.CheckReady()` should return `false` if the instance itself is not `ACTIVE`
2. If an instance is left in the ring on shutdown and the number of tokens in the config changes, the instance doesn't extends the tokens to match new configured `num-tokens`, because the `autoJoin()` logic is not triggered
3. The case "tokens already exist for this instance in the ring" is not necessarily an error condition (was logged as error), so I moved it to "info"

The ring lifecycler logic is a bit tricky, so I would be glad if several people could take a look.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * [CHANGE] Require explicit flag `-<prefix>.tls-enabled` to enable TLS in GRPC clients. Previously it was enough to specify a TLS flag to enable TLS validation. #3156
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
 * [CHANGE] Ingester: at startup, if the ingester is already registered in the ring with the `LEAVING` state, it will be switched back to `PENDING` keeping the previous tokens and then follow the normal lifecycler workflow. #3774
-* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651
+* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651 #3810
   - `-<prefix>.s3.sse.type`
   - `-<prefix>.s3.sse.kms-key-id`
   - `-<prefix>.s3.sse.kms-encryption-context`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 * [CHANGE] Ingester: don't update internal "last updated" timestamp of TSDB if tenant only sends invalid samples. This affects how "idle" time is computed. #3727
 * [CHANGE] Require explicit flag `-<prefix>.tls-enabled` to enable TLS in GRPC clients. Previously it was enough to specify a TLS flag to enable TLS validation. #3156
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
-* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651 #3810
+* [CHANGE] Ingester: at startup, if the ingester is already registered in the ring with the `LEAVING` state, it will be switched back to `PENDING` keeping the previous tokens and then follow the normal lifecycler workflow. #3774
+* [FEATURE] Adds support to S3 server side encryption using KMS. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651
   - `-<prefix>.s3.sse.type`
   - `-<prefix>.s3.sse.kms-key-id`
   - `-<prefix>.s3.sse.kms-encryption-context`
--- a/pkg/ring/lifecycler.go
+++ b/pkg/ring/lifecycler.go
@@ -567,7 +567,7 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 
 		// The instance may already be in the ring with one of the following states:
 		// - JOINING: this means it crashed due to a failed token transfer or some other reason during startup.
-		// - LEAVING: this means it crashed while shutting down and unregister from ring on shutdown is disabled.
+		// - LEAVING: this means it crashed while shutting down or unregister from ring on shutdown is disabled.
 		// In both cases, we want to set it back to PENDING in order to start the lifecycle from the beginning.
 		if instanceDesc.State == JOINING || instanceDesc.State == LEAVING {
 			level.Info(log.Logger).Log("msg", "instance already found in ring, changing state to PENDING", "state", instanceDesc.State, "ring", i.RingName)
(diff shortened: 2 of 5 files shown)


pkg/ring/lifecycler.go
```diff
@@ -556,21 +559,13 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 		// but we need to update the local state accordingly.
 		i.setRegisteredAt(instanceDesc.GetRegisteredAt())
 
-		// If the ingester is in the JOINING state this means it crashed due to
-		// a failed token transfer or some other reason during startup. We want
-		// to set it back to PENDING in order to start the lifecycle from the
-		// beginning.
-		if instanceDesc.State == JOINING {
```
pracucci, line 563: This was done in https://github.com/cortexproject/cortex/pull/2565

pkg/ring/lifecycler.go
```diff
@@ -556,21 +559,13 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
+		if instanceDesc.State == JOINING || instanceDesc.State == LEAVING {
+			level.Info(log.Logger).Log("msg", "instance already found in ring, changing state to PENDING", "state", instanceDesc.State, "ring", i.RingName)
 			instanceDesc.State = PENDING
-			return ringDesc, true, nil
-		}
-
-		// If the ingester failed to clean it's ring entry up in can leave it's state in LEAVING.
-		// Move it into ACTIVE to ensure the ingester joins the ring.
-		if instanceDesc.State == LEAVING && len(instanceDesc.Tokens) == i.cfg.NumTokens {
-			instanceDesc.State = ACTIVE
```
pracucci, line 573: This was done in https://github.com/cortexproject/cortex/pull/1921. It was updating `instanceDesc`, but the following `return nil, true, nil` was not really updating the ring. The reason why it was working was because the `instanceDesc.State` was set on the local state and then updated during the next heartbeat.

The new logic in this PR `return ringDesc, true, nil` if `instanceDesc` has changed, which I believe is easier to understand and less error prone.

bboreham:
I think #3154 is related

pkg/ring/lifecycler_test.go
```diff
@@ -43,6 +43,8 @@ func checkNormalised(d interface{}, id string) bool {
 }
 
 func TestLifecycler_HealthyInstancesCount(t *testing.T) {
+	t.Parallel()
```
bboreham, line 46: Please ensure all changes in a commit are explained by the commit message.
Specifically for making tests run in parallel, I would need strong justification to do it at the same time as changing the behaviour of the code being tested.

pracucci: Was just to speed up tests. Not related to these changes (a part from that I noticed while adding new tests). Anyway, I removed `t.Parallel()` and I will propose it in a separate PR to keep discussion isolated.

pracucci:
@bboreham What do you think about the logic change? Do you see any potential side effect?

pkg/ring/lifecycler.go
```diff
@@ -556,21 +559,13 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
-		// If the ingester is in the JOINING state this means it crashed due to
-		// a failed token transfer or some other reason during startup. We want
-		// to set it back to PENDING in order to start the lifecycle from the
-		// beginning.
-		if instanceDesc.State == JOINING {
-			level.Warn(log.Logger).Log("msg", "instance found in ring as JOINING, setting to PENDING",
-				"ring", i.RingName)
+		// The instance may already be in the ring with one of the following states:
+		// - JOINING: this means it crashed due to a failed token transfer or some other reason during startup.
+		// - LEAVING: this means it crashed while shutting down and unregister from ring on shutdown is disabled.
```
pstibrany, line 564: Should be "crashed OR unregister is disabled".

pstibrany (approved):
LGTM

pracucci:
I've tested with ring based on memberlist and I can't get the ingesters to start because of this:
```
level=error ts=2021-02-15T16:33:16.109587214Z caller=cortex.go:412 msg="module failed" module=ingester-service err="ingester subservice failed: service REDACTED failed: failed to join the ring ingester: failed to CAS-update key ring: no change detected"
```

I need to fix it.

pkg/ring/lifecycler.go
```diff
@@ -562,21 +565,13 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
-		// to set it back to PENDING in order to start the lifecycle from the
-		// beginning.
-		if instanceDesc.State == JOINING {
-			level.Warn(log.Logger).Log("msg", "instance found in ring as JOINING, setting to PENDING",
-				"ring", i.RingName)
+		// The instance may already be in the ring with one of the following states:
+		// - JOINING: this means it crashed due to a failed token transfer or some other reason during startup.
+		// - LEAVING: this means it crashed while shutting down or unregister from ring on shutdown is disabled.
+		// In both cases, we want to set it back to PENDING in order to start the lifecycle from the beginning.
+		if instanceDesc.State == JOINING || instanceDesc.State == LEAVING {
```
bboreham, line 572: In [this comment](https://github.com/cortexproject/cortex/pull/1921/files#r359214524) @pstibrany opines that PENDING is wrong because another leaving ingester may start to transfer tokens to this ingester.

I think I agree.  Can you clarify why, if we have a valid set of tokens, we would want to wait for a bit and then go active?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [CHANGE] Ingester: don't update internal "last updated" timestamp of TSDB if tenant only sends invalid samples. This affects how "idle" time is computed. #3727
 * [CHANGE] Require explicit flag `-<prefix>.tls-enabled` to enable TLS in GRPC clients. Previously it was enough to specify a TLS flag to enable TLS validation. #3156
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
+* [CHANGE] Ingester: at startup, if the ingester is already registered in the ring with the `LEAVING` state, it will be switched back to `PENDING` keeping the previous tokens and then follow the normal lifecycler workflow. #3774
```
pstibrany, line 8: Cortex release 1.8.0 is now in progress. Could you please rebase master and move the CHANGELOG entry under the master / unreleased section?

pracucci:
_Still valid, need to find the time to get back to this._

PR #3775, opened 2021-02-02
Upgrade Thanos to fix panic in store-gateway when index-header lazy loading is enabled

pracucci:
**What this PR does**:
I've fixed a race condition in Thanos https://github.com/thanos-io/thanos/pull/3760 which was leading to a panic when index-header lazy loading is enabled.

In this PR I'm upgrading Thanos to bring in the fix, as well as mapping the new config options supported by Thanos' S3 client. The default values of new S3 config options are the same values used before (previously hard-coded defaults in S3 client and Golang HTTP transport).

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany (approved):
LGTM

jtlisi (approved):
LGTM

PR #3778, opened 2021-02-02
fix: ensure global loggers are set

jtlisi:
**What this PR does**:

We have two global loggers in Cortex. However, only one is being set. This is causing `CheckFatal` error messages in the main.go file to not be logged.

A more permanent fix is to consolidate on a single global logger or remove the use of a global logger entirely. This PR ensures they are both set to the same underlying logger as a quick fix to this issue.

**Which issue(s) this PR fixes**:

**Before PR:**

```sh
$ cortex git:(fix_logger) ✗ ./cmd/cortex/cortex
level=info ts=2021-02-02T23:31:35.982989857Z caller=main.go:188 msg="Starting Cortex" version="(version=1.6.0, branch=fix_logger, revision=972bb3b1e)"
```

**After PR:**

```sh
$ cortex git:(fix_logger) ✗ ./cmd/cortex/cortex
ts=2021-02-02T23:26:24.904426209Z caller=log.go:49 test=test
level=info ts=2021-02-02T23:26:24.905421014Z caller=main.go:188 msg="Starting Cortex" version="(version=1.6.0, branch=fix_logger, revision=1ee520459)"
level=info ts=2021-02-02T23:26:24.905691041Z caller=modules.go:613 msg="RulerStorage is not configured in single binary mode and will not be started."
level=error ts=2021-02-02T23:26:24.905821125Z caller=log.go:27 msg="error running cortex" err="listen tcp :80: bind: permission denied\nerror initialising module: server\ngithub.com/cortexproject/cortex/pkg/util/modules.(*Manager).initModule\n\t/go/src/github.com/cortexproject/cortex/pkg/util/modules/modules.go:105\ngithub.com/cortexproject/cortex/pkg/util/modules.(*Manager).InitModuleServices\n\t/go/src/github.com/cortexproject/cortex/pkg/util/modules/modules.go:75\ngithub.com/cortexproject/cortex/pkg/cortex.(*Cortex).Run\n\t/go/src/github.com/cortexproject/cortex/pkg/cortex/cortex.go:366\nmain.main\n\t/go/src/github.com/cortexproject/cortex/cmd/cortex/main.go:190\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1373"
```

**Checklist**
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3781, opened 2021-02-03
Remove util.Logger and move util.InitLogger to util/log package.

pstibrany:
**What this PR does**: In the light of #3778 it's clear that keeping `util.Logger` for backwards compatibility was a mistake, and it will be better to remove it completely. This doesn't fix the problem with global logger, but at least we will be back to single one.

**Checklist**
- [na] Tests updated
- [na] Documentation added


pkg/ring/kv/memberlist/memberlist_logger.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // loggerAdapter wraps a Logger and allows it to be passed to the stdlib
 // logger's SetOutput. It understand and parses output produced by memberlist
-// library (esp. level). Timestamp from memberlist can be ignored (eg. util.Logger
+// library (esp. level). Timestamp from memberlist can be ignored (eg. util_log.Logger
```
pracucci, line 14:
```suggestion
// library (esp. level). Timestamp from memberlist can be ignored (eg. pkg/util/log.Logger
```

pracucci (approved):
LGTM (modulo a nit). Thanks!

PR #3782, opened 2021-02-03
Cleanup tenant data in ingester

pstibrany:
**What this PR does**: This PR does additional cleanup of in-memory data when tenant's TSDB is closed in ingester. Specifically:
- We now remove tenant's metadata still in memory, and update metadata-related metrics
- We remove all validation metrics for removed tenant.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/metrics.go
+++ b/pkg/ingester/metrics.go
@@ -225,7 +225,7 @@ func (m *ingesterMetrics) deletePerUserMetrics(userID string) {
 		m.memSeriesCreatedTotal.DeleteLabelValues(userID)
 	}
 
-	if m.memMetadataRemovedTotal != nil {
+	if m.memSeriesRemovedTotal != nil {
 		m.memSeriesRemovedTotal.DeleteLabelValues(userID)
 	}
 }
--- a/pkg/ingester/metrics.go
+++ b/pkg/ingester/metrics.go
@@ -223,6 +223,9 @@ func (m *ingesterMetrics) deletePerUserMetrics(userID string) {
 
 	if m.memSeriesCreatedTotal != nil {
 		m.memSeriesCreatedTotal.DeleteLabelValues(userID)
+	}
+
+	if m.memMetadataRemovedTotal != nil {
 		m.memSeriesRemovedTotal.DeleteLabelValues(userID)
 	}
 }
--- a/pkg/ingester/metrics.go
+++ b/pkg/ingester/metrics.go
@@ -216,6 +216,17 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
 	return m
 }
 
+func (m *ingesterMetrics) deletePerUserMetrics(userID string) {
+	m.memMetadataCreatedTotal.DeleteLabelValues(userID)
+	m.memMetadataRemovedTotal.DeleteLabelValues(userID)
+	m.activeSeriesPerUser.DeleteLabelValues(userID)
+
+	if m.memSeriesCreatedTotal != nil {
+		m.memSeriesCreatedTotal.DeleteLabelValues(userID)
+		m.memSeriesRemovedTotal.DeleteLabelValues(userID)
+	}
+}
+
 // TSDB metrics collector. Each tenant has its own registry, that TSDB code uses.
 type tsdbMetrics struct {
 	// Metrics aggregated from Thanos shipper.
--- a/pkg/util/metrics_helper.go
+++ b/pkg/util/metrics_helper.go
@@ -11,6 +11,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/prometheus/pkg/labels"
+	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
 
 	util_log "github.com/cortexproject/cortex/pkg/util/log"
 )
@@ -699,4 +700,44 @@ func GetSumOfHistogramSampleCount(families []*dto.MetricFamily, metricName strin
 	}
 
 	return sum
+}
+
+func GetLabels(c prometheus.Collector, filter map[string]string) ([]labels.Labels, error) {
+	ch := make(chan prometheus.Metric, 128)
+
+	go func() {
+		defer close(ch)
+		c.Collect(ch)
+	}()
+
+	errs := tsdb_errors.NewMulti()
+
+	var result []labels.Labels
+
+	dtoMetric := &dto.Metric{}
+nextMetric:
+	for m := range ch {
+		err := m.Write(dtoMetric)
+		if err != nil {
+			errs.Add(err)
+			// We cannot return here, to avoid blocking goroutine calling c.Collect()
+			continue
+		}
+
+		lbls := labels.NewBuilder(nil)
+		for _, lp := range dtoMetric.Label {
+			n := lp.GetName()
+			v := lp.GetValue()
+
+			filterValue, ok := filter[n]
+			if ok && filterValue != v {
+				continue nextMetric
+			}
+
+			lbls.Set(lp.GetName(), lp.GetValue())
+		}
+		result = append(result, lbls.Labels())
+	}
+
+	return result, errs.Err()
 }
--- a/pkg/util/metrics_helper.go
+++ b/pkg/util/metrics_helper.go
@@ -702,6 +702,8 @@ func GetSumOfHistogramSampleCount(families []*dto.MetricFamily, metricName strin
 	return sum
 }
 
+// GetLables returns list of label combinations used by this collector at the time of call.
+// This can be used to find and delete unused metrics.
 func GetLabels(c prometheus.Collector, filter map[string]string) ([]labels.Labels, error) {
 	ch := make(chan prometheus.Metric, 128)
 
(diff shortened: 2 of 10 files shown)


pkg/ingester/metrics.go
```diff
@@ -216,6 +216,17 @@ func newIngesterMetrics(r prometheus.Registerer, createMetricsConflictingWithTSD
...
 }
 
+func (m *ingesterMetrics) deletePerUserMetrics(userID string) {
+	m.memMetadataCreatedTotal.DeleteLabelValues(userID)
+	m.memMetadataRemovedTotal.DeleteLabelValues(userID)
+	m.activeSeriesPerUser.DeleteLabelValues(userID)
+
+	if m.memSeriesCreatedTotal != nil {
+		m.memSeriesCreatedTotal.DeleteLabelValues(userID)
+		m.memSeriesRemovedTotal.DeleteLabelValues(userID)
```
pracucci, line 226: I know the current code is correct, but just for paranoia (against possible future changes) and clarity I would add an extra `if m.memSeriesRemovedTotal != nil`

pkg/util/validation/validate.go
```diff
@@ -61,6 +61,29 @@ const (
 	TooManyHAClusters = "too_many_ha_clusters"
 )
 
+// allDiscardedSampleReasons is used when cleaning up metrics for a user.
+var allDiscardedSampleReasons = []string{
```
pracucci, line 65: We also have reasons defined in the ingester (sampleOutOfBounds, sampleOutOfOrder, newValueForTimestamp). The ingester also does this `validation.DiscardedSamples.WithLabelValues(ve.errorType, userID).Inc()` which we need to check all possible values there.

pkg/util/validation/validate.go
```diff
@@ -61,6 +61,29 @@ const (
...
+	invalidLabel,
+	labelNameTooLong,
+	duplicateLabelNames,
+	labelsNotSorted,
+	labelValueTooLong,
+	RateLimited,
+	TooManyHAClusters,
+}
+
+var allDiscardedMetadataReasons = []string{
```
pracucci, line 80: In the distributor could also be RateLimited, perUserMetadataLimit, perMetricMetadataLimit.

pkg/util/validation/validate.go
```diff
@@ -238,3 +261,13 @@ func formatLabelSet(ls []client.LabelAdapter) string {
 
 	return fmt.Sprintf("%s{%s}", metricName, strings.Join(labelStrings, ", "))
 }
+
+func DeletePerUserValidationMetrics(userID string) {
```
pracucci, line 265: Unfortunately this is a bit fragile, but I don't have smart ideas. The problem is that we rely on `allDiscardedSampleReasons` and `allDiscardedMetadataReasons` but they will de-sync as soon as we'll add another reason (we'll forget to add it to the list of reasons, especially if done in another package).

What if we discuss with Beorn the option to add a delete function which takes in input only some labels and delete all metrics matching it?

pkg/util/validation/validate.go
pstibrany: Good point, thanks!

pkg/util/validation/validate.go
pstibrany: > What if we discuss with Beorn the option to add a delete function which takes in input only some labels and delete all metrics matching it?

This would also help us to clean `cortex_distributor_deduped_samples_total` in distributor, which we cannot currently remove as we don't keep track of values for `cluster` label.

beorn7: Thanks for bringing this to my attention. I don't remember anyone has brought this up before. Which is in hindsight a bit surprising, given that there had been quite a few discussions about partial label specification (which lead to currying). I guess deletion just doesn't happen that often.

I collected my thoughts in https://github.com/prometheus/client_golang/issues/834 (which is at the same time the formal feature request for what you want).

pracucci: Given we're not expected to delete series too frequently, using `Collect()` to find out metric labels may be a viable option for us. WDYT @pstibrany ?

pstibrany: Yes, for this case, I think it's a good workaround. If there is different solution in the future, we can use that.

pstibrany: Implemented collect-based approach to find labels and deleting all user-specific labels.

pkg/util/metrics_helper.go
```diff
@@ -700,3 +701,43 @@ func GetSumOfHistogramSampleCount(families []*dto.MetricFamily, metricName strin
 
 	return sum
 }
+
+func GetLabels(c prometheus.Collector, filter map[string]string) ([]labels.Labels, error) {
```
pracucci, line 705: Please add a comment to explain what it does.

pkg/util/validation/validate.go
```diff
@@ -238,3 +241,27 @@ func formatLabelSet(ls []client.LabelAdapter) string {
...
```
pracucci, line 245: [nit, non-blocking] Given we're moving away from the global logger, what do you think taking the logger in input?

pracucci (approved):
You did great! Good job 👏

pkg/util/metrics_helper.go
```diff
@@ -700,3 +701,44 @@ func GetSumOfHistogramSampleCount(families []*dto.MetricFamily, metricName strin
...
+nextMetric:
+	for m := range ch {
+		err := m.Write(dtoMetric)
+		if err != nil {
+			errs.Add(err)
+			// We cannot return here, to avoid blocking goroutine calling c.Collect()
+			continue
+		}
+
+		lbls := labels.NewBuilder(nil)
```
nicolai86, line 728: what about re-using the same `*labels.Builder` here and calling `lbls.Reset(nil)` at the end of the loop, like this: 
```
	lbls := labels.NewBuilder(nil)
nextMetric:
	for m := range ch {
		// ...
		result = append(result, lbls.Labels())
		lbls.Reset(nil)
	}
```

This improves the overall runtime by `GetLabels` by ~24% for small label sets and ~50% for large label sets because it avoids allocating in a loop.

Compare these two benchmarks:

A
```
func BenchmarkGetLabels_SmallSet(b *testing.B) {
	m := prometheus.NewCounterVec(prometheus.CounterOpts{
		Name: "test",
		ConstLabels: map[string]string{
			"cluster": "abc",
		},
	}, []string{"reason", "user"})

	m.WithLabelValues("bad", "user1").Inc()
	m.WithLabelValues("worse", "user1").Inc()
	m.WithLabelValues("worst", "user1").Inc()

	m.WithLabelValues("bad", "user2").Inc()
	m.WithLabelValues("worst", "user2").Inc()

	m.WithLabelValues("worst", "user3").Inc()

	for i := 0; i < b.N; i++ {
		GetLabels(m, map[string]string{"user": "user1", "reason": "worse"})
	}
}
```

```
BEFORE
pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_SmallSet
BenchmarkGetLabels_SmallSet-8   	  269169	      3818 ns/op
PASS

AFTER
pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_SmallSet
BenchmarkGetLabels_SmallSet-8   	  445689	      2639 ns/op
PASS
```

B

```
func BenchmarkGetLabels_LargeSet(b *testing.B) {
	m := prometheus.NewCounterVec(prometheus.CounterOpts{
		Name: "test",
		ConstLabels: map[string]string{
			"cluster": "abc",
		},
	}, []string{"reason", "user"})

	for i := 1; i <= 1000; i++ {
		m.WithLabelValues("bad", fmt.Sprintf("user%d", i)).Inc()
		m.WithLabelValues("worse", fmt.Sprintf("user%d", i)).Inc()
		m.WithLabelValues("worst", fmt.Sprintf("user%d", i)).Inc()

		if i % 2 == 0 {
			m.WithLabelValues("bad", fmt.Sprintf("user%d", i)).Inc()
			m.WithLabelValues("worst", fmt.Sprintf("user%d", i)).Inc()
		} else {
			m.WithLabelValues("worst", fmt.Sprintf("user%d", i)).Inc()
		}
	}

	for i := 0; i < b.N; i++ {
		GetLabels(m, map[string]string{"user": "user1", "reason": "worse"})
	}
}
```

```
BEFORE
pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_LargeSet
BenchmarkGetLabels_LargeSet-8   	     710	   1698894 ns/op
PASS

AFTER
pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_LargeSet
BenchmarkGetLabels_LargeSet-8   	    1431	    876509 ns/op
PASS
```

pstibrany: This is not performance-critical code, since it's called pretty rarely. But low-hanging optimization like this looks good. Would you like to send PR? (I think we need to call `Reset` at the beginning of the loop, because not all iterations get to the end.  Interestingly unit tests don't catch this problem, but it's easy to add new test case – filtering on `"reason": "worse"`.)

nicolai86: I'll post a PR for this today. Thanks!

nicolai86: posted https://github.com/cortexproject/cortex/pull/3863 which just contains the 2 line change I mentioned, plus the benchmarks to show the diff.

PR #3783, opened 2021-02-03
Allow downstream projects to customise compactor

pracucci:
**What this PR does**:
In this PR I've refactored a bit the compactor to allow downstream projects to customise the compactor, specifically the grouper, planner and compactor. This refactoring is expected to be a no-op for Cortex (unless I'm introducing any bug).

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -34,7 +34,28 @@ import (
 var (
 	errInvalidBlockRanges = "compactor block range periods should be divisible by the previous one, but %s is not divisible by %s"
 	RingOp                = ring.NewOp([]ring.IngesterState{ring.ACTIVE}, nil)
+
+	DefaultGrouperFactory = func(logger log.Logger, cfg Config, bkt objstore.Bucket, reg prometheus.Registerer, blocksMarkedForDeletion prometheus.Counter, garbageCollectedBlocks prometheus.Counter) compact.Grouper {
+		return compact.NewDefaultGrouper(
+			logger,
+			bkt,
+			false, // Do not accept malformed indexes
+			true,  // Enable vertical compaction
+			reg,
+			blocksMarkedForDeletion,
+			garbageCollectedBlocks)
+	}
 )
+
+// GrouperFactory builds and returns the grouper to be used to compact a tenant's blocks.
+type GrouperFactory func(
+	logger log.Logger,
+	cfg Config,
+	bkt objstore.Bucket,
+	reg prometheus.Registerer,
+	blocksMarkedForDeletion prometheus.Counter,
+	garbageCollectedBlocks prometheus.Counter,
+) compact.Grouper
 
 // Config holds the Compactor config.
 type Config struct {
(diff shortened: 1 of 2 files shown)


pkg/compactor/compactor.go
```diff
@@ -165,22 +193,37 @@ type Compactor struct {
 
 // NewCompactor makes a new Compactor.
 func NewCompactor(compactorCfg Config, storageCfg cortex_tsdb.BlocksStorageConfig, logger log.Logger, registerer prometheus.Registerer) (*Compactor, error) {
-	createDependencies := func(ctx context.Context) (objstore.Bucket, tsdb.Compactor, compact.Planner, error) {
-		bucketClient, err := bucket.NewClient(ctx, storageCfg.Bucket, "compactor", logger, registerer)
-		if err != nil {
-			return nil, nil, nil, errors.Wrap(err, "failed to create the bucket client")
+	createBucketClient := func(ctx context.Context) (objstore.Bucket, error) {
```
pracucci, line 196: Decoupled from `createBlocksCompactor()` because I prefer to not let downstream projects customise it. The bucket is created based on the provided `storageCfg` and it should continue to be.

pkg/compactor/compactor.go
```diff
@@ -34,8 +34,29 @@ import (
 var (
 	errInvalidBlockRanges = "compactor block range periods should be divisible by the previous one, but %s is not divisible by %s"
 	RingOp                = ring.NewOp([]ring.IngesterState{ring.ACTIVE}, nil)
+
+	DefaultGrouperFactory = func(logger log.Logger, cfg Config, bkt objstore.Bucket, reg prometheus.Registerer, blocksMarkedForDeletion prometheus.Counter, garbageCollectedBlocks prometheus.Counter) compact.Grouper {
```
jtlisi, line 38: **question(non-blocking):** Is there a circumstance that would require the `cfg` to be a parameter here? Wouldn't the caller be able to initialize and pass the GrouperFactory function with the config pre set? I'm indifferent to leaving it here since it adds flexibility at little cost. Just curious if I was missing something.

jtlisi (approved):
LGTM

The pattern used is similar to other parts of the Prom/Cortex/Thanos stack.

pkg/compactor/compactor.go
pracucci: In the work I've done in the "downstream project" it looked handy having the config passed to the factory 😉

pkg/compactor/compactor.go
```diff
@@ -124,9 +168,11 @@ type Compactor struct {
 	// If empty, no users are disabled. If not empty, users in the map are disabled (not owned by this compactor).
 	disabledUsers map[string]struct{}
 
-	// Function that creates bucket client, TSDB planner and compactor using the context.
+	// Functions that creates bucket client, grouper, planner and compactor using the context.
 	// Useful for injecting mock objects from tests.
-	createDependencies func(ctx context.Context) (objstore.Bucket, tsdb.Compactor, compact.Planner, error)
+	createBucketClient     func(ctx context.Context) (objstore.Bucket, error)
```
pstibrany: `bucketClientFactory` ?

pstibrany (approved):
Thank you!

pkg/compactor/compactor.go
pracucci: Yes, already renamed 👍

Issue #2957, opened 2020-07-30
Add exporter for the overrides


codesome:
The exporter would expose the overrides as Prometheus metrics which can be then displayed on a dashboard.

I already have it implemented. Would it make sense to have a separate repo under `cortexproject` for this exporter?

pstibrany:
Would it make sense to make it part of Cortex metrics?

pracucci:
Also related to https://github.com/cortexproject/cortex/issues/1433.

pstibrany:
> Would be a separate module in Cortex? (because we don't want it to be exposed by >1 pod)

Good point. Separate module would work in my opinion. Even if it is an extra pod, it would be tiny, and doing no work other than reloading config and publishing it via metrics. I think that's no different than having extra exporter. And in single-binary mode, it would just be part of "big" Cortex.

pracucci:
> And in single-binary mode, it would just be part of "big" Cortex.

If we run it in the `-target=all`, doesn't this introduce the high cardinality issue which prevented us from exporting these metrics directly from Cortex? Even in a small 10 nodes cluster, it will generate 10x series which I'm not sure it's what we want.

pstibrany:
> If we run it in the `-target=all`, doesn't this introduce the high cardinality issue which prevented us from exporting these metrics directly from Cortex? Even in a small 10 nodes cluster, it will generate 10x series which I'm not sure it's what we want.

You're right. I forgot that we encourage people to use multiple single-binary instances.

weeco:
What about an option where users can configure in what tenants' metrics the user is interested in? All tenants which aren't part of this list will be grouped together as "other_tenants" or so?

simonswine:
Another solution I could think of would be to expose the limits under a completely different path like `/metrics/limits` by default and then users can decide to scrape it through a config that would only scrape it once per cluster. (Like a kubernetes `role: service` discovery rule)

pracucci:
> Another solution I could think of would be to expose the limits under a completely different path like /metrics/limits by default and then users can decide to scrape it through a config that would only scrape it once per cluster.

I'm not sure it solves the problem. To conditionally expose these metrics we don't need a separate endpoint, but just a configuration option. However, if you enable it and these metrics are exposed by every single Cortex service then in a large cluster you will have very high cardinality metrics. To solve this, a solution would be creating a new target (which is one of the proposed solution).

jtlisi:
_still valid_

[...]

PR #3785, opened 2021-02-03
feat: Emit per-tenant limit overrides as metrics

### LINKED ISSUE #1433: Add metrics to properly monitor validation limits
**Use case**

When running Cortex for multiple tenants (e. g. multiple departments within a larger company) one still wants to set reasonable validation limits to ensure a healthy Cortex cluster.

**The problem**

Due to the fact that all limits multiply with the number of distributor/ingester replicas it's not easy to understand the current validation limits. Additionally overriding the limits for specific users makes it more complex.

It would be very nice if I could monitor all validation limits (and their usage) per user, so that I know which user is close to it's validation limits. This way I can reach out to users ask them if they need higher limits, before they run into rejections which would cause lost metrics.

**Proposed solution**

We could add prometheus metricses for all limits which we can easily sum in a Grafana dashboard. This way we know the exact validation limits (e. g. total ingestion rate and burst size per user). This might be a problem for those who run Cortex for hundreds of users I assume as each user means a new metric series.

56quarters:
**What this PR does**:

Add a new target (not enabled as part of "all") that can be used to
expose per-tenant limit overrides as Prometheus metrics. These metrics
are updated as runtime configuration is periodically reloaded.

Exposing these metrics enables use cases like proactively alerting when
tenants start to approach their limits, instead of letting them hit their
limits and start to lose data.

Fixes #1433

See also #2957

**Checklist**
- [x] Tests updated
- [x] Documentation added


docs/guides/overrides-exporter.md
```diff
@@ -0,0 +1,67 @@
+---
```
56quarters, line 1: Wasn't really sure of where this should live or what docs are expected to look like. The exporter seemed big enough that it should probably have _some_ documentation

pkg/util/validation/exporter.go
```diff
@@ -0,0 +1,85 @@
...
+
+// NewOverridesExporter creates an OverridesExporter that reads updates to per-tenant
+// limits from the provided channel. A single instance of the tenant to limit mapping
+// is kept in memory at a time by this exporter.
+func NewOverridesExporter(overridesCh <-chan map[string]*Limits) *OverridesExporter {
+	exporter := &OverridesExporter{
+		overridesCh:  overridesCh,
+		overridesMtx: sync.RWMutex{},
+
+		// The name and labels for this metric are picked for compatibility with an existing tool
```
56quarters, line 33: Totally fine to change the name of this metric or the labels included. Like the comment mentions, this was picked for compatibility with the existing grafana/cortex-tools exporter

jtlisi (approved):
LGTM

docs/guides/overrides-exporter.md
jtlisi: This doc looks solid. Good work

pkg/util/validation/exporter.go
56quarters: Will do :+1:

pkg/cortex/modules.go
```diff
@@ -170,6 +171,23 @@ func (t *Cortex) initOverrides() (serv services.Service, err error) {
 	return nil, err
 }
 
+func (t *Cortex) initOverridesExporter() (services.Service, error) {
+	cfgChan := tenantLimitsRuntimeConfigChannel(t.RuntimeConfig)
+	if cfgChan == nil {
```
pstibrany, line 176:
```suggestion
	if t.Cfg.isModuleEnabled(OverridesExporter) && cfgChan == nil {
```

This checks if overrides exporter has been explicitly enabled.

pkg/cortex/modules.go
```diff
@@ -170,6 +171,23 @@ func (t *Cortex) initOverrides() (serv services.Service, err error) {
...
+		// This target isn't enabled by default ("all") and requires runtime configuration
+		// to work. Fail if it can't be setup correctly since the user explicitly wanted this
+		// target to run.
+		return nil, errors.New("runtime configuration not available for overrides exporter")
```
pstibrany, line 180: I'd suggest to clarify error message: "overrides-exporter has been enabled, but no runtime configuration file was configured"

pkg/util/validation/exporter.go
```diff
@@ -0,0 +1,85 @@
...
+	overridesMtx sync.RWMutex
+	overrides    map[string]*Limits
+
+	description *prometheus.Desc
+}
+
+// NewOverridesExporter creates an OverridesExporter that reads updates to per-tenant
+// limits from the provided channel. A single instance of the tenant to limit mapping
+// is kept in memory at a time by this exporter.
+func NewOverridesExporter(overridesCh <-chan map[string]*Limits) *OverridesExporter {
```
pstibrany, line 28: Using function to access current limits map would simplify the design, and not require `OverridesExporter` to be a service. Have you considered that?

We could then reuse this function in overrides, and reduce amount of code in `runtime_config.go` -- single function will do just fine.

pkg/util/validation/exporter.go
```diff
@@ -0,0 +1,85 @@
...
+		overridesCh:  overridesCh,
+		overridesMtx: sync.RWMutex{},
+
+		// The name and labels for this metric are picked for compatibility with an existing tool
+		// that serves the same purpose (though it lives outside of Cortex):
+		// https://github.com/grafana/cortex-tools/blob/98415629764544b96c7725a0282ff4240db5eb54/pkg/commands/overrides_exporter.go
+		description: prometheus.NewDesc(
+			"cortex_overrides",
+			"Resource limit overrides applied to tenants",
+			[]string{"limit_type", "type", "user"},
```
pstibrany, line 39: `limit_type` -> `limit_name`, and drop `type` entirely, since it's constant.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
   - `-<prefix>.s3.sse.kms-key-id`
   - `-<prefix>.s3.sse.kms-encryption-context`
 * [FEATURE] Querier: Enable `@ <timestamp>` modifier in PromQL using the new `-querier.at-modifier-enabled` flag. #3744
+* [FEATURE] Overrides Exporter: Add `-target overrides-exporter` target for exposing per-tenant resource limit overrides as metrics. #3785
```
pstibrany, line 12:
```suggestion
* [FEATURE] Overrides Exporter: Add `overrides-exporter` module for exposing per-tenant resource limit overrides as metrics. It is not included in `all` target, and must be explicitely enabled. #3785
```

pstibrany (commented):
Nice work. I'd suggest to simplify the design of `OverridesExporter`, since it doesn't need to react on changing values immediately.

pkg/util/validation/exporter.go
56quarters: :+1: I'll see what I can do

pkg/cortex/runtime_config.go
```diff
@@ -45,17 +45,38 @@ func loadRuntimeConfig(r io.Reader) (interface{}, error) {
...
+		if tenantLimits == nil {
 			return nil
 		}
 
-		return cfg.TenantLimits[userID]
+		return tenantLimits[userID]
+	}
+}
+
+// tenantLimitsRuntimeConfigFunc returns a function that returns a mapping of all
```
56quarters, line 66: Using a function here like you mentioned and the code definitely seems simpler. Not sure if this is exactly what you had in mind @pstibrany

pkg/cortex/runtime_config.go
```diff
@@ -45,17 +45,38 @@ func loadRuntimeConfig(r io.Reader) (interface{}, error) {
...
 	if c == nil {
 		return nil
 	}
+
+	supplier := tenantLimitsRuntimeConfigFunc(c)
 	return func(userID string) *validation.Limits {
-		cfg, ok := c.GetConfig().(*runtimeConfigValues)
-		if !ok || cfg == nil {
+		tenantLimits := supplier()
+		if tenantLimits == nil {
```
pstibrany, line 58: Nit: No need for this check... it's actually fine if `tenantLimits` is `nil` here, `tenantLimits[userID]` will then return `nil` too (also fine).

pkg/util/validation/exporter.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+func (oe *OverridesExporter) Describe(ch chan<- *prometheus.Desc) {
+	ch <- oe.description
+}
+
+func (oe *OverridesExporter) Collect(ch chan<- prometheus.Metric) {
+	allLimits := oe.limitSupplier()
+	if allLimits == nil {
```
pstibrany, line 33: Nit: also no need for this check. Range over nil map is noop.

pkg/util/validation/exporter.go
```diff
@@ -0,0 +1,48 @@
...
+		return
+	}
+
+	for tenant, limits := range allLimits {
```
pstibrany, line 37: Plenty more limits we could return here. Are those not interesting?

docs/guides/overrides-exporter.md
```diff
@@ -0,0 +1,67 @@
...
+cortex -target overrides-exporter -runtime-config.file runtime.yaml -server.http-listen-port=8080
+```
+
+After the `overrides-exporter` starts, you can to use `curl` to inspect the tenant overrides.
+
+```text
+curl -s http://localhost:8080/metrics | grep cortex_overrides
+# HELP cortex_overrides Resource limit overrides applied to tenants
+# TYPE cortex_overrides gauge
+cortex_overrides{limit_type="ingestion_burst_size",type="tenant",user="user1"} 350000
```
pstibrany, line 55: `type="tenant"` is no longer included.

docs/guides/overrides-exporter.md
```diff
@@ -0,0 +1,67 @@
...
+Since Cortex is a multi-tenant system, it supports applying limits to each tenant to prevent
+any single one from using too many resources. In order to help operators understand how close
+to their limits tenants are, the `overrides-exporter` module can expose limits as Prometheus metrics.
+
+## Context
+
+To update configuration without restarting, Cortex allows operators to supply a `runtime_config`
+file that will be periodically reloaded. This file can be specified under the `runtime_config` section
+of the main [configuration file](../configuration/arguments.md#runtime-configuration-file) or using the `-runtime-config.file`
+command line flag. This file is used apply tenant-specific limits.
```
pstibrany, line 17: ... used **to** apply ...

pstibrany (approved):
LGTM, thank you!

pkg/util/validation/exporter.go
56quarters: The ones here are what the`cortex-tools` exporter was making available and seem to be ingester focused. I figure we'd start with these since these seem to be the limits that Grafana Cloud users bump into the most. I can add others now if you have some in mind

pracucci (approved):
You did great! 👏

PR #3792, opened 2021-02-04
Clarify DNS service discovery

### LINKED ISSUE #3788: Support for IP addresses in memcached_client_config
The current `memcached_client_config` configuration only accepts DNS names as a `host` or `addresses` value. It would be great if we can specify IP addresses too.

https://cortexmetrics.io/docs/configuration/configuration-file/#memcached_client_config

pracucci:
**What this PR does**:
The issue #3788 may me realised that's not clear that the DNS service discovery also support "no prefix", which means the provided IP or hostname is used as is. This PR tries to clarify the doc.

Fixes #3788

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pstibrany (approved):
👍

Issue #3362, opened 2020-10-19
Support DynamoDB in BlocksConvert Scanner


andersosthus:
Adding support for DynamoDB as an Index source in Blocksconvert Scanner.

I've started working on making the Blocksconvert Scanner work with `DynamoDB` over in my branch (https://github.com/andersosthus/cortex/tree/dynamodb-scanner).

The scanner works (seemingly) now, but there are some issues remaining:
1. The code is basically a duplication of the `BigTable` code adjusted for DynamoDB without much consideration if it's appropriate for the `DynamoDB` part (I removed some stuff that seemed to be `BigTable` specific)
1. As far as I know, `DynamoDB` doesn't have that "sample keys" to get out how many threads one should run, so I opted for a `Scan` for now (fixed to 1 segment since I'm not sure how best to decide how many segments we should run)
1. Due to the use of `Scan`, a series might end up on several `IndexEntryProcessor`, not sure how to handle that, but the quick workaround is to just run with 1 `IndexEntryProcessor` :) With more understanding about how the series names are created, I guess we could maybe add some filters to the `Scans`, and assign a scan per `IndexEntryProcessor`?

But I have produced plan files that validate with the code, so it's kinda working at least.

I want to finish up this, and was wondering if anyone have some pointers about the issues I've raised above?

pracucci:
This is a topic for @pstibrany ;)

pstibrany:
Anders has indicated on the Slack that he has PR in progress that will be posted soon. In the meantime I haven't looked more into these DynamoDB-specific questions, as I'm not familiar with DynamoDB at all, but plan to review the PR once it's ready.

andersosthus:
I did some tests with my code, but it seemed to never finish (8 workers on an 8 cpu machine. Ran for about 48 hours before i killed it). I have been busy with other things since then, but I do intend do look into it again.

All my DynamoDB tables are pretty big, so maybe it will be easier to test with the new date range functionality in 1.5.0 ?

andersosthus:
Yeah, I did use `-scanner.tables` to limit it to the last table, but it has 160m records or something like that, so was still slow. But I'll see if I can find a smaller table, and also run with `-scanner.scan-period-start/end` and see if I can get the size manageable :)

pracucci:
_Reopening because should be addressed by https://github.com/cortexproject/cortex/pull/3828 (at least for schema v9)._

PR #3795, opened 2021-02-05
Add Cassandra support to blocksconvert

ubcharron:
This is early work on a Cassandra scanner for blocksconvert. I'm looking for comments to help me shape this commit. Among other things, I didn't want to duplicate the code that sets up the Cassandra connections, but wasn't sure how to go about re-using it, hence the `GetReadSession()` :(

There are no tests as both functions depend on Cassandra, and I wasn't sure if I should create a Cas mock interface.

I copied the BigTable scanning logic, but it doesn't react well when Cassandra times out during scanning, and I'm unsure how to handle that case.

Please have a look and let me know what sucks about it this code.

Signed-off-by: Benjamin Charron <benjamin.charron@ubisoft.com>

**What this PR does**:
Adds support for scanning Cassandra indexes in tools/blocksconvert

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tools/blocksconvert/scanner/cassandra_index_reader.go
+++ b/tools/blocksconvert/scanner/cassandra_index_reader.go
@@ -1,0 +1,172 @@
+package scanner
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/chunk/cassandra"
+)
+
+/* Cassandra can easily run out of memory or timeout if we try to SELECT the
+ * entire table. Splitting into many smaller chunks help a lot. */
+const nbTokenRanges = 512
+const queryPageSize = 10000
+
+type cassandraIndexReader struct {
+	log                    log.Logger
+	cassandraStorageConfig cassandra.Config
+	schemaCfg              chunk.SchemaConfig
+
+	rowsRead                  prometheus.Counter
+	parsedIndexEntries        prometheus.Counter
+	currentTableRanges        prometheus.Gauge
+	currentTableScannedRanges prometheus.Gauge
+}
+
+func newCassandraIndexReader(cfg cassandra.Config, schemaCfg chunk.SchemaConfig, l log.Logger, rowsRead prometheus.Counter, parsedIndexEntries prometheus.Counter, currentTableRanges, scannedRanges prometheus.Gauge) *cassandraIndexReader {
+	return &cassandraIndexReader{
+		log:                    l,
+		cassandraStorageConfig: cfg,
+
+		rowsRead:                  rowsRead,
+		parsedIndexEntries:        parsedIndexEntries,
+		currentTableRanges:        currentTableRanges,
+		currentTableScannedRanges: scannedRanges,
+	}
+}
+
+func (r *cassandraIndexReader) IndexTableNames(ctx context.Context) ([]string, error) {
+	client, err := cassandra.NewTableClient(ctx, r.cassandraStorageConfig, nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "create cassandra client failed")
+	}
+
+	defer client.Stop()
+
+	return client.ListTables(ctx)
+}
+
+type tokenRange struct {
+	start int64
+	end   int64
+}
+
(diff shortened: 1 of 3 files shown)


pstibrany:
Thanks for your PR. I'm not very familiar with Cassandra, but I think your approach is correct.

To handle possible timeouts, I'd suggest to first buffer all index entries for given hash, and only once next hash is found, pass the buffered entries to the processor (one by one). This assumes that all entries for one hash are returned before entries for another hash – I'm not sure from the code if that's the case or not, but I believe that is requirement for `IndexEntryProcessor` to work correctly.

Comment for `IndexEntryProcessor` says that ALL entries for Hash + Range pair must all arrive first, but I'm not sure if that's correct – there can only be single entry for given Hash and Range. I think processor requires all entries for given Hash instead to arrive in the sequence. Reason is that `IndexEntryProcessor` is interested in finding all chunks for given series, and series info is encoded in Hash, while chunk info is encoded in Range.

I hope this helps.

tools/blocksconvert/scanner/cassandra_index_reader.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/cortexproject/cortex/pkg/chunk"
+	"github.com/cortexproject/cortex/pkg/chunk/cassandra"
+)
+
+const nbTokenRanges = 512
```
bboreham, line 19: Why is this a fixed number, rather than being related to the number of processors?

tools/blocksconvert/scanner/cassandra_index_reader.go
```diff
@@ -0,0 +1,157 @@
...
+	for ix := range processors {
+		p := processors[ix]
+
+		g.Go(func() error {
+			for rng := range rangesCh {
+				level.Debug(r.log).Log("msg", "reading rows", "range_start", rng.start, "range_end", rng.end, "table_name", tableName)
+
+				var query string
+
+				query = fmt.Sprintf("SELECT hash, range, value FROM %s WHERE token(hash) >= %v", tableName, rng.start)
```
bboreham, line 109: I don't think the value is used for series index entries.  Small optimisation.

bboreham (commented):
I don't know Cassandra either, but it seems plausible.

Re timeouts, I guess you should re-try the same query a few times?

I can't see how any buffering would be necessary; as long as all values for the same hash are delivered to one processor before moving to another hash that's enough.

I did a DynamoDB port at #3362

tools/blocksconvert/scanner/cassandra_index_reader.go
ubcharron: Thank you, added some comments for the mystery value. Cassandra doesn't like having the entire table scanned, so I split it up in many chunks.

tools/blocksconvert/scanner/cassandra_index_reader.go
```diff
@@ -0,0 +1,157 @@
...
+	r.currentTableRanges.Set(float64(len(rangesCh)))
+	r.currentTableScannedRanges.Set(0)
+
+	defer r.currentTableRanges.Set(0)
+	defer r.currentTableScannedRanges.Set(0)
+
+	g, gctx := errgroup.WithContext(ctx)
+
+	for ix := range processors {
+		p := processors[ix]
```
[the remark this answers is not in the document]
ubcharron: I copied it from the bigtable reader, and it crashes with a nil pointer deference if I change to `for _, p := range processors`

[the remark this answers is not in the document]
bboreham: Aha, you probably tripped over this problem: https://golang.org/doc/faq#closures_and_goroutines
to which your solution is ok, alternatively `p := p` after my suggested change.
Either way a comment pointing to the FAQ would help readers recall why such gyrations are necessary.

bboreham (approved):
Allowing for the fact that I don't know Cassandra, I think we could merge this.

Thanks for the PR!

tools/blocksconvert/scanner/cassandra_index_reader.go
pstibrany: > Cassandra doesn't like having the entire table scanned, so I split it up in many chunks.

Interesting. I've recently written code scanning entire Cassandra table with hundreds of millions of entries, and haven't run into any issues. Select was very simple: `SELECT <columns> FROM <table>`, with no other conditions. It took a while to process all the results (as expected), but it worked.

pstibrany (approved):
Thank you!

PR #3798, opened 2021-02-08
Moved common protobuf code to new cortexpb package and introduced `distributor.Distributor` service name.

pstibrany:
**What this PR does**: Goal of this PR is to rename `cortex.PushOnlyIngester` gRPC service to `distributorpb.Distributor`, however to achieve that, this PR also does the following:
- introduce new `cortexpb` package, with common messages moved from previous "pkg/ingester/client" (`cortex` in protobuf) package.
- move of types also required some move of code from "pkg/ingester/client" to "pkg/cortexpb" package. Previous types and function names were preserved, but marked as deprecated, and will be cleaned up in subsequent release.
- `cortex` protobuf package was preserved, but will be renamed to `ingesterpb` in the future (while keeping backwards compatibility for registered service for smooth migration between ingesters)

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/ingester/client/compat.go
```diff
@@ -1,42 +1,15 @@
...
-	for i, s := range samples {
-		ts := timeSeriesPool.Get().(*TimeSeries)
-		ts.Labels = append(ts.Labels, FromLabelsToLabelAdapters(lbls[i])...)
-		ts.Samples = append(ts.Samples, s)
-		req.Timeseries = append(req.Timeseries, PreallocTimeseries{TimeSeries: ts})
-	}
-
-	return req
-}
+var ToWriteRequest = cortexpb.ToWriteRequest
```
pracucci, line 12: Should be marked as deprecated

pkg/cortexpb/compat_test.go
```diff
@@ -0,0 +1,51 @@
+package cortexpb
```
pracucci, line 1: We need to move here some tests from `pkg/ingester/client/compat_test.go`

pkg/cortexpb/compat.go
```diff
@@ -0,0 +1,235 @@
...
+		return textparse.MetricTypeUnknown
+	}
+}
+
+// Only set from tests to get special behaviour to verify that custom marshal and unmarshal is used.
+var isTesting = false
+
+func init() {
+
+	jsoniter.RegisterTypeEncoderFunc("cortexpb.Sample", func(ptr unsafe.Pointer, stream *jsoniter.Stream) {
```
pracucci, line 183: We've changed `client.Sample` into `cortexpb.Sample`. Is it backward compatible?

pkg/ingester/client/ingester.pb.go
```diff
@@ -56,172 +60,17 @@ var MatchType_value = map[string]int32{
...
-
 type ReadRequest struct {
 	Queries []*QueryRequest `protobuf:"bytes,1,rep,name=queries,proto3" json:"queries,omitempty"`
 }
 
 func (m *ReadRequest) Reset()      { *m = ReadRequest{} }
 func (*ReadRequest) ProtoMessage() {}
 func (*ReadRequest) Descriptor() ([]byte, []int) {
-	return fileDescriptor_893a47d0a749d749, []int{2}
+	return fileDescriptor_60f6df4f3586b478, []int{0}
```
pracucci, line 73: Do you know what's the 2nd returned parameter and if changing it is a breaking change?

pracucci (approved):
Good job! To my knowledge of protobuf this refactoring shouldn't introduce any breaking change, but please look at my comments.

pkg/cortexpb/compat.go
pstibrany: This string is used to match type name with correct encoder/decoder when using jsoniter library. I've added tests to "old" `compat` package to verify that custom marshal/unmarshal functions are used when working with `client.Sample` type alias.

pkg/ingester/client/ingester.pb.go
pstibrany: First returned value is a full description of "proto" file, which includes all types. It is itself a protobuf message, `FileDescriptorProto` and is described [here](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto). Second returned parameter is an index into `message_type` array.

Given that some messages were moved around, this proto file now has fewer messages, hence the change in returned index.

pstibrany:
Thanks for review. I've now addressed your feedback, please take a look again.

PR #3803, opened 2021-02-09
Don't truncate error on long series

gouthamve:
errors.Wrapf prints the message first and error later. If the message is
long then the client never prints the error itself. This fixes it.


pracucci (approved):
LGTM, thanks!

PR #3805, opened 2021-02-09
RuleStore: implementation using thanos objstore bucket client

jtlisi:
**What this PR does**:

- Implement a rulestore client using the Thanos objstore client

Related #2262 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
 * [CHANGE] Store-gateway: the chunks pool controlled by `-blocks-storage.bucket-store.max-chunk-pool-bytes` is now shared across all tenants. #3830
 * [CHANGE] Ingester: return error code 400 instead of 429 when per-user/per-tenant series/metadata limits are reached. #3833
+* [FEATURE] Experimental Ruler Storage: Add a separate set of configuration options to configure the ruler storage backend under the `-ruler-storage.` flag prefix. All TSDB bucket clients and the config service are currently supported. Clients using this implementation will only be enabled if the existing `-ruler.storage` flags are left unset. #3805
 * [FEATURE] Adds support to S3 server-side encryption using KMS. The S3 server-side encryption config can be overridden on a per-tenant basis. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651 #3810 #3811
   - `-<prefix>.s3.sse.type`
   - `-<prefix>.s3.sse.kms-key-id`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
 * [CHANGE] Store-gateway: the chunks pool controlled by `-blocks-storage.bucket-store.max-chunk-pool-bytes` is now shared across all tenants. #3830
 * [CHANGE] Ingester: return error code 400 instead of 429 when per-user/per-tenant series/metadata limits are reached. #3833
-* [FEATURE] Experimental Ruler Storage: Add a separate set of configuration options to configure the ruler storage backend under the `-ruler-storage.` flag prefix. All TSDB bucket clients and the config service are currently supported. Clients using this implementation will only be enabled if the existing `-ruler.storage` flags are left unset. #3805
+* [FEATURE] Experimental Ruler Storage: Add a separate set of configuration options to configure the ruler storage backend under the `-ruler-storage.` flag prefix. All blocks storage bucket clients and the config service are currently supported. Clients using this implementation will only be enabled if the existing `-ruler.storage` flags are left unset. #3805
 * [FEATURE] Adds support to S3 server-side encryption using KMS. The S3 server-side encryption config can be overridden on a per-tenant basis. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651 #3810 #3811
   - `-<prefix>.s3.sse.type`
   - `-<prefix>.s3.sse.kms-key-id`
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -116,7 +116,7 @@ type Config struct {
 	TenantFederation tenantfederation.Config         `yaml:"tenant_federation"`
 
 	Ruler          ruler.Config                               `yaml:"ruler"`
-	RulerStorage   rulestore.Config                           `yaml:"ruler_storage" doc:"hidden"`
+	RulerStorage   rulestore.Config                           `yaml:"ruler_storage"`
 	Configs        configs.Config                             `yaml:"configs"`
 	Alertmanager   alertmanager.MultitenantAlertmanagerConfig `yaml:"alertmanager"`
 	RuntimeConfig  runtimeconfig.ManagerConfig                `yaml:"runtime_config"`
--- a/pkg/cortex/cortex.go
+++ b/pkg/cortex/cortex.go
@@ -47,6 +47,7 @@ import (
 	"github.com/cortexproject/cortex/pkg/ring/kv/memberlist"
 	"github.com/cortexproject/cortex/pkg/ruler"
 	"github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/ruler/rulestore"
 	"github.com/cortexproject/cortex/pkg/scheduler"
 	"github.com/cortexproject/cortex/pkg/storage/tsdb"
 	"github.com/cortexproject/cortex/pkg/storegateway"
@@ -115,6 +116,7 @@ type Config struct {
 	TenantFederation tenantfederation.Config         `yaml:"tenant_federation"`
 
 	Ruler          ruler.Config                               `yaml:"ruler"`
+	RulerStorage   rulestore.Config                           `yaml:"ruler_storage" doc:"hidden"`
 	Configs        configs.Config                             `yaml:"configs"`
 	Alertmanager   alertmanager.MultitenantAlertmanagerConfig `yaml:"alertmanager"`
 	RuntimeConfig  runtimeconfig.ManagerConfig                `yaml:"runtime_config"`
@@ -162,6 +164,7 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	c.TenantFederation.RegisterFlags(f)
 
 	c.Ruler.RegisterFlags(f)
+	c.RulerStorage.RegisterFlags(f)
 	c.Configs.RegisterFlags(f)
 	c.Alertmanager.RegisterFlags(f)
 	c.RuntimeConfig.RegisterFlags(f)
@@ -194,6 +197,9 @@ func (c *Config) Validate(log log.Logger) error {
 	}
 	if err := c.ChunkStore.Validate(log); err != nil {
 		return errors.Wrap(err, "invalid chunk store config")
+	}
+	if err := c.RulerStorage.Validate(); err != nil {
+		return errors.Wrap(err, "invalid rulestore config")
 	}
 	if err := c.Ruler.Validate(c.LimitsConfig, log); err != nil {
 		return errors.Wrap(err, "invalid ruler config")
--- a/pkg/ruler/rulestore/bucket_client.go
+++ b/pkg/ruler/rulestore/bucket_client.go
@@ -21,25 +21,28 @@ import (
 const (
 	delim      = "/"
 	rulePrefix = "rules"
+
+	loadConcurrency = 10
 )
 
+// BucketRuleStore is used to support the RuleStore interface against an object storage backend. It is implemented
+// using the Thanos objstore.Bucket interface
 type BucketRuleStore struct {
-	bucket          objstore.Bucket
-	loadConcurrency int
-
-	logger log.Logger
+	bucket      objstore.Bucket
+	cfgProvider bucket.TenantConfigProvider
+	logger      log.Logger
 }
 
-func NewBucketRuleStore(bkt objstore.Bucket, loadConcurrency int, logger log.Logger) *BucketRuleStore {
+func NewBucketRuleStore(bkt objstore.Bucket, cfgProvider bucket.TenantConfigProvider, logger log.Logger) *BucketRuleStore {
 	return &BucketRuleStore{
-		bucket:          bucket.NewPrefixedBucketClient(bkt, rulePrefix),
-		loadConcurrency: loadConcurrency,
-		logger:          logger,
+		bucket:      bucket.NewPrefixedBucketClient(bkt, rulePrefix),
+		cfgProvider: cfgProvider,
+		logger:      logger,
 	}
 }
 
 func (b *BucketRuleStore) listNamespacesForUser(ctx context.Context, user string) ([]string, error) {
-	userBucket := bucket.NewUserBucketClient(user, b.bucket)
+	userBucket := bucket.NewUserBucketClient(user, b.bucket, b.cfgProvider)
 
 	var namespaces []string
 	err := userBucket.Iter(ctx, "", func(namespace string) error {
@@ -62,7 +65,7 @@ func (b *BucketRuleStore) listNamespacesForUser(ctx context.Context, user string
 
 // If existing rule group is supplied, it is Reset and reused. If nil, new RuleGroupDesc is allocated.
 func (b *BucketRuleStore) getRuleGroup(ctx context.Context, userID, namespace, groupName string, rg *rules.RuleGroupDesc) (*rules.RuleGroupDesc, error) {
-	userBucket := bucket.NewUserBucketClient(userID, b.bucket)
+	userBucket := bucket.NewUserBucketClient(userID, b.bucket, b.cfgProvider)
 	objectKey := generateRuleObjectKey(namespace, groupName)
 
 	reader, err := userBucket.Get(ctx, objectKey)
@@ -130,7 +133,7 @@ func (b *BucketRuleStore) ListAllRuleGroups(ctx context.Context) (map[string]rul
 // ListRuleGroupsForUserAndNamespace returns all the active rule groups for a user from given namespace.
 // If namespace is empty, groups from all namespaces are returned.
 func (b *BucketRuleStore) ListRuleGroupsForUserAndNamespace(ctx context.Context, userID string, namespace string) (rules.RuleGroupList, error) {
-	userBucket := bucket.NewUserBucketClient(userID, b.bucket)
+	userBucket := bucket.NewUserBucketClient(userID, b.bucket, b.cfgProvider)
 
 	var namespaces []string
 	var err error
@@ -179,7 +182,7 @@ func (b *BucketRuleStore) LoadRuleGroups(ctx context.Context, groupsToLoad map[s
 	// download all rule groups in parallel. We limit the number of workers to avoid a
 	// particular user having too many rule groups rate limiting us with the object storage.
 	g, gCtx := errgroup.WithContext(ctx)
-	for i := 0; i < b.loadConcurrency; i++ {
+	for i := 0; i < loadConcurrency; i++ {
 		g.Go(func() error {
 			for gr := range ch {
 				user, namespace, group := gr.GetUser(), gr.GetNamespace(), gr.GetName()
@@ -189,8 +192,7 @@ func (b *BucketRuleStore) LoadRuleGroups(ctx context.Context, groupsToLoad map[s
 
 				gr, err := b.getRuleGroup(gCtx, user, namespace, group, gr) // reuse group pointer from the map.
 				if err != nil {
-					level.Error(b.logger).Log("msg", "failed to get rule group", "namespace", namespace, "name", group, "user", user, "err", err)
-					return err
+					return fmt.Errorf("failed to get rule group: user=%q, namespace=%q, name=%q, err=%w", user, namespace, group, err)
 				}
 
 				if user != gr.User || namespace != gr.Namespace || group != gr.Name {
@@ -226,7 +228,7 @@ func (b *BucketRuleStore) GetRuleGroup(ctx context.Context, userID string, names
 }
 
 func (b *BucketRuleStore) SetRuleGroup(ctx context.Context, userID string, namespace string, group *rules.RuleGroupDesc) error {
-	userBucket := bucket.NewUserBucketClient(userID, b.bucket)
+	userBucket := bucket.NewUserBucketClient(userID, b.bucket, b.cfgProvider)
 	data, err := proto.Marshal(group)
 	if err != nil {
 		return err
@@ -236,7 +238,7 @@ func (b *BucketRuleStore) SetRuleGroup(ctx context.Context, userID string, names
 }
 
 func (b *BucketRuleStore) DeleteRuleGroup(ctx context.Context, userID string, namespace string, group string) error {
-	userBucket := bucket.NewUserBucketClient(userID, b.bucket)
+	userBucket := bucket.NewUserBucketClient(userID, b.bucket, b.cfgProvider)
 	err := userBucket.Delete(ctx, generateRuleObjectKey(namespace, group))
 	if b.bucket.IsObjNotFoundErr(err) {
 		return rules.ErrGroupNotFound
@@ -254,11 +256,11 @@ func (b *BucketRuleStore) DeleteNamespace(ctx context.Context, userID string, na
 		return rules.ErrGroupNamespaceNotFound
 	}
 
+	userBucket := bucket.NewUserBucketClient(userID, b.bucket, b.cfgProvider)
 	for _, rg := range ruleGroupList {
 		if err := ctx.Err(); err != nil {
 			return err
 		}
-		userBucket := bucket.NewUserBucketClient(userID, b.bucket)
 		objectKey := generateRuleObjectKey(rg.Namespace, rg.Name)
 		level.Debug(b.logger).Log("msg", "deleting rule group", "namespace", namespace, "key", objectKey)
 		err = userBucket.Delete(ctx, objectKey)
--- a/pkg/ruler/rulestore/bucket_client.go
+++ b/pkg/ruler/rulestore/bucket_client.go
@@ -1,0 +1,285 @@
+package rulestore
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/gogo/protobuf/proto"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/storage/bucket"
+)
+
+const (
+	delim      = "/"
+	rulePrefix = "rules"
+)
+
+type BucketRuleStore struct {
+	bucket          objstore.Bucket
+	loadConcurrency int
+
+	logger log.Logger
+}
+
+func NewBucketRuleStore(bkt objstore.Bucket, loadConcurrency int, logger log.Logger) *BucketRuleStore {
+	return &BucketRuleStore{
+		bucket:          bucket.NewPrefixedBucketClient(bkt, rulePrefix),
+		loadConcurrency: loadConcurrency,
+		logger:          logger,
+	}
+}
+
+func (b *BucketRuleStore) listNamespacesForUser(ctx context.Context, user string) ([]string, error) {
+	userBucket := bucket.NewUserBucketClient(user, b.bucket)
+
+	var namespaces []string
+	err := userBucket.Iter(ctx, "", func(namespace string) error {
+		namespace = strings.TrimSuffix(namespace, delim)
+		decodedNamespace, err := base64.URLEncoding.DecodeString(namespace)
+		if err != nil {
+			return fmt.Errorf("failed to decode namespace '%s': %w", namespace, err)
+		}
+
+		namespaces = append(namespaces, string(decodedNamespace))
+		return nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return namespaces, nil
+}
+
+// If existing rule group is supplied, it is Reset and reused. If nil, new RuleGroupDesc is allocated.
+func (b *BucketRuleStore) getRuleGroup(ctx context.Context, userID, namespace, groupName string, rg *rules.RuleGroupDesc) (*rules.RuleGroupDesc, error) {
+	userBucket := bucket.NewUserBucketClient(userID, b.bucket)
+	objectKey := generateRuleObjectKey(namespace, groupName)
+
+	reader, err := userBucket.Get(ctx, objectKey)
+	if userBucket.IsObjNotFoundErr(err) {
+		level.Debug(b.logger).Log("msg", "rule group does not exist", "name", objectKey)
+		return nil, rules.ErrGroupNotFound
+	}
+
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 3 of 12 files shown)


pkg/cortex/cortex.go
```diff
@@ -109,6 +110,7 @@ type Config struct {
 	TenantFederation tenantfederation.Config         `yaml:"tenant_federation"`
 
 	Ruler          ruler.Config                               `yaml:"ruler"`
+	RuleStore      rulestore.Config                           `yaml:"rule_store" doc:"hidden"`
```
pracucci, line 113: What do you think about naming the YAML (and CLI flags) `ruler_storage` to keep it consistent with `blocks_storage`?

pkg/storage/rulestore/config.go
```diff
@@ -0,0 +1,122 @@
...
+
+	f.StringVar(&cfg.Backend, prefix+"backend", "s3", fmt.Sprintf("Backend storage to use. Supported backends are: %s.", strings.Join(supportedBackends, ", ")))
+}
+
+func (cfg *Config) Validate() error {
+	if !util.StringsContain(supportedBackends, cfg.Backend) {
+		return ErrUnsupportedStorageBackend
+	}
+
+	if cfg.Backend == bucket.S3 {
```
pracucci, line 62: We shouldn't duplicate the bucket config `Validate()` but use it. I understanding you did it because the `Backend` validation wouldn't pass if it's configured `configdb`.

An option:
1. Create `bucket.Config.ValidateWithExtraBackends(extras []string)`
2. Pass `[]string{configdb}` to `ValidateWithExtraBackends()`
3. This function just calls `cfg.Bucket.Validate()`

pkg/storage/rulestore/config.go
```diff
@@ -0,0 +1,122 @@
...
+type Config struct {
+	bucket.Config `yaml:",inline"`
+	ConfigDB      client.Config `yaml:"configdb"`
+}
+
+// RegisterFlags registers the backend storage config.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.ConfigDB.RegisterFlagsWithPrefix(prefix, f)
+
+	cfg.S3.RegisterFlagsWithPrefix(prefix, f)
```
pracucci, line 48: You should call `cfg.Bucket.RegisterFlagsWithPrefix()` instead of duplicating here.

pkg/storage/rulestore/config.go
```diff
@@ -0,0 +1,122 @@
...
+		if err := cfg.S3.Validate(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// NewRuleStore creates a new bucket client based on the configured backend
+func NewRuleStore(ctx context.Context, cfg Config, logger log.Logger, reg prometheus.Registerer) (rules.RuleStore, error) {
```
pracucci, line 72: Why this function contains a bunch of duplicated code from `bucket.NewClient()`? We should reuse it.

Also configdb is not supported yet.

pkg/storage/rulestore/bucket_client.go
```diff
@@ -0,0 +1,311 @@
...
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/tenant"
+)
+
+const (
+	delim      = "/"
+	rulePrefix = "rules" + delim
+)
+
+type BucketClient struct {
```
pracucci, line 26: This is not a bucket client but an implementation of the rule store. I think we should name it accordingly.

pkg/storage/rulestore/bucket_client.go
```diff
@@ -0,0 +1,311 @@
...
+
+func NewBucketClient(bucket objstore.Bucket, loadConcurrency int, logger log.Logger) *BucketClient {
+	return &BucketClient{
+		bucket:          bucket,
+		loadConcurrency: loadConcurrency,
+		logger:          logger,
+	}
+}
+
+func (b *BucketClient) listNamespacesForUser(ctx context.Context, user string) ([]string, error) {
```
pracucci, line 41: I'm sorry. In all functions you will have to use `UserBucketClient` for every operation involving a specific user. We're leveraging on it to add extra functionalities and it's now required that we use it everywhere we leverage on the Thanos objstore client.

_We can discuss offline if you need more info about it._

pracucci (commented):
Thanks a lot @jtlisi for working on this! I left few high level comments, before jumping on a deeper 2nd pass review. 🙏

pkg/cortex/cortex.go
jtlisi: 👍

pkg/storage/rulestore/config.go
jtlisi: Makes sense. The only thing I would say is wouldn't we need `cfg.Bucket.Validate()` to call `cfg.Bucket. ValidateWithExtraBackends([]string{})` instead of vice versa?

pkg/storage/rulestore/config.go
jtlisi: The only reason I didn't call this function is because it wouldn't print `configdb` as an option in the help text.

Between this issue and https://github.com/cortexproject/cortex/pull/3805#discussion_r574604170 I think I found a solution that will accommodate the use case case.

1. Add `ExtraBackends []string` field to the bucket.Config struct.
2. In `rulestore.Config.RegisterFlags` we set the `bucket.Config.ExtraBackends` field to `[]string{ConfigDB}`
3. The `bucket.Config.RegisterFlagsWithPrefix` and `bucket.Config.Validate` functions are adjusted to merge the `supportedBackends` with the embedded `ExtraBackends` before validating/printing-help-text

Let me know what you think? I think this approach will avoid any additional functions at the cost of an additional hidden field in the config that must be set before the `bucket.Config.Validate` or `bucket.Config.RegisterFlags` functions are called.

pkg/storage/rulestore/config.go
jtlisi: Makes sense, I'll refactor to reuse upstream and implement the configdb.

pkg/storage/rulestore/bucket_client.go
jtlisi: How does `BucketRuleStore` work?

pkg/storage/rulestore/bucket_client.go
jtlisi: I transitioned all user operations to use a UserBucketClient. However, I had to make some additional changes to the UserBucketClient to allow for a global prefix, since all rules are currently stores on the path `rules/<user-id>`.

pkg/storage/bucket/user_bucket_client.go
```diff
@@ -22,16 +23,24 @@ type UserBucketClient struct {
...
 			userID: userID,
 			bucket: bucket,
+			prefix: prefix,
 		},
 		bucket: bucket,
 	}
 }
 
 func (b *UserBucketReaderClient) fullName(name string) string {
+	if b.prefix != "" {
```
pstibrany, line 41: `return path.Join(b.prefix, b.userID, name)` -- this also skips empty elements

pkg/storage/bucket/user_bucket_client.go
```diff
@@ -59,7 +68,7 @@ func (b *UserBucketReaderClient) Iter(ctx context.Context, dir string, f func(st
 			Since all objects are prefixed with the userID we need to strip the userID
```
pstibrany, line 68: Comment should be updated too.

pkg/storage/bucket/user_bucket_client.go
```diff
@@ -12,6 +12,7 @@ import (
 // UserBucketReaderClient is a wrapper around a objstore.BucketReader that reads from user-specific subfolder.
 type UserBucketReaderClient struct {
 	userID string
+	prefix string
```
pstibrany, line 15: I would suggest to keep prefix-functionality separate from `UserBucketClient`. Once that is done, `UserBucketClient` could be implemented by using `PrefixedBucketClient`. In PR #3811 `UserBucketClient` is getting extra user-specific functionality, but only on write path. WDYT?

That would also allow us to combine multiple prefixed bucket clients together by wrapping them, instead of adding multiple prefix-levels here.

pkg/storage/rulestore/bucket_client.go
```diff
@@ -0,0 +1,308 @@
+package rulestore
```
pstibrany, line 1: This belongs under ruler package, perhaps `pkg/ruler/rules/rulestore`? (which is honestly pretty bad, but local and objectclient live there too). 

Instead of rulestore, we could call it `bucketstore`, since `local` and `objectclient` are also called based on implementation.

pstibrany (commented):
Some initial comments from first quick look.

pkg/storage/bucket/user_bucket_client.go
jtlisi: Cool, I'll update it to use that function.

pkg/storage/bucket/user_bucket_client.go
jtlisi: 👍

pkg/storage/rulestore/bucket_client.go
jtlisi: I would prefer to move the functionality from that package here once the old `objectclient` is deprecated. I think it would make sense to centralize packages that revolve around object storage under the `storage/` package prefix. As for the relevant protos from that package, I think a new `rulestorepb` package could be created either under `pkg/storage/rulestore` or `pkg/ruler/rulestorepb`.

pkg/storage/bucket/user_bucket_client.go
jtlisi: I removed any changes to the `UserBucketClient` and implemented the necessary functionality using a newly created `PrefixedBucketClient`.

pkg/storage/rulestore/bucket_client.go
pstibrany: > I think it would make sense to centralize packages that revolve around object storage client under the `storage/` package prefix.

This is a wrong approach to modularization. Implementation detail (using bucket client) is not main categorization here – but the fact that it is part of ruler component is.

> As for the relevant protos from that package, I think the rulestore proto definitions can be moved to a `rulestorepb` package under `pkg/ruler/rulestorepb`.

I don't think packages under `pkg/storage` should depend on packages under `pkg/ruler`.

jtlisi: > This is a wrong approach to modularization. Implementation detail (using bucket client) is not main categorization here – but the fact that it is part of ruler component is.

Fair enough. It should be straightforward to move this to the `pkg/ruler/rulestore`.

jtlisi: Done

jtlisi:
@pstibrany Thanks for the feedback, this should be ready for another look.

pkg/ruler/rulestore/config.go
```diff
@@ -0,0 +1,58 @@
...
+// RegisterFlags registers the backend storage config.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.ExtraBackends = []string{ConfigDB}
+	cfg.ConfigDB.RegisterFlagsWithPrefix(prefix, f)
+	cfg.RegisterFlagsWithPrefix(prefix, f)
+}
+
+// NewRuleStore creates a new bucket client based on the configured backend
+func NewRuleStore(ctx context.Context, cfg Config, logger log.Logger, reg prometheus.Registerer) (rules.RuleStore, error) {
+	if cfg.Backend == ConfigDB {
```
pstibrany, line 37: Why do we need another way of configuring obsolete ConfigDB rule store? It's already possible to use it without having this option.

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+func NewBucketRuleStore(bucket objstore.Bucket, loadConcurrency int, logger log.Logger) *BucketRuleStore {
+	return &BucketRuleStore{
+		bucket:          bucket,
+		loadConcurrency: loadConcurrency,
+		logger:          logger,
+	}
+}
+
+func newRuleStoreUserBucketClient(userID string, bkt objstore.Bucket) objstore.Bucket {
+	return bucket.NewUserBucketClient(userID, bucket.NewPrefixedBucketClient(bkt, rulePrefix))
```
pstibrany, line 43: I'd suggest adding a test to verify that user-bucket and prefixed-bucket are composed correctly, and produce expected paths (rules/user1/...).

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+
+func (b *BucketRuleStore) listNamespacesForUser(ctx context.Context, user string) ([]string, error) {
+	userBucket := newRuleStoreUserBucketClient(user, b.bucket)
+
+	var namespaces []string
+	err := userBucket.Iter(ctx, "", func(namespace string) error {
+		namespace = strings.TrimSuffix(namespace, delim)
+		decodedNamespace, err := base64.URLEncoding.DecodeString(namespace)
+		if err != nil {
+			return err
```
pstibrany, line 54:
```suggestion
			return errors.Wrap(err, "failed to decode namespace")
```

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return rg, nil
+}
+
+func (b *BucketRuleStore) ListAllUsers(ctx context.Context) ([]string, error) {
+	var users []string
+	err := b.bucket.Iter(ctx, rulePrefix, func(user string) error {
```
pstibrany, line 105: Why not use prefixed bucket here? (Also b.bucket could itself be prefixed bucket already). No need to have "prefix" handling at more than 1 place.

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+		user = strings.TrimSuffix(strings.TrimPrefix(user, rulePrefix+delim), delim)
+		iterErr := tenant.ValidTenantID(user)
```
pstibrany, line 107: We don't do this validation in objectclient RuleStore. Does it really belong here?

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+		// rules for everyone else.
+		if iterErr != nil {
+			level.Error(b.logger).Log("msg", "invalid tenant id found in rule store", "err", iterErr, "user", user)
+		}
+
+		users = append(users, user)
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("unable to list users in rule store bucket, %w", err)
```
pstibrany, line 119: Nit:
```suggestion
		return nil, fmt.Errorf("unable to list users in rule store bucket: %w", err)
```

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+
+func (b *BucketRuleStore) ListAllRuleGroups(ctx context.Context) (map[string]rules.RuleGroupList, error) {
+	users, err := b.ListAllUsers(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	perUserRuleGroupListMap := map[string]rules.RuleGroupList{}
+
+	for _, user := range users {
```
pstibrany, line 133: We can call `ListRuleGroupsForUserAndNamespace` for each user with empty namespace, and put results to a map. No need to reimplement it here. Also we could do it concurrently to run faster.

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+
+	var namespaces []string
+	var err error
+
+	if namespace != "" {
+		namespaces = []string{namespace}
+	} else {
+		namespaces, err = b.listNamespacesForUser(ctx, userID)
+		if err != nil {
+			return nil, err
```
pstibrany, line 177:
```suggestion
			return nil, errors.Wrapf(err, "failed to list namespaces for user %v", userID)
```

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+
+	groupList := rules.RuleGroupList{}
+
+	for _, namespace := range namespaces {
+		prefix := generateRuleObjectKey(namespace, "")
+		err = userBucket.Iter(ctx, prefix, func(group string) error {
+			group = strings.TrimSuffix(strings.TrimPrefix(group, prefix), delim)
+			decodedGroup, err := base64.URLEncoding.DecodeString(group)
+			if err != nil {
+				return err
```
pstibrany, line 189:
```suggestion
				return err
				return errors.Wrap(err, "failed to decode group")
```

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+				}
+
+				user, namespace, group := gr.GetUser(), gr.GetNamespace(), gr.GetName()
+				if user == "" || namespace == "" || group == "" {
+					return fmt.Errorf("invalid rule group: user=%q, namespace=%q, group=%q", user, namespace, group)
+				}
+
+				gr, err := b.getRuleGroup(gCtx, user, namespace, group, gr) // reuse group pointer from the map.
+				if err != nil {
+					level.Error(b.logger).Log("msg", "failed to get rule group", "namespace", namespace, "name", group, "user", user)
```
pstibrany, line 230: We don't log the error here. I think we should. (Also in objectclient/rule_store.go)

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+	ch := make(chan *rules.RuleGroupDesc)
+
+	// Given we store one file per rule group. With this, we create a pool of workers that will
+	// download all rule groups in parallel. We limit the number of workers to avoid a
+	// particular user having too many rule groups rate limiting us with the object storage.
+	g, gCtx := errgroup.WithContext(ctx)
+	for i := 0; i < b.loadConcurrency; i++ {
+		g.Go(func() error {
+			for gr := range ch {
+				if gr == nil {
```
pstibrany, line 219: Nit: It is actually bit strange to send `nil` to the channel. Perhaps we should just skip it in the `for` loop below.

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,312 @@
...
+	return b.getRuleGroup(ctx, userID, namespace, group, nil)
+}
+
+func (b *BucketRuleStore) SetRuleGroup(ctx context.Context, userID string, namespace string, group *rules.RuleGroupDesc) error {
+	userBucket := newRuleStoreUserBucketClient(userID, b.bucket)
+	data, err := proto.Marshal(group)
+	if err != nil {
+		return err
+	}
+	return userBucket.Upload(ctx, generateRuleObjectKey(namespace, group.Name), bytes.NewReader(data))
```
pstibrany, line 269: Please use `bytes.NewBuffer(data)` instead of `bytes.NewReader` as bucket `Upload` function tries to find the size of the data before upload, and it cannot do it for `bytes.Reader`. (See `objstore.TryToGetSize` for details)

pstibrany (commented):
Good job! I've left some comments to take a look at.

pkg/storage/bucket/prefixed_bucket_client.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Close implements io.Closer
+func (b *PrefixedBucketClient) Close() error {
+	return b.bucket.Close()
+}
+
+// Upload the contents of the reader as an object into the bucket.
+func (b *PrefixedBucketClient) Upload(ctx context.Context, name string, r io.Reader) (err error) {
+	fmt.Println(b.fullName(name))
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/storage/bucket/prefixed_bucket_client.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Name returns the bucket name for the provider.
+func (b *PrefixedBucketClient) Name() string { return b.bucket.Name() }
+
+// Iter calls f for each entry in the given directory (not recursive.). The argument to f is the full
+// object name including the prefix of the inspected directory. The configured prefix will be stripped
+// before supplied function is applied.
+func (b *PrefixedBucketClient) Iter(ctx context.Context, dir string, f func(string) error) error {
+	if b.prefix != "" {
```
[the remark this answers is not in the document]
jtlisi: Makes sense, I'll remove it.

pkg/ruler/rulestore/config.go
jtlisi: Based on [this discussion](https://github.com/cortexproject/cortex/discussions/3771#discussioncomment-327738) we would eventually deprecate the config options currently configured using the `-ruler.storage.*` flag prefix in favor of the the newly added config field. However, the config service is still in use and needs to be carried over. I'm in favor of deprecating the config service but I think it would be best to scope that separately.

pkg/ruler/rulestore/bucket_client.go
jtlisi: Any opposition to using the standard library error wrapping instead?

```go
return fmt.Errorf("failed to decode namespace '%s', %w", namespace, err)
```

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍 

I updated it to use `fmt.Errorf` with with the `%w` verb

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍

pkg/ruler/rulestore/bucket_client.go
jtlisi: Makes sense, I'll keep in sequential for now and make it concurrent in a future PR.

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍  Makes sense, the context on tenants should be left to high level modules.

pkg/ruler/rulestore/bucket_client.go
pstibrany: No, although at some point it would be nice to unify this, and we use `errors` package quite heavily already.

pkg/ruler/rulestore/config.go
pstibrany: I didn't mean to deprecate configdb (although perhaps we should), just that we don't need a separate way of configuring it.

The way I was thinking about the transition was to add config options to existing `RuleStoreConfig`, and adding new types like "bucket-s3". But I see that it could create more confusion.

Ideally we will do this without breaking changes for ConfigDB users.

pkg/ruler/rulestore/bucket_client.go
jtlisi: Makes sense, `b.bucket` should be fine as a PrefixedBucket

pkg/storage/bucket/prefixed_bucket_client.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"fmt"
+	"io"
+	"path"
+	"strings"
+
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+type PrefixedBucketClient struct {
```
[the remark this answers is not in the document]
jtlisi: 👍 , Is it ok if that is done in a future PR? There is already an outstanding PR, #3811, that would likely have a merge conflict if I did an embedded approach as things stand.

pkg/ruler/rulestore/config.go
jtlisi: > The way I was thinking about the transition was to add config options to existing RuleStoreConfig, and adding new types like "bucket-s3". But I see that it could create more confusion.

I also felt it would be more confusing to try and fit this change into the existing config struct instead of creating a new top level struct. As for the configdb, the upgrade path would be identical to the bucket upgrade path, except only the prefix of the flags would change.

However, another option would be to:
1. Not add the configdb as an option for the new RuleStore config struct
2. When deprecating the existing objectclient flags/fields, maintain the config options for the `configdb`. However, give a clear warning that this client is not recommended for new deployments.
3. Figure out the future of the configdb service separately

pkg/ruler/rulestore/bucket_client.go
jtlisi: > No, although at some point it would be nice to unify this, and we use errors package quite heavily already.

I agree, I created a thread to center the discussion on this topic: https://github.com/cortexproject/cortex/discussions/3834

pkg/ruler/rulestore/bucket_client.go
jtlisi: I ended up removing this function since it's superflous when the prefixed bucket is a field on the rulestore. In terms of a test for this, wouldn't it be enough to pass the existing rulestore tests to verify the user/prefix bucket are composed correctly?

pkg/storage/bucket/prefixed_bucket_client.go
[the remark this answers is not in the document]
pstibrany: Sure, separate PR is fine.

pkg/ruler/rulestore/bucket_client.go
pstibrany: > In terms of a test for this, wouldn't it be enough to pass the existing rulestore tests to verify the user/prefix bucket are composed correctly?

You are right. I can see that tests already check for full path of some objects.

Would it make sense to refactor tests and reuse them for both bucket-based rulestore and objectclient-based one?

jtlisi: I think that would make sense. However, I may update the tests to use the rule store `Get` method instead of directly checking the underlying bucket if an object has been deleted. This will make it easier to write a reusable test having to consider the underlying implementation.

jtlisi: I agree, I'm hesitant to do it in this PR since it will push the LOC changes in this PR beyond where it's already at. Is it ok if I do this in a separate PR?

pstibrany: Sounds good to me.

pkg/storage/bucket/client.go
```diff
@@ -55,6 +55,15 @@ type Config struct {
 	// Not used internally, meant to allow callers to wrap Buckets
 	// created using this config
 	Middlewares []func(objstore.Bucket) (objstore.Bucket, error) `yaml:"-"`
+
+	// Used to inject additional backends into the config. Allows for this config to
+	// be embedded in multiple contexts and support non-object storage based backends.
+	ExtraBackends []string `yaml:"-"`
```
pstibrany, line 61: This is quite smart way of extending bucket.Config with more options.

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/storage/bucket"
+)
+
+const (
+	delim      = "/"
+	rulePrefix = "rules"
+)
+
+type BucketRuleStore struct {
```
pstibrany, line 26: Nit: comment this?

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if len(ruleGroupList) == 0 {
+		return rules.ErrGroupNamespaceNotFound
+	}
+
+	for _, rg := range ruleGroupList {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		userBucket := bucket.NewUserBucketClient(userID, b.bucket)
```
pstibrany, line 261: We should only create `userBucket` once, not for every rule group.

pstibrany (approved):
LGTM, good job!

pkg/ruler/rulestore/bucket_client.go
```diff
@@ -0,0 +1,285 @@
...
+		g.Go(func() error {
+			for gr := range ch {
+				user, namespace, group := gr.GetUser(), gr.GetNamespace(), gr.GetName()
+				if user == "" || namespace == "" || group == "" {
+					return fmt.Errorf("invalid rule group: user=%q, namespace=%q, group=%q", user, namespace, group)
+				}
+
+				gr, err := b.getRuleGroup(gCtx, user, namespace, group, gr) // reuse group pointer from the map.
+				if err != nil {
+					level.Error(b.logger).Log("msg", "failed to get rule group", "namespace", namespace, "name", group, "user", user, "err", err)
```
pracucci, line 192: We log in this case, but not other `return` error cases. What if we don't log any error here, but wrap the error with info about the namespace/group/user and let the caller log it?

pracucci (approved):
Good job! 👏 ! 👏  Please mention it in the list of experimental features at `docs/configuration/v1-guarantees.md`.

pkg/cortex/cortex.go
```diff
@@ -115,6 +116,7 @@ type Config struct {
 	TenantFederation tenantfederation.Config         `yaml:"tenant_federation"`
 
 	Ruler          ruler.Config                               `yaml:"ruler"`
+	RulerStorage   rulestore.Config                           `yaml:"ruler_storage" doc:"hidden"`
```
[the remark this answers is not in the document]
jtlisi: Makes sense, should be good to expose since we explicitly declare it experimental in the guarantees doc

pkg/ruler/rulestore/config.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/cortexproject/cortex/pkg/configs/client"
+	"github.com/cortexproject/cortex/pkg/ruler/rules"
+	"github.com/cortexproject/cortex/pkg/storage/bucket"
+)
+
+const (
+	ConfigDB = "configdb"
+
+	name   = "rule-store"
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/rulestore/config.go
```diff
@@ -0,0 +1,58 @@
...
+
+		return rules.NewConfigRuleStore(c), nil
+	}
+
+	bucketClient, err := bucket.NewClient(ctx, cfg.Config, name, logger, reg)
+	if err != nil {
+		return nil, err
+	}
+
+	store := NewBucketRuleStore(bucketClient, 10, logger)
```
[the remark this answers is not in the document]
jtlisi: Makes sense. I think we should avoid exposing more config options for this client unless there is a compelling need to do so.

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
 * [CHANGE] Store-gateway: the chunks pool controlled by `-blocks-storage.bucket-store.max-chunk-pool-bytes` is now shared across all tenants. #3830
 * [CHANGE] Ingester: return error code 400 instead of 429 when per-user/per-tenant series/metadata limits are reached. #3833
+* [FEATURE] Experimental Ruler Storage: Add a separate set of configuration options to configure the ruler storage backend under the `-ruler-storage.` flag prefix. All TSDB bucket clients and the config service are currently supported. Clients using this implementation will only be enabled if the existing `-ruler.storage` flags are left unset. #3805
```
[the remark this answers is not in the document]
jtlisi: 👍

pkg/ruler/rulestore/bucket_client.go
jtlisi: 👍

pracucci:
Fantastic job @jtlisi. LGTM! 🚀

PR #3810, opened 2021-02-10
Add S3 KMS support to blocks storage client

pracucci:
**What this PR does**:
In the PR https://github.com/cortexproject/cortex/pull/3651 we've added S3 server-side encryption support for the chunks storage client. This PR adds the support (with same exact CLI flags / YAML config naming) to the blocks storage client too.

**Which issue(s) this PR fixes**:
N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3811, opened 2021-02-10
Add per-tenant S3 KMS SSE encryption key support to blocks storage

pracucci:
**What this PR does**:
In the PR https://github.com/cortexproject/cortex/pull/3810 I'm proposing to add S3 KMS SSE encryption key support with a global config. On top of that, we'll need to also add support for a per-tenant encryption key override.

In this PR I'm proposing a refactoring to add the support in the blocks storage. The idea is to use `UserBucketClient` everywhere and take in input `overrides`. The `UserBucketClient.Upload()` will read the per-tenant encryption key from overrides and then inject it when calling the parent `Upload()` (idea is to inject via context).

_I've manually tested it and looks working._

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/storage/bucket/user_bucket_client.go
```diff
@@ -9,6 +9,13 @@ import (
 	"github.com/thanos-io/thanos/pkg/objstore"
 )
 
+// UserBucketOverrides defined config overrides supported by UserBucketClient.
+type UserBucketOverrides interface {
```
pstibrany, line 13: It's implemented by `validation.Overrides`, but we can give it a better name here. (Preferably something without "overrides". What about `TenantKeyProvider`?)

pkg/chunk/purger/blocks_purger_api.go
```diff
@@ -17,24 +17,30 @@ import (
...
 )
 
 type BlocksPurgerAPI struct {
 	bucketClient objstore.Bucket
 	logger       log.Logger
+	overrides    *validation.Overrides
 }
 
-func NewBlocksPurgerAPI(storageCfg cortex_tsdb.BlocksStorageConfig, logger log.Logger, reg prometheus.Registerer) (*BlocksPurgerAPI, error) {
+func NewBlocksPurgerAPI(storageCfg cortex_tsdb.BlocksStorageConfig, overrides *validation.Overrides, logger log.Logger, reg prometheus.Registerer) (*BlocksPurgerAPI, error) {
```
pstibrany, line 29: I'd just use new `UserBucketOverrides` type, to better express dependencies and reduce unnecessary imports. (Applies to all modified files, where applicable)

pkg/querier/blocks_store_queryable.go
```diff
@@ -84,10 +84,11 @@ type BlocksStoreClient interface {
 	RemoteAddress() string
 }
 
-// BlocksStoreLimits is the interface that should be implemented by the limits provider.
-type BlocksStoreLimits interface {
+// BlocksStoreConfigOverrides is the interface that should be implemented by the config overrides provider.
+type BlocksStoreConfigOverrides interface {
```
pstibrany, line 88: We should just make this interface extend `UserBucketOverrides` interface.

pkg/storage/bucket/user_bucket_client.go
```diff
@@ -9,6 +9,13 @@ import (
...
+	// S3SSEKMSKeyID returns the per-tenant S3 KMS-SSE key id or an empty string
+	// if no override is set.
+	S3SSEKMSKeyID(userID string) string
```
pstibrany, line 16: Not sure how to improve this method name, but thank god for autocomplete. (Adding underscores would not work, right? Other than that, no good suggestion.)

pstibrany (commented):
LGTM so far, except I wouldn't rename existing interfaces to include "Overrides", and instead of adding the `S3SSEKMSKeyID` method everywhere, we can extend existing interfaces with newly introduced one.

pkg/storage/bucket/user_bucket_client.go
pracucci: I thought to keep it more generic, cause we may need to pass further config overrides in the future (eg. we already need to also support "encryption context" override). I can remove the "overrides" naming an call it `UserBucketConfigProvider` if you prefer.

pstibrany: I'm not fan of too generic naming too early, but `UserBucketConfigProvider` is better than `UserBucketOverrides` in my opinion.

pracucci: Discussed offline. We agreed on `TenantConfigProvider`.

pkg/storage/bucket/user_bucket_client.go
pracucci: Yes, that name is ugly but don't think underscores are an option.

pkg/querier/blocks_store_queryable.go
```diff
@@ -84,8 +84,10 @@ type BlocksStoreClient interface {
...
```
pstibrany, line 88: No need to change this name (we're not overriding anything here).

pstibrany (approved):
LGTM.

I would still prefer if we didn't rename existing `BlocksStoreLimits` type to `BlocksStoreConfigOverrides`.

pstibrany (approved):
LGTM, nice work!

pkg/storage/bucket/user_bucket_client.go
alvinlin123: One way might be to spell out SSE like `S3ServierSideEncryptionKMSKeyID` it's kinda long, but maybe it's more readable then bunch of `S` and `3` stuffed together :-) (thank got for autocomplete in this case too). 

Other ways may be to swap terms around like 

`SSEKMSKeyIDS3` but I don't think this helps much. 

Another approach may be to use nested interface like: 

```go
type S3ServerSideEncryptionConfigProvider interface {
     KMSKeyID() string
     KMSType() string
}

type TenantConfigProvider interface {
      S3ServerSideEncryptionConfig(user string) S3ServerSideEncryptionConfigProvider
}
```

This is a lot of typing, but maybe seeing `config.S3ServerSideEncryptionConfig(userId).KMSKeyID()` in the code is worth the typing?

docs/blocks-storage/store-gateway.md
```diff
@@ -283,6 +283,10 @@ blocks_storage:
     # CLI flag: -blocks-storage.s3.endpoint
     [endpoint: <string> | default = ""]
 
+    # S3 region. If unset an S3 API call is made to autodetect it.
```
alvinlin123, line 286: Which S3 API call is made to auto detect region?

pracucci: The Minio client uses this API: https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html

alvinlin123: I see. Do you think it's worth it to be more clean in the doc and just spell out that GetBucketLocation API is called?

pracucci: Yes, done.

pkg/storage/bucket/user_bucket_client.go
pracucci: I've the feeling we're just going to add complexity for the sake of clarifying what SSE means, but if you're into Cortex internals I believe you quickly realise what it means.

pstibrany (approved):
Reviewed latest changes – great job! I think CHANGELOG.md should mention the fact that SSE config can be changed for each tenant individually, using overrides mechanism in the runtime config.

Issue #3779, opened 2021-02-03
Remove global logger


jtlisi:
A global logger from `pkg/util/log` is used throughout the codebase. This issue is meant to track removing usage of the global logger from the codebase.

See: 
https://github.com/cortexproject/cortex/blob/3df3149ef131982e8f9038fb291d005e85caa16f/pkg/util/log.go#L17-L19

pracucci:
_Still valid. We've done much work, but global logger is still not fully eradicated._

pracucci:
Still valid. This is also a pre-requisite for all components we'll move to dskit.

pracucci:
> so I override this public logger to my own logger right now, what's the proper way to let cortex use customized logger?

That's expected way to do it, until we'll get rid of the global logger.

PR #3821, opened 2021-02-12
chore: remove global logger from the querier package

jtlisi:
**What this PR does**:

- Remove the util_log.Logger from all files in the querier package and subpackages.
- Add a faillint rule to trigger when the global logger is used in the querier package.

Related #3779

### DIFF
--- a/cmd/cortex/main.go
+++ b/cmd/cortex/main.go
@@ -141,7 +141,7 @@ func main() {
 
 	util_log.InitLogger(&cfg.Server)
 
-	Set the Go standard library logger to use the same logger as Cortex
+	//	Set the Go standard library logger to use the same logger as Cortex
 	stdlog.SetOutput(log.NewStdlibAdapter(util_log.Logger))
 
 	// Allocate a block of memory to alter GC behaviour. See https://github.com/golang/go/issues/23044
--- a/cmd/cortex/main.go
+++ b/cmd/cortex/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	stdlog "log"
 	"math/rand"
 	"os"
 	"runtime"
@@ -12,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -138,6 +140,9 @@ func main() {
 	}
 
 	util_log.InitLogger(&cfg.Server)
+
+	Set the Go standard library logger to use the same logger as Cortex
+	stdlog.SetOutput(log.NewStdlibAdapter(util_log.Logger))
 
 	// Allocate a block of memory to alter GC behaviour. See https://github.com/golang/go/issues/23044
 	ballast := make([]byte, ballastBytes)
--- a/cmd/cortex/main.go
+++ b/cmd/cortex/main.go
@@ -141,7 +141,7 @@ func main() {
 
 	util_log.InitLogger(&cfg.Server)
 
-	//	Set the Go standard library logger to use the same logger as Cortex
+	// Set the Go standard library logger to use the same logger as Cortex.
 	stdlog.SetOutput(log.NewStdlibAdapter(util_log.Logger))
 
 	// Allocate a block of memory to alter GC behaviour. See https://github.com/golang/go/issues/23044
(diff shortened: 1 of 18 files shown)


pstibrany (commented):
This PR removes some log messages that can be useful (in parallel.go). Instead of util_log.Logger, we could use another logger which is available in all Go programs, namely logger in stdlib "log" package. We can use that instead (and configure it in main.go to pass through go-kit's logger). It's not ideal, but better than losing the log message, or panicking (the other PR). WDYT?

jtlisi:
> This PR removes some log messages that can be useful (in parallel.go). Instead of util_log.Logger, we could use another logger which is available in all Go programs, namely logger in stdlib "log" package. We can use that instead (and configure it in main.go to pass through go-kit's logger). It's not ideal, but better than losing the log message, or panicking (the other PR). WDYT?

Love this idea. There are some edge cases like this where forcing global logging is tricky. I feel relying on the standard library is the correct solution

cmd/cortex/main.go
```diff
@@ -139,6 +141,9 @@ func main() {
 
 	util_log.InitLogger(&cfg.Server)
 
+	//	Set the Go standard library logger to use the same logger as Cortex
```
pracucci, line 144: [nit] Just removing whitespaces.

```suggestion
	// Set the Go standard library logger to use the same logger as Cortex.
```

pkg/querier/astmapper/parallel.go
```diff
@@ -90,7 +87,7 @@ func CanParallelize(node parser.Node) bool {
 		return true
 
 	default:
-		level.Error(util_log.Logger).Log("err", fmt.Sprintf("CanParallel: unhandled node type %T", node))
+		log.Printf("err=\"CanParallel: unhandled node type %T\"", node)
```
pracucci, line 90: I have a mixed feeling about it. We're just going from a global logger to another global one. I would keep the previous code until we figure out a better solution. To skip the `faillint` check in this specific case you can add the comment `//lint:ignore faillint`

pkg/querier/querier.go
```diff
@@ -167,7 +167,7 @@ func New(cfg Config, limits *validation.Overrides, distributor Distributor, stor
 	})
 
 	engine := promql.NewEngine(promql.EngineOpts{
-		Logger:             util_log.Logger,
+		Logger:             log.NewNopLogger(),
```
pracucci, line 170: Why noop logger here? It's not a test.

pkg/querier/querier.go
```diff
@@ -199,7 +199,7 @@ func createActiveQueryTracker(cfg Config) *promql.ActiveQueryTracker {
 	dir := cfg.ActiveQueryTrackerDir
 
 	if dir != "" {
-		return promql.NewActiveQueryTracker(dir, cfg.MaxConcurrent, util_log.Logger)
+		return promql.NewActiveQueryTracker(dir, cfg.MaxConcurrent, log.NewNopLogger())
```
pracucci, line 202: Why noop logger here? It's not a test.

pracucci (commented):
Thanks Jacob for working on it! I left few comments

cmd/cortex/main.go
jtlisi: 👍

pkg/querier/astmapper/parallel.go
jtlisi: Will do

pkg/querier/querier.go
jtlisi: Wow, good catch. I must have accidentally ran the test find and replace code across the package and missed it. Thanks for spotting this. It would have been a major regression.

Will fix this up ASAP.

pkg/querier/querier.go
jtlisi: Will fix this up ASAP.

cmd/cortex/main.go
```diff
@@ -141,7 +141,7 @@ func main() {
 
 	util_log.InitLogger(&cfg.Server)
 
-	//	Set the Go standard library logger to use the same logger as Cortex
+	// Set the Go standard library logger to use the same logger as Cortex.
 	stdlog.SetOutput(log.NewStdlibAdapter(util_log.Logger))
```
pracucci, line 145: This is no more required, right?

pracucci (approved):
LGTM (modulo a nit). Thanks!

cmd/cortex/main.go
jtlisi: 👍

PR #3825, opened 2021-02-15
Prevent failed ingestion from affecting rate limiting in distributor.

### LINKED ISSUE #3804: Distributor ingestion rate limit increased for retries due to ingestion failure
The distributor ingestion rate limit increases the number of "consumed tokens" in the rate limiter once the request is received and before writing to ingesters:
https://github.com/cortexproject/cortex/blob/527f9b58b93ce234e8d8624724efeeebcdf7de78/pkg/distributor/distributor.go#L581

In the event of an ingesters outage (eg. 2+ ingesters are unavailable), this means that each tenant remote write request will consume tokens from its rate limiter even if samples have not been successfully ingested. The client (eg. Prometheus) will retry writes and this will further consume tokens from the rate limiter, until it will eventually hit the rate limit, regardless any samples has been actually ingested.

The burst should protect from this, but in the event of a relatively long outage we would end up consuming the burst too (eg. we set burst to 10x the rate limit).

I'm wondering if a better approach would be checking if enough tokens are still available in the rate limiter once the request is received but actually consuming them from the rate limiter only after samples have been successfully written to ingesters. Due to concurrency, the actual accepted rate could be higher than the limit, but we would err in favour of the customer instead of rate limiting for writes we haven't actually ingested.

Related discussions:
- [Distributor rate-limiter should call Wait() not Allow()](https://github.com/cortexproject/cortex/issues/879)

stevesg:
**What this PR does**:
Change the handling of the rate limiting in the distributor such that when the ingestion fails, the reservation in the rate limiter is canceled, returning the tokens for future use in the proceeding requests.

In the process, a change was made to the `util/limiter` package to use `ReserveN`. To match the semantics of `ApplyN`, we have to check the `Delay()` on the returned `rate.Reservation` is zero. I also chose to only expose a limited subset of the rate.Reservation struct (to use `CancelAt`), following from how the existing code does not expose the `x/time/rate` package.

Unit tests added to `rate_limiter_test.go` and `distributor_test.go` to check the cancel behaviour.

Fixes #3804

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor_test.go
+++ b/pkg/distributor/distributor_test.go
@@ -414,7 +414,7 @@ func TestDistributor_PushIngestionRateLimiter(t *testing.T) {
 				{metadata: 1, expectedError: httpgrpc.Errorf(http.StatusTooManyRequests, "ingestion rate limit (5) exceeded while adding 0 samples and 1 metadata")},
 			},
 		},
-		"unhappy ingestors: rate limit should be unaffected when ingestion fails": {
+		"unhappy ingesters: rate limit should be unaffected when ingestion fails": {
 			distributors:          1,
 			ingestionRateStrategy: validation.LocalIngestionRateStrategy,
 			ingestionRate:         10,
--- a/pkg/distributor/distributor_test.go
+++ b/pkg/distributor/distributor_test.go
@@ -369,6 +369,7 @@ func TestDistributor_PushIngestionRateLimiter(t *testing.T) {
 		ingestionRateStrategy string
 		ingestionRate         float64
 		ingestionBurstSize    int
+		ingestionFailing      bool
 		pushes                []testPush
 	}{
 		"local strategy: limit should be set to each distributor": {
@@ -413,6 +414,17 @@ func TestDistributor_PushIngestionRateLimiter(t *testing.T) {
 				{metadata: 1, expectedError: httpgrpc.Errorf(http.StatusTooManyRequests, "ingestion rate limit (5) exceeded while adding 0 samples and 1 metadata")},
 			},
 		},
+		"unhappy ingestors: rate limit should be unaffected when ingestion fails": {
+			distributors:          1,
+			ingestionRateStrategy: validation.LocalIngestionRateStrategy,
+			ingestionRate:         10,
+			ingestionBurstSize:    10,
+			ingestionFailing:      true,
+			pushes: []testPush{
+				{samples: 10, expectedError: errFail},
+				{samples: 10, expectedError: errFail},
+			},
+		},
 	}
 
 	for testName, testData := range tests {
@@ -425,10 +437,15 @@ func TestDistributor_PushIngestionRateLimiter(t *testing.T) {
 			limits.IngestionRate = testData.ingestionRate
 			limits.IngestionBurstSize = testData.ingestionBurstSize
 
+			happyIngesters := 3
+			if testData.ingestionFailing {
+				happyIngesters = 0
+			}
+
 			// Start all expected distributors
 			distributors, _, r := prepare(t, prepConfig{
 				numIngesters:     3,
-				happyIngesters:   3,
+				happyIngesters:   happyIngesters,
 				numDistributors:  testData.distributors,
 				shardByAllLabels: true,
 				limits:           limits,
--- a/pkg/util/limiter/rate_limiter.go
+++ b/pkg/util/limiter/rate_limiter.go
@@ -61,7 +61,7 @@ func (l *RateLimiter) AllowN(now time.Time, tenantID string, n int) (bool, Reser
 	// the semantics are subtly different to AllowN.
 	r := l.getTenantLimiter(now, tenantID).ReserveN(now, n)
 	if !r.OK() {
-		return false, r
+		return false, nil
 	}
 
 	// ReserveN will still return OK if the necessary tokens are
@@ -72,7 +72,7 @@ func (l *RateLimiter) AllowN(now time.Time, tenantID string, n int) (bool, Reser
 		// Having decided not to use the reservation, return the
 		// tokens to the rate limiter.
 		r.CancelAt(now)
-		return false, r
+		return false, nil
 	}
 
 	return true, r
--- a/pkg/util/limiter/rate_limiter.go
+++ b/pkg/util/limiter/rate_limiter.go
@@ -26,12 +26,15 @@ type RateLimiter struct {
 	tenants     map[string]*tenantLimiter
 }
 
-// Reservation encapsulates rate.Reservation to exclude interfaces which do not
-// make sense to expose, because we are following the semantics of AllowN, being
-// an immediate reservation, i.e. not delayed into the future.
-type Reservation struct {
-	ok          bool
-	reservation *rate.Reservation
+// Reservation is similar to rate.Reservation but excludes interfaces which do
+// not make sense to expose, because we are following the semantics of AllowN,
+// being an immediate reservation, i.e. not delayed into the future.
+type Reservation interface {
+	// CancelAt returns the reservation to the rate limiter for use by other
+	// requests. Note that typically the reservation should be canceled with
+	// the same timestamp it was requested with, or not all the tokens
+	// consumed will be returned.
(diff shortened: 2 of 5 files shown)


stevesg:
Fixed a couple of typos in comments, and added issue number to changelog.

stevesg:
Rebase, change issue number to PR number.

stevesg:
Fix lint

pkg/util/limiter/rate_limiter.go
```diff
@@ -42,9 +50,29 @@ func NewRateLimiter(strategy RateLimiterStrategy, recheckPeriod time.Duration) *
 	}
 }
 
-// AllowN reports whether n tokens may be consumed happen at time now.
-func (l *RateLimiter) AllowN(now time.Time, tenantID string, n int) bool {
-	return l.getTenantLimiter(now, tenantID).AllowN(now, n)
+// AllowN reports whether n tokens may be consumed happen at time now. The
+// reservation of tokens can be canceled using CancelAt on the returned object.
+func (l *RateLimiter) AllowN(now time.Time, tenantID string, n int) *Reservation {
```
pracucci, line 55: This is used in the write path, which is an hot path. I think we could avoid extra allocations for `type Reservation struct` doing the following:

1. Change `type Reservation struct` to `type Reservation interface { CancelAt(now time.Time) }` (`*rate.Reservation` will already implements this interface)
2. Change this `AllowN()` return from `*Reservation` to `(bool, Reservation)` (no pointer because now it's an interface) and directly return `(ok, *rate.Reservation)`

The `type Reservation interface` wouldn't be strictly required, but it's a nice to have to abstract away the underlying `*rate.Reservation` so we don't leak it.

pkg/distributor/distributor_test.go
```diff
@@ -413,6 +414,17 @@ func TestDistributor_PushIngestionRateLimiter(t *testing.T) {
 				{metadata: 1, expectedError: httpgrpc.Errorf(http.StatusTooManyRequests, "ingestion rate limit (5) exceeded while adding 0 samples and 1 metadata")},
 			},
 		},
+		"unhappy ingestors: rate limit should be unaffected when ingestion fails": {
```
pracucci, line 417: They're called "ingesters" not "ingestors". It's a very typical typo.

```suggestion
		"unhappy ingesters: rate limit should be unaffected when ingestion fails": {
```

pracucci (commented):
Good job! I left a comment, but overall LGTM!

pkg/util/limiter/rate_limiter.go
stevesg: Agreed - good point.

pkg/util/limiter/rate_limiter.go
```diff
@@ -42,9 +53,29 @@ func NewRateLimiter(strategy RateLimiterStrategy, recheckPeriod time.Duration) *
...
-	return l.getTenantLimiter(now, tenantID).AllowN(now, n)
+// AllowN reports whether n tokens may be consumed happen at time now. The
+// reservation of tokens can be canceled using CancelAt on the returned object.
+func (l *RateLimiter) AllowN(now time.Time, tenantID string, n int) (bool, Reservation) {
+
+	// Using ReserveN allows cancellation of the reservation, but
+	// the semantics are subtly different to AllowN.
+	r := l.getTenantLimiter(now, tenantID).ReserveN(now, n)
+	if !r.OK() {
+		return false, r
```
pracucci, line 64: I think the reservation should be nil when failed

```suggestion
		return false, nil
```

pkg/util/limiter/rate_limiter.go
```diff
@@ -42,9 +53,29 @@ func NewRateLimiter(strategy RateLimiterStrategy, recheckPeriod time.Duration) *
...
+
+	// ReserveN will still return OK if the necessary tokens are
+	// available in the future, and tells us this time delay. In
+	// order to mimic the semantics of AllowN, we must check that
+	// there is no delay before we can use them.
+	if r.DelayFrom(now) > time.Duration(0) {
+		// Having decided not to use the reservation, return the
+		// tokens to the rate limiter.
+		r.CancelAt(now)
+		return false, r
```
pracucci, line 75:
```suggestion
		return false, nil
```

pkg/util/limiter/rate_limiter_test.go
```diff
@@ -127,3 +163,7 @@ func (s *staticLimitStrategy) Burst(tenantID string) int {
 
 	return tenant.burst
 }
+
+func OK(ok bool, r Reservation) bool {
```
pracucci, line 167: No need to expose it. You can rename it to `isOK()` so that doesn't start with an uppercase letter and doesn't clash with `ok` variable too.

pracucci (approved):
Good job, thanks! LGTM  (modulo a couple of nits)

pkg/util/limiter/rate_limiter.go
```diff
@@ -42,9 +53,29 @@ func NewRateLimiter(strategy RateLimiterStrategy, recheckPeriod time.Duration) *
...
+	r := l.getTenantLimiter(now, tenantID).ReserveN(now, n)
+	if !r.OK() {
+		return false, nil
+	}
+
+	// ReserveN will still return OK if the necessary tokens are
+	// available in the future, and tells us this time delay. In
+	// order to mimic the semantics of AllowN, we must check that
+	// there is no delay before we can use them.
+	if r.DelayFrom(now) > time.Duration(0) {
```
pstibrany, line 71:
```suggestion
	if r.DelayFrom(now) > 0 {
```

pkg/util/limiter/rate_limiter.go
```diff
@@ -26,6 +26,17 @@ type RateLimiter struct {
 	tenants     map[string]*tenantLimiter
 }
 
+// Reservation is similar to rate.Reservation but excludes interfaces which do
+// not make sense to expose, because we are following the semantics of AllowN,
+// being an immediate reservation, i.e. not delayed into the future.
+type Reservation interface {
```
pstibrany, line 32: Interface here is not necessary and is actually what adds extra allocation – which is what it tried to prevent in the first place. I would suggest to go back to previous design:

```go
type Reservation struct {
    r *Reservation
}
```

and return it directly as a value and not pointer (!). I would also suggest to add a unit test proving that we don't allocate more than necessary (see `testing.AllocsPerRun()`).

pkg/util/limiter/rate_limiter_test.go
```diff
@@ -127,3 +163,7 @@ func (s *staticLimitStrategy) Burst(tenantID string) int {
 
 	return tenant.burst
 }
+
+func isOK(ok bool, r Reservation) bool {
```
pstibrany, line 167: This function is not necessary.

pkg/util/limiter/rate_limiter.go
stevesg: That's unfortunate! Makes sense to change it back then.

pstibrany: Turns out interface with pointer value doesn't need to do extra allocation, so my objection is not valid.

I still don't think we need the interface here... interfaces in Go are best used to give clients a way to provide different implementations. Interfaces don't hide the underlying implementation, as client code can still do `val.(*rate.Reservation)` on returned value. Custom struct as implemented originally achieves that goal better.

pstibrany (approved):
Thanks!

pkg/util/limiter/rate_limiter.go
stevesg: Got confused about this so did a little experiment: https://play.golang.org/p/iHSGnvXzbEx

It seems assigning a pointer (to something which is already allocated) to an interface does not cause an additional allocation. This makes sense, as the interface is sized to hold pointers, but cannot be sized for structs of arbitrary size, hence in the struct case, a heap allocated copy of the struct must be made.

bboreham:
What happens in the case that some samples succeed and some fail?
Looks like distributor acts as if they all failed.

PR #3826, opened 2021-02-15
Cleanup of query-frontend and query-scheduler metrics for inactive users.

pstibrany:
**What this PR does**: This PR adds cleanup of metrics for inactive users in query-frontend and query-scheduler.

Frontend v1 is now a service on its own (previously it was wrapped into a service), that must be started. Periodic cleanup of inactive users was moved from distributor to a service as well, so that it can be reused between components.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pracucci (approved):
Good job! 👏

PR #3828, opened 2021-02-15
Implement blocksconvert scanner for DynamoDB v9 schema

bboreham:
Previously it was only done for BigTable.

I don't have any v10 or v11 schema tables so haven't spent any time thinking about that.

DynamoDB table scans are un-ordered, so we use a map to remember which series have been processed. In small-scale tests (5 million series in a week) that map is ~300MB, so if your tables are 100x bigger you might need a different implementation.
Each time we encounter a new series ID, we do a separate query to fetch all chunk IDs for that series, so that we can meet the expectation that all chunks for one series come together in the plan file.

[After doing some experiments, it seems that DynamoDB _mostly_ returns all items with the same hash-key together, so we can do this much more efficiently by scanning the entire DynamoDB table as it comes and writing plan files which have some repeated series, then post-processing those files to bring the part-series together.  At worst we only have to hold info in memory for one day for one tenant, rather than all tenants for 7 days.  I don't plan to implement this; just an idea.]

This PR includes a few changes that are broader than DynamoDB:
 * moved the `IndexProcessor` interfaces to `chunk` package, so they can be used by both Bigtable and DynamoDB scanners. (DynamoDB version makes use of unexported code in the `chunk/aws` package).
 * documented a couple more flags
 * make use of the map of ignored users in `scanner_processor.go`

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   - `-alertmanager.receivers-firewall.block.cidr-networks` renamed to `-alertmanager.receivers-firewall-block-cidr-networks`
   - `-alertmanager.receivers-firewall.block.private-addresses` renamed to `-alertmanager.receivers-firewall-block-private-addresses`
 
+## Blocksconvert
+
+* [ENHANCEMENT] Scanner: add support for DynamoDB (v9 schema only). #3828
+
+
 ## 1.9.0 in progress
 
 * [CHANGE] Fix for CVE-2021-31232: Local file disclosure vulnerability when `-experimental.alertmanager.enable-api` is used. The HTTP basic auth `password_file` can be used as an attack vector to send any file content via a webhook. The alertmanager templates can be used as an attack vector to send any file content because the alertmanager can load any text file specified in the templates list. #4129
--- a/pkg/chunk/aws/dynamodb_index_reader.go
+++ b/pkg/chunk/aws/dynamodb_index_reader.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"strings"
 	"sync"
@@ -53,8 +54,17 @@ func (r *dynamodbIndexReader) IndexTableNames(ctx context.Context) ([]string, er
 }
 
 type seriesMap struct {
-	mutex           sync.Mutex          // protect concurrent access to map
-	seriesProcessed map[string]struct{} // set of all series processes (NOTE: does not scale indefinitely)
+	mutex           sync.Mutex           // protect concurrent access to maps
+	seriesProcessed map[string]sha256Set // map of userID/bucket to set showing which series have been processed
+}
+
+// Since all sha256 values are the same size, a fixed-size array
+// is more space-efficient than string or byte slice
+type sha256 [32]byte
+
+// an entry in this set indicates we have processed a series with that sha already
+type sha256Set struct {
+	series map[sha256]struct{}
 }
 
 // ReadIndexEntries reads the whole of a table on multiple goroutines in parallel.
@@ -64,7 +74,7 @@ func (r *dynamodbIndexReader) ReadIndexEntries(ctx context.Context, tableName st
 	projection := hashKey + "," + rangeKey
 
 	sm := &seriesMap{ // new map per table
-		seriesProcessed: make(map[string]struct{}),
+		seriesProcessed: make(map[string]sha256Set),
 	}
 
 	var readerGroup sync.WaitGroup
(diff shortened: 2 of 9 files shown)


bboreham:
Wouldn't it look better if you add something like this at the top?

```
  // special-case situation created by step bigger than range
  if req.GetStart() == req.GetEnd() {
    // do we have a cached extent that matches this zero-length request?
    if len(extents) == 1 && extents[0].Start == req.GetStart() && extents[0].End == req.GetEnd() {
      res, err := extent.toResponse()
      cachedResponses = append(cachedResponses, res)
    } else {
      // cache didn't match this request; pass it through to querier
      requests = append(requests, req)
    }
    return requests, cachedResponses, nil
  }
```

tools/blocksconvert/scanner/scanner.go
```diff
@@ -213,6 +214,19 @@ func (s *Scanner) running(ctx context.Context) error {
 			}
 
 			reader = newBigtableIndexReader(bigTable.Project, bigTable.Instance, s.logger, s.indexReaderRowsRead, s.indexReaderParsedIndexEntries, s.currentTableRanges, s.currentTableScannedRanges)
+		case "aws-dynamo":
+			cfg := s.storageCfg.AWSStorageConfig
+
+			if cfg.DynamoDB.URL == nil { // TODO: what else do we need?
```
pracucci, line 220: Can we remove/solve this TODO?

pkg/chunk/aws/dynamodb_index_reader.go
```diff
@@ -0,0 +1,238 @@
+package aws
```
pracucci, line 1: Given this is used only by the blocksconvert, why not keeping it in its package?

pkg/chunk/aws/dynamodb_index_reader.go
```diff
@@ -0,0 +1,238 @@
...
+			err := r.DynamoDB.ScanPagesWithContext(ctx, input, func(page *dynamodb.ScanOutput, lastPage bool) bool {
+				if cc := page.ConsumedCapacity; cc != nil {
+					r.metrics.dynamoConsumedCapacity.WithLabelValues("DynamoDB.ScanTable", *cc.TableName).
+						Add(float64(*cc.CapacityUnits))
+				}
+				r.processPage(ctx, sm, processor, tableName, page)
+				return true
+			}, withRetrys)
+			if err != nil {
+				outerErr = err
```
pracucci, line 104: Shouldn't we protect `outerErr` with a mutex?

pkg/chunk/aws/dynamodb_index_reader.go
```diff
@@ -0,0 +1,238 @@
...
+				// TODO: abort all segments
```
pracucci, line 105: I think this is a good case for `errgroup.WithContext()` (we already use it in some places). You can create an `errgroup`, spawn the goroutines with `group.Go()` and the whole group context would be canceled as soon as the first gorouting returns an error. Then the error will be returned to `group.Wait()`.

pracucci (approved):

pkg/chunk/aws/dynamodb_index_reader.go
bboreham: It uses several things which are not exported by `aws`

tools/blocksconvert/scanner/scanner.go
bboreham: Yeah I think that is the only mandatory setting.

CHANGELOG.md
```diff
@@ -81,6 +81,11 @@
 * [BUGFIX] Prevent panic at start if the http_prefix setting doesn't have a valid value. #3796
 * [BUGFIX] Memberlist: fixed panic caused by race condition in `armon/go-metrics` used by memberlist client. #3724
 
+## Blocksconvert
+
+* [ENHANCEMENT] Scanner: Implement for DynamoDB (v9 schema only). #3828
```
pstibrany, line 86: Cortex release 1.8.0 is now in progress. Could you please rebase master and move the CHANGELOG entry under the master / unreleased section?

pracucci: @bboreham Could you address this, so we can merge it, please?

bboreham:
I rebased, moved the CHANGELOG and converted from WaitGroup to errgroup.
Only thing now is I should really try it out, but I already converted all the production data.
Will try to give it a run on some staging system.

pracucci:
> Will try to give it a run on some staging system.

Great, thanks!

bboreham:
I have rebased and am now running a test on staging environment.
Can we try to get into 1.9 release @pracucci ?

bboreham:
I regret my test says this version does not work.
It goes through the motions, but the resulting blocks are quite small compared to previous conversions, and do not seem to contain the data I looked for.

pracucci:
> I regret my test says this version does not work.
> It goes through the motions, but the resulting blocks are quite small compared to previous conversions, and do not seem to contain the data I looked for.

Have you got a chance to figure out what's going on?

bboreham:
> Have you got a chance to figure out what's going on?

No, sorry, I am on vacation this week so don't expect to get back to that problem soon.

bboreham:
It turns out that the failures were down to schema configuration; I thought I had corrected that and re-run, but apparently did not re-run the full set of days so when I checked the data was missing.

However, good news is I believe this PR does work!

CHANGELOG.md
```diff
@@ -73,6 +73,8 @@
 ## Blocksconvert
 
 * [ENHANCEMENT] Builder: add `-builder.timestamp-tolerance` option which may reduce block size by rounding timestamps to make difference whole seconds. #3891
+* [ENHANCEMENT] Scanner: add support for DynamoDB (v9 schema only). #3828
```
pracucci, line 76: Could you rebase `master` and move this under the "master / unreleased", please?

pracucci (approved):
Thanks @bboreham for re-testing it. Could you just rebase `master` and move the CHANGELOG entry before merging it, please? Thanks!

CHANGELOG.md
bboreham: done

PR #3837, opened 2021-02-17
Introduce NamedService interface that allows service to return a name.

pstibrany:
**What this PR does**: This PR introduces `NamedService` interface, which extends Service with a name. `BasicService` implements this, and has new `WithName` method to set the name while service is still in `New` state. Failure watcher reports the name for failed services, if one is available.

**Checklist**
- [x] Tests updated
- [na] Documentation added

### DIFF
--- a/pkg/util/services/services.go
+++ b/pkg/util/services/services.go
@@ -2,6 +2,7 @@ package services
 
 import (
 	"context"
+	"fmt"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/util/services/services.go
```diff
@@ -130,3 +131,15 @@ func StopAndAwaitTerminated(ctx context.Context, service Service) error {
 	// can happen e.g. if context was canceled
 	return err
 }
+
+// DescribeService returns name of the service, if it has one, or returns string representation of the service.
+func DescribeService(service Service) string {
+	name := ""
```
jtlisi, line 137: **suggestion(optional):** should be fine to use the zero value here.
```suggestion
	var name string
```

jtlisi (approved):
LGTM

pracucci (approved):
Thanks!

Issue #2650, opened 2020-05-29
Reuse ring memberlist client in Alertmanager


jtlisi:
In order to run the alertmanager in a single binary the user will have to run two separate gossip ports. One for the alertmanager and another for the ring kv client. We should explore reusing the the ring memberlist implementation in place of the alertmanager cluster implementation.

pracucci:
@pstibrany Isn't reusing the ring client something we already did for other components?

pstibrany:
> @pstibrany Isn't reusing the same memberlist client something we already did for other components?

My understanding of this issue is different. Our existing work in Cortex allowed to use single memberlist-based between different components.

What I believe Jacob describes here is that AlertManager itself uses memberlist library to setup a cluster (with AM nodes). I think Jacob proposes to use Ring for AlertManagers instead, which would allow it to use our memberlist-based Ring implementation as well.

PR #3839, opened 2021-02-18
Alertmanager: Replicate state using the Ring

gotjosh:
**What this PR does**:

Alertmanager typically uses the memberlist gossip based protocol to
replicate state across replicas. In cortex, we used the same fundamentals
to provide some sort of high availability mode.

Now that we have support for sharding instances across many machines, we
can leverage the ring to find the corresponding instances and send the
updates via gRPC.

This is part of the proposal #3574 

And follows up #3664 , #3671 

Marking it as a draft as I have a few todos left that I need to address but the logic is pretty much what you see on the tin at the moment.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/alertmanager/alertmanager.go
```diff
@@ -286,6 +346,22 @@ func (am *Alertmanager) StopAndWait() {
...
+	return am.state.MergePartialState(part)
+}
+
+func (am *Alertmanager) Position() int {
+	am.positionMtx.Lock()
+	defer am.positionMtx.Unlock()
+	return am.position
+}
+
+func (am *Alertmanager) SetPosition(p int) {
```
pracucci, line 359: Question (don't do it, let's just talk about it). Do we need the `SetPosition()`? What if `Alertmanager` was ring-aware and query the ring on demand each time `Position()` is called?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+		Help: "Number of times we have failed to merge a partial state received for a key.",
+	}, []string{"key"})
+	stateReplicationTotal := promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+		Name: "alertmanager_state_replication_total",
+		Help: "Number of times we have tried to replicate a state to other alertmanagers.",
+	}, []string{"key"})
+	stateReplicationFailed := promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+		Name: "alertmanager_state_replication_failed_total",
+		Help: "Number of times we have failed to replicate a state to other alertmanagers.",
+	}, []string{"key"})
```
pracucci, line 68: Can be moved below, within the `&state{}` initialisation.

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+// stateChannel allows a state publisher to send messages that will be broadcasted to all other alertmanagers that a tenant
+// belongs to.
+type stateChannel struct {
+	key    string
+	logger log.Logger
+
+	send func([]byte)
+}
+
+func newStateChannel(key string, send func([]byte), logger log.Logger, reg prometheus.Registerer) *stateChannel {
```
pracucci, line 199: `key`, `logger` and `reg` are unused

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+func (s *state) Address() string {
+	return ""
+}
+
+func (s *state) Peers() []cluster.Node {
+	return nil
+}
+
+// newReplicatedStates creates a new state struct, which manages state to be replicated between alertmanagers.
+func newReplicatedStates(userID string, rf int, f func(context.Context, string, *clusterpb.Part) error, stopc chan struct{}, l log.Logger, r prometheus.Registerer) *state {
```
pracucci, line 52: Design wise, I'm not a big fan of taking in input `stopc`. I think `state` should self-manage its lifecycler (eg. could be a service).

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+func (s *state) MergePartialState(p *clusterpb.Part) error {
+	s.partialStateMergesTotal.WithLabelValues(p.Key).Inc()
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	st, ok := s.states[p.Key]
+	if !ok {
+		level.Debug(s.logger).Log("msg", "state key not found while merging", "key", p.Key)
+		s.partialStateMergesFailed.WithLabelValues(p.Key).Inc()
+		return nil
```
pracucci, line 119: Why not returning error?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	if err := st.Merge(p.Data); err != nil {
+		level.Error(s.logger).Log("msg", "failed to merge state", "err", err, "key", p.Key)
+		s.partialStateMergesFailed.WithLabelValues(p.Key).Inc()
+		return nil
```
pracucci, line 125: Why not returning error?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+	stateReplicationTotal    *prometheus.CounterVec
+	stateReplicationFailed   *prometheus.CounterVec
+
+	msgc   chan *clusterpb.Part
+	stopc  chan struct{}
+	readyc chan struct{}
+}
+
+func (s *state) Name() string {
+	return ""
```
pracucci, line 40: What's the impact of returning empty string?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (s *state) Address() string {
+	return ""
```
pracucci, line 44: No need to be implemented? If so, comment why.

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (s *state) Peers() []cluster.Node {
+	return nil
```
pracucci, line 48: No need to be implemented? If so, comment why.

pkg/alertmanager/alertmanager.go
```diff
@@ -122,11 +136,22 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
...
+	// 2) Alertmanager clustering with upstream Gossip
+	// 3) Alertmanager sharding
+	// #1 and #2 are covered by the 1st conditional, then the 2nd conditional covers #3
 	if cfg.Peer != nil {
 		c := cfg.Peer.AddState("nfl:"+cfg.UserID, am.nflog, am.registry)
 		am.nflog.SetBroadcast(c.Broadcast)
 	}
 
+	if cfg.Peer == nil && cfg.ShardingEnabled {
+		c := am.state.AddState("nfl:"+cfg.UserID, am.nflog)
```
pracucci, line 151: Why do we need duplicated logic here? I'm wondering if we could refactor `cfg.Peer` to be an interface and we set it either to the alertmanager clustering peer or `am.state` so that we keep the same logic between the two, but the underlying implementation changes.

pkg/alertmanager/multitenant.go
```diff
@@ -419,7 +426,8 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
 
 		am.grpcServer = server.NewServer(&handlerForGRPCServer{am: am})
 
-		am.distributor, err = NewDistributor(cfg.AlertmanagerClient, cfg.MaxRecvMsgSize, am.ring, nil, log.With(logger, "component", "AlertmanagerDistributor"), am.registry)
+		am.alertmanagerClientsPool = newAlertmanagerClientsPool(client.NewRingServiceDiscovery(am.ring), ClientConfig{}, logger, am.registry)
```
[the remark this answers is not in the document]
gotjosh: Good catch, this was a bad rebase.

pkg/alertmanager/alertmanager.go
```diff
@@ -189,14 +229,19 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 	}
 
 	am.dispatcherMetrics = dispatch.NewDispatcherMetrics(am.registry)
+	// TODO: Make sure the alertmanager is ready to receive requests and that we have gather notifications and silences from others at this point.
 	return am, nil
 }
 
+type PeerWait interface {
```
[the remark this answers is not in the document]
gotjosh: Good catch, I managed to get of rid it in favour of a callback

pkg/alertmanager/alertmanager.go
gotjosh: I thought about that but my initial impression is that would be too intense on the ring. We'd basically need to do a lookup for every alert that _tries_ to fire for each tenant across all replicas. 

This seemed like a good enough trade-off to test its term of behaviour and see where it puts us (as this a consistency trade-off)

pkg/alertmanager/state_replication.go
gotjosh: Alertmanager isn't so it seemed like overkill, happy to do it if you think otherwise.

pkg/alertmanager/state_replication.go
gotjosh: Good catch, I was thinking I wouldn't return any merge errors given these messages are supposed to come over UDP (we settle the gossip before sending an alert) but I think it's worth saying something here.

pkg/alertmanager/state_replication.go
gotjosh: Ditto from above.

pkg/alertmanager/multitenant.go
```diff
@@ -845,13 +876,81 @@ func (am *MultitenantAlertmanager) GetStatusHandler() StatusHandler {
...
+		if err != nil {
+			return err
+		}
+
+		switch resp.Status {
+		case alertmanagerpb.OK:
+			return nil
+		case alertmanagerpb.MERGE_ERROR:
+			level.Error(am.logger).Log("msg", "state replication failed", "user", userID, "key", part.Key, "err", resp.Error)
+			return nil
```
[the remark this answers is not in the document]
gotjosh: I don't think we should error here. [According to the algorithm](https://github.com/prometheus/alertmanager/blob/master/cluster/delegate.go#L150-L168) partial messages are sent over UDP - in this case, we don't care if the message gets lost as we'll settle the gossip before sending an alert.

The trade-off here will be eventual consistency on the API for silences (and maybe alert in the future?) but we're planning on providing a union of results for reads via the distributor.

pkg/alertmanager/state_replication.go
gotjosh: I added a comment on 

```
		// Cortex should not expose cluster information back to its tenants.
		Peer:     &state{},
```

Seems like my attempt to make it more visible was futile, I'll add one to these methods.

pkg/alertmanager/alertmanager.go
gotjosh: I think that's very reasonable, however, we're not done just yet. Given the next step is to implement the:

- Full state replication
- Persistence of state to object storage

I feel like we should wait until those two lands so that we can have an idea of how the final interface looks like.

pkg/alertmanager/multitenant.go
```diff
@@ -845,13 +876,81 @@ func (am *MultitenantAlertmanager) GetStatusHandler() StatusHandler {
...
+		case alertmanagerpb.USER_NOT_FOUND:
```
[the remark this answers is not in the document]
gotjosh: My comment about UDP messages and gossip settling apply here. This is a valid use-case that would happen due to a topology change and it wouldn't necessary be an error state.

pkg/alertmanager/multitenant.go
```diff
@@ -737,20 +744,44 @@ func (am *MultitenantAlertmanager) setConfig(cfg alerts.AlertConfigDesc) error {
 			return fmt.Errorf("unable to apply Alertmanager config for user %v: %v", cfg.User, err)
 		}
 	}
+
+	// Now that we're sure we have an alertmanager up and running, we need to make sure we update which position it
+	// holds in the ring.
+	if currentAM != nil && am.cfg.ShardingEnabled && am.ring.ReplicationFactor() > 1 {
```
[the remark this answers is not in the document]
gotjosh: There was a test that was screaming at me for this IIRC, but now it seems like it is not...

pkg/alertmanager/alertmanager.go
pracucci: > would be too intense on the ring

It's not. The ring is an in-memory structure. You can call `.Get()` as many times as you want (we do it thousands time / sec in the distributor).

gotjosh: TIL - This makes perfect sense.

pkg/alertmanager/multitenant.go
```diff
@@ -845,13 +888,82 @@ func (am *MultitenantAlertmanager) GetStatusHandler() StatusHandler {
 	}
 }
 
+// replicateStateForUser attempts to replicate a partial state sent by an alertmanager to its other replicas through the ring.
+func (am *MultitenantAlertmanager) replicateStateForUser(ctx context.Context, userID string, part *clusterpb.Part) error {
```
pracucci, line 892: The chain of calls leading to this function is a bit complicated. What if we move this function to `state`? You need to pass `selfAddress`, the ring client (not lifecycler) and the clients pool.

pkg/alertmanager/alertmanager.go
```diff
@@ -161,9 +200,10 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 		Alerts:     am.alerts,
 		Silences:   am.silences,
 		StatusFunc: am.marker.Status,
-		Peer:       cfg.Peer,
-		Registry:   am.registry,
-		Logger:     log.With(am.logger, "component", "api"),
+		// Cortex should not expose cluster information back to its tenants.
+		Peer:     &state{},
```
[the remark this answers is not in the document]
gotjosh: Fair, I'll create a [null object](https://en.wikipedia.org/wiki/Null_object_pattern)

pkg/alertmanager/alertmanager.go
```diff
@@ -152,6 +182,15 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 		am.wg.Done()
 	}()
 
+	if cfg.Peer == nil && cfg.ShardingEnabled {
+		// At this point, we're ready to ensure state can be sync by asking the state to settle.
+		am.wg.Add(1)
+		go func() {
+			go am.state.Settle(context.Background(), time.Second)
```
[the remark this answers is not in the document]
gotjosh: In my eyes, this seemed potentially dangerous. Making this call synchronous _might_ significantly slow down the syncing process of the alertmanager depending on how long it takes to settle so I opted for a separate goroutine - seem like one of those details that might be easy to miss can cause undesired side-effects. WDYT?

pkg/alertmanager/alertmanager.go
```diff
@@ -189,14 +229,15 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 	}
 
 	am.dispatcherMetrics = dispatch.NewDispatcherMetrics(am.registry)
+	// TODO: Make sure the alertmanager is ready to receive requests and that we have gather notifications and silences from others at this point.
```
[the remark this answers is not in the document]
gotjosh: I'll leave this note, I left it as a reminder that is not really until we call `ApplyConfig` that the alert manager is ready to receive requests but that is currently done synchronously at the moment.

pkg/alertmanager/alertmanager.go
gotjosh: Done, resolving this conversation.

pkg/alertmanager/alertmanager.go
gotjosh: Adding to this argument is the fact that `Peer`, as defined by upstream, is something that's global - it is always the same for _all_ the alertmanagers thus initialized in the multitenant manager. Whereas in this design `state` is something local to the tenent's alertmanager.

jtlisi:
Fixes #2650

pkg/alertmanager/alertmanager.go
gotjosh: I took this for a spin - is this better?

pkg/alertmanager/alertmanager.go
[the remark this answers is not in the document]
gotjosh: Is this better?

pkg/alertmanager/multitenant.go
```diff
@@ -846,13 +881,82 @@ func (am *MultitenantAlertmanager) GetStatusHandler() StatusHandler {
...
+			return nil
+		}
+	}, func() {})
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+
```
stevesg, line 921:
```suggestion
	return err
```

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+	for {
+		select {
+		case p := <-s.msgc:
+			// If the replication factor is <= 1, we don't need to replicate any state anywhere else.
+			if s.replicationFactor <= 1 {
+				return
+			}
+
+			s.stateReplicationTotal.WithLabelValues(p.Key).Inc()
+			ctx := context.Background() //TODO: I probably need a better context
```
stevesg, line 159: Did you plan to fix this one before merging, or was it intended for the next commit/commits?

gotjosh: It is intended for the next set of commits - I'm not sure whenever we need a timeout here or what are the implications. There might be a world where background is good enough but didn't want this to go unnoticed.

pkg/alertmanager/alertmanager.go
```diff
@@ -189,14 +239,16 @@ func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 	}
 
 	am.dispatcherMetrics = dispatch.NewDispatcherMetrics(am.registry)
+
+	am.state.WaitReady()
```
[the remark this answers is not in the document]
pracucci: We may remove `WaitReady()` call from this PR to unblock it and address it separately.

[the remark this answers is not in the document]
pstibrany: > _Generally speaking I believe it's a bad design "waiting for something" in a constructor function (like this `New()`) but I would like to better understand why we need it._

100%.

(Personally I would like to see this to follow our `Services` model, but I am not sure it makes sense here.)

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+		s.partialStateMergesFailed.WithLabelValues(p.Key).Inc()
+		return errors.Wrap(err, "unable to merge state")
+	}
+
+	return nil
+}
+
+// The clusterWait function uses the position to determine how long should it wait before sending a notification.
+// In the Cortex case, we copy the function directly and use the manager to figure out which position we hold.
+func (s *state) Position() int { return 0 }
```
pstibrany, line 120: Given the comment, does this method need to be part of `State` interface?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Settle waits until the alertmanagers are ready (and sets the appropriate internal state when it is).
+// The idea is that we don't want to start working" before we get a chance to know most of the notifications and/or silences.
+func (s *state) Settle(ctx context.Context, _ time.Duration) {
+	level.Info(s.logger).Log("msg", "Waiting for notification and silences to settle...")
+
+	// TODO: Make sure that the state is fully synchronised at this point.
```
pstibrany, line 127: Is this going to be implemented in this PR?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+	// We can check other alertmanager(s) and explicitly ask them to propagate their state to us if available.
+	close(s.readyc)
+}
+
+// WaitReady is needed for the pipeline builder to know whenever we've settled and the state is up to date.
+func (s *state) WaitReady() {
+	//TODO: At the moment, we settle in a separate go-routine (see multitenant.go as we create the Peer) we should
+	// mimic that behaviour here once we have full state replication.
+	s.Settle(context.Background(), time.Second)
```
pstibrany, line 136: `WaitReady` needs to be passed a context to solve this.

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+	return nil
+}
+
+// The clusterWait function uses the position to determine how long should it wait before sending a notification.
+// In the Cortex case, we copy the function directly and use the manager to figure out which position we hold.
+func (s *state) Position() int { return 0 }
+
+// Settle waits until the alertmanagers are ready (and sets the appropriate internal state when it is).
+// The idea is that we don't want to start working" before we get a chance to know most of the notifications and/or silences.
+func (s *state) Settle(ctx context.Context, _ time.Duration) {
```
pstibrany, line 124: Settle ignores context currently. It should not do that, and should report error back if context is finished before settling has finished.

pkg/alertmanager/state_replication.go
pstibrany, line 159: Hint: If `state` was a `services.Service` (= object with lifecycle in Cortex), it would already have its own context. :)

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+		replicationFactor:  rf,
+		states:             make(map[string]cluster.State, 2), // we use two, one for the notifications and one for silences.
+		msgc:               make(chan *clusterpb.Part),
+		readyc:             make(chan struct{}),
+		stopc:              stopc,
+		reg:                r,
+		partialStateMergesTotal: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+			Name: "alertmanager_partial_state_merges_total",
+			Help: "Number of times we have received a partial state to merge for a key.",
+		}, []string{"key"}),
```
pstibrany, line 58: Do we need per-key metrics or would per-user (i.e. per-state) be enough?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+		key:    key,
+		logger: logger,
+		send:   send,
+	}
+
+	return c
+}
+
+func (c *stateChannel) Broadcast(b []byte) {
+	//TODO: We probably need a bounded queue at some point to avoid a single tenant overflowing with messages
```
pstibrany, line 190: This shouldn't be responsibility of "stateChannel", but `replicateStateFunc` (ie.  `MultitenantAlertmanager.replicateStateForUser`), which can track queues per user and implement some fairness, if needed.

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+	level.Debug(c.logger).Log("msg", "message received for broadcasting", "key", c.key)
```
pstibrany, line 191: We can move this logging into `MultitenantAlertmanager.replicateStateForUser` and simplify this struct.

pkg/alertmanager/alertmanager_metrics.go
```diff
@@ -147,6 +152,22 @@ func newAlertmanagerMetrics() *alertmanagerMetrics {
 			"cortex_alertmanager_config_hash",
 			"Hash of the currently loaded alertmanager configuration.",
 			[]string{"user"}, nil),
+		partialMerges: prometheus.NewDesc(
+			"cortex_alertmanager_partial_state_merges_total",
+			"Number of times we have received a partial state to merge for a key.",
+			[]string{"key"}, nil),
```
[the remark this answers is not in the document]
gotjosh: We don't have user as part of these labels because `key` already includes the user.  It's a combination of `prefix + userID`. So by using key we "technically get both". Even though it breaks the nomenclature (of always using `user`).

pkg/alertmanager/state_replication.go
gotjosh: No, this one is not.

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+
+// MergePartialState merges a received partial message with an internal state.
+func (s *state) MergePartialState(p *clusterpb.Part) error {
+	s.partialStateMergesTotal.WithLabelValues(p.Key).Inc()
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	st, ok := s.states[p.Key]
+	if !ok {
+		level.Error(s.logger).Log("msg", "state key not found while merging", "key", p.Key)
```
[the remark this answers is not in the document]
gotjosh: Is this better?

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,193 @@
...
+		s.partialStateMergesFailed.WithLabelValues(p.Key).Inc()
+		return fmt.Errorf("unable to merge state, key '%s' not found while merging", p.Key)
```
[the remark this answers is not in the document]
gotjosh: Ditto.

pkg/alertmanager/state_replication.go
gotjosh: It does, I need to update the comment. We actually use the upstream implementation on that gossip-based clustering.

gotjosh: Thanks for pointing this out, it made me realise that a) I needed a test as part of this PR and b) that the above implementation is completely wrong.

pkg/alertmanager/multitenant_test.go
```diff
@@ -837,3 +842,175 @@ func TestAlertmanager_InitialSyncFailureWithSharding(t *testing.T) {
 	require.False(t, am.ringLifecycler.IsRegistered())
 	require.NotNil(t, am.ring)
 }
+
+// TODO: Test the position thing
```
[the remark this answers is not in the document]
gotjosh: Done! It was needed.

pkg/alertmanager/state_replication.go
gotjosh: Is this better?

pkg/alertmanager/state_replication.go
gotjosh: I'm leaving the implementation of `WaitReady` and `Settle` to the next PR - is it OK If we skip it for now? These two involve a full state replication and I fear we need the implementation of those to make sense of the big picture.

pkg/alertmanager/state_replication.go
gotjosh: Ditto from my comment below - Moreso, `WaitReady` as defined comes from upstream so we'd need to change that too.

pkg/alertmanager/state_replication.go
gotjosh: per-key includes both user and state so it would always be at most 2 per user.

pkg/alertmanager/alertmanager.go
```diff
@@ -42,6 +44,23 @@ import (
 
 const notificationLogMaintenancePeriod = 15 * time.Minute
 
+// NilPeer and NilChannel implements the Alertmanager clustering interface used by the API to expose cluster information.
```
pstibrany, line 47: Nit: I would move this type to the bottom of the file, to keep most important types on the top.

pkg/alertmanager/alertmanager_metrics.go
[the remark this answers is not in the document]
pstibrany: We have per-user registries, so we could do per-user output, aggregating over all keys. WDYT?

pkg/alertmanager/multitenant.go
pstibrany: If I understand it correctly, moving this to `state` can make it more difficult to implement fairness between different users.

pkg/alertmanager/state_replication.go
pstibrany: We don't need to copy `WaitReady` from upstream. We can define our own interface with what makes sense for us, and adapt upstream type to our interface via a wrapper.

Introducing context to upstream's `WaitReady` seems pretty straightforward too.

pkg/alertmanager/alertmanager.go
```diff
@@ -95,6 +123,13 @@ func init() {
 	}()
 }
 
+// State helps with replication and synchronization of notifications and silences across several alertmanager replicas.
+type State interface {
+	AddState(string, cluster.State, prometheus.Registerer) cluster.ClusterChannel
+	Position() int
+	WaitReady()
```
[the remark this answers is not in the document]
gotjosh: The implementation for `WaitReady` and `Settle` are part of the next PR, if you don't mind I'd like to leave it out of this PR for now.

pkg/alertmanager/state_replication.go
gotjosh: DItto from my comment above, it is certainly possible but I would like to avoid it as part of this PR - The implementation of `WaitReady` and `Settle` for the new replication method are yet to be determined.

pkg/alertmanager/alertmanager_metrics.go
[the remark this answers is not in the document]
gotjosh: Is that different from the current approach? An aggregation across all keys is not possible if the keys are per user e.g. `sil:user-3` or `nfl:user-2`.

pkg/alertmanager/state_replication.go
gotjosh: @pracucci had an idea where he wanted to move the implementation of `replicateStateFunc` to the state itself - we'll probably need to have this conversation then. is it OK if it stays here? I think it's worth considering whenever you want to implement it closer to where it is received as opposed to closer before it is replicated.

pkg/alertmanager/state_replication.go
gotjosh: Can you help me break this down? 

I have not seen us use services within services (That's different from services with sub-services) and even if we do, at its current state it seems like overkill. It's more code to manage the service, in particular, because the State itself it's satisfied by two different things (our `state` or the upstream `Peer`) which in my eyes brings us little benefits.

gotjosh:
Needs #3903

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,179 @@
...
+	s.stateReplicationTotal.WithLabelValues(key)
+	s.stateReplicationFailed.WithLabelValues(key)
+
+	send := func(b []byte) {
+		s.msgc <- &clusterpb.Part{Key: key, Data: b}
+	}
+
+	return &stateChannel{
+		send: send,
+	}
```
stevesg, line 95: More of a question than anything, now I've spent more time walking through this code. (Apologies if I'm missing something obvious here...)

Do you think we'll need the flexibility of having the `stateChannel` being passed a function in this way? Could the `stateChannel` just take `s.msgc` and `key`, and the `Broadcast` function be implemented concretely? (Illustrated this in a separate comment)

```suggestion
	return &stateChannel{
		msgc: s.msgc,
		key: key,
	}
```

pkg/alertmanager/state_replication.go
```diff
@@ -0,0 +1,179 @@
...
+// stateChannel allows a state publisher to send messages that will be broadcasted to all other alertmanagers that a tenant
+// belongs to.
+type stateChannel struct {
+	send func([]byte)
+}
+
+// Broadcast receives a message to be replicated by the state.
+func (c *stateChannel) Broadcast(b []byte) {
+	c.send(b)
+}
```
stevesg, line 179: (see last comment)

```suggestion
// stateChannel allows a state publisher to send messages that will be broadcasted to all other alertmanagers that a tenant
// belongs to.
type stateChannel struct {
	msgc chan *clusterpb.Part
	key  string
}

// Broadcast receives a message to be replicated by the state.
func (c *stateChannel) Broadcast(b []byte) {
	c.msgc <- &clusterpb.Part{Key: c.key, Data: b}
}
```

pkg/alertmanager/alertmanager_metrics.go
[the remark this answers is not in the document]
pstibrany: If I understand it correctly, we have per-user registries, which only use "key" label. During `alertmanagerMetrics.Collect`, when then call `SendSumOfCountersWithLabels` with `key` label. We could instead call `SendSumOfCountersPerUser` (eg. `data.SendSumOfCountersPerUser(out, m.partialMerges, "alertmanager_partial_state_merges_total")`), which would return `sum(alertmanager_partial_state_merges_total)` per user-registry, and then add `user` label to the output. I think that would be enough. WDYT?

pkg/alertmanager/state_replication.go
pstibrany: I'm only talking about `state` being a service (and not `Peer`), since it runs background process and its lifecycle needs to be managed.

Another possibility is to use custom context initialized in new, and canceled when stopping.

pkg/alertmanager/state_replication.go
pstibrany: ok, let's keep as-is in this PR.

pracucci (approved):

pstibrany (approved):
LGTM in current state, with caveat that sharding-support is still work in progress (don't use it yet! :)). Non-sharding code seems to be unaffected though.

Issue #3841, opened 2021-02-19
Compactor took 70 minutes to start work


bboreham:
**Describe the bug**
Compactor was stopped on one node by `kubectl drain` then moved to another node.
On startup, it did some kind of work for ~70 minutes before starting "real" work.

Compactor logs after startup:

```
level=info ts=2021-02-19T06:11:18.739889355Z caller=main.go:187 msg="Starting Cortex" version="(version=1.7.0-rc.0, branch=HEAD, revision=af2c64e)"
level=info ts=2021-02-19T06:11:18.740461621Z caller=server.go:229 http=[::]:80 grpc=[::]:9095 msg="server listening on addresses"
level=info ts=2021-02-19T06:11:18.742489959Z caller=module_service.go:58 msg=initialising module=memberlist-kv
level=info ts=2021-02-19T06:11:18.741951771Z caller=module_service.go:58 msg=initialising module=server
level=info ts=2021-02-19T06:11:18.742692682Z caller=module_service.go:58 msg=initialising module=compactor
level=info ts=2021-02-19T06:11:18.75085362Z caller=lifecycler.go:521 msg="not loading tokens from file, tokens file path is empty"
level=info ts=2021-02-19T06:11:18.751278858Z caller=compactor.go:317 component=compactor msg="waiting until compactor is ACTIVE in the ring"
level=info ts=2021-02-19T06:11:18.78962424Z caller=lifecycler.go:550 msg="instance not found in ring, adding with no tokens" ring=compactor
level=info ts=2021-02-19T06:11:18.804099438Z caller=lifecycler.go:397 msg="auto-joining cluster after timeout" ring=compactor
level=info ts=2021-02-19T06:11:20.508101255Z caller=compactor.go:331 component=compactor msg="waiting until compactor ring topology is stable" min_waiting=1m0s max_waiting=5m0s
level=info ts=2021-02-19T06:11:20.507975886Z caller=compactor.go:321 component=compactor msg="compactor is ACTIVE in the ring"
level=info ts=2021-02-19T06:12:20.508583011Z caller=blocks_cleaner.go:129 component=cleaner msg="started blocks cleanup and maintenance"
level=info ts=2021-02-19T06:12:20.508278927Z caller=compactor.go:335 component=compactor msg="compactor is ring topology is stable"
level=info ts=2021-02-19T06:12:24.10330928Z caller=blocks_cleaner.go:284 component=cleaner org_id=10445 msg="started blocks cleanup and maintenance"
[...]

PR #3842, opened 2021-02-19
Mention to disable marks migration in compactor production tips

pracucci:
**What this PR does**:
The issue #3841 made me realise we haven't added an entry about disabling global markers migration in the production tips, so this PR adds it.

N/A

**Checklist**
- [ ] Tests updated
- [x] Documentation added


gouthamve:
Just to understand it better, why is it not disabled by default?

pracucci:
> Just to understand it better, why is it not disabled by default?

@gouthamve The reason why we enabled it by default is to cover the case of a Cortex user upgrading from Cortex 1.6 to 1.7 and enabling the bucket index without accurately reading the CHANGELOG or doc.

pstibrany (approved):

PR #3843, opened 2021-02-19
Create an interface to access per-tenant limit overrides

56quarters:
**What this PR does**:

This change creates an interface to access per-tenant limits by
either user ID or in bulk. This allows the `Overrides` struct and
the overrides exporter to make use of the same functionality for
accessing limit overrides.

**Checklist**
- [x] Tests updated


pkg/util/validation/limits.go
```diff
@@ -183,9 +183,35 @@ func SetDefaultLimitsForYAMLUnmarshalling(defaults Limits) {
...
+type TenantLimits interface {
+	// ByUserID gets limits specific to a particular tenant or nil if there are none
+	ByUserID(userID string) *Limits
+
+	// AllByUserID gets a mapping of all tenant IDs and limits for that user
+	AllByUserID() map[string]*Limits
+}
+
+// MockTenantLimits exposes per-tenant limits based on a provided map. Useful for unit testing
+type MockTenantLimits struct {
```
pstibrany, line 196: This is only used by tests, so we can move it to _test.go file.

pkg/cortex/modules.go
```diff
@@ -165,22 +165,27 @@ func (t *Cortex) initRuntimeConfig() (services.Service, error) {
 }
 
 func (t *Cortex) initOverrides() (serv services.Service, err error) {
-	t.Overrides, err = validation.NewOverrides(t.Cfg.LimitsConfig, tenantLimitsFromRuntimeConfig(t.RuntimeConfig))
+	// If runtime configuration isn't enabled it doesn't make sense to configure per-tenant
+	// limits. Any consumers of TenantLimits will handle it being `nil` correctly.
+	if t.RuntimeConfig != nil {
```
pstibrany, line 170: If you want to be able to override `t.TenantLimits` in project based on Cortex, then we need to set `t.TenantLimits` in `initRuntimeConfig` (which also makes sense, since we only need to set it when actually using runtime config). `initOverrides` will only use, but not set, `t.TenantLimits`.

Then in your project, you can add dependency from `Overrides` to `CustomTenantLimits`, with its own init method, which will be called *before* `initOverrides`, so it has a chance to modify `t.TenantLimits`.

pstibrany (approved):
Nice job! I like how this interface simplifies runtime config. Please take a look at comment in `initOverrides` to make sure it works for you.

gouthamve (approved):
LGTM modulo Peter's comments. Further, can we set `t.TenantLimits = emptyTenantLimits{}` so that downstream users can avoid the nil check?

pstibrany:
> Further, can we set `t.TenantLimits = emptyTenantLimits{}` so that downstream users can avoid the nil check?

I think `nil` check is fine -- it's only at single place, while emptyTenantLimits requires entire new type with methods returning `nil`.

gouthamve (approved):
LGTM! Thanks!

PR #3849, opened 2021-02-22
Delete cached shuffle-shard subring when user is inactive.

pstibrany:
**What this PR does**: This PR implements deletion of cached shuffle-sharded subring in distributor, when user is inactive. In other components we don't need to do this, since we don't cache shuffle-sharded subrings in them.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -305,6 +305,8 @@ func (d *Distributor) running(ctx context.Context) error {
 }
 
 func (d *Distributor) cleanupMetricsForUser(userID string) {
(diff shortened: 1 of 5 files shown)


pkg/distributor/distributor.go
```diff
@@ -305,6 +305,8 @@ func (d *Distributor) running(ctx context.Context) error {
 }
 
 func (d *Distributor) cleanupMetricsForUser(userID string) {
```
pracucci, line 307: It doesn't cleanup only metrics now. We could rename the function for more clarity.

pracucci (approved):
LGTM (modulo a nit)

PR #3850, opened 2021-02-22
Apply jitter to the configured timeout for performing idle compactions.

### LINKED ISSUE #3832: Make TSDB idle-flush more dynamic by compacting only time range that cannot receive any samples.

stevesg:
**What this PR does**:
Attempt to prevent idle compaction from happening concurrently by
introducing a 25% jitter to the configured timeout. This jitter only
increases the timeout, such that the configured minimum timeout is
always adhered to.

Fixes #3832

**Checklist**
- [X] Tests updated
- [X] Documentation added - N/A


pkg/ingester/ingester_v2.go
```diff
@@ -1611,6 +1617,10 @@ func (i *Ingester) compactionLoop(ctx context.Context) error {
 	ticker := time.NewTicker(i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval)
 	defer ticker.Stop()
 
+	// Apply positive jitter only to ensure that the minimum timeout is adhered to.
+	i.TSDBState.compactionIdleTimeout = util.DurationWithPositiveJitter(i.cfg.BlocksStorageConfig.TSDB.HeadCompactionIdleTimeout, compactionIdleTimeoutJitter)
```
stevesg, line 1621: Up for discussion here: Should we change the timeout e.g. each time the timeout is hit? This would mean that if each ingester picked similar timeouts, then we we would diverge next time around. The tricky bit here, is that we would need a timeout per-user, as each user can hit the timeout at different times (depending on `lastUpdateTime`).

pkg/ingester/ingester_v2.go
```diff
@@ -355,6 +358,9 @@ type TSDBState struct {
 	forceCompactTrigger chan chan<- struct{}
 	shipTrigger         chan chan<- struct{}
 
+	// Timeout chosen for idle compactions.
+	compactionIdleTimeout time.Duration
```
stevesg, line 362: Should this be per-`userTSDB`? I figured it might be overkill, but open to opinions.

stevesg:
Rebase

pkg/ingester/ingester_v2.go
pracucci: A global should be fine.

pkg/ingester/ingester_v2.go
pracucci: I would start simple setting it at startup, as you did.

pracucci (commented):
Please updated the CLI flag description at `pkg/storage/tsdb/config.go` to mention a jitter is applied.

pracucci (approved):
LGTM, thanks!

stevesg:
Rebase

docs/blocks-storage/store-gateway.md
```diff
@@ -814,7 +814,9 @@ blocks_storage:
     # CLI flag: -blocks-storage.tsdb.head-compaction-concurrency
     [head_compaction_concurrency: <int> | default = 5]
 
-    # If TSDB head is idle for this duration, it is compacted. 0 means disabled.
+    # If TSDB head is idle for this duration, it is compacted. Note that up to
+    # 25% jitter is added to the value to avoid ingesters compacting
```
alvinlin123, line 818: I am just wondering, what is the reason to leave the jitter percentage unconfigurable  to the user?

alvinlin123 (approved):

docs/blocks-storage/store-gateway.md
stevesg: We didn't have a strong opinion here, other than following how jitter has typically been added in other places.

I don't think there would be opposition to be being configurable though.

pstibrany (approved):
LGTM, thanks!

docs/blocks-storage/store-gateway.md
alvinlin123: I see.  Make sense. I agree we can make it configurable when the need arises in a separate PR.

PR #3851, opened 2021-02-22
Cleanup meta sync files in compactor for un-owned users.

pstibrany:
**What this PR does**: This PR implements cleanup of meta-sync files on compactor, after compactor discovers that it no longer "owns" the given user.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"math/rand"
+	"os"
 	"path"
 	"strings"
 	"time"
@@ -480,6 +481,7 @@ func (c *Compactor) compactUsers(ctx context.Context) {
 		users[i], users[j] = users[j], users[i]
 	})
 
+	var skippedTenants []string
 	for _, userID := range users {
 		// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
 		if ctx.Err() != nil {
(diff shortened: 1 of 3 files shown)


pkg/compactor/compactor.go
```diff
@@ -520,6 +523,24 @@ func (c *Compactor) compactUsers(ctx context.Context) {
 		level.Info(c.logger).Log("msg", "successfully compacted user blocks", "user", userID)
 	}
 
+	// Delete local files for skipped tenants, if there are any.
+	for _, userID := range skippedTenants {
```
pracucci, line 527: This will delete data for tenants which don't belong anymore to the shard, but will keep data on disk if a tenant is deleted or a resharding occured while this compactor replica was not running (but disk is persisted). WDYT?

pstibrany: Resharding is not a problem I think, but deleted tenants are. I will fix that.

pkg/compactor/compactor_test.go
```diff
@@ -1192,3 +1193,88 @@ func TestAllowedUser(t *testing.T) {
...
+		require.NoError(t, inmem.Upload(context.Background(), userID+"/"+id.String()+"/meta.json", strings.NewReader(mockBlockMetaJSON(id.String()))))
+	}
+
+	// Create a shared KV Store
+	kvstore := consul.NewInMemoryClient(ring.GetCodec())
+
+	// Create two compactors
+	var compactors []*Compactor
+
+	for i := 1; i <= 2; i++ {
```
pracucci, line 1219: The two compactors should run with a different `DataDir` and the dir each compactor should be a temporarily dir removed once the tests ends.

pracucci (approved):
Good job! LGTM (modulo a couple of nits)

pkg/compactor/compactor_test.go
pstibrany: `prepare` function used in the loop already does that.

pkg/compactor/compactor.go
pstibrany: I think this is resolved. Please reopen if not.

pracucci: I agree it's solved now 👍

PR #3852, opened 2021-02-22
Rename Blocks Purger to Tenant Deletion.

pstibrany:
**What this PR does**: This PR just renames "Blocks Purger" to "Tenant Deletion". It doesn't do anything else. Really. I hope.

**Checklist**
- [x] Tests updated
- [na] Documentation added

### DIFF
--- a/pkg/cortex/modules.go
+++ b/pkg/cortex/modules.go
@@ -75,7 +75,7 @@ const (
 	StoreGateway             string = "store-gateway"
 	MemberlistKV             string = "memberlist-kv"
 	ChunksPurger             string = "chunks-purger"
-	BlocksPurger             string = "blocks-purger"
+	TenantDeletion           string = "blocks-purger"
 	Purger                   string = "purger"
 	QueryScheduler           string = "query-scheduler"
 	TenantFederation         string = "tenant-federation"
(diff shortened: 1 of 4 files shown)


pkg/cortex/modules.go
```diff
@@ -75,7 +75,7 @@ const (
 	StoreGateway             string = "store-gateway"
 	MemberlistKV             string = "memberlist-kv"
 	ChunksPurger             string = "chunks-purger"
-	BlocksPurger             string = "blocks-purger"
+	TenantDeletion           string = "blocks-purger"
```
pracucci, line 78: Can we also safely rename the value `blocks-purger`?

pracucci (approved):
LGTM with a question

pkg/cortex/modules.go
pstibrany: Yes. Given that it's marked as "invisible" to user and thus considered internal, I think that's perfectly fine.

bboreham:
#4370 seems to be undoing this rename.  Can you explain why it was done please, to help judge whether undoing it is a good idea?

pstibrany:
This PR was done because "blocks-purger" wasn't doing any "block purging". All "blocks purger" was doing was to accept API request to delete a tenant, and store marker file to the bucket. Hence "TenantDeletionAPI" seemed like better description.

Blocks cleanup logic itself is implemented in compactor instead.

PR #3856, opened 2021-02-23
Add functions to access cortexpb pool

colega:
**What this PR does**:
This commit adds functions to retrieve items from both poolsi in `cortexpb` package.

`ToWriteRequest` already exposes entities (slice of PreallocTimeseries and \*TimeSeries) retrieved from the pool, and encourages usage of ReuseSlice on them, however there was no way to obtain them without using that function.

If that package's user wants to build a WriteRequest itself (for example, if more than one Sample per metric is desired) then it's useful to retrieve those slice and struct pointers from the pool.

**Checklist**
- [x] Tests updated
- [ ] Documentation added (does not apply)
- [ ] `CHANGELOG.md` updated (not updated as this doesn't change the functionality)


gouthamve (approved):
LGTM!

colega:
(I force pushed to fix tests and use new functions in previous accesses to the pool)

jtlisi (approved):
LGTM

pracucci (approved):

PR #3860, opened 2021-02-23
Fixed TimeToMilliseconds() used in integration tests

pracucci:
**What this PR does**:
I've got an integration test failing because of the timestamp comparison (the expected timestamp was diff by 1 millis). After some investigation I've found out there 1 case where `TimeToMilliseconds()` used in integration tests doesn't actually match the rounding done in Prometheus. The case is when the nanoseconds part of the timestamp is like `xxx500000`.

This PR fixes it, using the same exact formula used by Prometheus. The test cases have been created running the Prometheus code.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/util.go
+++ b/integration/e2e/util.go
@@ -88,7 +88,8 @@ func PostRequest(url string) (*http.Response, error) {
 
 // TimeToMilliseconds returns the input time as milliseconds, using the same
 // formula used by Prometheus in order to get the same timestamp when asserting
-// on query results.
+// on query results. The formula we're mimicking here is Prometheus parseTime().
+// See: https://github.com/prometheus/prometheus/blob/df80dc4d3970121f2f76cba79050983ffb3cdbb0/web/api/v1/api.go#L1690-L1694
 func TimeToMilliseconds(t time.Time) int64 {
 	// Convert to seconds.
 	sec := float64(t.Unix()) + float64(t.Nanosecond())/1e9
--- a/integration/e2e/util.go
+++ b/integration/e2e/util.go
@@ -86,12 +86,21 @@ func PostRequest(url string) (*http.Response, error) {
 	return client.Post(url, "", strings.NewReader(""))
 }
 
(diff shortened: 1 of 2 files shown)


integration/e2e/util_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+func TestTimeToMilliseconds(t *testing.T) {
+	input := time.Unix(0, 1614091978491118000)
+	assert.Equal(t, int64(1614091978491), TimeToMilliseconds(input))
+
+	input = time.Unix(0, 1614091978505374000)
+	assert.Equal(t, int64(1614091978505), TimeToMilliseconds(input))
+
+	input = time.Unix(0, 1614091921796500000)
+	assert.Equal(t, int64(1614091921796), TimeToMilliseconds(input))
```
pracucci, line 18: This would fail in `master`.

MichelHollands (approved):
Nice catch

integration/e2e/util.go
```diff
@@ -86,12 +86,21 @@ func PostRequest(url string) (*http.Response, error) {
 	return client.Post(url, "", strings.NewReader(""))
 }
 
-// timeToMilliseconds returns the input time as milliseconds, using the same
+// TimeToMilliseconds returns the input time as milliseconds, using the same
```
jtlisi, line 89: **suggestion(optional):** To give context to future readers it may be worth adding a comment with a permalink to the [Prometheus parseTime function](https://github.com/prometheus/prometheus/blob/df80dc4d3970121f2f76cba79050983ffb3cdbb0/web/api/v1/api.go#L1690-L1694) this is referencing. There is a `FromTime` function in the Prometheus timestamp package that looks like it would accomplish the same thing but doesn't round the same way. However, I imagine someone reading may be unclear if the method below is up to date as Prometheus changes.

jtlisi (approved):
LGTM

integration/e2e/util.go
pracucci: Good point, done.

PR #3863, opened 2021-02-23
avoid allocating new label Builder in loop

nicolai86:
**What this PR does**: it avoids allocation a new `label.Builder` in the loop when `GetLabels` is called. This small change improves runtime for this particular method up to 50%:

### before

```
pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_SmallSet
BenchmarkGetLabels_SmallSet-8   	  273496	      3827 ns/op
PASS

pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_MediumSet
BenchmarkGetLabels_MediumSet-8   	     654	   1624579 ns/op
PASS
```

### after 

```
pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_SmallSet
BenchmarkGetLabels_SmallSet-8   	  407936	      2681 ns/op
PASS

pkg: github.com/cortexproject/cortex/pkg/util
BenchmarkGetLabels_MediumSet
BenchmarkGetLabels_MediumSet-8   	    1357	    853821 ns/op
PASS
```

**Checklist**
- [x] Tests updated
- ~[ ] Documentation added~
- ~[ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`~

### DIFF
--- a/pkg/util/metrics_helper_test.go
+++ b/pkg/util/metrics_helper_test.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"bytes"
+	"fmt"
 	"math/rand"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


pracucci:
Thanks @nicolai86. To better compare benchmarks, please do the following:
1. Run each benchmark with `-count=5` and store the output to a file (eg. `before.txt` and `after.txt`)
2. Install [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) and run `benchstat before.txt after.txt`

Benchstat will include statistical significance in the comparison.

pracucci (approved):
LGTM, thanks!

pkg/util/metrics_helper_test.go
```diff
@@ -999,6 +1000,56 @@ func uint64p(v uint64) *uint64 {
...
+	m.WithLabelValues("worst", "user1").Inc()
+
+	m.WithLabelValues("bad", "user2").Inc()
+	m.WithLabelValues("worst", "user2").Inc()
+
+	m.WithLabelValues("worst", "user3").Inc()
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		GetLabels(m, map[string]string{"user": "user1", "reason": "worse"})
```
pracucci, line 1022: The linter wants you to check the return error. To have you negative perf impact you can do:

```suggestion
		if err := GetLabels(m, map[string]string{"user": "user1", "reason": "worse"}); err != nil {
		    b.Fatal(err.Error())
		}
```

nicolai86:
```
❯ benchstat before.txt after.txt
name                   old time/op  new time/op  delta
GetLabels_SmallSet-8   3.67µs ± 3%  2.60µs ± 2%  -29.12%  (p=0.008 n=5+5)
GetLabels_MediumSet-8  1.66ms ±10%  0.79ms ± 8%  -52.65%  (p=0.008 n=5+5)
```

pkg/util/metrics_helper_test.go
nicolai86: updated in new commit. thanks

pracucci:
> benchstat before.txt after.txt

@nicolai86 What you provided is fine, thanks! For the next time, please also run tests with `-benchmem` so the memory allocations / allocated is compared as well.

Could you also sign your commit (DCO), otherwise we can't merge?

nicolai86:
thanks for the hand holding @pracucci .

signed off. also, here's an updated benchstat:

```
❯ benchstat before.txt after.txt
name                   old time/op    new time/op    delta
GetLabels_SmallSet-8     3.73µs ± 4%    2.75µs ± 7%  -26.31%  (p=0.000 n=10+10)
GetLabels_MediumSet-8    1.83ms ±16%    0.93ms ±15%  -49.23%  (p=0.000 n=10+10)

name                   old alloc/op   new alloc/op   delta
GetLabels_SmallSet-8     2.40kB ± 0%    1.20kB ± 0%  -50.00%  (p=0.000 n=10+10)
GetLabels_MediumSet-8     889kB ± 0%     169kB ± 0%  -81.00%  (p=0.000 n=9+6)

name                   old allocs/op  new allocs/op  delta
GetLabels_SmallSet-8       30.0 ± 0%      20.0 ± 0%  -33.33%  (p=0.000 n=10+10)
GetLabels_MediumSet-8     12.0k ± 0%      6.0k ± 0%  -49.96%  (p=0.000 n=10+10)
```

pstibrany (approved):
Thanks.

bboreham:
Would it be even better to put off building until we are sure this metric matches the filter?
On the assumption that the most common case it that it does not match.

nicolai86:
with my naive implementation - introducing a variable, creating and resetting only once `lbls` is used - it improves the case without match ~15%. 

However it adds one allocation again, which makes the case with matches slightly slower to what it is with the default allocation right now.

bboreham:
Don’t those numbers depend on how many labels are in each metric?

nicolai86:
It should - the number is just for the small set benchmark I added as part of this PR.

PR #3870, opened 2021-02-24
Refactoring: use PrefixedBucketClient in UserBucketClient

pracucci:
**What this PR does**:
This is just an internal refactoring PR to use `PrefixedBucketClient` in `UserBucketClient`.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/storage/bucket/prefixed_bucket_client.go
+++ b/pkg/storage/bucket/prefixed_bucket_client.go
@@ -3,7 +3,6 @@ package bucket
 import (
 	"context"
 	"io"
-	"path"
 	"strings"
 
 	"github.com/thanos-io/thanos/pkg/objstore"
@@ -23,7 +22,7 @@ func NewPrefixedBucketClient(bucket objstore.Bucket, prefix string) *PrefixedBuc
 }
 
 func (b *PrefixedBucketClient) fullName(name string) string {
-	return path.Join(b.prefix, name)
+	return b.prefix + objstore.DirDelim + name
 }
 
 // Close implements io.Closer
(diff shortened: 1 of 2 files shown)


pkg/storage/bucket/prefixed_bucket_client.go
```diff
@@ -23,7 +22,7 @@ func NewPrefixedBucketClient(bucket objstore.Bucket, prefix string) *PrefixedBuc
 }
 
 func (b *PrefixedBucketClient) fullName(name string) string {
-	return path.Join(b.prefix, name)
+	return b.prefix + objstore.DirDelim + name
```
pracucci, line 25: To keep the same exact logic we had in `UserBucketClient`. If name is "", the delimiter must be added anyway.

jtlisi: I see, sorry I made this change.

jtlisi (approved):
LGTM

pkg/storage/bucket/prefixed_bucket_client.go
pstibrany: Based on my suggestion, so sorry about that :) 

I'm not sure empty name should be a valid choice though and I wonder how do object stores work in that case.

pracucci: > I'm not sure empty name should be a valid choice
It's used to list the root. Thanos does it.

pstibrany: Ah right, I misunderstood and was thinking about empty prefix.

pstibrany (approved):

Issue #3669, opened 2021-01-08
Limit series per query in the blocks storage


pracucci:
**Is your feature request related to a problem? Please describe.**
The chunks storage support the `-ingester.max-series-per-query` limit, which allows to configure the maximum number of series for which a query can fetch samples from ingesters. It's not a full support to limit series per query, because the limit is enforced in the ingesters only.

Blocks storage doesn't support it at all.

**Describe the solution you'd like**
Ideally, we should limit series per query in the blocks storage, both when fetching samples from ingesters and long term storage. The store-gateway internally supports a series limit (we're currently not using it), but that would be enforced in a single store-gateway only. We could do it, plus enforced in ingesters and queriers, to have a real limit implementation.

PR #3873, opened 2021-02-24
Enforce series and sample limits on streaming queries to ingester from querier

bboreham:
**What this PR does**:
Count how many series and samples have been received by the querier, and stop processing if the limit is exceeded.

I check the sample limit only when samples are received (not chunks) and only after de-duplicating, which means the limit isn't very good protection against memory blow-up.

I also refactored the way `userID` is received by `Query()` and `QueryStream()` functions: I changed it from an indirect property of the `Context` to an explicit parameter. I needed `userID` to find the limits and couldn't bear adding another call to `tenant.TenantID()`.

I haven't renamed the CLI flags from `ingester.max...`; thought I would post this as a draft and see what people think.

Partial fix for #3669

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -2,6 +2,7 @@ package distributor
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"time"
 
@@ -167,7 +168,10 @@ func (d *Distributor) queryIngesters(ctx context.Context, replicationSet ring.Re
 }
 
 // queryIngesterStream queries the ingesters using the new streaming API.
-func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ring.ReplicationSet, req *ingester_client.QueryRequest) (*ingester_client.QueryStreamResponse, error) {
+func (d *Distributor) queryIngesterStream(ctx context.Context, userID string, replicationSet ring.ReplicationSet, req *ingester_client.QueryRequest) (*ingester_client.QueryStreamResponse, error) {
+	maxSeries := d.limits.MaxSeriesPerQuery(userID)
+	maxSamples := d.limits.MaxSamplesPerQuery(userID)
+
 	// Fetch samples from multiple ingesters
 	results, err := replicationSet.Do(ctx, d.cfg.ExtraQueryDelay, func(ctx context.Context, ing *ring.InstanceDesc) (interface{}, error) {
 		client, err := d.ingesterPool.GetClientFor(ing.Addr)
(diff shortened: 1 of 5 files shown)


pkg/distributor/query.go
```diff
@@ -19,15 +20,15 @@ import (
 )
 
 // Query multiple ingesters and returns a Matrix of samples.
-func (d *Distributor) Query(ctx context.Context, from, to model.Time, matchers ...*labels.Matcher) (model.Matrix, error) {
+func (d *Distributor) Query(ctx context.Context, userID string, from, to model.Time, matchers ...*labels.Matcher) (model.Matrix, error) {
```
pracucci, line 23: Shouldn't we enforce the limits here as well?

pkg/distributor/query.go
```diff
@@ -204,6 +203,10 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSet ri
 
 			result.Chunkseries = append(result.Chunkseries, resp.Chunkseries...)
 			result.Timeseries = append(result.Timeseries, resp.Timeseries...)
+
+			if len(result.Chunkseries) > maxSeries || len(result.Timeseries) > maxSeries {
+				return nil, fmt.Errorf("exceeded maximum number of series in a query (limit %d)", maxSeries)
```
pracucci, line 208: What are the implications on the returned status code? I've the feeling this may be detected as a storage error and we return a 5xx error (while it should be a 4xx) but I haven't deeply checked it.

pracucci (commented):
A couple of things:
- Please update the MaxSeriesPerQuery and MaxSamplesPerQuery  CLI flags description (and then `make doc`)
- Add a CHANGELOG entry

> I also refactored the way userID is received by Query() and QueryStream() functions

Makes sense to me.

pkg/distributor/query.go
bboreham: Yes; my focus was on things that blow up, which a range query is much more likely to.

pkg/distributor/query.go
bboreham: Do you think it would work better as an httpgrpc error?

pkg/distributor/query.go
pracucci: The `Query()` is used for range queries too, no? It's used when the "gRPC streaming" is disabled, while "QueryStream()" is called when it's enabled.

bboreham: Oh, sorry, I got confused.

For non-streaming queries and chunks, ingester enforces the limits already.

pkg/distributor/query.go
pracucci: @bboreham I don't remember how the error code propagation works in detail but I would start testing it. Do you have time/interest to work on it, otherwise we can takeover, cause we're interested into this limit too.

pracucci: I checked it and we have to return a `validation.LimitError`.

pkg/distributor/query.go
pracucci: > For non-streaming queries and chunks, ingester enforces the limits already.

Where is it done? I can't find it.

bboreham:
The `max-fetched-chunks-per-query` limit from #4125 is close to a limit on samples - each chunk is max 120 samples - so this limit is not needed in practice.

PR #3877, opened 2021-02-25
Added metrics to track partitioner behaviour

pracucci:
**What this PR does**:
We currently don't have any visibility on the partitioner behaviour (how much ranges are expanded and how many bucket API requests we actually save). This PR adds some metrics that should help us to better understand how it works.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pstibrany (approved):
Nice, I'm looking forward to see what the real numbers from this will look like.

(It would also be interesting to see how much bytes we fetch extra due to used 16000 bytes chunk size)

pracucci:
> (It would also be interesting to see how much bytes we fetch extra due to used 16000 bytes chunk size)

Yeah, we can't track this in Cortex. Will track it in Thanos. This will be surprising, I'm sure :)

PR #3879, opened 2021-02-25
Per-tenant retention periods

### LINKED ISSUE #3622: Feature request: be able to specify the retention time for blocks

stevesg:
**What this PR does**:
Implement per-tenant retention periods by performing block deletion in the Compactor.

Note that I've split the change into three commits to make reviewing it easier:
1) Add the new configuration
2) Add the new metrics
3) The change + unit tests

Fixes #3622 

**Checklist**
- [X] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
 * [CHANGE] Store-gateway: the chunks pool controlled by `-blocks-storage.bucket-store.max-chunk-pool-bytes` is now shared across all tenants. #3830
 * [CHANGE] Ingester: return error code 400 instead of 429 when per-user/per-tenant series/metadata limits are reached. #3833
-* [CHANGE] Compactor: add `reason` label to `cortex_compactor_blocks_marked_for_deletion_total` metric. The existing total is labelled as `compaction`. Deletion marks due to blocks passing the retention period are labelled as `retention`. #3879
+* [CHANGE] Compactor: add `reason` label to `cortex_compactor_blocks_marked_for_deletion_total` metric. Source blocks marked for deletion by compactor are labelled as `compaction`, while blocks passing the retention period are labelled as `retention`. #3879
 * [FEATURE] Experimental Ruler Storage: Add a separate set of configuration options to configure the ruler storage backend under the `-ruler-storage.` flag prefix. All blocks storage bucket clients and the config service are currently supported. Clients using this implementation will only be enabled if the existing `-ruler.storage` flags are left unset. #3805 #3864
 * [FEATURE] Adds support to S3 server-side encryption using KMS. The S3 server-side encryption config can be overridden on a per-tenant basis for the blocks storage and ruler. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651 #3810 #3811 #3870
   - `-<prefix>.s3.sse.type`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
     * `-alertmanager.alertmanager-client.tls-ca-path`
     * `-alertmanager.alertmanager-client.tls-server-name`
     * `-alertmanager.alertmanager-client.tls-insecure-skip-verify`
+* [FEATURE] Compactor: implement per-tenant retention periods by performing block deletion in the Compactor. This is configured via `-compactor.retention-period`, and can be overriden on a per-tenant basis. #3879
 * [ENHANCEMENT] Ruler: Add TLS and explicit basis authentication configuration options for the HTTP client the ruler uses to communicate with the alertmanager. #3752
   * `-ruler.alertmanager-client.basic-auth-username`: Configure the basic authentication username used by the client. Takes precedent over a URL configured username.
   * `-ruler.alertmanager-client.basic-auth-password`: Configure the basic authentication password used by the client. Takes precedent over a URL configured password.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
 * [CHANGE] Store-gateway: the chunks pool controlled by `-blocks-storage.bucket-store.max-chunk-pool-bytes` is now shared across all tenants. #3830
 * [CHANGE] Ingester: return error code 400 instead of 429 when per-user/per-tenant series/metadata limits are reached. #3833
+* [CHANGE] Compactor: add `reason` label to `cortex_compactor_blocks_marked_for_deletion_total` metric. The existing total is labelled as `compaction`. Deletion marks due to blocks passing the retention period are labelled as `retention`. #3879
 * [FEATURE] Experimental Ruler Storage: Add a separate set of configuration options to configure the ruler storage backend under the `-ruler-storage.` flag prefix. All blocks storage bucket clients and the config service are currently supported. Clients using this implementation will only be enabled if the existing `-ruler.storage` flags are left unset. #3805 #3864
 * [FEATURE] Adds support to S3 server-side encryption using KMS. The S3 server-side encryption config can be overridden on a per-tenant basis for the blocks storage and ruler. Deprecated `-<prefix>.s3.sse-encryption`, you should use the following CLI flags that have been added. #3651 #3810 #3811 #3870
   - `-<prefix>.s3.sse.type`
@@ -26,7 +27,7 @@
     * `-alertmanager.alertmanager-client.tls-ca-path`
     * `-alertmanager.alertmanager-client.tls-server-name`
     * `-alertmanager.alertmanager-client.tls-insecure-skip-verify`
-* [FEATURE] Compactor: implement per-tenant retention periods by performing block deletion in the Compactor. This is configured via `-compactor.retention-period`, and can be overriden on a per-tenant basis. #3879
+* [FEATURE] Compactor: added blocks storage per-tenant retention support. This is configured via `-compactor.retention-period`, and can be overridden on a per-tenant basis. #3879
 * [ENHANCEMENT] Ruler: Add TLS and explicit basis authentication configuration options for the HTTP client the ruler uses to communicate with the alertmanager. #3752
   * `-ruler.alertmanager-client.basic-auth-username`: Configure the basic authentication username used by the client. Takes precedent over a URL configured username.
   * `-ruler.alertmanager-client.basic-auth-password`: Configure the basic authentication password used by the client. Takes precedent over a URL configured password.
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -90,8 +90,8 @@ func NewBlocksCleaner(cfg BlocksCleanerConfig, bucketClient objstore.Bucket, use
 			Help: "Total number of blocks failed to be deleted.",
 		}),
 		blocksMarkedForDeletion: promauto.With(reg).NewCounter(prometheus.CounterOpts{
-			Name:        "cortex_compactor_blocks_marked_for_deletion_total",
-			Help:        "Total number of blocks marked for deletion in compactor.",
+			Name:        blocksMarkedForDeletionName,
+			Help:        blocksMarkedForDeletionHelp,
 			ConstLabels: prometheus.Labels{"reason": "retention"},
 		}),
 
@@ -322,9 +322,9 @@ func (c *BlocksCleaner) cleanUser(ctx context.Context, userID string, firstRun b
 	// The trade-off being that retention is not applied if the index has to be
 	// built, but this is rare.
 	if idx != nil {
-		if err := c.applyUserRetentionPeriod(ctx, idx, userID, userBucket, userLogger); err != nil {
-			level.Warn(userLogger).Log("msg", "failed to apply user retention period ", "err", err)
-		}
+		// We do not want to stop the remaining work in the cleaner if an
+		// error occurs here. Errors are logged in the function.
+		c.applyUserRetentionPeriod(ctx, idx, userID, userBucket, userLogger)
 	}
 
 	// Generate an updated in-memory version of the bucket index.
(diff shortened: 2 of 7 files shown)


pkg/compactor/blocks_cleaner_test.go
```diff
@@ -406,6 +410,231 @@ func TestBlocksCleaner_ShouldRemoveMetricsForTenantsNotBelongingAnymoreToTheShar
...
+
+	ctx := context.Background()
+	logger := log.NewNopLogger()
+	reg := prometheus.NewPedanticRegistry()
+	scanner := tsdb.NewUsersScanner(bucketClient, tsdb.AllUsers, logger)
+	cfgProvider := newMockConfigProvider()
+
+	cleaner := NewBlocksCleaner(cfg, bucketClient, scanner, cfgProvider, logger, reg)
+
+	assertBlocksExist := func(exists1, exists2, exists3, exists4 bool) {
```
pracucci, line 502: Mixed feeling about this function. Maybe it's easier to read when you call it, if you change it to pick in input only 1 block, and you call `assertBlockExists(block meta, exists bool)` multiple times, one per block

pkg/compactor/blocks_cleaner_test.go
```diff
@@ -406,6 +410,231 @@ func TestBlocksCleaner_ShouldRemoveMetricsForTenantsNotBelongingAnymoreToTheShar
...
+		cortex_bucket_blocks_marked_for_deletion_count{user="user-1"} 0
+		cortex_bucket_blocks_marked_for_deletion_count{user="user-2"} 0
+		# HELP cortex_compactor_blocks_marked_for_deletion_total Total number of blocks marked for deletion in compactor.
+		# TYPE cortex_compactor_blocks_marked_for_deletion_total counter
+		cortex_compactor_blocks_marked_for_deletion_total{reason="retention"} 0
+	`),
+		"cortex_bucket_blocks_count",
+		"cortex_bucket_blocks_marked_for_deletion_count",
+		"cortex_compactor_blocks_marked_for_deletion_total",
+	))
```
pracucci, line 547: Showing as an example here. When you have multiple tests in the same function and you're not using the testing table approach, then may be easier to read wrapping into blocks like:

```suggestion
{
	cfgProvider.UserRetentionPeriods["user-1"] = 0
	cfgProvider.UserRetentionPeriods["user-2"] = 0

	require.NoError(t, cleaner.cleanUsers(ctx, false))
	assertBlocksExist(true, true, true, true)

	assert.NoError(t, prom_testutil.GatherAndCompare(reg, strings.NewReader(`
		# HELP cortex_bucket_blocks_count Total number of blocks in the bucket. Includes blocks marked for deletion, but not partial blocks.
		# TYPE cortex_bucket_blocks_count gauge
		cortex_bucket_blocks_count{user="user-1"} 2
		cortex_bucket_blocks_count{user="user-2"} 2
		# HELP cortex_bucket_blocks_marked_for_deletion_count Total number of blocks marked for deletion in the bucket.
		# TYPE cortex_bucket_blocks_marked_for_deletion_count gauge
		cortex_bucket_blocks_marked_for_deletion_count{user="user-1"} 0
		cortex_bucket_blocks_marked_for_deletion_count{user="user-2"} 0
		# HELP cortex_compactor_blocks_marked_for_deletion_total Total number of blocks marked for deletion in compactor.
		# TYPE cortex_compactor_blocks_marked_for_deletion_total counter
		cortex_compactor_blocks_marked_for_deletion_total{reason="retention"} 0
	`),
		"cortex_bucket_blocks_count",
		"cortex_bucket_blocks_marked_for_deletion_count",
		"cortex_compactor_blocks_marked_for_deletion_total",
	))
}
```

pracucci (commented):
Very good job 👏 👏 👏  The PR LGTM! I left few nits, but everything should be quick to address 🚀

pkg/compactor/blocks_cleaner.go
```diff
@@ -391,3 +408,58 @@ func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map
...
+	if failed > 0 {
+		return errors.Errorf("failed to mark %d blocks for deletion", failed)
+	}
+
+	return nil
+}
+
+// listBlocksOutsideRetentionPeriod determines the blocks which have aged past
+// the specified retention period, and are not already marked for deletion.
+func listBlocksOutsideRetentionPeriod(idx *bucketindex.Index, now time.Time, retention time.Duration) (result bucketindex.Blocks) {
```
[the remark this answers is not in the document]
stevesg: I want as much logic as possible in the function for sake of testing. This puts less of a requirement on testing the calling code, which is harder to validate.

pkg/compactor/blocks_cleaner.go
```diff
@@ -391,3 +408,58 @@ func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map
 		level.Info(userLogger).Log("msg", "deleted partial block marked for deletion", "block", blockID)
 	}
 }
+
+// applyUserRetentionPeriod marks blocks for deletion which have aged past the retention period.
+func (c *BlocksCleaner) applyUserRetentionPeriod(ctx context.Context, idx *bucketindex.Index, userID string, userBucket *bucket.UserBucketClient, userLogger log.Logger) error {
+	retention := c.cfgProvider.CompactorRetentionPeriod(userID)
```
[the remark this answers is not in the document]
stevesg: I've put this behavior at the top of `listBlocksOutsideRetentionPeriod` to make the test for this case easier.

pkg/compactor/blocks_cleaner.go
```diff
@@ -391,3 +408,58 @@ func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map
...
+	// a warning log message. Use the block deletion marks already in-memory
+	// to prevent marking blocks already marked for deletion.
+	marked := make(map[ulid.ULID]struct{})
+	for _, d := range idx.BlockDeletionMarks {
+		marked[d.ID] = struct{}{}
+	}
+
+	for _, b := range idx.Blocks {
+		maxTime := time.Unix(b.MaxTime/1000, 0)
+		if now.After(maxTime.Add(retention)) {
```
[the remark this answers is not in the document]
stevesg: I assume the reason for this was to avoid subtracting from a timestamps, which is problematic with unsigned timestamps. But we can change it.

pkg/compactor/blocks_cleaner.go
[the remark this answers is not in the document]
pracucci: Testability will not change.

pkg/compactor/blocks_cleaner.go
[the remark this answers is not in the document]
pracucci: I don't get it, sorry. What's the problem with `.Before()`?

pkg/compactor/blocks_cleaner_test.go
stevesg: I agree, the meaning of each true/false is not clear. A nice alternative would be to list the blocks explicitly:

```
assertBlocksExist(block1, block2, block3)
```

pkg/compactor/blocks_cleaner.go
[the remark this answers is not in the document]
stevesg: I think it does - the testing of `cleanBlocks` is "hard" - it requires external resources (files/blocks) and is non-deterministic due to the use of `time.Now`. The testing of `listBlocksOutsideRetentionPeriod` does not have these issues.

The point is that the calling code can be "dumb" - it simply has to pass the timestamp and the configuration.

stevesg:
Rebase

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [CHANGE] Query-frontend: removed `-querier.split-queries-by-day` (deprecated in Cortex 0.4.0). You should use `-querier.split-queries-by-interval` instead. #3813
 * [CHANGE] Store-gateway: the chunks pool controlled by `-blocks-storage.bucket-store.max-chunk-pool-bytes` is now shared across all tenants. #3830
 * [CHANGE] Ingester: return error code 400 instead of 429 when per-user/per-tenant series/metadata limits are reached. #3833
+* [CHANGE] Compactor: add `reason` label to `cortex_compactor_blocks_marked_for_deletion_total` metric. The existing total is labelled as `compaction`. Deletion marks due to blocks passing the retention period are labelled as `retention`. #3879
```
pracucci, line 10:
```suggestion
* [CHANGE] Compactor: add `reason` label to `cortex_compactor_blocks_marked_for_deletion_total` metric. Source blocks marked for deletion by compactor are labelled as `compaction`, while blocks passing the retention period are labelled as `retention`. #3879
```

pkg/util/validation/limits.go
```diff
@@ -144,6 +147,8 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.RulerMaxRulesPerRuleGroup, "ruler.max-rules-per-rule-group", 0, "Maximum number of rules per rule group per-tenant. 0 to disable.")
 	f.IntVar(&l.RulerMaxRuleGroupsPerTenant, "ruler.max-rule-groups-per-tenant", 0, "Maximum number of rule groups per-tenant. 0 to disable.")
 
+	f.DurationVar(&l.CompactorRetentionPeriod, "compactor.retention-period", 0, "Delete blocks containing samples older than the specified retention period. 0 to disable.")
```
pracucci, line 150: The CLI flag should match the YAML field name.

```suggestion
	f.DurationVar(&l.CompactorRetentionPeriod, "compactor.blocks-retention-period", 0, "Delete blocks containing samples older than the specified retention period. 0 to disable.")
```

pkg/util/validation/limits.go
```diff
@@ -86,6 +86,9 @@ type Limits struct {
 	// Store-gateway.
 	StoreGatewayTenantShardSize int `yaml:"store_gateway_tenant_shard_size" json:"store_gateway_tenant_shard_size"`
 
+	// Compactor.
+	CompactorRetentionPeriod time.Duration `yaml:"compactor_blocks_retention_period" json:"compactor_blocks_retention_period"`
```
pracucci, line 90:
```suggestion
	CompactorBlocksRetentionPeriod time.Duration `yaml:"compactor_blocks_retention_period" json:"compactor_blocks_retention_period"`
```

pkg/util/validation/limits.go
```diff
@@ -415,6 +420,11 @@ func (o *Overrides) EvaluationDelay(userID string) time.Duration {
 	return o.getOverridesForUser(userID).RulerEvaluationDelay
 }
 
+// CompactorRetentionPeriod returns the retention period for a given user.
+func (o *Overrides) CompactorRetentionPeriod(userID string) time.Duration {
```
pracucci, line 424:
```suggestion
func (o *Overrides) CompactorBlocksRetentionPeriod(userID string) time.Duration {
```

pkg/compactor/blocks_cleaner.go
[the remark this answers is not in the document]
pracucci: If you pass a threshold which is a `time.Time` and you will compare it against the block `MaxTime` they will be both deterministic. Anyway, it's not a big deal.

pkg/compactor/blocks_cleaner.go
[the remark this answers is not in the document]
pracucci: I see, but the log below will be misleading ("applying retention"). I think we should also check if retention <= 0 to circuit break.

pkg/compactor/blocks_cleaner.go
```diff
@@ -391,3 +408,50 @@ func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map
...
+
+	level.Debug(userLogger).Log("msg", "applying retention", "retention", retention.String())
+	blocks := listBlocksOutsideRetentionPeriod(idx, now, retention)
+
+	// Attempt to mark all blocks. It is not critical if a marking fails, as
+	// the cleaner will retry applying the retention in its next cycle.
+	for _, b := range blocks {
+		level.Info(userLogger).Log("msg", "applied retention: marking block for deletion", "block", b.ID, "maxTime", b.MaxTime)
+		if err := block.MarkForDeletion(ctx, userLogger, userBucket, b.ID, fmt.Sprintf("block exceeding retention of %v", retention), c.blocksMarkedForDeletion); err != nil {
+			level.Warn(userLogger).Log("msg", "failed to mark block", "block", b.ID, "err", err)
```
pracucci, line 425: We should mention "mark for what..." (eg. in Thanos there's also a marker for "no compaction").

```suggestion
			level.Warn(userLogger).Log("msg", "failed to mark block for deletion", "block", b.ID, "err", err)
```

pracucci (approved):
Good job, LGTM! (modulo a couple of nits)

pkg/compactor/blocks_cleaner.go
```diff
@@ -391,3 +408,49 @@ func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map
 		level.Info(userLogger).Log("msg", "deleted partial block marked for deletion", "block", blockID)
 	}
 }
+
+// applyUserRetentionPeriod marks blocks for deletion which have aged past the retention period.
+func (c *BlocksCleaner) applyUserRetentionPeriod(ctx context.Context, idx *bucketindex.Index, userID string, userBucket *bucket.UserBucketClient, userLogger log.Logger) {
```
pstibrany, line 413: Nit: it feels like this method is taking too many reduntant parameters. It could create userLogger and userBucket on its own, since it already has userID.

Alternatively we could remove userID and pass it retention directly, since we're already passing userBucket and logger -- no more need for userID.

pkg/compactor/blocks_cleaner.go
```diff
@@ -391,3 +408,49 @@ func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map
...
+		level.Info(userLogger).Log("msg", "applied retention: marking block for deletion", "block", b.ID, "maxTime", b.MaxTime)
+		if err := block.MarkForDeletion(ctx, userLogger, userBucket, b.ID, fmt.Sprintf("block exceeding retention of %v", retention), c.blocksMarkedForDeletion); err != nil {
+			level.Warn(userLogger).Log("msg", "failed to mark block for deletion", "block", b.ID, "err", err)
+		}
+	}
+}
+
+// listBlocksOutsideRetentionPeriod determines the blocks which have aged past
+// the specified retention period, and are not already marked for deletion.
+func listBlocksOutsideRetentionPeriod(idx *bucketindex.Index, now time.Time, retention time.Duration) (result bucketindex.Blocks) {
```
pstibrany, line 436: Nit: There is no need to pass both `now` and `retention` to this function. The only value that we really need here is "now - retention".

pkg/compactor/compactor.go
```diff
@@ -34,6 +34,11 @@ import (
 	"github.com/cortexproject/cortex/pkg/util/services"
 )
 
+const (
+	blocksMarkedForDeletionName = "cortex_compactor_blocks_marked_for_deletion_total"
```
pstibrany, line 38: I'm not sure mixing this metric between compactor and blocks cleaner is a good idea. It it perhaps time for BlocksCleaner to be a separate component from Compactor (not necessarily run separately from compactor, but top-level component in Cortex), as it's doing more and more work. It's unfortunate that blocks cleaner metrics start with "cortex_compactor_" prefix. (But that's a topic for a separate PR.)

pstibrany (approved):
LGTM, nice job. Thank you!

pkg/compactor/blocks_cleaner.go
stevesg: The only reason for doing this was so that the unit test for the function had coverage of the subtraction, but I'm happy to change it as I think @pracucci has the same comment.

pkg/compactor/blocks_cleaner.go
stevesg: I mainly followed the pattern of cleanUserPartialBlocks here:

```
func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map[ulid.ULID]error, idx *bucketindex.Index, userBucket *bucket.UserBucketClient, userLogger log.Logger) {
```

For consistency, I think I prefer the option of passing retention, and removing `userID`. What do you think?

pkg/compactor/compactor.go
stevesg: It could be a separate metric, I don't have a particularly strong opinion. Thoughts @pracucci ?

I guess the question is whether the BlocksCleaner is an implementation detail of the Compactor, or a service in it's own right. I think the reasoning behind putting it in the BlocksCleaner was to avoid having another service to manage.

stevesg: Leaving as-is for now.

pstibrany (approved):
Thanks.

PR #3888, opened 2021-02-26
AlertStore: implementation using thanos objstore bucket client

pracucci:
**What this PR does**:
Similarly to #3805, in this PR I'm introducing the Thanos object store client support to the alertmanager backend clients. The config structure follows the same pattern used for the ruler in #3805.

**Important**: the `BucketAlertStore` is not using `UserBucketClient` yet (so it's not supporting S3 SSE per-tenant overrides) because it will need some changes, due to the structure of the alertmanager config in the bucket. I prefer to do it in a separate PR to keep the diff easier to review.

The `ListAlertConfigs` is not optimised (configs are loaded 1-by-1). I plan to optimise it in a follow up PR. The problem is that the caller is not optimised (we load all tenants configs and then we filter by shard's ownership). I will fix this in a separate PR.

_I suggest to review changes with "Hide whitespace changes" enabled._

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/alertmanager/alertstore/bucketclient/bucket_client.go
```diff
@@ -0,0 +1,110 @@
...
+// ListAlertConfigs implements alertstore.AlertStore.
+func (s *BucketAlertStore) ListAlertConfigs(ctx context.Context) (map[string]alertspb.AlertConfigDesc, error) {
+	cfgs := map[string]alertspb.AlertConfigDesc{}
+
+	err := s.bucket.Iter(ctx, "", func(key string) error {
+		userID := key
+
+		cfg, err := s.getAlertConfig(ctx, userID)
+		if err != nil {
+			return errors.Wrapf(nil, "failed to fetch alertmanager config for user %s", userID)
```
stevesg, line 47: I think `Wrapf(nil, ...`) will always return `nil` here - https://pkg.go.dev/github.com/pkg/errors#Wrapf

Did you mean `Errorf()` here?

pkg/alertmanager/alertstore/config.go
```diff
@@ -51,31 +59,79 @@ func (cfg *Config) Validate() error {
...
-	switch cfg.Type {
-	case "configdb":
+// IsDefaults returns true if the storage options have not been set.
+func (cfg *LegacyConfig) IsDefaults() bool {
+	return cfg.Type == ConfigDB && cfg.ConfigDB.ConfigsAPIURL.URL == nil
+}
+
+// NewLegacyAlertStore returns a new alertmanager storage backend poller and store
+func NewLegacyAlertStore(cfg LegacyConfig, logger log.Logger) (AlertStore, error) {
+	if cfg.Type == "configdb" {
```
stevesg, line 69: Use the `ConfigDB` constant here?

pkg/alertmanager/alertstore/config.go
```diff
@@ -15,10 +18,15 @@ import (
 	"github.com/cortexproject/cortex/pkg/chunk/azure"
 	"github.com/cortexproject/cortex/pkg/chunk/gcp"
 	"github.com/cortexproject/cortex/pkg/configs/client"
+	"github.com/cortexproject/cortex/pkg/storage/bucket"
 )
 
-// Config configures the alertmanager backend
-type Config struct {
+const (
+	ConfigDB = "configdb"
```
stevesg, line 25: Does this need to be exposed?

pkg/alertmanager/alertstore/config.go
```diff
@@ -51,31 +59,79 @@ func (cfg *Config) Validate() error {
...
+	}
+
+	bucketClient, err := bucket.NewClient(ctx, cfg.Config, "alertmanager-storage", logger, reg)
 	if err != nil {
 		return nil, err
 	}
-	return objectclient.NewAlertStore(client), nil
+
+	store := bucketclient.NewBucketAlertStore(bucketClient, cfgProvider, logger)
+	if err != nil {
```
stevesg, line 132: `NewBucketAlertStore` doesn't return an error - no need to check `err`.

Can just `return  bucketclient.NewBucketAlertStore(...), nil`

pkg/alertmanager/alertstore/bucketclient/bucket_client.go
pracucci: Damn, good job spotting it! I meant `Wrapf(err, ...)`.

pkg/alertmanager/alertstore/config.go
pracucci: No, it doesn't.

pkg/alertmanager/multitenant_test.go
```diff
@@ -43,38 +47,6 @@ receivers:
   - name: dummy`
 )
 
-// basic easily configurable mock
-type mockAlertStore struct {
```
pracucci, line 47: Removed in favour of using a real implementation (the one backed by filesystem) or the `bucket.ClientMock` when we need to mock the backend (eg. simulate a failure).

stevesg (approved):
LGTM!

pkg/alertmanager/multitenant_test.go
jtlisi: 👍

pkg/alertmanager/alertstore/store.go
```diff
@@ -20,3 +34,59 @@ type AlertStore interface {
...
+		c, err := client.New(cfg.ConfigDB)
+		if err != nil {
+			return nil, err
+		}
+		return configdb.NewStore(c), nil
+	}
+
+	if cfg.Backend == local.Name {
+		return local.NewStore(cfg.Local)
+	}
```
jtlisi, line 84: **thought:** When I updated the ruler client, I didn't port over the local implementation, just the configdb. My main thought was that it is readonly and mainly used for testing and the new client will already have a filesystem/inmem implementation out of the box. I figured I'd leave it in the legacy config for the time being to avoid confusions with the `filesystem` backend. However, I'm fine keeping it since there are some desirable properties to a statically configured read-only Alertmanager/RuleStorage backend. Up to you if you think it's worth porting over to the config style or if we should deprecate it when we remove the old config in the future.

If we add it here, we should also port it over to the new ruler config backend as well.

jtlisi (approved):
LGTM, great work

pkg/alertmanager/alertstore/store.go
pracucci: Good point. My point was to deprecate the old config as soon as we complete the rollout in prod (and we confirm the new config works fine). Unless we want to completely remove the `local` support, then I think the new config should support `local` too so that we can completely remove the old config in 2 releases.

jtlisi: 👍

pstibrany (approved):
Very nice, thanks!

PR #3889, opened 2021-02-27
Stream chunks from blocks-ingester to querier

### LINKED ISSUE #3831: Transfer encoded chunks from ingesters to querier in the blocks storage
The chunks storage was transferring encoded chunks from ingesters to querier at query time, while the blocks storage is transferring decoded ones. Having to decode chunks in the ingesters increases the memory utilisation at query time and increases the likelihood of getting ingesters OOMKilled because of an heavy query.

Historically, TSDB didn't allow to query for chunks, reason why we had to transfer decoded time-series from ingesters to querier, but now it's possible and we should it.

pstibrany:
**What this PR does**: This PR implements streaming of TSDB chunks from blocks-based ingester to querier. It reuses chunks-based querier, and just introduces new type of chunk: read-only wrapper around TSDB chunk.

~State of PR: manual testing shows that it works, unit tests not updated yet.~

PR is ready for review. Feature is currently protected by the flag (both CLI and runtime-config to avoid rollout of ingesters when playing with this), but will be enabled by default in the future.

Fixes #3831

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/chunk/encoding/factory.go
+++ b/pkg/chunk/encoding/factory.go
@@ -52,7 +52,7 @@ const (
 	Varbit
 	// Bigchunk encoding
 	Bigchunk
-	// Read-only wrapper around Prometheus XOR-encoded chunk.
+	// PrometheusXorChunk is a wrapper around Prometheus XOR-encoded chunk.
 	PrometheusXorChunk
 )
 
--- a/pkg/chunk/encoding/factory.go
+++ b/pkg/chunk/encoding/factory.go
@@ -52,6 +52,8 @@ const (
 	Varbit
 	// Bigchunk encoding
 	Bigchunk
+	// Read-only wrapper around Prometheus XOR-encoded chunk.
+	PrometheusXorChunk
 )
 
 type encoding struct {
@@ -76,6 +78,12 @@ var encodings = map[Encoding]encoding{
 		Name: "Bigchunk",
 		New: func() Chunk {
 			return newBigchunk()
+		},
+	},
+	PrometheusXorChunk: {
+		Name: "PrometheusXorChunk",
+		New: func() Chunk {
+			return newPrometheusXorChunk()
 		},
 	},
 }
--- a/pkg/chunk/encoding/prometheus_chunk.go
+++ b/pkg/chunk/encoding/prometheus_chunk.go
@@ -8,8 +8,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 )
 
-// Wrapper around Prometheus chunk. While it supports adding more samples, that is only implemented
-// to make tests work, and should not be used in production.
+// Wrapper around Prometheus chunk.
 type prometheusXorChunk struct {
 	chunk chunkenc.Chunk
 }
@@ -18,6 +17,9 @@ func newPrometheusXorChunk() *prometheusXorChunk {
 	return &prometheusXorChunk{}
 }
 
+// Add adds another sample to the chunk. While Add works, it is only implemented
+// to make tests work, and should not be used in production. In particular, it appends
+// all samples to single chunk, and uses new Appender for each Add.
 func (p *prometheusXorChunk) Add(m model.SamplePair) (Chunk, error) {
 	if p.chunk == nil {
 		p.chunk = chunkenc.NewXORChunk()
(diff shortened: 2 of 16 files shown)


pstibrany:
Benchmark comparing old (samples based) and new (chunk based) streaming:

```
name                      old time/op    new time/op    delta
Ingester_v2QueryStream-4    6.21ms ± 1%    1.13ms ± 2%  -81.88%  (p=0.000 n=10+10)

name                      old alloc/op   new alloc/op   delta
Ingester_v2QueryStream-4    3.55MB ± 0%    0.36MB ± 0%  -89.92%  (p=0.000 n=10+10)

name                      old allocs/op  new allocs/op  delta
Ingester_v2QueryStream-4     5.66k ± 0%     3.85k ± 0%  -31.88%  (p=0.000 n=10+10)
```

jtlisi:
Fixes #2485

pkg/ingester/ingester_v2.go
```diff
@@ -1190,14 +1242,101 @@ func (i *Ingester) v2QueryStream(req *client.QueryRequest, stream client.Ingeste
...
+		}
+
+		it := series.Iterator()
+		for it.Next() {
+			// Chunks are ordered by min time.
+			meta := it.At()
+
+			// It is not guaranteed that chunk returned by iterator is populated.
+			// For now just return error. We could also try to figure out how to read the chunk.
+			if meta.Chunk == nil {
```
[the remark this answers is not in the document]
pstibrany: Asked here: https://github.com/prometheus/alertmanager/issues/2504

pkg/chunk/encoding/factory.go
```diff
@@ -52,6 +52,8 @@ const (
 	Varbit
 	// Bigchunk encoding
 	Bigchunk
+	// PrometheusXorChunk is a wrapper around Prometheus XOR-encoded chunk.
+	PrometheusXorChunk
```
pracucci, line 56: Why not just calling it `XORChunk` (as named in Prometheus)?

pkg/chunk/encoding/prometheus_chunk.go
```diff
@@ -0,0 +1,174 @@
...
+
+import (
+	"io"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/tsdb/chunkenc"
+)
+
+// Wrapper around Prometheus chunk. While it supports adding more samples, that is only implemented
```
pracucci, line 11: > While it supports adding more samples, that is only implemented
> to make tests work, and should not be used in production.

I would move this comment to `Add()`.

pkg/chunk/encoding/prometheus_chunk.go
```diff
@@ -0,0 +1,174 @@
...
+func (p *prometheusXorChunk) Utilization() float64 {
+	// Used for reporting when chunk is used to store new data.
+	return 0
+}
+
+func (p *prometheusXorChunk) Slice(_, _ model.Time) Chunk {
+	return p
+}
+
+func (p *prometheusXorChunk) Rebound(from, to model.Time) (Chunk, error) {
```
pracucci, line 80: Given we don't need it, why not simplify it and return `errors.New("rebound is unsupported)`? Or is it required to get unit tests pass?

pkg/chunk/encoding/prometheus_chunk.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (p *prometheusChunkIterator) Batch(size int) Batch {
+	var batch Batch
+	j := 0
+	for j < size {
+		t, v := p.it.At()
+		batch.Timestamps[j] = t
+		batch.Values[j] = v
+		j++
+		if j < size && !p.it.Next() {
```
pracucci, line 155: Why do you also need to check `j < size` here?

pkg/ingester/ingester.go
```diff
@@ -79,8 +79,11 @@ type Config struct {
 	ActiveSeriesMetricsIdleTimeout  time.Duration `yaml:"active_series_metrics_idle_timeout"`
 
 	// Use blocks storage.
-	BlocksStorageEnabled bool                     `yaml:"-"`
-	BlocksStorageConfig  tsdb.BlocksStorageConfig `yaml:"-"`
+	BlocksStorageEnabled        bool                     `yaml:"-"`
+	BlocksStorageConfig         tsdb.BlocksStorageConfig `yaml:"-"`
+	StreamChunksWhenUsingBlocks bool                     `yaml:"-"`
```
pracucci, line 84: Why do we also need this if we have a runtime override? Can't we just used that?

pracucci (approved):
Wow, very good job! And good tests 👏  I left a couple of nits, but nothing particularly relevant.

pkg/chunk/encoding/factory.go
pstibrany: I wanted to make it clear that this is using XorChunk from Prometheus, hence the name.

pkg/chunk/encoding/prometheus_chunk.go
pstibrany: I added it to make tests pass. I can also make it return error instead, and just skip Rebounds test for this chunk type. I think that's fair.

pkg/chunk/encoding/prometheus_chunk.go
pstibrany: To avoid calling `p.it.Next()` if we cannot put more samples to the batch. (Note that `j` is now increased since last check). Honestly, I didn't think too much about it and just copied implementation from other chunks. Fortunately removing this check makes tests fail.

pkg/ingester/ingester.go
pstibrany: I used CLI flag to make integration test simpler to write. We could move this to change default in runtime config, but it seemed easier to have it here.

pracucci:
Thanks @pstibrany for addressing my feedback! LGTM. I don't feel strong about the double CLI flags considering they're temporarily I'm fine with that.

pstibrany:
Using this feature makes average time for rule group evaluation to go up:
`avg (cortex_prometheus_rule_group_last_duration_seconds)`

PR #3891, opened 2021-03-01
blocksconvert: optionally reduce timestamp jitter

bboreham:
Run with `-builder.timestamp-tolerance=5` to adjust sample timestamps by up to 5ms to align them to an exact number of seconds apart, thereby reducing the size of blocks in memory or on disk.

Prometheus started doing this in version 2.22, so this option is most useful for data gathered from older versions.

See https://github.com/prometheus/prometheus/issues/7846 for background info.

I have not run before/after tests with this code; I didn't know how to re-run over the same data and keep both versions for comparison.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
 * [BUGFIX] Ruler Storage: ignore objects with empty namespace or group in the name. #3999
 * [BUGFIX] Distributor: fix issue causing distributors to not extend the replication set because of failing instances when zone-aware replication is enabled. #3977
 
+## Blocksconvert
+
+* [ENHANCEMENT] Builder: add `-builder.timestamp-tolerance` option which may reduce block size by rounding timestamps to make difference whole seconds. #3891
+
 ## 1.8.0 in progress
 
 * [CHANGE] Alertmanager: Don't expose cluster information to tenants via the `/alertmanager/api/v1/status` API endpoint when operating with clustering enabled.
--- a/tools/blocksconvert/builder/builder.go
+++ b/tools/blocksconvert/builder/builder.go
@@ -38,10 +38,11 @@ type Config struct {
 	OutputDirectory string
 	Concurrency     int
 
-	ChunkCacheConfig cache.Config
-	UploadBlock      bool
-	DeleteLocalBlock bool
-	SeriesBatchSize  int
+	ChunkCacheConfig   cache.Config
+	UploadBlock        bool
+	DeleteLocalBlock   bool
+	SeriesBatchSize    int
+	TimestampTolerance int
 
 	PlanProcessorConfig planprocessor.Config
 }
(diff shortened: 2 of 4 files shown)


roidelapluie:
Note: Prometheus does 2ms, if the scrape interval is higher than 200ms.

tools/blocksconvert/builder/builder.go
```diff
@@ -38,10 +38,11 @@ type Config struct {
...
 
-	ChunkCacheConfig cache.Config
-	UploadBlock      bool
-	DeleteLocalBlock bool
-	SeriesBatchSize  int
+	ChunkCacheConfig   cache.Config
+	UploadBlock        bool
+	DeleteLocalBlock   bool
+	SeriesBatchSize    int
+	TimestampTolerance int
```
pracucci, line 45: I would make it a `time.Duration`. Would be easier to understand how to use it.

roidelapluie: note that the `prometheus/tsdb` package itself is time-agnostic.

pstibrany:
> I have not run before/after tests with this code; I didn't know how to re-run over the same data and keep both versions for comparison.

You can delete "<dayindex>.finished.<blockID>" file from the storage with the plans, and then the plan will be rebuilt. Note that you should probably remove previous block, or compactor will merge the two blocks together -- but they will have different timestamps.

bboreham:
> compactor will merge the two blocks together

Yeah this is very scary.
If I could run the builder such that it just reported the size and didn't upload, I might do that, for the comparison.

pstibrany:
> > compactor will merge the two blocks together
> 
> Yeah this is very scary.
> If I could run the builder such that it just reported the size and didn't upload, I might do that, for the comparison.

`-builder.upload=false` flag should do the trick.

tools/blocksconvert/builder/builder.go
pracucci: Using `time.Duration` instead of `int` doesn't make it time specific. Maybe I'm not getting the feedback 🤔

pstibrany: I think Julien's point is that TSDB is doesn't specify any time units. But Cortex (and Prometheus) only work with milliseconds  timestamps anyway. Here we're introducing this "knowledge" of timestamp units into blocks-builder.

pracucci: Sure, but for the Cortex final user I think it's easier to reason about time units than an `int`. You can convert `time.Duration` to milliseconds and keep the logic time-agnostic.

pstibrany: > You can convert time.Duration to milliseconds and keep the logic time-agnostic.

It's not unit-agnostic if you convert it to milliseconds :) (Btw, I think `time.Duration` makes sense here)

bboreham:
OK, now I have run a comparison; this was data sent last July from Prometheus 2.19.2:

`-builder.timestamp-tolerance=0`: `plan=migration/910/18354.plan.snappy msg="successfully built block for a plan" ulid=01EZQ8SANYAZWBCVCC4RP6ZC0M size=1553414435`

`-builder.timestamp-tolerance=2`: `plan=migration/910/18354.plan.snappy msg="successfully built block for a plan" ulid=01EZQB04SENZTTH95S0ZFWN2Z5 size=1314057565`

`-builder.timestamp-tolerance=5`: `plan=migration/910/18354.plan.snappy msg="successfully built block for a plan" ulid=01EZNYPJAEA0AA1TVQ5F548ENB size=1218364506`

So 2ms tolerance gives about a 15% reduction in block size for this data; 5ms gives 21%.

CHANGELOG.md
```diff
@@ -91,6 +91,10 @@
 * [BUGFIX] Prevent panic at start if the http_prefix setting doesn't have a valid value. #3796
 * [BUGFIX] Memberlist: fixed panic caused by race condition in `armon/go-metrics` used by memberlist client. #3724
 
+## Blocksconvert
+
+* [ENHANCEMENT] Builder: add `-builder.timestamp-tolerance` option which may reduce block size. #3891
```
pstibrany, line 96:
```
* [ENHANCEMENT] Builder: add `-builder.timestamp-tolerance` option which may reduce block size by adjusting timestamps to make timestamp diffeference rounded to seconds. #3891
```

pstibrany (approved):
LGTM, thank you.

CHANGELOG.md
pstibrany, line 96: Cortex release 1.8.0 is now in progress. Could you please rebase master and move the CHANGELOG entry under the master / unreleased section?

bboreham:
Rebased and changed the param to a Duration.

pracucci (approved):

PR #3892, opened 2021-03-01
Reuse FetcherMetrics from Thanos

pracucci:
**What this PR does**:
In the PR https://github.com/cortexproject/cortex/pull/3625 I had to duplicate some internal Thanos code to track metadata fetcher metrics. I then submit a couple of Thanos PRs (https://github.com/thanos-io/thanos/pull/3660 and https://github.com/thanos-io/thanos/pull/3842) to expose it and in this PR I'm upgrading Thanos to reuse it.

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/storegateway/bucket_index_metadata_fetcher.go
+++ b/pkg/storegateway/bucket_index_metadata_fetcher.go
@@ -9,14 +9,17 @@ import (
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/thanos-io/thanos/pkg/block"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
-	"github.com/thanos-io/thanos/pkg/extprom"
 	"github.com/thanos-io/thanos/pkg/objstore"
 
 	"github.com/cortexproject/cortex/pkg/storage/bucket"
 	"github.com/cortexproject/cortex/pkg/storage/tsdb/bucketindex"
+)
+
+const (
+	corruptedBucketIndex = "corrupted-bucket-index"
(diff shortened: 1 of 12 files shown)


pkg/storegateway/bucket_index_metadata_fetcher.go
```diff
@@ -49,37 +52,37 @@ func NewBucketIndexMetadataFetcher(
...
 		modifiers:   modifiers,
-		metrics:     newFetcherMetrics(reg),
+		metrics:     block.NewFetcherMetrics(reg, [][]string{{corruptedBucketIndex}, {noBucketIndex}}, nil),
 	}
 }
 
 // Fetch implements metadata.MetadataFetcher.
 func (f *BucketIndexMetadataFetcher) Fetch(ctx context.Context) (metas map[ulid.ULID]*metadata.Meta, partial map[ulid.ULID]error, err error) {
-	f.metrics.resetTx()
+	f.metrics.ResetTx()
```
jtlisi, line 61: **thought(non-blocking):** The [upstream ResetTx function](https://github.com/thanos-io/thanos/blob/8cf129b1451d9cad160dd440bd8064a89b49d96d/pkg/extprom/tx_gauge.go#L47) is not goroutine safe. Would it make sense to propagate that warning here with a comment indicating the same?

jtlisi (approved):
LGTM

pkg/storegateway/bucket_index_metadata_fetcher.go
pracucci: The whole `MetadataFetcher.Fetch()` is not required to be goroutine-safe (in Thanos it's not implemented in a goroutine-safe way too). I've updated the function comment.

gouthamve (approved):

PR #3894, opened 2021-03-02
feat: instrument the query layer to track rate-limited queries

jtlisi:
**What this PR does**:

Instrument the query-frontend and query-scheduler to track the queries that were discarded because of the configured outstanding queries rate limiter. I pre-emptively added a reason label to allow for flexibility in adding other reasons for why a query may be discarded.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,8 +29,8 @@
     * `-alertmanager.alertmanager-client.tls-insecure-skip-verify`
 * [FEATURE] Compactor: added blocks storage per-tenant retention support. This is configured via `-compactor.retention-period`, and can be overridden on a per-tenant basis. #3879
 * [ENHANCEMENT] Queries: Instrument queries that were discarded due to the configured `max_outstanding_requests_per_tenant`. #3894
-  * `cortex_query_frontend_discarded_queries_total`
-  * `cortex_query_scheduler_discarded_queries_total`
+  * `cortex_query_frontend_discarded_requests_total`
+  * `cortex_query_scheduler_discarded_requests_total`
 * [ENHANCEMENT] Ruler: Add TLS and explicit basis authentication configuration options for the HTTP client the ruler uses to communicate with the alertmanager. #3752
   * `-ruler.alertmanager-client.basic-auth-username`: Configure the basic authentication username used by the client. Takes precedent over a URL configured username.
   * `-ruler.alertmanager-client.basic-auth-password`: Configure the basic authentication password used by the client. Takes precedent over a URL configured password.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@
     * `-alertmanager.alertmanager-client.tls-server-name`
     * `-alertmanager.alertmanager-client.tls-insecure-skip-verify`
 * [FEATURE] Compactor: added blocks storage per-tenant retention support. This is configured via `-compactor.retention-period`, and can be overridden on a per-tenant basis. #3879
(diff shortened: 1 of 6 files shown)


pkg/scheduler/queue/queue.go
```diff
@@ -91,6 +95,7 @@ func (q *RequestQueue) EnqueueRequest(userID string, req Request, maxQueriers in
 		}
 		return nil
 	default:
+		q.discardedQueries.WithLabelValues(userID, validation.RateLimited).Inc()
```
pstibrany, line 98: I think "rate limited" is misleading here. We only drop the request if queue is full, which can be for various reasons (slow queries, queriers crashed, no querier connected or all queriers are busy). I'd suggest to start without "reason" label.

Also I don't think we should use "validation" reasons (used for validating incoming samples) here.

pkg/frontend/v1/frontend.go
```diff
@@ -83,14 +84,18 @@ func New(cfg Config, limits Limits, log log.Logger, registerer prometheus.Regist
 			Name: "cortex_query_frontend_queue_length",
 			Help: "Number of queries in the queue.",
 		}, []string{"user"}),
+		discardedQueries: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
```
pstibrany, line 87: Please cleanup this metric in `cleanupInactiveUserMetrics` method.

pkg/scheduler/scheduler.go
```diff
@@ -100,7 +101,12 @@ func NewScheduler(cfg Config, limits Limits, log log.Logger, registerer promethe
 		Name: "cortex_query_scheduler_queue_length",
 		Help: "Number of queries in the queue.",
 	}, []string{"user"})
-	s.requestQueue = queue.NewRequestQueue(cfg.MaxOutstandingPerTenant, s.queueLength)
+
+	s.discardedQueries = promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
```
pstibrany, line 105: Please cleanup this metric in `cleanupMetricsForInactiveUser` and add it to tests to verify that.

pstibrany (commented):
We already have metric to monitor queue length per user, but that doesn't tell us max queue length, so I can see how this new metric could be useful.

I'd suggest to start without "reason" label – we can always add it later, when needed.

pkg/scheduler/queue/queue.go
jtlisi: Makes sense.

pkg/frontend/v1/frontend.go
jtlisi: 👍

pkg/scheduler/scheduler.go
jtlisi: 👍

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@
     * `-alertmanager.alertmanager-client.tls-server-name`
     * `-alertmanager.alertmanager-client.tls-insecure-skip-verify`
 * [FEATURE] Compactor: added blocks storage per-tenant retention support. This is configured via `-compactor.retention-period`, and can be overridden on a per-tenant basis. #3879
+* [ENHANCEMENT] Queries: Instrument queries that were discarded due to the configured `max_outstanding_requests_per_tenant`. #3894
```
pstibrany, line 31: Nit: since limit has "requests" in the name, should metrics use "requests" instead of "queries" too?

pstibrany (approved):
LGTM, thank you.

CHANGELOG.md
jtlisi: 👍

